Issue #8, opened 2017-11-07
New submission script for github


rucio:
We need a new submission script for pull-requests on github.

PR #14, opened 2017-11-07
Testing #8: New submission procedure for github

bari12:


tools/submit-pull-request
```diff
@@ -0,0 +1,86 @@
...
+# Authors:
+# - Martin Barisits, <martin.barisits@cern.ch>, 2017
+
+
+import commands
+import sys
+
+
+def submit_pull_request(branch, message, label):
+    op = commands.getstatusoutput('hub pull-request -m "%s" -b %s -l %s' % (message, branch, label))
```
vingar: label not supported with the default version (not pre release)
I think we can drop it ? 
in the future the link to the issue can be also added there: at the endhttps://github.com/rucio/rucio/issues/<number>

if the association pull request <-> issue is missing

bari12: -label is supported in hub since 2016, I think you are just using a very old version.
The linking of pull-requests to issues is a strongly discouraged behaviour, even so much that it is deprecated in the API, that's why I took it out.

vingar: well it's the default and latest version in homebrew. the pre releases are maybe not considered final by them there. Why are the possible labels here ?

bari12: I would like to have it as it is immediately clear if it is a patch/feature/hotfix. But I see that it is a bit difficult to require the binary. For CC7 you need the binary anyway, as the default version does not work. Alternatively I can try to make the script just directly with the REST commands, but that will require some more work.

mlassnig:
Strongly in favour of using git only. No external command should be necessary to submit a request.

bari12:
If everyone prefers it I can try to remove hub completely from the submission script and do everything via their REST API, like in the good old gitlab days 😄

bari12:
Ok, I had a bit of a look on the API, it is not difficult at all, so I will replicate the behaviour with the API. Stay tuned 😃

tools/submit-pull-request
vingar: well to do pull request on github, you need need hub installed. Then to have with git I just have the alias:
alias git='hub'
then
git pull-request -m 'readme' ...

vingar:
well to do pull request on github, you need hub installed. Then to have with git I just have the alias:
alias git='hub'
then
git pull-request -m 'readme' ...

bari12:
No it also works completely via curl calls, no hub required. (But whoever wants to use hub, it's also fine)

vingar:
yes, hub just goes against the same rest calls I guess

vingar (approved):

PR #18, opened 2017-11-07
documentation: [RUCIO-2860] Added requirements.txt

vingar:


bari12:
Just a side-comment. I think we should add it to the build procedure that this file (based on the pip-requires files) is being updated. So it is always up to date.

vingar:
yes, each time a dep. is touched the requirements.txt should be regenerated (cat ...) as readthedocs is defined against the master

bari12 (approved):

Issue #23, opened 2017-11-08
File download from WebUI


bari12:
Lots of sites already support WebDAV downloads, so it would be a nice feature if we would have download directly from the WebUI. So when users list the files in a dataset their can just click a button and the files / one file would be directly downloaded to their machine.

_Migrated from JIRA [RUCIO-1630](https://its.cern.ch/jira/browse/RUCIO-1630)_

PR #89, opened 2017-11-09
WebUI: file download via webui #23

sprenner:


tools/bootstrap_tests.py
```diff
@@ -50,6 +50,7 @@
                                 ('/DC=ch/DC=cern/OU=Organic Units/OU=Users/CN=barisits/CN=692443/CN=Martin Barisits', 'x509', 'martin.barisits@cern.ch'),
                                 ('/DC=ch/DC=cern/OU=Organic Units/OU=Users/CN=tbeerman/CN=722011/CN=Thomas Beermann', 'x509', 'thomas.beermann@cern.ch'),
                                 ('/DC=ch/DC=cern/OU=Organic Units/OU=Users/CN=ruciobuildbot/CN=692443/CN=Robot: Rucio build bot', 'x509', 'rucio.build.bot@cern.ch'),
+                                ('CN=Stefan Prenner,CN=822876,CN=sprenner,OU=Users,OU=Organic Units,DC=cern,DC=ch', 'x509', 'stefan.prenner@cern.ch'),
```
mlassnig, line 53: this doesn't look right?

tbeerman: It works, at least on his dev machine. If ordered in the 'right' way the server does not accept the certificate.

lib/rucio/web/ui/static/did.js
```diff
@@ -99,6 +100,7 @@ load_replicas = function(scope, name) {
     r.list_replicas({
         'scope': scope,
         'name': name,
+	'schemes': ['davs'],
```
tbeerman: I'm afraid you will have to do this call twice. If you do it like this only replicas which are on an RSE that supports davs are listed. But in the table it should always show all available replicas and then mark those that are on an RSE that is browser enabled.

bari12:
Can you rebase this to a current master. Github shows that this is pulling in 250+ commits and 150k lines of code, so this seems a bit fishy 😄 It also shows conflicts. For the PR against next it seems to be correct though.
```
git fetch --all
git checkout patch-1630-file_download_webui
git rebase upstream/master
git push origin patch-1630-file_download_webui --force
```

bari12:
Forget my comments, now with the repaired next it is fine. 😄

tbeerman (approved):

PR #97, opened 2017-11-09
Core & Internals: Add files to support T3C Fix #79

### LINKED ISSUE #79: Transfer Time To Complete Service
T³C service for transfer time to complete prediction.
ToDo: Rucio DB access to retrieve the submitted_at for all the active transfers for a (src,dst,act) link.

_Migrated from JIRA [RUCIO-2843](https://its.cern.ch/jira/browse/RUCIO-2843)_

jwackito:
Core & Internals: Add files to support T3C Fix #79


bari12 (changes requested):
```
ERROR: Failure: ImportError (No module named extensions.forecast)
  File "/usr/lib/python2.7/site-packages/nose/loader.py", line 418, in loadTestsFromName
    addr.filename, addr.module)
  File "/usr/lib/python2.7/site-packages/nose/importer.py", line 47, in importFromPath
    return self.importFromDir(dir_path, fqname)
  File "/usr/lib/python2.7/site-packages/nose/importer.py", line 94, in importFromDir
    mod = load_module(part_fqname, fh, filename, desc)
  File "/opt/rucio/lib/rucio/tests/test_forecast.py", line 11, in <module>
    from rucio.extensions.forecast import T3CModel
ImportError: No module named extensions.forecast
Ran 378 tests in 440.114s
FAILED (errors=1)
```
Please fix :)

lib/rucio/extensions/forecast.py
```diff
@@ -0,0 +1,113 @@
...
+request/site/query/list/ddmendpoints?json')
+        j = r.json()
+        self._rse2site = {}
+        self._site2rses = {}
+        for i in j:
+            self._site2rses[i['name']] = i['ddmendpoints'].keys()
+            for ep in i['ddmendpoints']:
+                self._rse2site[ep] = i['name']
+
+    def ewma(self, x, span):
```
bari12, line 44: All the method docstrings are missing.

lib/rucio/extensions/forecast.py
```diff
@@ -0,0 +1,113 @@
...
+        except IOError:
+            self.model = {"STANDARD_LINK": {"r2": 0.0,
+                                            "rmse": 0.0,
+                                            "rate": 23399639.38262837,
+                                            "datalen": 0.0,
+                                            "overhead": 12.025538382153206,
+                                            "diskrw": 12046990.897099394}}
+
+        # Create RSE to site name dictionary
+        r = requests.get('http://atlas-agis-api.cern.ch/\
```
bari12, line 34: Few thing about this:
- The url should probably be put into the config somewhere.
- We should also think about the generic case for this, which should not be much work to implement. What I am saying is, the t3c will be interesting for non-ATLAS Rucio users too, but right now your setup basically requires AGIS.
- I think we should cache this information somewhere too. Depending on where/how this object is constructed, we could be hammering AGIS....

bari12 (changes requested):

lib/rucio/extensions/forecast.py
mlassnig: @jwackito something to keep in mind: right now for ATLAS we do link calculations site-by-site. to have this completely standalone, we need this functionality RSE-to-RSE, or otherwise we are always dependent on an informaiton catalogue. you should keep this in mind when scaling this out for one of the next iterations. right now we focus on getting the ATLAS case to work, but keep it in mind.

jwackito: @mlassnig, @bari12: Actualy if I do this on RSE basis, I will not need AGIS at all. It's clear to me why we need to do this on site level for ATLAS, but I do not know if there is any real difference doing it one way or the other.
In order to calculate the model for the network time, I think I can even do it on the fly from Rucio DB instead of calculate it from elasticsearch. But to do it at site level need some mapping from AGIS. I will check other daemons to see where they get AGIS data from...

vingar: site names are in rucio, e.g. rse_attributes. no need to have this url dependency here.

jwackito: @vingar: Then I can calculate everything from Rucio DB. Thanks!

bari12:
Hi Joaquin, due to a problem on the next branch, which we had to manually fix your PR is not out of date. Please re-create it with a new branch. Sorry for the inconvenience.

Issue #98, opened 2017-11-10
Minor fixes in the readme files


bari12:
Some minor fixes and links are not written consistently (sometimes URL, sometimes names)

PR #100, opened 2017-11-10
Documentation #98: Minor fixes in readme files

bari12:
Documentation #98: Minor fixes in readme files

### DIFF
--- a/README.client.rst
+++ b/README.client.rst
@@ -1,12 +1,12 @@
 Rucio Clients
 ==============
 
- Data Management for science in the Big Data era.
+Data Management for science in the Big Data era.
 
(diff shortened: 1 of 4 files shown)


bari12:
⚠️  So this is now one of the conflicts I was talking about. In that case never click the resolve conflict button.

vingar:
how was it done  ?  I'll suggest to start a contributor guide as it sounds scary..

bari12:
It's the same as on gitlab, I have to do it on command line. Sadly the automatic conflict resolver is not all too good and it finds (in my opinion) conflicts where there are none (As the file is the same on master/next, so why does it fail on next) - But this is a git issue, it was already the same on gitlab, just that gitlab did not offer the "web resolve conflicts"

PR #103, opened 2017-11-10
Release management #0: Preparation of the version file for 1.13.3 release

bari12:
Release management #0: Preparation of the version file for 1.12.4 release


bari12:
Just FYI, I didn't make a ticket number for this as it is just preparing the version file. Otherwise we would have an issue in every release note "Preparing the version file" which is a bit silly I think.

vingar:
well the issue can go to a permanent release management milestone (never closed). but more importantly, release notes in the repo can be something to consider (not only in the github release part)

bari12:
Yes, it can also be an issue which is not assigned to any milestone. IMHO this is also a good solution. Adding it to the (release) milestone just looks a bit silly, like artificially enlarging the release notes.

vingar (approved):

PR #115, opened 2017-11-13
Release: Change helper scripts to fully reset against upstream. Closes #113

### LINKED ISSUE #113: Create-patch/next-branch script should make full reset of origin against next
Sometimes it happens that the origin branches are a bit out of sync with the upstream. In these cases just rebasing upstream (what the script does now) will not make the situation better, instead it should make a full reset against upstream, those making the origin/upstream the same.

bari12:
Release: Change helper scripts to fully reset against upstream. Closes #113

### DIFF
--- a/tools/create-feature-branch
+++ b/tools/create-feature-branch
@@ -15,7 +15,7 @@ echo "Updating next"
 git pull --all --prune --progress
 
 echo "Rebasing next"
-git rebase upstream/next next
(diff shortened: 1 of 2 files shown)


vingar:
file changed 1,041...

bari12:
Yep, this is the case with all pull-requests against next at the moment. It seems like something is wrong with next. We are still investigating.

PR #118, opened 2017-11-14
documentation: More work on documentation Fix #86

### LINKED ISSUE #86: More work on documentation (reset)
_Migrated from JIRA [RUCIO-2861](https://its.cern.ch/jira/browse/RUCIO-2861)_

vingar:


ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,8 @@
+Motivation
```
bari12, line 1: It's not clear to me why these files are shown here in the pull-request again, they are already part of next. They are not shown in the one against master. I wonder if this is an indication that there is a problem and we are diverging the two branches again or if this is just how it is.

vingar: Pull requests done from a clean repo after a fork. 188 files modified looks normal as I reset the doc directory, before it was more something like 1500 files. After I don't know, we should just try ;)

bari12: It might very well be completely fine, it's just weird as the other PR shows 182 files, which is correct, as the 6 additional files here are already on next, so they should not be displayed. But it might be that github just displays this weirdly and from a git point of view it is fine.

bari12:
As discussed today, I would propose we merge this and then work on the documentation iteratively. It's very hard to review like this anyway and for now we keep the old doc on readthedocs.

bari12 (approved):

PR #130, opened 2017-11-20
WebUI: Fixes bug which caused wrong download links with datasets. #23

sprenner:
[RUCIO-1630] Fixes bug which caused wrong download links with datasets.


vingar:
please squash commits

tbeerman (approved):

PR #143, opened 2017-11-22
Rules: Command to move a replication rule; Closes #133

### LINKED ISSUE #133: Option to "move" a rule in CLI
Motivation
----------
The idea is basically the same principle as with BB8. A specific rule should be moved to another RSE, including all it's properties.

_Migrated from [JIRA](https://its.cern.ch/jira/browse/RUCIO-2721)_

bari12:
Rules: Command to move a replication rule; Closes #133

### DIFF
--- a/bin/rucio
+++ b/bin/rucio
@@ -2092,6 +2092,18 @@ def update_rule(args):
 
 
 @exception_handler
+def move_rule(args):
+    """
+    %(prog)s move-rule [options] <ruleid> <rse_expression>
+
+    Update a rule.
+    """
+    client = get_client(args)
+    print client.move_replication_rule(rule_id=args.rule_id, rse_expression=args.rse_expression)
+    return SUCCESS
+
+
+@exception_handler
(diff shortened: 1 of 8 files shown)


bin/rucio
```diff
@@ -2139,6 +2151,7 @@ def info_rule(args):
         print "Purge replicas:             %s" % rule['purge_replicas']
         print "Notification:               %s" % rule['notification']
         print "End of life:                %s" % rule['eol_at']
+        print "Child rule Id:              %s" % rule['child_rule_id']
```
mlassnig: capital R in rule or lowercase I in id just to be consistent ;-)

mlassnig:
look straightforward to me, any implications?

bin/rucio
bari12: 🙄  will fix it 😄

bari12:
No it should be fine, I did not test it massively, but will ask site admins to test it first a bit before large scale use.

PR #162, opened 2017-11-23
Deletion: Reaper check for suspended/waiting for approval rules when it deletes datasets Fix #142

### LINKED ISSUE #142: Deletion: Reaper check for suspended/waiting for approval rules when it deletes dataset
Motivation
----------
When the last file in a dataset is removed by the reaper, the reaper also tries to delete the dataset. However, in certain cases it can happen that there is a SUSPENDED/Waitint for approval rule with 0 locks on the dataset, thus the deletion always fails with a child_constraint error from the database.

Modification
------------
The reaper should check for rules when it is deleting the dataset and remove them in case there are any.

vingar:


mlassnig (approved):

vingar:
@bari12 the conflict is expected as the patch goes on master and next. Would you fix it ?

bari12:
Yes, this I will do manually.

Issue #158, opened 2017-11-22
Fix flake8 command used in travis builds


tbeerman:
Motivation
----------

The flake8 command used in travis is currently wrong and excludes everything.

Currently:
`flake8 --ignore=E501 --exclude="*.cfg bin/* lib/ tools/*.py tools/probes/common/*"`

But it should be:
`flake8 --ignore=E501 --exclude="*.cfg" bin/* lib/ tools/*.py tools/probes/common/*`

PR #172, opened 2017-11-23
Testing: fix flake8 exclude list for travis builds #158

tbeerman:
Testing: fix flake8 exclude list for travis builds #158

### DIFF
--- a/tools/travis_scripts.sh
+++ b/tools/travis_scripts.sh
@@ -31,7 +31,7 @@ fi
 echo '==============================='
 echo 'Running flake8                 '
 echo '==============================='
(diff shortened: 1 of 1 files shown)


bari12:
I don't get it.
`flake8 --ignore=E501 --exclude="*.cfg" bin/* lib/ tools/*.py tools/probes/common/*`
Gives no errors for me on the current master...

mlassnig:
different flake8 version? i've seen newer flake8 be much more conservative

bari12:
No, it's way more complicated then this. 😄 
In contrast to Gitlab, where only the branch was tested, travis actually merges the branch into master/next and tests that. Because there are still 2 errors which are not fixed on next (See my open PR), this fails. Very interesting and took some time to figure out.

PR #200, opened 2017-11-27
probes: check_quota probe does not create quota for new SCRATCHDIIKs for old accounts; Fix #199

### LINKED ISSUE #199: check_quota probe does not create quota for new SCRATCHDISKS for old accounts
Motivation
----------
The probe only checks if an account has quota on some SCRATCHDISK, thus for old accounts, they will never get quota on new scratchdisks.

Modification
------------
This has to be changed to actually check each quota specifically.

cserf:
probes: check_quota probe does not create quota for new SCRATCHDIIKs for old accounts; Fix #199


bari12 (changes requested):
tools/probes/common/check_quotas:174:5: E303 too many blank lines (2)

bari12 (approved):

Issue #211, opened 2017-11-28
Add eventType and usrDN parameters to CLI up/download


tbeerman:
Motivation
----------

When the pilot uses the rucio mover the traces are sent out as type 'download' and with the account 'pilot'. This messes up the monitoring in the as they will show up as 'CLI download' in the Dashboard and we are also losing the user information.

Modification
------------

Add two new paramters the rucio download cli command: `--usrdn` and `--eventType` which will then be propagated to the traces.

Expected behavior
-----------------

mlassnig:
Proposal for consistency:

--trace-user
--trace-event-type

PR #229, opened 2017-11-29
Clients: add environment variables for rucio mover to set trace parameters #211

tbeerman:
Clients: add environment variables for rucio mover to set trace parameters #211

### DIFF
--- a/bin/rucio
+++ b/bin/rucio
@@ -1672,6 +1672,21 @@ def download(args):
                      'eventType': 'download',
                      'eventVersion': version.RUCIO_VERSION[0]}
 
+    if args.trace_appid:
+        trace_pattern['appid'] = args.trace_appid
+    if args.trace_dataset:
+        trace_pattern['dataset'] = args.trace_dataset
+    if args.trace_datasetscope:
+        trace_pattern['datasetScope'] = args.trace_datasetscope
+    if args.trace_eventtype:
+        trace_pattern['eventType'] = args.trace_eventtype
+    if args.trace_pq:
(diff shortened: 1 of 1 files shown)


bin/rucio
```diff
@@ -1672,6 +1672,29 @@ def download(args):
...
                      'eventVersion': version.RUCIO_VERSION[0]}
 
+    t_appid = os.environ.get('RUCIO_TRACE_APPID')
+    t_dataset = os.environ.get('RUCIO_TRACE_DATASET')
+    t_datasetscope = os.environ.get('RUCIO_TRACE_DATASETSCOPE')
+    t_eventtype = os.environ.get('RUCIO_TRACE_EVENTTYPE')
+    t_pq = os.environ.get('RUCIO_TRACE_PQ')
+    t_taskid = os.environ.get('RUCIO_TRACE_TASKID')
+    t_usrdn = os.environ.get('RUCIO_TRACE_USRDN')
+
```
vingar: Getting the env. variable in argparse and propagating them as argument/dictionary to the download function will be cleaner, e.g.:
`parser.add_argument('--appid', default=os.environ.get('RUCIO_TRACE_APPID', None))`

vingar (changes requested):

bin/rucio
mlassnig: if we do it that way, we must hide the options from the user.

..., help=argparse.SUPPRESS, ...

vingar: that's a really good point. Moving toward a download api that can be called in a multi-threads env. was one of the motivation for the requested change.

PR #247, opened 2017-12-01
remove hardcoded support references; Fix #243

### LINKED ISSUE #243: Removing reference to  https://its.cern.ch/jira/browse/RUCIO in rucio clis
Motivation
----------
We should replace https://its.cern.ch/jira/browse/RUCIO by the github issue link in the client code

2017-11-30 16:01:50,357 ERROR [Rucio exited with an unexpected/unknown error. Please rerun the last command with the '-v' option and submit a ticket with all the necessary information at https://its.cern.ch/jira/browse/RUCIO]

mlassnig:
remove hardcoded support references; Fix #243

### DIFF
--- a/etc/rucio.cfg.atlas.client.template
+++ b/etc/rucio.cfg.atlas.client.template
@@ -6,6 +6,7 @@
 #
 # Authors:
 # - Vincent Garonne, <vincent.garonne@cern.ch>, 2013
+# - Mario Lassnig, <mario.lassnig@cern.ch>, 2017
 
 [common]
 
@@ -15,3 +16,5 @@ auth_host = https://voatlasrucio-auth-prod.cern.ch:443
 ca_cert = $RUCIO_HOME/etc/ca.crt
 client_x509_proxy = $X509_USER_PROXY
 request_retries = 3
+support_experiment = hn-atlas-dist-analysis-help@cern.ch
+support_rucio = https://github.com/rucio/rucio/issues
--- a/bin/rucio
+++ b/bin/rucio
@@ -226,11 +226,10 @@ def exception_handler(function):
                     pass
                 logger.error('Specified account %s does not have an associated identity.' % used_account)
             else:
-                logger.error(error)
                 logger.debug(traceback.format_exc())
-                logger.error("This means that the object doesn't have the property " + str(error) +
-                             ". This should never happen. Please rerun the last command with the '-v'" +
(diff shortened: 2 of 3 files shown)


etc/rucio.cfg.atlas.client.template
```diff
@@ -14,4 +15,6 @@ rucio_host = https://voatlasrucio-server-prod.cern.ch:443
 auth_host = https://voatlasrucio-auth-prod.cern.ch:443
 ca_cert = $RUCIO_HOME/etc/ca.crt
 client_x509_proxy = $X509_USER_PROXY
-request_retries = 3
\ No newline at end of file
+request_retries = 3
+support_experiment = hn-atlas-dist-analysis-help@cern.ch
```
bari12, line 19: I think `support_vo` would be a better name and more in line what we have now with the policy/permission as well.

vingar, line 19: An random user can change/edit this directly. Maybe in some policy files (like in permission ?) looks better

bari12: They can edit the python files as well 😄  Imho the rucio.cfg is the right place for it, as this will be provided by the VO anyway.

mlassnig: is that really realistic? ;-)

vingar: Few things like permission, webui pages, user support are related to a vo/policy. Maybe having all this information handled in the same way is beneficial.

mlassnig: That argument makes more sense.

However, we still need the vo-distinction in the rucio.cfg, which we don't have right now (only for permissions).

vingar: yes good point. Now it's :
[permission]
policy=atlas

can we make it more generic ? like 

[policy]
organyzation=atlas

and have some policies files

vingar: organization_support ?

mlassnig: yeah, i was thinking something like that. also sounds more generic and reasonable to me.

bari12: organization_support sounds good to me. But we should change this consistently throughout all files, so VO and organization is not mixed.

Issue #246, opened 2017-12-01
Restructure Dockerfile


cserf:
Motivation
----------
The DB part is used in server, daemons and UI image. Can be factorized by moving it to CC7 image

Modification
------------
Move the DB part to CC7 image

PR #256, opened 2017-12-01
Core & Internals: get_submitted_at_rucio Fix #232

### LINKED ISSUE #232: get_submitted_at_rucio should return an array of submitted_at time stamps
Motivation
----------
get_submitted_at_rucio function in forecast.py extension is buggy. 

Modification
------------
The function have access to rucio database, but the query is tricky and need testing on production to see if the time to retrieve the submitted_at time stamps for a link in not an overkill.

Expected behavior
-----------------
Should return an array on time stamps ordered descendent (the oldest one at the end).

jwackito:
Core & Internals: get_submitted_at_rucio Fix #232

### DIFF
--- a/lib/rucio/extensions/forecast.py
+++ b/lib/rucio/extensions/forecast.py
@@ -12,6 +12,7 @@
 Forecast library to predict the Transfers Time To Complete (T3C)
 """
 
+import datetime as dt
 import json
 import logging
 import time
@@ -20,8 +21,6 @@ import numpy as np
 from rucio.core.rse import list_rses, list_rse_attributes
 from rucio.db.sqla.session import read_session
 from rucio.db.sqla import models
-
-from sqlalchemy.sql.expression import and_
 
 
 def ewma(x_values, span):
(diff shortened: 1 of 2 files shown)


bari12:
Hmm travis fails, is this related to #246? @cserf @tbeerman

tbeerman:
Yes, @cserf changed the base docker image in #246 to already install the Oracle instantclient. In the docker build of travis it then also tries to install the client but it is already there and yum returns with an non-zero exit code, which breaks the build.

mlassnig:
I'm wondering if all these rse2site-like methods wouldn't be a better fit for common/util.py ?

vingar:
site is a rse attribute and the core methods related to rse_attributes are in in core/rse.  There are already some code duplication between the methods in core/rse and the rse2site methods.  If it is needed for the clients, I would rather put them in rse.clients

mlassnig:
@bari12 feel free to undo my approve if you want @jwackito to externalise the site/rse methods beforehand

mlassnig (approved):

PR #283, opened 2017-12-05
add extra dependencies Fix #281

### LINKED ISSUE #281: Add Extras dependency on the possible rucio backend (oracle, mysql, postgresql)
Motivation
----------
Express the "recommended" dependencies for the db backend

Modification
------------
Change of the main setup.py script to include extra dependencies

Expected behavior
-----------------
pip install rucio[oracle]
pip install rucio[postgresql]
pip install rucio[mysql]

or maybe define later some meta packages like rucio-oracle with install_requires = ["rucio[oracle]"]

vingar:


setup_rucio.py
```diff
@@ -113,6 +113,14 @@ def write_requirements():
         req_file.close()
 
 
+oracle_extras = ['cx_oracle>=5.1']
```
mlassnig, line 116: is it necessary to give the versions here?

vingar: necessary, I don't think. More recommended as sometimes the cx_oracle version are not strictly compatible. At some point, it has to be in sync with the versions used in CI/travis and in production

bari12: The extras are only installed if you give this as a parameter on the pip install, correct?

vingar: yes, it's one possibility:
    pip install rucio[oracle]

One additional is to have a meta package requiring this installation. like:
package rucio-oracle 
with:
install_requires = ["rucio[oracle]"]

specified in the setup.py

mlassnig: for the versions, ok, as long as we don't force an old version, i think it's fine.

metapackages: would be good imho. i didn't know about this pip install module[extra] mechanism for example. it's not obvious i think.

vingar: cf. http://peak.telecommunity.com/DevCenter/setuptools#declaring-extras-optional-features-with-their-own-dependencies

bari12 (approved):

PR #286, opened 2017-12-05
setup.py fixes for buildthedocs fix #285

### LINKED ISSUE #285: Adapt setup.py for buildthedocs
Motivation
----------
The kerberos packages are blocking the doc generation on buildthedocs

Modification
------------
Modify the main setup.py to not include kerberos packages by default but in extra requires

Expected behavior
-----------------
Working documentation generation on buidthedocs

vingar:


bari12 (dismissed):
This is killing the packaging like this. It should be only in setup_rucio.py

vingar:
this version differs from setup_rucio.py as it excludes the kerberos packages and put them in extra_dependencies to get buildthedocs working (with a setup.py without kerberos)

bari12:
Ah understood. So this is only our readthedocs setup.py - but it actually is never used in the normal build procedure. Can you add this as a comment to the top of the file?

bari12 (approved):

PR #292, opened 2017-12-05
Include postgres client into the CC7 docker image; Fix #282

### LINKED ISSUE #282: Include postgres client into the CC7 docker image
Motivation
----------
Need to provide postgres support

Modification
------------
Include postgres client into the CC7 docker image

cserf:
Include postgres client into the CC7 docker image; Fix #282


vingar:
once the new setup.py are there and published on pypi, we should be able to do:

pip install rucio[postgresql]

PR #306, opened 2017-12-06
protocols: add fix for generic parse_pfns

mlassnig:
protocols: add fix for generic parse_pfns


bari12 (approved):

vingar:
I'm not so sure we need a parse_pfns functions per protocol impl. I would assume that the one defined in protocol.py covers almost all cases.  We can try to delete the one in gfal/srm and try with the inherited one ?

PR #316, opened 2017-12-08
dependencies: pin paramiko for rucio-clients to 1.18.4

mlassnig:
dependencies: pin paramiko for rucio-clients to 1.18.4


tools/pip-requires-client
```diff
@@ -17,4 +17,4 @@ bz2file>=0.98               # Read and write bzip2-compressed files.
 python-magic>=0.4.13        # File type identification using libmagic
 futures>=3.1.1              # Clean single-source support for Python 3 and 2
 six>=1.11.0                 # Python 2 and 3 compatibility utilities
-paramiko>=2.4.0             # SSH2 protocol library
+paramiko==1.18.4            # SSH2 protocol library
```
vingar, line 20: moving it in extra-requires together with kerberos would be good..

bari12:
Can you prepare it as a feature please. I know it is just a downgrade, but I think it is better to stick to that rule and we can make a feature release just after the holidays anyway.

PR #324, opened 2017-12-11
client: Automatically discover X509 proxy using GSI rules Fix #323

### LINKED ISSUE #323: Automatically find user's X509 proxy
Motivation
----------
Currently, direct users of the `rucio` tools must setup the `$X509_USER_PROXY` environment variable and/or set a line in their `rucio.cfg`.  There are various standard ways to auto-discover the proxy in the user's environment.  Let's use them!

bbockelm:
With this, the rucio client will automatically discover the X509 proxy when
not in `rucio.cfg`, using the classic GSI rules:
- `$X509_USER_PROXY` if set,
- `/tmp/x509up_u$(id -u)` if it exists.

If none of these work, the client will give an error along the following lines:

```
2017-12-11 11:23:19,385 ERROR [Client parameters are missing.
Details: Cannot find a valid X509 proxy; not in /tmp/x509up_u1221, $X509_USER_PROXY not set, and 'x509_proxy' not set in the configuration file.]
```

### DIFF
--- a/lib/rucio/client/baseclient.py
+++ b/lib/rucio/client/baseclient.py
@@ -12,6 +12,7 @@
 # - Ralph Vigne, <ralph.vigne@cern.ch>, 2015
 # - Martin Barisits, <martin.barisits@cern.ch>, 2017
 # - Mario Lassnig, <mario.lassnig@cern.ch>, 2017
+# - Brian Bockelman, <bbockelm@cse.unl.edu>, 2017
 #
 # Client class for callers of the Rucio system
 
--- a/lib/rucio/client/baseclient.py
+++ b/lib/rucio/client/baseclient.py
@@ -25,7 +25,7 @@ from rucio.common.utils import build_url, get_tmp_dir, my_key_generator, parse_r
 from rucio import version
 
 from logging import getLogger, StreamHandler, ERROR
-from os import environ, fdopen, path, makedirs
+from os import environ, fdopen, path, makedirs, geteuid
 from shutil import move
 from tempfile import mkstemp
 from urlparse import urlparse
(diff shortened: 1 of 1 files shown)


mlassnig:
Hi Brian,

fantastic, thanks.

One more thing, did you do a git push instead of tools/submit-pull-request? This is a patch, so there should be one against the master branch and one against the next branch, but i can only see this one against master though. tools/submit-pull-request takes care of all that.

@bari12 we have to make this more clear in the contributing guidelines

lib/rucio/client/baseclient.py
```diff
@@ -25,7 +25,7 @@
 from rucio import version
 
 from logging import getLogger, StreamHandler, ERROR
-from os import environ, fdopen, path, makedirs
+from os import environ, fdopen, path, makedirs, geteuid
```
mlassnig, line 29: can you please add yourself to the list of authors at the top of the file?

lib/rucio/client/baseclient.py
```diff
@@ -136,7 +136,22 @@ def __init__(self, rucio_host=None, auth_host=None, account=None, ca_cert=None,
...
+                        # - /tmp/x509up_u`id -u` otherwise.
+                        # If neither exists (at this point, we don't care if it exists but is invalid), then rethrow
+                        if 'X509_USER_PROXY' in environ:
+                            self.creds['client_proxy'] = environ['X509_USER_PROXY']
+                        else:
+                            fname = '/tmp/x509up_u%d' % geteuid()
+                            if path.exists(fname):
+                                self.creds['client_proxy'] = path
+                            else:
+                                raise MissingClientParameter('Cannot find a valid X509 proxy; not in %s, $X509_USER_PROXY not set, and '
```
mlassnig, line 154: can you just verify if this gets a unique exit code for the client? there are a few people who rely on the exit codes of the clients and the more "unique" we can make them the better.

vingar: self.error_code = 42 ?

mlassnig: a fitting error code ;-)

vingar: not clear. Do you mean a different exception ?

bbockelm: Yeah - I can confirm that this raises the exit code expected for the `MissingClientParameter` exception:
```
$ rucio ping
2017-12-12 12:49:13,237 ERROR   Client parameters are missing.
Details: Cannot find a valid X509 proxy; not in /tmp/x509up_u1221, $X509_USER_PROXY not set, and 'x509_proxy' not set in the configuration file.
$ echo $?
42
```

@mlassnig - from your comment, it's not clear that you want me to simply verify that fact or if you want a unique exit code for this particular missing parameter.

bbockelm:
@bari12 @mlassnig - for the `tools/submit-pull-request` stuff: if you could, humor me a bit and let me do things manually so I can figure out the guts of how things flow.  As I get used to it, I'm sure I'll switch over to the normal toolset.

Anyhow, I think I addressed both comments from Mario's review and it should be ready for re-review.

vingar:
you can have a look on https://github.com/rucio/rucio/pull/333 

I extended a bit the contributing notes for the contributors not using the helper script (in favor of hub or the github interface)

bbockelm:
yup - trying out `hub` right now.  I'll work through a few variants and see which I like the best.

mlassnig (approved):

PR #326, opened 2017-12-11
Add simple dynamic bash completion code for RSEs.  Fix #325

### LINKED ISSUE #325: client: dynamic bash completion
Motivation
----------
Bash completion (e.g., hitting `<tab>`) provides a powerful way to improve the the user experience.

With `argcomplete`, most of the static suggestions are present.  However, dynamic items like `rucio list-rse-attributes <tab>` (which should suggest available RSEs) don't work.

Let's add some smarter auto-complete logic!

bbockelm:
When an RSE argument is needed, simply query the server to have it provide some suggestions.

### DIFF
--- a/bin/rucio
+++ b/bin/rucio
@@ -2420,6 +2420,15 @@ def touch(args):
         client.touch(scope, name, args.rse)
 
 
+# Auto-completers for argument parsing.
+def rse_completer(prefix, parsed_args, **kwargs):
+    """
+    Completes the argument with a list of RSEs
+    """
+    client = get_client(parsed_args)
+    return ["%(rse)s" % rse for rse in client.list_rses()]
+
+
 if __name__ == '__main__':
     usage = """
 usage: %(prog)s <command> [options] [args]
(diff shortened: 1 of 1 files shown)


bin/rucio
```diff
@@ -2420,6 +2420,16 @@ def touch(args):
...
 
 
+# Auto-completers for argument parsing.
+def rse_completer(prefix, parsed_args, **kwargs):
+    """
+    Completes the argument with a list of RSEs
+    """
+    client = get_client(parsed_args)
+    rses = ["%(rse)s" % rse for rse in client.list_rses(None)]
+    return rses
```
vingar: return [rse for rse in client.list_rses()]
or maybe just 
return client.list_rses() 
I don't remember if list_rses is a generator or not

bbockelm: It is a generator, and `argcomplete` didn't seem to like generators.

Cleaned up now -- one still needs the `"%(rse)s" % rse` part because each RSE returned is actually a dictionary (with a key named `rse`), not a string.

mlassnig:
Thanks! Same comment as for the other patch (--> tools/submit-pull-request)

mlassnig (approved):

PR #328, opened 2017-12-11
[client] Add rucio-admin support for protocols

bbockelm:
Currently, the only way protocols can be changed for a particular RSE is through a direct invocation of the REST or python APIs.

This adds support to the `rucio-admin` CLI.  Currently, this is limited to `add` / `delete`.  If the general approach for `add` looks good, then `update` should be straightforward.

For the reviewer - it's not clear how to best handle the `domains` key.  Advice is appreciated - right now, it's just a simply JSON dictionary passed from the CLI (a tad distasteful).

### DIFF
--- a/bin/rucio-admin
+++ b/bin/rucio-admin
@@ -16,6 +16,7 @@
  - Cedric Serfon, <cedric.serfon@cern.ch>, 2014-2016
  - Cheng-Hsi Chao, <cheng-hsi.chao@cern.ch>, 2014
  - Joaquin Bogado, <joaquin.bogado@cern.ch>, 2014-2015
+ - Brian Bockelman, <bbockelm@cse.unl.edu>, 2017
 """
 
 import argparse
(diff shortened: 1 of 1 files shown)


bbockelm:
Example `add-protocol` invocation:
```
rucio-admin rse add-protocol --hostname red-gridftp.unl.edu --scheme gsiftp --prefix /dropfiles/store T2_US_NEBRASKA_USERDISK --domain-json '{"wan": {"read": 1, "write": 1, "delete": 1}}'
```

Example `delete-protocol`:
```
rucio-admin rse delete-protocol T2_US_NEBRASKA_USERDISK --scheme gsiftp
```

mlassnig:
Thanks once more! (see comment on the other patch for --> tools/submit-pull-request)

bin/rucio-admin
```diff
@@ -501,6 +501,54 @@ def delete_attribute_rse(args):
     return SUCCESS
 
 
+@exception_handler
```
mlassnig, line 505: please add yourself as author to the top of the file

mlassnig (approved):

PR #337, opened 2017-12-12
clients: Import extra modules only if they are available Fix #336

### LINKED ISSUE #336: Import extra modules only if they are available
Motivation
----------
Some modules are not necessary in the clients and should be imported only if available or installed in the environment

Modification
------------
* Define a list of extra packages, check for their presence and import them if there
* Add a new exception missingModuleException and raise it from the method which can not worked with the missing module

vingar:


mlassnig:
Lots of "ValueError: No JSON object could be decoded" which points to a server error? Looks unrelated to the patch. I'm restarting the CI.

vingar:
it should be because of missing dependencies which should go in the main requirements files now as they are optional in the clients

vingar:
one test failing against postgresql, weird 
FAIL: DATA IDENTIFIERS (CLIENT): List dids by pattern.
  File "/usr/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/opt/rucio/lib/rucio/tests/test_did.py", line 187, in test_list_dids
    assert_equal(len(results), 2)
AssertionError: 3 != 2

bari12:
I have restarted the tests here too, lets see if it helps.

bari12:
Ok now its a bunch of pylint/flake output. Did somebody change the tests scripts?

vingar:
I moved the flake8/pylint tests at the beginning as it's faster to get those check first. 
If the pylint output is there it means that there is one pylint error somewhere

bari12:
Hmm I don't see that change in the changed files. Also we would need that on master as well then.

bari12 (approved):

PR #357, opened 2017-12-14
core:support for multiple permission and schema policy Fix #356

### LINKED ISSUE #356: support for multiple permission and schema policy
Motivation
----------
Schema is not configurable per collaboration (cms use case)

Modification
------------

Extend the permission and schema part to allow 
[policy]
permision = cms
schema = cms

from the configuration

vingar:


lib/rucio/common/schema/__init__.py
```diff
@@ -0,0 +1,26 @@
+# Copyright European Organization for Nuclear Research (CERN)
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# You may not use this file except in compliance with the License.
+# You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+
```
mlassnig: author missing

lib/rucio/common/schema/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+if config.config_has_section('policy'):
+    try:
+        POLICY = config.config_get('policy', 'schema')
+    except (NoOptionError, NoSectionError) as error:
+        POLICY = 'generic'
+else:
+    POLICY = 'generic'
+
+if POLICY.lower() == 'generic':
+    from .generic import *  # NOQA pylint:disable=wildcard-import
```
mlassnig, line 25: pylint complains for me on all the relative imports with W0403, better to do full imports

lib/rucio/core/permission/__init__.py
```diff
@@ -17,17 +17,23 @@
...
-
-    if POLICY.lower() == 'generic':
-        from rucio.core.permission.generic import *  # NOQA pylint:disable=wildcard-import
-    elif POLICY.lower() == 'atlas':
-        from rucio.core.permission.atlas import *  # NOQA pylint:disable=wildcard-import
-    else:
-        from rucio.core.permission.generic import *  # NOQA pylint:disable=wildcard-import
+elif config.config_has_section('policy'):
+    try:
+        POLICY = config.config_get('policy', 'permission')
```
mlassnig, line 24: update to rucio.cfg.template needed

lib/rucio/common/schema/__init__.py
vingar: weird!
pylint lib/rucio/common/schema/__init__.py- --> Your code has been rated at 10.00/10 (previous run: 10.00/10, +0.00)
and no complains.

$ pylint --version
pylint 1.7.5, 
astroid 1.5.3
Python 2.7.10 (default, Jul 15 2017, 17:16:57) 
[GCC 4.2.1 Compatible Apple LLVM 9.0.0 (clang-900.0.31)]

lib/rucio/common/schema/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+elif POLICY.lower() == 'atlas':
```
bari12, line 26: Can't we do the import completely dynamic? I mean in this case we still basically need to maintain a list of all organisations we support.

mlassnig: this is what i did for pilot2:


 workflow = __import__('pilot.workflow.%s' % args.workflow, globals(), locals(), [args.workflow], -1)
workflow.run(args)

so, whatever module name string you give to __import__, it will import with the correct python context

same logic can be applied here

vingar: with policy loader and entry_point, this should be doable (for the next feature release)
unless you have some immediate suggestions ?

vingar: setuptools can provide this. Example:
import pkg_resources
for impl in pkg_resources.iter_entry_points(
                    self.group, name):
                self.impls[name] = impl.load
                return impl.load()

bari12 (approved):

PR #379, opened 2017-12-14
Allow more explicit control of the LFN and PFN.  Fixes #377

### LINKED ISSUE #377: CMS: Allow `rucio upload` to accept alternate LFNs
Currently, `rucio upload` is a bit tied to the ATLAS concept that the filename on disk (`foo.root`) is going to match the filename in Rucio (`user.bbockelm:foo.root`).

This is a bit broken by CMS, where the LFN differs significantly from the name on disk.  We should allow the LFN to be specified via a CLI flag, just like the GUID.

The `rucio upload` interface should, by default, retain the current behavior.

bbockelm:
This allows the upload path to track the "file name" (on local disk) and the LFN separately, allowing the `rucio upload` client to expose the option of explicitly providing a LFN that differs from the file
name.

Additionally, to help upload files with '/' in them that are effectively pre-hashed, this allows the RSE to be marked as having the 'identity' determinism type, meaning to skipping the hashing step of generating a PFN.

This is the `next` equivalent of #378.

### DIFF
--- a/bin/rucio
+++ b/bin/rucio
@@ -901,7 +901,6 @@ def upload(args):
         return FAILURE
     list_files = []
     files_to_list = []
-    lfns = {}
     revert_dict = {}
     if args.scope:
         fscope = args.scope
@@ -968,9 +967,6 @@ def upload(args):
                 logger.debug('Automatically setting new GUID')
                 current_file_info['meta']['guid'] = generate_uuid()
             list_files.append(current_file_info)
-            directory_worklist = lfns.setdefault(dir_name, [])
-            # TODO: why is this different from current_file_info?
-            directory_worklist.append({'filename': base_name, 'name': name, 'scope': fscope, 'adler32': checksum, 'filesize': size})
             revert_dict[fscope, name] = dir_name
 
         except OSError as error:
--- a/bin/rucio
+++ b/bin/rucio
@@ -932,24 +932,24 @@ def upload(args):
         logger.error("A single GUID was specified on the command line, but there are multiple files to upload.")
         logger.error("If GUID auto-detection is not used, only one file may be uploaded at a time")
         return FAILURE
-    if len(files) > 1 and args.lfn:
+    if len(files) > 1 and args.name:
         logger.error("A single LFN was specified on the command line, but there are multiple files to upload.")
         logger.error("If LFN auto-detection is not used, only one file may be uploaded at a time")
         return FAILURE
-    for name in files:
+    for filename in files:
         try:
-            size = os.stat(name).st_size
-            checksum = adler32(name)
-            base_name = os.path.basename(name)
-            dir_name = os.path.dirname(name)
-            lfn = base_name
-            if args.lfn:
-                lfn = args.lfn
-            logger.debug('Extracting filesize (%s) and checksum (%s) for file %s:%s' % (str(size), checksum, fscope, lfn))
-            files_to_list.append({'scope': fscope, 'lfn': lfn, 'name': base_name})
-            current_file_info = {'scope': fscope, 'name': base_name, 'lfn': lfn, 'bytes': size, 'adler32': checksum, 'state': 'C', 'meta': {}}
-            if not args.guid and 'pool.root' in name.lower() and not args.no_register:  # is a root file, getting the GUID
-                status, output, err = execute('pool_extractFileIdentifier {0}'.format(name))
+            size = os.stat(filename).st_size
+            checksum = adler32(filename)
+            base_name = os.path.basename(filename)
+            dir_name = os.path.dirname(filename)
+            name = base_name
+            if args.name:
+                name = args.name
+            logger.debug('Extracting filesize (%s) and checksum (%s) for file %s:%s' % (str(size), checksum, fscope, name))
+            files_to_list.append({'scope': fscope, 'lfn': name, 'name': base_name})
+            current_file_info = {'scope': fscope, 'name': base_name, 'lfn': name, 'bytes': size, 'adler32': checksum, 'state': 'C', 'meta': {}}
+            if not args.guid and 'pool.root' in filename.lower() and not args.no_register:  # is a root file, getting the GUID
+                status, output, err = execute('pool_extractFileIdentifier {0}'.format(filename))
                 if status != 0:
                     logger.error('Trying to upload ROOT files but pool_extractFileIdentifier tool can not be found.')
                     logger.error('Setup your ATHENA environment and try again.')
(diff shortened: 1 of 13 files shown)


vingar:
bin/rucio:934:21: E225 missing whitespace around operator
bin/rucio:1153:56: E127 continuation line over-indented for visual indent
lib/rucio/rse/rsemanager.py:343:4: W291 trailing whitespace
lib/rucio/rse/rsemanager.py:346:12: W291 trailing whitespace

bari12:
There are some implications of the determinism_type which we have to check, because this reaches much further than just the protocol as it reaches into the transfers too. But I have some basic questions to this as well, will post them to the issue.

vingar:
there are some comments on the issue as well. We need to discuss and think a bit here

bin/rucio
```diff
@@ -918,12 +918,26 @@ def upload(args):
     trace['hostname'] = socket.getfqdn()
     trace['scope'] = fscope
     trace['uuid'] = generate_uuid()
+    if len(files) > 1 and args.guid:
+        logger.error("A single GUID was specified on the command line, but there are multiple files to upload.")
+        logger.error("If GUID auto-detection is not used, only one file may be uploaded at a time")
+        return FAILURE
+    if len(files) > 1 and args.lfn:
```
bari12: To be consistent with rucio terminology `lfn` should be renamed to `name`. And the parts below which use the variable `name` should be renamed to `filename`. Thus if no --name is provided, the base_name of the filename is used as name, otherwise the explicitly defined name. This should also be consistent throughout the protocols, as now the name is not necessarily the filename anymore.

lib/rucio/rse/protocols/protocol.py
```diff
@@ -65,7 +65,8 @@ def lfns2pfns(self, lfns):
 
         lfns = [lfns] if type(lfns) == dict else lfns
         for lfn in lfns:
-            scope, name = lfn['scope'], lfn['name']
+            print "generating lfn for", lfn
```
bari12: Remove print

lib/rucio/rse/protocols/protocol.py
```diff
@@ -99,7 +100,8 @@ def __lfns2pfns_client(self, lfns):
         lfns = [lfns] if type(lfns) == dict else lfns
         for lfn in lfns:
             scope = lfn['scope']
-            name = lfn['name']
+            name = lfn.get('lfn', lfn['name'])
+            print "Processing LFN", lfn
```
bari12: Remove print

lib/rucio/rse/protocols/protocol.py
```diff
@@ -45,7 +45,7 @@ def __init__(self, protocol_attr, rse_settings):
             if getattr(rsemanager, 'SERVER_MODE', None):
                 setattr(self, '_get_path', self._get_path_nondeterministic_server)
         else:
-            self.attributes['determinism_type'] = 'default'
+            self.attributes['determinism_type'] = self.rse.get('determinism_type', 'default')
```
bari12: Need to check if `determinsm_type` is part of the rse_settings. I am not sure.

vingar: it's deterministic, it is defined on rse and should be in rse_seetings

lib/rucio/rse/protocols/protocol.py
```diff
@@ -120,7 +122,10 @@ def _get_path(self, scope, name):
         hstr = hashlib.md5('%s:%s' % (scope, name)).hexdigest()
         if scope.startswith('user') or scope.startswith('group'):
             scope = scope.replace('.', '/')
-        return '%s/%s/%s/%s' % (scope, hstr[0:2], hstr[2:4], name)
+        if self.attributes.get('determinism_type', 'default') == 'identity':
```
bari12: For this part I really wonder if it would be beneficial if it comes from the policy and not the determinism_type from the RSE. e.g. for CMS the 'default' policy could be defined as `'%s/%s' % (scope, name)` while for ATLAS it is the hash part. Both have advantages, but I am mostly wondering if you forget to set the determinism_type for a newly created RSE you will get the hashed pathes until you fix it, while by setting the default via the policy that won't happen. Any preferences?

lib/rucio/rse/protocols/protocol.py
bari12: Yes, but deterministic is a boolean flag of the RSE. So for this you would need to have both the flag plus the determinism_type.

bbockelm:
Hm.  For some reason, GitHub didn't automatically link this to #377.  This comment should do the trick..

bari12:
Now that everyone is back in the office, coming back to this.

I think what still should be done is:
- Remove the prints
- `lfn` should be renamed to `name`, so that the CLI parameter is `--name`. This is more consistent with Rucio terminology. Thus internally, everywhere where we use the variable `name` now, it has to be renamed to `filename`. If `--name` is not given, the `base_name` of the file is used as `name`.
- `self.rse.get('determinism_type', 'default')` will not work, as determinism_type is not part of `core.rse.get_rse_protocols` return value. This has to be added too.
  
Otherwise I think the patch is ready.

bbockelm:
(note: I just rebased the branch in order to maintain sanity)

bbockelm:
Travis failures appear unrelated?

bin/rucio
```diff
@@ -1003,13 +1018,28 @@ def upload(args):
     else:
         logger.debug('Skipping dataset registration')
 
+    def make_valid_did(lfn_dict):
```
bari12: I think you do not need the `make_valid_did` part anymore (same for the rsemanager) as the dictionary describing the file will always have the `scope`, `name`, `filename` now?

bin/rucio
```diff
@@ -946,17 +960,18 @@ def upload(args):
...
                 logger.debug('Automatically setting new GUID')
-                list_files.append({'scope': fscope, 'name': os.path.basename(name), 'bytes': size, 'adler32': checksum, 'state': 'C', 'meta': {'guid': generate_uuid()}})
-            if not os.path.dirname(name) in lfns:
-                lfns[os.path.dirname(name)] = []
-            lfns[os.path.dirname(name)].append({'name': os.path.basename(name), 'scope': fscope, 'adler32': checksum, 'filesize': size})
-            revert_dict[fscope, os.path.basename(name)] = os.path.dirname(name)
+                current_file_info['meta']['guid'] = generate_uuid()
+            list_files.append(current_file_info)
+            directory_worklist = lfns.setdefault(dir_name, [])
+            # TODO: why is this different from current_file_info?
```
bari12: I don't know. To me it looks like we do not need the `directory_worklist` at all?

bin/rucio
bari12: Do we still need the `make_valid_did` method? Also in the rsemanager? It looks like to me that the file dictionary now always consists of `scope`, `name` and `filename`, thus it is not needed anymore?

bbockelm: `filename` isn't a valid key in the JSON schema for a DID.  Hence, if we don't remove it, it fails the JSON validation later...

bari12: Right, understood. In that case I would suggest we add it to rucio.common.utils, so the code doesn't need to be duplicated. Then the patch is ready from my side 🎉  🕺 🎆

bbockelm:
Alright, I think this one is ready to roll!

mlassnig (commented):
great! thanks a lot, brian! i think this is ready, but i leave the final signoff to @bari12

vingar (approved):
phew what a pull request ! all comments from Martin have been implemented and looks fine for me.

bbockelm:
Indeed, in retrospect, it conflated two issues that could have been made into separate PRs.  Hopefully can avoid that next time!

PR #416, opened 2017-12-20
core: Configurable scope-name regexp per organization policy Fix #411

### LINKED ISSUE #411: Client: Check URL escaping code for handling of /
If we want `/` to be a valid character in the LFN, URL escaping in the client suddenly becomes more relevant.

For example, on the CMS test instance, I noticed the following GET request in the logs:

```
GET /dids/user.bbockelm//store/mc/SAM/GenericTTbar/AODSIM/CMSSW_9_2_6_91X_mcRun1_realistic_v2-v1/00000/AE237916-5D76-E711-A48C-FA163EEEBFED.root/meta HTTP/1.1"
```

We probably wanted to issue this request:

```
GET /dids/user.bbockelm/%2Fstore%2Fmc%2FSAM%2FGenericTTbar%2FAODSIM%2FCMSSW_9_2_6_91X_mcRun1_realistic_v2-v1%2F00000%2FAE237916-5D76-E711-A48C-FA163EEEBFED.root/meta HTTP/1.1"
```

vingar:


lib/rucio/common/schema/cms.py
```diff
@@ -339,6 +339,8 @@
                      "type": "string",
                      "pattern": r'^[a-z0-9-_]{1,30}$'}
 
+SCOPE_NAME_REGEXP = '/([^/]*)(?=/)(.*)'
```
bari12, line 342: Doesn't this add a `/` in front of every `name`, even if it is not wanted?

vingar: All names in CMS start with a / (See name regular expression in schema).

mlassnig (approved):

PR #434, opened 2018-01-09
client: Recreate session after authorization failure. Fix #433

### LINKED ISSUE #433: Avoid session reuse when new token is needed
Motivation
----------
In testing of `rucio upload`, I've sporadically hit cases where:
1. In a sequence of REST API calls, one call requires the client to retrieve a new Rucio token.
2. The client will contact the token issuer endpoint to retrieve a new token.
3. Since the client reuses the session object, an existing SSL session may be utilized.
4. If the existing SSL session wasn't created with an X509 user proxy (or established via TLS session reuse), then `mod_gridsite` isn't able to pass along the appropriate headers about the client DN.
5. Without the client DN, the issuer (correctly) refuses to issue a new token.

Modification
------------
Whenever the client receives an authorization failure, it should create a new session object prior to retrying the token retrieval.

bbockelm:
I've encountered sporadic cases where session reuse against the token endpoint causes failures because the session was not established using the client certificate (meaning issuing a new token will fail).

With this patch, failures are properly detected and - if it's an auth failure - will trigger the creation of a fresh session.

### DIFF
--- a/lib/rucio/client/baseclient.py
+++ b/lib/rucio/client/baseclient.py
@@ -277,6 +277,7 @@ class BaseClient(object):
                 continue
 
             if result is not None and result.status_code == codes.unauthorized:  # pylint: disable-msg=E1101
+                self.session = session()
                 self.__get_token()
                 hds['X-Rucio-Auth-Token'] = self.auth_token
             else:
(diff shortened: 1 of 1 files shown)


bbockelm:
Travis-CI failure appears unrelated (`yum` failure).  Could an admin please trigger a new build?

bari12:
Restarted

bari12 (approved):

PR #440, opened 2018-01-10
database: bootstrap mismatch between SSH and GSS

mlassnig:
database: bootstrap mismatch between SSH and GSS


bari12:
Just on a side note, you don't need to change it, but please put the issue number in the commit message. It is really much easier as then you can just click on the link from the merge request and get to the issue.

bari12 (approved):

PR #442, opened 2018-01-10
core: make list_replicas domain aware; Fix #64

### LINKED ISSUE #64: Include the network domain (wan, lan) in list_replicas
Motivation
-----------

For local usage, a combination of wan/lan attributes with rse_expressions will provide the local or remote replicas to a site.

Modification
-------------
added lan/wan in  list_replicas (clients/rest/server). default wan

mlassnig:
core: make list_replicas domain aware; Fix #64

### DIFF
--- a/lib/rucio/api/replica.py
+++ b/lib/rucio/api/replica.py
@@ -6,7 +6,7 @@
 #
 # Authors:
 # - Vincent Garonne, <vincent.garonne@cern.ch>, 2012-2016
-# - Mario Lassnig, <mario.lassnig@cern.ch>, 2012, 2017
+# - Mario Lassnig, <mario.lassnig@cern.ch>, 2012, 2017-2018
 # - Cedric Serfon, <cedric.serfon@cern.ch>, 2014-2016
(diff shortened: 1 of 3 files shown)


vingar:
The corresponding tests are missing. 
Also wondering if it's not better to consider wan/lan as an ordered list ,i.e., domains instead of a single domain value as listing replicas with lan and then wan (per job) will generate much more load on the servers.

mlassnig:
There's very few cases where the LAN override actually happens and can be controlled directly from the pilot, so not a problem. For the rucio-clients we need to re-engineer the replica listing JSON reply to take into account the WAN/LAN, but I would much prefer to go to metalink for this anyway.

vingar:
After checking the code, the answer can be extended and includes directly the domain for which the pfn is valid:
```
file['pfns'][pfn]` = {'rse': rse,
                                       'type': str(rse_type),
                                       'volatile': volatile,
                                       'domains': ('lan', 'wan')}
```
as such the clients won't need to call several times list_replicas

bari12 (approved):

PR #446, opened 2018-01-10
Clients: Pass lifetime on from add_container to add_did Fix #444

### LINKED ISSUE #444: Lifetime in add_container ignore
Motivation
----------
Lifetime is a valid parameter of add_container(), but is ignored

Modification
------------

Expected behavior
-----------------
Should be passed through to add_did()

ericvaandering:


bari12:
Hi @ericvaandering ,
You do not need to make two different branches (one for next, one for master), the one you already made: `ericvaandering:patch-444-container_lifetime` should be merged against both branches (next and master) - So this PR should be closed and you should create a new PR which merges the existing ericvaandering:patch-444-container_lifetime to next.
Cheers,
Martin

ericvaandering:
Ok, thanks   Will do.

Issue #457, opened 2018-01-11
Crash when declaring bad file replicas already declared bad on a non-deterministic RSE


cserf:
Motivation
----------
__declare_bad_file_replicas crash when one declare bad file replicas already declared bad on non-deterministic RSE because of the check : rowcount != len(declared_replicas)

Modification
------------
Remove the files already declared from the declared_replicas list.

PR #458, opened 2018-01-11
Recovery : Crash when declaring bad file replicas already declared bad on a non-deterministic RSE #457

cserf:
Recovery : Crash when declaring bad file replicas already declared bad on a non-deterministic RSE #457


bari12:
Add bad replicas test fails.

bari12:
Push, would be nice to get this into todays release. 😄

bari12 (approved):

PR #470, opened 2018-01-12
Manipulate distance information via CLI.  Fix #469

### LINKED ISSUE #469: clients: Add CLI for manipulating distance information
Motivation
----------
We added a REST API for manipulating the distance information between two links.  However, we did not add the corresponding CLI.

Modification
------------
Add corresponding CLI.

Expected behavior
-----------------
Set very basic distance information (ranking and distance) via CLI.  More detailed attribute manipulation can remain part of the REST API.

bbockelm:
This adds the ability to set, update, and get distance information between two RSEs via the `rucio-admin` CLI.

This only exposes the `ranking` and `distance` attributes: others will require usage of the REST API still.

Fixes #469.

### DIFF
--- a/bin/rucio-admin
+++ b/bin/rucio-admin
@@ -503,6 +503,58 @@ def delete_attribute_rse(args):
 
 
 @exception_handler
+def set_distance_rses(args):
+    """
+    %(prog)s set-distance [options] SOURCE_RSE DEST_RSE
+
+    Set the distance between two RSEs.
+    """
+    client = get_client(args)
+    params = {'ranking': args.ranking, 'distance': args.distance}
+    client.add_distance(args.source, args.destination, params)
+    print 'Set distance from %s to %s to %d with ranking %d' % (args.source, args.destination, args.distance, args.ranking)
+    return SUCCESS
+
+
+@exception_handler
+def get_distance_rses(args):
+    """
+    %(prog)s get-distance SOURCE_RSE DEST_RSE
+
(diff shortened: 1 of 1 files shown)


mlassnig (commented):
there isn't really a difference from the user point of view between set and update. i'd remove the update.

bbockelm:
@mlassnig - I thought the same thing.  Of course, internally, the map to two different APIs.

That would imply a loss of efficiency and atomicity as one would have to effectively first query whether the link already exists, then change the API call.  I think that's OK for the CLI ... thoughts?

I figure for the utmost efficiency and correctness, users should utilize the REST directly.

bari12:
I think it is fine to have both, I am just wondering about naming it `set` and `update`. Update is fine I think, but setting something to me means updating it if it exists, or creating it if it doesn't. But here it just creates and does not update? So I wonder if it wouldn't be better to name it `add` (respective the CLI command `add-distance`) as well, similar to how the method is named in the python client.

vingar:
yes, `add` is more consistent (e.g., add rse, add account for HTTP POST operation).
For the update operation (HTTP PUT), `update` looks better for me here.

bbockelm:
@vgaronne @mlassnig @bari12 - sorry for the delay (got sidetracked on other PRs).  The rename of the command is done, as requested.

Anything else I missed for this one?

bari12 (approved):

Issue #140, opened 2017-11-21
extend download/upload for cloud checksums


berghaus:
Motivation
----------
Cloud storage solutions have opted to use MD5 to verify file integrity. To support cloud storage and ensuring file integrity the MD5 digests of files added to the data catalogue should be computed.

Modification
------------
 - [x] Client computes md5 checksum on file upload and adds this to the file metadata
 - [x] Client download checks the file adler32 or md5 against the metadata - preference of adler32

Expected behavior
-----------------
The MD5 field of the file metadata in the Rucio database will be filled with the MD5 digest of the file for new files.

mlassnig:
+1

bari12:
Hi Frank, just to confirm (as nobody is assigned to this ticket): You are planning to implement this, right? 😃

berghaus:
There is no one assigned to the ticket because I can't assign anyone. I am planning to implement this. Sorry about the long delay, I got absorbed by something else for a moment. Back at it now - starting with dealing with the repo reset ;-)

bari12:
I just added you as collaborator for the repo :) You should be able to assign now.

berghaus:
That worked, thanks!

berghaus:
Going to work on #322 first to set up my dev environment (if you guys like the suggestions that is).

PR #472, opened 2018-01-15
Clients: Validate against MD5 if no ADLER32 exists #140

berghaus:
Clients: Validate against MD5 if no ADLER32 exists #140


bari12 (commented):
To me it looks good 👍  Let's see if @mlassnig finds something :)

mlassnig (approved):
tiny comments, maybe for further cleanup patch:

- add yourself as author to the headers (and it's 2018, not 2017) :-)
- sanitize variables (eg., adler32 checksum vs md5_digest)
- you might actually test that the "hello test\n" md5sum is correctly reported

berghaus:
Thank you for the feedback Mario, I'm happy to make to add that cleanup. I was shy about renaming the `checksum` variable, hence the not so sane variable names. The 'hello world' test is a good idea, had not thought of that :-)

PR #491, opened 2018-01-18
Make used of SSL configurable for server containers

tbeerman:
Make used of SSL configurable for server containers

### DIFF
--- a/etc/docker/server/rucio.conf.j2
+++ b/etc/docker/server/rucio.conf.j2
@@ -1,11 +1,13 @@
+{% if RUCIO_ENABLE_SSL|default('False') == 'True' %}
 LoadModule ssl_module /usr/lib64/httpd/modules/mod_ssl.so
+{% endif %}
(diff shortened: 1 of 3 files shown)


etc/docker/server/rucio.conf.j2
```diff
@@ -30,16 +33,15 @@ CacheRoot /tmp
...
  #AB: SSLv3 disable
  SSLProtocol              all -SSLv2 -SSLv3
  #AB: for Security
  SSLCipherSuite           HIGH:!CAMELLIA:!ADH:!aNULL:!eNULL:!EXPORT:!DES:!RC4:!MD5:!PSK:!3DES
-
+{% endif %}
  LogLevel info
  CustomLog logs/access_log combinedrucio
 
- WSGIScriptAlias /auth                   /usr/lib/python2.7/site-packages/rucio/web/rest/authentication.py
```
cserf, line 42: If you remove the authentication endpoint then we need to provide a specific image for this

cserf (approved):

PR #503, opened 2018-01-22
Clients: Protocol failover in download Fix #502

### LINKED ISSUE #502: Automatic protocol failover broken in rucio download
Motivation
----------
In version 1.14.7 `rucio download` always uses the same protocol from an RSE

Modification
------------

Expected behavior
-----------------
`rucio download` should fail over to a different protocol when the first one fails. Fail over used to work until version 1.14.5

nikmagini:
Clients: Protocol failover in download Fix #502

### DIFF
--- a/bin/rucio
+++ b/bin/rucio
@@ -17,6 +17,7 @@
  - Joaquin Bogado <joaquin.bogado@cern.ch>, 2014, 2015
  - Evangelia Liotiri <evangelia.liotiri@cern.ch>, 2015
  - Tobias Wegner <tobias.wegner@cern.ch>, 2017
+ - Nicolo Magini <nicolo.magini@cern.ch>, 2018
 """
 
 import argparse
@@ -1299,7 +1300,6 @@ def _downloader(args, input_queue, output_queue, threadnb, total_threads, trace_
 
             while not success and len(protocols):
                 protocol = protocols.pop()
-                rse['read_protocol'] = protocol
 
                 logger.debug('%s : Trying protocol %s at %s' % (thread_prefix, protocol['scheme'], rse_name))
                 trace['protocol'] = protocol['scheme']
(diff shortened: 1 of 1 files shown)


nikmagini:
Hi all,

a comment: I noticed that there are two ways to choose the protocol for download; either the `force_scheme` argument of rsemanager.download; or through `rse_settings['read_protocol']`

I have used both, but please advise if it's preferred to use one or the other

N.

TWAtGH:
Hi,
ok I think I finally understood how `rse_settings['<operation>_protocol']` is meant to be used.
It shouldnt store a full protocol description. It only stores the priority of the protocol that should be used. DEFAULT_PROTOCOL is 1 (0 means disabled).
It was used completely wrong in the downloader.
Edit: I'm not sure if these operation_protocol fields are usefull at all... The priorities are defined per protocol per domain but the operation_protocol fields don't have the domain information.

bari12:
Yes, to force the protocol really `force_scheme` should be used. Manipulating the rse_settings is not the best way I would think. You could change (As Tobias described) the protocol priority in the rse_settings, but probably it is best to just run through with force_scheme.

TWAtGH:
Actually without this commit you cannot just change the protocol prio in the rse_settings (at least not for download/upload) because these `<operation>_protocol` fields are ignored in `select_protocol`. I wonder if they are used elsewhere.

bari12:
Yes, you are right, it would have to be changed in another part of the rse_settings (under the domains) - I think the `<operation>_protocol` is completely unused and should probably be removed. I will investigate this a bit in the afternoon. Anyway, how Nicolo is doing it now in this patch is fine and works.

bari12:
Yes right, actually I think the `<operation>_protocol` is a dead setting and should not really be used.

mlassnig:
can you deprecate `<operation>_protocol`? just rename the variable `deprecated_<operation>_protocol` and then we see if/what breaks

bari12:
Yes we should. But lets do this in a different patch. I think the operation_protocol is a leftover from times before we had domain-support.

nikmagini:
Ok I'll just use ‘force_scheme‘ and remove the ‘read_protocol‘ manipulation


Il mar 23 gen 2018, 09:34 Martin Barisits <notifications@github.com> ha
scritto:

TWAtGH (approved):

PR #507, opened 2018-01-23
Transfers: Only destination scheme should be used to find compatible source schemes; Fix #73

### LINKED ISSUE #73: Conveyor should only use schemes compatible to the destination scheme
_Migrated from JIRA [RUCIO-2817](https://its.cern.ch/jira/browse/RUCIO-2817)_

bari12:
Transfers: Only destination scheme should be used to find compatible source schemes; Fix #73


vingar (approved):
tests to be added in the future.

PR #517, opened 2018-01-24
Transfers: Conveyor receiver not processing messages; Fix #513

### LINKED ISSUE #513: Conveyor receiver does not consume messages
Motivation
----------
The conveyor receiver does not consume messages at the moment.

Modification
------------
The 'vo' field moved from headers to messages and needs to be checked there. Also the check for `false` on `msg['job_m_replica']` and the check for `true` on `msg['job_m_replica']` needs to be lowered.

bari12:
Transfers: Conveyor receiver not processing messages; Fix #513

### DIFF
--- a/lib/rucio/daemons/conveyor/receiver.py
+++ b/lib/rucio/daemons/conveyor/receiver.py
@@ -9,7 +9,7 @@
 # - Mario Lassnig, <mario.lassnig@cern.ch>, 2013-2015
 # - Vincent Garonne, <vincent.garonne@cern.ch>, 2014
 # - Wen Guan, <wen.guan@cern.ch>, 2014-2016
-# - Martin Barisits, <martin.barisits@cern.ch>, 2017
+# - Martin Barisits, <martin.barisits@cern.ch>, 2017-2018
 
 """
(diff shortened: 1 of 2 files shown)


lib/rucio/daemons/conveyor/receiver.py
```diff
@@ -60,17 +60,21 @@ def on_error(self, headers, message):
     def on_message(self, headers, message):
         record_counter('daemons.conveyor.receiver.message_all')
 
-        if 'vo' not in headers or headers['vo'] != 'atlas':
+        try:
+            msg = json.loads(message)
+        except Exception, e:
+            msg = json.loads(message[:-1])  # Note: I am not sure if this is needed anymore, this was due to an unparsable EOT character
+
+        if 'vo' not in msg or msg['vo'] != 'atlas':
```
cserf: what about using 
vo = rucio.common.policy.get_policy()
?

bari12: Good idea, I will adapt.

mlassnig (approved):

Issue #343, opened 2017-12-13
WebUI: Metalink download for datasets #343


sprenner:
At the moment it is not possible to download entire datasets at once. To make downloading datasets possible in a single step, the WebUI will be extended to create a metalink containing download links for all files contained in a dataset.

PR #541, opened 2018-01-26
WebUI: Creates metalinks for all files in an available dataset #343

sprenner:
WebUI: Creates metalinks for all files in an available dataset #343


mlassnig:
@tbeerman ping

tbeerman (approved):

PR #544, opened 2018-01-28
Clients: Include TTC for a rule in rucio rule-info Fix #530

### LINKED ISSUE #530: Functionality to include TTC for a rule in rucio rule-info
Motivation
----------
Rucio rule-info should display the TTC.

jwackito:
Clients: Include TTC for a rule in rucio rule-info Fix #530


bari12 (changes requested):
Hi JJ, I think you misunderstood what we talked about querying the requests 😄 Right now this would not work as you are trying to query the Requests directly from the client. (And the user does not have access to the DB)
What I meant is that the client does the normal client.get_replication_rule() call to the server, and if asked, the server calculates the TTC and sends it back in the output dictionary to the client. We can chat about this in more detail if you want.

PR #584, opened 2018-02-04
Transfers: Implement pluggable LFN2PFN algoirthms.  Fixes #570

### LINKED ISSUE #570: Pluggable LFN2PFN algorithms
Previously, we introduced a mechanism to specify the algorithm used for LFN2PFN at the RSE.  The options now are `default` and `identity`.  The `default` algorithm creates subdirectories based on hashes of the LFN, while `identity` is the identity function: LFN is used directly.

As new communities come in, each has their own twist on the LFN2PFN mapping.  Currently, there's no good way to add community-specific code for the mapping.  I propose we:

* In the Rucio server configuration, provide a module that is loaded on startup which can register additional plugins (mapping the algorithm name to a Python callable).
* Continue to allow the algorithm to be set per-RSE as an attribute.
* In the Rucio server configuration, specify the default algorithm to use if not specified as an attribute.

So, Rucio could continue to offer one or two stock algorithms -- but as experiments mature, this can become part of their policies.

bbockelm:
This adds the ability for a Rucio server configuration to specify a module that can register additional lfn2pfn algorithm implementations.  The configuration file may also specify the default LFN2PFN algorithm (for example, replacing `hash` with `identity` or a VO-specific algorithm).

The callout algorithms receive a scope, LFN name, RSE name, RSE attributes, and protocol attributes.  It should return the path portion of a PFN (not the full PFN).

The server instance's default can still be overridden by setting the `lfn2pfn_algorithm` attribute on the RSE.


lib/rucio/tests/lfn2pfn_module_test.py
```diff
@@ -0,0 +1,18 @@
+''' Copyright European Organization for Nuclear Research (CERN)
+ Licensed under the Apache License, Version 2.0 (the "License");
+ You may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Authors:
+ - Brian Bockelman, <bbockelm@cse.unl.edu>, 2019
```
mlassnig, line 8: time traveller? :-)

mlassnig (commented):
This looks fine to me. The upgrade path is simply making sure that the default config is set in the rucio.cfg before running a server with this code.

I'm not particularly fond of having fallbacks in the code (like line 38 in protocol.py), because this might go unnoticed by people. As this is a supercrucial thing to the system, it's better to fail early and hard if it's not set explicitly to what the instance should be doing.

Therefore, can you please add the required entries to the rucio.cfg.template file so we don't forget?

vingar:
@mlassnig not sure to understand your point. You want to get a failure if the mapping lfn to pfn is not explicitly set ? 
If yes, I looks fine to be rather explicit than implicit. On the other hand, this will break the backward compatibility and current behavior (hash mapping by default).  In both cases, it needs to be super well documented ;)

bari12:
I think the default = hash should be in the code. The implicit mapping is not the most beautiful thing to do, but it might be a nightmare with clients who get a newer version but stay with the old config, especially if this is in a patch release, which brings me to my second point.

This change should actually be a feature, as it changes something quite core to the system. @bbockelm how soon would you (ligo?) need this code in your evaluation? (The options basically are to do an earlier feature release, or leave it as a patch; But I would prefer the former)

mlassnig:
agree with martin here. i didn't want the line to be removed ;-) just to make have it explicitly added to the rucio.cfg

bbockelm:
As far as the LIGO use case goes - I don't think there's much of a preference either way.  LIGO just has a testing server, so there's not much risk in feature-vs-bugfix.

That said, I don't see much potential for disruption.  There's already a concept of per-RSE LFN2PFN algorithm.  This simply adds:

* Ability to make a different default.
* Ability to load additional algorithms.

where the default behavior is identical to the current release.

(But again, I don't have a strong preference here.)

bari12:
> I don't see much potential for disruption
Famous last words? 😆 

The change is a bit borderline between patch/feature, but if the defaults are there (as they are now) I am comfortable to put it as a patch too.

bbockelm:
Ah - after some testing for LIGO, I hit an important "gotcha": the Rucio client can no longer do the LFN2PFN translation by itself, even for deterministic RSEs.

Unfortunately, I'm not sure this is a difference we can pave over -- since we don't know what the remote server has set as the `default` algorithm, we'll always have to make a query.

Thoughts?

bari12:
Ok, some (not so structured) thoughts on this. This change is obviously not backward compatible, as the old clients cannot upload (download should work) as they do not know the lfn2pfn algorithm. But this is fine.

- I think the fact that the client cannot do the lfn2pfn translation completely independent (due to lack of knowledge of the server default) is also fine. For both upload as well as for download the rse_settings dictionary is loaded from the server, which can include the lfn2pfn algorithm. So the fact that the client can do all this work independently without contacting the server is not true anyway, as it is always loading the rse_settings dictionary.
- I think the client could just go for the `hash` algorithm if no algorithm is mentioned in the rse_settings. Thus, this is the **default**. We just have to make sure that on the server side, if the server has a different default (e.g. identity), that this is returned in the rse_settings if no explicit algorithm is set for the RSE.
- All algorithms should probably be part of the rucio package and not completely independently imported, to make sure that the client knows the algorithm; 
- I think this change should be a feature. (For 1.15.0)

I probably have some more thoughts on this. But this is my first brain dump 😄

bbockelm:
So, here's what I'd propose:

1.  Rucio server always populates the "real, used" algorithm in `rse_settings`.  So, instead of returning `default`, it'll return `hash`.
2.  When the protocol object is created, the client will check whether it understands the algorithm.
3.  If not, it will query the server for the PFN instead of generating it locally.

Thoughts?

vingar:
it looks good to me. 
One technical thing that we have to check is when uploading data and replica location registration the pfns should be provided. This will trigger a server side pfns validation.

bari12:
@bbockelm I think that's good.
@vingar I couldn't find a check which validates the format of the pfn (only some clean surls etc.) - do you remember where this is done in the code?

vingar:
@bari12 see:
lib/rucio/core/replica.py: add_replicas

                raise exception.InvalidPath('One of the PFNs provided does not match the Rucio expected PFN : %s vs %s (%s)' % (str(pfns), str(expected_pfns), str(lfns)))

bari12:
Yes, but that just checks against clean surls with the lfn2pfn function from the protocol, which then would already use the correct algorithm to create the pfns, so I think this is fine. I was more thinking about some kind of regexp somewhere?

bbockelm:
Ah - one challenge here: when the client wants to get a PFN from the server:

https://github.com/rucio/rucio/blob/master/lib/rucio/rse/protocols/protocol.py#L103

it invokes `client.list_replicas`.  This fails in `rucio upload` because, of course, there's no replica registered!

It seems the Rucio API is missing a "translate LFN2PFN" call.

bbockelm:
Ok, with the latest pushes (still waiting for CI to run...), the client will call a new `lfns2pfns` REST API to perform the mapping if it is unable to do the mapping locally.

bbockelm:
Oh - since we decided this goes against `next`, I'm going to close this PR.  Please follow-up in #585.

Issue #587, opened 2018-02-05
Flag to enable/disable the check of busy links in FTS probe


cserf:
Motivation
----------
The FTS probe checks the "busy" links (i.e. links with more than X files submitted). It can put some load on the FTS DB. 

Modification
------------
Introduce a flag to enable/disable the check of busy link in FTS probe

PR #588, opened 2018-02-05
Probe : Flag to enable/disable the check of busy links in FTS probe #587

cserf:
Probe : Flag to enable/disable the check of busy links in FTS probe #587


bari12 (changes requested):
tools/probes/common/check_fts_backlog:45:5: E303 too many blank lines (2)

bari12 (approved):

PR #591, opened 2018-02-05
clients: allow wan,lan,all selection of the network domain

mlassnig:
clients: allow wan,lan,all selection of the network domain

### DIFF
--- a/lib/rucio/core/replica.py
+++ b/lib/rucio/core/replica.py
@@ -739,27 +739,47 @@ def _list_replicas(dataset_clause, file_clause, state_clause, show_pfns, schemes
                     rse_schemes = schemes or []
                     if not rse_schemes:
                         try:
-                            rse_schemes = [rsemgr.select_protocol(rse_settings=rse_info[rse],
-                                                                  operation='read',
-                                                                  domain=domain)['scheme']]
+                            if domain == 'all':
+                                rse_schemes.append(rsemgr.select_protocol(rse_settings=rse_info[rse],
+                                                                          operation='read',
+                                                                          domain='wan')['scheme'])
+                                rse_schemes.append(rsemgr.select_protocol(rse_settings=rse_info[rse],
+                                                                          operation='read',
+                                                                          domain='lan')['scheme'])
+                            else:
+                                rse_schemes.append(rsemgr.select_protocol(rse_settings=rse_info[rse],
+                                                                          operation='read',
+                                                                          domain=domain)['scheme'])
                         except:
                             print format_exc()
 
                     protocols = []
                     for s in rse_schemes:
                         try:
-                            protocols.append(rsemgr.create_protocol(rse_settings=rse_info[rse],
-                                                                    operation='read',
-                                                                    scheme=s,
-                                                                    domain=domain))
+                            if domain == 'all':
+                                protocols.append(('lan', rsemgr.create_protocol(rse_settings=rse_info[rse],
+                                                                                operation='read',
+                                                                                scheme=s,
+                                                                                domain='lan')))
+                                protocols.append(('wan', rsemgr.create_protocol(rse_settings=rse_info[rse],
+                                                                                operation='read',
+                                                                                scheme=s,
+                                                                                domain='wan')))
+                            else:
+                                protocols.append((domain, rsemgr.create_protocol(rse_settings=rse_info[rse],
+                                                                                 operation='read',
+                                                                                 scheme=s,
+                                                                                 domain=domain)))
                         except exception.RSEProtocolNotSupported:
                             pass  # no need to be verbose
                         except:
                             print format_exc()
+
                     tmp_protocols[rse] = protocols
 
                 # get pfns
-                for protocol in tmp_protocols[rse]:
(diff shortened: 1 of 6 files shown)


lib/rucio/core/replica.py
```diff
@@ -739,27 +739,47 @@ def _list_replicas(dataset_clause, file_clause, state_clause, show_pfns, schemes
...
-                            rse_schemes = [rsemgr.select_protocol(rse_settings=rse_info[rse],
-                                                                  operation='read',
-                                                                  domain=domain)['scheme']]
+                            if domain == 'all':
+                                rse_schemes.append(rsemgr.select_protocol(rse_settings=rse_info[rse],
+                                                                          operation='read',
+                                                                          domain='wan')['scheme'])
+                                rse_schemes.append(rsemgr.select_protocol(rse_settings=rse_info[rse],
+                                                                          operation='read',
+                                                                          domain='lan')['scheme'])
```
vingar, line 748: the same protocol can be defined on wan and lan which will can create some duplicates in the rse_schemes list.

lib/rucio/core/replica.py
```diff
@@ -739,27 +739,47 @@ def _list_replicas(dataset_clause, file_clause, state_clause, show_pfns, schemes
...
-                                                                    operation='read',
-                                                                    scheme=s,
-                                                                    domain=domain))
+                            if domain == 'all':
+                                protocols.append(('lan', rsemgr.create_protocol(rse_settings=rse_info[rse],
+                                                                                operation='read',
+                                                                                scheme=s,
+                                                                                domain='lan')))
+                                protocols.append(('wan', rsemgr.create_protocol(rse_settings=rse_info[rse],
+                                                                                operation='read',
```
vingar, line 765: same comments about the duplicate entries.

lib/rucio/core/replica.py
```diff
@@ -816,11 +840,12 @@ def _list_replicas(dataset_clause, file_clause, state_clause, show_pfns, schemes
...
-                    for pfn in pfns:
-                        file['pfns'][pfn] = {'rse': rse,
-                                             'type': str(rse_type),
-                                             'volatile': volatile}
+                    file['rses'][rse] = list(set([tmp_pfn[1] for tmp_pfn in pfns]))  # extract properly the pfn from the (domain, pfn)
+                    for tmp_pfn in pfns:
+                        file['pfns'][tmp_pfn[1]] = {'rse': rse,
+                                                    'type': str(rse_type),
+                                                    'volatile': volatile,
+                                                    'domain': tmp_pfn[0]}  # extract properly the domain from the (domain, pfn)
```
vingar, line 848: hard to know the possible values in tmp_pfn[0] from the review interface. I guess is 'wan', 'lan', 'all' ? 
One alternative is to have a dictionary like {'wan': True, 'lan': True }

lib/rucio/core/replica.py
mlassnig: There's a set operation at the end to filter out the duplicates.

lib/rucio/core/replica.py
mlassnig: it's either wan or lan. all is just an option to the client to retrieve both. an extra dictionary would just make it even more confusing imho.

vingar: a pfn can be on all domains: wan and lan no ?  What will be the tmp_pfn value if the pfn is only on lan, wan and both on wan and lan ?

mlassnig: the domain is just a description of the access. if there's a both "lan" and "wan" access available for a pfn, but it's the same, then it's "wan". there's no need to separate it?

(unless we want to provide a usecase where we have something that is lan-only accessible, but not accessible via wan)

vingar: that's the question :) Does the client need to know that a pfn is accessible locally and then remotely ? e.g. If the clients first tries the lan replica and then fallbacks to a remote one ?

mlassnig: those will be two different pfns in the list_replica reply ;-)

vingar: pfn is the dictionary key thus unique now ?

mlassnig: yes, of course. either you have a lan-accessible PFN, or a lan+wan-accessible PFN. you cannot have a wan-accessible PFN that is not lan-accessible. 

and depending if you request domain lan,wan,all you get back either 1 pfn (tagged with domain:lan, or tagged with domain:wan) or potentially 2 pfns (lan-only pfn: tagged with domain:lan; lan+wan pfn tagged with domain:wan).

default is still to reply with wan-accessible only.

vingar: yes, a wan pfn can be accessible from the lan but it might not reflect the preferences and settings as expressed in Rucio (https for wan but not defined in lan): if the value 'all' should be there or not and if it brings more information.

mlassnig: If sites have LAN preferences then they just need to set them. Those that do have the settings in the RSE configs already.

bari12 (approved):

PR #592, opened 2018-02-05
Clients: Clients: disable certification verification on s3 boto protocol

wguanicedew:
To fix the ssl verification problem on centos7 machines

### DIFF
--- a/lib/rucio/rse/protocols/s3boto.py
+++ b/lib/rucio/rse/protocols/s3boto.py
@@ -176,6 +176,9 @@ class Default(protocol.RSEProtocol):
                                      aws_secret_access_key=secret_key,
                                      is_secure=is_secure,
                                      calling_format=OrdinaryCallingFormat())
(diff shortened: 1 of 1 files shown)


mlassnig:
This completely disables SSL remote server verification. Is this what we want?

bbockelm (changes requested):
Seems to almost definitely be the wrong approach. I don’t think security should be sacrificed even as a convenient work-around.  These things have a way of being embedded and forgotten for years.

wguanicedew:
S3 boto is using S3_ACCESS_KEY and S3_SECRET_KEY to verify connections. It's not using proxy to verify.

bbockelm:
I think it uses `S3_ACCESS_KEY` and `S3_SECRET_KEY` to authorize the connection.  Authentication of the remote server (as well as the integrity and secrecy of the channel) are based on TLS still.

PR #616, opened 2018-02-08
REST: first commit with new flask api and sphinx documentation #586

tbeerman:
REST: first commit with new flask api and sphinx documentation #586
Also Fixes #630

PR #631, opened 2018-02-09
core: multi rest backend support Fix #630

### LINKED ISSUE #630: multiple rest backend support
Motivation
----------
Introduce a backward compatible directory struture to have severl REST backends (webpy, flask, pyramide, django)
```
lib/rucio/web/rest/webpy/v1/__init__.py
lib/rucio/web/rest/flask/v1/__init__.py
```

vingar:


bari12:
I wonder if it shouldn't be the other way around:
Move everything to the webpy/v1 directory and symlink from the old one. By doing this we can remove the deprecated directory at some point, when we change the config. But this probably should be coordinated with the work from @tbeerman now, as there will be merge conflicts otherwise.

vingar:
yes, good point. It should be the other way around. I just changed it. 
In both cases, the deployment is backward compatible and will allow us to start documenting incrementally the REST API with flask (when this PR merged and  #616 rebased).

@tbeermaan ?

tbeerman:
The symlink looks fine to me. The only problem for me is the naming of the folders, at least for the flask one. `webpy` can stay like this because the official module name is just `web`. But for flask this introduces problems. So, we need a different name there. If you look at #616 you can see I used `flaskapi`. Maybe someone has a better idea but just flask will not work.

bari12:
I am closing this as it is included in #616

Issue #641, opened 2018-02-12
better handling NoSectionError in db/sqla/session


vingar:
Motivation
----------
  File "/home/docs/checkouts/readthedocs.org/user_builds/rucio-blueprint/envs/latest/local/lib/python2.7/site-packages/rucio-1.14.8-py2.7.egg/rucio/db/sqla/session.py", line 51, in <module>
    DEFAULT_SCHEMA_NAME = config_get(DATABASE_SECTION, 'schema')
  File "/home/docs/checkouts/readthedocs.org/user_builds/rucio-blueprint/envs/latest/local/lib/python2.7/site-packages/rucio-1.14.8-py2.7.egg/rucio/common/config.py", line 32, in config_get
    return __CONFIG.get(section, option)
  File "/usr/lib/python2.7/ConfigParser.py", line 607, in get
    raise NoSectionError(section)
NoSectionError: No section: 'database'
Modification

PR #642, opened 2018-02-12
core: better handling NoSectionError in db/sqla/session #641

vingar:


bari12:
> lib/rucio/db/sqla/session.py:53:23: F811 redefinition of unused 'NoSectionError' from line 19

bari12 (approved):

Issue #551, opened 2018-01-30
Improve doc for the CLI


cserf:
Motivation
----------
Missing doc for the CLI

Modification
------------
Improve doc for the CLI

nataliaratnikova:
HI, there seems to be a typo  in the example for adding protocols: underscores are used instead of dashes in --web_service_path :  
https://github.com/rucio/rucio/blame/3d9fe6112c64b155734cf9e7c8cb7d9bfedf5b69/bin/rucio-admin#L1442

bbockelm:
Hi Natalia,

Can you send a PR in for that change?

Thanks!

Brian

vingar:
not a big change... Just sent two PRs about it #814 #815

I'm sure @nataliaratnikova will find other inconsistencies and submit new PRs :)

nataliaratnikova:
Hi Brian,
Vincent already corrected this one, I believe.
I noticed that Rucio client is not very consistent about using _ and - 
in the options and names.  It's not always clear to me what the intended 
conventions are, and this particular code is under rapid development. 
May be it is worth going over the whole client code, once it is stable 
and the conventions are clarified.
Thanks,
    Natalia.

vingar:
Hi,
yes, I did that one as the example value was not appropriate as well.  I would be happy if you report more such inconsistency.

Vincent

mlassnig:
The convention is that "-" (minus) should be used in CLI arguments instead of "_" (underscore). Anything else is a typo/mistake.

nataliaratnikova:
Submitted https://github.com/rucio/rucio/pull/818 with two more fixes, 
hopefully all are caught now.

PR #650, opened 2018-02-13
Documentation : Improve doc for the CLI (3) #551

cserf:
Documentation : Improve doc for the CLI (3) #551

### DIFF
--- a/bin/rucio
+++ b/bin/rucio
@@ -359,7 +359,6 @@ def list_file_replicas(args):
 
     table = []
     dids = []
-    rse_dict = {}
     if args.missing and not args.selected_rse:
         print 'Cannot use --missing without specifying a RSE'
         return FAILURE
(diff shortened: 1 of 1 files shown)


bin/rucio
```diff
@@ -358,16 +359,9 @@ def list_file_replicas(args):
 
     table = []
     dids = []
-    rse_dict = {}
     if args.missing and not args.selected_rse:
         print 'Cannot use --missing without specifying a RSE'
         return FAILURE
-    if args.selected_rse and args.list_collections:
```
mlassnig, line 365: are these removals on purpose?

cserf: Yes, the option was deprecated for a long time :
    list_file_replicas_parser.add_argument('--list-collections', dest='list_collections', action='store_true', default=False, help='To have the number of files of the dataset by site\
        (' + Color.BOLD + 'DEPRECATED: ' + Color.END + 'please use list-dataset-replicas instead.)', required=False)

bin/rucio
```diff
@@ -524,6 +498,7 @@ def detach(args):
         cscope, cname = extract_scope(did)
         dids.append({'scope': cscope, 'name': cname})
     client.detach_dids(scope=scope, name=name, dids=dids)
+    print 'DIDs successfully detached from %s:%s' % (scope, name)
```
bari12, line 501: I am not for or against this, but we are not very consistent with these messages in general. For some commands we follow the "bash" mantra that we do not give a message on a successful command. On other we do give a message. I think we should be consistent here.

cserf: The idea here was to be consistent with the attach command:

    client.attach_dids(scope=scope, name=name, dids=dids)
    print 'DIDs successfully attached to %s:%s' % (scope, name)

I agree we need to be consistent on every commands. We will need to work on this after the workshop.

bari12 (approved):

Issue #525, opened 2018-01-25
Include the number of accesses column in DID table to count dataset accesses


bari12:
Motivation
----------
To calculate the number of accesses of a file or dataset, kronos should also update this information.

- [x] Schema change needed to ad `nbaccesses` column
- [x] Changes in Kronos needed to update the column

PR #651, opened 2018-02-13
Core : Include the number of accesses column in DID table to count dataset accesses #525

cserf:
Core : Include the number of accesses column in DID table to count dataset accesses #525


lib/rucio/db/sqla/models.py
```diff
@@ -353,6 +353,7 @@ class DataIdentifier(BASE, ModelBase):
     eol_at = Column(DateTime)
     is_archive = Column(Boolean(name='DIDS_ARCHIVE_CHK'))
     constituent = Column(Boolean(name='DIDS_CONSTITUENT_CHK'))
+    nbaccesses = Column(Integer())
```
mlassnig: just a minor comment, i've rarely seen other projects use "nbsomething", it always seems to be "nr_something".

but i have no particular strong opinion on this

bari12: Actually, to be most consistent with other names in our schema, access_cnt is probably the best name, as we have several other columns named _cnt

mlassnig: 👍

bari12 (approved):

Issue #574, opened 2018-02-01
Do not require space tokens for SRM


bbockelm:
At many CMS sites, space tokens are not used even when SRM is utilized.  However, it's not currently possible to skip this setting.

It may be as simple as setting the space token to the empty string (``), which could be done automatically if not specified.

However, there might be hidden dependencies that want a non-empty-string space token.  Needs investigation.

nataliaratnikova:
Currently Rucio unconditionally passes  spacetoken json parameter to FTS job.  No matter what value is passed, this causes fts error on CMS sites which do not support spacetokens.
The desired behavior is: if space_token is not in the 'extended_attributes' , do not add spacetoken to fts json.

bbockelm:
Lines that need fixing:

* https://github.com/rucio/rucio/blob/master/lib/rucio/transfertool/fts3.py#L171
* https://github.com/rucio/rucio/blob/master/lib/rucio/daemons/conveyor/common.py#L165

nataliaratnikova:
Hi @bbockelm ,   how do I now define an RSE so that it would not pass the spacetoken  to FTS? 
Adding 'srm' protocol still requires a non-empty spacetoken argument : 
`# rucio-admin  rse add-protocol --hostname 'stormfe1.pi.infn.it' --scheme 'srm' --prefix '/cms/store/test/rucio'  --web-service-path '/srm/managerv2?SFN='  --port '8444' --impl 'rucio.rse.protocols.gfalv2.Default' --domain-json '{"wan": {"read": 1, "write": 1, "delete": 1, "third_party_copy": 1}}' --space-token ''  T2_IT_Pisa `
`Error: space-token and web-service-path must be provided for SRM endpoints.`

PR #655, opened 2018-02-13
clients: Make space tokens optional (next branch). Fix #575

### LINKED ISSUE #575: Misleading warning in srm scheme
Motivation
----------
The srm scheme currently requires space_token as  extended_attribute, however when space_token is not set, rucio complains about  extended_attributes and web_service_path not specfied.

Modification
------------
Happens in these lines: 
https://github.com/rucio/rucio/blob/5fa4bb25b6f5b75913773b65217a5fcde6e94f1c/lib/rucio/core/rse.py#L691-L692

Expected behavior
-----------------
Ideally , space_token would be an optional attribute,  and if not used, no spacetoken should be passed to FTS transfer job submission command.    See also  #574

bbockelm:
Outside ATLAS (particularly, for CMS), space tokens are optional at some SRM endpoints.  Previously, for #574, we fixed the conveyor to handle empty space tokens.  Additionally, the fix for the client-side of #575 got mistakenly wrapped up in #570 (pluggable LFN2PFN) and put into `next`.

This is the missing piece for `master`.  As I tweaked the fix, this causes a merge conflict... is it best to wait for `master` to get merged first?  @bari12?


bari12:
The merge conflict I will resolve when the PR is approved.
Although it is almost not worth doing it, as our next release is 1.15.0 thus next (Which already has the space token fix) will be merged into master anyway :)

PR #674, opened 2018-02-15
client: implement uploadclient #671

### LINKED ISSUE #671: Implement uploadclient
Motivation
----------
Implement upload capabilities in uploadclient

Modification
------------

Expected behavior
-----------------

TWAtGH:
Fixes #671

### DIFF
--- a/lib/rucio/client/uploadclient.py
+++ b/lib/rucio/client/uploadclient.py
@@ -22,9 +22,9 @@ import time
 
 from rucio.client.client import Client
 from rucio.common.exception import (RucioException, RSEBlacklisted, DataIdentifierAlreadyExists,
-                                    DataIdentifierNotFound, Duplicate, FileAlreadyExists,
-                                    ServiceUnavailable, ResourceTemporaryUnavailable, InputValidationError)
-from rucio.common.utils import adler32, generate_uuid, md5, execute
+                                    DataIdentifierNotFound, ServiceUnavailable,
+                                    ResourceTemporaryUnavailable, InputValidationError)
+from rucio.common.utils import adler32, execute, generate_uuid, md5, send_trace
 from rucio.rse import rsemanager as rsemgr
 from rucio import version
 
@@ -32,11 +32,15 @@ from rucio import version
 class UploadClient:
     """ This class cover all functionality related to file uploads into Rucio."""
 
-    def __init__(self, _client=None):
-        self.logger = logging.getLogger('uploadclient')
-        logging.basicConfig()
-        self.logger.setLevel(logging.DEBUG)
+    def __init__(self, _client=None, user_agent='rucio_clients', logger=None):
+        if not logger:
+            logger = logging.getLogger('rucio_uploadclient')
+            logger.addHandler(logging.NullHandler())
+
+        self.logger = logger
+        self.user_agent = user_agent
         self.client = _client if _client else Client()
+        print(self.client.host)
         self.account = self.client.account
         self.default_file_scope = 'user.' + self.client.account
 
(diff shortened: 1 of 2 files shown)


mlassnig (approved):

lib/rucio/common/utils.py
```diff
@@ -598,3 +600,31 @@ def make_valid_did(lfn_dict):
     lfn_copy['name'] = lfn_copy.get('name', lfn_copy['filename'])
     del lfn_copy['filename']
     return lfn_copy
+
+
+def send_trace(trace, trace_endpoint, user_agent, retries=5, logger=None, log_prefix=''):
```
vingar, line 605: why not putting that in a trace_clients class and file?

lib/rucio/client/uploadclient.py
```diff
@@ -9,79 +9,333 @@
...
+        replica['name'] = file['did_name']
+        replica['bytes'] = file['bytes']
+        replica['adler32'] = file['adler32']
+        replica['md5'] = file['md5']
+        replica['meta'] = file['meta']
+        replica['state'] = file['state']
+        return replica
+
+    def upload(self, sources_with_settings, summary_file_path=None):
+        """
```
vingar, line 145: wrong docstring format

lib/rucio/client/uploadclient.py
```diff
@@ -9,79 +9,333 @@
...
+                files.append(file)
+            else:
+                logger.warning('No such file or directory: %s' % path)
+
+        if not len(files):
+            raise InputValidationError('No valid input files given')
+
+        return files
+
+    def convert_file_for_api(self, file):
```
vingar, line 133: missing docstring

lib/rucio/client/uploadclient.py
```diff
@@ -9,79 +9,333 @@
...
+        file['md5'] = md5(filepath)
+        file['meta'] = {'guid': self.get_file_guid(file)}
+        file['state'] = 'C'
+        file.setdefault('did_scope', self.default_file_scope)
+        file.setdefault('did_name', file['basename'])
+        file.setdefault('lifetime', None)
+
+        return file
+
+    def collect_and_validate_file_info(self, sources_with_settings):
```
vingar, line 90: missing docstring

lib/rucio/client/uploadclient.py
```diff
@@ -9,79 +9,333 @@
...
-                # ToDo the REST call for: recommendation = rucio_server.recommend_storage(account, sources[src])
-                # ToDo the REST call for: rucio_server.prepare_upload(account, sources[src], recommendation)
+                raise RucioException('Error extracting GUID from ouput of pool_extractFileIdentifier')
+        elif guid:
+            guid = guid.replace('-', '')
+        else:
+            guid = generate_uuid()
+        return guid
+
+    def collect_file_info(self, filepath, settings):
```
vingar, line 73: missing docsting

vingar (dismissed):
incomplete, badly formed and missing docstrings

TWAtGH:
I thought after the discussion in the morning meeting that we will first get it to run with the site mover and then do the beautifying. The downloadclient doesn't have documentation neither so far.

vingar:
Writing decent docstring documentation should be considered just normal not just as 'beautifying', no ? or at least having a second issue referring to this.

One comment is not related to the docstring but having proper tracer client to post traces.

bari12:
The uploadclient should be done properly right away. With the downloadclient we quickly wanted to get it merged so Tomas can test it on the grid before he leaves. But as we won't test the uploadclient on the grid like this we have more time for this.

vingar:
With production and deployment constraints is perfectly fine but then we need to have the corresponding issues to not lose track of the comments/next actions for the next release (and to get better from the past...)

TWAtGH:
@bari12 Please make this sure with Tomas, since I still want to do some changes before commenting and that maybe take another day.
@vingar I think decent docstrings are not really helpful. Then you have something like `x = 1 # set x to 1`. Especially for APIs (and in this case basically there goes in a huge dictionary with all settings and the settings depend in different ways on each other) there is a bit more effort needed to do this properly. But you are right we should keep track of this in an issue.
For the trace client thing I'm not sure. I mean send_traces is really just a very small utility function and I don't think it will get more extended. I would say we keep this in mind and if there needs to be more functionality in the tracing mechanism we put it in an extra class.

vingar:
@TWAtGH About docstring describing the method parameters I have in mind the documentation automatically generated  (e.g. in readthedocs or pydoc) which is usually really good for other developers (discussion, etc). 

For the tracer client, we got recently some questions about the API and respectively the client. Exposing it clearly in one place will be beneficial and we can just point on it the next time we got question.

bbockelm:
Hi,

As a newer collaborator, let me second the request for docstrings / documentation.  It's not just documenting the API parameters (for some things, like file description, you can simply refer to existing definitions of the schema) but also usage examples and mentioning the various aspects of the semantics

I wouldn't have any idea of how to use this class with the current documentation.  I'd have to Google around to find another script that uses it, then take an educated guess.

Brian

TWAtGH:
No question that docstrings are very important. But I thought this code is time critical (it isn't meant to be used in production yet anyway) so that Tomas can use it in the site mover for tests. Then it would have been best to create an issue for the task of creating the docstrings. 
Otherwise when you have to do it fast you get docstrings like here: 
https://github.com/rucio/rucio/blob/master/lib/rucio/client/didclient.py#L132
https://github.com/rucio/rucio/blob/master/lib/rucio/client/didclient.py#L174
I know these because I had to dig through them myself.
But now since Martin said Tomas is only testing with the downloadclient there is no need to rush things and I will do my changes and put in the docstrings :)

bari12:
FYI this is being merged so that @TomasJavurek can send it to prioritised testing.

PR #676, opened 2018-02-15
config: customisable support references; Fix #243

### LINKED ISSUE #243: Removing reference to  https://its.cern.ch/jira/browse/RUCIO in rucio clis

mlassnig:
config: customisable support references; Fix #243

### DIFF
--- a/bin/rucio
+++ b/bin/rucio
@@ -230,9 +230,9 @@ def exception_handler(function):
             else:
                 logger.error(error)
                 logger.debug(traceback.format_exc())
-                logger.error('The object is missing this property: ' + str(error))
-                logger.error('This should never happen. Please rerun the last command with the "-v" option to gather more information.')
-                logger.error('Please followup with all relevant information at: %s' % config_get('policy', 'support_email'))
+                logger.error('\nThe object is missing this property: %s\n'
+                             'This should never happen. Please rerun the last command with the "-v" option to gather more information.\n'
+                             'Please followup with all relevant information at: %s' % (str(error), config_get('policy', 'support')))
             raise
         except RucioException as error:
             logger.error(error)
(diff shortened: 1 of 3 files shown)


bin/rucio
```diff
@@ -229,20 +229,20 @@ def exception_handler(function):
...
                 logger.error(error)
                 logger.debug(traceback.format_exc())
-                logger.error("This means that the object doesn't have the property " + str(error) +
-                             ". This should never happen. Please rerun the last command with the '-v'" +
-                             " option and submit a ticket with all the necessary information at https://its.cern.ch/jira/browse/RUCIO")
-            return FAILURE
+                logger.error('The object is missing this property: ' + str(error))
+                logger.error('This should never happen. Please rerun the last command with the "-v" option to gather more information.')
+                logger.error('Please followup with all relevant information at: %s' % config_get('policy', 'support_email'))
+            raise
```
vingar: just for info, it's recommended to split one string on several lines like that
```
s = ("this is a very"
      "long string too"
      "for sure ..."
     )
```
here:

```
logger.error('The object is missing this property: ' + str(error)
                  'logger.error('This should never happen. Please rerun the last command with the "-v" option to gather more information.')
```

bari12:
Two things come to my mind:
- I think the variables should be called `support_url` and `rucio_url`. Could be both an email, or url, so I would just call it URL.
- "If it is a problem concerning Rucio itself" is maybe not strict enough. Maybe write: "If you are sure it is a porblem concerning Rucio itself" ... :)

vingar:
Martin's comment makes me think that maybe  config_get('policy', 'support_email') can come from the server through the config api rather than the local configuration file.

bari12:
But not being able to connect to the server might be an actual reason that somebody wants to contact the support 😆

vingar:
:)

mlassnig:
job is ok in travis, but github never gets the notification...

https://travis-ci.org/rucio/rucio/builds/342412518

bari12 (approved):

Issue #678, opened 2018-02-15
Improvements in finisher


cserf:
Motivation
----------
Try to improve the finisher speed

Modification
------------
Add more debug statements + remove unneeded sleep

PR #694, opened 2018-02-16
Transfers : Improvements in finisher #678

cserf:
Transfers : Improvements in finisher #678

### DIFF
--- a/lib/rucio/daemons/conveyor/finisher.py
+++ b/lib/rucio/daemons/conveyor/finisher.py
@@ -9,7 +9,7 @@
 # - Wen Guan, <wen.guan@cern.ch>, 2015-2016
 # - Mario Lassnig, <mario.lassnig@cern.ch>, 2015
 # - Vincent Garonne, <vincent.garonne@cern.ch>, 2015-2018
-# - Cedric Serfon, <cedric.serfon@cern.ch>, 2017
+# - Cedric Serfon, <cedric.serfon@cern.ch>, 2017-2018
 # - Martin Barisits, <martin.barisits@cern.ch>, 2017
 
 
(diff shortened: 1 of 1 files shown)


lib/rucio/daemons/conveyor/finisher.py
```diff
@@ -80,24 +80,20 @@ def finisher(once=False, process=0, total_processes=1, thread=0, total_threads=1
...
         start_time = time.time()
         try:
             hb = heartbeat.live(executable, hostname, pid, hb_thread, older_than=3600)
             prepend_str = 'Thread [%i/%i] : ' % (hb['assign_thread'] + 1, hb['nr_threads'])
-
+            logging.debug(prepend_str + 'Starting new cycle')
             if activities is None:
                 activities = [None]
 
-            if sleeping:
```
bari12, line 95: Taking the sleeping completely out is not a good idea for smaller experiments or if there is not that much work to do. This might severely hammer the db.

mlassnig: yep. we cannot take out the sleep, this will kill the DB if there's nothing to do for a thread.

cserf: The sleep was there twice (check line 133)

mlassnig: Argh, you're right. Fixed-time sleep vs diff-time sleep.

mlassnig (approved):

lib/rucio/daemons/conveyor/finisher.py
bari12: Oh yes 🙄

bari12: Ok coming back to this, I don't really like how the sleeping is done with the `time_diff`. Basically what the daemon does is it goes to a sleep if it was quick checking the last batch of requests, but there is no measure taken if there are any other requests left to check. So worst case, it is able to finish the batch quickly and then always goes to a sleep, even if there is still a lot of requests in the queue.
I think the other way is better, only go to a sleep if you got an empty response (Or alternatively not a full batch).

cserf: You are very optimistic how fast the finisher is able to update:
tail -n1000000 /var/log/rucio/conveyor-finisher.log | grep "\[9/15\]" | grep 19362
2018-02-20 14:12:49,119 19362   DEBUG   Thread [9/15] : Updating 1000 requests for activity User Subscriptions
2018-02-20 14:19:46,444 19362   DEBUG   Thread [9/15] : Finish to update 1000 finished requests for activity User Subscriptions
2018-02-20 14:19:49,827 19362   DEBUG   Thread [9/15] : Updating 1000 requests for activity User Subscriptions
2018-02-20 14:27:06,623 19362   DEBUG   Thread [9/15] : Finish to update 1000 finished requests for activity User Subscriptions
2018-02-20 14:27:12,998 19362   DEBUG   Thread [9/15] : Updating 1000 requests for activity User Subscriptions
2018-02-20 14:33:19,440 19362   DEBUG   Thread [9/15] : Finish to update 1000 finished requests for activity User Subscriptions
2018-02-20 14:33:22,980 19362   DEBUG   Thread [9/15] : Updating 1000 requests for activity User Subscriptions
2018-02-20 14:39:13,201 19362   DEBUG   Thread [9/15] : Finish to update 1000 finished requests for activity User Subscriptions
2018-02-20 14:39:16,117 19362   DEBUG   Thread [9/15] : Updating 1000 requests for activity User Subscriptions

i.e. 7 minutes to update 1000 requests :(

OTOH, in the case you sleep when there are no requests, and if for each cycle you have 1 or 2 requests to process that means the daemon will never sleep.

bari12: I think this is still the wrong approach because the operators basically need to scale the sleeping time according to the average time it takes to calculate a batch.

Imho this should only be based on two things. Are there any requests to work on or was a full batch received to work on? If there is only 1 request each time, maybe still sleep because not a full batch was received.

bari12: Coming back to this. We can merge it for now to get the PR done, as this will work. But the point still stands that I don't think that this is the best principle. I checked the other daemons, and we are basically using a number of different methods to pause the daemons, which is not good. We should converge to one principle there.

Issue #586, opened 2018-02-05
Add REST API frontend using flask


tbeerman:
Motivation
----------
Currently, the REST API frontend of Rucio is provided using the web.py framework. This framework is already quite dated and not in development anymore. On the other hand Flask is popular framework that is still actively supported. Overall, the effort to also add a flask frontend is not very high. Additionally, flask provides the possibility to autogenerate sphinx REST API documentation directly from comments in the code.

Modification
------------
Change the structure under `lib/rucio/web/rest` and add two subfolders `webpy` and `flask`. The old content goes into `webpy` and the flask interface will be ported to the other folder. The configuration of aliases in rucio.conf has then to be changed to the new folders.

PR #706, opened 2018-02-19
documentation: added default configuration values #586

vingar:


lib/rucio/common/config.py
```diff
@@ -76,9 +76,23 @@ def config_get_float(section, option):
     return __CONFIG.getfloat(section, option)
 
 
-def config_get_bool(section, option):
-    """Return the boolean value for a given option in a section"""
-    return __CONFIG.getboolean(section, option)
+def config_get_bool(section, option, raise_exception=True, default=None):
```
mlassnig, line 79: i don't get the reason for raise_exception, that should be covered by the default?

lib/rucio/core/transfer_limits.py
```diff
@@ -7,43 +7,29 @@
 #
 # Authors:
 # - Martin Barisits, <martin.barisits@cern.ch>, 2017
+# - Vincent Garonne vgaronne@gmail.com, 2018
```
mlassnig: missing brackets

lib/rucio/common/config.py
vingar: no, it's not covered by the default as the default value is only considered if raise_exception is set and None can be a possible default value. raise_exception has been also introduced to be backward compatible.

mlassnig: no, it's not covered by the default as the default value is only considered if raise_exception is set and None can be a possible default value. raise_exception has been also introduced to be backward compatible.

mlassnig: my browser crashed, somehow my review broke ;-)

ok.

just one thing, we need to check that the default is of the actual datatype (or can be converted), and otherwise throw a TypeError like .getboolean(), .getfloat(), etc. would. otherwise by mistake one could continue with the wrong datatype.

vingar: just added bool conversion if not None to be protected in that case.

mlassnig: thx.

do we want to extend get_int and get_float in the same way?

vingar: maybe later, it's not anymore a blocker for readthedocs now

mlassnig (approved):

PR #714, opened 2018-02-19
documentation: client exception and error codes

mlassnig:
documentation: client exception and error codes


vingar (approved):

bari12:
Just on a side-note: Please put the ticket number in the commit message. It's much easier to navigate back to the issue this way.

Issue #534, opened 2018-01-25
Consistent headers for all source files


bari12:
Motivation
----------
Docstrings and Copyright.

https://github.com/PyCQA/pylint/issues/1164
https://stackoverflow.com/questions/7877522/how-do-i-disable-missing-docstring-warnings-at-a-file-level-in-pylint

vingar:
The docstring headers are not handled properly by sphinx for the API documentation. The clients source files will be updated to:
```
# Copyright (c) 20XX-2018 CERN for the benefit of the ATLAS collaboration.
#
# Licensed under the Apache License, Version 2.0 (the "License");
# You may not use this file except in compliance with the License.
# You may obtain a copy of the License at
# http://www.apache.org/licenses/LICENSE-2.0
#
# Authors:
# - ...
```

mlassnig:
Back to block comments ;-)

Correct dates according to git are 2012-2018. I'd also modify the first to lines to be a bit more explicit (no need for (c) for example)


```
 # Copyright 2012-2018 European Organization for Nuclear Research (CERN)
 # For the benefit of the ATLAS Collaboration
 # 

```...

(also makes it nicer to modify the second line later on, if necessary)

vingar:
yes, back to block comments and we can hope that pylint will offer a way to mute the missing docstring at the top of the file

I was thinking having the starting date based on the file creation date.

According to http://hepsoftwarefoundation.org/notes/HSF-TN-2016-01.pdf
The format must be:

(c) Copyright [year] CERN [for the benefit of the [Name of appropriate
group] Collaboration]

mlassnig:
Right, then that's how it is ;-)

vingar:
this one is not true everywhere ...

bari12:
Ah I thought with your recent patches you iterated all files.

vingar:
only the ones I've modified to get the doc generation working

PR #728, opened 2018-02-20
Release management: utility script to generate headers with various options (in place, etc) #534

vingar:


tools/add_header
```diff
@@ -0,0 +1,151 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Authors:
+'''
+        for author in authors:
+            if int(authors[author]['min']) == int(authors[author]['max']):
+                header += '# - %(name)s, %(mail)s, %(min)s\n' % authors[author]
```
mlassnig: i propose the RFC-822 style with brackets for emails

 %(name)s <%(mail)s>, %(min)s\n

bari12: The list of mails is already filled with the brackets. So its fine.

bari12: Line 51

mlassnig: right, but the comma after the name needs to go

vingar: The comma after the name is historically present in almost all files.
Example: lib/rucio/common/utils.py

mlassnig: yep, but it should go ;-)

vingar: let it go

mlassnig (approved):

PR #756, opened 2018-02-23
clients: Update dependencies  Fix #754

### LINKED ISSUE #754: client dependencies update

vingar:


bari12:
Now the magic question, are these python 2.6 compatible ;)

vingar:
The usual problematic dependencies have been made optional, but as I don't have py26 around I'm not 100% sure. To be certain, you can check them by doing `pip install -r <requirements file>` with py26.

hugovk:
Is Python 2.6 still needed? It's 4 years past its EOL and no longer receives security updates.

It's also little used.

Here's the pip installs for rucio from PyPI for last month:

| python_version | percent | download_count |
| -------------- | ------: | -------------: |
| 2.7            |  95.92% |             94 |
| 3.5            |   3.06% |              3 |
| 2.6            |   1.02% |              1 |

Source: `pypinfo --start-date -55 --end-date -25 --percent --pip --markdown rucio pyversion`

ghost:
I've blocked the hugovk spam bot.

tools/pip-requires-client
```diff
@@ -1,16 +1,16 @@
...
+argcomplete>=1.9.4          # Bash tab completion for argparse
 requests==2.18.4            # Python HTTP for Humans.
 urllib3==1.22               # HTTP library with thread-safe connection pooling, file post, etc.
 dogpile.cache>=0.6.4        # Caching API plugins
 nose>=1.3.7                 # For rucio test-server
 boto>=2.48.0                # S3 boto protocol
-python-swiftclient>=3.4.0   # OpenStack Object Storage API Client Library
-tabulate>=0.8.1             # Pretty-print tabular data
-progressbar2>=3.34.3        # Text progress bar
+python-swiftclient>=3.5.0   # OpenStack Object Storage API Client Library
```
bari12:
```
Downloading/unpacking python-swiftclient>=3.5.0 (from -r tools/pip-requires-client (line 10))
  Downloading python-swiftclient-3.5.0.tar.gz (182kB): 182kB downloaded
  Running setup.py egg_info for package python-swiftclient
    Sorry, Python < 2.7 is not supported for python-swiftclient>=3.0
    Complete output from command python setup.py egg_info:
    Sorry, Python < 2.7 is not supported for python-swiftclient>=3.0

```

bari12 (changes requested):

tools/pip-requires-client
bari12: However, I get the same one with 3.4.0 which we used before?

bari12: I only get this running with `python-swiftclient<3.0.0`
I find this a bit strange though because I am quite sure that in the tarball we create it had 3.4.0. So I can only assume that technically, the require python 2.7 but that the files are actually still 2.6 compatible.

vingar: yes, it's just an update but for now we can move them to extra dependencies and drop it. Is the new version fine ?

vingar: but keep in mind that swift is not used at all for now in the rse protocol classes

bari12: Yes I think using the EXTRAS mechanism is best, since we are not using swift in protocols anyway.

bari12: And like this it's clean, since you can also install it (in theory) from a python 2.6 pip.

bari12 (approved):

Issue #665, opened 2018-02-15
webui: add tracing for webui downloads


mlassnig:
Motivation
----------

We want to able to monitor also downloads done directly from the WebUI.

Modification
------------

Send trace when user clicks on download link.

PR #777, opened 2018-02-26
'WebUI: adds tracing for webui file downloads #665

sprenner:
'WebUI: adds tracing for webui file downloads #665

### DIFF
--- a/lib/rucio/web/ui/static/did.js
+++ b/lib/rucio/web/ui/static/did.js
@@ -189,7 +189,7 @@ download_file = function(link, file_scope, filename, rse, filesize, dataset_scop
     var options = {
         eventVersion: 'webui_' + get_version(),
         protocol: 'davs',
-        uuid: uuid.replace(/-/g,''),
+        uuid: uuid,
         datasetScope: dataset_scope,
         eventType: 'download',
         remoteSite: rse,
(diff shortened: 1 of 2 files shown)


lib/rucio/web/ui/static/did.js
```diff
@@ -181,6 +181,63 @@ load_replicas = function(scope, name) {
...
+    document.body.removeChild(element);
+};
+
+uuid4 = function () {
+    var uuid = '';
+    for (var i = 0; i < 32; i ++) {
+        switch (i) {
+            case 8:
+            case 20:
+                uuid += '-';
```
tbeerman: As you are removing the dashes later on anyway, why not just omitting them here in the first place?

lib/rucio/web/ui/static/rucio.js
```diff
@@ -1391,6 +1392,43 @@ RucioClient.prototype.delete_rse_attribute = function(options) {
     });
 };
 
+/* send trace */
+RucioClient.prototype.send_trace = function(options) {
+    check_token();
```
tbeerman: You don't need the token for the traces, so you don't have to check it here.

lib/rucio/web/ui/static/rucio.js
```diff
@@ -1391,6 +1392,43 @@ RucioClient.prototype.delete_rse_attribute = function(options) {
...
+    var url = this.url + '/traces/';
+    if (options.async == null) { options.async = true; }
+    jQuery.ajax({
+        url: url,
+        async: options.async,
+        crossDomain: true,
+        headers: this.headers,
```
tbeerman: See above. The token and account are not needed, so you don't have to send them in the header.

tbeerman (changes requested):

lib/rucio/web/ui/static/did.js
sprenner: To have a function that meets the RFC4122 requirements - but I can omit them here already if you think thats better :)

tbeerman: OK I see your point. But here it's not a utility function which will be used anywhere else but rather just there to create the uuid for the traces, which are without dashes.

tbeerman (approved):

PR #809, opened 2018-03-05
Documentation: Fixed rse_expressions filename in RSE overview Fix #808

### LINKED ISSUE #808: RSE_Expressions link broken in documentation
Motivation
----------
Following the [RSE Expressions](http://rucio.readthedocs.io/RSE_Expressions.html) on http://rucio.readthedocs.io/overview_Rucio_Storage_Element.html leads to a non-existing page.

Modification
------------
Either the link in https://github.com/rucio/rucio/blob/master/doc/source/overview_Rucio_Storage_Element.rst needs to be made lower-case or the file https://github.com/rucio/rucio/blob/master/doc/source/rse_expressions.rst needs to be renamed to `RSE_Expressions`.

kreczko:
Fixing issue #808.

I mostly wanted to try out the tools for making patches, but `tools/submit-pull-request` is not working for me (not Python 3 compatible).
A patch for that is available [here](https://gist.github.com/kreczko/57cfa4752aa47143e7cdeb49d2007ffd) if you are interested, but it is also missing the `commands` module so I left it at replacing `print`.

### DIFF
--- a/doc/source/overview_Rucio_Storage_Element.rst
+++ b/doc/source/overview_Rucio_Storage_Element.rst
@@ -16,4 +16,4 @@ Rucio Cache RSE
 A cache is storage service which keeps additional copies of files to reduce response time and bandwidth usage. In Rucio, a cache is an RSE, tagged as volatile. The control of the cache content is usually handled by an external process or applications (e.g. the Workflow management systems) and not by Rucio. Thus, as Rucio doesn’t control all file movements on these RSEs, the application populating the cache must register and unregister these file replicas in Rucio. The information about replica location on volatile RSEs can have a lifetime. Replicas registered on volatile RSEs are excluded from the Rucio replica management system (replication rules, quota, replication locks) described in the section Replica management. Explicit transfer requests can be made to Rucio in order to populate the cache.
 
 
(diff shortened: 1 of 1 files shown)


mlassnig:
Hi Luke, thanks for your patch!

I can confirm that you py3 patch works for our production infrastructure (py 2.6.6), so I would suggest you "git commit --amend" your patch and have it included so everyone can benefit.

bari12:
Since this is a patch, please also open a second Pull Request against the next branch. 😄

kreczko:
@bari12 Thanks, will do.

bari12:
Hi @kreczko, The PR against `master` was correct, what I meant is create a second PR against `next`.  I know the fact that we have two separate PR (which are the same) against two branches is a bit confusing. https://github.com/rucio/rucio/blob/master/CONTRIBUTING.rst explains a little bit why we do that; Essentially we want to expose problems at the earliest possible points in the development chain, that's why we make PR against both branches.

kreczko:
@bari12 Thanks, I should have read more carefully :).

bari12 (approved):

PR #810, opened 2018-03-05
patch: replacing print with print_function in tools/submit-pull-request

kreczko:
As mentioned in #809, in order to make `tools/submit-pull-request` work with Python 2 & 3, changed `print` to the print function.

### DIFF
--- a/tools/submit-pull-request
+++ b/tools/submit-pull-request
@@ -8,7 +8,8 @@
 #
 # Authors:
 # - Martin Barisits, <martin.barisits@cern.ch>, 2017
-
+# - Luke Kreczko, <kreczko@cern.ch>, 2018
+from __future__ import print_function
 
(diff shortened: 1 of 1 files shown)


tools/submit-pull-request
```diff
@@ -8,7 +8,7 @@
 #
 # Authors:
 # - Martin Barisits, <martin.barisits@cern.ch>, 2017
```
bari12, line 10: Don't forget to add your name 😄

kreczko: For a small patch like this? If you say so :).

kreczko: BTW: If you enable Zenodo integration, then the author list will be compiled for you from the commits.

bari12 (approved):

PR #816, opened 2018-03-06
Core: add access_cnt update to touch_dids #525

tbeerman:
Core: add access_cnt update to touch_dids #525

### DIFF
--- a/lib/rucio/core/did.py
+++ b/lib/rucio/core/did.py
@@ -13,7 +13,7 @@
   - Cedric Serfon, <cedric.serfon@cern.ch>, 2013-2015, 2017
   - Martin Barisits, <martin.barisits@cern.ch>, 2013-2017
   - Ralph Vigne, <ralph.vigne@cern.ch>, 2013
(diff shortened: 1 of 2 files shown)


lib/rucio/core/did.py
```diff
@@ -1452,8 +1465,12 @@ def touch_dids(dids, session=None):
     now = datetime.utcnow()
     try:
         for did in dids:
-            session.query(models.DataIdentifier).filter_by(scope=did['scope'], name=did['name'], did_type=did['type']).\
-                update({'accessed_at': did.get('accessed_at') or now}, synchronize_session=False)
+            row = session.query(models.DataIdentifier).filter_by(scope=did['scope'], name=did['name'], did_type=did['type']).one()
```
bari12: I think a conditional update would be better here. Something like https://github.com/rucio/rucio/blob/master/lib/rucio/core/replica.py#L1467
In that case you don't need to read, calculate, update but can just du an update with access_cnt+1

mlassnig (approved):

Issue #819, opened 2018-03-06
Client migration to Python3


jwackito:
Motivation
----------
We need to make Rucio clients Python3 compatible

Modification
------------
Proposed incremental scenario (each step should be one or several pull requests):

- [x] Enable 2to3 in travis to report py3k incompatibility
- [x] change print by print()
- [x]  change xrange -> range
- [x] Fix dict.keys()/dict.values() to list()
- [x] Fixed  exception to be py3K compatible
- [x] compatible imports
- [x] float, Long, etc
- [x] Unicode vs byte string
- [x] having a py3K packaging +ping test in the test suite for the clients once the migration is complete (e.g. tox)
- [x] Check compatibility with different Python 3 versions; Currently works with 3.5 but not with 3.6

others maybe, python3 interpreter will tell...

bari12:
There is a general ticket tracking the py3k migration.
https://github.com/rucio/rucio/issues/67
You can use this for general tracking.

jwackito:
I should close this issue then?

kreczko:
BTW: I found that the `six` python module makes these kinds of things easier during the transition period:
https://pythonhosted.org/six/#module-six.moves

It is also easier to find them later (searching for `six`) whenever the project is ready to go Python 3 only.

cserf:
I think all the client files are now py3k compatible. @hahahannes can you confirm ?

hahahannes:
The last ones missing are the didclient, fileclient, dq2client and the downloadclient. According to pylint there are just small changes needed. I can fix them.

PR #824, opened 2018-03-07
Clients: print to print() migration #819

jwackito:
Clients: print to print() migration #819


kreczko:
If you want this code to work the same as before in Python 2.X, you will need to add `from __future__ import print_function` in all changed files.

jwackito:
You are right :+1:

bari12 (approved):

bari12 (approved):

PR #851, opened 2018-03-08
Transfer: Protocol mismatch error in the conveyor + some py3k fixes: Close #849

### LINKED ISSUE #849: Protocol mismatch error in the conveyor
Motivation
----------
Sometimes transfers end-up in protocol mismatch error despite there are matching schemes

Modification
------------
Fix it

cserf:
Transfer: Protocol mismatch error in the conveyor + some py3k fixes: Close #849

### DIFF
--- a/lib/rucio/core/transfer.py
+++ b/lib/rucio/core/transfer.py
@@ -664,13 +664,6 @@ def get_transfer_requests_and_source_replicas(process=None, total_processes=None
                 if retry_other_fts:
                     external_host = fts_list[retry_count % len(fts_list)]
 
-                if id in reqs_no_source:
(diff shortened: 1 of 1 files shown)


lib/rucio/core/transfer.py
```diff
@@ -824,6 +827,14 @@ def get_transfer_requests_and_source_replicas(process=None, total_processes=None
             logging.critical("Exception happened when trying to get transfer for request %s: %s" % (id, traceback.format_exc()))
             break
 
+    for id in transfers:
```
bari12, line 823: With this block we can remove the one above completely?

bari12 (approved):

PR #866, opened 2018-03-12
Clients: upload always valid if validate_checksum is False Fix #847

### LINKED ISSUE #847: Rucio Upload Should Respect verify_checksum Flag
Motivation
----------
When uploading to `CERN-EXTENSION_GRIDDISK` the Rucio client fails because it attempts to verify the checksum. At the moment the `verify_checksum` flag is set for the RSE and not ask the storage for the checksum.

The suggested modification will then also skip the verification of the file size on storage. The stat call checking the file size should be OK, and in this scenario is even more important to be verified (in [rsemgr](https://github.com/rucio/rucio/blob/e708f55772d74c782d36f79cfda521570190111a/lib/rucio/rse/rsemanager.py#L398)).

I loathe to put this modification at the protocol level hence the suggested modification, please let me know if you have better ideas.

Modification
------------
- [x] copy `verify_checksum` RSE attribute to RSE info
- [x] get `verify_checksum` from RSE info in `rsemgr.upload`
- [x] always declare uploaded file as valid if `ignore_checksum` is True.

**Some notes**
On line [405](https://github.com/rucio/rucio/blob/e708f55772d74c782d36f79cfda521570190111a/lib/rucio/rse/rsemanager.py#L405) of the rsemgr add
```python
valid = True if ignore_checksum is True else False
```
After discussion with Martin we decided that `verify_checksum` should conceptually be RSE info and to copy it analagously to the `lfn2pfn_algorithm` attribute/info.

berghaus:
Clients: upload always valid if validate_checksum is False Fix #847

### DIFF
--- a/lib/rucio/core/rse.py
+++ b/lib/rucio/core/rse.py
@@ -14,6 +14,7 @@
 # - Thomas Beermann, <thomas.beermann@cern.ch>, 2014, 2017
 # - Wen Guan, <wen.guan@cern.ch>, 2015-2016
 # - Brian Bockelman, <bbockelm@cse.unl.edu>, 2018
+# - Frank Berghaus, <frank.berghaus@cern.ch>, 2018
 
 from re import match
 from StringIO import StringIO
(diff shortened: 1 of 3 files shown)


mlassnig (approved):
looks good!

PR #873, opened 2018-03-13
clients: 3kpythonization of the adler32 function

vingar:


mlassnig (approved):
confirmed correct on overflowing 4GB file

vingar:
good. I did some comparison with xrdadler32  but on small files

PR #895, opened 2018-03-13
Core: Workaround for importlib incompatibility with python 2.6; Closes #875

### LINKED ISSUE #875: importlib not python 2.6 compatible
Motivation
----------
For the dynamic lfn2pfn function the module `importlib` is used. However, importlib is not part of python 2.6, thus it breaks the clients used on the ATLAS local root base. One option would be to use the backport of importlib for python 2.6 (on pypi) - needs to be investigated.

https://pypi.python.org/pypi/importlib

* Adding importlib to the dependencies works for python 2.6 and 2.7, but a install on python 3.x fails like this. 

For now we move the import to a conditional part of the code, as no module gets dynamically loaded at the moment anyway; This needs to be addressed later on with the dynamic loading of permissions and policies, as this should be handled in a similar way.

bari12:
Core: Workaround for importlib incompatibility with python 2.6; Closes #875

### DIFF
--- a/lib/rucio/rse/protocols/protocol.py
+++ b/lib/rucio/rse/protocols/protocol.py
@@ -13,13 +13,13 @@ Authors:
 - Cheng-Hsi Chao, <cheng-hsi.chao@cern.ch>, 2014
 - Mario Lassnig, <mario.lassnig@cern.ch>, 2017
 - Brian Bockelman, <bbockelm@cse.unl.edu>, 2018
+- Martin Barisits, <martin.barisits@cern.ch>, 2018
 
 This module defines the base class for implementing a transfer protocol,
 along with some of the default methods for LFN2PFN translations.
 """
 
 import hashlib
-import importlib
 
 from urlparse import urlparse
 from ConfigParser import NoOptionError, NoSectionError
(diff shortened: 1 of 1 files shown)


lib/rucio/rse/protocols/protocol.py
```diff
@@ -143,7 +143,13 @@ def _module_init_(cls):
         except (NoOptionError, NoSectionError):
             pass
         if policy_module:
-            importlib.import_module(policy_module)
+            # TODO: The import of importlib is done like this due to a depdency issue with python 2.6 and incompatibility of the module with py3.x
```
vingar: depdency -> dependency

lib/rucio/rse/protocols/protocol.py
```diff
@@ -143,7 +143,13 @@ def _module_init_(cls):
...
+            # More information https://github.com/rucio/rucio/issues/875
+            try:
+                import importlib
+                importlib.import_module(policy_module)
+            except:
+                pass
```
vingar: should it fail silently or with:
`raise exception.MissingDependency('Missing dependency : importlib')` ?

lib/rucio/rse/protocols/protocol.py
```diff
@@ -143,7 +143,13 @@ def _module_init_(cls):
         except (NoOptionError, NoSectionError):
             pass
         if policy_module:
-            importlib.import_module(policy_module)
+            # TODO: The import of importlib is done like this due to a depdency issue with python 2.6 and incompatibility of the module with py3.x
+            # More information https://github.com/rucio/rucio/issues/875
+            try:
+                import importlib
+                importlib.import_module(policy_module)
+            except:
```
vingar: except ImportError:

lib/rucio/rse/protocols/protocol.py
bari12: Right, it should just raise it through...

vingar (approved):

PR #903, opened 2018-03-13
tests: Attempt to speed up tests by --processes=2

vingar:


mlassnig:
On my dev-node I got an improvement from 362 to 361 seconds ;-)

bari12:
On Travis:  Ran for 40 min 51 sec 😄

vingar:
The parameter was at the wrong place and now it fails. From the errors, it looks like apache is configured to have only one session and all tests are waiting for it.  
@tbeerman, Is it correct ?

bari12:
I think this can be closed? The changes to the AUTHORS file should be done in a separate PR.

PR #924, opened 2018-03-15
replicas: automatic selection of wan or lan door; Fix #668

### LINKED ISSUE #668: Sorting of replicas in list_replicas
Motivation
----------
The `list_replicas` method orders the replicas, based on the client location, on the server side, but then puts them into a dictionary, hence the order of the replicas is lost. For each replica the dictionary should be expanded with an additional value (e.g. order) which represents the sorting order. This can be used by the client to order the replicas.

mlassnig:
replicas: automatic selection of wan or lan door; Fix #668

### DIFF
--- a/lib/rucio/core/replica.py
+++ b/lib/rucio/core/replica.py
@@ -1,21 +1,26 @@
-"""
-  Copyright European Organization for Nuclear Research (CERN)
-
-  Licensed under the Apache License, Version 2.0 (the "License");
-  You may not use this file except in compliance with the License.
-  You may obtain a copy of the License at
-  http://www.apache.org/licenses/LICENSE-2.0
-
-  Authors:
-  - Vincent Garonne, <vincent.garonne@cern.ch>, 2013-2017
-  - Cedric Serfon <cedric.serfon@cern.ch>, 2013-2018
-  - Ralph Vigne <ralph.vigne@cern.ch>, 2013-2014
-  - Martin Barisits <martin.barisits@cern.ch>, 2013-2016
-  - Mario Lassnig <mario.lassnig@cern.ch>, 2014-2018
-  - David Cameron <d.g.cameron@gmail.com>, 2014
-  - Thomas Beermann <thomas.beermann@cern.ch>, 2014-2016
-  - Wen Guan, <wen.guan@cern.ch>, 2014-2015
-"""
+# Copyright 2013-2018 CERN for the benefit of the ATLAS collaboration.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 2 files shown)


lib/rucio/core/replica.py
```diff
@@ -734,6 +739,13 @@ def _list_replicas(dataset_clause, file_clause, state_clause, show_pfns, schemes
                 if rse not in rse_info:
                     rse_info[rse] = rsemgr.get_rse_info(rse, session=session)
 
+                if domain is None:
+                    domain = 'wan'
+                    if client_location and 'site' in client_location and client_location['site']:
+                        site = get_rse_attribute('site', rse_info[rse]['id'], session=session)
+                        if site and site[0].lower() == client_location['site'].lower():
+                            domain = 'lan'
+
```
vingar: it would be better (and more powerful) to consider client_location as rse expression and check if the site is part of the list of rses with a simple presence test inside the loop. The list of rses must be resolved from the rse expression/client location outside the main loop. This will save O(replicas) calls to get_rse_attribute and the db.

vingar (changes requested):

lib/rucio/core/replica.py
mlassnig: I don't see how a test for presence replaces an equality check, especially if the RSE changes throughout the loop.

But anyway, this is micro optimisation and not important for the use case that we need to solve right now. We can improve this later and do it correctly: elevate "site" as a top-level RSE attribute and have it returned by get_rse_info.

vingar: With this approach, the call to get_rse_attribute is done for each replica and list_replicas is one of the most frequent call. Not sure there is any caching for `get_rse_attribute`actually.

Another alternative is to get the list of rses with `get_rses_with_attribute_value` before the loop and then check if the replica rse is in the list in the loop.

bari12: I am not sure how `parse_expression` compares execution timewise against eg. 3x `get_rse_attribute`. But the parse_expression result is cached, so it might be good to do it like this.

mlassnig: Ok, I see what you mean now. The ``get_rses_with_attribute_value`` is not so good, because it has a potential index range scan, I want to avoid this as much as possible.

Then it boils down to what is faster, single call to ``parse_expression`` (complex but memcached) or multiple calls to ``get_rse_attribute`` (unique id lookup).

vingar: good comment. I remember that some files have sometimes more than 100 replicas, but one should look at the statistics. For the panda jobs (and then the pilot), we usually have more than 3 replicas per file. It was 1.6 in avg but largely biased by user files.

The comparison will  be X calls to `get_rses_with_attribute_value` (with X the number of replicas) vs one call to `get_rses_with_attribute_value` or `parse_expression` if `client_location` is provided. The last approach is independent from the number of replicas.

bari12 (approved):

vingar (approved):

Issue #912, opened 2018-03-14
Add release notes to repository


bari12:
Motivation
----------
Add the release notes to the repository, so that they are available on the documentation.

tushar-97:
@bari12 I would like to take this up if possible. Where exactly do the release notes have to be added?

vingar:
Part of the work is to replicate the changelog from here https://github.com/rucio/rucio/releases into individual rst files. I guess you can help on this but first @bari12 has to define the directory and file name conventions.

bari12:
Hi, I have already started with this and prepared a structure and kind of template. (PR for that will come today) - If you would like to you could then fill all the missing release notes. I would be very thankful for that 😄

tushar-97:
Sure @bari12  I will wait for your PR then to get started.

bari12:
The PR is added now. Let's wait a bit of some people have comments to the general structure etc. Once it is merged I think you can go through the github releases and generate the releases notes similar to the 3 I already created. (You can use the `tools/generate-release-notes <release> --rst` script for the rst output) - Thanks

tushar-97:
When I run the script, it asks for a Github token file. How am I supposed to get that? From what I understand it is needed for authentication to use the Github API.

mlassnig:
Go here and create a new one, and put it in the .githubtoken file ;-)

https://github.com/settings/tokens

tushar-97:
Thanks @mlassnig it's working now.
@everyone If I understand the task correctly, I am to create rst files for ALL releases on https://github.com/rucio/rucio/releases? If so I should also club clients & webui releases in the same file. And do I create files for each milestone number or separate files for hotfix releases as well?

bari12:
Yes essentially create one file for each release (including webui-only, client-only and hotfix releases). If you want you can create a subset of the files first and submit the PR so we can comment them; so you won't have to change all your files if there are some repeating issues. 😄

tushar-97:
@bari12 @mlassnig @vingar I have added a few files via PR #946. Is the formatting correct? I wrote a small script to create these files. Also I am unable to generate release notes for tags of type *.post1. How do I proceed? Also the release-notes tool is not returning anything for releases like 1.14.10. Do I skip these or enter them manually?

bari12:
Hi @tushar-97 The formatting is fine 👍 
Let me check what's broken with the release notes script.

bari12:
[...]

PR #926, opened 2018-03-15
Documentation: Adding release notes to repo #912

bari12:
Documentation: Adding release notes to repo #912

### DIFF
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -33,6 +33,7 @@ For documentation specific to the any of these three, please see the subsequent 
     concepts
     replica_workflow
     architecture
+    releasenotes
 
 Developer Documentation
 =======================
--- a/doc/source/releasenotes/1.14.0.rst
+++ b/doc/source/releasenotes/1.14.0.rst
@@ -1,0 +1,81 @@
+====================
+1.14.0 "Professor D"
+====================
+
+-------
+General
+-------
+
+********
+Features
+********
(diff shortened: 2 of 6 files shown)


doc/source/index.rst
```diff
@@ -73,7 +74,7 @@ Operator Documentation
     installing_server
     installing_daemons
     man/daemons
-    monitoring
+    monitoring
```
mlassnig: extra spaces ^^

doc/source/releasenotes/1.14.0.rst
```diff
@@ -0,0 +1,81 @@
+====================
+1.14.0 "Professor D"
+====================
+
+-----
+Rucio
```
mlassnig: Rucio Server


?
(as below we have clients)

bari12: This is how we do it since a long time. These are the issues from the main package (1.15.0 instead of 1.15.0-clients or 1.15.0-webui) - The thing is server is not really correct, as it is also daemons and very often basically affects everything, also clients, thus calling it Server/Daemons is also not really correct. In the beginning I was thinking considering calling it core, but this is confusing so I just called it purely Rucio.

mlassnig: "Main changes" ?

vingar (approved):

PR #927, opened 2018-03-15
Documentation: Adding release notes to repo #912

bari12:
Documentation: Adding release notes to repo #912

### DIFF
--- a/tools/generate-release-notes
+++ b/tools/generate-release-notes
@@ -16,6 +16,7 @@
 # Authors:
 # - Martin Barisits, <martin.barisits@cern.ch>, 2017-2018
 
+from __future__ import print_function
 
 import commands
 import sys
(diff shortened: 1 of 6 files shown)


tools/generate-release-notes
```diff
@@ -99,17 +109,40 @@ enhancements = sorted(enhancements, key=lambda k: "%s %d" % (k['component'], k['
...
-        print format_issue(issue)
-    print ''
-if bugs:
-    print '## Bugs\n'
-    for issue in bugs:
-        print format_issue(issue)
+if output_type == 'github':
+    print 'a'
+    if features:
+        print '## Features\n'
```
vingar: print() to be py3 compatible but why not writing into the release note file directly  ?

bari12: Because releases notes hold the text from multiple milestones. Thus, the best way to do this is just piping them into the file in bash ;)

vingar: then write the print for py3...

bari12: I really don't think we should focus on that. I mean I can change it now, but this is really not the point. We defined our strategy as keeping the client 3.x compatible and the server/daemons 2.7. The tools, especially this one which is only used by one or two people, are not part of the group of modules we want to make 3.x compatible. So pointing this out for every single PR now is not helpful. If somebody wants to go the step for non-client code I think it is fine, but we should refrain from pushing everyone doing that, as it is not part of the strategy we agreed on.

vingar: It makes sense that we discipline ourselves to write py3 compatible code everywhere, no ?

bari12: Yes, but as we just discussed in the morning, besides the clients it's not a requirement now so I think we should not focus on this. When people get more used all the code will be written in py3.x anyway, so the transition will happen natural. But if we point out the py3 incompatibilities for every PR now we are just holding ourselves up I feel because we essentially treat a non-requirement as a requirement then. - eventually this requirement will come anyway, but I don't think we should make it a focus now.

vingar: yes, I agree. On the other hand, just knowing the right syntax now and putting it can avoid or reduce some painful migration in the future

PR #933, opened 2018-03-15
Transfer: Move conveyor transfers to third_party_copy_operation: Close #535

### LINKED ISSUE #535: Move conveyor transfers to third_party_copy_operation
Motivation
----------
At the moment read is used for source and write is used for destination, for both third_party_copy should be used.

cserf:
Transfer: Move conveyor transfers to third_party_copy_operation: Close #535

### DIFF
--- a/lib/rucio/core/rse.py
+++ b/lib/rucio/core/rse.py
@@ -17,7 +17,7 @@
 # - Ralph Vigne <ralph.vigne@cern.ch>, 2012-2015
 # - Mario Lassnig <mario.lassnig@cern.ch>, 2012-2017
 # - Martin Barisits <martin.barisits@cern.ch>, 2013-2018
-# - Cedric Serfon <cedric.serfon@cern.ch>, 2013-2016
+# - Cedric Serfon <cedric.serfon@cern.ch>, 2013-2018
 # - Thomas Beermann <thomas.beermann@cern.ch>, 2014-2017
 # - Wen Guan <wguan.icedew@gmail.com>, 2015-2016
 # - Brian Bockelman <bbockelm@cse.unl.edu>, 2018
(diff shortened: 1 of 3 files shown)


lib/rucio/core/rse.py
```diff
@@ -757,7 +753,6 @@ def get_rse_protocols(rse, schemes=None, session=None):
             'rse_type': str(_rse.rse_type),
             'credentials': None,
             'volatile': _rse.volatile,
-            'verify_checksum': verify_checksum[0] if verify_checksum else True,
```
mlassnig: hmmmm, did you remove this on purpose? i thought frank put this explicitly to make webdav transfers work.

mlassnig (commented):
this matches src:read_wan with dst:third_party_copy, correct?

lib/rucio/core/transfer.py
```diff
@@ -559,16 +559,15 @@ def get_transfer_requests_and_source_replicas(process=None, total_processes=None
                 allow_tape_source = True
 
                 # Find matching scheme between destination and source
-                # TODO: In the future, move to domain "third_party_copy"
                 try:
                     matching_scheme = rsemgr.find_matching_scheme(rse_settings_dest=rses_info[dest_rse_id],
                                                                   rse_settings_src=rses_info[source_rse_id],
                                                                   operation_src='read',
```
bari12: I was under the impression we do third_part_copy to third_party_copy, thus also using this as the source operation?

lib/rucio/core/rse.py
cserf: Oups, no. Will put it back

lib/rucio/core/transfer.py
mlassnig: we should, that's why i asked ;-)

lib/rucio/core/rse.py
mlassnig: also, line 742 above, just saw it

lib/rucio/core/transfer.py
mlassnig: this should also be third_party_copy

mlassnig (dismissed):

bari12:
@cserf please rebase

bari12 (approved):

Issue #938, opened 2018-03-16
 Add more python version support in setup_rucio_client.py


vingar:
Motivation
----------

Extend setup_rucio_client.py to support more python version.

PR #939, opened 2018-03-16
clients: Add more python version support in setup_rucio_client.py  (fix issue #938)

vingar:


setup_rucio_client.py
```diff
@@ -189,7 +189,8 @@ def get_file_list(self):
     description=DESCRIPTION,
     license="Apache License, Version 2.0",
     url="http://rucio.cern.ch/",
-    classifiers=[
+    python_requires=">=2.6, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*",
```
bari12, line 192: This would mean only python 3.4 support upwards? But no Python 3.0 support?

vingar: yes, python 3 is fine but not python 3.0.*. 

I preferred to follow the version restrictions defined in python requests:

    https://github.com/requests/requests/blob/master/setup.py#L75

as this is one of the main dependency module for the clients.

mlassnig: py34 is the minimum recommended py3 version (and also the default on RHEL), so it shouldn't be a problem.

bari12 (approved):

PR #950, opened 2018-03-19
Rules: Judge cleaner db time check does not work with certain versions of postgres; Fix #722

### LINKED ISSUE #722: rucio-judge-cleaner traceback against postgres
Motivation
----------

```# /bin/rucio-judge-cleaner 
  File "/bin/rucio-judge-cleaner", line 30, in <module>
    run(once=args.run_once, threads=args.threads)
  File "/usr/lib/python2.7/site-packages/rucio/daemons/judge/cleaner.py", line 143, in run
    if db_time - client_time > max_offset or client_time - db_time > max_offset:
TypeError: unsupported operand type(s) for -: 'datetime.date' and 'datetime.datetime' ```

### PR DESCRIPTION
**Pull request by bari12**
Rules: Judge cleaner db time check does not work with certain versions of postgres; Fix #722

### DIFF
--- a/lib/rucio/daemons/judge/cleaner.py
+++ b/lib/rucio/daemons/judge/cleaner.py
@@ -151,9 +151,10 @@ def run(once=False, threads=1):
     """
     client_time, db_time = datetime.utcnow(), get_db_time()
     max_offset = timedelta(hours=1, seconds=10)
-    if db_time - client_time > max_offset or client_time - db_time > max_offset:
-        logging.critical('Offset between client and db time too big. Stopping Cleaner')
-        return
+    if type(db_time) is datetime:
+        if db_time - client_time > max_offset or client_time - db_time > max_offset:
+            logging.critical('Offset between client and db time too big. Stopping Cleaner')
+            return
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: lib/rucio/daemons/judge/cleaner.py [hunk 1]
```diff
@@ -151,9 +151,10 @@ def run(once=False, threads=1):
     """
     client_time, db_time = datetime.utcnow(), get_db_time()
     max_offset = timedelta(hours=1, seconds=10)
-    if db_time - client_time > max_offset or client_time - db_time > max_offset:
-        logging.critical('Offset between client and db time too big. Stopping Cleaner')
-        return
+    if type(db_time) is datetime:
```
mlassnig, line 154: what if it's not? safe to continue?

bari12: It's the best you can do. Certain version of postgres only return a date, thus you cannot find out the actual db time that easily. (At least not via the sqlalchemy method). But this check is really an overly precautious safeguard. Not really needed.

mlassnig (approved):

Issue #954, opened 2018-03-19
Define a proper interface for the rucio transfertool


bari12:
Motivation
----------
Currently there is no interface definition for the transfertool; The fts3transfertool is the only implementation and de-facto the interface definition. However, there are many proprietary methods and especially not necessary methods in the fts3tranfertool. Thus it is beneficial to specify a proper interface and to also cleanup the fts3tranfertool.

PR #962, opened 2018-03-20
Transfers: Definition of a transfertool interface #954

bari12:
[WIP] Transfers: Definition of a transfertool interface #954

### DIFF
--- a/lib/rucio/transfertool/transfertool.py
+++ b/lib/rucio/transfertool/transfertool.py
@@ -1,0 +1,82 @@
+# Copyright 2018 CERN for the benefit of the ATLAS collaboration.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Authors:
+# - Martin Barisits <martin.barisits@cern.ch>, 2018
+
+from abc import ABCMeta, abstractmethod
+
+
+class Transfertool(object):
+    """
+    Interface definition of the Rucio transfertool
+    """
+
+    __metaclass__ = ABCMeta
+
+    def __init__(self, external_host):
+        """
+        Initializes the transfertool
+
+        :param external_host:   The external host where the transfertool API is running
+        """
+
+        self.external_host = external_host
+
+    @abstractmethod
+    def submit(self, files, job_params, timeout=None):
+        """
+        Submit transfers to the transfertool.
+
+        :param files:        List of dictionaries describing the file transfers.
+        :param job_params:   Dictionary containing key/value pairs, for all transfers.
+        :param timeout:      Timeout in seconds.
+        :returns:            Transfertool internal identifiers.
+        """
+        pass
+
+    @abstractmethod
+    def query(self, transfer_ids, details=False, timeout=None):
+        """
+        Query the status of transfers in the transfertool.
+
+        :param transfer_ids: List of transfertool internal identifiers as a string.
+        :param details:      Switch if detailed information should be listed.
+        :param timeout:      Timeout in seconds.
+        :returns:            Transfer status information as list of dictionaries.
+        """
+        pass
+
+    @abstractmethod
+    def cancel(self, transfer_ids, timeout=None):
+        """
+        Cancel transfers that have been submitted to the transfertool.
+
+        :param transfer_ids: Transfertool internal transfer identifiers as list of strings.
+        :param timeout:      Timeout in seconds.
+        :returns:            True if cancellation was successful.
+        """
+        pass
+
+    @abstractmethod
+    def update_priority(self, transfer_id, priority, timeout=None):
+        """
+        Update the priority of a transfer that has been submitted to the transfertool.
+
+        :param transfer_id: Transfertool internal transfer identifier as a string.
+        :param priority:    Job priority as an integer from 1 to 5.
+        :param timeout:     Timeout in seconds.
+        :returns:           True if update was successful.
+        """
+        pass
--- a/.pycodestyle
+++ b/.pycodestyle
@@ -1,0 +1,4 @@
+[pycodestyle]
+max-line-length=256
+ignore=E731,E722
+
--- a/lib/rucio/core/request.py
+++ b/lib/rucio/core/request.py
@@ -35,7 +35,7 @@ from rucio.core.rse import get_rse_id, get_rse_name
 from rucio.db.sqla import models
 from rucio.db.sqla.constants import RequestState, RequestType, FTSState, ReplicaState, LockState, RequestErrMsg
 from rucio.db.sqla.session import read_session, transactional_session
-from rucio.transfertool import fts3
+from rucio.transfertool.fts3 import FTS3Transfertool
 
 """
 The core request.py is specifically for handling requests.
@@ -360,7 +360,7 @@ def query_request(request_id, transfertool='fts3', session=None):
     if transfertool == 'fts3':
         try:
             ts = time.time()
-            response = fts3.query(req['external_id'], req['external_host'])
+            response = FTS3Transfertool(external_host=req['external_host']).query(transfer_ids=[req['external_id']], details=False)
             record_timer('core.request.query_request_fts3', (time.time() - ts) * 1000)
             req_status['details'] = response
         except Exception:
@@ -404,7 +404,7 @@ def query_request_details(request_id, transfertool='fts3', session=None):
 
     if transfertool == 'fts3':
         ts = time.time()
-        tmp = fts3.query_details(req['external_id'], req['external_host'])
+        tmp = FTS3Transfertool(external_host=req['external_host']).query(transfer_ids=[req['external_id']], details=True)
         record_timer('core.request.query_details_fts3', (time.time() - ts) * 1000)
         return tmp
 
@@ -659,11 +659,14 @@ def cancel_request_did(scope, name, dest_rse_id, request_type=RequestType.TRANSF
     except IntegrityError as error:
         raise RucioException(error.args)
 
+    transfertool_map = {}
     for req in reqs:
         # is there a transfer already in FTS3? if so, try to cancel it
         if req[1] is not None:
             try:
-                fts3.cancel(req[1], req[2])
+                if req[2] not in transfertool_map:
+                    transfertool_map[req[2]] = FTS3Transfertool(external_host=req[2])
+                transfertool_map[req[2]].cancel(transfer_ids=[req[1]])
             except Exception as error:
                 logging.warn('Could not cancel FTS3 transfer %s on %s: %s' % (req[1], req[2], str(error)))
         archive_request(request_id=req[0], session=session)
@@ -679,7 +682,7 @@ def cancel_request_external_id(transfer_id, transfer_host):
 
     record_counter('core.request.cancel_request_external_id')
     try:
-        fts3.cancel(transfer_id, transfer_host)
+        FTS3Transfertool(external_host=transfer_host).cancel(transfer_ids=[transfer_id])
     except:
         raise RucioException('Could not cancel FTS3 transfer %s on %s: %s' % (transfer_id, transfer_host, traceback.format_exc()))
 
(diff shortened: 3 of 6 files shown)


bari12:
This is WIP. Please have a look especially on the definition of the transfertool interface. This interface should be followed by all future transfertools. I also converged (and cleaned up a bit) the fts3 transfertool and implemented it against this new interface.
If the general structure/interface definition is fine with everyone, I will switch out the current fts3 transfertool against this one and also adapt all the conveyor calls.

lib/rucio/transfertool/transfertool.py
```diff
@@ -0,0 +1,76 @@
...
+    @abstractmethod
+    def submit(self, files, job_params, timeout=None):
+        """
+        Submit transfers to the transfertool.
+
+        :param files:        List of dictionaries describing the file transfers.
+        :param job_params:   Dictionary containing key/value pairs, for all transfers.
+        :returns:            Transfertool internal identifiers.
+        """
+        pass
```
vingar: raise NotImplementedError

lib/rucio/transfertool/transfertool.py
```diff
@@ -0,0 +1,76 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Authors:
+# - Martin Barisits <martin.barisits@cern.ch>, 2018
+
+from abc import ABCMeta, abstractmethod
+
+
+class Transfertool():
```
vingar: class Transfertool(Object):

provocative question: do we need a class  ? to keep the host as a class member but otherwise why?

lib/rucio/transfertool/transfertool.py
```diff
@@ -0,0 +1,76 @@
...
+        Submit transfers to the transfertool.
+
+        :param files:        List of dictionaries describing the file transfers.
+        :param job_params:   Dictionary containing key/value pairs, for all transfers.
+        :returns:            Transfertool internal identifiers.
+        """
+        pass
+
+    @abstractmethod
+    def query(self, transfer_ids, details=False):
```
vingar: @param timeout in all methods is a good idea

lib/rucio/transfertool/transfertool.py
bari12: It's an abstract class, does not need the NotImplementedError as it cannot be used like this anyway.

lib/rucio/transfertool/transfertool.py
vingar: About`transfer_ids`, if the transfer tool is simple webdav 3rd party transfer command or local cp, then `tranfersIds` are actually the same than `requestids`, correct ? btw Where is done the mapping destination=transferstool implementation ?

lib/rucio/transfertool/transfertool.py
bari12: Will add object.
I don't super like it as an object as well, but there is no other good way in python to enforce interfaces really.

lib/rucio/transfertool/transfertool.py
bari12: Will add.

bari12: the generation of the transfer_ids is really up to the transfertool. But in theory a simple transfertool could just hand back the request_ids yes.
The mapping must be done in the conveyor, but from these code changes we are a little bit away still.

lib/rucio/transfertool/transfertool.py
vingar: I'm not sure. Nothing prevent from instantiating an object from this class and submitting transfers. raise or return NotImplemented is more explicit and will fail also for the sub classes derived from that one if the method is not defined.  It's safer.

bari12: Python does ;)
```
from rucio.transfertool.transfertool import Transfertool
class a(Transfertool):
    pass
x = a()
TypeError: Can't instantiate abstract class a with abstract methods cancel, query, submit, update_priority
```

vingar: I didn't see `from abc import ABCMeta, abstractmethod`. Good to know :)

lib/rucio/transfertool/fts3_new.py
```diff
@@ -0,0 +1,637 @@
...
+            job_params["sid"] = files[0]['metadata']['request_id']
+            expected_transfer_id = self.__get_deterministic_id(job_params["sid"])
+            logging.debug("Submit bulk transfers in deterministic mode, sid %s, expected transfer id: %s" % (job_params["sid"], expected_transfer_id))
+
+        # bulk submission
+        params_dict = {'files': files, 'params': job_params}
+        params_str = json.dumps(params_dict)
+
+        r = None
+        if self.external_host.startswith('https://'):
```
cserf: This can be done in the constructor

lib/rucio/transfertool/fts3_new.py
```diff
@@ -0,0 +1,637 @@
...
+            job_params["id_generator"] = "deterministic"
+            job_params["sid"] = files[0]['metadata']['request_id']
+            expected_transfer_id = self.__get_deterministic_id(job_params["sid"])
+            logging.debug("Submit bulk transfers in deterministic mode, sid %s, expected transfer id: %s" % (job_params["sid"], expected_transfer_id))
+
+        # bulk submission
+        params_dict = {'files': files, 'params': job_params}
+        params_str = json.dumps(params_dict)
+
+        r = None
```
cserf: pylint don't like one char varaibles ;-)

lib/rucio/transfertool/fts3_new.py
```diff
@@ -0,0 +1,637 @@
...
+        :returns:            True if cancellation was successful.
+        """
+
+        if len(transfer_ids) > 1:
+            raise NotImplementedError('Bulk cancelling not implemented')
+        transfer_id = transfer_ids[0]
+
+        job = None
+
+        if self.external_host.startswith('https://'):
```
cserf: This can be done in the constructor

lib/rucio/transfertool/fts3_new.py
```diff
@@ -0,0 +1,637 @@
...
+        :param priority:    FTS job priority as an integer from 1 to 5.
+        :param timeout:     Timeout in seconds.
+        :returns:           True if update was successful.
+        """
+
+        job = None
+        params_dict = {"params": {"priority": priority}}
+        params_str = json.dumps(params_dict)
+
+        if self.external_host.startswith('https://'):
```
cserf: Sam comment

lib/rucio/transfertool/transfertool.py
```diff
@@ -0,0 +1,82 @@
...
+    def update_priority(self, transfer_id, priority, timeout=None):
+        """
+        Update the priority of a transfer that has been submitted to the transfertool.
+
+        :param transfer_id: Transfertool internal transfer identifier as a string.
+        :param priority:    Job priority as an integer from 1 to 5.
+        :param timeout:     Timeout in seconds.
+        :returns:           True if update was successful.
+        """
+        pass
```
cserf, line 82: Shouldn't we add a delegation method ?

lib/rucio/transfertool/fts3_new.py
```diff
@@ -0,0 +1,637 @@
...
+            job = requests.post('%s/jobs/%s' % (self.external_host, transfer_id),
+                                verify=False,
+                                data=params_str,
+                                cert=(__USERCERT, __USERCERT),
+                                headers={'Content-Type': 'application/json'},
+                                timeout=3)
```
cserf: You don't propagate the timeout

cserf (commented):
Some first comments

lib/rucio/transfertool/fts3_new.py
bari12: The fts3 code is the copy&paste of the current fts3 code with some minor changes. I am not planning to change this massively at the moment is it is working...

lib/rucio/transfertool/transfertool.py
bari12: I was thinking about it, but thinking a bit in the future I don't think that all transfertools would need delegation, so I think it should just be added in the ones who need it.

lib/rucio/transfertool/fts3_new.py
bari12: Will change.

lib/rucio/transfertool/fts3_new.py
bari12: I am not sure what you mean? What exactly should be moved to the constructor?

cserf: I means something like 
```python
if self.external_host.startswith('https://'):
   self.cert=(__USERCERT, __USERCERT)
else:
   self.cert = None
```
Then just call:
```python
r = requests.post('%s/jobs' % self.external_host,
                                  verify=False,
                                  cert=self.cert,
                                  data=params_str,
                                  headers={'Content-Type': 'application/json'},
                                  timeout=timeout)
```

bari12: Ah ok, but this doesn't really reduce the lines of code at all? Because you need the if block anyway to generate the requests.post request once with cert and once without.

lib/rucio/transfertool/fts3_new.py
```diff
@@ -0,0 +1,639 @@
...
+                logging.warn('Could not submit transfer to %s - %s' % (self.external_host, str(traceback.format_exc())))
+        else:
+            try:
+                ts = time.time()
+                r = requests.post('%s/jobs' % self.external_host,
+                                  data=params_str,
+                                  headers={'Content-Type': 'application/json'},
+                                  timeout=timeout)
+                record_timer('transfertool.fts3.submit_transfer.%s' % self.__extract_host(self.external_host), (time.time() - ts) * 1000 / len(files))
+            except:
```
cserf: Would be good to catch explicitely the following error : 

SSLError: HTTPSConnectionPool(host='fts.usatlas.bnl.gov', port=8446): Max retries exceeded with url: /jobs (Caused by SSLError(SSLError("bad handshake: SysCallError(104, 'ECONNRESET')",),))

bari12:
No longer WIP, please review

mlassnig (commented):
I think this is fine for a first step. If we see that we need to adapt the interface it's also not so much work since there's only one implementation to drag along.

lib/rucio/transfertool/transfertool.py
```diff
@@ -0,0 +1,82 @@
...
+        """
+        Initializes the transfertool
+
+        :param external_host:   The external host where the transfertool API is running
+        """
+
+        self.external_host = external_host
+
+    @abstractmethod
+    def submit(self, files, job_params, timeout=None):
```
vingar, line 38: Just a remark on the semantics: `job_params` looks weird. `request_params` or maybe `**params` at the end of the signature ?

lib/rucio/transfertool/transfertool.py
```diff
@@ -0,0 +1,82 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Authors:
+# - Martin Barisits <martin.barisits@cern.ch>, 2018
+
+from abc import ABCMeta, abstractmethod
+
+
+class Transfertool(object):
```
vingar, line 21: I've seen some examples with with_metaclass. 
```
from abc import ABCMeta, abstractmethod
from six import with_metaclass
 
class Transfertool(with_metaclass(ABCMeta)):
```
and usually they are explicit (rather then implicit) with method definition, i.e.,
```
 @abstractmethod
    def methods(self):
        raise NotImplementedError
```

bari12:
Yes, it's just a first step. If a new transfertool comes along we might need some (minor) adaptions.

mlassnig:
There's also some work by Diego to improve the delegation. This might also have some impact in the future.

lib/rucio/transfertool/transfertool.py
vingar: A global `request_id=None` can be added to the signature I think. this allows specifying the uuid from the client. It was mentioned few times with FTS in the past.

mlassnig (approved):

vingar (commented):
Few comments ignored ! 
Just hoping that this one won't be ignored either...

bari12:
What was ignored that specifically needed to be addressed?

lib/rucio/transfertool/transfertool.py
bari12: I used `job_params` as it was like this before and it is more clear, as it specifically asks for job parameters. `**params` is too ambiguous and generally not a good style if not absolutely needed.
The `request_id`, or better `transfer_id`, we can still add.

lib/rucio/transfertool/transfertool.py
bari12: Yes, the `with_metaclass` we should probably add for Py2/3 compatibility. The explicit raising of the error is not needed.

PR #967, opened 2018-03-21
replicas: bug in domain selection with proxy

mlassnig:
replicas: bug in domain selection with proxy


mlassnig:
"""The job exceeded the maximum time limit for jobs, and has been terminated."""

mlassnig:
Maximum travis run time is 50 minutes.

bari12 (approved):

PR #969, opened 2018-03-21
clients: Few Fixes (Fix #965)

### LINKED ISSUE #965: Client AttributeError when the server returns nothing
Motivation
----------

```
# /usr/rucio/tools/probes/common/check_ping_rucio_servers rucio-server-int-01.cern.ch 
  File "/usr/rucio/tools/probes/common/check_ping_rucio_servers", line 18, in <module>
    print 'Rucio version installed : %s ' % c.ping()['version']
  File "/usr/lib/python2.7/site-packages/rucio/client/pingclient.py", line 45, in ping
    if r.status_code == codes.ok:
AttributeError: 'NoneType' object has no attribute 'status_code'
```

vingar:


lib/rucio/common/exception.py
```diff
@@ -811,3 +820,13 @@ def __init__(self, *args, **kwargs):
         super(MissingModuleException, self).__init__(args, kwargs)
         self._message = "The module is not installed."
         self.error_code = 75
+
+
+class RucioServerException(RucioException):
```
bari12: I think this should be renamed to ConnectionFailureException. Or FailedConnectionException or something like this.

vingar: I thought about it but failure/Failed sounds to me like a temporary glitch and that it can be retried later on while here it can be a misconfiguration for example. I'm looking for a better name...

mlassnig:
How about ServiceMisconfigurationError ?

vingar:
it's not necessary a misconfiguration as the server can be down and give the same symptom

vingar:
maybe  RucioServerConnectionException  ?

mlassnig:
sounds ok to me

bari12:
I think it should be just ConnectionException. Maybe ServerConnectionException. But we do not have Rucio in all the other exceptions, it is just redundant. Server is also a bit redundant, because what else can it be? I feel ConnectionException would be better, and describing it a bit more in the message (That's already there in the PR)

vingar:
ah just see you comment now.  I put `ConnectionServerException` finally

bari12:
ConnectionServerException is worse 🤣 
ServerConnectionException or ConnectionException.
Actually, we should not use Exception but Error, according to PEP8, but we are very inconsistent with this in the exceptions anyway.

vingar:
both looks fine to me..but I can offer you a `ServerConnectionException`. I prefer Error as well but it's a backward incompatible change to introduce. Maybe with the API v2

lib/rucio/common/exception.py
```diff
@@ -811,3 +820,13 @@ def __init__(self, *args, **kwargs):
...
+
+
+class ServerConnectionException(RucioException):
+    """
+    RucioException
+    """
+    def __init__(self, *args, **kwargs):
+        super(ServerConnectionException, self).__init__(args, kwargs)
+        self._message = "Cannot connect to the Rucio server."
+        self.error_code = 76
```
mlassnig: NotAllFilesDownloaded already has errorcode 76 ;-)

and MissingModuleException just above has 75, which is already claimed by NoFilesDownloaded.

Can you please set:  MissingModuleException=77, and ServerConnectionException=78

mlassnig (approved):

Issue #992, opened 2018-03-22
Add rucio daemons CLIS as command-Line applications and entry points


vingar:
Motivation
----------

Add entry points for the command rucio and rucio-admin to install them automatically onto the user’s $PATH, so that they can be executed directly from a terminal session.

vingar:
it's easier to do it for the daemons first as there are no tests using the clis.

cf. http://python-packaging.readthedocs.io/en/latest/command-line-scripts.html

PR #993, opened 2018-03-22
core: Add rucio daemons CLIS as command-Line applications and entry points (issue: #992)

vingar:


setup_rucio.py
```diff
@@ -224,4 +224,10 @@ def get_file_list(self):
     install_requires=requires,
     extras_require=extras_require,
     dependency_links=depend_links,
+    entry_points = {
+    'console_scripts':
+        ['rucio-reaper=rucio.daemons.reaper.clis.reaper:main',
```
bari12: The actual locations of this clis we should still think about. Like this it feels very fragmented.
Do I understand right that everything in bin/* would go away as it is automatically created when the package is installed? This has some implications in a test environment where you are just working from source and do not install the package. Some of the tests using the CLI directly also have to be changed possibly.

vingar: I got some hesitation to put them in `rucio/daemons/clis` but it looks fine like that imi. Yes for the tests. The test `test_bin_rucio.py` should be rewritten to not spawn processes or we change our tests to install rucio with python setup.py at the beginning which will validate setup.py and the packaging as well.

vingar:
One advantage of having the command line in lib:
```
$ flake8 bin/rucio
```
gives nothing. but:

```
$ flake8 lib/rucio/client/clis/rucio.py
lib/rucio/client/clis/rucio.py:2225:16: F821 undefined name 'result'
```

setup_rucio.py
bari12: I think it could make sense to have all these centrally in `lib/rucio/clis`
I dont't like `lib/rucio/client/clis` a lot as we established `client/` as the location for the python apis.

vingar: It's a **convention** to decide imo. Several options:
1. All entry point source files(daemon+clients) go in `lib/rucio/clis`. It is quiet similar to what we have today as all scripts are centralized in `rucio/bin`.

2. The client ones in `lib/rucio/client/clis` and the daemon ones in `lib/rucio/daemons/clis`.

3. The client ones can also go in `lib/rucio/client/clis`  and the daemons one in each daemon directory `lib/rucio/daemons/{daemon name}/clis`. But the cons is the fragmentation you mentioned and I agree with it. On the other hand, it looks cleaner and self-contained.

4. You name it.

I prefer option 2.

bari12: I prefer 1 a little bit over 2 for the reason that all the clis are in one directory. To me it feels weird that we point people to the API directory `lib/rucio/client`, when they ask for the rucio APIs and then there is another directory in there which has nothing to do with the API. 
For 1 we could still do a `daemons/` and `clients/` - but I think all in one is fine as well.
3 and 4 I don't like 😄

vingar: yup, another option might be 4) `lib/rucio/clis/daemons/` and `lib/rucio/clis/clients/` which I prefer now.

bari12: Agreed, let's do it like this.

setup.py
```diff
@@ -1,227 +0,0 @@
-# Copyright 2012-2018 CERN for the benefit of the ATLAS collaboration.
```
bari12, line 1: For the general setup.py, I thought we need this for readthedocs? Is this no longer needed and the link to the setup_rucio.py is fine?

vingar: yes, it's fine.

bari12:
Can you try to rebase this? travis is blocked (Although it might be that the rebase won't help; we had this with another PR last week)

Issue #1007, opened 2018-04-02
CMS schema is too permissive


ericvaandering:
Motivation
----------
The CMS schema is too permissive on RSE names, scopes, and possibly LFNs

Modification
------------
Make these match actual CMS requirements and add testing to verify it

PR #1008, opened 2018-04-02
Common: test CMS Schema #1007

ericvaandering:
Would like some feedback on this from @bbockelm @nataliaratnikova and others. 

First I propose to enforce that the scopes are "cms" and user.HNNAME where HNNAME follows our established regex. If we need another scope, we should add it explicitly, I think. (Eventually the PFN algorithm will pay attention to this as well).

Then we have the Site/RSE which should match what we have in WMCore for a Regex (the RSE previously was allowed to be things that did not conform to that).

Then, you can see in the test script was is allowed/disallowed

### DIFF
--- a/lib/rucio/tests/test_schema_cms.py
+++ b/lib/rucio/tests/test_schema_cms.py
@@ -1,0 +1,140 @@
+# Copyright European Organization for Nuclear Research (CERN)
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# You may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Authors:
+# - Eric Vaandering, <ewv@fnal.gov>, 2018
+
+from __future__ import absolute_import, division, print_function
+
+from nose.tools import assert_raises
+
+from rucio.common.exception import InvalidObject
+from rucio.common.schema.cms import validate_schema
+
+
+# Some tests adapted from https://github.com/dmwm/WMCore/blob/master/test/python/WMCore_t/Lexicon_t.py
+
+class TestSchemaCMS(object):
+
+    def __init__(self):
+        pass
+
+    def test_site_names(self):
+        """ CMS SCHEMA (COMMON): Test site/RSE names """
+        validate_schema('rse', 'T2_US_Nebraska')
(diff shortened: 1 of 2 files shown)


ericvaandering:
Please don't merge this yet.

ericvaandering:
Fixed one minor issue here too. Please merge when it comes back OK.

vingar:
`pylint` is failing now and It took me a while to spot the error, but I got it:

```
************* Module rucio.tests.test_schema_cms
E: 40, 4: Method should have "self" as first argument (no-self-argument)
```

lib/rucio/tests/test_schema_cms.py
```diff
@@ -0,0 +1,140 @@
...
+        with assert_raises(InvalidObject):
+            validate_schema('rse', 'T1_US')
+        with assert_raises(InvalidObject):
+            validate_schema('rse', 'T1_US_')
+        with assert_raises(InvalidObject):
+            validate_schema('rse', 'T1_US_FNAL__Disk')
+        with assert_raises(InvalidObject):
+            validate_schema('rse', 'T2_US_Nebraska-Subpart')
+
+    def test_dids(selfs):
```
vingar: extra s

ericvaandering:
I'm going to blame my IDE. :-) Fixed it.

mlassnig (approved):

nataliaratnikova:
Hi Eric, all 
we'll have to see how PhEDEx subscription groups translate into this schema. 
PhEDEx groups are currently used for two purposes: to mark replica ownership and as activity tag, and it is possible to reassign the subscription group on the existing replica, see discussion in  https://hypernews.cern.ch/HyperNews/CMS/get/dmDevelopment/2024.html . 
In Rucio these are two separate concepts, the activity is implemented via a rule attribute.
Having group scopes would allow to maintain group quotas. I wonder how Rucio/Atlas are doing this.

vingar:
Not sure to have the complete picture but In Rucio, the quota is per account and it can be set in the subscriptions or rules. Activity is mostly for monitoring and applied fairshares within FTS.

ericvaandering:
In rucio scope:did is unique, not did so scope always, somehow, has to factor into the PFN. But I don't think this is a problem. Rules have account owners.

Issue #1009, opened 2018-04-03
Add python2.7 conditional requirements to core dependencies to support pip py3 installation


vingar:
Motivation
----------

Add conditional requirements to `./tools/pip-requires` to mark the non compatible python2.7 packages and support `pip install -r ./tools/pip-requires` for python 3.

PR #1010, opened 2018-04-03
[WIP] Release mgt: Add python2.7 conditional requirements to core dependencies to support pip py3 installation (Fix issue #1009)

vingar:


bari12:
Please resubmit as a feature; I know it's not a dependency change, but I think to make these changes only on features was quite helpful; (Also this fixes the conflict on the other PR)

vingar:
A patch is needed here to have separate tests for python2.7/3 in travis that will need it.

bari12:
The separate tests can also be on the next branch?

vingar:
If we don't want to regress our code wrt python 3, I would keep it on the two branches. 

BTW, For the conflict I can rebase against next once the PR on master is approved.

bari12:
It will eventually go to the master branch with the next feature release. This is already in 3 weeks, we can wait that long for the python 3 tests, I don't think it is worth to break policy for this.

vingar:
It doesn't break  *the convention* as the changes here are backward compatible :relaxed:.  
My ideas were to have in Travis some specific tests for various python versions. We can start by having `pip install -r ./tools/pip-requires-clients` done with 2.6, 2.7 and > 3. Then we can extend the tests by running `nosetests  lib/rucio/tests/test_clients.py` against the local rucio server running within the container. 
For the later, this patch is needed as some dependencies are in `tools/pip-requires-tests` and depending of `rucio.cfg` the ones in `tools/pip-requires`.  All these changes for me were compatible  with `master` and `next` branches.

bari12:
Did you try if it works with an old pip version < 6.0? 😄 

The idea is really great, but it doesn't need to have these changes on master immediately, since it also works if it is next only.
This is really more a point of principle, why to take the risk and throw policy over board for no advantage? We had issues with changes like this before, where we were absolutely sure that it doesn't break anything and then it did - exactly why we decided on this policy.

vingar:
I think that I'll iterate directly on this pull request with the changes instead  and put it as WIP. I did some experimentations in #998, which will be included in this pull request. 
No opinion on the question of master and next as I really feel it's secondary for now, regarding the target. What do you think ?

Issue #27, opened 2017-11-08
Possibility to pass a timeout for a protocol


bari12:
_Migrated from JIRA [RUCIO-2247](https://its.cern.ch/jira/browse/RUCIO-2247)_

nikmagini:
Proposed implementation:

1) For protocols/tools that natively support timeout (e.g. gfal), pass the timeout to the underlying command. Targeting this for the next client release.

2) Later, for all protocols implement a watchdog in rucio download/upload that kills the transfer after the timeout expiration. The plan is to do this only for the new downloadclient/uploadclient libraries.

nikmagini:
Checking the code, the gfal protocol already has a hardcoded 3600 s timeout

PR #1015, opened 2018-04-05
Clients: --transfer-timeout option for upload/download #27

nikmagini:
Clients: --transfer-timeout option for upload/download #27

### DIFF
--- a/lib/rucio/client/downloadclient.py
+++ b/lib/rucio/client/downloadclient.py
@@ -16,6 +16,7 @@
 # - Tomas Javurek <tomasjavurek09@gmail.com>, 2018
 # - Vincent Garonne <vgaronne@gmail.com>, 2018
 # - Joaquin Bogado <jbogado@linti.unlp.edu.ar>, 2018
+# - Nicolo Magini <nicolo.magini@cern.ch>, 2018
 
 import os
 import os.path
@@ -63,7 +64,7 @@ class DownloadClient(BaseClient):
         self.trace_taskid = os.environ.get('RUCIO_TRACE_TASKID', None),
         self.trace_usrdn = os.environ.get('RUCIO_TRACE_USRDN', None)
 
-    def download(self, dids, rse, protocol='srm', pfn=None, nrandom=None, nprocs=None, user_agent='rucio_clients', dir='.', no_subd=False):
+    def download(self, dids, rse, protocol='srm', pfn=None, nrandom=None, nprocs=None, user_agent='rucio_clients', dir='.', no_subd=False, transfer_timeout=None):
 
         trace_endpoint = client.host
         trace_pattern = {'hostname': socket.getfqdn(),
@@ -221,7 +222,7 @@ class DownloadClient(BaseClient):
                     input_queue.put(f)
 
         try:
-            self.download_rucio(pfn, protocol, input_queue, output_queue, trace_pattern, trace_endpoint, nprocs, user_agent, dir, no_subd)
+            self.download_rucio(pfn, protocol, input_queue, output_queue, trace_pattern, trace_endpoint, nprocs, user_agent, dir, no_subd, transfer_timeout)
         except Exception as error:
             logger.error('Exception during download: %s' % str(error))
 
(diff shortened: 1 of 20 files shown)


nikmagini:
Hi,

first iteration of the transfer timeout implementation. Currently the timeout is passed only to the gfal and srm protocols; for all others it's still dummy. But I opened already a pull request to review the interface changes since the commit is pretty big already. Send your comments!

Thanks
Nicolo

lib/rucio/client/downloadclient.py
```diff
@@ -288,7 +289,7 @@ def download_rucio(self, pfn, protocol, input_queue, output_queue, trace_pattern
                 thread.kill_received = True
         logger.debug('All threads finished')
 
-    def _downloader(self, pfn, protocol, human, input_queue, output_queue, user_agent, threadnb, total_threads, trace_endpoint, trace_pattern):
+    def _downloader(self, pfn, protocol, human, input_queue, output_queue, user_agent, threadnb, total_threads, trace_endpoint, trace_pattern, transfer_timeout=3600):
```
bbockelm: Why is the timeout `3600` in some places and `None` at others?

Seems that if the user doesn't specify one, we should ask the configuration for a timeout rather than scatter a "magic integer" through the codebase.

nikmagini: The ```None``` timeout is just a placeholder for the protocols where no timeout is implemented yet.

Where the protocol is implemented, I selected the magic value of ```3600``` just because it's what is hardcoded in the gfal protocol in the current production version, but of course I'm all for more sensible suggestions.

bbockelm: Oh, I'm not sure I have more sensible suggestion of a value -- I just want to aggregate it in one place instead of distributing it throughout the codebase.

vingar: I'm wondering if it's not better to extend the protocols `get` and `put` functions with **kwargs instead in the signature. It's more generic and allow more possible extensions

bari12: I think it should be explicit, wherever possible, as it increases the readability a lot. But when we have a usecase for a protocol we might want to add the **kwargs on top of that.

bari12 (changes requested):
Let's remove the default (3600) transfer_timeout everywhere and default the parameter in the various methods to None. In bin/rucio we try to get one from the config, if this is not found we use 3600.

vingar:
@bari12: Just to understand what you mean by config and clarify the policy there. 

One possible scenario:

Timeout is a per operation parameter and can be different depending on the site, environment, etc. It should be settable in `rucio download/upload` or propagated to the APIs (pilot, arc, etc). 

If the timeout is not defined, then the default one is taken from `rucio.cfg` for `rucio download/upload`. Remark: I don't think we have such upload/download section now in `rucio.cfg`.

If the timeout is not defined in `rucio.cfg`. The default value is `None` (?)

bari12:
Yes, that was pretty much my thought, with one correction:
If timeout is not defined in `rucio.cfg` I would still use a hardcoded default of 3600. The reason is that this is the behaviour right now and we might not be able to update the `rucio.cfg` in all environments, thus this would change the standard behaviour. But the only place where this "magic value" would be stated is in `bin/rucio` anyway, so it is not distributed all over the codebase.

vingar:
it's more clear for me, now 👍 

I would suggest to put only the default timeout value in the argparse options of `rucio` in this pull request and have a separate issue/pull request for the config changes.

bari12:
yes, that's also fine by me. Lets see how @nikmagini prefers it 😄

nikmagini:
Hi,

I set the default transfer_timeout to None everywhere except in bin/rucio argparse. I'll open a separate issue for the new config params.

Side note: the interfaces for the new DownloadClient.download and UploadClient.upload are completely different; I implemented transfer_timeout in each of the following the current interfaces but maybe it's worth harmonising them.

bari12 (approved):

PR #1030, opened 2018-04-09
Clients: added raise if api upload fails Fix #1029

### LINKED ISSUE #1029: rucio upload api doesn't raise an error in some error cases
Motivation
----------

Modification
------------

TWAtGH:
Clients: added raise if api upload fails Fix #1029


mlassnig (dismissed):

bari12 (changes requested):
For this it would be good if @TWAtGH can add the exception as discussed in the other PR.

bari12 (approved):

PR #1031, opened 2018-04-09
Clients: added raise if api upload fails Fix #1029

### LINKED ISSUE #1029: rucio upload api doesn't raise an error in some error cases

TWAtGH:
Clients: added raise if api upload fails Fix #1029


lib/rucio/client/uploadclient.py
```diff
@@ -312,7 +312,7 @@ def upload(self, sources_with_settings, summary_file_path=None):
                 else:
                     logger.error('Failed to upload file %s' % basename)
                     # TODO trace?
-                    continue  # skip attach_did and update_states for this file
+                    raise RucioException('Failed to upload file %s' % basename)
```
vingar: it might be worth to add a new exception in `lib/rucio/common/exception.py` like FailedUpload and raises it there.

bari12: Yes, this should be an exception. I am actually surprised that we didn't have one before 😄

bari12 (approved):

Issue #1025, opened 2018-04-09
add documentation build validation in travis


vingar:
Motivation
----------

Add `sphinx-build -T doc/source/  doc/build/html` and check if the status is not different than 0.

Issue #1024, opened 2018-04-09
setuptools package update to 39.0.1   


vingar:
Motivation
----------

Fix setuptools package updaye to 39.0.1 to avoid issue with old setuptools version like:
```install fails with: `install_requires` must be a string or list of strings```

Issue #1022, opened 2018-04-09
add sqlite backend tests to Travis


vingar:
Motivation
----------

Add sqlite as a backend to test in travis

bari12:
I am not sure if we really need to add this in the days of containers. The argument for sqlite was always that it is easy as you do not need to install mysql etc for quick development. But now in the days of containers the point is not really valid anymore. If the code runs fine on mysql/postgres/oracle it is fine I think.

Issue #1020, opened 2018-04-09
Extend setup.py with dev requirements/dependencies in extra-requires


vingar:
Motivation
----------

Provide all dev. dependencies in extra-require for setup.py:

`pip install .[dev]`

PR #1033, opened 2018-04-10
Release management: More work on travis (Fix issues #1009 #1017 #1017 #1018 #1019 #1020 #1012 #1022 #1024 #1025)

vingar:


bari12:
For me this is good to go, but as it is quite a big PR, if anyone else wants to have a look, please do it today, I will merge tomorrow/tonight.

mlassnig:
Looks good to me.

Just one improvement, it might be worth to run the flake/syntax check first, before spawning all the subjobs? Because if that one doesn't work, then we can fail quickly and not need to spawn all the matrix jobs.

vingar:
It's the concept of build stages: https://docs.travis-ci.com/user/build-stages/
It can come later if someone is motivated to have a look on it. It'll save CPU in travis but the improvement for us is already there: total waiting time around ~20 min with much more coverage.

mlassnig (approved):

PR #1041, opened 2018-04-12
Transfer : verify_checksum is not properly passed to FTS + py3k fix . Closes #1040

### LINKED ISSUE #1040: Transfer : verify_checksum is not properly passed to FTS
Motivation
----------
The verify_checksum RSE attribute is passed as a string instead of a bool to FTS

Modification
------------
Cast the string to bool

cserf:
Transfer : verify_checksum is not properly passed to FTS + py3k fix . Closes #1040

### DIFF
--- a/lib/rucio/core/transfer.py
+++ b/lib/rucio/core/transfer.py
@@ -596,7 +596,7 @@ def get_transfer_requests_and_source_replicas(process=None, total_processes=None
 
                 # Compute the destination url
                 if rses_info[dest_rse_id]['deterministic']:
-                    dest_url = protocols[dest_rse_id].lfns2pfns(lfns={'scope': scope, 'name': name}).values()[0]
+                    dest_url = list(protocols[dest_rse_id].lfns2pfns(lfns={'scope': scope, 'name': name}).values())[0]
                 else:
                     # compute dest url in case of non deterministic
                     # naming convention, etc.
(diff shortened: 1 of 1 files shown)


lib/rucio/core/transfer.py
```diff
@@ -677,7 +677,7 @@ def get_transfer_requests_and_source_replicas(process=None, total_processes=None
                                  'filesize': bytes,
                                  'md5': md5,
                                  'adler32': adler32,
-                                 'verify_checksum': rse_attrs[dest_rse_id].get('verify_checksum', True)}
+                                 'verify_checksum': json.loads(rse_attrs[dest_rse_id].get('verify_checksum', 'True').lower())}
```
bari12, line 680: I am not sure if this works? For the default (if it is not set, it is fine) - but if verify_checksum is set to `True` or `False` I don't think this works.

cserf: I tested it on integration nodes and it works ;-)

mlassnig: wouldn't it suffice to cast it to bool() instead of running it through the json parser?

mlassnig: ok, this is confusing as hell, apparently the json.loads really is the best variant to do this. bool() does not do what i expected it to do.

mlassnig (approved):

PR #1049, opened 2018-04-13
Transfers : Modify the verify_checksum parameter passed to FTS. Closes #1048

### LINKED ISSUE #1048: Transfers : Modify the verify_checksum parameter passed to FTS
Motivation
----------
```
Hi Cedric,

the checksum validation was improved to have all checksum possibilities. Before, when you where indicating verify checksum True was applying the checksum to both (source and destination) and if you were using False the verification was not applied.
I just keep the True and False for compatibility with the API. But, we encouraged when this new feature was presented in the steering, to use both if you want a checksum verification in source and destination, source when you only wants the verification on the source, destination just to verify in destination and none for no verification.

So, you can provide the verification that adapts better to your needs as it is documented.

Cheers,
Maria
```

cserf:
Transfers : Modify the verify_checksum parameter passed to FTS. Closes #1048

### DIFF
--- a/lib/rucio/core/transfer.py
+++ b/lib/rucio/core/transfer.py
@@ -533,6 +533,8 @@ def get_transfer_requests_and_source_replicas(process=None, total_processes=None
                 if source_rse_id not in rses_info:
                     source_rse = get_rse_name(rse_id=source_rse_id, session=session)
                     rses_info[source_rse_id] = rsemgr.get_rse_info(source_rse, session=session)
+                if source_rse_id not in rse_attrs:
+                    rse_attrs[source_rse_id] = get_rse_attributes(source_rse_id, session=session)
 
                 attr = None
                 if attributes:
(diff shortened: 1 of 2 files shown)


mlassnig (commented):
looks good to me. @bari12 ?

lib/rucio/core/transfer.py
```diff
@@ -659,6 +661,18 @@ def get_transfer_requests_and_source_replicas(process=None, total_processes=None
                     retry_count = 0
                 fts_list = fts_hosts.split(",")
 
+                verify_checksum = 'both'
+                if rse_attrs[dest_rse_id].get('verify_checksum', 'True') == 'False':
```
bari12: I think we should put .upper() or .lower() into these, just to make it save.

bari12 (changes requested):

bari12:
Ok travis is a bit weird one these ones. I tested this manually and it works. Will force merge.

PR #1058, opened 2018-04-18
core: rse add protocols does not set third_party_copy; Fix 1057

mlassnig:
core: rse add protocols does not set third_party_copy


cserf (approved):

bari12:
Please add issue number to commit message 😄

vingar:
The required metadata for the change are now bigger than the change itself...😁

PR #1061, opened 2018-04-18
Transfer : Fix compatible transfers mapping: Closes #1060

### LINKED ISSUE #1060: Transfer : Fix compatible transfers mapping
Motivation
----------
The table mapping compatible mapping is erroneous 

Modification
------------
Fix the mapping table

cserf:
Transfer : Fix compatible transfers mapping: Closes #1060

### DIFF
--- a/lib/rucio/common/constants.py
+++ b/lib/rucio/common/constants.py
@@ -6,7 +6,7 @@
 #
 # Authors:
 # - Vincent Garonne, <vincent.garonne@cern.ch>, 2013
-# - Cedric Serfon, <cedric.serfon@cern.ch>, 2015
+# - Cedric Serfon, <cedric.serfon@cern.ch>, 2015-2018
 # - Mario Lassnig, <mario.lassnig@cern.ch>, 2018
 
 """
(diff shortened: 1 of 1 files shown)


mlassnig:
We cannot do this, as this would block all transfers to Google until we have signed URL support in FTS. Suggest to put on hold.

cserf:
Since more than 80 % of the sites have a WebDAV door, enabling S3 <-> davs  only would be enough.

vingar:
To elaborate more on this: Introducing the distinction between stream and third party copy can be useful for the scheme mapping.

bari12:
I don't think getting this stream information in there is particular useful. If we really need to regulate on this I would suggest to introduce a parameter in the conveyor-submitter where you can specify the protocol pairs the conveyor is allowed to use (or not allowed to use).

cserf:
What can be done to regulate this is that the conveyor marks the non-matching protocol transfers, then it can release them slowly depending on how many transfers are already streaming through FTS.

bari12:
That sounds a bit complicate for something which we should only use for integration/testing anyway. Maybe we can put these rules just in a highly throttled activity?

vingar:
Do we need to know that we have stream party copy ? For example to limit them, be protected against, monitor ? If yes, a conveyor might have a flag to allow them or not and should be able to make the distinction between the two. If no, nothing needed :)

mlassnig:
One other thing, FTS server will always go PULL->PUSH->STREAM, if either fails. We already might be streaming quite a bit.

mlassnig (approved):

vingar:
@mlassnig you approved the change which removes the compatibility srm-s3. Is it expected ?

mlassnig:
Yep, instead https/davs -> S3 was added, which is now supported by FTS.

PR #1121, opened 2018-05-03
Transfers : Bug with verify checksum : Closes #1118

### LINKED ISSUE #1118: Transfers : Bug with verify checksum
Motivation
----------
The conveyor crash if verify_checksum is a bool instead of a string

cserf:
Transfers : Bug with verify checksum : Closes #1118

### DIFF
--- a/lib/rucio/core/transfer.py
+++ b/lib/rucio/core/transfer.py
@@ -662,13 +662,13 @@ def get_transfer_requests_and_source_replicas(process=None, total_processes=None
                 fts_list = fts_hosts.split(",")
 
                 verify_checksum = 'both'
-                if rse_attrs[dest_rse_id].get('verify_checksum', 'True').lower() == 'false':
-                    if rse_attrs[source_rse_id].get('verify_checksum', 'True').lower() == 'false':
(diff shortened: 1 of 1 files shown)


lib/rucio/core/transfer.py
```diff
@@ -662,13 +662,13 @@ def get_transfer_requests_and_source_replicas(process=None, total_processes=None
                 fts_list = fts_hosts.split(",")
 
                 verify_checksum = 'both'
-                if rse_attrs[dest_rse_id].get('verify_checksum', 'True').lower() == 'false':
-                    if rse_attrs[source_rse_id].get('verify_checksum', 'True').lower() == 'false':
+                if not rse_attrs[dest_rse_id].get('verify_checksum', True):
```
mlassnig, line 665: is there a reason why it's not

if rse_attrs.get(...   , False):

?

bari12: Because if the attribute is not set, we still want checksum checks (per default)?

mlassnig (approved):

bari12:
This works for `bool` now, but not for string?

PR #1146, opened 2018-05-08
credentials: make config load optional; Fix #1144

### LINKED ISSUE #1144: Credentials config/file should be optional not mandatory
Motivation
----------
Currently the [credentials] section in the config must be filled, otherwise Rucio does not work. However, as this is an optional feature the config should also be optional.

mlassnig:
credentials: make config load optional; Fix #1144


lib/rucio/web/rest/webpy/v1/replica.py
```diff
@@ -287,6 +287,7 @@ def POST(self):
 
         dids, schemes, select, unavailable, limit = [], None, None, False, None
         ignore_availability, rse_expression, all_states, domain = False, None, False, None
+        lifetime = 600
```
bari12: Doesnt this introduce a default lifetime for the replicas?

mlassnig: this is the default lifetime of the signed url

bbockelm: Can you re-route the defaults through the config system at least?  Seems to be a bit strange how buried this default is...

vingar: TBC: lifetime might be used to express the lifetime of the replica location for the cache use case and possibly set in the db.

bari12: Ah I got confused as this is a POST REST method, thus I confused it with replica registration. However, this is just the list replica method (for which we use POST too...) 🤦‍♂️ 
I would rename this parameter to signature_lifetime or something like that. Lifetime as a parameter name is too easily confused with replica/rule/did lifetime.
+1 on config

vingar: you confused me too ! 'signed_urls_lifetime' is more explicit.

bari12 (approved):

PR #1166, opened 2018-05-15
Functional tests: update dev container build Fix #1156

### LINKED ISSUE #1156: Functional tests: dev container fails on userpass authentication
Motivation
----------
When spooling up a new dev environment using the `rucio/rucio-dev` many tests fail. The first message is an exception that authentication with username and password fails.

Modification
------------
- [x] Update Dockerfile
   - [x] Update packages installed by yum
   - [x] Install all requirements with pip (such that `tools/run_tests_docker.sh` may be used)
   - [x] Remove python-ipaddress package to handle pip dependencies
- [x] Update `rucio.cfg` to streamline with travis container
- [x] Change mysql credentials in `docker-compose.yml` to rucio.cfg

berghaus:
Functional tests: update dev container build Fix #1156

### DIFF
--- a/etc/docker/dev/rucio.cfg
+++ b/etc/docker/dev/rucio.cfg
@@ -1,14 +1,3 @@
-# Copyright European Organization for Nuclear Research (CERN)
-#
-# Licensed under the Apache License, Version 2.0 (the "License");
-# You may not use this file except in compliance with the License.
-# You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
-#
-# Authors:
-# - Mario Lassnig, <mario.lassnig@cern.ch>, 2012-2014
-# - Thomas Beermann, <thomas.beermann@cern.ch>, 2012, 2015-2016
-# - Cedric Serfon, <cedric.serfon@cern.ch>, 2013
-
 [common]
 logdir = /var/log/rucio
 loglevel = DEBUG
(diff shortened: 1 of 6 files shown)


berghaus:
Something odd is going on here. One of the tests consistently seems to end up with authentication issues during the test, while the other all pass. I had seen these authentication issues in a misconfigured dev container in the past ... I wonder if I can reproduce this.

vingar:
I vaguely remember something like that due to wrong CA certificate(not the same on the server) in the configuration and had to fix it for the travis test.

bari12:
I re-executed the test now and it passed. Maybe this is one of the occasional hickups we have on travis sometime...

berghaus:
Thanks Martin!

etc/docker/dev/rucio.cfg
```diff
@@ -53,9 +37,9 @@ gss_identity = ddmlab@CERN.CH
 gss_email = ph-adp-ddm-lab@cern.ch
 
 [monitor]
-carbon_server = localhost
+carbon_server = rucio-graphite-int.cern.ch
```
mlassnig: won't this spam our graphite server with random people firing up dev containers?

berghaus: Indeed. What is the best way to turn this off entirely for the dev container? Remove the config or point to localhost?

mlassnig: leaving it at localhost is fine. maybe later we can even include graphite in the dev container (so people can actually look at the metrics). i think @tbeerman already did that.

tbeerman: yes, I've played around with it. You can just add the `graphiteapp/graphite-statsd` image in docker-compose and link it. That should do it.

berghaus: Thanks @tbeerman I'll do that :-)

bari12: Ping 😄

vingar (approved):

Issue #1177, opened 2018-05-22
FTS3 transfertool: proxy delegation function


dciangot:
Motivation
----------
Avoid using cronjobs for periodically delegating proxy to FTS3 and, more generally, make delegation function available within the daemons. 

Modification
------------
Include fts3 proxy delegation function in fts3 transfertool using fts3 python bindings.

rcarpa:
Should we re-open this one ?

rcarpa:
@dchristidis

PR #1178, opened 2018-05-22
Transfers: proxy delegation function in FTS3 transfertool #1177

dciangot:
This include FTS3 proxy delegation function in the corresponding transfertool using FTS3 python bindings (added in the pip requirements).

### DIFF
--- a/lib/rucio/transfertool/fts3.py
+++ b/lib/rucio/transfertool/fts3.py
@@ -11,7 +11,9 @@
 # - Wen Guan, <wen.guan@cern.ch>, 2014-2016
 # - Martin Barisits, <martin.barisits@cern.ch>, 2017-2018
 # - Eric Vaandering, <ewv@fnal.gov>, 2018
+# - Diego Ciangottini <diego.ciangottini@pg.infn.it>, 2018
 
+from __future__ import absolute_import
 import datetime
 import json
 import logging
@@ -27,6 +29,8 @@ from requests.packages.urllib3 import disable_warnings  # pylint: disable=import
 from dogpile.cache import make_region
 from dogpile.cache.api import NoValue
 
+from fts3.rest.client.easy import Context, delegate
+from fts3.rest.client.exceptions import BadEndpoint, ClientError, ServerError
 from rucio.common.config import config_get, config_get_bool
 from rucio.core.monitor import record_counter, record_timer
 from rucio.db.sqla.constants import FTSState
@@ -41,9 +45,6 @@ logging.basicConfig(stream=sys.stdout,
                                              raise_exception=False,
                                              default='DEBUG').upper()),
                     format='%(asctime)s\t%(process)d\t%(levelname)s\t%(message)s')
-
-USERCERT = config_get('conveyor', 'usercert', False, None)
-USE_DETERMINISTIC_ID = config_get_bool('conveyor', 'use_deterministic_id', False, False)
 
 REGION_SHORT = make_region().configure('dogpile.cache.memory',
                                        expiration_time=1800)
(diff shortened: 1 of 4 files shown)


lib/rucio/transfertool/fts3.py
```diff
@@ -60,17 +61,61 @@ def __init__(self, external_host):
...
-            self.cert = (USERCERT, USERCERT)
+            self.cert = (usercert, usercert)
             self.verify = False
         else:
             self.cert = None
             self.verify = True  # True is the default setting of a requests.* method
 
     # Public methods part of the common interface
 
+    def delegate_proxy(self, proxy, ca_path='/etc/grid-security/certificates/', duration_hours=48, timeleft_hours=12):
```
mlassnig: can we set the defaults to duration_hours=96 and timeleft_hours=72? That should be better for weekend scenarios.

lib/rucio/transfertool/fts3.py
```diff
@@ -60,17 +61,61 @@ def __init__(self, external_host):
...
+        :param timeleft_hours: int, optional
+
+        :return: delegation ID
+        :rtype: str
+        """
+
+        logging.info("Delegating proxy %s to %s", proxy, self.external_host)
+
+        try:
+            context = Context(self.external_host,
```
mlassnig: can you add a call to graphite with the core.monitor so we can watch when/what is being delegated

lib/rucio/transfertool/fts3.py
dciangot: sure no problem

lib/rucio/transfertool/fts3.py
dciangot: need to take a look at core.monitor, but I see no problem in adding it

dciangot:
@mlassnig please check if the graphite part is what you meant.

lib/rucio/transfertool/fts3.py
```diff
@@ -60,17 +61,62 @@ def __init__(self, external_host):
...
+                              ucert=proxy,
+                              ukey=proxy,
+                              verify=True,
+                              capath=ca_path)
+            delegation_id = delegate(context,
+                                     lifetime=datetime.timedelta(hours=duration_hours),
+                                     delegate_when_lifetime_lt=datetime.timedelta(hours=timeleft_hours))
+            record_timer('transfertool.fts3.delegate_proxy.%s' % proxy, (time.time() - start_time))
```
mlassnig: looks good!

can you maybe add extend it like this:

transfertool.fts3.delegate_proxy.success.%s

and below for the exceptions

transfertool.fts3.delegate_proxy.fail.%s

?

dciangot: sure, done

mlassnig (dismissed):

tools/pip-requires-test
```diff
@@ -23,3 +23,4 @@ pytz==2017.3                # World timezone definitions, modern and historical
 Babel==2.5.1                # Internationalization utilities - Dependency of sphinx
 subprocess32==3.2.7; python_version <= '3.0' # A backport of the subprocess module from Python 3.2/3.3 for use on 2.x.
 pycodestyle==2.3.1          # New package replacing pep8
+fts3-rest-API==3.7.1        # FTS rest API
```
bari12: This should be moved to pip-requires as well, as it is not only used for testing.

bari12 (changes requested):

tools/pip-requires-test
vingar: yes. 
side comment It brings M2Crypto and pycurl. Not the most pythonic dependency..

dciangot: mmm, it is already in pip-requires, isn't it (*)?
Regarding the dependencies I could agree, but we also did agree to push for having these bindings from FTS people. So I don't know how you prefer to proceed at this point.

(*)https://github.com/rucio/rucio/pull/1178/files#diff-3f6c71b838e7a93232a1c4b91bb669f7R44

bari12: Ooops 🙄 Sorry 😄 
In that case you can remove it from pip-requires-test as pip-requires will be installed for sure anyway.

dciangot: np ^^ Removed from test then.

requirements.readthedocs.txt
```diff
@@ -56,6 +56,9 @@ bz2file>=0.98                                     # Read and write bzip2-compres
 python-magic>=0.4.15                              # File type identification using libmagic
 futures>=3.2.0                                    # Clean single-source support for Python 3 and 2
 six>=1.11.0                                       # Python 2 and 3 compatibility utilities
+fts3-rest-API==3.7.1                              # FTS rest API
+MyProxyClient==2.0.1                              # myproxy client bindings
```
bari12: You did not add MyProxyClient to pip-requires (And I guess it should not be for the moment) - but in that case it can be removed from the requirements.readthedocs.txt too.

dciangot: @bari12  right, that's not used here. Removed.

bari12 (approved):

PR #1182, opened 2018-05-23
RSE manager: handle RSEChecksumUnavailabale depending on RSE settings Fix #1149

### LINKED ISSUE #1149: Raise Meaningful Exception When Protocol Fails To Get Checksum
Motivation
----------
In issue #847 I handled an exception to respect the `verify_checksum` flag with the rucio cleint. Unfortunately the `protocol.stat` function raises a `ServiceUnavailable` when the endpoint does not return a checksum. In this case the client should raise a more specific exception that can be be handled in the rsemanager.

Modification
------------
* [x] In `rucio.common.exceptions` define `ChecksumUnavailable` exception.
* [x] In `stat` function of protocols raise `ChecksumUnavailable` In place of  `ServiceUnavailable` when the checksum query fails.
* [x] In `rsemanager` catch `ChecksumUnavailable` and depending on `verify_checksum` flag:
   * set file as valid if verify_checksum is false
   * else raise `ServiceUnavailable` with same message

berghaus:
RSE manager: handle RSEChecksumUnavailabale depending on RSE settings Fix #1149

### DIFF
--- a/lib/rucio/common/exception.py
+++ b/lib/rucio/common/exception.py
@@ -851,11 +851,11 @@ class NotAllFilesUploaded(RucioException):
         self.error_code = 80
 
 
-class RSEChecksumUnavailabale(RucioException):
+class RSEChecksumUnavailable(RucioException):
     """
     Cannot retrieve checksum from RSE
     """
     def __init__(self, *args, **kwargs):
-        super(RSEChecksumUnavailabale, self).__init__(*args, **kwargs)
-        self._message = "RSE Checksum unavailabale."
+        super(RSEChecksumUnavailable, self).__init__(*args, **kwargs)
+        self._message = "RSE checksum unavailable."
         self.error_code = 81
--- a/lib/rucio/common/exception.py
+++ b/lib/rucio/common/exception.py
@@ -849,3 +849,13 @@ class NotAllFilesUploaded(RucioException):
         super(NotAllFilesUploaded, self).__init__(args, kwargs)
         self._message = "Not all of the given files have been uploaded."
         self.error_code = 80
+
+
+class RSEChecksumUnavailabale(RucioException):
+    """
+    Cannot retrieve checksum from RSE
+    """
+    def __init__(self, *args, **kwargs):
+        super(RSEChecksumUnavailabale, self).__init__(*args, **kwargs)
+        self._message = "RSE Checksum unavailabale."
+        self.error_code = 81
--- a/lib/rucio/rse/rsemanager.py
+++ b/lib/rucio/rse/rsemanager.py
@@ -38,6 +38,7 @@ except ImportError:
     from urllib.parse import urlparse
 
 from rucio.common import exception, utils, constants
+from rucio.common.constraints import STRING_TYPES
 from rucio.common.utils import make_valid_did
 
 
@@ -317,7 +318,7 @@ def exists(rse_settings, files):
     files = [files] if not type(files) is list else files
     for f in files:
         exists = None
-        if isinstance(f, str) or isinstance(f, type(u'')):
+        if isinstance(f, STRING_TYPES):
             exists = protocol.exists(f)
             ret[f] = exists
         elif 'scope' in f:  # a LFN is provided
@@ -420,7 +421,7 @@ def upload(rse_settings, lfns, source_dir=None, force_pfn=None, force_scheme=Non
                             valid = stats['adler32'] == lfn['adler32']
                         if (valid is None) and ('filesize' in stats) and ('filesize' in lfn):
                             valid = stats['filesize'] == lfn['filesize']
-                    except exception.RSEChecksumUnavailabale as e:
+                    except exception.RSEChecksumUnavailable as e:
                         if rse_settings['verify_checksum'] is False:
                             valid = True
                         else:
--- a/lib/rucio/rse/rsemanager.py
+++ b/lib/rucio/rse/rsemanager.py
@@ -317,7 +317,7 @@ def exists(rse_settings, files):
     files = [files] if not type(files) is list else files
     for f in files:
         exists = None
-        if (type(f) is str) or (type(f) is unicode):
+        if isinstance(f, str) or isinstance(f, type(u'')):
             exists = protocol.exists(f)
             ret[f] = exists
         elif 'scope' in f:  # a LFN is provided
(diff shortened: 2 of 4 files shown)


lib/rucio/rse/rsemanager.py
```diff
@@ -317,7 +317,7 @@ def exists(rse_settings, files):
     files = [files] if not type(files) is list else files
     for f in files:
         exists = None
-        if (type(f) is str) or (type(f) is unicode):
+        if isinstance(f, str) or isinstance(f, type(u'')):
```
berghaus: Flake8 complained here, so I added this. seems OK. Not pretyt with the `type(u'')` though ...

lib/rucio/rse/rsemanager.py
```diff
@@ -414,13 +414,17 @@ def upload(rse_settings, lfns, source_dir=None, force_pfn=None, force_scheme=Non
...
                 valid = None
                 try:  # Get metadata of file to verify if upload was successful
-                    stats = protocol.stat('%s.rucio.upload' % pfn)
-                    if ('adler32' in stats) and ('adler32' in lfn):
-                        valid = stats['adler32'] == lfn['adler32']
-                    if (valid is None) and ('filesize' in stats) and ('filesize' in lfn):
-                        valid = stats['filesize'] == lfn['filesize']
-                except NotImplementedError:
-                    valid = True if rse_settings['verify_checksum'] is False else False
+                    try:
```
berghaus, line 418: The only other way I could think of doing this is an `isinstance` in the except clause, these nested try's seemed cleaner.

lib/rucio/common/exception.py
```diff
@@ -849,3 +849,13 @@ def __init__(self, *args, **kwargs):
...
         self.error_code = 80
+
+
+class RSEChecksumUnavailabale(RucioException):
+    """
+    Cannot retrieve checksum from RSE
+    """
+    def __init__(self, *args, **kwargs):
+        super(RSEChecksumUnavailabale, self).__init__(*args, **kwargs)
+        self._message = "RSE Checksum unavailabale."
```
mlassnig: typo ;-)

mlassnig: oh, just realised that the class itself has the same typo

lib/rucio/rse/rsemanager.py
mlassnig: isinstance("asdfasdf", basestring)

tests for either string or unicode

TWAtGH: I think basestring doesn't exist in python 3

mlassnig: yep, in py3 string doesn't exist anymore ;-)

for py3 we only do isinstance(..., unicode)

lib/rucio/common/exception.py
berghaus: Herp, time for me to go fix some typos.

lib/rucio/rse/rsemanager.py
berghaus: In py3 only str exists, in py2 str and unicode, hence (one of the) falke8 complaints. How about:

```python
if sys.version_info.major == 3:
    string_types = (str,)
else:
    string_types = (str, unicode)
if isinstance(f, string_types):
```

berghaus: Mind you flake still complains about this: `F821 undefined name 'unicode'`

berghaus: How about:
```python
try:
    unicode
except NameError:
    unicode = str
```

berghaus: So this is giving me a bad of a bad smell: I need to do this in each source file where we use the `unicode` keyword ... but I don't quite know how to do this as a module.

berghaus: So what I could do is make a `common/python3workaround.py` containing:
```python
try:
    unicode
except NameError:
    unicode = str
```
then in any file which uses `unicode`
```python
try:
    from rucio.common.python3workaround import unicode
except ImportError:
    pass
```
Long live exceptions?!

vingar: we did something like that in some part of the code:
```
try:
    AUTHORIZED_VALUE_TYPES = (float, int, unicode)
except NameError:
    AUTHORIZED_VALUE_TYPES = (float, int, str)
```
generating the possible type values and then check against them

TWAtGH:
```
if sys.version_info[0] == 2:
	strclass = basestring
elif sys.version_info[0] == 3:
	strclass = str
else:
	raise
```
Wondering why people in Python always try to replaces if's with exceptions.

vingar: it's actually the recommended way. I read several times this advice.

TWAtGH: Only because a lot of people do it or it's marked as solution on stackoverflow it doesn't need to be recommended :D

vingar: ... mentioned with some motivation about the performance ;) but I doubt I can convince you anyway ;)

berghaus: @TWAtGH this is python [EAFP](https://docs.python.org/3/glossary.html#term-eafp) style.

TWAtGH: Yes I know but this code will only be executed once per start of a Rucio process. So this is no place were we should prefer performance over readability. Also EAFP should only be used when you know that a specific case is very exceptional and you don't know this in our case do you?

berghaus: At the moment, the rare case is running in python3. Plus the exception avoids adding the sys module.

TWAtGH: Not adding the sys module is a good point but I would guess that it's very unlikely that sys is not already loaded into memory, since it gets imported in some core areas of Rucio.
Again this is no point where performance matters because the impact is nearly nonexistent. Even if performance would matter the except cases should be really really rare.
The impact on the readability and maintainability would be huge though. One can immediately see that you're doing a differentiation between Python versions there. And think about adding more stuff that needs to be distinguished. With your approach you would end up adding a lot of try/excepts.

vingar: Feature detection is already at various places of the code, following this:
https://docs.python.org/3/howto/pyporting.html#use-feature-detection-instead-of-version-detection

I would prefer to be consistent 

About unicode:
https://docs.python.org/3/howto/pyporting.html#text-versus-binary-data

mlassnig (approved):

Issue #501, opened 2018-01-19
Traceback error in rucio upload


vingar:
Motivation
----------
```
[root@9c1d917353f9 /]# rucio -v --verbose upload --scope user.ewv --rse T2_US_NEBRASKA --name /store/data/Run2017A/SingleMuon/MINIAOD/PromptReco-v3/000/296/980/00000/3C8C839C-3256-E711-AD6B-02163E01A717.root /tmp/test.root 
2018-01-19 15:50:03,264 DEBUG   Extracting filesize (2161712084) and checksum (86e5e384) for file user.ewv:/store/data/Run2017A/SingleMuon/MINIAOD/PromptReco-v3/000/296/980/00000/3C8C839C-3256-E711-AD6B-02163E01A717.root
2018-01-19 15:50:03,267 DEBUG   Automatically setting new GUID
2018-01-19 15:50:03,346 DEBUG   Using account ewv
2018-01-19 15:50:03,346 DEBUG   Skipping dataset registration
2018-01-19 15:50:03,346 DEBUG   Processing file user.ewv:/store/data/Run2017A/SingleMuon/MINIAOD/PromptReco-v3/000/296/980/00000/3C8C839C-3256-E711-AD6B-02163E01A717.root for upload
2018-01-19 15:50:03,364 INFO    Adding replicas in Rucio catalog
2018-01-19 15:50:03,404 INFO    Replicas successfully added
2018-01-19 15:50:03,404 INFO    Adding replication rule on RSE T2_US_NEBRASKA for the file user.ewv:/store/data/Run2017A/SingleMuon/MINIAOD/PromptReco-v3/000/296/980/00000/3C8C839C-3256-E711-AD6B-02163E01A717.root
2018-01-19 15:50:06,514 ERROR   list index out of range
2018-01-19 15:50:06,515 ERROR   Rucio exited with an unexpected/unknown error. Please rerun the last command with the '-v' option and submit a ticket with all the available information to reproduce this error at https://its.cern.ch/jira/browse/RUCIO
2018-01-19 15:50:06,519 DEBUG   Traceback (most recent call last):
  File "/usr/bin/rucio", line 168, in new_funct
    return function(*args, **kwargs)
  File "/usr/bin/rucio", line 1082, in upload
    trace['protocol'] = rse_settings['protocols'][0]['scheme']
IndexError: list index out of range
Completed in 8.8367 sec.
```
Modification
------------

Expected behavior
-----------------

vingar:
[...]

Issue #468, opened 2018-01-12
Automatic protocol failover for upload


bari12:
Motivation
----------
Similar to `rucio download` there should be a protocol failover mechanism in `rucio upload`, if the process does not work. This can probably be c/p in large parts.

mlassnig:
rucio download first tries on the same site with different protocols, then hops to another site and tries the protocols available there. for the upload, we only want the different protocols, we don't want to automatically upload to another site.

nikmagini:
Comment about the c/p suggestion: are you sure that the protocol failover in `rucio download` actually works? From a few tests and looking at the code, I don't think it's doing anything... AFAICT only the site failover works in `rucio download` but the protocol failover will always select the default (though hard to test for sure since I need to find a site where protocol 1 fails and then protocol 2 works...)

mlassnig:
pinging @TWAtGH , the download one should work with both protocol and rse failover. something for monday ;-)

TWAtGH:
You mean because the protocol is assigned to the 'read_protocol' attribute and then 'read_protocol' is ignored afterwards? I think you're right but for any reason its working in my tests:

>rucio -v download --ndownloader 1 user.mlassnig:user.mlassnig.pilot.test.multi.hits
2018-01-22 13:30:06,722	DEBUG	Thread 1/1 : 2 possible protocol(s) for read
2018-01-22 13:30:06,722	DEBUG	Thread 1/1 : Trying protocol davs at IN2P3-LAPP-MWTEST_DATADISK
2018-01-22 13:30:06,722	INFO	Thread 1/1 : File mc15_14TeV:HITS.10075481._000455.pool.root.1 trying from IN2P3-LAPP-MWTEST_DATADISK
2018-01-22 13:31:06,812	WARNING	The requested service is not available at the moment.
Details: An unknown exception occurred.
Details: Could not open source: Connection timed out
2018-01-22 13:31:06,812	DEBUG	Thread 1/1 : Failed attempt 1/2
2018-01-22 13:31:06,812	INFO	Thread 1/1 : File mc15_14TeV:HITS.10075481._000455.pool.root.1 trying from IN2P3-LAPP-MWTEST_DATADISK
2018-01-22 13:32:06,899	WARNING	The requested service is not available at the moment.
Details: An unknown exception occurred.
Details: Could not open source: Connection timed out
2018-01-22 13:32:06,899	DEBUG	Thread 1/1 : Failed attempt 2/2
[...]

Issue #29, opened 2017-11-08
Option to specify protocol implementation for upload


bari12:
_Migrated from JIRA [RUCIO-2349](https://its.cern.ch/jira/browse/RUCIO-2349)_

TWAtGH:
Hi, if you don't mind I will take this too, since I need it for the new api upload.

nikmagini:
Hi,

sure, sorry for the delay. Maybe you can simply reuse my changes to rsemanager.py here:

https://github.com/rucio/rucio/commit/7883be7685d906b3edbb4dc9dfdb098980ad8968#diff-8777b68ebcb342123d7ce30a3ff039b5

Cheers
N.

TWAtGH:
Ah great! Thanks for  the hint!

bari12:
Done in bin/rucio, but not in uploadclient

TWAtGH:
It is
https://github.com/rucio/rucio/blob/master/lib/rucio/client/uploadclient.py#L218
https://github.com/rucio/rucio/blob/master/lib/rucio/client/uploadclient.py#L276
https://github.com/rucio/rucio/blob/master/lib/rucio/client/uploadclient.py#L296

bari12:
So this has to be prepared in the list of dicts given to the uploadclient?

TWAtGH:
Exactly. Most options are file specific in the API.

bari12:
I see, thanks 👍 . Closing

PR #1194, opened 2018-05-25
Clients: using upload api in bin/rucio Fix #670 #468 #501 #29

### LINKED ISSUE #670: Switch bin/rucio upload to uploadclient
Motivation
----------
As the uploadclient is implemented now, the upload code in bin/rucio should use the uploadclient.

TWAtGH:
Clients: using upload api in bin/rucio Fix #670 #468 #501 #29

### DIFF
--- a/lib/rucio/tests/test_bin_rucio.py
+++ b/lib/rucio/tests/test_bin_rucio.py
@@ -25,7 +25,7 @@
 
 from __future__ import print_function
 
-from os import remove, unlink, listdir, rmdir, stat
+from os import remove, unlink, listdir, rmdir, stat, path
 
 import nose.tools
 import re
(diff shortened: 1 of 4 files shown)


mlassnig (commented):
looks good to me, but need a second pair of eyes :-) @bari12

lib/rucio/tests/test_bin_rucio.py
```diff
@@ -259,14 +264,16 @@ def test_upload_repeated_file(self):
         remove(tmp_file1)
         remove(tmp_file2)
         remove(tmp_file3)
-        nose.tools.assert_not_equal("File %s:%s successfully uploaded on the storage" % (self.user, tmp_file1[5:]) in out, None)
+        nose.tools.assert_true((self.upload_success_str % tmp_file1_name) in out)
```
nikmagini, line 267: Why was this changed from assert_not_equal to assert_true? Did the logic change for this case? (IIUC trying to upload a file already in the catalog is not an error anymore?)

nikmagini (commented):
Looks OK to me - I just have one comment about a change in the test logic

lib/rucio/tests/test_bin_rucio.py
TWAtGH: No, probably a copy paste error. Thanks!

TWAtGH: But this was wrong anyway wasn't it? Assuming the `... in out` expression returns either True or False, it's always not equal to None.

bari12 (approved):

lib/rucio/tests/test_bin_rucio.py
nikmagini: I agree that the old test logic didn't make sense because the test was always returning success. IIUC with the new logic, the test will return OK if the file is successfully uploaded after being previously removed - if this is the intended test behavior then it looks fine now.

PR #1209, opened 2018-05-30
clients: workaround for url signatures in download; Fix #1208

### LINKED ISSUE #1208: clients: workaround url signature for rucio download until download-api is deployed
Motivation
----------

The old method for rucio download does not use the list_replicas PFNs, and instead constructs them manually. This workaround allows to use the list_replicas PFNs when necessary and will be removed again once we switch to download-api next month.

mlassnig:
clients: workaround for url signatures in download; Fix #1208


TWAtGH (approved):
This code will be removed when we switch to the download API with the next feature release.

Issue #1137, opened 2018-05-08
xrdcp transfer a single file part of a zip file


bari12:
Motivation
----------
The CLI should be able to download a content file of an archive via xroot.

At first the client should be able to do this by explicitly parameterizing both, the content file name as well as the archive name. In a second step, once the list-replicas issue #1138 is done the client should be able to do this transparently by just giving the content file name.

bari12:
The code for bin/rucio @TWAtGH will add.

PR #1237, opened 2018-06-06
#1137 - extracting files from archive

TomasJavurek:
#1137 - extracting files from archive

### DIFF
--- a/lib/rucio/client/downloadclient.py
+++ b/lib/rucio/client/downloadclient.py
@@ -98,106 +98,92 @@ class DownloadClient:
         """
         logger = self.logger
         trace = copy.deepcopy(self.trace_tpl)
-        log_prefix = 'Extracting files : '
+        log_prefix = 'Extracting files: '
 
         logger.info('Processing %d item(s) for input' % len(items))
-        input_items = []
         for item in items:
             archive = item.get('archive')
             file_extract = item.get('did')
             rse_name = item.get('rse')
+            if not archive or not file_extract:
+                raise InputValidationError('File DID and archive DID are mandatory')
+            if '*' in archive:
+                logger.debug(archive_did_str)
+                raise InputValidationError('Cannot use PFN download with wildcard in DID')
+
             file_extract_scope, file_extract_name = self._split_did_str(file_extract)
             archive_scope, archive_name = self._split_did_str(archive)
 
-            # if not specified, taking first availbale replica
-            if not rse_name:
-                replicas = self.client.list_replicas([{'scope': archive_scope, 'name': archive_name}],
-                                                     schemes='root',
-                                                     rse_expression='istape=False',
-                                                     unavailable=False,
-                                                     client_location=detect_client_location())
-                for r in replicas:
-                    for rse in r['states'].keys():
-                        if r['states'][rse] == 'AVAILABLE':
-                            rse_name = rse
-                            break
-                    if rse_name and rse_name is not None:
-                        break
+            # listing all available replicas of given archhive file
+            rse_expression = 'istape=False' if not rse_name else '(%s)&istape=False' % rse_name
+            archive_replicas = self.client.list_replicas([{'scope': archive_scope, 'name': archive_name}],
+                                                  schemes=['root'],
+                                                  rse_expression=rse_expression,
+                                                  unavailable=False,
+                                                  client_location=detect_client_location())
 
-            rse = {}
-            try:
-                rse = rsemgr.get_rse_info(rse_name)
-            except RSENotFound:
-                logger.warning('%sCould not get info of RSE %s' % (log_prefix, rse_name))
-                continue
 
-            logger.debug('Preparing pfn of archive for download of %s (%s) from %s' % (archive, file_extract, rse))
-            pfn_archive = ''
-            try:
-                pfns_dict = rsemgr.lfns2pfns(rse,
-                                             lfns={'name': archive_name, 'scope': archive_scope},
-                                             operation='read',
-                                             scheme='root')
-                pfn_archive = pfns_dict[archive]
-            except:
-                raise InputValidationError('The pfn of archive file not found.')
-
-            if '*' in archive:
-                logger.debug(archive)
-                raise InputValidationError('Cannot use PFN download with wildcard in DID')
+            # preparing trace
+            trace['uuid'] = generate_uuid()
+            trace['scope'] = archive_scope
+            trace['dataset'] = archive_name
+            trace['filename'] = file_extract
 
             # preparing output directories
             dest_dir_path = self._prepare_dest_dir(item.get('base_dir', '.'),
                                                    os.path.join(archive_scope, archive_name + '.extracted'), file_extract,
                                                    item.get('no_subdir'))
-            logger.debug('Preparing output destination %s' % dest_dir_path)
+            logger.debug('%sPreparing output destination %s' % (log_prefix, dest_dir_path))
 
-            item['file_extract'] = file_extract_name
-            item['scope_archive'] = archive_scope
-            item['name_archive'] = archive_name
-            item['rse'] = rse_name
-            item['pfn_archive'] = pfn_archive
-            item['force_scheme'] = pfn_archive.split(':')[0]
-            item['dest_dir_path'] = dest_dir_path
-            item.setdefault('ignore_checksum', True)
-            input_items.append(item)
+            # validation and customisation of list of replicas
+            archive_pfns = []
+            replicas = next(archive_replicas)
+            for rse in replicas['rses']:
+                archive_pfns.extend(replicas['rses'][rse])
+                if len(archive_pfns) == 0:
+                    raise InputValidationError('No PFNs for replicas of archive %s' % archive)
+            archive_pfns.reverse()
 
-        for item in input_items:
-            trace['uuid'] = generate_uuid()
-            trace['scope'] = item['scope_archive']
-            trace['dataset'] = item['name_archive']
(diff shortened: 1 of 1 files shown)


lib/rucio/client/downloadclient.py
```diff
@@ -73,6 +73,134 @@ def __init__(self, _client=None, user_agent='rucio_clients', logger=None):
...
+                  clientState can be one of the following: ALREADY_DONE, DONE, FILE_NOT_FOUND, FAIL_VALIDATE, FAILED
+
+        :raises InputValidationError: if one of the input items is in the wrong format
+        :raises NoFilesDownloaded: if no files could be downloaded
+        :raises NotAllFilesDownloaded: if not all files could be downloaded
+        :raises RucioException: if something unexpected went wrong during the download
+        """
+        logger = self.logger
+        trace = copy.deepcopy(self.trace_tpl)
+        log_prefix = 'Unarchiving'
```
mlassnig: "Extracting" maybe better? :-D

lib/rucio/client/downloadclient.py
```diff
@@ -73,6 +73,134 @@ def __init__(self, _client=None, user_agent='rucio_clients', logger=None):
...
+            except RSENotFound:
+                logger.warning('%sCould not get info of RSE %s' % (log_prefix, rse_name))
+                continue
+
+                raise InputValidationError('The keys archive (zip file), file_extract, and rse are mandatory')
+
+            logger.debug('Preparing pfn of archive for download of %s (%s) from %s' % (archive, file_extract, rse))
+            pfn_archive = ''
+            try:
+                pfns_dict = rsemgr.lfns2pfns(rse,
```
mlassnig: no need to build the pfn again? it comes already from list_replica

lib/rucio/client/downloadclient.py
```diff
@@ -73,6 +73,134 @@ def __init__(self, _client=None, user_agent='rucio_clients', logger=None):
...
+                for r in replicas:
+                    for rse in r['states'].keys():
+                        if r['states'][rse] == 'AVAILABLE':
+                            rse_attr = self.client.list_rse_attributes(rse)
+                            if rse_attr['istape'] == 'False':
+                                rse_name = rse
+                                break
+                    if rse_name and rse_name is not None:
+                        break
+
```
TWAtGH: You can use list_replicas with `unavailable=False` to exclude unavailable replicas
and set an rse_expression to `'istape=False' if not rse else '(%s)&istape=False' % rse` to exclude tape endpoints

lib/rucio/client/downloadclient.py
```diff
@@ -73,6 +73,134 @@ def __init__(self, _client=None, user_agent='rucio_clients', logger=None):
...
+            rse = {}
+            try:
+                rse = rsemgr.get_rse_info(rse_name)
+            except RSENotFound:
+                logger.warning('%sCould not get info of RSE %s' % (log_prefix, rse_name))
+                continue
+
+                raise InputValidationError('The keys archive (zip file), file_extract, and rse are mandatory')
```
TWAtGH: this raise will probably not be reached

lib/rucio/client/downloadclient.py
```diff
@@ -73,6 +73,134 @@ def __init__(self, _client=None, user_agent='rucio_clients', logger=None):
...
+                status, out, err = execute(cmd)
+                end_time = time.time()
+                trace['transferStart'] = start_time
+                trace['transferEnd'] = end_time
+                if status == 54:
+                    trace['clientState'] = 'FAILED'
+                    raise SourceNotFound()
+                elif status != 0:
+                    trace['clientState'] = 'FAILED'
+                    raise RucioException(err)
```
cserf: Propagate the xrdcp error

lib/rucio/client/downloadclient.py
```diff
@@ -73,6 +73,134 @@ def __init__(self, _client=None, user_agent='rucio_clients', logger=None):
...
+                else:
+                    trace['clientState'] = 'DONE'
+            except Exception as e:
+                trace['clientState'] = 'FAILED'
+                raise ServiceUnavailable(e)
+            send_trace(trace, self.client.host, self.user_agent)
+
```
cserf: No checksum validation ?

lib/rucio/client/downloadclient.py
```diff
@@ -73,6 +73,134 @@ def __init__(self, _client=None, user_agent='rucio_clients', logger=None):
...
+                  clientState can be one of the following: ALREADY_DONE, DONE, FILE_NOT_FOUND, FAIL_VALIDATE, FAILED
+
+        :raises InputValidationError: if one of the input items is in the wrong format
+        :raises NoFilesDownloaded: if no files could be downloaded
+        :raises NotAllFilesDownloaded: if not all files could be downloaded
+        :raises RucioException: if something unexpected went wrong during the download
+        """
+        logger = self.logger
+        trace = copy.deepcopy(self.trace_tpl)
+        log_prefix = 'Extracting'
```
TWAtGH: log_prefix is just used once and it's missing a whitespace at the end. You could write it directly in the message or do you plan to use it more often later?

lib/rucio/client/downloadclient.py
TomasJavurek: from some reason pfns is not populated, and also I would need to do mapping between pfn and rse:

output of list_file_replicas:

{u'adler32': u'48e8ce74', u'name': u'DAOD_BPHY5.13781702._000102.zip.1', u'rses': {u'CERN-PROD_DATADISK': [], u'BNL-OSG2_DATADISK': [], u'UNI-FREIBURG_DATADISK': [], u'PRAGUELCG2_DATADISK': []}, u'bytes': 1106077679, u'states': {u'CERN-PROD_DATADISK': u'AVAILABLE', u'BNL-OSG2_DATADISK': u'AVAILABLE', u'UNI-FREIBURG_DATADISK': u'AVAILABLE', u'PRAGUELCG2_DATADISK': u'AVAILABLE'}, u'pfns': {}, u'scope': u'data16_valid', u'md5': None}

TWAtGH: Don't know why PFNs are not populated but I don't think you would need this mapping.
Either you get the RSE from the input options or you try to use RSE after RSE. List replicas should already give them in a sorted order.

lib/rucio/client/downloadclient.py
TomasJavurek: unavailable seems to be False by default, then perhaps also line:
 if r['states'][rse] == 'AVAILABLE':
seems to be useless. But only if understand corrctly tha availability of site = availability of replica.

lib/rucio/client/downloadclient.py
TomasJavurek: ok, but for now I leave rather as it is, since it is working. Once we figure out, why pfns are not populated, we can remove the lfns2pfns part from the code.

lib/rucio/client/downloadclient.py
TWAtGH: You are not using a site state here are you?
I'm pretty sure `r['states'][rse]` is related to the replica `r`
And `unavailable=False` also:
https://github.com/rucio/rucio/blob/master/lib/rucio/client/replicaclient.py#L101

TomasJavurek: If rse_name is defined, I am not calling list_replicas at all. Thus I left only:
rse_expression='istape=False'

lib/rucio/client/downloadclient.py
TomasJavurek: the raise is redundant anyway, and rse is actualy optional. I removed this line.

lib/rucio/client/downloadclient.py
TomasJavurek: we have no metadata for files inside archive yet, I'll add checksum validation as soon as we have metadata.

lib/rucio/client/downloadclient.py
TomasJavurek: I placed: SourceNotFound(err)

lib/rucio/client/downloadclient.py
TomasJavurek: Yes I do. Once we add transparent replica listing at some far future.

lib/rucio/client/downloadclient.py
mlassnig: if pfns are empty, then the protocol does not exist for the RSE.

and i now see the bug:      `schemes='root'` should be  `schemes=['root']`

lib/rucio/client/downloadclient.py
TomasJavurek: I am also thinking about some validation of input at the first place. log_prefix is now: 'Extracting files : '

lib/rucio/client/downloadclient.py
```diff
@@ -73,6 +73,118 @@ def __init__(self, _client=None, user_agent='rucio_clients', logger=None):
...
+                logger.debug(archive_did_str)
+                raise InputValidationError('Cannot use PFN download with wildcard in DID')
+
+            file_extract_scope, file_extract_name = self._split_did_str(file_extract)
+            archive_scope, archive_name = self._split_did_str(archive)
+
+            # listing all available replicas of given archhive file
+            rse_expression = 'istape=False' if not rse_name else '(%s)&istape=False' % rse_name
+            archive_replicas = self.client.list_replicas([{'scope': archive_scope, 'name': archive_name}],
+                                                  schemes=['root'],
```
mlassnig: indentation looks wrong?

TWAtGH (approved):

Issue #1179, opened 2018-05-22
Features for user data transfers in CMS


dciangot:
Motivation
----------
CMS data transfers for user analysis outputs need additional features for theirs integration:
- user proxy retrieval from myproxy.cern.ch
- user FTS proxy delegation
- keep central processing conveyor separate from the user one

Modification
------------
- creation of a transfertool FTS3 subclass supporting:
-- get user DN starting from scope name
-- get proxy from myproxy:  key = hash(username+some key from configuration)
-- use fts3 python bindings for submitting transfers and delegating proxy only if needed
- for certain activities and if specified in the Rucio configuration, allow conveyor to act in "user mode":
-- split payload by scopes in order to separate connections to FTS server by user
-- submit transfers using the FTS3 subclass created above

PR #1239, opened 2018-06-08
Transfers: Features for user data transfers integration in CMS #1179

dciangot:
CMS data transfers for user analysis outputs need additional features for theirs integration:

- user proxy retrieval from myproxy.cern.ch
- user FTS proxy check and eventual delegation is done via FTS bindings at job submission time with a single call
- keep central processing conveyor separate from the user one by activities


lib/rucio/core/transfer.py
```diff
@@ -45,9 +46,10 @@
 REGION_SHORT = make_region().configure('dogpile.cache.memcached',
                                        expiration_time=600,
                                        arguments={'url': "127.0.0.1:11211", 'distributed_lock': True})
+USER_TRANSFERS = config_get('conveyor', 'user_transfers', False, None)
 
 
-def submit_bulk_transfers(external_host, files, transfertool='fts3', job_params={}, timeout=None):
+def submit_bulk_transfers(external_host, files, transfertool='fts3', job_params={}, timeout=None, user_transfer_job=False):
```
vingar, line 52: user_transfer_job is to give the mapping with a transfertool implementation. Won't it be possible to extend the usage of the transfertool field for this purpose ? transfertool='FTS' by default. This will give more flexibility. In that case, the value to be propagated will be transfertool='fts_user_cms'  for example.

lib/rucio/core/transfer.py
```diff
@@ -45,9 +46,10 @@
 REGION_SHORT = make_region().configure('dogpile.cache.memcached',
                                        expiration_time=600,
                                        arguments={'url': "127.0.0.1:11211", 'distributed_lock': True})
+USER_TRANSFERS = config_get('conveyor', 'user_transfers', False, None)
```
vingar, line 49: user transfers here is a configuration option. user transfer is also an activity which is attached with the transfer. The mapping activity-> transfertool implementation can be done possibly upstream imo. see my next comment. Unless I miss something this line can be removed here ?

lib/rucio/core/transfer.py
```diff
@@ -76,7 +78,13 @@ def submit_bulk_transfers(external_host, files, transfertool='fts3', job_params=
                 else:
                     job_file[key] = file[key]
             job_files.append(job_file)
-        transfer_id = FTS3Transfertool(external_host=external_host).submit(files=job_files, job_params=job_params, timeout=timeout)
+        if not user_transfer_job:
```
vingar, line 81: As mentioned before, the transfertool field there is more relevant for me. The mapping transfertool label <-> implementation can be better defined in a dictionary but can stay hardcoded with if for now.

lib/rucio/daemons/conveyor/common.py
```diff
@@ -45,7 +49,9 @@ def submit_transfer(external_host, job, submitter='submitter', process=0, thread
 
     # prepare submitting
     xfers_ret = {}
+
     try:
+        print job
```
vingar: print statement to remove

lib/rucio/transfertool/cms_fts3.py
```diff
@@ -0,0 +1,358 @@
+# Copyright European Organization for Nuclear Research (CERN)
```
vingar: I wrote the script ./tools/generate_header to get the header decided within ATLAS and the right apache v2 text. I would propose that you use it for now..

lib/rucio/transfertool/cms_fts3.py
```diff
@@ -0,0 +1,358 @@
...
+                                  config_get('common', 'loglevel',
+                                             raise_exception=False,
+                                             default='DEBUG').upper()),
+                    format='%(asctime)s\t%(process)d\t%(levelname)s\t%(message)s')
+
+REGION_SHORT = make_region().configure('dogpile.cache.memory',
+                                       expiration_time=1800)
+
+
+class CMSUserTransfer(FTS3Transfertool):
```
mlassnig: In principle this is a generic transfertool that happens to deal with MyProxy. Maybe better to call it something like `FTS3MyProxyTransfertool` or something similar? Essentially making CMS the first "client" of this transfertool.

lib/rucio/transfertool/cms_fts3.py
```diff
@@ -0,0 +1,358 @@
...
+            logging.error("Wrong FTS endpoint: %s", self.external_host)
+            return None
+
+        if jobID:
+            record_counter('transfertool.fts3.%s.submission.success' % self.__extract_host(self.external_host), len(files))
+            transfer_id = jobID
+        else:
+            transfer_id = jobID
+            logging.error("Unexpected failure during FTS job submission.")
+            record_counter('transfertool.fts3.%s.submission.failure' % self.__extract_host(self.external_host), len(files))
```
mlassnig: a separate subcategory for these might be useful, e.g., `transfertool.fts3myproxy.%s.submission` ?

lib/rucio/transfertool/cms_fts3.py
dciangot: well, the key definition for retrieving the proxy from myproxy is quite CRAB bound, nevertheless I think it can be overwritten in other subclasses if needed. So agreed.

lib/rucio/daemons/conveyor/common.py
```diff
@@ -164,7 +172,21 @@ def bulk_group_transfer(transfers, policy='rule', group_bulk=200, fts_source_str
...
+            grouped_transfers[external_host] = {}
+            if USER_TRANSFERS not in ['cms'] or activity not in USER_ACTIVITY:
+                grouped_jobs[external_host] = []
+            elif activity in USER_ACTIVITY:
+                grouped_jobs[external_host] = {}
+                if scope not in grouped_transfers[external_host]:
+                    grouped_transfers[external_host][scope] = {}
+                    grouped_jobs[external_host][scope] = []
+
+        job_params = {'verify_checksum': True if file['checksum'] and file['metadata'].get('verify_checksum', True) else False,
```
mlassnig: you should just use the `verify_checksum` variable that was created a few lines above here. otherwise you'll nullify the `source`/`destination`/`both` checksum capability

lib/rucio/core/transfer.py
dciangot: @vingar Just to keep track and summarize the discussion on slack about this point. The idea here is to optimize the mapping at a second round of iteration (dedicated issue for that? what do you think?) since it involves the splitting of transfers by scope that could be potentially dangerous to include for the production conveyor without proper tests. Is this correct?

lib/rucio/transfertool/cms_fts3.py
dciangot: @vingar actually I just copied the transfertool/fts3.py one and I cannot find the generate_header script. Can you send me the link for that?

vingar: it's in the repo `./tools/generate_header`

`./tools/generate_header  <file>`

lib/rucio/core/transfer.py
vingar: I think you are right ;) it should come after and possibly iteratively.

lib/rucio/transfertool/cms_fts3.py
mlassnig: almost ;-D  `tools/add_header`

dciangot: oh, I found it I think. But it is add_header right? https://github.com/rucio/rucio/blob/master/tools/add_header

vingar: yes, that's correct.

lib/rucio/core/transfer.py
bari12: 👍 Fully agree to the iterative implementation.

lib/rucio/daemons/conveyor/common.py
dciangot: right, changing it.

lib/rucio/transfertool/cms_fts3.py
dciangot: ok, I agree, no problem

lib/rucio/core/transfer.py
dciangot: I would keep this change on the same line of activity mapping comment. So, putting it on a dedicated ticket for the second iteration to avoid forgetting it, while not pushing for the inclusion on this PR.

dciangot:
@mlassnig @bari12 @vingar can someone re-trigger the checks? It seems to me like something failed again for code-unrelated issues.

bari12:
Done ;)

vingar (approved):
this can go for further integration

bari12:
👍 I am merging this now. Let's create some further issues for the changes we want to address at a later point.

PR #1243, opened 2018-06-13
Client: Add a 600s timeout to all client commands; Fix #1204

### LINKED ISSUE #1204: Add a timeout to rucio client commands
Motivation
----------
Protection on Apache with a 10 minute timeout, thus it makes sense to put a 10 minute client timeout.

bari12:
Client: Add a 600s timeout to all client commands; Fix #1204


bari12:
`E: 75, 4: Unable to import 'requests_kerberos' (import-error)`
It fails one syntax test, but I think that one can be ignored. Not sure why this is even coming up as the extra is not installed.

vingar:
There is a typo. Ask tobi

bari12:
The typo is not the problem though, as the error is raised at a much earlier point in the code (also when the typo is fixed the error is still raised)

bari12:
I think pylint is just doing something weird here ignoring the conditional.

mlassnig (approved):

PR #1246, opened 2018-06-14
core: correct replica sorting across wan and lan; Fix #668

### LINKED ISSUE #668: Sorting of replicas in list_replicas

mlassnig:
core: correct replica sorting across wan and lan; Fix #668

### DIFF
--- a/lib/rucio/core/replica.py
+++ b/lib/rucio/core/replica.py
@@ -757,6 +757,10 @@ def _list_replicas(dataset_clause, file_clause, state_clause, show_pfns, schemes
                 if rse not in rse_info:
                     rse_info[rse] = rsemgr.get_rse_info(rse, session=session)
 
+                # assign scheme priorities
+                rse_info[rse]['priority_wan'] = {p['scheme']: p['domains']['wan']['read'] for p in rse_info[rse]['protocols']}
+                rse_info[rse]['priority_lan'] = {p['scheme']: p['domains']['lan']['read'] for p in rse_info[rse]['protocols']}
+
                 # select the lan door in autoselect mode, otherwise use the wan door
                 if domain is None:
                     domain = 'wan'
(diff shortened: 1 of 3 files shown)


lib/rucio/tests/test_replica_sorting.py
```diff
@@ -0,0 +1,195 @@
...
+                                 'prefix': '//lambda/complex/',
+                                 'impl': 'rucio.rse.protocols.xrootd.Default',
+                                 'domains': {
+                                     'lan': {'read': 1, 'write': 1, 'delete': 1},
+                                     'wan': {'read': 3, 'write': 3, 'delete': 3}}})
+
+        replicas = [r for r in self.rc.list_replicas(dids=[{'scope': 'mock',
+                                                            'name': f['name'],
+                                                            'type': 'FILE'} for f in self.files],
+                                                     schemes=['root', 'gsiftp', 'davs'],
```
bari12, line 98: What happens when schemes=None or schemes=[]?

lib/rucio/core/replica.py
```diff
@@ -861,13 +868,14 @@ def _list_replicas(dataset_clause, file_clause, state_clause, show_pfns, schemes
...
+                    file['rses'][rse] += list(set([tmp_pfn[0] for tmp_pfn in pfns]))  # extract properly the pfn from the (pfn, domain, priority)
                     file['states'][rse] = str(state)
                     for tmp_pfn in pfns:
-                        file['pfns'][tmp_pfn[1]] = {'rse': rse,
+                        file['pfns'][tmp_pfn[0]] = {'rse': rse,
                                                     'type': str(rse_type),
                                                     'volatile': volatile,
-                                                    'domain': tmp_pfn[0]}  # extract properly the domain from the (domain, pfn)
+                                                    'domain': tmp_pfn[1],
+                                                    'priority': tmp_pfn[2]}
```
bari12, line 878: I wonder if priority shouldn't be a total order. Because now you repeat 1-N for every RSE, but for the client it would be easier to have a total order I think.

lib/rucio/tests/test_replica_sorting.py
mlassnig: you will only get the highest ranking protocol per domain

lib/rucio/core/replica.py
mlassnig: like in metalink?

bari12: Yes basically. Because otherwise a client doesn't really know how to order the RSEs. Feels like we only get half the way we want to go in this case.

mlassnig: It's correctly sorted though, but it will make Tobi's life easier I agree ;-D

bari12: Is it? I thought all the PFNs are in one dictionary and the priority repeats 1-N?

mlassnig: ah, now I see what you mean. yes, it's sorted per domain.

bari12:
👍 Just a note to my comment: Mario and I discussed the total order and we both agreed that this would beneficial, but it will come in a later patch.

bari12 (approved):

Issue #1222, opened 2018-06-03
rucio download does not respect protocol order


mlassnig:
Motivation
----------

The order of selected protocols is not correct in rucio download. (Using list_replicas directly produces the correct order, however)

TWAtGH:
This is not a bug but a missing feature. Rucio download **does** respect the order but it's not able to determine if lan or wan should be used. Instead it always uses the default (=wan).

Issue #669, opened 2018-02-15
Switch bin/rucio download to downloadclient


bari12:
Motivation
----------
As the downloadclient is implemented now, the download code in bin/rucio should use the downloadclient.

TWAtGH:
This will also fix this issue
#567 
#490 
#1222

Issue #567, opened 2018-01-31
bin/rucio upload/download still have a lot of return FAILURE statements


TWAtGH:
Motivation
----------
Pilot2 test cases are parsing the output of rucio download in case of errors. The tests are failing because there are still cases in rucio download where return FAILURE is used instead of raising the appropriate exception.

Modification
------------
Change return FAILURE statements to raising the correct exceptions.

Expected behavior
-----------------

bari12:
As discussed today, this should probably be coordinated with @TomasJavurek as he is already working on the download function.

PR #1254, opened 2018-06-15
Clients: switched bin/rucio download to API Fix #490 #567 #669 #1222 #1263

### LINKED ISSUE #490: Rucio download randomly shuffles the rses, even if they are retrieved in a sorted way
Motivation
----------
The shuffling makes sense if no client location can be detected, to distributed the load, however, if a client location can be detected, the list_replicas call returns the replica locations already in a sorted order and they must not be shuffled.

TWAtGH:
Clients: switched bin/rucio download to API Fix #490 #567 #669 #1222

### DIFF
--- a/lib/rucio/client/downloadclient.py
+++ b/lib/rucio/client/downloadclient.py
@@ -19,42 +19,28 @@
 # - Nicolo Magini <nicolo.magini@cern.ch>, 2018
 # - Tobias Wegner <tobias.wegner@cern.ch>, 2018
 
-import os
-import os.path
 
 import copy
 import logging
+import os
+import os.path
 import random
-import time
 import socket
 import signal
-
-from rucio.common.exception import (InputValidationError, NoFilesDownloaded, ServiceUnavailable,
-                                    NotAllFilesDownloaded, RSENotFound, RucioException, SourceNotFound)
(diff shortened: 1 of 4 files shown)


lib/rucio/client/downloadclient.py
```diff
@@ -646,6 +559,387 @@ def download(self, dids, rse=None, archive=None, protocol=None, pfn=None, nrando
...
+
+            logger.debug('Processing: %s' % item)
+
+            # since we are using metalink we need to explicitly
+            # give all schemes (probably due to a bad server site implementation)
+            force_scheme = item.get('force_scheme')
+            if force_scheme:
+                schemes = force_scheme if isinstance(force_scheme, list) else [force_scheme]
+            else:
+                schemes = ['davs', 'gsiftp', 'https', 'root', 'srm', 'file']
```
bari12, line 896: As just discussed, so that this is documented here too. This hardcoding of schemes is bad. I understand that this is how the `list_replicas` with `metalink=True` requires it, but it really shouldn't. The default behaviour of list_replicas should be if schemes = [] is given, all protocols should be supplied. That's the only way how we can make proper protocol failover work. Due to release considerations, I am in favour of merging this, but we should address this eventually. See #1262

bari12 (approved):
I am approving this as we need it in the 1.17.0 release, but due to the length it is difficult to review ;)

Issue #716, opened 2018-02-20
json based import for RSEs, protocols, distances


bari12:
Motivation
----------
To have a structured way of importing RSEs, their protocols and even distances a json based importer would be beneficial.

mlassnig:
JJ "volunteered", can you give him more details pls.

jwackito:
Should this be a rucio-admin command? I was thinking about something like 
```
rucio-admin rse export RSE_NAME -o/--output rse.json (default to rse_name.json)
rucio-admin rse import rse_name.json
```
Which attributes should be mandatory, in order to make the method fail proof?
For distances, maybe an update method that support json files will do the trick.
`rucio-admin update-distances --from-json distances.json`
May also work for protocols.

bari12:
The original idea of this came from Ale and what he had in mind was really a full importer. Thus essentially everything (rses, attributes, protocols, distances) can be represented as one big json and this can be imported by rucio to update/create/delete etc. I think we will need some further face2face discussion on this.

hahahannes:
I would propose to add the endpoints /import and /export endpoints to the API where the admin can upload the full json file. Instead of different rucio-admin commands, there would been one command  like `rucio-admin import data.json`.

mlassnig:
Other option, maybe as `/rse/configuration` with associated GET, POST, and DELETE?

hahahannes:
I think a more general endpoint like `/configuration` is better if we want to add more export functionality in the future. For example with users or scopes if this would be useful later.

bari12:
I think a separate endpoint might be better. It gives us more flexibility to widen the scope of the importer/exporter later on.

hahahannes:
What endpoint name should we use in this case? Maybe `configuration` is a bit confusing as there already is an endpoint `config`.

Also I would like to reopen the PR [https://github.com/rucio/rucio/pull/1259](https://github.com/rucio/rucio/pull/1259) to use the export_distances() function.

PR #1259, opened 2018-06-17
Core: Distance core export function #716

jwackito:
Core: Distance core export function #716

### DIFF
--- a/lib/rucio/core/distance.py
+++ b/lib/rucio/core/distance.py
@@ -14,7 +14,7 @@
 from sqlalchemy.exc import DatabaseError, IntegrityError
 
 from rucio.common import exception
-from rucio.db.sqla.models import Distance
+from rucio.db.sqla.models import Distance, RSE
 from rucio.db.sqla.session import transactional_session, read_session
 
 
@@ -139,5 +139,33 @@ def update_distances(src_rse_id=None, dest_rse_id=None, parameters=None, session
         if dest_rse_id:
             query = query.filter(Distance.dest_rse_id == dest_rse_id)
         query.update(params)
+    except IntegrityError as error:
+        raise exception.RucioException(error.args)
+
+
+@read_session
+def export_distances(session=None):
+    """
+    Get distances between all the rses.
+
+    :param session: The database session to use.
+
+    :returns distance: dictionary of dictionaries with all the distances.
+    """
+
+    rsemap = {}
+    for k, v in session.query(RSE.id, RSE.rse).all():
+        rsemap[k] = v
+    distances = {}
+    try:
+        query = session.query(Distance)
+        for distance in query.all():
+            src_name = rsemap[distance.src_rse_id]
+            dst_name = rsemap[distance.dest_rse_id]
+            key = src_name + '__' + dst_name
+            distances[key] = {}
+            for k, v in distance:
+                distances[key][k] = v
+        return distances
     except IntegrityError as error:
         raise exception.RucioException(error.args)


lib/rucio/core/distance.py
```diff
@@ -141,3 +142,28 @@ def update_distances(src_rse_id=None, dest_rse_id=None, parameters=None, session
...
+    distances = {}
+    try:
+        query = session.query(Distance)
+        for distance in query.all():
+            src_name = get_rse_name(distance.src_rse_id)
+            dst_name = get_rse_name(distance.dest_rse_id)
+            key = src_name + '__' + dst_name
+            distances[key] = {}
+            for k, v in distance:
+                distances[key][k] = v
```
vingar: a join with the rse table is more efficient.

vingar (changes requested):

lib/rucio/core/distance.py
```diff
@@ -141,3 +142,28 @@ def update_distances(src_rse_id=None, dest_rse_id=None, parameters=None, session
         query.update(params)
     except IntegrityError as error:
         raise exception.RucioException(error.args)
+
+
+@read_session
+def export_distances(session=None):
```
vingar, line 147: def list_distances

lib/rucio/core/distance.py
jwackito: This is not to get the distance between two sites, but all the distances. I need to do a full table scan anyways.

lib/rucio/core/distance.py
jwackito: No, is export_distances as export_rses. What is wrong is the comment. It should say "Export all the distance table". This will be the methods used for export RSE/distances from Rucio in json format from the clients.

lib/rucio/core/distance.py
bari12: What Vincent means is that you should join in the RSE table to resolve the rse_ids directly with the name. Because this way you do basically O(#rse ^ 2) queries to just resolve the rse ids to names (sans caching).

vingar: it's more about the performance. For each rse_id, there is two selects in the rse table to get dest and src. With joins, this will speed up the total time and will benefit from db internal optimization. Just have a look on rse.py and how to do a double joins to to get the src and dest name directly in one query.

jwackito: Aaaaaah... I see.. I will fix it. Thanks for the clarification!

lib/rucio/core/distance.py
vingar: list_distances to list all distances works as well, :) no ? and looks more coherent with the current conventions (list_dids, list_rses, etc)

lib/rucio/core/distance.py
jwackito: Solved. I created a local map from RSE.id --> RSE.name so only query RSE table once.

lib/rucio/core/distance.py
```diff
@@ -141,3 +141,31 @@ def update_distances(src_rse_id=None, dest_rse_id=None, parameters=None, session
...
+
+    :returns distance: dictionary of dictionaries with all the distances.
+    """
+
+    rsemap = {}
+    for k, v in session.query(RSE.id, RSE.rse).all():
+        rsemap[k] = v
+    distances = {}
+    try:
+        query = session.query(Distance)
```
vingar, line 161:
```
from sqlalchemy.orm import aliased
sources = aliased(models.RSE)
destinations = aliased(models.RSE)
query = session.query(sources.rse, destinations.rse, Distance).\
   join(sources, sources.id == Distance.src_rse_id).\
   join(destinations, destinations.id == Distance.dest_rse_id)
```

lib/rucio/core/distance.py
```diff
@@ -141,3 +141,31 @@ def update_distances(src_rse_id=None, dest_rse_id=None, parameters=None, session
...
+        for distance in query.all():
+            src_name = rsemap[distance.src_rse_id]
+            dst_name = rsemap[distance.dest_rse_id]
+            key = src_name + '__' + dst_name
```
vingar, line 165: A matrix in a dictionary is more intuitive, i.e, `dict[src][dst] = distances dict`

lib/rucio/core/distance.py
vingar: list_distances can be also extended with a filters={} parameters that can be used later on.

vingar (changes requested):

PR #1260, opened 2018-06-18
core: archive support for list replicas; Fix #1138

### LINKED ISSUE #1138: Transparent list-replicas support of zip contents
Motivation
----------
When querying for a did which is also content of an archive, list-replicas should return the did of the archive file in a way that the client can download the file. (e.g. use xroot to directly download the content of a zip)

mlassnig:
core: archive support for list replicas; Fix #1138

### DIFF
--- a/lib/rucio/core/replica.py
+++ b/lib/rucio/core/replica.py
@@ -38,7 +38,7 @@ from sqlalchemy.sql.expression import case, bindparam, select, text, false, true
 import rucio.core.lock
 
 from rucio.common import exception
-from rucio.common.utils import chunks, clean_surls, str_to_date
+from rucio.common.utils import chunks, clean_surls, str_to_date, add_url_query
 from rucio.core.credential import get_signed_url
 from rucio.core.rse import get_rse, get_rse_id, get_rse_name, get_rse_attribute, get_rses_with_attribute_value
 from rucio.core.rse_counter import decrease, increase
@@ -580,9 +580,10 @@ def _resolve_dids(dids, unavailable, ignore_availability, all_states, session):
                 archive = session.query(models.ConstituentAssociation.scope,
                                         models.ConstituentAssociation.name)\
                                  .filter(models.ConstituentAssociation.child_scope == scope,
-                                         models.ConstituentAssociation.child_name == name).one()
-                constituents['%s:%s' % (scope, name)] = {'scope': archive[0],
-                                                         'name': archive[1]}
+                                         models.ConstituentAssociation.child_name == name)\
+                                 .with_hint(models.ConstituentAssociation, "INDEX(ARCHIVE_CONTENTS ARCH_CONTENTS_PK)", 'oracle')\
+                                 .all()
+                constituents['%s:%s' % (scope, name)] = [{'scope': tmp[0], 'name': tmp[1]} for tmp in archive]
 
             if did_type == DIDType.FILE:
                 files.append({'scope': scope, 'name': name})
@@ -758,12 +759,58 @@ def _list_replicas(dataset_clause, file_clause, state_clause, show_pfns,
     for replicas in filter(None, files):
         for scope, name, bytes, md5, adler32, path, state, rse, rse_type, volatile in replicas:
 
+            pfns = []
+
             # reset the domain selection to original user's choice (as this could get overwritten each iteration)
             domain = deepcopy(original_domain)
 
-            pfns = []
-            if show_pfns and rse:
+            # if the file is a constituent, find the available archives and add them to the list of possible PFNs
+            if '%s:%s' % (scope, name) in constituents:
+                archive_result = list_replicas(dids=constituents['%s:%s' % (scope, name)],
+                                               schemes=schemes, client_location=client_location,
+                                               domain=domain, sign_urls=sign_urls,
+                                               signature_lifetime=signature_lifetime,
+                                               session=session)
 
+                # is it the only instance (i.e., no RSE for the replica)?
+                # then yield and continue as the replica does not exist and we don't want to return
+                # an additional empty pfn to the client
+                if rse is None:
+                    for archive in archive_result:
+                        for tmp_archive in archive['pfns']:
+                            archive['pfns'][tmp_archive]['domain'] = 'zip'
+                            # at this point we don't know the protocol of the parent, so we have to peek
+                            if tmp_archive.startswith('root://') and 'xrdcl.unzip' not in tmp_archive:
+                                # use direct addressable path for root
+                                new_pfn = add_url_query(tmp_archive, {'xrdcl.unzip': name})
+                                archive['pfns'][new_pfn] = archive['pfns'].pop(tmp_archive)
+                                tmp_archive = new_pfn
+                            if 'xrdcl.unzip' not in tmp_archive:
+                                archive['pfns'][tmp_archive]['client_extract'] = True
+                        yield archive
+                    continue
+
+                # now, repeat the procedure slightly different in case if there are replicas available
+                # and make the archive just an additional available PFN
+                available_archives = [r['pfns'] for r in archive_result]
+                for archive in available_archives:
+                    for archive_pfn in archive:
+                        # at this point we don't know the protocol of the parent, so we have to peek
+                        if archive_pfn.startswith('root://'):
+                            # use direct addressable path for root
+                            pfn = add_url_query(archive_pfn, {'xrdcl.unzip': name})
+                            client_extract = False
+                        else:
+                            # otherwise just use the zip and tell the client to extract
+                            pfn = archive_pfn
+                            client_extract = True
+
+                        # use zip domain for sorting, and pass through all the information
+                        # we need it later to reconstruct the final PFNS
+                        # ('pfn', 'domain', 'priority', 'client_extract', archive-passthrough)
+                        pfns.append((pfn, 'zip', archive[archive_pfn]['priority'], client_extract, archive))
+
+            if show_pfns and rse:
                 if rse not in rse_info:
                     rse_info[rse] = rsemgr.get_rse_info(rse, session=session)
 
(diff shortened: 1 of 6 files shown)


lib/rucio/core/replica.py
```diff
@@ -852,6 +862,16 @@ def _list_replicas(dataset_clause, file_clause, state_clause, show_pfns, schemes
...
 
+                        # is the file part of an archive and can be extract directly?
+                        if protocol.attributes['scheme'] == 'root' and '%s:%s' % (scope, name) in constituents:
+                            # parent archive can have a different access path! need to run full PFN resolve for it
+                            archive_pfn = list(list_replicas(dids=[constituents['%s:%s' % (scope, name)]],
+                                                             schemes=['root'],
+                                                             rse_expression=rse,
+                                                             client_location=client_location,
+                                                             domain=domain))[0]['pfns'].keys()[0]
+                            pfn = '%s?xrdcl.unzip=%s' % (archive_pfn, name)
```
vingar: only for scheme == xroot. what about the other schemes ?

lib/rucio/core/replica.py
```diff
@@ -569,9 +569,21 @@ def _resolve_dids(dids, unavailable, ignore_availability, all_states, session):
...
+                                                                models.DataIdentifier.did_type,
+                                                                models.DataIdentifier.constituent)\
+                                                         .with_hint(models.DataIdentifier, "INDEX(DIDS DIDS_PK)", 'oracle')\
+                                                         .filter(or_(*did_clause)):
+            if constituent:
+                # file is a constituent, resolve to any parent archive
+                archive = session.query(models.ConstituentAssociation.scope,
+                                        models.ConstituentAssociation.name)\
+                                 .filter(models.ConstituentAssociation.child_scope == scope,
+                                         models.ConstituentAssociation.child_name == name).one()
```
vingar: + `with_hint` index to stabilize the query.

lib/rucio/core/replica.py
```diff
@@ -569,9 +569,21 @@ def _resolve_dids(dids, unavailable, ignore_availability, all_states, session):
...
+                constituents['%s:%s' % (scope, name)] = {'scope': archive[0],
+                                                         'name': archive[1]}
```
bari12: I think this is not enough. You have to get the archive somehow into the `file_clause`.
Otherwise the replicas are resolved in line 732, but essentially for most constituent files there will be no replica resolved, as there is none (it is only part of the archive).
So the whole resolving has to be done with the archive (pointing to the constituent)

lib/rucio/core/replica.py
mlassnig: Only root supports direct extraction, so it only makes sense here.

One option could be to just return the archive.zip for all other protocols, but then the client has to realise that it didn't get the file via root and has to to manually unzip.

vingar: and What about, the `rucio-download` use case, e.g.,`rucio-download  MyScope.file.zip` ?

mlassnig: `list_replicas` will construct the right PFN, for `rucio download` it will be transparent

vingar: To wrap up: one option is that `list_replicas` gives back nested urls (`http://example.org/package.zip#path=path/to/the/file` or `http://example.org/package.zip#path/to/the/file`) for non root schemes and that `rucio donwload` does the extraction client side ?

bari12: \# might be a problem if it is part of the filename?
In any way, it does not need to be nested but can just be part of the dictionary. There is already multiple keys in there, we can add another one.

vingar: Certain clients like xrdcp might support nested urls directly.  An option of `list_replicas`  for the format of the zip file url can be relevant as well.

bbockelm: It might be worthwhile to use URL parsing libraries to parse out the URL, then add the `xrdcl.unzip` option, then serialize it back to a string.

Direct manipulation of URLs as strings always seems to end up in tears (or, at least, multiple `?` and/or inadvertently repeated options).

mlassnig: I'm rewriting this part already ;-)

lib/rucio/core/replica.py
```diff
@@ -569,9 +569,22 @@ def _resolve_dids(dids, unavailable, ignore_availability, all_states, session):
...
+                                                                models.DataIdentifier.constituent)\
+                                                         .with_hint(models.DataIdentifier, "INDEX(DIDS DIDS_PK)", 'oracle')\
+                                                         .filter(or_(*did_clause)):
+            if constituent:
+                # file is a constituent, resolve to any parent archive
+                archive = session.query(models.ConstituentAssociation.scope,
+                                        models.ConstituentAssociation.name)\
+                                 .filter(models.ConstituentAssociation.child_scope == scope,
+                                         models.ConstituentAssociation.child_name == name)\
+                                 .with_hint(models.ConstituentAssociation, "INDEX(ARCHIVE_CONTENTS ARCH_CONTENTS_PK)", 'oracle')\
```
vingar, line 584: Did you check how I define the PK index with child_scope, child_name first to avoid range scans ?  I think I did to avoid the secondary reverse index but better to check.

lib/rucio/core/replica.py
```diff
@@ -748,12 +759,58 @@ def _list_replicas(dataset_clause, file_clause, state_clause, show_pfns, schemes
...
+            if '%s:%s' % (scope, name) in constituents:
+                archive_result = list_replicas(dids=constituents['%s:%s' % (scope, name)],
+                                               schemes=schemes, client_location=client_location,
+                                               domain=domain, sign_urls=sign_urls,
+                                               signature_lifetime=signature_lifetime,
+                                               session=session)
+
+                # is it the only instance (i.e., no RSE for the replica)?
+                # then yield and continue as the replica does not exist and we don't want to return
+                # an additional empty pfn to the client
```
vingar, line 777: It's not clear if this applies only to nested archive files, so it's better to clarify this comment: The pfn entry without replicas is there also to report the completness of a collection(dataset, container). It is compatible ?

lib/rucio/core/replica.py
```diff
@@ -868,15 +924,25 @@ def _list_replicas(dataset_clause, file_clause, state_clause, show_pfns, schemes
...
                                                     'domain': tmp_pfn[1],
                                                     'priority': tmp_pfn[2]}
+                        # instruct the client that the archive needs to be manually extracted
+                        if tmp_pfn[3]:
+                            file['pfns'][tmp_pfn[0]]['client_extract'] = True
                 else:
+                    # quick exit, but don't forget to set the total order for the priority
+                    # --> exploit that L(AN) comes before W(AN) before Z(IP) alphabetically
+                    # and use 1-indexing to be compatible with metalink
+                    tmp = sorted([(file['pfns'][p]['domain'], file['pfns'][p]['priority'], p) for p in file['pfns']])
```
vingar, line 943: In general, I think that too much business logic is applied to list_replicas which can have some non negligible impacts on the maintenance and performance. I'm in favor of splitting the db access: get all possible information from the db as fast as possible and then the ordering part, url constructions, etc. the later can be applied on top with some pluggable decorators for example and possibly sometimes client side.

lib/rucio/core/replica.py
mlassnig: child comes first, yes. But as a file can be registered to multiple zips, we have to get all of them.

lib/rucio/core/replica.py
mlassnig: this is only in the archive workflow, and zips can never be datasets, only files, so it's ok.

lib/rucio/core/replica.py
vingar: yes, I think I assumed that getting parent zip files was a more popular use case than getting archive content. Better to keep an eye on it wrt wfms access patterns.

lib/rucio/core/replica.py
mlassnig: (and i still think that the empty pfn reply for collection completeness was a very bad design, we should get rid of that in the future)

lib/rucio/core/replica.py
mlassnig: well, that's what it is. getting the stuff from the database is in a two different functions.
this one here is the business logic function (and we can't do anything on the client side). my principal problem here is that list_replicas by itself returns a non-optimal datastructure that's why it's so complex. this could be much better once we introduce a list_replicas2 or something, as soon as we've settled on the major use cases (we already added location handling, proxy handling, and archive handling)

lib/rucio/core/replica.py
mlassnig: in my tests it was ok, let's see how it scales, i'd like to avoid another index on replicas :-D

lib/rucio/core/replica.py
vingar: Which options would you propose then ? 
Resolving the content first and creating a dictionary with all entries (present or not) is intuitive no ? It does the job to report the completeness efficiently in one method without having to look-up the content separately. Feature requested by the wfms..

lib/rucio/core/replica.py
vingar: yes, I agree on the structure returned by list_replicas. Historically, the structure had to change for the pilot needs. Which optimal structure would you propose ?

lib/rucio/core/replica.py
vingar: The concerned table is archive_content, not the replicas table. No?

mlassnig: yes, of course, i typed too quickly

lib/rucio/core/replica.py
mlassnig: to me, it breaks what list_replicas should do (= listing replicas) :-D

another call, is_complete(did) or similar would probably be the way go. i have to think about it.

lib/rucio/core/replica.py
mlassnig: i have a few ideas, i'll post a ticket

lib/rucio/core/replica.py
vingar: It gives also the details of which files is there or not for a collection at a site. Now if the collection concept is considered sub-optimal...

lib/rucio/core/replica.py
vingar: Ok putting this PR on hold then

bari12: We should progress on this PR; this cannot really wait until list_replicas2 is out.
I don't really think that too much stuff is happening in list_replicas, I would make the argument that at least ordering is a central part of the list_replicas method. Other things (url construction etc. is maybe different). Without having done any measurements, I don't think that we loose too much performance on the ordering part.

bari12:
I know there are some ongoing discussions about this, but I am merging this now so we can test this feature still in June, as we promised.
There should not be any impact on the normal list_replicas calls, only the ones involving files in archive should be affected. Lets address the discussions in an issue.

bari12 (approved):

PR #1300, opened 2018-06-27
RSE Manager: handle RSEChecksumUnavailable when renaming is false Fix #1299

### LINKED ISSUE #1299: Handle RSEChecksumUnavailable for protocols without renaming
Motivation
----------

For #1149 we introduced the `RSEChecksumUnavailable` to make the rucio upload client respect the `verify_checksum` RSE flag. The same exception handling should be done for protocols that do not support renaming, i.e. when `protocol.renaming == False`

Modification
------------
 - [x] `rse.rsemanager` handles `RSEChecksumUnavailable` when `protocol.renaming == False`

berghaus:
RSE Manager: handle RSEChecksumUnavailable when renaming is false Fix #1299

### DIFF
--- a/lib/rucio/rse/rsemanager.py
+++ b/lib/rucio/rse/rsemanager.py
@@ -460,13 +460,17 @@ def upload(rse_settings, lfns, source_dir=None, force_pfn=None, force_scheme=Non
 
                 valid = None
                 try:  # Get metadata of file to verify if upload was successful
-                    stats = protocol.stat(pfn)
-                    if ('adler32' in stats) and ('adler32' in lfn):
-                        valid = stats['adler32'] == lfn['adler32']
-                    if (valid is None) and ('filesize' in stats) and ('filesize' in lfn):
-                        valid = stats['filesize'] == lfn['filesize']
-                except NotImplementedError:
-                    valid = True  # If the protocol doesn't support stat of a file, we agreed on assuming that the file was uploaded without error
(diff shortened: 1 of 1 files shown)


lib/rucio/rse/rsemanager.py
```diff
@@ -460,13 +460,17 @@ def upload(rse_settings, lfns, source_dir=None, force_pfn=None, force_scheme=Non
...
                 valid = None
                 try:  # Get metadata of file to verify if upload was successful
-                    stats = protocol.stat(pfn)
-                    if ('adler32' in stats) and ('adler32' in lfn):
-                        valid = stats['adler32'] == lfn['adler32']
-                    if (valid is None) and ('filesize' in stats) and ('filesize' in lfn):
-                        valid = stats['filesize'] == lfn['filesize']
-                except NotImplementedError:
-                    valid = True  # If the protocol doesn't support stat of a file, we agreed on assuming that the file was uploaded without error
+                    try:
```
mlassnig, line 463: can't you just pull this exception block one level up?

berghaus: The reason I did this nesting is that I wanted to re-raise the exception when `verify_checksum == True` so the standard exception handling can take over. Of, course it is possible to copy the code from there into the `else` part of that if statement.

mlassnig (approved):

PR #1303, opened 2018-07-02
METADATA: Adding generic meta support for oracle

asketagarwal:
METADATA: Adding generic meta support for oracle

### DIFF
--- a/lib/rucio/core/did.py
+++ b/lib/rucio/core/did.py
@@ -23,10 +23,10 @@
 # - Joaquin Bogado <jbogado@linti.unlp.edu.ar>, 2014-2015
 # - Wen Guan <wguan.icedew@gmail.com>, 2015
 
+import json
 import logging
 import random
 import sys
-import json
 
 from datetime import datetime, timedelta
 from hashlib import md5
--- a/lib/rucio/core/did.py
+++ b/lib/rucio/core/did.py
@@ -1288,6 +1288,11 @@ def get_did_meta(scope, name, session=None):
     """
     if session.bind.dialect.name == 'sqlite':
         raise NotImplementedError
+    if session.bind.dialect.name == 'oracle':
+        oracle_version = int(session.connection.connection.version.split('.')[0])
+        if oracle_version < 12:
+            raise NotImplementedError
+
     try:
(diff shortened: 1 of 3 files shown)


lib/rucio/core/did.py
```diff
@@ -26,6 +26,7 @@
 import logging
 import random
 import sys
+import json
```
mlassnig: sort list of imports

lib/rucio/db/sqla/types.py
```diff
@@ -113,5 +113,7 @@ def load_dialect_impl(self, dialect):
             return dialect.type_descriptor(JSONB())
         elif dialect.name == 'mysql':
             return dialect.type_descriptor(types.JSON())
+        elif dialect.name == 'oracle':
+        	return dialect.type_descriptor(CLOB())
```
mlassnig: lib/rucio/db/sqla/types.py:117:9: E101 indentation contains mixed spaces and tabs
lib/rucio/db/sqla/types.py:117:9: W191 indentation contains tabs

mlassnig:
for my understanding, this works by just storing the JSON as a clob in oracle? and we would need oracle12 working to get full JSON support, right?

asketagarwal:
This does not work yet, when you see the oracle test you see it fails as it can't create the table. Oracle 12 is needed for that. Basically for json in oracle there needs to be a constraint on the clob type with IS JSON keyword which is not there in oracle 11

vingar (approved):

mlassnig (approved):

lib/rucio/core/did.py
```diff
@@ -1285,12 +1286,17 @@ def get_did_meta(scope, name, session=None):
     :param scope: the scope of did
     :param name: the name of the did
     """
-    if session.bind.dialect.name == 'oracle' or session.bind.dialect.name == 'sqlite':
+    if session.bind.dialect.name == 'sqlite':
         raise NotImplementedError
+    if session.bind.dialect.name == 'oracle':
```
bari12, line 1291: Just an idea for later: why might think about doing these db compatibility checks via a function decorator?

Issue #1307, opened 2018-07-04
Update oracle xe image for travis to 12c


tbeerman:
Motivation
----------
Currently travis is using the 11g version of Oracle XE which doesn't support json but a new 12c version of the docker image is available is already available on docker hub.

PR #1309, opened 2018-07-04
change travis oracle version to 12c #1307

tbeerman:
change travis oracle version to 12c #1307

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,7 +50,8 @@ script:
   - if [[ $SUITE == "all" ]]; then docker exec -it rucio /bin/sh -c "/opt/rucio/tools/run_tests_docker.sh -1q" ; fi
 
 env:
-    - SUITE=all RDBMS=oracle
+    - SUITE=all RDBMS=oracle11
+    - SUITE=all RDBMS=oracle12
     - SUITE=all RDBMS=mysql
     - SUITE=all RDBMS=postgres
     - SUITE=all RDBMS=sqlite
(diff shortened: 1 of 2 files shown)


vingar (dismissed):
I remind that there are few other places with sleep statement. it would be good to have the same good approach than there. Reducing the sleep time to 10 can be considered too.

tbeerman:
For Oracle 12c the sleep time of 30 should be ok. It anyway takes a couple of minutes to start up the container. For the other dbs it can be lower. 
This PR is still WIP and shouldn't be merged, yet. The tests haven't passed, yet. It shows green but the oracle test actually never ran and already failed with the DB initialization.  Furthermore, the Oracle test should be split into two different ones: one for 11g and one for 12c to make sure that the code remains backwards compatible.

vingar:
One side comment: the json type support needs 12c and will fail with 11g. Meaning that the tests for 11g has to exclude them once merged.

vingar (dismissed):

tbeerman:
Please do not merge, the tests are still not running fine even though it showed green. Sometimes it works and sometimes it doesn't. It's not really stable. Before it just stalled during the tests and this time it failed to reset the DB. But the script still returns 0 so it looks like a success. That should be changed.

bari12:
Yes, I noticed. I restarted the tests several times; One time it ran through green, without actually running all the tests due to some db connection timeout. Will leave it open.

PR #1327, opened 2018-07-10
Clients: Optional load of paramiko; Fix #1324

### LINKED ISSUE #1324: common/utils fails when loading dysfunctional paramiko version
Motivation
----------
Paramiko needs to be imported in a try/except block.

bari12:
Clients: Optional load of paramiko; Fix #1324

### DIFF
--- a/lib/rucio/common/utils.py
+++ b/lib/rucio/common/utils.py
@@ -97,7 +97,10 @@ if EXTRA_MODULES['web']:
     from web import HTTPError
 
 if EXTRA_MODULES['paramiko']:
-    from paramiko import RSAKey
+    try:
+        from paramiko import RSAKey
+    except Exception:
+        EXTRA_MODULES['paramiko'] = False
 
 if EXTRA_MODULES['flask']:
     from flask import Response
@@ -660,7 +663,7 @@ def ssh_sign(private_key, message):
     :return: Base64 encoded signature as a string.
     """
     if not EXTRA_MODULES['paramiko']:
-        raise MissingModuleException('The paramiko module is not installed.')
+        raise MissingModuleException('The paramiko module is not installed or faulty.')
     sio_private_key = StringIO(private_key)
     priv_k = RSAKey.from_private_key(sio_private_key)
     sio_private_key.close()


lib/rucio/common/utils.py
```diff
@@ -97,7 +97,10 @@
     from web import HTTPError
 
 if EXTRA_MODULES['paramiko']:
-    from paramiko import RSAKey
+    try:
+        from paramiko import RSAKey
+    except Exception:
+        EXTRA_MODULES['paramiko'] = False
```
vingar, line 103: The package is detected in the environment but the package is faulty... logging a warning should be added like could not load the module blah blah. This would be useful when paramiko is used ,e.g.,  s3 interactions.

vingar (dismissed):

lib/rucio/common/utils.py
bari12: Will add it, I hope this is not too spammy for some reason, but I think it should only be given once on module load.

vingar: yes, just realized that all ATLAS users will see it but in certain case like when paramiko is needed it should be exposed which is also useful. opinion?

TWAtGH: Maybe just add a hint in the MissingModuleException? e.g. `raise MissingModuleException('The paramiko module is not installed or faulty.')`

vingar: The motivation of the PR is to not fail or raise exception when getting a faulty paramiko.

TWAtGH: https://github.com/bari12/rucio/blob/091879d38fb4d814873ba1ee9c6bd0eb6d1f9d54/lib/rucio/common/utils.py#L666

bari12: The problem is, that even if you install the clients without the 'ssh' extra (which requires paramiko) it is loaded anyway in utils because the existence of the package is detected. So basically, nobody wants paramiko, but we load it anyway. But we cannot detect at that point if the user asked for the extra, in which case the error should be at least displayed, or if it was just loaded due to existence, in which case it should be silent.

TWAtGH: If you fail to load it and really need it you will get an exception anyway. I talked about adding a hint to this exception not to raise a new one.
But it was just a suggestion :)

bari12: It is a bit of a weird case. Thinking about it, I am in favour of leaving it like it is in this PR, thus to exclude it silently. My thinking is that we are essentially showing an error to all ATLAS users, and they cannot do anything about it, as they have no influence on the environment on lxplus. The only downside is, that users who install the rucio-clients with [ssh], and for some reasons their paramiko package is faulty as well, they would have a hard way to find out what's going wrong.

bari12: Actually, chatting with Tobias about this, he is right, we do not need the warning when we try to load the module. For the people who actually want to use ssh auth, it will call the ssh_sign method and thus it will raise the exception in line 666. Thus I think we can just fail it silently on line 100.

vingar: when paramiko is needed in the clients and pilot for object store interactions, showing this error as soon as possible is better imo. But it should fail anyway right? Would you know which error message do we get then ? Another use case for paramiko, ssh authentication?

bari12: When actually ssh_sign is used it will fail with `raise MissingModuleException('The paramiko module is not installed.')` So the user is pointed in the right direction what is going wrong.
I agree with the fail early, but in this case, we cannot distinguish if the user wants paramiko or not. As, by far, the majority will not want it, I think it is wrong to spam them with a warning they cannot do anything about. For the ones who want it, it will fail, just not early.

mlassnig: I would argue in favour of the average user not needing to see the import error and changing the error message for the missing module to "... not found or faulty" as @TWAtGH suggested.

vingar: Same clients for the average user and the pilot. Did someone check how it fails when interacting with S3 (X2) ?

bari12: I didn't try, but I imagine for S3 it fails immediately, as paramiko is not loaded directly, but via the S3 library. Thus it would just fail when importing the S3 module.

bari12: The only place where we use paramiko directly is in utils.py. There the time between the import at the top and the time sign_ssh is called is probably very small, thus in this case, the fail early problematic is not really the case. I will change the error message in line 666 still.

mlassnig: s3 doesn't use paramiko, and boto falls back to using a non-paramiko implementation, so it's fine

vingar: strange .. it's part of the dependency https://github.com/boto/boto/blob/develop/requirements.txt

mlassnig: The import is try/excepted, it will continue without paramiko if necessary.

TWAtGH (approved):

mlassnig (approved):

Issue #1346, opened 2018-07-12
Deleting left-overs of first attempt of upload by correct protocol


TomasJavurek:
Motivation
----------
upload() in rsemgr.py tries to delete left-overs of previous attempt. But if force-scheme is set, which is almost all the time to the write_wan protocol (case of upload), the upload function try to delete these left-overs with the force-scheme protocol. But this can be different from 1st priority deletion protocol in AGIS, e.g. webdav. That's also the case of NDGF-T1

Modification
------------
https://github.com/rucio/rucio/blob/master/lib/rucio/rse/rsemanager.py#L375
->
protocol_delete = create_protocol(rse_settings, 'delete')

PR #1347, opened 2018-07-12
#1346 - upload in rsemgr will use correct protocol for deletions of left-overs

TomasJavurek:
#1346 - upload in rsemgr will use correct protocol for deletions of left-overs


bari12:
Patch is fine, just a (pedantic) reminder that the commit message format should be `<component>: <short_change_message> Fix #<issue number>` 😄

TWAtGH (approved):

PR #1387, opened 2018-07-26
Change conveyors to heartbeat (1st step) : poller

cserf:
Change conveyors to heartbeat (1st step) : poller

### DIFF
--- a/lib/rucio/core/transfer.py
+++ b/lib/rucio/core/transfer.py
@@ -189,6 +189,82 @@ def set_transfers_state(transfers, submitted_at, session=None):
 
 
 @read_session
+def get_next_transfers_new(request_type, state, limit=100, older_than=None, rse=None, activity=None,
+                           total_workers=0, worker_number=0,
+                           activity_shares=None, session=None):
+    """
+    Retrieve the next transfers matching the request type and state.
+    Workers are balanced via hashing to reduce concurrency on database.
+
+    :param request_type:      Type of the request as a string or list of strings.
+    :param state:             State of the request as a string or list of strings.
+    :param limit:             Integer of requests to retrieve.
+    :param older_than:        Only select requests older than this DateTime.
+    :param total_workers:     Number of total workers.
+    :param worker_number:     Id of the executing worker.
+    :param activity_shares:   Activity shares dictionary, with number of requests
+    :param session:           Database session to use.
+    :returns:                 List of a {external_host, external_id, request_id} dictionary.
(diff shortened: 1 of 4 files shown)


lib/rucio/daemons/conveyor/poller.py
```diff
@@ -74,29 +72,28 @@ def poller(once=False,
...
     hostname = socket.getfqdn()
     pid = os.getpid()
     hb_thread = threading.current_thread()
     heartbeat.sanity_check(executable=executable, hostname=hostname)
-    hb = heartbeat.live(executable, hostname, pid, hb_thread)
+    heart_beat = heartbeat.live(executable, hostname, pid, hb_thread)
+    prepend_str = 'Thread [%i/%i] : ' % (heart_beat['assign_thread'] + 1, heart_beat['nr_threads'])
+    logging.info(prepend_str + 'Poller starting - db_bulk (%i) fts_bulk (%i) timeout (%s)' % (db_bulk, fts_bulk, timeout))
+
+    time.sleep(10)  # To prevent running on the same partition if all the poller restart at the same time
```
mlassnig, line 84: but they will all wait 10 seconds? shouldn't it include a random time delay instead?

cserf: Yes, they all wait 10 seconds (the time that all the threads start and report their heartbeat). I'm not sure what we gain with a random time sleep.

bari12: Yes this pause is just done once so that the heartbeat done in line 80 is done by all threads and when they read it again in line 85 they all know about each other.

lib/rucio/core/transfer.py
```diff
@@ -188,6 +188,82 @@ def set_transfers_state(transfers, submitted_at, session=None):
         raise RucioException(error.args)
 
 
+@read_session
+def get_next_transfers_new(request_type, state, limit=100, older_than=None, rse=None, activity=None,
```
bari12, line 192: For me the code looks good, please just add some kind of TODO comment, both at the get_next_transfers_new and get_next_transfers that one of them will be deprecated and replaced by the other.

bari12 (changes requested):

bari12:
Also a general observation (no need to change): Please stick to the message format for the commit messages, the issue number is nowhere mentioned 😄 
For me the PR is 👍 now

bari12 (approved):

PR #1389, opened 2018-07-26
Core : Extend the bad_replicas table with bytes : Closes #1370

### LINKED ISSUE #1370: Extend the bad_replicas table with bytes
Motivation
----------

The bad_replicas table doesn't contain the bytes associated to lost or corrupted files, making it more complicated to provide summaries on lost files.

cserf:
Core : Extend the bad_replicas table with bytes : Closes #1370


etc/sql/oracle/schema.sql
```diff
@@ -1646,6 +1646,7 @@ CREATE TABLE BAD_REPLICAS(
    REASON VARCHAR2(255),
    STATE CHAR(1),
    ACCOUNT VARCHAR2(25),
+   BYTES NUMBER(19),
```
bari12, line 1649: Should there be a Not Null check as well?

bari12 (approved):

PR #1395, opened 2018-07-30
uploadclient: upload with pfn and registration to non-deter. sites Fixes #1377

### LINKED ISSUE #1377: rucio upload --no-register --pfn fails
Motivation
----------
rucio upload --no-register --pfns fails while it works with 1.16.4.

```
[wguan@lxplus067 wguan]$ /usr/bin/env rucio -v upload --rse CERN-PROD_ES --scope transient --no-register --pfn s3://cs3.cern.ch:443//atlas-eventservice/05cc36b8-383f-444f-a384-592fad3b918e /bin/hostname
2018-07-25 14:54:14,552 WARNING Upload with given pfn implies that no_register is True
2018-07-25 14:54:14,658 INFO    Preparing upload for file hostname
2018-07-25 14:54:16,608 ERROR   An unknown exception occurred.
Details: no error information passed (http status code: 500 ('internal_server_error', 'server_error', '/o\\', '\xe2\x9c\x97'))
Completed in 2.1462 sec.
```

The problem is related to rsemanager.exists call, which trys to resolve the did on the server with lfns2pfns, which does not work as the replica does not exists. In 1.16.4 with bin/rucio it was first checked if the did exists, and if not, rsemanager.exists was not even called.

TomasJavurek:
uploadclient: upload with pfn and registration to non-deter. sites
Fixes #1377

### DIFF
--- a/lib/rucio/client/uploadclient.py
+++ b/lib/rucio/client/uploadclient.py
@@ -143,12 +143,11 @@ class UploadClient:
             rse_settings = self.rses[rse]
             is_deterministic = rse_settings.get('deterministic', True)
             if not is_deterministic and not pfn:
-                logger.warning('PFN has to be defined for NON-DETERMINISTIC site.')
+                logger.error('PFN has to be defined for NON-DETERMINISTIC site.')
                 continue
-            if pfn:
-                if no_register and not is_deterministic:
-                    logger.warning('Upload with given pfn implies that no_register is True, except non-deterministic sites')
-                    no_register = True
+            if pfn and not is_deterministic:
+                logger.warning('Upload with given pfn implies that no_register is True, except non-deterministic sites')
+                no_register = True
 
             if not no_register:
                 self._register_file(file, registered_dataset_dids)
--- a/lib/rucio/client/uploadclient.py
+++ b/lib/rucio/client/uploadclient.py
@@ -144,7 +144,7 @@ class UploadClient:
 
             rse = file['rse']
             rse_settings = self.rses[rse]
-            is_deterministic = rse_settings['deterministic']
+            is_deterministic = rse_settings.get('deterministic', True)
             if not is_deterministic and not pfn:
                 logger.warning('PFN has to be defined for NON-DETERMINISTIC site.')
                 continue
@@ -424,6 +424,7 @@ class UploadClient:
         replica['md5'] = file['md5']
         replica['meta'] = file['meta']
         replica['state'] = file['state']
-        if 'pfn' in file:
+        pfn = file.get('pfn')
+        if pfn:
             replica['pfn'] = file['pfn']
         return replica
--- a/lib/rucio/client/uploadclient.py
+++ b/lib/rucio/client/uploadclient.py
@@ -143,10 +143,10 @@ class UploadClient:
             rse_settings = self.rses[rse]
             is_deterministic = rse_settings.get('deterministic', True)
             if not is_deterministic and not pfn:
-                logger.error('PFN has to be defined for NON-DETERMINISTIC site.')
+                logger.error('PFN has to be defined for NON-DETERMINISTIC RSE.')
                 continue
             if pfn and not is_deterministic:
-                logger.warning('Upload with given pfn implies that no_register is True, except non-deterministic sites')
+                logger.warning('Upload with given pfn implies that no_register is True, except non-deterministic RSEs')
                 no_register = True
 
             if not no_register:
--- a/lib/rucio/client/uploadclient.py
+++ b/lib/rucio/client/uploadclient.py
@@ -139,26 +139,28 @@ class UploadClient:
             file_did = {'scope': file['did_scope'], 'name': file['did_name']}
             dataset_did_str = file.get('dataset_did_str')
 
-            if not no_register:
-                self._register_file(file, registered_dataset_dids)
-
             rse = file['rse']
             rse_settings = self.rses[rse]
             is_deterministic = rse_settings.get('deterministic', True)
             if not is_deterministic and not pfn:
                 logger.warning('PFN has to be defined for NON-DETERMINISTIC site.')
                 continue
+            if pfn:
+                if no_register and not is_deterministic:
+                    logger.warning('Upload with given pfn implies that no_register is True, except non-deterministic sites')
+                    no_register = True
+
+            if not no_register:
+                self._register_file(file, registered_dataset_dids)
 
             # if file already exists on RSE we're done
             if not is_deterministic and not no_register:
                 if rsemgr.exists(rse_settings, pfn):
-                    logger.info('File already exists on RSE with given pfn. Skipping upload. Re-upload is not supported. In such case, existing replica has to be removed first.')
+                    logger.info('File already exists on RSE with given pfn. Skipping upload. Existing replica has to be removed first.')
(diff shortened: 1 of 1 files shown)


lib/rucio/client/uploadclient.py
```diff
@@ -366,8 +381,8 @@ def _collect_and_validate_file_info(self, items):
                 logger.warning('Skipping file %s because no rse was given' % path)
                 continue
             if pfn:
-                if item.get('no_register'):
-                    logger.warning('Upload with given pfn implies that no_register is True')
+                if item.get('no_register') and item.get('deterministic'):
```
TWAtGH: `item` probably does not have a 'deterministic' key. I would suggest to do this check later when you have the rse_settings

lib/rucio/client/uploadclient.py
```diff
@@ -144,11 +144,26 @@ def upload(self, items, summary_file_path=None):
...
 
+            # if file already exists on RSE we're done
+            if not is_deterministic and not no_register:
+                if rsemgr.exists(rse_settings, pfn):
+                    logger.info('File already exists on RSE with given pfn. Skipping upload. Re-upload is not supported. In such case, existing replica has to be removed first.')
+                    continue
+                elif rsemgr.exists(rse_settings, file_did):
+                    logger.info('File already exists on RSE with different pfn. Skipping upload')
+                    continue
+                else:
```
TWAtGH: this can be removed

lib/rucio/client/uploadclient.py
```diff
@@ -409,4 +424,7 @@ def _convert_file_for_api(self, file):
         replica['md5'] = file['md5']
         replica['meta'] = file['meta']
         replica['state'] = file['state']
+        pfn = file.get('pfn')
+        if pfn:
+            replica['pfn'] = file['pfn']
```
TWAtGH: you can directly use `pfn` instead of `file['pfn']` here

TWAtGH (changes requested):

lib/rucio/client/uploadclient.py
TomasJavurek: Easiest solution:
          placing that part of code directly to upload() method.
Correct solution:
        move resolving rse_settings  to _collect_and_validate_file_info and publishing it to items

TomasJavurek: Another solution:
  calling self.rses.setdefault(rse, rsemgr.get_rse_info(rse)) in _collect_and_validate method and    put  only is_deterministic flag to item dict.

TomasJavurek: self.rses.setdefault(rse, rsemgr.get_rse_info(rse)) should be perhpas actually the first thing that we call in upload()

lib/rucio/client/uploadclient.py
TomasJavurek: it cannot. Threre are three unrelated cases:
1) file exists at site at given pfn
2) file exists somewhere at the site (which you woulldn't catch by first if)
3) file is not at the site and we can upload it.

lib/rucio/client/uploadclient.py
TWAtGH: No do it in the upload function please. I had a look and I remember why I did things that way. You loop 3 times over the files:
1. check if input is correct
2. check if all RSEs are ready
3. upload

This was to reduce the server calls in case something fails. It should be fine to do it in the second loop after the blacklisting check

lib/rucio/client/uploadclient.py
TWAtGH: It cannot? What impact does it have?
```
else:
  pass
```

TomasJavurek: ok, you mean only last two lines, ok, they can be removed

TomasJavurek:
I am not really happy with lsat commit I did because: 
I moved no_register flag evaluation to main upload method, but this has a negative consequence: guid will be always calculated, also in cases when pfn is set. This wasn't the case before.

lib/rucio/client/uploadclient.py
```diff
@@ -139,15 +139,32 @@ def upload(self, items, summary_file_path=None):
             file_did = {'scope': file['did_scope'], 'name': file['did_name']}
             dataset_did_str = file.get('dataset_did_str')
 
+            rse = file['rse']
+            rse_settings = self.rses[rse]
+            is_deterministic = rse_settings.get('deterministic', True)
+            if not is_deterministic and not pfn:
+                logger.warning('PFN has to be defined for NON-DETERMINISTIC site.')
+                continue
+            if pfn:
```
TWAtGH: you don't need this `if pfn` here. You made sure that a PFN is given if `not is_deterministic` evaluates to True .

lib/rucio/client/uploadclient.py
```diff
@@ -139,15 +139,32 @@ def upload(self, items, summary_file_path=None):
             file_did = {'scope': file['did_scope'], 'name': file['did_name']}
             dataset_did_str = file.get('dataset_did_str')
 
+            rse = file['rse']
+            rse_settings = self.rses[rse]
+            is_deterministic = rse_settings.get('deterministic', True)
+            if not is_deterministic and not pfn:
+                logger.warning('PFN has to be defined for NON-DETERMINISTIC site.')
```
TWAtGH: I would prefer to give an error here instead of a warning because the upload of this iteration will not be continued

lib/rucio/client/uploadclient.py
TomasJavurek: there is a bug, but different one. Actually, if no_register shouldn't be here. But otherwise, I moved this part form _collect_and_validate here. We have to be sure, that if pfn was given, no_register is set to True (but only in cases of non-deterministic sites)

TWAtGH (dismissed):

lib/rucio/client/uploadclient.py
```diff
@@ -139,15 +139,31 @@ def upload(self, items, summary_file_path=None):
             file_did = {'scope': file['did_scope'], 'name': file['did_name']}
             dataset_did_str = file.get('dataset_did_str')
 
+            rse = file['rse']
+            rse_settings = self.rses[rse]
+            is_deterministic = rse_settings.get('deterministic', True)
+            if not is_deterministic and not pfn:
+                logger.error('PFN has to be defined for NON-DETERMINISTIC site.')
```
bari12: Change `site` to `RSE`.

lib/rucio/client/uploadclient.py
```diff
@@ -139,15 +139,31 @@ def upload(self, items, summary_file_path=None):
...
+                continue
+            if pfn and not is_deterministic:
+                logger.warning('Upload with given pfn implies that no_register is True, except non-deterministic sites')
```
bari12: Change `site` to `RSE`.

bari12 (changes requested):
Just a small change-request to change "site" to "RSE".

lib/rucio/client/uploadclient.py
TWAtGH: Suggesting:
`logger.warning('Skipping upload because a PFN must be specified to upload to the non-deterministic RSE: %s' % rse)`

bari12 (approved):

Issue #1168, opened 2018-05-15
Misleading error message from conveyor-transfer-submitter


briedel:
The `conveyor-transfer-submitter` throws errors like

```
ERROR   Operation "third_party_copy" not supported by <RSE> with schemes <transfer schema>
```
 in two instances:
https://github.com/rucio/rucio/blob/master/lib/rucio/core/transfer.py#L572
https://github.com/rucio/rucio/blob/master/lib/rucio/core/transfer.py#L584

This is misleading as this error gets thrown twice for different checks, and if any of the other settings in `wan` in `"protocol": "domain":{}` are not set properly. It also blames the destination when the source is the issue.

Issue #63, opened 2017-11-08
Threadpool package is obsolete, needs to be replaced by something newer in conveyor


bari12:
- [x] Poller
- [x] Submitter

https://pypi.python.org/pypi/threadpool

_Migrated from JIRA [RUCIO-2785](https://its.cern.ch/jira/browse/RUCIO-2785)_

bari12:
Multiproces in Python 3

PR #1400, opened 2018-08-01
Conveyor : Change conveyors to heartbeat (submitter) :  #58, #63,#1168

### LINKED ISSUE #58: Change conveyors to heartbeat
_Migrated from JIRA [RUCIO-2761](https://its.cern.ch/jira/browse/RUCIO-2761)_

cserf:
Conveyor : Change conveyors to heartbeat (submitter) : Fixes #58

### DIFF
--- a/lib/rucio/core/staging.py
+++ b/lib/rucio/core/staging.py
@@ -7,6 +7,7 @@
 #
 # Authors:
 # - Martin Barisits, <martin.barisits@cern.ch>, 2017
+# - Cedric Serfon <cedric.serfon@cern.ch>, 2018
 
 
 """
@@ -28,16 +29,13 @@ from rucio.rse import rsemanager as rsemgr
 
 
 @read_session
-def get_stagein_requests_and_source_replicas(process=None, total_processes=None, thread=None, total_threads=None,
-                                             failover_schemes=None, limit=None, activity=None, older_than=None, rses=None,
-                                             mock=False, schemes=None, bring_online=43200, retry_other_fts=False, session=None):
+def get_stagein_requests_and_source_replicas(total_workers=0, worker_number=0, failover_schemes=None, limit=None, activity=None, older_than=None,
+                                             rses=None, mock=False, schemes=None, bring_online=43200, retry_other_fts=False, logging_prepend_str='', session=None):
     """
     Get staging requests and the associated source replicas
 
-    :param process:               Current process.
-    :param total_processes:       Total processes.
-    :param thread:                Thread number.
-    :param total_threads:         Total threads.
+    :param total_workers:         Number of total workers.
+    :param worker_number:         Id of the executing worker.
     :param failover_schemes:      Failover schemes.
     :param limit:                 Limit.
     :param activity:              Activity.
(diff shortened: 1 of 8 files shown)


lib/rucio/core/staging.py
```diff
@@ -194,22 +195,22 @@ def get_stagein_requests_and_source_replicas(process=None, total_processes=None,
...
-                                 'copy_pin_lifetime': attr.get('lifetime', -1) if attr else -1,
-                                 'external_host': external_host,
-                                 'selection_strategy': 'auto',
-                                 'rule_id': rule_id,
-                                 'file_metadata': file_metadata}
-                logging.debug("Transfer for request(%s): %s" % (id, transfers[id]))
-        except:
-            logging.critical("Exception happened when trying to get transfer for request %s: %s" % (id, traceback.format_exc()))
+                transfers[req_id] = {'request_id': req_id,
+                                     # 'src_urls': [source_url],
```
mlassnig, line 199: commented on purpose?

lib/rucio/daemons/conveyor/stager.py
```diff
@@ -92,23 +86,25 @@ def stager(once=False, rses=[], mock=False,
...
+    heart_beat = heartbeat.live(executable, hostname, pid, hb_thread)
+    prepend_str = 'Thread [%i/%i] : ' % (heart_beat['assign_thread'] + 1, heart_beat['nr_threads'])
+    logging.info(prepend_str + 'Stager starting with bring_online %s seconds' % (bring_online))
 
-    logging.info('Stager started - process (%i/%i) thread (%i/%i)' % (process, total_processes,
-                                                                      hb['assign_thread'], hb['nr_threads']))
-
-    threadPool = ThreadPool(total_threads)
-    activity_next_exe_time = defaultdict(time.time)
+    time.sleep(10)  # To prevent running on the same partition if all the poller restart at the same time
```
mlassnig, line 99: random time instead?

bari12: Random time is not necessary here. This is just to ensure that all threads have time to register so that the next read in line 100 sees all threads.

bbockelm: Isn’t that a classic race condition?

bari12: In principle it is, but there are other mechanisms (row-locking) which ensure that only one worker works at a certain request at a time. This timeout here should just ensure that it doesn't even come to this conflicts on start-up (daily restart etc.). But remember, these daemons can crash/startup at any time, so it can always happen that at some point in time the work partitions overlap until the end of the loop and the next heartbeat resync (line 107) - that's why there are the row-locking mechanisms as well.

bbockelm: Still, arbitrary sleeps catch a lot of attention (why 10 seconds?  why not 1?  why not 100?  etc).

How about using a barrier with a timeout?

https://docs.python.org/3/library/threading.html#barrier-objects (note that implementation is Python 3.x only but the implementation can be found in any OS book in the library...)

That will allow startup proceed as fast as possible -- but limit the lag possible from one thread.

bari12: The 10s are just a result of experience. The barrier-object will work within the process, but what we want to do here is to "sync" up multiple workers running on different machines, which can't be done with the barrier-object.
It's really just an artefact of the "synchronised" restart every day. The code would work without the sleep, there would just be a few row-lock conflicts in the first iteration of the loop.

lib/rucio/core/staging.py
cserf: Nope, it was already commented. I just modified the indentation

Issue #1195, opened 2018-05-28
Generic metadata support


mlassnig:
Motivation
----------

Currently, we only have a fixed set of metadata that is principally exposed via the core/meta.py, api/meta.py, client/metaclient.py modules. These metadata are strictly enforced based on the schema of the DID and require a schema-change for every new metadata field.

This task is to support the use of arbitrary key:value pairs per DID based on JSON.

Modification
------------
1. [x] re-enable and redo test/test_meta.py to conform to a new generic metadata interface
2. [x] adapt client/metaclient.py and bin/rucio to use the new metadata interface
3. [x] then, either, create new metadata table with three principal columns "scope", "name", "json metadata", or extend _did_ table with "json metadata" column (to be decided)
4. [x] adapt api/meta.py and core/meta.py to use the new schema

asketagarwal:
I'll be taking this up

tbeerman:
I'm currently still having troubles getting Oracle 12c reliably running on travis, i.e., you will need to add a switch in the code to enable and disable this feature based on the version of Oracle. So if the version is < 12 then disable json support otherwise enable it. This will be necessary anyway since our production DB is still 11g and we also would run tests both against 11g and 12c.

PR #1418, opened 2018-08-06
Metadata: Adding generic metadata support for sqlite #1195

asketagarwal:
Metadata: Adding generic metadata support for sqlite #1195

### DIFF
--- a/lib/rucio/core/did.py
+++ b/lib/rucio/core/did.py
@@ -1286,8 +1286,6 @@ def get_did_meta(scope, name, session=None):
     :param scope: the scope of did
     :param name: the name of the did
     """
-    if session.bind.dialect.name == 'sqlite':
-        raise NotImplementedError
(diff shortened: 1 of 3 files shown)


lib/rucio/core/did.py
```diff
@@ -1296,7 +1294,7 @@ def get_did_meta(scope, name, session=None):
     try:
         row = session.query(models.DidMeta).filter_by(scope=scope, name=name).one()
         meta = getattr(row, 'meta')
-        return json.loads(meta) if session.bind.dialect.name == 'oracle' else meta
+        return json.loads(meta) if session.bind.dialect.name == 'oracle' or session.bind.dialect.name == 'sqlite' else meta
```
mlassnig: can you try if it works with a statement like:
`.... if session.bind.dialect.name in ['oracle', 'sqlite'] ....`

mlassnig (approved):

PR #1426, opened 2018-08-07
Conveyor : Bug due to move to heartbeat : Fixes #1423

### LINKED ISSUE #1423: Bug due to move to heartbeat
Motivation
----------
The heart_beat in the conveyor is done using the full executable + arguments. One should only keep the executable + the activity option.
In addition a bug has been introduced by moving from `type(request_type) is not list`  to `isinstance(request_type, list)`

cserf:
Conveyor : Bug due to move to heartbeat : Fixes #1423


lib/rucio/daemons/conveyor/poller.py
```diff
@@ -72,7 +72,14 @@ def poller(once=False, activities=None, sleep_time=60,
     except NoOptionError:
         timeout = None
 
-    executable = ' '.join(sys.argv)
+    executable = sys.argv[0]
+    if activities:
+        activities.sort()
```
mlassnig, line 77: does this really come as a list? i think (may be mistaken) that this is just a single string, so you're sorting the characters in the string here?

bari12 (approved):

PR #1447, opened 2018-08-10
Core: Coerce Postgres uuid type to text before calling md5 function; Fix #1446

### LINKED ISSUE #1446: postgres requires ::text conversion of uuid for md5
Motivation
----------

It seems that some PSQL instances do not autoconvert uuid to text when calling md5(). We can force the conversion, e.g., this line:
               `query = query.filter(text('mod(abs((\'x\'||md5(id))::bit(32)::int), %s) = %s' % (total_workers + 1, worker_number)))`
should actually be:
               `query = query.filter(text('mod(abs((\'x\'||md5(id::text))::bit(32)::int), %s) = %s' % (total_workers + 1, worker_number)))`

illingwo:
Core: Coerce Postgres uuid type to text before calling md5 function; Fix #1446


mlassnig (approved):
Thabk you, @illingwo !

PR #1456, opened 2018-08-13
Metadata: Adding ui for generic metadata #1195

asketagarwal:
Metadata: Adding ui for generic metadata #1195

### DIFF
--- a/lib/rucio/web/ui/static/did.js
+++ b/lib/rucio/web/ui/static/did.js
@@ -861,7 +861,7 @@ build_page = function() {
             for(var i=0; i<sorted_keys.length; ++i) {
                 $('#t_generic_metadata').append($('<tr><th>' + sorted_keys[i] + '</th><td>' + data[sorted_keys[i]] + '</td></tr>'));
             }
-            $('#generic_metadata').css('style', 'visibility: visible') 
+            $('#generic_metadata').css('style', 'display: block') 
         },
         error: function(jqXHR, textStatus, errorThrown) {
             $('#loading').html('<font color="red">Could not find the DID.</font>');
--- a/lib/rucio/web/ui/static/did.js
+++ b/lib/rucio/web/ui/static/did.js
@@ -851,6 +851,21 @@ build_page = function() {
             $('#loading').html('<font color="red">Could not find the DID.</font>');
         }
     });
+
+    r.did_get_generic_metadata({
+        'scope': scope,
+        'name': name,
+        success: function(data) {
+            $("#loading").html(""); 
+            var sorted_keys = Object.keys(data).sort();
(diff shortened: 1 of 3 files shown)


lib/rucio/web/ui/templates/did.html
```diff
@@ -24,6 +24,13 @@ <h4>DID Metadata</h4>
     </table>
   </div>
 
+  <div id="metadata" class="columns panel">
+    <h4>Generic DID Metadata</h4>
+    <div id="loading">Loading, please wait...</div>
+    <table id="t_generic_metadata" style="word-wrap: break-word;">
```
bari12: I guess we need to be careful with this. For now using the metadata functions will not work on the production instance (No table, no support for json) - But with this the webui will always call the did_meta method. I wonder if this will just crash everything. 
In general we need to carefully test this that running 1.18.0 (with this metadata support) with non-supportive databases will not crash everything.

tbeerman: Also, on the webui side it will always show "Could not find DID". Which is a bit misleading. This div should be hidden by default and should only show when the generic did meta call is successful.

lib/rucio/web/ui/static/did.js
```diff
@@ -851,6 +851,21 @@ build_page = function() {
...
+        'name': name,
+        success: function(data) {
+            $("#loading").html(""); 
+            var sorted_keys = Object.keys(data).sort();
+            for(var i=0; i<sorted_keys.length; ++i) {
+                $('#t_generic_metadata').append($('<tr><th>' + sorted_keys[i] + '</th><td>' + data[sorted_keys[i]] + '</td></tr>'));
+            }
+        },
+        error: function(jqXHR, textStatus, errorThrown) {
+            $('#generic_metadata').hide();
```
asketagarwal: @tbeerman  I am hiding html if the rest api returns an error.

tbeerman: Please do it the other way around: hide the whole div by default (`style="visibility: hidden;"`) and only make it visible when the call was successful.

tbeerman:
I've just tried this on my dev machine. The hiding works fine but it leaves a big white space between the normal metadata panel and the panel, e.g., rules, which doesn't look nice. As an alternative it would be possible not define the `generic_metadata` div at all in the html file but instead only append it dynamically if needed. You could have a look at the `load_parents_dids` function. The panel there is also only added if there are actual parents to show.

mlassnig:
Hi @asketagarwal , will you still update this?

asketagarwal:
Yeah sorry for the delay. I'll try to finish by tomorrow

asketagarwal:
This should work i guess. @tbeerman

tbeerman (approved):

Issue #1454, opened 2018-08-13
provide helm charts to install rucio on kubernetes


tbeerman:
Motivation
----------
helm is a tool to manage kubernetes charts (packages). It packages pre-configures K8s resources for easy installation. It furthermore provides tools to manage deployed installations on a cluster. Helping to keep track of current installations and performing easy rollbacks.

Modification
------------
Add the chart to the repository so that other users can download it and use it to install Rucio.

PR #1461, opened 2018-08-15
add helm chart for rucio server installation #1454

tbeerman:
add helm chart for rucio server installation #1454


etc/docker/helm/rucio-server/Chart.yaml
```diff
@@ -0,0 +1,14 @@
+name: rucio-server
+version: 0.1.0
```
mlassnig: should this reflect the rucio version? or is it separate?

etc/docker/helm/rucio-server/README.md
```diff
@@ -0,0 +1,73 @@
+# Rucio
+
+##  Data Management for science in the Big Data era.
```
mlassnig, line 3: `  ## Scientific data management`

etc/docker/helm/rucio-server/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+Rucio is a project that provides services and associated libraries for allowing scientific collaborations to manage large volumes of data spread across facilities at multiple institutions and organisations. Rucio has been developed by the ATLAS. experiment. It offers advanced features, is highly scalable and modular. Rucio is a data management solution that could cover the needs of different communities in the scientific domain (e.g., HEP, astronomy, biology).
```
mlassnig: There's an additional dot behind ATLAS. And I would remove the last sentence (also from the main readme.md)

etc/docker/helm/rucio-server/README.md
```diff
@@ -0,0 +1,73 @@
...
+$ helm install rucio-server/
+```
+
+## Introduction
+
+This chart bootstraps a Rucio server deployment and service on a Kubernetes cluster using the Helm Package manager.
+
+## Installing the Chart
+
+To install the chart with the release name `my-release`:
```
mlassnig, line 25: what's the significance of the release name? is this tied to the release version mentioned above?

etc/docker/helm/rucio-server/Chart.yaml
tbeerman: There is a set of metadata one can use in the Chart.yaml. Some are required and some are optional. `version` is mandatory and it should reflect the version of the chart and not of the application it installs. For that there is another field `appVersion` which is necessary. But I deliberately omitted that one, as the Rucio version can be changed by setting the `image.tag` value without the need to update the chart itself.

etc/docker/helm/rucio-server/README.md
tbeerman: just copy/pasted the whole things from the main page. I'll change it on boths readme's

etc/docker/helm/rucio-server/README.md
tbeerman: A release is an actual installation of a chart on the kubernetes cluster. It just generally helps grouping together all of the resources created by this chart. It's possible to install multiple releases of the same chart on a cluster. It's not directly tied to the version number.

mlassnig (approved):

Issue #1470, opened 2018-08-17
Add downloadclient and uploadclient to main Client object


bari12:
Motivation
----------
Add the two clients to the main client object
https://github.com/rucio/rucio/blob/master/lib/rucio/client/client.py#L24

bari12:
@TWAtGH during the creation of the PR we saw that UploadClient and DownloadClient are not inherited from BaseClient. Is there a special reason for this?

TWAtGH:
I don't really remember to be honest. But looking at it now the download and upload client don't really fit into the api like the other classes do. The two classes only need to use the client and don't do rest calls themselves. 
I guess that were my thoughts when creating these classes. I probably wasn't aware of how the client object is meant to be used.
However changing this now will be a little bit challenging since the `__init__` calls are getting some custom parameters and should be downward compatible.

TWAtGH:
I would suggest to close this one. As I've written above adding these classes to the main client is not necessary and could be quite problematic

PR #1471, opened 2018-08-20
clients: introducing upload/download clients to the main client #1470

TomasJavurek:
clients: introducing upload/download clients to the main client #1470


bari12:
I am not sure if this is enough (I thought it is when I created the ticket) but in the meanwhile I saw that the UploadClient and the DownloadClient are not inherited from BaseClient, so it won't work like this I think. Not sure if this was forgotten or if it is on purpose?

TomasJavurek:
I don't think there is a purpose, but I rather ask Tobi. However, we always call upload/download clients directly in the other services.

bari12:
I am closing this now, as it won't work like this, we first would have to inherit the UploadClient and DownloadClient from BaseClient. Let's follow this up in #1470

Issue #1487, opened 2018-08-22
Upload with --no-register --pfn does not work


bari12:
Motivation
----------

The upload with --no-register --pfn does not work.

```
[wguan@lxplus078 wguan]$  /usr/bin/env rucio -v upload --rse CERN-PROD_ES --scope transient --no-register --pfn s3://cs3.cern.ch:443//atlas-eventservice/05cc36b8-383f-444f-a384-592fad3b918e123323 /bin/hostname
2018-08-22 17:21:13,312 INFO    Preparing upload for file hostname
2018-08-22 17:21:13,312 WARNING Upload with given pfn implies that no_register is True, except non-deterministic RSEs
2018-08-22 17:21:13,530 INFO    Trying upload with s3 to CERN-PROD_ES
2018-08-22 17:21:14,590 ERROR   list indices must be integers, not str
2018-08-22 17:21:14,590 DEBUG   This means the parameter you passed has a wrong type.
Completed in 1.3510 sec.
```

TomasJavurek:
We have hole somewhere in rsemgr upload method. What is certainly wrong, that we don't have same structure on return, once list and once dictionary:
https://github.com/rucio/rucio/blob/master/lib/rucio/rse/rsemanager.py#L494

The case of upload to CERN-PROD_ES leads to the empty list, ret = [], so it returns [gs=True, []]
but this should actually never happen.

PR #1509, opened 2018-09-03
rsemanager: dirty dirty fix of upload method return - #1487

TomasJavurek:
rsemanager: dirty dirty fix of upload method return - #1487


TWAtGH:
We have a duplicated PR now: #1493
I guess you can delete the other one.

TWAtGH (approved):

Issue #312, opened 2017-12-07
Rebalancing breaks in decomissioning mode


cserf:
Motivation
----------
The query against the locks takes too much time and Oracle fails with "snapshot too old"
`
***************************
BB8 - Execution Summary
Mode:    DECOMISSION
Dry Run: False
***************************
scope:name rule_id bytes(Gb) target_rse child_rule_id
(cx_Oracle.DatabaseError) ORA-01555: snapshot too old: rollback segment number 7078 with name "_SYSSMU7078_1542158916$" too small
 [SQL: "SELECT r.scope, r.name, rawtohex(r.id) as rule_id, r.rse_expression as rse_expression, r.subscription_id as subscription_id, 0 as bytes, 0 as length FROM atlas_rucio.rules r\nWHERE\nr.id IN (SELECT rule_id FROM atlas_rucio.locks WHERE rse_id = atlas_rucio.rse2id(:rse) GROUP BY rule_id) and\nr.state = 'O' and\nr.child_rule_id is NULL"] [parameters: {'rse': 'xxxx']
`

Modification
------------
Use some dumps on Hadoop. The dumps must be grouped by rule_id, rse and state

TomasJavurek:
The solution is not streightforeward since we don't have dump of lock table in hadoop.

PR #1510, opened 2018-09-04
bb8: decommission with dumps and other changes - #312

TomasJavurek:
bb8: decommission with dumps and other changes - #312

### DIFF
--- a/bin/rucio-bb8
+++ b/bin/rucio-bb8
@@ -1,4 +1,4 @@
-#!/usr/bin/python2
+#!/usr/bin/env python
 # Copyright 2016-2018 CERN for the benefit of the ATLAS collaboration.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
--- a/bin/rucio-bb8
+++ b/bin/rucio-bb8
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/python2
 # Copyright 2016-2018 CERN for the benefit of the ATLAS collaboration.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -37,7 +37,7 @@ def get_parser():
     parser.add_argument('--exclude-expression', action='store', help='Exclude these rse_expression from being destinations')
     parser.add_argument('--comment', action='store', help='Add a comment to the new rules')
     parser.add_argument('--force-expression', action='store', help='For this rse_expression for rebalanced rules instead of letting BB8 decide')
-    parser.add_argument('--decomission', action='store_true', help='Run BB8 in decomission mode')
+    parser.add_argument('--decommission', action='store_true', help='Run BB8 in decommission mode')
     parser.add_argument('--priority', action='store', help='Priority for the newly created rules', type=int, default=3)
     parser.add_argument('--source-replica-expression', action='store', help='Source replica expression for the newly created rules')
     return parser
@@ -47,12 +47,12 @@ if __name__ == "__main__":
     parser = get_parser()
     args = parser.parse_args()
 
-    if args.decomission:
+    if args.decommission:
         rebalance_rse(rse=args.rse, max_bytes=args.bytes, dry_run=args.dry_run,
                       comment=args.comment, force_expression=args.force_expression,
                       priority=args.priority,
                       source_replica_expression=args.source_replica_expression,
-                      mode='decomission')
+                      mode='decommission')
     else:
         rebalance_rse(rse=args.rse, max_bytes=args.bytes, dry_run=args.dry_run,
                       comment=args.comment, force_expression=args.force_expression,
--- a/lib/rucio/daemons/bb8/common.py
+++ b/lib/rucio/daemons/bb8/common.py
@@ -110,29 +110,38 @@ def rebalance_rule(parent_rule, activity, rse_expression, priority, source_repli
 
 def __dump_url(rse):
     """
-    getting rse dump with different methods according to settings
+    getting potential urls of the dump over last week
 
-    :param rse          RSE where the dump is released.
+    :param rse:          RSE where the dump is released.
     """
 
     # get the date of the most recent dump
-    dump_production_day = config_get('bb8', 'dump_production_day', raise_exception=False, default='Sunday')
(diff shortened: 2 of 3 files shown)


TomasJavurek:
I forgot about flak8.

lib/rucio/daemons/bb8/common.py
```diff
@@ -101,12 +100,79 @@ def rebalance_rule(parent_rule_id, activity, rse_expression, priority, source_re
...
 
+def _list_rebalance_rule_candidates_dump(rse, mode=None):
+    """
+    Download dump to tmporary directory
+
+    :param rse:          RSE of the source.
+    :param mode:         Rebalancing mode.
+    """
+
+    from requests import get
```
bari12: I think this should be on top with the other imports

lib/rucio/daemons/bb8/common.py
```diff
@@ -101,12 +100,79 @@ def rebalance_rule(parent_rule_id, activity, rse_expression, priority, source_re
...
+
+    # get last sunday locks dump date:
+    today = date.today()
+    today_idx = (today.weekday() + 1) % 7
+    sun = today - timedelta(today_idx)
+    last_sunday_date = sun.strftime('%d-%m-%Y')
+
+    # expected location of a dump for given rse
+    dumps_location = config_get('bb8', 'dumps_location', raise_exception=False, default='http://rucio-analytix.cern.ch:8080/LOCKS/')
+    rse_dump = '%sGetFileFromHDFS?date=%s&rse=%s' % (dumps_location, str(last_sunday_date), rse)
```
bari12: I am not sure if we should do it like that.
This way of getting the dump is quite specific to how ATLAS stores the logfiles on rucio-analytix. I think it might be better if BB8 just takes a dump (or URL to a dump) and all this dump identification part we do outside. This way it is easier for non-ATLAS users to use bb8.

bin/rucio-bb8
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/python2
```
mlassnig: this has to stay with env python

lib/rucio/daemons/bb8/common.py
```diff
@@ -171,17 +245,24 @@ def select_target_rse(current_rse, rse_expression, subscription_id, rse_attribut
...
+            rses = parse_expression(expression='(%s)\\%s' % (force_expression, target_rse), filter={'availability_write': True}, session=session)
+        else:
+            # in order to avoid replication of the part of distributed dataset not present at rabalanced rse -> rses in force_expression
+            # this will be extended with development of delayed rule
+            rses = parse_expression(expression='((%s)|(%s))\\%s' % (force_expression, rse_expression, target_rse), filter={'availability_write': True}, session=session)
     elif len(rses) > 1:
         # Just define the RSE Expression without the current_rse
         return '(%s)\\%s' % (rse_expression, target_rse)
-    elif rse_attributes['tier'] is True or rse_attributes['tier'] == '1':
+    elif rse_attributes['tier'] is True or int(rse_attributes['tier']) == 1:
```
mlassnig, line 294: what happens if an RSE does not have the attribute 'tier' set?

lib/rucio/tests/test_bb8.py
```diff
@@ -75,7 +75,12 @@ def test_bb8_rebalance_rule(self):
 
         rule_id = add_rule(dids=[{'scope': scope, 'name': dataset}], account='jdoe', copies=1, rse_expression=self.rse1, grouping='NONE', weight='fakeweight', lifetime=None, locked=False, subscription_id=None)[0]
 
-        child_rule = rebalance_rule(rule_id, 'Rebalance', self.rse3, priority=3)
+        rule = {}
+        try:
+            rule = get_rule(rule_id)
+        except:
+            assert_raises(RuleNotFound, get_rule, rule_id)
+        child_rule = rebalance_rule(rule, 'Rebalance', self.rse3, priority=3)
```
mlassnig, line 83: some verification possible? :-)

lib/rucio/daemons/bb8/common.py
TomasJavurek: uff ... I wouldn't like to set some probe outside of bb8, on which the bb8 would be dependent.

lib/rucio/daemons/bb8/common.py
TomasJavurek: I get a KeyError, but it used to be there since ever. I'll fix that.

TomasJavurek: This is not easy. We have no such a case for our DATADISKs. If other customers won't use our notation of DATADISKs and tiers, this all will be completely broken and needed to be completely rewritten. There might be three solutions for now, 1) leave it be, 2) if tier not in attributes: rebalance to all datadisks 3) if tier not in attributes: skip the rse. I would prefer the first one for the moment.

mlassnig: ok, but please put a TODO comment in the code so we don't forget

lib/rucio/daemons/bb8/common.py
```diff
@@ -101,12 +101,98 @@ def rebalance_rule(parent_rule_id, activity, rse_expression, priority, source_re
...
-    update_rule(rule_id=parent_rule_id, options={'child_rule_id': child_rule, 'lifetime': 0})
+    update_rule(rule_id=parent_rule['id'], options={'child_rule_id': child_rule, 'lifetime': 0})
     return child_rule
 
 
+def __dump_url(rse):
+    """
+    getting rse dump with different methods according to settings
+
+    :param rse          RSE where the dump is released.
```
bari12: :param rse:

lib/rucio/daemons/bb8/common.py
```diff
@@ -101,12 +101,98 @@ def rebalance_rule(parent_rule_id, activity, rse_expression, priority, source_re
...
+    """
+
+    # get the date of the most recent dump
+    dump_production_day = config_get('bb8', 'dump_production_day', raise_exception=False, default='Sunday')
```
bari12: Do you expect that the config is always updated for this? I wonder if this is necessary, or if it is enough to just iterate back from today, one day at a time?

bari12 (changes requested):

bari12 (approved):

PR #1513, opened 2018-09-05
Clients: fixed exception handler if no policy section is present; Fix #1485

### LINKED ISSUE #1485: If policy section not available in client config, no exception should be raised
Motivation
----------
There are two `config_get('policy', ...)` calls in the bin/rucio code. However, if the policy section is not available in the config an error is raised. This should not happen. Default values should be used. ("")

TWAtGH:
Clients: fixed exception handler if no policy section is present; Fix #1485

### DIFF
--- a/bin/rucio
+++ b/bin/rucio
@@ -213,7 +213,7 @@ def exception_handler(function):
                 logger.error('Specified account %s does not have an associated identity.' % used_account)
             else:
                 logger.debug(traceback.format_exc())
-                contact= config_get('policy', 'support', raise_exception=False)
+                contact = config_get('policy', 'support', raise_exception=False)
                 support = ('Please followup with all relevant information at: ' + contact) if contact else ''
                 logger.error('\nThe object is missing this property: %s\n'
                              'This should never happen. Please rerun the last command with the "-v" option to gather more information.\n'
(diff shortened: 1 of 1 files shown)


bin/rucio
```diff
@@ -212,23 +212,26 @@ def exception_handler(function):
                     pass
                 logger.error('Specified account %s does not have an associated identity.' % used_account)
             else:
-                logger.error(error)
                 logger.debug(traceback.format_exc())
+                contact= config_get('policy', 'support', raise_exception=False)
+                support = ('Please followup with all relevant information at: ' + contact) if contact else ''
                 logger.error('\nThe object is missing this property: %s\n'
                              'This should never happen. Please rerun the last command with the "-v" option to gather more information.\n'
-                             'Please followup with all relevant information at: %s' % (str(error), config_get('policy', 'support')))
```
vingar: One possibility is to define a default value instead like  `config_get('policy', 'support', 'https://github.com/rucio/issues')`

TWAtGH: Yeah I know but what to write in there?

mlassnig: `https://github.com/rucio/rucio/issues`

:-D

TWAtGH: What I wanted to say is then we could hardcode it anyway ^^

bari12:
I think for a rucio problem we should hardcode https://github.com/rucio/rucio/issues

bari12 (approved):

PR #1530, opened 2018-09-12
Clients: added human friendly setting ofaccount limits per rucio admin cli; Fix #1529

### LINKED ISSUE #1529: Human friendly limit setting
Motivation
----------
Right now, setting limits via the command line (`rucio-admin account set-limits`) is a bit of black magic.  You have to specify in bytes (notoriously human-unfriendly) and there's an undocumented magic value of `-1` which is shorthand for "infinity".

Proposed Modification
------------
* Allow the CLI interface to accept units (`10GB` instead of `10000000000`)
* Accept "inf" and "infinity" (case-insensitive).

hahahannes:
Clients: added human friendly setting ofaccount limits per rucio admin cli; Fix #1529

### DIFF
--- a/bin/rucio-admin
+++ b/bin/rucio-admin
@@ -37,6 +37,7 @@ import signal
 import sys
 import time
 import traceback
+import re
 
 try:
     from ConfigParser import NoOptionError, NoSectionError
@@ -329,8 +330,36 @@ def set_limits(args):
     Set account limit for an account and rse.
     """
     client = get_client(args)
-    client.set_account_limit(account=args.account, rse=args.rse, bytes=args.bytes)
-    print('Set account limit for account %s on RSE %s: %s' % (args.account, args.rse, sizefmt(args.bytes, True)))
+    limit = args.limit
+    result = re.findall('^([0-9]+)([A-Z]+)$',limit)
+     
+    if limit == 'inf' or limit == 'infinity':
+        limit = -1
+    elif len(result) is not 0:
+        value = int(result[0][0])
+        unit = result[0][1]
+        
+        if unit == 'B':
+            limit = value
+        elif unit == 'KB':
+            limit = value * 1000
+        elif unit == 'MB':
+            limit = value * 1000000
+        elif unit == 'GB':
+            limit = value * 1000000000
+        elif unit == 'TB':
+            limit = value * 1000000000000
+        elif unit == 'PB':
+            limit = value * 1000000000000000
+        else:
+            print('The unit of your limit could not be recognized. Please use one of the following units: B,KB,MB,GB,TB,PB.')
+            return FAILURE
+    else:
+        print('The limit could not be set. Either you misspelled infinity or used a wrong pattern. Please use a format like 10GB with B,KB,MB,GB,TB,PB as units.')
+        return FAILURE
+        
+    client.set_account_limit(account=args.account, rse=args.rse, bytes=limit)
+    print('Set account limit for account %s on RSE %s: %s' % (args.account, args.rse, sizefmt(limit, True)))
     return SUCCESS
 
 
@@ -1114,7 +1143,7 @@ def get_parser():
     set_account_limits_parser.set_defaults(which='set_limits')
     set_account_limits_parser.add_argument('account', action='store', help='Account name')
     set_account_limits_parser.add_argument('rse', action='store', help='RSE boolean expression')
-    set_account_limits_parser.add_argument('bytes', action='store', type=int, help='The total number of bytes that can be stored')
+    set_account_limits_parser.add_argument('limit', action='store', help='The total number of bytes that can be stored. Value can be either the number of bytes as Integer like 10000 or in units like 10GB with the common units or inf/infinity.')
 
     # The account-limit subparser
     get_account_limits_parser = account_subparser.add_parser('get-limits',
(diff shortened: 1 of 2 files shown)


bin/rucio-admin
```diff
@@ -1114,7 +1143,7 @@ def get_parser():
     set_account_limits_parser.set_defaults(which='set_limits')
     set_account_limits_parser.add_argument('account', action='store', help='Account name')
     set_account_limits_parser.add_argument('rse', action='store', help='RSE boolean expression')
-    set_account_limits_parser.add_argument('bytes', action='store', type=int, help='The total number of bytes that can be stored')
+    set_account_limits_parser.add_argument('limit', action='store', help='The total number of bytes that can be stored. Value can be either the number of bytes as Integer like 10000 or in units like 10GB with the common units or inf/infinity.')
```
mlassnig: There might be scripts that still use the --bytes option. Can you add --limit as an alternative and not remove the old one?

bin/rucio-admin
```diff
@@ -37,6 +37,7 @@ import signal
 import sys
 import time
 import traceback
+import re
```
mlassnig: imports should be sorted alphabetically

bin/rucio-admin
```diff
@@ -329,8 +330,36 @@ def set_limits(args):
...
+    limit = args.limit
+    result = re.findall('^([0-9]+)([A-Z]+)$',limit)
+     
+    if limit == 'inf' or limit == 'infinity':
+        limit = -1
+    elif len(result) is not 0:
+        value = int(result[0][0])
+        unit = result[0][1]
+        
+        if unit == 'B':
```
mlassnig: not entirely sure if it makes sense, but should we check along the lines of  lower(unit) == 'b' ? (just so ignore case sensitivity) ?

bin/rucio-admin
mlassnig: oh, and don't forget to add yourself as an author to the top of the file

bin/rucio-admin
mlassnig: also, the transform of values could go into common/utils?

bin/rucio-admin
```diff
@@ -329,8 +330,36 @@ def set_limits(args):
     Set account limit for an account and rse.
     """
     client = get_client(args)
-    client.set_account_limit(account=args.account, rse=args.rse, bytes=args.bytes)
-    print('Set account limit for account %s on RSE %s: %s' % (args.account, args.rse, sizefmt(args.bytes, True)))
+    limit = args.limit
+    result = re.findall('^([0-9]+)([A-Z]+)$',limit)
+     
+    if limit == 'inf' or limit == 'infinity':
```
bari12: Would be good to move this part as a method to common.utils, in case we want to use this in other methods.

bin/rucio-admin
```diff
@@ -329,8 +330,36 @@ def set_limits(args):
...
+        elif unit == 'MB':
+            limit = value * 1000000
+        elif unit == 'GB':
+            limit = value * 1000000000
+        elif unit == 'TB':
+            limit = value * 1000000000000
+        elif unit == 'PB':
+            limit = value * 1000000000000000
+        else:
+            print('The unit of your limit could not be recognized. Please use one of the following units: B,KB,MB,GB,TB,PB.')
```
vingar: Suggest using the logging module and the appropriate logging level like in the rucio CLI.

vingar (commented):
One non-blocking comment.

bin/rucio-admin
bbockelm: Suggested tweak:

> Value can be specified in bytes ("10000"), with a storage unit ("10GB"), or "infinity"

bin/rucio-admin
```diff
@@ -1114,7 +1143,7 @@ def get_parser():
     set_account_limits_parser.set_defaults(which='set_limits')
     set_account_limits_parser.add_argument('account', action='store', help='Account name')
     set_account_limits_parser.add_argument('rse', action='store', help='RSE boolean expression')
-    set_account_limits_parser.add_argument('bytes', action='store', type=int, help='The total number of bytes that can be stored')
```
bbockelm, line 1117: For backward compatibility, please keep the `bytes` argument.

lib/rucio/common/utils.py
```diff
@@ -723,3 +724,18 @@ def add_url_query(url, query):
     mod_query.update(query)
     url_parts[4] = urllib.urlencode(mod_query)
     return urlparse.urlunparse(url_parts)
+
+
+def extract_storage_unit_and_value(input_string):
```
bari12: What I had in mind as moving all of the conversion part to the method, including the kb, mb, gb, etc. part. as it is really something we might want to use in different places as well.

hahahannes: Okay, makes sense. Then I would add a function to get the bytes value from a string, e.g. get_bytes_from_string()

bin/rucio-admin
```diff
@@ -329,8 +330,22 @@ def set_limits(args):
     Set account limit for an account and rse.
     """
     client = get_client(args)
-    client.set_account_limit(account=args.account, rse=args.rse, bytes=args.bytes)
-    print('Set account limit for account %s on RSE %s: %s' % (args.account, args.rse, sizefmt(args.bytes, True)))
+    byte_limit = None
+    limit_input = args.bytes
+
+    if limit_input == 'inf' or limit_input == 'infinity':
```
bbockelm: Can you do a case-insensitive match?

bari12 (approved):

PR #1536, opened 2018-09-14
Clients: added option to list rse usage per account; Fix #1218

### LINKED ISSUE #1218: New option for rucio list-rse-usage to list usage per account
Motivation
----------
rucio list-rse-usage returns the RSE usage from different sources (Rucio, storage, expired). It would be good to have an option to list the usage per account.

hahahannes:
Clients: added option to list rse usage per account; Fix #1218

### DIFF
--- a/bin/rucio
+++ b/bin/rucio
@@ -1437,7 +1437,7 @@ def list_rse_usage(args):
 
     """
     client = get_client(args)
-    usages = client.get_rse_usage(rse=args.rse, filters={'per_account': args.per_account})
+    usages = client.get_rse_usage(rse=args.rse, filters={'per_account': args.show_accounts})
     print('USAGE:')
     for usage in usages:
         if usage['source'] not in ['srm', 'gsiftp', 'webdav']:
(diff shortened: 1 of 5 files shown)


bin/rucio
```diff
@@ -1995,6 +2009,7 @@ You can filter by key/value, e.g.::
     list_rse_usage_parser.set_defaults(which='list_rse_usage')
     list_rse_usage_parser.add_argument(dest='rse', action='store', help='Rucio Storage Element (RSE) name.').completer = rse_completer
     list_rse_usage_parser.add_argument('--history', dest='history', default=False, action='store', help='List RSE usage history. [Unimplemented]')
+    list_rse_usage_parser.add_argument('--per-account', dest='per_account', action='store_true', default=False, help='List RSE usage per account')
```
bari12: `--per-account` sounds a bit weird to me. I think `--show-accounts` or something similar might be better.

lib/rucio/api/rse.py
```diff
@@ -243,17 +244,18 @@ def set_rse_usage(rse, source, used, free, issuer):
...
+def get_rse_usage(rse, issuer, source=None, per_account=False):
     """
     get RSE usage information.
 
     :param rse: The RSE name.
     :param issuer: The issuer account.
     :param source: dictionary of attributes by which the results should be filtered
+    :param per_account: Boolean whether usage should be calculated per account or not
 
     :returns: True if successful, otherwise false.
```
bari12: I know you did not write this line, but please change to what is actually happening.

lib/rucio/core/rse.py
```diff
@@ -485,30 +486,42 @@ def set_rse_usage(rse, source, used, free, session=None):
...
     """
     get rse usage information.
 
     :param rse: The rse name.
     :param source: The information source, e.g. srm.
     :param rse_id:  The RSE id.
     :param session: The database session in use.
+    :param per_account: Boolean whether the usage should be calculated per account or not.
 
     :returns: True if successful, otherwise false.
```
bari12: Same with this line

lib/rucio/api/rse.py
hahahannes: More like 'Show accounts usages of RSE'? Also shall I replace the parameter name to 'show_accounts'?

bari12: I mean the `:returns: True if successful, otherwise false.` should actually be a `:returns: List of RSE usage data`

bari12:
This will need a rebase, infact, it might be better to just re-submit this as an actual patch so things are in order.

hahahannes:
Okay, will do

PR #1546, opened 2018-09-19
Clients: chunk DIDs array with length above 1000; Fix #30

### LINKED ISSUE #30: Rucio attach should chunk the DIDSs array up accordingly to the 1000 server limits
_Migrated from JIRA [RUCIO-2418](https://its.cern.ch/jira/browse/RUCIO-2418)_

hahahannes:
Clients: chunk DIDs array with length above 1000; Fix #30

### DIFF
--- a/lib/rucio/core/did.py
+++ b/lib/rucio/core/did.py
@@ -490,6 +490,18 @@ def attach_dids_to_dids(attachments, account, ignore_duplicate=False, session=No
     parent_did_condition = list()
     parent_dids = list()
     for attachment in attachments:
+        content = [{'scope': did['scope'], 'name': did['name']} for did in list_content(scope=attachment['scope'], name=attachment['name'], session=session)]
+        already_attached = []
+        not_attached = []
+        for did in attachment['dids']:
+            if {'name': did['name'], 'scope': did['scope']} in content:
+                already_attached.append(did)
+            else:
+                not_attached.append(did)
+        if len(not_attached) != 0:
+            attachment['dids'] = not_attached
+        else:
+            raise exception.DataIdentifierAlreadyAttached("Data Identifier already attached: {0}".format(','.join([did['scope'] + ':' + did['name'] for did in already_attached])))
         try:
             parent_did = session.query(models.DataIdentifier).filter_by(scope=attachment['scope'], name=attachment['name']).\
                 with_hint(models.DataIdentifier, "INDEX(DIDS DIDS_PK)", 'oracle').\
--- a/lib/rucio/core/did.py
+++ b/lib/rucio/core/did.py
@@ -490,18 +490,6 @@ def attach_dids_to_dids(attachments, account, ignore_duplicate=False, session=No
     parent_did_condition = list()
     parent_dids = list()
     for attachment in attachments:
-        content = [{'scope': did['scope'], 'name': did['name']} for did in list_content(scope=attachment['scope'], name=attachment['name'], session=session)]
-        already_attached = []
-        not_attached = []
-        for did in attachment['dids']:
-            if {'name': did['name'], 'scope': did['scope']} in content:
-                already_attached.append(did)
-            else:
-                not_attached.append(did)
-        if len(not_attached) != 0:
-            attachment['dids'] = not_attached
-        else:
-            raise exception.DataIdentifierAlreadyAttached("Data Identifier already attached: {0}".format(','.join([did['scope'] + ':' + did['name'] for did in already_attached])))
         try:
             parent_did = session.query(models.DataIdentifier).filter_by(scope=attachment['scope'], name=attachment['name']).\
                 with_hint(models.DataIdentifier, "INDEX(DIDS DIDS_PK)", 'oracle').\
--- a/bin/rucio
+++ b/bin/rucio
@@ -37,6 +37,7 @@ from __future__ import print_function
 
 import argparse
 import logging
+import math
 import os
 import signal
 import subprocess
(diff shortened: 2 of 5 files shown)


mlassnig:
`2018-09-26 15:10:56,311	ERROR	Database exception.
Details: (sqlite3.OperationalError) too many SQL variables [SQL: u'SELECT dids.scope AS dids_scope, dids.name AS dids_name, dids.did_type AS dids_did_type \nFROM dids \nWHERE dids.scope = ? AND dids.name = ? OR dids.scope = ? AND dids.name = ? OR dids.scope = ? AND dids.name = ? OR dids.scope = ? AND dids.name = ? OR dids.scope = ? AND dids.name = ? OR dids.scope = ? AND dids.name = ? OR dids.scope = ? AND dids.name = ? OR dids.scope = ? AND dids.name = ? OR dids.scope = ? AND dids.name = ? OR dids.scope = ? AND dids.name = ? OR dids.scope = ? AND dids.name = ? OR dids.scope = ? AND dids.name = ? OR dids.scope = ? AND dids.name = ? OR dids.scope = ? AND dids.name = ? OR dids.scope = ? AND dids.name = ? OR dids.scope = ? AND dids.name = ? OR dids.scope = ? AND dids.name = ? OR dids.scope = ? AND dids.name = ? OR dids.scope = ? AND dids.name = ? OR dids.scope = ? AND dids.name = ? OR dids.scope = ? AND dids.name = ? OR dids.scope = ? AND dids.name = ? OR dids.scope = ? AND dids.name = ? OR...`

bari12:
Essentially we never tried to attach a big number of dids in the tests, now @hahahannes is doing this. And with sqllite there is a maximum number of bind variables, which we are hitting. I would propose to disable this test in sqllite.

mlassnig:
FYI MySQL limit is 65k, PostgreSQL limit is 32k.

vingar:
limit for sqlite ?

mlassnig:
SQLite 999

vingar:
I would suggest to lower the chunk size to get it working on all backends including sqlite: 1000 -> 999

bari12:
I traced to code a bit and it seems like often the chunk sizes we use are not limited, we just take whatever list is given to the method. Thus the limit here is really given by the attach size in the test (what is given to the client) - this is not ideal and some chunk sizes should be enforced in the code. 
Limiting in the client for all to make the sqllite test work is not so good. Especially (needs to be tested) this might mean to limit it down to 300 or so dids per attach. Because it could be 3 bind variables per did (scope, name, did_type).

bari12:
There is a second issue which we should address right away in this PR:
For the user who calls attach-dids with 10k dids it looks like one transaction, however, in the background this is obviously multiple interactions/transactions. I think this is fine, but we should say it somewhere.
It gets problematic though if the command crashes/breaks e.g. after 5000 of 10000 attaches. To the user this looks like the whole command crashed, but actually the command is half finished. This should also be said somewhere in the output. Thirdly, now the user wants to re-execute the command, I think right now this would just fail as the dids are already partly attached to the dataset.

mlassnig:
I don't think it's necessary to do it so complicated. There's no need to flush the insert of 500+500+500+... each. If the final flush doesn't go through, it would rollback everything.

bari12:
But for each of these 500 these are different REST calls, so per design it's a new transaction, you cannot flush them together.

mlassnig:
Ah, I missed the multiple REST calls. Yes, that's bad then. POST anyone?

hahahannes:
If I understood correctly, there are several problems
1. Client chunk
Chunk inside the client with 499 instead of 1000 if we want to support SQLite or not.

2. Internal Chunk
The possibility that there can be lists inside the server that can also exceed the limit and there is no chunking internal. Then there the lists should be also chunked with length 300 with SQLite or the minimum limit if we drop SQLite. If we want to keep SQLite then the client could chunk the list with length 1000 because of the apache problem and then internal it would be again chunked with 300. If we drop SQLite the client would chunk with 1000 and then internal again with the minimum limit. 
 
3. One Transaction vs Multiple Transactions
I guess we could display the chunks as they are processed in the client like:
1/10 successful
2/10 successful
and if one fails the client could deattach them which would be also displayed.

bari12:
1. Yes, change it down to 499 for now to make the tests work.
2. The internal chunking you don't have to do anything about. This is a change for the future.
3. I think detaching we don't really need. If the attach failed it is quite likely that the detach will fail too (Network issues etc.). I think a better workflow is to clearly display what has been successfully done. But the important part is, when it fails, the user should be able to re-send the command and Rucio should not fail because some of the files are already attached. (You have to test this, I am not sure, but quite likely attach gives an error if you try to attach something twice)

hahahannes:
1.) Ok
3.) I tested it and yes, it fails with
```
[u'(_mysql_exceptions.IntegrityError) (1062, "Duplicate entry \'mock-bla-mock-bla2\' for key \'PRIMARY\'")']
```
I would then check the content of a collection if one of the dids is already there and then remove it from the list before running the query.

hahahannes:
I created a new exception class for the case, that a did already got attached. Not sure if this is correct, because I had to change some other test cases.

lib/rucio/core/did.py
```diff
@@ -488,6 +488,18 @@ def attach_dids_to_dids(attachments, account, ignore_duplicate=False, session=No
     parent_did_condition = list()
     parent_dids = list()
     for attachment in attachments:
+        content = [{'scope': did['scope'], 'name': did['name']} for did in list_content(scope=attachment['scope'], name=attachment['name'], session=session)]
```
bari12: This is getting too expensive on the server side, because it essentially adds O(n) queries for n dids to be attached, this is really a big increase for attaches. I think it is better to leave the server-side code simple and just let it try to insert and run into an IntegrityException
This entire block (491-502) should be removed. 

We should do this on bin/rucio side only.
Thus when you have to do the chunking: If the IntegrityException is raised, then you check for this chunk which files have already been added to the dataset and which don't, then you go to the next chunk and try to insert it etc.
For now I would even say to only do this this for the chunking and leave the non-chunked mode (eg. you only attach 5 dids) as it is already, just let it fail without trying to complete it.

bari12 (changes requested):

bari12 (approved):

mlassnig (approved):

PR #1547, opened 2018-09-19
Clients: chunk DIDs array with length above 1000; Fix #30

### LINKED ISSUE #30: Rucio attach should chunk the DIDSs array up accordingly to the 1000 server limits

hahahannes:
Clients: chunk DIDs array with length above 1000; Fix #30

### DIFF
--- a/bin/rucio
+++ b/bin/rucio
@@ -37,6 +37,7 @@ from __future__ import print_function
 
 import argparse
 import logging
+import math
 import os
 import signal
 import subprocess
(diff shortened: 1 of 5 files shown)


bin/rucio
```diff
@@ -451,24 +451,28 @@ def attach(args):
...
+            f = open(dids[0], 'r')
+            dids = [did.rstrip() for did in f.readlines()]
         except IOError:
-            logger.error("Can't open file '" + args.dids[0] + "'.")
+            logger.error("Can't open file '" + dids[0] + "'.")
             return FAILURE
-    else:
-        for did in args.dids:
+
+    chunks = [dids[i:i + limit] for i in range(0, len(dids), limit)]
```
TWAtGH: Hi I think there is a util function for this:
https://github.com/rucio/rucio/blob/master/lib/rucio/common/utils.py#L374

Could check it out whether it can be used here

hahahannes: Ah okay perfect. Thanks

hahahannes:
It seems that sqlite has a problem with too many variables in a query according to the [documentation of limits (Nr. 9)](https://www.sqlite.org/limits.html). The limit is by default 999 which means that the dids must be chunked with length 499 because the did gets stripped into two variables.

bari12:
We only use sqllite for testing, so I wonder if it is not better to only adapt the sqlite test and not limit all the other installations running a different database.
I know, in the real deployment the client will not know which database is run, but I think we can just chunk on a 1000 and rely that nobody will run sqlite in production.

bari12 (approved):

PR #1562, opened 2018-09-26
Transfers: Do not send checksum if verify_checksum is fully disabled; Fix #1522

### LINKED ISSUE #1522: Don't send checksum to FTS if verify_checksum is disabled
Motivation
----------
Currently, when submitting a job to FTS where both RSEs have `"verify_checksum": False` it will be set to `"none"` in the metadata send to FTS but the checksum are still sent. This leads to certain problems with xrootd 3rd copy on DPM where checksum validation is not working. If a checksum is set in the metadata FTS will currently ignore the `"verify_checksum"` and will try to validate anyway.

Modification
------------
Don't send the checksum in the metadata when `"verify_checksum": "none"`

bari12:
Transfers: Do not send checksum if verify_checksum is fully disabled; Fix #1522

### DIFF
--- a/lib/rucio/daemons/conveyor/common.py
+++ b/lib/rucio/daemons/conveyor/common.py
@@ -178,7 +178,7 @@ def bulk_group_transfer(transfers, policy='rule', group_bulk=200, fts_source_str
                 'selection_strategy': fts_source_strategy,
                 'request_type': transfer['file_metadata'].get('request_type', None),
                 'activity': str(transfer['file_metadata']['activity'])}
-        if file['metadata'].get('verify_checksum', True):
+        if verify_checksum != 'none':
             if 'md5' in list(file['metadata'].keys()) and file['metadata']['md5']:
                 file['checksum'] = 'MD5:%s' % str(file['metadata']['md5'])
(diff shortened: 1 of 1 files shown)


lib/rucio/daemons/conveyor/common.py
```diff
@@ -178,7 +178,7 @@ def bulk_group_transfer(transfers, policy='rule', group_bulk=200, fts_source_str
                 'selection_strategy': fts_source_strategy,
                 'request_type': transfer['file_metadata'].get('request_type', None),
                 'activity': str(transfer['file_metadata']['activity'])}
-        if file['metadata'].get('verify_checksum', True):
+        if verify_checksum != 'none':
```
mlassnig, line 181: ` != 'none'` looks weird? Shouldn't this be either `is not None` (Python) or `!= 'null'` (JSON) ?

bari12: The options are 'both', 'source', 'destination', 'none' 
See https://github.com/rucio/rucio/blob/aff3b74312bd6e5bc5839cca06c1f221aa340712/lib/rucio/core/transfer.py#L662
I know, it's a bit weird, but these are strings given to FTS.

mlassnig (approved):

PR #1565, opened 2018-09-26
replicas: allow listing on RSEs without site attribute; Fix #1528

### LINKED ISSUE #1528: Site attribute check for lan/wan selection needs to be optional
Motivation
----------
Currently list_replicas checks the `site` attribute of the RSE and matches it with the `client_location`. However, if the site attribute is not set the method crashes. This should be changed and just the wan replica should be returned.

mlassnig:
replicas: allow listing on RSEs without site attribute; Fix #1528

### DIFF
--- a/lib/rucio/core/replica.py
+++ b/lib/rucio/core/replica.py
@@ -900,8 +900,16 @@ def _list_replicas(dataset_clause, file_clause, state_clause, show_pfns,
                         # cannot be pushed into protocols because we need to lookup rse attributes.
                         # ultra-conservative implementation.
                         if domain == 'wan' and protocol.attributes['scheme'] == 'root' and client_location:
+
                             if 'site' in client_location and client_location['site']:
-                                replica_site = get_rse_attribute('site', rse_info[rse]['id'], session=session)[0]
+
+                                # is the RSE site-configured?
+                                rse_site_attr = get_rse_attribute('site', rse_info[rse]['id'], session=session)
+                                replica_site = ['']
+                                if isinstance(rse_site_attr, list) and len(rse_site_attr):
+                                    replica_site = rse_site_attr[0]
+
+                                # does it match with the client?
                                 if client_location['site'] != replica_site:
                                     root_proxy_internal = get_rses_with_attribute_value('site', client_location['site'],
                                                                                         'root-proxy-internal',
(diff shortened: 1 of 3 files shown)


lib/rucio/core/replica.py
```diff
@@ -900,8 +900,16 @@ def _list_replicas(dataset_clause, file_clause, state_clause, show_pfns,
...
                             if 'site' in client_location and client_location['site']:
-                                replica_site = get_rse_attribute('site', rse_info[rse]['id'], session=session)[0]
+
+                                # is the RSE site-configured?
+                                rse_site_attr = get_rse_attribute('site', rse_info[rse]['id'], session=session)
+                                replica_site = ['']
+                                if isinstance(rse_site_attr, list) and len(rse_site_attr):
+                                    replica_site = rse_site_attr[0]
+
+                                # does it match with the client?
```
vingar, line 912: I'm wondering if `get_rse_attribute` has some caching enabled. If not: If the same information is asked several times for the same site or rse, please consider caching the result in some top variable like `rse_attributes[rse_id]{'site': xxx, '' }` and only query the db if not found in the dictionary.

lib/rucio/core/rse.py
```diff
@@ -445,18 +445,26 @@ def get_rse_attribute(key, rse_id=None, value=None, session=None):
...
-            query = session.query(models.RSEAttrAssociation.value).filter_by(rse_id=rse_id, key=key, value=value).distinct()
-    else:
-        query = session.query(models.RSEAttrAssociation.value).filter_by(key=key).distinct()
-        if value:
-            query = session.query(models.RSEAttrAssociation.value).filter_by(key=key, value=value).distinct()
-    for attr_value in query:
-        rse_attrs.append(attr_value[0])
-    return rse_attrs
+
+    result = REGION.get('%s-%s-%s' % (key, rse_id, value))
```
bari12, line 449: I am wondering if this is conflicting with respect to line 407.
Because there the cache is populated with (key, value, lookup_key) but a list of rses is stored. You store (key, rse_id, value) but the rse_attrs are stored. I wonder if there is a combination where the tuple is the same, which would then crash the procedure.
It might be better in both caches to prefix it with ('get_rse_attribute-%s-%s-%s') or ('get_rses_with_attribute_value-%s-%s-%s') respectively.

bari12 (approved):

PR #1575, opened 2018-09-27
Clients: resolve archives per default; Fix #1560

### LINKED ISSUE #1560: Resolve archives per default in clients
Motivation
----------

Should be default true everywhere and also work for datasets, not only files.

hahahannes:
Clients: resolve archives per default; Fix #1560


TWAtGH:
Please change this also for the `rucio download`
https://github.com/rucio/rucio/blob/master/bin/rucio#L939
https://github.com/rucio/rucio/blob/master/bin/rucio#L1989

I will do the changes in downloadclient.py myself.

bari12 (approved):

PR #1611, opened 2018-10-03
Propagate the size of the file when declaring bad file : Closes #1610

### LINKED ISSUE #1610: Propagate the size of the file when declaring bad file
Motivation
----------
The `bad_replicas` table was extended with a bytes column for monitoring purpose. This column needs to be populated when a bad file is declared

cserf:
Propagate the size of the file when declaring bad file : Closes #1610

### DIFF
--- a/lib/rucio/core/replica.py
+++ b/lib/rucio/core/replica.py
@@ -119,19 +119,19 @@ def __exists_replicas(rse_id, scope=None, name=None, path=None, session=None):
             with_hint(models.RSEFileAssociation, "+ index(replicas REPLICAS_PATH_IDX", 'oracle').\
             filter(models.RSEFileAssociation.rse_id == rse_id).filter(or_(*path_clause))
     else:
-        query = session.query(models.RSEFileAssociation.path, models.RSEFileAssociation.scope, models.RSEFileAssociation.name, models.RSEFileAssociation.rse_id).\
+        query = session.query(models.RSEFileAssociation.path, models.RSEFileAssociation.scope, models.RSEFileAssociation.name, models.RSEFileAssociation.rse_id, models.RSEFileAssociation.bytes).\
             filter_by(rse_id=rse_id, scope=scope, name=name)
     if query.count():
         result = query.first()
-        path, scope, name, rse_id = result[0], result[1], result[2], result[3]
+        path, scope, name, rse_id, size = result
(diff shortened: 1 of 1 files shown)


lib/rucio/core/replica.py
```diff
@@ -115,7 +115,7 @@ def __exists_replicas(rse_id, scope=None, name=None, path=None, session=None):
             path_clause.append(models.RSEFileAssociation.path == path[1:])
         else:
             path_clause.append(models.RSEFileAssociation.path == '/%s' % path)
-        query = session.query(models.RSEFileAssociation.path, models.RSEFileAssociation.scope, models.RSEFileAssociation.name, models.RSEFileAssociation.rse_id).\
+        query = session.query(models.RSEFileAssociation.path, models.RSEFileAssociation.scope, models.RSEFileAssociation.name, models.RSEFileAssociation.rse_id, models.RSEFileAssociation.bytes).\
```
vingar, line 118: I would suggest to break this long line in several to improve readability.

vingar (commented):
One non blocking request. Otherwise looks Ok.

lib/rucio/core/replica.py
```diff
@@ -906,7 +906,7 @@ def _list_replicas(dataset_clause, file_clause, state_clause, show_pfns,
                                 # is the RSE site-configured?
                                 rse_site_attr = get_rse_attribute('site', rse_info[rse]['id'], session=session)
                                 replica_site = ['']
-                                if isinstance(rse_site_attr, list) and len(rse_site_attr):
```
mlassnig, line 909: this actually checks if there is an item in the list, you can't remove the `len` check

cserf: ??
The test is the same
>>> rse_site_attr = []
>>> rse_site_attr is False
False
>>> len(rse_site_attr) is False
False

mlassnig: aaargg, the original statement is wrong even that's why i got confused.
it really should be ... `and len(rse_site_attr)>0`

cserf: ???

`if len(rse_site_attr)>0` or `if len(rse_site_attr)` or `if rse_site_attr` should give the same result

mlassnig (approved):

PR #1628, opened 2018-10-10
Core & Internals: Porting PL/SQL procedure to abacus daemon; Fix #524

### LINKED ISSUE #524: Porting oracle jobs/procedures to non-oracle backends
Motivation
----------
At the moment there are two oracle procedures (cron jobs) being used to keep the data consistent. However, there needs to be a solution for non-oracle users. For Postgres it might be possible to also introduce PLSQL Cronjobs, but for mysql this is not possible, thus daemon implementations for this might be applicable.

- Update collection replicas: Exists only as PL/SQL
- Update of account counters: Exists as PL/SQL and in the Abacus daemon

hahahannes:
Core & Internals: Porting PL/SQL procedure to abacus daemon; Fix #524

### DIFF
--- a/bin/rucio-abacus-collection-replica
+++ b/bin/rucio-abacus-collection-replica
@@ -1,0 +1,51 @@
+#!/usr/bin/env python
+# Copyright 2014-2018 CERN for the benefit of the ATLAS collaboration.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Authors:
+# - Hannes Hansen, <hannes.jakob.hansen@cern.ch>, 2018
+
+"""
+Abacus collection replica is a daemon to update collection replicas.
+"""
+
+import argparse
+import signal
+
+from rucio.daemons.abacus.collection_replica import run, stop
+
+
+def get_parser():
+    """
+    Returns the argparse parser.
+    """
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--run-once", action="store_true", default=False, help='One iteration only')
+    parser.add_argument("--process", action="store", default=0, type=int, help='Concurrency control: current processes number')
+    parser.add_argument("--total-processes", action="store", default=1, type=int, help='Concurrency control: total number of processes')
+    parser.add_argument("--threads-per-process", action="store", default=1, type=int, help='Concurrency control: total number of threads per process')
+    return parser
+
+
+if __name__ == "__main__":
+
+    signal.signal(signal.SIGTERM, stop)
+    parser = get_parser()
+    args = parser.parse_args()
+    try:
+        run(once=args.run_once, process=args.process,
+            total_processes=args.total_processes,
+            threads_per_process=args.threads_per_process)
+    except KeyboardInterrupt:
+        stop()
--- a/bin/rucio-abacus-collection-replica
+++ b/bin/rucio-abacus-collection-replica
@@ -32,9 +32,7 @@ def get_parser():
     """
     parser = argparse.ArgumentParser()
     parser.add_argument("--run-once", action="store_true", default=False, help='One iteration only')
-    parser.add_argument("--process", action="store", default=0, type=int, help='Concurrency control: current processes number')
-    parser.add_argument("--total-processes", action="store", default=1, type=int, help='Concurrency control: total number of processes')
-    parser.add_argument("--threads-per-process", action="store", default=1, type=int, help='Concurrency control: total number of threads per process')
+    parser.add_argument("--threads", action="store", default=1, type=int, help='Concurrency control: total number of threads on this process')
     return parser
 
 
@@ -44,8 +42,6 @@ if __name__ == "__main__":
     parser = get_parser()
     args = parser.parse_args()
     try:
-        run(once=args.run_once, process=args.process,
-            total_processes=args.total_processes,
-            threads_per_process=args.threads_per_process)
+        run(once=args.run_once, threads=args.threads)
     except KeyboardInterrupt:
         stop()
--- a/lib/rucio/core/replica.py
+++ b/lib/rucio/core/replica.py
@@ -21,11 +21,11 @@
 # - David Cameron <d.g.cameron@gmail.com>, 2014
 # - Thomas Beermann <thomas.beermann@cern.ch>, 2014-2018
 # - Wen Guan <wguan.icedew@gmail.com>, 2014-2015
+# - Hannes Hansen <hannes.jakob.hansen@cern.ch>, 2018
 #
 # PY3K COMPATIBLE
 
 from __future__ import print_function
-
 from collections import defaultdict
 from copy import deepcopy
 from curses.ascii import isprint
@@ -36,6 +36,7 @@ from traceback import format_exc
 
 from sqlalchemy import func, and_, or_, exists, not_
 from sqlalchemy.exc import DatabaseError, IntegrityError
+from sqlalchemy.sql import label
 from sqlalchemy.orm.exc import FlushError, NoResultFound
 from sqlalchemy.sql.expression import case, bindparam, select, text, false, true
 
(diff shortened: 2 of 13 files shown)


tools/run_tests.sh
```diff
@@ -70,6 +70,9 @@ find lib -iname "*.pyc" | xargs rm
 echo 'Cleaning old authentication tokens'
 rm -rf /tmp/.rucio_*/
 
+echo 'Cleaning RSEs'
```
hahahannes, line 74: I added this because I had some problems with uploading files and restarting the tests as they are not in the database anymore.

lib/rucio/tests/test_dataset_replicas.py
```diff
@@ -124,3 +127,211 @@ def test_list_dataset_replicas_archive(self):
...
+        update_request = self.db_session.query(models.UpdatedCollectionReplica).filter_by(rse_id=self.rse_id, scope=self.scope, name=dataset_name).one()
+        update_collection_replica(update_request_id=update_request.id, session=self.db_session)
+        dataset_replica = self.db_session.query(models.CollectionReplica).filter_by(scope=self.scope, name=dataset_name).one()
+        assert_equal(dataset_replica['bytes'], len(files) * file_size)
+        assert_equal(dataset_replica['length'], len(files))
+        assert_equal(dataset_replica['available_bytes'], len(files) * file_size)
+        assert_equal(dataset_replica['available_replicas_cnt'], len(files))
+        assert_equal(str(dataset_replica['state']), 'AVAILABLE')
+
+        # Delete all file replicas -> dataset replica should be deleted
```
hahahannes, line 223: There is a difference between replica updates with and with out rse ids. If all file replicas of a collection are deleted, a collection replica gets also deleted if the replica update has a rse id, which makes sense to me. But the collection replica is not deleted if the replica update has no rse id. In this case nothing is happening, according to the procedure?

lib/rucio/tests/test_dataset_replicas.py
```diff
@@ -124,3 +127,211 @@ def test_list_dataset_replicas_archive(self):
...
+        assert_equal(dataset_replica['available_replicas_cnt'], len(files) - 1)
+        assert_equal(str(dataset_replica['state']), 'UNAVAILABLE')
+        dataset_replica = self.db_session.query(models.CollectionReplica).filter_by(scope=self.scope, name=dataset_name, rse_id=self.rse_id2).one()
+        assert_equal(dataset_replica['bytes'], len(files) * file_size)
+        assert_equal(dataset_replica['length'], len(files))
+        assert_equal(dataset_replica['available_bytes'], (len(files) - 1) * file_size)
+        assert_equal(dataset_replica['available_replicas_cnt'], len(files) - 1)
+        assert_equal(str(dataset_replica['state']), 'UNAVAILABLE')
+
+        # Delete first replica on second RSE -> file is not longer part of dataset -> both replicas should be available
```
hahahannes, line 286: Also if there are multiple replicas per file and the last file replica gets deleted, it gets also removed as a child of the dataset. If the replica update has no rse id, the columns 'bytes' and 'length' are then calculated without the replica and then the collection replica will be set as available which seems correct to me, because the file is no longer part of the dataset. But if the replica update has a rse id, the columns 'bytes' and 'length' are just copied from the last collection replica entry and it would be set as unavailable because the number of available replicas is less then the length (see line 197 where the collection replica will be unavailable).

tools/run_tests.sh
mlassnig: that's ok. but i think to be super-safe you could restrict it to

`rm -rf /tmp/rucio_rse/*` ?

lib/rucio/tests/test_abacus_rse.py
```diff
@@ -0,0 +1,88 @@
...
+        # Delete files -> rse usage should decrease
+        cleaner.run(once=True)
+        reaper.run(once=True, rses=[self.rse], greedy=True)
+        rse.run(once=True)
+        rse_usage = get_rse_usage(rse=self.rse)[0]
+        assert_equal(rse_usage['used'], 0)
+        rse_usage_from_rucio = get_rse_usage(rse=self.rse, source='rucio')[0]
+        assert_equal(rse_usage_from_rucio['used'], 0)
+        rse_usage_from_unavailable = get_rse_usage(rse=self.rse, source='unavailable')[0]
+        assert_equal(rse_usage_from_unavailable['used'], 2)
```
hahahannes: I would expect this to be 0 after the files are deleted, but was not sure about the sources stuff. I guess it is not updating because after the [group by](https://github.com/rucio/rucio/blob/e9b2d634c767d988e8620512fe3c92b04c4c1ab6/lib/rucio/core/rse_counter.py#L162) the list is empty and then the RSEUsage for source=unavailable does not get updated.

lib/rucio/tests/test_dataset_replicas.py
bari12: Yes, this is a bit specific to the workflow how the UpdatedCollectionReplicas are created. Essentially UpdatedCollectionReplicas with the rse_id not set, are never created when replicas are reduced, but only when they are increased. So the case where the CollectionReplica becomes size=0 is only when the rse_id of the UpdatedCollectionReplica is set.

lib/rucio/tests/test_dataset_replicas.py
bari12: I have to look a little bit into the workflow of this case.

bin/rucio-abacus-collection-replica
```diff
@@ -0,0 +1,51 @@
...
+
+import argparse
+import signal
+
+from rucio.daemons.abacus.collection_replica import run, stop
+
+
+def get_parser():
+    """
+    Returns the argparse parser.
```
bari12, line 31: Please do the arguments similar to https://github.com/rucio/rucio/blob/master/bin/rucio-judge-cleaner
So just --threads and --run-once
The rest should be figured out by heartbeats.

lib/rucio/daemons/abacus/collection_replica.py
```diff
@@ -0,0 +1,100 @@
...
+
+logging.basicConfig(stream=sys.stdout,
+                    level=getattr(logging,
+                                  config_get('common', 'loglevel',
+                                             raise_exception=False,
+                                             default='DEBUG').upper()),
+                    format='%(asctime)s\t%(process)d\t%(levelname)s\t%(message)s')
+
+
+def collection_replica_update(once=False, process=0, total_processes=1, thread=0, threads_per_process=1):
```
bari12: Add heartbeats e.g. similar to judge-cleaner

lib/rucio/core/replica.py
```diff
@@ -2261,3 +2263,175 @@ def mark_unlocked_replicas(rse, bytes, session=None):
...
+def get_cleaned_updated_collection_replicas(total_workers, worker_number, session=None):
+    """
+    Get update request for collection replicas.
+    :param total_workers:      Number of total workers.
+    :param worker_number:      id of the executing worker.
+    :param session:            Database session in use.
+    :returns:                  List of ids of update requests for collection replicas.
+    """
+    # Delete update requests which do not have collection_replicas and delete duplicates
+    replica_update_requests = session.query(models.UpdatedCollectionReplica).all()
```
bari12: Several changes with this.
- The cleanup has to select hashed too, right now it is not, which could potentially create conflicts with multiple threads.
- The cleanup is too complicated, it's too many queries as you basically have a single delete statement for each row.
You can replicate (+ hashing) these two queries here:
```
DELETE FROM ATLAS_RUCIO.UPDATED_COL_REP A WHERE A.rse_id IS NOT NULL AND NOT EXISTS(SELECT * FROM ATLAS_RUCIO.COLLECTION_REPLICAS B WHERE B.scope = A.scope AND B.name = A.name  AND B.rse_id = A.rse_id);
DELETE FROM ATLAS_RUCIO.UPDATED_COL_REP A WHERE A.rse_id IS NULL AND NOT EXISTS(SELECT * FROM ATLAS_RUCIO.COLLECTION_REPLICAS B WHERE B.scope = A.scope AND B.name = A.name);
```
This is than only one execution each to get rid of the ones which do not have collection replicas.
The duplicate removal You can than do like you do it now essentially.

lib/rucio/core/replica.py
```diff
@@ -2261,3 +2263,175 @@ def mark_unlocked_replicas(rse, bytes, session=None):
...
+    return [update_request.id for update_request in query.all()]
+
+
+@transactional_session
+def update_collection_replica(update_request_id, session=None):
+    """
+    Reads UpdatedCollectionReplicas and updates a collection replica.
+    :param update_request_id: update request from the upated_col_rep table.
+    """
+    update_request = session.query(models.UpdatedCollectionReplica).filter_by(id=update_request_id).one()
```
bari12: Also expensive, one extra query per row. It might be better if in the outer loop you already get a updated_collection_replica object which includes all this data.

lib/rucio/core/replica.py
bari12: Alternatively: Make a version without hashing first, and we do a second one later.

lib/rucio/core/rse_counter.py
```diff
@@ -128,15 +130,70 @@ def update_rse_counter(rse_id, session=None):
...
+                        used=sum_bytes,
+                        files=sum_files,
+                        source='rucio').save(session=session)
 
     for update in updated_rse_counters:
         update.delete(flush=False, session=session)
+
+
+@read_session
+def get_rse_usage_from_unavailable_replicas(total_workers, worker_number, session=None):
```
bari12: I am not quite sure where this requirement comes from? I don't think that the PL/SQL procedure fills unavailable replica info?

bari12 (changes requested):

lib/rucio/core/rse_counter.py
hahahannes: I thought it belongs to [this part of the procedure](https://github.com/rucio/rucio/blob/master/etc/sql/oracle/procedures.sql#L210)

bari12: Ahh I see; Hmm the issue is that in the procedure this is really an optimised query going against a special index in Oracle. For big tables this could be a bit of a show-stopper, but let's leave it for now. 👍

mlassnig:
ping. what's the status here?

bari12:
Still in review.

bari12 (approved):

PR #1629, opened 2018-10-10
Core & Internals: Porting PL/SQL procedure to abacus daemon; Fix #524

### LINKED ISSUE #524: Porting oracle jobs/procedures to non-oracle backends

hahahannes:
Core & Internals: Porting PL/SQL procedure to abacus daemon; Fix #524


lib/rucio/core/replica.py
```diff
@@ -2261,3 +2263,130 @@ def mark_unlocked_replicas(rse, bytes, session=None):
...
+    update_requests_without_rse_id = []
+    for update_request in replica_update_requests:
+        if update_request.rse_id is not None:
+            small_request = {'name': update_request.name, 'scope': update_request.scope, 'rse_id': update_request.rse_id}
+            if small_request not in update_requests_with_rse_id:
+                update_requests_with_rse_id.append(small_request)
+            else:
+                update_request.delete(session=session)
+                continue
+            found_replicas = session.query(models.CollectionReplica).filter(models.CollectionReplica.rse_id == update_request.rse_id, models.CollectionReplica.scope == update_request.scope, models.CollectionReplica.name == update_request.name).all()
```
mlassnig: you can cut line length, just put a backslash and start the next line with the dot of the function

(happens a couple of times, you can do it everywhere)

lib/rucio/core/replica.py
```diff
@@ -2261,3 +2263,130 @@ def mark_unlocked_replicas(rse, bytes, session=None):
...
+            if len(found_replicas) == 0:
```
mlassnig: shorter:
`if not found_replicas:`

mlassnig: (below as well)

lib/rucio/tests/test_abacus_account.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+class TestAbacusAccount():
+
+    def setUp(self):
+        self.account = 'root'
+        self.scope = 'mock'
+        self.upload_client = UploadClient()
+        self.file_sizes = 2
+        self.rse = 'MOCK4'
```
mlassnig, line 44: use rse_name_generator()

hahahannes: For this case I wanted to use one of the simulated RSEs where I can upload a real file. I am not sure if this is possible with a random RSE.

mlassnig: understood

PR #1640, opened 2018-10-11
WebUI: removed generic meta data from webui; Fix #1619

### LINKED ISSUE #1619: Searching for a DID shows "Could not find the DID" but did was found
Motivation
----------
See https://its.cern.ch/jira/browse/RUCIO-2883

hahahannes:
WebUI: removed generic meta data from webui; Fix #1619

### DIFF
--- a/lib/rucio/web/ui/static/did.js
+++ b/lib/rucio/web/ui/static/did.js
@@ -851,22 +851,6 @@ build_page = function() {
             $('#loading').html('<font color="red">Could not find the DID.</font>');
         }
     });
-
-    r.did_get_generic_metadata({
-        'scope': scope,
-        'name': name,
-        success: function(data) {
-            $("#loading").html(""); 
-            var sorted_keys = Object.keys(data).sort();
-            for(var i=0; i<sorted_keys.length; ++i) {
-                $('#t_generic_metadata').append($('<tr><th>' + sorted_keys[i] + '</th><td>' + data[sorted_keys[i]] + '</td></tr>'));
-            }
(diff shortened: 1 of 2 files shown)


lib/rucio/web/ui/static/did.js
```diff
@@ -851,22 +851,6 @@ build_page = function() {
             $('#loading').html('<font color="red">Could not find the DID.</font>');
         }
     });
-
```
mlassnig: not necessary to remove the function, is it?

hahahannes: shall I just comment it out for later use?

mlassnig: just leave it in. you removed the call in the other file

hahahannes: But the function is called on page load, I guess. Then the error message would be still visible.

bari12: The function can stay; I think what we should comment out is the error part (line 866-868) - so that the error is not displayed even if the did is there.

hahahannes: ok

bari12: I wonder though if we skip the error, if the table is still rendered, even if there is no data? (Maybe @tbeerman knows?)

hahahannes: The table seems to have a 'display: none' at the moment

bari12 (approved):

PR #1657, opened 2018-10-15
downloadclient: introducing tracing param. 1214

TomasJavurek:
downloadclient: introducing tracing param. 1214

### DIFF
--- a/lib/rucio/client/downloadclient.py
+++ b/lib/rucio/client/downloadclient.py
@@ -45,7 +45,7 @@ from rucio import version
 
 class DownloadClient:
 
-    def __init__(self, client=None, logger=None, check_admin=False):
+    def __init__(self, client=None, logger=None, tracing=True, check_admin=False):
         """
         Initialises the basic settings for an DownloadClient object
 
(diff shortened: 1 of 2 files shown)


lib/rucio/client/downloadclient.py
```diff
@@ -1089,3 +1090,15 @@ def _check_output(self, output_items):
...
+        """
+        Checks if sending trace is allowed and send the trace.
+
+        :param trace: the trace
+        """
+        logger = self.logger
+        if self.tracing:
+            send_trace(trace, self.client.host, self.client.user_agent)
+        else:
+            logger.info('Tracing was turned off.')
```
mlassnig: i would have this as debug only

TWAtGH: I would remove this completely at this point at output it just once in init()

lib/rucio/client/uploadclient.py
```diff
@@ -218,6 +220,8 @@ def upload(self, items, summary_file_path=None):
                     if not self.client.update_replicas_states(rse, files=[replica_for_api]):
                         logger.warning('Failed to update replica state')
             else:
+                self.trace['clientState'] = 'FAILED'
```
TWAtGH, line 225: @tbeerman said you need to fill the `self.trace['stateReason']` field. I think you can write the exception from `except (ServiceUnavailable, ResourceTemporaryUnavailable) as error:` into this as a string.

TWAtGH (dismissed):

Issue #1664, opened 2018-10-16
Rucio user account synchronization improvements (python paged search)


vokac:
Motivation
----------
Discussion in DDM mailing list and suboptimal LDAP access in synchronization script.

Modification
------------
Improved check_voms synchronization script - use python ldap module instead of parsing ldapsearch command line output. Updated code use TLS for paged LDAP search to get info about all account (in less than 10s).

PR #1666, opened 2018-10-16
patch: Account synchronization improvements (python paged search) #1664

vokac:
patch: Account synchronization improvements (python paged search) #1664


vokac:
Coding stile fixed in my local working area but I'm not completely sure how to push here updated version.

bari12:
It should be sufficient to commit it and then do a `git push origin patch-1664-ldap_paged_search --force`

PR #1678, opened 2018-10-17
Extract the suspicious files from the traces : Closes #1623

### LINKED ISSUE #1623: Extract the suspicious files from the traces
Motivation
----------
Currently the suspicious files are reported by the conveyor, the pilot and `rucio download`. We can centralize everything so that Kronos checks the traces and reports as suspicious all the files with errors that match a certain pattern

cserf:
Extract the suspicious files from the traces : Closes #1623

### DIFF
--- a/lib/rucio/daemons/tracer/kronos.py
+++ b/lib/rucio/daemons/tracer/kronos.py
@@ -139,7 +139,7 @@ class AMQConsumer(object):
         for report in self.__reports:
             try:
                 # Identify suspicious files
-                if self.__bad_files_patterns and report['eventType'] in ('get_sm', 'sm_get', 'get_sm_a', 'sm_get_a') and report['clientState'].find('FAILED') > -1:
+                if self.__bad_files_patterns and report['eventType'] in ['get_sm', 'get_sm_a', 'get'] and report['clientState'] not in ['DONE', 'FOUND_ROOT', 'ALREADY_DONE']:
                     for pattern in self.__bad_files_patterns:
                         if pattern.match(report['stateReason']):
                             reason = report['stateReason'][:255]
--- a/lib/rucio/daemons/tracer/kronos.py
+++ b/lib/rucio/daemons/tracer/kronos.py
@@ -18,33 +18,43 @@
 # - Vincent Garonne <vgaronne@gmail.com>, 2015-2018
 # - Mario Lassnig <mario.lassnig@cern.ch>, 2015
 # - Wen Guan <wguan.icedew@gmail.com>, 2015
+# - Cedric Serfon <cedric.serfon@cern,ch>, 2018
+#
+# PY3K COMPATIBLE
 
 """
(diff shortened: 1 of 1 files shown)


lib/rucio/daemons/tracer/kronos.py
```diff
@@ -127,6 +138,17 @@ def __update_atime(self):
         rses = []
         for report in self.__reports:
             try:
+                # Identify suspicious files
+                if self.__bad_files_patterns and report['eventType'] in ('get_sm', 'sm_get', 'get_sm_a', 'sm_get_a') and report['clientState'].find('FAILED') > -1:
```
tbeerman: You can remove `sm_get` and `sm_get_a`. They were traces sent with eventType some time ago by mistake but it is fixed now and they should not appear anymore.

tbeerman: Are you sure that get all needed clientStates with this? Looking at the last 30 days only at `get_sm/_a` I can see a lot of different error states and they do not all have "FAILED" in the name. Here is a list of all clientStates in the last 30 days that are not `DONE`, `FOUND_ROOT` or `ALREADY_DONE` ordered by count.

- `STAGEIN_ATTEMPT_FAILED`
- `NO_REPLICA`
- `STAGEIN_FILE_FAILED`
- `COPY_ERROR`
- `STAGEIN_FAILED`
- `NO_FILE`
- `CP_TIMEOUT`
- `GLOBUS_FAIL`
- `AD_MISMATCH`
- `BAD_COPYTOOL`
- `STAGEIN_BAD_COPYTOOL`
- `NO_SPACE`
- `NO_PROTOCOL_DEFINED`
- `ServiceUnavailable`
- `FILE_EXISTS`

lib/rucio/daemons/tracer/kronos.py
```diff
@@ -250,11 +272,20 @@ def kronos_file(once=False, thread=0, brokers_resolved=None, dataset_queue=None)
     chunksize = config_get_int('tracer-kronos', 'chunksize')
     prefetch_size = config_get_int('tracer-kronos', 'prefetch_size')
     subscription_id = config_get('tracer-kronos', 'subscription_id')
+    try:
+        bad_files_patterns = []
+        pattern = get(section='kronos', option='bad_files_patterns', session=None)
```
tbeerman, line 277: why not just use `config_get`?

tbeerman (dismissed):

lib/rucio/daemons/tracer/kronos.py
cserf: `config_get` gets the config from the `rucio.cfg`. It's better to get it from the config table.

lib/rucio/daemons/tracer/kronos.py
cserf: Where all the states are documented ?

cserf: OK

bari12 (approved):

PR #1699, opened 2018-10-23
Release: Requests 2.20 not python 2.6 compatible; Fix #1698

### LINKED ISSUE #1698: Requests 2.20.0 not Python 2.6 compatible
Motivation
----------
Need to fix to 2.19.* and to urllib3 < 1.24

bari12:
Release: Requests 2.20 not python 2.6 compatible; Fix #1698

### DIFF
--- a/tools/pip-requires-client
+++ b/tools/pip-requires-client
@@ -2,8 +2,8 @@
 setuptools>=36.8.0                                # Python packaging utilities (36.8.0 last py2.6 compatible version)
 argparse>=1.4.0; python_version == '2.6'          # Command-line parsing library
 argcomplete>=1.9.4                                # Bash tab completion for argparse
-requests>=2.6.0                                   # Python HTTP for Humans.
-urllib3==1.23                                     # HTTP library with thread-safe connection pooling, file post, etc.
+requests>=2.6.0,<2.20.0                           # Python HTTP for Humans.
+urllib3>=1.23,<1.24                               # HTTP library with thread-safe connection pooling, file post, etc.
 dogpile.cache>=0.6.5                              # Caching API plugins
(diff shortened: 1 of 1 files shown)


vingar:
Two package updates to consider as well:
 dogpile.cache package update to 0.6.7
 progressbar2 package update to 3.38.0

bari12:
I looked a bit more, and it is a bit more complicated.
- `argparse` We only install in 2.6 anyway
- `argcomplete` 1.9.4 is not flagged as 2.7 compatible. Seems to work fine. Actually none of the versions is flagged as 2.6 compatible. Would suggest to make it `<=1.9.4`
- `requests` already done with `2.6.0,<2.20.0`
- `urllib3` already done with `>=1.23,<1.24`
- `dogpile.cache` Even 0.6.5 is marked as only Python3 ;-) It seems to work; Would suggest to do `<=0.6.7`
- `boto` we only do for 2.7
- `tabulate` 0.8.2 not flagged as 2.6 compatible, seems to work fine; Would suggest to do `<=0.8.2`
- `progressbar2` 3.38.0 (newest, from May) is not marked 2.6 compatible, but seems to work just fine. The newest 2.6 version they flag is 3.1.0 from October 2015. I would propose here to allow anything `<=3.38.0` as it seems to work just fine.
- `bz2file` 0.98 is fine and compatible. Would suggest to do `<=0.98`
- `python-magic` 0.4.15 is fine --> `<=0.4.15`
- `futures` for python<3
- `six` should be fine as it is a compatibility library.

cserf:
Maybe we merge this one so that one can rebase the waiting PR that failed the py2.6 check.

mlassnig (approved):

PR #1706, opened 2018-10-25
Core & Internals: delete rse only if it is empty; Fix #26

### LINKED ISSUE #26: Protection in rucio-admin rse delete
We need a protection for deleting a RSE. In particular, it should fail if some rules, collection replicas exist.

_Migrated from JIRA [RUCIO-2103](https://its.cern.ch/jira/browse/RUCIO-2103)_

hahahannes:
Core & Internals: delete rse only if it is empty; Fix #26

### DIFF
--- a/lib/rucio/core/rse.py
+++ b/lib/rucio/core/rse.py
@@ -164,6 +164,9 @@ def del_rse(rse, session=None):
     :param session: The database session in use.
     """
 
+    old_rse = None
+    if not rse_is_empty(rse=rse, session=session):
+        raise exception.RSEOperationNotSupported('RSE \'%s\' is not empty' % rse)
     try:
         old_rse = session.query(models.RSE).filter_by(rse=rse).one()
     except sqlalchemy.orm.exc.NoResultFound:
@@ -173,6 +176,19 @@ def del_rse(rse, session=None):
         del_rse_attribute(rse=rse, key=rse, session=session)
     except exception.RSEAttributeNotFound:
         pass
+
+
+@read_session
+def rse_is_empty(rse, session=None):
+    """
+    Check if no file replica belongs to the RSE.
+
+    :param rse: the rse name.
+    :param session: the database session in use.
+    """
+
+    rse_id = get_rse(rse).id
+    return session.query(models.RSEFileAssociation).filter_by(rse_id=rse_id).count() == 0
 
 
 @read_session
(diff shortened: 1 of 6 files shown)


lib/rucio/core/rse.py
```diff
@@ -168,6 +171,19 @@ def del_rse(rse, session=None):
...
+def rse_is_empty(rse, session=None):
+    """
+    Check if no file replica belongs to the RSE.
+
+    :param rse: the rse name.
+    :param session: the database session in use.
+    """
+
+    rse_id = get_rse(rse).id
+    return session.query(models.RSEFileAssociation).filter_by(rse_id=rse_id).count() == 0
```
cserf: This can take a very long time. The table is not indexed by `rse_id` and it will go to full table scan. We might need to set an index. Other possibility : use the replicas dumps.

mlassnig: another index on replicas is maybe not so good, it will blow up the tablespace. what about using collection replicas? that would be much smaller and faster. for the result from collection replicas an additional deep check to replicas can be done afterwards?

mlassnig: or even rse counters?

hahahannes: ah yes, rse counters makes more sense. It already has the information about the rse usage. I guess there is some delay in updating it but I don't think this is a big problem?

mlassnig: that's not really a problem. the more problematic part is when the abacus fails and things are at zero and then this function would believe it's empty when in fact it's not.

hahahannes: I guess collection replicas can not be used also, because there can be files on the RSE that are not attached to a collection, if I understood it correctly.

And what is a replica dump actually?

bari12: Replica dump is not a good idea I think. It seems wrong to introduce this on server nodes.
I think rse_counters is a good idea. They are sometimes out of sync, but deleting an rse is a rare occassion. If the deletion fails because of the counters being out of sync, we just have to re-sync the counters.
👍 for counters

bari12: Btw. Just for the record: I tried to run this query on a deleted_rse and it takes about 50m to finish. On the ATLAS instance (10m timeout) it would never finish 😄

bari12 (approved):

Issue #368, opened 2017-12-14
Automatically add dark data to quarantine table


bari12:
_Migrated from [JIRA](https://its.cern.ch/jira/browse/RUCIO-2238)_

PR #1708, opened 2018-10-26
Consistency: Automatically add dark data to quarantine table #368

dchristidis:
Summary of changes:

1. Automatically process `DARK` data reported by the Auditor (issue #368).
2. Automatically compress produced files after their successful processing.
3. Lay down the groundwork for also processing `LOST` files.
4. Protect the Core from database errors.
5. Other small code improvements and fixes.

The patch was successfully tested on the production node.

### DIFF
--- a/lib/rucio/daemons/auditor/__init__.py
+++ b/lib/rucio/daemons/auditor/__init__.py
@@ -20,9 +20,10 @@
 # - Dimitrios Christidis <dimitrios.christidis@cern.ch>, 2018
 
 import Queue
+import bz2
 import glob
 import logging
-import os.path
+import os
 import select
 import sys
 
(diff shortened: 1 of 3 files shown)


lib/rucio/daemons/auditor/__init__.py
```diff
@@ -66,6 +65,113 @@ def consistency(rse, delta, configuration, cache_dir, results_dir):
...
+                    raise ValueError('unexpected label')
+    # Since the file is read immediately after its creation, any error
+    # exposes a bug in the Auditor.
+    except Exception as error:
+        logger.critical('Error processing "%s"', output, exc_info=True)
+        raise error
+
+    client = Client()
+    rse = os.path.basename(output[:output.rfind('_')])
+    usage = next(client.get_rse_usage(rse, filters={'source': 'rucio'}))
```
cserf: Why not using the core method ?

bari12: I think so too. Would be better to use Core methods here. (I don't think we have any daemon using client calls)

cserf (approved):

PR #1714, opened 2018-10-26
Probe : Probe to cancel FTS jobs for site in long downtime : Closes #543

### LINKED ISSUE #543: Probe to cancel FTS jobs for site in long downtime
Motivation
----------
If a site goes into a long downtime, but many transfers are already submitted to FTS, there will be a lot of failures.

Modification
------------
A probe collecting the downtimes in AGIS will cancel the FTS jobs if a site is supposed to be down for more than 24 hours.

cserf:
Probe to cancel FTS jobs for site in long downtime : Closes #543

### DIFF
--- a/lib/rucio/transfertool/fts3.py
+++ b/lib/rucio/transfertool/fts3.py
@@ -13,6 +13,8 @@
 # - Eric Vaandering, <ewv@fnal.gov>, 2018
 # - Diego Ciangottini <diego.ciangottini@pg.infn.it>, 2018
 # - Cedric Serfon <cedric.serfon@cern.ch>, 2018
+#
+# PY3K COMPATIBLE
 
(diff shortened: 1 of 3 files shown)


lib/rucio/transfertool/fts3.py
```diff
@@ -427,6 +433,76 @@ def bulk_query(self, transfer_ids, timeout=None):
 
         return responses
 
+    def list_se_status(self):
```
mlassnig, line 436: does this need to be upscaled to the parent transfertool interface?

bari12: I don't think so. Sounds like something quite specific to FTS and not really generic.

mlassnig (approved):

PR #1719, opened 2018-10-29
Core & Internals: set and get account properties; Fix #1685

### LINKED ISSUE #1685: Implement missing update_account method
Motivation
----------
There is a `set_account_status` to update the status of an account, but there is no `update_account` method to update the account email for instance. Proposal is to get rid of  `set_account_status` and replace it by `update_account(account, key, value)`. Change is needed at all levels (core, API, REST, CLI)

hahahannes:
Core & Internals: set and get account properties; Fix #1685 

Summary of changes:
- replaced `set_account_status` with `update_account`
- replaced `get_account_status` with `get_account_info` as it is only used in the tests
- added rucio-admin command to update account properties
- updated tests
- python3 compatibility

### DIFF
--- a/lib/rucio/api/account.py
+++ b/lib/rucio/api/account.py
@@ -12,6 +12,7 @@
 # - Martin Barisits, <martin.barisits@cern.ch>, 2014
 # - Joaquin Bogado, <joaquin.bogado@cern.ch>, 2015
 # - Cedric Serfon, <cedric.serfon@cern.ch>, 2015
+# - Hannes Hansen, <hannes.jakob.hansen@cern.ch>, 2018
 #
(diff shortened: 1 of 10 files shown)


lib/rucio/api/account.py
```diff
@@ -70,26 +71,18 @@ def get_account_info(account):
     return account_core.get_account(account)
 
 
-def get_account_status(account):
```
mlassnig, line 73: is this removed on purpose?

hahahannes: yes, as discussed with @cserf. Because it is only used in one test.

mlassnig (approved):

PR #1727, opened 2018-10-31
Clients: Possibility to register replica after successful upload; Fix #1599

### LINKED ISSUE #1599: Option to register replica only after upload
Motivation
----------

The current workflow when uploading a file is to register the replica as `UNAVAILABLE` first; Process the upload; Switch the replica to `AVAILABLE`. When the upload fails, the command can just be re-executed, however it is verified that a file with a similar checksum is uploaded.
We recently were told about official workflows where the re-upload, after a failure, would be done with a different file; Thus same filename but different checksum, which does not work in this workflow. The proposed solution is to offer a switch `--register-after-upload` (or something similar) where the client uploads first and registers the replica only as `AVAILABLE` when successful. Thus the pre-registration as `UNAVAILABLE` is skipped. In this case however, if the upload is repeated, the destination file always has to be over-written, as it could always be a new file being uploaded, as long as it is not registered in the catalog yet.

hahahannes:
Clients: Possibility to register replica after successful upload; Fix #1599

### DIFF
--- a/lib/rucio/client/uploadclient.py
+++ b/lib/rucio/client/uploadclient.py
@@ -19,6 +19,7 @@
 # - Martin Barisits <martin.barisits@cern.ch>, 2017
 # - Tobias Wegner <twegner@cern.ch>, 2018
 # - Nicolo Magini <nicolo.magini@cern.ch>, 2018
+# - Hannes Hansen <hannes.jakob.hansen@cern.ch>, 2018
 #
 # PY3K COMPATIBLE
 
@@ -71,18 +72,19 @@ class UploadClient:
         """
 
         :param items: List of dictionaries. Each dictionary describing a file to upload. Keys:
-            path             - path of the file that will be uploaded
-            rse              - rse name (e.g. 'CERN-PROD_DATADISK') where to upload the file
-            did_scope        - Optional: custom did scope (Default: user.<account>)
-            did_name         - Optional: custom did name (Default: name of the file)
-            dataset_scope    - Optional: custom dataset scope
-            dataset_name     - Optional: custom dataset name
-            force_scheme     - Optional: force a specific scheme (if PFN upload this will be overwritten) (Default: None)
-            pfn              - Optional: use a given PFN (this sets no_register to True, and no_register becomes mandatory)
-            no_register      - Optional: if True, the file will not be registered in the rucio catalogue
-            lifetime         - Optional: the lifetime of the file after it was uploaded
-            transfer_timeout - Optional: time after the upload will be aborted
-            guid             - Optional: guid of the file
+            path                  - path of the file that will be uploaded
+            rse                   - rse name (e.g. 'CERN-PROD_DATADISK') where to upload the file
+            did_scope             - Optional: custom did scope (Default: user.<account>)
+            did_name              - Optional: custom did name (Default: name of the file)
+            dataset_scope         - Optional: custom dataset scope
+            dataset_name          - Optional: custom dataset name
+            force_scheme          - Optional: force a specific scheme (if PFN upload this will be overwritten) (Default: None)
+            pfn                   - Optional: use a given PFN (this sets no_register to True, and no_register becomes mandatory)
+            no_register           - Optional: if True, the file will not be registered in the rucio catalogue
+            register_after_upload - Optional: if True, the file will be registered after successful upload
+            lifetime              - Optional: the lifetime of the file after it was uploaded
+            transfer_timeout      - Optional: time after the upload will be aborted
+            guid                  - Optional: guid of the file
         :param summary_file_path: Optional: a path where a summary in form of a json file will be stored
 
         :returns: 0 on success
@@ -131,6 +133,7 @@ class UploadClient:
             logger.info('Preparing upload for file %s' % basename)
 
             no_register = file.get('no_register')
+            register_after_upload = file.get('register_after_upload')
             pfn = file.get('pfn')
             force_scheme = file.get('force_scheme')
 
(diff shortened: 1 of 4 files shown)


lib/rucio/client/uploadclient.py
```diff
@@ -153,11 +156,23 @@ def upload(self, items, summary_file_path=None):
...
 
-            # if file already exists on RSE we're done
-            if not is_deterministic and not no_register:
+            # if register_after_upload, file should be overwritten if it is not registered
+            # otherwise if file already exists on RSE we're done
+            if register_after_upload:
+                if rsemgr.exists(rse_settings, file_did):
+                    logger.debug([replica for replica in self.client.list_replicas(dids=[{'name': file['did_name'], 'scope': file['did_scope']}])])
+
+                    if not [replica for replica in self.client.list_replicas(dids=[{'name': file['did_name'], 'scope': file['did_scope']}])]:
```
TWAtGH: Maybe it's possible to replace list_replicas with a less expensive call?
Also please remove the double call for the debug output.

lib/rucio/client/uploadclient.py
```diff
@@ -153,11 +156,23 @@ def upload(self, items, summary_file_path=None):
...
+                        logger.info('File already exists on RSE. Existing replica will be removed first.')
+                        rsemgr.delete(rse_settings, [{'name': file['did_name'], 'scope': file['did_scope']}])
+                        logger.info('Successfully removed replica from RSE')
+                    else:
+                        logger.info('File already registered. Skipping upload.')
+                        continue
+            elif not is_deterministic and not no_register and not register_after_upload:
```
TWAtGH: `not register_after_upload` can be removed since the `elif` implies it

lib/rucio/client/uploadclient.py
```diff
@@ -131,6 +133,7 @@ def upload(self, items, summary_file_path=None):
             logger.info('Preparing upload for file %s' % basename)
 
             no_register = file.get('no_register')
+            register_after_upload = file.get('register_after_upload')
```
TWAtGH: I think its better to force `register_after_upload = False` when `no_register = True`, e.g., `register_after_upload = file.get('register_after_upload') and not no_register`.

lib/rucio/client/uploadclient.py
```diff
@@ -153,11 +156,23 @@ def upload(self, items, summary_file_path=None):
...
 
-            if not no_register:
+            if not no_register and not register_after_upload:
                 self._register_file(file, registered_dataset_dids)
 
-            # if file already exists on RSE we're done
-            if not is_deterministic and not no_register:
+            # if register_after_upload, file should be overwritten if it is not registered
+            # otherwise if file already exists on RSE we're done
+            if register_after_upload:
```
TWAtGH, line 165: This should be fine if you force `register_after_upload = False` as suggested above

TWAtGH (changes requested):

TWAtGH (dismissed):

lib/rucio/client/uploadclient.py
```diff
@@ -153,11 +156,23 @@ def upload(self, items, summary_file_path=None):
...
+                if rsemgr.exists(rse_settings, file_did):
+                    try:
+                        dids = self.client.get_did(file['did_scope'], file['did_name'])
+                        logger.info('File already registered. Skipping upload.')
+                        logger.debug(dids)
```
bari12: Don't think we should leave the debug in?

lib/rucio/client/uploadclient.py
```diff
@@ -153,11 +156,23 @@ def upload(self, items, summary_file_path=None):
...
+                        continue
+                    except DataIdentifierNotFound:
+                        logger.info('File already exists on RSE. Existing replica will be removed first.')
+                        rsemgr.delete(rse_settings, [{'name': file['did_name'], 'scope': file['did_scope']}])
```
bari12: I don't think this workflow is correct.
rsemgr.upload(...) already takes care of the proper deletion if an incomplete file got uploaded (Check rsemanager.upload and the protocols). So you don't have to do this here. I think it might be enough to just call the upload and catch the exception and then do the registration

bari12 (dismissed):

lib/rucio/client/uploadclient.py
hahahannes: Does it only count for incomplete files?
Because in my current test I upload a file and only remove it from the dids to force the overwrite, but it does not work, because the file already exists according to the posix exist() check.

bari12:
Sorry for the late follow up on the review. For me it looks good. 👍 @TWAtGH can you please have a look again as well.
The tests seem to fail right now though

```
FAIL: CLIENT(USER): Rucio upload files with registration after upload
  File "/usr/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/opt/rucio/lib/rucio/tests/test_bin_rucio.py", line 270, in test_upload_file_register_after_upload
    nose.tools.assert_true((self.upload_success_str % path.basename(tmp_file4)) in out)
AssertionError: False is not true
```

hahahannes:
Yes, the test fails because of my simulation of a broken upload. I uploaded a file and removed the DID. When I then try to reupload the file it fails at the moment because the protocol existence check says that the file already exists. 
Before my last commit, I deleted the replica before reuploading it like this:
```python
logger.info('File already exists on RSE. Existing replica will be removed first.')
rsemgr.delete(rse_settings, [{'name': file['did_name'], 'scope': file['did_scope']}])
``` 
Should I put it back oder remove the test case? I was not sure if the test case was correct because it is not really a broken upload.

Edit: It's this [test](https://github.com/rucio/rucio/pull/1727/files#diff-90834bcf0a2db0761cdad97c256665dcR255)

bari12:
I think the problem here is that posix is a renaming protocol. With a renaming protocol the file is first uploaded as `filename.upload` and once successful renamed to `filename`. How the rsemanager works right now, if something is already in it's final state (`filename`) it is considered to be done. Only `filename.upload` would be re-tried again.

On the other hand you are testing a valid case where you upload (+rename) and then the process crashes (and the post-registration is never done); Maybe it is valid to say in this case (`exception.FileReplicaAlreadyExists`) that we delete the replica and re-upload it, since it is not registered yet, but I am not 100% sure. @cserf any input?

cserf:
OK to delete and reupload. Maybe we can have another exception for this e.g. `FileBrokenUpload` or whatever better name.

bari12:
Ok so in the `--registration-after-upload` case we delete the file, even if the full file (`filename`)already exists. I think however that this should be handled in the rsemanager than. Maybe with a new parameter in `upload` `delete_existing=True` or so. Not a huge fan of doing the deletion protocol handling directly in the uploadclient.

bari12 (approved):

PR #1742, opened 2018-11-02
Core & Internals: added python3 compatibility to API; Fix #1738

### LINKED ISSUE #1738: REST API migration to python3
Motivation
----------
Update REST API (webpy, flask) to be compatible with python3. 
Also needed for #1026 

Modification
------------

hahahannes:
Core & Internals: added python3 compatibility to API; Fix #1738

### DIFF
--- a/lib/rucio/web/rest/flaskapi/v1/archive.py
+++ b/lib/rucio/web/rest/flaskapi/v1/archive.py
@@ -17,7 +17,11 @@
 # - Vincent Garonne <vincent.garonne@cern.ch>, 2017
 # - Mario Lassnig <mario.lassnig@cern.ch>, 2018
 # - Thomas Beermann <thomas.beermann@cern.ch>, 2018
+# - Hannes Hansen, <hannes.jakob.hansen@cern.ch>, 2018
+#
+# PY3K COMPATIBLE
 
+from __future__ import print_function
 from json import dumps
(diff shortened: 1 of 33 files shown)


lib/rucio/web/rest/flaskapi/v1/archive.py
```diff
@@ -54,7 +58,7 @@ def get(self, scope, name):
                 data += dumps(file) + '\n'
             return Response(data, content_type="application/x-json-stream")
         except Exception, error:
```
mlassnig: shouldn't this be `except Exception as error` for py3? (there's three occurrences like this in this merge request)

hahahannes: yes, strangely running `pylint --py3k lib/rucio/web/rest/flaskapi/v1/archive.py` gives no error message about this.
But `2to3 lib/rucio/web/rest/flaskapi/v1/archive.py` does. I am checking the problem with pylint now. Otherwise we should maybe switch to 2to3.

mlassnig (approved):

lib/rucio/web/rest/flaskapi/v1/archive.py
hahahannes: Linking [this pylint issue](https://github.com/PyCQA/pylint/issues/2582) for the missing catch of the outdated exception handling.

PR #1744, opened 2018-11-02
schema: Single quotes shouldn't be allowed in scope name. Fix #1739

### LINKED ISSUE #1739: Scope name validation regex allows single quotes
The scope validation regex for atlas/generic/icecube (but not cms) allows single quotes, which seems like a bad idea.

`"^[a-zA-Z'_'-.0-9]{1,%s}$" % SCOPE_LENGTH` looks wrong, and I'm guessing it ought to be `"^[a-zA-Z_\\-.0-9]{1,%s}$" % SCOPE_LENGTH`. Unless someone really does have this requirement?

illingwo:
schema: Single quotes shouldn't be allowed in scope name. Fix #1739

### DIFF
--- a/lib/rucio/common/schema/generic.py
+++ b/lib/rucio/common/schema/generic.py
@@ -49,7 +49,7 @@ SCOPE_LENGTH = 25
 
 SCOPE = {"description": "Scope name",
          "type": "string",
-         "pattern": "^[a-zA-Z'_'-.0-9]{1,%s}$" % SCOPE_LENGTH}
+         "pattern": "^[a-zA-Z_\\-.0-9]{1,%s}$" % SCOPE_LENGTH}
 
 R_SCOPE = {"description": "Scope name",
            "type": "string",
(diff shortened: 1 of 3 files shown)


mlassnig (dismissed):

lib/rucio/common/schema/generic.py
```diff
@@ -49,7 +49,7 @@
 
 SCOPE = {"description": "Scope name",
          "type": "string",
-         "pattern": "^[a-zA-Z'_'-.0-9]{1,%s}$" % SCOPE_LENGTH}
+         "pattern": "^[a-zA-Z_\\-.0-9]{1,%s}$" % SCOPE_LENGTH}
```
bari12, line 52: This would add `\` as an allowed character? I don't think we want this?

illingwo:
No, it doesn't match a backslash:
```python
>>> bool(re.match("^[a-zA-Z_\\-.0-9]{1,25}$", "some\\thing"))
False
```

a hyphen is one of the few special characters inside a set, so it needs to be escaped when you want a literal backslash (or put at the end; it loses its special meaning in that case). To match an actual backslash you'd need a whole four backslashes in the regex in this case.

bari12:
I see, thanks for the explanation Robert. 👍

bari12 (approved):

Issue #1792, opened 2018-11-19
bb8 is failing with division by 0


TomasJavurek:
Motivation
----------
when calculating the ratio of secondaries and total:
https://github.com/rucio/rucio/blob/master/lib/rucio/daemons/bb8/t2_background_rebalance.py#L40
the bb8 is sometimes attempting dividing by zero.

Modification
------------
exclude such a cases.

TomasJavurek:
Traceback (most recent call last):
 File "/opt/rucio/lib/rucio/daemons/bb8/t2_background_rebalance.py", line 111, in <module>
   rebalance_rse(source_rse['rse'], max_bytes=available_target_rebalance_volume, dry_run=False, comment='T2 Background rebalancing', force_expression=destination_rse['rse'])
 File "/usr/lib/python2.7/site-packages/rucio/db/sqla/session.py", line 351, in new_funct
   result = function(*args, **kwargs)
 File "/usr/lib/python2.7/site-packages/rucio/daemons/bb8/common.py", line 339, in rebalance_rse
   for scope, name, rule_id, rse_expression, subscription_id, bytes, length in list_rebalance_rule_candidates(rse=rse, mode=mode):
 File "/usr/lib/python2.7/site-packages/rucio/db/sqla/session.py", line 360, in new_funct
   raise DatabaseException(str(error))
rucio.common.exception.DatabaseException: Database exception.
Details: (cx_Oracle.DatabaseError) ORA-01476: divisor is equal to zero
[...]

PR #1793, opened 2018-11-19
bb8: sql query ordered by file size - #1792

TomasJavurek:
bb8: sql query ordered by file size - 1792

### DIFF
--- a/lib/rucio/daemons/bb8/common.py
+++ b/lib/rucio/daemons/bb8/common.py
@@ -335,7 +335,7 @@ def rebalance_rse(rse, max_bytes=1E9, max_files=None, dry_run=False, exclude_exp
 
     print 'scope:name rule_id bytes(Gb) target_rse child_rule_id'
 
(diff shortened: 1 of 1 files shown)


bari12:
It's good, but as a side comment, please mention the ticket with the `#` eg #1792, otherwise it is not clickable here 😄

TomasJavurek:
yep, I noticed that only after commiting.

mlassnig:
Looks ok to me, but I'm a bit worried that this is with raw SQL queries, instead of SQLAlchemy. Will this be changed eventually?

bari12:
Yes should probably be changed eventually.

TomasJavurek:
Interesting is, that although there was a bug in the code, and running it with python faild, travis didn't fail.

bari12:
The BB8 test only covers a fraction of what BB8 actually does. So this was probably not covered by the test.

bari12 (approved):

Issue #1818, opened 2018-11-23
PK in history tables mostly wrong


bari12:
Motivation
----------
There is an inconsistency with the PKs in history tables which has basically two reasons:
- SQLAlchemy requires each table in models.py to have a PK (Otherwise it could not be a ORM)
- In ATLAS on Oracle, we do not have PKs on History tables to save storage

This creates an inconsistency how we (developers) expect these history tables to operate between instances which run the oracle schema (without PKs) and instances where the schema is generated from models.py (With PKs)

For example, there was instances where we make MERGE INSERTS in the history tables. This clearly works in cases where there is a PK, but it doesnt work for ATLAS, as there is no PK on the table. In the other way around, multiple inserts work on the tables for ATLAS, but do not work on tables created by the models.py, due to the uniqueness of the PK.

Modification
------------
I would propose to change all models.py History table to define the PK as `mapper_args` e.g.
```
__mapper_args__ = {
        'primary_key': [scope, name, child_scope, child_name]
    }
```
This basically tells SQLAlchmey that there is a PK, to satisfy this requirement, while in reality there is None. Secondly we would have to make an upgrade script which drops all PKs in these tables. 

Communities who query these history tables frequently, just need to create indices on these tables.

bari12:
As discussed in #1820: If there is an API/CLI to access history information, we should add an index per default. If there is no API/CLI we still might want to add one, if this information is frequently queried by operations.

PR #1820, opened 2018-11-23
Core: Removed merge commit as it does not work for history tables; Fix #1772

### LINKED ISSUE #1772: Undertaker crash when a DID is attach/detached many times
Motivation
----------
The undertaker crash when a DID is attach/detached many times. The problem is due to the archival in the history table : https://github.com/rucio/rucio/blame/master/lib/rucio/core/did.py#L714 .

bari12:
Core: Removed merge commit as it does not work for history tables; Fix #1772

### DIFF
--- a/lib/rucio/core/did.py
+++ b/lib/rucio/core/did.py
@@ -14,7 +14,7 @@
 #
 # Authors:
 # - Vincent Garonne <vgaronne@gmail.com>, 2013-2018
-# - Martin Barisits <martin.barisits@cern.ch>, 2013-2017
+# - Martin Barisits <martin.barisits@cern.ch>, 2013-2018
 # - Cedric Serfon <cedric.serfon@cern.ch>, 2013-2018
 # - Ralph Vigne <ralph.vigne@cern.ch>, 2013
 # - Mario Lassnig <mario.lassnig@cern.ch>, 2013-2018
@@ -728,7 +728,6 @@ def detach_dids(scope, name, dids, session=None):
                                                              created_at=associ_did.created_at,
                                                              updated_at=associ_did.updated_at,
                                                              deleted_at=datetime.utcnow())
-        new_detach = session.merge(new_detach)
         new_detach.save(session=session, flush=False)
 
         # Send message for AMI. To be removed in the future when they use the DETACH messages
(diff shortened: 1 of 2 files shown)


lib/rucio/db/sqla/models.py
```diff
@@ -579,8 +579,11 @@ class DataIdentifierAssociationHistory(BASE, ModelBase):
     rule_evaluation = Column(Boolean(name='CONTENTS_HIST_RULE_EVAL_CHK'))
     did_created_at = Column(DateTime)
     deleted_at = Column(DateTime)
-    _table_args = (PrimaryKeyConstraint('scope', 'name', 'child_scope', 'child_name', name='CONTENTS_HIST_PK'),
-                   CheckConstraint('DID_TYPE IS NOT NULL', name='CONTENTS_HIST_DID_TYPE_NN'),
+    __mapper_args__ = {
+        'primary_key': [scope, name, child_scope, child_name]
```
vingar, line 583: I suggest to add `deleted_at` or another date/id column in the PK, then the insertion of the same tuple (scope, name, child_scope, child_name) won't fail and will avoid the extra lookup with the advantage to keep the history of operations and favouring performance.

bari12: The point of doing it like this is, that there actually is no Primary Key at all, as we do not need/want them for history tables. See #1818 
For tables where there might be more frequent select queries, an index should be created instead of the PK.

vingar: ah. It's to micmic the pk creation. not obvious as `primary_key` is misleading but less hacky too :)  I think the lookup in the above query against the dataIdentifierAssociationHistory table can be removed -- didn't check the logic carefully but several entries for the same tuple is fine with  simple bulk inserts vs. selects + bulk inserts (imo).  The non unique index on [scope, name, child_scope, child_name] can be added there too, no ?

bari12: yes exactly, it's just to mimic to sqlalchemy that there is a PK, while in reality there isn't 😄 I checked the table create queries, like this no PK gets created, but sqlalchemy ORM still works as it thinks these columns are a PK. Unless you don't handle these objects too much (select + update) this should be completely fine.
The (non-unique) index creation we have to discuss. One argument could be that if we only rarely query these tables anyway (or even do it in hadoop) there is no point in "wasting" the space for an index (Which can be considerable for a big history table) and communities who query them often (or don't use hadoop) can create the index themselves. On the other hand we could just create them in the first place, and communities who don't need them, can drop them. Let's discuss this in #1818 😄

lib/rucio/core/did.py
```diff
@@ -864,9 +863,11 @@ def list_content_history(scope, name, session=None):
     :param session: The database session in use.
     """
     try:
+        # query = session.query(models.DataIdentifierAssociationHistory).\
+        #    with_hint(models.DataIdentifierAssociationHistory,
+        #              "INDEX(CONTENTS_HISTORY CONTENTS_HIST_PK)", 'oracle').\
+        #    filter_by(scope=scope, name=name)
         query = session.query(models.DataIdentifierAssociationHistory).\
-            with_hint(models.DataIdentifierAssociationHistory,
-                      "INDEX(CONTENTS_HISTORY CONTENTS_HIST_PK)", 'oracle').\
```
vingar, line 869: I would either keep this select together with the relevant index or drop this query without the need for the index. Like this, it goes in a full table scan for each detach operation. Keeping several entries can help to answer the question when files were removed by the production system for example.

bari12: Yes I fully agree, if there is an API (even CLI command) to access history information, we should have an index. I will add this as a comment for #1818 
The HINT here though should go anyway, as the PK doesnt exist anyway. But as soon as we have an index in the schema (With #1818) I will add it again. (For ATLAS, we actually do have a scope, name index)

PR #1823, opened 2018-11-26
Wrongly handled exception makes the poller crash : Closes #1788

### LINKED ISSUE #1788: Wrongly handled exception makes the poller crash
Motivation
----------
Traceback :
```
2018-11-15 00:13:40,539 23840   ERROR   Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/rucio/daemons/conveyor/poller.py", line 248, in poll_transfers
    for transfer_id in resps:
UnboundLocalError: local variable 'resps' referenced before assignment
```

cserf:
Wrongly handled exception makes the poller crash : Closes #1788

### DIFF
--- a/lib/rucio/daemons/conveyor/poller.py
+++ b/lib/rucio/daemons/conveyor/poller.py
@@ -232,20 +232,22 @@ def poll_transfers(external_host, xfers, prepend_str='', request_ids=None, timeo
             logging.info(prepend_str + 'Polling %i transfers against %s with timeout %s' % (len(xfers), external_host, timeout))
             resps = transfer_core.bulk_query_transfers(external_host, xfers, 'fts3', timeout)
             record_timer('daemons.conveyor.poller.bulk_query_transfers', (time.time() - tss) * 1000 / len(xfers))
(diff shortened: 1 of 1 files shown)


lib/rucio/daemons/conveyor/poller.py
```diff
@@ -234,6 +234,19 @@ def poll_transfers(external_host, xfers, prepend_str='', request_ids=None, timeo
             record_timer('daemons.conveyor.poller.bulk_query_transfers', (time.time() - tss) * 1000 / len(xfers))
         except (TransferToolTimeout, TransferToolWrongAnswer) as error:
             logging.error(prepend_str + str(error))
+            if isinstance(error, TransferToolWrongAnswer):
```
mlassnig: Shouldn't this be a separate exception handler instead of testing with `isinstance`?

I'm also not entirely convinced to have processing logic in the exception handler, can this be done differently?

bari12: Different handler would be better, makes it easier readable and the code duplication is minimal (logging.error twice)

bari12 (approved):

PR #1837, opened 2018-11-29
Core & Internals: added import/export of RSEs to API, client, CLI; Fix #716

### LINKED ISSUE #716: json based import for RSEs, protocols, distances

hahahannes:
Core & Internals: added import/export of RSEs to API, client, CLI; Fix #716

### DIFF
--- a/bin/rucio-admin
+++ b/bin/rucio-admin
@@ -57,7 +57,7 @@ from rucio.common.config import config_get
 from rucio.common.exception import (AccountNotFound, DataIdentifierAlreadyExists, AccessDenied, DataIdentifierNotFound, InvalidObject, ReplicaNotFound,
                                     RSENotFound, RSEOperationNotSupported, InvalidRSEExpression, DuplicateContent, RuleNotFound, CannotAuthenticate,
                                     Duplicate)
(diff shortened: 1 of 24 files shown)


hahahannes:
1) A few words to the workflow:
The import process creates new RSEs, limits, attributes and distances if they are not found, otherwise it overrides them or deletes/creates them new.  For example if there is no RSE with the given name, it wil be created, otherwise the properties that can be changed, will be updated.
Same for distances and protocols. Limits and attributes will be deleted if already there and will be new created.

2) For the distances, I created two new functions export_distances and list_distances. list_distances is there because there is also list_rse. The export_distances function is needed because the RSE ids can not be used for importing because if the RSE has to be created, it will get a new id. Therefor the RSE names are needed. 

3) I tried to add some schema validation for the json data, for example to check the outer structure and the RSE names but couldn't add validation for the distances matrix as the keys can change.

bari12:
Hi @hahahannes 
To me it looks very good 👍 
Can you please re-submit this as a feature though, instead of a patch. It's not a final version yet, but I think we can merge it already and use the export as a basis for our CRIC discussions. The are some open points for the import (conflict handling etc) I want to go through, but this we can do later as well.

PR #1840, opened 2018-11-29
RSE Manager: retry stat multiple times before failing Fix #1833

### LINKED ISSUE #1833: rsemgr: retries of protocol.stat
Motivation
----------
in some cases, the stat failed due to service unavailable, although the file was phsysically there.

Modification
------------
at line: https://github.com/rucio/rucio/blob/master/lib/rucio/rse/rsemanager.py#L468
try, fail, wait 1 second, try, fail, wait 2 seconds, try, fail wait 4 seconds, try fail, wait 16 seconds, try fail, wait 32 seconds, try, fail finally

berghaus:
RSE Manager: retry stat multiple times before failing Fix #1833


lib/rucio/rse/rsemanager.py
```diff
@@ -703,6 +703,26 @@ def find_matching_scheme(rse_settings_dest, rse_settings_src, operation_src, ope
...
 
 
+def _retry_protocol_stat(protocol, pfn):
+    """
+    try to stat file, on fail try again 1s, 2s, 4s, 8s, 16s, 32s later. Fail is all ail
+
+    :param protocol     The protocol to use to reach this file
+    :param pfn          Physical file name of the target for the protocol stat
+    """
+    from time import sleep
```
mlassnig, line 726: imports always go to the top of the module

bari12: Pinging @berghaus 😄

lib/rucio/rse/rsemanager.py
```diff
@@ -703,6 +703,26 @@ def find_matching_scheme(rse_settings_dest, rse_settings_src, operation_src, ope
...
+    for attempt in range(6):
```
bari12: In the baseclient there is a variable called `REQUEST_RETRIES`, I wonder if we should use this here as well.

berghaus: Looking through this it is specified in the Rucio configuration and used in the loop in line [298](https://github.com/rucio/rucio/blob/6d17a9dac8fb2844c10342dd74d39e9ff0f6161b/lib/rucio/client/baseclient.py#L298) of the baseclient. It sets a number of times to try opening a connection.

Using it here would overload the meaning: here the connection can be opened but it may take the storage element some time to recognise the data to be there. We could use this variable here, but should we be worried about introducing some interplay between what it does to connection retries and what it does in the RSE manager?

berghaus:
The commit has been rebased on the current next branch. Let me know what you think about my previous question.

berghaus:
Added a new configuration variable `protocol_stat_retries` set to 6 by default and defaulting to 6 if not picked up from the configuration. The forced push was necessary because of another rebase.

berghaus:
Replaced this feature by a patch since we agreed that by now this has become too urgent to wait until June.

PR #1854, opened 2018-11-30
‘expired’ and ‘obsolete’ counters are not correctly updated : Closes #1849

### LINKED ISSUE #1849: ‘expired’ and ‘obsolete’ counters are not correctly updated
This affects the `check_deletable_replicas` and `check_obsolete_replicas` probes. The SQL queries do not deal with RSEs where everything has already been deleted by the Reaper.

cserf:
‘expired’ and ‘obsolete’ counters are not correctly updated : Closes #1849

### DIFF
--- a/tools/probes/common/check_deletable_replicas
+++ b/tools/probes/common/check_deletable_replicas
@@ -7,11 +7,13 @@
 #
 # Authors:
 # - Mario Lassnig, <mario.lassnig@cern.ch>, 2013-2014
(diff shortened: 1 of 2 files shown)


mlassnig (approved):
will we need these queries for the other databases though?

bari12:
The RSE Usage calculation should maybe be it's own daemon (Similar to the abbacus stuff) 
All the other probes are more optional, but this one is really more fundamental.

PR #1858, opened 2018-12-03
db: boolean handling; Fix #1852

### LINKED ISSUE #1852: Verify boolean handling from clients to all databases
Motivation
----------

Modification
------------

mlassnig:
db: boolean handling; Fix #1852

### DIFF
--- a/lib/rucio/db/sqla/types.py
+++ b/lib/rucio/db/sqla/types.py
@@ -84,8 +84,13 @@ class BooleanString(TypeDecorator):
 
         if isinstance(value, bool):
             if value:
-                return '1'
-            return '0'
+                return 'true'
+            return 'false'
+        elif isinstance(value, str):
+            if value.lower() == 'true':
+                return 'true'
+            elif value.lower() == 'false':
+                return 'false'
 
         return str(value)
 
@@ -93,6 +98,7 @@ class BooleanString(TypeDecorator):
         if value is None:
             return value
 
+        # be backward compatible for now
         if value.lower() in ['1', 'true']:
             return True
         elif value.lower() in ['0', 'false']:
--- a/lib/rucio/db/sqla/types.py
+++ b/lib/rucio/db/sqla/types.py
@@ -6,7 +6,7 @@
 #
 # Authors:
 # - Vincent Garonne, <vincent.garonne@cern.ch>, 2013
-# - Mario Lassnig, <mario.lassnig@cern.ch>, 2013
+# - Mario Lassnig, <mario.lassnig@cern.ch>, 2013-2018
 #
 # PY3K COMPATIBLE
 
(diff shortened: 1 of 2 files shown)


lib/rucio/db/sqla/types.py
```diff
@@ -93,9 +93,9 @@ def process_result_value(self, value, dialect):
         if value is None:
             return value
 
-        if value == '1':
+        if value.lower() in ['1', 'true']:
             return True
-        elif value == '0':
+        elif value.lower() in ['0', 'false']:
             return False
```
vingar, line 105: Sometimes 0 /1 are simple numerical values like tier 1. What about dropping 0 / 1 to use only true/false for booleans  ? I t can help to have a clear type specification from the clients.  For the db non supporting the boolean type, one additional field/column/etc telling if the value is a boolean or a numerical value can be added to help the json serialisation imo.

mlassnig: I messed up the commit somehow, that's what I wanted to do. FOr now we have to keep '0','1' as backward compatible, but store only 'true'/'false' as string. For the type, we know that already, since they are set by the SQLA model.

vingar: yes, introducing a proper boolean support and maintaining the backward compatibility looks like a big mess there. Can you explain me what will be the output for the rse attribute stored with name=tier and value=1 in the db?

mlassnig: Like before: Python Boolean True. IMHO this is wrong conceptually, but that's what we have right now.

We need to gradually migrate installations over to the new 'true'/'false' behaviour, once this code is in place.

bari12: I think this should be a feature, looks like a potential impactful change.
I wonder if it is not worth to introduce an additional column `value_type` right away and not trick around with some intermediate states?

mlassnig: `value_type` won't really help, will it? unless we expect users to always define the type when setting the key/value pair, which is a bit weird.

but i agree on making this a feature. let me cancel this one then so it doesn't get too confusing.

bari12: If it comes from the API, we can extract the type. From the CLI we probably would need a `--boolean` parameter.

mlassnig: Isn't this how it should be:

- strings `true`/`false` are interpreted as boolean
- any number is a number (0 and 1 not interpreted as boolean, that was the mistake we made)
- anything else is string

No need for a parameter that way, and IMHO much more intuitive.

vingar: using true/false like in json should be good enough imo. With pseudo commands: `set attr attrname true`, `set attr attrname false`, `set attr name` can get `true` by default.

mlassnig: ok, so we agree :-D

bari12: Yes that sounds good, I have to check the RSE-Expression though as well, this might impact them. (e.g. CERN-PROD_DATADISK matches against a boolean attribute, need to check if this is done correctly)

mlassnig: RSE Expression continue to work as before only on MySQL. I'll address this in my patch.

Issue #1861, opened 2018-12-04
bb8 - replacing sql query with sql alchemy


TomasJavurek:
Motivation
----------
In order to be generic in terms of DB, sql query in bb8 needs to be replaced by sql alchemy query.

Modification
------------
rewrite the query.

PR #1862, opened 2018-12-04
bb8: sqlalchemy query, #1861

TomasJavurek:
bb8: sqlalchemy query, #1861

### DIFF
--- a/lib/rucio/daemons/bb8/common.py
+++ b/lib/rucio/daemons/bb8/common.py
@@ -23,17 +23,22 @@ import sys
 
 from datetime import datetime, date, timedelta
 from string import Template
+from sqlalchemy.orm import aliased
+from sqlalchemy import func, and_, or_, cast, Integer
+from sqlalchemy.sql.expression import case, select
 
 from rucio.core.lock import get_dataset_locks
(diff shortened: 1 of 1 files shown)


lib/rucio/daemons/bb8/common.py
```diff
@@ -220,36 +225,52 @@ def list_rebalance_rule_candidates(rse, mode=None, session=None):
...
+    rse_id = get_rse_id(rse)
+
+    # dumps can be applied only for decommission since the dumps doesn't contain info from dids
+    if mode == 'decommission':
         return _list_rebalance_rule_candidates_dump(rse, mode)
 
+    # the rest is done with sql query
+    from_date = datetime.utcnow() + timedelta(days=60)
+    to_date = datetime.now() - timedelta(days=60)
+    allowed_accounts = ['panda', 'root', 'ddmadmin']
```
bari12, line 237: That sqlalchemy query is a really nice job 😄 
In the long-term, there is a few things in there which we should maybe make configurable, so this can be used by other experiments, e.g. allowed_accounts, but for now it looks really good like this.

bari12 (approved):

Issue #1867, opened 2018-12-05
Incorrect calculation of expected free size in rule-approval e-mail


dchristidis:
Motivation
----------

The calculation does not take into account the replicas that are already available on the destination.

Modification
------------

Use `rucio.core.replica.list_replicas()`. The operation can be expensive, but since it’s not done frequently, it should be fine.

dchristidis:
### Example
On an RSE there are 100 TB free, dataset `A` is 10 TB and is already available on the RSE, dataset `B` is 5 TB and not available on the RSE, container `C` contains `A` and `B`. The e-mail sent when a user requests to replicate `C` should report that the expected free space after the approval is 95 TB. Currently, the e-mail will report 85 TB.

bari12:
Thank you for raising this. We have reviewed this issue as part of our regular issue review.

We see value in this request, but we do not have the capacity to work on it within the next twelve months.

In line with our [issue management guideline](https://rucio.cern/documentation/developer/issue_management), we are labelling it `wishlist` and closing it. This does not mean it is forgotten: wishlist items are revisited regularly during component reviews and roadmap planning, and this issue can be re-opened and added back to the pipeline as soon as interest and capacity align.

PR #1868, opened 2018-12-05
Rules: Consider available files in free-space calculation #1867

dchristidis:


lib/rucio/core/rule.py
```diff
@@ -2892,13 +2893,16 @@ def __create_rule_approval_email(rule, session=None):
...
 
         try:
-            for usage in get_rse_usage(rse=target_rse, session=session):
-                if usage['source'] == 'storage':
-                    free_space = sizefmt(usage['free'])
-                    if did['bytes'] is None:
-                        free_space_after = 'undefined'
-                    else:
-                        free_space_after = sizefmt(usage['free'] - did['bytes'])
+            replicas = rucio.core.replica.list_replicas(dids=[{'scope': rule.scope,
```
bari12, line 2896: This is really a quite heavy query, since you are reading/streaming all rows from the DB to the server, but you are actually only interested in an aggregate. For big datasets/containers, this is an issue.
What you can do is if `did['bytes'] is None`, query with `get_did(..., dynamic=True)`
https://github.com/rucio/rucio/blob/ba88eaf0f1de3c8cd6c44546568c6aafdfa784b8/lib/rucio/core/did.py#L1094

bari12 (changes requested):

lib/rucio/core/rule.py
dchristidis: I am not sure I understand your suggestion to use `rucio.core.did.get_did()`.

In the Rucio UI, `rucio.core.replica.list_dataset_replicas()` is used. It’s unfortunate that `CollectionReplica` doesn’t have information on containers. The only alternative I see is to use that and omit (i.e. set to undefined) the expected free space after the approval of the replication rule when the DID is a container.

bari12: What `get_did` does it also exposes the `bytes` value of the DID. If the value is `None` though, it would reply None unless you use the `dynamic=True` option, which calculates the bytes on demand, however in a much more efficient way than listing the replicas explicitly.

dchristidis: I’m sorry, I might have not described the motivation enough. What is needed in this particular case is to calculate the portion of a DID that is already available on the destination RSE. I will elaborate in the issue.

bari12: Ah, I see, should have read the code more carefully 😄 
I understand now, yes like this it doesn't work like that, but list_replicas is still not the right approach for this, the aggregation should be done on the database side. Listing 30k+ replicas will take way too long and slow a simple rule insertion (for approval) down a lot. So maybe a new method to do this would be appropriate.

dchristidis:
I am closing this while I work on a different approach.

PR #1875, opened 2018-12-05
Core & Internals: added import/export of RSEs to API, client, CLI; Fix #716

### LINKED ISSUE #716: json based import for RSEs, protocols, distances

hahahannes:
1. A few words to the workflow:
The import process creates new RSEs, limits, attributes and distances if they are not found, otherwise it overrides them or deletes/creates them new. For example if there is no RSE with the given name, it wil be created, otherwise the properties that can be changed, will be updated.
Same for distances and protocols. Limits and attributes will be deleted if already there and will be new created.

2. For the distances, I created two new functions export_distances and list_distances. list_distances is there because there is also list_rse. The export_distances function is needed because the RSE ids can not be used for importing because if the RSE has to be created, it will get a new id. Therefor the RSE names are needed.

3. I tried to add some schema validation for the json data, for example to check the outer structure and the RSE names but couldn't add validation for the distances matrix as the keys can change.

### DIFF
--- a/bin/rucio-admin
+++ b/bin/rucio-admin
@@ -57,7 +57,7 @@ from rucio.common.config import config_get
 from rucio.common.exception import (AccountNotFound, DataIdentifierAlreadyExists, AccessDenied, DataIdentifierNotFound, InvalidObject, ReplicaNotFound,
                                     RSENotFound, RSEOperationNotSupported, InvalidRSEExpression, DuplicateContent, RuleNotFound, CannotAuthenticate,
                                     Duplicate)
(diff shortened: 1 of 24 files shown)


bari12:
As I said in the other PR, for me this is good and can be used as the basis for the discussion of the json-scheme with CRIC. We can adapt the code from there then. I'll leave this open until tomorrow, if somebody else has some comments and then merge.

mlassnig:
Looks good to me as well.

bari12:
👍 I will merge it like this and we iterate from there.

Issue #403, opened 2017-12-19
Automatic recovery of suspicious files that have more than one replica


cserf:
Motivation
----------
If often happens that suspicious files have more than one replica. In that case an automatic recovery can be done without the need of a human confirmation.

Modification
------------
Automatic recovery of suspicious files that have more than one replica

bari12:
Don't we have this behaviour already for files marked as BAD? So wouldn't it be enough to mark a suspicious file as bad?

mlassnig:
@gumond has volunteered for this task

bari12:
This was implemented as a probe in #1871, but it should probably be it's own daemon.

gumond:
I have not had a look at this task yet. I considered it closed. When I come back tomorrow, I will see with you & Cedric how rucio daemons are implemented and what could I do about creating this one.

gumond:
bin/rucio-replica-recoverer:
- Contains, in an epilog variable, my current procedure for manual testing.  
- Once I learn more about rucio, should I write a test for this daemon ? 

lib/rucio/core/replica.py: 
- Although there are some differences, the two functions 'get_suspicious_files' and the new 'get_available_suspicious_replicas' seem to be aiming for similar outputs. Could we merge the two into one ?

PR #1876, opened 2018-12-05
ATLAS : Probe to automatically recover suspicious files that have multiple replicas : Closes #1871

### LINKED ISSUE #1871: Probe to automatically recover suspicious files that have multiple replicas
Motivation
----------
For the suspicious files that have multiple replicas, there is no need of human check to confirm that the replica is actually bad. One can have a probe to automatically recover them.

cserf:
ATLAS : Probe to automatically recover suspicious files that have multiple replicas : Closes #1871


bari12 (approved):
I think this is 👍 
However, we should still keep #403 open and transform this into a daemon at some point.

Issue #425, opened 2018-01-08
Several small BB8 optimisiations


TomasJavurek:
- automatic mode skips sites, where is some active rebalancing rule in STUCK and REPLICATING status. For the moment, only REPLICATING shall be used and later, it shall count the number of bytes that are currently replicating.

- test overlap between t2s and Nuclei, shall the rebalancing in these two groups completely seprate?

- development of softer conditions for rules that can be rebalanced.

- reduce background rebalancing, optimal set of params.

- very low efficiency of t2 bg rebalancing:
 https://monit-grafana.cern.ch/d/FtSFfwdmk/ddm-transfers?orgId=17&from=1543497236926&to=1544102036926&var-activity=Data%20Rebalancing&var-activity=Data%20rebalancing&var-src_cloud=All&var-src_site=All&var-src_tier=2&var-src_endpoint=All&var-dst_cloud=All&var-dst_site=All&var-dst_tier=2&var-dst_endpoint=All&var-binning=%24__auto_interval_binning&var-groupby=activity&var-columns=src_cloud&var-rows=dst_cloud&var-measurement=ddm_transfer&var-retention_policy=raw

PR #1886, opened 2018-12-06
bb8: reduc bg activity - #425

TomasJavurek:
bb8: reduc bg activity - #425


cserf:
Would be better to be able to set it with `rucio config`. Just add a section for BB8

bari12:
In the end this `nuclei_background_rebalance.py` module should not even be part of the core package. I put it there back then as it was just easier to deploy, but we should see to get it out of there, as this part of the code is really only useful for ATLAS (nuclei etc.) And the module is not really a daemon, more of a user-application on top of BB8, so it's more similar to a probe.
I will merge this now, but we should try to get this into the upcoming probes repository maybe?

bari12 (approved):

PR #1893, opened 2018-12-07
db: boolean handling; Fix #1852

### LINKED ISSUE #1852: Verify boolean handling from clients to all databases

mlassnig:
db: boolean handling; Fix #1852


bari12:
I know it's backward compatible, but I think this should be a feature too and we should do this migration over two feature versions essentially. Can you maybe detail in the ticket what the steps are for the migration? (We might want to have these as alembic upgrade queries actually?)

PR #1899, opened 2018-12-10
minor doc fixes

mlassnig:
minor doc fixes

### DIFF
--- a/doc/source/installing_daemons.rst
+++ b/doc/source/installing_daemons.rst
@@ -4,20 +4,142 @@ Installing Rucio daemons
 Prerequisites
 ~~~~~~~~~~~~~
 
-Rucio server runs on Python 2.6, 2.7.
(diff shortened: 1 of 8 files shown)


doc/source/installing_daemons.rst
```diff
@@ -4,20 +4,142 @@ Installing Rucio daemons
 Prerequisites
 ~~~~~~~~~~~~~
 
-Rucio server runs on Python 2.6, 2.7.
+The Rucio daemons run on Python 2.6 and 2.7 on any Unix-like platform.
```
bari12: 2.7 only. I mean they might run on 2.6, but we don't test against that anymore.

mlassnig: Servers and clients as well?

tbeerman (approved):

PR #1907, opened 2018-12-11
bug: Fix CMS RSE Name pattern for ATTACHMENTS Fix #1901

### LINKED ISSUE #1901: Wrong cms RSE name pattern used when attaching dids
Motivation
----------
in lib/rucio/common/schema/cms.py the RSE Name pattern is 

`^T[0-3]_[A-Z]{2}((_[A-Za-z0-9]+)+)$`

while the pattern for the rse property of ATTACHMENT is

`^([A-Z0-9]+([_-][A-Z0-9]+)*)$`

as a result when we add replicas to a CMS RSE whose name is not all in capitals (e.g. T1_UK_RAL_Buffer):
  * if we use `add_replicas` and the `attach_dids` it works fine
  * if we use `attach_dids` directly it fails with `InvalidObject`

Modification
------------
Just put the same regexp for the two patterns (PR coming)

sartiran:
bug: Fix CMS RSE Name pattern for ATTACHMENTS Fix #1901

### DIFF
--- a/lib/rucio/common/schema/cms.py
+++ b/lib/rucio/common/schema/cms.py
@@ -269,7 +269,7 @@ ATTACHMENT = {"description": "Attachement",
                              "name": NAME,
                              "rse": {"description": "RSE name",
                                      "type": ["string", "null"],
-                                     "pattern": "^([A-Z0-9]+([_-][A-Z0-9]+)*)$"},
+                                     "pattern": "^T[0-3]_[A-Z]{2}((_[A-Za-z0-9]+)+)$"},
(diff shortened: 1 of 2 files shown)


bbockelm (dismissed):
Looks correct to me.

bari12:
Andrea still wanted to test one thing so he asked me to wait with the merge. So waiting for the 👍 from @sartiran 😄

sartiran:
Hi Martin, with Eric we tested that this actually fixes the problem. I think you can merge.

bari12 (approved):

PR #1915, opened 2018-12-13
Dataset deletion: Undertaker should expire large rules and let judge-cleaner take over; Fix #1909

### LINKED ISSUE #1909: Undertaker handling of very larger rules
Motivation
----------
For very large rules, the undertaker is blocked a long time to delete them which leads to long backlogs.

Modification
------------
Instead, for rules > 10k files it could just flag the rule for expiration and set the timeout of the dataset 1 day in the future. Doing this, the judge-cleaner will take care of the rule deletion and thus free up the undertaker to do its job.

bari12:
Dataset deletion: Undertaker should expire large rules and let judge-cleaner take over; Fix #1909

### DIFF
--- a/lib/rucio/core/did.py
+++ b/lib/rucio/core/did.py
@@ -544,13 +544,16 @@ def attach_dids_to_dids(attachments, account, ignore_duplicate=False, session=No
 
 
 @transactional_session
-def delete_dids(dids, account, session=None):
+def delete_dids(dids, account, expire_rules=False, session=None):
     """
     Delete data identifiers
 
-    :param dids: The list of dids to delete.
-    :param account: The account.
-    :param session: The database session in use.
+    :param dids:          The list of dids to delete.
+    :param account:       The account.
+    :param expire_rules:  Expire large rules instead of deleting them right away. This should only be used in Undertaker mode, as it can be that
+                          the method returns normally, but a did was not deleted; This trusts in the fact that the undertaker will retry an
+                          expired did.
+    :param session:       The database session in use.
     """
     rule_id_clause, content_clause = [], []
     parent_content_clause, did_clause = [], []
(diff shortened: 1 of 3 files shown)


lib/rucio/core/did.py
```diff
@@ -613,19 +616,34 @@ def delete_dids(dids, account, session=None):
...
+                                                                                                                              models.ReplicationRule.locks_stuck_cnt).filter(or_(*rule_id_clause)):
                 logging.debug('Removing rule %s for did %s:%s on RSE-Expression %s' % (str(rule_id), scope, name, rse_expression))
+
                 # Propagate purge_replicas from did to rules
                 if (scope, name) in not_purge_replicas:
                     purge_replicas = False
                 else:
                     purge_replicas = True
-                rucio.core.rule.delete_rule(rule_id=rule_id, purge_replicas=purge_replicas, delete_parent=True, nowait=True, session=session)
+                if expire_rules and locks_ok_cnt + locks_replicating_cnt + locks_stuck_cnt > 10000:
```
cserf, line 636: Make it configurable in a section undertaker in the config table ?

lib/rucio/core/did.py
```diff
@@ -613,19 +616,34 @@ def delete_dids(dids, account, session=None):
...
+                    # Expire the rule (soft=True)
+                    rucio.core.rule.delete_rule(rule_id=rule_id, purge_replicas=purge_replicas, soft=True, delete_parent=True, nowait=True, session=session)
+                    # Update expiration of did
+                    set_metadata(scope=scope, name=name, key='lifetime', value=3600 * 24, session=session)
+                    skip_deletion = True
+                else:
+                    rucio.core.rule.delete_rule(rule_id=rule_id, purge_replicas=purge_replicas, delete_parent=True, nowait=True, session=session)
```
vingar, line 643: Performance wise, bulk methods will help. Do they exist, e.g.,  `delete_rules`, `bulk_set_metadata` ?
If not, then please consider storing the concerned rules in a list instead as a first step to introduce the logic.  The bulk update itself can be done with raw query instead or even better introducing the corresponding bulk methods starting with simple sequential implementation will help converging.

vingar (commented):
non blocking.

lib/rucio/core/did.py
bari12: No, there are no bulk methods for both of these. For `delete_rule` we should consider it, not necessarily for this workflow, but also for the judge-cleaner it would be beneficial to have a bulk method. But due to it's complexity this is on a longer timescale. 
For `set_metadata` we don't really need it (at least not for this workflow). This branch of the if statement is called so rarely that it is just extremely unlikely that there will be a second did in the same chunk where the condition applies. There are just not that many rules >10k, so it's not really worth it.

lib/rucio/core/did.py
bari12: I am merging this now, as it should be in todays release. The configuration will come with #1932

PR #1921, opened 2018-12-17
db: boolean handling

mlassnig:
db: boolean handling

### DIFF
--- a/lib/rucio/core/rse.py
+++ b/lib/rucio/core/rse.py
@@ -29,8 +29,6 @@
 
 from __future__ import division
 
-from re import match
-from six import string_types
 try:
     from StringIO import StringIO
 except ImportError:
@@ -39,6 +37,9 @@ except ImportError:
 import json
 import sqlalchemy
 import sqlalchemy.orm
+
+from re import match
+from six import string_types
 
 from dogpile.cache import make_region
 from dogpile.cache.api import NO_VALUE
@@ -51,7 +52,6 @@ from sqlalchemy.sql.expression import or_, false
 import rucio.core.account_counter
 
 from rucio.core.rse_counter import add_counter, get_counter
-
 from rucio.common import exception, utils
 from rucio.common.config import get_lfn2pfn_algorithm_default
 from rucio.db.sqla import models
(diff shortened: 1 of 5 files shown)


bari12:
I am not quite sure how the `tmp_atlas_` variables help? I still see a problem with this when we are in the INT / PROD inconsistency we will have for a week?
Also I wonder if we should add an upgrade script where we change all the True/False to true/false?

mlassnig:
We need a third "truth" state for the RSEs, which is represented by the tmp_atlas variables. That's why they are twice handled: once in the core, and once on the DB type.

Script: Alembic doesn't support content updates, would have to go through the SQLAlchemy interface. There's no way to downgrade though anymore, because that would update all "true" to "True", when they would only partially be "True" before.

Maybe it's better to just provide a SQL statement with the upgrade instructions?

bari12:
We should still add an alembic update file here (if possible) which changes `False/True` to `false/true`. We should also add a note to the release which clarifies that entries with `0/1` which are used as boolean (Thus not `tier` etc.), should also be changed to `false/true`

vingar:
I'm wondering if won't help to introduce some extra column with the right type directly or a column describing the type and the constraint. e.g; json schema.  Something like : atr, value , schema or description. 

The schema field can thus be used to convert the value to the right type, without touching the current value and be backward compatible.

bari12:
Pylint throws some errors here

```
E: 40, 4: Module 'alembic.op' has no 'execute' member (no-member)
E: 41, 4: Module 'alembic.op' has no 'execute' member (no-member)
E: 42, 4: Module 'alembic.op' has no 'execute' member (no-member)
E: 43, 4: Module 'alembic.op' has no 'execute' member (no-member)
E: 62, 4: Module 'alembic.op' has no 'execute' member (no-member)
E: 63, 4: Module 'alembic.op' has no 'execute' member (no-member)
E: 64, 4: Module 'alembic.op' has no 'execute' member (no-member)
E: 65, 4: Module 'alembic.op' has no 'execute' member (no-member)
```

mlassnig:
hmmm, seems like pylint does not properly resolve the alembic imports?

bari12:
Yes, this happens sometimes. `# pylint: disable=no-member` is your friend 😄

bari12:
Please rebase to make tests work.

mlassnig:
@bari12 we basically need to have this running in production before we can merge https://github.com/rucio/rucio/pull/2031 , otherwise we will have to make another workaround to deal with 0/1 in list_rses

bari12 (approved):

Issue #1845, opened 2018-11-29
import of builtins not Python 2.6 compatible in downloadclient


bari12:
Motivation
----------
https://github.com/rucio/rucio/blob/06e4f7e3ac03111f1ec94add1be3524d3d0fc4d9/lib/rucio/client/downloadclient.py#L35 is not compatible to Python 2.6 it seems, which is a problem for the deployment on the grid

Py2.6:
```
[desilva@atlas-sl6x64 ~]$ rucio --version
Traceback (most recent call last):
  File "/ALRB/atlasadmin/atlas-sl6x64.triumf.ca/ATLASLocalRootBase/x86_64/rucio-clients/1.18.6/bin/rucio", line 63, in <module>
    from rucio.client.downloadclient import DownloadClient
  File "/ALRB/atlasadmin/atlas-sl6x64.triumf.ca/ATLASLocalRootBase/x86_64/rucio-clients/1.18.6/lib/python2.7/site-packages/rucio/client/downloadclient.py", line 35, in <module>
    from builtins import round
ImportError: No module named builtins
```

bari12:
Not sure why this was not caught by the Python 2.6 test though?

hahahannes:
I will remove the imports and add a basic test to the python2.6 client test to check at least imports. Also the pylint check for python3 has to be changed to ignore round errors.

PR #1925, opened 2018-12-19
Core & Internals: added python3 compatibilty to core; Fix #1912

### LINKED ISSUE #1912: Core migration to python3
Motivation
----------
Make the core module of Rucio python3 compatible.

Modification
------------

hahahannes:
Core & Internals: added python3 compatibilty to core; Fix #1912

### DIFF
--- a/lib/rucio/core/authentication.py
+++ b/lib/rucio/core/authentication.py
@@ -20,6 +20,9 @@
 # - Ralph Vigne <ralph.vigne@cern.ch>, 2013-2014
 # - Cedric Serfon <cedric.serfon@cern.ch>, 2013
 # - Thomas Beermann <thomas.beermann@cern.ch>, 2017
+# - Hannes Hansen <hannes.jakob.hansen@cern.ch>, 2018
+#
+# PY3K COMPATIBLE
 
 import base64
 import datetime
@@ -282,7 +285,7 @@ def get_ssh_challenge_token(account, appid, ip=None, session=None):
     # Cryptographically secure random number.
     # This requires a /dev/urandom like device from the OS
     rng = random.SystemRandom()
-    crypto_rand = rng.randint(0, sys.maxint)
+    crypto_rand = rng.randint(0, sys.maxsize)
 
     # give the client 10 seconds max to sign the challenge token
     expiration = datetime.datetime.utcnow() + datetime.timedelta(seconds=10)
--- a/lib/rucio/core/credential.py
+++ b/lib/rucio/core/credential.py
@@ -14,12 +14,22 @@
 #
 # Authors:
 # - Mario Lassnig <mario@lassnig.net>, 2018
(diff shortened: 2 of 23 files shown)


lib/rucio/core/authentication.py
```diff
@@ -282,7 +285,7 @@ def get_ssh_challenge_token(account, appid, ip=None, session=None):
     # Cryptographically secure random number.
     # This requires a /dev/urandom like device from the OS
     rng = random.SystemRandom()
-    crypto_rand = rng.randint(0, sys.maxint)
+    crypto_rand = rng.randint(0, sys.maxsize)
```
hahahannes, line 288: Not sure about this. sys.maxint is not available in python3. Instead sys.maxsize is still available but not sure if it can be also used here ([https://docs.python.org/2/library/sys.html#sys.maxint](https://docs.python.org/2/library/sys.html#sys.maxint)).

lib/rucio/core/credential.py
```diff
@@ -59,9 +69,12 @@ def get_signed_url(service, operation, url, lifetime=600):
         raise UnsupportedOperation('URL must not be empty')
 
     if lifetime:
-        if not isinstance(lifetime, (int, long)):
+        if not isinstance(lifetime, integer_types):
             try:
-                lifetime = long(lifetime)
+                if PY2:
+                    lifetime = long(lifetime)
```
hahahannes, line 75: I guess the `long` type is needed here? Therefor I created an exception in the python3 check to exclude the long check, because otherwise the test will fail. But then it is possible that someone will use `long` somewhere else without the python version check.
There is the possbility to use `form builtins import int` to have a solution for py2 and py3 but its causing problems with python2.6 (see #1845 )

lib/rucio/core/authentication.py
hahahannes: Ok, the six package seems to have a constant for this

bari12: I think this is fine.

lib/rucio/core/credential.py
bari12: Python 2.6 is not an issue here, since this is only used in the server. I think it is fine like that, but @mlassnig should comment.

bari12:
To me everything looks good, I would prefer though if this goes as a feature to 1.19.0. So please resubmit as a feature. 😄

lib/rucio/core/authentication.py
mlassnig: Yes, this is ok.

lib/rucio/core/credential.py
mlassnig: I don't think that setting lifetimes longer than a couple million years will be problematic ;-)

hahahannes:
Close for #1936

PR #1938, opened 2019-01-08
Core & Internals: added configuration for amount of locks for rules to be expired by undertaker; Fix #1932

### LINKED ISSUE #1932: Make undertaker expire_rules size configurable
Motivation
----------
Currently the undertaker expires rules bigger than 10k locks, instead of deleting them immediately. This value should be configurable via the config table.

hahahannes:
Core & Internals: added configuration for amount of locks for rules to be expired by undertaker; Fix #1932

### DIFF
--- a/lib/rucio/core/did.py
+++ b/lib/rucio/core/did.py
@@ -45,7 +45,7 @@ import rucio.core.replica  # import add_replicas
 from rucio.common import exception
 from rucio.common.config import config_get
 from rucio.common.utils import str_to_date, is_archive, chunks
-from rucio.core import account_counter, rse_counter
+from rucio.core import account_counter, rse_counter, config as config_core
 from rucio.core.message import add_message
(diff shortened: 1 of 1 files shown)


cserf:
That would be better to have it configurable via the config table

etc/rucio.cfg.template
```diff
@@ -205,3 +206,6 @@ usercert = /opt/rucio/etc/usercert_with_key.pem
 [credentials]
 gcs = /opt/rucio/etc/google-cloud-storage-test.json
 signature_lifetime = 3600
+
+[undertaker]
```
mlassnig: can this go into the core/config module instead of the rucio.cfg?

hahahannes: yes, already done but will submit the new commit after PR #1950 got merged

bari12 (approved):

PR #1939, opened 2019-01-08
Core & Internals: added configuration for amount of locks for rules to be expired by undertaker; Fix #1932

### LINKED ISSUE #1932: Make undertaker expire_rules size configurable

hahahannes:
Core & Internals: added configuration for amount of locks for rules to be expired by undertaker; Fix #1932

### DIFF
--- a/lib/rucio/api/config.py
+++ b/lib/rucio/api/config.py
@@ -7,6 +7,7 @@
 #
 # Authors:
 # - Mario Lassnig, <mario.lassnig@cern.ch>, 2014
+# - Hannes Hansen, <hannes.jakob.hansen@cern.ch>, 2019
 #
 # PY3K COMPATIBLE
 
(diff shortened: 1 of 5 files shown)


etc/docker/dev/rucio.cfg
```diff
@@ -173,3 +173,6 @@ support_rucio = https://github.com/rucio/rucio/issues/
 
 [credentials]
 gcs = /opt/rucio/etc/google-cloud-storage-test.json
+
+[undertaker]
+expire_rules_locks_size = 10000
```
vingar: I would suggest to add this configuration parameter in the configuration table instead.

bari12: Yes, that was the original plan of the issue; I think I didn't describe this clearly and @hahahannes interpreted it wrong. In https://github.com/rucio/rucio/blob/master/lib/rucio/core/transfer_limits.py you can see how this is done with the config table (config_core)
On a side-note. I am wondering if we should introduce some caching to the config_core module, as this has to be done in the other modules otherwise.

hahahannes: alright, I am pushing the new code after the PR #1950 got merged

bari12 (approved):

PR #1950, opened 2019-01-09
Core & Internals: added default value to config core; Fix #1948

### LINKED ISSUE #1948: add default value to config table
Motivation
----------
It would be nice to define a default value when reading a config from the config table.
[https://github.com/rucio/rucio/blob/master/lib/rucio/core/config.py#L101](https://github.com/rucio/rucio/blob/master/lib/rucio/core/config.py#L101
)

Modification
------------
```python
def get(section, option, session=None)
```
to
```python
def get(section, option, default=None, session=None)
```

Also REST and API have to be updated.

hahahannes:
Core & Internals: added default value to config core; Fix #1948

### DIFF
--- a/lib/rucio/core/config.py
+++ b/lib/rucio/core/config.py
@@ -9,11 +9,11 @@
   Authors:
   - Mario Lassnig, <mario.lassnig@cern.ch>, 2014
   - Cedric Serfon, <cedric.serfon@cern.ch>, 2017
+  - Hannes Hansen, <hannes.jakob.hansen@cern.ch>, 2019
 
   PY3K COMPATIBLE
 '''
 
-from rucio.common.exception import ConfigNotFound
 from rucio.db.sqla import models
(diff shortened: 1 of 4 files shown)


lib/rucio/core/config.py
```diff
@@ -116,7 +117,7 @@ def get(section, option, session=None):
     if tmp is not None:
         return __convert_type(tmp[0])
     else:
-        raise ConfigNotFound()
+        return default
```
bari12, line 123: I think this should be handled differently. If `default=None` and nothing can be found the exception should still be raised. Only otherwise we raise the default.
This should be fine, since `None` is not an eligible default value anyway (since the value comes from the db)

bari12 (changes requested):

lib/rucio/core/config.py
hahahannes: ah yes make sense

bari12 (approved):

Issue #1512, opened 2018-09-05
Erroneous rucio download summary info


cohm:
Motivation
----------

I tried to download one file from a dataset, and the output from rucio doesn't make sense:

```
lxplus040:/tmp/cohm > rucio get --nrandom 1 data17_5TeV:data17_5TeV.00340634.calibration_PixelBeam.merge.DAOD_IDPIXLUMI.c1166_m1905
----------------------------------
Download summary
----------------------------------------
DID data17_5TeV:data17_5TeV.00340634.calibration_PixelBeam.merge.DAOD_IDPIXLUMI.c1166_m1905
Total files :                                 0
Downloaded files :                            0
Files already found locally :                 0
Files that cannot be downloaded :             0
```

Nothing was downloaded, so it should somehow say that there was a problem somewhere in the output. Could this be improved?

Thanks a lot!

Best,
Christian

Modification
------------

TWAtGH:
This dataset is only available at `CERN-PROD_DERIVED` which has the `istape` attribute set to `True`. Download is not allowed from tape endpoints by non-admin accounts.
Because of this no file is considered for download. Anyway this is probably already fixed with 1.17. Needs to be checked though.

cohm:
Thanks for the quick reply! 

Yes, I should have said that I figured out this much myself as well. I think the output is still wrong and should be fixed.

Best,
Christian

TWAtGH:
With 1.17 this case is checked and the `NoFilesDownloaded` exception is thrown. The CLI exit with a return code != 0. There is no error message though.

bari12:
@TWAtGH Can you add an error message for this case?

TWAtGH:
Yes I'm currently working on the downloadclient anyway

Issue #1427, opened 2018-08-07
Call list_replicas in downloadclient only once


bari12:
Motivation
----------
Right now `list_replicas` is called in a loop for every did. In cases where the pilot runs over N=15 files this creates a lot of calls on the server. This should be changed to just one `list_replicas` bulk call.

The reason why this is done is that the did could potentially be a dataset, and right now, if the call is made in bulk the association between resolved files and dataset gets lost. However, this association is needed for the traces. The idea would be to expand the list_replicas result with the used did. This requires changes in the client and the core.

- [x] Adapt list_replicas to include the dataset/container name in the result
- [x] Adapt the downloadclient to list replicas in bulk

TWAtGH:
Another problem with this is that the download API is currently designed to allow to specify parameters per DID (e.g. rse-expression, schemes, resolve_archives), while list_replicas takes those arguments per call.
Solutions are:
1. Merge the DIDs by the parameters to reduce the list_replicas calls to a minimum
2. Create a new method in the API that is more strict in specifying the parameters but fits to list_replicas
3. Drop the support of specifying parameters per DID in the current download API
4. Change list_replicas so that it is possible to specify parameters per DID

mlassnig:
Change for list_replicas():

Return (multiple) fields in metalink: `<parent>...did...</parent>`

Return single list field in json: `'parents': [did1, did2, ...]`

Convention is not to resolve all parents, only the ones the user is requesting.

Issue #1593, opened 2018-10-02
Download client should report suspicious files properly


cserf:
Motivation
----------

The download client only reports suspicious files if there is a checksum mismatch. In case the download fails because of missing files, it doesn't report anything. The client should parse the error reported by `protocol.get` and if it matches some pattern, report.

Modification
------------
Fix it

bari12:
Link to #1623 as only one of them has to be implemented.

cserf:
#1623 has been implemented and merged. Therefore all downloads with eventType `get_sm` or `get_sm_a` shouldn't report any suspicious files

PR #1956, opened 2019-01-10
Clients: refactored download workflow and added automatic zip extraction; Fix #1354 #1427 #1512

### LINKED ISSUE #1354: client_extract=True replicas need to be extracted from the zip file
Motivation
----------
The client needs to check for replicas with client_extract=True and extract them.

TWAtGH:
Clients: refactored download workflow and added automatic zip extraction; Fix #1354 #1427 #1512


mlassnig (approved):

bari12 (approved):

cserf:
The logic looks correct, but difficult to test ;-) . Just one comment, you can remove the part dedicated to the suspicious file declaration since this part is now handled on the server side : https://github.com/rucio/rucio/blob/master/lib/rucio/daemons/tracer/kronos.py#L143

cserf (approved):

bari12:
This will be done with #1593. Merging this now.

PR #1963, opened 2019-01-11
Core & Internals: added caching to config core; Fix #1958

### LINKED ISSUE #1958: Add caching (memcache) to config table methods
Motivation
----------
Currently the caching has to be handled by all modules using these methods, but it should be done natively. Option to force a readout from the DB should be there as well.

hahahannes:
Core & Internals: added caching to config core; Fix #1958

### DIFF
--- a/lib/rucio/core/config.py
+++ b/lib/rucio/core/config.py
@@ -151,13 +151,15 @@ def get(section, option, default=None, use_cache=True, expiration_time=3600, ses
     if isinstance(value, NoValue):
         tmp = session.query(models.Config.value).filter_by(section=section, opt=option).first()
         if tmp is not None:
-            value = tmp[0]
-            REGION.set(value_key, value)
+            value = __convert_type(tmp[0])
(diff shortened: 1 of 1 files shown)


lib/rucio/core/config.py
```diff
@@ -48,57 +61,73 @@ def add_section(section, session=None):
...
+    :param use_cache: Boolean if the cache should be used.
+    :param expiration_time: Time after that the cached value gets ignored.
     :param session: The database session in use.
     :returns: True/False
     """
-
-    query = session.query(models.Config).filter_by(section=section)
-
-    return True if query.first() else False
+    has_section_key = 'has_%s' % section
```
bari12: I would find it a bit easier yo understand if the lookup keys are always the method name.
Thus
- has_section_%
- options_%
- has_option_%
...
Make it just a bit easier to read and also understand, once you just dump the full cache contents.

bari12 (changes requested):

bari12 (approved):

PR #1966, opened 2019-01-11
Core: add ability to limit the total size usage of an RSE Resolves #1965

### LINKED ISSUE #1965: Limit the maximum space usage on an RSE
Motivation
----------

It should be possible to control the maximum space used for an RSE, in addition to account quotas.

Modification
------------

Add a configurable parameter for the maximum RSE usage and check 
against this in the RSE selector.

illingwo:
Core: add ability to limit the total size usage of an RSE Resolves #1965

### DIFF
--- a/lib/rucio/core/rse_selector.py
+++ b/lib/rucio/core/rse_selector.py
@@ -8,16 +8,18 @@
 # Authors:
 # - Martin Barisits, <martin.barisits@cern.ch>, 2013-2017
 # - Cedric Serfon, <cedric.serfon@cern.ch>, 2015
+# - Robert Illingworth, <illingwo@fnal.gov>, 2019
 #
 # PY3K COMPATIBLE
 
 from random import uniform, shuffle
 
-from rucio.common.exception import InsufficientAccountLimit, InsufficientTargetRSEs, InvalidRuleWeight
+from rucio.common.exception import InsufficientAccountLimit, InsufficientTargetRSEs, InvalidRuleWeight, RSEOverQuota
 from rucio.core.account import has_account_attribute
-from rucio.core.account_counter import get_counter
+from rucio.core.account_counter import get_counter as get_account_counter
 from rucio.core.account_limit import get_account_limit
-from rucio.core.rse import list_rse_attributes, has_rse_attribute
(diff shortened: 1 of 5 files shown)


lib/rucio/core/rse_selector.py
```diff
@@ -98,6 +110,7 @@ def select_rse(self, size, preferred_rse_ids, copies=0, blacklist=[], prioritize
         :param copies:                       Select this amount of copies, if 0 use the pre-defined rule value.
         :param blacklist:                    List of blacklisted rses. (Do not put replicas on these sites)
         :param prioritze_order_over_weight:  Prioritize the order of the preferred_rse_ids list over the picking done by weight.
+        :existing_rse_size:                 Dictionary of size of files already present at each rse
```
bari12: Space missing

bari12 (changes requested):
I think it's good like this. I only found 1 missing space. For the question if it should be an RSE-limit or RSE-Attribute. For now I think RSE-limit is the correct one to do, although it makes me wonder if we really need this separate table. For the sake of simplicity, it might be better to just make these RSE-attributes as well and get rid of the rse_limits table. But this is stuff for future thought.
Please add a test-case where the `RSEOverQuota` exception is raised.

lib/rucio/core/rse_selector.py
illingwo: I've fixed the comment spacing, added a test case, and fixed a bug in the NONE grouping repair function that I'd overlooked before.

I've been wondering what the motivation for the separate RSE limit table is - they're less discoverable than attributes (and there's no CLI commands). But given they do exist it was the obvious place to put this.

bari12: To me the separation doesn't make much sense and I can't quite remember our reasoning when we designed it that way. But for now it's the right place to put that information, we can do some consolidation (and get rid of the table) later.

vingar: RSE attributes are mostly tags/string attributes and limits numerical values with some constraints. I guess It looked cleaner to separate them. With a better support of various types and theirs respective constraints in the rse attributes table, it makes sense to converge on using only this table.

bari12 (approved):

Issue #1954, opened 2019-01-10
Automatically report LOST files as suspicious


dchristidis:
Motivation
----------

This will assist the cloud squads. Pull request #1708 laid the groundwork to process LOST files identified during the consistency check.

Modification
------------

Convert the LFNs to PFNs, then report them as suspicious. The `reason` will be set to `Reported by Auditor` so that it is easily identifiable.

PR #1978, opened 2019-01-15
Consistency: Automatically report LOST files as suspicious #1954

dchristidis:


lib/rucio/daemons/auditor/__init__.py
```diff
@@ -160,12 +164,33 @@ def process_output(output, sanity_check=True, compress=True):
...
+        logger.warning('Number of LOST files is exceeding threshold: "%s"',
+                       output)
+        found_error = True
+    if found_error and sanity_check:
+        raise AssertionError('sanity check failed')
+
+    # While converting LOST replicas to PFNs, entries that do not
+    # correspond to a replica registered in Rucio are silently dropped.
+    lost_pfns = []
+    for chunk in chunks(lost_replicas, 100):
```
mlassnig: you're going straight to the core here, i think it's better if you're not chunking. core internal list_replicas works better with fewer calls. chunking should be only beneficial over the REST interface.

mlassnig: (unless of course you expect thousands of replicas)

bari12:
Hi @dchristidis, can you please rebase this (against master) and push again.

bari12 (approved):

PR #1988, opened 2019-01-16
Core: Decrease chunksize for add_constituents; Fix #1987

### LINKED ISSUE #1987: Decrease chunksize in add_constituents_to_archive from 50 to 20
Motivation
----------
Chunksize of 50 seems to lead to weird ORA error.

bari12:
Core: Decrease chunksize for add_constituents; Fix #1987

### DIFF
--- a/lib/rucio/core/did.py
+++ b/lib/rucio/core/did.py
@@ -311,7 +311,7 @@ def __add_files_to_archive(scope, name, files, account, ignore_duplicate=False, 
     try:
         new_files and session.bulk_insert_mappings(models.DataIdentifier, new_files)
         if existing_files_condition:
-            for chunk in chunks(existing_files_condition, 50):
+            for chunk in chunks(existing_files_condition, 20):
                 session.query(models.DataIdentifier).\
                     with_hint(models.DataIdentifier, "INDEX(DIDS DIDS_PK)", 'oracle').\
                     filter(models.DataIdentifier.did_type == DIDType.FILE).\


vingar:
Weird! what are the ORA errors ? DBAs asks usually to have this limit around 1k - 5k. I would suggest to check the bind var values corresponding to these errors.

bari12:
Without limit we went with 1000 files in this query, but this went into a full partition scan (even with hint), which is a blocker for large scopes.
I added the chunking with a chunksize of 50 then which respected the hint and all went fine in my tests. For Panda though we consistently get ORA-25408 "can not safely replay call" which is something related to database failover. It's weird, as this seems completely unrelated, with a smaller chunksize it works fine though.

vingar:
It's very surprising here. The hint should have stabilised the query. It's worth to check with DBAs what's going on. Maybe some empty scopes are biasing the statistics and the optimiser. "Can not safely replay call" usually comes from a previous transaction not closed properly.

bari12:
We are already investigating, but in the meanwhile it needs to run. It's probably related to the partition statistics the database generates (like you said, maybe a bias between very large/very small scopes), why the hint is ignored, but the actual error is very misleading.
Usually the transaction ORAs are rollback segment related. E.g. ORA-01555 etc.
But when you read the documentation of ORA-25408, this is failover related (e.g. database goes down and switches to failover cluster)

mlassnig (approved):

PR #2008, opened 2019-01-21
Release: Added schema for Belle II; Fix #2007

### LINKED ISSUE #2007: Add schema and construct surl method for Belle2
Motivation
----------
- BelleII schema
- BelleII has also a specific construct_surl function for non-deterministic transfers. At the moment, these functions are part of the core repository, so it will be added there. In the medium term, this has to be changed similar to the lfn2pfn calls.

bari12:
Release: Added schema for Belle II; Fix #2007


bari12:
I am merging this now due the release build. Only failing test is caused by the oracle container issue #2010

PR #2016, opened 2019-01-21
Clients: added possibility to use metalink file in rucio download; Fix #1415

### LINKED ISSUE #1415: rucio download should be able to use locally cached metalink
Motivation
----------

Instead of calling list_replicas use locally cached metalink file.

Ref: @PalNilsson

hahahannes:
Clients: added possibility to use metalink file in rucio download; Fix #1415


bari12:
Please rebase to make tests work.

Issue #2022, opened 2019-01-22
Rucio client support for protocol environment variable


vokac:
Motivation
----------
Rucio download called directly from CLI can be forced to transfer data using specific protocol with "--protocol" command line option. I think it would be also useful to use environment variable (e.g. RUCIO_PROTOCOL) to achieve same behavior, because with this functionality it is pretty easy to pass this "protocol filter" to the jobs running in environment with limited network connectivity that allows only certain protocol to pass through.

Modification
------------
Just minor changes in rucio code to also look at environment variable RUCIO_PROTOCOL in addition to the "--protocol" command line option for download/upload

vokac:
Thinking a bit more about my implementation it is probably not the best/correct one, because it doesn't really apply in case DownloadClient/UploadClient is called through API. I'll first have a look how it is used in pilot and than I could come back with different solution for use case described above.

bari12:
Yeah, I was about to comment that. I think it really has to be added to the DownloadClient and Uploadclient.

vokac:
Now, I'm even thinking if we should polute rucio code with this functionality or if it is job that should be done directly in pilot ... I'll discuss that with @TomasJavurek

vokac:
I briefly discussed how to filter RSE protocols used by jobs on specific site and this functionality should go directly in pilot.

So drop this request.

PR #2023, opened 2019-01-22
Clients:Support for RUCIO_PROTOCOL env variable #2022

vokac:
Clients:Support for RUCIO_PROTOCOL env variable #2022


mlassnig (dismissed):

bari12:
This is still in discussion in #2022 
Adding the env check to bin/rucio probably does not do much, as the pilot never calls bin/rucio. So this either has to be added to the DownloadClient or the pilot directly.

PR #2024, opened 2019-01-22
Clients:Support for RUCIO_PROTOCOL env variable #2022

vokac:
Clients:Support for RUCIO_PROTOCOL env variable #2022


vokac:
How to get rid of this duplicate to #2023

bari12:
The two PR are intended. So all good here 😄

PR #2031, opened 2019-01-23
Core & Internals: added rse expression to resolve all RSEs; Fix #1455

### LINKED ISSUE #1455: Handling of istape=False in the client
Motivation
----------
Currently, the clients only allow downloads from RSEs which are flagged with `istape=False`. There are two issues with this. For some reason, in ATLAS, the `istape` attribute has been stored with the value `"False"` instead of the value `0`, which means the rse_expression only works with the direct string check on `"False"`. Secondly, as this is hardcoded in the client, for communities which do not have their RSEs flagged with `"istape=False"` the download will not work.
We need a better workflow for this.

Some options are:
- Extending the RSE Expression parameter with a system-default `*` parameter, which lists all RSEs. Currently, in ATLAS, the `ALL` flag is used for this, but again, this needs to be created by a script. A system internal parameter would be good. With this a new attribute could be used to flag tape rses: `tape=1`. The client can now do the download with the EXPRESSION `*\tape=1`. Even for communities who do not have the tape flag, this query will work.

hahahannes:
Core & Internals: added rse expression to resolve all RSEs; Fix #1455

### DIFF
--- a/lib/rucio/client/downloadclient.py
+++ b/lib/rucio/client/downloadclient.py
@@ -1151,7 +1151,7 @@ class DownloadClient:
             # extend RSE expression to exclude tape RSEs for non-admin accounts
             rse_expression = item.get('rse')
             if self.is_tape_excluded:
-                rse_expression = 'istape=False' if not rse_expression else '(%s)&istape=False' % rse_expression
+                rse_expression = '*\istape=1' if not rse_expression else '(%s)\istape=1' % rse_expression
             logger.debug('rse_expression: %s' % rse_expression)
 
             # get PFNs of files and datasets
--- a/lib/rucio/core/rse_expression_parser.py
+++ b/lib/rucio/core/rse_expression_parser.py
@@ -10,6 +10,7 @@
   - Vincent Garonne, <vincent.garonne@cern.ch>, 2013
   - Mario Lassnig, <mario.lassnig@cern.ch>, 2013
   - Cedric Serfon, <cedric.serfon@cern.ch>, 2014
+  - Hannes Hansen, <hannes.jakob.hansen@cern.ch>, 2019
 '''
 
 import abc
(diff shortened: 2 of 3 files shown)


hahahannes:
I think we need another name for the new tape attribute because '\t' is used for tabs and this causes problems with the regex, if I got it correctly.
```python
parse_expression('rse=MOCK\istape=1') # correct output
parse_expression('rse=MOCK\tape=1') # Expression does not comply to RSE Expression syntax
```

lib/rucio/client/downloadclient.py
```diff
@@ -1151,7 +1151,7 @@ def _get_sources(self, merged_items, resolve_archives=True):
             # extend RSE expression to exclude tape RSEs for non-admin accounts
             rse_expression = item.get('rse')
             if self.is_tape_excluded:
-                rse_expression = 'istape=False' if not rse_expression else '(%s)&istape=False' % rse_expression
+                rse_expression = '*\istape=1' if not rse_expression else '(%s)\istape=1' % rse_expression
```
mlassnig, line 1154: can you check if it already works with "true" and "false instead of "0" and "1"? we are trying to get rid of "0"/"1" for boolean variables.

hahahannes: ah yes, seems like it does not work with 'true'.
```python
from rucio.core.rse import get_rse_attribute, get_rse_id
rse = 'MOCK2'
id = get_rse_id(rse)
add_rse_attribute(rse, 'istape', 1)
parse_expression('*&istape=true') # empty result
```

mlassnig: ah, we need https://github.com/rucio/rucio/pull/1921 for this to work, yes

bari12: I will review this in a bit. Please close this in the meanwhile and re-submit it as a **feature** when #1921 is available.

hahahannes: ok

mlassnig: once #1921 is in, this can be a patch IMHO

bari12: If it is a patch it won't see the changes of #1921 (as that one is a feature)

hahahannes: Linking #2037 here

PR #2035, opened 2019-01-23
Tools to monitor account usage history : Closes #1952

### LINKED ISSUE #1952: Tools to monitor account usage history
Motivation
----------
There is currently no tool to provide the account usage history on on RSE. 

Modification
------------
Will implement :
- Core+API+REST
- CLI
- WebUI view

cserf:
Tools to monitor account usage history : Closes #1952

### DIFF
--- a/lib/rucio/core/account_counter.py
+++ b/lib/rucio/core/account_counter.py
@@ -9,6 +9,7 @@
 # - Vincent Garonne, <vincent.garonne@cern.ch>, 2013
 # - Mario Lassnig, <mario.lassnig@cern.ch>, 2013-2014
 # - Martin Barisits, <martin.barisits@cern.ch>, 2014
+# - Cedric Serfon, <cedric.serfon@cern.ch>, 2019
 # - Hannes Hansen, <hannes.jakob.hansen@cern.ch>, 2018
 #
 # PY3K COMPATIBLE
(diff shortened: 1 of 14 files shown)


lib/rucio/core/account_counter.py
```diff
@@ -172,3 +156,20 @@ def update_account_counter(account, rse_id, session=None):
 
     for update in updated_account_counters:
         update.delete(flush=False, session=session)
+
+
+@transactional_session
+def update_account_counter_history(account, rse_id, session=None):
```
bari12, line 165: I think this should be renamed to `account_usage_history`, since there are no account counters anymore. It should also be moved to `account.py`.

cserf: In that case, we should also move `update_account_counter`. But then there is almost nothing left in `account_counter.py`. Should we move everything to `account.py` ? But similarly, why do we keep `account_limit.py` ?

bari12: account_limit we shouldn't keep either I think. Everything account.py
The only thing where account_counter might make sense is for the methods which touch the updated_account_counter table. But even methods like add_account_counter are only used in the tests right now.

mlassnig: if we can get everything into `account.py` that would be great

cserf: OK, but the change is too big to be out for 1.19.0. Let's do it for 1.20.0

cserf: Rebased

bari12 (approved):

PR #2037, opened 2019-01-24
Core & Internals: added rse expression to resolve all RSEs; Fix #1455

### LINKED ISSUE #1455: Handling of istape=False in the client

hahahannes:
Core & Internals: added rse expression to resolve all RSEs; Fix #1455

### DIFF
--- a/doc/source/rse_expressions.rst
+++ b/doc/source/rse_expressions.rst
@@ -12,19 +12,37 @@ Simple RSE Expressions
 ^^^^^^^^^^^^^^^^^^^^^^
 Rucio allows to test RSE Expressions, using the command **list-rses**. The most simple RSE Expression is the one containing the name of a particular RSE.
 
-1) The following expression only return a set containing a single RSE::
+1) The following expression returns all RSEs::
+     jbogadog@lxplus0058:~$ rucio list-rses --expression '*'
+     IFIC-LCG2_LOCALGROUPDISK
+     IFAE_PRODDISK
+     PIC_SCRATCHDISK
+     EELA-UNLP_SCRATCHDISK
+     CERN-PROD_TZDISK
+     BNL-OSG2_MCTAPE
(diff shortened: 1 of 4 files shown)


hahahannes:
I think we need another name for the new tape attribute because '\t' is used for tabs and this causes problems with the regex, if I got it correctly.
```python
parse_expression('rse=MOCK\istape=1') # correct output
parse_expression('rse=MOCK\tape=1') # Expression does not comply to RSE Expression syntax
```

bari12:
Should be fine to just escape the backslash. `\\`

hahahannes:
Ah yes, this works. Should I change or do we use the old attribute?

bari12:
I would leave it like it is at the moment. I am not sure if the new attribute is consistently set yet.

mlassnig:
Can confirm that we use `istape` all over the place

bari12:
Can you please add the new `*` operator in the documentation about RSE Expressions as well. Could be useful also for other queries.

bari12 (approved):

Issue #2039, opened 2019-01-25
Implement pcaches into clients.


TomasJavurek:
Motivation
----------
LSM uses pcaches. In order to fulfil wishes of US admins, we need to implement them in rucio so, that rucio mover will use this feature. 

Modification
------------
1) stealing pcache script from lsm
2) implementing check, whether a file is listed in pcache, if so, make a hardlink to its location instead of downloading it again.
3) populating the pcache list.

PR #2043, opened 2019-01-25
common: pcaches for downloadclient - #2039

TomasJavurek:
common: pcaches for downloadclient - #2039


bari12:
I know I originally recommended putting this in `lib/rucio/common` but now thinking about it, I am not so sure anymore as it is a command line tool. Command line tools we only have as part of `bin/` and `tools/`. Optimal would be to have it as a pypi installable package somewhere, then it could just be in the dependency file of the rucio-clients. Any opinions?

TomasJavurek:
E741 ambiguous variable name 'l'

Checking that. I can relocate the script, that's not a problem.

mlassnig:
pcache would be called by the downloadclient, so it won't run as a standalone tool. I think common/utils is still fine.

PR #2055, opened 2019-01-28
Rules: New RuleNotification Mode PROGRESS; Closes #1922

### LINKED ISSUE #1922: New Rule Notification Mode
Motivation
----------
Currently Rucio has three notification modes:
- `YES`: Notify all state changes
- `NO`: Notify no state changes at all
- `CLOSE`: Only notify when rule is `OK` and dataset is closed.

For the TapeCarousel activity an additional mode would be beneficial, where a notification is sent once the rule reaches a certain percentage of completion. e.g. 75%. After that the mode can be changed to `CLOSE`.

bari12:
Rules: New RuleNotification Mode PROGRESS; Closes #1922


lib/rucio/db/sqla/migrate_repo/versions/01eaf73ab656_add_new_rule_notification_state_progress.py
```diff
@@ -0,0 +1,49 @@
...
+# revision identifiers used by alembic
+revision = '01eaf73ab656'       # pylint: disable=invalid-name
+down_revision = '9eb936a81eb1'  # pylint: disable=invalid-name
+
+
+def upgrade():
+    '''
+    Upgrade the database to this revision
+    '''
+    if context.get_context().dialect.name not in ('sqlite', 'mysql'):  # pylint: disable=no-member
```
mlassnig: why not in mysql?

lib/rucio/tests/test_rule.py
```diff
@@ -88,6 +88,17 @@ def check_dataset_ok_callback(scope, name, rse, rule_id, session=None):
     return False
 
 
+@transactional_session
+def check_rule_progress_callback(scope, name, progress, rule_id, session=None):
+    callbacks = session.query(models.Message.id).filter(models.Message.payload == json.dumps({'scope': scope,
```
mlassnig, line 93: there's no index on the message payload column. is this ok? in case we have a backlog of messages, wouldn't this slowdown?

lib/rucio/tests/test_rule.py
```diff
@@ -88,6 +88,17 @@ def check_dataset_ok_callback(scope, name, rse, rule_id, session=None):
...
+                                                                                              'name': name,
+                                                                                              'progress': progress,
+                                                                                              'rule_id': rule_id})).all()
+    if len(callbacks) > 0:
```
mlassnig: if callbacks:

lib/rucio/db/sqla/migrate_repo/versions/01eaf73ab656_add_new_rule_notification_state_progress.py
bari12: I just took this from a different revision. I was under the impression that some of the commands (drop PK?) do not work with mysql, but didn't try it myself. Will take it out.

lib/rucio/tests/test_rule.py
bari12: Yes, this method is just used for testing.

mlassnig (approved):

lib/rucio/core/rule.py
```diff
@@ -1849,27 +1849,29 @@ def update_rules_for_bad_replica(scope, name, rse_id, nowait=False, session=None
...
 
-    :param rule:     The rule object.
-    :param session:  The Database session
+    :param rule:                       The rule object.
+    :param replicating_locks_before:   Amount of replicating locks before the current state change.
+    :param session:                    The Database session
     """
 
     session.flush()
+    total_locks = rule.locks_replicating_cnt + rule.locks_ok_cnt
```
cserf, line 1862: You don't consider the stuck ones ?

lib/rucio/core/rule.py
```diff
@@ -1878,6 +1880,14 @@ def generate_rule_notifications(rule, session=None):
                                          'name': rule.name,
                                          'rule_id': rule.id},
                                 session=session)
+                    if rule.notification == RuleNotification.PROGRESS:
```
cserf, line 1883: It will only apply for the rules in state OK (if statement line 1864) ?

lib/rucio/core/rule.py
```diff
@@ -1911,6 +1921,21 @@ def generate_rule_notifications(rule, session=None):
                     except DataIdentifierNotFound:
                         pass
 
+    elif rule.state == RuleState.REPLICATING and rule.notification == RuleNotification.PROGRESS and replicating_locks_before:
+        # For RuleNotification PROGRESS rules, also notifiy when REPLICATING thresholds are passed
+        if __progress_class(replicating_locks_before, total_locks) != __progress_class(rule.locks_replicating_cnt, total_locks):
+            try:
+                did = rucio.core.did.get_did(scope=rule.scope, name=rule.name, session=session)
+                if not did['open']:
```
cserf, line 1929: Why do you need to check if the DID is open ?

lib/rucio/db/sqla/migrate_repo/versions/01eaf73ab656_add_new_rule_notification_state_progress.py
```diff
@@ -0,0 +1,49 @@
...
+# Revises: 9eb936a81eb1
+# Creation Date: 2019-01-28 10:34:46.605485
+
+from alembic.op import (create_check_constraint, drop_constraint)
+
+from alembic import context
+
+# revision identifiers used by alembic
+revision = '01eaf73ab656'       # pylint: disable=invalid-name
+down_revision = '9eb936a81eb1'  # pylint: disable=invalid-name
```
cserf, line 29: Be careful to change the down_revision when you merge the other PR ;)

lib/rucio/core/rule.py
bari12: Yes. The conveyor does not trigger callbacks for rules who are in STUCK mode and the judge-repairer only triggers the callback once the rules state has changed to REPLICATING again. Thus this will never be called for STUCK rules.

lib/rucio/core/rule.py
bari12: Yes, otherwise the branch in line 1924 covers the REPLICATING ones.

lib/rucio/core/rule.py
bari12: I thought similar to CLOSE, PROGRESS also only applies for closed datasets. Otherwise there might be this weird case where a rule flip/flops between under/over a threshold (e.g. 50%)

lib/rucio/db/sqla/migrate_repo/versions/01eaf73ab656_add_new_rule_notification_state_progress.py
bari12: Always; We have a test for that which cries loudly 😄

PR #2070, opened 2019-01-30
View for suspicious files : Closes #2054

### LINKED ISSUE #2054: View for suspicious files
Motivation
----------
The sites/clouds squads need a view to list the suspicious files on their site.

cserf:
View for suspicious files : Closes #2054

### DIFF
--- a/lib/rucio/api/replica.py
+++ b/lib/rucio/api/replica.py
@@ -14,7 +14,7 @@
 #
 # Authors:
 # - Vincent Garonne <vgaronne@gmail.com>, 2013-2016
-# - Cedric Serfon <cedric.serfon@cern.ch>, 2014-2016
+# - Cedric Serfon <cedric.serfon@cern.ch>, 2014-2019
 # - Thomas Beermann <thomas.beermann@cern.ch>, 2014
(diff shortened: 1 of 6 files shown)


lib/rucio/api/replica.py
```diff
@@ -212,3 +212,13 @@ def list_datasets_per_rse(rse, filters=None, limit=None):
...
+
+
+def get_suspicious_files(rse_expression, younger_than=None, nattempts=None):
+    """
+    List the list of suspicious files on a list of RSEs
+    :param rse_expression: The RSE expression where the suspicious files are located
+    :param younger_than: datetime object to select the suspicious replicas younger than this date.
+    :param nattempts: The number of time the replicas have been declared suspicious
+    """
+    return replica.get_suspicious_files(rse_expression=rse_expression, younger_than=younger_than, nattempts=nattempts)
```
mlassnig, line 224: in the other API calls we usually have seconds and not datetime objects to pass around. any particular reason why datetime objects are more fitting here? something like younger_than=1234 seconds is not enough?

cserf: The API is only used via the WebUI, and I found it more convenient to use datetime

PR #2072, opened 2019-01-30
common: pcache for download client - #2039

TomasJavurek:
common: pcache for download client - #2039


bari12:
Should we merge this already, or do you still want to do the changes in the downloadclient?

TomasJavurek:
I would merge now. At least, US guys can see that we are willing to go in that direction. To implement it into downloadclient might be easy, but there might be also some complications, as usual.

bari12 (approved):

PR #2077, opened 2019-01-31
Core: staging_area replicas including in file replica list and used as transfer source; closes #2069

### LINKED ISSUE #2069: Allow staging_area replicas to be used as transfer sources and include them in the list replicas output
Motivation
----------

Staging area replicas should be available for use as transfer sources and appear in the replica list. This makes the staging area behave more like a "virtual disk"

Modification
------------

Drop the staging_area==False restrictions on the queries.

illingwo:
Core: staging_area replicas including in file replica list and used as transfer source; closes #2069

### DIFF
--- a/lib/rucio/core/replica.py
+++ b/lib/rucio/core/replica.py
@@ -23,6 +23,7 @@
 # - Wen Guan <wguan.icedew@gmail.com>, 2014-2015
 # - Hannes Hansen <hannes.jakob.hansen@cern.ch>, 2018
 # - Dimitrios Christidis <dimitrios.christidis@cern.ch>, 2019
+# - Robert Illingworth, <illingwo@fnal.gov>, 2019
 #
 # PY3K COMPATIBLE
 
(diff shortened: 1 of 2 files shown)


mlassnig:
For me this is fine, but I'm not sure if there's some implication that I'm overseeing. @cserf , opinion?

illingwo:
As far as I know I'm the only one trying to use the staging area, which limits the potential issues if there are any problems. It's not something that's likely to get triggered accidentally.

bari12:
I think this should be fine too. We are handling the staging_areas in a very specific way due to the workflow we had in ATLAS back in the days; But there is no special reason in locking it exactly to this way, so I think this change is fine.
However, I won't merge this into 1.19.0 as I am already preparing the release. @illingwo I would suggest to close this and re-submit as a patch, so we can have it in 1.19.1.

illingwo:
There's a (trivial) merge conflict with the current master; I think it'll be simpler to wait until you've merged next in before resubmitting it.

PR #2083, opened 2019-01-31
metalink: resolve all schemes when None; Fix #1262

### LINKED ISSUE #1262: list_replicas with metalink should resolve all schemes, if no schemes are given
Motivation
----------
Currently the schemes have to be given explicitly. However, if no schemes are given all schemes should be replied, not only the top-priority one.

mlassnig:
metalink: resolve all schemes when None; Fix #1262

### DIFF
--- a/lib/rucio/web/rest/webpy/v1/replica.py
+++ b/lib/rucio/web/rest/webpy/v1/replica.py
@@ -109,7 +109,7 @@ class Replicas(RucioController):
 
         # Resolve all reasonable protocols when doing metalink for maximum access possibilities
         if metalink and schemes is None:
-            schemes = ['gsiftp', 'srm', 'root', 'davs', 'https', 'file']
+            schemes = supported_protocols
 
         try:
             # first, set the appropriate content type, and stream the header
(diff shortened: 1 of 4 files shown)


lib/rucio/web/rest/webpy/v1/replica.py
```diff
@@ -102,6 +102,10 @@ def GET(self, scope, name):
             if 'limit' in params:
                 limit = int(params['limit'][0])
 
+        # Resolve all reasonable protocols when doing metalink for maximum access possibilities
+        if metalink and schemes is None:
+            schemes = ['gsiftp', 'srm', 'root', 'davs', 'https', 'file']
```
bari12: We have lists of all possible schemes in several places in the code. I wonder if we should replace this with something centrally specified or maybe even something community specific?

mlassnig: While I agree a central place makes sense, I don't really see much need for a configurable list as the supported protocols per VO will be automatically resolved through the RSEs anyway?

bari12: Yes, I was just putting it in the room. A central list of all rucio supported protocols should be sufficient, the actual VO support is derived from the used protocols anyway.

bari12 (dismissed):
I think the comment got lost 😄 Please change the schemes list to some central defined list (Maybe also in other parts of the code base)

bari12:
Can you please rebase this and re-submit. travis seems to have problems.

mlassnig:
@bari12 done, travis is back in action now

bari12 (approved):

Issue #2076, opened 2019-01-30
transfer_timeout and other options not propagated to download functions


nikmagini:
Motivation
----------
In the current HEAD of the download client, the transfer_timeout is not propagated to protocol.get
(it is always None)

I think this issue is caused by this refactoring:

https://github.com/rucio/rucio/pull/1957

The transfer_timeout is now stored in the merged_options dictionary, but _download_item is not updated accordingly.

Modification
------------

nikmagini:
Note - this seems to affect also other options stored in merged_options e.g. ignore_checksum

PR #2087, opened 2019-01-31
Clients: propagate options to download protocols #2076

nikmagini:
Clients: propagate options to download protocols #2076

### DIFF
--- a/lib/rucio/client/downloadclient.py
+++ b/lib/rucio/client/downloadclient.py
@@ -16,7 +16,7 @@
 # - Tomas Javurek <tomasjavurek09@gmail.com>, 2018
 # - Vincent Garonne <vgaronne@gmail.com>, 2018
 # - Joaquin Bogado <jbogado@linti.unlp.edu.ar>, 2018
-# - Nicolo Magini <nicolo.magini@cern.ch>, 2018
+# - Nicolo Magini <nicolo.magini@cern.ch>, 2018-2019
 # - Tobias Wegner <tobias.wegner@cern.ch>, 2018
 # - Hannes Hansen <hannes.jakob.hansen@cern.ch>, 2018-2019
 #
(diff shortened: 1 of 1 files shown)


mlassnig:
@nikmagini need to rebase the patch on top of upstream/master

lib/rucio/client/downloadclient.py
```diff
@@ -1270,6 +1270,8 @@ def _prepare_items_for_download(self, did_to_options, merged_items_with_sources,
                 if options is None:
                     continue
                 file_item['merged_options'] = options
+                file_item.setdefault('transfer_timeout', options.get('transfer_timeout'))
```
mlassnig: if I see correctly, options could be `None` here, right? will this still work then?

TWAtGH: No there is the `if options is None: continue` directly above. Actually options should never be `None` except there is a bug in the input merging.
Anyhow I would prefer to change the `_download_item` function to access the transfer_timeout and ignore_checksum fields correctly e.g. `item.get('merged_options', {}).get('transfer_timeout')` instead of copying the values. Could you change this please @nikmagini ?

bari12: Ping @nikmagini

nikmagini: > Anyhow I would prefer to change the `_download_item` function to access the transfer_timeout and ignore_checksum fields correctly e.g. `item.get('merged_options', {}).get('transfer_timeout')` instead of copying the values. Could you change this please @nikmagini ?

I'll look into it - however, IIRC I didn't do that because the `options` dict isn't always copied to `merged_options` (e.g. for pfns download). So I think that I'll have to look both in `options` and in `merged_options` for these keys... I'll let you know

TWAtGH: You can just change it in the pfn download aswell

TWAtGH:
```
options = item.setdefault('merged_options', {})
options.setdefault('ignore_checksum', item.pop('ignore_checksum', True))
options.setdefault('transfer_timeout', item.pop('transfer_timeout', None))
```

nikmagini: Hi, read the suggestion too late - for now I changed it to check for the keys in both dicts.
I agree it would be better to make all methods consistent though.
Can look into it tomorrow...

mlassnig (dismissed):

lib/rucio/client/downloadclient.py
TWAtGH: Did you look into it? :D

nikmagini: Right now - finally pushed the new update...

TWAtGH (approved):

PR #2097, opened 2019-02-01
Core & Internals: added functionality to set tombstone manually; Fix #82

### LINKED ISSUE #82: rucio-admin command to put tombstone on a replica
When upload multiple files, if the number of file is too much so that cause error like "database: too many connections" or "database deadlock for MySQL server" or "can not allocate memory on server". Some upload process may be interrupted at the middle, leaving some DIDs without rule. If one want to delete these DIDs, one will fail, because rucio-undertaker give replicas tombstone when deleting rules:
> /opt/rucio/lib/rucio/core/did.py(604)delete_dids()

-> rucio.core.rule.delete_rule(rule_id=rule_id, purge_replicas=purge_replicas, delete_parent=True, nowait=True, session=session)
But if a DID does not have a rule, rucio-undertaker will go to 
/opt/rucio/lib/rucio/core/did.py(639)delete_dids()
if file_clause:
rowcount = session.query(models.DataIdentifier).filter(or_(*file_clause)).\
filter(models.DataIdentifier.did_type == DIDType.FILE).\
update({'expired_at': None}, synchronize_session=False)
and only reset expired_at to None. As a result, one can not delete these DIDs.

_Migrated from JIRA [RUCIO-2854](https://its.cern.ch/jira/browse/RUCIO-2854)_

hahahannes:
Core & Internals: added functionality to set tombstone manually; Fix #82


lib/rucio/core/replica.py
```diff
@@ -2733,3 +2733,26 @@ def get_suspicious_files(rse_expression, younger_than=None, nattempts=None, sess
...
+    :param session: database session in use.
+    """
+    if not rse_id:
+        rse_id = get_rse_id(rse)
+    replica = session.query(models.RSEFileAssociation).filter_by(rse_id=rse_id, name=name, scope=scope).first()
+    if not replica:
+        raise exception.ReplicaNotFound('Replica %s:%s on RSE %s could not be found.' % (scope, name, rse))
+    if not replica.lock_cnt:
+        replica.tombstone = OBSOLETE
+        replica.save(session=session)
```
vingar: (non blocking) Another way it 's to do the update with the clause `lock_cnt =0` and check the row-count after to know if the update was done or not.  This saves one interaction with the db.

lib/rucio/core/replica.py
```diff
@@ -2733,3 +2733,26 @@ def get_suspicious_files(rse_expression, younger_than=None, nattempts=None, sess
...
+
+    :param rse: name of the RSE.
+    :param scope: scope of the replica DID.
+    :param name: name of the replica DID.
+    :param rse_id: optional ID of RSE.
+    :param session: database session in use.
+    """
+    if not rse_id:
+        rse_id = get_rse_id(rse)
+    replica = session.query(models.RSEFileAssociation).filter_by(rse_id=rse_id, name=name, scope=scope).first()
```
bari12: .first() can actually be .one() as it can only be one (or none)

lib/rucio/core/replica.py
bari12: I think we should make it sure and really check the locks table here as well, just to make sure that there are no locks and that the lock_cnt is not inconsistent. Could also be done in one query with a join.

hahahannes: okay, shall I also change to one query? Makes sense to me as its saving interactions. But then I cannot distinguish if  the replica was not found or if its locked in case no rows get updated. Maybe useful information for error handling on the client side.

vingar: `update table where not exists (select 1 from another_table) etc` might work. 
If the update returns 0 rows affected, one can check the replicas presence, etc to give back a proper error message. It's still performant.

lib/rucio/api/replica.py
```diff
@@ -242,3 +243,17 @@ def get_suspicious_files(rse_expression, younger_than=None, nattempts=None):
...
+def set_tombstone(rse, scope, name, issuer):
+    """
+    Sets a tombstone on one replica.
+
+    :param rse: name of the RSE.
+    :param scope: scope of the replica DID.
+    :param name: name of the replica DID.
+    :param issuer: The issuer account
+    """
+    if not permission.has_permission(issuer=issuer, action='set_tombstone', kwargs={}):
```
bari12, line 257: We should add the permission (just check against admin) to the permission files too.

hahahannes: I wanted to use the default permission check: [https://github.com/rucio/rucio/blob/master/lib/rucio/core/permission/generic.py#L111](https://github.com/rucio/rucio/blob/master/lib/rucio/core/permission/generic.py#L111).
Should I add it anyway?

bari12: Oh yes right, just leave it like this then 😄 Sorry for the noise.

bari12 (approved):

PR #2098, opened 2019-02-01
Clients: added success message to declare-bad command; Fix #2038

### LINKED ISSUE #2038: print message when declaring files as bad
Motivation
----------

```
Done

> rucio-admin replicas declare-bad --reason "dcache does not know anything about this file" srm://lcg-se1.sfu.computecanada.ca:8443/srm/managerv2?SFN=/atlas/atlasdatadisk/rucio/data18_13TeV/19/db/DAOD_EXOT3.15351753._000211.pool.root.1
> echo $?
0


I did not get any feedback from the above command which is unnerving !  (DDM, would you consider a message printout as to what was done ?)   

Thanks !

regards,
Asoka
```

hahahannes:
Clients: added success message to declare-bad command; Fix #2038

### DIFF
--- a/bin/rucio-admin
+++ b/bin/rucio-admin
@@ -909,13 +909,23 @@ def declare_bad_file_replicas(args):
     # Group file list in separate sublists for different schemes
     bad_files_pfns.sort()
     bad_files_pfns_grouped = groupby(bad_files_pfns, lambda f: f[:f.find('://')])
+    all_non_declared = {}
 
     for sublist in bad_files_pfns_grouped:
         for chunk in chunks(list(sublist[1]), 500):
             non_declared = client.declare_bad_file_replicas(pfns=chunk, reason=args.reason)
             for rse in non_declared:
-                for pfn in non_declared[rse]:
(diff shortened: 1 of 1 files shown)


bin/rucio-admin
```diff
@@ -916,6 +916,11 @@ def declare_bad_file_replicas(args):
             for rse in non_declared:
                 for pfn in non_declared[rse]:
                     print('%s : PFN %s cannot be declared.' % (rse, pfn))
+
+    successfully_declared_pfns = [pfn for pfn in bad_files_pfns if pfn not in [not_declared_pfn.split(' ')[0] for entry in list(non_declared.values()) for not_declared_pfn in entry]]
```
mlassnig: Whoops, forgot to commit my previous review.

Wouldn't it make more sense to do this the other way round? Basically say "Everything went fine" and if not, display only those that didn't work? So it won't spam the screen for naught?

hahahannes: ah yes, makes more sense

mlassnig (approved):

Issue #2104, opened 2019-02-04
Lost files - make it uniqe


TomasJavurek:
Motivation
----------
There are double counts in the lost files reports when the same lost file is in many datasets. 

Modification
------------
easy

TomasJavurek:
The new version from Alexander Bogdanchikov is used from last week on.

PR #2110, opened 2019-02-04
probes: lost files double counts - #2104

TomasJavurek:
probes: lost files double counts - #2104


cserf:
Looks good, but can you make it py3k compatible ?

bari12:
Ping @TomasJavurek

TomasJavurek:
Pong @bari12

TomasJavurek:
Well the error says:
No module named 'retrying'
but that seems more like problem with pip install, or venv, than py3k, right?

bari12:
What Cedric means is making the file PY3K ready. You can for example use `pylint --py3k` to check what's wrong. print statements for example.

PR #2152, opened 2019-02-08
Keep the history of the susbcriptions : Closes #1810

### LINKED ISSUE #1810: Keep the history of the susbcriptions
Motivation
----------
When a subscription is changed, we don't keep the history. Will implement subscriptions history

cserf:
Keep the history of the susbcriptions : Closes #1810


bari12 (changes requested):
As discussed: We need a config-table switch for the history generation, as it will fail due to the missing `comments` column in the history table. Once the column is there (`1.20.0`) we can remove the switch.

bari12 (approved):

PR #2153, opened 2019-02-08
Keep the history of the susbcriptions : Closes #1810

### LINKED ISSUE #1810: Keep the history of the susbcriptions

cserf:
Keep the history of the susbcriptions : Closes #1810


bari12:
This has to go in as a feature, as the update_subscription part will block communities which use the models.py as their base of schema generation, due to the PK on subscriptions_history.
We need to get rid of the Versioned table first, add a SubscriptionsHistory model (without PK) and add an alembic upgrade script as well.

bari12 (approved):

PR #2164, opened 2019-02-12
Core & Internals: added email field to list identities of account; Fix #2142

### LINKED ISSUE #2142: Email missing from list identities
Motivation
----------
Looking at the output from list_identities in the account client, it does not return the e-mail address for the identity. However, the set identity takes this parameter and the field exists in the database.  Is this an oversight or is it intentional?

The underlying motivation for this is that I'm trying to migrate the accounts and identities from one server to another.

Modification
------------

hahahannes:
Core & Internals: added email field to list identities of account; Fix #2142


bari12 (dismissed):

bari12:
Please rebase.

bari12 (approved):

PR #2169, opened 2019-02-13
Core & Internals: added password option to identity add; Fix #2140

### LINKED ISSUE #2140: impossible to create userpass identity
Motivation
----------

userpass authentication requires a --password option, in addition to --id (which is for x509/ssh/kerberos).

This is available through the API, but not bin/rucio

hahahannes:
Core & Internals: added password option to identity add; Fix #2140


lib/rucio/web/rest/webpy/v1/identity.py
```diff
@@ -63,18 +63,20 @@ def PUT(self, account):
         """
         username = ctx.env.get('HTTP_X_RUCIO_USERNAME')
         password = ctx.env.get('HTTP_X_RUCIO_PASSWORD')
+        email = ''
```
hahahannes: Just added a blanc mail because the test I added to test_identity.py failed on postgres and oracle because of a DB constraint that email should not be none. Can also fix it in this PR properly or open a new issue.

bari12 (approved):

PR #2171, opened 2019-02-14
Core: staging_area replicas including in file replica list and used as transfer source; closes #2069

### LINKED ISSUE #2069: Allow staging_area replicas to be used as transfer sources and include them in the list replicas output

illingwo:
Core: staging_area replicas including in file replica list and used as transfer source; closes #2069


bari12:
Hi Robert, please do a quick rebase with master. We had to change the oracle test case as it was failing due to a pulled docker container.

bari12 (approved):

PR #2230, opened 2019-02-20
Core & Internals: added check if endpoint supports requested content-type; Fix #1730

### LINKED ISSUE #1730: REST API not returning full JSON
Motivation
----------

It seems like the Rucio REST API does not actually return true JSON. This is an e-mail from @vkuznet to me about this. The context is that CMS uses something called DAS to aggregate info across our data management system, data catalog, and other systems. It relies on REST APIs to do this in a reasonable fashion (not a client library for each system). Thoughts?

It seems Rucio APIs do not return proper JSON:

```
curl -H "X-Rucio-Auth-Token: $token" http://cms-rucio-test.cern.ch/rses/

{"rse_type": "DISK", "id": "c7ed36b8906c4808b090b35976373f88", "region_code": null, "city": null, "updated_at": "Fri, 28 Sep 2018 19:11:56 UTC", "rse": "T2_US_Nebraska_Test", "created_at": "Fri, 28 Sep 2018 19:11:56 UTC", "ISP": null, "deleted": false, "time_zone": null, "longitude": null, "availability": 7, "deterministic": true, "volatile": false, "country_name": null, "latitude": null, "deleted_at": null, "continent": null, "ASN": null, "staging_area": false}
{"rse_type": "DISK", "id": "1a8167d0f9c743749ecd3bf00d4fb04c", "region_code": null, "city": null, "updated_at": "Wed, 24 Oct 2018 12:06:07 UTC", "rse": "T3_US_NERSC", "created_at": "Wed, 24 Oct 2018 12:06:07 UTC", "ISP": null, "deleted": false, "time_zone": null, "longitude": null, "availability": 7, "deterministic": true, "volatile": false, "country_name": null, "latitude": null, "deleted_at": null, "continent": null, "ASN": null, "staging_area": false}

```

and

```

curl -H "X-Rucio-Auth-Token: $token" http://cms-rucio-test.cern.ch/accounts/
{"account": "das", "type": "USER", "email": null}
{"account": "dciangot", "type": "USER", "email": null}

```

Yes, they return JSON dicts, but it is not single JSON document, and therefore
it fails to be parsed. Here is example to prove it.

```
curl -H "X-Rucio-Auth-Token: $token" http://cms-rucio-test.cern.ch/accounts/ > account.json
python
Python 2.7.5 (default, Jul 13 2018, 13:06:57)
[GCC 4.8.5 20150623 (Red Hat 4.8.5-28)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
import json
data = json.load(open("accounts.json"))
 File "/usr/lib64/python2.7/json/__init__.py", line 290, in load
   **kw)
 File "/usr/lib64/python2.7/json/__init__.py", line 338, in loads
   return _default_decoder.decode(s)
 File "/usr/lib64/python2.7/json/decoder.py", line 369, in decode
   raise ValueError(errmsg("Extra data", s, end, len(s)))
ValueError: Extra data: line 2 column 1 - line 9 column 1 (char 50 - 419)

```
I also tried to use HTTP header Content-Type: application/json and results are
the same.

I don't want to crawl over strings (this is what APIs return now) and convert
them to JSON docs. Instead, I would [like]  a JSON document. For the record, the
proper JSON would be something like (based on accounts API output):

```
[
{"account": "das", "type": "USER", "email": null},
{"account": "dciangot", "type": "USER", "email": null}
]

```
It is also desired that all APIs will use the same data-type, a list of dicts,
even if one item in a list is present. This will allow to abstract parsing code
much easily and make it API independent.

Best,
Valentin.

Modification
------------

hahahannes:
Core & Internals: added check if endpoint supports requested content-type; Fix #1730

### DIFF
--- a/lib/rucio/web/rest/flaskapi/v1/replica.py
+++ b/lib/rucio/web/rest/flaskapi/v1/replica.py
@@ -46,11 +46,12 @@ from rucio.common.exception import (AccessDenied, DataIdentifierAlreadyExists, I
                                     RSENotFound, UnsupportedOperation, ReplicaNotFound)
 from rucio.common.replica_sorter import sort_random, sort_geoip, sort_closeness, sort_dynamic, sort_ranking
 from rucio.common.utils import generate_http_error_flask, parse_response, APIEncoder, render_json_list
-from rucio.web.rest.flaskapi.v1.common import before_request, after_request
+from rucio.web.rest.flaskapi.v1.common import before_request, after_request, check_accept_header_wrapper_flask
 
 
 class Replicas(MethodView):
 
+    @check_accept_header_wrapper_flask(['application/json', 'application/metalink4+xml'])
     def get(self, scope, name):
         """
         List all replicas for data identifiers.
@@ -71,6 +72,7 @@ class Replicas(MethodView):
         :status 200: OK.
         :status 401: Invalid auth token.
         :status 404: DID not found.
+        :status 406: Not Acceptable.
         :status 500: Internal Error.
         :returns: A dictionary containing all replicas information.
         :returns: A metalink description of replicas if metalink(4)+xml is specified in Accept:
@@ -272,6 +274,7 @@ class Replicas(MethodView):
 
 class ListReplicas(MethodView):
 
+    @check_accept_header_wrapper_flask(['application/json', 'application/metalink4+xml'])
     def post(self):
         """
         List all replicas for data identifiers.
@@ -295,6 +298,7 @@ class ListReplicas(MethodView):
         :status 400: Cannot decode json parameter list.
         :status 401: Invalid auth token.
         :status 404: DID not found.
+        :status 406: Not Acceptable.
         :status 500: Internal Error.
         :returns: A dictionary containing all replicas information.
         :returns: A metalink description of replicas if metalink(4)+xml is specified in Accept:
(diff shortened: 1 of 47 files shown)


mlassnig:
Looks good to me, second pair of eyes @bari12  ?

lib/rucio/web/rest/flaskapi/v1/replica.py
```diff
@@ -272,6 +274,7 @@ def delete(self):
 
 class ListReplicas(MethodView):
 
+    @check_accept_header_wrapper_flask(['application/json', 'application/metalink4+xml'])
```
bari12: Why `application/json` here and not `application/x-json-stream` as defined below?
It's a bit strange anyway, since we define `application/x-json-stream` but the data in line 370 is not yielded...

hahahannes: Ah yes. Because its for flask and we dont have streaming there at the moment. Should I change to `application/x-json-stream` as we are changing it in the future anyway?

bari12: Ah understood.
I would say we use the ones from weby 1:1, even if flask does not allow it. Since we are not using flask anyway (other than the doc generation)

bari12 (approved):

Issue #2112, opened 2019-02-05
Client: expose replicalocks to client


dciangot:
Motivation
----------
Allow python apis to list the DID status in a rule, in the context of a automatic script for monitoring the user file transfer for instance. 
AFAIU the rest call is already there, the only work needed is to expose that to the client

Modification
------------
Expose replicalocks call defined here: https://github.com/rucio/rucio/blob/master/lib/rucio/web/rest/webpy/v1/rule.py

PR #2273, opened 2019-03-05
Clients: expose replicalocks to the client #2112

dciangot:
Clients: expose replicalocks to the client #2112


dciangot:
closing since it is against master, instead of next. sorry.

PR #2283, opened 2019-03-06
Probe : Declare lost suspicious log files

cserf:
Probe : Declare lost suspicious log files


bari12 (changes requested):
`tools/probes/common/check_suspicious_logs:27:36: E271 multiple spaces after keyword`

bari12 (approved):

PR #2289, opened 2019-03-06
Clients: expose replicalocks to the client #2112

dciangot:


bari12:
Yes this one looks good 😄

lib/rucio/client/ruleclient.py
```diff
@@ -227,3 +227,18 @@ def examine_replication_rule(self, rule_id):
             return next(self._load_json_data(r))
         exc_cls, exc_msg = self._get_exception(r.headers, r.status_code)
         raise exc_cls(exc_msg)
+
+    def examine_replication_locks(self, rule_id):
```
bari12: I would suggest to name this `get_replica_locks` or `list_replica_locks`.
The `examine` from the previous method was a bit specific to finding transfer errors, as the return dictionary is specifically prepared for this. Also please adapt the description text.

dciangot: Done, sorry about that!

bari12 (approved):

PR #2300, opened 2019-03-08
Automated alembic downgrade upgrade on all databases

mlassnig:
oracle should work


mlassnig:
NB:

(1) SQLite is out of the alembic rotation, since it doesn't support ALTER TABLE statements.

(2) All DB travis runs will now initialise the model, run a complete downgrade, then a complete upgrade, and only then will run the unittest

(3) Most of the time the Oracle and MySQL scripts are similar, since they both tend to be more lax with the interpretation of the SQL standard. PostgreSQL is more strict, so it's a bit more cumbersome in the scripts.

(4) The `next` branch of this pull request will always fail. Needs a separate feature pull request afterwards to bring the one outstanding script on `next` up to speed.

lib/rucio/db/sqla/migrate_repo/versions/01eaf73ab656_add_new_rule_notification_state_progress.py
```diff
@@ -39,7 +39,7 @@ def upgrade():
 
     elif context.get_context().dialect.name == 'postgresql':
         schema = context.get_context().version_table_schema + '.' if context.get_context().version_table_schema else ''
-        op.execute('ALTER TABLE ' + schema + 'rules ALTER COLUMN notification TYPE CHAR')
+        op.execute('ALTER TABLE ' + schema + 'rules ALTER COLUMN notification TYPE CHAR')  # pylint: disable=no-member
```
vingar, line 42: To avoid writing this exclusion in the future,  the class `alembic.op` can be added in `pylintrc`:`ignored-classes`.

mlassnig: Thx!

bari12 (approved):

Issue #1787, opened 2018-11-15
Bring S3 and Swift signature support in line with rest of code


jamesp-epcc:
Motivation
----------
Rucio has support for S3-style objectstore download and upload, using shared secrets (with boto and 's3cmd'), as well as native support for S3 signed URLs.

There are a few implications:
* Shared secrets, as the name implies, require distribution of the secrets to clients which limits its usefulness. We cannot guarantee confidentiality as shared secrets can leak easily.
* The code itself is still from early days of R&D and follows an old architecture.
* Swift is an S3-style compatible protocol, targeted to OpenStack instead of Amazon, which exists as an untested prototype.

In the meantime, URL signatures were implemented natively in Rucio, with a Google Cloud Storage signature plugin,  which are similar to S3 signatures. The first objective would be to revisit the S3 and Swift signatures and bring it in line with the remainder of the code, such that it can be used transparently by the replica core.

Modification
------------

vingar:
One difference between the Google Cloud Storage signature plugin and S3/swift ones is the need to call an external service to generate the signed url. To no affect the performance on one of the most called APIs `list_replicas`, the signed urls operation has ben isolated in a dedicated endpoint. cf. ´lib/rucio/web/rest/objectstore.py´. For example, with the degradation of this signed url external service, the responses time can increase together with the number of simultaneous db sessions  resulting in less slots available on http frontends (snowball effect). A solution might be between the two approaches.

mlassnig:
There's the possibility to sign S3 offline without having to contact an external service. This should work for both private S3 servers (because we control the private key) and for AWS (similar to the delegated credential loading we do for GCS).

vingar:
Interesting. It looks new. For which S3 version ? spec ?  ceph (the most popular s3 object stores for ATLAS) is usually behind for new features unfortunately. And for swift ?

jamesp-epcc:
I'm not sure if it's mentioned in the documentation, but my experimentation shows that the signing operations in the boto3 API work fully offline (they still work if I disable the network on my VM) but the ones in the old boto API (currently used by Rucio) don't. So it might be necessary to use boto3 if we want offline url signing, unless it can somehow be made to work with boto 2.

PR #2331, opened 2019-03-14
Core & Internals: Implement URL signing for Swift object stores #1787

jamesp-epcc:
This adds support for Swift tempurls to the core Rucio code. I have tested this against our OpenStack object store at Edinburgh. It works very similarly to the S3 support in my previous pull request.

### DIFF
--- a/lib/rucio/client/credentialclient.py
+++ b/lib/rucio/client/credentialclient.py
@@ -38,7 +38,7 @@ class CredentialClient(BaseClient):
         """
         Return a signed version of the given URL for the given operation.
 
-        :param service: The service the URL points to (gcs, s3)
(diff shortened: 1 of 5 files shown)


mlassnig (commented):
looks good to me and will sign-off after confirmation from @bari12 about--> is this ok as a pure next->next merge?

bari12:
For me this is fine 👍 ; There is no formal requirement of which branch to use; In the contribution guide we just say "create a local branch", which `next` is as well 😄 - so as long as the changes are fine, this is ok.
I would still recommend to create a separate branch for each issue (maybe with the helper scripts) - otherwise it will become difficult if you ever work on multiple changes at once.

mlassnig (approved):

PR #2353, opened 2019-03-19
Conveyor : Submitter continues to submit transfers jobs even for sites blacklisted for write : Closes #2352

### LINKED ISSUE #2352: Submitter continues to submit transfers jobs even for sites blacklisted for write
Motivation
----------
Conveyor Submitter continues to submit transfers jobs even for sites blacklisted for write

cserf:
Conveyor : Submitter continues to submit transfers jobs even for sites blacklisted for write : Closes #2352

### DIFF
--- a/lib/rucio/core/transfer.py
+++ b/lib/rucio/core/transfer.py
@@ -8,7 +8,7 @@
 # Authors:
 # - Martin Barisits, <martin.barisits@cern.ch>, 2017-2018
 # - Mario Lassnig, <mario.lassnig@cern.ch>, 2017-2018
-# - Cedric Serfon, <cedric.serfon@cern.ch>, 2018
+# - Cedric Serfon, <cedric.serfon@cern.ch>, 2018-2019
 # - Hannes Hansen, <hannes.jakob.hansen@cern.ch>, 2018
 # - Robert Illingworth, <illingwo@fnal.gov>, 2019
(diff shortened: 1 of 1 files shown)


mlassnig (dismissed):
there is no space in front of : in english!!!

lib/rucio/core/transfer.py
```diff
@@ -889,22 +896,26 @@ def __set_transfer_state(external_host, transfer_id, new_state, session=None):
 
 
 @read_session
-def __get_unavailable_read_rse_ids(session=None):
+def __get_unavailable_rse_ids(activity, session=None):
```
bari12: This should be called `operation` not `activity`, since it might be confusing with the transfer activities. In other places in the code we already call this operation.

bari12 (changes requested):
One minor change. I was struggling a bit since we now also have this `third_party_copy` operation. But I think we just use it for the protocols, not for the actual RSE availability. So this should be fine.

cserf:
> there is no space in front of : in english!!!

"I'm French! Why do think I have this outrageous accent, you silly king!"

lib/rucio/core/transfer.py
cserf: Fixed

bari12 (approved):

PR #2385, opened 2019-03-25
Clients: handle sourceless file download correctly; Fix #2362

### LINKED ISSUE #2362: Format error with `rucio download`
Motivation
----------

```
# rucio --version 
rucio 1.19.4
# rucio download   "238357;238356;238359:manda4_cp1_3.00u_SW20140322.part1.tar"
2019-03-21 13:58:45,998    INFO    Processing 1 item(s) for input
2019-03-21 13:58:45,998    INFO    Getting sources of DIDs
2019-03-21 13:58:46,434    ERROR    %d format: a number is required, not NoneType
```

TWAtGH:
Clients: handle sourceless file download correctly; Fix #2362

### DIFF
--- a/lib/rucio/client/downloadclient.py
+++ b/lib/rucio/client/downloadclient.py
@@ -1146,7 +1146,7 @@ class DownloadClient:
             if force_scheme:
                 schemes = force_scheme if isinstance(force_scheme, list) else [force_scheme]
             else:
-                schemes = ['davs', 'gsiftp', 'https', 'root', 'srm', 'file']
+                schemes = ['davs', 'gsiftp', 'http', 'https', 'root', 'srm', 'file']
             logger.debug('schemes: %s' % schemes)
 
             # extend RSE expression to exclude tape RSEs for non-admin accounts
--- a/lib/rucio/client/downloadclient.py
+++ b/lib/rucio/client/downloadclient.py
@@ -17,7 +17,7 @@
 # - Vincent Garonne <vgaronne@gmail.com>, 2018
 # - Joaquin Bogado <jbogado@linti.unlp.edu.ar>, 2018
 # - Nicolo Magini <nicolo.magini@cern.ch>, 2018-2019
-# - Tobias Wegner <tobias.wegner@cern.ch>, 2018
+# - Tobias Wegner <tobias.wegner@cern.ch>, 2018-2019
 # - Hannes Hansen <hannes.jakob.hansen@cern.ch>, 2018-2019
 #
 # PY3K COMPATIBLE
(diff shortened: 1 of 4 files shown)


lib/rucio/client/downloadclient.py
```diff
@@ -1145,7 +1145,7 @@ def _get_sources(self, merged_items, resolve_archives=True):
             if force_scheme:
                 schemes = force_scheme if isinstance(force_scheme, list) else [force_scheme]
             else:
-                schemes = ['davs', 'gsiftp', 'https', 'root', 'srm', 'file']
+                schemes = ['davs', 'gsiftp', 'http', 'https', 'root', 'srm', 'file']
```
mlassnig: replace with the import from lib/rucio/rse/protocols/__init__.py ?

bari12: We should change all of these to a list in https://github.com/rucio/rucio/blob/master/lib/rucio/common/constants.py

TWAtGH: Wont this be set to `None` anyway as soon as list_replicas resolves this correctly?

vingar: won't it be better to reflect by default all the schemes reported by listing replicas instead of adding some filtering on top ?

bari12: It's probably fine to just give no schemes to `list_replicas` as this should resolve all schemes since #1262 was implemented.

vingar (dismissed):
Can you add or extend the relevant testcases to validate these changes ?

TWAtGH:
Not really until we have better error handling in the DownloadClient. I can add a test case for that the download should fail if there are no sources but I cannot check that it failed because of this specific scenario :man_shrugging:

vingar:
I meant having tests for the condition described in #2362:  Download of a file with the name ending with `.tar` and on a `http://` endpoint.

TWAtGH:
Ah that should be possible, although I don't know how. Won't make it for todays release then I guess.

lib/rucio/tests/test_replica.py
```diff
@@ -346,6 +346,32 @@ def test_list_replica_with_domain(self):
...
+                               'lan': {'read': 1, 'write': 1, 'delete': 1},
+                               'wan': {'read': 1, 'write': 1, 'delete': 1}}})
+
+        scope = 'mock'
+        name = 'element_%s' % generate_uuid()
+        file_item = {'scope': scope, 'name': name, 'bytes': 1234, 'adler32': 'deadbeef'}
+
+        add_replicas(rse=rse, files=[file_item], account='root')
+
+        replicas = list(rc.list_replicas([{'scope': scope, 'name': name}], schemes=['http']))
```
vingar: maybe removing the explicit`schemes=['http']` as it should work by default?

bari12 (approved):

PR #2406, opened 2019-03-28
Conveyor    :    Implement protection against double submission   :   Closes #1953

### LINKED ISSUE #1953: Implement protection against double submission
Motivation
----------
The Deterministic External id (eid) sent to FTS is generated using as seeds `voname`, `base_id` (both static, but obtained from a `whoami` against the FTS server) and the `request_id` of the first file of the FTS job. This Determinitic eid was introduced to prevent submitting 2 jobs for the same file : if there is already one job submitted for a given request_id, FTS will reject another attempt because they will have the same eid.
Problems can occur when a first submission is done to FTS and FTS doesn't respond even if the submission was successful. In that case a second attempt will be done with the same `request_id`. Here are the potential issues :  
- The `whoami` fails and in that case `voname` and `base_id` are set to None. In that case the external eid generated will be different even if the `request_id` is the same.
- Even if the `whoami` succeeds, the fact that only the `request_id` of the first file is considered leads to the fact that the transfers of [file1, file2] and [file2, file1] will generate 2 different eid. So basically the eid is useless for jobs with multiple destination replicas (i.e. more than 90 % of our transfers)

cserf:
Conveyor    :    Implement protection against double submission   :   Closes #1953

### DIFF
--- a/lib/rucio/common/exception.py
+++ b/lib/rucio/common/exception.py
@@ -19,7 +19,7 @@
 # - Mario Lassnig <mario.lassnig@cern.ch>, 2012-2017
 # - Vincent Garonne <vgaronne@gmail.com>, 2012-2018
 # - Ralph Vigne <ralph.vigne@cern.ch>, 2012-2013
-# - Cedric Serfon <cedric.serfon@cern.ch>, 2013-2018
+# - Cedric Serfon <cedric.serfon@cern.ch>, 2013-2019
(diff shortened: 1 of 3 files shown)


mlassnig (commented):
It looks ok to me, I'm just wondering if using the Exception for flow control is a good approach. But I don't see another way that wouldn't complicate the loop even more.

bari12 (approved):

PR #2430, opened 2019-04-03
Transfers: Automatically recover requests in state 'protocol mismatch' : Closes #2115

### LINKED ISSUE #2115: Automatically recover requests in state "protocol mismatch"
Motivation
----------

When RSE configurations are repaired then M-state requests are not picked up again automatically. Do this by default, but add switch for ATLAS to explicitly disable it.

cserf:
Transfers: Automatically recover requests in state 'protocol mismatch' : Closes #2115

### DIFF
--- a/lib/rucio/common/exception.py
+++ b/lib/rucio/common/exception.py
@@ -926,7 +926,7 @@ class MetalinkJsonParsingError(RucioException):
 
 class ReplicaIsLocked(RucioException):
     """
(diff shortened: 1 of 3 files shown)


mlassnig (commented):
i can't see any obvious problem, but the patch is huge. quite likely i missed something. @bari12 second pair of eyes?

bari12 (approved):

PR #2435, opened 2019-04-04
Core: Removal of ReplicaState.SOURCE as it is not used; Fix #1864

### LINKED ISSUE #1864: Add the possibility to remove suspicious files from the bad files summary plots
Motivation
----------
The number of suspicious is way more bigger than the actually big one. One should have the possibility to filter out the suspicious files from the bad files summary plots

bari12:
Core: Removal of ReplicaState.SOURCE as it is not used; Fix #1864


lib/rucio/db/sqla/migrate_repo/versions/7ec22226cdbf_new_replica_state_for_temporary_.py
```diff
@@ -55,6 +55,7 @@ def downgrade():
     elif context.get_context().dialect.name == 'postgresql':
         schema = context.get_context().version_table_schema + '.' if context.get_context().version_table_schema else ''
         op.execute('ALTER TABLE ' + schema + 'replicas ALTER COLUMN state TYPE CHAR')  # pylint: disable=no-member
+        drop_constraint('REPLICAS_STATE_CHK', 'replicas', type_='check')
```
mlassnig: maybe i'm not seeing it, but is there dialect='mysql' missing for the downgrade?

bari12: It's below, but there is no downgrade for the constraint (so no drop_constraint) which is needed in mysql now. But we should probably address these in a separate PR (It's partly already in @hahahannes PR)

mlassnig (approved):

PR #2451, opened 2019-04-08
Dataset deletion: Mechanism to decrease contention in Undertaker; Fix #2355

### LINKED ISSUE #2355: Pause dids in the undertaker which raise nowait errors
Motivation
----------
In case the rule-deletion in the undertaker raises  nowait DB error, the undertaker just retries the same did right now. Instead, the undertaker should pause these dids for 1-10min and only retry then.

bari12:
Dataset deletion: Mechanism to decrease contention in Undertaker; Fix #2355

### DIFF
--- a/lib/rucio/daemons/undertaker/undertaker.py
+++ b/lib/rucio/daemons/undertaker/undertaker.py
@@ -15,7 +15,7 @@
 # Authors:
 # - Vincent Garonne <vgaronne@gmail.com>, 2013-2018
 # - Cedric Serfon <cedric.serfon@cern.ch>, 2013-2015
-# - Martin Barisits <martin.barisits@cern.ch>, 2016-2018
+# - Martin Barisits <martin.barisits@cern.ch>, 2016-2019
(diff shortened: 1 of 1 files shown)


lib/rucio/daemons/undertaker/undertaker.py
```diff
@@ -81,8 +99,14 @@ def undertaker(worker_number=1, total_workers=1, chunk_size=5, once=False):
                     record_counter(counters='undertaker.delete_dids', delta=len(chunk))
                 except RuleNotFound as error:
                     logging.error(error)
-                except DatabaseException as error:
-                    logging.error('Undertaker(%s): Got database error %s.', worker_number, str(error))
+                except (DatabaseException, DatabaseError, UnsupportedOperation) as e:
+                    if match('.*ORA-00054.*', str(e.args[0])):
```
mlassnig: missing check for other databases?

bari12: Good point, I actually need to check if we consider other databases in the other daemons at all.

mlassnig (approved):

PR #2472, opened 2019-04-12
deletion: Reaper automatic mode fix #2469

### LINKED ISSUE #2469: Automatic mode for reaper
Motivation
----------

Cf. Reaper's proposal:
- Global performance: All reaper daemon instances should have a steady deletion performance. A situation where some of them are busy and some are idle should not occur, if there is a deletion backlog to operate on.
-  Protection of storage: Measures need to be in place to protect the storage from deletion overload. It should be possible to specify the maximum number of threads concurrently interacting with a storage element.
-  Intelligent distribution of workload: The global deletion workload needs to be distributed intelligently. RSEs having a large deletion backlog need to be handled with a larger priority than RSEs with a very small backlog. At the same time, starvation of RSEs cannot happen either.

Modification
------------

The principle is to have a pool of threads available. The method is then to assign dynamically a list of RSEs to each thread with several constraints (max threads per RSE).  This implementation is based on several hash functions and introduced feature flags to roll out gradually the reaper in the new auto mode in a backward compatible way and with reapers running without this `mode`.

- add `--auto` mode  for work automatically on RSE flagged with the reserved RSE attribute `tombstone`, e.g., `rucio-reaper --auto  --total-workers 6 --threads-per-worker 10 `
- Exclude RSE flagged with the reserved RSE attribute tombstone in the non auto mode,
- Add support for the `greedy` reserved RSE attribute in the `auto` mode
- Add `dry-mode` option to check the logic without doing deletion,  e.g., `rucio-reaper --auto --dry-mode  --total-workers 6 --threads-per-worker 10`
- add `--dynamic` mode to consider only the RSEs with something to delete
- add support for several RSE-threads mapping function `--map-function`: `list_rses_for_thread_hash` and `list_rses_for_thread_sequential`.

vingar:


bari12:
Hi,

 It would have been helpful to have some high-level explanation of the idea/concept similar to the two other proposals we had, would have accelerated the reading of this quite a bit. 😄 But I powered through it.

The **Protection of storage** requirement holds, as the hashing with the concatenation basically guarantees that you can't have more than N threads per RSE. You can't really define a max_number_of_threads per RSE, or at least not easily, as the space is split in the `list_rses_for_thread` method. It's not 100% clear to me if we need this though (Although it also doesn't hurt to have it)

For the **Global performance** and **Intelligent distribution of workload** I am not so sure though. As the total space is split strictly based on hashes in the `list_rses_for_thread`. I think this potentially creates the same issue we are having now, just on a smaller scale. Some threads will have no RSE allocation, some multiple and this is intermixed between RSEs which need a lot of deletion and others which do need none. 
This is the key issue, that `list_rses_for_thread` is not aware of the load of the RSEs. Maybe this can be included, but I couldn't think of a way to do it yet.
In general, it is statistically less likely to get into issues here, since the workload space is split over much more threads than right now, where one reaper exclusively deals with a set of RSEs, but it still happens (See simulation)

I ran a few simulations with this hash assigning of RSEs to threads with different number of threads and rses and different deletion backlogs. For the backlog I assumed a power law distribution with a maximum of 3E6 files. 

https://cernbox.cern.ch/index.php/s/EICD73vPsSkx8pn

You can see that some threads work on a backlog of RSEs with 7 million files while others only have backlogs of 1 million files. And this partitioning is basically constant and not re-distributed. In proposal 1 the distribution is basically re-calculated every cycle, which gets the distribution much closer to the red dotted line (Not exactly, but close enough)

I also did a second discrete event simulation with deletion cycles where I compared proposal 1 with your proposal (3). The assumption here is again a power law distribution of maximum of 3 million files to delete per RSE. Objective of the simulation is to see how many cycles it takes to get all deletion done on all RSEs. (assuming there are no new files to delete during the run)
One cycle is having each thread delete 500 (configurable) files on an RSE.

https://cernbox.cern.ch/index.php/s/ec08loZp5obku5s

The long tail is due to this imbalance of very full RSEs with not so full ones.

vingar:
Few random remarks:
- The hash method guarantees that there is exactly N thread (N = total threads per RSE =10) working on one RSE.
- In the study, the hash is applied on a random string whereas in the code it is applied on RSE.id / uuid which gives a known better distribution: in avg the same numbers of RSEs is allocated per thread. The more auto reaper threads, less RSEs there are per thread (horizontal scaling). The option `--dry-run` with `--auto` can  be used to study the distribution rses per threads.
- See the option dynamic to restrict the RSEs to the ones having a deletion backlog only. I didn't put by default as it makes the list of RSES more dynamic (and it might work without).
- Relying on heartbeats exposes the same issue than the rest of the daemons regarding the mutable view on the heartbeats and list of RSEs but one iteration is relatively short and the randomisation of RSEs should decrease the probability of having two threads working on the same replicas. 
- Another option can be to use one extra `payload` field in the heartbeat table to have an exclusive allocation of RSEs per threads (e.g. select for update) but it's not necessary imo as the hash function can provide this.

bari12:
> * The hash method guarantees that there is exactly N thread (N = total threads per RSE =10) working on one RSE.

Yes, agree. (See my first point) - it's also confirmed by the study.

> * In the study, the hash is applied on a random string whereas in the code it is applied on RSE.id / uuid which gives a known better distribution: in avg the same numbers of RSEs is allocated per thread. The more auto reaper threads, less RSEs there are per thread (horizontal scaling). The option `--dry-run` with `--auto` can  be used to study the distribution rses per threads.

That shouldn't matter since md5 is a cryptographic hash function (So it does support non-continuity of output values) - I have changed the study to use uuids, it's the same result.
In theory it should be the same number of RSEs per thread, but we can observe two effects her: The RSEs do not have the same deletion backlog (This is the main effect) and secondly the small cardinality of RSEs. Statistically, the hashing will create a uniform distribution, but we are dealing with a very small sample here, that's why we see such a big difference. (Some threads having 7 RSEs to deal with, some 0). You can easily test this by changing the `TOTAL_RSES`. The higher it gets, the more uniform the distribution gets. At 100k there is only a 10% difference between the "biggest" thread and the smallest.

> * See the option dynamic to restrict the RSEs to the ones having a deletion backlog only. I didn't put by default as it makes the list of RSES more dynamic (and it might work without).

While the dynamic mode helps, I don't think it changes the underlying issue, that the assignment of RSE to thread is not based on the load. Thus you cut away the ones which do not have a backlog, but the ones who do are still assigned randomly.  The cycle study actually already runs in the dynamic setting; In line 111 the rse is removed from the RSE list if there is nothing to delete anymore. If this wouldn't be done the result would be about 20% worse.

> * Relying on heartbeats exposes the same issue than the rest of the daemons regarding the mutable view on the heartbeats and list of RSEs but one iteration is relatively short and the randomisation of RSEs should decrease the probability of having two threads working on the same replicas.

The risk of working on the same replicas is dealt with pretty much in the same way with both proposals (similar to how it is done right now). Although in your one, as the assignment RSE -> thread is stable, there is much less change of partitions. I don't think this is an issue though in both proposals.

> * Another option can be to use one extra `payload` field in the heartbeat table to have an exclusive allocation of RSEs per threads (e.g. select for update) but it's not necessary imo as the hash function can provide this.

The idea of the payload function in the other proposal is really only to fulfil the **Storage protection** criteria. Since RSEs are not assigned to threads, but every thread works on essentially everything, some kind of limiting is needed. Since the assignment of RSEs to threads already takes care of this in your proposal, there is no added benefit of using the payload.



I have a few ideas how it might be possible to get load conditions into the distribution of RSEs to thread in the `list_rses_for_thread` method, but I have to think a bit about it, not yet sure if it is feasible.

vingar:
I have update the issue with more explanation. The method to assign a list of RSEs to each thread with the constraints can have several implementations. This implementation is based on a hash function but others implementation can exist.

About the study, I'm not sure about the numbers threads - list of RSEs distribution.  
If we consider 888 RSEs with something to delete and 600 threads, I got: `avg : 14.7 RSEs per thread, min.: 5 RSEs, max. 27 RSEs`

I don't see any threads with no rse assigned. Now, based on the current load: 273 RSEs need deletion, it gives: ` 9.08485856905158069883527454242928452579	1	21	601`

Notes: 
One reaper instance will mix the RSE types (datadisk, userdisk, etc) with different queue lengths, file sizes, etc. It's better. Now each thread can also adapt its strategy regarding RSE priority and deletion strategy (big files first, LRU, etc.).
The mapping can be done at the process level also and then each process can handle the load a bit better among their threads, but with threads it was the less intrusive and good incremental option.

Respective queries:

```
select avg(total_rses), min(total_rses), max(total_rses), count(1)
from(
select ora_hash(id||(n), 600) as thread_number, count(1) as total_rses
        from atlas_rucio.rses,      
        (SELECT LEVEL n
        FROM DUAL
        CONNECT BY LEVEL <= 10)
group by  ora_hash(id||(n), 600))
```

```select avg(total_threads), min(total_threads), max(total_threads), count(1)
from(
select ora_hash(atlas_rucio.rses.id||(n), 600) as thread_number, count(1) as total_threads
        from 
        atlas_rucio.rses,
        atlas_rucio.rse_usage,
        (SELECT LEVEL n
        FROM DUAL
        CONNECT BY LEVEL <= 20)
where   atlas_rucio.rses.id = atlas_rucio.rse_usage.rse_id
and  source = 'expired' and files > 0 and deleted = 0
group by  ora_hash(atlas_rucio.rses.id||(n), 600)       
)```

vingar:
btw, another deterministic rses-thread mapping giving in addition len(rses)*concurrency /total threads rses per thread is doable. I just took the first one coming to my mind.

bari12:
> I have update the issue with more explanation. The method to assign a list of RSEs to each thread with the constraints can have several implementations. This implementation is based on a hash function but others implementation can exist.

Thanks. Yes, that's what I am currently thinking about if we can get the load somehow into the method. Do you have an idea for that?

> About the study, I'm not sure about the numbers threads - list of RSEs distribution.
> If we consider 888 RSEs with something to delete and 600 threads, I got: `avg : 14.7 RSEs per thread, min.: 5 RSEs, max. 27 RSEs`

I have updated the study to that numbers (888 RSEs, 600 threads) It get's the same numbers as you: 
https://cernbox.cern.ch/index.php/s/h1eKJAN3CURRcr9
It is again a skewed distribution; The issue though is really the last plot, that there is no load consideration in the assignment. So you have some threads with a massive backlog but others with almost none, while implicitly it would be better if all have, more or less, the same.

Similar with the "current" load (273 RSEs, 600 threads): https://cernbox.cern.ch/index.php/s/EICD73vPsSkx8pn

> Notes:
> One reaper instance will mix the RSE types (datadisk, userdisk, etc) with different queue lengths, file sizes, etc. It's better. Now each thread can also adapt its strategy regarding RSE priority and deletion strategy (big files first, LRU, etc.).
> The mapping can be done at the process level also and then each process can handle the load a bit better among their threads, but with threads it was the less intrusive and good incremental option.

Yes the mixing of everything is the same in the other proposal. That's clearly something we should strive for. The manual assignment of RSEs to reapers should be avoided (One of the reasons we are having issues now)
The key here is that one thread can only optimise within it's assigned group. So it can decide to treat RSEx more often, but this means that RSEy (which might need a lot of deletion too) will have 1 thread less taking care of it. This is different in the other proposal since it can optimise on a global picture. The only way I can see doing this here is by considering the load in `list_rses_for_thread`

I have updated the cycle model also to 600 threads, 273 RSEs:
https://cernbox.cern.ch/index.php/s/hqDgNW1Dl7vehQc
At the end there is also a plot of the theoretical maximum (=Number of RSEs having a backlog * THREADS_PER_RSE) versus the used threads in the cycle. This shows the distance to the optimum nicely.

vingar:
The ideal solution is n(concurrency) threads constantly working for one RSE. Let have 400 RSEs, we need 4000 threads with concurrency = 10.  
With this implementation, with 20 auto reaper processes, it means 200 threads per process. It sounds reasonable. Then we have In avg 2.5 RSEs per thread, min 1,  max 9. Having at max 9 threads getting 1/10th of the queue length of RSE sounds good and Having threads doing nothing is not an issue imo. I would just increase the reaper processes and the nb threads. 
Yes, the matchmaking function RSE-thread have place for improvements (weight, map, queue, master slave, etc).

bari12:
What did you conceptually change?

vingar:
It's mainly support for a new method which assigns RSEs to the threads by keeping the constraint of having N threads per RSE and keeping in average  len(rses)*N /total RSEs  per thread. The description of the issue has been updated accordingly.

bari12:
Ah ok. I had one idea how to get the load into the `list_rses_for_thread` method:
The fact that once all threads are established the thread_number (eg. thread 75 out of 600) does not change anymore. (Other than crashes, etc.). Thus based on this knowledge if all threads do the pick of an RSE deterministically, we can get priority into the picture.
Thus thread 75 knows (deterministically) the RSE picks of thread 1 to 74, so it can make it's pick accordingly, etc.
The issue here is that the load-conditions change and the situation might be different when thread 75 makes the calculation than when thread 23 did it, but the only impact here is that there might be a few threads too many working on an RSE.

I am not convinced though that this is more advantageous than proposal 1, which is a bit more dynamic in that regard, since it does the limit check on execution.

vingar:
The method `list_rses_for_thread_sequential` introduces the load with the dynamic option and assigns only the loaded RSEs to threads. It also keeps in average the same number of RSEs per thread.
It's hard to make any suppositions on how several threads working on the same RSEs for few moment can have a bad impact when the list of loaded RSEs changed. It should be measured when deploying gradually the approach. Considering all RSEs and increasing the number of reaper processes solves naturally this.

bari12:
Closing this now, since we decided for the other proposal. I am keeping the branch locally for the changes with the sequential mode, since we might want to integrate these into the other proposal as well.

PR #2476, opened 2019-04-12
Clients: revert gfal/xrdcp workaround; Fix #1598

### LINKED ISSUE #1598: revert xrdcp workaround in downloadclient as soon as gfal is fixed
Motivation
----------
We implemented a workaround (#1596) since gfal doesnt correctly call xroot copy when an archive PFN is given.
This can be removed again as soon as gfal is fixed.

Modification
------------

TWAtGH:
Clients: revert gfal/xrdcp workaround; Fix #1598

### DIFF
--- a/lib/rucio/client/downloadclient.py
+++ b/lib/rucio/client/downloadclient.py
@@ -574,8 +574,9 @@ class DownloadClient:
             self._send_trace(trace)
             return item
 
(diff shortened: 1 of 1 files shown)


mlassnig (approved):
only merge together with metalink fix

PR #2478, opened 2019-04-12
Core & Internals: upgrade mysql version; Fix #2421

### LINKED ISSUE #2421: problems upgrading to latest mysql/mariadb version
Motivation
----------
I am trying to use the latest mysql version 8 which results in multiple erros when trying to setup the schema via running the tests in the docker dev environment.
This is needed for #2220.

Modification
------------
### problem with loading the authentication plugin
- `Unable to load authentication plugin 'caching_sha2_password'.`
- solved by `--default-authentication-plugin=mysql_native_password`

### too long primary keys
- `Specified key was too long; max key length is 3072 bytes`
- solved by `--character-set-server=latin1`

### syntax error in setting up the rules table
- `You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'IS NOT NULL`
- solved by escaping 

### missing index while applying upgrade 22d887e4ec0a
- `Failed to add the foreign key constraint. Missing index for constraint 'SOURCES_REPLICA_FK' in the referenced table 'replicas'") [SQL: u'ALTER TABLE sources ADD CONSTRAINT SOURCES_REPLICA_FK FOREIGN KEY(scope, name, rse_id) REFERENCES replicas (scope, name, rse_id)']`
- solved by changing the primary key while upgrading the replicas table

hahahannes:
Core & Internals: upgrade mysql version; Fix #2421

### DIFF
--- a/lib/rucio/db/sqla/migrate_repo/versions/1c45d9730ca6_increase_identity_length.py
+++ b/lib/rucio/db/sqla/migrate_repo/versions/1c45d9730ca6_increase_identity_length.py
@@ -14,6 +14,7 @@
 #
 # Authors:
 # - Mario Lassnig <mario.lassnig@cern.ch>, 2017-2019
+# - Hannes Hansen <hannes.jakob.hansen@cern.ch>, 2019
 
 ''' increase identity length '''
 
@@ -26,6 +27,8 @@ from alembic.op import alter_column, create_check_constraint, create_foreign_key
 revision = '1c45d9730ca6'
 down_revision = 'b4293a99f344'
 
+schema = context.get_context().version_table_schema if context.get_context().version_table_schema else ''
+
 
 def upgrade():
     '''
@@ -33,8 +36,6 @@ def upgrade():
     '''
 
     if context.get_context().dialect.name in ['oracle', 'postgresql']:
-
-        schema = context.get_context().version_table_schema if context.get_context().version_table_schema else ''
         alter_column('tokens', 'identity', existing_type=sa.String(255), type_=sa.String(2048), schema=schema)
         alter_column('identities', 'identity', existing_type=sa.String(255), type_=sa.String(2048), schema=schema)
         alter_column('account_map', 'identity', existing_type=sa.String(255), type_=sa.String(2048), schema=schema)
@@ -58,6 +59,8 @@ def upgrade():
         alter_column('account_map', 'identity', existing_type=sa.String(255), type_=sa.String(2048), nullable=False)
         create_foreign_key('ACCOUNT_MAP_ID_TYPE_FK', 'account_map', 'identities', ['identity', 'identity_type'], ['identity', 'identity_type'])
 
+        drop_constraint('ACCOUNT_MAP_ID_TYPE_CHK', 'account_map', type_='check')
+        drop_constraint('IDENTITIES_TYPE_CHK', 'identities', type_='check')
         create_check_constraint(constraint_name='IDENTITIES_TYPE_CHK',
                                 table_name='identities',
                                 condition="identity_type in ('X509', 'GSS', 'USERPASS', 'SSH')")
@@ -94,7 +97,6 @@ def downgrade():
         alter_column('identities', 'identity', existing_type=sa.String(2048), type_=sa.String(255))
 
     elif context.get_context().dialect.name == 'postgresql':
-        schema = context.get_context().version_table_schema + '.' if context.get_context().version_table_schema else ''
         execute("DELETE FROM " + schema + "account_map WHERE identity_type='SSH'")  # pylint: disable=no-member
         execute("DELETE FROM " + schema + "identities WHERE identity_type='SSH'")  # pylint: disable=no-member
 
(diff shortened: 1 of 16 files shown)


lib/rucio/db/sqla/models.py
```diff
@@ -774,7 +774,8 @@ class RSEFileAssociation(BASE, ModelBase):
                    CheckConstraint('bytes IS NOT NULL', name='REPLICAS_SIZE_NN'),
                    CheckConstraint('lock_cnt IS NOT NULL', name='REPLICAS_LOCK_CNT_NN'),
                    Index('REPLICAS_TOMBSTONE_IDX', 'tombstone'),
-                   Index('REPLICAS_PATH_IDX', 'path', mysql_length=NAME_LENGTH))
+                   Index('REPLICAS_PATH_IDX', 'path', mysql_length=NAME_LENGTH),
+                   Index('REPLICAS_PK_IDX', 'scope', 'name', 'rse_id'))
```
hahahannes: This seems a bit wrong, as it is already the primary key but it was the only fix to get rid of the error:
```
Failed to add the foreign key constraint. Missing index for constraint 'SOURCES_REPLICA_FK' in the referenced table 'replicas'") [SQL: u'ALTER TABLE sources ADD CONSTRAINT SOURCES_REPLICA_FK FOREIGN KEY(scope, name, rse_id) REFERENCES replicas (scope, name, rse_id)']
```
while applying the alembic upgrade 22d887e4ec0a

hahahannes: Also it is causing now an error with oracle because there is already an index

hahahannes: Fixed, was caused by a bug in a downgrade step.

lib/rucio/db/sqla/models.py
```diff
@@ -61,6 +61,15 @@ def compile_binary_oracle(type_, compiler, **kw):
     return "NUMBER(1)"
 
 
+@event.listens_for(Table, "before_create")
```
bari12: This is problematic like this, as it makes the models.py super hard to read. Doesn't it work if the string below is just
`"'GROUPING' IS NOT NULL"` ?

hahahannes: `"'GROUPING' IS NOT NULL"` seems not to work on mysql and ``"`GROUPING` IS NOT NULL"`` does not work on postgres unfortunately

illingwo: `CheckConstraint(grouping!=None, name='RULES_GROUPING_NN'),` should give the dialect appropriate identifier quoting.

hahahannes: Yes that worked, thanks!

lib/rucio/db/sqla/migrate_repo/versions/1c45d9730ca6_increase_identity_length.py
```diff
@@ -58,12 +58,12 @@ def upgrade():
         alter_column('account_map', 'identity', existing_type=sa.String(255), type_=sa.String(2048), nullable=False)
         create_foreign_key('ACCOUNT_MAP_ID_TYPE_FK', 'account_map', 'identities', ['identity', 'identity_type'], ['identity', 'identity_type'])
 
-        create_check_constraint(constraint_name='IDENTITIES_TYPE_CHK',
```
hahahannes: Removed the constraint for mysql because with the latest version, mysql actually implements the constraints and without dropping the constraint, the creation fails because of duplicates. Unfortunately the drop syntax is different and alembic/sqlalchemy dont have the right syntax yet (see [https://github.com/sqlalchemy/alembic/issues/554](https://github.com/sqlalchemy/alembic/issues/554))

hahahannes:
When the alembic version 1.0.11 is released on pypi we should test if it works, then we can use the normal `drop_constraint` method instead the raw sql

mlassnig (dismissed):
looks good to me. second pair of eyes @illingwo ?

lib/rucio/db/sqla/migrate_repo/versions/22d887e4ec0a_create_sources_table.py
```diff
@@ -54,7 +56,7 @@ def upgrade():
 
         create_primary_key('SOURCES_PK', 'sources', ['request_id', 'rse_id', 'scope', 'name'])
         create_foreign_key('SOURCES_REQ_ID_FK', 'sources', 'requests', ['request_id'], ['id'])
-        create_foreign_key('SOURCES_REPLICA_FK', 'sources', 'replicas', ['scope', 'name', 'rse_id'], ['scope', 'name', 'rse_id'])
+        create_foreign_key('SOURCES_REPLICA_FK', 'sources', 'replicas', ['rse_id', 'scope', 'name'], ['rse_id', 'scope', 'name'])
```
illingwo, line 59: Did you mean to reorder the foreign key columns here?

hahahannes: This was the old wrong order [rse_id,scope,name] which got fixed by the upgrade 3345511706b8 to [scope,name,rse_id]. But in this upgrade it should be the old one.

bari12 (changes requested):
I would like to wait with this one until https://github.com/sqlalchemy/alembic/issues/554 is in alembic
Then we have no need about this special handling of the CHECK Constraints in the update scripts.

bari12:
For me this looks good now with the new sqlalchemy version. The only thing I would suggest is in the tests, where you changed to `mysql-server:latest` to go back to explicit versions.

Issue #2480, opened 2019-04-15
Compiling issues with Python 3 (3.6.8) and hash sum calculation


XeBoris:
Motivation
----------
  - There are several compiling issues observed when installing Rucio in Python 3.6.8. These errors are coming mainly from Py2->Py3 translations so far.
  - An error in the encoding (utf8) is observed when calculating the hash for determination of the final file path.

Modification
------------
  - Modifications are necessary in storm.py and protocol.py are necessary.

hahahannes:
@bari12  Maybe we should change the python3 test to check all changed files now instead of only files with the py3 header

PR #2481, opened 2019-04-15
Clients: Fix installation issues with Py2->Py3 and hash sum #2480

XeBoris:
There are two issues observed in: 
  - Installation in Python 3.6.8 failed because of bad translated Py2->Py3 code in rucio/lib/rucio/rse/protocols/storm.py when it comes to evaluate the exceptions.
  - The hash sum calculation failed due to an encoding error when the hash sum is calculated for the final path definition. (protocols.py)

Both are addressed in this pull request.

### DIFF
--- a/lib/rucio/client/uploadclient.py
+++ b/lib/rucio/client/uploadclient.py
@@ -68,10 +68,8 @@ class UploadClient:
         self.trace['eventType'] = 'upload'
         self.trace['eventVersion'] = version.RUCIO_VERSION[0]
 
-
     def upload(self, items, summary_file_path=None):
         """
-
         :param items: List of dictionaries. Each dictionary describing a file to upload. Keys:
             path                  - path of the file that will be uploaded
             rse                   - rse name (e.g. 'CERN-PROD_DATADISK') where to upload the file
--- a/lib/rucio/client/uploadclient.py
+++ b/lib/rucio/client/uploadclient.py
@@ -43,7 +43,7 @@ from rucio import version
 
 class UploadClient:
 
-    def __init__(self, _client: object = None, logger: object = None, tracing: object = True) -> object:
+    def __init__(self, _client=None, logger=None, tracing=True):
         """
         Initialises the basic settings for an UploadClient object
 
@@ -317,7 +317,6 @@ class UploadClient:
             logger.debug('local checksum: %s, remote checksum: %s' % (file['adler32'], meta['adler32']))
 
             if meta['adler32'] != file['adler32']:
-                logger.info("Here")
                 logger.error('Local checksum %s does not match remote checksum %s' % (file['adler32'], meta['adler32']))
 
                 raise DataIdentifierAlreadyExists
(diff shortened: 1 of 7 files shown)


lib/rucio/client/uploadclient.py
```diff
@@ -43,7 +43,7 @@
 
 class UploadClient:
 
-    def __init__(self, _client=None, logger=None, tracing=True):
+    def __init__(self, _client: object = None, logger: object = None, tracing: object = True) -> object:
```
bari12: The type hints are not backwards compatible unfortunately 😞

lib/rucio/client/uploadclient.py
```diff
@@ -312,8 +315,11 @@ def _register_file(self, file, registered_dataset_dids):
             meta = self.client.get_metadata(file_scope, file_name)
             logger.info('File DID already exists')
             logger.debug('local checksum: %s, remote checksum: %s' % (file['adler32'], meta['adler32']))
+
             if meta['adler32'] != file['adler32']:
+                logger.info("Here")
```
bari12: Please remove

lib/rucio/common/pcache.py
```diff
@@ -873,11 +873,20 @@ def daemonize(self):
...
+
+        #try to be backwards compatible with Python2 (?)
+        try:
+            for fd in xrange(0, maxfd + 1):
+                try:
+                    os.close(fd)
+                except:
+                    pass
+        except NameError:
+            for fd in range(0, maxfd + 1):
```
bari12: Why not just use range in both cases?

bari12 (changes requested):

lib/rucio/common/pcache.py
XeBoris: Indeed, this is much simpler.

lib/rucio/client/uploadclient.py
XeBoris: oh! I forgot to delete it before uploading.

lib/rucio/client/uploadclient.py
XeBoris: Right! I changed it back

XeBoris:
I submitted the requested changes but the build still fails. The log files saying it is failing only for one test (py2.7 with Oracle) and Travis ends with these lines in the log file (everything before looks reasonable to me):
```
RSE (CLIENTS): get protocols for operations of rse. ... ok
RSE (CLIENTS): get protocols for operations of rse (RSEOperationNotSupported). ... 
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated
```

Furthermore, Python3.6 tests seems to fail since flake8 is missing:
```
Running flake8                 
./tools/travis/check_syntax.sh: line 23: flake8: command not found
```

bari12:
Hi Boris,
Yes, the oracle test hangs sometimes, I am usually just restarting them, but now on vacation so I didn't see this right away.
The missing flake8 for py36 we know about. This will be fixed as well.
Thanks!

Issue #2411, opened 2019-03-29
Use signed URLs when deleting from object stores


jamesp-epcc:
Motivation
----------
From our original plan: "For third-party-copy, FTS3 supports `s3://` as a protocol and can do the URL signing itself. Like for Google Cloud Storage, the credential will be stored in the FTS3 server additionally. The replica core and transfer daemon need to skip URL signatures if the request is for the transfer daemon, but correctly sign for the deletion daemon. There is a legacy `s3+https` protocol implementation in Rucio which should be removed in favour of this common workflow".

Modification
------------
Having looked at the code, it should be relatively straightforward to add a call to get the signed URL (if required for the RSE) to the reaper. The core credential code can already sign `DELETE` URLs for GCP, S3 and Swift. The legacy code could then be removed.

I am not sure whether the planned rewrite of the reaper will affect this.

PR #2499, opened 2019-04-18
Deletion: Sign URL for deletion in reaper when required #2411

jamesp-epcc:
This only works correctly with GFAL2 2.16.2 or later (earlier versions caused Davix to perform a stat operation before the actual delete, which fails for signed URLs). See https://its.cern.ch/jira/browse/DMC-1155 for details.


mlassnig:
ALRB is only shipping EMI4.0.3-1v3 right now, which pulls in GFAL2 2.15.4.

So while the code is fine, this won't work for ATLAS until the next EMI release.

mlassnig (approved):

PR #2509, opened 2019-04-29
pcache: v1.0 - #2039

TomasJavurek:
pcache: v1.0 - #2039

### DIFF
--- a/lib/rucio/client/downloadclient.py
+++ b/lib/rucio/client/downloadclient.py
@@ -1510,7 +1510,7 @@ class DownloadClient:
         :raises NoFilesDownloaded:
         :raises NotAllFilesDownloaded:
         """
-        success_states = ['ALREADY_DONE', 'DONE']
+        success_states = ['ALREADY_DONE', 'DONE', 'FOUND_IN_PCACHE']
         # failure_states = ['FILE_NOT_FOUND', 'FAIL_VALIDATE', 'FAILED']
         num_successful = 0
         num_failed = 0
--- a/lib/rucio/client/downloadclient.py
+++ b/lib/rucio/client/downloadclient.py
@@ -42,6 +42,7 @@ from threading import Thread
 from rucio.client.client import Client
 from rucio.common.exception import (InputValidationError, NoFilesDownloaded, ServiceUnavailable,
                                     NotAllFilesDownloaded, RSENotFound, RucioException, SourceNotFound)
+from rucio.common.pcache import Pcache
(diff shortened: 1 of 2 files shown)


lib/rucio/common/pcache.py
```diff
@@ -9,7 +9,7 @@
 import subprocess
 import signal
 
-from urllib import urlopen, urlencode
+from urllib import urlopen, urlencode  # not py3
```
bari12: The issue why the module_importer fails is because of this. (On the py35 test)

bari12: This is somewhat weird to me since urllib is imported as a dependency of requests, and requests is 35 compatible. So one option would be to just change these calls to requests.

lib/rucio/client/downloadclient.py
```diff
@@ -536,6 +538,8 @@ def _download_item(self, item, trace, log_prefix=''):
         :returns: dictionary with all attributes from the input item and a clientState attribute
         """
         logger = self.logger
+        pcache = Pcache()
```
TWAtGH: `pcache = Pcache() if self.check_pcache and len(item.get('archive_items', [])) == 0 else None`

lib/rucio/client/downloadclient.py
```diff
@@ -574,6 +578,43 @@ def _download_item(self, item, trace, log_prefix=''):
             self._send_trace(trace)
             return item
 
+        # checking Pcache
+        storage_prefix = None
+        if self.check_pcache:
```
TWAtGH: `if pcache:`

lib/rucio/client/downloadclient.py
```diff
@@ -660,6 +701,16 @@ def _download_item(self, item, trace, log_prefix=''):
         first_dest_file_path = next(dest_file_path_iter)
         logger.debug("renaming '%s' to '%s'" % (temp_file_path, first_dest_file_path))
         os.rename(temp_file_path, first_dest_file_path)
+
+        # if the file was downloaded with success, it can be linked to pcache
+        if self.check_pcache:
```
TWAtGH: `if pcache:`

TWAtGH (dismissed):
I think the pcache check should be disabled if download_item is going to download a client extract archive

lib/rucio/client/downloadclient.py
TomasJavurek: Let's do it like that for now, but that perhaps contradicts the main motivation for pcaches.

PR #2516, opened 2019-04-29
CLI : Strange error: No section: 'policy' : Fixes #2514

### LINKED ISSUE #2514: CLI : Strange error: No section: 'policy'
Motivation
----------

If no policy section exists in the `rucio.cfg`, `rucio` exists with :
```
rucio -v list-rse-attributes TAIWAN-LCG2_TAPE-STAGING
...

2019-04-29 16:03:49,002 ERROR   could not convert string to float: True
2019-04-29 16:03:49,002 ERROR   Strange error: No section: 'policy'
```

cserf:
CLI : Strange error: No section: 'policy' : Fixes #2514

### DIFF
--- a/bin/rucio
+++ b/bin/rucio
@@ -19,7 +19,7 @@
 # - Martin Barisits, <martin.barisits@cern.ch>, 2012-2018
 # - Thomas Beermann, <thomas.beermann@cern.ch>, 2012-2017
 # - Yun-Pin Sun, <winter0128@gmail.com>, 2012-2013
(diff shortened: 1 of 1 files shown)


TWAtGH:
This should already be fixed in the correct place:
#2463

TWAtGH (changes requested):
Better to just pass `raise_exception=False` instead of catching the exception. But since #2463 is merged we could just close this PR

cserf:
Ah, I didn't see that there was already a PR that was merged.

Issue #533, opened 2018-01-25
Better way to deal with configuration / permissions / policy


bari12:
Motivation
----------
Right now the different configuration/permissions (atlas, cms) are part of the core repository. This is not sustainable in the long term. There should be a better way to manage these e.g. separate packages and inclusion via entry points etc.

bari12:
This was discussed in the [Dev Meeting on 2018-10-11](https://indico.cern.ch/event/760955/)
Two things should be added:
- Mentioning a module in the Rucio CFG which should be loaded. This module can be part of a config package (e.g. from ATLAS or CMS) which is installed via PyPi
- Alternatively also directly mentioning the path of a python file should be possible, which is loaded (More suitable for small experiments who do not want to package their configs)

jamesp-epcc:
I've written some code to load the experiment-specific stuff from an external module: https://github.com/rucio/rucio/pull/2842. It will definitely need more work but I thought it might be a useful starting point for discussion.

mlassnig:
Thanks James, I'm going to have a look.

jamesp-epcc:
I am not sure how best to take this forward. My main concern is how we would switch to the new policy packages without breaking existing installations. We would need packages to be available for all the VOs that need them - I could create the initial packages but I don't think I am the right person to maintain them since I have no connection with most of the VOs. We also need to make sure the right packages get installed so that everything continues to work, including for the tests.
There is also the multi-VO issue. I think we will want each VO to be able to have a different policy package but this would need quite significant changes to the code and I am not sure how best to do this.

bari12:
I think for the beginning we need some kid of dual stack solution. We would keep the current policies we have in the core package (cms, atlas, belle,...) and use them and notify these organisations that we will take the policies out of the core package within one or two feature releases. (Thus 4-8 month time) - We would also show them how to produce their own policy packages. (So we need good documentation for that)
During that time they can either provide a policy package in the config, or still use the files included in the repo. I don't think we need to prepare these for everyone, just show them how to create them and have some best practice recommendation.

The Multi-VO I would leave out of the picture for now. The current multi VO development does not support multiple policies, it should eventually, but I would do this step by step: Let's get the single VO packages working first and then also see with Eli, who will continue the Multi VO development in February.

jamesp-epcc:
That all sounds sensible. I'll work on making my current code into a dual stack version. The pull request for this already contains some documentation about building the policy packages.

PR #2526, opened 2019-05-02
rse: Add_an_LFN2PFN_algorithm_for_LIGO; Fix #2519

### LINKED ISSUE #2519: Add an LFN2PFN algorithm for LIGO
Motivation
----------
LIGO frame files already follow an internal schema which determines their physical filename from the logical filename.  We would like to be able to support this LFN2PFN algorithm in rucio.  To support maximum flexibility, it seems advantageous to do so via an external module.

We propose adding a new method to the `RSEDeterministicTranslation` which supports a plug-able lfn2pfn algorithm.

[ I presume I don't have authority to assign this to anyone but @mlassnig said he'd take a look ]

Modification
------------
See [this commit](https://github.com/astroclark/rucio/commit/75767473f4cb2855259937a36abf0ad4aa2e0964) for a proposed starting point for this change.

The implementation of the algorithm itself can be found [here](https://git.ligo.org/james-clark/gwrucio/blob/master/gwrucio/lfn2pfn.py).   I (James) propose keeping the implementation separate from the upstream rucio repository to give flexibility for handling future exceptions and adaptation to other data types.

This lfn2pfn algorithm is then used by setting the following in the server `rucio.cfg`:
```
[policy]
permission = generic
schema = generic
lfn2pfn_algorithm_default = ligo
```

mlassnig:
rse: Add_an_LFN2PFN_algorithm_for_LIGO; Fix #2519

### DIFF
--- a/lib/rucio/rse/protocols/protocol.py
+++ b/lib/rucio/rse/protocols/protocol.py
@@ -169,7 +169,7 @@ class RSEDeterministicTranslation(object):
         del rse
         del rse_attrs
         del protocol_attrs
-        from ligo_rucio import lfn2pfn as ligo_lfn2pfn
+        from ligo_rucio import lfn2pfn as ligo_lfn2pfn  # pylint: disable=import-error
         return ligo_lfn2pfn.ligo_lab(scope, name, None, None, None)
 
     @classmethod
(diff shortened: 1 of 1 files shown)


bari12:
👍 In the medium term these community specific things should all be loaded in via the policy module of the experiment. See #533 so we do not have lines like `from ligo_rucio import lfn2pfn as ligo_lfn2pfn  # pylint: disable=import-error` in the code.

bari12 (approved):

Issue #2535, opened 2019-05-03
bb8 should not use tape replicas


TomasJavurek:
Motivation
----------
When other replica is not available, bb8 takes the tape one. This is not ideal since we need bb8 to be fast. 

Modification
------------
1) if rse attribute istape=True: bb8-enabled set to False
2) this is considered automatically by background modes
3) needs to be introduced in manual mode

PR #2537, opened 2019-05-06
bb8: excluding tape sources - #2535

TomasJavurek:
bb8: excluding tape sources - #2535


bari12 (approved):

mlassnig (approved):
check that the variable is set everywhere properly. i can see =false set on many ATLAS RSEs, but =true only set on one.

PR #2551, opened 2019-05-08
Core & Internals:  Changes core to exclusively use rse_id. API layer acts as a boundary where rse name is converted to rse_id and vice versa; Fix #2500

### LINKED ISSUE #2500: Replace use of RSE names with ids in core
Motivation
----------
Currently name/id are interchangeable for RSEs, however for a multi-VO setup  it will be important to only use the id in core methods for uniqueness.

Modification
------------
The API directory will be changed so that any rse passed into the core is done using the id. 
The methods in core will be edited to remove dependancy on rse name as inputs.

AndrewLister-STFC:
Core & Internals:  Changes core to exclusively use rse_id. API layer acts as a boundary where rse name is converted to rse_id and vice versa; Fix #2500


lib/rucio/api/account_limit.py
```diff
@@ -115,14 +120,15 @@ def get_account_usage(account, rse, issuer):
     :returns:        List of dicts {'rse_id', 'bytes_used', 'files_used', 'bytes_limit'}
     """
 
-    kwargs = {'account': account, 'rse': rse}
+    rse_id = None
+    print('get_acc_usage')
```
mlassnig: extra print

lib/rucio/api/replica.py
```diff
@@ -118,13 +147,28 @@ def list_replicas(dids, schemes=None, unavailable=False, request_id=None,
...
-                                 resolve_archives=resolve_archives, resolve_parents=resolve_parents)
+    replicas = replica.list_replicas(dids=dids, schemes=schemes, unavailable=unavailable,
+                                     request_id=request_id,
+                                     ignore_availability=ignore_availability,
+                                     all_states=all_states, rse_expression=rse_expression,
+                                     client_location=client_location, domain=domain,
+                                     sign_urls=sign_urls, signature_lifetime=signature_lifetime,
+                                     resolve_archives=resolve_archives, resolve_parents=resolve_parents)
+
+    for rep in replicas:
```
mlassnig, line 158: does this overwrite the field? the existing list_replicas response should not be modified (stuff can be added though). i'm not really following the code below

lib/rucio/core/rule.py
```diff
@@ -1850,7 +1850,7 @@ def update_rules_for_bad_replica(scope, name, rse_id, nowait=False, session=None
     if nlock:
         session.query(models.RSEFileAssociation).filter(models.RSEFileAssociation.scope == scope, models.RSEFileAssociation.name == name, models.RSEFileAssociation.rse_id == rse_id).update({'state': ReplicaState.COPYING})
     else:
-        logging.info('File %s:%s at site %s has no locks. Will be deleted now.', scope, name, rse)
+        logging.info('File %s:%s at site %s has no locks. Will be deleted now.', scope, name, rse_id)
```
mlassnig: for all the logging it would be good to have both RSE name and RSE ID

mlassnig (commented):
the mother of all patches ;-) looks fairly good to me, but a few more pairs of eyes will certainly be useful. testing this will be interesting.

just one question, i see that you've used dicts like `rse["id"]` in many places. would it make sense to just use namedtuples so it can be `rse.id` ? a bit easier on the eyes.

lib/rucio/api/replica.py
AndrewLister-STFC: Yes, "list_replicas" returns a dict with stucture dependant on keys which were the rse name (replicas['rses']['MOCK'], replicas['rses']['MOCK2']...). 
I changed it to rse_id so that if any other part of core needs to rely on it its still ids. 
Its then the job in api to translate those keys.
So it takes the result and replaces these keys by creating a new entry for 'rses' and 'states'. 

Why shouldn't it be modified?

AndrewLister-STFC:
The next patch will be a similar size for the account/scope version of this...

For rse["id"] vs rse.id I tried to just follow the lead of the code that was already there but I agree that named tuples look better.

lib/rucio/api/replica.py
mlassnig: This is the API here, there are clients that rely on the RSE being the RSE name.

lib/rucio/core/rule.py
AndrewLister-STFC: Good point, I can update it to a format like `logging.info('File %s:%s at site %s(%s) has no locks. Will be deleted now.', scope, name, get_rse_name(rse_id=rse_id, session=session), rse_id) ` throughout core

lib/rucio/api/replica.py
AndrewLister-STFC: Do those clients use core? The rse is id in core, then before it returns from api is convereted to name

mlassnig: Ah, no, if the reply of the RSE stays the same that's fine. Be careful though that the lists in here are all sorted, so rewriting them must not change the order.

AndrewLister-STFC: I hadn't spotted that they're sorted. This bit won't change the order but due to changes in the core the rses will be sorted by rse_id. But I thought dicts were unordered anyway?

lib/rucio/core/rule.py
bari12: We could even leave rse name for logging. I mean, in the case of multi-VO, the VO-Name would be encoded in the scope anyway, so it is somewhat clear which RSE is concerned.

AndrewLister-STFC: If we're using the external representation for rse, it'd be cleaner to use the external representation for all scope and accounts too and explicitly stating the vo. In which case just the name is fine here and when VO is added, it will need to be inserted into all the logging.

bari12:
There is one error in the syntax test `lib/rucio/api/did.py:140:1: W293 blank line contains whitespace`

lib/rucio/api/account_limit.py
AndrewLister-STFC: Fixed in new commit

lib/rucio/core/rule.py
AndrewLister-STFC: I've updated the core logging to use rse names instead of ids.

PR #2552, opened 2019-05-08
Core & Internals:  Changes core to exclusively use rse_id. API layer acts as a boundary where rse name is converted to rse_id and vice versa; Fix #2500

### LINKED ISSUE #2500: Replace use of RSE names with ids in core

AndrewLister-STFC:
Core & Internals:  Changes core to exclusively use rse_id. API layer acts as a boundary where rse name is converted to rse_id and vice versa; Fix #2500


AndrewLister-STFC:
Latest commit just updates, discussed patches with Martin are still to come

PR #2555, opened 2019-05-08
Core & Internals: Add support for per RSE checksum support setting; Fix #2498

### LINKED ISSUE #2498: Per-RSE checksum specification
Motivation
----------
Standard gridftp do not support multiple checksum algorithms without the installation of additional plugins.
This caused some hassle when configuring Rucio to operate in a context in which:
- vanilla gridftp offered only `md5`
- StoRM gridftp offered only `adler32`, but required VOMS auth

where provided.

The problem was related to the fact that the Rucio client writes to the DB entry both checksums (`md5` and `adler32`), hence any following operation on the data tried to check both checksums, failing alternatively depending on which gridftp was used.

Possible modification
------------
Several option have been partially discussed:

- Modify the Rucio Client in order to avoid the upload in the DB of unsupported checksums (e.g. adding support for a runtime flag such `--checksums md5,adler32`). This solution is weak and fast to implement, but delegates the handling of the RSE-checksum compatibility to the uploader (being it human or automatic);
- Insert in `rucio.cfg` a global list of supported checksums. This solution is more elegant and the setup is explicit, but making such selection global might be an issue in case of orthogonal compatibilities;
- Add to the DB an RSE field able to record a per-RSE checksum compatibility list. Such information can then be used in the Rucio demons to avoid unsupported checks.

Discussion
------------

This issue is aimed at triggering some discussion prior to the implementation of one of such solutions.

gabrielefronze:
Following issue available at https://github.com/rucio/rucio/issues/2498 I have modified Rucio to allow for the specification of per-RSE checksum compatibility.

This required:
- Implementation of utilities and cosnstants in `common.utils`
- Addition of specific methods to set the RSE attributes related to checksums compatibility in `core.rse`
- Rework of the checksum verification procedure to enable RSE compatibility-aware third party copies in `conveyor.common`
- Rework of the checksum verification procedure in the`upload` method in `rse.rsemanager`
- Extension of `rse.protocols.gfal` protocol to return all the available checksums in the `stats` object

All the modifications were implemented keeping the standard Rucio behaviour, hence compatibility with previous setups should be complete.

PLEASE NOTE: some testing is still undergoing and merging is not yet safe nor recommended.

### DIFF
--- a/lib/rucio/api/rse.py
+++ b/lib/rucio/api/rse.py
@@ -22,7 +22,6 @@
 from rucio.api import permission
 from rucio.common import exception
 from rucio.common.schema import validate_schema
-from rucio.common.utils import api_update_return_dict, CHECKSUM_KEY
 from rucio.core import distance as distance_module
 from rucio.core import rse as rse_module
 from rucio.core.rse_expression_parser import parse_expression
@@ -125,17 +124,6 @@ def del_rse_attribute(rse, key, issuer):
 
     return rse_module.del_rse_attribute(rse_id=rse_id, key=key)
 
-def reset_rse_checksum_support(rse, issuer):
-    """
-    Delete a RSE attribute.
-
-    :param rse: the name of the rse_module.
-
-    :return: True if RSE attribute was deleted successfully, False otherwise.
-    """
-
-    return del_rse_attribute(rse=rse, key=CHECKSUM_KEY, issuer=issuer)
-
 
 def add_rse_attribute(rse, key, value, issuer):
     """ Adds a RSE attribute.
@@ -154,27 +142,6 @@ def add_rse_attribute(rse, key, value, issuer):
         raise exception.AccessDenied('Account %s can not add RSE attributes' % (issuer))
 
     return rse_module.add_rse_attribute(rse_id=rse_id, key=key, value=value)
-
-
-def set_rse_checksum_support(rse, checksum_names, issuer):
-    """ Adds a RSE attribute.
-
-    :param rse: the rse name.
-    :param checksum_names: the checksum name[s] as a single item or a list.
-    :param issuer: The issuer account.
-
-    returns: True if successful, False otherwise.
-    """
-
-    if type(checksum_names) is not list:
-        checksum_names = [checksum_names]
-    supported_checksums = filter(lambda x: is_checksum_valid(x), checksum_names)
-    supported_checksums = map(lambda x: ''.join(x.split()), supported_checksums)
-    if any(supported_checksums):
-        supported_checksums_csv = ','.join(map(str, supported_checksums))
-        return add_rse_attribute(rse=rse, key=CHECKSUM_KEY, value=supported_checksums_csv, issuer=issuer)
-    else:
-        return False
 
 
 def list_rse_attributes(rse):
--- a/lib/rucio/api/rse.py
+++ b/lib/rucio/api/rse.py
@@ -125,6 +125,7 @@ def del_rse_attribute(rse, key, issuer):
 
     return rse_module.del_rse_attribute(rse_id=rse_id, key=key)
 
+
 def del_rse_checksum(rse, checksum_name, issuer):
     """
     Delete a RSE attribute.
@@ -155,6 +156,7 @@ def add_rse_attribute(rse, key, value, issuer):
         raise exception.AccessDenied('Account %s can not add RSE attributes' % (issuer))
 
     return rse_module.add_rse_attribute(rse_id=rse_id, key=key, value=value)
+
 
 def add_rse_checksum(rse, checksum_name, issuer):
     """ Adds a RSE attribute.
--- a/lib/rucio/api/rse.py
+++ b/lib/rucio/api/rse.py
@@ -22,7 +22,7 @@
 from rucio.api import permission
 from rucio.common import exception
 from rucio.common.schema import validate_schema
-from rucio.common.utils import api_update_return_dict, is_checksum_valid, get_checksum_attr_key
+from rucio.common.utils import api_update_return_dict, CHECKSUM_KEY
 from rucio.core import distance as distance_module
 from rucio.core import rse as rse_module
 from rucio.core.rse_expression_parser import parse_expression
@@ -125,18 +125,16 @@ def del_rse_attribute(rse, key, issuer):
 
     return rse_module.del_rse_attribute(rse_id=rse_id, key=key)
 
-
-def del_rse_checksum(rse, checksum_name, issuer):
+def reset_rse_checksum_support(rse, issuer):
     """
     Delete a RSE attribute.
 
     :param rse: the name of the rse_module.
-    :param checksum_name: the checksum name.
 
     :return: True if RSE attribute was deleted successfully, False otherwise.
     """
 
-    return del_rse_attribute(rse=rse, key=get_checksum_attr_key(checksum_name), issuer=issuer)
+    return del_rse_attribute(rse=rse, key=CHECKSUM_KEY, issuer=issuer)
 
 
 def add_rse_attribute(rse, key, value, issuer):
@@ -158,17 +156,22 @@ def add_rse_attribute(rse, key, value, issuer):
     return rse_module.add_rse_attribute(rse_id=rse_id, key=key, value=value)
 
 
-def add_rse_checksum(rse, checksum_name, issuer):
+def set_rse_checksum_support(rse, checksum_names, issuer):
     """ Adds a RSE attribute.
 
     :param rse: the rse name.
-    :param key: the key name.
+    :param checksum_names: the checksum name[s] as a single item or a list.
     :param issuer: The issuer account.
 
     returns: True if successful, False otherwise.
     """
-    if is_checksum_valid(checksum_name):
-        return add_rse_attribute(rse=rse, key=get_checksum_attr_key(checksum_name), value=True, issuer=issuer)
+
+    if type(checksum_names) is not list: checksum_names = [checksum_names]
+    supported_checksums = filter(lambda x: is_checksum_valid(x), checksum_names)
+    supported_checksums = map(lambda x: ''.join(x.split()), supported_checksums)
+    if any(supported_checksums):
+        supported_checksums_csv = ','.join(map(str, supported_checksums))
+        return add_rse_attribute(rse=rse, key=CHECKSUM_KEY, value=supported_checksums_csv, issuer=issuer)
     else:
         return False
 
--- a/lib/rucio/api/rse.py
+++ b/lib/rucio/api/rse.py
@@ -166,7 +166,8 @@ def set_rse_checksum_support(rse, checksum_names, issuer):
     returns: True if successful, False otherwise.
     """
 
-    if type(checksum_names) is not list: checksum_names = [checksum_names]
+    if type(checksum_names) is not list:
+        checksum_names = [checksum_names]
     supported_checksums = filter(lambda x: is_checksum_valid(x), checksum_names)
     supported_checksums = map(lambda x: ''.join(x.split()), supported_checksums)
     if any(supported_checksums):
--- a/lib/rucio/api/rse.py
+++ b/lib/rucio/api/rse.py
@@ -14,6 +14,7 @@
   - Thomas Beermann, <thomas.beermann@cern.ch>, 2014
   - Hannes Hansen, <hannes.jakob.hansen@cern.ch>, 2018
   - Andrew Lister, <andrew.lister@stfc.ac.uk>, 2019
+  - Gabriele Fronze' <gfronze@cern.ch>, 2019
 
   PY3K COMPATIBLE
 '''
@@ -21,7 +22,7 @@
 from rucio.api import permission
 from rucio.common import exception
 from rucio.common.schema import validate_schema
-from rucio.common.utils import api_update_return_dict
+from rucio.common.utils import api_update_return_dict, is_checksum_valid, get_checksum_attr_key
 from rucio.core import distance as distance_module
 from rucio.core import rse as rse_module
 from rucio.core.rse_expression_parser import parse_expression
@@ -30,7 +31,7 @@ from rucio.core.rse_expression_parser import parse_expression
 def add_rse(rse, issuer, deterministic=True, volatile=False, city=None, region_code=None,
             country_name=None, continent=None, time_zone=None, ISP=None,
             staging_area=False, rse_type=None, latitude=None, longitude=None, ASN=None,
-            availability=None):
+            availability=None, supported_checksums=None):
     """
     Creates a new Rucio Storage Element(RSE).
 
@@ -50,6 +51,7 @@ def add_rse(rse, issuer, deterministic=True, volatile=False, city=None, region_c
     :param longitude: Longitude coordinate of RSE.
     :param ASN: Access service network.
     :param availability: Availability.
+    :param supported_checksums: The checksums supported by the RSE.
     """
     validate_schema(name='rse', obj=rse)
     kwargs = {'rse': rse}
@@ -59,7 +61,7 @@ def add_rse(rse, issuer, deterministic=True, volatile=False, city=None, region_c
     return rse_module.add_rse(rse, deterministic=deterministic, volatile=volatile, city=city,
                               region_code=region_code, country_name=country_name, staging_area=staging_area,
                               continent=continent, time_zone=time_zone, ISP=ISP, rse_type=rse_type, latitude=latitude,
-                              longitude=longitude, ASN=ASN, availability=availability)
+                              longitude=longitude, ASN=ASN, availability=availability, supported_checksums=supported_checksums)
 
 
 def get_rse(rse):
@@ -124,6 +126,18 @@ def del_rse_attribute(rse, key, issuer):
 
     return rse_module.del_rse_attribute(rse_id=rse_id, key=key)
 
+def del_rse_checksum(rse, checksum_name, issuer):
+    """
+    Delete a RSE attribute.
+
+    :param rse: the name of the rse_module.
+    :param checksum_name: the checksum name.
+
+    :return: True if RSE attribute was deleted successfully, False otherwise.
+    """
+
+    return del_rse_attribute(rse=rse, key=get_checksum_attr_key(checksum_name), issuer=issuer)
+
 
 def add_rse_attribute(rse, key, value, issuer):
     """ Adds a RSE attribute.
@@ -142,6 +156,20 @@ def add_rse_attribute(rse, key, value, issuer):
         raise exception.AccessDenied('Account %s can not add RSE attributes' % (issuer))
 
     return rse_module.add_rse_attribute(rse_id=rse_id, key=key, value=value)
+
+def add_rse_checksum(rse, checksum_name, issuer):
+    """ Adds a RSE attribute.
+
+    :param rse: the rse name.
+    :param key: the key name.
+    :param issuer: The issuer account.
+
+    returns: True if successful, False otherwise.
+    """
+    if is_checksum_valid(checksum_name):
+        return add_rse_attribute(rse=rse, key=get_checksum_attr_key(checksum_name), value=True, issuer=issuer)
+    else:
+        return False
 
 
 def list_rse_attributes(rse):
--- a/lib/rucio/api/rse.py
+++ b/lib/rucio/api/rse.py
@@ -31,7 +31,7 @@ from rucio.core.rse_expression_parser import parse_expression
 def add_rse(rse, issuer, deterministic=True, volatile=False, city=None, region_code=None,
             country_name=None, continent=None, time_zone=None, ISP=None,
             staging_area=False, rse_type=None, latitude=None, longitude=None, ASN=None,
-            availability=None, supported_checksums=None):
+            availability=None):
     """
     Creates a new Rucio Storage Element(RSE).
 
@@ -51,7 +51,6 @@ def add_rse(rse, issuer, deterministic=True, volatile=False, city=None, region_c
     :param longitude: Longitude coordinate of RSE.
     :param ASN: Access service network.
     :param availability: Availability.
-    :param supported_checksums: The checksums supported by the RSE.
     """
     validate_schema(name='rse', obj=rse)
     kwargs = {'rse': rse}
@@ -61,7 +60,7 @@ def add_rse(rse, issuer, deterministic=True, volatile=False, city=None, region_c
     return rse_module.add_rse(rse, deterministic=deterministic, volatile=volatile, city=city,
                               region_code=region_code, country_name=country_name, staging_area=staging_area,
                               continent=continent, time_zone=time_zone, ISP=ISP, rse_type=rse_type, latitude=latitude,
-                              longitude=longitude, ASN=ASN, availability=availability, supported_checksums=supported_checksums)
+                              longitude=longitude, ASN=ASN, availability=availability)
 
 
 def get_rse(rse):
--- a/lib/rucio/api/rse.py
+++ b/lib/rucio/api/rse.py
@@ -22,6 +22,7 @@
 from rucio.api import permission
 from rucio.common import exception
 from rucio.common.schema import validate_schema
+from rucio.common.utils import api_update_return_dict
 from rucio.core import distance as distance_module
 from rucio.core import rse as rse_module
 from rucio.core.rse_expression_parser import parse_expression
--- a/lib/rucio/client/downloadclient.py
+++ b/lib/rucio/client/downloadclient.py
@@ -42,7 +42,7 @@ from threading import Thread
 from rucio.client.client import Client
 from rucio.common.exception import (InputValidationError, NoFilesDownloaded, NotAllFilesDownloaded, RucioException)
 from rucio.common.pcache import Pcache
-from rucio.common.utils import adler32, md5, detect_client_location, generate_uuid, parse_replicas_from_string, \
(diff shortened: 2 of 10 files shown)


lib/rucio/core/rse.py
```diff
@@ -106,6 +107,15 @@ def add_rse(rse, deterministic=True, volatile=False, city=None, region_code=None
     # Add rse name as a RSE-Tag
     add_rse_attribute(rse=rse, key=rse, value=True, session=session)
 
+    # # Add supported checksums in rse_attr_map if specified
```
mlassnig: forgot to uncomment?

lib/rucio/core/rse.py
```diff
@@ -520,6 +557,43 @@ def get_rse_attribute(key, rse_id=None, value=None, session=None):
...
+    Retrieve RSE attribute value.
+
+    :param rse_id: The RSE id.
+    :param session: The database session in use.
+
+    :returns: The list of checksums supported by the selected RSE.
+    """
+
+    # Collect enabled and disabled checksums
+    checksum_attr_list = list(filter(lambda x: (get_rse_attribute(get_checksum_attr_key(x), rse_id=rse_id, session=session) is not None), GLOBALLY_SUPPORTED_CHECKSUMS))
```
mlassnig: can't see it from the diff, but if this is a core method you need to pass through the session to get_rse_attribute  (also in the call below)

lib/rucio/core/rse.py
```diff
@@ -520,6 +557,43 @@ def get_rse_attribute(key, rse_id=None, value=None, session=None):
...
+
+    # If none either enabled or disabled, then all checksums are supported
+    if (checksum_attr_list is None):
+        return GLOBALLY_SUPPORTED_CHECKSUMS
+    # Otherwise return the list of enabled checksums (key=True)
+    else:
+        return list(filter(lambda x: (get_rse_attribute(get_checksum_attr_key(x), rse_id=rse_id, value=True, session=session) is not None), GLOBALLY_SUPPORTED_CHECKSUMS))
```
mlassnig: session passthrough

lib/rucio/core/rse.py
```diff
@@ -520,6 +557,43 @@ def get_rse_attribute(key, rse_id=None, value=None, session=None):
...
+    Retrieve RSE attribute value.
+
+    :param checksum_name: The desired checksum name for the attribute.
+    :param rse_id: The RSE id.
+    :param session: The database session in use.
+
+    :returns: True if required checksum is supported, False otherwise.
+    """
+    if is_checksum_valid(checksum_name):
+        return (checksum_name in  get_rse_supported_checksums(rse_id=rse_id, session=session))
```
mlassnig: session passthrough

lib/rucio/daemons/conveyor/common.py
```diff
@@ -262,11 +266,15 @@ def bulk_group_transfer(transfers, policy='rule', group_bulk=200, source_strateg
...
                   'request_type': transfer['file_metadata'].get('request_type', None),
                   'activity': str(transfer['file_metadata']['activity'])}
+        
         if verify_checksum != 'none':
-            if 'md5' in list(t_file['metadata'].keys()) and t_file['metadata']['md5']:
-                t_file['checksum'] = 'MD5:%s' % str(t_file['metadata']['md5'])
-            if 'adler32' in list(t_file['metadata'].keys()) and t_file['metadata']['adler32']:
-                t_file['checksum'] = 'ADLER32:%s' % str(t_file['metadata']['adler32'])
+            rse_id = get_rse_id(transfer['rse'], session=session)
+            supported_checksums = get_rse_supported_checksums(rse_id=rse_id, session=sql_session)
```
mlassnig: you don't need to create an explicit session. the @read_session wrapper will do it automatically for you

mlassnig (commented):
i like this. just a few comments on the session handling inline.

lib/rucio/core/rse.py
gabrielefronze: After some discussion with @bari12 we opted for not adding such parameters in the `add_rse` method. This is due to the fact that the new attributes are not (yet) in the `use` main schema, but saved as attributes instead.
Since I did already implement that feature I decided to leave it around commented. I can prune these lines if you prefer!

lib/rucio/core/rse.py
gabrielefronze: Nice point. I see the way that decorator works, but how should I change the arguments of the `get_rse_attribute`? Isn't session=session performing a passthrough? Sorry for the naive question :)

lib/rucio/api/rse.py
```diff
@@ -57,7 +60,7 @@ def add_rse(rse, issuer, deterministic=True, volatile=False, city=None, region_c
     return rse_module.add_rse(rse, deterministic=deterministic, volatile=volatile, city=city,
                               region_code=region_code, country_name=country_name, staging_area=staging_area,
                               continent=continent, time_zone=time_zone, ISP=ISP, rse_type=rse_type, latitude=latitude,
-                              longitude=longitude, ASN=ASN, availability=availability)
+                              longitude=longitude, ASN=ASN, availability=availability, supported_checksums=supported_checksums)
```
bari12: I would not add this in any of the `add_rse` at all. It should just stay a pure optional rse attribute. The `del_rse_checksum`, `add_rse_checksum` etc. are also not necessary, it can all be done via the `add_rse_attribute`. That would be more in line with other "configuration" attributes we have.

lib/rucio/client/rseclient.py
```diff
@@ -86,6 +87,7 @@ def add_rse(self, rse, **kwargs):
         :param longitude: Longitude coordinate of RSE.
         :param ASN: Access service network.
         :param availability: Availability.
+        :param supported_checksums: The checksums supported by the RSE.
```
bari12: See above

lib/rucio/core/rse.py
bari12: Yes please take out 😄

bari12 (changes requested):
👍 Thanks for the work Gabriele! A few change requests. I think it would be better to store the possible checksums in one attribute. E.g. `supported_checksums=md5,adler32` - This can be just a comma separated ordered list. For one this would encode the priority (first try md5, then adler32) and secondly it is easier to see in one grasp which checksums an RSE supports. You don't have to look for multiple attributes.

lib/rucio/core/rse.py
gabrielefronze: Roger that 🤣

lib/rucio/core/rse.py
gabrielefronze: It seems to me I should still add session=None in the args, but then a session global-like variable will be generated and passed to the inner method, right?

lib/rucio/api/rse.py
gabrielefronze: In my opinion `del_rse_checksum ` and `add_rse_checksum` are much needed to use the correct key syntax.
I agree on this peculiar call, this is a left around...

lib/rucio/client/rseclient.py
gabrielefronze: 👍

gabrielefronze (commented):
> 👍 Thanks for the work Gabriele! A few change requests. I think it would be better to store the possible checksums in one attribute. E.g. `supported_checksums=md5,adler32` - This can be just a comma separated ordered list. For one this would encode the priority (first try md5, then adler32) and secondly it is easier to see in one grasp which checksums an RSE supports. You don't have to look for multiple attributes.

Totally agree, I was planning on changing this "attribute setup" today 👍

gabrielefronze:
@bari12 I have implemented the change of the attribute format. Now it is a comma speared list. Let me know if it seems good to you. ;)

.gitignore
```diff
@@ -62,3 +62,16 @@ pylint.out
 # Conveyor test files
 tools/test.file.1G
 
+.vscode/
```
bari12: These seem very specific to your environment 😄

lib/rucio/api/rse.py
```diff
@@ -117,6 +118,17 @@ def del_rse_attribute(rse, key, issuer):
 
     return rse_module.del_rse_attribute(rse=rse, key=key)
 
+def reset_rse_checksum_support(rse, issuer):
```
bari12: Please remove all these. We have many functional attributes like these, and we just set them with add_rse_attribute etc. If we make specific methods to add/delete/update all these, we would completely pollute the API. What is missing though is in the documentation a page where all the different "functional" attributes are explained.

lib/rucio/client/rseclient.py
```diff
@@ -171,6 +171,22 @@ def add_rse_attribute(self, rse, key, value):
             exc_cls, exc_msg = self._get_exception(headers=r.headers, status_code=r.status_code, data=r.content)
             raise exc_cls(exc_msg)
 
+    def add_rse_checksum(self, rse, checksum_name):
```
bari12: Same as above. please remove.

lib/rucio/core/rse.py
```diff
@@ -359,6 +360,27 @@ def add_rse_attribute(rse, key, value, session=None):
     return True
 
 
+@transactional_session
+def set_rse_checksum_support(rse, checksum_names, session=None):
```
bari12: I think this one is also not needed.

lib/rucio/daemons/conveyor/common.py
```diff
@@ -262,11 +265,15 @@ def bulk_group_transfer(transfers, policy='rule', group_bulk=200, source_strateg
                   'selection_strategy': source_strategy if source_strategy else activity_source_strategy.get(str(transfer['file_metadata']['activity']), default_source_strategy),
                   'request_type': transfer['file_metadata'].get('request_type', None),
                   'activity': str(transfer['file_metadata']['activity'])}
+
         if verify_checksum != 'none':
-            if 'md5' in list(t_file['metadata'].keys()) and t_file['metadata']['md5']:
-                t_file['checksum'] = 'MD5:%s' % str(t_file['metadata']['md5'])
-            if 'adler32' in list(t_file['metadata'].keys()) and t_file['metadata']['adler32']:
-                t_file['checksum'] = 'ADLER32:%s' % str(t_file['metadata']['adler32'])
+            rse_id = get_rse_id(transfer['rse'], session=session)
```
bari12: Are you sure the `rse` key is being used here? (I didn't check, but usually this is dest_rse or something like that.

bari12 (changes requested):
Hi @gabrielefronze 
Sorry for the late reply. Some minor change requests. Essentially all these add/delete/update checksum commands are not needed. We update these functional attributes just with add/delete rse attribute. If we would have separate commands for each functional attribute, this would pollute the API too much. Thanks!

.gitignore
gabrielefronze: Yes it is, can you just avoid picking this peculiar commit?

lib/rucio/api/rse.py
gabrielefronze: I see. I just wanted to avoid errors in the key spelling: isn't leaving it free a bit of an issue?
Same for the following :)

lib/rucio/daemons/conveyor/common.py
gabrielefronze: I still have to test it, however I'll get access to a GridFTP and a SRM endpoints by the end of the week, so that I can perform some testing before giving your green light for merging 👍

.gitignore
bari12: Not really, I could mingle this afterwards, but it is easier to just accept a full PR :-)

gabrielefronze: Uhm, I see. A bit of a issue since the editor creates and points at them automatically. I have to figure out how to avoid their creation.

lib/rucio/api/rse.py
bari12: I get the concern, but all these settings are to be set by admins anyway, so the knowledge should be there. You cannot stop them from using `rucio-admin rse set-attribute` anyway, even if the special method is there. And currently, for all the other functional attributes there is no special method, so most admins would not expect one for this as well. So to me this is more about documenting this properly. 
Anyway, I see this more as a "special" setting anyway, because for most RSEs you won't need it, only for very few "special" ones.

gabrielefronze: Roger that then 👍

gabrielefronze:
Hi @mlassnig and @bari12,

I have finalized this PR, keeping the standard behavior in case the `supported_checksum` RSE attribute is not present in the DB, falling back to the standard adler32+md5.
The `PREFERRED_CHECKSUM` value helps keeping `adler32` as preferred choice, but can be easily set to make Rucio prefer another checksum algorithm.
It is even possible to have RSEs with no checksum support. This can be done setting `supported_checksum` to `none`.

As you can see the addition of a new algorithm can be done editing the `utils.py` file only:
- define the `checksum_name(file)` method
- add `checksum_name` as a string to `GLOBALLY_SUPPORTED_CHECKSUMS`
- insert the `checksum_name` to `checksum_name(file)` function mapping in `CHECKSUM_ALGO_DICT`

The only limit is the `gfal` compatibility, which is already wide enough to introduce things such as `crc` and `sha` with out of the box support.

Let me know if you figure we need additional improvements!

gabrielefronze:
Hello,

We just completed the testing of this PR.
Everything works fine and md5-to-md5 trasfers with source and destination checksum verification work as intended.
Since I had to manage the case in which there is no checksum algorithm in common between source and destination, the  behavior of Rucio after this PR will be the following:

 *   Source and destination have at least one compatible checksum algorithm in common -> the checksum is computed and verified at source and at destination. This is the usual Rucio behavior.
 *   Source and destination have no checksum algorithm in common -> the checksum is verified only at destination.

As a bonus I delighted myself adding sha256 and crc32 algorithms. This took around 5 minutes and may work as a demonstrator of how to add new algos. 
They are ready to be used (the additional checksums' values should be added to the DiDs, tho).

lib/rucio/common/utils.py
```diff
@@ -179,10 +180,37 @@ def clean_headers(msg):
     return msg
 
 
+GLOBALLY_SUPPORTED_CHECKSUMS = ['adler32', 'md5', 'sha256', 'crc32']
+CHECKSUM_ALGO_DICT = {}
+PREFERRED_CHECKSUM = GLOBALLY_SUPPORTED_CHECKSUMS[0]
+CHECKSUM_KEY = 'supported_checksums'
```
bari12, line 188: I think it is fine to just hardcode this in the code. While in principle I like the idea of having a table with all the "reserved" keys like this, we don't have it anywhere else.

lib/rucio/client/uploadclient.py
```diff
@@ -208,6 +209,7 @@ def upload(self, items, summary_file_path=None, traces_copy_out=None):
                 lfn['scope'] = file['did_scope']
                 lfn['name'] = file['did_name']
                 lfn['adler32'] = file['adler32']
+                lfn['md5'] = file['md5']
```
bari12: Is this always set?

lib/rucio/core/transfer.py
```diff
@@ -667,6 +667,13 @@ def get_transfer_requests_and_source_replicas(total_workers=0, worker_number=0,
                     else:
                         verify_checksum = 'both'
 
+                logging.info('source RSE checksum compatibility: {}'.format(rse_attrs[source_rse_id][CHECKSUM_KEY]))
+                logging.info('destination RSE checksum compatibility: {}'.format(rse_attrs[dest_rse_id][CHECKSUM_KEY]))
+                common_checksum_names = set(rse_attrs[dest_rse_id][CHECKSUM_KEY]).intersection(rse_attrs[source_rse_id][CHECKSUM_KEY])
+                if len(common_checksum_names) == 0:
+                    logging.info('No common checksum method. Verifying destination only.')
+                    verify_checksum = 'destination'
+
```
bari12: I think this will break production as it assumes that the `supported_checksums` is set in the `rse_attrs`. But it will not be unless we manually add it everywhere (Which we won't, it should be optional). Thus it will fall back everything to `verify_checksum` destination.

lib/rucio/daemons/conveyor/common.py
```diff
@@ -253,8 +256,42 @@ def bulk_group_transfer(transfers, policy='rule', group_bulk=200, source_strateg
...
+        source_rse_id = transfer['file_metadata']['src_rse_id']
+
+        dest_supported_checksums = get_rse_supported_checksums(rse_id=dest_rse_id, session=session)
+        source_supported_checksums = get_rse_supported_checksums(rse_id=source_rse_id, session=session)
+        common_checksum_names = set(source_supported_checksums).intersection(dest_supported_checksums)
+
+        logging.info('source RSE checksum compatibility: {}'.format(source_supported_checksums))
+        logging.info('destination RSE checksum compatibility: {}'.format(dest_supported_checksums))
+
+        if source_supported_checksums == ['none']:
```
bari12, line 271: Why is this a string?

bari12 (commented):
@TWAtGH please check the rsemanager and upload part of this.

lib/rucio/common/utils.py
gabrielefronze: I see the issue there.
For the moment we leave it like that, but we can change it when necessary.
I have also modified line 183: we should stick to standard checksums, even if the commented line is there for future reference.

lib/rucio/client/uploadclient.py
gabrielefronze: Totally right.
Here and below in the same method I modified the dictionary creation to programmatically add all the available checksums.

lib/rucio/core/transfer.py
gabrielefronze: That's right, I forgot I have implemented the `get_rse_supported_checksums` method for such usage in `core/rse.py:541`.
I have modified the logic to use such method that returns the custom attribute if set, otherwise the full set of supported checksums.

lib/rucio/daemons/conveyor/common.py
gabrielefronze: The `'none'` checksum is intended to be a special value of the RSE attribute `supported_checksums` to disable all checksums on a specific RSE.
I think this is necessary to make the PR "complete", but I can remove it.

lib/rucio/client/downloadclient.py
```diff
@@ -571,16 +573,17 @@ def _download_item(self, item, trace, traces_copy_out, log_prefix=''):
...
+                    local_checksum = 'local_checksum'
+
+                    for checksum_name in GLOBALLY_SUPPORTED_CHECKSUMS:
+                        rucio_checksum_temp = item.get(checksum_name)
+                        if rucio_checksum_temp is not None:
+                            rucio_checksum = rucio_checksum_temp
+                            if checksum_name in CHECKSUM_ALGO_DICT:
+                                local_checksum = CHECKSUM_ALGO_DICT[checksum_name](temp_file_path)
+                                if checksum_name == PREFERRED_CHECKSUM:
+                                    break
```
TWAtGH: I suggest we create a method for the verification like this:
```
def _verify_checksum(item, file_path):
    rucio_checksum = item.get(PREFERRED_CHECKSUM)
    checksum_algo = CHECKSUM_ALGO_DICT.get(PREFERRED_CHECKSUM)
    if rucio_checksum and checksum_algo:
        return rucio_checksum == checksum_algo(file_path)

    for checksum_name in GLOBALLY_SUPPORTED_CHECKSUM:
        rucio_checksum = item.get(checksum_name)
        checksum_algo = CHECKSUM_ALGO_DICT.get(checksum_name)
        if rucio_checksum and checksum_algo:
            return rucio_checksum == checksum_algo(file_path)

    return False
```
and then just do `success = self._verify_checksum(item, temp_file_path)`.

TWAtGH (changes requested):

lib/rucio/client/downloadclient.py
gabrielefronze: Sounds good to me, I'll do that by the end of the weekend!

gabrielefronze: Introduced with commit 746dedb0014c548ae244ce51bfe7cda260805c90

lib/rucio/client/downloadclient.py
```diff
@@ -1452,3 +1445,22 @@ def _send_trace(self, trace):
         """
         if self.tracing:
             send_trace(trace, self.client.host, self.client.user_agent)
+
+    def _verify_checksum(self, item, path, rucio_checksum, local_checksum):
```
TWAtGH: I don't think that `rucio_checksum` and `local_checksum` are passed by reference, i.e., the caller will not receive these checksums.

lib/rucio/client/downloadclient.py
```diff
@@ -1452,3 +1445,22 @@ def _send_trace(self, trace):
...
+        checksum_algo = CHECKSUM_ALGO_DICT.get(PREFERRED_CHECKSUM)
+
+        if rucio_checksum and checksum_algo:
+            local_checksum = checksum_algo(path)
+        else:
+            for checksum_name in GLOBALLY_SUPPORTED_CHECKSUMS:
+                rucio_checksum = item.get(checksum_name)
+                checksum_algo = CHECKSUM_ALGO_DICT.get(checksum_name)
+                if rucio_checksum and checksum_algo:
+                    local_checksum = checksum_algo(path)
```
TWAtGH: Without a `return` or `break` here it still calculates every available checksum and only considers the last one instead of just calculating and checking the first one.

TWAtGH (changes requested):
I forgot in the code that I suggested that you'll need the checksums outside the function so I'd suggest to change it to return a tuple instead of just True/Flase, e.g., 
```
def _verify_checksum(item, file_path):
    rucio_checksum = item.get(PREFERRED_CHECKSUM)
    checksum_algo = CHECKSUM_ALGO_DICT.get(PREFERRED_CHECKSUM)
    if rucio_checksum and checksum_algo:
        local_checksum = checksum_algo(file_path)
        return (rucio_checksum == local_checksum, rucio_checksum, local_checksum)

    for checksum_name in GLOBALLY_SUPPORTED_CHECKSUM:
        rucio_checksum = item.get(checksum_name)
        checksum_algo = CHECKSUM_ALGO_DICT.get(checksum_name)
        if rucio_checksum and checksum_algo:
            local_checksum = checksum_algo(file_path)
            return (rucio_checksum == local_checksum, rucio_checksum, local_checksum)

    return (False, rucio_checksum, None)
```

lib/rucio/client/downloadclient.py
gabrielefronze: Yep, just noticed that from the tests. Working on a fix.

lib/rucio/client/downloadclient.py
gabrielefronze: Same as before :)

gabrielefronze: Also: at line 590 I was testing the wrong condition. `If _verify_checksum`, instead of `if not _verify_checksum`... Now fixed!

lib/rucio/client/downloadclient.py
```diff
@@ -1452,3 +1444,22 @@ def _send_trace(self, trace):
...
+        return rucio_checksum == local_checksum, rucio_checksum, local_checksum
+    else:
+        for checksum_name in GLOBALLY_SUPPORTED_CHECKSUMS:
+            rucio_checksum = item.get(checksum_name)
+            checksum_algo = CHECKSUM_ALGO_DICT.get(checksum_name)
+            if rucio_checksum and checksum_algo:
+                local_checksum = checksum_algo(path)
+                break
+
+    return rucio_checksum == local_checksum, rucio_checksum, local_checksum
```
TWAtGH: The `else:` is not necessary.
I would use a `return` instead of a `break` and return the fixed `False` at the very end instead of the comparison of the checksums.
If you really want to use the `break` you need to change the last return to `return rucio_checksum is not None and rucio_checksum == local_checksum, ...`

TWAtGH (changes requested):

lib/rucio/client/downloadclient.py
gabrielefronze: You are right. I wanted to limit the number of returns, but it is widely better to follow your advice.

lib/rucio/client/downloadclient.py
```diff
@@ -1452,3 +1444,22 @@ def _send_trace(self, trace):
...
+        return rucio_checksum == local_checksum, rucio_checksum, local_checksum
+
+    for checksum_name in GLOBALLY_SUPPORTED_CHECKSUMS:
+        rucio_checksum = item.get(checksum_name)
+        checksum_algo = CHECKSUM_ALGO_DICT.get(checksum_name)
+        if rucio_checksum and checksum_algo:
+            local_checksum = checksum_algo(path)
+            return rucio_checksum == local_checksum, rucio_checksum, local_checksum
+
+    return False
```
TWAtGH: The last return should return a tuple to have the function always returning the same type

TWAtGH (changes requested):

bari12:
Please also rebase, since there were some changes in `core/transfer.py`

lib/rucio/client/downloadclient.py
gabrielefronze: Done!

TWAtGH (commented):
Alright thanks a lot! Looks good for me now.

bari12:
`lib/rucio/core/transfer.py:41:1: F401 'rucio.common.utils.CHECKSUM_KEY' imported but unused`
Please fix, otherwise I think we can merge

bari12 (approved):

TWAtGH (approved):

Issue #2436, opened 2019-04-04
Changing sql-alchemy enginge to create PostgreSQL check_constraints instead of ENUMS


bari12:
Motivation
----------
Currently sqlalchemy creates postgres ENUM columns instead of check-constraints (such as in Oracle). It might be good to change this to check-constraints too, to have the same behaviour as in oracle. It needs to be understood if there is any performance impact due to this.

Do we need to change any of the alembic upgrade/downgrade scripts?

PR #2559, opened 2019-05-08
Core: Change PostgreSQL to use check constraints rather than enums #2436

illingwo:
Core: Change PostgreSQL to use check constraints rather than enums #2436

### DIFF
--- a/lib/rucio/db/sqla/migrate_repo/versions/f1b14a8c2ac1_postgres_use_check_constraints.py
+++ b/lib/rucio/db/sqla/migrate_repo/versions/f1b14a8c2ac1_postgres_use_check_constraints.py
@@ -1,0 +1,134 @@
+# Copyright 2013-2019 CERN for the benefit of the ATLAS collaboration.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Authors:
+# - Robert Illingworth <illingwo@fnal.gov>, 2019
+
+''' postgres_use_check_constraints '''
+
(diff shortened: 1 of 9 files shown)


lib/rucio/db/sqla/migrate_repo/versions/f1b14a8c2ac1_postgres_use_check_constraints.py
```diff
@@ -0,0 +1,134 @@
...
+revision = 'f1b14a8c2ac1'
+down_revision = 'b8caac94d7f0'
+
+
+def upgrade():
+    '''
+    Upgrade the database to this revision
+    '''
+
+    if context.get_context().dialect.name != 'postgresql':
```
mlassnig, line 34: i'm a friend of negations as much as anyone else, but wouldn't an indented block with `dialect == "postgresql"` be more descriptive? :-D

illingwo: You can see that it only applies to PostgreSQL right at the beginning without having to scroll to the end to look for an else clause (and originally I thought that the whole thing would be even longer than it turned out).

mlassnig: it's fine with me ;-)

mlassnig (approved):

lib/rucio/db/sqla/migrate_repo/versions/f1b14a8c2ac1_postgres_use_check_constraints.py
bari12: Just to confirm, for every migrate script after this one, we can simply rely on the `create_constraint/drop_constraint` syntax of alembic, correct?

illingwo: I believe so. This migration script should make the constraints consistent, so there will always be something for `drop_constraint` to drop, and the type will be correct for `create_check_constraint`.

illingwo: As a quick test I took the migrate script from #1864, changed the postgresql upgrade to use the  same code block as Oracle with just `drop_constraint/create_check_constraint`, and it applies cleanly.

Issue #2600, opened 2019-05-23
Downloadclient: adding possibility to manipulate external trace objects


TomasJavurek:
Motivation
----------
In WM, the traces are sent by pilot. We still need to populate error msgs on file level there.

Modification
------------
More possibilities, needs some discusssion with downloadclient developer.

TomasJavurek:
The last commit is meant for discussion.
Especially, I am not sure whether it is not better to:
1) place and propagate external_traces directly in the methods of Download client in place of making it parameter of the class. 
2) not to create separate method _update_traces, in place of adding it to _send_trace method. It might be enough just to rename it to _send_or_update_trace()

did is populated directly in the traces, but we still can make it dict: {did:trace} if needed for pilot.

TomasJavurek:
The list is used as an object to which the reference is propagated from pilot to the client. I could propagate directly the trace objects from pilot, but that wouldn't be generic enought in my opinion. I don't see any issues with datasets/bulks/archives, but @TWAtGH might be of different opinion.

PR #2604, opened 2019-05-24
Downloadclient: minimal update, meant only for discusion - #2600

TomasJavurek:
Downloadclient: minimal update, meant only for discusion - #2600


TWAtGH (changes requested):
You can do it like this but:
- DID stored in trace will eventually not match any of the input DIDs. This association is maybe lost if your input contains wildcards, datasets, container. Client extract archive support is questionable. So a proper solution by exceptions/return values would be better.
- I would prefer that `external_traces` is not a member variable.
- I think `external_traces` is a bit misleading. Maybe `traces_copy_out` or something.
- Not sure if `trace` of `_send_trace` is eventually overridden at some point

TomasJavurek:
1) _DID stored in trace will eventually not match any of the input DIDs. ..._
the wildecards and datasets are resolved in methods:
_get_sources()
_prepare_items_for_download()
the trace is sent later in _download_item() method, where filename is populated into trace:
https://github.com/rucio/rucio/blob/master/lib/rucio/client/downloadclient.py#L435
this information can be used for merging traces with the certain file later in pilot where trace is effectively sent. I don't see a problem with that, could you create some example where you think it would fail? @TWAtGH 

Rest of the comments are ok, I already anticipated some:
https://github.com/rucio/rucio/issues/2600

lib/rucio/client/downloadclient.py
```diff
@@ -1416,9 +1428,19 @@ def _check_output(self, output_items):
...
             send_trace(trace, self.client.host, self.client.user_agent)
+
+    def _update_trace(self, traces_copy_out, trace):
+        """
+        If reference to external list of traces given, traces are appended there.
+
+        :param trace: the trace
+        """
+
+        if self.traces_copy_out is not None:
```
TWAtGH: `self.traces_copy_out` shouldn't exist

lib/rucio/client/downloadclient.py
```diff
@@ -575,6 +585,7 @@ def _download_item(self, item, trace, log_prefix=''):
                 if not success:
                     logger.warning('%sDownload attempt failed. Try %s/%s' % (log_prefix, attempt, retries))
                     self._send_trace(trace)
+                    self._update_trace(traces_copy_out, trace)
```
TWAtGH: This would add multiple references to the same trace object. Maybe not a big issue but remember that all traces will only have the highest attempt number.

TWAtGH (changes requested):

lib/rucio/client/downloadclient.py
TomasJavurek: yep. That's what we agreed with @bari12

lib/rucio/client/downloadclient.py
TomasJavurek: sorry, I missed this one.

lib/rucio/client/downloadclient.py
TWAtGH: How about just adding this to the top of download_item:
```
if traces_copy_out is not None:
   traces_copy_out.append(trace)
```

TWAtGH: and removing all the ` self._update_trace(traces_copy_out, trace)` of course

TomasJavurek: I'll think of it. I wanted to implement it in absolutely the same way, as _send_trace.

TWAtGH: But _send_trace works differently :man_shrugging:

TomasJavurek: the intention was, that whenever downloadclient would send the trace, it attaches the trace to the reference list instead (in case that such reference is given).

TomasJavurek: Right, it creates new trace reference with deepcopy in download_worker. But download worker still works with bulk of files (via queues). So effectively we would add always the same reference in download_item. I think we need to make another deepcopy when updating the trace. But this couldn't be done only at the beginning of the download_item, since we are updating the trace later.

TomasJavurek: actualy download_worker does it within the while loop. So no need for deepcopy. So right, the solution to place it just infront of the download_item should work.

lib/rucio/client/downloadclient.py
```diff
@@ -1416,7 +1428,7 @@ def _check_output(self, output_items):
 
     def _send_trace(self, trace):
         """
-        Checks if sending trace is allowed and send the trace.
+        If reference to external list of traces given, traces are appended there.
```
TWAtGH: Deleted the wrong function description? (:

TWAtGH (changes requested):

lib/rucio/client/downloadclient.py
TomasJavurek: That was just a test whether you notice :-D

TWAtGH (approved):

PR #2608, opened 2019-05-27
Conveyor : Filter out requests on blacklisted RSE : Closes #2607

### LINKED ISSUE #2607: Filter out requests on blacklisted RSE
Motivation
----------
If a site is blacklisted but has a huge number of requests in state `Q` it can lead to starvation on other RSEs.

Modification
------------
Exclude RSEs blacklisted for write from the conveyor loop

cserf:
Conveyor : Filter out requests on blacklisted RSE : Closes #2607

### DIFF
--- a/lib/rucio/core/transfer.py
+++ b/lib/rucio/core/transfer.py
@@ -791,11 +791,6 @@ def __list_transfer_requests_and_source_replicas(total_workers=0, worker_number=
     :param session:          Database session to use.
     :returns:                List.
     """
-    nonavailable_rses = session.query(models.RSE.id).filter(models.RSE.deleted == false()).filter(models.RSE.availability.in_((0, 1, 4, 5))).all()
-    rse_clause = []
-    for rse in nonavailable_rses:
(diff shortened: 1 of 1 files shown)


lib/rucio/core/transfer.py
```diff
@@ -791,6 +791,10 @@ def __list_transfer_requests_and_source_replicas(total_workers=0, worker_number=
     :param session:          Database session to use.
     :returns:                List.
     """
+    nonavailable_rses = session.query(models.RSE.id).filter(models.RSE.deleted == false()).filter(models.RSE.availability.in_((0, 1, 4, 5))).all()
+    rse_clause = []
+    for rse in nonavailable_rses:
+        rse_clause.append(models.Request.dest_rse_id != rse[0])
```
bari12: It would be better to do this via a JOIN in the `sub_requests` query below. More performant, and we don't run the risk of too many bind_variables in case there is lots of RSEs/blacklisted RSEs.

bari12 (changes requested):

bari12 (approved):

Issue #2614, opened 2019-05-29
Create username/password authentication for Rucio WebUI


Ruturaj123:
Motivation
----------
Rucio currently supports only x509 authentication. The need is to create a username/password authentication for the Web UI

Modification
------------
Files modified: `web/ui/common/utils.py`, `web/ui/main.py`, `web/ui/static/rucio.js`. 
New files: `web/ui/static/login.js`, `web/ui/templates/login.html`

PR #2615, opened 2019-05-29
WebUI: Create username/password authentication #2614

Ruturaj123:
WebUI: Create username/password authentication #2614

### DIFF
--- a/lib/rucio/web/ui/common/utils.py
+++ b/lib/rucio/web/ui/common/utils.py
@@ -7,11 +7,12 @@
 #
 # Authors:
 # - Thomas Beermann, <thomas.beermann@cern.ch>, 2014-2018
+# - Ruturaj Gujar, <ruturaj.gujar23@gmail.com>, 2019
 
 from json import dumps
 from os.path import dirname, join
 from time import time
-from web import cookies, ctx, input, setcookie, template
+from web import cookies, ctx, input, setcookie, template, seeother
 
 from rucio import version
 from rucio.api import authentication, identity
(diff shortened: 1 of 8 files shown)


tbeerman:
From a first look it seems okay but I still have to test it. But if I got it right it looks that with this only userpass would work since you replaced `check_token` with `log_in`. But we would need this to be configurable, so one could decide when deploying the webui to either use userpass or x509.

Ruturaj123:
Yes, Sir now I am working on it to make it configurable. I will update you on this soon.

lib/rucio/web/ui/common/utils.py
```diff
@@ -152,3 +153,116 @@ def check_token(rendered_tpl):
...
+    render = template.render(join(dirname(__file__), '../templates'))
+
+    if auth_type == 'x509':
+        return check_token(rendered_tpl)
+
+    elif auth_type == 'userpass':
+        if validate_token:
+            return log_in(None, rendered_tpl)
+
+        else:
```
mlassnig: these extra else are not needed, just unindent fallback return

lib/rucio/web/ui/common/utils.py
```diff
@@ -152,3 +153,116 @@ def check_token(rendered_tpl):
...
+
+    if attribs:
+        setcookie('rucio-account-attr', value=dumps(attribs), path='/')
+
+    return seeother('/')
+
+
+def authenticate(rendered_tpl):
+
+    auth_type = config_get('webui', 'auth_type')
```
tbeerman: please put a try/except in case no 'webui' section or 'auth_type' is defined and use 'x509' as default.

lib/rucio/web/ui/common/utils.py
Ruturaj123: Yes, Sir. I will resolve this soon.

lib/rucio/web/ui/common/utils.py
Ruturaj123: Okay Sir :+1:

bari12 (dismissed):
Please rebase this; Also there is still a syntax error
`lib/rucio/web/ui/common/utils.py:271:5: E303 too many blank lines (2)`

Ruturaj123:
I will resolve the syntax error. Should I rebase it to `master` or `next`?

bari12:
Next 😄

tbeerman (approved):

Issue #2266, opened 2019-03-01
Add chunking for delete_replicas dst_replica_condition


bari12:
Motivation
----------
https://github.com/rucio/rucio/blob/10e031551b9292605e47db5ea8729ec359f5b68b/lib/rucio/core/replica.py#L1389
This line needs chunking similar to the other queries.

Modification
------------

Issue #1804, opened 2018-11-22
mode should be an RSE attribute


mlassnig:
Motivation
----------

Right now the configuration of the reaper decides which mode of deletion is used. This causes us to run different types of reapers for no good reason. If the mode is an RSE attribute this is more flexible.

bari12:
This is actually a change which is more relevant to #2444 since the algorithm needs to decide if to include greey RSEs or not.

cserf:
The new reaper relies on the RSE attribute `greedyDeletion`, so I propose to close this ticket.

Issue #1637, opened 2018-10-11
Protection of sources too strict in the reaper


bari12:
Motivation
----------
Currently the reaper is over-protective with the protection of sources if a transfer is active. All possible sources of an active transfer are not eligible for deletion, which is too protective, if there are many possible sources.
This should be changed to:
- If only 1 source is available, protect it --> behaviour stays as it is
- If 2 or more sources are available, all are eligible for deletion, except the one with the alphabetically first RSE (or the one which is oldest) - some shared knowledge so that multiple reapers would not accidentally delete the 2 remaining sources.

This was discussed in the [dev meeting on 2018-10-11](https://indico.cern.ch/event/760955/)

PR #2641, opened 2019-06-07
Reaper 2.0 outer loop: Creation of list of RSES to delete on and list_unlocked_replicas changes : Closes #2444, #1637, #1804, #2266

### LINKED ISSUE #2444: Reaper 2.0 outer loop: Creation of list of RSES to delete on and list_unlocked_replicas changes
Motivation
----------
Creating the list of RSEs based on thresholds, epoche replicas and greedy mode. `list_unlocked_replicas changes` needs to change too for SKIP LOCKED queries.

cserf:
Reaper 2.0 outer loop: Creation of list of RSES to delete on and list_unlocked_replicas changes : Closes #2444, #1637

### DIFF
--- a/bin/rucio-reaper2
+++ b/bin/rucio-reaper2
@@ -1,0 +1,78 @@
+#!/usr/bin/env python
+# Copyright 2014-2019 CERN for the benefit of the ATLAS collaboration.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Authors:
+# - Cedric Serfon, <cedric.serfon@cern.ch>, 2019
+#
+# PY3K COMPATIBLE
+
+'''
+'''
+
+import argparse
+import signal
+
+from rucio.daemons.reaper.reaper2 import run, stop
+
+
+def get_parser():
+    """
+    Returns the argparse parser.
+    """
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--run-once", action="store_true", default=False,
+                        help='Runs one loop iteration')
+    parser.add_argument("--threads", action="store", default=1, type=int,
+                        help='Concurrency control: number of threads')
+    parser.add_argument("--chunk_size", action="store", default=100, type=int,
+                        help='The size used for a bulk deletion on on RSE')
+    parser.add_argument('--sleep-time', action="store", default=60, type=int,
+                        help='Minimum time between 2 consecutive cycles')
+    parser.add_argument('--greedy', action="store_true", default=False,
+                        help='To enable greedy mode. Deprecated, kept for compatibility reasons. Recommended way is to use RSE attribute greedyDeletion=True')
+    parser.add_argument('--exclude-rses', action="store", default=None, type=str,
+                        help='RSE expression to exclude')
+    parser.add_argument('--include-rses', action="store", default=None, type=str,
+                        help='RSE expression to include')
+    parser.add_argument('--rses', nargs='+', type=str,
+                        help='Explicit list of RSEs to include. If empty, it considers all RSEs')
+    parser.add_argument("--delay-seconds", action="store", default=600, type=int,
+                        help='The delay (seconds) to query replicas in BEING_DELETED state.')
+    parser.add_argument('--scheme', action="store", default=None, type=str,
+                        help='Force the reaper to use a particular protocol/scheme, e.g., mock')
+
+    return parser
+
+
+if __name__ == "__main__":
+
+    # Bind our callback to the SIGTERM signal and run the daemon:
+    signal.signal(signal.SIGTERM, stop)
+
+    parser = get_parser()
+    args = parser.parse_args()
+    try:
+        run(threads=args.threads,
+            chunk_size=args.chunk_size,
+            include_rses=args.include_rses,
+            exclude_rses=args.exclude_rses,
+            rses=args.rses,
+            once=args.run_once,
+            greedy=args.greedy,
+            scheme=args.scheme,
+            delay_seconds=args.delay_seconds,
+            sleep_time=args.sleep_time)
+    except KeyboardInterrupt:
+        stop()
--- a/lib/rucio/core/heartbeat.py
+++ b/lib/rucio/core/heartbeat.py
@@ -230,57 +230,6 @@ def list_payload_counts(executable, older_than=600, hash_executable=None, sessio
     return dict(query.all())
 
 
-@read_session
-def get_payload_partition(executable, hostname, pid, thread, payload, older_than=600, hash_executable=None, session=None):
-    """
-    Get the partition associated a a defined payload
-
-    :param executable: Executable name as a string, e.g., conveyor-submitter.
-    :param hostname: Hostname as a string, e.g., rucio-daemon-prod-01.cern.ch.
-    :param pid: UNIX Process ID as a number, e.g., 1234.
-    :param thread: Python Thread Object.
-    :param payload: Payload identifier which can be further used to identify the work a certain thread is executing.
-    :param older_than: Ignore specified heartbeats older than specified nr of seconds.
-    :param hash_executable: Hash of the executable.
-    :param session: The database session in use.
-
-    :returns heartbeats: Dictionary {assign_thread, nr_threads}
-    """
-
-    if not hash_executable:
-        hash_executable = calc_hash(executable)
-
-    # assign payload thread identifier
-    query = session.query(Heartbeats.hostname,
-                          Heartbeats.pid,
-                          Heartbeats.thread_id,
-                          Heartbeats.payload)\
-                   .with_hint(Heartbeats, "index(HEARTBEATS HEARTBEATS_PK)", 'oracle')\
-                   .filter(Heartbeats.executable == hash_executable)\
-                   .filter(Heartbeats.payload == payload)\
-                   .filter(Heartbeats.updated_at >= datetime.datetime.utcnow() - datetime.timedelta(seconds=older_than))\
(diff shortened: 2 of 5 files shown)


bin/rucio-reaper2
```diff
@@ -0,0 +1,78 @@
...
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--run-once", action="store_true", default=False,
+                        help='Runs one loop iteration')
+    parser.add_argument("--threads", action="store", default=1, type=int,
+                        help='Concurrency control: number of threads')
+    parser.add_argument("--chunk_size", action="store", default=100, type=int,
+                        help='The size used for a bulk deletion on on RSE')
+    parser.add_argument('--sleep-time', action="store", default=60, type=int,
+                        help='Minimum time between 2 consecutive cycles')
+    parser.add_argument('--greedy', action="store_true", default=False,
```
bari12: Maybe we should take this out right away?

lib/rucio/core/replica.py
```diff
@@ -1637,6 +1638,98 @@ def list_unlocked_replicas(rse, limit, bytes=None, rse_id=None, worker_number=No
...
+    :delay_seconds:            The delay to query replicas in BEING_DELETED state
+    :param session:            The database session in use.
+
+    :returns: a list of dictionary replica.
+    """
+    replicas_alias = aliased(models.RSEFileAssociation, name='replicas_alias')
+
+    # filter(models.RSEFileAssociation.state != ReplicaState.BEING_DELETED).\
+    none_value = None  # Hack to get pep8 happy...
+    query = session.query(replicas_alias.scope, replicas_alias.name, replicas_alias.path, replicas_alias.bytes, replicas_alias.tombstone, replicas_alias.state, func.count(models.RSEFileAssociation.scope)).\
```
bari12: This query should be with `with_for_update(skip_locked=True)` 
We need to do retrys though if this throws an exception

lib/rucio/core/replica.py
```diff
@@ -1637,6 +1638,98 @@ def list_unlocked_replicas(rse, limit, bytes=None, rse_id=None, worker_number=No
...
+            query = query.filter(text('mod(abs((\'x\'||md5(name))::bit(32)::int), %s) = %s' % (total_workers - 1, worker_number - 1)))
+
+    needed_space = bytes
+    total_bytes, total_files = 0, 0
+    rows = []
+    for (scope, name, path, bytes, tombstone, state, cnt) in query.yield_per(1000):
+        # Check if more than one replica is available
+        if cnt > 1:
+            if state != ReplicaState.UNAVAILABLE:
+
```
bari12: Remove empty line

lib/rucio/core/replica.py
```diff
@@ -1637,6 +1638,98 @@ def list_unlocked_replicas(rse, limit, bytes=None, rse_id=None, worker_number=No
...
+                         'bytes': bytes, 'tombstone': tombstone,
+                         'state': state})
+            session.query(models.RSEFileAssociation).filter(models.RSEFileAssociation == scope,
+                                                            models.RSEFileAssociation.name == name,
+                                                            models.RSEFileAssociation.rse_id == rse_id).\
+                update({'updated_at': datetime.utcnow(), 'state': ReplicaState.BEING_DELETED}, synchronize_session=False)
+
+        else:
+            # If this is the last replica, check if there are some requests
+            request_cnt = session.query(func.count()).\
```
bari12, line 1709: I find this rather heavy as a query since it is done for every replica in the loop. Wouldn't it be better to do this in the general query above?

lib/rucio/daemons/reaper/reaper2.py
```diff
@@ -0,0 +1,359 @@
...
+                    tot_needed_free_space += 1000000000000
+                else:
+                    if needed_free_space:
+                        dict_rses[(rse['rse'], rse['id'])] = [needed_free_space, max_being_deleted_files]
+                        tot_needed_free_space += needed_free_space
+                    else:
+                        logging.debug('%s Nothing to delete on %s', prepend_str, rse['rse'])
+
+            # Ordering the RSEs based on the needed free space
+            sorted_dict_rses = OrderedDict(sorted(dict_rses.items(), key=itemgetter(1), reverse=True))
```
bari12, line 332: I think this should be done differently. The list should be completely random, but weighed.
Thus the ones which need a lot of `needed_free_space` (relatively) should get more repetitions in the loop at line 235. But it should not be ordered by this weigh, otherwise all reapers will try to do the same RSEs at startup.

lib/rucio/core/heartbeat.py
```diff
@@ -223,6 +230,57 @@ def list_payload_counts(executable, older_than=600, hash_executable=None, sessio
     return dict(query.all())
 
 
+@read_session
+def get_payload_partition(executable, hostname, pid, thread, payload, older_than=600, hash_executable=None, session=None):
```
bari12: I don't think this is needed; I would propose to do the `list_and_mark_unlocked_replicas` completely without heartbeats, as it does not have any added benefit. And in that case, we do not need to do payload_partitions.

lib/rucio/daemons/reaper/reaper2.py
```diff
@@ -0,0 +1,359 @@
...
+                logging.debug('%s Payload count : %s', prepend_str, str(payload_cnt))
+                rse_hostname, rse_info = rses_hostname_mapping[rse_name]
+                rse_hostname_key = '%s,%s' % (rse_name, rse_hostname)
+
+                tot_threads_for_hostname = 0
+                for key in payload_cnt:
+                    if key and key.find(',') > -1 and key.split(',')[1] == rse_hostname:
+                        tot_threads_for_hostname += payload_cnt[key]
+
+                payload_threads = get_payload_partition(executable, hostname, pid, hb_thread, payload=rse_hostname_key, older_than=600, hash_executable=None, session=None)
```
bari12: See above, I think we can do without the payload partitions.

lib/rucio/daemons/reaper/reaper2.py
```diff
@@ -0,0 +1,359 @@
...
+
+                if rse_hostname_key in payload_cnt and (tot_threads_for_hostname >= max_deletion_thread or payload_threads['nr_threads'] >= max_workers):
+                    logging.debug('%s Too many deletion threads for %s on RSE %s. Back off', prepend_str, rse_hostname, rse_name)
+                    # Might need to reschedule a try on this RSE later in the same cycle
+                    continue
+                del_start_time = time.time()
+                logging.info('%s Starting new worker on RSE %s', prepend_str, rse_name)
+                live(executable, hostname, pid, hb_thread, older_than=600, hash_executable=None, payload=rse_hostname_key, session=None)
```
bari12, line 374: We should probably right away use rse_ids in the payload, or at least rsename,rse_id tuples

lib/rucio/daemons/reaper/reaper2.py
```diff
@@ -0,0 +1,359 @@
...
+            logging.debug('%s List of RSEs to process ordered by needed space desc : %s', prepend_str, str(sorted_dict_rses))
+
+            # Get the mapping between the RSE and the hostname used for deletion. The dictionary has RSE as key and (hostanme, rse_info) as value
+            rses_hostname_mapping = get_rses_to_hostname_mapping()
+            # logging.debug('%s Mapping RSEs to hostnames used for deletion : %s', prepend_str, str(rses_hostname_mapping))
+
+            # Loop over the RSEs. rse_key = (rse, rse_id)
+            for rse_key in sorted_dict_rses:
+                rse_name, rse_id = rse_key
+                max_workers = ceil(sorted_dict_rses[rse_key][0] / tot_needed_free_space * heart_beat['nr_threads'] * 10)
```
bari12: I don't quite get why we need this?

lib/rucio/core/replica.py
```diff
@@ -1637,6 +1638,98 @@ def list_unlocked_replicas(rse, limit, bytes=None, rse_id=None, worker_number=No
...
+        filter(or_(replicas_alias.state.in_((ReplicaState.AVAILABLE, ReplicaState.UNAVAILABLE, ReplicaState.BAD)),
+                   and_(replicas_alias.state == ReplicaState.BEING_DELETED, replicas_alias.updated_at < datetime.utcnow() - timedelta(seconds=delay_seconds)))).\
+        join(models.RSEFileAssociation, and_(models.RSEFileAssociation.scope == replicas_alias.scope,
+                                             models.RSEFileAssociation.name == replicas_alias.name,
+                                             models.RSEFileAssociation.rse_id != replicas_alias.rse_id,
+                                             models.RSEFileAssociation.state == ReplicaState.AVAILABLE)).\
+        group_by(replicas_alias.scope, replicas_alias.name, replicas_alias.path, replicas_alias.bytes, replicas_alias.tombstone, replicas_alias.state).\
+        order_by(replicas_alias.tombstone)
+
+    if worker_number and total_workers and total_workers - 1 > 0:
```
bari12: As described above, I think we can skip this.

bari12 (changes requested):

lib/rucio/core/replica.py
```diff
@@ -1637,6 +1638,99 @@ def list_unlocked_replicas(rse, limit, bytes=None, rse_id=None, worker_number=No
     return rows
 
 
+@transactional_session
+def list_and_mark_unlocked_replicas(limit, bytes=None, rse_id=None, delay_seconds=0, session=None):
```
bari12: I would add a higher default for the `delay_seconds` here; Maybe 10m?

lib/rucio/core/replica.py
bari12: Alternatively we can do this in a bulk step too. So for all the ones with `cnt==0` you do a bulked query.

lib/rucio/core/replica.py
```diff
@@ -1637,6 +1638,99 @@ def list_unlocked_replicas(rse, limit, bytes=None, rse_id=None, worker_number=No
...
+                    break
+
+                total_files += 1
+                if total_files > limit:
+                    break
+
+            rows.append({'scope': scope, 'name': name, 'path': path,
+                         'bytes': bytes, 'tombstone': tombstone,
+                         'state': state})
+            session.query(models.RSEFileAssociation).filter(models.RSEFileAssociation == scope,
```
bari12: I think we should consider to do this one, and the one in the `else` condition in a bulk update too

bari12:
Merging this prototype now for testing; We will have a final full review of the code at a later point;

bari12 (approved):

PR #2643, opened 2019-06-10
Clients: Don't disable file registration when uploading to non-deterministic RSE; Fix #2620

### LINKED ISSUE #2620: Upload fails to register file if pfn provided and non-deterministic RSE
Motivation
----------

If the RSE is non-deterministic and a pfn is provided then the client prints `Upload with given pfn implies that no_register is True, except non-deterministic RSEs` and fails to register the replica.

illingwo:
Clients: Don't disable file registration when uploading to non-deterministic RSE; Fix #2620

### DIFF
--- a/lib/rucio/client/uploadclient.py
+++ b/lib/rucio/client/uploadclient.py
@@ -156,7 +156,7 @@ class UploadClient:
             if not is_deterministic and not pfn:
                 logger.error('PFN has to be defined for NON-DETERMINISTIC RSE.')
                 continue
-            if pfn and not is_deterministic:
+            if pfn and is_deterministic:
                 logger.warning('Upload with given pfn implies that no_register is True, except non-deterministic RSEs')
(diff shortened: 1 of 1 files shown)


lib/rucio/client/uploadclient.py
```diff
@@ -166,7 +166,7 @@ def upload(self, items, summary_file_path=None):
             # if register_after_upload, file should be overwritten if it is not registered
             # otherwise if file already exists on RSE we're done
             if register_after_upload:
-                if rsemgr.exists(rse_settings, file_did):
+                if rsemgr.exists(rse_settings, pfn if pfn else file_did):
```
TomasJavurek, line 169: Why this change? pfn is preferred, otherwise rsemgr needs to extract pfn itself.

bari12:
Implemented in @TomasJavurek PR

PR #2644, opened 2019-06-10
Clients: Don't disable file registration when uploading to non-deterministic RSE; Fix #2620

### LINKED ISSUE #2620: Upload fails to register file if pfn provided and non-deterministic RSE

illingwo:
Clients: Don't disable file registration when uploading to non-deterministic RSE; Fix #2620

### DIFF
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -1,4 +1,4 @@
-.. Rucio documentation master file, created by
+g.. Rucio documentation master file, created by
    sphinx-quickstart on Fri Oct 27 14:25:40 2017.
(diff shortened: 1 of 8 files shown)


TomasJavurek:
The local branch would need to be rebased and pr submitted again due to travis issues that we had last days. I impelemented Robert's updated to another pr: 
https://github.com/TomasJavurek/rucio/commit/fb39f8073b2e9d3bdc157066d55dca6713ddf61b

PR #2645, opened 2019-06-12
clients: bug fix, non-deter - #2620

TomasJavurek:
clients: bug fix, non-deter - #2620


TomasJavurek:
merged from next, closing.

PR #2647, opened 2019-06-13
Core & Internals: updated RSE export structure; Fix #2237

### LINKED ISSUE #2237: update rse export format
Motivation
----------
update the json format of the RSE export in core/export.py

Modification
------------
- drop key 'ASN'
- drop key 'ISP'
- drop key 'continent'
- drop key 'transfer_limits'
- drop key 'limits'
- remove outer protocol key
- move key 'credentials' one level down to protocols
- move key 'lfn2pfn' one level up

hahahannes:
Core & Internals: updated RSE export structure; Fix #2237

### DIFF
--- a/lib/rucio/core/importer.py
+++ b/lib/rucio/core/importer.py
@@ -13,12 +13,113 @@
 # limitations under the License.
 #
 # Authors:
-# - Hannes Hansen <hannes.jakob.hansen@cern.ch>, 2018
+# - Hannes Hansen <hannes.jakob.hansen@cern.ch>, 2018-2019
 #
 # PY3K COMPATIBLE
 
+from six import string_types
+from rucio.common.exception import RSEOperationNotSupported
 from rucio.core import rse as rse_module, distance as distance_module
+from rucio.db.sqla.constants import RSEType
 from rucio.db.sqla.session import transactional_session
+
+
+@transactional_session
+def import_rses(rses, session=None):
+    new_rses = []
+    for rse_name in rses:
+        new_rses.append(rse_name)
+        rse = rses[rse_name]
+        if isinstance(rse.get('rse_type'), string_types):
+            rse['rse_type'] = RSEType.from_string(str(rse['rse_type']))
+        if not rse_module.rse_exists(rse_name, session=session):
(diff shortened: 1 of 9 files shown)


lib/rucio/core/rse.py
```diff
@@ -487,19 +492,22 @@ def get_rses_with_attribute_value(key, value, lookup_key, session=None):
 
 
 @read_session
-def get_rse_attribute(key, rse_id=None, value=None, session=None):
+def get_rse_attribute(key, rse_id=None, value=None, use_cache=True, session=None):
```
mlassnig, line 495: i don't quite understand why you want to make this optional. shouldn't the cache be used always?

hahahannes: I added it because I want to use `get_rse_attributes` in the tests (test_import_export.py) and the cache is also running on travis. I can also replace it there with a normal db query.

mlassnig: hmmmm, it's a core function so it should be fine, as long as it's not in the api ;-)

lib/rucio/core/importer.py
```diff
@@ -13,14 +13,115 @@
...
 
+from six import string_types
+from rucio.common.exception import RSEOperationNotSupported
 from rucio.core import rse as rse_module, distance as distance_module
+from rucio.db.sqla.constants import RSEType
 from rucio.db.sqla.session import transactional_session
 
 
+@transactional_session
+def import_rses(rses, session=None):
```
bari12, line 28: Method doc missing

lib/rucio/core/importer.py
```diff
@@ -13,14 +13,115 @@
...
+    for old_rse in old_rses:
+        if old_rse not in new_rses:
+            try:
+                rse_module.del_rse(rse=old_rse, session=session)
+            except RSEOperationNotSupported:
+                pass
+
+
+@transactional_session
+def import_distances(distances, session=None):
```
bari12, line 105: Method doc missing

bari12:
I am merging this now so we can use it for testing on the various instances and get some experience. Some things might still need minor modifications though.

bari12 (approved):

PR #2689, opened 2019-06-22
WebUI: Add new identity type (SAML) in core

Ruturaj123:
WebUI: Add new identity type (SAML) in core

### DIFF
--- a/bin/rucio
+++ b/bin/rucio
@@ -32,6 +32,7 @@
 # - Frank Berghaus, <frank.berghaus@cern.ch>, 2018
 # - Hannes Hansen <hannes.jakob.hansen@cern.ch>, 2018-2019
 # - Dimitrios Christidis <dimitrios.christidis@cern.ch>, 2018
+# - Ruturaj Gujar, <ruturaj.gujar23@gmail.com>, 2019
 #
 # PY3K COMPATIBLE
 
@@ -241,7 +242,7 @@ def get_client(args):
     """
     Returns a new client object.
     """
-    if args.auth_strategy == 'userpass':
+    if args.auth_strategy == 'userpass' or args.auth_strategy == 'saml':
         creds = {'username': args.username, 'password': args.password}
     else:
         creds = None
--- a/bin/rucio
+++ b/bin/rucio
@@ -242,7 +242,7 @@ def get_client(args):
     """
     Returns a new client object.
     """
-    if args.auth_strategy == 'userpass' or args.auth_strategy == 'saml':
+    if args.auth_strategy in ['userpass', 'saml']:
         creds = {'username': args.username, 'password': args.password}
     else:
         creds = None
--- a/lib/rucio/api/authentication.py
+++ b/lib/rucio/api/authentication.py
@@ -17,6 +17,7 @@
 # - Vincent Garonne <vgaronne@gmail.com>, 2012-2015
 # - Martin Barisits <martin.barisits@cern.ch>, 2017
 # - Andrew Lister, <andrew.lister@stfc.ac.uk>, 2019
+# - Ruturaj Gujar <ruturaj.gujar23@gmail.com>, 2019
 #
 # PY3K COMPATIBLE
 
@@ -159,6 +160,7 @@ def get_auth_token_saml(account, saml_nameid, appid, ip=None):
         raise exception.AccessDenied('User with identity %s can not log to account %s' % (saml_nameid, account))
 
     return authentication.get_auth_token_saml(account, saml_nameid, appid, ip)
+
 
 def validate_auth_token(token):
     """
--- a/lib/rucio/api/authentication.py
+++ b/lib/rucio/api/authentication.py
@@ -141,6 +141,25 @@ def get_ssh_challenge_token(account, appid, ip=None):
     return authentication.get_ssh_challenge_token(account, appid, ip)
 
 
+def get_auth_token_saml(account, saml_nameid, appid, ip=None):
+    """
+    Authenticate a Rucio account temporarily via SSO.
+
+    The token lifetime is 1 hour.
+
+    :param account: Account identifier as a string.
+    :param saml_nameid: NameId returned in SAML response as a string.
+    :param appid: The application identifier as a string.
+    :param ip: IP address of the client as a string.
+    :returns: Authentication token as a variable-length string.
+    """
+
+    kwargs = {'account': account, 'saml_nameid': saml_nameid}
+    if not permission.has_permission(issuer=account, action='get_auth_token_saml', kwargs=kwargs):
+        raise exception.AccessDenied('User with identity %s can not log to account %s' % (saml_nameid, account))
+
+    return authentication.get_auth_token_saml(account, saml_nameid, appid, ip)
+
 def validate_auth_token(token):
     """
     Validate an authentication token.
--- a/lib/rucio/api/authentication.py
+++ b/lib/rucio/api/authentication.py
@@ -159,6 +159,8 @@ def get_auth_token_saml(account, saml_nameid, appid, ip=None):
     if not permission.has_permission(issuer=account, action='get_auth_token_saml', kwargs=kwargs):
         raise exception.AccessDenied('User with identity %s can not log to account %s' % (saml_nameid, account))
 
+    account = InternalAccount(account)
+
     return authentication.get_auth_token_saml(account, saml_nameid, appid, ip)
 
 
--- a/lib/rucio/client/baseclient.py
+++ b/lib/rucio/client/baseclient.py
@@ -560,7 +560,7 @@ class BaseClient(object):
 
                 SAML_auth_url = SAML_auth_result.headers['X-Rucio-SAML-Auth-URL']
                 result = self.session.post(SAML_auth_url, data=userpass, verify=False, allow_redirects=True)
-                result = self.get(url, headers=headers)
+                result = self.session.get(url, headers=headers)
                 break
             except ConnectionError as error:
                 LOG.warning('ConnectionError: ' + str(error))
@@ -608,7 +608,7 @@ class BaseClient(object):
             elif self.auth_type == 'saml':
                 if not self.__get_token_saml():
                     raise CannotAuthenticate('saml authentication failed for account=%s with identity=%s' % (self.account,
-                                                                                                            self.creds))
+                                                                                                             self.creds))
             else:
                 raise CannotAuthenticate('auth type \'%s\' not supported' % self.auth_type)
 
--- a/lib/rucio/client/baseclient.py
+++ b/lib/rucio/client/baseclient.py
@@ -23,6 +23,7 @@
 # - Martin Barisits <martin.barisits@cern.ch>, 2016-2018
 # - Tobias Wegner <twegner@cern.ch>, 2017
 # - Brian Bockelman <bbockelm@cse.unl.edu>, 2017-2018
+# - Ruturaj Gujar, <ruturaj.gujar23@gmail.com>, 2019
 #
 # PY3K COMPATIBLE
 
@@ -152,8 +153,8 @@ class BaseClient(object):
         if auth_type is None:
             LOG.debug('no auth_type passed. Trying to get it from the environment variable RUCIO_AUTH_TYPE and config file.')
             if 'RUCIO_AUTH_TYPE' in environ:
-                if environ['RUCIO_AUTH_TYPE'] not in ('userpass', 'x509', 'x509_proxy', 'gss', 'ssh'):
-                    raise MissingClientParameter('Possible RUCIO_AUTH_TYPE values: userpass, x509, x509_proxy, gss, ssh, vs. ' + environ['RUCIO_AUTH_TYPE'])
+                if environ['RUCIO_AUTH_TYPE'] not in ('userpass', 'x509', 'x509_proxy', 'gss', 'ssh', 'saml'):
+                    raise MissingClientParameter('Possible RUCIO_AUTH_TYPE values: userpass, x509, x509_proxy, gss, ssh, saml, vs. ' + environ['RUCIO_AUTH_TYPE'])
                 self.auth_type = environ['RUCIO_AUTH_TYPE']
             else:
                 try:
@@ -165,7 +166,7 @@ class BaseClient(object):
             LOG.debug('no creds passed. Trying to get it from the config file.')
             self.creds = {}
             try:
-                if self.auth_type == 'userpass':
+                if self.auth_type == 'userpass' or self.auth_type == 'saml':
                     self.creds['username'] = config_get('client', 'username')
                     self.creds['password'] = config_get('client', 'password')
                 elif self.auth_type == 'x509':
@@ -538,6 +539,49 @@ class BaseClient(object):
         LOG.debug('got new token')
         return True
 
+    def __get_token_saml(self):
+        """
+        Sends a request to get an auth token from the server and stores it as a class attribute. Uses saml authentication.
+
+        :returns: True if the token was successfully received. False otherwise.
+        """
+
+        headers = {'X-Rucio-Account': self.account}
+        userpass = {'username': self.creds['username'], 'password': self.creds['password']}
+        url = build_url(self.auth_host, path='auth/sso')
+
+        result = None
+        for retry in range(self.AUTH_RETRIES + 1):
+            try:
+                SAML_auth_result = self.session.get(url, headers=headers)
+
+                if SAML_auth_result.headers['X-Rucio-Auth-Token']:
+                    return SAML_auth_result.headers['X-Rucio-Auth-Token']
+
+                SAML_auth_url = SAML_auth_result.headers['X-Rucio-SAML-Auth-URL']
+                result = self.session.post(SAML_auth_url, data=userpass, verify=False, allow_redirects=True)
+                result = self.get(url, headers=headers)
+                break
+            except ConnectionError as error:
+                LOG.warning('ConnectionError: ' + str(error))
+                self.ca_cert = False
+                if retry > self.request_retries:
+                    raise
+
+        if not result or 'result' not in locals():
+            LOG.error('cannot get auth_token')
+            return False
+
+        if result.status_code != codes.ok:  # pylint: disable-msg=E1101
+            exc_cls, exc_msg = self._get_exception(headers=result.headers,
+                                                   status_code=result.status_code,
+                                                   data=result.content)
+            raise exc_cls(exc_msg)
+
+        self.auth_token = result.headers['X-Rucio-Auth-Token']
+        LOG.debug('got new token')
+        return True
+
     def __get_token(self):
         """
         Calls the corresponding method to receive an auth token depending on the auth type. To be used if a 401 - Unauthorized error is received.
@@ -560,6 +604,10 @@ class BaseClient(object):
             elif self.auth_type == 'ssh':
                 if not self.__get_token_ssh():
                     raise CannotAuthenticate('ssh authentication failed for account=%s with identity=%s' % (self.account,
+                                                                                                            self.creds))
+            elif self.auth_type == 'saml':
+                if not self.__get_token_saml():
+                    raise CannotAuthenticate('saml authentication failed for account=%s with identity=%s' % (self.account,
                                                                                                             self.creds))
             else:
                 raise CannotAuthenticate('auth type \'%s\' not supported' % self.auth_type)
@@ -639,6 +687,9 @@ class BaseClient(object):
                 raise NoAuthInformation('The SSH private key has to be defined')
         elif self.auth_type == 'gss':
             pass
+        elif self.auth_type == 'saml':
+            if self.creds['username'] is None or self.creds['password'] is None:
+                raise NoAuthInformation('No SAML username or password passed')
         else:
             raise CannotAuthenticate('auth type \'%s\' not supported' % self.auth_type)
 
--- a/lib/rucio/client/baseclient.py
+++ b/lib/rucio/client/baseclient.py
@@ -153,7 +153,7 @@ class BaseClient(object):
         if auth_type is None:
             LOG.debug('no auth_type passed. Trying to get it from the environment variable RUCIO_AUTH_TYPE and config file.')
             if 'RUCIO_AUTH_TYPE' in environ:
-                if environ['RUCIO_AUTH_TYPE'] not in ('userpass', 'x509', 'x509_proxy', 'gss', 'ssh', 'saml'):
+                if environ['RUCIO_AUTH_TYPE'] not in ['userpass', 'x509', 'x509_proxy', 'gss', 'ssh', 'saml']:
                     raise MissingClientParameter('Possible RUCIO_AUTH_TYPE values: userpass, x509, x509_proxy, gss, ssh, saml, vs. ' + environ['RUCIO_AUTH_TYPE'])
                 self.auth_type = environ['RUCIO_AUTH_TYPE']
             else:
@@ -166,7 +166,7 @@ class BaseClient(object):
             LOG.debug('no creds passed. Trying to get it from the config file.')
             self.creds = {}
             try:
-                if self.auth_type == 'userpass' or self.auth_type == 'saml':
+                if self.auth_type in ['userpass', 'saml']:
                     self.creds['username'] = config_get('client', 'username')
                     self.creds['password'] = config_get('client', 'password')
                 elif self.auth_type == 'x509':
@@ -548,7 +548,7 @@ class BaseClient(object):
 
         headers = {'X-Rucio-Account': self.account}
         userpass = {'username': self.creds['username'], 'password': self.creds['password']}
-        url = build_url(self.auth_host, path='auth/sso')
+        url = build_url(self.auth_host, path='auth/saml')
 
         result = None
         for retry in range(self.AUTH_RETRIES + 1):
--- a/lib/rucio/core/authentication.py
+++ b/lib/rucio/core/authentication.py
@@ -323,7 +323,7 @@ def get_auth_token_saml(account, saml_nameid, appid, ip=None, session=None):
 
     # remove expired tokens
     session.query(models.Token).filter(models.Token.expired_at < datetime.datetime.utcnow(),
-                                       models.Token.account == account).delete()
+                                       models.Token.account == account).with_for_update(skip_locked=True).delete()
 
     tuid = generate_uuid()  # NOQA
     token = '%(account)s-%(saml_nameid)s-%(appid)s-%(tuid)s' % locals()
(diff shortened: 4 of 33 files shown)


bari12:
Hi Ruturaj, there were some errors showing up in the travis test; Please fix 😄 

```
lib/rucio/web/ui/common/utils.py:96:57: E128 continuation line under-indented for visual indent
lib/rucio/web/ui/common/utils.py:97:57: E128 continuation line under-indented for visual indent
lib/rucio/web/ui/common/utils.py:98:57: E128 continuation line under-indented for visual indent
lib/rucio/web/ui/common/utils.py:128:57: E128 continuation line under-indented for visual indent
lib/rucio/web/ui/common/utils.py:129:57: E128 continuation line under-indented for visual indent
lib/rucio/web/ui/common/utils.py:130:57: E128 continuation line under-indented for visual indent
lib/rucio/api/authentication.py:150:1: E302 expected 2 blank lines, found 1
lib/rucio/core/authentication.py:310:1: W293 blank line contains whitespace
```

Ruturaj123:
Yes Sir I have fixed the errors locally. Actually I have been getting some errors in the webui so I will push the changes once those errors have been fixed.

bin/rucio
```diff
@@ -241,7 +242,7 @@ def get_client(args):
     """
     Returns a new client object.
     """
-    if args.auth_strategy == 'userpass':
+    if args.auth_strategy == 'userpass' or args.auth_strategy == 'saml':
```
mlassnig: `if args.auth_strategy in ['userpass', 'saml']:`

lib/rucio/client/baseclient.py
```diff
@@ -152,8 +153,8 @@ def __init__(self, rucio_host=None, auth_host=None, account=None, ca_cert=None,
         if auth_type is None:
             LOG.debug('no auth_type passed. Trying to get it from the environment variable RUCIO_AUTH_TYPE and config file.')
             if 'RUCIO_AUTH_TYPE' in environ:
-                if environ['RUCIO_AUTH_TYPE'] not in ('userpass', 'x509', 'x509_proxy', 'gss', 'ssh'):
-                    raise MissingClientParameter('Possible RUCIO_AUTH_TYPE values: userpass, x509, x509_proxy, gss, ssh, vs. ' + environ['RUCIO_AUTH_TYPE'])
+                if environ['RUCIO_AUTH_TYPE'] not in ('userpass', 'x509', 'x509_proxy', 'gss', 'ssh', 'saml'):
```
mlassnig: should be list, not tuple (original was wrong already)

lib/rucio/client/baseclient.py
```diff
@@ -165,7 +166,7 @@ def __init__(self, rucio_host=None, auth_host=None, account=None, ca_cert=None,
             LOG.debug('no creds passed. Trying to get it from the config file.')
             self.creds = {}
             try:
-                if self.auth_type == 'userpass':
+                if self.auth_type == 'userpass' or self.auth_type == 'saml':
```
mlassnig: `if self.auth_type in ['userpass', 'saml']:`

lib/rucio/client/baseclient.py
```diff
@@ -538,6 +539,49 @@ def __get_token_gss(self):
...
+    def __get_token_saml(self):
+        """
+        Sends a request to get an auth token from the server and stores it as a class attribute. Uses saml authentication.
+
+        :returns: True if the token was successfully received. False otherwise.
+        """
+
+        headers = {'X-Rucio-Account': self.account}
+        userpass = {'username': self.creds['username'], 'password': self.creds['password']}
+        url = build_url(self.auth_host, path='auth/sso')
```
mlassnig: maybe the endpoint should actually be called `auth/saml` ? the mental jump from sso to saml is not obvious for people not used to AAI

lib/rucio/core/authentication.py
```diff
@@ -301,6 +302,26 @@ def get_ssh_challenge_token(account, appid, ip=None, session=None):
...
 
+@transactional_session
+def get_auth_token_saml(account, saml_nameid, appid, ip=None, session=None):
+
+    # Make sure the account exists
+    if not account_exists(account, session=session):
+        return None
+
+    # remove expired tokens
+    session.query(models.Token).filter(models.Token.expired_at < datetime.datetime.utcnow(),
```
mlassnig, line 340: please add the `for_update` statement, like in https://github.com/rucio/rucio/pull/2721/files

lib/rucio/core/permission/atlas.py
```diff
@@ -314,6 +315,19 @@ def perm_get_auth_token_x509(issuer, kwargs):
     return False
 
 
+def perm_get_auth_token_saml(issuer, kwargs):
+    """
+    Checks if a user can request a token with user_pass for an account.
```
mlassnig: mean to say saml here

lib/rucio/db/sqla/migrate_repo/versions/1c45d9730ca6_increase_identity_length.py
```diff
@@ -81,13 +82,13 @@ def downgrade():
         drop_constraint('IDENTITIES_TYPE_CHK', 'identities', type_='check')
         create_check_constraint(constraint_name='IDENTITIES_TYPE_CHK',
                                 table_name='identities',
-                                condition="identity_type in ('X509', 'GSS', 'USERPASS')")
+                                condition="identity_type in ('X509', 'GSS', 'USERPASS', 'SAML')")
```
mlassnig: in the downgrade, i think you ought to remove SAML here and keep SSH :-)

(same below)

lib/rucio/db/sqla/util.py
```diff
@@ -128,6 +130,8 @@ def create_root_account():
         gss_email = config_get('bootstrap', 'gss_email')
         ssh_id = config_get('bootstrap', 'ssh_identity')
         ssh_email = config_get('bootstrap', 'ssh_email')
+        saml_id = config_get('bootstrap', 'saml_id')
```
mlassnig: these should be added to the template in etc/

lib/rucio/web/rest/webpy/v1/authentication.py
```diff
@@ -441,6 +450,103 @@ def GET(self):
...
+            200 OK
+
+        HTTP Error:
+            401 Unauthorized
+
+        :param Rucio-Account: Account identifier as a string.
+        :param Rucio-Username: Username as a string.
+        :param Rucio-Password: Password as a string.
+        :param Rucio-AppID: Application identifier as a string.
+        :returns: "Rucio-Auth-Token" as a variable-length string header.
```
mlassnig: returns X-Rucio-SAML-Auth-URL, no?

lib/rucio/web/rest/webpy/v1/authentication.py
```diff
@@ -441,6 +450,103 @@ def GET(self):
...
+                raise InternalError(error)
+
+            if not result:
+                raise generate_http_error(401, 'CannotAuthenticate', 'Cannot authenticate to account %(account)s with given credentials' % locals())
+
+            header('X-Rucio-Auth-Token', result.token)
+            header('X-Rucio-Auth-Token-Expires', date_to_str(result.expired_at))
+            return str()
+
+        SAML_PATH = '/opt/rucio/lib/rucio/web/ui/common/saml/'
```
mlassnig: ok, we can't really depend on a path like this. for what exactly is it needed?

lib/rucio/web/ui/common/saml/advanced_settings.json
```diff
@@ -0,0 +1,33 @@
+{
```
mlassnig, line 1: this will have to be customised for the CERN SAML provider, right?

lib/rucio/web/ui/common/utils.py
```diff
@@ -20,6 +20,51 @@
...
+
+def prepare_webpy_request(request, data):
+    """
+    Prepare a webpy request for SAML
+
+    :param request: webpy request object
+    :param data: GET or POST data
+    """
+    return {
+        'https': 'on' if request['wsgi.url_scheme'] == 'https' else 'off',
```
mlassnig: i'm strongly in favour of failing hard if https is not available.

lib/rucio/web/ui/common/utils.py
```diff
@@ -20,6 +20,51 @@
...
+        # Uncomment if using ADFS as IdP
+        # 'lowercase_urlencoding': True,
+        'post_data': data
+    }
+
+
+def set_cookies(token, cookie_accounts, attribs, ui_account=None):
+    # if there was no valid session token write the new token to a cookie.
+    if token:
+        setcookie('x-rucio-auth-token', value=token, path='/')
```
mlassnig: cookies are case-sensitive, and for consistency we should keep them in line with the headers

lib/rucio/web/ui/common/utils.py
```diff
@@ -20,6 +20,51 @@
...
+        setcookie('rucio-auth-token-created-at', value=int(time()), path='/')
```
mlassnig: int is enough? or better long?

tools/pip-requires-client
```diff
@@ -14,3 +14,4 @@ futures>=3.2.0; python_version < '3.0'            # Clean single-source support
 six>=1.12.0<1.13.0                                # Python 2 and 3 compatibility utilities
 boto3>=1.9.130,<1.10.0                            # S3 boto protocol (new version)
 pysftp>=0.2.9,<0.3                                # Used by sftp protocol
+#python3-saml>=1.6.0								  # Used for making SAML requests using python
```
mlassnig: this should go to all requires files, no?

mlassnig (commented):
looks good from an auth workflow perspective to me. still unclear how we can register the app to the CERN SAML provider, but that's for us to figure out. some tiny comments inside.

lib/rucio/web/rest/webpy/v1/authentication.py
Ruturaj123: This is needed as a pointer to SAML config file (settings.json) and the SP certificate & key.

lib/rucio/web/ui/common/saml/advanced_settings.json
Ruturaj123: Yes, it needs to be customized as per CERN SAML provider

lib/rucio/web/ui/common/utils.py
Ruturaj123: So if https is not available, an error should be raised?

lib/rucio/db/sqla/migrate_repo/versions/1c45d9730ca6_increase_identity_length.py
Ruturaj123: Remove SAML means delete SAML from account map and identities table or just remove SAML in the condition above? As of now, SSH keys are deleted from the table so instead of that SAML credentials need to be deleted?

lib/rucio/web/ui/common/utils.py
mlassnig: ah wait, this is for everything, or only for the authentication? authentication must be encrypted, the rest doesn't matter

lib/rucio/db/sqla/migrate_repo/versions/1c45d9730ca6_increase_identity_length.py
mlassnig: the downgrade should bring the database in the same state as it was before. what you are effectively doing here is adding the SAML type in the upgrade, then when you downgrade, instead of removing the SAML you removed the SSH.

mlassnig: this applies to the whole migrate script

Ruturaj123: Oh okay, I got it. Sorry Sir.

lib/rucio/web/ui/common/utils.py
Ruturaj123: This function is used to create the request format for SAML login. So this is used to login to IdP

lib/rucio/web/rest/webpy/v1/authentication.py
Ruturaj123: Also, should we change this path or keep it as it is?

bari12:
Hi @Ruturaj123 
Can you please fix the syntax and python3 tests which still fail. (Have a look on travis) Otherwise I think we can merge.

Ruturaj123:
Actually Sir the syntax tests are failing because there is a problem while setting up the docker image. The package xmlsec1 is not getting installed even though I have added it in `etc/docker/travis/Dockerfile`. How can we resolve this? When I install these packages in the dev environment they are installed successfully.

lib/rucio/db/sqla/migrate_repo/versions/1c45d9730ca6_increase_identity_length.py
bari12: Please create a new migrate script for the adding of `SAML`. This should not be appended to the existing one 😄

bari12:
Ah I see. The syntax tests are actually not executed out of the docker container, thus adding the packages there doesn't do anything. Please check with @hahahannes how to add it to the workflow of the syntax tests. I think this can be just added to one of the test prepare scripts.

hahahannes:
Mh but if the package is added to the travis dockerfile, pylint should also see it

Ruturaj123:
So @hahahannes how can the pylint error be fixed?

hahahannes:
Just realized that I mixed up the docker images. So after a discussion with @tbeerman, we think the easiest would be to create an own tools/pip-requires-syntax to only install sphinx, pylint and flake and not all the dependencies for running the rucio tests. Then this error would not occur. Or we just install the required packages in the tools/install_script.sh before starting pip. Or we start the syntax tests in an own container like the rucio tests. 
But an own pip-requires-syntax would be the cleanest solution

Ruturaj123:
So should I remove the packages from travis Dockerfile and add it to tools/travis/install_script.sh?

tbeerman:
No, leave it in the Dockerfile. If you remove it there all the DB tests would fail. Instead, also install those packages in the Travis VM. I don't know if it works but an `apt install` should work since it's running on Ubuntu. If if doesn't work a separate pip-requires-syntax, as proposed by @hahahannes, would be the way to go.

lib/rucio/db/sqla/migrate_repo/versions/1c45d9730ca6_increase_identity_length.py
Ruturaj123: Sir I was trying to create a new alembic migrate script for SAML now, but it is throwing an error. After talking to hahahannes, it seems that there is a problem with the alembic version on the next branch. So, for now, I have kept this on hold and I will create a new script as soon as the bug on next is fixed. If there is any other workaround then just suggest me and I will do that asap.

hahahannes: I opened #2872 for this

tbeerman:
The tests failed again. You'll have to use sudo with apt-get.

tbeerman:
Failed again. This time it couldn't find the packages. Seems like somebody had the problem already [1]. Try a `sudo apt-get update` before the install.

[1] https://github.com/travis-ci/travis-ci/issues/7351

Ruturaj123:
Yes Sir, I was going to do the same. So should I add it to before_script.sh?

tbeerman:
No, in the install_script. Just before you run `apt-get install`

Ruturaj123:
Okay so the syntax tests are now passing. The error is in python3 tests:
```
PYLINT FAILED
cat: pylint.out: No such file or directory
```

hahahannes:
The pylint output is saying, that you are using the `input` statement which seems to be not compatible with python2 and 3. Try maybe:
```
import six
six.moves.input()
```

Ruturaj123:
Can someone please restart the travis build? All the tests are now passing just the oracle test failed because no output was received

Ruturaj123:
@bari12 Sir, can you please restart the travis build? I have made the changes and added a new alembic script too.

hahahannes:
You can login on Travis with your Github credentials. Then you can restart the tests :)

Ruturaj123:
@hahahannes Sir, thank you for the help :smile:  And sorry for the trouble

tbeerman (dismissed):

lib/rucio/api/authentication.py
```diff
@@ -141,6 +142,26 @@ def get_ssh_challenge_token(account, appid, ip=None):
...
+    :param appid: The application identifier as a string.
+    :param ip: IP address of the client as a string.
+    :returns: Authentication token as a variable-length string.
+    """
+
+    kwargs = {'account': account, 'saml_nameid': saml_nameid}
+    if not permission.has_permission(issuer=account, action='get_auth_token_saml', kwargs=kwargs):
+        raise exception.AccessDenied('User with identity %s can not log to account %s' % (saml_nameid, account))
+
+    return authentication.get_auth_token_saml(account, saml_nameid, appid, ip)
```
bari12, line 164: Change account to `InternalAccount`
Similar to line 141 in this file.

lib/rucio/db/sqla/migrate_repo/versions/f1b14a8c2ac1_postgres_use_check_constraints.py
```diff
@@ -45,7 +46,7 @@ def upgrade():
     did_types = ('A', 'C', 'D', 'F', 'Y', 'X', 'Z')
     types_to_drop = set()
     for table, column, constraint_name, constraint in (
-            ('account_map', 'identity_type', "ACCOUNT_MAP_ID_TYPE_CHK", ('X509', 'GSS', 'USERPASS', 'SSH')),
+            ('account_map', 'identity_type', "ACCOUNT_MAP_ID_TYPE_CHK", ('X509', 'GSS', 'USERPASS', 'SSH', 'SAML')),
```
bari12: I don't think this should be added there, since it is an historic migrate script. All your changes should come from the `9a1b149a2044_add_saml_identity_type.py ` migrate script you wrote.

lib/rucio/db/sqla/migrate_repo/versions/f1b14a8c2ac1_postgres_use_check_constraints.py
```diff
@@ -63,7 +64,7 @@ def upgrade():
             ('did_keys', 'key_type', "DID_KEYS_TYPE_CHK", ('ALL', 'CONTAINER', 'DERIVED', 'COLLECTION', 'DATASET', 'FILE')),
             ('dids', 'did_type', "DIDS_TYPE_CHK", did_types),
             ('dids', 'availability', "DIDS_AVAILABILITY_CHK", ('A', 'D', 'L')),
-            ('identities', 'identity_type', "IDENTITIES_TYPE_CHK", ('X509', 'GSS', 'USERPASS', 'SSH')),
+            ('identities', 'identity_type', "IDENTITIES_TYPE_CHK", ('X509', 'GSS', 'USERPASS', 'SSH', 'SAML')),
```
bari12: See above

lib/rucio/db/sqla/migrate_repo/versions/9a1b149a2044_add_saml_identity_type.py
```diff
@@ -0,0 +1,80 @@
...
+                                condition="identity_type in ('X509', 'GSS', 'USERPASS', 'SSH', 'SAML')")
+
+
+def downgrade():
+    '''
+    Downgrade the database to the previous revision
+    '''
+
+    if context.get_context().dialect.name in ['oracle', 'mysql']:
+        execute("DELETE FROM account_map WHERE identity_type='SAML'")  # pylint: disable=no-member
```
bari12: Please remove the `DELETE FROM` lines. Of course the downgrade wouldn't work without them, but we do all the migrate scripts with the assumptions that the admins clean the tables up themselves.

lib/rucio/db/sqla/migrate_repo/versions/9a1b149a2044_add_saml_identity_type.py
```diff
@@ -0,0 +1,80 @@
...
+
+        drop_constraint('ACCOUNT_MAP_ID_TYPE_CHK', 'account_map', type_='check')
+
+        create_check_constraint(constraint_name='ACCOUNT_MAP_ID_TYPE_CHK',
+                                table_name='account_map',
+                                condition="identity_type in ('X509', 'GSS', 'USERPASS', 'SSH')")
+
+    elif context.get_context().dialect.name == 'postgresql':
+        schema = context.get_context().version_table_schema + '.' if context.get_context().version_table_schema else ''
+        execute("DELETE FROM " + schema + "account_map WHERE identity_type='SAML'")  # pylint: disable=no-member
```
bari12: Same as above

lib/rucio/db/sqla/migrate_repo/versions/9a1b149a2044_add_saml_identity_type.py
```diff
@@ -0,0 +1,80 @@
...
+        execute("DELETE FROM " + schema + "identities WHERE identity_type='SAML'")  # pylint: disable=no-member
+
+        drop_constraint('ACCOUNT_MAP_ID_TYPE_FK', 'account_map', type_='foreignkey')
+        op.execute('ALTER TABLE ' + schema + 'identities DROP CONSTRAINT IF EXISTS "IDENTITIES_TYPE_CHK", ALTER COLUMN identity_type TYPE VARCHAR')  # pylint: disable=no-member
```
bari12: You should be able to do these just with `drop_constraint` too?

lib/rucio/web/rest/webpy/v1/authentication.py
bari12: Let's load the path from the rucio.cfg where you specify it in a `[saml]` section.
Thus: `config_get('saml', 'config')` (or similar)

lib/rucio/web/ui/common/saml/advanced_settings.json
bari12: Should we rename this specific one then to template?

tools/pip-requires
```diff
@@ -21,3 +21,4 @@ Flask==1.0.2                                      # Python web framework
 fts3-rest-API==3.7.1; python_version == '2.7'     # FTS rest API
 M2Crypto<=0.35.2                                  # Dependency of FTS rest API; Temporary fix since 0.33 does not install on CC7
 MyProxyClient==2.1.0                              # myproxy client bindings
+#python3-saml>=1.6.0						  	  # Used for making SAML requests using python
```
bari12: Please remove

tools/pip-requires-client
```diff
@@ -14,3 +14,4 @@ futures>=3.2.0; python_version < '3.0'            # Clean single-source support
 six>=1.12.0<1.13.0                                # Python 2 and 3 compatibility utilities
 boto3>=1.9.130,<1.10.0                            # S3 boto protocol (new version)
 pysftp>=0.2.9,<0.3                                # Used by sftp protocol
+#python3-saml>=1.6.0						  	  # Used for making SAML requests using python
```
bari12: Please remove

tools/pip-requires-test
```diff
@@ -19,3 +19,4 @@ subprocess32==3.5.3; python_version <= '3.0'  # A backport of the subprocess mod
 pycodestyle==2.3.1                            # New package replacing pep8
 mock==2.0.0                                   # Mock package needed for readthedocs build for mocking pycurls
 s3cmd==2.0.2                                  # Used for testing of S3
+python3-saml>=1.6.0						  	  # Used for making SAML requests using python
```
bari12: Please remove it here and put it directly to the `setup_rucio.py`
There is the possiblity to define extras (Similar to the oracle, kerberos, etc. extras) Please specify one for `saml` with this dependency.

bari12 (changes requested):
Hi Ruturaj, 
Apologies for the late review comments. I just have a few minor changes, @tbeerman had already a look on the webui things and approved it. It would be awesome if you could still have a look and make the adjustments! Thanks!

Ruturaj123:
Yes Sir, I will update and push the changes soon!

tools/pip-requires-test
Ruturaj123: Removing this causes the travis syntax test case to fail as it cannot import the module.

Ruturaj123:
Also the other cases are failing with the error `Error: No such container: rucio`. What could be the reason for this?

tools/pip-requires-test
bari12: Hi Ruturaj, yes you have to add this to the `setup.py` as an extra and then also to the Dockerfile https://github.com/rucio/rucio/blob/master/etc/docker/travis/Dockerfile Thus in the dockerfile you also install it with the extra (`pip install .[oracle,...,saml]`)

lib/rucio/db/sqla/migrate_repo/versions/9a1b149a2044_add_saml_identity_type.py
Ruturaj123: After removing this, the test cases aren't passing as downgrade is failing.

lib/rucio/db/sqla/util.py
```diff
@@ -182,13 +187,17 @@ def create_root_account():
     identity4 = models.Identity(identity=ssh_id, identity_type=IdentityType.SSH, email=ssh_email)
     iaa4 = models.IdentityAccountAssociation(identity=identity4.identity, identity_type=identity4.identity_type, account=account.account, is_default=True)
 
+    # SAML authentication
+    identity5 = models.Identity(identity=saml_id, identity_type=IdentityType.SAML, email=saml_email)
```
bari12: I think we have to remove these, and you have to add them to the unit-tests.
Currently the downgrade test is not running through, because we add the new SAML identities at setup and then obviously the downgrade run is failing. Can you please change that. Otherwise it seems good to me!

ericvaandering:
This PR is causing a problem at this line: https://github.com/rucio/rucio/blob/870d36024db27f709d66e9b7c12974bf0edc6fe2/lib/rucio/web/rest/webpy/v1/authentication.py#L49

It assumes a "client" section exists in the config file for the server. And the server images are defined such that there is no client section.

bari12:
Yes I think this is done wrong and should probably be done similar to https://github.com/rucio/rucio/blob/7e32dee2c3af1fd51417481483449ca3d9c3bf28/lib/rucio/client/baseclient.py#L67
Thus check if the extra modules are on path and only load them if they are.
@Ruturaj123 do you possibly have time to submit a PR for this minor change?

bari12:
See #3209

PR #2690, opened 2019-06-24
Core & Internals: fixed python3 travis tests; Fix #2597

### LINKED ISSUE #2597: fix setup problems in python3 travis tests
Motivation
----------
Now that the travis tests are using python3, there are still some problems while setting up the tests
- psycopg2 throwing a type error for the binary type in the identity table while setting up the accounts
- py3 incompatible helper scripts like `tools/sync_rses.py`

Modification
------------

hahahannes:
Core & Internals: fixed python3 travis tests; Fix #2597


etc/docker/travis/alembic_mysql.ini
```diff
@@ -24,7 +24,7 @@ script_location = lib/rucio/db/sqla/migrate_repo/
 # the 'revision' command, regardless of autogenerate
 # revision_environment = false
 
-sqlalchemy.url = mysql://root:secret@mysql/mysql
+sqlalchemy.url = mysql+pymysql://root:secret@mysql/mysql
```
hahahannes: MySQL-python seems to be not python3 compatible and was installed in the Dockerfiles. I removed the installation and changed the driver to pyMySQL which gets already installed by setup.py.

lib/rucio/core/scope.py
```diff
@@ -43,11 +43,13 @@ def add_scope(scope, account, session=None):
         new_scope.save(session=session)
     except IntegrityError as e:
         if match('.*IntegrityError.*ORA-00001: unique constraint.*SCOPES_PK.*violated.*', e.args[0]) \
-           or match('.*IntegrityError.*1062, "Duplicate entry.*for key.*', e.args[0]) \
+           or match('.*IntegrityError.*Duplicate entry.*for key.*', e.args[0]) \
```
hahahannes, line 46: The error message from pymysql is slightly different and did not get catched

lib/rucio/core/monitor.py
```diff
@@ -22,14 +22,14 @@
 
 import time
 
-from pystatsd import Client
+from statsd import StatsClient
```
hahahannes, line 25: pystatsd seems to be not python3 compatible, so I choose another library (https://github.com/jsocol/pystatsd) with almost the same methods

etc/docker/travis/alembic_mysql.ini
```diff
@@ -24,7 +24,7 @@ script_location = lib/rucio/db/sqla/migrate_repo/
 # the 'revision' command, regardless of autogenerate
 # revision_environment = false
 
-sqlalchemy.url = mysql://root:secret@mysql/mysql
+sqlalchemy.url = mysql+pymysql://root:secret@mysql/mysql?charset=latin1
```
bari12, line 27: Why latin1? Shouldn't we use utf

lib/rucio/core/scope.py
mlassnig: this match happens in a lot of places, is it fixed everywhere?

etc/docker/travis/alembic_mysql.ini
hahahannes: For some reason the database schema gets initialised with latin1. Without the specific charset pymysql tries to use utf8mb4 and throws a lot of warnings. 
I also remember that I had the same problem with upgrading MySQL in #2478 where all strings with a length of 3000 were causing errors because the key length was too long with utf8 (3000 * 3 bytes > 3072 max key length)

lib/rucio/core/scope.py
hahahannes: seems like in most places the correct regex is used. There is one wrong regex in core/account which I will change.

etc/docker/travis/alembic_mysql.ini
mlassnig: there's no workaround for this problem in mysql/mariadb, so latin1 would be acceptable IMHO

lib/rucio/core/identity.py
```diff
@@ -47,7 +49,12 @@ def add_identity(identity, type, email, password=None, session=None):
 
     if type == IdentityType.USERPASS and password is not None:
         salt = os.urandom(255)  # make sure the salt has the length of the hash
-        password = hashlib.sha256('%s%s' % (salt, password.encode('utf-8'))).hexdigest()  # hash it
+        if six.PY3:
+            salt = salt.decode()
+            salted_password = ('%s%s' % (salt, password)).encode()
+        else:
+            salted_password = '%s%s' % (salt, str(password))
```
mlassnig, line 57: is it not necessary to do the encode('utf-8') in py2 anymore like previously?

lib/rucio/core/monitor.py
```diff
@@ -39,7 +39,17 @@ def record_counter(counters, delta=1):
     :param counters: The counter or a list of counters to be updated.
     :param delta: The increment for the counter, by default increment by 1.
     """
-    CLIENT.update_stats(counters, delta)
+    if type(counters) == 'list':
```
mlassnig: why not isinstance() ?

lib/rucio/core/scope.py
```diff
@@ -43,11 +43,13 @@ def add_scope(scope, account, session=None):
...
            or match('.*IntegrityError.*UNIQUE constraint failed: scopes.scope.*', e.args[0]) \
            or match('.*IntegrityError.*duplicate key value violates unique constraint.*', e.args[0])\
            or match('.*sqlite3.IntegrityError.*is not unique.*', e.args[0]):
             raise Duplicate('Scope \'%s\' already exists!' % scope)
+        else:
+            raise RucioException()
```
mlassnig: should raise with e? otherwise there's no detailed exception message

lib/rucio/db/sqla/session.py
```diff
@@ -95,26 +95,33 @@ def mysql_ping_listener(dbapi_conn, connection_rec, connection_proxy):
...
+        converter[FIELD_TYPE.NEWDECIMAL] = float
+    except ImportError:
+        try:
+            from pymysql.constants import FIELD_TYPE
+            from pymysql.converters import conversions as conv
+            converter = conv.copy()
+            converter[FIELD_TYPE.DECIMAL] = float
+            converter[FIELD_TYPE.NEWDECIMAL] = float
+        except ImportError:
+            raise RucioException('Trying to use MySQL without mysql-python or pymysql installed!')
```
mlassnig, line 123: should we foresee dropping mysql-python at all? does pymysql work in py2? otherwise maybe adding something in parentheses would help

e.g.,  ...  without mysql-python (for Python 2) or pymysql (for Python3) ...

lib/rucio/core/scope.py
hahahannes: ah yes, makes sense

lib/rucio/db/sqla/session.py
hahahannes: pymysql should be py2 compatible, too. So we could drop mysql-python. I have read somewhere that it is faster than pymysql because pymysql is only a pure python binding, but it is easier to install via pip.

lib/rucio/core/identity.py
hahahannes: I have added the encoding for py3 compatibility but at the wrong place, because the whole string needs to be encoded, not the password. But back then, my wrong fix was enough to let the pylint python3 test pass.

lib/rucio/core/scope.py
hahahannes: ok, done

lib/rucio/core/monitor.py
hahahannes: done

lib/rucio/core/identity.py
```diff
@@ -47,7 +49,12 @@ def add_identity(identity, type, email, password=None, session=None):
 
     if type == IdentityType.USERPASS and password is not None:
         salt = os.urandom(255)  # make sure the salt has the length of the hash
-        password = hashlib.sha256('%s%s' % (salt, password.encode('utf-8'))).hexdigest()  # hash it
+        if six.PY3:
+            salt = salt.decode()
+            salted_password = ('%s%s' % (salt, password)).encode()
```
mlassnig: i'm still not so super-happy about the whole `.encode()` business all over the codebase. is there a sensible way to do it more centrally? it's easy to forget to do it

bari12:
Sorry for the long wait @hahahannes 
Could you please rebase this.

lib/rucio/core/identity.py
hahahannes: I think there is no easier way to do it, unfortunately

hahahannes:
This PR makes now the test covered parts of rucio python3 compatible. Last thing for python3.6 is the deprecated fts3 package which will be removed in a later PR. For python3.5 there is a last problem with web.py which should be fixed with the next release soon

Issue #2612, opened 2019-05-28
Rucio user authentication via OIDC protocol (XDC IAM), getting user info and JWT tokens


gumond:
Motivation
----------
Main motivation to integrate OpenID Connect into Rucio can be found here:
https://twiki.cern.ch/twiki/bin/viewauth/LCG/WLCGAuthorizationWG

As a first step, Rucio needs to be able to use OIDC protocol as authentication service. It means redirect to an Identity Provider (IdP), e.g. IAM XDC, to login there and (if for the first time) authorise Rucio to get user information, then Rucio checks if the user is known and ask the IdP for an access and refresh tokens.

In order to do that, Rucio needs to:
- implement a OIDC client 
- cache tokens for their period of validity
- refresh tokens in case they expire using refresh tokens 
- eventually, on behalf of the user, perform the requested downstream actions using these tokens

Modification
------------
- implement a OIDC client in the authorisation scheme (running on the auth server)
- implement the OpenID information to be recognised and mapped to the rucio accounts & identities
- implement access token refresh mechanism 
- implement token validation checks (if not already part of the above)
- implement token caching

gumond:
This issue will be used only for modifications of rucio/core libraries. Second issue will be created for the IAM prototype script.

PR #2699, opened 2019-06-25
Authentication: Rucio user authentication via OIDC client #2612

gumond:
Authentication: Rucio user authentication via OIDC client #2612


gumond:
This is the first prototype of OIDC authentication for Rucio.

Some issues to follow upon: 
- access tokens: currently can not be saved in the DB due to column length limit being too small. access tokens currently have 458 characters and they will not get shorter. It is more probable they will get longer once we introduce groups/capabilities etc. In IAM the token value size is limited to 4096 characters (could be our DB limit at the moment).

- refresh tokens: authentication flow using refresh tokens is currently not implemented, as it needs some discussion

gumond:
As I will be leaving until 18.8., here is what one needs to test the current functionalities of this branch:
 
- Rucio auth server, it can be an open stack machine with the usual production installation of Rucio docker containers

On this auth server one needs (done from the bash of dev_rucio_1 docker in my case) :
  - presence of `/opt/rucio/etc/IAM_OIDC_client_secret.json` file with Rucio OIDC Client secret info content

The Rucio OIDC Client described in the file mentioned above needs to be configured also on the IAM XDC side consistently. This configuration must include the exact same host names in the redirection URIs. Also, the client must be instructed to revoke validity of refresh tokens once they have been used automatically (done by XDC IAM admin, not visible via the current IAM config web interface).

  - comment out line `SSLVerifyClient` in `/etc/httpd/conf.d/rucio.conf` (this needs to be thought through and configured appropriately with corresponding code changes before being put in any production - see TO DO section at the end of this comment !)
  - Installed the OIDC certified library on the auth server being tested: `$ pip install oic`
  - if your auth server is running in docker, it is important to also disable firewall on that host for the port 443: 
`$ sudo firewall-cmd --zone=public --add-port=443/tcp --permanent`
`$ sudo firewall-cmd --reload`

Than one should be allowed to test the following: 

* Adding OIDC user identity = user’s SUB + issuers URL (globally unique identifier), e.g.:
  - `$ rucio-admin identity add --account root --type OIDC --id "SUB=b3127dc7-2be3-417b-9647-6bf61238ad01, ISS=https://iam.extreme-datacloud.eu/" --email "Jaroslav.Guenther@gmail.com"`
  - `$ rucio-admin account list-identities root`

* The Authorization Code Flow (after removing locally stored access token):
  - `rm /tmp/root/.rucio_root/auth_token_root`
  - `rucio -a=root -S=’OIDC’ -u=<username> -pwd=<password> -sc=’openid’ --auth-host='https://rucio-auth-dev-guenther.cern.ch' -v whoami`
  - now test just access token read and validation, e.g. by `rucio whoami`

* The Authorization Code Flow With Refresh Token Request (after removing locally stored access token), e.g.:
  - `rm /tmp/root/.rucio_root/auth_token_root`
  - `rucio -a=root -S=’OIDC’ -u=<username> -pwd=<password> -sc=’openid offline_access’ --auth-host='https://rucio-auth-dev-guenther.cern.ch' -v whoami`

  - refresh token is not returned to the user, it is stored in the Rucio DB only for later use (see test below). New access token is returned and returned to the user at the same time.
  - now test just access token read and validation, e.g. by `rucio whoami`

* The Access Token refresh via an Refresh Token from the Rucio DB, e.g.:
  - `rm /tmp/root/.rucio_root/auth_token_root`
  - `rucio -a=root -S=’OIDC_refresh’ -sub=<user’s OIDC sub claim> --auth-host='https://rucio-auth-dev-guenther.cern.ch' -v whoami`

  - Refresh token from the Rucio DB has been used to get new refresh and access token for the user. The new refresh token is stored in the Rucio DB (the used one is deleted) and a new access token is returned to the user to be stored locally. 
  - now test just access token read and validation, e.g. by `rucio whoami`

TO DO LIST:
- protect/move elsewhere the OIDC_refresh mechanism. In the current implementation, anyone knowing a sub claim for an account name (public info), could attempt from the CLI to ask for an access token and perform operations as someone else.
- consider the necessity of SSL client verification between:
    Rucio auth server → Identity Provider
    Identity Provider (automatic redirect to)→ Rucio auth server
- check how pyoidc library handles issuers public keys ( need to reinitialise regularly the Rucio OIDC client ?) 
- need to add monitor to the core authentication to monitor auth server activity for different requests of OIDC ! 
- discuss which standalone tests should be written - most of the functionality depends on the Identity Provider connectivity and configuration
- consider closure mechanisms (logout OIDC, session closure) are these handled/needed ?
- ‘nonce’ and ‘state’ has to go to a designated place withint the Rucio DB (not to the respective identity and token columns as it is now)
- revise the exceptions being thrown and their catching 
- add permissions  layer to the api 
- implementation of SCIM Rucio provisioning test client as a probe daemon syncing Rucio accounts and their and XDC IAM identities (Andrea already provided one for testing). With SCIM client credentials from Andrea, we were able to get an access token and retrieve user information from SCIM endpoint (currently using curl from command line only, implementation into a daemon is the next step).

gumond:
OIDC_refresh mechanism was removed from CLI (not accessible to users anymore). 
Access tokens are newly saved in tokens DB together with refresh tokens. New DB columns for the 'tokens' table were introduced [1]. 

+ TO-DO: to allow Rucio to automatically use user tokens for performing asynchronously authorised operations on behalf of them, we need a Rucio daemon checking token validity and performing deletion (expired)/refresh (with refresh flag True) operations automatically. Separate probe will be created - to be referenced later. 

[1]
New DB Requirements: 
ALTER TABLE mysql.tokens MODIFY token VARCHAR(3072);
ALTER TABLE mysql.tokens ADD COLUMN token_type VARCHAR(25) AFTER account;
ALTER TABLE mysql.tokens ADD COLUMN refresh bool AFTER token_type;

gumond:
Utility function 'build_url' in lib/rucio/common/utils.py was also fixed in recent commits to my branch. Now it is adding '/' to URL only if path is provided (not always as it was the case previously). The issue was that, if there was no path, but URL parameters were provided, we finished with e.g. 

`https://address/?`

 which is being refused by servers which control such conventions.

gumond:
Closing as a new pull request will be created shortly ( 'feature-2612-OIDC-AuthZN-base-next'  --> 'rucio/next' based on the current rucio/next branch.

PR #2726, opened 2019-07-12
Core & Internals: added source RSE handling to throttler; Fix #2611

### LINKED ISSUE #2611: Add Source rse mode to throttler
Motivation
----------
The throttler should be able to throttle based on source RSEs and not only destination RSEs.

Modification
------------
- this requires prefilling of source RSEs in the requests as the information is not available yet
- and the possibility to release datasets after a defined waiting time
- a way to configure the different modes to switch between source-destination and actvity-all_activities (see #2542)

hahahannes:
Core & Internals: added source RSE handling to throttler; Fix #2611

### DIFF
--- a/lib/rucio/core/request.py
+++ b/lib/rucio/core/request.py
@@ -28,12 +28,13 @@ from six import string_types
 
 from sqlalchemy import and_, or_, func, update
 from sqlalchemy.exc import IntegrityError
+from sqlalchemy.sql import tuple_
 from sqlalchemy.sql.expression import asc, bindparam, text, false, true
 
 from rucio.common.exception import RequestNotFound, RucioException, UnsupportedOperation
 from rucio.common.types import InternalAccount, InternalScope
-from rucio.common.utils import generate_uuid, chunks
-from rucio.core import transfer_limits as transfer_limits_core
+from rucio.common.utils import generate_uuid, chunks, get_parsed_throttler_mode
+from rucio.core.config import get
 from rucio.core.message import add_message
 from rucio.core.monitor import record_counter, record_timer
 from rucio.core.rse import get_rse_name, get_rse_transfer_limits
@@ -124,7 +125,7 @@ def queue_requests(requests, session=None):
     logging.debug("queue requests")
 
     request_clause = []
-    transfer_limits = {}
+    transfer_limits, rses = get_rse_transfer_limits(session=session), {}
     for req in requests:
 
         if isinstance(req['attributes'], string_types):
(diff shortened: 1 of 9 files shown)


hahahannes:
There seems to be a problem with sqlalchemy translating the `WHERE IN` clause which is causing this test to fail.
See https://github.com/sqlalchemy/sqlalchemy/issues/4766

hahahannes:
> There seems to be a problem with sqlalchemy translating the `WHERE IN` clause which is causing this test to fail.
> See [sqlalchemy/sqlalchemy#4766](https://github.com/sqlalchemy/sqlalchemy/issues/4766)

Should be fixed with sqlalchemy 1.3.6

lib/rucio/core/request.py
```diff
@@ -163,6 +156,30 @@ def queue_requests(requests, session=None):
             for request in query_existing_requests:
                 existing_requests.append(request)
 
+    request_scopes_names = [(request['scope'], request['name']) for request in requests]
+    results = session.query(models.RSEFileAssociation.scope, models.RSEFileAssociation.name, models.RSEFileAssociation.rse_id, models.Distance.dest_rse_id, models.Distance.ranking)\
```
bari12: So this query is basically to fill the `src_rse_id` since we do not have it at this point.
I am wondering if we should do this for all requests, or only the ones affected by the throttler, but it is probably good to do it just for all. I hope that the query does not slow down the workload too much, but that we have to see.
Som change requests:
- Please add chunking here (Look at other modules where this is done) The requests list can potentially be quite long, so you would be querying for thousands of replicas in one query. I wouldn't do chunks larger than 50 here.
- Right now you are selecting a random source, but we should at least make sure that a disk source is selected, if it is available (and not a tape)

lib/rucio/core/request.py
```diff
@@ -163,6 +156,30 @@ def queue_requests(requests, session=None):
...
+        if scope not in source_rses:
+            source_rses[scope] = {}
+        if name not in source_rses[scope]:
+            source_rses[scope][name] = {}
+        if dest_rse_id not in source_rses[scope][name]:
+            source_rses[scope][name][dest_rse_id] = {}
+        if src_rse_id not in source_rses[scope][name][dest_rse_id]:
+            source_rses[scope][name][dest_rse_id][src_rse_id] = distance
+
+    throttler_mode = get('throttler', 'mode', default='DEST_PER_ACT', use_cache=False)
```
bari12: Shouldn't the default be no throttler to be used at all?

lib/rucio/core/request.py
```diff
@@ -868,41 +893,94 @@ def get_heavy_load_rses(threshold, session=None):
...
         raise RucioException(error.args)
 
 
+@transactional_session
+def release_waiting_requests_per_deadline(rse_id=None, deadline=1, session=None):
+    """
+    Release waiting requests that were waiting too long and exceeded the maximum waiting time to be released.
+    If the DID of a request is attached to a dataset, the oldest requested_at date of all requests related to the dataset will be used for checking and all requests of this dataset will be released.
+    :param rse_id:           The source RSE id.
+    :param deadline:         Maximal waiting time until a dataset gets released.
```
bari12: Please add wait time in **hours**

lib/rucio/core/request.py
```diff
@@ -1079,7 +1165,7 @@ def release_waiting_requests_fifo(rse_id, activity=None, count=None, account=Non
 
 
 @transactional_session
-def release_waiting_requests_grouped_fifo(rse_id, count=None, session=None):
+def release_waiting_requests_grouped_fifo(rse_id, count=None, direction='destination', deadline=1, volume=0, session=None):
```
bari12, line 1174: Add the new parameters to the description

lib/rucio/core/request.py
```diff
@@ -1113,16 +1204,14 @@ def release_waiting_requests_grouped_fifo(rse_id, count=None, session=None):
...
-    amount_updated_requests += release_waiting_requests_per_free_volume(rse_id, volume=volume, session=session)
+    if volume:
+        amount_updated_requests += release_waiting_requests_per_free_volume(rse_id=rse_id, volume=volume, session=session)
 
     return amount_updated_requests
 
 
 @transactional_session
-def release_all_waiting_requests(rse_id, activity=None, account=None, session=None):
+def release_all_waiting_requests(rse_id, activity=None, account=None, direction='destination', session=None):
```
bari12, line 1223: Add new parameters to description

lib/rucio/core/rse.py
```diff
@@ -665,12 +665,16 @@ def set_rse_transfer_limits(rse_id, activity, rse_expression=None, max_transfers
     :param transfers: Current number of tranfers.
     :param waitings: Current number of waitings.
     :param volume: Maximum transfer volume in bytes.
+    :param deadline: Maximum waiting time until a datasets gets released.
```
bari12: in hours

bari12 (changes requested):
Hi Hannes, Sorry for the late review of this. I have a few comments 😄

bari12:
For me this looks good. @cserf please have a look as well on the mechanic.

PR #2735, opened 2019-07-18
Enhancement: Add models and methods for listing users, adding and removing dids from followed table

Ruturaj123:
Enhancement: Add models and methods for listing users, adding and removing dids from followed table


lib/rucio/core/did.py
```diff
@@ -1829,3 +1830,115 @@ def list_archive_content(scope, name, session=None):
...
+    :param account: The account owner.
+    :param session: The database session in use.
+    """
+    try:
+        for did in dids:
+            did = session.query(models.DataIdentifier).filter_by(scope=did['scope'], name=did['name']).one()
+            new_did_followed = models.DidsFollowed(scope=did.scope, name=did.name, account=account,
+                                                   did_type=did.did_type, expired_at=did.expired_at)
+
+            new_did_followed.save(session=session, flush=True)
```
mlassnig: any reason why you flush to the DB every iteration? just curious

lib/rucio/core/did.py
```diff
@@ -1829,3 +1830,115 @@ def list_archive_content(scope, name, session=None):
...
+    """
+    Return list of dids followed by an user
+
+    :param scope: The scope name.
+    :param name: The data identifier name.
+    :param session: The database session in use.
+    """
+    try:
+        query = session.query(models.DidsFollowed).filter_by(scope=scope, name=name).all()
+        users = list()
```
mlassnig: shouldn't be necessary to convert here? can you try without?

lib/rucio/db/sqla/models.py
```diff
@@ -1251,6 +1252,37 @@ class LifetimeExceptions(BASE, ModelBase):
...
+
+class FollowEvents(BASE, ModelBase):
+    """Represents the events affecting the datasets which are followed"""
+    __tablename__ = 'follow_events'
+    scope = Column(String(SCOPE_LENGTH))
+    name = Column(String(NAME_LENGTH))
+    account = Column(String(25))
+    did_type = Column(DIDType.db_type(name='FOLLOW_EVENTS_TYPE_CHK'))
+    event_type = Column(String(1024))
+    payload = Column(LargeBinary(255))
```
mlassnig: is the payload definition and size correct here?

lib/rucio/core/did.py
Ruturaj123: No Sir, not any specific reason as such. I will make the changes and flush after the loop is completed.

lib/rucio/core/did.py
Ruturaj123: Without creating a list? Actually I created a list of dictionaries so that it can be rendered as json in the rest api

lib/rucio/db/sqla/models.py
Ruturaj123: Martin Sir had asked to keep the payload as a blob so that's why LargeBinary. And for the size I checked similar definitions in the models.py file and they had used 255 for LargeBinary. So I kept it the same.. Should it be increased?

bari12: For other tables we use `Column(Text)` for this. I would just do this here as well.

Ruturaj123: Okay Sir. I will change it to `Column(Text)`

lib/rucio/core/did.py
Ruturaj123: Here should I used yield instead? So that no need to create a list

lib/rucio/core/did.py
```diff
@@ -1829,3 +1830,115 @@ def list_archive_content(scope, name, session=None):
...
+
+    :param dids: A list of dids.
+    :param account: The account owner.
+    :param session: The database session in use.
+    """
+    try:
+        for did in dids:
+            did = session.query(models.DataIdentifier).filter_by(scope=did['scope'], name=did['name']).one()
+            new_did_followed = models.DidsFollowed(scope=did.scope, name=did.name, account=account,
+                                                   did_type=did.did_type, expired_at=did.expired_at)
```
bari12: expired_at is not really needed here.

lib/rucio/api/did.py
```diff
@@ -409,3 +410,57 @@ def list_archive_content(scope, name):
...
+    """
+    Return list of dids followed by an user
+
+    :param account: The account owner.
+    :param session: The database session in use.
+    """
+    return did.get_users_following_did(name=name, scope=scope, session=session)
+
+
+def remove_did_from_followed(scope, name, session=None):
```
bari12: The core method has an account, this is needed here too.

lib/rucio/core/did.py
```diff
@@ -1829,3 +1830,115 @@ def list_archive_content(scope, name, session=None):
...
+    :param name: The data identifier name.
+    :param event_type: The type of event affecting the did.
+    :param payload: Any message to be stored along with the event.
+    :param session: The database session in use.
+    """
+    try:
+        dids = session.query(models.DidsFollowed).filter_by(scope=scope, name=name).all()
+
+        for did in dids:
+            new_event = models.DidsFollowed(scope=scope, name=name, account=did.account,
```
bari12: I suppose this should be FollowEvents

lib/rucio/daemons/follower/follower.py
```diff
@@ -0,0 +1,116 @@
...
+    logging.info('event_aggregation: started')
+
+    hostname = socket.gethostname()
+    pid = os.getpid()
+    current_thread = threading.current_thread()
+    live(executable='rucio-follower', hostname=hostname, pid=pid, thread=current_thread)
+
+    while not graceful_stop.is_set():
+        try:
+            accounts = session.query(FollowEvents.account).distinct()
```
bari12: Don't use SQLAlchemy in the daemon directly; Also do this via core calls.

lib/rucio/db/sqla/models.py
```diff
@@ -1251,6 +1252,37 @@ class LifetimeExceptions(BASE, ModelBase):
...
 
 
+class DidsFollowed(BASE, ModelBase):
+    """Represents the datasets followed by an user"""
+    __tablename__ = 'dids_followed'
+    scope = Column(String(SCOPE_LENGTH))
+    name = Column(String(NAME_LENGTH))
+    account = Column(String(25))
+    did_type = Column(DIDType.db_type(name='DIDS_FOLLOWED_TYPE_CHK'))
+    expires_at = Column(DateTime)
```
bari12: I don't think we need `expires_at` here.

lib/rucio/db/sqla/models.py
```diff
@@ -1251,6 +1252,37 @@ class LifetimeExceptions(BASE, ModelBase):
...
+    _table_args = (PrimaryKeyConstraint('scope', 'name', 'account', name='DIDS_FOLLOWED_PK'),
+                   CheckConstraint('SCOPE IS NOT NULL', name='DIDS_FOLLOWED_SCOPE_NN'),
+                   CheckConstraint('NAME IS NOT NULL', name='DIDS_FOLLOWED_NAME_NN'),
+                   CheckConstraint('DID_TYPE IS NOT NULL', name='DIDS_FOLLOWED_DID_TYPE_NN'),
+                   ForeignKeyConstraint(['account'], ['accounts.account'], name='DIDS_FOLLOWED_ACCOUNT_FK'))
```
bari12: Also add FK to scope name in dids

lib/rucio/db/sqla/models.py
```diff
@@ -1251,6 +1252,37 @@ class LifetimeExceptions(BASE, ModelBase):
...
+    name = Column(String(NAME_LENGTH))
+    account = Column(String(25))
+    did_type = Column(DIDType.db_type(name='FOLLOW_EVENTS_TYPE_CHK'))
+    event_type = Column(String(1024))
+    payload = Column(Text)
+    _table_args = (PrimaryKeyConstraint('scope', 'name', 'account', name='FOLLOW_EVENTS_PK'),
+                   CheckConstraint('SCOPE IS NOT NULL', name='FOLLOW_EVENTS_SCOPE_NN'),
+                   CheckConstraint('NAME IS NOT NULL', name='FOLLOW_EVENTS_NAME_NN'),
+                   CheckConstraint('DID_TYPE IS NOT NULL', name='FOLLOW_EVENTS_DID_TYPE_NN'),
+                   ForeignKeyConstraint(['account'], ['accounts.account'], name='FOLLOW_EVENTS_ACCOUNT_FK'))
```
bari12: If you query this by account (like in the daemon) you will need an index on accoun too.

lib/rucio/web/rest/webpy/v1/did.py
```diff
@@ -73,6 +75,7 @@
     '/new', 'NewDIDs',
     '/resurrect', 'Resurrect',
     '/list_dids_by_meta', 'ListByMeta',
+    '%s/follow/(.*)' % SCOPE_NAME_REGEXP, 'Follow',
```
bari12: only `%s/follow' 
There shouldn't be another parameter in the REST. the account should be part of the data

lib/rucio/api/did.py
```diff
@@ -409,3 +410,57 @@ def list_archive_content(scope, name):
...
+    """
+    Mark a did as not followed
+
+    :param scope: The scope name.
+    :param name: The data identifier name.
+    :param session: The database session in use.
+    """
+    return did.remove_did_from_followed(scope=scope, name=name, session=session)
```
bari12: Please also add permission checks here

bari12: And on the above; Have a look how this is done with the issuer in other APIs.

bari12 (changes requested):
A few comments in the code; Please also have a look on the alembic migrate scripts; You need to provide an upgrade/downgrade script as well for the table generation.

lib/rucio/db/sqla/models.py
Ruturaj123: Index means a new `CheckConstraint` for account too right?

bari12: No, really an Index similar to other tables; Alternatively I suppose we could re-order the PK to account, scope, name as well.

bari12:
Ah before I forget it, there is also things which came up in the syntax test in travis; Check the ones marked as Errors; (`E:...`)
It would also be good if you can add some test cases for adding/listing/removing followers etc.

lib/rucio/db/sqla/models.py
Ruturaj123: Okay Sir.

Ruturaj123:
> Ah before I forget it, there is also things which came up in the syntax test in travis; Check the ones marked as Errors; (`E:...`)
> It would also be good if you can add some test cases for adding/listing/removing followers etc.

Ruturaj123:
Yes Sir. I will make all the changes soon

lib/rucio/daemons/follower/follower.py
```diff
@@ -0,0 +1,81 @@
...
+    logging.info('event_aggregation: started')
+
+    hostname = socket.gethostname()
+    pid = os.getpid()
+    current_thread = threading.current_thread()
+    live(executable='rucio-follower', hostname=hostname, pid=pid, thread=current_thread)
+
+    while not graceful_stop.is_set():
+        # Create reports of events for all the accounts.
+        create_reports()
```
cserf: Each thread will send one report for all the accounts. Need partitioning

lib/rucio/core/did.py
```diff
@@ -1829,3 +1830,157 @@ def list_archive_content(scope, name, session=None):
...
+                    body += "\n"
+
+                body += "Thank You."
+
+                # Get the email associated with the account.
+                email = session.query(models.Account.email).filter_by(account=account)
+                add_message('email', {'to': email,
+                                      'subject': 'Report of affected dataset(s)',
+                                      'body': body})
+
```
cserf: One should clean the events when the message was successfully sent, otherwise at the next iteration the same list of events will be sent. Another possibility is to keep all the events, but only send the recent ones, based on created_at.

lib/rucio/core/permission/atlas.py
```diff
@@ -1042,3 +1045,35 @@ def perm_add_bad_pfns(issuer, kwargs):
...
+    """
+    Checks if an account can remove did from followed table.
+
+    :param issuer: Account identifier which issues the command.
+    :param kwargs: List of arguments for the action.
+    :returns: True if account is allowed, otherwise False
+    """
+    return _is_root(issuer)\
+        or has_account_attribute(account=issuer, key='admin')\
+        or rucio.core.scope.is_scope_owner(scope=kwargs['scope'], account=issuer)\
```
cserf: I would remove the condition on the scope and add one on `kwargs['account] == issuer`

lib/rucio/core/permission/atlas.py
```diff
@@ -1042,3 +1045,35 @@ def perm_add_bad_pfns(issuer, kwargs):
...
+    :param issuer: Account identifier which issues the command.
+    :param kwargs: List of arguments for the action.
+    :returns: True if account is allowed, otherwise False
+    """
+    if _is_root(issuer) or has_account_attribute(account=issuer, key='admin'):
+        return True
+    scopes = [did['scope'] for did in kwargs['dids']]
+    scopes = list(set(scopes))
+    for scope in scopes:
+        if not rucio.core.scope.is_scope_owner(scope, issuer):
```
cserf: Same comment as above (remove the condition on the scope and add one on `kwargs['account] == issuer`

lib/rucio/core/permission/cms.py
```diff
@@ -875,3 +878,35 @@ def perm_add_bad_pfns(issuer, kwargs):
...
+    """
+    Checks if an account can remove did from followed table.
+
+    :param issuer: Account identifier which issues the command.
+    :param kwargs: List of arguments for the action.
+    :returns: True if account is allowed, otherwise False
+    """
+    return _is_root(issuer)\
+        or has_account_attribute(account=issuer, key='admin')\
+        or rucio.core.scope.is_scope_owner(scope=kwargs['scope'], account=issuer)\
```
cserf: Same comment as above

lib/rucio/core/permission/cms.py
```diff
@@ -875,3 +878,35 @@ def perm_add_bad_pfns(issuer, kwargs):
...
+    :param issuer: Account identifier which issues the command.
+    :param kwargs: List of arguments for the action.
+    :returns: True if account is allowed, otherwise False
+    """
+    if _is_root(issuer) or has_account_attribute(account=issuer, key='admin'):
+        return True
+    scopes = [did['scope'] for did in kwargs['dids']]
+    scopes = list(set(scopes))
+    for scope in scopes:
+        if not rucio.core.scope.is_scope_owner(scope, issuer):
```
cserf: Same comment as above

lib/rucio/core/permission/generic.py
```diff
@@ -866,3 +869,35 @@ def perm_add_bad_pfns(issuer, kwargs):
...
+    """
+    Checks if an account can remove did from followed table.
+
+    :param issuer: Account identifier which issues the command.
+    :param kwargs: List of arguments for the action.
+    :returns: True if account is allowed, otherwise False
+    """
+    return _is_root(issuer)\
+        or has_account_attribute(account=issuer, key='admin')\
+        or rucio.core.scope.is_scope_owner(scope=kwargs['scope'], account=issuer)\
```
cserf: Same comment as above

lib/rucio/core/permission/generic.py
```diff
@@ -866,3 +869,35 @@ def perm_add_bad_pfns(issuer, kwargs):
...
+    :param issuer: Account identifier which issues the command.
+    :param kwargs: List of arguments for the action.
+    :returns: True if account is allowed, otherwise False
+    """
+    if _is_root(issuer) or has_account_attribute(account=issuer, key='admin'):
+        return True
+    scopes = [did['scope'] for did in kwargs['dids']]
+    scopes = list(set(scopes))
+    for scope in scopes:
+        if not rucio.core.scope.is_scope_owner(scope, issuer):
```
cserf: Same comment as above

cserf:
To keep in mind for you Martin : `delete_dids` should remove the rows in the followed table otherwise, we will have FK constraint violation

Ruturaj123:
> To keep in mind for you Martin : `delete_dids` should remove the rows in the followed table otherwise, we will have FK constraint violation

So should I make changes in `delete_dids` for removing the rows from followed table?

bari12:
Yes please.

lib/rucio/daemons/follower/follower.py
Ruturaj123: So how to go about doing the partitioning?

lib/rucio/api/did.py
```diff
@@ -530,3 +531,67 @@ def list_archive_content(scope, name):
     dids = did.list_archive_content(scope=scope, name=name)
     for d in dids:
         yield api_update_return_dict(d)
+
+
+def add_did_to_followed(scope, name, account, session=None):
```
bari12, line 536: Please change all scopes and accounts to `InternalScope` and `InternalAccount`. E.g. have a look on the `create_did_sample` method in this module.

lib/rucio/core/did.py
```diff
@@ -1833,3 +1841,171 @@ def list_archive_content(scope, name, session=None):
...
+                body += "\n"
+            body += "Thank You."
+            # Get the email associated with the account.
+            email = session.query(models.Account.email).filter_by(account=event.account)
+            add_message('email', {'to': email,
+                                  'subject': 'Report of affected dataset(s)',
+                                  'body': body})
+
+            # Clean up the events after creating the report
+            session.query(models.FollowEvents).filter_by(account=event.account).\
```
bari12: I don't think this deletion works, since you only processed a partition of the events of this account, but you are deleting all of them. @cserf suggested to delete the processed events when sending them.

lib/rucio/db/sqla/migrate_repo/versions/277bec1ddba9_update_dids_followed_and_follow_events_.py
```diff
@@ -0,0 +1,55 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Authors:
+# - Ruturaj Gujar, <ruturaj.gujar23@gmail.com>, 2019
+
+''' update dids_followed and follow_events table '''
```
bari12: Please merge this with the other migrate script. It's enough to do it in one 😄

bari12 (changes requested):
Hi Ruturaj, again a few comments. Sorry for the delay. Would be awesome if you could still integrate them! Thanks! 😄

bari12:
Looks good to me too. @cserf or @mlassnig  please also have a last look on this one.

lib/rucio/core/did.py
```diff
@@ -1833,3 +1841,173 @@ def list_archive_content(scope, name, session=None):
...
+    """
+    Create a summary report of the events affecting a dataset, for its followers.
+
+    :param session: The database session in use.
+    """
+    # Query the FollowEvents table
+    query = session.query(models.FollowEvents)
+
+    # Use hearbeat mechanism to select a chunck of events based on the hashed account
+    if total_workers > 0:
```
cserf, line 1973: Not blocking, but can switch to filter_thread_work

lib/rucio/core/did.py
```diff
@@ -1833,3 +1841,173 @@ def list_archive_content(scope, name, session=None):
...
+                account = event.account
+                # Clean up the event after creating the report
+                session.query(models.FollowEvents).\
+                    filter_by(scope=event.scope, name=event.name, account=event.account).\
+                    delete(synchronize_session=False)
+
+            body += "Thank You."
+            # Get the email associated with the account.
+            email = session.query(models.Account.email).filter_by(account=account)
+            add_message('email', {'to': email,
```
cserf, line 2008: Each thread will send a separate email. Can be improved.

lib/rucio/daemons/follower/follower.py
```diff
@@ -0,0 +1,96 @@
...
+    hostname = socket.gethostname()
+    sanity_check(executable='rucio-follower', hostname=hostname)
+
+    if once:
+        logging.info("executing one follower iteration only")
+        aggregate_events(once)
+    else:
+        logging.info("starting follower threads")
+        # Run the follower daemon thrice a day
+        threads = [get_thread_with_periodic_running_function(28800, aggregate_events, graceful_stop) for i in range(threads)]
```
cserf, line 90: Interval should be parametrizable

lib/rucio/db/sqla/migrate_repo/versions/7541902bf173_add_didsfollowed_and_followevents_table.py
```diff
@@ -0,0 +1,83 @@
...
+
+from alembic import context
+from alembic.op import (create_table, create_primary_key, create_check_constraint,
+                        create_foreign_key, create_index, drop_table)
+
+from rucio.db.sqla.constants import DIDType
+
+# Alembic revision identifiers
+revision = '7541902bf173'
+down_revision = '2cbee484dcf9'
```
cserf, line 31: Down revision must be changed (for you Martin)

cserf (approved):
A few non-blocking comments that can be addressed in a future PR. The only thing to change is the alembic revision number.
I would approve for now.

Issue #2717, opened 2019-07-05
S3 URL signing code should use RSE name to index credentials


jamesp-epcc:
Currently, the S3 URL signing code uses the hostname plus bucket name to index its credentials. The RSE name should be used for this instead. This will mean adding a new RSE name parameter to the `get_signed_url` function.

PR #2740, opened 2019-07-19
Core & Internals: look up objectstore credentials by RSE ID #2717

jamesp-epcc:
This pull request modifies the URL signing code to look up objectstore credentials based on the RSE ID rather than components of the URL. The RSE name is used at the API level and above.


jamesp-epcc:
Closed and reopened to force checks to rerun due to spurious failure.

lib/rucio/core/credential.py
```diff
@@ -134,8 +135,8 @@ def get_signed_url(service, operation, url, lifetime=600):
         if colon >= 0:
             host = host[:colon]
 
-        # look up in RSE account configuration by <hostname>_<bucketname>
-        cred_name = host + "_" + bucket
+        # look up in RSE account configuration by RSE ID
+        cred_name = rse
```
mlassnig: shouldn't this variable be called `rse_id` ?

lib/rucio/rse/__init__.py
```diff
@@ -46,12 +46,23 @@ def get_rse_client(rse, **kwarg):
     return RSEClient().get_rse(rse)
 
 
-def get_signed_url_client(service, op, url):
+def get_signed_url_client(rse, service, op, url):
```
mlassnig, line 50: this should also be `rse_id`, right?

lib/rucio/core/credential.py
jamesp-epcc: I have now changed the name to `rse_id` for consistency with the rest of the code.

lib/rucio/rse/__init__.py
jamesp-epcc: In this case it is an RSE name rather than an ID. The API layer translates it to an ID and passes that into the core.

jamesp-epcc:
Closing and reopening to force timed-out Oracle test to rerun

lib/rucio/api/credential.py
```diff
@@ -40,10 +42,14 @@ def get_signed_url(account, appid, ip, service, operation, url, lifetime):
 
     kwargs = {'account': account}
```
bari12: The account needs to be transformed into an `InternalAccount` Have a look on the api/did.py

bari12 (changes requested):
Hi @jamesp-epcc 
Apologies for the late review. I just have one minor change related to the Multi-VO changes which were introduced lately.

lib/rucio/api/credential.py
jamesp-epcc: I have added the conversion.

bari12 (approved):

PR #2778, opened 2019-08-02
Core & Internals: Add vo table and vo column to rse table. Fix #2727

### LINKED ISSUE #2727: Need VO table and VO column in RSE table
Motivation
----------
For the multi-VO work we will need to be able to store data on VOs

Modification
------------
Add table "vos" with columns:
    - vo (3 char tag for the vo, e.g SKA, ATL, CMS, ...) 
    - description (full name/description of vo, e.g. "Square Kilometre Array, ATLAS, ...)
    - email 
Add column "vo" to RSE table as FK into "vos". This will change the unique constraint on RSE table. (As 2 vos should be able to both have an RSE named the same thing)

AndrewLister-STFC:
Core & Internals: Add vo table and vo column to rse table. Fix #2727


AndrewLister-STFC:
I forgot features are branched from next. I'll rebase and force push

mlassnig (commented):
looks good to me. but i let @bari12 approve

lib/rucio/db/sqla/migrate_repo/versions/a118956323f8_added_vo_table_and_vo_col_to_rse.py
```diff
@@ -0,0 +1,84 @@
...
+    '''
+    Upgrade the database to this revision
+    '''
+
+    if context.get_context().dialect.name in ['oracle', 'postgresql', 'mysql']:
+        schema = context.get_context().version_table_schema if context.get_context().version_table_schema else ''
+        # add a vo table
+        vos = create_table('vos',
+                           sa.Column('vo', String(3), primary_key=True),
+                           sa.Column('description', String(255)),
```
vingar, line 48: `String(3)` looks too small for me. Maybe 50 or move the length to the constant file.

lib/rucio/db/sqla/models.py
```diff
@@ -1263,6 +1265,15 @@ class LifetimeExceptions(BASE, ModelBase):
                    ForeignKeyConstraint(['account'], ['accounts.account'], name='LIFETIME_EXCEPT_ACCOUNT_FK'))
 
 
+class VO(BASE, ModelBase):
```
vingar, line 1318: Won't VO can be an account attribute too  ? maybe a map table account.vo map can be useful ?

lib/rucio/db/sqla/migrate_repo/versions/a118956323f8_added_vo_table_and_vo_col_to_rse.py
vingar: Remark: If the VO name can change but needs to be referenced in other tables then the primary key should independent from the name like rse/id

bari12: It's only 3 characters because these are used to prefix (Rucio internally, not externally) all accounts and scopes of that VO. If we would put a longer one that would potentially eat too much space of the scope and account.

lib/rucio/db/sqla/models.py
AndrewLister-STFC: Yes, accounts will be linked to a specific VO. This will be done by encoding the VO in the "account" field in the db. e.g. "jdoe@vo1"
Due to restrictions from the fact that some existing users have instances with very large tables that would be time consuming to rebuild (if the structure of some tables changed), the best option is to encode this in the PK.
(This is also true of scopes)

The issue with having a separate table to capture this is that we want to allow vos to work without competing over namespace, and since the name is the PK in account, this means including information from the VO to make it unique. Think of "John Smith" and "Jane Smith" in 2 different VOs, they should both be able to use "jsmith" as they won't see the other in the system.

bari12: It won't be an account attribute, but a `vo` column in the accounts table. A vo map-account table is not really needed and it makes things a bit complicated since you could have tricky corner cases to e.g. separate data/rules/metadata etc from the different VOs

So if a person has access to multiple VOs, it will be an account in each of this VOs. The user can still use the same proxy/DN to login to both VOs, the VO just has to be named when authenticating.

lib/rucio/db/sqla/migrate_repo/versions/a118956323f8_added_vo_table_and_vo_col_to_rse.py
vingar: ok, it's a short vo name then. In a way, it looks more than an autogenerated id then. vo ->id , description -> vo and referenced as vo-id in the other tables if we follow the logic of some current objects imo. 

For illustration, what can be the column values for some of the current existing VOs, e.g., alice, atlas, ops, dteam, swegrid.se, ops.ndgf.org, nordugrid.org, skavengers.eu, eiscat3d ?

lib/rucio/db/sqla/models.py
vingar: If a person is part of several VOS, the same account / person will be in several entries in the account table, e.g., jdoe@vo1, jdoe@vo2, etc. Right?
About encoding the vo identifier in the PK, it breaks the relational model and won't enforce any contraints like foreign keys. Will it concern the scope column of the big tables like dids, replicas, etc ?

elichad:
Hi all,

I'm now picking up from Andrew on implementing the multi-VO functionality. Can anyone offer a timescale for the review of this PR?

Eli

bari12:
Hi Eli,
Let's have a chat about this on Slack. I mainly didn't merge this since it most likely will need some iterations and I wanted to wait for that until your start. But in principle this can go in very soon.
Cheers,
Martin

lib/rucio/db/sqla/models.py
```diff
@@ -1313,6 +1315,15 @@ class LifetimeExceptions(BASE, ModelBase):
...
+    """Represents the VOS in a MultiVO setup"""
+    __tablename__ = 'vos'
+    vo = Column(String(3))
+    description = Column(String(255))
+    email = Column(String(255))
+    _table_args = (PrimaryKeyConstraint('vo', name='VO_PK'), )
```
bari12: I think this should be called `VOS_PK`

lib/rucio/db/sqla/migrate_repo/versions/a118956323f8_added_vo_table_and_vo_col_to_rse.py
```diff
@@ -0,0 +1,84 @@
...
+def upgrade():
+    '''
+    Upgrade the database to this revision
+    '''
+
+    if context.get_context().dialect.name in ['oracle', 'postgresql', 'mysql']:
+        schema = context.get_context().version_table_schema if context.get_context().version_table_schema else ''
+        # add a vo table
+        vos = create_table('vos',
+                           sa.Column('vo', String(3), primary_key=True),
```
bari12: It should be possible to add the PK key name/hint here as well. Thus `VOS_PK`

bari12 (approved):

PR #2813, opened 2019-08-12
<Core & Internals>: Change caching endpoint address to be obtained from Rucio config or use localhost. Fix #2812

### LINKED ISSUE #2812: The caching endpoint address should be a configurable value.
Motivation
----------
The value specifying the address of the caching endpoint used by Dogpile should allow for values other than 127.0.0.1 so that remote caching services can be used.

bjwhite-fnal:
<Core & Internals>: Change caching endpoint address to be obtained from Rucio config or use localhost. Fix #2812

### DIFF
--- a/lib/rucio/common/objectstore.py
+++ b/lib/rucio/common/objectstore.py
@@ -46,7 +46,7 @@ logging.getLogger("boto.s3.connection").setLevel(logging.WARNING)
 
 REGION = make_region().configure('dogpile.cache.memcached',
                                  expiration_time=3600,
-                                 arguments={'url': "127.0.0.1:11211", 'distributed_lock': True})
+                                 arguments={'url': config.config_get('cache', 'url', False, '127.0.0.1:11211'), 'distributed_lock': True})
 
 # for local test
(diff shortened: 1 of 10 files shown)


lib/rucio/common/objectstore.py
```diff
@@ -46,7 +46,7 @@
 
 REGION = make_region().configure('dogpile.cache.memcached',
                                  expiration_time=3600,
-                                 arguments={'url': "127.0.0.1:11211", 'distributed_lock': True})
+                                 arguments={'url': config.config_get('cache', 'url', True, '127.0.0.1:11211'), 'distributed_lock': True})
```
mlassnig: the True should be False, otherwise we get an exception and not the default value ;-)

bjwhite-fnal: Well that was silly. Should be fixed!

mlassnig: oh no, github+travis are choking again and not starting the automated unittests . the only reliable way to get it to run is to delete the pull requests, also delete the branches on your github fork, and then do tools/submit-pull-request again

PR #2817, opened 2019-08-13
<Core & Internals>: Change caching endpoint address to be obtained from Rucio config or use localhost. Fix #2812

### LINKED ISSUE #2812: The caching endpoint address should be a configurable value.

bjwhite-fnal:
<Core & Internals>: Change caching endpoint address to be obtained from Rucio config or use localhost. Fix #2812


mlassnig:
sorry, travis dead again ;-(

mlassnig:
i wish we could pay them some money to provide a more robust and fast CI service ;-(

PR #2820, opened 2019-08-13
<Deletion> Reaper 2.0 will check if an RSE is greedy before looking up its limits. Fix #2819

### LINKED ISSUE #2819: Reaper 2.0 Should Reap Greedy RSE's Without Needing MinFreeSpace/MaxBeingDeletedFiles Set
Motivation
----------
The current version of the Reaper 2.0 will not reap an RSE marked as greedy that does not have MinFreeSpace and MaxBeingDeletedFiles set. MinFreeSpace should not matter for a greedy RSE, and MaxBeingDeletedFiles is not used by the Reaper 2.

Modification
------------
Move the check for greedy mode in __check_rse_usage() to before the acquisition of RSE limits.

bjwhite-fnal:
<Deletion> Reaper 2.0 will check if an RSE is greedy before looking up its limits. Fix #2819

### DIFF
--- a/lib/rucio/daemons/reaper/reaper2.py
+++ b/lib/rucio/daemons/reaper/reaper2.py
@@ -195,6 +195,14 @@ def __check_rse_usage(rse, rse_id, prepend_str):
     if result is NO_VALUE:
         max_being_deleted_files, needed_free_space, used, free, obsolete = None, None, None, None, None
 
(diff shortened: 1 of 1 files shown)


mlassnig:
Looks ok to me. Just don't forget to add yourself as an author at the top of the file. Second pair of eyes, @cserf ?

bari12 (approved):

PR #2842, opened 2019-08-19
Core & Internals: Load experiment-specific code from external package #533

jamesp-epcc:
I don't think this is ready to be merged yet, but I thought it would be useful to open a discussion before I go any further with it.

I have added a new configuration item "package" to the "policy" section. It specifies the name of a Python package that contains the experiment-specific code. The schema and permission code will then load modules from this package instead of from the core Rucio code. The "generic" code is still there in Rucio and it will fall back to this if the package is unspecified or unavailable.

The package format is very simple. It just needs to contain a module named `schema.py` and one named `permission.py`, which have exactly the same contents as the ones currently in the codebase. If one of them is omitted (because the experiment doesn't need to customise that module), the generic version will be used instead. It should also be possible to register `lfn2pfn` algorithms in the package's `__init__.py`. The package could be installed from Pypi, or it can just be in a local directory that's added to the `PYTHONPATH`.

Before this change is made it will probably be necessary to create proper packages for the existing experiment customisations (ATLAS, CMS, etc.) and make sure everyone is aware of the need to install them. I have created versions for testing but I don't think I am the right person to maintain them as I have no connection to those experiments.

Does this look like a sensible approach? I am not sure about the method I used to add the module contents to the global namespace. I couldn't find any other way to effectively do `from <module> import *` using a module name determined at run time, but maybe there is a better way.


bari12:
Hi @jamesp-epcc 
Sorry for the late review again...
I think the idea like this is good, I had the same thing in mind. We probably have to provide some kind of recipe how to build these policy packages to the community, so they can build their own.

Some questions/actions/ideas from me (not only to you, but for everyone to discuss):
- I think there shouldn't be a failover. If a policy package is provided it has to be enforced. If it cannot be loaded, Rucio must exit with an error and not load the generic one.
- We need to find a way to include the surl functions to the policy packages as well. Thus https://github.com/rucio/rucio/blob/master/lib/rucio/common/utils.py all the `construct_surl_DQ2`, `construct_surl_BelleII`, `construct_surl_T0`. These are somewhat similar to the lfn2pfn methods, but then again a bit more special.
- I am wondering if we need some kind of `supported version` field. Something to make sure that the policy package includes/considers the information of the respective Rucio version. I don't have a clear idea how this should look like, but I would like to make sure that e.g. we include new sets of methods for the permissions, that the policy package excludes these methods knowingly, and not only because it was not updated with the newest Rucio version in mind.

bjwhite-fnal:
Hey James, I'm trying to find some time to look over your changes this week before the Rucio workshop. I was thinking maybe we could work on this next week since we need it badly here at FNAL also.

jamesp-epcc:
Sounds good... I am planning to try and develop this further next week.

jamesp-epcc:
@bari12 Thanks for your comments and sorry for late response. I have been thinking about how to address these points.

- Removing the failover should be easy and I agree this makes more sense. I will add an exception if a policy package is specified and not found.
- For the surl functions it would be possible to add a registry, similar to the `RSEDeterministicTranslation` class for lfn2pfn. The policy packages could then register new ones when they initialise. We would need to make sure that the initialisation happens before any code that uses this functionality. (As an aside, Robert Illingworth said that he would like more functionality from the surl functions, including a way to signal failure, and possibly more arguments).
- I am not sure how best to handle the `supported_version` field. At present I think the code will just crash if a required method is missing. We would need some way to detect the missing methods and decide what to do about them. Maybe the package should provide a list of the methods it implements and the missing ones can either fallback to the built-in generic policy (if they were knowingly excluded) or cause an exception (if required by a newer, unsupported Rucio version).

jamesp-epcc:
After some discussion this morning I'm also wondering whether there are multi VO implications of this. What if a multi VO Rucio instance needs to have separate policy packages for each VO? Should this be supported?

vingar:
I think so as each VO might come with different naming conventions, permission, etc.

bari12:
Yes, I think we have to foresee that different VOs have different packages. But to make progress, I would suggest to prepare this in the single-VO implementation first. The adaptions to the multi-VO should not be too heavy then.

jamesp-epcc:
I've been having a think about the `supported_version` field. At present there is no risk of the code falling back to a generic version of a method because there is no mechanism to do this - if a method that's not implemented is requested, the code will crash. However it could happen in the future if (for example) a whole new module was added to the policy packages but an old package didn't implement it.

I can see a few different options here:

1. Include the `supported_version` field and exit with an error if it doesn't match the Rucio version in use. I think this is probably undesirable since every policy package would then have to be updated even for minor updates to Rucio that don't touch the policy related code.
2. Instead of an error, simply print a warning if the versions don't match. However I'm not sure if people would take notice of this.
3. Include the field in the specification for policy packages but don't use it yet. It could be used in future if a new module is added to the policy, for example.
4. Change the way Rucio interfaces with the permission and schema code so that individual methods can be left unimplemented and will fall back to the generic version. Use the `supported_version` field to ensure that newly added methods don't unknowingly fall back.

I'm also open to other suggestions.

bari12:
I think matching the exact Rucio version will be difficult, because you would really have to produce new policy packages all the time.

Thinking a bit more about this, maybe it is just fine to always fall back to the generic policy then. The thing is, for the permissions this is somewhat easy, for the schema it is not so obvious anymore. I think in that case we should just let it crash.

jamesp-epcc:
I've added some more details to the documentation and fixed the flake8 and pylint failures. I added `rucio.core.permission` and `rucio.common.schema` to pylint's ignored modules list because the way they are loaded now means imports from them are flagged up as errors by pylint. I think the remaining Travis failures are because the generic permission module is being used instead of the previous ATLAS one. To fix this we will need to either make the tests work with the generic permissions module, or somehow ensure that the ATLAS policy is installed for the CI tests.

mlassnig (approved):

bari12 (approved):

PR #2879, opened 2019-09-02
Core & Internals: added new endpoint and core function to list requests by RSE and state; Fix #2878

### LINKED ISSUE #2878: Request listing interface
Motivation
----------

We need a new API for the network people, such that they can query the queued and submitted requests for a given link. The complication comes that the network people basically only have the site name, and do not know about RSEs.

Modification
----------

Suggestion is to create a new endpoint in `web/request.py`:

`/request/list`

with the HTTP parameters:

`request_state=[Q,S,....]`

`src_rse`,`dst_rse`

`src_site`,`dst_site`  (where if the client uses sites instead of RSEs, then they have to be resolved to RSEs internally)

hahahannes:
Core & Internals: added new endpoint and core function to list requests by RSE and state; Fix #2878

### DIFF
--- a/lib/rucio/api/request.py
+++ b/lib/rucio/api/request.py
@@ -153,4 +153,5 @@ def list_requests(src_rse, dst_rse, states, issuer):
     if not permission.has_permission(issuer=issuer, action='list_requests', kwargs=kwargs):
         raise exception.AccessDenied('%(issuer)s cannot list requests from RSE %(src_rse)s to RSE %(dst_rse)s' % locals())
 
-    return request.list_requests(src_rse_id, dst_rse_id, states)
+    for req in request.list_requests(src_rse_id, dst_rse_id, states):
+        yield api_update_return_dict(req)
--- a/lib/rucio/api/request.py
+++ b/lib/rucio/api/request.py
@@ -135,3 +135,22 @@ def get_request_by_did(scope, name, rse, issuer):
     req = request.get_request_by_did(scope, name, rse_id)
 
     return api_update_return_dict(req)
+
+
+def list_requests(src_rse, dst_rse, states, issuer):
+    """
+    List all requests in a specific state from a source RSE to a destination RSE.
+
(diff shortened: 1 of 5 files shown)


lib/rucio/api/request.py
```diff
@@ -135,3 +135,22 @@ def get_request_by_did(scope, name, rse, issuer):
     req = request.get_request_by_did(scope, name, rse_id)
 
     return api_update_return_dict(req)
+
+
+def list_requests(src_rse, dst_rse, state, issuer):
```
mlassnig: can you make state --> states ? so we can ask for a list of request states?

hahahannes: ah yes, of course

lib/rucio/api/request.py
```diff
@@ -135,3 +135,22 @@ def get_request_by_did(scope, name, rse, issuer):
...
+    :param issuer: Issuing account as a string.
+    """
+    src_rse_id = get_rse_id(rse=src_rse)
+    dst_rse_id = get_rse_id(rse=dst_rse)
+
+    kwargs = {'src_rse_id': src_rse_id, 'dst_rse_id': dst_rse_id, 'issuer': issuer}
+    if not permission.has_permission(issuer=issuer, action='list_requests', kwargs=kwargs):
+        raise exception.AccessDenied('%(issuer)s cannot list requests from RSE %(src_rse)s to RSE %(dst_rse)s' % locals())
+
+    return request.list_requests(src_rse_id, dst_rse_id, states)
```
bari12: For Multi-VO Rucio you have to do something similar to
https://github.com/rucio/rucio/blob/491263e035db97a84dde783b11f5111692b2c3cb/lib/rucio/api/did.py#L62
Otherwise the scope and account would not be correct for Multi-VO

bari12 (changes requested):

mlassnig (approved):

bari12 (approved):

PR #2936, opened 2019-10-01
docker: remove old demo

mlassnig:
docker: remove old demo


doc/source/rucio_demo.rst
```diff
@@ -1,259 +0,0 @@
-..  Copyright 2018 CERN for the benefit of the ATLAS collaboration.
```
bari12, line 1: Aren't we loosing all of this info below? Thus even the demo steps etc. 
I think it is worth to keep this, just linking it to the dev container though.

mlassnig: All the demo steps are obsolete and partially not applicable at all anymore, and there's a script in the dev container that does all of that and more.

And the commands are in the tutorial docs as well.

bari12 (approved):

PR #2948, opened 2019-10-03
RSE Manager: Turn double slashes in identity algorithm into a single slash. 

ericvaandering:
This is especially useful for CEPH which does not treat extra slashes as a single.

Pull request title
------------------

Valid component names are listed in the `label list <https://github.com/rucio/rucio/labels>`_.


mlassnig:
Ah sorry, we can't do that generally, since root requires double slashes in some cases. 

I'm actually preparing a pull request right at this moment, which solves it through passing the correct gfal option.

PR #2949, opened 2019-10-03
RSE Manager: Turn double slashes in identity algorithm into a single slash. 

ericvaandering:
This is especially useful for CEPH which does not treat extra slashes as a single.''

### DIFF
--- a/lib/rucio/rse/protocols/protocol.py
+++ b/lib/rucio/rse/protocols/protocol.py
@@ -35,6 +35,7 @@ along with some of the default methods for LFN2PFN translations.
 """
 
 import hashlib
+import re
 
 try:
     # PY2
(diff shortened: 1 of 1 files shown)


KatyEllis:
Just to elaborate on why RAL Echo storage struggles with this - 

Ceph Echo is an object store, so the part of the file name we normally call the path is actually just part of the object's name. If you enter the file with a double slash in the PFN, when you come to retrieve it, the double slash has to be there. This would be inconvenient.

mlassnig:
Yes, I understand, we are on it with Tim. We are having the same trouble in ATLAS. The problem is fixed by https://github.com/rucio/rucio/pull/2950 which does not try to mangle paths anymore.

On the other hand, this patch will remove doubleslashes for everyone and everywhere, which is not good ;-)

bari12:
👍 With #2950 merged, I would suggest we close this. @ericvaandering is this fine for you?

PR #2953, opened 2019-10-04
Clients: return error info in traces Fix #2925

### LINKED ISSUE #2925: Better error propagation for download and upload client
Motivation
----------

Pilots using rucio mover do not report useful information on why a transfer failed. This is because on failure the download and upload client throw an exception with no information inside. The real reason for the error should be propagated better back to the caller of the client.

Modification
------------

The dowloadclient and uploadclient should return a dictionary of file status (containing appropriate error information) instead of raising an exception.

davidgcameron:
Clients: remove unused imports Fix #2925


mlassnig:
I've restarted the travis build for Oracle. This one is prone to crashing due to whatever reasons...

bari12 (approved):

PR #2971, opened 2019-10-10
More readable version of CMS permissions file

ericvaandering:
Not clear if there is a problem with the other one or if something odd was going on. This one verified to work.


bari12:
Hi @ericvaandering 
You are always opening two different branches to make the two PR. While in principle this is fine, it is sufficient (for a patch) to just make a branch based on `master` and then make the PR against both `master` and `next`, so you don't have to do it twice. If there are merge conflicts, I will resolve them during merging.

ericvaandering:
I thought I could. I tried that and and the resulting PR to next seemed to changing 100 different files. I must be doing something wrong, of course. I wanted to ask you about that this week,.

Issue #2990, opened 2019-10-15
Add a docker-compose that brings services up only listening on localhost


bjwhite-fnal:
Motivation
----------
The current development docker-compose makes services listen to all incoming IP addresses, which is incompatible with security policies at some locations.

Modification
------------
Provide a separate docker-compose file that listens only to the loopback address.

PR #2991, opened 2019-10-15
Testing: Provide a localhost-only docker-compose file #2990

bjwhite-fnal:
Testing: Provide a localhost-only docker-compose file #2990

### DIFF
--- a/etc/docker/dev/docker-compose-storage-localhost.yml
+++ b/etc/docker/dev/docker-compose-storage-localhost.yml
@@ -1,0 +1,81 @@
+version: "2"
+services:
+  rucio:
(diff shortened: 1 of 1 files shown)


etc/docker/dev/docker-compose-localhost.yml
```diff
@@ -0,0 +1,81 @@
+version: "2"
```
bari12: We should probably rename this to `docker-compose-storage-localhost.yml`
@tbeerman changed this in the recent PR, so it would reflect that change as well.

bari12 (approved):

PR #3001, opened 2019-10-16
Add optional monitoring stack to docker dev environment

tbeerman:
Add optional monitoring stack to docker dev environment


etc/docker/dev/docker-compose-storage-monit.yml
```diff
@@ -0,0 +1,128 @@
...
+      - ../../../lib:/opt/rucio/lib
+    environment:
+      - X509_USER_CERT=/opt/rucio/etc/usercert.pem
+      - X509_USER_KEY=/opt/rucio/etc/userkey.pem
+  ruciodb:
+    image: postgres:11
+    environment:
+      - POSTGRES_USER=rucio
+      - POSTGRES_DB=rucio
+      - POSTGRES_PASSWORD=secret
```
vingar, line 29: Putting DB data on volume is better in case of restart / kill containers. It can avoid to recreate it. e.g. 
```
volumes:
      - pg_data:/var/lib/postgresql/data
volumes:
  pg_data:
    external: true
  mysql_data:
    external: true
```

etc/docker/dev/docker-compose-storage.yml
```diff
@@ -0,0 +1,81 @@
...
+      - POSTGRES_PASSWORD=secret
+    ports:
+      - "5432:5432"
+    command: ["-c", "fsync=off","-c", "synchronous_commit=off","-c", "full_page_writes=off"]
+  graphite:
+    image: graphiteapp/graphite-statsd
+    ports:
+      - "80:80"
+  fts:
+    image: rucio/fts
```
vingar, line 37: Maybe using the docker images provided by FTS instead/ see https://gitlab.cern.ch/fts/fts3/tree/develop/packaging/docker
```
image: gitlab-registry.cern.ch/fts/fts3:latest
image: gitlab-registry.cern.ch/fts/fts-rest:latest
gitlab-registry.cern.ch/fts/fts-monitoring:latest
```

vingar (approved):

vingar:
huuu Is there a chance to get answers even on non blocking comments/questions ?

etc/docker/dev/docker-compose-storage.yml
mlassnig: yeah, we cooked up our own because the gitlab registry at CERN was having some problems with outside connections. i think that's much better now, so most likely i'll migrate that soon

etc/docker/dev/docker-compose-storage-monit.yml
mlassnig: we had a couple of cases where changes to the database schema were screwing with the dev environment if we didn't start with a clean slate. most likely the best approach would be a separate docker-compose that explicitly puts this on a volume, so people can use that when they want.

PR #3016, opened 2019-10-17
client: Change requests.session to requests.Session Fixes #3011

### LINKED ISSUE #3011: session() replaced by Session() in requests
Noticed this going through the requests documentation

ericvaandering:
.session is deprecated

### DIFF
--- a/lib/rucio/client/baseclient.py
+++ b/lib/rucio/client/baseclient.py
@@ -315,7 +315,7 @@ class BaseClient(object):
                 continue
 
             if result is not None and result.status_code == codes.unauthorized:  # pylint: disable-msg=E1101
-                self.session = session()
+                self.session = Session()
                 self.__get_token()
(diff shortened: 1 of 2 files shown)


lib/rucio/client/baseclient.py
```diff
@@ -126,7 +126,7 @@ def __init__(self, rucio_host=None, auth_host=None, account=None, ca_cert=None,
         self.host = rucio_host
         self.list_hosts = []
         self.auth_host = auth_host
-        self.session = session()
+        self.session = Session()
```
bari12, line 129: On line 318 this has to be changed as well.

bari12 (changes requested):

lib/rucio/client/baseclient.py
ericvaandering: Done. I must have hit undo at some point...

bari12 (approved):

PR #3017, opened 2019-10-17
Core & Internals bug: Modify hasing algorithm for PostgreSQL and MySQL. Fix #2993

### LINKED ISSUE #2993: Thread workload hashing is incorrect for MySQL and PostgreSQL
Motivation
----------
For both the MySQL and PostgreSQL database types, the hashing algorithm used to partition worker thread responsibilities has the divisor of the modulo operation being incorrectly computed, erroneously using total_threads-1 instead of the raw value of total threads. This has the consequence of preventing some threads from ever being assigned work. 

Modification
------------
I will provide a fix for this. I intend to create a function in db/sqla/util.py to do this, as this pattern for workload partitioning is repeated 21 times throughout the codebase.

bjwhite-fnal:
Core & Internals bug: Modify hasing algorithm for PostgreSQL and MySQL. Fix #2993

### DIFF
--- a/etc/docker/dev/docker-compose-storage-localhost.yml
+++ b/etc/docker/dev/docker-compose-storage-localhost.yml
@@ -1,0 +1,81 @@
+version: "2"
+services:
+  rucio:
+    image: rucio/rucio-dev
+    hostname: rucio
+    ports:
+      - "127.0.0.1:443:443"
+    links:
+      - ruciodb:ruciodb
+      - graphite:graphite
+      - fts:fts
+      - ftsdb:ftsdb
+      - xrd1:xrd1
+      - xrd2:xrd2
+      - xrd3:xrd3
+    volumes:
+      - ../../../tools:/opt/rucio/tools
+      - ../../../bin:/opt/rucio/bin
+      - ../../../lib:/opt/rucio/lib
+    environment:
+      - X509_USER_CERT=/opt/rucio/etc/usercert.pem
+      - X509_USER_KEY=/opt/rucio/etc/userkey.pem
+  ruciodb:
+    image: postgres:11
+    environment:
+      - POSTGRES_USER=rucio
+      - POSTGRES_DB=rucio
+      - POSTGRES_PASSWORD=secret
+    ports:
+      - "127.0.0.1:5432:5432"
+    command: ["-c", "fsync=off","-c", "synchronous_commit=off","-c", "full_page_writes=off"]
+  graphite:
+    image: graphiteapp/graphite-statsd
+    ports:
+      - "127.0.0.1:80:80"
+  fts:
+    image: rucio/fts
+    hostname: fts
+    ports:
+      - "127.0.0.1:8446:8446"
+      - "127.0.0.1:8449:8449"
+  ftsdb:
+    image: mysql:5
+    hostname: ftsdb
+    environment:
+      - MYSQL_USER=fts
+      - MYSQL_PASSWORD=fts
+      - MYSQL_ROOT_PASSWORD=fts
+      - MYSQL_DATABASE=fts
+    ports:
+      - "127.0.0.1:3306:3306"
+  xrd1:
+    image: rucio/xrootd
+    hostname: xrd1
+    environment:
+      - XRDPORT=1094
+    ports:
+      - "127.0.0.1:1094:1094"
+    volumes:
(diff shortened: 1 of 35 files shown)


etc/docker/dev/docker-compose-storage-localhost.yml
```diff
@@ -0,0 +1,81 @@
...
+      - X509_USER_CERT=/opt/rucio/etc/usercert.pem
+      - X509_USER_KEY=/opt/rucio/etc/userkey.pem
+  ruciodb:
+    image: postgres:11
+    environment:
+      - POSTGRES_USER=rucio
+      - POSTGRES_DB=rucio
+      - POSTGRES_PASSWORD=secret
+    ports:
+      - "127.0.0.1:5432:5432"
```
vingar, line 30: "5432:5432" should be enough

vingar (dismissed):
looks good. One minor and non blocking comment.

vingar:
I meant : looks good after having fixed the flake8 errors

lib/rucio/db/sqla/__init__.py
```diff
@@ -3,3 +3,27 @@
...
+            if not hash_variable:
+                query = query.filter(text('mod(md5(id), %s) = %s' % (total_threads, thread)))
+            else:
+                query = query.filter(text('mod(md5(%s), %s) = %s' % (hash_variable, total_workers, worker_number)))
+        elif session.bind.dialect.name == 'postgresql':
+            if not hash_variable:
+                query = query.filter(text('mod(abs((\'x\'||md5(id::text))::bit(32)::int), %s) = %s' % (total_threads, thread)))
+            else:
+                query = query.filter(text('mod(abs((\'x\'||md5(%s::text))::bit(32)::int), %s) = %s' % (hash_variable, total_workers, worker_number)))
+    return query
```
bari12: Hi @bjwhite-fnal 
This looks very good to me. Instead of `__init__` I would add this to the `sqla/util.py` though.

In the other files I see there was big confusion between `total_workers + 1` and `total_workers -1` etc.
There is one remaining issue which would be great if you can check. In the daemons calling these core functions you changed we usually do `total_workers=heartbeat['nr_threads'] - 1` and `worker_number=heartbeat['assign_thread']` - Please check if this is done in all daemons consistently or if we have some +/-1 mixup there as well.

Jimbat: My first thought was also to include this definition in `util.py`. I tried that first, however it created a cyclic importing issue with another file. `util.py` wanted to import a module that ended up using `filter_thread_work()`, which had not been defined yet. There are no forward declarations in Python, so I moved it to `__init__.py` as opposed to creating a new module.

I'll check on the rest of that for you.

bjwhite-fnal: My first thought was also to include this definition in `util.py`. I tried that first, however it created a cyclic importing issue with another file. `util.py` wanted to import a module that ended up using `filter_thread_work()`, which had not been defined yet. There are no forward declarations in Python, so I moved it to `__init__.py` as opposed to creating a new module.

I'll check on the rest of that for you.

etc/docker/dev/docker-compose-storage-localhost.yml
Jimbat: That will listen on 0.0.0.0. By specifying the bind address `docker-compose-storage-localhost.yml` can be used to deploy a Rucio instance that listens only on the loopback.

bjwhite-fnal: That will listen on 0.0.0.0. By specifying the bind address `docker-compose-storage-localhost.yml` can be used to deploy a Rucio instance that listens only on the loopback.

bjwhite-fnal:
So all of the tests currently pass, and my simple checks of various functionality all looked ok, but this probably needs some additional careful review as I made changes to quite a few spots. However @bari12, I believe that I was able to smooth out the consistency issues across the daemons. In the future, it should be possible to filter the assigned work for each thread simply by passing the raw values of heartbeat['nr_threads'] and hearbeat['assign_thread'] directly to the hashing function.

bjwhite-fnal:
Looks like the Travis builds found a couple spots I missed. I'll keep cleaning this up.

bari12:
It's running now: https://travis-ci.org/rucio/rucio/builds/608398241

Issue #2911, opened 2019-09-16
Allow RSEs to be re-enabled


arisfkiaras:
Motivation
----------
Right now there are methods to 'soft' delete an RSE and disable it but no methods to bring it back. 
Wouldn't it make sense to allow for an RSE to be enabled again without doing manual operations on the database? This might be useful for synchronization with CRIC https://github.com/rucio/rucio/issues/2896

Modification
------------
Add method that sets up deleted RSE as active again. 
Would a CLI command `rucio-admin rse recover` make sense?

bari12:
To me that makes sense, but @cserf is mostly an advocate of not making these commands too accessibly. 😄 
But I think you can go ahead and implement that.

PR #3043, opened 2019-10-28
Core & Internals: Make importer configurable Fix #2896 #2911

### LINKED ISSUE #2896: Importer fails to import RSEs that were deleted
Motivation
----------
I was experimenting with the Rucio-CRIC integration and while doing so, I was disabling and re enabling RSEs.
What I noticed is that when the importer tries to configure an RSE that already exists on the database but is marked as 'Deleted', it will throw an error: 
`Details: RSE with id 'f9c7592cdc6344eabce00e8ea4d2641a' cannot be found`

After further investigation I found out that when the importer sees an RSE on the json it will check if it already exists to avoid creating it again. It does so using the `rse_module.get_rse_id` method. This method though returns the id even if the RSE is already marked as deleted.
https://github.com/rucio/rucio/blob/9388a53bf2b0b9b82b0727d897ef0ab818ade60f/lib/rucio/core/importer.py#L37
Further down on the code that the rse_id (of the deleted RSE) is referenced, of course it throws an error.

Modification
------------
Possible  solution: Check if the RSE is marked as 'deleted' and enable it again before trying to modify it.

arisfkiaras:
Core & Internals: Make importer configurable Fix #2896 #2911
------------------
After @bari12 request I made the importer configurable on a fine grained basis.
The current available sync modes are: hard, edit, append.
Those sync modes can be configured on an RSE, Attributes, Protocol level.

On hard mode:
- new RSE/Attributes/Protocols are created
- RSE/Attributes/Protocols that appear modified on the JSON, are modified on Rucio as well
- RSE/Attributes/Protocols that exist on Rucio but not on the JSON, are deleted

On edit mode:
- new RSE/Attributes/Protocols are created
- RSE/Attributes/Protocols that appear modified on the JSON, are modified on Rucio as well
- RSE/Attributes/Protocols that exist on Rucio but not on the JSON, are **NOT** deleted

On append mode:
- new RSE/Attributes/Protocols are created
- RSE/Attributes/Protocols that appear modified on the JSON, are **NOT** modified on Rucio
- RSE/Attributes/Protocols that exist on Rucio but not on the JSON, are **NOT** deleted

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -41,6 +41,7 @@ Individual contributors to the source code
 - Boris Bauermeister <Boris.Bauermeister@gmail.com> 2019
 - Ruturaj Gujar <ruturaj.gujar23@gmail.com> 2019
 - Andrew Lister <andrew.lister@stfc.ac.uk>, 2019
+- Aristeidis Fkiaras <aristeidis.fkiaras@cern.ch>, 2019
 
 Organisations employing contributors
 ------------------------------------
--- a/lib/rucio/core/importer.py
+++ b/lib/rucio/core/importer.py
@@ -15,35 +15,48 @@
 # Authors:
 # - Hannes Hansen <hannes.jakob.hansen@cern.ch>, 2018-2019
 # - Andrew Lister, <andrew.lister@stfc.ac.uk>, 2019
+# - Aristeidis Fkiaras <aristeidis.fkiaras@cern.ch>, 2019
 #
 # PY3K COMPATIBLE
 
 from six import string_types
-from rucio.common.exception import RSEOperationNotSupported, RSENotFound
+from rucio.common.exception import RSEOperationNotSupported
 from rucio.core import rse as rse_module, distance as distance_module, account as account_module, identity as identity_module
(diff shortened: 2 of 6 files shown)


bari12:
I think the rse_expression parser is failing because in the hard mode you are removing some RSEs which were created for test purposes and the test expects to be there. In the rse_expression parser tests we just have to make sure that these still exist and possibly re-create them.

arisfkiaras:
Thanks for the comment.
I gave a try enabling all the RSEs and another one deleting the ones my test is creating and rse_expression parser is still giving that error.

From the looks of it, its my attribute hard sync method that is causing the issue. I guess for some RSEs I might be removing attributes that are 'mandatory'.

I will change the default attribute sync method to a softer one, find a way to test the hard sync without affecting rse_expression_parser, run the full test suite again locally and push to the branch.

bari12:
I think we can also just re-write the rse_expression parser test so that the test just creates the RSEs it needs.

arisfkiaras:
Opening new PR to be submitted as patch.
#3104 #3105 
Closing this one.

PR #3064, opened 2019-11-01
Transfers: Multihop support; Fix #2632

### LINKED ISSUE #2632: Generic Multi-Hop transfers
Motivation
----------
For CTA RSEs the conveyor needs to submit multihop transfers.

bari12:
Transfers: Multihop support; Fix #2632

### DIFF
--- a/lib/rucio/core/replica.py
+++ b/lib/rucio/core/replica.py
@@ -1753,13 +1753,14 @@ def get_sum_count_being_deleted(rse_id, session=None):
 
 
 @transactional_session
-def update_replicas_states(replicas, nowait=False, session=None):
+def update_replicas_states(replicas, nowait=False, add_tombstone=False, session=None):
     """
     Update File replica information and state.
 
-    :param replicas: The list of replicas.
-    :param nowait:   Nowait parameter for the for_update queries.
-    :param session:  The database session in use.
+    :param replicas:        The list of replicas.
+    :param nowait:          Nowait parameter for the for_update queries.
+    :param add_tombstone:   To set a tombstone in case there is no lock on the replica.
+    :param session:         The database session in use.
     """
 
     for replica in replicas:
@@ -1790,7 +1791,7 @@ def update_replicas_states(replicas, nowait=False, session=None):
         elif replica['state'] == ReplicaState.AVAILABLE:
             rucio.core.lock.successful_transfer(scope=replica['scope'], name=replica['name'], rse_id=replica['rse_id'], nowait=nowait, session=session)
             # If No locks we set a tombstone in the future
-            if lock_cnt == 0:
+            if add_tombstone and lock_cnt == 0:
                 set_tombstone(rse_id=replica['rse_id'], scope=replica['scope'], name=replica['name'], tombstone=datetime.utcnow() + timedelta(hours=2), session=session)
 
         elif replica['state'] == ReplicaState.UNAVAILABLE:
@@ -1800,7 +1801,7 @@ def update_replicas_states(replicas, nowait=False, session=None):
                                             broken_message=replica.get('broken_message', None),
                                             nowait=nowait, session=session)
             # If No locks we set a tombstone in the future
-            if lock_cnt == 0:
+            if add_tombstone and lock_cnt == 0:
                 set_tombstone(rse_id=replica['rse_id'], scope=replica['scope'], name=replica['name'], tombstone=datetime.utcnow() + timedelta(hours=2), session=session)
         elif replica['state'] == ReplicaState.TEMPORARY_UNAVAILABLE:
             query = query.filter(or_(models.RSEFileAssociation.state == ReplicaState.AVAILABLE, models.RSEFileAssociation.state == ReplicaState.TEMPORARY_UNAVAILABLE))
--- a/lib/rucio/core/replica.py
+++ b/lib/rucio/core/replica.py
@@ -1764,9 +1764,13 @@ def update_replicas_states(replicas, nowait=False, session=None):
 
     for replica in replicas:
         query = session.query(models.RSEFileAssociation).filter_by(rse_id=replica['rse_id'], scope=replica['scope'], name=replica['name'])
+        lock_cnt = 0
         try:
             if nowait:
-                query.with_for_update(nowait=True).one()
+                rep = query.with_for_update(nowait=True).one()
+            else:
+                rep = query.one()
+            lock_cnt = rep.lock_cnt
         except NoResultFound:
             # remember scope, name and rse
             raise exception.ReplicaNotFound("No row found for scope: %s name: %s rse: %s" % (replica['scope'], replica['name'], get_rse_name(replica['rse_id'], session=session)))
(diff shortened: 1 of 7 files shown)


lib/rucio/core/replica.py
```diff
@@ -1753,20 +1753,25 @@ def get_sum_count_being_deleted(rse_id, session=None):
...
-    :param session:  The database session in use.
+    :param replicas:        The list of replicas.
+    :param nowait:          Nowait parameter for the for_update queries.
+    :param add_tombstone:   To set a tombstone in case there is no lock on the replica.
+    :param session:         The database session in use.
     """
 
     for replica in replicas:
         query = session.query(models.RSEFileAssociation).filter_by(rse_id=replica['rse_id'], scope=replica['scope'], name=replica['name'])
+        lock_cnt = 0
```
bari12, line 1768: That's not needed, since the code would either reach L1774 or raise

lib/rucio/core/replica.py
```diff
@@ -1753,20 +1753,25 @@ def get_sum_count_being_deleted(rse_id, session=None):
 
 
 @transactional_session
-def update_replicas_states(replicas, nowait=False, session=None):
+def update_replicas_states(replicas, nowait=False, add_tombstone=False, session=None):
```
vingar, line 1756: I would suggest to have`'tombstone=None` instead in the signature. If not set (==None), then it should be like the `add_tombstone=False` behaviour. If set, use this value for the tombstone instead ,e.g., `tombstone=datetime.utcnow() + timedelta(hours=2)`.

lib/rucio/core/replica.py
```diff
@@ -1785,12 +1790,19 @@ def update_replicas_states(replicas, nowait=False, session=None):
             values['tombstone'] = OBSOLETE
         elif replica['state'] == ReplicaState.AVAILABLE:
             rucio.core.lock.successful_transfer(scope=replica['scope'], name=replica['name'], rse_id=replica['rse_id'], nowait=nowait, session=session)
+            # If No locks we set a tombstone in the future
```
vingar, line 1793: why ? Is there a need to have an intermediate replica protection, here for 2 hours ?

lib/rucio/core/replica.py
```diff
@@ -1785,12 +1790,19 @@ def update_replicas_states(replicas, nowait=False, session=None):
...
+            if add_tombstone and lock_cnt == 0:
+                set_tombstone(rse_id=replica['rse_id'], scope=replica['scope'], name=replica['name'], tombstone=datetime.utcnow() + timedelta(hours=2), session=session)
+
         elif replica['state'] == ReplicaState.UNAVAILABLE:
             rucio.core.lock.failed_transfer(scope=replica['scope'], name=replica['name'], rse_id=replica['rse_id'],
                                             error_message=replica.get('error_message', None),
                                             broken_rule_id=replica.get('broken_rule_id', None),
                                             broken_message=replica.get('broken_message', None),
                                             nowait=nowait, session=session)
+            # If No locks we set a tombstone in the future
```
vingar, line 1803: Same comment than before.

lib/rucio/core/replica.py
```diff
@@ -2934,19 +2946,20 @@ def get_suspicious_files(rse_expression, **kwargs):
 
 
 @transactional_session
-def set_tombstone(rse_id, scope, name, session=None):
+def set_tombstone(rse_id, scope, name, tombstone=OBSOLETE, session=None):
```
vingar, line 2949: non blocking: maybe consider in the future a bulk method for performance.

lib/rucio/core/replica.py
```diff
@@ -2934,19 +2946,20 @@ def get_suspicious_files(rse_expression, **kwargs):
...
     """
     Sets a tombstone on a replica.
 
     :param rse_id: ID of RSE.
     :param scope: scope of the replica DID.
     :param name: name of the replica DID.
+    :param tombstone: the tombstone to set. Default is OBSOLETE
     :param session: database session in use.
     """
     conn = get_engine().connect()
```
vingar, line 2959: not part of this patch but `conn = get_engine().connect()` ? It should be handled by the `transactional_session` decorator ? this can potentially lead to connection leak or abnormal transaction behaviour.

bari12: Yes, that part looks strange indeed. If the engine object is needed, we should try to resolve this from the session directly. I will add a separate issue for this.

lib/rucio/core/replica.py
bari12: Sounds reasonable. We can add this in a later patch.

lib/rucio/core/replica.py
bari12: It's just added protection. The point is we need to add a tombstone since the replica would otherwise stay forever, since it has no rule. An immediate tombstone sounded too aggressive though, since the submission machinery could take too much time. In theory, the replica is protected by the sources table FK as well, but it's just added protection, since we know for sure that we do not have to delete this replica right away.

lib/rucio/core/replica.py
bari12: #3071

lib/rucio/core/replica.py
vingar: Maybe getting this value from the internal config table instead would give some flexibility there. Default can be 2 hours if not defined.

lib/rucio/daemons/conveyor/common.py
```diff
@@ -322,38 +326,58 @@ def bulk_group_transfer(transfers, policy='rule', group_bulk=200, source_strateg
...
-                    grouped_transfers[external_host][job_key][policy_key]['files'].append(t_file)
+                    if multihop:
+                        # The parent transfer should be the first of the list
+                        # TODO : Only work for a single hop now, need to be able to handle multiple hops
+                        if transfer['parent_request']:  # This is the child
+                            grouped_transfers[external_host][job_key][policy_key]['files'].append(t_file)
+                        else:
+                            grouped_transfers[external_host][job_key][policy_key]['files'].insert(0, t_file)
+                    else:
+                        grouped_transfers[external_host][job_key][policy_key]['files'].append(t_file)
```
rcarpa, line 369: can somebody explain why this "else" code block doesn't exist as an "else" at line 381 ? Why don't we want to append the file if `USER_TRANSFERS == 'cms' and activity in USER_ACTIVITY` ?

rcarpa: @cserf

bari12: I think it should be there, but was forgotten.
The `USER_ACTIVITY` code is basically unused though. So it probably was never noticed. That's something we should address in the re-factor. Also all the `USER_TRANSFERS == 'cms' ` parts are not used as far as I know (Maybe @ericvaandering can confirm) - So also something we should re-factor.

ericvaandering:
Correct. We explored this and abandoned it. It should be removed to simplify the code.

lib/rucio/daemons/conveyor/common.py
ericvaandering: Correct. We explored this and abandoned it. It should be removed to simplify the code.

Issue #3074, opened 2019-11-06
Remove explicit references to config.yml in globusLibrary.py


maatthias:
Motivation
----------
Currently to transfer files via Globus Online the globusLibrary.py makes explicit path references to config.yml to retrieve client_id and refresh_token. 
client_id = cfg['globus']['apps']['SDK Tutorial App']['client_id']
refresh_token = cfg['globus']['apps']['SDK Tutorial App']['refresh_token']

Modification
------------
Retrieve globus_auth_app name in rucio.cfg conveyor property.
client_id = cfg['globus']['apps'][GLOBUS_AUTH_APP]['client_id']
refresh_token = cfg['globus']['apps'][GLOBUS_AUTH_APP]['refresh_token']

PR #3077, opened 2019-11-06
Core & Internals: added globus_auth_app #3074

maatthias:


bari12:
Hi Matt, sorry for the late reply on this. Can you please close this and re-submit as a patch. We had the 1.21 release in the meanwhile and it would be good to add this to the current release line already, not waiting for the next feature release. Thanks!

maatthias:
Hi Martin- not sure how I missed this!  Sure I will close and re-submit as patch.

PR #3104, opened 2019-11-21
make importer configurable fix #2896

### LINKED ISSUE #2896: Importer fails to import RSEs that were deleted

arisfkiaras:
make importer configurable fix #2896

### DIFF
--- a/lib/rucio/core/importer.py
+++ b/lib/rucio/core/importer.py
@@ -15,35 +15,48 @@
 # Authors:
 # - Hannes Hansen <hannes.jakob.hansen@cern.ch>, 2018-2019
 # - Andrew Lister, <andrew.lister@stfc.ac.uk>, 2019
+# - Aristeidis Fkiaras <aristeidis.fkiaras@cern.ch>, 2019
 #
 # PY3K COMPATIBLE
 
 from six import string_types
-from rucio.common.exception import RSEOperationNotSupported, RSENotFound
+from rucio.common.exception import RSEOperationNotSupported
 from rucio.core import rse as rse_module, distance as distance_module, account as account_module, identity as identity_module
 from rucio.db.sqla import models
 from rucio.db.sqla.constants import RSEType, AccountType, IdentityType
 from rucio.db.sqla.session import transactional_session
+from rucio.common.config import config_get
 
 
 @transactional_session
-def import_rses(rses, session=None):
+def import_rses(rses, rse_sync_method='edit', attr_sync_method='edit', protocol_sync_method='edit', session=None):
     new_rses = []
     for rse_name in rses:
         rse = rses[rse_name]
         if isinstance(rse.get('rse_type'), string_types):
             rse['rse_type'] = RSEType.from_string(str(rse['rse_type']))
-        try:
+
+        if rse_module.rse_exists(rse_name, include_deleted=False, session=session):
+            # RSE exists and is active
             rse_id = rse_module.get_rse_id(rse=rse_name, session=session)
-        except RSENotFound:
+            rse_module.update_rse(rse_id=rse_id, parameters=rse, session=session)
+        elif rse_module.rse_exists(rse_name, include_deleted=True, session=session):
+            # RSE exists but in deleted state
+            # Should only modify the RSE if importer is configured for edit or hard sync
+            if rse_sync_method in ['edit', 'hard']:
+                rse_id = rse_module.get_rse_id(rse=rse_name, include_deleted=True, session=session)
+                rse_module.restore_rse(rse_id, session=session)
+                rse_module.update_rse(rse_id=rse_id, parameters=rse, session=session)
+            else:
+                # Config is in RSE append only mode, should not modify the disabled RSE
+                continue
+        else:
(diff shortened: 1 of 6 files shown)


lib/rucio/core/importer.py
```diff
@@ -15,35 +15,48 @@
...
 from rucio.core import rse as rse_module, distance as distance_module, account as account_module, identity as identity_module
 from rucio.db.sqla import models
 from rucio.db.sqla.constants import RSEType, AccountType, IdentityType
 from rucio.db.sqla.session import transactional_session
+from rucio.common.config import config_get
 
 
 @transactional_session
-def import_rses(rses, session=None):
+def import_rses(rses, session=None, rse_sync_method='edit', attr_sync_method='edit', protocol_sync_method='edit'):
```
bari12: `session` should be the last parameter in the list

lib/rucio/core/rse.py
```diff
@@ -119,7 +120,7 @@ def add_rse(rse, deterministic=True, volatile=False, city=None, region_code=None
 
 
 @read_session
-def rse_exists(rse, session=None):
+def rse_exists(rse, session=None, include_deleted=False):
```
bari12: `session` should be the last parameter in the list

lib/rucio/core/importer.py
```diff
@@ -15,35 +15,48 @@
...
     new_rses = []
     for rse_name in rses:
         rse = rses[rse_name]
         if isinstance(rse.get('rse_type'), string_types):
             rse['rse_type'] = RSEType.from_string(str(rse['rse_type']))
-        try:
+
+        if rse_module.rse_exists(rse_name, include_deleted=False):
```
bari12: session has to be passed through, otherwise a new transaction is started

lib/rucio/core/importer.py
```diff
@@ -15,35 +15,48 @@
...
+            # RSE exists and is active
             rse_id = rse_module.get_rse_id(rse=rse_name, session=session)
-        except RSENotFound:
+            rse_module.update_rse(rse_id=rse_id, parameters=rse, session=session)
+        elif rse_module.rse_exists(rse_name, include_deleted=True):
```
bari12: session has to be passed through

lib/rucio/core/importer.py
```diff
@@ -15,35 +15,48 @@
...
+            # RSE exists but in deleted state
+            # Should only modify the RSE if importer is configured for edit or hard sync
+            if rse_sync_method in ['edit', 'hard']:
+                rse_id = rse_module.get_rse_id(rse=rse_name, session=session, include_deleted=True)
```
bari12: session should be last parameter (readability)

lib/rucio/core/importer.py
```diff
@@ -15,35 +15,48 @@
...
+                rse_module.restore_rse(rse_id)
```
bari12: session has to be passed through

lib/rucio/core/importer.py
```diff
@@ -201,10 +229,13 @@ def import_data(data, session=None):
...
     """
-    # RSEs
+    rse_sync_method = config_get('importer', 'rse_sync_method', False, 'edit')
+    attr_sync_method = config_get('importer', 'attr_sync_method', False, 'edit')
+    protocol_sync_method = config_get('importer', 'rse_sync_method', False, 'edit')
+
     rses = data.get('rses')
     if rses:
-        import_rses(rses, session=session)
+        import_rses(rses, session=session, rse_sync_method=rse_sync_method, attr_sync_method=attr_sync_method, protocol_sync_method=protocol_sync_method)
```
bari12: session should be last parameter (readability)

bari12 (changes requested):

lib/rucio/db/sqla/models.py
```diff
@@ -665,6 +665,12 @@ class RSE(BASE, SoftModelBase):
                    CheckConstraint('RSE IS NOT NULL', name='RSES_RSE__NN'),
                    CheckConstraint('RSE_TYPE IS NOT NULL', name='RSES_TYPE_NN'),)
 
+    def restore(self, session=None):
```
bari12: Sorry for the late comment @arisfkiaras 
Please move that one to the `restore_rse` method directly. (so just to change `.deleted` and `.deleted_at` and then save)
I don't want to put (more) methods into the models directly.

bari12 (approved):

Issue #3124, opened 2019-11-26
Add support for prometheus in core.monitor


tbeerman:
Motivation
----------
Currently, the monitor module only supports statsd but the CERN Kubernetes clusters come with a Prometheus server that can also be used for auto-scaling.

Modification
------------
Make the monitor module configurable to either report to statd or to prometheus.

PR #3145, opened 2019-11-28
Monitoring: add some prometheus metrics to the daemons #3124

tbeerman:
Monitoring: add some prometheus metrics to the daemons #3124


bari12:
Since this introduces a new dependency, I would suggest to close the `master` part of the PR and we just merge against `next`, so this will be included in the coming feature release.

PR #3146, opened 2019-11-28
Monitoring: add some prometheus metrics to the daemons #3124

tbeerman:
Monitoring: add some prometheus metrics to the daemons #3124


bari12:
Needs to be rebased;

Issue #3084, opened 2019-11-08
Moving upload method from rsemgr to client


TomasJavurek:
Motivation
----------
Some usless duplicate functionality. Rather keeping in the Client.

Modification
------------
Keeping basic functinalities, but:
1) removing the file loop
2) duplicate check for existence of a file
3) renaming - we can do better, duplicate code
4) adding proper exceptions
5) input validation

TomasJavurek:
Tests performed on the latest version. All tests have been performed on CERN-PROD_SCRATCHDISK and UNI-FREIBURG_SCRATCHDISK
1) upload with registration in several attempts
2) upload without registration to the same rses
3) upload of existing file that is not registered
4) upload of a file that is already registered
5) upload of non-existing file
All these tests performed well. Although it is not all we should be checking, e.g. uploading corrupted file, uploading to not-existing rse etc.

TomasJavurek:
Following the comments from @TWAtGH I dropped the summary string form the NotAllFilesUploaded exception - #3158 

The suggestion from Tobi is to add attributes directly to the exception:
https://github.com/rucio/rucio/blob/master/lib/rucio/common/exception.py#L847
whether as a dict or standalone attributes corresponding to the quantities in the summary, needs to be discussed and decided.

TomasJavurek:
Just noticed misbehaviour:
`[root@rucio-daemon-int-02 javor]# rucio -v upload --lifetime 1000 --scope test test_upload_1 --rse UNI-FREIBURG_SCRATCHDISK
2020-06-23 11:38:45,435	DEBUG	Num. of files that upload client is processing: 1
2020-06-23 11:38:45,549	DEBUG	Input validation done.
2020-06-23 11:38:45,549	INFO	Preparing upload for file test_upload_1
2020-06-23 11:38:45,584	DEBUG	wan is used for the upload
2020-06-23 11:38:45,584	DEBUG	Registering file
2020-06-23 11:38:45,585	DEBUG	Skipping dataset registration
2020-06-23 11:38:45,598	DEBUG	File DID does not exist
2020-06-23 11:38:45,662	ERROR	An unknown exception occurred.
Details: [u'(cx_Oracle.DatabaseError) ORA-14400: inserted partition key does not map to any partition']`

TomasJavurek:
There should be some proper exception if the scope does not exist.

TomasJavurek:
otherwise, tested on 10 different sites on the grid.

[...]

PR #3158, opened 2019-11-29
clients: upload from rsemgr to uploadclient - #3084

TomasJavurek:
clients: upload from rsemgr to uploadclient - #3084

### DIFF
--- a/lib/rucio/client/uploadclient.py
+++ b/lib/rucio/client/uploadclient.py
@@ -32,12 +32,15 @@ import socket
 
 import logging
 import time
+import collections
 
 from rucio.client.client import Client
-from rucio.common.exception import (RucioException, RSEBlacklisted, DataIdentifierAlreadyExists,
-                                    DataIdentifierNotFound, NoFilesUploaded, NotAllFilesUploaded,
+from rucio.common.config import config_get_int
+from rucio.common.exception import (RucioException, RSEBlacklisted, RSEChecksumUnavailable, RSEOperationNotSupported,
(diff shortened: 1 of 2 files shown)


bari12:
@TWAtGH please have a look as well.

lib/rucio/client/uploadclient.py
```diff
@@ -289,9 +300,9 @@ def upload(self, items, summary_file_path=None, traces_copy_out=None):
                 json.dump(final_summary, summary_file, sort_keys=True, indent=1)
 
         if num_succeeded == 0:
-            raise NoFilesUploaded()
+            raise NoFilesUploaded(str(collections.Counter(summary_exception)))
```
TWAtGH, line 321: Could we omit these changes for now and think again about how we pass these error information back?

TWAtGH (changes requested):

lib/rucio/client/uploadclient.py
TomasJavurek: Well. I did already two versions. We greed on previous one, we agreed on this one. I am not going to apply another already third one.

TWAtGH: I didn't agree ^^ Back then I started doing my own version based on your commit but I wasn't able to finish it and now I also don't have time to finish it.

TomasJavurek:
Since Exceptions are using kwargs, I thought I could do just easily:
 raise NotAllFilesUploaded(summary=summary_exception)
and the summary becomes attribute of the exception. But I got:
TypeError: must be type, not classobj

bari12:
Closing this since it has been resubmitted

Issue #2630, opened 2019-06-06
rucio.cfg vs config table


bari12:
Motivation
----------
Automatic layering of rucio.cfg vs. config table. On server/daemon all config lookups first go to rucio.cfg, if not found to the config table. This way common configuration can be overwritten with local configurations.

On client just rucio.cfg will be used.

bari12:
[Google doc](https://docs.google.com/document/d/1ZJwmP971SGo1X3lS8gbJEMrpgQU0FzEUiJF01yK68MA/edit#heading=h.3lnruokyb4eg) to follow this activity.

PR #3179, opened 2019-12-10
Reaper : Move to snake_case for RSE attributes for total and used space : Closes #3178

### LINKED ISSUE #3178: Move to snake_case for RSE attributes for total and used space
Motivation
----------

After discussion, it looks like many people prefer snake_case instead of CamelCase for RSE attributes for total and used space. That's obviously a wrong choice similar to choosing emacs instead of vi (my 2 satoshis), but since the majority decides, let it be...

cserf:
Reaper : Move to snake_case for RSE attributes for total and used space : Closes #3178

### DIFF
--- a/lib/rucio/daemons/reaper/reaper2.py
+++ b/lib/rucio/daemons/reaper/reaper2.py
@@ -276,8 +276,8 @@ def __check_rse_usage(rse, rse_id, prepend_str):
 
         # Check from which sources to get used and total spaces
         # Default is storage
-        source_for_total_space = attributes.get('sourceForTotalSpace', 'storage')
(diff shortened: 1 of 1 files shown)


vingar:
It might be worth to extend the patch with an alembic script duplicating the attributes 
like sourceForTotalSpace -> source_for_total_space for a smooth migration.

snake_case format can also be enforced by a formal validation  like regexp.

bari12:
We should collect all these parameters in #2630 and then change the mismatched ones with an alembic script as you described.
For this one, since this is reaper2 prototype I will merge this through. The old one (for reaper1) won't stop working.

bari12 (approved):

Issue #3257, opened 2020-01-24
PanDA prefers different return from list_dataset_replicas_vp


ivukotic:
Motivation
----------

when VP replica is not created the list_dataset_replicas_vp should not return anything. Currently it returns regular replicas.

Modification
------------
It should not return anything if no VP replicas.

PR #3259, opened 2020-01-24
now VP call does not return anything if no VP replica created.

ivukotic:
replica: change in VP return #3257
------------------

It also returns nothing if the only regular replicas are on TAPE.


bari12:
To me it looks good like this. @mlassnig should have a second look as the master-of-(list)-replicas though

mlassnig (approved):
this is fine, approving

ivukotic:
Great. Thanks. So when can I expect to see this change in production?

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

bari12:
This will be part of the patch release 1.21.7 next week and go to integration testing on Tuesday next week. Production deployment is the week after.

Issue #3076, opened 2019-11-06
Globus transfers fail when fts_hosts is empty string


maatthias:
Motivation
----------
Error is thrown when fts_hosts is empty string.  The 'fts' attribute is empty for Globus transfers resulting in an error.  

Modification
------------
I made 2 changes to fix here but cannot sure if they're ideal for FTS transfers.
https://github.com/maatthias/rucio/blob/a86cc1080da03c065f360b0105ed401898f49088/lib/rucio/core/transfer.py#L764

PR #3332, opened 2020-02-27
Core & Internals: check for empty fts_hosts #3076

maatthias:
Core & Internals: check for empty fts_hosts #3076


maatthias:
I will re-submit these pull requests after running through Travis.

Issue #3346, opened 2020-02-28
Adding list and exists methods to GlobusRSEProtocol


maatthias:
Motivation
----------
Allow client to query globus storage via rse protocol wrapper 

Modification
------------
Adding list and exists methods to GlobusRSEProtocol

PR #3351, opened 2020-03-03
Transfers: Adding list and exists methods to GlobusRSEProtocol #3346

maatthias:
Transfers: Adding list and exists methods to GlobusRSEProtocol #3346


bari12:
Reminder for the syntax error; By now will probably need a rebase as well 😄

Issue #3350, opened 2020-03-03
Move config.yml to ./etc directory with other rucio configuration files


maatthias:
Motivation
----------
Move the config.yml file from its current place at ./lib/rucio/transfertool/ to ./etc where the other Rucio configuration files are stored. 

Modification
------------
Modify the code and documentation that references the location of config.yml:
./lib/rucio/transfertool/globusLibrary.py
./doc/source/configure_globus_transfertool.rst

PR #3353, opened 2020-03-03
Transfers: move globus config file #3350

maatthias:
Transfers: move globus config file #3350


lib/rucio/transfertool/globusLibrary.py
```diff
@@ -21,7 +21,7 @@
 import os
```
mlassnig, line 21: nitpick comment:

we generally don't use camelCase. any strong opinion to keep it like that?

maatthias: Yes let's change that.  In fact, this separate globusLibrary.py is really a relic from when I was trying to do this outside Rucio.  These methods could probably be rolled into ./lib/rucio/transfertool/globus.py  What do you think?  (And it's rarely nitpicky when it's correct.)

bari12: The globus library is a bit special, since you are also using it from the globus protocol. Since the client distribution does include protocols, but not the `transfertool/` tree, the library should probably be put to a common place. E.g. `lib/rucio/common/`

bari12:
Needs rebase and move of the `globusLibrary.py` 😄

PR #3403, opened 2020-03-20
Clients: Load config specified by environment or argument Fix #3377

### LINKED ISSUE #3377: Specify client configuration file via env var and/or CLI option
Motivation
----------
The current ways of specifying a client config file are not very flexible and don't make it easy to use two different rucio systems from the same machine.

Modification
------------
Use an environment variable like `RUCIO_CFG` to specify a config file, and add a command line option like `-c CONFFILE`.

bziemons:
Clients: Load config specified by environment or argument Fix #3377

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -42,6 +42,7 @@ Individual contributors to the source code
 - Ruturaj Gujar <ruturaj.gujar23@gmail.com> 2019
 - Andrew Lister <andrew.lister@stfc.ac.uk>, 2019
 - Aristeidis Fkiaras <aristeidis.fkiaras@cern.ch>, 2019
+- Benedikt Ziemons <benedikt.ziemons@cern.ch>, 2020
 
 Organisations employing contributors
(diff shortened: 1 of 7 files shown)


bziemons:
This is a draft, command line argument parsing is not yet implemented. A few questions/notes:

* Parsing the configuration file in an cmd argument would require doing that before parsing any other arguments, because other argument defaults use `config_get` while parsing. Is that in scope? It would be added first in main: https://github.com/rucio/rucio/blob/next/bin/rucio#L2252
* Using `RUCIO_CONFIG` should in my opinion stop the search for any other configuration files - supplying a non-existing config should not cause the program to just load another config.
* Using `__CONFIGFILES` in https://github.com/rucio/rucio/blob/next/lib/rucio/daemons/auditor/srmdumps.py#L23 looks like a bad idea to me. Shouldn't this be refactored?

bari12:
Please also resubmit as a patch.

bari12 (approved):

PR #3407, opened 2020-03-23
Testing: Convert pre-commit hook to Python 3; Fix #3405

### LINKED ISSUE #3405: Supplied git pre-commit hook is using Python 2
Motivation
----------
The CONTRIBUTING document [describes to set-up git hooks](https://github.com/rucio/rucio/blob/next/CONTRIBUTING.rst#git-hooks), while the `pre-commit` is using Python 2.

Modification
------------
Either:
* Use `#!/usr/bin/env python2`

or (considering Python 3 should already be installed in a dev env):
* Port it to Python 3 (quite the easy task) and use `#!/usr/bin/env python3`

bziemons:
Testing: Convert pre-commit hook to Python 3; Fix #3405

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -42,6 +42,7 @@ Individual contributors to the source code
 - Ruturaj Gujar <ruturaj.gujar23@gmail.com> 2019
 - Andrew Lister <andrew.lister@stfc.ac.uk>, 2019
 - Aristeidis Fkiaras <aristeidis.fkiaras@cern.ch>, 2019
+- Benedikt Ziemons <benedikt.ziemons@cern.ch>, 2020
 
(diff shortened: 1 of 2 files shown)


bari12:
I think we should still allow python2 for this. Not everyone is using the dev containers and (at the moment) we still do support python 2.7.

bziemons:
This is just the git-hook that is used with git-commit. This doesn't affect the containers and only runs on the developer's machine. I think it is safe to say that everyone should have Python 3 at least installed side-by-side, but it's your call.

bari12:
Also re-submit as a patch please. 👍

bziemons:
@bari12 done, see #3483

bari12 (approved):

Issue #3383, opened 2020-03-12
Selection page for authentication method if not specified in the rucio.cfg file


gumond:
Motivation
----------
Currently Rucio WebUI does not offer a user the possibility to choose how they authenticate before authorized to use the WebUI. The WebUI authentication method has to be pre-defined in the rucio.cfg file. 

Modification
------------
changes in code under rucio/lib/rucio/web/ui/

gumond:
+ new login method selection page was introduced for cases when Rucio servers would not configure default login method
+ methods (except SAML) tested also for the usual case when Rucio server has default auth_type provided
+ multiple OIDC providers are supported dynamically on the login method selection page
+ added an option to specify Rucio account for all authentication methods
+ saving last accessed path into cookie for 2 minutes in case of invalid token which allows to come back to the previous page in case of successful authentication (goes to root path if not found)
+ tried to make the messy code shorter, more understandable, less buggy and added docs. 

SAML method was not tested for CERN SSO (neither in the implementation before these changes), is there some documentation about it somewhere so that I could test it ?

gumond:
replaced by a patch #3428, closing this issue.

PR #3414, opened 2020-03-26
Feature 3383 rucio web ui auth method selection

gumond:
Authentitcation & Authorization: WeubUI Auth method selection page #3383


gumond:
Canceling request as it has been replaced by #3429

Issue #3410, opened 2020-03-25
Fixes for multi IdP support for OAuth2/OIDC


gumond:
Motivation
----------
From first local tests, there are few fixes necessary to make client_credential flow work with multiple issuers. This is due to the selection of admin account in order to request Rucio app admin token (e.g. for FTS transfers) which is currently random with hard-coded account names. This needs to change to a choice of an account which does have the Rucio app admin OIDC identity instead. In the near future, there might be other fixes coming up relevant for TPC testing Rucio rule -> FTS -> dCache. 

Modification
------------
lib/rucio/core/oidc.py

gumond:
- the minimal scope and audience claims expected in the token by Rucio in order to proceed for token request via Rucio client/WebUI or in order to validate a token by Rucio, were made fully configurable on the server side
- the request scope and audience for token exchange or for getting Rucio admin token for submitting transfer to FTS were made fully configurable on the server side
- adding checks on minimal scope and audience also to the initialisation of the request
- adapting correspondingly the bin/rucio and bin/rucio-admin files as well as web/ui/common/utils.py (!!! this 2 line change in the latter has to be made correspondingly when merging the recent WebUI login pull request !!!)
- removing hard coded admin account names for admin token requests and changing the mechanism the account and identity is found for admin/user for the purpose of authentication the FTS transfer (rucio/core/transfer.py)

gumond:
replaced by a patch #3428, closing this issue.

PR #3421, opened 2020-03-27
OAuth2/OIDC: fixes to the release 1.22.0 and more tests #3410

gumond:


gumond:
Canceling request as it has been replaced by #3429

Issue #3096, opened 2019-11-14
Unification of metadata interfaces


bari12:
Motivation
----------
Currently there are several interfaces to address metadata in Rucio. This should be unified into one interface which intelligently selects which backend to use. (Currently: Hardcoded columns or JSON columns) In the future this enables us to extend Rucio to external metadata catalogs which are accessed via the unified frontend.

PR #3424, opened 2020-03-30
Core & Internals: Unification of metadata interfaces #3096

arisfkiaras:
Core & Internals: Unification of metadata interfaces #3096
------------------

Opening merge request so we can discuss also here. 

The branch I have put on this merge request is a clean version of https://github.com/arisfkiaras/rucio/tree/feature-3096-Unification_of_metadata_interfaces
Unfortunately when I tried to run the testing suite on it, the AUTH issue on the dev container appeared again, so we might see a couple of errors in the travisCI checks. Before the cleanup all tests were passing.

As a first step I think we should discuss the Documentation: doc/source/did_meta.rst and see if this is the functionality we want.

Then I would like to discuss the usage of DIDKey table which is a requirement that come after our initial discussion.

Then I have some technical questions were input from senior rucio devs would be helpful.


bari12:
Hi Aris,

Sorry for the late reply but here some points in non-prioritized order 😄 

- The loading of the metadata plugins is good, we should more strongly align this to the policy modules maybe, so they can be added there as well
- I don't think we should have the switch for `fixed_metadata`; There should be a list of super_hardcoded fields which always go to the column plugin (Or maybe let's even handle this in the core directly). You can't even turn this off. This should be `lifetime, guid, events, adler32, bytes, length, md5, deleted_at, transient, accessed_at, closed_at, eol_at, is_archive, constituent, access_cnt` Maybe I am forgetting some. But these are metadata values core to rucio, it shouldn't be possible to turn them off. Of course there are some other metadata column which are somewhat optional `phys_group, provenance, panda_id` which are in the dids table. I think it should be even fine if a VO think of adding a further column to the schema. But these ones should be handled by the column plugin.
- I don't think there should be a `generic_metadata_fixed_keys` list or any kind of key list in the cfg. This should be derived from the list stored in the plugins. So each plugin defines (can be overwritten by the VO) which metadata keys it is in charge of (Similar to the `HARDCODED_KEYS` you definied in did_meta/hardcoded. It should probably be allowed to use wildcards. E.g. `*` would mean that this metadata module takes care of any key not defined by any other plugin. This means when storing a metadata value, you need to load the plugins, check the keys each one is in charge of and then decide to which one to relay the metadata.
- The only option in the cfg really needed is a list of enabled metadata modules I think. Again, if none are in the list, the super_hardcoded metadata columns from above should still be used.
- The keys defined by rucio-admin add_key etc. should still be enforced on top of the modules. Thus we first check if a key/value validates these schemas, if yes then we pick the plugin to store it to.

arisfkiaras:
Follow up of a meeting with Martin:
* part of the plugin loading can be moved to the policy module
* each plugin should publish the list of keys it is responsible to, should be able to use wildcards
* the dir/module should be renamed to something indicating more that the things inside are 'plugins'
* the meta.py can stay were it was

bari12:
Hi Aris,

Here a few comments to the latest changes.
- Please add an Abstract Class/Interface which the different plugins implement - so it is easier for somebody making a new plugin.
- The hardcoded plugin we should rename to `didscolumns` (or something like this)
- In core/did.py we can remove the add/get/delete_did_meta methods. We should completely unify this to set_meta get_meta (and probably a new delete_meta) call
- With the list_dids I am struggling a bit. We really have to check that this is done correctly, but in principle I think you also did the right thing of unifying it. We should probably also remove the `list_dids_by_meta` call, since it can be fully done by `list_dids`.

Overall looks very good and basically what I had in mind as well.

arisfkiaras:
While rebasing to latest upstream/next realized there was an addition of a related metadata API call.
https://github.com/rucio/rucio/pull/3678

If I move it as well to the plugin mechanism it might break what @cserf developed it for.
After we have verified it won't break anything, it might make sense to move this method in the plugin mechanism after the release as a patch.

As we discussed, kept list-dids as it was and created a list-dids-extended call.

bari12:
I agree, lets put the change from #3678 to the plugin later. But it definitely should go there eventually.

bari12 (approved):

PR #2862, opened 2019-08-22
Infrastructure: Add MinIO objectstore to dev Docker #2855

jamesp-epcc:
This pull request adds a MinIO objectstore to the development `docker-compose.yml`. I have tested this and it works with the signed URL support in Rucio (with minor change to include the port number in the endpoint URL).

Should there be documentation somewhere to let people know how to set it up? I didn't see any for the existing XRootD servers so I haven't added anything yet.

Issue #2855, opened 2019-08-21
Add objectstore to development Docker image


jamesp-epcc:
It would be desirable to add an S3-compatible objectstore (based on https://min.io/) to the development `docker-compose`, in addition to the 3 Xrootd servers currently included.

PR #3440, opened 2020-04-03
Infrastructure: Restore min.io objectstore to dev docker #2855

jamesp-epcc:
The min.io objectstore that I added to the development docker-compose in PR #2862 seems to have been lost at some point, probably when the docker-compose files were rearranged. This PR adds it back.

### DIFF
--- a/etc/docker/dev/docker-compose-storage-localhost.yml
+++ b/etc/docker/dev/docker-compose-storage-localhost.yml
@@ -13,6 +13,7 @@ services:
       - xrd1:xrd1
       - xrd2:xrd2
       - xrd3:xrd3
+      - minio:minio
     volumes:
       - ../../../tools:/opt/rucio/tools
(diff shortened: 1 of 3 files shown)


tbeerman (approved):

bari12:
Hi James, can you maybe repeat this one was a patch. It's more an enhancement than a feature and otherwise we have to wait for June for this to make it to master.

jamesp-epcc:
> Hi James, can you maybe repeat this one was a patch. It's more an enhancement than a feature and otherwise we have to wait for June for this to make it to master.

Sure, I will submit a PR against master as well (that's what I need to do to make this a patch, right?).

bari12:
Technically you should do a new PR with a branch based of master, otherwise you will pull changes over from next to master. (Right now this is not the case, since they just recently merged, but in principle that should be done :-)

Issue #3441, opened 2020-04-03
Cannot access ActiveMQ Web Console


viveknigam3003:
Motivation
----------
While using the Rucio Storage with Monitoring docker development environment, ActiveMQ is server is available on port `- 61613:61613` but [http://localhost:61613](http://localhost:61613) ends in Error 404.

```BASH=
74b81515c543    webcenter/activemq:latest       "/app/run.sh"    59 seconds ago    Up 50 seconds       1883/tcp, 5672/tcp, 8161/tcp, 61614/tcp, 61616/tcp, 0.0.0.0:61613->61613/tcp
```

Modification
------------

PR #3442, opened 2020-04-06
Messaging: Added Port for ActiveMQ in docker-compose file #3441

viveknigam3003:
Messaging: Added Port for ActiveMQ in docker-compose file #3441


tbeerman (dismissed):
Remove the .vscode file. The rest looks good.

PR #3444, opened 2020-04-06
Infrastructure: Restore min.io objectstore to dev docker #2855

jamesp-epcc:
The min.io objectstore that I added to the development docker-compose in PR #2862 seems to have been lost at some point, probably when the docker-compose files were rearranged. This PR adds it back.


bari12:
I cherry-picked this one into master. Closing

PR #3475, opened 2020-04-15
Release: Fix Python3 setup.py file; Fix #3347

### LINKED ISSUE #3347: setup.py for main rucio package is not Python 3 compatible
Motivation
----------
The `setup.py` for the main Rucio package is not Python 3 compatible. 

Modification
------------
- [x] Fix compatibility
- [x] Add Python3 Test for server to travis

bari12:
Release: Fix Python3 setup.py file; Fix #3347

### DIFF
--- a/setup_rucio.py
+++ b/setup_rucio.py
@@ -15,7 +15,7 @@
 # Authors:
 # - Vincent Garonne <vgaronne@gmail.com>, 2014-2018
 # - Thomas Beermann <thomas.beermann@cern.ch>, 2014-2018
-# - Martin Barisits <martin.barisits@cern.ch>, 2016-2019
+# - Martin Barisits <martin.barisits@cern.ch>, 2016-2020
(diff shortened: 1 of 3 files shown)


setup_rucio.py
```diff
@@ -82,13 +82,13 @@ def run_git_command(cmd):
 
 if using_pip and using_git:
     git_version_cmd = '''git describe --dirty=-dev`date +%s`'''
-    pkg_version = run_git_command(git_version_cmd).decode()
+    pkg_version = run_git_command(git_version_cmd)
```
vingar: It should be better to add the `decode()` call directly in the `run_git_command` function.

bari12: I don't think it is needed at all. It was removed from the `tools/generate_version.py` and it runs fine. In principle I think this if branch can be removed completely, since we always create the `vcsversion.py` file by hand anyway.

vingar: right, I overlooked the change and think it was the other way around.. brrrrrr

PR #3483, opened 2020-04-15
Testing: Convert pre-commit hook to Python 3; Fix #3405

### LINKED ISSUE #3405: Supplied git pre-commit hook is using Python 2

bziemons:
Testing: Convert pre-commit hook to Python 3; Fix #3405

### DIFF
--- a/tools/pre-commit
+++ b/tools/pre-commit
@@ -25,9 +25,10 @@ def main():
     if currentbranch == 'master' or currentbranch == 'next' or currentbranch == 'hotfix':
         print("You are trying to commit on your master/next/hotfix, that's probably a mistake. Exiting...")
         sys.exit(1)
-    modified = re.compile(r'^[AM]+\s+(?P<name>.*\.py)', re.MULTILINE)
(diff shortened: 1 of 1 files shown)


bziemons:
Some error when connecting to the IPv6 of www.sqlite.org: https://travis-ci.org/github/rucio/rucio/jobs/675442926#L1426. The log reads

```
curl: (7) Failed to connect to 2600:3c00::f03c:91ff:fe96:b959: Cannot assign requested address
```

I think it would be better if `tools/travis/install_script.sh` exits with non-zero when the docker build fails.

bari12 (approved):

PR #3538, opened 2020-04-27
Replicas registration: removing adler32 requirement in add_replica methods #3494

### LINKED ISSUE #3494: adler32 checksum should be optional
Motivation
----------
A community may decide to use md5 checksums instead of adler32, but using the generic schema, giving a None adler32 to the replica API is not allowed:

```
Rucio exception: Provided object does not match schema.
Details: Problem validating dids : None is not of type 'string'

Failed validating 'type' in schema['items']['properties']['adler32']:
    {'description': 'adler32',
     'pattern': '^[a-fA-F\\d]{8}$',
     'type': 'string'}

On instance[0]['adler32']:
    None
```

Modification
------------

adler32 checksums should be optional at the schema level and also in the client API:

`def add_replica(self, rse, scope, name, bytes, adler32, pfn=None, md5=None, meta={}):`

gabrielefronze:
Replicas registration: removing adler32 requirement in add_replica methods #3494

### DIFF
--- a/lib/rucio/client/replicaclient.py
+++ b/lib/rucio/client/replicaclient.py
@@ -184,11 +184,11 @@ class ReplicaClient(BaseClient):
         :return: True if files were created successfully.
 
         """
-        dict = {'scope': scope, 'name': name, 'bytes': bytes, 'meta': meta}
+        dictio = {'scope': scope, 'name': name, 'bytes': bytes, 'meta': meta}
 
-        set_replica_checksums(dict, md5=md5, adler32=adler32, pfn=pfn)
+        set_replica_checksums(dictio, md5=md5, adler32=adler32, pfn=pfn)
 
-        return self.add_replicas(rse=rse, files=[dict])
+        return self.add_replicas(rse=rse, files=[dictio])
 
     def add_replicas(self, rse, files, ignore_availability=True):
         """
--- a/lib/rucio/client/replicaclient.py
+++ b/lib/rucio/client/replicaclient.py
@@ -36,6 +36,7 @@ from requests.status_codes import codes
 from rucio.client.baseclient import BaseClient
 from rucio.client.baseclient import choice
 from rucio.common.utils import build_url, render_json
+from rucio.common.utils import set_replica_checksums
 
 
 class ReplicaClient(BaseClient):
@@ -167,7 +168,7 @@ class ReplicaClient(BaseClient):
         exc_cls, exc_msg = self._get_exception(headers=r.headers, status_code=r.status_code, data=r.content)
         raise exc_cls(exc_msg)
 
-    def add_replica(self, rse, scope, name, bytes, adler32, pfn=None, md5=None, meta={}):
+    def add_replica(self, rse, scope, name, bytes, adler32=None, pfn=None, md5=None, meta={}):
         """
         Add file replicas to a RSE.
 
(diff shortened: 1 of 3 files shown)


lib/rucio/client/replicaclient.py
```diff
@@ -167,7 +168,7 @@ def list_replicas(self, dids, schemes=None, unavailable=False,
         exc_cls, exc_msg = self._get_exception(headers=r.headers, status_code=r.status_code, data=r.content)
         raise exc_cls(exc_msg)
 
-    def add_replica(self, rse, scope, name, bytes, adler32, pfn=None, md5=None, meta={}):
+    def add_replica(self, rse, scope, name, bytes, adler32=None, pfn=None, md5=None, meta={}):
```
bari12, line 171: I think in principle this is fine now, however, it would allow adding replicas without any checksum at all. 
Currently the schema validation would allow that.
https://github.com/rucio/rucio/blob/6df1b039e5158ef77c80f121d3eed7283c7d757c/lib/rucio/api/replica.py#L202
It does call the schema validation there, but since it compares with `dids` where `adler32` is not mandatory (due to datasets/containers) we can end up also with files without checksums.
@davidgcameron @mlassnig @cserf @dchristidis would be good to check this too. I am not sure if we want to allow that, I think there needs to be at least an option to require one, or the other, checksum.

davidgcameron: Can the requirement be specified via a community-specific policy or configuration? For ATLAS checksums should be enforced but I can imagine other cases where it should be optional, like registering files residing on a storage which doesn't support checksums, mutable files, or pseudo-files like say a dynamic web page. I think we should be flexible and not restrict the way rucio can be (ab)used by any community :)

bari12: I think we should do that. There is a few possibilities, one of them would be to place it in the `schema` (Which can already be community specific). The difficulty with that is that we would have to change this quite a bit though, since for now you could not just make the `adler32` field mandatory. (That part of the schema is validated for datasets/containers too, and would fail them, since they do not have checksums)
Another option is to put this into rucio.cfg or config table. Something like `required_checksum`.
I am a little bit more leaning to the `schema` direction, since this makes this work natively also for multi-vo mode, where different VOs might have different requirements.

gabrielefronze: I would like to add few cents.
I'm not sure, but can you confirm making one checksum mandatory might limit the functionalities of per-RSE checksum compatibility?
In that case enforcing adler32 will throw away compatibility with, e.g., legacy gridftp sites (the triggering reason for the per-RSE compatibility).
For this reason I'd tend to the solution which would check all the provided values and verify that there is at least one, more than having a given algorithm set up statically as VO o deployment setting.
If this is not the case, then the per VO setting seems to me the best one, while having to interact with the schema is in mi mind not (read never) ideal.

mlassnig: I was not following this closely, but just to throw my few cents into the discussion.

There should be a way to express the following things:
- "every file must have a checksum, doesn't matter which one"
- "every file must have a checksum, and it must be expliclity this one"
- "no checksums are needed"
- "no checksums are allowed on special RSEs"
- "no checksums are allowed on special files"

I guess a combination of both approaches makes sense here. A schema validation for the global "decision", and then more specific attributes to express the more fine-grained selections.

bari12:
Closing, followed up in @robbarnsley PR

bari12:
Closing; Currently re-discussing a different approach

Issue #3472, opened 2020-04-14
StoRM - archives and logging.


TomasJavurek:
Motivation
----------
1) When a constituent of archive is an input file for the download, xrootd is used as of today. But we need to develop a procedure for StoRM.

2) logging needs to be introduced 

Modification
------------

TomasJavurek:
Just an internal remark:
logging is not used even in the rsemgr. I need to figure out which logger object can be used in the storm protocol:
`loggers = [logging.getLogger(name) for name in logging.root.manager.loggerDict]`

PR #3539, opened 2020-04-28
storm, gfal, downloadclient: adding logging - #3472

TomasJavurek:
storm, gfal, downloadclient: adding logging - #3472

### DIFF
--- a/lib/rucio/client/downloadclient.py
+++ b/lib/rucio/client/downloadclient.py
@@ -544,8 +544,8 @@ class DownloadClient:
             logger.info('%sTrying to download with %s from %s: %s ' % (log_prefix, scheme, rse_name, did_str))
 
             try:
-                protocol = rsemgr.create_protocol(rse, operation='read', scheme=scheme, auth_token=self.auth_token)
-                protocol.connect(logger=logger)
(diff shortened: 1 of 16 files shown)


mlassnig (commented):
this looks ok to me. second patch will come with actual logging messages or will you update this one?

@TWAtGH eyes pls

TomasJavurek:
> this looks ok to me. second patch will come with actual logging messages or will you update this one?

We will add the logging messages in the individual protocols when they will be needed. For now, I added just one to the storm, that serves as a check that it really gets into the pilot log. This was requested by @davidgcameron

davidgcameron:
Please add the check that the link exists after it was created, and feel free to be as verbose as possible in these logs. For example:

Creating link from x -> y
Created link from x-> y
Checking link x exists
Output of stat(x)
Link exists ok

TomasJavurek:
> Please add the check that the link exists after it was created, and feel free to be as verbose as possible in these logs. For example:
> 
> Creating link from x -> y
> Created link from x-> y
> Checking link x exists
> Output of stat(x)
> Link exists ok

Right. I would do it step by step. Let's firstly get the logs into the protocols. Checking the link is different issue/suggestion. A good one, but I'll make separate ticket for that.

bari12 (approved):

Issue #3552, opened 2020-05-01
Mock transfer tool


nsmith-:
Motivation
----------
I want to simulate a rucio installation without the need to run FTS and have physical files moving around (even if only within docker containers).

Modification
------------
Implement `mock` transfer tool to always succeed transfers.

Possible future improvement: tool could delay finishing based on file size, to simulate finite network bandwidth.

PR #3553, opened 2020-05-01
Transfers: implement a mock transfertool that always succeeds #3552

nsmith-:
Probably for the long-term it would be good to better abstract the transfertool, this had to touch more spots than one would expect.


nsmith-:
Fixes #3552

bari12 (approved):

PR #3573, opened 2020-05-11
Core: add/inject_rule uses apply_rule, Fixes #3571

### LINKED ISSUE #3571: use sharded apply_rule in add/inject_rule
this (partly) addresses #3418 by limiting the amount of memory used during the transaction (usage is 20-30 KB per file!)

lgoossen:
Core: add/inject_rule uses apply_rule, Fixes #3571

### DIFF
--- a/lib/rucio/core/rule.py
+++ b/lib/rucio/core/rule.py
@@ -1,4 +1,4 @@
-# Copyright 2012-2018 CERN for the benefit of the ATLAS collaboration.
+# Copyright 2012-2020 CERN for the benefit of the ATLAS collaboration.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


lgoossen:
first test round revealed indentation bug in rule_grouping.apply_rule
(code to update usage counters was indented 1 level too much and hence not executed for individual files)

mlassnig (dismissed):
This looks OK to me, but I can't really test this. @bari12 ?

bari12:
We are still working on the testing and deployment scheme of this; It would be quite risky to just put it into the code like this.

lgoossen:
modified implementation allowing both old and new rule engine algorithm to be used
configuration switch is rules:use_new_rule_algorithm [False by default]
to allow nose tests to test both algorithms without config tampering, a module var USE_NEW_RULE_ALGORITHM
[False by default] can also be used to force the choice

bari12 (approved):

Issue #3542, opened 2020-04-29
Policy packages should support multi-VO Rucio installations


jamesp-epcc:
The policy package code needs work to make it compatible with multi-VO Rucio installations. Because each VO will likely want a different policy package, we need to be able to load multiple packages simultaneously and ensure that any access to the policy package accesses the appropriate one.

Loading multiple policy packages will mean allowing a separate package per VO to be specified (in `rucio.cfg` or elsewhere?) and importing the schema and permissions modules from the packages into separate namespaces. Ensuring that all usage of the packages goes to the correct VO's package will be more involved. Policy packages typically contain up to four elements (permissions module, schema module, lfn2pfn functions and surl functions) and these are discussed in detail below.

## Permissions Module

Policy packages contain a permissions module, the contents of which are loaded into the `rucio.core.permission` namespace. All accesses to this module go through a single function, `has_permission` which takes an `issuer` parameter identifying a user account. A central implementation of `has_permission` could be provided which will determine which VO this user belongs to and dispatch the call to the appropriate policy package, allowing code that calls this function to remain unchanged. Alternatively, if the user account name is not sufficient to determine the active VO, a separate VO parameter can be added.

## Schema Module

Similar to the permissions module, the contents of the policy package schema module are loaded into the `rucio.common.schema` namespace. However, the schema module is more complex in that it contains multiple items that are accessed externally. The `validate_schema` function is called from numerous places within Rucio, similar to the `has_permission` function. A central implementation should again be provided to dispatch to the appropriate policy package implementation.

However, this time none of the function's existing parameters will allow us to determine the VO, so an extra parameter will need to be added to specify the VO.

Some data items in the schema module are also directly accessed from elsewhere in the code. These are `ACTIVITY` (used in the conveyer daemon), `SCOPE_LENGTH` (used in various database related code), `SCOPE_NAME_REGEXP` (used by the REST interface), and `NAME_LENGTH` (used in database related code). In some cases (for example, `SCOPE_LENGTH` and `NAME_LENGTH`) it makes no sense to have a per-VO version of these values since they are fixed at the database level, so they should not be accessed from individual policy packages at all.

In other cases (for example, `SCOPE_NAME_REGEXP`) it does make sense to have a different value for each VO, so a getter function should be provided to get the value from the policy package for a given VO, and this should be called instead of the value being accessed directly from the schema module.

## lfn2pfn and surl functions

[...]

PR #3591, opened 2020-05-14
Core & Internals: prepare policy package code for multi-VO #3542

jamesp-epcc:
This PR prepares the groundwork for making policy packages work with multi-VO Rucio. Instead of loading the schema and permission modules from the policy packages directly into the relevant namespaces, it instead stores them in dictionaries so that we can later load a separate module per VO. All access to them now goes through centralised functions that pass through to the correct policy package. Getter functions have been added for schema data since it is no longer possible to access it directly from outside the module in this model.

Please leave the issue open as there is still more work to be done before this is fully functional.

### DIFF
--- a/bin/rucio-cache-client
+++ b/bin/rucio-cache-client
@@ -37,7 +37,7 @@ from rucio.client.didclient import DIDClient
 from rucio.client.rseclient import RSEClient
 from rucio.common.config import config_get, config_get_int
 from rucio.common.exception import DataIdentifierNotFound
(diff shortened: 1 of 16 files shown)


jamesp-epcc:
I am not sure why Github thinks these commits were made by James Joaquin... they were definitely made by me!

jamesp-epcc:
I'm closing this PR now and submitting an updated one with a single getter method for the schema and the git log problems fixed.

PR #3598, opened 2020-05-14
Core & Internals: Fix py3 VO table initialisation

elichad:
In progress, as discussed with Martin.

This initial commit gets past the alembic table initialisation, but there are a number of test failures which I don't know the cause of so I'm not sure if it's an appropriate fix or not.


bari12:
I think this fix is good 👍
Reading https://docs.sqlalchemy.org/en/13/core/type_basics.html#sqlalchemy.types.String.params.convert_unicode
the issue seems to be that the option `"force"` we used just does not make any sense under Python 3. Even in Python 2 I wonder if `True` would not have been the better setting instead of forcing the SQLAlchemy internal conversion. My guess is that it just does not make any difference and Force was set without a special reason.
My assessmnet would be that everything will just run as is with the `True` setting, but some more eyes on this would be good. @mlassnig @cserf

mlassnig (approved):
this is correct. should have never been "force" in the first place ;-)

Issue #3623, opened 2020-05-21
monitor.py fails with non-default metrics_port


astroclark:
Motivation
----------
Enabling graphite monitors with a non-default `metrics_port` causes all daemons and the server to immediately crash when they call `start_http_server`.

The input to `start_http_server` in [monitor.py](https://github.com/rucio/rucio/blob/master/lib/rucio/core/monitor.py#L38) should be of type `int` but is [parsed](https://github.com/rucio/rucio/blob/master/lib/rucio/core/monitor.py#L37) from `rucio.cfg` as a string.

Modification
------------
Change `config_get` -> `config_get_int` when the configuration is [read](https://github.com/rucio/rucio/blob/master/lib/rucio/core/monitor.py#L37)

astroclark:
Sorry, minor addition worth mentioning here: 
```
ENABLE_METRICS = config_get..
```
Should probably be:
```
ENABLE_METRICS = config_get_bool
```
Otherwise setting `metrics_enabled=False` in `rucio.cfg` still activates

PR #3625, opened 2020-05-22
Read metrics port from config as int #3623

tbeerman:
Read metrics port from config as int #3623


mlassnig (dismissed):

ericvaandering:
@tbeerman as long as you are in this code, I think enable metrics is not being read as a bool either. I cannot seem to turn it off.

tbeerman:
@ericvaandering you're right. I've fixed it.
Just as a temporary fix: you can still turn if off just by omitting `enable_metrics` in rucio.cfg.

bari12 (approved):

Issue #3626, opened 2020-05-22
Rucio upload does not honour write_lan


dchristidis:
Motivation
----------

Even with `SITE_NAME` or a similar environmental variable set, the Rucio client does not propagate the domain correctly and `write_wan` is used instead of `write_lan`.

This is supposed to be addressed with the RSE Manager rewrite in issue #3147, but it is a high-priority item and should be fixed as soon as possible.

Modification
------------

Add some basic logging in order to be able to better identify which destination is used. Propagate the domain when calling `create_protocol()` in the RSE Manager.

davidgcameron:
Can you give an example of when it doesn't work? When I tested it myself it worked ok. This part of the upload client should make sure lan domain is used:
https://github.com/rucio/rucio/blob/master/lib/rucio/client/uploadclient.py#L205

davidgcameron:
Also, +1 for more logging. I had to put print() statements in the RSE manager and protocols to see what was going on because there isn't even a logger object there.

dchristidis:
This sets `domain =  'lan'`. But, from my observations while investigating, the call to `upload()` a few lines below is not affected.

davidgcameron:
Ok, I see it now, the domain is not set in `create_protocol()`. I guess my tests with Glasgow worked because write_wan was also set to the internal proxy.

vokac:
I made tests with DESY-HH where everything seems to work fine, but it is just because of force_protocol that is passed to upload function. With different scheme set for preferred write_lan/0 vs. write_wan/0 transfers are done differently for LAN vs. WAN.

Basically it is just necessary to make upload call a bit more ugly with additional "force_domain" parameter and add domain parameters in these places
https://github.com/rucio/rucio/blob/2a1d8c41debcebc0ad2fa8f1ef7f44519767f2f2/lib/rucio/client/uploadclient.py#L236-L244
https://github.com/rucio/rucio/blob/2a1d8c41debcebc0ad2fa8f1ef7f44519767f2f2/lib/rucio/rse/rsemanager.py#L279
https://github.com/rucio/rucio/blob/2a1d8c41debcebc0ad2fa8f1ef7f44519767f2f2/lib/rucio/rse/rsemanager.py#L309-L311

This doesn't look too difficult unless you want to come cleaner solution.

dchristidis:
Thanks Petr, the linked PR does exactly that.

https://github.com/rucio/rucio/pull/3628

PR #3629, opened 2020-05-22
RSE Manager: Honour write_lan #3626

dchristidis:
Please be advised that I have little experience with this part of the code and have done minimal testing (essentially running it on LXPLUS and stubbing out some function calls so that the execution could progress).


mlassnig (approved):
I can't find anything obviously wrong with this, so green light from my side. We still need an extra pair of eyes on the clients testing release for this though.

Issue #3646, opened 2020-05-27
remove all non-essential parameters from the heartbeat executable


tbeerman:
Motivation
----------
The executable across the daemons is not uniform. Some only use the daemon name, some use the full path and others the path plus all cli arguments. E.g.,

```
rucio-undertaker
/usr/bin/rucio-reaper2
/usr/bin/rucio-conveyor-receiver --total-threads 5
```

The sharding is based on the executable and in the example above if two receivers would be started, one with 5 threads and one with 10 threads, they wouldn't share the load. 

Modification
------------
Remove the path and all non-essential cli arguments.

PR #3647, opened 2020-05-27
remove unnecessary parts from the heartbeats #3646

tbeerman:
remove unnecessary parts from the heartbeats #3646

### DIFF
--- a/lib/rucio/daemons/conveyor/poller_latest.py
+++ b/lib/rucio/daemons/conveyor/poller_latest.py
@@ -58,7 +58,7 @@ def poller_latest(external_hosts, once=False, last_nhours=1, fts_wait=1800):
     Main loop to check the status of a transfer primitive with a transfertool.
     """
 
-    executable = 'rucio-poller-latest'
+    executable = 'conveyor-poller-latest'
     hostname = socket.getfqdn()
     pid = os.getpid()
     hb_thread = threading.current_thread()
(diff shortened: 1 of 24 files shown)


lib/rucio/daemons/conveyor/poller_latest.py
```diff
@@ -58,7 +58,7 @@ def poller_latest(external_hosts, once=False, last_nhours=1, fts_wait=1800):
     Main loop to check the status of a transfer primitive with a transfertool.
     """
 
-    executable = ' '.join(sys.argv)
+    executable = 'rucio-poller-latest'
```
bari12: Change to `conveyor-poller-latest`

lib/rucio/daemons/conveyor/submitter.py
```diff
@@ -114,7 +114,7 @@ def submitter(once=False, rses=None, mock=False,
     logging.debug("Maximum time in queue for different activities: %s", max_time_in_queue)
 
     activity_next_exe_time = defaultdict(time.time)
-    executable = sys.argv[0]
+    executable = "conveyor-transfer-submitter"
```
bari12: Change to `conveyor-submitter`

bari12 (changes requested):
I think in general this is fine, but for the ATLAS deployment with int/prod we will have an issue that we will run on different partitions for a week, due to the renaming of all executables. For some of them which pull the `argv` in it is somewhat unavoidable, but for the others where we just renamed the `executable` this will happen as well.

bari12 (approved):

PR #3660, opened 2020-05-27
Client : Bulk get_metadata method : Closes #3645

### LINKED ISSUE #3645: Bulk get_metadata method
Motivation
----------
A bulk get_metadata method is needed for some BelleII workflow

cserf:
Client : Bulk get_metadata method : Closes #3645

### DIFF
--- a/lib/rucio/web/rest/webpy/v1/did.py
+++ b/lib/rucio/web/rest/webpy/v1/did.py
@@ -20,7 +20,7 @@
 # - Mario Lassnig <mario.lassnig@cern.ch>, 2012-2018
 # - Yun-Pin Sun <yun-pin.sun@cern.ch>, 2013
 # - Cedric Serfon <cedric.serfon@cern.ch>, 2014-2018
-# - Martin Baristis <martin.barisits@cern.ch>, 2014-2015
+# - Martin Baristis <martin.barisits@cern.ch>, 2014-2020
 # - Hannes Hansen <hannes.jakob.hansen@cern.ch>, 2018-2019
 # - Ruturaj Gujar, <ruturaj.gujar23@gmail.com>, 2019
 #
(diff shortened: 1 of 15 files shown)


lib/rucio/api/did.py
```diff
@@ -379,6 +379,22 @@ def get_metadata(scope, name):
...
+def bulk_get_metadata(dids, session=None):
+    """
+    Get metadata for a list of dids
+
+    :param dids: A list of dids.
+    :param session: The database session in use.
+    """
+
+    validate_schema(name='dids', obj=dids)
+    for entry in dids:
```
mlassnig: i don't think this does what you intend? the entry variable won't be persisted back to dids, won't it?

lib/rucio/core/did.py
```diff
@@ -1325,6 +1325,30 @@ def get_metadata(scope, name, session=None):
         raise exception.DataIdentifierNotFound("Data identifier '%(scope)s:%(name)s' not found" % locals())
 
 
+@stream_session
+def bulk_get_metadata(dids, session=None):
```
mlassnig: maybe `get_metadata_bulk` instead?

lib/rucio/web/rest/webpy/v1/did.py
```diff
@@ -653,6 +654,43 @@ def POST(self, scope, name, key):
...
+
+        :returns: A list of dictionaries containing all meta.
+        """
+        header('Content-Type', 'application/x-json-stream')
+        json_data = data()
+        try:
+            params = parse_response(json_data)
+            if 'dids' in params:
+                dids = params['dids']
+            for meta in bulk_get_metadata(dids):
```
mlassnig: if dids is not in params, then this variable won't be found and it will crash

lib/rucio/api/did.py
cserf:
```
>>> dids = [{'scope': 'AA'}, {'scope': 'BB'}]
>>> for entry in dids:
...     entry['scope'] = 'blah'
... 
>>> dids
[{'scope': 'blah'}, {'scope': 'blah'}]
```

lib/rucio/core/did.py
cserf: OK

lib/rucio/web/rest/webpy/v1/did.py
cserf: Will add a protection

Issue #3696, opened 2020-06-05
Write domain used for storm protocol.


TomasJavurek:
Motivation
----------
report from @davidgcameron 
[dcameron@lxplus7127] ~> SITE_NAME=UKI-LT2-QMUL rucio -v upload --rse UKI-LT2-QMUL_SCRATCHDISK --scope user.dcameron --name dasghbnwofsdfdfds /bin/bash
2020-06-03 20:02:44,157    INFO    Preparing upload for file bash
2020-06-03 20:02:44,184    DEBUG    Registering file
2020-06-03 20:02:44,184    DEBUG    Skipping dataset registration
2020-06-03 20:02:44,213    INFO    File DID already exists
2020-06-03 20:02:44,213    DEBUG    local checksum: 3d1ae86e, remote checksum: 3d1ae86e
2020-06-03 20:02:44,292    DEBUG    Traceback (most recent call last):
  File "/cvmfs/atlas.cern.ch/repo/ATLASLocalRootBase/x86_64/rucio-clients/1.22.7/bin/rucio", line 162, in new_funct
    return function(*args, **kwargs)
  File "/cvmfs/atlas.cern.ch/repo/ATLASLocalRootBase/x86_64/rucio-clients/1.22.7/bin/rucio", line 898, in upload
    upload_client.upload(items, summary_file_path)
  File "/cvmfs/atlas.cern.ch/repo/ATLASLocalRootBase/x86_64/rucio-clients/1.22.7/lib/python2.7/site-packages/rucio/client/uploadclient.py", line 205, in upload
    if rsemgr.exists(rse_settings, pfn if pfn else file_did, domain=domain, auth_token=self.auth_token):
  File "/cvmfs/atlas.cern.ch/repo/ATLASLocalRootBase/x86_64/rucio-clients/1.22.7/lib/python2.7/site-packages/rucio/rse/rsemanager.py", line 264, in exists
    pfn = list(protocol.lfns2pfns(f).values())[0]
  File "/cvmfs/atlas.cern.ch/repo/ATLASLocalRootBase/x86_64/rucio-clients/1.22.7/lib/python2.7/site-packages/rucio/rse/protocols/storm.py", line 58, in lfns2pfns
    path = lfn['path'] if 'path' in lfn and lfn['path'] else self._get_path(scope=lfn['scope'].external,
AttributeError: 'str' object has no attribute 'external'

2020-06-03 20:02:44,292    ERROR    'str' object has no attribute 'external'
[...]

PR #3699, opened 2020-06-05
storm: write domains should not be used - #3696

TomasJavurek:
storm: write domains should not be used - #3696

### DIFF
--- a/lib/rucio/rse/protocols/storm.py
+++ b/lib/rucio/rse/protocols/storm.py
@@ -55,7 +55,7 @@ class Default(protocol.RSEProtocol):
 
         lfns = [lfns] if isinstance(lfns, dict) else lfns
         for lfn in lfns:
-            path = lfn['path'] if 'path' in lfn and lfn['path'] else self._get_path(scope=lfn['scope'].external,
+            path = lfn['path'] if 'path' in lfn and lfn['path'] else self._get_path(scope=lfn['scope'],
                                                                                     name=lfn['name'])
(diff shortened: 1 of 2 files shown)


lib/rucio/rse/protocols/storm.py
```diff
@@ -55,7 +55,7 @@ def lfns2pfns(self, lfns):
 
         lfns = [lfns] if isinstance(lfns, dict) else lfns
         for lfn in lfns:
-            path = lfn['path'] if 'path' in lfn and lfn['path'] else self._get_path(scope=lfn['scope'].external,
+            path = lfn['path'] if 'path' in lfn and lfn['path'] else self._get_path(scope=lfn['scope'],
```
mlassnig: isn't this needed for multi-vo?

bari12: In the protocols we assume that it is always the external scope already. It's difficult here, since these methods are both used at the client and server, but since the main usage is on client we decided that we assume external scopes here.

bari12:
I am not sure why travis still shows the tests as pending. But the tests actually failed, please have a look.

TomasJavurek:
> I am not sure why travis still shows the tests as pending. But the tests actually failed, please have a look.

protocol.exists(None)

is the problem. I wanted to test whether exists method is implemented with this, but this was stupid.

bari12:
I think this is fine. @mlassnig please have a look as well.

mlassnig (approved):

Issue #3692, opened 2020-06-03
Add dirac_add_files method for the Rucio/Dirac integration


cserf:
Motivation
----------
The goal of the method is to have a atomic way to create files for Dirac. Use case : I would like to create a file named `/belle//user/serfon/dir/mytest.root`. This should be part of dataset `/belle/user/serfon/dir`. The method needs to : 
- Check if `/belle/user/serfon/dir`  and all it's ascendants `/belle/user/serfon`, `/belle/user`, etc. exist. If not it creates and attachs them. 
- Create the file+replicas and a rule on the dataset.

Modification
------------
Create a new client, REST, API, core for dirac methods

mlassnig:
Don't we have something like this for ATLAS Tier-0 already?

cserf:
Nope. T0 just does a only `add_replicas` + `attach` and it's not atomic. What we want here is different, because we need to reconstruct the hierarchical namespace : 
Container `/belle` containing container `/belle/user` (and so on)  containing dataset `/belle/user/serfon/dir` containing file `/belle//user/serfon/dir/mytest.root` . The method should be able to create any of the non existing directories in an atomic way.

Issue #3690, opened 2020-06-03
Belleii : Adding permission file + fixing regular expression in schema.py


cserf:
Motivation
----------
Missing permission file for Belleii

PR #3709, opened 2020-06-09
Add dirac_add_files method for the Rucio/Dirac integration + Belleii : Adding permission file + Define extract_scope based on policy : Closes #3679, #3690, #3692

### LINKED ISSUE #3679: Define extract_scope based on policy
Motivation
----------
The `extract_scope` method defined in bin is ATLAS specific. It should depend on the policy

cserf:
Add dirac_add_files method for the Rucio/Dirac integration + Belleii : Adding permission file + Define extract_scope based on policy : Closes #3679, #3690, #3692


bari12 (approved):
Looks good to me, I think this can be added like this.

PR #3731, opened 2020-06-12
Core & Internals: prepare policy package code for multi VO #3542

jamesp-epcc:
This pull request lays the groundwork for making policy packages work in multi VO Rucio. Instead of loading the schema and permission modules directly into Python namespaces as before, they are now stored in dictionaries indexed by the VO name, and all accesses to them go through a few centralised methods.


jamesp-epcc:
Created against master in error.

PR #3736, opened 2020-06-16
Core: Add bulk method for list dataset replicas; Fix #2459

### LINKED ISSUE #2459: Support bulk requests in the list_dataset_replicas CLI API
Motivation
----------
Given that a Rucio dataset corresponds to a CMS block, it would be extremely helpful if the rucio-clients `list_dataset_replicas` API would accept a list of dataset names, such that we can fetch all CMS block replicas that belong to a CMS dataset in a single call (thus having a better performance at multiple layers).

For the record, I've placed the same issue under the CMSRucio project:
https://github.com/dmwm/CMSRucio/issues/41

FYI @ericvaandering 

Modification
------------

Support a list of datasets in the `list_dataset_replicas` API, from the rucio-clients package.

bziemons:
Core: Add bulk method for list dataset replicas; Fix #2459

- [x] add API access to bulk method
- [x] add ReplicaClient method
- [x] add `test_list_dataset_replicas_bulk` in `test_dataset_replicas.py`
- [ ] ~add list-dataset-replicas tests in `test_bin_rucio.py`~ (I wasn't able to mock/run the replication process)

### DIFF
--- a/bin/rucio
+++ b/bin/rucio
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
-# Copyright 2012-2018 CERN for the benefit of the ATLAS collaboration.
+# Copyright 2012-2020 CERN for the benefit of the ATLAS collaboration.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


bziemons:
@bari12 ready to review. Just fixed and rebased it and now minds multi-vo

bari12 (approved):

Issue #3716, opened 2020-06-10
Gfal logging


TomasJavurek:
Motivation
----------
parent issue - #3684 

Modification
------------

TomasJavurek:
It seems that this method:
https://github.com/rucio/rucio/blob/master/lib/rucio/common/utils.py#L580
is being used somewhere, but I couldn't find it. It causes duplicates of the logger messages.

TomasJavurek:
actually it is this module that introduces new fromatter:
https://github.com/rucio/rucio/blob/master/lib/rucio/common/policy.py#L41

PR #3741, opened 2020-06-17
logging: it produces duplicate messages, needs a fix - #3716

TomasJavurek:
logging: it produces duplicate messages, needs a fix - #3716

### DIFF
--- a/bin/rucio
+++ b/bin/rucio
@@ -89,6 +89,7 @@ def setup_logger(logger):
         def func(*args):
             if 'RUCIO_LOGGING_FORMAT' not in os.environ:
                 levelno = args[0].levelno
+                format_str = '%(asctime)s\t%(levelname)s\t%(message)s\033[0m'
                 if levelno >= logging.CRITICAL:
                     color = '\033[31;1m'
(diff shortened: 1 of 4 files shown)


mlassnig (approved):

bin/rucio
```diff
@@ -99,9 +100,10 @@ def setup_logger(logger):
                     color = '\033[32;1m'
                 elif levelno >= logging.DEBUG:
                     color = '\033[36;1m'
+                    format_str = '%(asctime)s\t%(levelname)s\t%(filename)s\t%(funcName)s\t%(message)s\033[0m'
```
mlassnig, line 103: @TomasJavurek this will print the filename and funcname also for INFO

i don't think this is intended?

TomasJavurek: It is not. We already discussed with @Martin that we don't do it like this. However, we have, I think, more ugly problem with the loggin, and that's the duplicate messages.

TomasJavurek: but although logging.INFO >= logging.DEBUG evaluates to True, it wouldn't be used for the logging.INFO due to the elif clause.

PR #3765, opened 2020-06-23
Core: When validating a DID, supply the DID type

ericvaandering:
CMS DID validation needs to know the DID type. These are the two places it's not supplied


bari12:
Hi @ericvaandering 
The syntax tests show `lib/rucio/tests/test_schema_cms.py:17:1: F401 'pdb' imported but unused`

ericvaandering:
Fixed. Thanks!

bari12 (approved):

PR #3770, opened 2020-06-24
GFAL implementation with CLI instead of API; Fix #3537

### LINKED ISSUE #3537: GFAL implementation with CLI instead of API
Motivation
----------
Due to the C-native binding, it's de-facto not feasible to cancel an ongoing download that's timing out.

Modification
----------
Provide an alternative GFAL implementation that uses gfal-copy instead.

mlassnig:
GFAL implementation with CLI instead of API; Fix #3537


bari12:
Closed due to new code management model (Master PR still open)

Issue #3775, opened 2020-06-26
Return files size when listing scope DIDs


gabrielefronze:
Motivation
----------
The API response when listing the DIDs belonging solely to a scope or to a container is very dissimilar.
While in the first case the bytes size is not shown:

```
{"scope": "O3", "type": "FILE", "name": "K-K1_llhoft-1266855936-4096.gwf", "parent": null, "level": 0}
```

in the second it is:

```
{"adler32": "3011a75c", "name": "K-K1_C10-1268027392-4096.gwf", "bytes": 34248761, "scope": "O3", "type": "FILE", "md5": "3a46a528ff1d618f5791e3f0eceb42f8"}
```

Having a similar pattern (or at least the bytes size) would greatly improve the operations of the FUSE-Posix module, since it would enable bulk file size retrieval even at the scope level (and not only within a dataset or container).

Modification
------------
It should suffice to edit the API implementation around https://github.com/rucio/rucio/blob/master/lib/rucio/core/did.py#L1098 in order to introduce the retrieval of other fields, as done here https://github.com/rucio/rucio/blob/master/lib/rucio/core/did.py#L1141 .

viveknigam3003:
I'll handle this one.

PR #3792, opened 2020-06-30
enhancement: Added 'bytes' field for GET /dids/(scope)/ #3775

viveknigam3003:
enhancement: Added 'bytes' field for GET /dids/(scope)/ #3775

### DIFF
--- a/lib/rucio/core/did.py
+++ b/lib/rucio/core/did.py
@@ -28,6 +28,7 @@
 # - Ruturaj Gujar <ruturaj.gujar23@gmail.com>, 2019
 # - Brandon White <bjwhite@fnal.gov>, 2019
 # - Luc Goossens <luc.goossens@cern.ch>, 2020
+# - Vivek Nigam <viveknigam.nigam3@gmail.com>, 2020
 #
 # PY3K COMPATIBLE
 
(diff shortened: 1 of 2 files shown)


lib/rucio/core/did.py
```diff
@@ -1092,10 +1093,13 @@ def scope_list(scope, name=None, recursive=False, session=None):
 
     def __topdids(scope):
         c = session.query(models.DataIdentifierAssociation.child_name).filter_by(scope=scope, child_scope=scope)
-        q = session.query(models.DataIdentifier.name, models.DataIdentifier.did_type).filter_by(scope=scope)  # add type
+        q = session.query(models.DataIdentifier.name, models.DataIdentifier.did_type, models.DataIdentifier.bytes).filter_by(scope=scope)  # add type
         s = q.filter(not_(models.DataIdentifier.name.in_(c))).order_by(models.DataIdentifier.name)
         for row in s.yield_per(5):
-            yield {'scope': scope, 'name': row.name, 'type': row.did_type, 'parent': None, 'level': 0}
+            if str(row.did_type) == 'FILE':
```
gabrielefronze, line 1099: The `did_type` can be directly compared with `DIDType.FILE`.
Please use if in the if statement!

gabrielefronze (changes requested):
Seems good! Just fix the if statement please :)

PR #3862, opened 2020-07-20
enhancement : Added 'bytes' field for scope DIDs #3775

viveknigam3003:
enhancement: Added 'bytes' field for GET /dids/(scope)/ #3775


bari12:
Please do not make PRs against the release branches. 😄

PR #3864, opened 2020-07-21
Authentication & Authorisation: Added authentication options for m-VO, Closes #3855

### LINKED ISSUE #3855: Add authentication options for multi-VO
Motivation
----------
Currently the OIDC method of authentication hasn't been updated to be compatible with m-VO mode.

Also, the CLI can be used to specify the account you wish to authenticate against, so it makes sense to allow the option to specify the VO to authenticate against as well. This will take precedence over the setting in `rucio.cfg`. For cases where an individual is associated with multiple VOs and wants to easily swap between them.

Modification
------------
Make sure VO is passed through the OIDC code, and tests. Add explicit m-VO testing for OIDC.

Add VO as a CLI option like account, and explicit testing of this.

patrick-austin:
Added authentication options for m-VO
------------------

Have added option specify VO on the CLI, if used then the clients created will have the VO passed to their constructor, otherwise the VO from the config file is used as before.

Have made some minor changes to the OIDC code to ensure VO is passed through all layers.

Added tests to check for interference between 2 VOs for both of the above features, as well as modifying the existing OIDC tests to use the VO from the config file when multi-VO tests are run.

### DIFF
--- a/bin/rucio
+++ b/bin/rucio
@@ -35,6 +35,7 @@
 # - Ruturaj Gujar, <ruturaj.gujar23@gmail.com>, 2019
 # - Jaroslav Guenther <jaroslav.guenther@gmail.com>, 2019-2020
 # - Benedikt Ziemons <benedikt.ziemons@cern.ch>, 2020
+# - Patrick Austin <patrick.austin@stfc.ac.uk>, 2020
 #
 # PY3K COMPATIBLE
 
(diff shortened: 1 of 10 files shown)


bari12:
Please squash into 1 commit

bari12:
Although marked as a feature, I will still merge this to the current release line (1.23) as well.

bari12 (approved):

Issue #3805, opened 2020-07-01
rucio mover is not retrying stageout 3 times as it should


ivukotic:
Motivation
----------

rucio mover does not retry stage out. If it fails job is failed. Fixing this would save ~10-20 jobs/h from failing.

Modification
------------

bari12:
I think it does at least one retry
https://github.com/rucio/rucio/blob/975d8f7cefdc647ec51ce88f81478c84a2503203/lib/rucio/client/downloadclient.py#L558
@TomasJavurek @mlassnig ?

ivukotic:
On stage in it works. On stage out it does not. I have never seen an evidence of it in jobs logs. 
Here one: https://aipanda023.cern.ch/condor_logs_2/20-07-01_00/grid.7520343.2.out
from this job: https://bigpanda.cern.ch/job?pandaid=4774970626

mlassnig:
Just checked the uploadclient. There's no retry there, correct. @TomasJavurek can you please look into this.

TomasJavurek:
There used to be retries in rucio mover in pilot1. But we decided to keep it fully in clients for the pilot2.

ivukotic:
@mlassnig can you revisit with @TomasJavurek what was agreed with Paul and Alexey? If I remember well it was on rucio mover to implement retries.

bari12:
https://github.com/rucio/rucio/blob/3eeeca8f99e8f58fd24925e014289f23edc0c534/bin/rucio#L1043
Broke the possibility to just `rucio download <name>` without giving the `<scope>:<name>` pair. This needs to be fixed asap, since we cannot deploy the clients for users like this.

The cause of the issue is related to #3824 actually (not this ticket) - but this was all put together in one PR. Followup in #3824

PR #3870, opened 2020-07-23
Clients: merging 3805 and 3824 - Closes #3824 #3805

### LINKED ISSUE #3824: Incomplete download looks "complete" when restricting RSEs
Motivation
----------
As a user one may naively assume that a dataset/container is completely downloaded when seeing an output like
```
Total files :                                93
Downloaded files :                            0
Files already found locally :                93
Files that cannot be downloaded :             0
```
and a return code of 0. However, this example is taken from a case where I restricted the RSEs to download from, e.g., via `rucio download --rse '*\SOME_PROBLEMATIC_SITE' someContainer` and the full container contains more than the 93 files reported. (Downloads from the problematic site in this case keep failing for a subset of the files involved)

The effect is the same if some RSE is explicitly selected (`--rse 'SOME_GOOD_RSE'`) or individual RSEs are vetoed as in the example above.

Modification
------------
The total number of files should likely be determined without the RSE restriction to avoid files only present on the excluded RSEs being forgotten about. If the current counting is intentional it would be good to at least print a warning in case only a restricted set of RSEs was considered.

TomasJavurek:
Just mergint two previous branches that already succeeded in travis.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -53,7 +53,13 @@ Contributing
 
 **Step 1**: If not exist, create an `issue <https://github.com/rucio/rucio/issues/new>`_ with the description of the contribution (motivation, modification and expected results). Every issue will get a **unique issue number**.
 
-**Step 2**: Create a local branch that corresponds to the issue. There are utility scripts to help you with this::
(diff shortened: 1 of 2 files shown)


bari12:
Please squash commits ;-)

PR #3883, opened 2020-07-24
Testing: Changed m-VO tests to use postgres, Closes #3865

### LINKED ISSUE #3865: Change m-VO tests to use postgres
Motivation
----------
PostgreSQL is a more widely used DB dialect than the currently used SQLite. It should still be sufficient to only test m-VO explicitly against one of the supported DB dialects however.

Modification
------------
Alter the `.cfg`s used in the test suite to use PostgreSQL, and rename files appropriately.

patrick-austin:
Changed m-VO tests to use postgres
------------------

- Modified and renamed the files for m-VO tests to run against PostgreSQL instead of SQLite.
- Added a line to the normal `run_tests_docker.sh` that allows single-VO tests to be run after multi-VO tests in dev containers, without manually modifying the config files . Currently `run_multi_vo_tests_docker.sh` overwrites `rucio.cfg` when setting `multi_vo = True`, so any future actions, including `run_tests_docker.sh` would be in multi-VO mode.
- Fixed the VO not being passed through the REST layer by `delete_metadata`

### DIFF
--- a/etc/docker/test/extra/rucio_multi_vo_ts2_postgres12.cfg
+++ b/etc/docker/test/extra/rucio_multi_vo_ts2_postgres12.cfg
@@ -19,10 +19,13 @@ request_retries = 3
 vo = ts2
 
 [database]
-default = sqlite:////tmp/rucio.db
+default = postgresql://postgres:secret@postgres12/postgres
+schema = dev
 pool_recycle=3600
 echo=0
 pool_reset_on_return=rollback
+pool_size = 20
+max_overflow = 20
 
 [bootstrap]
 # Hardcoded salt = 0, String = secret, Python: hashlib.sha256("0secret").hexdigest()
--- a/etc/docker/test/extra/rucio_multi_vo_tst_postgres12.cfg
+++ b/etc/docker/test/extra/rucio_multi_vo_tst_postgres12.cfg
@@ -19,10 +19,13 @@ request_retries = 3
 vo = tst
 
 [database]
-default = sqlite:////tmp/rucio.db
+default = postgresql://postgres:secret@postgres12/postgres
+schema = dev
 pool_recycle=3600
 echo=0
 pool_reset_on_return=rollback
+pool_size = 20
+max_overflow = 20
 
 [bootstrap]
 # Hardcoded salt = 0, String = secret, Python: hashlib.sha256("0secret").hexdigest()
--- a/etc/docker/test/matrix.yml
+++ b/etc/docker/test/matrix.yml
@@ -22,4 +22,4 @@ suites:
   - id: client
     RDBMS: sqlite
   - id: multi_vo
-    RDBMS: sqlite
+    RDBMS: postgres12
--- a/lib/rucio/web/rest/webpy/v1/did.py
+++ b/lib/rucio/web/rest/webpy/v1/did.py
@@ -732,7 +732,7 @@ class Meta(RucioController):
                 raise generate_http_error(404, 'KeyNotFound', 'No key provided to remove')
 
         try:
-            delete_metadata(scope=scope, name=name, key=key)
+            delete_metadata(scope=scope, name=name, key=key, vo=ctx.env.get('vo'))
         except KeyNotFound as error:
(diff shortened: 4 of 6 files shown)


bziemons:
Can we add a parameter to `run_tests_docker.sh` to do the overriding? The autotests are using the same script to run the tests and although it seems to run fine at the moment, I fear it is a detail that can be missed easily. Usually I would expect the configs to be where they are before I start a script to run tests. Before you change anything, lets wait for @bari12's opinion though.

patrick-austin:
In principle I think making it an optional parameter is a good idea - presumably with the old behaviour as the default?

As you say the autotests use the same script, but as `rucio_single_vo.cfg` isn't present it outputs a complaint when it tries to copy the file, but then continues to run the tests (or at least this is how it worked with Travis, and it looks like the new system behaves the same way).

From a dev container perspective, I can see that it might be confusing to change the cfg files as part of the test script, especially if you just changed the cfg for whatever reason (and so expected that changed value to be used in the tests). Copying cfg files for the m-VO test script was necessary as we change between two VOs during the tests, and so needed to change to cfg file to reflect this. It then seemed sensible to have the original test script do the same thing and overwrite the test file at the start so running the m-VO tests didn't change anything for future testing.

What might make more sense than changing `run_tests_docker.sh`, is to change `run_multi_vo_tests_docker.sh` so that it makes a copy of the current `rucio.cfg` at the start, then after it completes the tests, it reverts back to that file. That way the existing and more widely used script is unchanged, and running the multi-VO tests won't affect the state of the config file after they're run.

bziemons:
> As you say the autotests use the same script, but as rucio_single_vo.cfg isn't present it outputs a complaint when it tries to copy the file, but then continues to run the tests (or at least this is how it worked with Travis, and it looks like the new system behaves the same way).

I see so the `rucio_single_vo.cfg` is created when running the multi-vo test suite? I haven't looked through it yet, I guess.

> What might make more sense than changing run_tests_docker.sh, is to change run_multi_vo_tests_docker.sh so that it makes a copy of the current rucio.cfg at the start, then after it completes the tests, it reverts back to that file. That way the existing and more widely used script is unchanged, and running the multi-VO tests won't affect the state of the config file after they're run.

I think the config should actually be copied to the right place by a setup script. Looking at https://github.com/rucio/rucio/blob/master/tools/test/before_script.sh I thought this was the case for the multi-vo suite?

Edit: Oh right, obviously they aren't called `rucio.cfg` both. Then, instead I suggest different configuration directories and setting `RUCIO_HOME` for those tests before executing them. For me it would be much clearer what happens in that case.

patrick-austin:
> I see so the rucio_single_vo.cfg is created when running the multi-vo test suite? I haven't looked through it yet, I guess.

`rucio_single_vo.cfg` [is only in the dev docker container](https://github.com/rucio/containers/blob/master/dev/Dockerfile#L81
), I put it there along with the two multi-VO `.cfg`s so that it'd be possible to run the multi_vo tests. `rucio_single_vo.cfg` isn't directly used by `run_multi_vo_tests_docker.sh`, it's just another copy of the normal `.cfg` so that the only copy doesn't get overwritten by the m-VO tests (and making it easier to switch between the two test scripts).

> Edit: Oh right, obviously they aren't called rucio.cfg both. Then, instead I suggest different configuration directories and setting RUCIO_HOME for those tests before executing them. For me it would be much clearer what happens in that case.

I hadn't looked into `RUCIO_HOME` as a possible way round this (which is weird because it looks like I copied everything from `rucio_single_vo.cfg` _except_ the line `export RUCIO_HOME=/opt/etc/test`). But yes it looks like using it would do everything without copying over any files. I'll set it up to use something like `/opt/rucio/etc/multi_vo/tst/etc/rucio.cfg`  for the m-VO tests and then make PR for the dev container to match this structure (assuming there aren't any more issues with it).

bari12:
To me it looks good like this. But please also squash into one commit :-) (There seems to be a conflict in the meanwhile, probably due to my merging PRs)

bari12 (approved):

PR #3886, opened 2020-07-27
WebUI: Make m-VO compatible with WebUI, Closes #3885

### LINKED ISSUE #3885: Make m-VO compatible with WebUI
Motivation
----------

Currently m-VO isn't supported when using the WebUI, as VO cannot be selected at login and the VO is not passed by the function calls from within the WebUI.

Modification
------------

Add a cookie for VO, and add it to the headers used in function calls made by the `RucioClient`.
Add additional logic to the process of account selection so we automatically choose the VO where possible (based on the identity and optionally provided account). If the VO cannot be uniquely determined (e.g. one person has access to the account `root` at two VOs) then display a list of the appropriate VOs for them to choose from.

patrick-austin:
Make m-VO compatible with WebUI
------------------

VO is now either determined based on identity (with account name narrowing the options if provided) at login. If multiple VOs are possible for the identity given, then the user stays on the login screen and has to manually choose their VO from a drop down list.

This VO is then set as a cookie and used in the headers so information displayed only applies to the user's VO.

Should be no change for single-VO operation.

### DIFF
--- a/lib/rucio/web/ui/common/utils.py
+++ b/lib/rucio/web/ui/common/utils.py
@@ -11,6 +11,7 @@
 # - Jaroslav Guenther, <jaroslav.guenther@cern.ch>, 2019-2020
 # - Eli Chadwick, <eli.chadwick@stfc.ac.uk>, 2020
 # - Benedikt Ziemons <benedikt.ziemons@cern.ch>, 2020
+# - Patrick Austin <patrick.austin@stfc.ac.uk>, 2020
 #
 # PY3K COMPATIBLE
 #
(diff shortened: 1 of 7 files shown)


lib/rucio/web/ui/common/utils.py
```diff
@@ -73,6 +75,7 @@
     except:
         AUTH_ISSUERS = []
 
+MULTI_VO = config_get('common', 'multi_vo', default=False)
```
tbeerman: please use `config_get_bool` here. Otherwise multi_vo will always be enabled as soon it's added to the config file, even for `multi_vo = False`

lib/rucio/web/ui/main.py
```diff
@@ -83,6 +84,7 @@
 
 )
 
+MULTI_VO = config_get('common', 'multi_vo', default=False)
```
tbeerman: see above

tbeerman (changes requested):
Looks good to me. It couldn't really test the multi_vo features, since I don't have a multi_vo test env, but at least it doesn't break anything in my single vo test env.

lib/rucio/web/ui/common/utils.py
patrick-austin: Done.

lib/rucio/web/ui/main.py
patrick-austin: And done.

tbeerman (dismissed):

patrick-austin:
Updating from `config_get` to `config_get_bool` broke all the tests because I didn't set `raise_exception=False`. Hopefully the tests should pass now I've changed that.

I also added a missing default argument for `ui_vo` in the `oidc_auth` function.

bari12 (approved):

PR #3888, opened 2020-07-28
Deletion: VO handling for daemons, Closes #3887

### LINKED ISSUE #3887: VO handling for daemons
Motivation
----------

Currently the following daemons can accept RSE names or expressions as arguments:
- Reaper (original, light, dark, 2)
- Conveyor submitter/stager
- Suspicious replica recoverer

In a m-VO instance, this raises issues as an RSE name/expression could match storage for different VOs at the same time.

Also, in light of changes to policy packages, conveyor submitter sometimes needs the VO to get the default list of activities. Adding VO as an argument will make it possible to get the activities in a m-VO instance.

Further changes to daemons may also be needed in the future to make them fully m-VO compatible.

Modification
------------

Add VO as an argument to the above daemons, and use it when they get/parse RSEs (or get from the policy).

patrick-austin:
VO handling for daemons
------------------

Added VO as an argument for reaper(s), conveyor-submitter/stager and replica recover. This is mainly used when determining which RSEs to act on. If the VO is specified in `rucio.cfg` then this is also taken into account, e.g. can only run on multiple VOs if the VO isn't specified or the VO in the config is `def`.

Also added some explicit tests of running m-VO reaper to the m-VO test suite.

### DIFF
--- a/bin/rucio-dark-reaper
+++ b/bin/rucio-dark-reaper
@@ -16,6 +16,7 @@
 # Authors:
 # - Vincent Garonne <vgaronne@gmail.com>, 2016-2018
 # - Mario Lassnig <mario.lassnig@cern.ch>, 2019
+# - Patrick Austin <patrick.austin@stfc.ac.uk>, 2020
 
 """
 Reaper is a daemon to manage file on disk but not in Rucio(dark data) deletion
@@ -38,6 +39,9 @@ def get_parser():
     parser.add_argument("--chunk-size", action="store", default=10, type=int, help='Chunk size')
     parser.add_argument("--scheme", action="store", default=None, type=str, help='Force the reaper to use a particular protocol, e.g., mock.')
     parser.add_argument('--rses', nargs='+', type=str, help='List of RSEs')
+    parser.add_argument('--exclude-rses', action="store", default=None, type=str, help='RSEs expression to exclude RSEs')
+    parser.add_argument('--include-rses', action="store", default=None, type=str, help='RSEs expression to include RSEs')
+    parser.add_argument('--vos', nargs='+', type=str, help='Optional list of VOs to consider. Only used in multi-VO mode.')
     return parser
 
 
@@ -47,7 +51,7 @@ if __name__ == '__main__':
     args = parser.parse_args()
     try:
         run(total_workers=args.total_workers, chunk_size=args.chunk_size,
-            once=args.run_once, scheme=args.scheme, rses=args.rses,
-            all_rses=args.all_rses)
+            once=args.run_once, scheme=args.scheme, rses=args.rses, all_rses=args.all_rses,
+            exclude_rses=args.exclude_rses, include_rses=args.include_rses, vos=args.vos)
     except KeyboardInterrupt:
         stop()
--- a/lib/rucio/daemons/conveyor/common.py
+++ b/lib/rucio/daemons/conveyor/common.py
@@ -29,6 +29,7 @@
 # - Gabriele Fronze' <gfronze@cern.ch>, 2019
 # - Jaroslav Guenther <jaroslav.guenther@cern.ch>, 2019-2020
 # - Benedikt Ziemons <benedikt.ziemons@cern.ch>, 2020
+# - Patrick Austin <patrick.austin@stfc.ac.uk>, 2020
 #
 # PY3K COMPATIBLE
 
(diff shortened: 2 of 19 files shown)


lib/rucio/daemons/conveyor/common.py
```diff
@@ -440,32 +443,55 @@ def bulk_group_transfer(transfers, policy='rule', group_bulk=200, source_strateg
...
 
     :param rses:          List of rses (Single-VO only)
     :param include_rses:  RSEs to include
     :param exclude_rses:  RSEs to exclude
+    :param vos:           VOs on which to look for RSEs. Only used in multi-VO mode.
+                          If None, we either use all VOs if run from "def", or the current VO otherwise.
     :return:              List of working rses
     """
+    multi_vo = config_get_bool('common', 'multi_vo', raise_exception=False, default=False)
+    vo = config_get('client', 'vo', raise_exception=False, default='def')
```
bari12: I am not sure why the client-vo setting is relevant here at all?

patrick-austin: The way I'd been handling RSE selection for the daemons was to limit the daemon to the VO set in the [client] section (if it was set). But to allow action across any/all VO specified if the client section doesn't have a VO set (i.e. the `super_root` at `def` is running it).
I suppose this would be unnecessary if the daemons are never run from "client" locations associated with a specific VO, or maybe this kind of check is less relevant for the conveyor than it is for the reaper(s).

bari12: I think that should be removed. On a server/daemon machine the `[client]` section of the cfg should not have any impact on the running of the daemon, that will just lead to confusion. If needed, that should be moved to a specific section. e.g. a `vo` attribute in the `[conveyor]` section etc.

patrick-austin: My understanding was that server/daemon machines wouldn't have the `[client]` section, and so this part of the logic would only affect a client with daemons installed. But I suppose installing daemons on a client might not be a thing commonly done anyway? I may have got the wrong impression from the dev container, where you have access to daemons alongside the CLI / admin CLI.
I'll remove this part of the logic, and allow the daemon to run on any VO(s) passed as an argument (provided `multi_vo` is True). This happens for the replica-recoverer and reaper(s) too, so I'll make the same change in all places (and modify the tests/documentation to reflect this).

bin/rucio-dark-reaper
```diff
@@ -33,11 +34,13 @@ def get_parser():
     """
     parser = argparse.ArgumentParser(description="The Dark-Reaper daemon is responsible for the deletion of quarantined replicas.")
     parser.add_argument("--run-once", action="store_true", default=False, help='One iteration only')
-    parser.add_argument("--all-rses", action="store_true", default=False, help='Select RSEs from the quarantined queues')
```
bari12: The removal of this option is a bit problematic. I understand why you do it, but adding this in a patch release is a bit problematic, since this might fail unexpectedly for some communities.
Can you maybe add this in two steps.
1. In this PR you still leave the `--all-rses` in, but if used in multi-vo mode you give an error. You also add the new arguments (exclude-rses, include-rses, vos)
2. Once this one is merged, you preapre a feature PR which removes the `--all-rses`. 

Not only for the dark-reaper but also for the other daemons where changed.

bari12 (changes requested):

lib/rucio/daemons/reaper/dark_reaper.py
```diff
@@ -178,47 +181,62 @@ def stop(signum=None, frame=None):
...
     :param rses: List of RSEs the reaper should work against. If empty, it considers all RSEs (Single-VO only).
     :param scheme: Force the reaper to use a particular protocol/scheme, e.g., mock.
     :param exclude_rses: RSE expression to exclude RSEs from the Reaper.
     :param include_rses: RSE expression to include RSEs.
+    :param vos: VOs on which to look for RSEs. Only used in multi-VO mode.
+                If None, we either use all VOs if run from "def", or the current VO otherwise.
     """
     logging.info('main: starting processes')
 
-    all_rses = list_rses()
```
patrick-austin, line 197: @bari12 just to be clear, when putting the option for `--all-rses` back in for the light/dark reaper I should also revert to this logic of listing the RSEs and overwriting the Boolean value of `all_rses` with a list (meaning it will always behave as though `--all-rses` was passed)?
I agree with your comment about making it a patch followed by feature, I just want to make sure I've understood what needs to be changed for this PR.

bari12: Yes, I think so. We just keep it as it was and address the change of options in a feature.

bari12 (approved):

Issue #3843, opened 2020-07-16
Cannot deny rule which is made with a lock


ericvaandering:
Motivation
----------
CMS is going to use the approve/deny rule functionality to make sure sites have tape families for datasets before allowing them onto tape. Those rules are made in a locked state (by Rucio, we think). 

Denying these rules does not work. 

Modification
------------

I think the way to solve this is to add a parameter ignore_locks to delete_rule in core/rule.py and make deny rule use it.

There is a separate issue that when using the web UI, there is no useful message output, just a red bar.

bari12:
I think that solution is good. I would just call the parameter `ignore_rule_lock` so there is no confusion with replica locks.
Alternatively, within `deny_rule` you can also just check if there is a lock active and remove it. (Might be easier)

PR #3892, opened 2020-07-29
Rules: Add a parameter to ignore rule locks and use it when denying rules #3843

ericvaandering:
Fixes the issue where CMS is not able to use the Web UI to deny rules on tape sites.

Please apply to 1.23 branch and master

### DIFF
--- a/lib/rucio/core/rule.py
+++ b/lib/rucio/core/rule.py
@@ -923,26 +923,30 @@ def list_associated_rules_for_file(scope, name, session=None):
 
 
 @transactional_session
-def delete_rule(rule_id, purge_replicas=None, soft=False, delete_parent=False, nowait=False, session=None):
+def delete_rule(rule_id, purge_replicas=None, soft=False, delete_parent=False, nowait=False, session=None,
+                ignore_rule_lock=False):
     """
     Delete a replication rule.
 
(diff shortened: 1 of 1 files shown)


lib/rucio/core/rule.py
```diff
@@ -923,26 +923,30 @@ def list_associated_rules_for_file(scope, name, session=None):
...
+    :param session:           The database session in use.
+    :param ignore_rule_lock:  Ignore any locks on the rule
+    :raises:                  RuleNotFound if no Rule can be found.
+    :raises:                  UnsupportedOperation if the Rule is locked.
     """
 
     with record_timer_block('rule.delete_rule'):
         try:
-            rule = session.query(models.ReplicationRule).filter(models.ReplicationRule.id == rule_id).with_for_update(nowait=nowait).one()
+            rule = (session.query(models.ReplicationRule)
```
bari12: It's only cosmetics, but can you please remove the enclosing parenthesis? Just for symmetry with the rest of the code. (The line-breaks are fine)

bari12 (changes requested):

ericvaandering:
I will. But this is now the preferred way. :-)

bari12 (approved):

Issue #3420, opened 2020-03-27
Python 3 overview ticket


bari12:
- [x] Activate Python 3.6 tests for server
- [x] Activate Python 3.7 tests for server (Which currently do not work)
- [x] Activate Python 3.8 tests for server (Which currently do not work)
- [x] DB initialisation fails with Python 3 #3480
- [x] Fix OIDC basestring issues #3714 
- [x] Fix all remaining 3.6 errors

bziemons:
Again, watch out: the tests with the `all` suite that look to be running on Python 3.7 (e.g. https://travis-ci.org/github/rucio/rucio/jobs/695958938) or Python 3.8 on Travis actually run with Python 2.7

bari12:
Yes I know, but this will be solved with the series of GH-Actions changes then 😄

bziemons:
Just mentioning it, because the merged PRs and the ticks above are kind of misleading :)

bari12:
Updated the text :-)

bziemons:
As the python3 tests are running on my branch for moving to GitHub Actions, here is a snapshot of the current problems. This is based on the next branch.

[10_test (centos7, 3.6, syntax).txt](https://github.com/rucio/rucio/files/4858217/10_test.centos7.3.6.syntax.txt)
[12_test (centos7, 3.6, all, oracle).txt](https://github.com/rucio/rucio/files/4858229/12_test.centos7.3.6.all.oracle.txt)
[13_test (centos7, 3.6, all, mysql5).txt](https://github.com/rucio/rucio/files/4858230/13_test.centos7.3.6.all.mysql5.txt)
[14_test (centos7, 3.6, all, mysql8).txt](https://github.com/rucio/rucio/files/4858231/14_test.centos7.3.6.all.mysql8.txt)
[15_test (centos7, 3.6, all, postgres9).txt](https://github.com/rucio/rucio/files/4858232/15_test.centos7.3.6.all.postgres9.txt)
[16_test (centos7, 3.6, all, postgres12).txt](https://github.com/rucio/rucio/files/4858234/16_test.centos7.3.6.all.postgres12.txt)
[17_test (centos7, 3.6, all, sqlite).txt](https://github.com/rucio/rucio/files/4858235/17_test.centos7.3.6.all.sqlite.txt)
[18_test (centos7, 3.6, client, sqlite).txt](https://github.com/rucio/rucio/files/4858236/18_test.centos7.3.6.client.sqlite.txt)
[...]

PR #3896, opened 2020-07-29
Core & Internals: Fix Python 3 syntax; #3420

bziemons:
Apart from fixing the Python 3 syntax problems, fixes isort dependency below version 5 as version 5 seems to be the default for Python 3 and pylint can currently not use isort >= 5.

### DIFF
--- a/bin/rucio
+++ b/bin/rucio
@@ -14,27 +14,32 @@
 # limitations under the License.
 #
 # Authors:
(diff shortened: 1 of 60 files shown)


cserf:
Oughh... Jumbo patch. Going through the list of files, there are a few of them that could be deleted (e.g. `generateDatasetsforPanda.py ` or `dq2client.py`). @bari12 we need to do a bit of cleanup at some point.

bari12 (approved):

PR #3907, opened 2020-08-05
Clients: CAUTION - change for requests call - #3866

### LINKED ISSUE #3866: Poor error message when the Rucio server certificate has expired.
Motivation
----------

I would like to understand why the Rucio client is unable to connect to the server, as this may facilitate diagnosing the problem and speed up its resolution.  For example, some problems the user can fix by themselves while other problems require changes on the server.

Modification
------------

The Rucio server at escape-dios-dl.cern.ch currently has an expired certificate:

```
paul@sprocket:~$ host escape-dios-dl.cern.ch
escape-dios-dl.cern.ch has address 188.184.86.179
paul@sprocket:~$ openssl s_client -connect escape-dios-dl.cern.ch:443 2>/dev/null </dev/null | openssl x509 -noout -dates
notBefore=Jun  5 08:52:03 2019 GMT
notAfter=Jul  9 08:52:03 2020 GMT
paul@sprocket:~$ 
```

While this is unfortunately, what is interesting is how the rucio client reacts to this problem, and reports back to the user:

```
paul@sprocket:~$ rucio whoami
2020-07-21 19:16:11,245 ERROR   Cannot connect to the Rucio server.
paul@sprocket:~$ 
```

While this is technically correct, it provides no information about why the client cannot connect to the server.  From the limited available information, the problem could be due to some networking problem, the rucio server's certificate not being trusted, certificate has expired, or some other problem.

It would be helpful if the error message contained some more details on why the connection failed.

TomasJavurek:
This just a suggestion how to fix #3866. Pls check the code carefuly, since the change could cause serious troubles.

### DIFF
--- a/lib/rucio/client/baseclient.py
+++ b/lib/rucio/client/baseclient.py
@@ -346,7 +346,8 @@ class BaseClient(object):
         elif 'content-type' in response.headers and response.headers['content-type'] == 'application/json':
             yield parse_response(response.text)
         else:  # Exception ?
-            yield response.text
+            if response.text:
+                yield response.text
 
     def _send_request(self, url, headers=None, type='GET', data=None, params=None, stream=False):
         """


mlassnig (approved):

bari12:
But does this fix the issue in a way Paul had in mind? E.g. does the Exception now include that the server certificate expired? (I did not test this, just asking)

mlassnig (changes requested):
There's a couple more changes needed across the whole baseclient. This is not the only case where self.request_retries is tested.

TomasJavurek:
> But does this fix the issue in a way Paul had in mind? E.g. does the Exception now include that the server certificate expired? (I did not test this, just asking)

@bari12 @mlassnig 
in my opinion yes, because with this patch, we avoid `ServerConnectionException` that Paul has seen, but there's the propper error message printed. Pls read my explanation at the parrent issue #3866 . I however haven't found a way to simulate exactly Paul's case in the Docker. Nevertheless, there is only this place in the baseclient where the ServerConnectionException is risen, so this has to be it.

bari12:
I think this solves a problem, but not the problem described in the issue. It is true that the exception is not raised, but for an identified error, it shouldn't be raised with the stacktrace, it just should describe the error. The stacktrace is helpful if it is an unknown problem.

But Pauls problem is that he is lacking information, e.g. in his case he would like to have something like: 
`Cannot connect to the Rucio server: The Server certificate expired.` 
The problem here is that this cannot be, fully, solved a the client level. The server must already give this information to the client. But we can prepare the client already to at least be ready to display that information.

PR #3926, opened 2020-08-10
Added m-VO database conversion script, Closes #3914

### LINKED ISSUE #3914: Script to convert DB between VOs
Motivation
----------

In a multi-VO instance accounts, scopes and RSEs need to be associated with a VO in the database. To preserve existing database entries for a community wishing to go from a single to multi VO instance, a script which can add the VO to all existing entries in the database would be useful.

Likewise if one VO in a multi-VO instance grows, it may be desirable for them to break off and setup their own dedicated instance. This should be possible with a similar script.

Modification
------------

- Scripts for renaming/removing entries associated with a particular VO (or not associated with any)
- Documentation on how to use them, and practical considerations (i.e. skipping large tables of historical data)

patrick-austin:
Added m-VO database conversion script
------------------

- Added script that generates SQL commands using SQLAlchemy that change references to the VO in the database so that the data from an existing instance can be used in a m-VO instance (or vice versa).
- Made some changes to `create_root_account()` so creating account counters is optional, and it doesn't raise an exception if identities are already present in the database (both would be an issue when using it for a m-VO database conversion).
- Added some documentation on how to use the script, and practical considerations (such as skipping large historical data to make the process faster).

### DIFF
--- a/doc/source/multi_vo_rucio.rst
+++ b/doc/source/multi_vo_rucio.rst
@@ -59,3 +59,116 @@ Managing VOs
 
 In addition to creating VOs, the description and email for a VO can be altered using ``update_vo``. If the root user of a VO loses access to their account, the super_root can
 associate a new identity with it using ``recover_vo_root_identity``. Finally, a list of current VOs and their descriptions is accessible via ``list_vos``.
+
+
+Converting Existing Instances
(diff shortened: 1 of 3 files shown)


bari12 (approved):
I am pulling this in as a patch to the current release branch as well. No need to wait for the next feature release.

Issue #3929, opened 2020-08-12
Upload fails on some sites due to Filesize mismatch - follow up 3927 


TomasJavurek:
Motivation
----------
This happens on some (DPM?) sites. Possibly due to a stat race condition, but not very clear what is happening. E.g.
https://aipanda184.cern.ch/condor_logs_2/20-08-11_01/grid.6845186.19.out

```
2020-08-11 02:19:39,649 | INFO     | Thread-6            | pilot.copytool.rucio             | _upload_item              | Successful upload of temporary file. gsiftp://bohr3226.tier2.hep.manchester.ac.uk:2811/dpm/tier2.hep.manchester.ac.uk/home/atlas/atlasscratchdisk/rucio/user/gsumi/ba/1d/user.gsumi.22262233.EXT0._091513.DAOD_MUON1.DAOD.pool.root.rucio.upload
2020-08-11 02:19:40,744 | WARNING  | Thread-6            | pilot.copytool.rucio             | _stage_out_api            | *** rucio API upload client failed ***
2020-08-11 02:19:40,745 | WARNING  | Thread-6            | pilot.copytool.rucio             | _stage_out_api            | caught exception: An unknown exception occurred.
Details: Filesize mismatch. Source: 73443479 Destination: 0
2020-08-11 02:19:40,867 | ERROR    | Thread-6            | pilot.copytool.rucio             | _stage_out_api            | Traceback (most recent call last):
  File "/scratch/condor_pool/condor/dir_14304/atlas_79CVDatu/pilot2/pilot/copytool/rucio.py", line 573, in _stage_out_api
    result = upload_client.upload([f], summary_file_path=summary_file_path, traces_copy_out=trace_report_out)
  File "/cvmfs/atlas.cern.ch/repo/ATLASLocalRootBase/x86_64/rucio-clients/1.23.1/lib/python2.7/site-packages/rucio/client/uploadclient.py", line 252, in upload
    sign_service=sign_service)
  File "/cvmfs/atlas.cern.ch/repo/ATLASLocalRootBase/x86_64/rucio-clients/1.23.1/lib/python2.7/site-packages/rucio/client/uploadclient.py", line 618, in _upload_item
    raise error
RucioException: An unknown exception occurred.
Details: Filesize mismatch. Source: 73443479 Destination: 0
```

Modification
------------
Firstly functional tests on xrood.stat for DPM sites in order to reproduce the cases when it returns 0 filesize. Solution accordint to this analysis.

[...]

Issue #3927, opened 2020-08-11
Upload fails on some sites due to wrong Filesize mismatch


bari12:
Motivation
----------
This happens on some (DPM?) sites. Possibly due to a stat race condition, but not very clear what is happening. E.g. 
https://aipanda184.cern.ch/condor_logs_2/20-08-11_01/grid.6845186.19.out

```
2020-08-11 02:19:39,649 | INFO     | Thread-6            | pilot.copytool.rucio             | _upload_item              | Successful upload of temporary file. gsiftp://bohr3226.tier2.hep.manchester.ac.uk:2811/dpm/tier2.hep.manchester.ac.uk/home/atlas/atlasscratchdisk/rucio/user/gsumi/ba/1d/user.gsumi.22262233.EXT0._091513.DAOD_MUON1.DAOD.pool.root.rucio.upload
2020-08-11 02:19:40,744 | WARNING  | Thread-6            | pilot.copytool.rucio             | _stage_out_api            | *** rucio API upload client failed ***
2020-08-11 02:19:40,745 | WARNING  | Thread-6            | pilot.copytool.rucio             | _stage_out_api            | caught exception: An unknown exception occurred.
Details: Filesize mismatch. Source: 73443479 Destination: 0
2020-08-11 02:19:40,867 | ERROR    | Thread-6            | pilot.copytool.rucio             | _stage_out_api            | Traceback (most recent call last):
  File "/scratch/condor_pool/condor/dir_14304/atlas_79CVDatu/pilot2/pilot/copytool/rucio.py", line 573, in _stage_out_api
    result = upload_client.upload([f], summary_file_path=summary_file_path, traces_copy_out=trace_report_out)
  File "/cvmfs/atlas.cern.ch/repo/ATLASLocalRootBase/x86_64/rucio-clients/1.23.1/lib/python2.7/site-packages/rucio/client/uploadclient.py", line 252, in upload
    sign_service=sign_service)
  File "/cvmfs/atlas.cern.ch/repo/ATLASLocalRootBase/x86_64/rucio-clients/1.23.1/lib/python2.7/site-packages/rucio/client/uploadclient.py", line 618, in _upload_item
    raise error
RucioException: An unknown exception occurred.
Details: Filesize mismatch. Source: 73443479 Destination: 0
```

bari12:
Needs more investigation.

PR #3930, opened 2020-08-12
Clients: stats after upload with write prot. - #3927

TomasJavurek:
Quick temporary fix. Needs to be follwed up in #3929


mlassnig (commented):
can you close this? i've fixed the google cloud problem, and already include this change as well

PR #3935, opened 2020-08-13
Core: Adding missing indices which only existed on Oracle as FB indic…

bari12:


mlassnig (commented):
what about the oracle sql extra file?

should be in there with CREATE INDEX ONLINE statement?

bari12:
They are already part of the oracle one as quite special functional indices. But they are not reflected at all in the models file. However, some of them are quite important for performant operations, so I put them now as (non-functional) indices.

mlassnig (approved):

PR #3947, opened 2020-08-21
Core & Internals: Add is_old_db method; Fix #1157

### LINKED ISSUE #1157: Add check that alembic version of DB works with rucio release
Motivation
----------
* Avoid errors when `alembic_version` is not set in the DB
* Make sure that the server code can understand the DB schema. 

Modification
------------

At start up the RESTful API checks whether the `alembic_version` is in a whitelist, or whether the code version is supported.

bziemons:
Add is_old_db method to test for an old alembic revision in the database.
Move alembic migration to tools/alembic_migration.sh.
Add is_old_db test in alembic_migration.sh.
Add startup tests for old DB in all daemons.
Add test_daemons.py for testing all daemons on that behavior.

### DIFF
--- a/lib/rucio/alembicrevision.py
+++ b/lib/rucio/alembicrevision.py
@@ -1,0 +1,19 @@
+# -*- coding: utf-8 -*-
+# Copyright 2020 CERN
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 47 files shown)


bziemons:
@cserf importing hermes and hermes2 at the same time now fails tests, because the prometheus metrics are named the same and are initialized on module import. Can this be fixed?

bari12 (approved):

PR #3949, opened 2020-08-24
core: fix long parameter in dids/search REST #3762

### LINKED ISSUE #3762: long=True query string does not work for /dids/{scope}/dids/search REST API
Motivation
----------
According to the RESTFul API documentation in:
https://rucio.readthedocs.io/en/latest/restapi/did.html#get--dids-(scope)-dids-search

one could pass `long=True` as a query string to get further information about DIDs, otherwise only the DID name gets returned. It seems to have no effect on the output, and I get only the DID names regardless of setting that flag to True or not.

Modification
------------
Either fix the code such that the `long` parameter acts as expected; or fix the documentation in case this parameter is not supposed to be there.

amaltaro:
Patch: Properly deal with the `long` parameter in the `dids/search` and `dids/search_extended` RESTful APIs (it expects a boolean, not 1 | 0).

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -47,6 +47,7 @@ Individual contributors to the source code
 - Eli Chadwick <eli.chadwick@stfc.ac.uk>, 2020
 - Patrick Austin <patrick.austin@stfc.ac.uk>, 2020
 - Rob Barnsley <R.Barnsley@skatelescope.org>, 2020
+- Alan Malta Rodrigues <alan.malta@cern.ch>, 2020
 
 Organisations employing contributors
 ------------------------------------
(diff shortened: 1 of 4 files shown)


amaltaro:
I had a look at a few automated tests logs and problems are apparently unrelated to this change.
Given that it's my first PR against this repository, could the Rucio team please let me know if anything needs attention and/or to be fixed?

bziemons:
Hi, thanks for taking this up! It is much appreciated.

The tests do seem to be failing because of this change. See for example the current didclient: https://github.com/rucio/rucio/blob/c6e5cf340753a40c1d774d786a6ff83645332582/lib/rucio/client/didclient.py#L72
it is still using the numeric 1 here. It would be great if you could look for all usages of the endpoint you are changing and then looking at the long parameter usage in those usages.

amaltaro:
Thanks for your feedback. I have updated the other missing locations and tests look better now.

bziemons (approved):

bari12:
Thanks for this PR 👍

PR #3953, opened 2020-08-25
Clients, bug: explicitly sort protocols by scheme, Fixes #3854

### LINKED ISSUE #3854: rucio-admin rse info does not return protocols and/or RSE usages
Motivation
----------

```
'<' not supported between instances of 'dict' and 'dict'
This means the parameter you passed has a wrong type.
```

Modification
------------

patrick-austin:
Explicitly sort protocols by scheme
------------------

When sorting the RSE protocols for the CLI to print, sort them by the value of `'scheme'` to prevent the error `'<' not supported between instances of 'dict' and 'dict'` that was occurring when simply sorting the dicts directly.

### DIFF
--- a/bin/rucio-admin
+++ b/bin/rucio-admin
@@ -561,13 +561,13 @@ def info_rse(args):
         print('  ' + attribute + ': ' + str(attributes[attribute]))
     print('Protocols:')
     print('==========')
-    for protocol in sorted(rseinfo['protocols']):
+    for protocol in sorted(rseinfo['protocols'], key=lambda x: x['scheme']):
         print('  ' + protocol['scheme'])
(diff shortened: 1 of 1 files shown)


dchristidis:
Would you kindly apply a similar fix for `usage` a few lines below?

patrick-austin:
Of course, I've pushed the change.

I only ever had a problem with `protocols` so I thought the issue was that when it couldn't sort those, it cut off everything below it (which included the `usage` section). But looking at it now, it's obvious the same problem could occur for `usage` as well.

dchristidis (approved):

dchristidis:
Thanks!

Issue #3954, opened 2020-08-25
Off by one error in core/transfer.py


ericvaandering:
Motivation
----------

Reported by Patrick Austin

It looks like this commit https://github.com/rucio/rucio/commit/b519409488ad2af867c866e4e72cd96ecdbf86bc
introduced a bug in core/transfer.py where another field was selected out of the database, but the destination RSE in line 1254 was not updated to reflect this.

Modification
------------

Change 9 to 10 in line 1254

PR #3962, opened 2020-08-26
Transfers: Fix an off by one bug introduced. # 3954

ericvaandering:
Pull request title
------------------

An earlier PR introduced a bug by shifting this index. See the description in #3954 

Should go into 1.23.5 as well as 1.24 (and hot fix if produced)


ericvaandering:
Oops. I see @patrick-austin also made this PR.

patrick-austin:
Ah sorry, I should've commented or self-assigned on the issue (I thought I might as well make the PR given I'd already made the change on the RAL instance). At least it's only a one line change so it's not too much time wasted...

PR #3963, opened 2020-08-27
Core & Internals: support policy packages for multi-VO Rucio #3542

jamesp-epcc:
This adds full support for policy packages in multi-VO Rucio. The main change is to the REST APIs where it is now necessary to generate separate URLs for each unique SCOPE_NAME_REGEXP in use. Permission and schema modules are now loaded on-demand when multi_vo is enabled. This is because the test scripts import the schema before the database is initialised so it is not possible to iterate over the VOs at that point.

### DIFF
--- a/lib/rucio/common/schema/__init__.py
+++ b/lib/rucio/common/schema/__init__.py
@@ -32,41 +32,110 @@ import importlib
 # dictionary of schema modules for each VO
 schema_modules = {}
 
-# TODO: load schema module for each VO in multi-VO installations
+# list of unique SCOPE_NAME_REGEXP values from all schemas
+scope_name_regexps = []
 
 try:
-    if config.config_get_bool('common', 'multi_vo'):
-        GENERIC_FALLBACK = 'generic_multi_vo'
-    else:
-        GENERIC_FALLBACK = 'generic'
+    multivo = config.config_get_bool('common', 'multi_vo')
 except (NoOptionError, NoSectionError):
+    multivo = False
+
+# multi-VO version loads schema per-VO on demand
+# we can't get a list of VOs here because the database might not
+# be available as this is imported during the bootstrapping process
+if not multivo:
     GENERIC_FALLBACK = 'generic'
 
-if config.config_has_section('policy'):
-    try:
-        POLICY = config.config_get('policy', 'package') + ".schema"
-    except (NoOptionError, NoSectionError):
-        # fall back to old system for now
+    if config.config_has_section('policy'):
         try:
-            POLICY = config.config_get('policy', 'schema')
+            POLICY = config.config_get('policy', 'package') + ".schema"
         except (NoOptionError, NoSectionError):
-            POLICY = GENERIC_FALLBACK
-        POLICY = 'rucio.common.schema.' + POLICY.lower()
-else:
-    POLICY = 'rucio.common.schema.' + GENERIC_FALLBACK.lower()
+            # fall back to old system for now
+            try:
+                POLICY = config.config_get('policy', 'schema')
+            except (NoOptionError, NoSectionError):
+                POLICY = GENERIC_FALLBACK
+            POLICY = 'rucio.common.schema.' + POLICY.lower()
+    else:
+        POLICY = 'rucio.common.schema.' + GENERIC_FALLBACK.lower()
 
-try:
-    module = importlib.import_module(POLICY)
-except ImportError:
-    raise exception.PolicyPackageNotFound('Module ' + POLICY + ' not found')
+    try:
+        module = importlib.import_module(POLICY)
+    except ImportError:
+        raise exception.PolicyPackageNotFound('Module ' + POLICY + ' not found')
 
-schema_modules["def"] = module
+    schema_modules["def"] = module
+    scope_name_regexps.append(module.SCOPE_NAME_REGEXP)
 
 
-# TODO: in both of these functions, verify that named module exists
+def load_schema_for_vo(vo):
+    GENERIC_FALLBACK = 'generic_multi_vo'
+    if config.config_has_section('policy'):
+        try:
+            POLICY = config.config_get('policy', 'package-' + vo) + ".schema"
+        except (NoOptionError, NoSectionError):
+            # fall back to old system for now
+            try:
+                POLICY = config.config_get('policy', 'schema')
+            except (NoOptionError, NoSectionError):
+                POLICY = GENERIC_FALLBACK
+            POLICY = 'rucio.common.schema.' + POLICY.lower()
+    else:
+        POLICY = 'rucio.common.schema.' + GENERIC_FALLBACK.lower()
+
+    try:
+        module = importlib.import_module(POLICY)
+    except ImportError:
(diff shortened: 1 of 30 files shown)


bari12:
Hi @jamesp-epcc 
In general this looks good to me; Please rebase your branch and have a look on the failing tests as well.

jamesp-epcc:
> Hi @jamesp-epcc
> In general this looks good to me; Please rebase your branch and have a look on the failing tests as well.

@bari12 The tests are all passing now that the rebase is done. I think it was just the git cherry pick failing before.

bari12:
Hi @jamesp-epcc 
Sorry, I have another change-request. In the meanwhile I merged #3975 which was a long-planned change since we want to move from web.py to Flask as web backend. You added the functionality already to the web.py files, but didn't do it for the Flask ones (yet). At this point we should really try to keep flask and web.py completely in sync, since it should be interchangeable and eventually allow us to drop web.py as a backend completely. Can you please amend your PR with this. Sorry for the late notice about this.

lib/rucio/common/schema/__init__.py
```diff
@@ -32,41 +32,106 @@
...
     schema_modules[vo].validate_schema(name, obj)
 
 
 def get_schema_value(key, vo='def'):
+    if vo not in schema_modules:
+        load_schema_for_vo(vo)
     return getattr(schema_modules[vo], key)
+
+
+# returns a list of all unique SCOPE_NAME_REGEXPs from all schemas
```
bziemons: function comments should be added as doc strings to the function, like this:
```python
def get_scope_name_regexps():
    """ load schemas for all VOs here and add unique scope_name_regexps to list """
    …
```

lib/rucio/common/schema/__init__.py
```diff
@@ -32,41 +32,106 @@
...
+                load_schema_for_vo(vo['vo'])
+            scope_name_regexp = schema_modules[vo['vo']].SCOPE_NAME_REGEXP
+            if scope_name_regexp not in scope_name_regexps:
+                scope_name_regexps.append(scope_name_regexp)
+    return scope_name_regexps
+
+
+# given a tuple of URLs for webpy with '%s' as a placeholder for SCOPE_NAME_REGEXP,
+# return a finalised tuple of URLs that will work for all SCOPE_NAME_REGEXPs in all
+# schemas
```
bziemons: move to docstr, see above

lib/rucio/core/rse_expression_parser.py
```diff
@@ -33,15 +33,14 @@
...
 from rucio.common.config import config_get
 from rucio.common.exception import InvalidRSEExpression, RSEBlacklisted
 from rucio.core.rse import list_rses, get_rses_with_attribute, get_rse_attribute
 from rucio.db.sqla.session import transactional_session
 
 
-DEFAULT_RSE_ATTRIBUTE = schema.get_schema_value('DEFAULT_RSE_ATTRIBUTE')['pattern']
-RSE_ATTRIBUTE = schema.get_schema_value('RSE_ATTRIBUTE')['pattern']
+DEFAULT_RSE_ATTRIBUTE = r'([A-Z0-9]+([_-][A-Z0-9]+)*)'
+RSE_ATTRIBUTE = r'([A-Za-z0-9\._-]+[=<>][A-Za-z0-9_-]+)'
```
bziemons, line 43: is this known to all communities who might have that configured in their schema?

lib/rucio/web/rest/flaskapi/v1/common.py
```diff
@@ -105,11 +105,13 @@ def parse_scope_name(scope_name):
...
-    if scope_name is None:
-        raise ValueError('cannot parse scope and name')
-    return scope_name.group(1, 2)
+    scope_name_regexps = get_scope_name_regexps()
+    for regexp in scope_name_regexps:
+        # why again does that regex start with a slash?
+        result = re.match(regexp, '/' + scope_name)
+        if result is not None:
+            return result.group(1, 2)
+    raise ValueError('cannot parse scope and name')
```
bziemons: shouldn't the method instead take the vo as a parameter and select the correct `scope_name` regex from that? Also please keep the documentation of the function up-to-date.

lib/rucio/web/rest/webpy/v1/did.py
```diff
@@ -1099,6 +1099,11 @@ def DELETE(self, scope, name):
    Web service startup
 ----------------------"""
 
+logfile = open('/tmp/rucio_log_jtp', 'w')
+for url in URLS:
+    logfile.write(url + "\n")
+logfile.close()
```
bziemons: should this still be in here? debug code?

bziemons (changes requested):
From what I can see you might have forgotten at least the URLS in `webpy/v1/main.py`. Please check, if there are more like that.

lib/rucio/common/schema/__init__.py
jamesp-epcc: Thanks, I will fix that.

lib/rucio/common/schema/__init__.py
jamesp-epcc: I will also fix this one.

lib/rucio/core/rse_expression_parser.py
jamesp-epcc: I discussed this with Martin a few weeks ago, he said it had never been used and could safely be removed.

lib/rucio/web/rest/flaskapi/v1/common.py
jamesp-epcc: Yes, that's probably a better approach for the flask version. (It wasn't possible to do it that way for webpy since the URLs have to be registered at start up). I will make the change.

lib/rucio/web/rest/webpy/v1/did.py
jamesp-epcc: Well spotted, it should not be there. I will remove it.

jamesp-epcc:
Thanks also for pointing out the URLs in `main.py` - those weren't there when I started working on this so I hadn't noticed them.

lib/rucio/web/rest/flaskapi/v1/common.py
bziemons: not all endpoints have the `request.environ['vo']` set, because that is set with the same hook that checks authentication. You could get the header manually in those cases.

jamesp-epcc: I've done that for the redirects.

bziemons (dismissed):
Thanks for the requested changes! Looks good to me now

PS: @bari12 would probably want to have the changes squashed
PPS: But he can talk for himself :D

jamesp-epcc:
I have squashed the commits to tidy this up a bit.

ericvaandering:
This commit breaks CMS. @nsmith-  this is where the issue is coming from. The problem is in the changes to lib/rucio/core/rse_expression_parser.py 

CMS RSEs are not all upper case. It looks like instead of querying the schema, something was chosen and what was chosen was from the generic schema not the schema actually in use. For CMS this could be remedied with a simple addition of a-z in a place or two, but there should be a more general fix.

PR #3964, opened 2020-08-27
Testing: Fix tests in docker-compose dev env

bziemons:
Tests currently fail when running tools/run_tests_docker.sh inside the docker dev environment:
* XFAIL multi_vo tests without multi_vo instead of logging a warning.
* Add RDBMS variable to rucio-dev container for tests with database-specific checks.


patrick-austin (approved):
The changes to `test_multi_vo.py` look good to me

PR #3969, opened 2020-08-28
Update language

ericvaandering:
This PR is in two commits. The first corrects some user facing language I noticed. The second commit refactors the code base to change the name of the problematic exception name

### DIFF
--- a/lib/rucio/web/ui/static/request_rule.js
+++ b/lib/rucio/web/ui/static/request_rule.js
@@ -7,6 +7,7 @@
  *
  * Authors:
  * - Thomas Beermann, <thomas.beermann@cern.ch>, 2015-2018
+ * - Eric Vaandering, <ewv@fnal.gov>, 2020
  */
 
 var selected_dids = [];
@@ -886,7 +887,7 @@ create_rules = function(approval) {
         if (error_details['ExceptionClass'] == 'DuplicateRule') {
             html_check = '<font color="red">The rule that you are trying to create already exists, please go back and choose another DID or RSE.</font>';
         } else if (error_details['ExceptionClass'] == 'RSEBlacklisted') {
-            html_check = '<font color="red">The RSE that you chose is currently blacklisted. Please check <a href="http://atlas-agis.cern.ch/agis/ddmblacklisting/list/" target="_blank">here</a> for more information.</font>';
+            html_check = '<font color="red">The RSE that you chose is currently not available for writing. Please check <a href="http://atlas-agis.cern.ch/agis/ddmblacklisting/list/" target="_blank">here</a> for more information.</font>';
         } else {
             html_check = '<font color="red">' + error_details['ExceptionClass'] + ': ' + error_details['ExceptionMessage'] + '</font>';
         }
--- a/lib/rucio/web/ui/static/request_rule.js
+++ b/lib/rucio/web/ui/static/request_rule.js
@@ -886,7 +886,7 @@ create_rules = function(approval) {
         error_details = JSON.parse(jqXHR.responseText);
         if (error_details['ExceptionClass'] == 'DuplicateRule') {
             html_check = '<font color="red">The rule that you are trying to create already exists, please go back and choose another DID or RSE.</font>';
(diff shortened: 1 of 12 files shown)


lib/rucio/web/ui/static/request_rule.js
```diff
@@ -885,8 +886,8 @@ create_rules = function(approval) {
         error_details = JSON.parse(jqXHR.responseText);
         if (error_details['ExceptionClass'] == 'DuplicateRule') {
             html_check = '<font color="red">The rule that you are trying to create already exists, please go back and choose another DID or RSE.</font>';
-        } else if (error_details['ExceptionClass'] == 'RSEBlacklisted') {
-            html_check = '<font color="red">The RSE that you chose is currently blacklisted. Please check <a href="http://atlas-agis.cern.ch/agis/ddmblacklisting/list/" target="_blank">here</a> for more information.</font>';
+        } else if (error_details['ExceptionClass'] == 'RSENoWriting') {
+            html_check = '<font color="red">The RSE that you chose is currently not available for writing. Please check <a href="http://atlas-agis.cern.ch/agis/ddmblacklisting/list/" target="_blank">here</a> for more information.</font>';>>>>>>> c78798acb... More language
```
bziemons: a little error has sneaked in here

ericvaandering: Indeed!

bziemons:
Also if I may suggest: instead of `NoWriting` I find `ReadOnly` much more natural, but that may be a personal preference (or perhaps `RSEDeniesWrite` if this is not as simple as read/write).

bari12:
Thanks @ericvaandering 
I was also planning to submit a PR for this :-) Next to `RSENoWriting`, `RSEReadOnly`, and `RSEDeniesWrite` I am also throwing `RSEWriteBlocked` 😄 
`RSEReadOnly` sounds most natural, but might be misleading, since the RSE might actually be completely blocked, thus also not allows Reading.

ericvaandering:
Just let me know and I can easily update. I like RSEWriteBlocked.

ericvaandering:
Exception name changed to RSEWriteBlocked. Please take once you are satisfied.

bari12:
Thanks @ericvaandering 
There is a bit of a backwards-compatibility issue here which we probably can't easily address.
On the client side the [translation](https://github.com/rucio/rucio/blob/bfafc84bd616857d853cec5b5bd5b3cc41f45c80/lib/rucio/client/baseclient.py#L308) of the Exception sent by the REST interface is based on the exception class name. So if the server suddenly starts to send `RSEWriteBlocked` older clients will not be able to understand it and raise generic `RucioException` instead. 
In retrospective it would have been better to do the exception translation by the unique exception code, this would allow easy rename. Of course that doesn't help us now with old clients.

ericvaandering:
I thought that might be the case, which is why I did two commits. The first one should be completely safe. Can you cherry pick it or do you want me to make a new PR?

bari12:
I would suggest, we rename whatever we can without breaking backwards-compatibility. (E.g. the messages etc.) We do add a new `RSEWriteBlocked` exception, add it in all the catch cases, but do not raise it on the server yet. In some, as short as possible, time (e.g. with 1.24?) we remove the old exception and only use `RSEWriteBlocked`. This way we give the clients a time window to upgrade.
Older clients will not break completely, they will just raise a `RucioException` instead, but I would imagine that the workflow of certain external applications would not work as intended with that.

ericvaandering:
OK, I'll update with that in mind.

ericvaandering:
OK, this should be OK with one possible exception: I left the error_code for the exception the same since I don't really know what that is and they will be used to signal the same thing.

bari12:
```
lib/rucio/daemons/sonar/sonar/send_sonar_tests.py:86:43: E241 multiple spaces after ','
lib/rucio/common/exception.py:589:1: E302 expected 2 blank lines, found 1
```
:-)

ericvaandering:
Fixed.

bari12 (approved):

Issue #1817, opened 2018-11-23
Add functional Flask endpoints


hahahannes:
Motivation
----------
The Flask API has problems with dealing exceptions when using JSON streaming. This should be fixed as it is essential for handling lot of data.

Modification
------------
See [documentation](http://flask.pocoo.org/docs/1.0/patterns/streaming/?highlight=stream) for implementing JSON streaming to change the Flask API.
Investigate how to deal with exceptions.

hahahannes:
It seems that it is not possible to catch exceptions inside the streaming.
To reproduce, run this minimal flask server. The exception gets raised on flask side but does not get catched. An internal server error is generated instead of the 'error' string.

```python
from flask import Flask, Response, jsonify

app = Flask(__name__)

@app.route("/")
def hello():
    def generate(): 
        for i in range(0, 100):   
            raise Exception
            yield '1' + '\n'    
    
    try:
        return Response(generate())
    except Exception as error:
        return jsonify({'error': error})

if __name__ == '__main__':
    app.run(passthrough_errors=True)
```

```
FLASK_APP=file.py flask run
```

One solution would be to rework the logic and check for possible problems like non existing account before querying the database. 
Another solution would be to call the same function inside a for loop but with a break after one loop, where we check for exceptions. Afterward the streaming can start if there was no exception catched.

bari12:
To me the accurate path looks like the one where we change the logic (and checking) of the APIs which use streaming. Thus we do all the conflict checking first and then just iterate the stream.

Conceptually, the error handling we do right now with streaming is wrong and possibly leads to malformed replies. Because the HTTP return code is calculated (and sent!) at the header anyway. So let's say this is a `200`, whatever you stream afterwards (and throws an exception) is not properly reflected in the error code.

But this will take time, we probably have to more carefully select which APIs we offer as a streaming API (list-contents, list-replicas, ...) and not be so generous like we are now with them.

vingar:
Worth mentioning here to think about proper support for HTTP Chunked Encoding.

bziemons:
@bari12 *after* my current tasks I would like to take care of this. I debugged a little on the issue where there is only a generic internal server error transmitted to the client and what you pointed out above is most probably the cause of the problem.

bari12:
👍 much appreciated. We need to plan and approach this carefully though. Especially with list_replicas it won't be easy.

bziemons:
Just talked to Mario about it. He suggested the following tasks and their order:
[...]

PR #3975, opened 2020-09-02
Core & Internals: Fix Flask endpoints; #1817

bziemons:
Rename lock, replica blueprints, because of duplicate naming resulting in errors when adding both to the same application.
Fix Flask endpoints compared to web.py.
Add global account limit endpoint to Flask.
Add `parse_scope_name` for parsing scope and name following the schema.
Add `try_stream` for all streaming endpoints.
Fix did endpoint and add missing follow endpoint.
Fix archive endpoint.
Add missing authentication endpoints.

Caveats
-----------

OIDC authentication templates were created for web.py only. The outputs on the Flask side are simple one-liners in text instead of HTML. As those endpoints shouldn't be handled as REST APIs, this doesn't break anything.

### DIFF
--- a/lib/rucio/web/rest/flaskapi/v1/archive.py
+++ b/lib/rucio/web/rest/flaskapi/v1/archive.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python
-# Copyright 2018 CERN for the benefit of the ATLAS collaboration.
+# -*- coding: utf-8 -*-
+# Copyright 2017-2020 CERN
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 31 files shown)


bziemons:
Can be reviewed, but I am going to have another look at the generated documentation, since I am changing a few endpoints here.

bziemons:
ready for review & merge. I would suggest merging this as a patch, because that will prevent issues with cherry-picking to the release branch & fix documentation issues.

bari12:
@tbeerman please have a look on this as well.

lib/rucio/web/rest/flaskapi/v1/archive.py
```diff
@@ -42,27 +46,32 @@ class Archive(MethodView):
     """ REST APIs for archive. """
 
     @check_accept_header_wrapper_flask(['application/x-json-stream'])
-    def get(self, scope, name):
+    def get(self, scope_name):
```
tbeerman, line 49: why this change?

bziemons: because schemas can configure an arbitary `SCOPE_NAME_REGEX`[(1)] and this is backwards-compatible.

[(1)]: https://github.com/rucio/rucio/blob/4ab624fd77949d97a3498e2d31ce58bfd6eef135/lib/rucio/common/schema/cms.py#L391

tbeerman: ah right, I forgot about the schema.

bziemons: I am all open for a better solution, if you have any :D

tbeerman: not really. I just missed the step with the schema matching. all good from my side.

tbeerman (approved):

lib/rucio/web/rest/flaskapi/v1/exporter.py
```diff
@@ -71,6 +74,7 @@ def get(self):
 
 export_view = Export.as_view('scope')
 bp.add_url_rule('/', view_func=export_view, methods=['get', ])
+# FIXME: Add '' rule
```
bari12, line 77: What are these for?

bari12 (approved):
This looks very good. Thanks a lot @bziemons 👍 
Due to the size of it we will have to test it carefully though. Next by adding it to the tests but then also by partial integration deployment.

lib/rucio/web/rest/flaskapi/v1/exporter.py
bziemons: Because the blueprints do not have the url prefix, adding this rule would currently break the blueprint since it would add a rule for an empty url. These are basically notes for myself I followed up on #3990

Issue #3965, opened 2020-08-27
Integrity error when deleting last replica


davidgcameron:
Motivation
----------
When deleting the last replica of a file I see the following traceback in the reaper log:

```
2020-08-27 13:05:44,961 6       INFO    Thread [0/1] :  Deletion ATTEMPT of test-v2.1:mc_v12-4GeV-1e-ecal_photonuclear-test_t1594763716.root as file:///projects/hep/fs9/shared/ldmx/ldcs/output/l
dmx/mc-data/v12/4.0GeV/v2.1test-2020-07-14/mc_v12-4GeV-1e-ecal_photonuclear-test_t1594763716.root on LUND
2020-08-27 13:05:44,961 6       WARNING Thread [0/1] :  Deletion NOTFOUND of test-v2.1:mc_v12-4GeV-1e-ecal_photonuclear-test_t1594763716.root as file:///projects/hep/fs9/shared/ldmx/ldcs/output/
ldmx/mc-data/v12/4.0GeV/v2.1test-2020-07-14/mc_v12-4GeV-1e-ecal_photonuclear-test_t1594763716.root on LUND
2020-08-27 13:05:44,961 6       INFO    Thread [0/1] :  1 files processed in 0.000700950622559 seconds
(psycopg2.IntegrityError) update or delete on table "dids" violates foreign key constraint "DID_META_FK" on table "did_meta"
DETAIL:  Key (scope, name)=(test-v2.1, mc_v12-4GeV-1e-ecal_photonuclear-test_t1594763716.root) is still referenced from table "did_meta".

[SQL: DELETE FROM dids WHERE dids.scope = %(scope_1)s AND dids.name = %(name_1)s]
[parameters: {'name_1': 'mc_v12-4GeV-1e-ecal_photonuclear-test_t1594763716.root', 'scope_1': 'test-v2.1'}]
(Background on this error at: http://sqlalche.me/e/gkpj)
2020-08-27 13:05:45,182 6       CRITICAL        Thread [0/1] :  Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/rucio/daemons/reaper/reaper2.py", line 523, in reaper
    delete_replicas(rse_id=rse_id, files=deleted_files)
  File "/usr/lib/python2.7/site-packages/rucio/db/sqla/session.py", line 374, in new_funct
    raise DatabaseException(str(error))
DatabaseException: Database exception.
[...]

PR #3976, opened 2020-09-02
Deletion: also remove did metadata when deleting a replica #3965

tbeerman:


lib/rucio/core/replica.py
```diff
@@ -1584,6 +1586,12 @@ def delete_replicas(rse_id, files, ignore_availability=True, session=None):
                                                      RuleState.WAITING_APPROVAL))).\
             delete(synchronize_session=False)
 
+    # Remove DID Metadata
+    for chunk in chunks(deleted_did_meta, 100):
```
bari12, line 1596: Here something similar to 
https://github.com/rucio/rucio/blob/6189936fd27ada4c729b7cc15e061387ec7121cb/lib/rucio/core/did.py#L616
should be done.

bari12 (approved):

PR #3979, opened 2020-09-03
Core: Fix for list-dids backwards compatibility issue; Fix #3978

### LINKED ISSUE #3978: Fixing backwards compatibility issue with list-dids
Motivation
----------
#3762 broke backwards compatibility for old clients for the `list-dids` command. 

Modification
------------
Check for `long=='1'` additionally to `long=='True'`.

bari12:


lib/rucio/web/rest/webpy/v1/did.py
```diff
@@ -203,7 +206,7 @@ def GET(self, scope):
                 elif k == 'long':
                     long = v[0] == 'True'
                 elif k == 'recursive':
-                    recursive = v[0] == 'True'
+                    recursive = v[0] in ['True', '1']
```
bziemons: wrong line ;)

bari12: 🤦

bziemons (commented):
Looks fine to me. Personally I find it hard to differentiate what has priority: documentation or implementation. To me it would make sense to add this problem to the documentation and to advise against using the numeric 1 from now on, so that it can be deprecated and removed at some point.

Issue #3912, opened 2020-08-06
Unconditionally reducing the log level omits important information even under '-v' 


knservis:
The following reduces the log level, even when the user has requested '-v' for verbose (which also sets the log level to DEBUG).

https://github.com/rucio/rucio/blob/7820df402a54bcf1436173fa517f79b9bcc8986b/lib/rucio/client/baseclient.py#L96

In addition the following line:
https://github.com/rucio/rucio/blob/7820df402a54bcf1436173fa517f79b9bcc8986b/lib/rucio/client/baseclient.py#L376

Would be much more informative using `LOG.exception()`

TomasJavurek:
Exception is perhaps not the ideal one, since we use custom Formatter. I guess exception would print the warning as an error, which might be inpropriete (I believe the developer that coded this placed the warning for a reason). But the question is why the debug is temporarily switched off, and maybe rather switch that back here.

TomasJavurek:
An obvious solution would be to set the log level globaly: 
`logging.basicConfig(level=logging.DEBUG)`
on line:
https://github.com/rucio/rucio/blob/release-1.23-LTS/bin/rucio#L2365
but that would produce uncontralled tracks with no correct formater like the following, rucio - whoami reseulted to:

`DEBUG:rucio.client.baseclient:no creds passed. Trying to get it from the config file.
DEBUG:rucio.client.baseclient:no ca_cert passed. Trying to get it from the config file.
DEBUG:rucio.client.baseclient:no account passed. Trying to get it from the config file.
DEBUG:rucio.client.baseclient:no vo passed. Trying to get it from the config file.
DEBUG:rucio.client.baseclient:got token from file
DEBUG:dogpile.lock:NeedRegenerationException
DEBUG:dogpile.lock:no value, waiting for create lock
DEBUG:dogpile.lock:value creation lock <dogpile.cache.region._LockWrapper object at 0x7f72e0c7c8d0> acquired
DEBUG:dogpile.lock:Calling creation function for not-yet-present value
DEBUG:dogpile.lock:Released creation lock
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): rucio:443
DEBUG:urllib3.connectionpool:https://rucio:443 "GET /accounts/whoami HTTP/1.1" 303 0
DEBUG:urllib3.connectionpool:https://rucio:443 "GET /accounts/root HTTP/1.1" 200 197`

TomasJavurek:
[...]

PR #3984, opened 2020-09-04
Clients: log level propagation - #3912

TomasJavurek:
Propagation of log level
------------------------

- the log level has to be propagated through all clients downstream to baseclient, which is inheriting from them
- further propagation to rsemgr still has to be done
- distinguishing log level by server/client mode and verbosity set in bin/rucio
    - the full logic can be seen at the end of bin/rucio
- the log level is decided centrally in bin/rucio for all the downstream modules
- if cli is not used, there's a default within setup_logger method in common/utils.py, this would be used for api calls.

Outcomes:
----------
- #3912 is solved by this
- The general formatter is now applied to almost all loggers in clients.
- There should be way less logger load on the servers when testing.

### DIFF
--- a/lib/rucio/common/.vim_runtime/my_configs.vim
+++ b/lib/rucio/common/.vim_runtime/my_configs.vim
@@ -1,0 +1,1 @@
+set mouse-=a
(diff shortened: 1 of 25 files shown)


lib/rucio/common/.vim_runtime/my_configs.vim
```diff
@@ -0,0 +1 @@
+set mouse-=a
```
bziemons, line 1: I think this file shouldn't make it into the repo

TomasJavurek: No, not yet until the the propper levels for testing and server mode are set. Pls check: #3912

bari12: No, that vim file shouldn't be in the repo at all 😄

bari12:
Closing; Will be re-submitted with the changes discussed in dev meeting

PR #3995, opened 2020-09-15
Core & Internals: Fix userpass salts with py3; Fix #3994

### LINKED ISSUE #3994: Userpass authentication not backwards compatible nor unicode capable
Motivation
----------
The current implementation for userpass authentication results in two different hashes for Python 2 and Python 3. The following code lines are affected (and can be found in add_identity as well):

https://github.com/rucio/rucio/blob/7ea877bd8b226012fe2441611aaf82b35c9df82b/lib/rucio/core/authentication.py#L97-L101

Running on Python 2:
```
db_salt = b'as\xab as'  # salt from the database is a bytes object: Column(LargeBinary(255))
password = 'asdasd'
salted_password = '%s%s' % (db_salt, password)
```
Results in `salted_password = 'as\xab asasdasd'` and the hash `'13a429a9643f23f5e0b5ec9cbad8da39d30d2939b391379b9b6763a65a8d0d80'`

In Python 3 (db_salt and password from above):
```
db_salt = b64encode(db_salt).decode()
salted_password = ('%s%s' % (db_salt, password)).encode()
```
Results in `salted_password = 'YXOrIGFzasdasd'` and the hash `'82e5c392dd4dccef16c64c6a58c6059a5cf34c6311e379b342205f35616b958f'`

which means that passwords already hashed on Python 2 and stored in the database are not compatible with using a Python 3 server.

Modification
------------
Instead of unnecessary differentiating between Python 2 and Python 3, bytes should be used directly:
`salted_password = db_salt + password.encode()`

bziemons:
Fix userpass salts with Python 3.

**This invalidates existing userpass password hashes stored in databases with Python 3 servers!**
Tested the new method successfully with Python 2.7.5, 3.6.8 and 3.8.5.

### DIFF
--- a/lib/rucio/core/authentication.py
+++ b/lib/rucio/core/authentication.py
@@ -33,10 +33,9 @@ import hashlib
 import random
 import sys
 import traceback
-from base64 import b64decode, b64encode
+from base64 import b64decode
 
 import paramiko
-import six
 from dogpile.cache import make_region
 from dogpile.cache.api import NO_VALUE
 from sqlalchemy import and_, or_
@@ -94,12 +93,7 @@ def get_auth_token_user_pass(account, username, password, appid, ip=None, sessio
     db_salt = result['salt']
     db_password = result['password']
 
-    if six.PY3:
-        db_salt = b64encode(db_salt).decode()
-        salted_password = ('%s%s' % (db_salt, password)).encode()
-    else:
-        salted_password = '%s%s' % (db_salt, password)
-
+    salted_password = db_salt + password.encode()
     if db_password != hashlib.sha256(salted_password).hexdigest():
         return None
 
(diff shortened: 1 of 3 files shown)


bari12:
This is a bit problematic, since these communities need to completely re-collect their users passwords and salt them again. To be fair, I don't think there is anyone who really uses username+password in production, but there might be some established testing instances which are impacted by this. So we should do this only with a feature release.
There is no way to get py2-py3 in sync otherwise I assume?

bziemons:
>  There is no way to get py2-py3 in sync otherwise I assume?

Checking user passwords against two hashes seems harmful to security and since the damage is already done, I guess not.

bari12:
Right, cutting the search-space in half is definitely not something we should do. I'll merge this for the 1.24 release and put an advise in the release note.

bari12 (approved):

bziemons:
Let's also not forget that this is fixing a bug for people switching between Python versions on the server.

bari12:
True, let's mention it in the next Dev Meeting as well.

PR #3999, opened 2020-09-16
Transfers: Multihop graph protection for negative ranking; Fix #3997

### LINKED ISSUE #3997: Protection for negative ranking needed in multihop resolve algorithm
Motivation
----------
Currently the algorithm might end up in an infinite loop if negative `ranking` values are stored to the `distances` table.

Modification
------------
Protection against negative values. Assume 0 instead.

bari12:


lib/rucio/core/transfer.py
```diff
@@ -1448,10 +1448,13 @@ def __load_distance_edges_node(rse_id, session=None):
         for distance in session.query(models.Distance).join(models.RSE, models.RSE.id == models.Distance.dest_rse_id)\
                                .filter(models.Distance.src_rse_id == rse_id)\
                                .filter(models.RSE.deleted == false()).all():
+            ranking = None
```
bziemons: does this make sense? Is the algorithm okay with `None` values in the distance_graph?

bari12: It is covered by the algorithm which is just skipping these entries. But it is a good point actually, it is better to just skip `None` entries here entirely, so they don't even have to be iterated in the algorithm later.

lib/rucio/core/transfer.py
```diff
@@ -1448,10 +1448,13 @@ def __load_distance_edges_node(rse_id, session=None):
         for distance in session.query(models.Distance).join(models.RSE, models.RSE.id == models.Distance.dest_rse_id)\
                                .filter(models.Distance.src_rse_id == rse_id)\
                                .filter(models.RSE.deleted == false()).all():
+            if ranking is None:
```
bziemons: `if distance.ranking is None:`

bziemons (changes requested):

lib/rucio/core/transfer.py
bari12: 🤯

bziemons (approved):

PR #4001, opened 2020-09-16
Core: Fix caching collision for rse_info and rse_attributes; Fix #3998

### LINKED ISSUE #3998: Caching collisions with rse_attributes and rse_info
Motivation
----------
There is a caching collision with the rse_info values as well as the rse_attributes in the conveyor. Both dictionaries are stored in memcache under the `rse_id` key, thus leading to collisions.

https://github.com/rucio/rucio/blob/e7c4def359e1aeb413966e383b83ac9887f3c728/lib/rucio/rse/rsemanager.py#L96

https://github.com/rucio/rucio/blob/e7c4def359e1aeb413966e383b83ac9887f3c728/lib/rucio/common/rse_attributes.py#L52

Modification
------------
Use proper keys for both.

bari12:


lib/rucio/common/rse_attributes.py
```diff
@@ -1,13 +1,25 @@
-# Copyright European Organization for Nuclear Research (CERN)
+# -*- coding: utf-8 -*-
```
mlassnig, line 1: what's this line?

bziemons: Setting the encoding of the file to utf-8 for Python 2. Python 3 already has this as a default. It allows special characters in the file. I added it to the add_header script because some author names had special characters and that's now fine with this.

mlassnig: hmmmm, not sure about this... i'd rather not add anything for python2 at this stage anymore. @bari12 ?

bari12: I would leave it, probably even post 1.24. We do still keep Python 2.7 compatibility for the clients, so in theory it might be needed there and doesn't hurt really. What I would remove though, at least with 1.24 is the `# PY3k Compatible` line.

bziemons: Funny that you mention the `# PY3K COMPATIBLE` line, because I removed that from the add_header script along with the python3 test suite in my Python 3.7 branch :)

lib/rucio/common/rse_attributes.py
```diff
@@ -44,12 +52,13 @@ def get_rse_attributes(rse_id, session=None):
     :returns: A dictionary with RSE attributes for a RSE.
     """
 
-    result = REGION.get(rse_id)
+    key = 'rse_attributes_%s' % (str(rse_id))
```
bziemons: why convert `rse_id` to `str` explicitly?

lib/rucio/rse/rsemanager.py
```diff
@@ -35,6 +36,7 @@
 # - Dimitrios Christidis <dimitrios.christidis@cern.ch>, 2020
 # - Patrick Austin <patrick.austin@stfc.ac.uk>, 2020
 # - Benedikt Ziemons <benedikt.ziemons@cern.ch>, 2020
+# - javor <33832672+TomasJavurek@users.noreply.github.com>, 2020
```
bziemons: this should probably be added to the `add_header`s special mappings

bziemons (commented):
since `REGION` is just a cache, the rest seems fine to me

bziemons (approved):

PR #4005, opened 2020-09-16
xrootd: stat broken on py3; Fix #4004

### LINKED ISSUE #4004: xrootd: stat broken on py3
Motivation
----------
The python3 response from the stat is of type byte, which requires a modification of split() and decode(). This is backward compatible to py2.

mlassnig:


bziemons (approved):
Looks good :) btw. just because I stumbled accross `six.ensure_text` and in case someone does not know: `six.ensure_text` is a nice and readable way of ensuring that the input (str/byte/unicode) is converted to unicode-only as the output for both py2/3. There is also `six.ensure_str` and `six.ensure_binary`.

Issue #4039, opened 2020-09-30
Scope check in upload client crashes in certain conditions


bari12:
Motivation
----------
https://github.com/rucio/rucio/blob/fe6eb2e9cd7e1bc9464b955aa0a7d974c961fe82/lib/rucio/client/uploadclient.py#L350
Here the uploadclient checks if the account owns the scope it is uploading to and if not displays a warning message.
However, due to this check is done, if the account does not own any scopes, the `ScopeNotFound` exception is raised which crashes the uploadclient at that point.

Modification
------------
Catch that exception, display the warning but continue with the upload otherwise.

TomasJavurek:
The suggestion from Martin is implemented, but I think it didn't solve the problem entirely. If we don't raise exception here, we get the prblem at db level later:

`Details: [u'(psycopg2.errors.ForeignKeyViolation) insert or update on table "dids" violates foreign key constraint "DIDS_SCOPE_FK"\nDETAIL:  Key (scope)=(test) is not present in table "scopes".\n']`

PR #4050, opened 2020-10-02
Uploadclient: upload without socpe - #4039

TomasJavurek:
Following suggestion of Martin of having possibility of upload without any scope assigned to an account.

### DIFF
--- a/lib/rucio/client/uploadclient.py
+++ b/lib/rucio/client/uploadclient.py
@@ -352,7 +352,7 @@ class UploadClient:
         try:
             account_scopes = self.client.list_scopes_for_account(self.client.account)
         except ScopeNotFound:
-            logger.warning('No socpe belonging to the account {}. Processing upload with given scope or no scope at all.'.format(self.client.account))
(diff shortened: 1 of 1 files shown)


lib/rucio/client/uploadclient.py
```diff
@@ -347,8 +348,12 @@ def _register_file(self, file, registered_dataset_dids):
         logger.debug('Registering file')
 
         # verification whether the scope exists
-        account_scopes = self.client.list_scopes_for_account(self.client.account)
-        if file['did_scope'] not in account_scopes:
+        account_scopes = []
+        try:
+            account_scopes = self.client.list_scopes_for_account(self.client.account)
+        except ScopeNotFound:
+            logger.warning('No socpe belonging to the account {}. Processing upload with given scope or no scope at all.'.format(self.client.account))
```
bari12: In the `except` part we don't need the separate message, so a `pass` here is sufficient.

bari12 (approved):

PR #4079, opened 2020-10-22
Transfers: Add conveyor-preparer daemon; #4056

### LINKED ISSUE #4056: Re-Write of Conveyor-Throttler and the throtteling system
Motivation
----------
Due to scalability issue this entire workflow needs to be re-written.

bziemons:
- Add preparer daemon
- Searches for minimum distance source and sets QUEUED or WAITING state based on throttler mode and limits.
- Add PREPARING request state.
- Add conveyor.use_preparer config option.
- Add tests for preparer.


lib/rucio/db/sqla/constants.py
```diff
@@ -1,19 +1,27 @@
...
 #
 # Authors:
-# - Vincent Garonne, <vincent.garonne@cern.ch>, 2013-2017
-# - Mario Lassnig, <mario.lassnig@cern.ch>, 2014, 2017
-# - Martin Barisits, <martin.barisits@cern.ch>, 2014-2019
-# - Cedric Serfon, <cedric.serfon@cern.ch>, 2015-2018
-# - Wen Guan, <wen.guan>, 2016
+# - Vincent Garonne <vincent.garonne@cern.ch>, 2015-2017
+# - Wen Guan <wen.guan@cern.ch>, 2015-2016
+# - Cedric Serfon <cedric.serfon@cern.ch>, 2016
```
cserf: Why do you change the dates ? Is it based on git blame ?

lib/rucio/core/transfer.py
```diff
@@ -1260,8 +1258,8 @@ def get_transfer_requests_and_source_replicas(total_workers=0, worker_number=0,
 
 
 @read_session
-def __list_transfer_requests_and_source_replicas(total_workers=0, worker_number=0,
-                                                 limit=None, activity=None, older_than=None, rses=None, session=None):
+def __list_transfer_requests_and_source_replicas(total_workers=0, worker_number=0, limit=None, activity=None,
+                                                 older_than=None, rses=None, request_state=None, session=None):
```
cserf: This method is called by `get_transfer_requests_and_source_replicas`, but you changed the signature by introducing a request_state. The call should be changed in `get_transfer_requests_and_source_replicas`

lib/rucio/daemons/conveyor/preparer.py
```diff
@@ -0,0 +1,224 @@
...
+                shortest_item = req_sources[idx]
+            elif req_sources[idx][distance_col] < shortest_item[distance_col]:
+                shortest_item = req_sources[idx]
+        yield shortest_item
+
+    count = _update_request_chunk(minimum_distance_request(), session=session)
+    return f'updated {count}/{limit} requests'
+
+
+@transactional_session
```
cserf: Shouldn't it be moved to core ?

lib/rucio/daemons/conveyor/preparer.py
```diff
@@ -0,0 +1,224 @@
...
+        worker_number=worker_number,
+        limit=limit,
+        request_state=RequestState.PREPARING,
+        session=session
+    )
+    if not req_sources:
+        return 'had nothing to do'
+
+    # sort by Request.id, should be pretty quick since the database should return it this way (tim sort)
+    req_sources = sorted(req_sources, key=lambda t: t[request_id_col])
```
cserf: What type is `req_sources` ?

lib/rucio/core/transfer.py
```diff
@@ -1273,6 +1271,10 @@ def __list_transfer_requests_and_source_replicas(total_workers=0, worker_number=
     :param session:          Database session to use.
     :returns:                List.
     """
+
+    if request_state is None:
+        request_state = RequestState.QUEUED
+
```
cserf, line 1288: Ah, sorry, forgot my comment above :D

cserf (commented):
A first few comments

lib/rucio/daemons/conveyor/preparer.py
bziemons: Good question. It is preparer-specific though - I feel like it would be misplaced in core.

lib/rucio/db/sqla/constants.py
bziemons: its based on tools/add_header and therefore the git history - is there a problem with these lines? The history isn't always correct, especially when files were moved around

lib/rucio/core/transfer.py
bziemons: will do this anyway, although it works fine with the default

lib/rucio/daemons/conveyor/preparer.py
bziemons: list

lib/rucio/core/transfer.py
bziemons: just saw this `request_state=RequestState.QUEUED` in the `get_transfer_requests_and_source_replicas` call …? Nothing to do here.

lib/rucio/daemons/conveyor/preparer.py
bziemons: Replaced `sorted` with `list.sort`.

lib/rucio/daemons/conveyor/preparer.py
bari12: It's debatable, but in general we tried to put everything with database interactions to the core, even if only used by the one daemon. (It's the same for the judges etc.)
I would move it.

lib/rucio/db/sqla/constants.py
bari12: The constants file was probably (content) copied once when we move to the sqla folder. Sometimes the dates in these files are also wrongly inserted by people. I think it is fine to just rely on the add_header one.

lib/rucio/daemons/conveyor/preparer.py
bziemons: will do

bari12 (approved):

PR #4081, opened 2020-10-22
Testing: Disable most py2 tests

bziemons:
Disable most py2 tests and make it possible to run tests without httpd in the autotesting framework.

The work in this PR got a bit out of hand and therefor this wandered a bit to a "move to Python 3".

### DIFF
--- a/.github/workflows/autotest.yml
+++ b/.github/workflows/autotest.yml
@@ -5,20 +5,6 @@ on:
   - push
 
 jobs:
-  pr_setup:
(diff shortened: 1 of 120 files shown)


bziemons:
As a reference, this is how the CI looks like when Python 2 core is broken in a non-feature branch:

image

I also made the opposite test - all green when Python 2 is not broken in a non-feature branch :)

bari12 (approved):

Issue #1797, opened 2018-11-19
Support checking of tape staging status


ericvaandering:
Motivation
----------

Concern by CMS that data which is only on the tape buffer may not be migrated to tape before additional disk copies are deleted which could result in the loss of data.

Modification
------------

Undetermined, but as I understand it

- [ ] Add a new, optional, state "Waiting for tape check"
- [ ] FTS transfers to tape systems are put in this state when the transfer finishes
- [ ] A new daemon checks the state with the storage system, moves state to the one which can be picked up by the finisher

bari12:
FTS is thinking about adding this to FTS; On hold until confirmation

bari12:
Just updating tickets. FTS is implementing this at the moment. We will have to add a parameter to these jobs.

PR #4088, opened 2020-10-26
Transfers: Add archive_timeout parameter #1797

ericvaandering:
This has been tested against CTA and an SRM site and the correct parameter is passed to FTS and the behavior is as expected. Operation is unchanged if "archive_timeout" is not set as an RSE attribute.

Should go to 1.23 series and 1.24


bari12:
Looks good to me. I am merging this for 1.23.9 (today)

bari12 (approved):

PR #4089, opened 2020-10-27
Clients: Tighten down permission check for x509 even more; Fix #4071

### LINKED ISSUE #4071: Unhelpful error message for certificates with incorrect permissions
Motivation
----------
Using auth_type=x509, authentication fails when the personal certificate and private key have not the permission 400. Mines had 600. And I got the following error message.

```
$ rucio whoami
Internal error: Request for authentication token returned no result!
2020-10-13 08:58:00,389    ERROR    Cannot authenticate.
Details: x509 authentication failed for account=pchanial with identity={'client_cert': '/opt/rucio/etc/client.crt', 'client_key': '/opt/rucio/etc/client.key'}

```

Modification
------------
Certificate permissions should be checked and the user should be told what to do to fix it if they are not 400, to prevent wasting time on checking the validity of his/her certificates.

mlassnig:


bari12 (approved):

davidgcameron:
Hi,

Using version 1.26.2 I came up against this error:

```
$ rucio whoami
2021-08-18 09:53:12,802	ERROR	Cannot authenticate.
Details: X.509 authentication selected, but private key (/tmp/x509up_u1000) permissions are liberal (required: 400, found: 600)
```
The standard tools for creating proxies like arcproxy and voms-proxy-init give the proxy 600 permissions so this check is too strict and I think should be reverted.

davidgcameron:
I see the problem. I copied the atlas template config which has

```
client_cert = $X509_USER_PROXY
client_key = $X509_USER_PROXY
client_x509_proxy = $X509_USER_PROXY
```
The error message is complaining about `client_key`. When I remove that things work again. I'll make a separate issue to request removing this from the template since it is unnecessary.

PR #4098, opened 2020-10-30
Transfers: Fix rare protocol mismatch bug for transfer sources; Fix #…

bari12:


bari12:
Two things I noticed during creating this PR:
- We need to work a bit on the source resolve procedure, right now it does an early selection and sticks to it; This is not most optimal
- The caching in `get_hops` does not cache if there is NoProtocol match or NoDistance. This can be cached as well.
These will be addressed in a later PR.

PR #4114, opened 2020-11-09
Making tools execution dir independent Fix #4113

### LINKED ISSUE #4113: Make tool execution-dir independent
# Motivation
All the python tool scripts are not directory independent and need to be run as `tools/<script_name>.py`, while they fail without clear explanation if run as `./<script_name>.py` while inside the `tools` directory.

This behaviour causes (and caused) some misconceptions and multiple support requests.

# Modification
Adding on top of the scripts something in the lines of:

```python
import os.path, sys
base_path = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
print(base_path)
sys.path.append(base_path)
os.chdir(base_path)
```

should make them much more portable and easy to use.

gabrielefronze:
Implementing what is described in the related issue.

### DIFF
--- a/tools/bootstrap.py
+++ b/tools/bootstrap.py
@@ -22,8 +22,7 @@ base_path = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 sys.path.append(base_path)
 os.chdir(base_path)
 
-from rucio.db.sqla.util import (build_database, create_base_vo,
-                                create_root_account)
+from rucio.db.sqla.util import (build_database, create_base_vo, create_root_account)  # noqa: E402
 
 if __name__ == '__main__':
     build_database()
--- a/tools/bootstrap.py
+++ b/tools/bootstrap.py
@@ -8,12 +8,18 @@
 # Authors:
 # - Mario Lassnig, <mario.lassnig@cern.ch>, 2012
 # - Vincent Garonne,  <vincent.garonne@cern.ch>, 2013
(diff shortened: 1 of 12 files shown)


bziemons:
There must be an easier way to satisfy flake8 than writing `# noqa: E402` after each import

gabrielefronze:
> There must be an easier way to satisfy flake8 than writing `# noqa: E402` after each import

@bziemons any suggestion? 
I think this shouldn't be a merge stopper since that modification really eases putting in place a test setup for newcomers, but if you have a immediate solution to avoid that specific comment please let me know.

bziemons:
> any suggestion?

Yes, writing `# flake8: noqa: E402` in a line before the block of imports instead of after every import. https://flake8.pycqa.org/en/3.1.1/user/ignoring-errors.html#ignoring-entire-files

> *I think this shouldn't be a merge stopper* since that modification really eases putting in place a test setup for newcomers, but if *you* have a immediate solution to avoid that specific comment please let me know.

I do not decide whether this should be merged or not.

gabrielefronze:
@bziemons ,

thank you for the suggestion!
I am more in favour of a line-by-line instruction. Having a whole file setup would allow the same violation even when not desired or understood by the developer.
Let's wait or @bari12 and @mlassnig opinion to decide which approach to adopt!

mlassnig:
to be honest, i don't care much. this is the tools/ directory and not the core. do as you see fit :-)

bari12:
Fine by me too. 👍

PR #4128, opened 2020-11-12
Allow protocol swap - discussion only

ericvaandering:
Please don't merge this, I want to check this. The CMS mechanism to generate a PFN allows different protocols for different paths. For instance, Fermilab has one portion which has these protocols for two paths:
```
{u'cms:/store/user': u'srm://cmsdcadisk.fnal.gov:8443/srm/managerv2?SFN=/dcache/uscmsdisk/store/user'}
{u'cms:/store/temp/user': u'gsiftp://cmseos-gridftp.fnal.gov/eos/uscms/store/temp/user'}
```

where it wants to switch protocol itself to gsiftp. 

To hack this in, I changed gfal.py such that if it gets something that looks like a URI already instead of a path from the LFN to PFN algorithm, it just passes it along. What do you think of this @bari12 ? Can we turn this into a real pull request?

I don't think we need SRM to do this, but perhaps.

You can see this is rough, I still have lots of debug messages in it and so forth.

Pull request title
------------------

### DIFF
--- a/lib/rucio/rse/protocols/gfal.py
+++ b/lib/rucio/rse/protocols/gfal.py
@@ -97,14 +97,25 @@ class Default(protocol.RSEProtocol):
                 path = lfn['path'] if 'path' in lfn and lfn['path'] else self._get_path(scope=scope, name=name)
                 if self.attributes['scheme'] != 'root' and path.startswith('/'):  # do not modify path if it is root
                     path = path[1:]
-                pfns['%s:%s' % (scope, name)] = ''.join([self.attributes['scheme'], '://', hostname, web_service_path, prefix, path])
+                print('With port %s' % path)
+                if re.match('\w+://', path):
+                    pfns['%s:%s' % (scope, name)] = path  # Algorithm gave us a full URI. Use it
+                else:
+                    pfns['%s:%s' % (scope, name)] = ''.join([self.attributes['scheme'], '://', hostname,
+                                                             web_service_path, prefix, path])
         else:
             for lfn in lfns:
                 scope, name = str(lfn['scope']), lfn['name']
                 path = lfn['path'] if 'path' in lfn and lfn['path'] else self._get_path(scope=scope, name=name)
                 if self.attributes['scheme'] != 'root' and path.startswith('/'):  # do not modify path if it is root
                     path = path[1:]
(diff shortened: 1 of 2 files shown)


bari12:
Hi Eric,

Can you maybe explain a bit how you would be using this? Is it mostly for list-replicas? Or also for downloads or even third-party-copy.

I think for list-replicas and downloads this would probably work like it is (Would need to check corner cases), for third-party-copy though it kills the logic, as you couldn't properly do the protocol matching anymore. So that's partly dangerous as it could really lead to some unexpected behaviour.

I wonder if it would conceptually be better to abstract this in a `multi-protocol` scheme/implementation (can be based on the gfal class), to make it clear already from the outside (E.g. when viewing protocol priorities) what is going to happen. But it really depends a bit how you would like to use such a functionality.

Cheers,
Martin

ericvaandering:
What we need this for at the moment is really silly. We need a service to do LFN to PFN conversions and Phedex used to do it. The PFN generated in this particular case will never be managed by Rucio, but Rucio has the information to tell the user where to write/find it.

In this particular case, TPC would work, right, since gfal handles SRM and gsiftp?

bari12:
Sorry Eric, I missed your comment.
In this particular case it would work (although still might break protocol priorities), but as soon as it breaks the protocol compatibility map it would break completely.

bari12:
Hi @ericvaandering any followup on this? just wondering what to do with the PR :-)

ericvaandering:
Let’s close it.

Sent from a mobile device. Please excuse my brevity or transcription errors.

On Apr 6, 2021, at 7:29 AM, Martin Barisits ***@***.***> wrote:

﻿

Hi @ericvaandering<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_ericvaandering&d=DwMCaQ&c=gRgGjJ3BkIsb5y6s49QqsA&r=EHaoB-POFWGrYFvPXoj1bQ&m=MXbpifQtSGfiguqblXwfWgnRgDiLYjCLLwsPZUNMu9I&s=hadaJ1recDI_GFr-Wb3Ei-AylMq4wcGJJ55eqqznyEQ&e=> any followup on this? just wondering what to do with the PR :-)

Issue #4129, opened 2020-11-13
Rucio upload with rse_expression


bari12:
Motivation
----------
For certain workflows it would be useful to do an upload to an rse_expression, where the rucio clients then select the RSE to upload to. E.g. `country=de` and a (random) RSE in germany is selected. This should be addable somewhere here:
https://github.com/rucio/rucio/blob/9f7a71b03dcf90c3b4c28123de51cb8a60f41000/lib/rucio/client/uploadclient.py#L134

I checked a little bit for any corner-cases here, but I think as long as the RSE is selected at the beginning and further on the selected RSE is used, this workflow should be fine.

TomasJavurek:
A separate option for this in the item list? e.g. dst_rse_expression? Or leaving the rse option to be optionally an expression?
This might depend on the way how items in the upload workflow are created (outside of the uploadclient). In case of cli: https://github.com/rucio/rucio/blob/master/bin/rucio#L946, both options should be fine.

PR #4131, opened 2020-11-13
uploadclient: adding rse_expression option - #4129

TomasJavurek:
I inspired by the downloadclient and left the rse option to be either rse name or rse expression. I decided for using the rseclient method list_rses(), in place of core methods. The destination is selected randomly from the list provided by the mentioned method.

### DIFF
--- a/lib/rucio/client/uploadclient.py
+++ b/lib/rucio/client/uploadclient.py
@@ -84,6 +84,7 @@ class UploadClient:
             logger.debug('Tracing is turned off.')
         self.default_file_scope = 'user.' + self.client.account
         self.rses = {}
+        self.rse_expressions = {}
 
         self.trace = {}
         self.trace['hostname'] = socket.getfqdn()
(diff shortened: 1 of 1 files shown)


lib/rucio/client/uploadclient.py
```diff
@@ -129,15 +130,22 @@ def upload(self, items, summary_file_path=None, traces_copy_out=None):
...
         registered_dataset_dids = set()
         registered_file_dids = set()
+        rses = []
+        rse_expression = None
         for file in files:
-            rse = file['rse']
+            rse_expression = file['rse']
+            if not rses:
+                rses = [r['rse'] for r in self.client.list_rses(rse_expression)]  # can raise InvalidRSEExpression
+                random.shuffle(rses)
```
mlassnig: this is weird, but i guess that's what it's supposed to do. @bari12 opinion?

bari12: I don't think this entirely works. The `files` list might have files where the rse is already set. In this case the algorithm overwrites it. I think it needs to be a bit smarter than that. 
But it also shouldn't re-evaluate the rse_expression (if it is the same) over and over again. So a bit more logic is needed here.

bari12 (approved):

PR #4136, opened 2020-11-17
Validate against r_dids instead of dids

ericvaandering:
I didn't open an issue for this bug. This was introduced in code by @bziemons and it causes problems for CMS. R_DIDS is the right thing to validate against, not DIDS since the DIDS already exist. 

The reason this is an issue for CMS is that on DIDs we require a type to be provided during the validation because we have different RegEx for files, datasets, and containers. For RDIDs we don't care. 

As long as there is no argument that r_dids is the right thing to validate against, please get this into 1.23.11

### DIFF
--- a/lib/rucio/api/replica.py
+++ b/lib/rucio/api/replica.py
@@ -23,6 +23,7 @@
 # - Luc Goossens <luc.goossens@cern.ch>, 2020
 # - Patrick Austin, <patrick.austin@stfc.ac.uk>, 2020
 # - Benedikt Ziemons <benedikt.ziemons@cern.ch>, 2020
+# - Eric Vaandering <ewv@fnal.gov>, 2020
 #
 # PY3K COMPATIBLE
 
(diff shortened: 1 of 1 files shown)


bziemons:
I wouldn't know the difference. @bari12 the 1.23 release does not have the vo parameter for `validate_schema` so mind that when merging this commit.

bziemons (dismissed):

ericvaandering:
A prefix of R_ is always for "read", validating things already there rather than creating new.

ericvaandering:
I don't understand this failure as it is a one line change branched off from the current master. Where is the conflict from?

bziemons:
> I don't understand this failure as it is a one line change branched off from the current master. Where is the conflict from?

I already mentioned it briefly above. `master` and `release-1.23-LTS` diverged in the changed line. I'll link the two lines so it is easier to see.

master:
https://github.com/rucio/rucio/blob/f711667ee4a4ea8c898098f8380c311460dc2642/lib/rucio/api/replica.py#L324

release-1.23-LTS:
https://github.com/rucio/rucio/blob/a33dbde1ed15f4322fca3953e23cb1f0319260cd/lib/rucio/api/replica.py#L324

For git this is not automatically resolvable, since git only looks at complete lines of files for automatic resolving of conflicts.

bari12 (approved):

PR #4139, opened 2020-11-19
Core & Internals: Add support to inequality filters in metadata queries Fix #3750

### LINKED ISSUE #3750: Support DID metadata queries with inequality operators:
Motivation
----------
This is a request for support for did metadata queries using inequalities like:
```
rucio list-dids-by-meta --scope blah --key>some_value
```
where `key` is something like a `gps-start-time` with a numerical value (e.g. 1276523729) associated with the DID.

The particular use case here is for LIGO/Virgo where our rucio-managed files contain time-series data and are associated with some time segment.  

Modification
------------
Introduce inequalities and integer/float values to the metadata query in [did.py](https://github.com/rucio/rucio/blob/master/lib/rucio/core/did.py#L1477)

gabrielefronze:
PR about the integration of the Rucio Inequality Engine (https://github.com/gabrielefronze/rucio-inequality-engine) within rucio/rucio.

### DIFF
--- a/bin/rucio
+++ b/bin/rucio
@@ -43,6 +43,7 @@
 # - Patrick Austin <patrick.austin@stfc.ac.uk>, 2020
 # - Radu Carpa <radu.carpa@cern.ch>, 2021
 # - Rahul Chauhan <omrahulchauhan@gmail.com>, 2021
+# - Gabriele Gaetano Fronze' <gabriele.fronze@to.infn.it>, 2020-2021
 
 from __future__ import print_function
 
@@ -535,26 +536,18 @@ def detach(args):
 @exception_handler
 def list_dids(args):
     """
-    %(prog)s list-dids scope[:*|:name] [--filter 'key=value' | --recursive]
+    %(prog)s list-dids scope[:*|:name] [--filter 'value' | --recursive]
 
     List the data identifiers for a given scope.
     """
     client = get_client(args)
-    filters = {}
+    filters = args.filter
+
+    if not filters:
+        filters = ''
+
     type = 'collection'
     table = []
-    if args.filter:
-        try:
-            filters, type = parse_did_filter_from_string(args.filter)
-        except InvalidType as error:
-            logger.error(error)
-            return FAILURE
-        except ValueError as error:
-            logger.error(error)
-            return FAILURE
-        except Exception:
-            logger.error("Invalid Filter. Filter must be 'key=value', 'key>=value', 'key>value', 'key<=value', 'key<value'")
-            return FAILURE
 
     try:
         scope, name = get_scope(args.did[0], client)
@@ -570,10 +563,17 @@ def list_dids(args):
     if args.recursive and '*' in name:
         logger.error('Option recursive cannot be used with wildcards')
         return FAILURE
-    elif ('name' in filters) and (name != '*'):
-        logger.error('You cannot use a wildcard query and a filter by name')
-        return FAILURE
-    filters['name'] = name
+    else:
+        if filters:
+            if ('name' in filters) and (name != '*'):
+                logger.error('You cannot use a wildcard query and a filter by name')
+                return FAILURE
+
+    if name:
+        if ';' in filters:
+            filters.replace(';', ', name == {};'.format(name))
+        else:
+            filters += ', name == {}'.format(name)
 
     for did in client.list_dids(scope, filters=filters, type=type, long=True, recursive=args.recursive):
         table.append(['%s:%s' % (did['scope'], did['name']), did['did_type']])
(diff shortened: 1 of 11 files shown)


bin/rucio
```diff
@@ -580,34 +585,73 @@ def list_dids(args):
...
+        if scope not in client.list_scopes():
+            logger.error('Scope not found')
+            return FAILURE
+
+        if args.recursive and '*' in name:
+            logger.error('Option recursive cannot be used with wildcards')
+            return FAILURE
+        elif ('name' in filters) and (name != '*'):
+            logger.error('You cannot use a wildcard query and a filter by name')
+            return FAILURE
```
bziemons: use functions, avoid duplicate code

lib/rucio/api/did.py
```diff
@@ -60,20 +61,44 @@ def list_dids(scope, filters, type='collection', ignore_case=False, limit=None,
     :param recursive: Recursively list DIDs content.
     :param vo: The VO to act on.
     """
```
bziemons, line 63: missing docstring for filterstr

bziemons (commented):
one general question before I continue reviewing: what can you do with `--filter` that you cannot do with the new `--filterstr`? From looking at the issue and reading all comments I cannot wrap my head around this.

gabrielefronze:
> one general question before I continue reviewing: what can you do with `--filter` that you cannot do with the new `--filterstr`? From looking at the issue and reading all comments I cannot wrap my head around this.

Hi @bziemons ,

The idea behind this PR is to pave the way for much more versatile filters allowing an experiment to define additional SQL columns or other generic metadata catalogs and be able to query the Rucio file catalog using the standard interface.

I guess a much deeper explanation of what the `inequality_engine` allows one to perform can be found in this repo and its readme: https://github.com/gabrielefronze/rucio-inequality-engine

Basically, at the moment, without the addition of additional columns, `filterstr` has few advantages over `filters` even though it allows to evaluate inequalities involving `length` without the use of the additional columns `'length.gt', 'length.lt', 'length.lte', 'length.gte'`.
Right now the only real advantage is that, since it creates one query for each `OR group` (see linked repo), it is able to get in one query file satisfying multiple filter sets at once.

Ideally in the future one might set metadata values such as the colliding stars masses, the number of Weak bosons or anything relevant and use this `filterstr` parameter to obtain the relevant data with a single (complex?) rucio catalog query.

Nevertheless this addition enables a "variable swept" configuration of the `FUSE posix` module.
Right now it represents the Rucio catalog in a posix-like way using a `server/scope/container-or-dataset/file` branching.
Thanks to this addition it will be able to allow a user defined branching, e.g. `server/cms_energy/n_part/n_muons` or even `server/is_star_a_blackhole/mass_star_a/mass_star_b/date`. This would enable a complete new way of representing datasets.

Let me know if that's what you'd like to know!

bziemons:
If I understand you correctly, the new parameter can do everything the old parameter can do. Why do we need two parameters then?

gabrielefronze:
That's right. I discussed with @bari12 about this and we opted to introduce that gradually without removing the legacy parameter.
If you prefer to have only one parameter that's perfectly ok with me.
If I'm not mistaken (I'll test that, just in case) the syntax does not even change either, the `inequality_engine` just allows ternary inequalities and `;` as OR-groups separator, so it's basically a superset of `filters`.

gabrielefronze:
@bziemons I just realised I interpreted your first question/comment the other way around... These two parameters are definitely creating some confusion 😂

bziemons:
I agree that the similarity is confusing. If we'd keep the old parameter as is, the new one needs to have a name that is better in identifying it. The easiest to me would be something along `--filter-v2`. I would still like to hear about the motivation to keep both parameters as this adds complexity and doesn't seem necessary to me.

gabrielefronze:
Yeah, both suggestions are fine to me. Let's hear from Martin and move on! Thanks!

lib/rucio/tests/test_inequality_engine.py
```diff
@@ -0,0 +1,307 @@
...
+    def test_FlipIfNeeded_NotNeeded(self):
+        string = "length < 87"
+        splitted = condition_split(string)
+        splitted = flip_if_needed(splitted)
+        control = ["length", "<", "87"]
+        for i, w in enumerate(splitted):
+            self.assertEqual(w, control[i])
+
+
+class TestInequalityEngineOffline(unittest.TestCase):
```
mlassnig, line 175: o.O

gabrielefronze: This was needed because I thought I had to use the generation of query filters using a tuple formed as `key, op, value`.
It should not be needed anymore.

mlassnig: :laughing:

gabrielefronze: 🤪🤪🤪

bari12:
Just to jump in to the `filter` vs. `filterstr` discussion. I think we can just replace `--filter` and make sure that it is backwards compatible as long as we are careful with the backwards compatibility here.

gabrielefronze:
@bari12 are the tests enough to verify backward compatibility?
In case they aren't, can you provide a list of legacy `filter` values which are supposed to be supported?
Thanks

bari12:
I think the only ones which work are basically the ones which down the code get translated to 
`created_before', 'created_after', 'length.gt', 'length.lt', 'length.lte', 'length.gte', 'length'`
There is this
https://github.com/rucio/rucio/blob/3698c52044d96515a7c55ac1c5564d18c09610d6/lib/rucio/tests/test_utils.py#L67
which is essentially what is supported.

gabrielefronze:
Hey there,
I finally made through switching from the default filtering algorithm to the new one.
There's one test failing (`test_InequalityEngineRetrocompatibilityLength`) which needs a manual call to sqlalchemy to manually set the datasets length in order to verify retrocompatibility. This is understood and will be fixed soon.
Cheers,

bari12:
Closing, followed up in @robbarnsley PR

PR #4147, opened 2020-11-23
Consider lower-priority protocols in add_replicas - Fix #3940

### LINKED ISSUE #3940: add_replicas with PFN does not consider lower-priority protocols
Motivation
----------

If the primary protocol is not available via WAN, it immediately jumps to first priority LAN, and fails afterwards if not possible.
The secondary and later protocols are never tried.

gabrielefronze:
Uses `rsemgr` routines to iterate over all available `wan` protocols before using `lan` ones. Standard behaviour is retained.

!!! Must be tested

### DIFF
--- a/lib/rucio/rse/rsemanager.py
+++ b/lib/rucio/rse/rsemanager.py
@@ -194,8 +194,8 @@ def create_protocol(rse_settings, operation, scheme=None, domain='wan', auth_tok
         protocol_attr = select_protocol(rse_settings, operation, scheme, domain)
     else:
         candidates = _get_possible_protocols(rse_settings, operation, scheme, domain)
-        if not protocol_attr in candidates:
-             raise exception.RSEProtocolNotSupported('Protocol %s operation %s on domain %s not supported' % (protocol_attr, operation, domain))
+        if protocol_attr not in candidates:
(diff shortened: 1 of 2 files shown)


mlassnig:
This looks ok, can we un-draft this to run the test suite again?

gabrielefronze:
Ciao @mlassnig ,

Yes, sure!

mlassnig (dismissed):

lib/rucio/rse/rsemanager.py
```diff
@@ -169,7 +169,7 @@ def select_protocol(rse_settings, operation, scheme=None, domain='wan'):
     return min(candidates, key=lambda k: k['domains'][domain][operation])
 
 
-def create_protocol(rse_settings, operation, scheme=None, domain='wan', auth_token=None, logger=logging.log):
+def create_protocol(rse_settings, operation, scheme=None, domain='wan', auth_token=None, logger=logging.log, protocol_attr=None):
```
bari12: `logger` should be the last one, add `protocol_attr` before. Also describe it below in the doc.

bari12 (changes requested):

lib/rucio/rse/rsemanager.py
gabrielefronze: It's done!

bari12 (approved):

mlassnig (approved):

PR #4166, opened 2020-11-27
Testing: Use pytest-xdist and -n auto; Fix #4134

### LINKED ISSUE #4134: Parallelise unit tests
Motivation
----------

Install the xdist plugin for pytest with `pip3 install pytest-xdist`, then start pytest with `pytest -n auto`.

Currently this fails because there's a handful of tests that are not threadsafe, these need to be fixed before enabling this.

### LINKED ISSUE #4400: ValueError: hour must be in 0..23
Motivation
----------
Will close this issue myself. Opened it for testing purposes. 
Actually saw the same integration test failure for different commits. 
Trying to understand the problem. 

______ TestRequestCoreRelease.test_release_waiting_requests_grouped_fifo _______
lib/rucio/tests/test_throttler.py:1809: in test_release_waiting_requests_grouped_fifo
    'requested_at': datetime.now().replace(hour=current_hour - 2),
E   ValueError: hour must be in 0..23

Modification
------------
Testing builds with the addition of a single comment.

bziemons:
Testing: Use pytest-xdist and -n auto; Fix #4134

### DIFF
--- a/lib/rucio/tests/test_abacus_collection_replica.py
+++ b/lib/rucio/tests/test_abacus_collection_replica.py
@@ -18,12 +18,14 @@
 # - Andrew Lister <andrew.lister@stfc.ac.uk>, 2019
 # - Patrick Austin <patrick.austin@stfc.ac.uk>, 2020
 # - Eli Chadwick <eli.chadwick@stfc.ac.uk>, 2020
-# - Benedikt Ziemons <benedikt.ziemons@cern.ch>, 2020
+# - Benedikt Ziemons <benedikt.ziemons@cern.ch>, 2020-2021
 # - Martin Barisits <martin.barisits@cern.ch>, 2020-2021
 # - Mario Lassnig <mario.lassnig@cern.ch>, 2020
 
 import os
 import unittest
+
+import pytest
 
 from rucio.client.didclient import DIDClient
 from rucio.client.replicaclient import ReplicaClient
@@ -44,6 +46,7 @@ from rucio.db.sqla.constants import DIDType, ReplicaState
 from rucio.tests.common import file_generator, rse_name_generator
 
 
+@pytest.mark.noparallel(reason='uses pre-defined RSE, fails when run in parallel')
 class TestAbacusCollectionReplica(unittest.TestCase):
     account = 'root'
     scope = 'mock'
--- a/lib/rucio/tests/test_replica_sorting.py
+++ b/lib/rucio/tests/test_replica_sorting.py
@@ -183,17 +183,20 @@ def test_sort_geoip_wan(vo, rest_client, auth_token, protocols_setup, content_ty
         assert sorted_hosts == sorted_replica_hosts, 'assert sorting of result as distance suggested'
 
 
-argvalues = [
-    (Mime.METALINK, 0),
-    (Mime.METALINK, 1),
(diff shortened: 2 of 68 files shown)


lib/rucio/tests/test_abacus_collection_replica.py
```diff
@@ -44,6 +46,7 @@
 from rucio.tests.common import file_generator, rse_name_generator
 
 
+@pytest.mark.noparallel(reason='uses pre-defined RSE, fails when run in parallel')
```
mlassnig, line 49: should add a new issue to fix this later on, i think. no predefined entities should be used in any the tests IMHO (also valid for a couple other tests).

lib/rucio/tests/test_replica_sorting.py
```diff
@@ -183,17 +183,20 @@ def fake_get_distance(se1, se2, *args, **kwargs):
...
+        (Mime.METALINK, 1),
+        pytest.param(Mime.JSON_STREAM, 0, marks=pytest.mark.xfail(reason='see https://github.com/rucio/rucio/issues/4105')),
+        pytest.param(Mime.JSON_STREAM, 1, marks=pytest.mark.xfail(reason='see https://github.com/rucio/rucio/issues/4105')),
+    ]
+    rargvalues = map(lambda p: p.values if hasattr(p, 'values') else p, argvalues)
+    ids = [f'mime={repr(mime)}, lan-site={repr(base_rse_info[iid]["site"])}' for mime, iid in rargvalues]
+    return dict(argvalues=argvalues, ids=ids)
+
+
+@pytest.mark.noparallel(reason='fails when run in parallel, uses site=… as rse expression with non-unique sites')
```
mlassnig, line 198: is that a valid character?

mlassnig (approved):
LGTM two tiny comments for later

bziemons:
Some things to note:
- Tests cannot be run in parallel with sqlite as database, since it's erroring about locks rarely (maybe follow up with a general retry on locked database?)
- I tried to minimize the "flakyness" of tests by generously marking tests as noparallel if they use shared resources like the "MOCK", "MOCK2", … RSEs. This currently means that over one third of tests are not run in parallel and many tests can be easily adapted. Should be handled in a follow-up.
- It is possible to add timeouts for tests in a follow-up in the special pytest plugin I added here. I am in favor of this
- Parallel tests are shuffled and ran in random order, since they must be able to do this to support parallel testing :eyes: 
- It was a pain.

lib/rucio/tests/test_replica_sorting.py
bziemons: Yes, hello utf-8

lib/rucio/tests/test_abacus_collection_replica.py
bziemons: just added a comment to this PR regarding this. Yes, there should be follow-ups. And easy fixes like these should be addressed then

bziemons:
btw. first measurements show a 5 minute runtime improvement for each multi-vo test suites. That is one fifth of the suite's runtime. Adding this up, we are saving a lot of GH Actions minutes.

Edit: The tests running on my local machine with 8 cores (and 8 parallel testers) reduces the overall time of `tools/run_tests_docker.sh -a` roughly from 6 to 4 minutes.

bari12 (approved):

PR #4186, opened 2020-12-07
WebUI: Fix misinterpretation of the date picker #4184

### LINKED ISSUE #4184: Suspicious replicas page is misinterpreting the date picker
Motivation
----------

The code expects the return string from the date picker to be in `DD/MM/YYYY` format but it’s actually in `MM/DD/YYYY`.

Modification
------------

Swap the array indices.

dchristidis:


mlassnig (dismissed):
this is .... o.O

bziemons:
seems suspicious to me since date -> str functions in JavaScript are locale-controlled and this might be switched in FR/EN 
sure about this change?

mlassnig:
On my default Firefox (US), it creates this for December 7:

`https://rucio-ui.cern.ch:443/proxy/replicas/suspicious/?rse_expression=tier=1&younger_than=2020-07-12T00:00:00&nattempts=10`

If I change my browser language and locale to DE, it creates `2020-07-12` still.

mlassnig:
This forces the locale for the datepicker, regardless what the browser wants. See here: https://api.jqueryui.com/datepicker/#method-option

bziemons:
it will return the date as `2020-12-24` though, so split by `-` instead of `/`, see https://jqueryui.com/datepicker/#date-formats

Edit: split is not needed anymore instead, which is even better

bari12:
Ping, what's the status here?

dchristidis:
I’ll come back to this after things calm down. :D

bari12:
Another ping 🛎️ 😄

dchristidis:
Replaced with #4556 (patch instead of feature).

PR #4199, opened 2020-12-11
docker: activate some in-memory performance patches for testing; Fix #4198

### LINKED ISSUE #4198: Improve docker image performance
- Switch databases to in-memory
- Remove fsync
- Activate async I/O

mlassnig:


bari12:
What's the status of this? (Since the tests are failing)

mlassnig:
Ah, I will look into this again. This is useful.

bziemons:
I think asking for another gigabyte of ram per job with a database goes against the fair use of GitHub Actions and since it does not seem to work anyway, I would suggest to make docker-compose override configs for this (letting you run `docker-compose -f etc/docker/dev/docker-compose.yml -f etc/docker/dev/docker-compose-database-tmpfs.yml …`) so this can be used where possible.

bari12:
Please update / close :-)

PR #4203, opened 2020-12-16
Clients: Check subcommand in rucio-admin for py3; Fix #4202

### LINKED ISSUE #4202: Python3 client subparser failure
Subparser is broken on Python3. This works on Python2.

```
[root@client bin]# rucio --version
rucio 1.23.8
```
```
[root@client bin]# rucio-admin rse
Strange error: 'Namespace' object has no attribute 'which'
```
```
[root@client bin]# rucio-admin account
Strange error: 'Namespace' object has no attribute 'which'
```
etc.etc.

```
[root@client bin]# head -n1 rucio-admin
#!/usr/bin/python3.6
```

bziemons:
Clients: Check subcommand in rucio-admin for py3; Fix #4202

### DIFF
--- a/bin/rucio-admin
+++ b/bin/rucio-admin
@@ -36,6 +36,7 @@
 # - Jaroslav Guenther <jaroslav.guenther@cern.ch>, 2019-2020
 # - Patrick Austin <patrick.austin@stfc.ac.uk>, 2020
 # - Rob Barnsley <robbarnsley@users.noreply.github.com>, 2020
+# - Benedikt Ziemons <benedikt.ziemons@cern.ch>, 2020
 
 from __future__ import division
 from __future__ import print_function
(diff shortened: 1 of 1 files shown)


bin/rucio-admin
```diff
@@ -2229,6 +2222,6 @@ if __name__ == '__main__':
         if args.verbose:
             print("Completed in %-0.4f sec." % (end_time - start_time))
         sys.exit(result)
-    except Exception as error:
-        logger.error("Strange error: {0}".format(error))
+    except Exception:
+        logger.exception("Uncaught exception")
```
mlassnig, line 2226: Should print the actual exception, no? Otherwise you have no clue what's actually failing?

mlassnig: Or does .exception() do that automatically?

bziemons: exactly, `.exception` prints the exception with stacktrace while before there was no stacktrace and just the description of the exception as you can see in the issue.

mlassnig (approved):

PR #4206, opened 2020-12-17
Replicas: implement sort by distance #4020

### LINKED ISSUE #4020: Distance based sorting of replicas
Motivation
----------
Currently `list_replicas` offers two ways of sorting replicas which happens on the API level: geoip and random, with random being the default. The idea of this development would be to offer a distance based sorting (Based on the distance table) similar to how it is done for third party transfers. At first that sorting method should just be optional, but eventually it should be the default sorting method to bring this in line with how the source sorting is done for third party transfers.

nikmagini:
First implementation of replica sorting by distance. Tested OK in a private rucio deployment.

Implementation details:

1) Takes the client location from the SITE_NAME env. WARNING: if SITE_NAME is undefined or not known to Rucio (e.g. if client is running on the WN of a diskless site), sort_distance will throw InvalidRSEExpression. Need to decide how to define client location in this case.

2) Distances are measured between the RSEs hosting each remote replica and one (arbitrary) disk RSE at SITE_NAME. If there are multiple disk RSEs at SITE_NAME with different distance settings, sorting results may vary...

2) If distance between the RSEs is not defined (i.e. there is no direct link) the replica is skipped, since multi-hop would be needed to download it

### DIFF
--- a/lib/rucio/core/replica_sorter.py
+++ b/lib/rucio/core/replica_sorter.py
@@ -22,6 +22,7 @@
 # - Cedric Serfon <cedric.serfon@cern.ch>, 2019
 # - Martin Barisits <martin.barisits@cern.ch>, 2019
 # - Benedikt Ziemons <benedikt.ziemons@cern.ch>, 2020
+# - Nicolo Magini <nicolo.magini@cern.ch>, 2020
 
 # This product includes GeoLite data created by MaxMind,
 # available from <a href="http://www.maxmind.com">http://www.maxmind.com</a>
@@ -45,6 +46,8 @@ from dogpile.cache.api import NO_VALUE
 from rucio.common import utils
 from rucio.common.config import config_get
 from rucio.common.exception import InvalidRSEExpression
+from rucio.core.distance import get_distances
+from rucio.core.rse import get_rse_id
 from rucio.core.rse_expression_parser import parse_expression
 
 if TYPE_CHECKING:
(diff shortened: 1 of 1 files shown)


mlassnig (approved):

bari12:
Hi @nikmagini 
Sorry for the late reply, it took quite a while to get the `1.25` release out of the window to get back to this.
In my opinion this won't work since it creates a new query for each replica, there is just not enough caching in it. I know that you moved on to new things in the meanwhile. @mlassnig is this something you want to pickup or should we close it for now?

lib/rucio/core/replica_sorter.py
```diff
@@ -223,6 +228,28 @@ def distance(pfn):
...
+    :param client_location: Location dictionary containing {'ip', 'fqdn', 'site'}
+    """
+
+    client_rse_id = parse_expression("site=%s&rse_type=DISK" % client_location['site'])[0]['id']
+
+    replica_distances = {}
+    for pfn in dictreplica:
+        src_rse = dictreplica[pfn][2]
+        try:
+            replica_distances[pfn] = get_distances(src_rse_id=get_rse_id(src_rse),
```
mlassnig, line 244: missing session pass-through

mlassnig:
So, in principle this should work (with the `,session=session,..` added), but I'm also worried about the caching here.

bari12:
Let's close this for now, and add it with caching later. 
There is not too much urgency, given that the geoip sorting works well in production.

PR #4211, opened 2021-01-04
Core: Fix bad file declaration on non-deterministic RSEs #4209

### LINKED ISSUE #4209: Minos does not work for non-deterministic RSEs
Motivation
----------

Commit 5cc96812352f27ae7096bdc4a39b658a2f494d23 fixes comparing the value of `status` against different items of `BadFilesStatus` by converting both to `str`s. However, this was only applied for deterministic RSEs.

Upon closer inspection, it appears that Minos is calling `declare_bad_file_replicas()` with instances of `BadPFNStatus` instead of `BadFilesStatus`.

Modification
------------

~~Apply the same logic for non-deterministic RSEs.~~

Fix the bug in Minos, then remove the conversions done in the core for deterministic RSEs.

dchristidis:
Closes #4209.

### DIFF
--- a/lib/rucio/core/replica.py
+++ b/lib/rucio/core/replica.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Copyright 2013-2020 CERN
+# Copyright 2013-2021 CERN
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


lib/rucio/core/replica.py
```diff
@@ -356,7 +356,7 @@ def __declare_bad_file_replicas(pfns, rse_id, reason, issuer, status=BadFilesSta
         for pfn in parsed_pfn:
             path = '%s%s' % (parsed_pfn[pfn]['path'], parsed_pfn[pfn]['name'])
             __exists, scope, name, already_declared, size = __exists_replicas(rse_id, scope=None, name=None, path=path, session=session)
-            if __exists and ((status == BadFilesStatus.BAD and not already_declared) or status == BadFilesStatus.SUSPICIOUS):
+            if __exists and ((str(status) == str(BadFilesStatus.BAD) and not already_declared) or str(status) == str(BadFilesStatus.SUSPICIOUS)):
```
bari12, line 359: That seems like a bad idea to compare strings here. Why doesn't the constant comparison work?

dchristidis: Minos is passing a `BadPFNStatus` instance. :facepalm:
Thanks for taking the time to review the patch.

PR #4231, opened 2021-01-12
Recovery: Fix bad file declaration on non-deterministic RSEs #4209

### LINKED ISSUE #4209: Minos does not work for non-deterministic RSEs

dchristidis:
Closes #4209.


mlassnig (approved):

bari12:
Merged as patch.

PR #4240, opened 2021-01-14
Intermediary fix for FTSState to return the actual string name; Fix #…

mlassnig:


bziemons (approved):
Are there going to be tests for these errors?

mlassnig:
Mayank is working on a complete third-party-copy test suite.

PR #4244, opened 2021-01-18
Transfers: Adapt submitter and poller daemons for preparer third-party copy #4216

### LINKED ISSUE #4216: Stronger integration of Globus Online transfer tool

maatthias:
Adapt submitter and poller daemons for preparer third-party copy


lib/rucio/db/sqla/migrate_repo/versions/f85a2962b021_adding_transfertool_column_to_requests_.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def upgrade():
+    '''
+    Upgrade the database to this revision
+    '''
+
+    if context.get_context().dialect.name in ['oracle', 'postgresql', 'mysql']:
+        schema = context.get_context().version_table_schema + '.' if context.get_context().version_table_schema else ''
+        add_column('requests', sa.Column('transfer_tool', sa.String(64)), schema=schema)
```
bziemons: please stick with `transfertool`

lib/rucio/db/sqla/migrate_repo/versions/f85a2962b021_adding_transfertool_column_to_requests_.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def downgrade():
+    '''
+    Downgrade the database to the previous revision
+    '''
+
+    if context.get_context().dialect.name in ['oracle', 'postgresql', 'mysql']:
+        schema = context.get_context().version_table_schema + '.' if context.get_context().version_table_schema else ''
+        drop_column('requests', 'transfer_tool', schema=schema)
```
bziemons: see above

lib/rucio/core/request.py
```diff
@@ -317,6 +322,13 @@ def get_next(request_type, state, limit=100, older_than=None, rse_id=None, activ
         elif activity:
             query = query.filter(models.Request.activity == activity)
 
+        # if conveyor configured for preparer third-party copy then filter queued requests by transfertool for conveyor-submitter daemon
+        preparer_enabled = config_get_bool('conveyor', 'use_preparer', raise_exception=False, default=False)
+        TRANSFER_TOOL = config_get('conveyor', 'transfertool', False, None)
```
bziemons: use lower-case names for variables, this is not a module-wide constant

bziemons (changes requested):
@maatthias added a few notes. Also you need to update `lib/rucio/alembicrevision.py` when adding a new alembic revision

lib/rucio/core/request.py
```diff
@@ -317,6 +322,12 @@ def get_next(request_type, state, limit=100, older_than=None, rse_id=None, activ
         elif activity:
             query = query.filter(models.Request.activity == activity)
 
+        # if conveyor configured for preparer third-party copy then filter queued requests by transfertool for conveyor-submitter daemon
+        preparer_enabled = config_get_bool('conveyor', 'use_preparer', raise_exception=False, default=False)
+        transfertool_config = config_get('conveyor', 'transfertool', False, None)
```
bari12: `transfertool` should be a method parameter, not fetched via config.

lib/rucio/core/request.py
```diff
@@ -317,6 +322,12 @@ def get_next(request_type, state, limit=100, older_than=None, rse_id=None, activ
         elif activity:
             query = query.filter(models.Request.activity == activity)
 
+        # if conveyor configured for preparer third-party copy then filter queued requests by transfertool for conveyor-submitter daemon
+        preparer_enabled = config_get_bool('conveyor', 'use_preparer', raise_exception=False, default=False)
```
bari12: Not needed, can be removed

lib/rucio/core/request.py
```diff
@@ -317,6 +322,12 @@ def get_next(request_type, state, limit=100, older_than=None, rse_id=None, activ
...
+        transfertool_config = config_get('conveyor', 'transfertool', False, None)
+        if preparer_enabled:
+            query = query.filter(models.Request.transfertool == transfertool_config)
```
bari12: I think this won't be sufficient because we should force another index to be used.
So in line 309 we need to replace the entire query, add a new index (`REQUESTS_TYP_STA_TRANS_ACT_IDX`) and hint that one to be used. 
So in the alembic/model.py file we also need to create that index.

lib/rucio/core/transfer.py
```diff
@@ -1328,6 +1327,11 @@ def __list_transfer_requests_and_source_replicas(total_workers=0, worker_number=
     if activity:
         sub_requests = sub_requests.filter(models.Request.activity == activity)
 
+    # if conveyor configured for preparer third-party copy then filter queued requests by transfertool for conveyor-submitter daemon
+    preparer_enabled = config_get_bool('conveyor', 'use_preparer', raise_exception=False, default=False)
```
bari12: Same comment as with `get_next`

lib/rucio/db/sqla/models.py
```diff
@@ -1166,6 +1166,7 @@ class Request(BASE, ModelBase, Versioned):
     account = Column(InternalAccountString(get_schema_value('ACCOUNT_LENGTH')))
     requested_at = Column(DateTime)
     priority = Column(Integer)
+    transfertool = Column(String(64))
```
bari12: Need to add an index below as well
`Index('REQUESTS_TYP_STA_TRANS_ACT_IDX', 'request_type', 'state', 'transfertool', 'activity', ),`

bari12 (commented):
Just added a few comments as discussed in our meeting.
Important aswell:
- Heartbeats in the submitter need to include the transfertool, if one is used. https://github.com/rucio/rucio/blob/c15716f9494cbbfb1da70e5a77f2d907a61e01a0/lib/rucio/daemons/conveyor/submitter.py#L126
- `get_transfers` needs to include transfertool https://github.com/rucio/rucio/blob/c15716f9494cbbfb1da70e5a77f2d907a61e01a0/lib/rucio/daemons/conveyor/submitter.py#L168
- which then in turn needs to be handed down to the core https://github.com/rucio/rucio/blob/c15716f9494cbbfb1da70e5a77f2d907a61e01a0/lib/rucio/core/transfer.py#L632
- When you finally arrive in the `list_transfer_requests_and_source_replicas` method, we need to remove sources which do not support the transfertool. So this needs to be filtered at the end of the method https://github.com/rucio/rucio/blob/c15716f9494cbbfb1da70e5a77f2d907a61e01a0/lib/rucio/core/transfer.py#L1290

lib/rucio/db/sqla/models.py
bziemons: just noticed: do we want to keep the file `etc/sql/oracle/schema.sql` recent? Then this also needs to be updated in this PR.

lib/rucio/core/request.py
```diff
@@ -280,6 +285,7 @@ def get_next(request_type, state, limit=100, older_than=None, rse_id=None, activ
     :param hash_variable:     The variable to use to perform the partitioning. By default it uses the request id.
     :param activity_shares:   Activity shares dictionary, with number of requests
     :param session:           Database session to use.
+    :transfertool:            The transfer tool as specified in rucio.cfg.
```
bziemons: needs to start with `:param transfertool:`, see the other parameter descriptions.

lib/rucio/core/transfer.py
```diff
@@ -642,6 +641,7 @@ def get_transfer_requests_and_source_replicas(total_workers=0, worker_number=0,
     :parm retry_other_fts:        Retry other fts servers.
     :param failover_schemes:      Failover schemes.
     :session:                     The database session in use.
+    :transfertool:                The transfer tool as specified in rucio.cfg.
```
bziemons: see above. While you're at it, it would be great if you could fix the other wrong parameter descriptions.

lib/rucio/core/transfer.py
```diff
@@ -1318,6 +1320,7 @@ def __list_transfer_requests_and_source_replicas(total_workers=0, worker_number=
         .with_hint(models.Request, "INDEX(REQUESTS REQUESTS_TYP_STA_UPD_IDX)", 'oracle') \
         .filter(models.Request.state == request_state) \
         .filter(models.Request.request_type == RequestType.TRANSFER) \
+        .filter(models.Request.transfertool == transfertool) \
```
bziemons: I think this line needs a check for `None`, so that you could disable the filter for a specific transfertool.

bziemons (changes requested):
As far as I can tell Martin's following comment was not addressed entirely:

- When you finally arrive in the `list_transfer_requests_and_source_replicas` method, we need to remove sources which do not support the transfertool. So this needs to be filtered at the end of the method https://github.com/rucio/rucio/blob/c15716f9494cbbfb1da70e5a77f2d907a61e01a0/lib/rucio/core/transfer.py#L1290

I would have expected fetching the transfertool RSE attribute for all source RSEs and an additional manual filter at the end of the function `__list_transfer_requests_and_source_replicas` if the transfertool parameter is not None.

lib/rucio/core/request.py
maatthias: done

lib/rucio/core/transfer.py
maatthias: I've removed this and added the filter for source RSEs where the transfertool attribute == the transfertool specified in rucio.cfg

lib/rucio/core/transfer.py
maatthias: done

lib/rucio/core/transfer.py
bziemons: I think we need both filters. Maybe we can chat about this again.

maatthias: Yes looking at it again I think you are correct.  Added the filter as show above back in.

lib/rucio/core/transfer.py
```diff
@@ -652,7 +652,8 @@ def get_transfer_requests_and_source_replicas(total_workers=0, worker_number=0,
                                                                older_than=older_than,
                                                                rses=rses,
                                                                request_state=RequestState.QUEUED,
-                                                               session=session)
+                                                               session=session,
+                                                               transfertool=transfertool)
```
bari12: session should stay last argument

lib/rucio/core/transfer.py
```diff
@@ -641,7 +640,8 @@ def get_transfer_requests_and_source_replicas(total_workers=0, worker_number=0,
     :param bring_online:          Bring online timeout.
     :parm retry_other_fts:        Retry other fts servers.
```
bari12: 🐔

lib/rucio/core/transfer.py
```diff
@@ -1286,7 +1291,7 @@ def get_transfer_requests_and_source_replicas(total_workers=0, worker_number=0,
 
 @read_session
 def __list_transfer_requests_and_source_replicas(total_workers=0, worker_number=0, limit=None, activity=None,
-                                                 older_than=None, rses=None, request_state=None, session=None) -> "List[Tuple]":
+                                                 older_than=None, rses=None, request_state=None, session=None, transfertool=None) -> "List[Tuple]":
```
bari12: session should stay last

lib/rucio/db/sqla/models.py
```diff
@@ -1176,7 +1178,8 @@ class Request(BASE, ModelBase, Versioned):
                    Index('REQUESTS_TYP_STA_UPD_IDX', 'request_type', 'state', 'activity'),
                    Index('REQUESTS_RULEID_IDX', 'rule_id'),
                    Index('REQUESTS_EXTERNALID_UQ', 'external_id'),
-                   Index('REQUESTS_DEST_RSE_ID_IDX', 'dest_rse_id'))
+                   Index('REQUESTS_DEST_RSE_ID_IDX', 'dest_rse_id'),
+                   Index('REQUESTS_TYP_STA_TRANS_ACT_IDX', 'request_type', 'state', 'transfertool'))
```
bari12: Add `activity` as well to the end of index.

bari12: Length of `REQUESTS_TYP_STA_TRANS_ACT_IDX` might be too long.

lib/rucio/core/transfer.py
bari12: 1. We need to make this conditional to `transfertool` as well and use the other Index in case transfertool is used.
2. If transfertool is not used, it should not be part of the query.

lib/rucio/core/transfer.py
```diff
@@ -1370,6 +1377,15 @@ def __list_transfer_requests_and_source_replicas(total_workers=0, worker_number=
...
         .with_hint(models.Distance, "+ index(distances DISTANCES_PK)", 'oracle')
 
+    # if transfertool specified, select only the requests where the source rses are set up for the transfer tool
+    if transfertool:
+        query = query.subquery()
+        subq = session.query(query) \
+            .join(models.RSEAttrAssociation, models.RSEAttrAssociation.rse_id == query.c.rse_id) \
+            .filter(models.RSEAttrAssociation.key == 'transfertool',
+                    models.RSEAttrAssociation.value == transfertool)
+        return subq.all()
```
bari12: Replace subq by query and do not return there, it will be returned later.

lib/rucio/core/transfer.py
```diff
@@ -1370,6 +1377,15 @@ def __list_transfer_requests_and_source_replicas(total_workers=0, worker_number=
                                          models.RSEFileAssociation.rse_id == models.Distance.src_rse_id)) \
         .with_hint(models.Distance, "+ index(distances DISTANCES_PK)", 'oracle')
 
+    # if transfertool specified, select only the requests where the source rses are set up for the transfer tool
+    if transfertool:
+        query = query.subquery()
+        subq = session.query(query) \
+            .join(models.RSEAttrAssociation, models.RSEAttrAssociation.rse_id == query.c.rse_id) \
+            .filter(models.RSEAttrAssociation.key == 'transfertool',
+                    models.RSEAttrAssociation.value == transfertool)
```
bari12: Possibility of having multiple transfertools in the key. Needs to check for transfertool in value.

lib/rucio/daemons/conveyor/finisher.py
```diff
@@ -97,6 +98,7 @@ def finisher(once=False, sleep_time=60, activities=None, bulk=100, db_bulk=1000)
     logging.debug("Suspicious patterns: %s" % [pat.pattern for pat in suspicious_patterns])
 
     retry_protocol_mismatches = conveyor_config.get('retry_protocol_mismatches', False)
+    transfertool = config_get('conveyor', 'transfertool', False, None)
```
bari12: Can be removed.

lib/rucio/daemons/conveyor/finisher.py
```diff
@@ -135,7 +137,8 @@ def finisher(once=False, sleep_time=60, activities=None, bulk=100, db_bulk=1000)
                                              total_workers=heart_beat['nr_threads'],
                                              worker_number=heart_beat['assign_thread'],
                                              mode_all=True,
-                                             hash_variable='rule_id')
+                                             hash_variable='rule_id',
+                                             transfertool=transfertool)
```
bari12: Can be removed

lib/rucio/daemons/conveyor/poller.py
```diff
@@ -133,7 +135,8 @@ def poller(once=False, activities=None, sleep_time=60,
                                                 total_workers=heart_beat['nr_threads'], worker_number=heart_beat['assign_thread'],
```
bari12, line 129: To the logging line above (L130) add transfertool to the logging.

lib/rucio/daemons/conveyor/submitter.py
```diff
@@ -332,6 +335,7 @@ def __get_transfers(total_workers=0, worker_number=0, failover_schemes=None, lim
     :param max_sources:      Max sources.
     :bring_online:           Bring online timeout.
```
bari12: Add `param`

lib/rucio/db/sqla/migrate_repo/versions/7df670114c42_adding_index_for_transfertool_on_.py
```diff
@@ -0,0 +1,43 @@
...
+down_revision = 'f85a2962b021'
+
+
+def upgrade():
+    '''
+    Upgrade the database to this revision
+    '''
+
+    if context.get_context().dialect.name in ['oracle', 'mysql', 'postgresql']:
+        create_index('REQUESTS_TYP_STA_TRANS_ACT_IDX', 'requests', ['request_type', 'state', 'transfertool'])
```
bari12: Add `activity` aswell

lib/rucio/db/sqla/migrate_repo/versions/f85a2962b021_adding_transfertool_column_to_requests_.py
```diff
@@ -0,0 +1,47 @@
+# Copyright 2013-2020 CERN for the benefit of the ATLAS collaboration.
```
bari12, line 1: Merge the two alembic files

lib/rucio/db/sqla/models.py
```diff
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
```
bari12: Please add the changes (new index and new column in requests table) to oracle sql file as well
https://github.com/rucio/rucio/blob/master/etc/sql/oracle/schema.sql

etc/docker/dev/docker-compose.yml
```diff
@@ -26,6 +29,8 @@ services:
       - "5432:5432"
     command: ["-c", "fsync=off","-c", "synchronous_commit=off","-c", "full_page_writes=off"]
   graphite:
+    hostname: graphite
+    container_name: graphite
```
bziemons: please restore this file to the original :)

lib/rucio/core/transfer.py
bziemons: unfortunately, this wasn't resolved

lib/rucio/db/sqla/migrate_repo/versions/f85a2962b021_adding_transfertool_column_to_requests_.py
bziemons: unfortunately this is not resolved

bziemons (changes requested):
Most of the points were addressed. This is almost good to go :)

lib/rucio/db/sqla/migrate_repo/versions/f85a2962b021_adding_transfertool_column_to_requests_.py
maatthias: Ah yes.  I misunderstood this as merging your original alembic files with mine.  Yes we can combine the column and index creation files.

etc/docker/dev/docker-compose.yml
maatthias: :smile:

lib/rucio/core/request.py
```diff
@@ -270,7 +271,7 @@ def temp_serializer(obj):
 @read_session
 def get_next(request_type, state, limit=100, older_than=None, rse_id=None, activity=None,
              total_workers=0, worker_number=0, mode_all=False, hash_variable='id',
-             activity_shares=None, session=None):
+             activity_shares=None, session=None, transfertool=None):
```
bziemons: move session to be the last argument.

lib/rucio/alembicrevision.py
```diff
@@ -16,5 +16,6 @@
 # Authors:
 # - Benedikt Ziemons <benedikt.ziemons@cern.ch>, 2020
 # - Martin Barisits <martin.barisits@cern.ch>, 2020
+# - Matt Snyder <msnyder@bnl.gov>, 2021
 
-ALEMBIC_REVISION = 'd23453595260'  # the current alembic head revision
+ALEMBIC_REVISION = '7df670114c42'  # the current alembic head revision
```
bziemons: This needs to be `f85a2962b021`

bziemons (changes requested):

bziemons (dismissed):

lib/rucio/core/transfer.py
```diff
@@ -719,10 +720,14 @@ def get_transfer_requests_and_source_replicas(total_workers=0, worker_number=0,
         # In case the source_rse and the dest_rse are connected, the list contains only the destination RSE
         # In case of non-connected, the list contains all the intermediary RSEs
         list_hops = []
+        include_multihop = core_config_get('transfers', 'use_multihop', default=False, expiration_time=600, session=session)
+        if transfertool:
```
bari12: Needs to be checked if transfertool is `fts`, only for non-fts transfertools we want to disable this for sure.

lib/rucio/core/transfer.py
```diff
@@ -1331,6 +1337,9 @@ def __list_transfer_requests_and_source_replicas(total_workers=0, worker_number=
     if activity:
         sub_requests = sub_requests.filter(models.Request.activity == activity)
 
+    if transfertool:
+        sub_requests = sub_requests.filter(models.Request.transfertool == transfertool)
```
bari12, line 1342: You need to make the index hint (Line 1321) conditional here too.
E.g. if no transfertool, use the hint from above, if transfertool, use the new index.

lib/rucio/core/transfer.py
```diff
@@ -1373,6 +1382,14 @@ def __list_transfer_requests_and_source_replicas(total_workers=0, worker_number=
                                          models.RSEFileAssociation.rse_id == models.Distance.src_rse_id)) \
         .with_hint(models.Distance, "+ index(distances DISTANCES_PK)", 'oracle')
 
+    # if transfertool specified, select only the requests where the source rses are set up for the transfer tool
+    if transfertool:
+        query = query.subquery()
+        query = session.query(query) \
+            .join(models.RSEAttrAssociation, models.RSEAttrAssociation.rse_id == query.c.rse_id) \
+            .filter(models.RSEAttrAssociation.key == 'transfertool',
+                    models.RSEAttrAssociation.value.like(transfertool))
```
bari12: Shouldnt this be a LIKE against `%transfertool%`? Since an RSE could have the tag `tranfertool1,transfertool2,transfertool3`?

lib/rucio/db/sqla/migrate_repo/versions/f85a2962b021_adding_transfertool_column_to_requests_.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def upgrade():
+    '''
+    Upgrade the database to this revision
+    '''
+
+    if context.get_context().dialect.name in ['oracle', 'postgresql', 'mysql']:
+        schema = context.get_context().version_table_schema if context.get_context().version_table_schema else ''
+        add_column('requests', sa.Column('transfertool', sa.String(64)), schema=schema)
```
bari12, line 37: This column needs to be added (and removed) to the `requests_history` table as well. (Sorry did not tell you about this). In the `models.py` this is done automatically, but you need to add it to the upgrade scripts too. Please also add it to the oracle schema.sql

bari12 (changes requested):
Hi @maatthias 
I was about to merge this, and address the missing changes later, but then I noticed that we are missing changes for the `requests_history` table. (We completely neglected this :-) See comments. We should adapt this before merging it, but it should be rather minor.

In `requests.py:archive_request(...)` we need to add the new `transfertool` column as well - so it get's filled in `requests_history` too.

There is also the `add_monitor_message` call, but here we would need to see how to properly handle that. Let's leave it like it is for now.

lib/rucio/core/transfer.py
maatthias: Since include_multihop defaults to false can't we just remove the transfertool check entirely?

bari12: The problem is, e.g. in the ATLAS deployment it will evaluate to `True`, since we use multihop in fts. But we don't (actually can't) use multhop in GO.

lib/rucio/core/transfer.py
maatthias: You are correct.  I changed the check to '%' + transfertool + '%' so that the attribute can have multiple transfer tools.  I tested and transfers were successful.  However, does this raise risk of injection?  Not sure how/if sqlalchemy sanitizes but just a thought.

bari12: It's not a huge risk here, since the `transfertool` setting comes from the config / command line parameters on the node. With that level of access you can as well read the database credentials from the config.

lib/rucio/core/transfer.py
```diff
@@ -719,10 +720,14 @@ def get_transfer_requests_and_source_replicas(total_workers=0, worker_number=0,
         # In case the source_rse and the dest_rse are connected, the list contains only the destination RSE
         # In case of non-connected, the list contains all the intermediary RSEs
         list_hops = []
+        include_multihop = False
+        if transfertool == 'fts':
```
bari12: We need to cover the default case as well (`ransfertool=None`)
let's change this to `if transfertool in ['fts', None]:`

bari12 (approved):

Issue #528, opened 2018-01-25
Temporarily exclude RSEs with a timeout to not impact general deletion rate


bari12:
Motivation
----------
If a RSE performs badly on deletion, exclude it temporarily to not impact the general deletion rate.

PR #4245, opened 2021-01-19
Deletion: automatically exclude unavailable RSEs #528

tbeerman:

### DIFF
--- a/bin/rucio-reaper2
+++ b/bin/rucio-reaper2
@@ -1,5 +1,5 @@
 #!/usr/bin/env python3
-# Copyright 2014-2019 CERN for the benefit of the ATLAS collaboration.
+# Copyright 2014-2021 CERN for the benefit of the ATLAS collaboration.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -16,6 +16,7 @@
 # Authors:
 # - Cedric Serfon, <cedric.serfon@cern.ch>, 2019
 # - Patrick Austin <patrick.austin@stfc.ac.uk>, 2020
+# - Thomas Beermann <thomas.beermann@cern.ch>, 2021
 #
 # PY3K COMPATIBLE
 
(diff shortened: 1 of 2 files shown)


lib/rucio/daemons/reaper/reaper2.py
```diff
@@ -594,31 +616,33 @@ def stop(signum=None, frame=None):
...
-    :param exclude_rses:       RSE expression to exclude RSEs from the Reaper.
-    :param include_rses:       RSE expression to include RSEs.
-    :param vos:                VOs on which to look for RSEs. Only used in multi-VO mode.
-                               If None, we either use all VOs if run from "def",
-                               or the current VO otherwise.
-    :param delay_seconds:      The delay to query replicas in BEING_DELETED state.
-    :param sleep_time:         Time between two cycles.
+    :param threads:                The total number of workers.
+    :param chunk_size:             The size of chunk for deletion.
+    :param noaccess_threshold:     Number of service unavailable exceptions after which the RSE gets temporarily excluded.
```
bziemons: `noaccess_threshold` doesn't exist as parameter

bziemons (changes requested):

lib/rucio/daemons/reaper/reaper2.py
tbeerman: thanks, I missed that one.

bziemons (dismissed):

bin/rucio-reaper2
```diff
@@ -55,6 +56,10 @@ def get_parser():
                         help='The delay (seconds) to query replicas in BEING_DELETED state.')
     parser.add_argument('--scheme', action="store", default=None, type=str,
                         help='Force the reaper to use a particular protocol/scheme, e.g., mock')
+    parser.add_argument("--auto_exclude_threshold", action="store", default=100, type=int,
```
bari12, line 59: As discussed, we should right away also add a config_get from the config table alternatively, so we can more easily experiment with these values.

bari12 (changes requested):

bin/rucio-reaper2
tbeerman: Which one should be preferred? CLI argument or config table?

bari12 (approved):

PR #4250, opened 2021-01-19
Core & Internals: Check json support before falling back to json

bziemons:
Check json support before falling back to json. <del>This means all unknown meta data will be silently ignored instead of always trying to set them in the JSONDidMeta.</del> An error will be thrown on unknown metadata keys, added check in #4255 
Additionally this adds the (optional) session argument to `DidMetaPlugin.manages_key`.


bziemons:
For reviewers: removed redundant docs from the `manages_key` methods, because the doc is already present in the base class.

bari12 (approved):

PR #4254, opened 2021-01-19
Multi VO: Test for multi-VO before checking access_rule_vo  Fix #4253

### LINKED ISSUE #4253: Rucio single-VO should not check for permission on access_rule_vo
Motivation
----------
Problems in a single-VO instance when checking permissions on various rule commands. Single-VO instances, apart from ATLAS, do not have the necessary access_rule_vo permission available.

Modification
------------
Only check for the access_rule_vo permission if the instance is running in multi-VO mode. The new function is_multi_vo() tests the instance for multi-VO configuration. If this returns False, the permission on access_rule_vo is not checked.

ijjorama:
Changed code in lib/rucio/api/rule.py so that single-VO instances don't check for access_rule_vo.

### DIFF
--- a/lib/rucio/api/rule.py
+++ b/lib/rucio/api/rule.py
@@ -22,6 +22,14 @@ from rucio.common.schema import validate_schema
 from rucio.common.types import InternalAccount, InternalScope
 from rucio.common.utils import api_update_return_dict
 from rucio.core import rule
+from rucio.common.config import config_get_bool
+
+def is_multi_vo():
+    """
+    Check whether this instance is configured for multi-VO
+    returns: Boolean True if running in multi-VO 
+    """
+    return config_get_bool('common', 'multi_vo', raise_exception=False, default=False)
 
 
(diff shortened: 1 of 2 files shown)


lib/rucio/api/rule.py
```diff
@@ -22,6 +22,14 @@
...
 from rucio.common.utils import api_update_return_dict
 from rucio.core import rule
+from rucio.common.config import config_get_bool
+
+def is_multi_vo():
+    """
+    Check whether this instance is configured for multi-VO
+    returns: Boolean True if running in multi-VO 
+    """
+    return config_get_bool('common', 'multi_vo', raise_exception=False, default=False)
```
bziemons: This would fit better into `lib/rucio/core/vo.py` and imported from there

bari12:
Thanks @ijjorama - there are two small syntax tests failing. Please update the PR. Otherwise this is ready to merge. 👍

bziemons (commented):
looks good

bari12 (approved):

PR #4260, opened 2021-01-21
API: Provide setting for loading endpoints; Fix #4043

### LINKED ISSUE #4043: Flask dynamic endpoint loading
Motivation
----------
Currently Flask is used with a single endpoint in testing. This is required for some of the endpoint blueprints, creating certain URL rules for backwards compatibility with the v1 API structure. The single endpoint is defined here: https://github.com/rucio/rucio/blob/a1ca9008067f49c3fcbddc769f218d7078f07a71/lib/rucio/web/rest/flaskapi/v1/main.py

Modification
------------
The standard list of endpoints should be loaded by default: https://github.com/rucio/containers/blob/fef758e0f90f9cc4e359078cbd9af8c95db46393/server/rucio.conf.j2#L89-L110 and errors thrown while loading them should be propagated immediately. This is to avoid silent errors and unexpected behavior. Endpoints may be added to or removed from this list in the Rucio configuration.

bziemons:
Add [api] section and endpoints configuration variable for configuring the dynamic addition/removal of Flask endpoints.
Also rename all Flask blueprint modules so they can be identified by their api prefix.
I reversed the testing matrix here as well, so that Flask is now tested fully and webpy only on a few cases.
Closes #4043


bziemons:
Follow-up: Need to adjust config templates and configs in [rucio/containers](https://github.com/rucio/containers)

bari12 (approved):

Issue #4220, opened 2021-01-07
Logging review


bari12:
Motivation
----------
Over the years the log messages in Rucio evolved into an inconsistent state. This leads to difficulties in reading the logs and in some cases also to a large number of log messages being produced. The aim of this review is to
- Enforce consistency between the different log levels. DEBUG / INFO / WARNING / ERROR / CRITICAL
  - Especially between INFO and DEBUG: INFO should be used for information useful to the user or for stable operation of a daemon/server, otherwise messages should go to DEBUG. However, this should still be generally still useful information and not just any log statements leftover from development - these should be removed.
  - In general daemons/server should only need to be operated in INFO mode.
- Log format
  - make the log format configurable. The log output should be different depending if the logging is used inside the servers or daemons.
  - Daemon messages should be in the format `daemon_name[heartbeat['assign_thread']/heartbeat['nr_threads']]: ...`
- Reduce number of messages 

Note: At the moment logging.basicConfig is called in a lot of different modules in the daemons and in the core. If a daemon sets the logging and then imports a module it could potentially overwrite the config. This should be removed from all modules and instead put in a central place that is only called once when the application starts.

Modification
------------
List of components to be checked. The individual component leaders are in charge of checking their components.
- [ ] Authentication & Authorization
- [X] Clients
- [ ] Consistency checks
- [x] Core & Internals
- [ ] Data Pre-Placement
- [ ] Database
- [ ] Dataset deletion
- [ ] Deletion
- [ ] DIRAC
- [x] Docker & Kubernetes
- [x] Documentation
- [ ] Functional tests
- [ ] Life time model
- [ ] Messaging
- [x] Monitoring & Logging
- [ ] Multi VO
- [ ] Policies
- [ ] Probes & Alarms
- [ ] Protocols
- [ ] Rebalancing
- [ ] Recovery
- [x] Release management
- [x] Replicas
- [x] REST & API
- [x] Rules
- [x] Subscriptions
- [x] Testing
- [x] Traces
- [x] Transfers
- [ ] WebUI

dchristidis:
I understand how the daemon logging is to be implemented inside the threads, but what about before (e.g. in `start()`)?

PR #4272, opened 2021-01-27
Monitoring & Logging: Unify logging setup #4220

tbeerman:


tbeerman:
@bari12 the release-patch-setup test will fail until you cherry-pick the flask dynamic endpoint loading patch into the release branch.

lib/rucio/common/logging.py
```diff
@@ -0,0 +1,50 @@
...
+
+from rucio.common.config import config_get
+
+config_loglevel = getattr(logging, config_get('common', 'loglevel', raise_exception=False, default='DEBUG').upper())
+config_logformat = config_get('common', 'logformat', raise_exception=False, default='%(asctime)s\t%(name)s\t%(process)d\t%(levelname)s\t%(message)s')
+
+
+logging.basicConfig(stream=sys.stdout,
+                    level=config_loglevel,
+                    format=config_logformat)
```
bziemons: should be moved into a function and called from a function. To get rid of running code on import

etc/rucio.cfg.template
```diff
@@ -1,6 +1,7 @@
 [common]
 logdir = /var/log/rucio
 loglevel = DEBUG
+logformat = %%(asctime)s\t%%(process)d\t%%(levelname)s\t%%(message)s
```
bziemons, line 4: Does the `config_get` actually unquote the `%%`?

lib/rucio/core/nongrid_trace.py
```diff
@@ -31,18 +31,34 @@
 from rucio.common.config import config_get, config_get_int
 from rucio.core.monitor import record_counter
 
-ERRLOG = logging.getLogger('errlog')
-ERRLOG.setLevel(logging.ERROR)
+CONFIG_COMMON_LOGLEVEL = getattr(logging, config_get('common', 'loglevel', raise_exception=False, default='DEBUG').upper())
+CONFIG_COMMON_LOGFORMAT = config_get('common', 'logformat', raise_exception=False, default='%(asctime)s\t%(process)d\t%(levelname)s\t%(message)s')
```
bziemons, line 35: why do we need this here again? Can't we re-use this from `lib/rucio/common/logging.py`?

lib/rucio/core/trace.py
```diff
@@ -37,18 +37,35 @@
 from rucio.common.config import config_get, config_get_int
 from rucio.core.monitor import record_counter
 
-ERRLOG = logging.getLogger('errlog')
-ERRLOG.setLevel(logging.ERROR)
+
+CONFIG_COMMON_LOGLEVEL = getattr(logging, config_get('common', 'loglevel', raise_exception=False, default='DEBUG').upper())
+CONFIG_COMMON_LOGFORMAT = config_get('common', 'logformat', raise_exception=False, default='%(asctime)s\t%(process)d\t%(levelname)s\t%(message)s')
```
bziemons, line 42: see above.

lib/rucio/daemons/conveyor/poller.py
```diff
@@ -311,9 +306,9 @@ def poll_transfers(external_host, xfers, prepend_str='', request_ids=None, timeo
                     transfer_core.touch_transfer(external_host, transfer_id)
                 except (DatabaseException, DatabaseError) as error:
                     if re.match('.*ORA-00054.*', error.args[0]) or re.match('.*ORA-00060.*', error.args[0]) or 'ERROR 1205 (HY000)' in error.args[0]:
-                        logging.warning(prepend_str + "Lock detected when handling request %s - skipping" % request_id)
+                        logger(logging.WARNING, "Lock detected when handling request %s - skipping" % request_id)
                     else:
-                        logging.error(traceback.format_exc())
-            logging.debug(prepend_str + 'Finished updating %s transfer requests status (%i requests state changed) in %s seconds' % (len(xfers), cnt, (time.time() - tss)))
+                        logger(logging.ERROR, exc_info=True)
```
bziemons: I think this needs a message, at least an empty string

lib/rucio/daemons/conveyor/poller.py
```diff
@@ -311,9 +306,9 @@ def poll_transfers(external_host, xfers, prepend_str='', request_ids=None, timeo
...
+            logger(logging.DEBUG, 'Finished updating %s transfer requests status (%i requests state changed) in %s seconds' % (len(xfers), cnt, (time.time() - tss)))
     except Exception:
-        logging.error(traceback.format_exc())
+        logger(logging.ERROR, exc_info=True)
```
bziemons: see above

lib/rucio/web/rest/flaskapi/v1/replicas.py
```diff
@@ -178,7 +176,7 @@ def generate(vo):
         except RucioException as error:
             return generate_http_error_flask(500, error.__class__.__name__, error.args[0])
         except Exception as error:
-            print(format_exc())
+            logging.log(logging.ERROR, "Internal Error", exc_info=True)
```
bziemons: why not use `logging.exception("Internal Error")` like everywhere?

lib/rucio/core/transfer.py
```diff
@@ -1076,7 +1078,7 @@ def get_transfer_requests_and_source_replicas(total_workers=0, worker_number=0,
                     transfers[req_id]['use_ipv4'] = True
 
         except Exception:
-            logging.critical("Exception happened when trying to get transfer for request %s: %s" % (req_id, traceback.format_exc()))
+            logger(logging.CRITICAL, "Exception happened when trying to get transfer for request %s: %s" % (req_id, traceback.format_exc()))
```
bziemons: use `exc_info=True` instead of `traceback.format_exc()`

lib/rucio/daemons/conveyor/common.py
```diff
@@ -450,7 +446,7 @@ def bulk_group_transfer(transfers, policy='rule', group_bulk=200, source_strateg
     return grouped_jobs
 
 
-def get_conveyor_rses(rses=None, include_rses=None, exclude_rses=None, vos=None):
+def get_conveyor_rses(rses=None, include_rses=None, exclude_rses=None, vos=None, logger=logging.getLogger):
```
bziemons: this wouldn't work. This either needs some special handling or `logger=logging.log`.

idea for special handling (could also be done on a higher level, if useful):
```py
LOGGER = logging.getLogger(…)

def get_conveyor_rses(…, logger=LOGGER.log): 
```

bziemons (changes requested):
Overall looks good. What bothers me is the logging setup plain in the logging module instead of a function. This is basically a replacement for the previous bad behavior on setting up the logger each time the logging module is imported.

tbeerman:
The module only provides the formatted logger and the logging setup. So when you explicitly import the module you would anyway want logging and then the setup would be okay. I don't see the issue.

etc/rucio.cfg.template
tbeerman: Yes. With a single % it doesn't work. Another problem is the \t. It's not converted to an actual tab but it ends up as \t in the logs. I still have to figure out how to do that.

lib/rucio/core/nongrid_trace.py
tbeerman: The two trace modules are a bit special. I want a specific logger that write only the raw traces two a rotating log file as a buffer. The problem is, if I only setup the rotating logger, then these logs will also be duplicated to the root logger. That's why I added this hack to remove the handlers from the root logger and reconfigure it here, therefore I need these confit values again.

lib/rucio/daemons/conveyor/common.py
tbeerman: Yeah, that's just a plain typo.

lib/rucio/web/rest/flaskapi/v1/replicas.py
tbeerman: Yeah, no idea why I did it differently here. Will fix it.

lib/rucio/tests/test_bin_rucio.py
```diff
@@ -308,8 +308,8 @@ def test_upload_file_register_after_upload(self):
         cmd = 'rucio -v upload --rse {0} --scope {1} {2} {3} {4} --register-after-upload'.format(self.def_rse, self.user, tmp_file1, tmp_file2, tmp_file3)
         print(self.marker + cmd)
         exitcode, out, err = execute(cmd)
-        print(out)
-        print(err)
+        print("out")
+        print("err")
```
bziemons: this doesn't make sense

bziemons (changes requested):

bziemons (approved):

bziemons:
Looks good, thanks!

Edit: also all tests against release-1.24 are running fine now as well.

bari12:
This is extremely nice :-) Also thanks for testing @bziemons

PR #4278, opened 2021-01-29
Policy: RSE Expression regexp too restrictive; Fix #4277

### LINKED ISSUE #4277: Too restrictive RSE Expression regular expressions for RSE names
Motivation
----------
The default regular expression for RSE Expression RSE Names is too restrictive.
For example CMS uses lowercase characters in their RSE Names.

bari12:
=


bziemons (approved):

dchristidis:
Would you mind providing some context, either here or in the issue? I don’t understand why this expression can be anything other than https://github.com/rucio/rucio/blob/master/lib/rucio/common/schema/generic.py#L92 (minus the line anchors).

_Edit:_ I see that CMS allows small letters in their RSE names. Still, I would suggesting documenting this motivation in the issue.

PR #4281, opened 2021-01-29
API: Provide setting for loading endpoints; Fix #4043

### LINKED ISSUE #4043: Flask dynamic endpoint loading

bziemons:
Rename all Flask blueprint modules so they can be identified by their api prefix.

Backport to `release-1.24`: no change of the test matrix.


tbeerman (dismissed):

bziemons:
I had to replace the `None` default with an empty string in https://github.com/rucio/rucio/blob/55d5ef1238187852a312eef61dea2adb5d82e6e7/lib/rucio/web/rest/flaskapi/v1/main.py#L67

tbeerman (approved):

PR #4299, opened 2021-02-04
REST & API: Add templates for oidc in Flask; Fix #4042

### LINKED ISSUE #4042: OIDC authentication endpoint on Flask
Motivation
----------
Currently the OIDC authentication endpoint on Flask returns simple one-liner texts. Since the webpy version returned HTML via an HTML template, this needs to be evaluated for necessity and ported accordingly. This is probably related to #4041 since it uses templates as well. This is a leftover task from #3975 since the OIDC endpoints weren't created for Flask before.

Modification
------------
Create HTML templates for the Flask OIDC authentication endpoints and render them instead of returning simple text.

bziemons:
Closes #4042


mlassnig (approved):
LGTM

Issue #4097, opened 2020-10-30
Develop a lightweight client for Rucio


jamesp-epcc:
It would be desirable to have a lightweight version of the Rucio client with fewer dependencies to make installation by end users easier. Mario suggests the following things need to happen for this:

1. Get rid of libraries used for convenience and instead use Python standard libraries (e.g. network interaction, printing out, progress bar, tabulate, dogpile, etc.).
2. Remove server side mapping of library to be used for access to each RSE, instead make this a client side option so client can use (e.g.) HTTP or Xrootd instead of GFAL.
3. Client should dynamically check its environment to see what libraries are available for accessing the storage.

Questions to be resolved:

1. **Should the lightweight client be a replacement for the standard Rucio client, or an additional standalone alternative client?** If we can do this without too much disruption and still support all of the required features, it could be a modification to the standard client. This would avoid having to maintain two separate client code bases and keep them in sync.
2. **Which features does the lightweight client need to support?** Upload, download and list replicas are definitely required. Upload and download are the difficult ones since they involve direct interaction with the storage; other features are mostly simple REST calls to the server so it might be feasible to support them all. On the other hand, if the lightweight client is standalone, fewer features would be easier to keep in sync with the main client.
3. **Should this be Python 3 only, or also support Python 2.7?** Dropping Python 2 would allow us to rely on the richer features available in the Python 3 standard library (e.g. urllib to replace requests), simplifying the code, but would limit the environments in which the client could run, which may be unacceptable since the main purpose of this change is to allow the client to easily be installed anywhere.

As a first step I will study the relevant code in detail and come up with a plan (or maybe multiple options for further discussion).

mlassnig:
Hi James, thanks. I think the ideal way would be to modify the client directly, and phase out usage of existing dependencies over time, because there's no reason to keep a "heavy" client in any case. This would simplify everyone's life. For Py2/3 I have a slight tendency to drop support for Py2 in the client, but we should maybe see how much we can already drop and still stick to Py2.

jamesp-epcc:
I have looked at the client code to see where all the current dependencies are used. I thought I would write up my ideas on how to deal with each one. Please let me know if you have other suggestions.

- `setuptools` - only used in `setup.py`. I guess we need to keep this, but if we make the client Py3 only we could probably get rid of the upper version limit.
- `argcomplete` - can't remove this without losing tab completion of RSEs, but could make it optional.
[...]

PR #4300, opened 2021-02-04
Clients: remove several dependencies from client #4097

jamesp-epcc:
This pull request removes several of the "easier" dependencies from the Rucio client. Some have been moved to the server or test dependencies since they were not actually needed for the client (e.g. boto3, python-magic, futures). Others have been made optional.

boto and pysftp only need to be installed if RSEs using those protocols need to be accessed.

argcomplete will be used if it's there but is no longer a hard dependency.

tabulate and progressbar have also been made optional by adding very basic substitutes that will be used if they are not present. I know some people were against this kind of change so I am open to discussion about whether this is a good idea or how it could be done better.

Would it be a good idea to document the optional dependencies somewhere, so that people are aware of what they are and why they might want to install them?


bin/rucio-admin
```diff
@@ -72,6 +69,34 @@ try:
...
+            result += "\n"
+            for i in range(0, len(headers)):
+                result += ("-" * collens[i]) + "  "
+            result += "\n"
+        for row in table:
+            for i in range(0, len(row)):
+                val = str(row[i])
+                result += val + (" " * ((collens[i] + 2) - len(val)))
+            result += "\n"
+        return result
```
bziemons: Please do not create duplicate code. Move this to some module and import it from there

etc/pip-requires
```diff
@@ -16,3 +16,9 @@ paramiko==2.7.2                                   # SSH2 protocol library
 Flask==1.1.2                                      # Python web framework
 oic==1.2.1                                        # for authentication via OpenID Connect protocol
 prometheus_client==0.9.0                          # Python client for the Prometheus monitoring system
+python-magic>=0.4.15,<0.5.0                       # File type identification using libmagic
+futures>=3.2.0; python_version < '3.0'            # Clean single-source support for Python 3 and 2
+boto>=2.49.0,<2.50.0                              # S3 boto protocol
+boto3>=1.9.130,<1.17.0                            # S3 boto protocol (new version)
```
bziemons, line 22: I don't think we want both versions

etc/pip-requires-client
```diff
@@ -1,18 +1,8 @@
 # All dependencies needed to run rucio client should be defined here
 setuptools>=36.8.0,<45.0.0; python_version < '3.0'          # Python packaging utilities (36.8.0 last py2.6 compatible version, 45.0 only Py3 compatible)
 setuptools>=45.0.0,<=50.3.2; python_version >= '3.0'        # Python packaging utilities
-argcomplete>=1.9.0,<=1.12.2                                 # Bash tab completion for argparse
 requests>=2.20.0,<=2.25.0                                   # Python HTTP for Humans.
 urllib3>=1.24.2,<1.26                                       # HTTP library with thread-safe connection pooling, file post, etc.
 dogpile.cache>=0.6.5,<0.7.0; python_version < '3.0'         # Caching API plugins (0.7.0 only Py3 compatible)
 dogpile.cache>=0.6.5,<=1.1.1; python_version >= '3.0'       # Caching API plugins
-boto>=2.49.0,<2.50.0                                        # S3 boto protocol
-tabulate>=0.8.0,<0.9.0                                      # Pretty-print tabular data
```
bziemons, line 10: These should not be removed here, but moved to optional dependencies in the client's setup.py

bziemons (changes requested):

bziemons:
Sorry if this PR wasn't meant for being reviewed yet. If that's the case, please convert it to a draft (see _Reviewers_ panel on the top-right).

bin/rucio-admin
jamesp-epcc: I have moved this to `rucio.common.utils`.

etc/pip-requires
jamesp-epcc: boto is used by the legacy S3 code in Rucio. For the new version using signed URLs I moved to boto3 because it supports offline signing which is useful. It might be possible to retire the legacy code and move entirely to the new code, then the boto dependency could go.

etc/pip-requires-client
jamesp-epcc: I have moved these to optional dependencies. I am open to suggestions for the names of them, and also whether it would be better to merge some of the minor ones into a single extra that people could install if they don't care about the number of dependencies and want all the original functionality, rather than having several different ones.

etc/pip-requires-client
```diff
@@ -1,18 +1,8 @@
...
-progressbar2>=3.39.0,<=3.53.1                               # Text progress bar
-bz2file>=0.98,<0.99                                         # Read and write bzip2-compressed files.
-python-magic>=0.4.15,<0.5.0                                 # File type identification using libmagic
-futures>=3.2.0; python_version < '3.0'                      # Clean single-source support for Python 3 and 2
 six>=1.12.0<1.16.0                                          # Python 2 and 3 compatibility utilities
-boto3>=1.9.130,<1.17.0                                      # S3 boto protocol (new version)
-pysftp>=0.2.9,<0.3                                          # Used by sftp protocol
-jsonschema>=3.2.0                                           # For JSON schema validation (Policy modules)
```
bari12, line 18: Is this not used anywhere?

etc/pip-requires-client
bari12: I think to put these as optional is the way to go. No need to include a specific protocol per default, if you are not using it.

lib/rucio/common/utils.py
```diff
@@ -1435,3 +1435,36 @@ def __call__(self, mtries=3, logger=None):
                     logger.debug(str(e))
                 attempt -= 1
         return self.func(*self.args, **self.kwargs)
+
+
+def tabulate(table, tablefmt='plain', headers=None):
```
bari12, line 1440: While this works, I am really wondering how much we gain by doing it. I just cannot fathom that including `tabulate` as a pypi dependency would be an issue for anyone. I could somewhat understand not using any dependencies at all, since there is some merit in that (Although very unrealistic to achieve), but since we are using some anyway, I don't see how these super simple ones are causing any issues. I am not advocating to change it, just noting it.

lib/rucio/rse/protocols/webdav.py
```diff
@@ -36,13 +36,32 @@
...
 from urllib3.poolmanager import PoolManager
 
 from rucio.common import exception
 from rucio.rse.protocols import protocol
 
+try:
+    from progressbar import ProgressBar
+except ImportError:
+    # make progressbar an optional dependency
+    class ProgressBar:
```
bari12, line 50: Why do we have progressbar here at all? Do we have it in any of the other protocols?

etc/pip-requires-client
jamesp-epcc: boto3 is only used by the server to sign URLs. I don't think it was ever needed on the client side. pysftp is only needed for the sftp protocol so I made it optional. jsonschema seemed to be only used in the policy packages which I don't think are needed for the client. If I'm wrong and they are needed, this dependency should be kept.

lib/rucio/common/utils.py
jamesp-epcc: I agree it would probably be unrealistic to aim for zero dependencies. I am happy to leave the tabulate dependency if people think it's unlikely to be a problem in practice.

lib/rucio/rse/protocols/webdav.py
jamesp-epcc: It's only used in webdav. I'm not sure why that protocol uses it but none of the others do.

bari12: @mlassnig any idea here? I think we should remove this.

mlassnig: o.O no idea.

i'd rather have a progressbar that is universally usable for all protocols

bari12: Exactly, but that should happen outside the protocol and not within.

jamesp-epcc:
Closing this PR as I have now submitted a new one, taking on board the discussion above.

PR #4301, opened 2021-02-04
Monitoring & Logging: add formatted_logger to protocols #4220

tbeerman:

### DIFF
--- a/bin/rucio
+++ b/bin/rucio
@@ -17,7 +17,7 @@
 # - Mario Lassnig <mario.lassnig@cern.ch>, 2012-2020
 # - Vincent Garonne <vincent.garonne@cern.ch>, 2012-2018
 # - Martin Barisits <martin.barisits@cern.ch>, 2012-2020
-# - Thomas Beermann <thomas.beermann@cern.ch>, 2012-2017
+# - Thomas Beermann <thomas.beermann@cern.ch>, 2012-2021
 # - Yun-Pin Sun <winter0128@gmail.com>, 2012-2013
 # - Cedric Serfon <cedric.serfon@cern.ch>, 2013-2020
 # - Ralph Vigne <ralph.vigne@cern.ch>, 2013
@@ -96,7 +96,7 @@ setup_gfal2_logger(gfal2_logger)
 
 
 def signal_handler(sig, frame):
-    logger.warning('You pressed Ctrl+C! Exiting gracefully')
+    logger(logging.WARNING, 'You pressed Ctrl+C! Exiting gracefully')
     child_processes = subprocess.Popen('ps -o pid --ppid %s --noheaders' % os.getpid(), shell=True, stdout=subprocess.PIPE)
     child_processes = child_processes.stdout.read()
     for pid in child_processes.split("\n")[:-1]:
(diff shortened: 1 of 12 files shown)


tbeerman:
I "just" wanted to add the prefixed logger to the gfal protocol but I had to change a some more places from the clients to automatix and the reaper since it's all connected somehow.

lib/rucio/rse/rsemanager.py
```diff
@@ -55,6 +55,7 @@
 from rucio.common import exception, utils, constants
 from rucio.common.config import config_get_int
 from rucio.common.constraints import STRING_TYPES
+from rucio.common.logging import formatted_logger
 from rucio.common.utils import make_valid_did, GLOBALLY_SUPPORTED_CHECKSUMS
 
 _logger = logging.getLogger(__name__)
```
bziemons: remove unused code

bziemons (dismissed):
Changes a few log outputs. As long as any third-party program that may or may not read them is fine with it, this looks good

lib/rucio/client/downloadclient.py
```diff
@@ -134,7 +135,7 @@ def __init__(self, client=None, logger=None, tracing=True, check_admin=False, ch
         self.logger = logger
         self.tracing = tracing
         if not self.tracing:
-            logger.debug('Tracing is turned off.')
+            logger(logging.DEBUG, 'Tracing is turned off.')
```
bziemons, line 134: above:
```diff
 if not logger:
             logger = logging.getLogger('%s.null' % __name__)
             logger.disabled = True
+            logger = logger.log
```

lib/rucio/client/uploadclient.py
```diff
@@ -62,7 +62,7 @@
 
 class UploadClient:
 
-    def __init__(self, _client=None, logger=None, tracing=True):
+    def __init__(self, _client=None, logger=logging.log, tracing=True):
```
bziemons, line 65: Check consistency see above. If you add `logger=logging.log` here, check that no-one explicitly passes `None` and remove the `if not logger:` below.

lib/rucio/client/uploadclient.py
```diff
@@ -686,15 +686,15 @@ def _retry_protocol_stat(self, protocol, pfn):
...
                 fail_str = ['The requested service is not available at the moment', 'Permission refused']
                 if any(x in str(error) for x in fail_str):
                     raise error
-                self.logger.debug('stat: unknown edge case, retrying in %ss' % 2**attempt)
+                self.logger(logging.DEBUG, 'stat: unknown edge case, retrying in %ss' % 2**attempt)
                 time.sleep(2**attempt)
         return protocol.stat(pfn)
 
-    def _create_protocol(self, rse_settings, operation, force_scheme=None):
+    def _create_protocol(self, rse_settings, operation, force_scheme=None, logger=logging.log):
```
bziemons: be consistent in this function as well. I would suggest using `self.logger` if it is available, instead of a logger parameter.

bin/rucio
```diff
@@ -2310,13 +2310,13 @@ if __name__ == '__main__':
 
     logger = None
```
bziemons, line 2311: this module-wide variable is shadowed by the parameter "logger" and makes it hard to track usage of either in the review. I would suggest renaming this to something like `module_logger`

bziemons (changes requested):
The usage of the name `logger` as a parameter, class attribute and global module variable makes it extremely hard to see through this. I know this is old code, but it would be extremely helpful if you could clean this up.

lib/rucio/client/uploadclient.py
```diff
@@ -62,17 +62,13 @@
 
 class UploadClient:
 
-    def __init__(self, _client=None, logger=None, tracing=True):
+    def __init__(self, _client=None, logger=logging.log, tracing=True):
         """
         Initialises the basic settings for an UploadClient object
 
         :param _client:     - Optional: rucio.client.client.Client object. If None, a new object will be created.
         :param logger:      - logging.Logger object to use for uploads. If None nothing will be logged.
```
bziemons: just saw this doc string here, should be updated wherever present and probably added when absent

lib/rucio/daemons/automatix/automatix.py
```diff
@@ -233,34 +232,34 @@ def automatix(sites, inputfile, sleep_time, account, worker_number=1, total_work
...
                     physical_fname = '%s/%s' % (tmpdir, "".join(fname.split('/')))
                     physical_fnames.append(physical_fname)
-                    generate_file(physical_fname, filesize)
+                    generate_file(physical_fname, filesize, logger=logger)
                     fnames.append(fname)
-                logging.info('%s Upload %s to %s', prepend_str, dsn, site)
+                logger(logging.INFO, 'Upload %s to %s', dsn, site)
                 dsn = '%s:%s' % (scope, dsn)
-                status = upload(files=lfns, scope=scope, metadata=metadata, rse=site, account=account, source_dir=tmpdir, worker_number=worker_number, total_workers=total_workers, dataset_lifetime=dataset_lifetime, did=dsn, set_metadata=set_metadata)
+                status = upload(files=lfns, scope=scope, metadata=metadata, rse=site, account=account, source_dir=tmpdir, dataset_lifetime=dataset_lifetime, did=dsn, set_metadata=set_metadata)
```
bziemons: missing a `logger=logger` ?

bziemons (changes requested):

bziemons (approved):

PR #4308, opened 2021-02-09
Monitoring & Logging: fix logging bug in finisher #4220

tbeerman:

### DIFF
--- a/lib/rucio/daemons/reaper/reaper2.py
+++ b/lib/rucio/daemons/reaper/reaper2.py
@@ -144,7 +144,7 @@ def get_rses_to_process(rses, include_rses, exclude_rses, vos):
     return rses
 
 
-def delete_from_storage(replicas, prot, rse_info, staging_areas, auto_exclude_threshold, logger):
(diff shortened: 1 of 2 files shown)


lib/rucio/daemons/reaper/reaper2.py
```diff
@@ -144,7 +144,7 @@ def get_rses_to_process(rses, include_rses, exclude_rses, vos):
     return rses
 
 
-def delete_from_storage(replicas, prot, rse_info, staging_areas, auto_exclude_threshold, logger):
+def delete_from_storage(replicas, prot, rse_info, staging_areas, auto_exclude_threshold, logger=logging.log):
```
mlassnig, line 147: is this on purpose, here `logger=logging.log` and the ones below `logger=logger` ?

tbeerman: Yes, the default is set to logging.log so that this function will also work without passing a logger. Below I’m then passing the decorated logger which is just a wrapper around logging.log that adds the prefix.

mlassnig (dismissed):

mlassnig (approved):

bziemons (approved):

PR #4310, opened 2021-02-09
core: invalidate config caches on option creation and deletion. Closes #3916

### LINKED ISSUE #3916: Problem with config endpoint
Motivation
----------
The autotests to test the return value of `config.get_config(None, None)` fail, because the returned `dict` does not contain the same sections that were saved to it beforehand.

The simple test can be seen here: https://github.com/rucio/rucio/blob/d3418b28a0dfc87d96a92eae5e4786dddf147575/lib/rucio/tests/test_config.py#L68-L79
and this is the previously existing and more complex test: https://github.com/rucio/rucio/blob/d3418b28a0dfc87d96a92eae5e4786dddf147575/lib/rucio/tests/test_config.py#L104-L117

rcarpa:
The config read operations are accelerated by cache. This cache is never updated and can return obsolete data.

The commit also changes the format of the options cache key. Until now it was "options", but it doesn't seem right, because there a separate options cache per section.

### DIFF
--- a/lib/rucio/core/config.py
+++ b/lib/rucio/core/config.py
@@ -18,11 +18,13 @@
 # - Cedric Serfon <cedric.serfon@cern.ch>, 2017
 # - Hannes Hansen <hannes.jakob.hansen@cern.ch>, 2018-2019
 # - Brandon White <bjwhite@fnal.gov>, 2019-2020
+# - Radu Carpa <radu.carpa@cern.ch>, 2021
 #
 # PY3K COMPATIBLE
 
 from dogpile.cache import make_region
 from dogpile.cache.api import NoValue
+from sqlalchemy import func
 
 from rucio.common.exception import ConfigNotFound
 from rucio.common.config import config_get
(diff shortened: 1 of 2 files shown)


lib/rucio/core/config.py
```diff
@@ -23,6 +23,7 @@
```
mlassnig, line 24: looks good to me, don't forget to add yourself to the authorlist at the top

rcarpa: Done, updated the MR.

bziemons (approved):

bziemons:
https://github.com/rucio/rucio/blob/master/CONTRIBUTING.rst#human-review

> Merging of approved pull requests is done by the Rucio development lead.

rcarpa:
Hi. It is my bad. I did read it, but forgot due to an old habit. Spoke already with @mlassnig on this subject and will leave it merged this time, but will be careful for other MR

PR #4318, opened 2021-02-10
Testing: Update docker-compose configs; Fix #4316

### LINKED ISSUE #4316: Replace `links` in docker-compose
Motivation
----------
The links flag in docker is deprecated: https://docs.docker.com/network/links/
While for docker-compose this does not necessarily apply, podman-compose fails to run our configurations.

Modification
------------
Switch to a single network namespace for all the docker-compose files in the repo.

bziemons:
Use a single network namespace and expose only rootless ports (80 -> 8080, 443 -> 8443). Add all hostnames for all services to the rucio container. This makes the configurations runnable rootless and with podman-compose. Also adjust documentation in `etc/docker/dev/README.rst`.
Fixes #4316


bziemons:
fyi: this needs a fairly recent Docker and docker-compose version.

etc/docker/dev/docker-compose-storage-alldb.yml
```diff
@@ -1,129 +1,106 @@
...
+      - ../../../lib:/opt/rucio/lib:Z
+    environment:
+      - X509_USER_CERT=/opt/rucio/etc/usercert.pem
+      - X509_USER_KEY=/opt/rucio/etc/userkey.pem
+      - RDBMS=postgres11
   ruciomy5:
-    image: mysql:5
-    hostname: ruciomy5
+    image: docker.io/mysql:5
+    network_mode: "service:rucio"
```
rcarpa, line 40: A thing to be taken into consideration: because the network binds to the service "rucio", re-creating only this container breaks connectivity between it and other containers. It is necessary to re-create the whole docker-compose environment if the rucio container is restarted.

tbeerman (changes requested):
Overall looks good. But can you please add back the ports? It can be quite useful to access some services directly from the host, e.g., for webui development, directly checking the DB or accessing Kibana/Grafana. Most ports should be fine even without root and for the server you could take something like 8443 and for grafana maybe 8080.

etc/docker/dev/docker-compose-storage-alldb.yml
bziemons: I can avoid this by adding an infra container and attach everything to the infra container. Would that be better?

rcarpa: In my local config it works correctly without any network_mode at all. And without any extra_hosts. By default all containers are put into the same network and can access each-other by their name = hostname. Any reason why we want to diverge from the default behavior ?

bziemons: The answer to that is in the PR description

rcarpa: ACK. it's thus due to the podman restrictions. I have no preferences at all, I'll have to use my own docker-compose file override anyway because i'll customize it for my personal workflow. Good for me to leave it as-is

tbeerman (approved):

PR #4338, opened 2021-02-15
Core: Stop meta variable from being reused and re-encoded over and over Fix #4337

### LINKED ISSUE #4337: core/rule.py reencodes meta over and over again
Motivation
----------

core/replica.py reencodes meta to JSON over and over again inside a loop on DIDs. Eventually this makes a string so big it causes an exception in oracle. It's not clear if this also corrupts the intended metadata

Modification
------------

Use a dedicated variable for the results of the JSON encoding

ericvaandering:


ericvaandering:
This bug appears to be years old. May be a good candidate for porting back to LTS releases

bari12 (approved):

Issue #2311, opened 2019-03-12
Functional download/upload tests with xrootd containers


bari12:
Motivation
----------
At the moment we only have very limited upload/download tests either using mock calls or direct posix protocols. One possibility to expand these to actualy functional tests would be to use xrootd docker containers and start up two xrootd storages in the test environment. This would allow us to test upload/download also with different protocols, such as gfal, which would also help enormously to validate gfal for python 3.6.

rcarpa:
I'm working on increasing the coverage of the tests in test_download.py. I need better tests to be able to work more confidently on #4153 . Let me know if I risk to do duplicate work.

PR #4340, opened 2021-02-16
Testing: improve download client test coverage; #2311

rcarpa:
- check the status of different results to ensure that files are
  correctly overwritten or not
- check the resulting path with/without no_subdir set
- add test for downloading from an archive
- implicitly test XRD download because it is used in the archive test
- add tests for downloading multiple files via a wildcard
- implicitly test some upload logic
- some verifications that trace_copy_out works as expected
- ensure that checksum validation is enforced (or ignored if option set)

### DIFF
--- a/lib/rucio/tests/test_download.py
+++ b/lib/rucio/tests/test_download.py
@@ -20,24 +20,30 @@
 # - Patrick Austin <patrick.austin@stfc.ac.uk>, 2020
 # - Benedikt Ziemons <benedikt.ziemons@cern.ch>, 2020
 # - Thomas Beermann <thomas.beermann@cern.ch>, 2021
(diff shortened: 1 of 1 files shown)


lib/rucio/tests/test_download.py
```diff
@@ -48,40 +54,271 @@ def setUp(self):
         logger.addHandler(logging.StreamHandler())
         logger.setLevel(logging.DEBUG)
         self.client = Client()
+        self.did_client = DIDClient()
```
bari12, line 57: Only a minor detail the `Client()` includes all clients, except the `UploadClient` and `DownloadClient` - so you do not explicitly need to import it.

bari12 (approved):

PR #4351, opened 2021-02-23
Clients: re-download existing files if checksum is wrong. Fixes #4323

### LINKED ISSUE #4323: "rucio download --no-subdir" behavior is different from its documentation
Motivation
----------
Rucio documentation for file download states that the "--no-subdir" will overwrite files if they exist in the destination folder.
```
  --no-subdir           Don't create a subdirectory for the scope of the
                        files. Existing files in the directory will be
                        overwritten.

```

However, the files are not overwitten, as shown by the following example : 
```
$ echo "some_content" > /tmp/test/a_file
$ rucio upload --rse XRD1 --scope test /tmp/test/a_file
$ echo "other_content" > /tmp/test/a_file 
$ rucio download --no-subdir --dir /tmp/test/ test:a_file
   ----------------------------------
   Download summary
   ----------------------------------------
   DID mock:a_file
   Total files:                                  1
   Downloaded files:                             0
   Files already found locally:                  1
   Files that cannot be downloaded:              0
$ cat /tmp/test/a_file 
other_content
$ rm /tmp/test/a_file
$ rucio download --no-subdir --dir /tmp/test/ test:a_file
   ----------------------------------
   Download summary
   ----------------------------------------
   DID mock:a_file
   Total files:                                  1
   Downloaded files:                             1
   Files already found locally:                  0
   Files that cannot be downloaded:              0
# cat /tmp/test/a_file 
some_content
```

Modification
------------
Either the documentation must be updated to note that existing files will not be overwritten, or the files must be overwritten.

rcarpa:
Also update the "no-subdir" option documentation.

In the past, rucio was changing its behavior of overwriting existing
files depending on the 'no-subdir' option. If this option was set,
local files were always overwritten. If the option was not set, the
files were always let intact. This was counter-intuitive and was fixed
in a recent commit as being a bug. Leaving an incorrect documentation.

The discussion in the linked issue suggested that it will be the most
intuitive to check the checksum of existing files. If their checksum
is correct: leave them intact, but overwrite files in case of a
miss-match.

### DIFF
--- a/bin/rucio
+++ b/bin/rucio
@@ -41,6 +41,7 @@
 # - Jason Nielsen <jnielsen@ucsc.edu>, 2020
 # - Aristeidis Fkiaras <aristeidis.fkiaras@cern.ch>, 2020
 # - Patrick Austin <patrick.austin@stfc.ac.uk>, 2020
+# - Radu Carpa <radu.carpa@cern.ch>, 2021
 
 from __future__ import print_function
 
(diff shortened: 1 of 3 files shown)


mlassnig (approved):
LGTM

bziemons (approved):

PR #4352, opened 2021-02-23
Core & Api & Clients: Possibility to inject rules delayed. Closes #2639

### LINKED ISSUE #2639: Possibility to inject rules delayed
Motivation
----------
Creation date of rule in the future, injector will only inject it then.

rcarpa:
Add a new API option and associated CLI implementation to enable
delayed creation of requests for replication rules.

Rules are inserted with "created_at" set in the future. The injector
is also changed to take into consideration this case. It creates
requests for rules in "INJECT" state only if their created_at time has
arrived.

Also fix the test_create_rule in rucio bin tests. On errors, stdout
was empty, so the search regexp was empty and didn't catch errors.

### DIFF
--- a/bin/rucio
+++ b/bin/rucio
@@ -1192,7 +1192,8 @@ def add_rule(args):
                                                activity=args.activity,
                                                comment=args.comment,
                                                ask_approval=args.ask_approval,
-                                               asynchronous=args.asynchronous)
+                                               asynchronous=args.asynchronous,
+                                               delay_injection=args.delay_injection)
     except DuplicateRule as error:
         if args.ignore_duplicate:
             for did in dids:
@@ -1210,7 +1211,8 @@ def add_rule(args):
                                                           activity=args.activity,
                                                           comment=args.comment,
                                                           ask_approval=args.ask_approval,
-                                                          asynchronous=args.asynchronous)
+                                                          asynchronous=args.asynchronous,
+                                                          delay_injection=args.delay_injection)
                     rule_ids.extend(rule_id)
                 except DuplicateRule:
                     print('Duplicate rule for %s:%s found; Skipping.' % (did['scope'], did['name']))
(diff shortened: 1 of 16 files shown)


bin/rucio
```diff
@@ -2184,6 +2187,7 @@ You can filter by key/value, e.g.::
     add_rule_parser.add_argument('--comment', dest='comment', action='store', help='Comment about the replication rule')
     add_rule_parser.add_argument('--ask-approval', dest='ask_approval', action='store_true', help='Ask for rule approval')
     add_rule_parser.add_argument('--asynchronous', dest='asynchronous', action='store_true', help='Create rule asynchronously')
+    add_rule_parser.add_argument('--when-to-inject', dest='when_to_inject', action='store', help='Time when to start applying the rule (UTC)')
```
mlassnig: add example in the help how the user has to write the time

rcarpa: Good idea. Added an example.

lib/rucio/api/rule.py
```diff
@@ -13,17 +13,30 @@
...
+
+def __parse_datetime(time_str):
+    if not time_str:
+        return None
+    try:
+        if hasattr(datetime, 'fromisoformat'):
+            return datetime.fromisoformat(time_str)
+        return datetime.strptime(time_str, '%Y-%m-%dT%H:%M:%S')
+    except ValueError as error:  # NOQA, pylint: disable=W0612
+        raise InvalidObject("Problem parsing time stamp: %(error)s" % locals())
```
mlassnig: just wondering if repeating the correct time stamp format should also be added here?

something like      `expected format is YYYY-MM-DDTHH:MM:S`  ?

bari12: Also careful with UTC here. (All timestamps in the DB should be in UTC)
For most other commands where we use time/duration, we always used basically durations. e.g. for rules `--lifetime` instead of `--expires-at` etc. We could do something similar here with a `--delay-injection <seconds>`

rcarpa: Done. I initially tried to remove the `asynchronous` parameter from all core functions, but it became quickly very counter-intuitive to have to set `delay_injection=0` to mean "asynchronous" and  `delay_injection=None` to mean "synchronous"  As a result, I keep both parameters, but ensure that setting delay_injection implies asynchronous

PR #4363, opened 2021-02-24
Wider imp import for smoother forward/backward compatibility; Fix #4361

### LINKED ISSUE #4361: imp / importlib incompatibility
Motivation
----------

The fallback load of `imp` is not good enough, this would only start working on Python 3.9.

mlassnig:


bziemons:
Wouldn't it be better to add one function to import imp and have that comment? And replace all imports of imp with it?

mlassnig:
> Wouldn't it be better to add one function to import imp and have that comment? And replace all imports of imp with it?

There's a few "import imp" statements in other modules, which would need a more thorough rewrite of those parts. I want to avoid this for now as we need to get on with migrating the nodes and this part is blocking. The others will start "automatically failing" once we add Py3.9 tests :-)

bari12 (approved):

PR #4367, opened 2021-02-25
Transfers: tolerate missing distances in path computation. Closes #4132

### LINKED ISSUE #4132: Bug in transfer core when no outbound distances
Motivation
----------
If an RSE has no outgoing links defined, then the code at https://github.com/rucio/rucio/blob/f711667ee4a4ea8c898098f8380c311460dc2642/lib/rucio/core/transfer.py#L1431-L1444 returns an empty dictionary, and the calling code at https://github.com/rucio/rucio/blob/f711667ee4a4ea8c898098f8380c311460dc2642/lib/rucio/core/transfer.py#L564 raises a KeyError.

Modification
------------
I imagine setting `distance_graph = {rse_id: {}}` would fix this but I am not familiar with the calling code logic and how best to handle this error.

rcarpa:
The links without distance will be ignored.

__load_distance_edges_node was returning a dict containing a single key:
the current node. This un-needed indirection was the root cause of the
issue from the linked ticket.

Also, remove the check for the case that a destination RSE is isolated.
The chosen criteria was "absence of outgoing links", which doesn't seem
right: for a destination node, we are interested in the inbound links.
I was split between implementing the check using inbound links and
removing the check all-together. It is located in the critical path of
the execution and ads a database access for all path computations.

### DIFF
--- a/lib/rucio/tests/test_transfer.py
+++ b/lib/rucio/tests/test_transfer.py
@@ -1,0 +1,139 @@
+# -*- coding: utf-8 -*-
+# Copyright 2021 CERN
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Authors:
+# - Radu Carpa <radu.carpa@cern.ch>, 2021
+
+import pytest
+
+from rucio.common.exception import NoDistance
+from rucio.core.distance import add_distance
+from rucio.core.transfer import get_hops
+
+
+def test_get_hops(rse_factory):
+    # Build the following topology:
+    # +------+           +------+     10    +------+
+    # |      |     40    |      +-----------+      |
+    # | RSE0 |  +--------+ RSE1 |           | RSE2 +-------------+
+    # |      |  |        |      |      +----+      |             |
+    # +------+  |        +------+      |    +------+             | <missing_cost>
+    #           |                      |                         |
+    #           |                      |                         |
+    #           |                      |                         |
+    # +------+  |        +------+  10  |    +------+           +-+----+
+    # |      +--+        |      +------+    |      |   --20->  |      |
+    # | RSE3 |   --10->  | RSE4 |           | RSE5 +-->--->--->+ RSE6 |
+    # |      +-->--->--->+      +-----------+      |           |      |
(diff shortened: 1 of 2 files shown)


rcarpa:
Tests fail because the cache layer ([here](https://github.com/rucio/rucio/blob/40ca9a9b3f33d540a65d0279e61ab09734a8164c/lib/rucio/core/transfer.py#L503)) return a previously computed solution for a call with different value of the `multihop_rses` argument. Is it a bug we want to fix? Adding a new rse during the lifetime of the cache will not start using it for transfers.

bari12:
We do live with these cache-lifetime inconsistencies all over the code, while annoying for testing it's not really an issue in practice. For a new RSE, the cache won't hold any value, so it will just fully calculate.

What is true though, that depending on the settings (E.g. `multihop_rses`) the output is different. I mean `limit_dest_schemes` is part of the key for that reason. I don't really see a case though where we would run a deployment with different `multihop_rses`, but it might be better to just add them to the key should that ever be the case. (Not sure if key length is limited, maybe we need to hash this)

rcarpa:
My understanding is as follows: if we add a new RSE with `available_for_multihop=true` , it will not be used as a new intermediate hop until the cache expires. I completely miss the knowledge of how things work in production to judge if this particularity can be a problem.
If it is worth fixing, I'll sort the list and hash it; and add the hash to the cache key. Should I do it ?

bari12:
This is true, but it's not an issue in production since these things change very rarely. So waiting 10m for the cache to update is not an issue. Also `parse_expression` is cached as well, so it would not show up there either. The case of adding a new RSE is not really problematic. 

Thinking about it, I don't think it is worth adding it, since we also couldn't run multiple submitters where every one is running with different `multihop_rses`, it's all derived from the same rse_expression anyway. Just leave it like that.

lib/rucio/tests/test_transfer.py
```diff
@@ -0,0 +1,200 @@
...
+from lib.rucio.core.transfer import get_hops
+from rucio.common.config import config_get, config_get_bool
+from rucio.common.exception import NoDistance
+from rucio.common.types import InternalAccount
+from rucio.core.importer import import_data
+from rucio.core.rse import get_rse_id, del_rse
+from rucio.tests.common import rse_name_generator
+
+
+class TestTransfer(unittest.TestCase):
```
bziemons: just a side-note: do you know about pytest fixtures? https://docs.pytest.org/en/stable/fixture.html
I try to gradually replace `unittest.TestCase`, because we can save a lot of test code and therefore time with fixtures.

rcarpa: I already used them in the past. Now that I know it's OK to use them in rucio, I'll definitely use them more often.

bziemons (dismissed):
lgtm

bari12 (dismissed):

bari12:
Looks good, just re-submitting the tests because something seems to be failing.

rcarpa:
Found why test where failing. I used import() in tests, which (by default) removes all rses which are not in the imported list. This impacted other tests due to parallel runs.

bari12 (approved):

PR #4368, opened 2021-02-26
WebUI: Removed direct DB dependence from WebUI Fix #4319

### LINKED ISSUE #4319: Remove direct DB dependence from the WebUI
Motivation
----------
The WebUI has a direct dependence on the core/api, e.g. https://github.com/rucio/rucio/blob/master/lib/rucio/web/ui/main.py#L41 . This make it not possible to deploy an instance on a machine with not connected to the DB.

Modification
------------
Remove all direct DB dependence from the WebUI

DhruvSondhi:
Removed direct DataBase dependence from WebUI
------------------
This PR aims to Fix #4319 [Thanks, @bziemons] 
Direct DB dependence has been removed from `main.py` & `flask/bp.py`

### DIFF
--- a/lib/rucio/web/ui/flask/bp.py
+++ b/lib/rucio/web/ui/flask/bp.py
@@ -20,31 +20,14 @@
 
 from flask import Blueprint, request, render_template, make_response
 
-from rucio.api.authentication import get_auth_token_x509
 from rucio.common.config import config_get, config_get_bool
 from rucio.web.rest.flaskapi.v1.common import generate_http_error_flask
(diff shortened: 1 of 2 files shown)


tbeerman (commented):
I'm afraid that removing the DB dependence from the WebUI is a bit more complicated than that. This change will just break the account switcher. The place where the DB is mainly used is in common/utils.py and it's needed to get the initial auth token, the list of accounts and to select the default account. It's quite integrated with the Javascript part of the code so it would need some major rewrite.

DhruvSondhi:
I will try to dig deeper as you have mentioned with the file & try to make changes accordingly. Thank you very much for the review :) & for mentioning my mistake.

bari12:
Closing this in the meanwhile 👍

PR #4369, opened 2021-02-26
Expose signature_lifetime in replica client; Fix #3972

### LINKED ISSUE #3972: replica client: expose signature_lifetime
This is available in the API and REST endpoint, but the clients are not setting the value so it always defaults to the 600s in the REST endpoint.

mlassnig:
On purpose not exposed in the CLI.


bziemons (approved):
lgtm

Issue #4364, opened 2021-02-24
Add command list-suspicious-replicas to API


ChristophAmes:
Motivation
----------
Allows the user to get a list of all replicas marked as suspicious, as is possible on the rucio-ui.cern.ch website.

Modification
------------
Added a new method to lib/rucio/client/replicaclient.py
Added a new method and a new command to bin/rucio

PR #4371, opened 2021-03-01
API: Add command list-suspicious-replicas #4364

ChristophAmes:
API: Add command list-suspicious-replicas #4364

### DIFF
--- a/bin/rucio
+++ b/bin/rucio
@@ -42,6 +42,7 @@
 # - Aristeidis Fkiaras <aristeidis.fkiaras@cern.ch>, 2020
 # - Patrick Austin <patrick.austin@stfc.ac.uk>, 2020
 # - Radu Carpa <radu.carpa@cern.ch>, 2021
+# - Christoph Ames <christoph.ames@cern.ch>, 2021
 
 from __future__ import print_function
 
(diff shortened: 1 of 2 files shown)


cserf (dismissed):

bin/rucio
```diff
@@ -1489,6 +1490,42 @@ def list_rses(args):
...
+        replicas = i
+
+    data = [['', '', '', '', ''], ['RSE Expression:', 'Scope:', 'Created at:', 'Nattempts:', 'File Name:'], ['', '', '', '', '']]
+    for rep in replicas:
+        data.append([rep['rse'], rep['scope'], rep['created_at'], rep['cnt'], rep['name']])
+    col_width_0 = max(len(str(row[0])) for row in data) + 6
+    col_width_1 = max(len(str(row[1])) for row in data) + 6
+    col_width_2 = max(len(str(row[2])) for row in data) + 6
+    col_width_3 = max(len(str(row[3])) for row in data) + 6
+    for row in data:
```
bari12: Why not use `tabulate` here?

ChristophAmes: I hadn't heard of tabulate before and this was a simple solution that makes the output easy to read.

bari12: Please change it to tabulate. It will produce the same output, but in that case all the commands use the same workflow, which makes it easier to read.

bari12 (approved):

bari12:
Actually I just noticed (now after merging):
@ChristophAmes I think your commits are submitted under @rcarpa eMail/id - I think you might have something misconfigured...

ChristophAmes:
Oh, I didn't catch that. Something must have gone wrong whilst squashing commits.

PR #4373, opened 2021-03-01
Core: Added importlib for compatibility; Fix #4372

### LINKED ISSUE #4372: imp deprecation warning in favour of importlib
Motivation
----------
imp was deprecated long back in favour of importlib with python 3.3. 
Since Python 3.9 will be the last version providing backwards python2 compatibility, these libraries may cause an error. 

long

Modification
------------
Implement dynamic loading using importlib, try to ensure backward compatibility for the time being

dynamic-entropy:
Core: Added importlib for compatibility; Fix #4372
------------------

Used sys.version_info to determine the version of the interpreter in order to use a suitable package for dynamic importing. 
Bypassed use of importlib.find_loader [https://docs.python.org/3/library/importlib.html#importlib.find_loader](url) to use imp, Since py3.3 and 3.4 are backwards compatible with imp.

### DIFF
--- a/lib/rucio/core/transfer.py
+++ b/lib/rucio/core/transfer.py
@@ -39,10 +39,10 @@ from __future__ import division
 
 import copy
 import datetime
-import imp
 import json
 import logging
 import re
+import sys
 import time
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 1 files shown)


bari12:
This kind of code is also called in several other modules, so it should be changed there aswell.
@mlassnig any input on this, since you changed it for the clients in #4363 recently?

mlassnig:
This looks good, but needs a double-check after the merge of #4363 to make sure the imp/importlib situation is correct across the whole codebase.

mlassnig (approved):
LGTM

Issue #2961, opened 2019-10-09
list_replicas for archive


bari12:
Motivation
----------
From Rod (See [JIRA](https://its.cern.ch/jira/browse/RUCIO-2891))

> This file
> data15_13TeV:data15_13TeV.00280673.physics_CosmicCalo.merge.RAW._lb0080._SFO-ALL._0001.1
> has 3 tape replicas, plus in an archive on tape and disk.
> rucio list-file-replicas shows only the replica from the staged zip file on IN2P3-CC_DATADISK
> not the archive on IN2P3-CC_DATATAPE and not the 3 constituent replicas on tape. 
> The see the latter again,
> --no-resolve-archives
> The python client shows all 5 replicas.
> I don`t think it will break prod - just confusing.
> I had another example with no replicas, before adding --no-resolve-archives
> data17_13TeV.00341294.physics_Background.merge.RAW._lb0130._SFO-7._0001.1
> Cheers,
> Rod.

rcarpa:
This is one is probably fixed

Issue #2313, opened 2019-03-12
list_replicas should prioritise root protocol for zips


bari12:
Motivation
----------
Currently `list_replicas` orders replicas based on their protocol priorities. For zip replicas this is not efficient though, as root offers a more efficient way to access the constituent files in the zip.

Modification
------------
`list_replicas` should prioritise root protocol for zip replicas over all other protocols.

PR #4380, opened 2021-03-05
Core: don't filter to 'root' for list_replicas on archives. #2961

rcarpa:
Because of the filter, the 'pfn' field is not filled correctly, so the client doesn't
show the replica.

The change was introduced in #2313, resulting in list_replicas being
broken for any RSE which is not 'root'. The reason of the breakage is
that 'schemes' is a filter which means "all protocols" if empty.
Initializing it to ['root'] restricts the search to only this protocol.
This doesn't seem to be the intended goal of the commit introducing the
regression.
Later in the code, the root protocol is already prioritized if it
exists in the returned 'archive_result'.

### DIFF
--- a/lib/rucio/core/replica.py
+++ b/lib/rucio/core/replica.py
@@ -36,6 +36,7 @@
 # - Patrick Austin <patrick.austin@stfc.ac.uk>, 2020
 # - Eric Vaandering <ewv@fnal.gov>, 2020-2021
 # - Benedikt Ziemons <benedikt.ziemons@cern.ch>, 2020
+# - Radu Carpa <radu.carpa@cern.ch>, 2021
 
 from __future__ import print_function
 
@@ -834,15 +835,6 @@ def _list_replicas(dataset_clause, file_clause, state_clause, show_pfns,
             # if the file is a constituent, find the available archives and add them to the list of possible PFNs
             # taking into account the original rse_expression
             if '%s:%s' % (scope.internal, name) in constituents:
-
-                # special protocol handling for constituents
-                # force the use of root if client didn't specify
-                if not schemes:
-                    schemes = ['root']
-                else:
-                    # always add root for archives
-                    schemes.append('root')
-                    schemes = list(set(schemes))
 
                 archive_result = list_replicas(dids=constituents['%s:%s' % (scope.internal, name)],
                                                schemes=schemes, client_location=client_location,
--- a/lib/rucio/tests/test_archive.py
+++ b/lib/rucio/tests/test_archive.py
@@ -1,4 +1,5 @@
-# Copyright 2017-2020 CERN for the benefit of the ATLAS collaboration.
+# -*- coding: utf-8 -*-
+# Copyright 2017-2021 CERN
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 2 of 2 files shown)


mlassnig:
Hmmm, we can only use the root protocol to read constituents from archives, for everything else it should fail. The other protocols do not support remote constituent extraction. Am I missing something?

rcarpa:
I saw that the download client is capable to download the archive and extract it. Is that function broken ?

mlassnig:
No, that should work, but is sooooo soooo discouraged, that we don't want to use it :-D

rcarpa:
Ok, so the solution is to make it clear in the documentation that download of files from archives is only supported for xroot; and make 'rucio list-file-replicas' correctly list files from archives, but without a pfn ?

bari12:
I am not 100% sure about just forcing `root`.
In principle we should not force only root for archive, just give it top-priority. After all there might be an RSE which does not offer a root door.
Further down after @rcarpa code there is re-prioritization of protocols done, but I don't exactly remember why we basically overwrite this. #2313 also points out re-prioritization, not forcing it. @mlassnig do you remember?

mlassnig:
It had something to do with the pilot, but that's some time ago. I honestly can't remember.

rcarpa:
@rodwalker recently sent an e-mail about list-file-replicas not showing some replicas. And it looks related.

bari12:
Yes this is very likely related.

rodwalker:
Preferring root to retrieve from a zip archive makes sense, but did not need such a change.
Ilija was pushing to choose root protocol on VP sites, regardless or the preferences. I trust this did not get into rucio code, as the correct solution is for the pilot to only ask for root replicas on VP sites, and this was implemented.

rcarpa:
In this case, I believe we should revert the commit from #2313 , as it is a huge hack which is not needed anymore.

bari12:
I don't think we can completely revert it, since the prioritisation wouldn't work then. But the part where we force root, we can/should remove. We should however add some test-cases to make sure this works as expected - I still wonder why we decided for forcing root, instead of the re-prioritisation.

rodwalker:
I was not very clear. I meant preferring root to download from a zip file is needed, but does not need to lead to the other effects.
Such as not showing 1 replica per RSE. Showing root replica when asking for srm.

mlassnig:
I can't remember why I did this.

The pilot always calls list_replicas explicitly with `'schemes': ['srm', 'root', 'davs', 'gsiftp', 'https', 'storm']`, so this must have been something different.

mlassnig:
Ok, I checked a bit more the pilot code. I think I found it.

```
        if use_vp:
            query['schemes'] = ['root']
```

I guess this was not there back then in the pilot, so this might have been a hack in Rucio to make VP work.

rcarpa:
If I understand correctly Rod's message:

> Ilija was pushing to choose root protocol on VP sites, regardless or the preferences. I trust this did not get into rucio code, as the correct solution is for the pilot to only ask for root replicas on VP sites, and this was implemented.

The needed filtering was implemented in pilot since you did the commit. Now we must change things back to not alter the schema list. But we have to keep the code which increases the priority of root.

mlassnig:
So this PR is fine then as-is? I only see the removal of the workaround.

rcarpa:
Yes, this is my understanding, but I need a confirmation. Moreover, I also propose to amend the PR and completely remove that if (not only the else).

mlassnig:
Ok, but double check please with the testcases for constituents of archives.

rcarpa:
I updated the Pull Request. 
However, I discovered that the case of multiple protocols per RSE is not handled correctly because of pre-filtering here :

https://github.com/rucio/rucio/blob/1fa2532c78d3d9fb5701a5f7bc6c7ebfc4fab613/lib/rucio/core/replica.py#L950

First, the protocol with the best priority per RSE is selected; and only afterwards the pfns of different RSEs are ordered. With the current implementation we cannot easily change this behavior for archives without a big refactoring.

rodwalker:
Please don`t break the current behaviour allowing download from a zip file in the pilot.

rcarpa:
Merging this PR has potential to break things without fixing the real problem. After discussion with @bari12 , we decided to close this ticket and open #4473 in order to correctly fix the underlying issues.

Issue #1431, opened 2018-08-08
Handling of Archives in the Reaper


bari12:
Motivation
----------
When deleting the last replica of a file, reaper deletes the file logically as well but does not delete the contents in the archive_contents table. This causes a foreign key constraint error and the reaper gets stuck.
There are three changes needed in total for this.

- [x] #1433 When deleting the file logically, the rows from the archive_contents table need to be deleted too
- [x] The just deleted archive_content rows have to be added to the archive_contents_history table
- [x] For the files whose archive just got deleted, it needs to be checked if they do not have any other replicas, in which case they need to be removed logically (If there are no rules on them, which is actually a bit pointless)
- [x] #2213 When the last replica of a file is removed in the reaper, it should only be removed from the dataset if it is not a constituent of an existing archive replica. This also implies, if a archive replica is removed and there are no replicas on the constituents, the constituents need to be removed from their parent datasets.

PR #4381, opened 2021-03-05
Testing: Reorganize reaper2 tests #1431

rcarpa:
Split the existing test into two: single vo; and multi vo. Skip the
later in single-vo environments.

The existing test was difficult to follow and understand. I have a
strong feeling that some parts of the tests were not working as
intended. For example: always using rse_names[0] and all_file_names[0]
(and never the second elements of the arrays) in the multi-vo code path
doesn't seem right. Also, exclude_rses was provided an empty list,
which suggests that the author expected include_rses to also take a
list. This is not the case: both take an rse expression string.

Move the relevant multi_vo tests to tests_reaper2.py and merge
duplicate code.

### DIFF
--- a/lib/rucio/tests/test_multi_vo.py
+++ b/lib/rucio/tests/test_multi_vo.py
@@ -19,10 +19,10 @@
 # - Benedikt Ziemons <benedikt.ziemons@cern.ch>, 2020-2021
 # - Mario Lassnig <mario.lassnig@cern.ch>, 2020
 # - Martin Barisits <martin.barisits@cern.ch>, 2020-2021
+# - Radu Carpa <radu.carpa@cern.ch>, 2021
 
 import os
 import unittest
-from datetime import datetime, timedelta
 from logging import getLogger
 from os import remove
 from random import choice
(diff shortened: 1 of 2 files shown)


bziemons:
I am all for cleaning up :+1:
Can you check if the tests in https://github.com/rucio/rucio/blob/e049b184b55fa4b1304db04d5e88d945d82969e8/lib/rucio/tests/test_multi_vo.py#L1092-L1210 can be included here?

rcarpa:
Didn't know about these ones. I see that one of the tests does almost the same thing as the test multi-vo one in test_reaper.py.
Should I move/merge the reaper tests from `test_multi_vo.py` into `test_reaper.py` and skip them in single-vo tests?

bziemons:
> Didn't know about these ones. I see that one of the tests does almost the same thing as the test multi-vo one in test_reaper.py.
> Should I move/merge the reaper tests from `test_multi_vo.py` into `test_reaper.py` and skip them in single-vo tests?

sounds like the right thing to do, yes

bari12:
👍 I think that would be good. For MultiVO functionality we do have tests in all the individual test_* files anyway. They are not concentrated to test_multi_vo.py.

bziemons:
Please rebase on master.

bari12 (approved):

Issue #4221, opened 2021-01-07
Identify and cleanup unused functionality and code


bari12:
Motivation
----------
Over the years we produced code and functionality which is not used anymore. These code parts and functionality should be identified and eventually cleaned up.

For unused functionality, please document this in this ticket first, so we can discuss if removal of that functionality is appropriate or should be kept. For more simple unused code cases, just submit a PR.

Modification
------------
All Component leaders are responsible to do this for their components:

- [ ] Authentication & Authorization
- [ ] Clients
  - Removal of DQ2 clients
- [ ] Consistency checks
- [ ] Core & Internals
  - Removal of the extensions/forecast module (T3CModel) and also numpy as a dependency.
  - Temporary dids
- [ ] Data Pre-Placement
- [ ] Database
- [ ] Dataset deletion
- [ ] Deletion
  - `MaxBeingDeletedFiles` is unused in reaper2
- [x] DIRAC
- [ ] Docker & Kubernetes
- [ ] Documentation
- [ ] Functional tests
- [ ] Life time model
- [ ] Messaging
- [ ] Monitoring & Logging
- [ ] Multi VO
- [ ] Policies
- [ ] Probes & Alarms
- [ ] Protocols
- [ ] Rebalancing
- [ ] Recovery
- [ ] Release management
- [ ] Replicas
- [ ] REST & API
  - Temporary dids endpoint
- [ ] Rules
- [ ] Subscriptions
- [ ] Testing
- [ ] Traces
- [ ] Transfers
- [ ] WebUI

Issue #3311, opened 2020-02-19
Reaper seems to skip the deletion of the collection replica


cserf:
Motivation
----------
The reaper seems to skip the deletion of the collection replica when the last replica of a dataset is deleted on a RSE

PR #4383, opened 2021-03-05
Core: reorganize delete_replicas #1431 #4221

rcarpa:
Refactor the delete_replicas function
- reorganize the very long lines of queries to make them easier to read
- add comments
- split the function in two

### DIFF
--- a/lib/rucio/core/replica.py
+++ b/lib/rucio/core/replica.py
@@ -36,6 +36,7 @@
 # - Patrick Austin <patrick.austin@stfc.ac.uk>, 2020
 # - Eric Vaandering <ewv@fnal.gov>, 2020-2021
 # - Benedikt Ziemons <benedikt.ziemons@cern.ch>, 2020
+# - Radu Carpa <radu.carpa@cern.ch>, 2021
 
 from __future__ import print_function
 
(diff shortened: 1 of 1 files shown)


lib/rucio/core/replica.py
```diff
@@ -1462,63 +1463,108 @@ def delete_replicas(rse_id, files, ignore_availability=True, session=None):
...
         raise exception.ReplicaNotFound("One or several replicas don't exist.")
 
+    __update_metadata_after_replica_deletion(rse_id=rse_id, files=files, session=session)
+
+    # Decrease RSE counter
+    decrease(rse_id=rse_id, files=delta, bytes=bytes, session=session)
+
+
+@transactional_session
+def __update_metadata_after_replica_deletion(rse_id, files, session=None):
```
bari12: I would call this `__cleanup_after_replica_deletion`
`metadata` has a bit of a double-loaded meaning in Rucio. I think cleanup is cleaner 😄

rcarpa: Done

lib/rucio/core/replica.py
```diff
@@ -1544,21 +1590,47 @@ def delete_replicas(rse_id, files, ignore_availability=True, session=None):
...
+                # (it was the last children), metadata cleanup has to be done:
+                #
+                # 1) Schedule to remove the replicas of this empty collection
+                clt_replica_condition.append(
+                    and_(models.CollectionReplica.scope == parent_scope,
+                         models.CollectionReplica.name == parent_name,
+                         exists(select([1]).prefix_with("/*+ INDEX(DIDS DIDS_PK) */", dialect='oracle')).where(
+                             and_(models.DataIdentifier.scope == parent_scope,
+                                  models.DataIdentifier.name == parent_name,
+                                  models.DataIdentifier.is_open == False)),  # NOQA
```
cserf, line 1610: Does anyone remember why we have this ? I suspect this is the cause of this bug: https://github.com/rucio/rucio/issues/3311

cserf (commented):
That looks good, but I think I found a bug (not related to your code, but in the previous code)

lib/rucio/core/replica.py
bari12: I can't remember, but thinking about this, it is probably wrong and should not filter on `is_open == False`.

bari12: Let's continue discussing this in #3311
I think @rcarpa can have a look at this one too during his reaper2 works.

bari12 (approved):

PR #4388, opened 2021-03-09
adding xcache prefix determination through VP mechanism

ivukotic:
Now multinode caches are possible.  VP service has data on all caches, servers and their sizes. Based on that it can calculate the best servers.
Rucio will once per minute refresh info on xcache sites and servers. Rucio will do the final (very quick) determination of which server to use. 

This PR adds one more unit test:
test_replica.py::TestReplicaCore::test_cache_replicas

for the test to pass one needs to run: 
rucio-admin config set --section clientcachemap --option BLACKMESA --value AGLT2


lib/rucio/core/replica.py
```diff
@@ -28,7 +28,7 @@
 # - James Perry <j.perry@epcc.ed.ac.uk>, 2019
 # - Jaroslav Guenther <jaroslav.guenther@cern.ch>, 2019
 # - Andrew Lister <andrew.lister@stfc.ac.uk>, 2019
-# - Ilija Vukotic <ivukotic@cern.ch>, 2020
+# - Ilija Vukotic <ivukotic@cern.ch>, 2020, 2021
```
bziemons: 2020-2021

lib/rucio/core/replica.py
```diff
@@ -644,16 +648,16 @@ def _resolve_dids(dids, unavailable, ignore_availability, all_states, resolve_ar
...
                 archive = session.query(models.ConstituentAssociation.scope,
                                         models.ConstituentAssociation.name)\
-                                 .filter(models.ConstituentAssociation.child_scope == scope,
-                                         models.ConstituentAssociation.child_name == name)\
-                                 .with_hint(models.ConstituentAssociation, "INDEX(ARCHIVE_CONTENTS ARCH_CONTENTS_PK)", 'oracle')\
-                                 .all()
+                    .filter(models.ConstituentAssociation.child_scope == scope,
+                            models.ConstituentAssociation.child_name == name)\
+                    .with_hint(models.ConstituentAssociation, "INDEX(ARCHIVE_CONTENTS ARCH_CONTENTS_PK)", 'oracle')\
+                    .all()
```
bziemons: please revert unrelated changes

lib/rucio/core/replica.py
```diff
@@ -799,6 +803,37 @@ def _list_replicas_for_files(file_clause, state_clause, files, rse_clause, updat
...
+        # print('reloading xcache sites and ranges.')
+        vp_endpoint = 'https://vps.cern.ch/serverRanges'
+        response = requests.get(vp_endpoint, verify=False)
+        if response:
+            # print('Success! Got XCache ranges.')
+            x_caches = response.json()
+            # print(x_caches)
+            write_to_cache('CacheSites', x_caches)
+        else:
+            print('Could not reload from vps.')
```
bziemons: use a logger instead of print, see https://github.com/rucio/rucio/commit/e433c472abcc54536d111371466f6af934c3cede#diff-1085ef0a38068abd1b394d8a69b43fa31b34a3a3eb1686fc3727f3638853c7cdR558 on how this is done in core now.

lib/rucio/core/replica.py
```diff
@@ -2242,14 +2286,14 @@ def get_source_replicas_for_dataset(scope, name, source_rses=None,
...
-                               .filter(models.DataIdentifierAssociation.scope == scope,
-                                       models.DataIdentifierAssociation.name == name)
+                    .with_hint(models.DataIdentifierAssociation, "INDEX_RS_ASC(CONTENTS CONTENTS_PK) NO_INDEX_FFS(CONTENTS CONTENTS_PK)", 'oracle')\
+                    .outerjoin(models.RSEFileAssociation,
+                               and_(models.DataIdentifierAssociation.child_scope == models.RSEFileAssociation.scope,
+                                    models.DataIdentifierAssociation.child_name == models.RSEFileAssociation.name,
+                                    models.RSEFileAssociation.state == ReplicaState.AVAILABLE,
+                                    or_(*rse_clause)))\
+                    .filter(models.DataIdentifierAssociation.scope == scope,
+                            models.DataIdentifierAssociation.name == name)
```
bziemons: revert these and other unrelated changes

lib/rucio/core/replica.py
```diff
@@ -799,6 +803,37 @@ def _list_replicas_for_files(file_clause, state_clause, files, rse_clause, updat
             {'scope': scope, 'name': name} in files and files.remove({'scope': scope, 'name': name})
 
 
+def get_multi_cache_prefix(cache_site, filename):
+
+    # print('Looking up prefix for cache:', cache_site, 'and filename:', filename)
+    x_caches = read_from_cache('CacheSites', expiration_time=60)
+    if not x_caches:
+        # print('reloading xcache sites and ranges.')
+        vp_endpoint = 'https://vps.cern.ch/serverRanges'
```
bziemons: this would run any test of this functionality against a live instance? Where is the documentation on this endpoint located? This will probably not work for organizations using Rucio outside of CERN.

bziemons (changes requested):

lib/rucio/core/replica.py
ivukotic: Yes. Once we have all of VP integrated in Rucio this won't be needed. But for now even someone testing from outside won't see any issues. btw... this is not the only reference to it... there is one in line 2592..

lib/rucio/core/replica.py
ivukotic: OK

lib/rucio/core/replica.py
ivukotic: I'll try... it's hard when yours and mine autoformat differ.

lib/rucio/core/replica.py
ivukotic: I will try.

lib/rucio/core/replica.py
```diff
@@ -799,6 +811,37 @@ def _list_replicas_for_files(file_clause, state_clause, files, rse_clause, updat
             {'scope': scope, 'name': name} in files and files.remove({'scope': scope, 'name': name})
 
 
+def get_multi_cache_prefix(cache_site, filename, logger=logging.log):
+
+    # print('Looking up prefix for cache:', cache_site, 'and filename:', filename)
+    x_caches = read_from_cache('CacheSites', expiration_time=60)
```
bziemons: please cache the error case where the http response is not successful, too. This would spam the log and the vp endpoint server address otherwise.

lib/rucio/core/replica.py
```diff
@@ -799,6 +811,37 @@ def _list_replicas_for_files(file_clause, state_clause, files, rse_clause, updat
...
+        # print('reloading xcache sites and ranges.')
+        vp_endpoint = 'https://vps.cern.ch/serverRanges'
+        response = requests.get(vp_endpoint, verify=False)
+        if response:
+            # print('Success! Got XCache ranges.')
+            x_caches = response.json()
+            # print(x_caches)
+            write_to_cache('CacheSites', x_caches)
+        else:
+            logger(logging.ERROR, 'Could not reload from vps.')
```
bziemons: Please make this error more useful. Include this method namely and e. g. the requested http address and the status code.

lib/rucio/core/replica.py
bziemons: Please have this configurable anyway (also the other reference) and disable requesting http if the configuration value is not set as it is done in other parts of Rucio.

lib/rucio/core/replica.py
```diff
@@ -72,6 +77,13 @@
                                    DEFAULT_SCHEMA_NAME, BASE)
 from rucio.rse import rsemanager as rsemgr
 
+logging.basicConfig(stream=sys.stdout,
+                    level=getattr(logging,
+                                  config_get('common', 'loglevel',
+                                             raise_exception=False,
+                                             default='ERROR').upper()),
+                    format='%(asctime)s\t%(process)d\t%(levelname)s\t%(message)s')
+
```
bziemons: This must be removed. `setup_logging` from `rucio.common.logging` is run at entrypoints.

bziemons (changes requested):
I am guessing the commented print statements are there, because this is all 'temporary'. It would be nice if you could remove them and instead add non-spammy and useful debug messages.

lib/rucio/core/replica.py
```diff
@@ -61,6 +65,7 @@
 from rucio.common.types import InternalScope
 from rucio.common.utils import chunks, clean_surls, str_to_date, add_url_query
 from rucio.core.config import get as config_get
+from rucio.core.config import read_from_cache, write_to_cache
```
bari12: It's only a minor detail, but it would be better not to use these methods. (They should not be public, but internal to `core.config` anyway. 
The issue with this is that you are basically using the cache region of the config module. In principle this works, but it may have some side-effects. It would be better to establish a REGION directly in core/replica.

lib/rucio/core/replica.py
bari12: Yes, just put it in the config table aswell. 
E.g.
```
[virtualplacement]
vp_endpoint= https://vps.cern.ch/serverRanges
```

lib/rucio/core/replica.py
ivukotic: agree. will fix

lib/rucio/core/replica.py
ivukotic: Agree

lib/rucio/core/replica.py
ivukotic: agree

lib/rucio/core/replica.py
ivukotic: OK. Fixed.

lib/rucio/core/replica.py
```diff
@@ -799,6 +808,48 @@ def _list_replicas_for_files(file_clause, state_clause, files, rse_clause, updat
             {'scope': scope, 'name': name} in files and files.remove({'scope': scope, 'name': name})
 
 
+def get_vp_endpoint():
+    vp_endpoint = config_get('virtual_placement', 'vp_endpoint', default='')
+    print('vp endpoint: ', vp_endpoint)
```
bziemons: leftover?

lib/rucio/core/replica.py
```diff
@@ -799,6 +808,48 @@ def _list_replicas_for_files(file_clause, state_clause, files, rse_clause, updat
...
+    if x_caches is NO_VALUE:
+        logger(logging.DEBUG, 'reloading xcache sites and ranges.')
+        try:
+            response = requests.get('{}/serverRanges'.format(vp_endpoint), verify=False)
+            if response.status_code == 200:
+                x_caches = response.json()
+                logger(logging.DEBUG, x_caches)
+                REGION.set('CacheSites', x_caches)
+            else:
+                return ''
```
bziemons, line 831: again, this will not cache any http errors. Please at least run `response.raise_for_status()` after `response = requests.get(…)`

lib/rucio/core/replica.py
```diff
@@ -799,6 +808,48 @@ def _list_replicas_for_files(file_clause, state_clause, files, rse_clause, updat
...
+
+    # print('Looking up prefix for cache:', cache_site, 'and filename:', filename)
+    x_caches = REGION.get('CacheSites')
+    if x_caches is NO_VALUE:
+        logger(logging.DEBUG, 'reloading xcache sites and ranges.')
+        try:
+            response = requests.get('{}/serverRanges'.format(vp_endpoint), verify=False)
+            if response.status_code == 200:
+                x_caches = response.json()
+                logger(logging.DEBUG, x_caches)
```
bziemons: this would just dump some object to the log and when a server has load this will not directly be after `reloading xcache sites and ranges.`. This does not seem to be useful.

lib/rucio/core/replica.py
```diff
@@ -799,6 +808,48 @@ def _list_replicas_for_files(file_clause, state_clause, files, rse_clause, updat
...
+                REGION.set('CacheSites', x_caches)
+            else:
+                return ''
+        except requests.exceptions.RequestException as re:
+            REGION.set('CacheSites', {'could not reload': ''})
+            logger(logging.ERROR, 'In get_multi_cache_prefix, could not access {}. Error:{}'.format(vp_endpoint, re))
+            return ''
+
+    if cache_site not in x_caches:
+        logger(logging.DEBUG, 'cache site: ' + cache_site + ' not active.')
```
bziemons: is this actually a case which is useful when debugging a problem or does this happen often?

lib/rucio/core/replica.py
```diff
@@ -799,6 +808,48 @@ def _list_replicas_for_files(file_clause, state_clause, files, rse_clause, updat
...
+def get_multi_cache_prefix(cache_site, filename, logger=logging.log):
+
+    vp_endpoint = get_vp_endpoint()
+    if not vp_endpoint:
+        return ''
+
+    # print('Looking up prefix for cache:', cache_site, 'and filename:', filename)
+    x_caches = REGION.get('CacheSites')
+    if x_caches is NO_VALUE:
+        logger(logging.DEBUG, 'reloading xcache sites and ranges.')
```
bziemons: this is not useful for debugging and just indicates normal operation

lib/rucio/core/replica.py
```diff
@@ -799,6 +808,48 @@ def _list_replicas_for_files(file_clause, state_clause, files, rse_clause, updat
...
+            response = requests.get('{}/serverRanges'.format(vp_endpoint), verify=False)
+            if response.status_code == 200:
+                x_caches = response.json()
+                logger(logging.DEBUG, x_caches)
+                REGION.set('CacheSites', x_caches)
+            else:
+                return ''
+        except requests.exceptions.RequestException as re:
+            REGION.set('CacheSites', {'could not reload': ''})
+            logger(logging.ERROR, 'In get_multi_cache_prefix, could not access {}. Error:{}'.format(vp_endpoint, re))
```
bziemons: maybe downgrade this to warning, since the server operation does not seem to be disturbed

lib/rucio/core/replica.py
```diff
@@ -39,13 +39,17 @@
...
 
+import logging
 from collections import defaultdict
 from copy import deepcopy
 from curses.ascii import isprint
 from datetime import datetime, timedelta
 from json import dumps
 from re import match
 from traceback import format_exc
+import requests
```
bziemons, line 49: requests is currently not a server dependency

lib/rucio/tests/test_replica.py
```diff
@@ -83,6 +85,57 @@ def setUp(self):
...
+
+        name = 'element_%s' % generate_uuid()
+        hstr = hashlib.md5(('%s:%s' % (tmp_scope, name)).encode('utf-8')).hexdigest()
+        pfn = 'http://root.aperture.com:1409//test/chamber/mock/%s/%s/%s' % (hstr[0:2], hstr[2:4], name)
+        files.append({'scope': tmp_scope, 'name': name, 'bytes': 1234, 'adler32': 'deadbeef', 'pfn': pfn})
+
+        add_replicas(rse_id=rse_id, files=files, account=root)
+
+        cconfig_set('clientcachemap', 'BLACKMESA', 'AGLT2')
+        cconfig_set('virtual_placement', 'vp_endpoint', 'https://vps.cern.ch')
```
bziemons, line 129: again, mock this for the tests. This may or may not fail when the server is not available. You could mock/patch the requests.get method for testing `get_multi_cache_prefix`. It would also be nice to have tests for error cases.

bziemons (changes requested):

lib/rucio/core/replica.py
bari12: It is (loaded via the `pip-requires-client` file 🙄 ...)
But please put the imports still in alphabetical order

lib/rucio/core/replica.py
bari12: I would probably remove this. In theory it should not happen too often, but when it happens it is really quite spammy.

lib/rucio/tests/test_replica.py
bari12: Yes, I agree. Although this is a prototype feature for now, it would be better to mock this right away. The vps server might be unavailable, or it might not be reachable from the GH test farm etc. and then our tests fail because of it. Better to mock it right away.

lib/rucio/core/replica.py
bziemons: whoops, my bad. I thought that was only for the clients.

lib/rucio/core/replica.py
ivukotic: yes. will remove.

lib/rucio/core/replica.py
ivukotic: fixed.

lib/rucio/core/replica.py
ivukotic: it could happen that a client was pointing to a cache that had only one cache node and that one went dead. Or have it wrongly configured. In that case it would be spammy. So I removed the printout.

lib/rucio/core/replica.py
ivukotic: true. removed.

lib/rucio/core/replica.py
ivukotic: true. now WARNING

lib/rucio/core/replica.py
ivukotic: OK. removed

lib/rucio/core/replica.py
ivukotic: now in alphabetical order

lib/rucio/tests/test_replica.py
ivukotic: this will fail if the server is not available. it must return a prefix which is not '', that is what is looked for in the assert.

lib/rucio/tests/test_replica.py
```diff
@@ -83,6 +85,57 @@ def setUp(self):
...
+
+        for rep in list_replicas(
+                dids=[{'scope': f['scope'], 'name': f['name'], 'type': DIDType.FILE} for f in files],
+                schemes=['root'],
+                domain='wan',
+                client_location={'site': 'BLACKMESA'}):
+            print('Got replica!', rep['pfns'].keys())
```
bari12: Remove print

bari12 (changes requested):
One minor change (Remove print)

To me it looks good now. I think it would be good to have the mocking of the VPS service in the tests, but I am ok with skipping this until we either move it into production or remove it from the codebase anyway.

lib/rucio/tests/test_replica.py
ivukotic: done.

bari12:
Also needs a rebase (Forgot to mention earlier) 😄

bziemons:
I wouldn't merge this as is. If there is a problem with the domain in the test, it will fail like this:
```
_____________________ TestReplicaCore.test_cache_replicas ______________________
[gw1] linux -- Python 3.6.8 /usr/bin/python
test_replica.py:136: in test_cache_replicas
    assert list(rep['pfns'].keys())[0].count('root://') == 2
E   AssertionError: assert 1 == 2
E     +1
E     -2
```

My suggestion would be to either fix it before merging this PR or (less preferably) mark the test with xfail.

ivukotic:
> I wouldn't merge this as is. If there is a problem with the domain in the test, it will fail like this:
> 
> ```
> _____________________ TestReplicaCore.test_cache_replicas ______________________
> [gw1] linux -- Python 3.6.8 /usr/bin/python
> test_replica.py:136: in test_cache_replicas
>     assert list(rep['pfns'].keys())[0].count('root://') == 2
> E   AssertionError: assert 1 == 2
> E     +1
> E     -2
> ```


I am not sure I understand this comment. what domain are we talking about here?

bziemons:
> I am not sure I understand this comment. what domain are we talking about here?

I am talking about the only review comment left (which I unresolved). This line:

https://github.com/rucio/rucio/pull/4388/files#diff-e850f539a4f5999c303caa50ac07cfdb3ca9c21d9b7b0aa817fda925102701b5R129

bari12:
The problem with the test is, if the VPS service fails/is not reachable/github blocks it, etc. our Rucio tests, for unrelated PRs will fail too, for quite unclear reasons to the developer. That's why Ben said a mock service would be ideal for the test, so you don't have to connect to anything outside of the test container.
But as I wrote above, it is fine for me if we add this mock test later, since right now this is just a prototype we want to try. But to circumvent any possible issues, marking it as `xfail` is probably best.

ivukotic:
> One minor change (Remove print)
> 
> To me it looks good now. I think it would be good to have the mocking of the VPS service in the tests, but I am ok with skipping this until we either move it into production or remove it from the codebase anyway.

I have added a mock VPS server. Now it passes the test.

ivukotic:
mistake ...

bari12:
Work continues in #4469
Closing this

Issue #4335, opened 2021-02-15
Revisit support for download to multiple destination in downloadclient


rcarpa:
Motivation
----------

The rucio downloadclient (used by the rucio download command, but also by [pilot](https://github.com/search?q=org%3APanDAWMS+download_dids&type=code) ) has internal support for downloading to multiple destination folders. Which basically means downloading to one destination, followed by performing local copies between destinations. This support is implemented in many of the private methods. However, 3 out 4 public members of the downloadclient class don't allow to use this behavior at all : 

`download_pfns` sets destination internally to a [single path](https://github.com/rucio/rucio/blob/326f6eca8473bc05318e4b4ca13fb277c94bcc5d/lib/rucio/client/downloadclient.py#L226)
`download_dids` and `download_aria2c` both use a helper function which does [a similar thing](https://github.com/rucio/rucio/blob/2a9f137b0615448f785a83b36c189be0803f20de/lib/rucio/client/downloadclient.py#L1012). Moreover, the aria2c related functions don't implement sync between folders. So the behavior would not be consistent even if accessible via public methods. 
`download_from_metalink_file` does allow to pass an existing ['destinations' list](https://github.com/rucio/rucio/blob/2a9f137b0615448f785a83b36c189be0803f20de/lib/rucio/client/downloadclient.py#L323) but it looks like an undocumented unwanted side-effect/bug. 

Modification
------------
Decide if we want to maintain this logic. If yes, it must be made available via a public interface and tested. If not, this unused code must be removed.

rcarpa:
This function can actually be used, but indirectly by download_dids. It happens when requesting to download the same did twice to different base_dir. It is useful when two wildcard dids (which overlap) are requested. In this case, the overlapping pfns will only be downloaded once and then synced between directories. I'll create a test and verify if this function is also needed for aria2c downloads.

PR #4390, opened 2021-03-09
Testing: add test for downloading to multiple paths. #4335

rcarpa:
Whenever the same did is resolved by two items in download_did,
the downloadclient will download the file only once and then sync it
between folders.

Add a test to ensure that the file ends downloaded to both paths.


bari12 (approved):
Merging as a `patch`

PR #4395, opened 2021-03-10
Transfers: Fix for transfertool filtering with conventional submissio…

bari12:


bziemons:
you want to change line 724 in `lib/rucio/core/transfer.py`

bziemons (approved):

PR #4398, opened 2021-03-11
Monitoring & Logging: Add two probes into abacus-rse instead Fix #4261

### LINKED ISSUE #4261: Move deletable and obsolete replica checks from probes into Abacus-RSE
Motivation
----------

Modification
------------

ericvaandering:
This adds the functionality of two probes into abacus-rse. There was discussion that other things could be added?

### DIFF
--- a/lib/rucio/core/rse_counter.py
+++ b/lib/rucio/core/rse_counter.py
@@ -11,12 +11,17 @@
 # - Martin Barisits, <martin.barisits@cern.ch>, 2014
 # - Hannes Hansen, <hannes.jakob.hansen@cern.ch>, 2018-2019
 # - Brandon White, <bjwhite@fnal.gov>, 2019
+# - Eric Vaandering <ewv@fnal.gov>, 2021
 
-from sqlalchemy.orm.exc import NoResultFound
+import datetime
 
 from rucio.common.exception import CounterNotFound
 from rucio.db.sqla import models, filter_thread_work
+from rucio.db.sqla.constants import OBSOLETE
 from rucio.db.sqla.session import read_session, transactional_session
+from sqlalchemy import func
+from sqlalchemy.orm.exc import NoResultFound
+from sqlalchemy.sql.expression import null
 
 
 @transactional_session
@@ -27,8 +32,7 @@ def add_counter(rse_id, session=None):
     :param rse_id:  The id of the RSE.
     :param session: The database session in use.
     """
-    models.RSEUsage(rse_id=rse_id, source='rucio', files=0, used=0).\
-        save(session=session)
+    models.RSEUsage(rse_id=rse_id, source='rucio', files=0, used=0).save(session=session)
 
 
 @transactional_session
@@ -41,8 +45,7 @@ def increase(rse_id, files, bytes, session=None):
     :param bytes:   The number of added bytes.
     :param session: The database session in use.
     """
-    models.UpdatedRSECounter(rse_id=rse_id, files=files, bytes=bytes).\
-        save(session=session)
+    models.UpdatedRSECounter(rse_id=rse_id, files=files, bytes=bytes).save(session=session)
 
 
 @transactional_session
@@ -67,8 +70,7 @@ def del_counter(rse_id, session=None):
     :param session: The database session in use.
     """
 
-    session.query(models.RSEUsage).filter_by(rse_id=rse_id, source='rucio').\
-        delete(synchronize_session=False)
+    session.query(models.RSEUsage).filter_by(rse_id=rse_id, source='rucio').delete(synchronize_session=False)
 
 
 @read_session
@@ -84,8 +86,7 @@ def get_counter(rse_id, session=None):
     """
 
     try:
-        counter = session.query(models.RSEUsage).\
-            filter_by(rse_id=rse_id, source='rucio').one()
+        counter = session.query(models.RSEUsage).filter_by(rse_id=rse_id, source='rucio').one()
         return {'bytes': counter.used,
                 'files': counter.files,
                 'updated_at': counter.updated_at}
@@ -103,10 +104,10 @@ def get_updated_rse_counters(total_workers, worker_number, session=None):
     :param session:            Database session in use.
     :returns:                  List of rse_ids whose rse_counters need to be updated.
     """
-    query = session.query(models.UpdatedRSECounter.rse_id).\
-        distinct(models.UpdatedRSECounter.rse_id)
+    query = session.query(models.UpdatedRSECounter.rse_id).distinct(models.UpdatedRSECounter.rse_id)
 
-    query = filter_thread_work(session=session, query=query, total_threads=total_workers, thread_id=worker_number, hash_variable='rse_id')
+    query = filter_thread_work(session=session, query=query, total_threads=total_workers, thread_id=worker_number,
+                               hash_variable='rse_id')
     results = query.all()
     return [result.rse_id for result in results]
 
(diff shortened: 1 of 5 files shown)


lib/rucio/core/rse_counter.py
```diff
@@ -27,7 +30,7 @@ def add_counter(rse_id, session=None):
     :param rse_id:  The id of the RSE.
     :param session: The database session in use.
     """
-    models.RSEUsage(rse_id=rse_id, source='rucio', files=0, used=0).\
+    models.RSEUsage(rse_id=rse_id, source='rucio', files=0, used=0). \
```
bziemons: if you want to format these lines, please move the dot to the next line and the escaping backslash wont be needed anymore.

lib/rucio/core/rse_counter.py
```diff
@@ -12,11 +12,14 @@
 # - Hannes Hansen, <hannes.jakob.hansen@cern.ch>, 2018-2019
 # - Brandon White, <bjwhite@fnal.gov>, 2019
```
bziemons, line 15: please update the copyright either by using `tools/add_header -i <filename>` or manually

lib/rucio/core/rse_counter.py
```diff
@@ -138,6 +142,49 @@ def update_rse_counter(rse_id, session=None):
...
+        obsolete_files = 0
+
+    # Merge (INSERT or UPDATE) the values
+    deletable = models.RSEUsage(rse_id=rse_id, used=deletable_bytes, files=deletable_files, source='expired')
+    session.merge(deletable)
+    obsolete = models.RSEUsage(rse_id=rse_id, used=obsolete_bytes, files=obsolete_files, source='obsolete')
+    session.merge(obsolete)
+
+    # fill_rse_counter_history_table takes care of the history for all types
+    return
```
bziemons: remove unnecessary return

lib/rucio/core/rse_counter.py
```diff
@@ -147,4 +194,5 @@ def fill_rse_counter_history_table(session=None):
     """
     RSEUsageHistory = models.RSEUsage.__history_mapper__.class_
     for usage in session.query(models.RSEUsage).all():
-        RSEUsageHistory(rse_id=usage['rse_id'], used=usage['used'], files=usage['files'], source=usage['source']).save(session=session)
+        RSEUsageHistory(rse_id=usage['rse_id'], used=usage['used'], files=usage['files'], source=usage['source']).save(
+            session=session)
```
bziemons: please don't break a line here, this does not make it more readable

lib/rucio/daemons/abacus/rse.py
```diff
@@ -71,15 +72,23 @@ def rse_update(once=False):
...
                         break
                     start_time = time.time()
                     update_rse_counter(rse_id=rse_id)
-                    logging.debug('rse_update[%s/%s]: update of rse "%s" took %f' % (heartbeat['assign_thread'], heartbeat['nr_threads'] - 1, rse_id, time.time() - start_time))
+                    logging.debug('rse_update[%s/%s]: counter update of rse "%s" took %f' %
+                                  (heartbeat['assign_thread'], heartbeat['nr_threads'] - 1, rse_id,
+                                   time.time() - start_time))
+                    start_time = time.time()
+                    update_replica_counts(rse_id=rse_id)
+                    logging.debug('rse_update[%s/%s]: replica update of rse "%s" took %f' %
```
bziemons: if you're adding new log messages anyway, could you please create a `logger` after fetching the heartbeat in line 65 and use that?

lib/rucio/core/rse_counter.py
bziemons: also please update the copyright in the other files

lib/rucio/daemons/abacus/rse.py
```diff
@@ -37,7 +37,8 @@
 from rucio.common.logging import setup_logging
 from rucio.common.utils import get_thread_with_periodic_running_function
 from rucio.core.heartbeat import live, die, sanity_check
-from rucio.core.rse_counter import get_updated_rse_counters, update_rse_counter, fill_rse_counter_history_table
+from rucio.core.rse_counter import (fill_rse_counter_history_table, get_updated_rse_counters,
+                                    update_replica_counts, update_rse_counter)
```
bziemons, line 42: I remember there was a discussion about this notation already. Is this just autoformatting?

lib/rucio/core/rse_counter.py
```diff
@@ -138,6 +142,49 @@ def update_rse_counter(rse_id, session=None):
...
+    """
+    none_value = None
+
+    # Get the counds of obsolete and deletably (expired) files
+    obsolete_files, obsolete_bytes = (session
+                                      .query(func.count(), func.sum(models.RSEFileAssociation.bytes))
+                                      .with_hint(models.RSEFileAssociation,
+                                                 "INDEX_FFS(REPLICAS REPLICAS_TOMBSTONE_IDX)", 'oracle')
+                                      .filter(models.RSEFileAssociation.rse_id == rse_id,
+                                              models.RSEFileAssociation.tombstone != none_value,
```
bziemons: please use `null()` from sqlalchemy instead

lib/rucio/core/rse_counter.py
```diff
@@ -138,6 +142,49 @@ def update_rse_counter(rse_id, session=None):
...
+    Read the replica counts and update rse_usage.
+
+    :param rse_id:   The rse_id to update.
+    :param session:  Database session in use.
+    """
+    none_value = None
+
+    # Get the counds of obsolete and deletably (expired) files
+    obsolete_files, obsolete_bytes = (session
+                                      .query(func.count(), func.sum(models.RSEFileAssociation.bytes))
```
bziemons, line 153: is there a way to make these blocks a bit better readable?

bziemons (changes requested):
my comments are mostly cosmetics, functionality-wise I don't see a problem. I am not sure how `RSEUsage` is updated so far. This would override any values set by other means and is not configurable.

lib/rucio/core/rse_counter.py
ericvaandering: Not sure what's unreadable about it. Using () to enclose multiple lines is now preferred and it clearly separates the clauses. Indentation is determined by PyCharm

lib/rucio/core/rse_counter.py
ericvaandering: These were all short enough to combine on a single line.

ericvaandering:
I think the only test which didn't work was the merging test. But that should be solved now. Any comments on the general approach of where this is located?

bziemons:
> I think the only test which didn't work was the merging test. But that should be solved now. Any comments on the general approach of where this is located?

you're not following the [branch naming guidelines](https://github.com/rucio/rucio/blob/master/CONTRIBUTING.rst#contributing). Branches prefixed with `feature-` will not be cherry-picked to the latest release branch. The release-patch-setup fails if the cherry-picking fails and master and the latest release branch diversed too much for your changes.

lib/rucio/daemons/abacus/rse.py
```diff
@@ -62,32 +64,43 @@ def rse_update(once=False):
...
             start = time.time()  # NOQA
             rse_ids = get_updated_rse_counters(total_workers=heartbeat['nr_threads'],
                                                worker_number=heartbeat['assign_thread'])
-            logging.debug('Index query time %f size=%d' % (time.time() - start, len(rse_ids)))
+            logger(logging.DEBUG, 'Index query time %f size=%d' % (time.time() - start, len(rse_ids)))
 
             # If the list is empty, sent the worker to sleep
             if not rse_ids and not once:
-                logging.info('rse_update[%s/%s] did not get any work' % (heartbeat['assign_thread'], heartbeat['nr_threads'] - 1))
+                logger(logging.INFO, 'rse_update[%s/%s] did not get any work' %
```
bziemons, line 79: double prefix

lib/rucio/daemons/abacus/rse.py
```diff
@@ -62,32 +64,43 @@ def rse_update(once=False):
...
+                       (heartbeat['assign_thread'], heartbeat['nr_threads'] - 1))
                 time.sleep(10)
             else:
                 for rse_id in rse_ids:
                     if graceful_stop.is_set():
                         break
                     start_time = time.time()
                     update_rse_counter(rse_id=rse_id)
-                    logging.debug('rse_update[%s/%s]: update of rse "%s" took %f' % (heartbeat['assign_thread'], heartbeat['nr_threads'] - 1, rse_id, time.time() - start_time))
+                    logger(logging.DEBUG, 'rse_update[%s/%s]: counter update of rse "%s" took %f' %
```
bziemons, line 88: double prefix

lib/rucio/daemons/abacus/rse.py
```diff
@@ -62,32 +64,43 @@ def rse_update(once=False):
...
+                           (heartbeat['assign_thread'], heartbeat['nr_threads'] - 1, rse_id,
+                            time.time() - start_time))
+                    start_time = time.time()
+                    update_replica_counts(rse_id=rse_id)
+                    logger(logging.DEBUG, 'rse_update[%s/%s]: replica update of rse "%s" took %f' %
```
bziemons, line 93: double prefix

bziemons (changes requested):

lib/rucio/core/rse_counter.py
bziemons: How about this:
```diff
+    obsolete_files, obsolete_bytes = (
+        session.query(func.count(), func.sum(models.RSEFileAssociation.bytes))
+               .with_hint(models.RSEFileAssociation, "INDEX_FFS(REPLICAS REPLICAS_TOMBSTONE_IDX)", 'oracle')
+               .filter(models.RSEFileAssociation.rse_id == rse_id,
+                       models.RSEFileAssociation.tombstone != null(),
+                       models.RSEFileAssociation.tombstone == OBSOLETE)
+               .one()
+    )
```

bari12: 👍 I prefer that one too. Both are PEP8 compatible I guess, but for me the lonely `session` looks strange. Bens version has the advantage of saving line-length space as well.

ericvaandering:
Ok I have taken the code (after resolving comments) and put it into https://github.com/rucio/rucio/pull/4437 with the correct branch name.

PR #4404, opened 2021-03-12
Core: FTSState moved to common/constants Fix #4241

### LINKED ISSUE #4241: Replace FTSState
Motivation
----------

FTSState is not a database column, but still in `db/constants.py`, that's why it has to be used with `.name` all over the code.

Modification
------------

Migrate from `db/constants.py` to `common/constants.py` and replace the appropriate code parts with better checks.

dynamic-entropy:
Named tuples are used for readability - allows dot notation
Named tuples prevent direct modification of constant, thus preventing accidental changes to constants

Additional insight - 
str() is redundant around enum.VAR.name
> class color(Enum):
>>   RED = 1
>>   GREEN = 2

> type(color.RED.name)
>><class 'str'>

### DIFF
--- a/lib/rucio/common/constants.py
+++ b/lib/rucio/common/constants.py
@@ -15,6 +15,7 @@
 Constants.
 
 """
+from collections import namedtuple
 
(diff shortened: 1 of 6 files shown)


bari12:
Thanks for submitting this. 👍 It was merged via PR #4470

PR #4420, opened 2021-03-15
Clients: remove several dependencies from client #4097

jamesp-epcc:
This is a second try at removing some dependencies from the client and making others optional. There are two differences from my first PR:

1. tabulate is no longer being removed
2. progressbar has been removed for consistency since it was only used in one protocol, instead of being made optional

### DIFF
--- a/etc/pip-requires
+++ b/etc/pip-requires
@@ -16,7 +16,5 @@ Flask==1.1.2                                      # Python web framework
 oic==1.2.1                                        # for authentication via OpenID Connect protocol
 prometheus_client==0.9.0                          # Python client for the Prometheus monitoring system
 python-magic>=0.4.15,<0.5.0                       # File type identification using libmagic
-futures>=3.2.0; python_version < '3.0'            # Clean single-source support for Python 3 and 2
-boto>=2.49.0,<2.50.0                              # S3 boto protocol
 boto3>=1.9.130,<1.17.0                            # S3 boto protocol (new version)
-pysftp>=0.2.9,<0.3                                # Used by sftp protocol
+jsonschema>=3.2.0                                 # For JSON schema validation (Policy modules)
--- a/etc/pip-requires
+++ b/etc/pip-requires
@@ -17,4 +17,3 @@ oic==1.2.1                                        # for authentication via OpenI
 prometheus_client==0.9.0                          # Python client for the Prometheus monitoring system
 python-magic>=0.4.15,<0.5.0                       # File type identification using libmagic
 boto3>=1.9.130,<1.17.0                            # S3 boto protocol (new version)
-jsonschema>=3.2.0                                 # For JSON schema validation (Policy modules)
--- a/etc/pip-requires
+++ b/etc/pip-requires
@@ -15,3 +15,8 @@ paramiko==2.7.2                                   # SSH2 protocol library
 Flask==1.1.2                                      # Python web framework
 oic==1.2.1                                        # for authentication via OpenID Connect protocol
 prometheus_client==0.9.0                          # Python client for the Prometheus monitoring system
+python-magic>=0.4.15,<0.5.0                       # File type identification using libmagic
+futures>=3.2.0; python_version < '3.0'            # Clean single-source support for Python 3 and 2
+boto>=2.49.0,<2.50.0                              # S3 boto protocol
+boto3>=1.9.130,<1.17.0                            # S3 boto protocol (new version)
+pysftp>=0.2.9,<0.3                                # Used by sftp protocol
--- a/etc/pip-requires-client
+++ b/etc/pip-requires-client
@@ -1,18 +1,9 @@
 # All dependencies needed to run rucio client should be defined here
 setuptools>=36.8.0,<45.0.0; python_version < '3.0'          # Python packaging utilities (36.8.0 last py2.6 compatible version, 45.0 only Py3 compatible)
 setuptools>=45.0.0,<=50.3.2; python_version >= '3.0'        # Python packaging utilities
-argcomplete>=1.9.0,<=1.12.2                                 # Bash tab completion for argparse
 requests>=2.20.0,<=2.25.0                                   # Python HTTP for Humans.
 urllib3>=1.24.2,<=1.26.4                                    # HTTP library with thread-safe connection pooling, file post, etc.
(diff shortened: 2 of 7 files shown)


mlassnig (dismissed):
LGTM

at one point in the future we need to overhaul this version pinning in a more general sense. many of them come from "not meddling with OS-installed versions"

jamesp-epcc:
I squashed the changes into a single commit, but otherwise this is unchanged since Mario reviewed.

bziemons:
I am still unhappy about the removal of dependencies from the *pip-requires* files (except when they are not needed anymore), since that means those dependencies wont be available for testing.

jamesp-epcc:
> I am still unhappy about the removal of dependencies from the _pip-requires_ files (except when they are not needed anymore), since that means those dependencies wont be available for testing.

If any of the ones I have removed are useful for testing, I am happy to move them to `pip-requires-test` instead of removing them altogether.

bari12:
I think we have to address this differently (Ben and I already talked about this a bit)

Right now `pip-requires-client` (or on the server `pip-requires` + `pip-requires-client`) only hold the dependencies which are absolutely required to run Rucio. The setup.py loads them from there. The optional ones (Defined as extras) are just defined in the setup.py files. 

But Ben is right, this makes it much harder for the tests, since there is not a single file which includes all the dependencies (mandatory and optional) Rucio uses. To escape that, the tests do run the setup.py with all the options, but that's something easy to forget if you add a new option (Like here)

To move forward: I think we should merge this like @jamesp-epcc proposed, but in parallel work on a better way to manage the dependencies.

etc/pip-requires
```diff
@@ -16,3 +16,8 @@ paramiko==2.7.2                                   # SSH2 protocol library
 Flask==1.1.2                                      # Python web framework
 oic==1.2.1                                        # for authentication via OpenID Connect protocol
 prometheus_client==0.9.0                          # Python client for the Prometheus monitoring system
+python-magic>=0.4.15,<0.5.0                       # File type identification using libmagic
```
bari12, line 18: Where do we use this?

etc/pip-requires
```diff
@@ -16,3 +16,8 @@ paramiko==2.7.2                                   # SSH2 protocol library
...
+futures>=3.2.0; python_version < '3.0'            # Clean single-source support for Python 3 and 2
```
bari12: It doesn't make much sense to just include this for the server package, since this is already Python 3 only. Why is this not needed for clients anymore?

etc/pip-requires
```diff
@@ -16,3 +16,8 @@ paramiko==2.7.2                                   # SSH2 protocol library
...
+boto>=2.49.0,<2.50.0                              # S3 boto protocol
+boto3>=1.9.130,<1.17.0                            # S3 boto protocol (new version)
+pysftp>=0.2.9,<0.3                                # Used by sftp protocol
```
bari12: Also seems weird to me to just include these protocols on the server?

etc/pip-requires-client
```diff
@@ -1,18 +1,9 @@
...
-boto>=2.49.0,<2.50.0                                        # S3 boto protocol
 tabulate>=0.8.0,<0.9.0                                      # Pretty-print tabular data
-progressbar2>=3.39.0,<=3.53.1                               # Text progress bar
-bz2file>=0.98,<0.99                                         # Read and write bzip2-compressed files.
-python-magic>=0.4.15,<0.5.0                                 # File type identification using libmagic
-futures>=3.2.0; python_version < '3.0'                      # Clean single-source support for Python 3 and 2
 six>=1.12.0<1.16.0                                          # Python 2 and 3 compatibility utilities
-boto3>=1.9.130,<1.17.0                                      # S3 boto protocol (new version)
-pysftp>=0.2.9,<0.3                                          # Used by sftp protocol
-jsonschema>=3.2.0                                           # For JSON schema validation (Policy modules)
```
bari12, line 18: jsonschema is imported in several common modules and a bin/ aswell. Why is this going to `pip-requires-test` - that way it will never be installed in the package?

etc/pip-requires
mlassnig: This might come in as a secondary dependency. In general, we should remove all entries that are not directly needed.

etc/pip-requires
jamesp-epcc: As far as I could tell, they are not used in any of the client code. It's possible they are not used on the server either in which case they could be completely removed, but I did not fully check this.

etc/pip-requires
jamesp-epcc: boto3 is needed server side to do the URL signing. Maybe boto and pysftp are not needed though. I moved them to the server dependencies rather than removing them altogether in case it broke something but it might be better to remove them completely.

etc/pip-requires-client
jamesp-epcc: I may have misunderstood how the dependency files worked - I moved jsonschema to `pip-requires-test` because it appeared to be needed by one of the tests. I thought this would also make it available to the server but if not it should go in `pip-requires`.

jamesp-epcc:
I've made a few changes to `pip-requires` and `pip-requires-test`. I've moved `jsonschema` back to the client dependencies for now as I think it is needed there after all. I've removed `boto` and `pysftp` from the server dependencies but added them to `pip-requires-test` since some of the tests depend on them. I've removed `futures` altogether since it doesn't seem to be used anywhere, but left `python-magic` as the dumper uses it.

etc/pip-requires
bari12: I looked this up, it is used by the auditor to identify mimetypes. I guess we need to keep this.

bari12: Just saw that @jamesp-epcc commented this as well :-)

mlassnig:
@bari12 not sure what to do here... I've lost track of the dependency mgmt side of things.

bari12 (dismissed):

bari12:
I think it is good like that, it does decrease the dependencies a bit. @bziemons is working on a new dependency management model in parallel, but we can merge this PR already.
@jamesp-epcc please rebase once though.

jamesp-epcc:
I have rebased this now. Sorry for the delay, it's been a chaotic sort of week.

PR #4423, opened 2021-03-15
Client : Add an option to export_data not to export distance : Closes #4411

### LINKED ISSUE #4411: Add an option to export_data not to export distance
Motivation
----------
By default `export_data` export the distance which considerably slow down the response. There should be an option to disable the distance export, but the default behaviour will stay the same.

cserf:
Client : Add an option to export_data not to export distance : Closes #4411

### DIFF
--- a/lib/rucio/web/rest/webpy/v1/exporter.py
+++ b/lib/rucio/web/rest/webpy/v1/exporter.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 # -*- coding: utf-8 -*-
-# Copyright 2018-2020 CERN
+# Copyright 2018-2021 CERN
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -21,6 +21,7 @@
 # - Benedikt Ziemons <benedikt.ziemons@cern.ch>, 2020
 # - Cedric Serfon <cedric.serfon@cern.ch>, 2021
 
+from urllib.parse import parse_qs
 from web import (application, header, loadhook, ctx)
 
 from rucio.api.exporter import export_data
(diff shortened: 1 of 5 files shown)


bari12 (dismissed):

bari12:
It is good, but can you run the `tools/add_header -i` on all these files, to make the header consistent please.

lib/rucio/web/rest/webpy/v1/exporter.py
```diff
@@ -28,6 +29,11 @@
 from rucio.web.rest.common import rucio_loadhook, RucioController, check_accept_header_wrapper
 from rucio.web.rest.utils import generate_http_error
 
+try:
+    from urlparse import parse_qs
+except ImportError:
+    from urllib.parse import parse_qs
```
bziemons: no more Python 2 :) just use `from urllib.parse import parse_qs`

cserf: :+1:

cserf:
> It is good, but can you run the `tools/add_header -i` on all these files, to make the header consistent please.

Done, but as I already mentioned it in the past: the add_header changes the date wrongly, e.g.:
```
-# - Benedikt Ziemons <benedikt.ziemons@cern.ch>, 2020-2021
+# - Benedikt Ziemons <benedikt.ziemons@cern.ch>, 2021
```
Would be good to fix it. In addition:
```
#!/usr/bin/env python
```
should be change to 
```
#!/usr/bin/env python3
```
The native python in my dev node is still python2 :grin:

bziemons:
> Done, but as I already mentioned it in the past: the add_header changes the date wrongly, e.g.:

it does it according to the git history of that specific file name :) It just does not pick up file moves, PRs are welcome

Edit: I think the point is to just change the header to be in the exact same format as the add_header script would output. I revert the copyright dates for my commits.

bari12:
In the grand scheme of things the `add_header` script does it correctly. I think it is fine and the utility outweighs the small inconsistencies. The important part is anyway that the persons names are added, that the dates are not 100% correct is a rather minor issue (to me)

bziemons (approved):

PR #4424, opened 2021-03-15
WebUI: R2D2 wrongly creates a container Fix #583

### LINKED ISSUE #583: R2D2 wrongly creates a container if the list of dids is containing only files
Motivation
----------

If a user specifies only files in the list of DIDs field R2D2 will still create a container and will try to put the files in the container.

Modification
------------

Check the type of DIDs and create a container of dataset accordingly.

rak108:
R2D2 wrongly creates a container if the list of dids is containing only files Fix #583
------------------
The change is that instead of directly calling request_container, it first checks if there is is even one in the list of dids which is not of file type, and then calls request_container.

### DIFF
--- a/lib/rucio/web/ui/static/request_rule.js
+++ b/lib/rucio/web/ui/static/request_rule.js
@@ -829,7 +829,12 @@ create_rules = function(approval) {
 
     var request_container = null;
     if ((selected_dids.length > 1) && (approval)) {
-        request_container = create_request_container(selected_dids);
+        $.each(selected_dids, function(index, did){
+            if (did['did_type']!='FILE'){
+                request_container = create_request_container(selected_dids);
+                return false;
+            }
+        });
     }
 
     rse_expr = storage.get('rse_expr');


tbeerman:
I'm afraid this is more complicated than that. As you implemented it you would still put files in a container if the selected_dids are mixed, i.e., both files and datasets, which won't work. Instead there are three cases that should be handled:
- all DIDs are files -> create dataset
- all DIDs are datasets -> create container
- DIDs are mixed -> create dataset for files and add it together with the remaining datasets to a container.

rak108:
> I'm afraid this is more complicated than that. As you implemented it you would still put files in a container if the selected_dids are mixed, i.e., both files and datasets, which won't work. Instead there are three cases that should be handled:
> 
> * all DIDs are files -> create dataset
> * all DIDs are datasets -> create container
> * DIDs are mixed -> create dataset for files and add it together with the remaining datasets to a container.

Ohh alright, seems I had misunderstood the implementation. I'll work on this

rak108:
I'll make the required changes to this PR for this by the end of April/beginning of May, I am a little preoccupied with exams and end of the year submissions for university right now. Sorry for the delay!!

PR #4427, opened 2021-03-16
WebUI: Removed Jira Issue Reference & Redirected to GitHub Issues Fix #4426

### LINKED ISSUE #4426: Remove reference to JIRA from Web UI
Motivation
----------

image

The Rucio Jira project has been archived.

Modification
------------

I would suggest removing this completely. Alternatively, we can point users to the GitHub issues instead.

DhruvSondhi:
Fixes Jira Reference in reporting an error message 
------------------
This PR aims to fix the reference to the Jira archived project in the error message

### DIFF
--- a/lib/rucio/web/ui/flask/templates/request_rule.html
+++ b/lib/rucio/web/ui/flask/templates/request_rule.html
@@ -26,8 +26,8 @@
         If you are new to this interface you might want to take the <a id="start_tour">tour</a>.
         <a href="#" id="tour_alert_close" class="close">&times;</a>
       </div>
-      <div data-alert style="font-size:120%" id="jira_alert" class="alert-box secondary radius">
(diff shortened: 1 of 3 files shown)


dchristidis (commented):
Thanks for the patch. However, my personal preference is to remove this entirely. My main motivation is that there’s little gain suggesting users to submit issues exclusively for this part of the UI. Additionally, the Rucio Web UI has been used in production for a long time.

Maybe other reviewers have a different opinion.

mlassnig:
I think it's fine to keep it. Sometimes users need a little convincing to report things, and extra messages might convince them to do that.

The ID should be renamed though :-) That's still called `jira_alert` there.

bziemons (changes requested):
For the time being, please also change the file `lib/rucio/web/ui/templates/request_rule.html`

bziemons (approved):

PR #4437, opened 2021-03-16
Monitoring & Logging: Add two probes into abacus-rse instead of probes Fix #4261

### LINKED ISSUE #4261: Move deletable and obsolete replica checks from probes into Abacus-RSE

ericvaandering:
Pull request title
------------------

Valid component names are listed in the `label list <https://github.com/rucio/rucio/labels>`_.

### DIFF
--- a/lib/rucio/core/rse_counter.py
+++ b/lib/rucio/core/rse_counter.py
@@ -11,12 +11,17 @@
 # - Martin Barisits, <martin.barisits@cern.ch>, 2014
 # - Hannes Hansen, <hannes.jakob.hansen@cern.ch>, 2018-2019
 # - Brandon White, <bjwhite@fnal.gov>, 2019
+# - Eric Vaandering <ewv@fnal.gov>, 2021
 
-from sqlalchemy.orm.exc import NoResultFound
+import datetime
 
 from rucio.common.exception import CounterNotFound
 from rucio.db.sqla import models, filter_thread_work
+from rucio.db.sqla.constants import OBSOLETE
 from rucio.db.sqla.session import read_session, transactional_session
+from sqlalchemy import func
+from sqlalchemy.orm.exc import NoResultFound
+from sqlalchemy.sql.expression import null
 
 
 @transactional_session
@@ -27,8 +32,7 @@ def add_counter(rse_id, session=None):
     :param rse_id:  The id of the RSE.
     :param session: The database session in use.
     """
-    models.RSEUsage(rse_id=rse_id, source='rucio', files=0, used=0).\
-        save(session=session)
+    models.RSEUsage(rse_id=rse_id, source='rucio', files=0, used=0).save(session=session)
 
 
 @transactional_session
@@ -41,8 +45,7 @@ def increase(rse_id, files, bytes, session=None):
     :param bytes:   The number of added bytes.
     :param session: The database session in use.
     """
-    models.UpdatedRSECounter(rse_id=rse_id, files=files, bytes=bytes).\
-        save(session=session)
+    models.UpdatedRSECounter(rse_id=rse_id, files=files, bytes=bytes).save(session=session)
 
 
 @transactional_session
@@ -67,8 +70,7 @@ def del_counter(rse_id, session=None):
     :param session: The database session in use.
     """
 
-    session.query(models.RSEUsage).filter_by(rse_id=rse_id, source='rucio').\
-        delete(synchronize_session=False)
+    session.query(models.RSEUsage).filter_by(rse_id=rse_id, source='rucio').delete(synchronize_session=False)
 
 
 @read_session
@@ -84,8 +86,7 @@ def get_counter(rse_id, session=None):
     """
 
     try:
-        counter = session.query(models.RSEUsage).\
-            filter_by(rse_id=rse_id, source='rucio').one()
+        counter = session.query(models.RSEUsage).filter_by(rse_id=rse_id, source='rucio').one()
         return {'bytes': counter.used,
                 'files': counter.files,
                 'updated_at': counter.updated_at}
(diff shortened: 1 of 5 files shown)


lib/rucio/core/rse_counter.py
```diff
@@ -138,6 +139,44 @@ def update_rse_counter(rse_id, session=None):
...
+    Read the replica counts and update rse_usage.
+
+    :param rse_id:   The rse_id to update.
+    :param session:  Database session in use.
+    """
+
+    # Get the counds of obsolete and deletably (expired) files
+    obsolete_files, obsolete_bytes = (session.query(func.count(), func.sum(models.RSEFileAssociation.bytes))
+                                      .with_hint(models.RSEFileAssociation,
+                                                 "INDEX_FFS(REPLICAS REPLICAS_TOMBSTONE_IDX)", 'oracle')
```
cserf: This index is a function based index. I'm not sure it will be used like this. You can have a look how it is done for the reaper:
https://github.com/rucio/rucio/blob/55ddd574709092220f0ae70911e0964538863c46/lib/rucio/core/replica.py#L1718

lib/rucio/core/rse_counter.py
```diff
@@ -138,6 +139,44 @@ def update_rse_counter(rse_id, session=None):
...
+                                      .filter(models.RSEFileAssociation.rse_id == rse_id,
+                                              models.RSEFileAssociation.tombstone != null(),
+                                              models.RSEFileAssociation.tombstone == OBSOLETE)
+                                      .one())
+    deletable_files, deletable_bytes = (session.query(func.count(), func.sum(models.RSEFileAssociation.bytes))
+                                        .with_hint(models.RSEFileAssociation,
+                                                   "INDEX_FFS(REPLICAS REPLICAS_TOMBSTONE_IDX)", 'oracle')
```
cserf: Same comment as above

lib/rucio/core/rse_counter.py
```diff
@@ -138,6 +139,44 @@ def update_rse_counter(rse_id, session=None):
...
+    if not obsolete_files:
+        obsolete_files = 0
+
+    # Merge (INSERT or UPDATE) the values
+    # fill_rse_counter_history_table takes care of the history for all types
+    deletable = models.RSEUsage(rse_id=rse_id, used=deletable_bytes, files=deletable_files, source='expired')
+    session.merge(deletable)
+    obsolete = models.RSEUsage(rse_id=rse_id, used=obsolete_bytes, files=obsolete_files, source='obsolete')
+    session.merge(obsolete)
+
```
cserf, line 174: Should we move the previous values into the history table ?

ericvaandering:
Hints were removed. Performance is unchanged.

ericvaandering:
I think this is ready @bari12

bari12:
I'll have a look asap.

lib/rucio/core/rse_counter.py
```diff
@@ -138,6 +139,40 @@ def update_rse_counter(rse_id, session=None):
...
+@transactional_session
+def update_replica_counts(rse_id, session=None):
+    """
+    Read the replica counts and update rse_usage.
+
+    :param rse_id:   The rse_id to update.
+    :param session:  Database session in use.
+    """
+
+    # Get the counds of obsolete and deletably (expired) files
```
bari12, line 151: counts

lib/rucio/core/rse_counter.py
```diff
@@ -138,6 +139,40 @@ def update_rse_counter(rse_id, session=None):
...
+    obsolete_files, obsolete_bytes = (session.query(func.count(), func.sum(models.RSEFileAssociation.bytes))
```
bari12, line 152: I think @cserf mentioned it before, we need to pull in the hint here. Similar to 
https://github.com/rucio/rucio/blob/740b825cac747280ca159089bca4635264c07fc1/lib/rucio/core/replica.py#L1848
The issue is that the replicas table is not indexed. Although I am planing to add an rse_id index anyway, we have an index exactly for this kind of query (where tombstones are not null).

lib/rucio/daemons/abacus/rse.py
```diff
@@ -62,32 +64,38 @@ def rse_update(once=False):
...
             else:
                 for rse_id in rse_ids:
                     if graceful_stop.is_set():
                         break
                     start_time = time.time()
                     update_rse_counter(rse_id=rse_id)
-                    logging.debug('rse_update[%s/%s]: update of rse "%s" took %f' % (heartbeat['assign_thread'], heartbeat['nr_threads'] - 1, rse_id, time.time() - start_time))
+                    logger(logging.DEBUG, 'counter update of rse "%s" took %f' % (rse_id, time.time() - start_time))
+                    start_time = time.time()
+                    update_replica_counts(rse_id=rse_id)
```
bari12, line 89: I think this wouldn't work here.

The problem is that the `get_updated_rse_counters` gives you RSEs which have new replicas added or removed. But it doesn't give you RSEs whose replicas changed (from no tombstone to obsolete tombstone for example) - So it might be if nothing else happens with an RSE, that the obsolete/expired counter would never be updated.

I think this needs to be done completely separately and I would probably just do it in one big query, and not go RSE by RSE.

bari12 (changes requested):
Hi Eric, sorry for the late reply.
I don't think this would work like expected; See one of my comments about it.

bari12:
@ericvaandering I am closing this now (Can be resubmitted later, but it is just idling at the moment).
As in my last comment, I don't think it will work like this alone and needs some bigger re-work. The main issue I see is that the abacus right now is just not made to fully support that use-cases, so just patching it in won't work, since the current events will not trigger all changes. We probably need to start there.

PR #4446, opened 2021-03-17
Testing: Looking into CODEOWNER files;

bari12:
Having a first look into CODEOWNER files.


CODEOWNERS
```diff
@@ -0,0 +1,114 @@
...
+lib/rucio/api/rule.py                   @bari12
+lib/rucio/core/rule*                    @bari12
+lib/rucio/daemons/judge/                @bari12
+
+# Subscriptions
+lib/rucio/core/subscription.py          @cserf
+lib/rucio/client/subscriptionclient.py  @cserf
+lib/rucio/daemons/transmogrifier/       @cserf
+
+# Testing
```
bziemons: I would add
```
.github/                                @bziemons
etc/docker/test/                        @bziemons
lib/rucio/tests/conftest.py             @bziemons
lib/rucio/tests/ruciopytest/            @bziemons
tools/github/                           @bziemons
tools/test/                             @bziemons
```

CODEOWNERS
```diff
@@ -0,0 +1,120 @@
...
+
+# Core & Internals (This goes first and is a catch-all for core)
+lib/rucio/core/                         @bari12
+lib/rucio/common/                       @bari12
+
+# Clients
+lib/rucio/client/baseclient.py          @mlassnig
+lib/rucio/client/uploadclient.py        @mlassnig
+lib/rucio/client/downloadclient.py      @mlassnig
+lub/rucio/client/                       @mlassnig
```
bziemons: -lub +lib ;)

bziemons (changes requested):
Looking through the CODEOWNERS specification, I saw

```
# In this example, @octocat owns any file in an apps directory
# anywhere in your repository.
apps/ @octocat

# In this example, @doctocat owns any file in the `/docs`
# directory in the root of your repository and any of its
# subdirectories.
/docs/ @doctocat
```

that means the line

```
bin/                                    @mlassnig
```

would assign any bin directory to Mario. This is in line with the `.gitignore` syntax. I would suggest prefixing all entries with `/` so that they are resolved from the repository root.


Additionally, the `CODEOWNERS` file (as well as the `pull_request_template.md`) can reside in the `.github` directory. Maybe that is preferred since they are GitHub specific and to not have too many files in the root directory?

CODEOWNERS
bziemons: oh and then remove the other entries pointing to files inside `lib/rucio/client/` ?

bari12: Will change; Right the other ones can be removed. 👍

bari12:
For `bin/` I planned to do it similar to `lib/rucio/clients` but then never did. But I think that's fine for now, there is not that much change with these anyway.
Will move them to `.github/` and add the preceeding `/`

bziemons (approved):

PR #4449, opened 2021-03-18
Testing: make more tests parallel. #4134

rcarpa:
For multiple issues I'm working right now, I realize that I
copy-paste the same code for initializing rses, rules,
requests in test cases.

Add factory classes allowing to create RSEs and upload files
(implicitly creating replication rules, with correct checksums, etc).
Add an associated fixture to easily use it in multiple tests. 
Auto-cleanup created resources at the end of the test.

Migrate a couple of tests to using this fixture. This allows
to also remove some "noparallel" marks on download tests.

### DIFF
--- a/lib/rucio/tests/test_download.py
+++ b/lib/rucio/tests/test_download.py
@@ -25,361 +25,348 @@
 
 import logging
 import shutil
-import unittest
 from tempfile import TemporaryDirectory
 from zipfile import ZipFile
 
 import pytest
 
-from rucio.client.client import Client
-from rucio.client.didclient import DIDClient
 from rucio.client.downloadclient import DownloadClient
-from rucio.client.uploadclient import UploadClient
-from rucio.common.config import config_get, config_get_bool
 from rucio.common.exception import InputValidationError, NoFilesDownloaded
 from rucio.common.utils import generate_uuid
 from rucio.rse import rsemanager as rsemgr
 from rucio.rse.protocols.posix import Default as PosixProtocol
-from rucio.tests.common import file_generator
 
 
-@pytest.mark.noparallel(reason='uses pre-defined RSEs, fails when run in parallel')
-class TestDownloadClient(unittest.TestCase):
-    def setUp(self):
-        if config_get_bool('common', 'multi_vo', raise_exception=False, default=False):
-            self.vo = {'vo': config_get('client', 'vo', raise_exception=False, default='tst')}
-        else:
-            self.vo = {}
+@pytest.fixture
+def download_client():
+    logger = logging.getLogger('dlul_client')
+    logger.addHandler(logging.StreamHandler())
+    logger.setLevel(logging.DEBUG)
+    return DownloadClient(logger=logger)
 
-        logger = logging.getLogger('dlul_client')
-        logger.addHandler(logging.StreamHandler())
-        logger.setLevel(logging.DEBUG)
-        self.client = Client()
-        self.did_client = DIDClient()
-        self.upload_client = UploadClient(_client=self.client, logger=logger)
-        self.download_client = DownloadClient(client=self.client, logger=logger)
 
-    def _upoad_test_file(self, rse, scope, name, path=None):
-        item = {
-            'path': path if path else file_generator(),
-            'rse': rse,
-            'did_scope': scope,
-            'did_name': name,
-            'guid': generate_uuid(),
-        }
-        assert self.upload_client.upload([item]) == 0
-        return item
+def _check_download_result(actual_result, expected_result):
+    assert len(expected_result) == len(actual_result)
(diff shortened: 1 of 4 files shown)


lib/rucio/tests/temp_test_env.py
```diff
@@ -0,0 +1,205 @@
...
+from rucio.core import replica as replica_core
+from rucio.core import rse as rse_core
+from rucio.core import rule as rule_core
+from rucio.db.sqla import models
+from rucio.db.sqla.session import transactional_session
+from rucio.tests.common import file_generator
+from rucio.tests.common import rse_name_generator
+
+
+class TemporaryRSEFactory:
```
bziemons: Can the `GeneratedRSE` class be merged with this or replaced by this?
https://github.com/rucio/rucio/blob/63846a61555e6949d43da99bf3e0deedea877e73/lib/rucio/tests/test_preparer.py#L44-L70

lib/rucio/tests/test_download.py
```diff
@@ -25,361 +25,347 @@
...
+                    'did': did100_str,
+                    'clientState': 'ALREADY_DONE',
+                    'dest_file_paths': ['%s/%s' % (tmp_dir, item100['did_name'])],
+                }
+            ],
+        )
+
+
+@pytest.mark.xfail(reason='XRD1 must be initialized https://github.com/rucio/rucio/pull/4165/')
 @pytest.mark.noparallel(reason='uses pre-defined RSEs, fails when run in parallel')
```
bziemons: please update the reason to point out that the necessary XRD1 RSE is used from the integration tests. I am not sure if this test cannot run in parallel, but it definitely uploads a test file and never deletes it (which would mean this is `@pytest.mark.dirty` as well)

lib/rucio/tests/temp_test_env.py
```diff
@@ -0,0 +1,205 @@
...
+        self.upload_client.upload([item])
+        did = {'scope': scope, 'name': name}
+        self.created_dids.append(did)
+        return item if return_full_item else did
+
+
+class TemporaryTestEnv(TemporaryRSEFactory, TemporaryFileFactory):
+    def __init__(self, vo, default_scope):
+        TemporaryRSEFactory.__init__(self, vo=vo)
+        TemporaryFileFactory.__init__(self, vo=vo, default_scope=default_scope)
```
bziemons: why not use `super(…)` here and in the cleanup? Also I think a fixture for each that just looks like this would be a better fit and seems more modular:
```python3
@pytest.fixture
def rse_factory(vo):
    factory = TemporaryRSEFactory(vo=vo)
    yield factory
    factory.cleanup()
```

bziemons (commented):
This looks a lot like what I had in mind :+1: Just a few things I noticed while looking through it

bziemons:
Also, not directly related to this PR, but I planned to make the conftest.py a bit more useful for us. I was thinking of replacing the following

https://github.com/rucio/rucio/blob/351206acb336cda9dc712247ce17ab5c09e4afa4/lib/rucio/tests/conftest.py#L28

with three files, each loaded like this:

```python
try:
    import rucio.core
    
    from .server_conftest import *
except ImportError:
    pass

try:
    import rucio.common
    
    from .common_conftest import *
except ImportError:
    pass

try:
    import rucio.client
    
    from .client_conftest import *
except ImportError:
    pass
```

that said, I haven't tested that yet, but it seems more comfortable to not worry about local imports too much.

Edit: also not sure if the common is useful at all. Maybe just server/client conftest files..

lib/rucio/tests/test_download.py
rcarpa: Done, but not sure I did it the way you have it in mind.

lib/rucio/tests/temp_test_env.py
rcarpa: You are right, I started by having a single class, then split it into 2. It's better to have separate fixtures. Split it and amended the commits

lib/rucio/tests/temp_test_env.py
rcarpa: It can be easily done. Can we leave it for a separate MR though ?

rcarpa:
I would leave those imports for now. It looks much more confusing the way you propose it.

lib/rucio/tests/test_download.py
bziemons: Yup, looks good now. Imo it is not bad if a few tests are noparallel - it is just perfect to reduce this number as much as possible. For the integration tests (which don't take very long anyway right now), it is fine. It's just good to have proper documentation of why.

bziemons (approved):

bziemons:
> I would leave those imports for now. It looks much more confusing the way you propose it.

let's have a quick talk about this then

bari12:
👍 Thanks a lot for this 😃

PR #4458, opened 2021-03-19
Transfers: Submit requests in creation order. Closes #4375

### LINKED ISSUE #4375: Submitter should try to process requests in a FIFO order
Motivation
----------

As far as I can tell, the Submitter does not necessarily process the queued requests based on the time they were created. This can lead to significant delays when there’s a backlog.

Modification
------------

Consider ordering the requests by the appropriate timestamp column.

rcarpa:
If requests are not submitted in FIFO order, it can lead to delay when
there is a backlog.

To be able to test the submitter, do some small fixes to existing code.

### DIFF
--- a/lib/rucio/core/transfer.py
+++ b/lib/rucio/core/transfer.py
@@ -33,6 +33,7 @@
 # - Patrick Austin <patrick.austin@stfc.ac.uk>, 2020
 # - Benedikt Ziemons <benedikt.ziemons@cern.ch>, 2020
 # - Thomas Beermann <thomas.beermann@cern.ch>, 2021
+# - Radu Carpa <radu.carpa@cern.ch>, 2021
 
 from __future__ import division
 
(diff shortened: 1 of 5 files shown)


lib/rucio/core/transfer.py
```diff
@@ -1323,12 +1323,14 @@ def __list_transfer_requests_and_source_replicas(total_workers=0, worker_number=
                                  models.Request.previous_attempt_id,
                                  models.Request.dest_rse_id,
                                  models.Request.retry_count,
-                                 models.Request.account) \
+                                 models.Request.account,
+                                 models.Request.created_at) \
```
bziemons, line 1327: fyi, if this changes the returned tuple of `__list_transfer_requests_and_source_replicas` it might break the preparer https://github.com/rucio/rucio/blob/327571d841bd36d0712acf2b5ce4ae4e3ba44897/lib/rucio/core/request.py#L1530

rcarpa: It doesn't change the output format. Here I change the internal subquery to propagate the field towards the main query. And the main one uses the field only for order_by

bari12 (approved):

PR #4474, opened 2021-03-22
Clients: retry on 502/503/504 errors. Closes #4444

### LINKED ISSUE #4444: Base client should retry when receiving a 502 error
Motivation
----------

ATLAS has observed a couple of times when requests to Rucio would fail with a 502 Bad Gateway error. This does not seem to be a wide-spread problem, but during those times it was happening consistently. Initial discussions pointed to HAProxy. 

Depending on which use of the Rucio client is failing, this can have detrimental effects on DPA. We could ask the users of the client to implement some protection around this, but there’s no doubt that it’s better for Rucio to handle it internally. The base client already has a mechanism for retrying requests that fail with a connection error.

Modification
------------

We should consider extending the retrying mechanism to also cover this type of error.

rcarpa:
Perform an exponential backoff when retrying.

Test it for password authentication and for get requests using a simple mock server 
which always returns 502.

### DIFF
--- a/lib/rucio/client/baseclient.py
+++ b/lib/rucio/client/baseclient.py
@@ -34,7 +34,8 @@
 # - Andrew Lister <andrew.lister@stfc.ac.uk>, 2019
 # - Eli Chadwick <eli.chadwick@stfc.ac.uk>, 2020
 # - Patrick Austin <patrick.austin@stfc.ac.uk>, 2020
-# - Benedikt Ziemons <benedikt.ziemons@cern.ch>, 2020
+# - Benedikt Ziemons <benedikt.ziemons@cern.ch>, 2020-2021
+# - Radu Carpa <radu.carpa@cern.ch>, 2021
 
 '''
  Client class for callers of the Rucio system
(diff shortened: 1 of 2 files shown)


lib/rucio/client/baseclient.py
```diff
@@ -114,6 +115,13 @@
     expiration_time=60,
 )
 
+STATUS_CODES_TO_RETRY = [502, 503, 504]
```
mlassnig, line 118: should this go into common.constants?

lib/rucio/client/baseclient.py
```diff
@@ -114,6 +115,13 @@
...
+MAX_RETRY_BACK_OFF_SECONDS = 10
```
mlassnig, line 119: this too?

lib/rucio/client/baseclient.py
```diff
@@ -402,6 +410,9 @@ def _send_request(self, url, headers=None, type='GET', data=None, params=None, s
                     result = self.session.delete(url, headers=hds, data=data, verify=self.ca_cert, timeout=self.timeout)
                 else:
                     return
+                if result.status_code in STATUS_CODES_TO_RETRY:
+                    back_off(retry)
```
mlassnig: log warning about this and state how long you will wait

mlassnig: (and the other instances where this happens as well)

mlassnig (commented):
LGTM, just two tiny comments

lib/rucio/client/baseclient.py
rcarpa: Doesn't look like some common constant for all rucion. I'd keep it here.

lib/rucio/client/baseclient.py
rcarpa: done

bari12 (approved):

Issue #4410, opened 2021-03-12
Configparser loads entire ENV space


bari12:
Motivation
----------
For probably historic reasons our config parser (For rucio.cfg) also loads the entire ENV space into the config parser. This causes certain issues at a few sites, since the parser breaks do to their environment setup (Duplicate keys etc.)
https://github.com/rucio/rucio/blob/253c2f12f6e99e5eca5f59c6538e5a2bcd5f7c48/lib/rucio/common/config.py#L224-L227

I do not think that it is necessary to load the entire ENV space in there, since the ENV variables we are interested in, are usually directly accessed from the environment.

1. Investigate if this is really needed
2. If not take it out from the Configparser

PR #4481, opened 2021-03-23
Clients: don't load env into config defaults. #4410

rcarpa:


bziemons (approved):
LGTM as long as no-one was actually using this feature ;)

bari12:
We didn't really advertise this as a feature, so I hope not :-)

PR #4484, opened 2021-03-23
Core & Internals: cleanup old heartbeats. Closes #4459

### LINKED ISSUE #4459: Heartbeat cleanup
Motivation
----------
Due to the more dynamic `hostnames` with kubernetes the current cleanup functions of heartbeats are not sufficient, and the table is polluted with old heartbeats which never get removed. Heartbeats do continue to work normally, but the table will grow indefinitely.

Modification
------------
Add a cleanup of all heartbeats older than 24h to the `sanity_check` method.

rcarpa:
Due to the more dynamic hostnames with kubernetes the current cleanup
functions of heartbeats are not sufficient, and the table is polluted
with old heartbeats which never get removed.

Add a cleanup of all heartbeats older than 24h.

### DIFF
--- a/lib/rucio/core/heartbeat.py
+++ b/lib/rucio/core/heartbeat.py
@@ -1,17 +1,25 @@
-# Copyright European Organization for Nuclear Research (CERN)
+# -*- coding: utf-8 -*-
+# Copyright 2015-2021 CERN
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
-# You may not use this file except in compliance with the License.
+# you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
-# http://www.apache.org/licenses/LICENSE-2.0
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


lib/rucio/core/heartbeat.py
```diff
@@ -70,6 +83,9 @@ def _sanity_check(executable, hostname, hash_executable=None, session=None):
             if not pid_exists(pid):
                 session.query(Heartbeats).filter_by(hostname=hostname, pid=pid).delete()
 
+    if expiration_delay:
+        cardiac_arrest(older_than=expiration_delay, session=session)
```
bari12, line 87: I wonder if we should put this into a try/except. Reason being that the `sanity_check` is often called in a daemon outside of the while loop which would recover from errors. And there might be some row-contention/deadlock errors if this is called at the same time.

rcarpa: _sanity_check is wrapped by sanity_check() which does exactly that. Do we need to catch any exceptions other than "databaseError" ?

bari12: Right, I have misread this, ignore what I said :-)

mlassnig (approved):

bari12:
Will be included in `1.25` release line

Issue #2213, opened 2019-02-18
Removing last replica which is a constituent


bari12:
Motivation
----------
When the last replica of a file is removed in the reaper, it should only be removed from the dataset if it is not a constituent of an existing archive replica. This also implies, if a archive replica is removed and there are no replicas on the constituents, the constituents need to be removed from their parent datasets.

PR #4487, opened 2021-03-23
Deletion: handle archives in reaper. #1431 #2213

rcarpa:
Ensure that replicas are removed from datasets only if they don't exists
in an archive.

When removing an archive, handle the removal of its constituents.
Only if:
- they don't exist in another archive
- they don't have replicas outside any archive

This is achieved by recursively calling the cleanup function after
removing the archive (and thus the constituent replicas).

### DIFF
--- a/lib/rucio/tests/test_reaper2.py
+++ b/lib/rucio/tests/test_reaper2.py
@@ -24,16 +24,22 @@ from datetime import datetime, timedelta
 
 import pytest
 
+from rucio.api import replica as replica_api
+from rucio.api import rse as rse_api
 from rucio.common.config import config_get_bool
+from rucio.common.exception import ReplicaNotFound, DataIdentifierNotFound
 from rucio.common.types import InternalAccount, InternalScope
 from rucio.common.utils import generate_uuid
-from rucio.api import rse as rse_api
-from rucio.api import replica as replica_api
+from rucio.core import did as did_core
 from rucio.core import replica as replica_core
 from rucio.core import rse as rse_core
+from rucio.core import rule as rule_core
 from rucio.core import scope as scope_core
 from rucio.core import vo as vo_core
-from rucio.daemons.reaper.reaper2 import reaper, REGION, run as run_reaper
+from rucio.daemons.reaper.reaper2 import reaper, REGION
+from rucio.daemons.reaper.reaper2 import run as run_reaper
+from rucio.db.sqla.models import ConstituentAssociationHistory
+from rucio.db.sqla.session import read_session
 from rucio.tests.common import rse_name_generator
 
 __mock_protocol = {'scheme': 'MOCK',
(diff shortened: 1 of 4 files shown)


mlassnig (commented):
LGTM, second pair of eyes @cserf @bari12 ?

bari12:
Review is already request ;-)

lib/rucio/tests/test_reaper2.py
```diff
@@ -203,3 +209,130 @@ def test_reaper_multi_vo(vo):
...
+def test_archive_removal_impact_on_constituents(rse_factory, file_factory, mock_scope, root_account):
+    rse_name, rse_id = rse_factory.make_mock_rse()
+    scope = mock_scope
+    account = root_account
+
+    # Create an 2 archives and 4 files:
+    # - One only exists in the first archive
+    # - One in both, plus another replica, which is not in an archive
+    # - One in both, plus another replica, which is not in an archive; and this replica has expired
+    # - One in both, plus another replica, which is not in an archive; and this replica has expired; but a replication rule exists on this second replica
```
cserf, line 223: This condition shouldn't occur if the rule already exists

lib/rucio/tests/test_reaper2.py
```diff
@@ -203,3 +209,130 @@ def test_reaper_multi_vo(vo):
...
+    assert len(list(did_core.list_content(**dataset1))) == 3
+    assert len(list(did_core.list_archive_content(**archive1))) == 0
+    assert len(list(did_core.list_archive_content(**archive2))) == 3
+    assert __get_archive_contents_history_count(archive1) == 4
+    assert __get_archive_contents_history_count(archive2) == 0
+
+    # Expire the second archive replica and run reaper another time
+    # c_with_expired_replica is removed because its external replica got removed at previous step
+    # and it exist only inside the archive now.
+    # If not open, Dataset2 will be removed because it will be empty.
```
cserf, line 312: We should remove the condition on the dataset open

cserf (commented):
Also looks good to me. Just 2 small non blocking comments

lib/rucio/tests/test_reaper2.py
rcarpa: Could you please elaborate ?

lib/rucio/tests/test_reaper2.py
rcarpa: I believe there is related to #3311 . So should be fixed when working on that issue. For now it works like that.

lib/rucio/tests/test_reaper2.py
bari12: I think what Cedric means is that the forth one 
```
One in both, plus another replica, which is not in an archive; and this replica has expired; but a replication rule exists on this second replica
```
cannot exist, since there cannot be an expired replica covered by a replication rule.

rcarpa: Should I remove this case from the test then ? it ensures that replicas protected by rules are not removed by reaper. But it's better to remove it completely if such case cannot arrive.

bari12: In theory such a case cannot arrive, but in pricniple it makes sense that the reaper protects for it as well, just in case some other workflow is not acting as expected. I would leave it. @cserf what do you think?

bari12 (approved):

bari12:
I am merging this as an enhancement in the ongoing `1.25` release line.

PR #4505, opened 2021-03-29
Improvements to the declaration of lost files : Closes #3235

### LINKED ISSUE #3235: Improvements to the declaration of lost files
I would like to make the following suggestions:

 1. Use the synchronous API in case the number of files is below some thresholds. We had a case where a request to mark a large amount of files as temporarily unavailable was causing long delays for single-file requests which should have had higher priority.
 2. Change the message to something other than `replicas successfully declared`. This is misleading as the PFNs could be invalid and the user will move on thinking that everything is done.
 3. Do some checking on the PFNs, at the very least that they are valid URLs. It should be noted that on the side of Minos this isn’t handled gracefully.
```
  File "/usr/lib/python2.7/site-packages/rucio/daemons/badreplicas/minos.py", li
ne 182, in minos
    dict_rse[rse_id].extend(tmp_dict_rse[rse_id])
  File "/usr/lib/python2.7/site-packages/retrying.py", line 49, in wrapped_f
    return Retrying(*dargs, **dkw).call(f, *args, **kw)
  File "/usr/lib/python2.7/site-packages/retrying.py", line 206, in call
    return attempt.get(self._wrap_exception)
  File "/usr/lib/python2.7/site-packages/retrying.py", line 247, in get
    six.reraise(self.value[0], self.value[1], self.value[2])
  File "/usr/lib/python2.7/site-packages/retrying.py", line 200, in call
    attempt = Attempt(fn(*args, **kwargs), attempt_number, False)
  File "/usr/lib/python2.7/site-packages/rucio/db/sqla/session.py", line 284, in
 new_funct
  File "/usr/lib/python2.7/site-packages/rucio/core/replica.py", line 437, in ge
t_pfn_to_rse
    split_se = surl.split('/')[2].split(':')
```

cserf:
Improvements to the declaration of lost files : Closes #3235

### DIFF
--- a/bin/rucio-admin
+++ b/bin/rucio-admin
@@ -19,7 +19,7 @@
 # - Martin Barisits <martin.barisits@cern.ch>, 2012-2020
 # - Vincent Garonne <vincent.garonne@cern.ch>, 2012-2018
 # - Thomas Beermann <thomas.beermann@cern.ch>, 2012-2013
-# - Cedric Serfon <cedric.serfon@cern.ch>, 2013-2020
+# - Cedric Serfon <cedric.serfon@cern.ch>, 2013-2021
 # - Wen Guan <wen.guan@cern.ch>, 2014
 # - Ralph Vigne <ralph.vigne@cern.ch>, 2014
 # - David Cameron <david.cameron@cern.ch>, 2014-2015
@@ -65,7 +65,8 @@ from rucio.common.exception import (AccountNotFound, DataIdentifierAlreadyExists
                                     RSENotFound, RSEOperationNotSupported, InvalidRSEExpression,
                                     DuplicateContent, RuleNotFound, CannotAuthenticate,
                                     Duplicate, ReplicaIsLocked, ConfigNotFound, ScopeNotFound)
-from rucio.common.utils import chunks, construct_surl, sizefmt, get_bytes_value_from_string, render_json, parse_response, extract_scope
+from rucio.common.utils import (chunks, construct_surl, sizefmt, get_bytes_value_from_string,
+                                render_json, parse_response, extract_scope, clean_surls)
 from rucio.rse import rsemanager as rsemgr
 
 try:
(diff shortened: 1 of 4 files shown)


bin/rucio-admin
```diff
@@ -1003,18 +1004,57 @@ def declare_bad_file_replicas(args):
...
+        print('Getting the information about RSE protocols. It can take several seconds')
+        dict_rse = client.export_data(distance=False)
+        prot_dict = {}
+        for rse, dict_attr in dict_rse['rses'].items():
+            protocols = dict_attr['protocols']
+            for prot in protocols:
+                if prot['scheme'] == 'srm':
+                    prot_dict[str('%s://%s%s' % (prot['scheme'], prot['hostname'], prot['prefix']))] = rse
+                else:
+                    prot_dict[str('%s://%s:%s%s' % (prot['scheme'], prot['hostname'], prot['port'], prot['prefix']))] = rse
```
dchristidis: Sorry, I didn’t try to run the new code, but my understanding is that this would require explicitly setting the port in the PFNs. Is that the case?

bin/rucio-admin
```diff
@@ -1003,18 +1004,57 @@ def declare_bad_file_replicas(args):
...
+        cnt = 0
+        nchunk = math.ceil(tot_files / chunk_size)
+        for chunk in chunks(bad_files_pfns, chunk_size):
+            list_bad_pfns = []
+            cnt += 1
+            previous_pattern = None
+            for surl in clean_surls(chunk):
+                unknown = True
+                if previous_pattern:
+                    if surl.find(previous_pattern) > -1:
```
dchristidis: May I suggest expressing this as `previous_pattern in surl` instead? Or maybe `surl.startswith(previous_pattern)`?

bin/rucio-admin
```diff
@@ -1003,18 +1004,57 @@ def declare_bad_file_replicas(args):
...
+                        list_bad_pfns.append(surl)
+                        unknown = False
+                        continue
+                for pattern in prot_dict:
+                    if surl.find(pattern) > -1:
```
dchristidis: Ditto.

bin/rucio-admin
```diff
@@ -1003,18 +1004,55 @@ def declare_bad_file_replicas(args):
...
-    print('%s replicas successfully declared' % tot_files)
+    if len(bad_files_pfns) < 100:
+        # Using the old API to declare
+        non_declared = client.declare_bad_file_replicas(pfns=bad_files_pfns, reason=args.reason)
+        for rse in non_declared:
+            for pfn in non_declared[rse]:
+                print('%s : PFN %s cannot be declared.' % (rse, pfn))
+    else:
+        print('Getting the information about RSE protocols. It can take several seconds')
+        dict_rse = client.export_data(distance=False)
```
dchristidis, line 1021: Has the export feature been tested? For one thing, the appropriate permission is not defined, so it defaults to `_is_root()`. For another, it didn’t work for me.
```
>>> c = Client(account='ddmadmin')
>>> d = c.export_data(distance=False)
2021-04-07 13:15:21,720 WARNING Waiting 0.25s due to reason: server returned 502 
2021-04-07 13:20:22,381 WARNING Waiting 0.5s due to reason: server returned 502 
2021-04-07 13:25:23,212 WARNING Waiting 1.0s due to reason: server returned 502 
  File "/var/home/dchristidis/.local/lib/python3.9/site-packages/rucio/client/exportclient.py", line 54, in export_data
    raise exc_cls(exc_msg)
rucio.common.exception.RucioException: An unknown exception occurred.
Details: no error information passed (http status code: 502)
```

cserf: Yes, I noticed the missing permission. It works on the dev container

bin/rucio-admin
cserf: Constraint relaxed. Both are accepted

bin/rucio-admin
cserf: Ditto

bin/rucio-admin
bari12: I don't think we should use the export API here. That's really not the use-case it was made for. Can't you use rse_info?

cserf: How ?? We have no reverse method to get a RSE from a hostname.

bari12: As discussed offline, let's go ahead with it like this.

dchristidis: Issue still stands. I am afraid I cannot approve this dependence on the export API until I can confirm that it works. It appears that, in addition to the problems described in my previous comment, the endpoint might not be properly registered on Flask.
```
>>> c = Client(account='ddmadmin')
>>> d = c.export_data(distance=False)
  File "/var/home/dchristidis/.local/lib/python3.9/site-packages/rucio/client/exportclient.py", line 54, in export_data
    raise exc_cls(exc_msg)
rucio.common.exception.RucioException: An unknown exception occurred.
Details: no error information passed (http status code: 404)
```

bari12: @cserf can you please check if the endpoint is registered? A testcase would of course help here to show these errors 😏

bziemons: I think there is a misconception here. As far as I can see the endpoint is well-defined. It needs to be configured in the server's rucio.cfg though:

```
[api]
endpoints = …, export, …
```

This is already done in the latest dev container, so make sure you have that if you're testing this.
https://github.com/rucio/containers/blob/master/dev/rucio.cfg#L183

cserf: The 502 error was due to a problem with the exporter API in Flask. It was working fine on web.py :(
Fixed by https://github.com/rucio/rucio/pull/4658

cserf: Rebase + Fix permission

bari12 (approved):

Issue #4499, opened 2021-03-25
Rework transfers workflow


rcarpa:
Motivation
----------

Transfers has grown to be quite complex. Many parts of the code are difficult to follow. 

Modification
------------
Expected changes (more or less in the expected order)
- add functional tests for transfers. This will also allow to ave better understanding of functional requirements. 
- perform cosmetic refactoring of existing code without changing its behavior (split gigantic functions; remove un-used code; factorize duplicate code)
- create a clear interface between the transfer tools and the conveyor
- try to unify staging with transfers
- modularize transfertools: create some abstraction interface and implement this interface in existing transfer tools : globus/fts

Issue #3254, opened 2020-01-23
Switch protocol selection to third_party_copy read/write


bari12:
Motivation
----------
This needs to be changed in various parts of the code from `third_party_copy` to use `third_party_copy_read` and `third_party_copy_write`. In a later feature release, the `third_party_copy` column can then be removed.

PR #4507, opened 2021-03-30
Transfers: refactor get_transfer_requests_and_source_replicas. #4499

rcarpa:
Try to reduce the amount of duplicate code and reorganize over-complicated sections while keeping the same functional behavior. 

The merge request is split into multiple commits for easier review. But should be squashed into one on merge.

### DIFF
--- a/lib/rucio/core/transfer.py
+++ b/lib/rucio/core/transfer.py
@@ -791,6 +791,9 @@ def get_transfer_requests_and_source_replicas(total_workers=0, worker_number=0, 
         if source_rse_id is None or rse is None:
             continue
 
+        if rses and dest_rse_id not in rses:
+            continue
+
         dest_rse_name = ctx.rse_name(dest_rse_id)
         source_rse_name = ctx.rse_name(source_rse_id)
 
@@ -856,9 +859,6 @@ def get_transfer_requests_and_source_replicas(total_workers=0, worker_number=0, 
 
         allow_tape_source = True
         try:
-            if rses and dest_rse_id not in rses:
-                continue
-
             # Get source protocol
             source_protocol = ctx.protocol(source_rse_id, source_scheme, 'read')
 
--- a/lib/rucio/core/transfer.py
+++ b/lib/rucio/core/transfer.py
@@ -627,13 +627,11 @@ def get_dsn(scope, name, dsn):
     return 'other'
 
 
-def __build_source_url(scope, name, path, protocol):
-    source_url = list(protocol.lfns2pfns(lfns={'scope': scope.external, 'name': name, 'path': path}).values())[0]
-    return source_url
-
-
 def __build_dest_url(scope, name, protocol, dest_rse_attrs, dest_is_deterministic, dest_is_tape, dict_attributes, retry_count, activity):
-    # I.3 - Compute the destination url
+    """
+    Private helper function to build destination URL when retrieving transfers to execute.
+    """
+
     if dest_is_deterministic:
         dest_url = list(protocol.lfns2pfns(lfns={'scope': scope.external, 'name': name}).values())[0]
     else:
@@ -653,6 +651,9 @@ def __build_dest_url(scope, name, protocol, dest_rse_attrs, dest_is_deterministi
 
 
 def __rewrite_source_url(source_url, source_sign_url, dest_sign_url, source_scheme):
+    """
+    Parametrize source url for some special cases of source and destination schemes
+    """
     if dest_sign_url == 'gcs':
         if source_scheme in ['davs', 'https']:
             source_url += '?copy_mode=push'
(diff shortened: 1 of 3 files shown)


lib/rucio/core/transfer.py
```diff
@@ -627,6 +627,70 @@ def get_dsn(scope, name, dsn):
...
+    Private helper function to build destination URL when retrieving transfers to execute.
+    """
+
+    if dest_is_deterministic:
+        dest_url = list(protocol.lfns2pfns(lfns={'scope': scope.external, 'name': name}).values())[0]
+    else:
+        # compute dest url in case of non deterministic
+        # naming convention, etc.
+        dsn = get_dsn(scope, name, dict_attributes.get('dsn', None))
+        # DQ2 path always starts with /, but prefix might not end with /
```
cserf, line 641: This comment can be removed. It is ATLAS specific

lib/rucio/core/transfer.py
```diff
@@ -658,15 +722,57 @@ def get_transfer_requests_and_source_replicas(total_workers=0, worker_number=0,
...
+            _rse_info = self.rse_info(rse_id)
+            if _rse_info['rse_type'] == RSEType.TAPE or _rse_info['rse_type'] == 'TAPE':
+                return True
+            return False
+
+        def protocol(self, rse_id, scheme, operation):
+            protocol_key = '%s_%s_%s' % (operation, rse_id, scheme)
+            protocol = self.protocols.get(protocol_key)
+            if not protocol:
+                protocol = rsemgr.create_protocol(self.rse_info(rse_id), 'third_party_copy', scheme)
```
cserf, line 764: third_party_copy should be replaced by operation ?

lib/rucio/core/transfer.py
```diff
@@ -750,40 +857,21 @@ def get_transfer_requests_and_source_replicas(total_workers=0, worker_number=0,
...
-                rses_info[source_rse_id] = rsemgr.get_rse_info(rse=source_rse_name,
-                                                               vo=get_rse_vo(rse_id=source_rse_id, session=session),
-                                                               session=session)
-            if source_rse_id not in rse_attrs:
-                rse_attrs[source_rse_id] = get_rse_attributes(source_rse_id, session=session)
             # Get source protocol
-            source_rse_id_key = 'read_%s_%s' % (source_rse_id, source_protocol)
-            if source_rse_id_key not in protocols:
-                protocols[source_rse_id_key] = rsemgr.create_protocol(rses_info[source_rse_id], 'third_party_copy', source_protocol)
+            source_protocol = ctx.protocol(source_rse_id, source_scheme, 'read')
```
cserf, line 867: @bari12 should we use third_party_copy_read ? https://github.com/rucio/rucio/issues/3254

lib/rucio/core/transfer.py
```diff
@@ -979,115 +1028,54 @@ def get_transfer_requests_and_source_replicas(total_workers=0, worker_number=0,
...
+                    continue
+
+                if current_source_is_tape and not new_source_is_tape or \
+                        new_source_is_tape and not current_source_is_tape:
+                    # Tape and Disk sources must not be used at the same time.
+                    # Either keep existing sources unchanged, or substitute all existing source with the new source.
+
+                    # Find the best ranking among existing sources
+                    avail_top_ranking = None
+                    for founded_source in transfers[req_id]['sources']:
```
cserf, line 1050: found_source is enough :smile:

lib/rucio/core/transfer.py
rcarpa: This would change the existing behavior. It may be correct to change it, but it is out of the scope of this PR and is a bug which must be fixed separately.

lib/rucio/core/transfer.py
rcarpa: Also changes the current behavior. Probably correct idea, but must be fixed separately. Otherwise we introduce new features in a PR which is about refactoring. If problems due to this PR are discovered, it will be difficult to distinguish between it being caused by the refactoring; or by the new feature.

lib/rucio/core/transfer.py
rcarpa: I completely agree. Was already like this and I left it unchanged. However, these lines will be removed in a later PR I am preparing.

lib/rucio/core/transfer.py
bari12: Let's leave it in for now. Radu just restructured the code here, so I think it's not bad to leave things the same as much as possible. But let's take it out in a later PR.

bari12 (dismissed):
Looks very nice to me. I would merge it like this. Waiting for @cserf last comments.

cserf:
If it's just about restructuring, that looks good. Would be good to address the non-resolved comment in another PR

cserf (dismissed):

bari12 (approved):

Issue #2063, opened 2019-01-30
Versioned (History) Tables should be defined explicitly


bari12:
Motivation
----------
There are currently 7 history tables which are defined by using the SQLA model base `Versioned`. The advantage of this is that the history table does not have to be created by hand, but it is just mirrored by the primary base. However, by doing this we loose full control about the indices and primary keys of the history table, as it is also just mirrored from the primary model. This already creates inconsistencies, as the history tables we define on oracle, do not have indices and primary keys.

Modification
------------
Two possibilities:

- Find a way to adapt indices/PKs for versioned tables (AFAIK not possible)
- Do not use Versioned at all and define the history tables explicitly; This is already done for some tables anyway.

Tables to change:

- [x] `quarantined_replicas`
- [x] `rse_usage`
- [x] `account_usage`
- [x] `requests`
- [x] `sources`
- [x] `subscriptions`
- [x] `configs`

And then

- [x] Removal of Versioned dependency entirely

PR #4517, opened 2021-04-05
Deletion : Fix condition preventing to delete collection replica for open datasets : Closes #3311

### LINKED ISSUE #3311: Reaper seems to skip the deletion of the collection replica

cserf:
Deletion : Fix condition preventing to delete collection replica for open datasets : Closes #3311

### DIFF
--- a/lib/rucio/core/replica.py
+++ b/lib/rucio/core/replica.py
@@ -1777,8 +1777,7 @@ def __cleanup_after_replica_deletion(rse_id, files, session=None):
                          models.CollectionReplica.name == parent_name,
                          exists(select([1]).prefix_with("/*+ INDEX(DIDS DIDS_PK) */", dialect='oracle')).where(
                              and_(models.DataIdentifier.scope == parent_scope,
-                                  models.DataIdentifier.name == parent_name,
-                                  models.DataIdentifier.is_open == False)),  # NOQA
+                                  models.DataIdentifier.name == parent_name)),
                          ~exists(select([1]).prefix_with("/*+ INDEX(CONTENTS CONTENTS_PK) */", dialect='oracle')).where(
                              and_(models.DataIdentifierAssociation.scope == parent_scope,
                                   models.DataIdentifierAssociation.name == parent_name))))
@@ -1792,16 +1791,28 @@ def __cleanup_after_replica_deletion(rse_id, files, session=None):
                                   models.DataIdentifierAssociation.name == parent_name))))
 
                 # 3) Schedule removal of the entry from the DIDs table
-                did_condition.append(
-                    and_(models.DataIdentifier.scope == parent_scope,
-                         models.DataIdentifier.name == parent_name,
-                         models.DataIdentifier.is_open == False,  # NOQA
-                         ~exists([1]).where(
-                             and_(models.DataIdentifierAssociation.child_scope == parent_scope,
-                                  models.DataIdentifierAssociation.child_name == parent_name)),
-                         ~exists([1]).where(
-                             and_(models.DataIdentifierAssociation.scope == parent_scope,
-                                  models.DataIdentifierAssociation.name == parent_name))))
+                remove_open_did = config_get('reaper', 'remove_open_did', default=False, session=session)
+                if remove_open_did:
+                    did_condition.append(
+                        and_(models.DataIdentifier.scope == parent_scope,
+                             models.DataIdentifier.name == parent_name,
+                             ~exists([1]).where(
+                                 and_(models.DataIdentifierAssociation.child_scope == parent_scope,
+                                      models.DataIdentifierAssociation.child_name == parent_name)),
+                             ~exists([1]).where(
+                                 and_(models.DataIdentifierAssociation.scope == parent_scope,
+                                      models.DataIdentifierAssociation.name == parent_name))))
+                else:
+                    did_condition.append(
+                        and_(models.DataIdentifier.scope == parent_scope,
+                             models.DataIdentifier.name == parent_name,
+                             models.DataIdentifier.is_open == False,  # NOQA
+                             ~exists([1]).where(
+                                 and_(models.DataIdentifierAssociation.child_scope == parent_scope,
+                                      models.DataIdentifierAssociation.child_name == parent_name)),
+                             ~exists([1]).where(
+                                 and_(models.DataIdentifierAssociation.scope == parent_scope,
+                                      models.DataIdentifierAssociation.name == parent_name))))
 
         if child_did_condition:
 
--- a/lib/rucio/tests/test_abacus_collection_replica.py
+++ b/lib/rucio/tests/test_abacus_collection_replica.py
@@ -22,6 +22,7 @@
 # - Martin Barisits <martin.barisits@cern.ch>, 2020-2021
 # - Mario Lassnig <mario.lassnig@cern.ch>, 2020
 # - Simon Fayer <simon.fayer05@imperial.ac.uk>, 2021
+# - Cedric Serfon <cedric.serfon@cern.ch>, 2021
 
 import os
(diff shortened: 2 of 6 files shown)


mlassnig (dismissed):

mlassnig:
Patch looks ok, not sure why it doesn't apply to release-patch-test set? @bari12

bari12:
Because there are tests failing :-)

bari12:
@cserf please check failing tests.

lib/rucio/tests/test_replica.py
```diff
@@ -257,15 +257,12 @@ def test_delete_replicas_from_datasets(self, rse_factory, mock_scope, root_accou
         _, rse_id = rse_factory.make_mock_rse()
```
rcarpa, line 254: Seems like this test was explicitly ensuring that open datasets are not removed, while closed one are removed. Going all the way back to 2013 ... https://github.com/rucio/rucio/commit/c93322d0641f36f165288e1a9477bf83f8ca1692 . It's sad that there is absolutely no context in the commit message to understand why it was done like that.

rcarpa (dismissed):

lib/rucio/tests/test_replica.py
cserf: @mlassnig and @bari12 should comment. AFAIR, the purpose was to address this kind of workflow: You have a dataset, all the rules for this dataset expires, all the replica from the dataset are removed, then the dataset is deleted and the dataset name cannot be reused. If the dataset is open that means that the production system might try to add new files, so we don't want that the dataset completely disappear.
But I don't really think that it's useful to keep the dataset if all the other files are already gone.

mlassnig: I think that depends, e.g., for user datasets it makes sense, for officially produced maybe not (as the associated task has finished/failed anyway).

cserf: Yeah, but it leaves a lot of zombies, for maybe one or two users datasets. I would really prefer to delete everything. Let's discuss in tomorrow devops

bari12: I agree, it probably was the use-case that an open dataset might be re-used. For user ones it actually makes sense.
The zombies I am not too worried about, at least in terms of DB rows used, since the total stays the same with the row moved to `deleted_dids`.

lib/rucio/core/replica.py
```diff
@@ -1728,7 +1727,6 @@ def __cleanup_after_replica_deletion(rse_id, files, session=None):
                 did_condition.append(
                     and_(models.DataIdentifier.scope == parent_scope,
                          models.DataIdentifier.name == parent_name,
-                         models.DataIdentifier.is_open == False,  # NOQA
```
bari12, line 1798: As discussed in the dev meeting, this should be made conditional so communities can decide if they want to get rid of empty open datasets. @cserf can you adapt this? (Or maybe submit it in two PRs?)

bari12 (changes requested):

cserf:
WT??
```
E   sqlalchemy.exc.IntegrityError: (pymysql.err.IntegrityError) (1062, "Duplicate entry 'reaper-remove_open_did-2021-05-28 12:39:26' for key 'configs_history.PRIMARY'")
E   [SQL: INSERT INTO dev.configs_history (section, opt, value, created_at, updated_at) VALUES (%(section)s, %(opt)s, %(value)s, %(created_at)s, %(updated_at)s)]
E   [parameters: {'section': 'reaper', 'opt': 'remove_open_did', 'value': 'False', 'created_at': datetime.datetime(2021, 5, 28, 12, 39, 25, 952863), 'updated_at': datetime.datetime(2021, 5, 28, 12, 39, 25, 952866)}]
```
The primary key for the config history table is contatenation of section, option and timestamp ???

cserf:
And I didn't touch this test
```
 ______________ ERROR at teardown of test_archive_on_dataset_level ______________
/usr/local/lib64/python3.6/site-packages/sqlalchemy/engine/base.py:1284: in _execute_context
    cursor, statement, parameters, context
/usr/local/lib64/python3.6/site-packages/sqlalchemy/engine/default.py:590: in do_execute
    cursor.execute(statement, parameters)
E   cx_Oracle.DatabaseError: ORA-00908: missing NULL keyword
```

cserf:
For the PK error, will have to wait for https://github.com/rucio/rucio/issues/2063

lib/rucio/core/replica.py
```diff
@@ -907,6 +907,19 @@ def _list_files_wo_replicas(files_wo_replica, session):
             yield scope, name, bytes, md5, adler32
 
 
+@read_session
+def remove_empty_open_did(force_refresh=False, session=None):
+    """
+    Check if empty open dids must be removed (default Yes)
```
rcarpa: I like that we put it default to "yes". But it must be very clear in release notes that the default behavior changes.

lib/rucio/tests/test_dataset_replicas.py
```diff
@@ -277,14 +279,28 @@ def test_update_collection_replica(self):
...
         delete_replicas(rse_id=self.rse_id, files=files, session=self.db_session)
-        update_request = self.db_session.query(models.UpdatedCollectionReplica).filter_by(rse_id=self.rse_id, scope=self.scope, name=dataset_name).one()  # pylint: disable=no-member
-        update_collection_replica(update_request=update_request.to_dict(), session=self.db_session)
-        dataset_replica = self.db_session.query(models.CollectionReplica).filter_by(scope=self.scope, name=dataset_name).all()  # pylint: disable=no-member
-        assert len(dataset_replica) == 0
+        with pytest.raises(NoResultFound):
+            update_collection_replica(update_request=update_request.to_dict(), session=self.db_session)
+
+        # New behaviour, open empty datasets are deleted
+        cconfig_set(section='reaper', option='remove_open_did', value=True, session=self.db_session)
```
rcarpa: The config option will persist in the database after the test. Did you try using the config table override as in this example : https://github.com/rucio/rucio/blob/f38bfe391f0c07e023408de6a826c9affba1d5af/lib/rucio/tests/test_conveyor_submitter.py#L96 ?  Config_set will then go a temporary table. Don't foget to also override the cache as here https://github.com/rucio/rucio/blob/f38bfe391f0c07e023408de6a826c9affba1d5af/lib/rucio/tests/test_conveyor_submitter.py#L101
otherwise the cached value can pollute the tests.

lib/rucio/tests/test_abacus_collection_replica.py
```diff
@@ -145,14 +148,72 @@ def test_abacus_collection_replica(self):
...
+        assert dataset_replica[0]['available_length'] == 0
+
+
+@pytest.mark.parametrize("core_config_mock", [{"table_content": [
+    ('reaper', 'remove_open_did', True)
+]}], indirect=True)
+@pytest.mark.parametrize("caches_mock", [{"caches_to_mock": [
+    'rucio.core.config', 'rucio.core.replica'
+]}], indirect=True)
+def test_abacus_collection_replica_new(vo, rse_factory, rucio_client, did_factory, core_config_mock, caches_mock):
```
rcarpa, line 171: I wouldn't use the "_new" suffix. Probably better to name the test by its function. For example : _with_open_did_cleanup ; or something like that.

rcarpa (approved):

lib/rucio/tests/test_abacus_collection_replica.py
rcarpa: same in all other places where you refer to "old behavior" or "new"

bari12 (approved):

PR #4538, opened 2021-04-13
Testing: only leave one submitter test

rcarpa:
Due to job assignment to concurrent submitters by hash, having multiple
parallel submitters in tests result in flacky tests.

The problem should be fixed more in-depth later. However, as a rapid
workaround, having a single submitter test will reduce the number of
failing pipelines.

Update the single test to have the same coverage as the initial two.


bziemons:
Why not mark all submitter tests as noparallel ?

rcarpa:
To avoid marking everything noparallel. These test can be run in parallel with everything else, just not between them

bari12 (approved):

PR #4541, opened 2021-04-13
Recovery : Reason is not maintained for temporarily unavailable replicas : Closes #4023

### LINKED ISSUE #4023: Reason is not maintained for temporarily unavailable replicas
Motivation
----------

https://github.com/rucio/rucio/blob/master/lib/rucio/daemons/badreplicas/minos.py#L222

This doesn’t look intentional. The entries in `BadPFNs` are dropped, so the reason is lost forever.

Modification
------------

Extra care should be taken for entries with different reason in the same chunk, though the chance of this occurring should be very low.

cserf:
Recovery : Reason is not maintained for temporarily unavailable replicas : Closes #4023

### DIFF
--- a/lib/rucio/daemons/badreplicas/minos.py
+++ b/lib/rucio/daemons/badreplicas/minos.py
@@ -68,13 +68,13 @@ def minos(bulk=1000, once=False, sleep_time=60):
     hb_thread = threading.current_thread()
     heartbeat.sanity_check(executable=executable, hostname=hostname)
     heart_beat = heartbeat.live(executable, hostname, pid, hb_thread)
-    prefix = 'Thread [%i/%i] : ' % (heart_beat['assign_thread'], heart_beat['nr_threads'])
+    prefix = 'minos[%i/%i] ' % (heart_beat['assign_thread'], heart_beat['nr_threads'])
     logger = formatted_logger(logging.log, prefix + '%s')
     logger(logging.INFO, 'Minos starting')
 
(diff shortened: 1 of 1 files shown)


lib/rucio/daemons/badreplicas/minos.py
```diff
@@ -69,24 +68,27 @@ def minos(bulk=1000, once=False, sleep_time=60):
     hb_thread = threading.current_thread()
     heartbeat.sanity_check(executable=executable, hostname=hostname)
     heart_beat = heartbeat.live(executable, hostname, pid, hb_thread)
-    prepend_str = 'Thread [%i/%i] : ' % (heart_beat['assign_thread'], heart_beat['nr_threads'])
-    logging.info(prepend_str + 'Minos starting')
+    prefix = 'Thread [%i/%i] : ' % (heart_beat['assign_thread'], heart_beat['nr_threads'])
```
dchristidis: Please consider splitting the logging changes into a separate commit and referencing issue #4220. Additionally, the message format should be `daemon_name[heartbeat['assign_thread']/heartbeat['nr_threads']]: …`

dchristidis (changes requested):

lib/rucio/daemons/badreplicas/minos.py
cserf: I would prefer not to revert the changes already done, but point taken for next time. I'll fix the format.

dchristidis (approved):

PR #4495, opened 2021-03-25
Transfers: Refactor list transfer requests

bziemons:
Since I was a bit annoyed by the returned tuple of `__list_transfer_requests_and_source_replicas` and floating variables to look up certain columns, I introduce a named tuple here. This should make further changes in these parts less painful.

Based on #4458

PR #4542, opened 2021-04-14
Transfers: further rework get transfer requests. #4499. Closes #4516.

### LINKED ISSUE #4516: Request with one tape and one multihop tape source will crash submitter
Motivation
----------
Suppose in the `get_transfer_requests_and_source_replicas` main loop, a request is added with a multihop tape source. Then in another iteration, a direct tape source is identified. We end up in https://github.com/rucio/rucio/blob/740b825cac747280ca159089bca4635264c07fc1/lib/rucio/core/transfer.py#L968
Since the first source was multihop, we clear `transfers[req_id]['sources']` at https://github.com/rucio/rucio/blob/740b825cac747280ca159089bca4635264c07fc1/lib/rucio/core/transfer.py#L1005-L1009
Then shortly after, if the second source is tape, we will end up in https://github.com/rucio/rucio/blob/740b825cac747280ca159089bca4635264c07fc1/lib/rucio/core/transfer.py#L1042-L1050
where there is an unchecked access to `transfers[req_id]['sources'][0]` that crashes.

Modification
------------
In this situation, I believe we still want to prefer the direct tape source over the multihop tape source and want to take the path at L1047. So I would suggest modifying L1045 to
```python
if len(transfers[req_id]["sources"]) and (
    transfers[req_id]["sources"][0][3] > ranking
    or (
        transfers[req_id]["sources"][0][3] == ranking
        and transfers[req_id]["sources"][0][4] <= link_ranking
    )
):
    continue
```

rcarpa:
Continue rework of transfers internal code. This PR has multiple functional changes: 

- there is no early binding when iterating over sources. All candidate transfers, including intermediate hops, are generated between possible sources and destination, followed by a merge operation. This should have a higher computational and memory cost: a small dict will be generated in memory for each hop, instead of one dict per transfer request. Also, generating this dict takes cpu time, but we mostly speak about string operations. The positive side is that this change should fix multiple issues with previous code. For example: impossibility to find the best source for multipath transfers; impossibility to correctly chose between mixed disk and tape sources; inconsistent transfer definition generated for intermediate jumps in multipaths; etc

- source and destination generation for multipath transfers now use common code path with singlepath ones. The common code  is capable to handle more cases than needed for multipath transfers. For example: it can correctly handle source and destination schemes which are not used for multipath today. However, it allows to reduce the amount of duplicate code.

- the OIDC support is activated after creating multihop transfers. Meaning it could be activated for each intermediate hop separately.

The PR is split into multiple commits for easier review, but it's probably better to squash them on merge. Some of them cannot be safely reverted individually. 

Resolves #4495

### DIFF
--- a/lib/rucio/core/request.py
+++ b/lib/rucio/core/request.py
@@ -1614,7 +1614,9 @@ def __throttler_request_state(activity, source_rse_id, dest_rse_id, session: "Op
 
 
 def reduce_requests(
-    req_sources: "List[RequestWithSources]", sort_reduce_funcs: "List[ReduceFunction]", logger: "Callable",
+    req_sources: "List[RequestWithSources]",
+    sort_reduce_funcs: "List[ReduceFunction]",
+    logger: "Callable",
 ) -> "Iterator[RequestResultOrState]":
     """
     Reduces the passed requests & sources objects by using the sort-reduce
--- a/lib/rucio/core/request.py
+++ b/lib/rucio/core/request.py
@@ -1633,9 +1633,9 @@ def reduce_requests(
         result = [
             {
                 'request_id': rws.request_id,
-                'dest_rse_id': rws.dest_rse_id,
+                'dest_rse_id': rws.dest_rse.id,
                 'activity': rws.activity,
-                'src_rse_id': source.rse_id,
+                'src_rse_id': source.rse.id,
                 'distance_ranking': source.distance_ranking
             }
             for source in rws.sources
--- a/lib/rucio/core/request.py
+++ b/lib/rucio/core/request.py
@@ -62,7 +62,7 @@ from rucio.transfertool.fts3 import FTS3Transfertool
 RequestAndState = namedtuple('RequestAndState', ['request_id', 'request_state'])
 
 if TYPE_CHECKING:
-    from rucio.core.transfer import RequestWithSource
+    from rucio.core.transfer import RequestWithSources
     from typing import Any, Dict, Iterable, Iterator, List, Optional, Callable, Set, Union
     from sqlalchemy.orm import Session
 
@@ -1614,7 +1614,7 @@ def __throttler_request_state(activity, source_rse_id, dest_rse_id, session: "Op
 
 
 def reduce_requests(
-    req_sources: "List[RequestWithSource]", sort_reduce_funcs: "List[ReduceFunction]", logger: "Callable",
+    req_sources: "List[RequestWithSources]", sort_reduce_funcs: "List[ReduceFunction]", logger: "Callable",
 ) -> "Iterator[RequestResultOrState]":
     """
     Reduces the passed requests & sources objects by using the sort-reduce
@@ -1624,11 +1624,17 @@ def reduce_requests(
     """
     assert len(req_sources) != 0, 'parameter request sources must be non-empty'
 
-    # sort by Request.id for partitioning later
-    req_sources.sort(key=lambda rws: rws.request_id)
-
-    def pick_result(items: "List[RequestWithSource]") -> "Optional[RequestResult]":
-        result = [item._asdict() for item in items]
+    def pick_result(rws: "RequestWithSources") -> "Optional[RequestResult]":
+        result = [
+            {
+                'request_id': rws.request_id,
+                'dest_rse_id': rws.dest_rse_id,
+                'activity': rws.activity,
+                'src_rse_id': source.rse_id,
+                'distance_ranking': source.distance_ranking
+            }
+            for source in rws.sources
+        ]
         debug_log = []
         for sort_reduce in sort_reduce_funcs:
             newresult = list(sort_reduce(result))
@@ -1636,7 +1642,7 @@ def reduce_requests(
             result = newresult
 
         if len(result) == 0:
-            logger(logging.WARNING, 'all available sources were filtered for requests with id %s', items[0].request_id)
+            logger(logging.WARNING, 'all available sources were filtered for requests with id %s', rws.request_id)
             logger(logging.DEBUG, 'the following filters ran:\n' + '\n'.join(debug_log))
         else:
             return result[0]
@@ -1647,17 +1653,9 @@ def reduce_requests(
         else:
             return result
 
-    cur_request_id = req_sources[0].request_id
-
-    # partition the req_sources by request_id and yield the best result from each group
-    current_items = [req_sources[0]]
-    for idx in range(1, len(req_sources)):
-        if cur_request_id != req_sources[idx].request_id:
-            yield result_or_no_sources(pick_result(current_items))
-            cur_request_id = req_sources[idx].request_id
-            current_items.clear()
-        current_items.append(req_sources[idx])
-    yield result_or_no_sources(pick_result(current_items))
+    for rws in req_sources:
+        cur_request_id = rws.request_id
+        yield result_or_no_sources(pick_result(rws))
 
 
 def get_supported_transfertools(rse_id: str, session=None) -> "Set[str]":
--- a/lib/rucio/core/request.py
+++ b/lib/rucio/core/request.py
@@ -1552,6 +1552,9 @@ def list_requests(src_rse_ids, dst_rse_ids, states=[RequestState.WAITING], sessi
 
 @transactional_session
 def preparer_update_requests(source_iter: "Iterable[RequestResultOrState]", session: "Optional[Session]" = None) -> int:
+    """
+    Update transfer requests according to preparer settings.
+    """
     count = 0
     for rws in source_iter:
         update_dict = dict()
--- a/lib/rucio/core/request.py
+++ b/lib/rucio/core/request.py
@@ -17,7 +17,7 @@
 # - Mario Lassnig <mario.lassnig@cern.ch>, 2013-2021
 # - Vincent Garonne <vincent.garonne@cern.ch>, 2013-2017
 # - Cedric Serfon <cedric.serfon@cern.ch>, 2014-2020
-# - Martin Barisits <martin.barisits@cern.ch>, 2014-2020
+# - Martin Barisits <martin.barisits@cern.ch>, 2014-2021
 # - Wen Guan <wen.guan@cern.ch>, 2014-2016
 # - Joaquín Bogado <jbogado@linti.unlp.edu.ar>, 2015-2019
 # - Thomas Beermann <thomas.beermann@cern.ch>, 2016-2021
@@ -28,16 +28,17 @@
 # - Andrew Lister <andrew.lister@stfc.ac.uk>, 2019
 # - Brandon White <bjwhite@fnal.gov>, 2019
 # - Benedikt Ziemons <benedikt.ziemons@cern.ch>, 2020-2021
-# - Matt Snyder <msnyder@bnl.gov>, 2021
 # - Radu Carpa <radu.carpa@cern.ch>, 2021
+# - Matt Snyder <msnyder@bnl.gov>, 2021
 
 import datetime
 import json
 import logging
 import time
 import traceback
+from collections import namedtuple
 from itertools import filterfalse
-from typing import TYPE_CHECKING, Set
+from typing import TYPE_CHECKING
 
 from six import string_types
 from sqlalchemy import and_, or_, func, update
@@ -58,11 +59,16 @@ from rucio.db.sqla.constants import RequestState, RequestType, ReplicaState, Loc
 from rucio.db.sqla.session import read_session, transactional_session, stream_session
 from rucio.transfertool.fts3 import FTS3Transfertool
 
+RequestAndState = namedtuple('RequestAndState', ['request_id', 'request_state'])
+
 if TYPE_CHECKING:
-    from typing import List, Tuple, Iterable, Iterator, Optional, Callable, Sequence
+    from rucio.core.transfer import RequestWithSource
+    from typing import Any, Dict, Iterable, Iterator, List, Optional, Callable, Set, Union
     from sqlalchemy.orm import Session
 
-    RowIterator = Iterator[Sequence]
+    RequestResult = Dict[str, Any]
+    RequestResultOrState = Union[RequestResult, RequestAndState]
+    RowIterator = Iterator[RequestResult]
     ReduceFunction = Callable[[RowIterator], RowIterator]
 
 """
@@ -1544,38 +1550,30 @@ def list_requests(src_rse_ids, dst_rse_ids, states=[RequestState.WAITING], sessi
         yield request
 
 
-# important column indices from __list_transfer_requests_and_source_replicas
-request_id_col = 0
-activity_col = 7
-dest_rse_id_col = 10
-source_rse_id_col = 12
-distance_col = 20
-
-extra_transfertool_col = 21
-
-
 @transactional_session
-def preparer_update_requests(source_iter: "Iterable[Sequence]", session: "Optional[Session]" = None) -> int:
+def preparer_update_requests(source_iter: "Iterable[RequestResultOrState]", session: "Optional[Session]" = None) -> int:
     count = 0
-    for req_source in source_iter:
+    for rws in source_iter:
         update_dict = dict()
-        if len(req_source) == 2:
+        if isinstance(rws, RequestAndState):
             # special case where the first entry is the request id and the second is the new state
             # (see handling of RequestState.NO_SOURCES in reduce_requests)
-            update_dict[models.Request.state] = req_source[1]
+            request_id = rws.request_id
+            update_dict[models.Request.state] = rws.request_state
         else:
+            request_id = rws['request_id']
             update_dict[models.Request.state] = __throttler_request_state(
-                activity=req_source[activity_col],
-                source_rse_id=req_source[source_rse_id_col],
-                dest_rse_id=req_source[dest_rse_id_col],
+                activity=rws['activity'],
+                source_rse_id=rws['src_rse_id'],
+                dest_rse_id=rws['dest_rse_id'],
                 session=session,
             )
-            update_dict[models.Request.source_rse_id] = req_source[source_rse_id_col]
+            update_dict[models.Request.source_rse_id] = rws['src_rse_id']
 
-        if len(req_source) > extra_transfertool_col:
-            update_dict[models.Request.transfertool] = req_source[extra_transfertool_col]
+            if 'transfertool' in rws:
+                update_dict[models.Request.transfertool] = rws['transfertool']
 
-        session.query(models.Request).filter_by(id=req_source[request_id_col]).update(update_dict, synchronize_session=False)
+        session.query(models.Request).filter_by(id=request_id).update(update_dict, synchronize_session=False)
         count += 1
     return count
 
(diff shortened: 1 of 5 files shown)


rcarpa:
The new integration tests I added recently showed some regressions. Working on fixing them

lib/rucio/core/transfer.py
```diff
@@ -713,376 +1057,203 @@ def get_transfer_requests_and_source_replicas(total_workers=0, worker_number=0,
...
-    ctx = _LocalContext(session)
+    request_with_sources = __list_transfer_requests_and_source_replicas(total_workers=total_workers,
+                                                                        worker_number=worker_number,
+                                                                        limit=limit,
+                                                                        activity=activity,
+                                                                        older_than=older_than,
+                                                                        rses=rses,
+                                                                        request_state=RequestState.QUEUED,
+                                                                        transfertool=transfertool,
+                                                                        session=session)
```
bziemons: imo better:
```py
    request_with_sources = __list_transfer_requests_and_source_replicas(
        total_workers=total_workers,
        worker_number=worker_number,
        limit=limit,
        activity=activity,
        older_than=older_than,
        rses=rses,
        request_state=RequestState.QUEUED,
        transfertool=transfertool,
        session=session,
    )
```
see Black formatter.

lib/rucio/core/transfer.py
```diff
@@ -1278,19 +1288,28 @@ def protocol(self, rse_id, scheme, operation):
...
+    :param total_workers:    Number of total workers.
+    :param worker_number:    Id of the executing worker.
     :param limit:            Integer of requests to retrieve.
     :param activity:         Activity to be selected.
     :param older_than:       Only select requests older than this DateTime.
     :param rses:             List of rse_id to select requests.
     :param transfertool:     The transfer tool as specified in rucio.cfg.
     :param session:          Database session to use.
-    :returns:                List.
+    :returns:                List of RequestWithSource namedtuple.
```
bziemons: adjust comment

lib/rucio/core/request.py
```diff
@@ -1592,48 +1590,48 @@ def __throttler_request_state(activity, source_rse_id, dest_rse_id, session: "Op
...
 
 
-def reduce_requests(req_sources: "List[Tuple]", sort_reduce_funcs: "List[ReduceFunction]", logger: "Callable") -> "RowIterator":
+def reduce_requests(
+    req_sources: "List[RequestWithSources]", sort_reduce_funcs: "List[ReduceFunction]", logger: "Callable",
+) -> "Iterator[RequestResultOrState]":
     """
     Reduces the passed request sources tuples by using the sort-reduce functions,
     returning the first of the remaining items or a simple
     (request_id, RequestState.NO_SOURCES) tuple, if no sources were found.
```
bziemons: doc needs adjustment
```py
    """
    Reduces the passed requests & sources objects by using the sort-reduce
    functions, yielding the best RequestResult object or a RequestAndState
    object. If all sources were filtered, a RequestAndState object with
    RequestState.NO_SOURCES is yielded.
    """
```

lib/rucio/core/transfer.py
```diff
@@ -1382,14 +1398,25 @@ def __list_transfer_requests_and_source_replicas(total_workers=0, worker_number=
...
+
+        # rses (of unknown length) should be a temporary table to check against instead of this special case
+        if rses and dest_rse_id not in rses:
+            continue
+
+        request = requests_by_id.get(request_id)
+        if not request:
+            request = RequestWithSources(id=request_id, rule_id=rule_id, scope=scope, name=name, md5=md5, adler32=adler32, byte_count=byte_count,
+                                         activity=activity, attributes=attributes, previous_attempt_id=previous_attempt_id, dest_rse_id=dest_rse_id,
+                                         account=account, retry_count=retry_count)
```
bziemons: is there any specific reason not to use a `namedtuple` and get rid of great unwrapping in the for-loop?

lib/rucio/core/request.py
rcarpa: You even provided the entire correct docstring ! Thanks. Updated.

lib/rucio/core/transfer.py
rcarpa: I agree that it's better this way

lib/rucio/core/transfer.py
rcarpa: done

lib/rucio/core/transfer.py
rcarpa: Namedtuples are immutable. There were a couple of places where having a fully mutable class allowed to simplify the logic. For example: to sanitize the source_ranking and set it to 0 in the case when it is returned as None. I'm however not very sure the benefit is really justified right now. Named tuples could do the trick quite well. It ads complexity in some places, but makes life easier in others.... However, I'm thinking of (maybe) bringing some logic into this class in the future for a more object-oriented design. But I'm far from that design for now, and I'll maybe revert back to named tuples if I find it better suited latter.

bziemons (dismissed):

mlassnig:
@rcarpa LGTM, just this here:

Conflicting files:
`lib/rucio/core/transfer.py`

rcarpa:
> @rcarpa LGTM, just this here:
> 
> Conflicting files:
> `lib/rucio/core/transfer.py`

Rebased. The conflict was with one of my recent hotfixes. The hotfix doesn't apply to this PR, as the affected code was completely reworked and removed in this PR.

lib/rucio/core/request.py
```diff
@@ -1544,38 +1550,30 @@ def list_requests(src_rse_ids, dst_rse_ids, states=[RequestState.WAITING], sessi
...
-dest_rse_id_col = 10
-source_rse_id_col = 12
-distance_col = 20
-
-extra_transfertool_col = 21
-
-
 @transactional_session
-def preparer_update_requests(source_iter: "Iterable[Sequence]", session: "Optional[Session]" = None) -> int:
+def preparer_update_requests(source_iter: "Iterable[RequestResultOrState]", session: "Optional[Session]" = None) -> int:
```
bari12, line 1554: Please add some documentation what the method is supposed to do. (I know, we didn't have this before either, but let's make it better readable for the next generation 😄 ) - Maybe this is more for @bziemons actually.

lib/rucio/core/request.py
```diff
@@ -1615,48 +1613,49 @@ def __throttler_request_state(activity, source_rse_id, dest_rse_id, session: "Op
     return RequestState.WAITING if limit_found else RequestState.QUEUED
 
 
-def reduce_requests(req_sources: "List[Tuple]", sort_reduce_funcs: "List[ReduceFunction]", logger: "Callable") -> "RowIterator":
+def reduce_requests(
```
bari12, line 1619: Just out of personal preference, break the parameters in individual lines instead?

lib/rucio/core/request.py
```diff
@@ -1675,37 +1674,38 @@ def get_supported_transfertools(rse_id: str, session=None) -> "Set[str]":
 
 
 def get_transfertool_filter(
```
bari12, line 1681: See above

lib/rucio/core/transfer.py
```diff
@@ -119,6 +119,83 @@
 WEBDAV_TRANSFER_MODE = config_get('conveyor', 'webdav_transfer_mode', False, None)
 
 
+class TransferSource:
```
bari12: `@dataclass` ?

lib/rucio/core/transfer.py
```diff
@@ -702,6 +781,271 @@ def __rewrite_dest_url(dest_url, dest_sign_url, dest_scheme):
     return dest_url
 
 
+def __prepare_source_dest_config(ctx, request_id, src_rse_id, dest_rse_id, source_scheme, dest_scheme, dict_attributes, activity,
```
bari12: Add a short description what this does.

lib/rucio/core/transfer.py
```diff
@@ -702,6 +781,271 @@ def __rewrite_dest_url(dest_url, dest_sign_url, dest_scheme):
...
+    dest_spacetoken = None
+    if dest_protocol.attributes and 'extended_attributes' in dest_protocol.attributes and \
+            dest_protocol.attributes['extended_attributes'] and 'space_token' in dest_protocol.attributes['extended_attributes']:
+        dest_spacetoken = dest_protocol.attributes['extended_attributes']['space_token']
+
+    return transfer_src_type, transfer_dst_type, source_url, dest_url, dest_spacetoken, overwrite, bring_online
+
+
+@transactional_session
+def __prepare_transfer_definition(ctx, rws, source, computed_distance, dict_attributes, transfertool, retry_other_fts, list_hops, activity,
```
bari12: Also add quick description

lib/rucio/core/transfer.py
```diff
@@ -702,6 +781,271 @@ def __rewrite_dest_url(dest_url, dest_sign_url, dest_scheme):
...
+            transfer['sources'] = [(source_rse_name, source_url, source_rse_id, source.source_ranking, computed_distance)]
+            transfer['account'] = rws.account
+            if rws.previous_attempt_id:
+                file_metadata['previous_attempt_id'] = rws.previous_attempt_id
+
+        transfer_path.append(transfer)
+    return transfer_path
+
+
+def __merge_or_discard_tranfer_definitions(candidate_paths):
```
bari12: Description

lib/rucio/core/transfer.py
```diff
@@ -723,379 +1067,205 @@ def get_transfer_requests_and_source_replicas(total_workers=0, worker_number=0,
...
-                if ctx.is_tape_rse(source_rse_id) or ctx.rse_attrs(source_rse_id).get('staging_required', False):
-                    bring_online = bring_online_local
-                    transfer_src_type = "TAPE"
-                    if not allow_tape_source:
-                        if req_id not in reqs_only_tape_source:
-                            reqs_only_tape_source.append(req_id)
-                        if req_id in reqs_no_source:
-                            reqs_no_source.remove(req_id)
-                        continue
+            # rse_name will be None if rse is staging area
```
bari12: Where is this enforced?

lib/rucio/core/transfer.py
```diff
@@ -702,6 +781,271 @@ def __rewrite_dest_url(dest_url, dest_sign_url, dest_scheme):
...
+        best_candidate[0]['sources'].extend(candidate[0]['sources'])
+        # If one source has force ipv4, force it for all sources
+        if best_candidate[0].get('use_ipv4') or candidate[0].get('use_ipv4'):
+            best_candidate[0]['use_ipv4'] = True
+
+    return best_candidate
+
+
 @transactional_session
 def get_transfer_requests_and_source_replicas(total_workers=0, worker_number=0, limit=None, activity=None, older_than=None, rses=None, schemes=None,
```
bari12: Add a general description here about the further workflow. Nothing huge, just to ilustrate the steps being done to build the chosen transfers.

bari12 (changes requested):
Thanks for the excellent PR Radu and sorry for the delay, just a few comments. (The preparer ones might actually be for @bziemons )
I went over it, back and forth for some time now, and I think it is correct. But it's a big PR, so be ready for hotfixes anyway 😄 
I especially like the new context classes.
Please still add a bit more documentation to the internal methods. No need to do it for each parameter, but just a rough description what the method does, to make it easier for others to understand. 
Also add some general description of the workflow. Although it is easier now, it is still not easy to understand. I think on the entry point of the `get_transfer_requests_and_source_replicas` you should add something.
Thanks a lot.

lib/rucio/core/transfer.py
```diff
@@ -702,6 +781,271 @@ def __rewrite_dest_url(dest_url, dest_sign_url, dest_scheme):
...
+            if request_id in reqs_no_source:
+                reqs_no_source.remove(request_id)
+            return
+
+    if ctx.is_tape_rse(dest_rse_id):
+        overwrite = False
+        transfer_dst_type = "TAPE"
+
+    # Get destination protocol
+    dest_protocol = ctx.protocol(dest_rse_id, dest_scheme, 'write')
```
bari12: write -> third_party_copy

lib/rucio/core/transfer.py
```diff
@@ -702,6 +781,271 @@ def __rewrite_dest_url(dest_url, dest_sign_url, dest_scheme):
     return dest_url
 
 
+def __prepare_source_dest_config(ctx, request_id, src_rse_id, dest_rse_id, source_scheme, dest_scheme, dict_attributes, activity,
+                                 scope, name, file_path, retry_count, reqs_only_tape_source, reqs_no_source, bring_online_local):
+
+    # Get source protocol
+    source_protocol = ctx.protocol(src_rse_id, source_scheme, 'read')
```
bari12: read -> third_party_copy 
Was this read before in the old code?

rcarpa: it was . it's just that ` ctx.protocol(src_rse_id, source_scheme, 'read')` doesn't correctly use this parameter . This issue was mentioned by @cserf in the previous refactoring. However, the previous refactoring tried to keep the same behavior as before. This time it's the moment to fix it. I'll make the change.

lib/rucio/core/transfer.py
```diff
@@ -702,6 +781,271 @@ def __rewrite_dest_url(dest_url, dest_sign_url, dest_scheme):
...
+
+        if best_candidate[0]['file_metadata']['src_type'] != candidate[0]['file_metadata']['src_type']:
+            # Tape and Disk sources must not be used at the same time.
+            continue
+
+        if len(candidate) > 1:
+            # do not add multihop sources
+            continue
+
+        if not set(best_candidate[0]['schemes']).intersection(candidate[0]['schemes']):
```
bari12: I have to think a bit more about this, but I am not 100% sure if this comparison is correct.
Because the `schemes` is already the compatible schemes of a specific chosen destination_scheme. But here you are iterating candidates (src-dest pairs) whose compatible_schemes is based on a different selected destination_scheme. I am not 100% sure if this is correct.

lib/rucio/core/request.py
rcarpa: Done with a quite generic docstring. Didn't have much inspiration for this one. The function does a couple of different things which are difficult to summarize into a docstring.

lib/rucio/core/request.py
rcarpa: done

lib/rucio/core/request.py
rcarpa: not sure what to do with this one. There is a single argument on a separate line. The function name is expressive enough that I don't have good ideas for a docstring.

lib/rucio/core/transfer.py
rcarpa: Would love it. But I think it's a python 3.7+ feature.

lib/rucio/core/transfer.py
rcarpa: merged this function into the calling context. I feel it is better this way.

lib/rucio/core/transfer.py
rcarpa: done

lib/rucio/core/transfer.py
rcarpa: done

lib/rucio/core/transfer.py
rcarpa: done

lib/rucio/core/transfer.py
rcarpa: done

lib/rucio/core/transfer.py
rcarpa: done. added a short workflow description at the end of the docstring

lib/rucio/core/transfer.py
rcarpa: Didn't think this through. The comment was there before the refactoring. Do you feel like it is wrong ?

rcarpa (commented):
Took into consideration all comments except one. Will have to dig more attentively into the multisource scheme merge. Will add a separate commit for that comment latter.

lib/rucio/core/request.py
bziemons: I mean, it updates the requests in the database based on its input essentially.. I still think the function name is descriptive enough

rcarpa:
I rebased and added a couple of simple cleanup commits. No changes to previous commits.

lib/rucio/core/transfer.py
```diff
@@ -721,381 +1058,239 @@ def get_transfer_requests_and_source_replicas(total_workers=0, worker_number=0,
...
+        filtered_sources = rws.sources
+        # Only keep allowed sources
+        if allowed_source_rses is not None:
+            filtered_sources = filter(lambda s: s.rse_id in allowed_source_rses, filtered_sources)
+        filtered_sources = filter(lambda s: s.rse_name is not None, filtered_sources)
+        # Ignore blacklisted RSEs
+        filtered_sources = filter(lambda s: s.rse_id not in unavailable_read_rse_ids, filtered_sources)
+        # Ignore tape sources if they are not desired
+        filtered_sources = list(filtered_sources)
+        had_tape_sources = len(filtered_sources) > 0
```
cserf, line 1258: I don't get it. From the name I understand what it should be (files that had tape sources), but I cannot relate it to `filtered_sources`

lib/rucio/core/transfer.py
```diff
@@ -637,6 +719,44 @@ def get_dsn(scope, name, dsn):
...
+        if source.rse_id not in inbound_distances:
+            # There is no direct connection between this source and the destination
+            continue
+
+        try:
+            matching_scheme = rsemgr.find_matching_scheme(
+                rse_settings_src=__load_rse_settings(rse_id=source.rse_id, session=session),
+                rse_settings_dest=__load_rse_settings(rse_id=dest_rse_id, session=session),
+                operation_src='third_party_copy',
+                operation_dest='third_party_copy',
```
cserf, line 953: Not to forget. We should use `third_party_copy_write`  and `third_party_copy_read` at some point

lib/rucio/core/transfer.py
```diff
@@ -702,6 +822,223 @@ def __rewrite_dest_url(dest_url, dest_sign_url, dest_scheme):
...
+        filtered_candidate_paths.append(path)
+    candidate_paths = filtered_candidate_paths
+
+    def __transfer_order_key(transfer_path):
+        # higher source_ranking first,
+        # on equal source_ranking, prefer DISK over TAPE
+        # on equal type, prefer lower distance_ranking
+        # on equal distance, prefer single hop
+        return (
+            - transfer_path[-1]['sources'][0][3],
```
cserf: So if there are 2 different transfer paths:
1 - A -> dest where A hosts the file but has lower ranking than B
2 - C -> B -> dest where B doesn't host the file but has higher ranking than A
In that case the multihop will be chosen ?

lib/rucio/core/transfer.py
```diff
@@ -721,381 +1058,239 @@ def get_transfer_requests_and_source_replicas(total_workers=0, worker_number=0,
...
+                      'md5': rws.md5,
+                      'state': 'C'}]
+            try:
+                add_replicas(rse_id=dest_rse_id,
+                             files=files,
+                             account=InternalAccount('root', vo=dest_rse_vo),
+                             ignore_availability=False,
+                             dataset_meta=None,
+                             session=session)
+            except Exception as error:
```
cserf, line 1355: In case of exception, the session is rollback if I'm not wrong

lib/rucio/core/transfer.py
rcarpa: yes, but source_ranking is only decreased at failed transfers. Meaning that A was tried in the past, but failed. So this current transfer is probably result of re-scheduling the failed one.

lib/rucio/core/transfer.py
rcarpa: @bari12 , any opinion ? It's not a behavior which changed. It looks like new code because github detects the diff wrongly. But this behavior is old (doesn't mean it's correct though)

lib/rucio/core/transfer.py
rcarpa: check the usage of `had_tape_sources` bellow and it will become clear. It's used like "if not filtered_sources and had_tape_sources .... then". Do you have a better idea on how to achieve this goal ?

rcarpa:
As agreed with @bari12 , added 5 additional commits. They fit into this refactoring conceptually; shouldn't introduce behavioral changes; but are important for latter work on generic transfertool handling. So bring them into this already big PR. Had to rebase due to conflict with https://github.com/rucio/rucio/pull/4651/.

lib/rucio/core/transfer.py
bari12: This is fine. It is the same as before. It actually checks if there already is a replica in `add_replicas`. For the case that it did not detect it at that point, but due to another session adding one exactly at this time, this will fire an integrity-constraint error, and the session will be rolled back. But also that should be fine.

bari12 (approved):

PR #4543, opened 2021-04-15
Core & Internals: archive information on dataset level. Closes #4385

### LINKED ISSUE #4385: Archive information on dataset level
Motivation
----------
In recent discussion with developers from ATLAS prodsys the use case was raised that it would be helpful to know if a dataset holds any files which are archives. Eg. via a simple `get_metadata(...)` query.

Right now, Rucio does not propagate this information to the dataset level. There are two fields associated with **file** dids:

- `constituent` which evaluates to `True` if a file is available in an archive (But can have conventional physical replicas as well)
- `is_archive` which evaluates to `True` if a file is an archive, and holds constituent file.

Modification
------------

I think we have two options:

1. Re-use the existing `is_archive` field and propagate this information from the content dids to the archive. Thus, if at least one content file in a dataset has the `is_archive` flag, the dataset will get it too. Needs to be removed again in case of detach operations.
2. Essentially the same as option (1) but with a new field. `contains_archives` - this would need a schema change.

Then this information could simply be queried with the dataset did.

rcarpa:
re-use the is_archive field on the dataset DIDs. If any file inside the
dataset is a an archive, the dataset will also be marked is_archive=True.

Also some refactoring tests to make addition of the new test easier.

### DIFF
--- a/lib/rucio/core/did.py
+++ b/lib/rucio/core/did.py
@@ -323,12 +323,23 @@ def __add_files_to_archive(scope, name, files, account, ignore_duplicate=False, 
     except IntegrityError as error:
         raise exception.RucioException(error.args)
 
-    # mark tha archive file as is_archive
-    session.query(models.DataIdentifier). \
+    archive_did = session.query(models.DataIdentifier). \
         filter(models.DataIdentifier.did_type == DIDType.FILE). \
         filter(models.DataIdentifier.scope == scope). \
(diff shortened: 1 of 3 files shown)


bari12 (dismissed):

bari12 (approved):

bari12:
I am merging this as a patch in the `1.25` release line.

PR #4553, opened 2021-04-19
Conveyor : Fix bug affecting http+srm protocol : Closes #4540

### LINKED ISSUE #4540: Conveyor : Fix bug affecting http+srm protocol
Motivation
----------
Fix bug affecting http+srm protocol in the conveyor

cserf:
Conveyor : Fix bug affecting http+srm protocol : Closes #4540

### DIFF
--- a/lib/rucio/core/transfer.py
+++ b/lib/rucio/core/transfer.py
@@ -671,8 +671,8 @@ def __rewrite_source_url(source_url, source_sign_url, dest_sign_url, source_sche
         source_url = re.sub('davs', 's3s', source_url)
         source_url = re.sub('https', 's3s', source_url)
 
-    if source_scheme == 'https+srm':
-        source_url = re.sub('srm', 'https+srm', source_url)
+    if source_url[:12] == 'srm+https://':
+        source_url = 'srm' + source_url[10:]
     return source_url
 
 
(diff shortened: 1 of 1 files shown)


lib/rucio/core/transfer.py
```diff
@@ -671,8 +671,8 @@ def __rewrite_source_url(source_url, source_sign_url, dest_sign_url, source_sche
         source_url = re.sub('davs', 's3s', source_url)
         source_url = re.sub('https', 's3s', source_url)
```
rcarpa: I believe all these previous lines potentially have the same issue. Why don't we fix them too ?

lib/rucio/core/transfer.py
```diff
@@ -664,15 +664,15 @@ def __rewrite_source_url(source_url, source_sign_url, dest_sign_url, source_sche
...
-        source_url = re.sub('https', 's3s', source_url)
-
-    if source_scheme == 'https+srm':
-        source_url = re.sub('srm', 'https+srm', source_url)
+    source_sign_url_map = {'gcs': 'gclouds', 's3': 's3s'}
+    if source_sign_url in source_sign_url_map:
+        if source_url[:7] == 'davs://':
+            source_url = source_sign_url_map[source_sign_url] + source_url[4:]
+        if source_url[:8] == 'https://':
+            source_url = source_sign_url_map[source_sign_url] + source_url[5:]
```
rcarpa, line 672: I find the new organisation with a map; and iteration over it; much too difficult to follow for the little benefit it bring. If we don't intend to add many more protocols in the future to this map, it's probably better to keep it explicit and easy.

rcarpa: That being said, I'm not against to merge it without changes.

rcarpa (approved):

Issue #4491, opened 2021-03-24
Allow specification of a default tombstone when registering replicas


bari12:
Motivation
----------
Currently, replicas are registered without a default tombstone, a rule needs to touch them first for them to be eligible for deletion. This creates issues in certain workflows.

Modification
------------
Allow the registration of replicas with a default tombstone. (Parameter name to be decided)

PR #4554, opened 2021-04-19
[WIP] Testing: rework test_replicas. #4491

rcarpa:
Until now, almost entire test_replicas was marked noparallel, which is unfortunate, because it contains 30 different tests; many not very fast. Rework the entire file and remove the noparallel mark when possible. 

Add minor changes to temp factories to cleanup bad replicas and allow to easily create 'srm' rses which are used all-over this test. 
The usage of rse_factory also allows to remove the dirty flag from most tests, because it takes care of cleanup after the test ends.

The patch is split into multiple commits for easier review. Otherwise github just shows that "everything" changed in test_replicas.py, 
but a squash is definitely needed when merging.


rcarpa:
The branch name doesn't follow the convention (path-* instead of patch-*). Close PR and reopen later with correct branch

PR #4559, opened 2021-04-20
Testing: rework temp factories #4385, #4491, #1431

rcarpa:
Extract and merge multiple changes/improvements done to temp factories while working on other issues into a independent PR. 

To be able to use those improvements in other unrelated PRs.

### DIFF
--- a/lib/rucio/tests/temp_factories.py
+++ b/lib/rucio/tests/temp_factories.py
@@ -31,6 +31,7 @@ from rucio.db.sqla import models
 from rucio.db.sqla.session import transactional_session
 from rucio.tests.common import file_generator
 from rucio.tests.common import rse_name_generator
+from rucio.db.sqla.constants import DIDType
 
 
 class TemporaryRSEFactory:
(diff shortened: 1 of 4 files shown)


lib/rucio/tests/temp_factories.py
```diff
@@ -108,27 +111,32 @@ def __cleanup_rses(self):
             # So running test in parallel results in some tests failing on foreign key errors.
             rse_core.del_rse(rse_id)
 
-    def _make_rse(self, scheme, protocol_impl, add_rse_kwargs):
+    def _make_rse(self, scheme, protocol_impl, parameters={}, add_rse_kwargs={}):
```
bziemons: I advice against using mutable values as default arguments

bziemons (dismissed):
looks good

lib/rucio/tests/temp_factories.py
rcarpa: I initially did it on purpose, because these default arguments are rapidly dereferenced. But you are right, better do it correctly. Updated the PR

bziemons (approved):

PR #4567, opened 2021-04-21
Transfers: modify argument passing in submitter. #4499

rcarpa:
This change was initially part of the bigger transfer
refactoring already waiting for review, but I need the same
changes for other issues now. So extract them separately.

- remove useless mock argument in submitter. Substitute it with 
"partition_wait_time" for the small part where it is relevant.
Tests work correctly without the mock argument. Is it used by 
somebody in production? why would it be?

- pass use_multihop as argument from submitter:
Retrieve it from config at the same place as other config values.
The main motivation is to allow easily set this argument in tests.
The commit changes the current behavior: the configuration will
only be retrieved once, on daemon startup; and the cached value is
expired slower. But this makes the behavior of use_multihop coherent
with other configuration values in submitter.

The configuration value is called "use_multihop", while internally
the arguments were "include_multihop". Homogenize and change to 
"use_multihop" everywhere.

### DIFF
--- a/lib/rucio/core/transfer.py
+++ b/lib/rucio/core/transfer.py
@@ -483,13 +483,13 @@ def update_transfer_state(external_host, transfer_id, state, session=None, logge
 
 
 @transactional_session
-def get_hops(source_rse_id, dest_rse_id, include_multihop=False, multihop_rses=None, limit_dest_schemes=None, session=None):
+def get_hops(source_rse_id, dest_rse_id, use_multihop=False, multihop_rses=None, limit_dest_schemes=None, session=None):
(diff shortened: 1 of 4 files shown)


rcarpa:
Re-checking the PR now, I have a huge doubt about it. Do I understand correctly that there are two kind of "configs" in rucio ? One in the database. The other not ?

bari12:
Yes two configs. The idea is to simplify this though and only have one interface, which then checks these in order.

rcarpa:
This pull request was incorrect and wrong from the start. Close it.

PR #4568, opened 2021-04-21
Conveyor : Replace incorrect https+srm scheme to srm+https : Fix #4506

### LINKED ISSUE #4506: Implement http-srm protocol
Motivation
----------
Some sites want to be able to enable HTTP as File Transfer Protocol instead of GridFTP. We need to implement a new http-srm protocol and add it to the list of compatible protocol for HTTP(S)/DAV(S)

vokac:


bari12:
This has been fixed in the meanwhile in another PR. Closing

PR #4569, opened 2021-04-22
Testing: TPC + UploadClient

### LINKED ISSUE #4190: Testing TPC transfers with Rucio
Motivation
----------
End to end tests for TPC workflows 

Modification
------------

### LINKED ISSUE #4451: Test UploadClient and update integration tests
Motivation
----------
Write tests for UploadClient that work with mock rse's and containerized rse's
Replace rsemgr.upload() with uploadclient in the integration tests.

maany:
----- Comments below have been Addressed ---- 
1. The TPC workflow uses fts-rest-cli to validate the functions performed by rucio-daemons.
For the fts-rest-cli to work correctly inside the rucio-dev container, we still need the following PR to be merged:
https://github.com/rucio/containers/pull/129

2. Remaining task: The TPC workflow needs to create an artefact for the fts_transfer's request-id (fts job id). The job-id will be used to validate that the transfer was actually TPC. This validation will be done by checking the fts transfer logs in the fts container. To realize this feature, I'll add a new GH Action workflow for TPC that shares and cleans up the artefacts. Converting this PR to a draft.

### DIFF
--- a/lib/rucio/common/utils.py
+++ b/lib/rucio/common/utils.py
@@ -77,7 +77,7 @@ from uuid import uuid4 as uuid
 from xml.etree import ElementTree
 
 import requests
-from six import string_types, text_type, PY3
+from six import string_types, text_type, binary_type, ensure_text, PY3
 
 from rucio.common.config import config_get
 from rucio.common.exception import MissingModuleException, InvalidType, InputValidationError, MetalinkJsonParsingError, RucioException
@@ -1242,9 +1242,9 @@ def run_cmd_process(cmd, timeout=3600):
         process.kill()
 
     stdout, stderr = process.communicate()
-    if isinstance(stdout, bytes):
-        stdout = stdout.decode()
-        stderr = stderr.decode()
+    if isinstance(stdout, binary_type):
+        stdout = ensure_text(stdout, errors='replace')
+        stderr = ensure_text(stderr, errors='replace')
     if not stderr:
         stderr = ''
     if not stdout:
--- a/lib/rucio/common/utils.py
+++ b/lib/rucio/common/utils.py
@@ -1241,6 +1241,9 @@ def run_cmd_process(cmd, timeout=3600):
         process.kill()
 
     stdout, stderr = process.communicate()
+    if isinstance(stdout, bytes):
+        stdout = stdout.decode()
+        stderr = stderr.decode()
     if not stderr:
         stderr = ''
     if not stdout:
--- a/lib/rucio/common/utils.py
+++ b/lib/rucio/common/utils.py
@@ -38,6 +38,7 @@
 # - Radu Carpa <radu.carpa@cern.ch>, 2021
 # - Anil Panta <47672624+panta-123@users.noreply.github.com>, 2021
 
+
 from __future__ import absolute_import, print_function
 
 try:
--- a/lib/rucio/tests/conftest.py
+++ b/lib/rucio/tests/conftest.py
@@ -123,11 +123,18 @@ def root_account(vo):
 def containerized_rses(rucio_client):
     """
     Detects if containerized rses for xrootd are available in the testing environment.
-    :return: An array of (rse_name, rse_id) tuples.
+    :return: A list of (rse_name, rse_id) tuples.
     """
-    rses = rucio_client.list_rses()
-    available_xrd_containerized_rses = [(rse_obj['rse'], rse_obj['id']) for rse_obj in rses if "xrd" in rse_obj['rse'].lower()]
-    return available_xrd_containerized_rses
+    rses = []
+    try:
+        xrd_rses = [x['rse'] for x in rucio_client.list_rses(rse_expression='test_container_xrd=True')]
+        xrd_rses = [rucio_client.get_rse(rse) for rse in xrd_rses]
+        xrd_containerized_rses = [(rse_obj['rse'], rse_obj['id']) for rse_obj in xrd_rses if "xrd" in rse_obj['rse'].lower()]
+        xrd_containerized_rses.sort()
+        rses.extend(xrd_containerized_rses)
+    except Exception:
+        traceback.print_exc()
+    return rses
 
 
 @pytest.fixture
--- a/lib/rucio/tests/conftest.py
+++ b/lib/rucio/tests/conftest.py
@@ -46,6 +46,12 @@ def replica_client():
 
 
 @pytest.fixture(scope='module')
+def rucio_client():
+    from rucio.client import Client
+    return Client()
+
+
+@pytest.fixture(scope='module')
 def did_client():
     from rucio.client.didclient import DIDClient
 
@@ -111,6 +117,17 @@ def root_account(vo):
     from rucio.common.types import InternalAccount
 
     return InternalAccount('root', vo=vo)
+
+
+@pytest.fixture(scope="module")
+def containerized_rses(rucio_client):
+    """
+    Detects if containerized rses for xrootd are available in the testing environment.
+    :return: An array of (rse_name, rse_id) tuples.
+    """
+    rses = rucio_client.list_rses()
+    available_xrd_containerized_rses = [(rse_obj['rse'], rse_obj['id']) for rse_obj in rses if "xrd" in rse_obj['rse'].lower()]
+    return available_xrd_containerized_rses
 
 
 @pytest.fixture
--- a/lib/rucio/tests/conftest.py
+++ b/lib/rucio/tests/conftest.py
@@ -16,7 +16,7 @@
 # Authors:
 # - Benedikt Ziemons <benedikt.ziemons@cern.ch>, 2020
 # - Radu Carpa <radu.carpa@cern.ch>, 2021
-# - Mayank Sharma <mayank.sharma@cern.ch> 2021
+# - Mayank Sharma <mayank.sharma@cern.ch>, 2021
 
 from __future__ import print_function
 
--- a/lib/rucio/tests/conftest.py
+++ b/lib/rucio/tests/conftest.py
@@ -23,8 +23,6 @@ from __future__ import print_function
 import traceback
 import pytest
 
-from rucio.common.exception import InvalidRSEExpression
-
 # local imports in the fixtures to make this file loadable in e.g. client tests
 
 
@@ -125,6 +123,8 @@ def containerized_rses(rucio_client):
     Detects if containerized rses for xrootd are available in the testing environment.
     :return: A list of (rse_name, rse_id) tuples.
     """
+    from rucio.common.exception import InvalidRSEExpression
+
     rses = []
     try:
         xrd_rses = [x['rse'] for x in rucio_client.list_rses(rse_expression='test_container_xrd=True')]
@@ -152,6 +152,14 @@ def did_factory(vo, mock_scope):
     from rucio.tests.temp_factories import TemporaryDidFactory
 
     with TemporaryDidFactory(vo=vo, default_scope=mock_scope) as factory:
+        yield factory
+
+
+@pytest.fixture
+def file_factory(tmp_path_factory):
+    from rucio.tests.temp_factories import TemporaryFileFactory
+
+    with TemporaryFileFactory(pytest_path_factory=tmp_path_factory) as factory:
         yield factory
 
 
--- a/lib/rucio/tests/conftest.py
+++ b/lib/rucio/tests/conftest.py
@@ -132,8 +132,9 @@ def containerized_rses(rucio_client):
         xrd_containerized_rses = [(rse_obj['rse'], rse_obj['id']) for rse_obj in xrd_rses if "xrd" in rse_obj['rse'].lower()]
         xrd_containerized_rses.sort()
         rses.extend(xrd_containerized_rses)
-    except InvalidRSEExpression as ex:
-        # containerized RSEs will not be available in non-containerized test environments
+    except InvalidRSEExpression as invalid_rse_expression:
+        print("{ex}. Note that containerized RSEs will not be available in non-containerized test environments"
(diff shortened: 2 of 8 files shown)


bari12:
In principle looks good to me, but there is a bunch of tests failing 😅

lib/rucio/common/utils.py
```diff
@@ -1238,6 +1239,9 @@ def run_cmd_process(cmd, timeout=3600):
         process.kill()
 
     stdout, stderr = process.communicate()
+    if isinstance(stdout, bytes):
+        stdout = stdout.decode()
+        stderr = stderr.decode()
```
bziemons: Please replace this with `six.ensure_text(…, errors='replace')` which is more resilient (since there could be non-decodable bytes) and it marks this as a Python 2/3 compatibility thing :)

lib/rucio/tests/conftest.py
```diff
@@ -98,13 +105,38 @@ def mock_scope(vo):
...
+    """
+    rses = []
+    try:
+        xrd_rses = [x['rse'] for x in rucio_client.list_rses(rse_expression='test_container_xrd=True')]
+        xrd_rses = [rucio_client.get_rse(rse) for rse in xrd_rses]
+        xrd_containerized_rses = [(rse_obj['rse'], rse_obj['id']) for rse_obj in xrd_rses if "xrd" in rse_obj['rse'].lower()]
+        xrd_containerized_rses.sort()
+        rses.extend(xrd_containerized_rses)
+    except Exception:
+        traceback.print_exc()
```
bziemons: this seems like some failures will be silently ignored

lib/rucio/tests/test_tpc.py
```diff
@@ -0,0 +1,168 @@
...
+from rucio.core.rule import add_rule
+from rucio.core.transfer import get_transfer_requests_and_source_replicas
+from rucio.common.utils import generate_uuid
+from rucio.daemons.judge.evaluator import re_evaluator
+from rucio.daemons.conveyor import submitter, poller, finisher
+from rucio.client.rseclient import RSEClient
+from rucio.client.ruleclient import RuleClient
+from rucio.common.utils import run_cmd_process
+
+# Fixtures
```
bziemons: I think this is not needed

lib/rucio/tests/test_tpc.py
```diff
@@ -0,0 +1,168 @@
...
+def rse2(containerized_rses):
+    if len(containerized_rses) > 1:
+        rse, rse_id = containerized_rses[1]
+        return {
+            'rse_name': rse,
+            'rse_id': rse_id,
+        }
+    return None
+
+# Utility functions
```
bziemons: please remove these hanging comments. I don't think they are respected in the future and will lead to confusion

lib/rucio/tests/test_tpc.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+def poll_fts_transfer_status(request_id, timeout=30):
+    rcode, out = run_cmd_process(f"/usr/bin/python2 /usr/bin/fts-rest-transfer-status -v -s https://fts:8446 {request_id}", timeout=timeout)
+    transfer_status = None
+    if rcode == 0:
+        transfer_status = re.search("Status: (.*)", out).group(1)
+    return transfer_status
+
+# TPC tests
```
bziemons: see above

lib/rucio/tests/test_tpc.py
```diff
@@ -0,0 +1,168 @@
...
+# Utility functions
+
+
+def check_url(pfn, hostname, path):
+    assert hostname in pfn
+    assert path in pfn
+
+
+def list_fts_transfer(timeout=30):
+    time_start = datetime.datetime.now().second
```
bziemons: `datetime.second` returns seconds of a clock (0-60), this will break the second you look at it. Please use `time.time()` for this specific use-case and timedelta (simply `date_end - date_start`) otherwise

lib/rucio/tests/test_tpc.py
```diff
@@ -0,0 +1,168 @@
+# -*- coding: utf-8 -*-
+# Copyright 2012-2021 CERN
```
bziemons: nothing big, but the start year seems off. Also in other files

lib/rucio/tests/test_tpc.py
```diff
@@ -0,0 +1,168 @@
...
+        rse, rse_id = containerized_rses[0]
+        return {
+            'rse_name': rse,
+            'rse_id': rse_id,
+        }
+    return None
+
+
+@pytest.fixture
+def rse2(containerized_rses):
```
bziemons: why not use the `containerized_rses` directly? This seems a bit like an overuse of fixtures

lib/rucio/tests/test_tpc.py
```diff
@@ -0,0 +1,168 @@
...
+    rule = rule_client.get_replication_rule(rule_id[0])
+    assert rule['locks_ok_cnt'] == 1
+    assert rule['locks_replicating_cnt'] == 0
+
+    if artifact is not None:
+        date = datetime.date.today().strftime("%Y-%m-%d")
+        with open(artifact, 'w') as artifact_file:
+            artifact_file.write(
+                f"/var/log/fts3/{date}/{rse1['rse_name'].lower()}__{rse2['rse_name'].lower()}/*__{fts_transfer_id}"
+            )
```
bziemons: this seems a bit weird, since it builds a file name with a date (which could be wrong) and just assumes that the file is there. I think there should at least be an existence check here.

bziemons (commented):
Generally looks pretty good. A few notes though.

bari12:
@maany are you still addressing Ben's comments? In general it looks good to me too.

lib/rucio/tests/test_tpc.py
maany: Aye! I've removed these fixtures for now. I wanted to create tuples of rse_pairs and parametrize the test_tpc with it. It was a bit complicated since fixtures (containerized_rses) cannot be called as regular functions in non-test functions and also pytest.mark.parameterize does not accept return values from fixtures. For the time being, rse_1 and rse_2 were placeholders for the values that would have been passed. I will create a separate PR for this.

lib/rucio/tests/test_tpc.py
maany: Hey @bziemons, this piece of code is executed inside the dev_rucio container. It writes the location of the expected fts transfer log file inside the artifact file (/tmp/test_tpc.artifact in this case) for the transfer created by the this test. This artifact is used in the 'Verify TPC transfers..' step of the workflow. If the file is not present, that step will fail:
https://github.com/rucio/rucio/pull/4569/files#diff-d917339141f036bb9e9a987412f09fc4c6780f8c0cd5ec1d89c4652c90032127R91

I agree that the date could be wrong, however, since the host machine of fts and rucio containers is the same, they would get the same value for date (unless the container timezones are configured differently, which is not the case as per their Dockerfile's). FTS uses a directory named ${date} to store the transfer logs. If FTS changes the directory structure for its logs, the test would fail as well. Open to suggestions for handling this. :)

maany:
> @maany are you still addressing Ben's comments? In general it looks good to me too.

should be good to go :)

lib/rucio/tests/test_tpc.py
bziemons: I get that, but I would suggest failing as early as possible which also helps debugging problems that may arise here; translating assumptions to asserts basically. The following would be my proposal:

```py
date = datetime.date.today().strftime("%Y-%m-%d")
fts_log_file = f"/var/log/fts3/{date}/{rse1_name.lower()}__{rse2_name.lower()}/*__{fts_transfer_id}"
assert os.path.exists(fts_log_file), 'FTS log file was not found'
with open(artifact, 'w') as artifact_file:
    artifact_file.write(fts_log_file)
```

maany: Hey @bziemons, thanks for the comments :). I think I need to elaborate since the suggested change
```
assert os.path.exists(fts_log_file), 'FTS log file was not found'
```

will always fail as the `fts_log_file` does not exist inside the `dev_rucio_1` container.  

```
[root@rucio-devmaany rucio]# docker exec -it dev_rucio_1 bash
[root@dc95631ee443 log]# ls -la /var/log/
total 368
drwxr-xr-x 1 root root     32 Apr 28 07:49 .
drwxr-xr-x 1 root root     28 Apr 28 07:47 ..
-rw------- 1 root utmp      0 Nov 13  2020 btmp
-rw-r--r-- 1 root root    193 Nov 13  2020 grubby_prune_debug
drwx------ 1 root root     41 May 27 07:50 httpd
-rw-r--r-- 1 root root 292000 Apr 28 07:47 lastlog
drwxrwxrwx 1 root root     66 May 27 07:50 rucio
-rw------- 1 root root  64000 Apr 28 07:47 tallylog
-rw-rw-r-- 1 root utmp      0 Nov 13  2020 wtmp
-rw------- 1 root root  12235 Apr 28 07:47 yum.log
```

It only exists inside the `dev_fts_1` container. 

```
[root@rucio-devmaany rucio]# docker exec -it dev_fts_1 bash
[root@dc95631ee443 /]# ls -la /var/log/
total 372
drwxr-xr-x 1 root root     47 Apr 28 07:18 .
drwxr-xr-x 1 root root     39 Apr 28 07:17 ..
-rw------- 1 root utmp      0 Nov 13  2020 btmp
drwxr-xr-x 1 fts3 root     46 May 27 09:17 fts3
drwxr-xr-x 1 fts3 fts3     26 Apr 28 07:18 fts3rest
-rw-r--r-- 1 root root    193 Nov 13  2020 grubby_prune_debug
drwx------ 1 root root     94 May 27 07:50 httpd
-rw-r--r-- 1 root root 292000 Apr 28 07:18 lastlog
-rw------- 1 root root  64000 Apr 28 07:18 tallylog
-rw-rw-r-- 1 root utmp      0 Nov 13  2020 wtmp
-rw------- 1 root root  14503 Apr 28 07:18 yum.log
```

It is created inside the `dev_fts_1` container when `test_tpc` gets executed inside the `dev_rucio_1` container. 

Since all tests run inside the `dev_rucio_1` container:

https://github.com/rucio/rucio/blob/8a0035701af3e79b905dbbed914607ea46983d20/.github/workflows/integration_tests.yml#L90-L92

the test environment cannot access the file system inside the `dev_fts_1` container. 

**Unless there is some way to access the filesystem of `dev_fts_1` container from inside of the `dev_rucio_1` container, I believe we cannot make this assertion in the `dev_rucio_1` container.** 

Since there is no straightforward way to do ^,  I export the artifact file from the 'dev_rucio_1' container that contains a string. This string is the location of the fts transfer log file inside the `dev_fts_1` container that was generated during the test.

The artifact file is then used in a subsequent step of integration_test.yml to check for the existence and contents of the fts transfer log file inside the `dev_fts_1` container:

https://github.com/rucio/rucio/blob/8a0035701af3e79b905dbbed914607ea46983d20/.github/workflows/integration_tests.yml#L93-L103 

The goal of the artifact mechanism is allow tests from 'dev_rucio_1' container to export information that can be used to perform additional checks on the states of the other containers and ensure that overall system level picture (xrd, fts, minio, ...) is looking good.

Let me know if we can get on a call and clarify this :)

bziemons:
Since there is no good way to share the fts3 log directory between the two containers with compose, all my comments are resolved :) There is still the standing issue with the failing test. @rcarpa any idea on that?

maany:
@bziemons already synced up with @rcarpa . Fix is on the way ;)

maany:
@bziemons The CI was failing earlier because `test_upload.py` and `test_download.py` were using and **deleting** the same `basedir` for files downloaded via the `DownloadClient`.  This led to a race condition, making the 2 tests flaky when executed in parallel. Following @rcarpa's suggestion on using [tmpdir fixtures](https://docs.pytest.org/en/6.2.x/tmpdir.html) available in pytest, I've added a TemporaryFileFactory 

https://github.com/rucio/rucio/blob/8adf0df3c9f5c8f86e55f2e5fb84d5cb98319f81/lib/rucio/tests/temp_factories.py#L301-L306

It can also generate and cleanup temp_files outside of pytest's `tmp_path_factory` managed base directories. 

Once we can migrate enough of the existing tests to use it, we could potentially deprecate the following functions:

https://github.com/rucio/rucio/blob/f1159fbf1e6002715a77b613d35d037c55082803/lib/rucio/tests/common.py#L75-L98

lib/rucio/tests/test_tpc.py
```diff
@@ -0,0 +1,144 @@
...
+
+def poll_fts_transfer_status(request_id, timeout=30):
+    rcode, out = run_cmd_process(f"/usr/bin/python2 /usr/bin/fts-rest-transfer-status -v -s https://fts:8446 {request_id}", timeout=timeout)
+    transfer_status = None
+    if rcode == 0:
+        transfer_status = re.search("Status: (.*)", out).group(1)
+    return transfer_status
+
+
+@pytest.mark.dirty(reason="Creates artifact /tmp/test_tpc.artifact in dev_rucio_1 container")
```
rcarpa, line 87: should mark it noparallel, multiple submiters cannot run in parallel because the workload is partitioned by the number of submitters. Otherwise, tests will fail randomly. See conveyor_submitter tests for the decorator to add.

rcarpa (commented):
Nice work.

lib/rucio/tests/test_tpc.py
maany: Gotcha... and done ;)

maany:
@bari12 @bziemons this is ready to be merged unless there are some further comments.

bziemons (approved):
lgtm

bari12 (approved):
👍

PR #4573, opened 2021-04-23
Testing: allow to override core.config and caches per test

rcarpa:
Introduce two fixtures which can take parameters. 

The override only works on core function calls because it acts by monkey-patching using mock.patch().


bziemons (approved):
Awesome

bari12:
Very nice 👍

PR #4577, opened 2021-04-26
Core & Internals: delete unused replicas functions. #4491

rcarpa:
- get_sum_count_being_deleted
- update_replica_lock_counter
- update_replicas_paths
- list_unlocked_replicas
- mark_unlocked_replicas


bari12:
LGTM, I also double checked with the probes but can't find these anywhere. I think this is partly reaper1 stuff and partly methods which just were never used. Second 👁️ please @cserf 😄

cserf:
I was almost sure that `update_replicas_paths` was used somewhere for the non-deterministic RSEs, but it doesn't seem to be the case. So LGTM

cserf (approved):

Issue #4575, opened 2021-04-23
Allow bulk deletion of files transferred by globus


maatthias:
Motivation
----------
Currently a single task per file is submitted to globus for deletion.

Modification
------------
Modify the reaper2 daemon to capture a group of files before submitting the deletion request.

bari12:
Close for the moment, will re-open once needed.

PR #4581, opened 2021-04-27
Deletion: allow bulk deletion of files transferred by globus #4575

maatthias:
- added feature to reaper2 daemon to submit a bulk delete job to globus
- daemon takes optional run-time option `--bulk`
- renamed `transfertool/globusLibrary.py` -> `transfertool/globus_library.py`


bari12:
Before going into review details, this needs a rebase to get the tests working ;-)

lib/rucio/rse/protocols/globus.py
```diff
@@ -223,6 +225,38 @@ def delete(self, path):
             print('delete_task not accepted by Globus')
             print('delete_response: %s' % delete_response)
 
+    def bulk_delete(self, replicas):
```
rcarpa: Could you please extend the abstract class in lib/rucio/rse/protocols/protocol.py  with this additional contract method and its documentation.

lib/rucio/daemons/reaper/reaper2.py
```diff
@@ -177,7 +180,10 @@ def delete_from_storage(replicas, prot, rse_info, staging_areas, auto_exclude_th
                     # sign the URL if necessary
                     if prot.attributes['scheme'] == 'https' and rse_info['sign_url'] is not None:
                         pfn = get_signed_url(rse_id, rse_info['sign_url'], 'delete', pfn)
-                    prot.delete(pfn)
+                    if not bulk:
+                        prot.delete(pfn)
+                    else:
+                        replicas_to_bulk_delete.append(replica)
```
rcarpa: I don't think we must act on replicas here. There seem to be an existing separation between the core (working on replicas), and protocols (acting on pfns). You only need pfns anyway inside the protocol class.

lib/rucio/daemons/reaper/reaper2.py
```diff
@@ -628,7 +639,7 @@ def stop(signum=None, frame=None):
     GRACEFUL_STOP.set()
 
 
-def run(threads=1, chunk_size=100, once=False, greedy=False, rses=None, scheme=None, exclude_rses=None, include_rses=None, vos=None, delay_seconds=0, sleep_time=60, auto_exclude_threshold=100, auto_exclude_timeout=600):
+def run(threads=1, chunk_size=100, once=False, greedy=False, rses=None, scheme=None, exclude_rses=None, include_rses=None, vos=None, delay_seconds=0, sleep_time=60, auto_exclude_threshold=100, auto_exclude_timeout=600, bulk=False):
```
rcarpa, line 642: A test verifying that the new argument  works as expected would be welcomed. You'll probably need to implement a dummy "bulk_delete()" implementation in  lib/rucio/rse/protocols/mock.py. You could start from one of the simpler tests in lib/rucio/tests/test_reaper2.py

lib/rucio/rse/protocols/globus.py
maatthias: Yes good idea.  I've extended the abstract base with `bulk_delete` method.

lib/rucio/daemons/reaper/reaper2.py
maatthias: Indeed.  I've pushed a change that only handles pfns instead of a list of replicas.

lib/rucio/daemons/reaper/reaper2.py
maatthias: To truly mock test this there would need to be some access to a globus service to initiate a transfer then delete.  I will take a look to see how far we could get with this functionally.

rcarpa: Truly mocking may be a nice bonus, but is probably complex. However, ensuring that the generic(non-protocol-specific) code inside lib/rucio/daemons/reaper/reaper2.py works as expected should be quite easy to verify.

bin/rucio-reaper2
```diff
@@ -60,6 +61,8 @@ def get_parser():
                         help='Number of service unavailable exceptions after which the RSE gets temporarily excluded.')
     parser.add_argument("--auto_exclude_timeout", action="store", default=600, type=int,
                         help='Timeout for temporarily excluded RSEs.')
+    parser.add_argument("--bulk", action="store_true", default=False,
```
mlassnig, line 64: Is this only valid for Globus deletion? If so, maybe this should be renamed to something like --globus-bulk, or alternatively change the comment "Enable bulk deletion in Globus". Because otherwise I would even go as far as recommending this to be set to default=true.

mlassnig: Also, might be worth to actually make this an integer instead of a bool, to set the bulk size?

maatthias: Yes this is to bundle multiple file deletes into 1 async job for Globus.  I like the idea to make this an integer because that removes the dependency on `chunk_size`.  Your idea to make this generic is also a good one.  Another service could easily use the reaper to do the same.

bari12: I wonder if we should keep this or instead just assume that with globus, everything is bulk up to the `--_chunk_size`; Opinions?

lib/rucio/daemons/reaper/reaper2.py
```diff
@@ -145,11 +146,13 @@ def get_rses_to_process(rses, include_rses, exclude_rses, vos):
     return rses
 
 
-def delete_from_storage(replicas, prot, rse_info, staging_areas, auto_exclude_threshold, logger=logging.log):
+def delete_from_storage(replicas, prot, rse_info, staging_areas, auto_exclude_threshold, bulk, logger=logging.log):
```
bari12, line 149: See comment above, if we assume everything with globus is bulk, we don't need that part.

lib/rucio/daemons/reaper/reaper2.py
bari12: As discussed, would be good to have a mock part here (Not an actual integration test, just mock)

lib/rucio/rse/protocols/protocol.py
```diff
@@ -492,6 +492,14 @@ def delete(self, path):
         """
         raise NotImplementedError
 
+    def bulk_delete(self, pfns):
```
bari12, line 495: I am not sure if we need to add this to the protocol interface, since it is really something more special to globus.

maatthias:
Apologies @bari12 @rcarpa @mlassnig I deleted this branch in my rebase.  Which I guess results in deleted pull.  Sorry to complicate the conversation history.  I believe all above is addressed in new pull https://github.com/rucio/rucio/pull/4676

PR #4582, opened 2021-04-27
Transfers: minor daemon changes and add more conveyor tests. #4491

rcarpa:
Preparation for the tombstone refactoring. But these tests will be useful in any case. So send them in a separate PR.


etc/certs/generate.sh
```diff
@@ -37,6 +37,11 @@ openssl req -new -newkey rsa:2048 -nodes -keyout hostcert_xrd3.key.pem -subj "/C
 openssl x509 -req -days 9999 -CAcreateserial -extfile <(printf "subjectAltName=DNS:xrd3,DNS:localhost,DNS:xrd3.default.svc.cluster.local") -in hostcert_xrd3.csr -CA rucio_ca.pem -CAkey rucio_ca.key.pem -out hostcert_xrd3.pem -passin env:PASSPHRASE
 
 
+# XrootD Server 3
```
bziemons: 4?

rcarpa: Absolutely right. Fixed; also fixed in the DNS field a couple of lines lower.

bari12 (approved):

PR #4584, opened 2021-04-28
Core & Internals: fix dictionary changed while iterating. Closes #4583

### LINKED ISSUE #4583: Python 3 bug in core list_replicas
Motivation
----------

Found the following bug while trying to get Rucio server running on Python 3.8 (and this might also affect Python 3 versions below 3.8):

```
  File "/usr/local/lib/python3.8/site-packages/rucio/web/rest/flaskapi/v1/common.py", line 64, in dispatch_request
    return super(ErrorHandlingMethodView, self).dispatch_request(*args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/flask/views.py", line 163, in dispatch_request
    return meth(*args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/rucio/web/rest/flaskapi/v1/common.py", line 137, in decorated
    return f(*args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/rucio/web/rest/flaskapi/v1/replicas.py", line 412, in post
    return try_stream(generate(request_id=request.environ.get('request_id'),
  File "/usr/local/lib/python3.8/site-packages/rucio/web/rest/flaskapi/v1/common.py", line 184, in try_stream
    peek = next(it)
  File "/usr/local/lib/python3.8/site-packages/rucio/web/rest/flaskapi/v1/replicas.py", line 340, in generate
    for rfile in list_replicas(dids=dids, schemes=schemes,
  File "/usr/local/lib/python3.8/site-packages/rucio/api/replica.py", line 186, in list_replicas
    for rep in replicas:
  File "/usr/local/lib/python3.8/site-packages/rucio/db/sqla/session.py", line 339, in new_funct
    for row in function(*args, **kwargs):
  File "/usr/local/lib/python3.8/site-packages/rucio/core/replica.py", line 1271, in list_replicas
    for f in _list_replicas(dataset_clause, file_clause, state_clause, pfns,
  File "/usr/local/lib/python3.8/site-packages/rucio/core/replica.py", line 936, in _list_replicas
    for tmp_archive in archive['pfns']:
RuntimeError: dictionary keys changed during iteration
```

Modification
------------

Fix the corresponding code: https://github.com/rucio/rucio/blob/8de8259d74204221b4cce5b4eb4e3dab55525813/lib/rucio/core/replica.py#L936-L953

rcarpa:
In the case when we append the xrdcl.uzip query string, we just move
the existing value to a new key. It seems safe to just iterate over
a copy of the old keys.

Moving the two lines of code after the "if" doesn't change anything,
but it is more clear this way that adding domain = zip is done on
the new key(pfn with query string).

### DIFF
--- a/lib/rucio/core/replica.py
+++ b/lib/rucio/core/replica.py
@@ -933,9 +933,7 @@ def _list_replicas(dataset_clause, file_clause, state_clause, show_pfns,
                         archive['md5'] = constituent_meta['md5']
                         archive['bytes'] = constituent_meta['bytes']
 
-                        for tmp_archive in archive['pfns']:
-                            # declare the constituent as being in a zip file
-                            archive['pfns'][tmp_archive]['domain'] = 'zip'
+                        for tmp_archive in list(archive['pfns']):
                             # at this point we don't know the protocol of the parent, so we have to peek
                             if tmp_archive.startswith('root://') and 'xrdcl.unzip' not in tmp_archive:
                                 # use direct addressable path for root
@@ -946,6 +944,9 @@ def _list_replicas(dataset_clause, file_clause, state_clause, show_pfns,
                                 archive['pfns'][tmp_archive]['priority'] -= 99
                             if 'xrdcl.unzip' not in tmp_archive:
                                 archive['pfns'][tmp_archive]['client_extract'] = True
+
+                            # declare the constituent as being in a zip file
+                            archive['pfns'][tmp_archive]['domain'] = 'zip'
 
                             pfns.append((tmp_archive, 'zip',
                                          archive['pfns'][tmp_archive]['priority'],


mlassnig:
while this is semantically correct, does this really fix the underlying problem?

rcarpa:
My understanding is that it will not trigger the error. While it doesn't change the behavior. 
Do I miss something?

bziemons:
> while this is semantically correct, does this really fix the underlying problem?

it should. The `list(archive['pfns'])` creates a copy of the keys (plus the `str` is immutable) and therefore `archive['pfns'].pop(tmp_archive)` does not affect the iteration anymore. I can run it through my tests in a minute

mlassnig:
ok, good

bziemons (approved):
Tests liked it as well

mlassnig (approved):

PR #4587, opened 2021-04-28
Testing: Disable parallel testing on mysql8

bziemons:
:parrot:


rcarpa (approved):
At this stage, we can probably put "if database =~ postgresql; then parallel; else not parallel;" :trollface:

Issue #3431, opened 2020-03-30
list-file-repicas --all-states should show the states


dchristidis:
Motivation
----------

The WebUI shows this information. Without it, `--all-states` isn’t that helpful.

Modification
------------

Add an additional column when `--all-states` is used.

PR #4589, opened 2021-04-29
Core & Internals: blacklisted rses in list_replicas. Closes #4353

### LINKED ISSUE #4353: Core list_replicas() does not make use of ignore_availability
Motivation
----------

If an RSE is excluded for reading, the the download client should not attempt to download from that RSE.

The REST layer already does this, but once `ignore_availability` propagates to the core, it is not actually used.
https://github.com/rucio/rucio/blob/960855d6e86f4e38ebe584b1d26e65751d52a210/lib/rucio/web/rest/webpy/v1/replica.py#L336
https://github.com/rucio/rucio/blob/960855d6e86f4e38ebe584b1d26e65751d52a210/lib/rucio/core/replica.py#L1164

Modification
------------

rcarpa:
The default in the clients is already to request ignoring the
replicas.
Add the missing pieces in core to respect this parameter.

### DIFF
--- a/lib/rucio/core/replica.py
+++ b/lib/rucio/core/replica.py
@@ -704,7 +704,7 @@ def _resolve_dids(dids, unavailable, ignore_availability, all_states, resolve_ar
     return file_clause, dataset_clause, state_clause, files, constituents
 
 
-def _list_replicas_for_datasets(dataset_clause, state_clause, rse_clause, updated_after, session):
+def _list_replicas_for_datasets(dataset_clause, state_clause, rse_clause, ignore_availability, updated_after, session):
     """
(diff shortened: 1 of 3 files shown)


mlassnig (approved):
nice!

bari12 (approved):

dchristidis (commented):
Thanks Radu. I am under the impression that we need to pair this with a patch to the client; it would be a terrible regression if `rucio list-file-replicas` does not return the ones on excluded RSEs. I think the Web UI is already fine.

PS: Users might then complain that a replica appears as available with `list-replicas` but can’t be downloaded. I think I will consider modifying #3431 to have this additional information even without `--all-states`.

Issue #4436, opened 2021-03-16
Make the time before tombstone is set configurable


KatyEllis:
Motivation
----------
Recently CMS had a problem with multihop transfers to CTA. Because of queuing on the tape buffer, the files on EOS were being deleted by Rucio before they had a chance to get to tape. In this case, the config on FTS was incorrect and the system got into a bad state, but it still might be that in the future we want more control over the deletion time of files multi hopping through EOS. The EOS disk files (the middle hop) have a tombstone set 2 hours in the future. This may not be long enough for CTA to keep up when we are taking data.

Modification
------------
Make the tombstone time configurable (per RSE?) so we can find an appropriate value. During data taking we may find we need to change this value urgently!

Issue #4188, opened 2020-12-07
Multihop intermediate replica tombstone


nsmith-:
Motivation
----------
From https://github.com/rucio/rucio/pull/3064/files#r343021930 it looks like, in addition to the transfer source replica protection, an additional protection is placed on files likely to be created by a multihop transfer in setting the tombstone to (now+2h). I think these intermediate replicas should have epoch tombstone so they are cleaned up immediately rather than flushing popular but unlocked replicas from the intermediate disk RSE.

Modification
------------
There may be other uses of `update_replicas_states` besides multihop that would hit this code, so I am not sure if we want to set the tombstone to epoch always. Perhaps somewhere else we can set this? Or we can add a flag to indicate the replica is being used for multihop?

nsmith-:
Somewhat related, I think the tombstone set on a replica that was made at the end of a transfer in https://github.com/rucio/rucio/blob/master/lib/rucio/daemons/conveyor/finisher.py#L462 should also be epoch.

bari12:
You could also argue since the replicas was used, for the multihop, it is a valid access. But I get your point that, since the data is migrated to tape, it is just less likely that a workflow will re-access it.
We would carefully have to check if we could end up in some kind of race-condition with the epoche tombstone though, thus the intermediate replica is removed before the second hop happens. For normal transfers, this is protected, but for MH everything is a bit different. @cserf what do you think?

cserf:
See also https://github.com/rucio/rucio/issues/4436
I would propose to have the tombstone configurable

cserf:
Proposal:
When the finisher marks the intermediate replica as AVAILABLE in https://github.com/rucio/rucio/blob/398df650f9cd75b3befac60d3b71e031c39061da/lib/rucio/core/replica.py#L1900 , it will check what tombstone should be set using the tombstone defined in the config table. Adding a protection against race condition with the reaper(2) in case the tombstone is set to Epoch would require a change in the logic of `list_and_mark_unlocked_replicas`. I need to think what is the best solution.

bari12:
Yes, as we have seen setting an epoch tombstone here would probably not work correctly, since it will remove the replica way to soon (See #4436)
Having a configurable tombstone in the configtable would solve both #4436 and #4188 (You can set a much shorter one, if you are sure that this works in your deployment)

nsmith-:
I'm confused, are you saying there is a race condition in the logic and you can configure how risky you want to be with respect to it?

[...]

PR #4593, opened 2021-04-30
Core & Internals: rework tombstone handling. Closes #4491, #4436 and #4188

### LINKED ISSUE #4491: Allow specification of a default tombstone when registering replicas

rcarpa:
Always set a default tombstone on any replica creation. Introduce two
rse attributes which allow to customize the behavior per rse:
- One allows to define the normal replica creation tombstone
- The other used to configure the multihop temporary replica tombstone

As a tombstone will now be added when the intermediate multihop replica
is created, remove the redundant protection which was touching the
tombstone in the finisher.
Add a submitter test which ensures that multihop sources are created;
and that the multihop request are created with a tombstone.


rcarpa:
The first commit was submitted as a separate PR (#4577)

rcarpa:
Closing this PR to re-submit it again as "patch"

PR #4594, opened 2021-04-30
Core & Internals: rework tombstone handling. Closes #4491, #4436 and #4188

### LINKED ISSUE #4491: Allow specification of a default tombstone when registering replicas

rcarpa:
Allow setting a default tombstone on any replica creation. Introduce two
rse attributes which allow to customize the behavior per rse:
- One allows to define the normal replica creation tombstone
(defaults to 0: "no tombstone")
- The other used to configure the multihop temporary replica tombstone
(defaults to 2 hours in the future)
Use a negative value to set the epoch tombstone. Or 0 to explicitly
set a null tombstone.

Protect replicas which are used as sources from deletion even if their
tombstone is set to epoch. Add an integration test which verifies this
behavior.
Add a submitter test ensuring that multihop sources are created;
and that the temp multihop replicas are created with a tombstone.

Remove the redundant behavior which touches the tombstone in the
finisher. The 2 remaining protections for intermediate multihop replicas
are now: 1) entries in the source table; 2) the default multihop
tombstone delay of 2 hours into the future.

### DIFF
--- a/lib/rucio/core/replica.py
+++ b/lib/rucio/core/replica.py
@@ -1406,6 +1406,26 @@ def __bulk_add_file_dids(files, account, dataset_meta=None, session=None):
     return new_files + available_files
 
 
+def tombstone_from_delay(tombstone_delay):
+    # Tolerate None for tombstone_delay
+    if not tombstone_delay:
+        return None
+
+    if not isinstance(tombstone_delay, timedelta):
+        try:
+            tombstone_delay = timedelta(seconds=int(tombstone_delay))
+        except ValueError:
+            return None
+
+    if not tombstone_delay:
+        return None
+
+    if tombstone_delay < timedelta(0):
+        return datetime(1970, 1, 1)
+
+    return datetime.utcnow() + tombstone_delay
+
+
 @transactional_session
 def __bulk_add_replicas(rse_id, files, account, session=None):
     """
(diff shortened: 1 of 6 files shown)


rcarpa:
The first commit was submitted as a separate PR (#4577)

mlassnig (commented):
looks good from a code point of view. @bari12 can you check for semantics as well pls.

bziemons (dismissed):
lgtm

bari12 (changes requested):
I think there is a misconception here 😄 

The default_tombstone should only be applied for RSEs which have the `tombstone_delay` set. For all the other RSEs, current behaviour should be kept. As far as as I can see here, we would put the 2 week tombstone on everything which does not have a `tombstone_delay` set.

For the intermediate replicas of multihops I think we need to do it a bit different. There we should have a global (config table) variable which specifies the tombstone delay.

rcarpa:
I updated thing to not set any default tombstone on normal replica creation. And to set it two hours into the future for the multihop tombstone. I'll work on a separate commit now to make the multihop_tombstone_delay also available as a config variable, which will be used for RSEs which don't have it set per-rse.

bari12 (approved):
Note: for the more strict replica deletion selection (which filters out sources) we have to see if this affects deletion too negatively. Perhaps we have to switch here to only exclude sources we are actually using.

bari12:
Merging as a feature (For 1.26)

PR #4597, opened 2021-05-02
WebUI: R2D2 wrongly creates a container Fix #583

### LINKED ISSUE #583: R2D2 wrongly creates a container if the list of dids is containing only files

rak108:
As mentioned by @tbeerman, Three cases that should be handled:

- all DIDs are files -> create dataset
- all DIDs are datasets -> create container
- DIDs are mixed -> create dataset for files and add it together with the remaining datasets to a container.

please do review and let me know if there are any changes necessary!

### DIFF
--- a/lib/rucio/web/ui/static/request_rule.js
+++ b/lib/rucio/web/ui/static/request_rule.js
@@ -786,6 +786,99 @@ create_request_container = function(dids) {
     return [{'scope': scope, 'name': name}]
 }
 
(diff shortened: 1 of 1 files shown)


tbeerman (approved):
LGTM

PR #4598, opened 2021-05-03
Transfers: reset multihop sources after checking tape conditions

rcarpa:
Otherwise we can end with an empty sources dict. A multihop source is
discarded, but no other source is added because it's a tape or has worse ranking.


bari12:
Closes #4516

bari12 (approved):

PR #4600, opened 2021-05-03
Clients: bring python 2.7 support back to bin/rucio

rcarpa:
fixes following two errors:
- File "bin/rucio", line 212
  SyntaxError: invalid syntax
- File "bin/rucio", line 25
  SyntaxError: Non-ASCII character '\xc3' in file bin/rucio on line 25


bziemons (approved):
lgtm

PR #4611, opened 2021-05-05
Core : Non optimal rule reevaluation when attach_dids_to_dids is used with ignore_duplicate option : Closes #4178

### LINKED ISSUE #4178: Non optimal rule reevaluation when attach_dids_to_dids is used with ignore_duplicate option
Motivation
----------
`attach_dids_to_dids` has  an `ignore_duplicate` option to not trigger an exception if an attempt to reattach the same DID is done. But if this option is used, it will always insert a row into the `updated_dids` table.

cserf:
Core : Non optimal rule reevaluation when attach_dids_to_dids is used with ignore_duplicate option : Closes #4178

### DIFF
--- a/lib/rucio/core/did.py
+++ b/lib/rucio/core/did.py
@@ -356,13 +356,14 @@ def __add_files_to_dataset(scope, name, files, account, rse_id, ignore_duplicate
     """
     Add files to dataset.
 
-    :param scope: The scope name.
-    :param name: The data identifier name.
-    :param files: .
-    :param account: The account owner.
-    :param rse_id: The RSE id for the replicas.
-    :param ignore_duplicate: If True, ignore duplicate entries.
-    :param session: The database session in use.
+    :param scope:              The scope name.
+    :param name:               The data identifier name.
+    :param files:              The list of files.
+    :param account:            The account owner.
+    :param rse_id:             The RSE id for the replicas.
+    :param ignore_duplicate:   If True, ignore duplicate entries.
+    :param session:            The database session in use.
+    :returns:                  List of files attached (excluding the ones that were already attached to the dataset).
     """
     # Get metadata from dataset
     try:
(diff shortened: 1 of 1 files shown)


lib/rucio/core/did.py
```diff
@@ -577,7 +577,8 @@ def attach_dids_to_dids(attachments, account, ignore_duplicate=False, session=No
         except NoResultFound:
             raise exception.DataIdentifierNotFound("Data identifier '%s:%s' not found" % (attachment['scope'], attachment['name']))
 
-        session.bulk_insert_mappings(models.UpdatedDID, parent_dids)
+    parent_dids = [dict(tup) for tup in set(tuple(dictionary.items()) for dictionary in parent_dids)]
```
mlassnig, line 595: oof, please add comment what this does :-)

bari12: Quick explenation would be good.
But also, is this enough @cserf AFAIK this only filters parent duplicates, so that they are just added once, but it would still produce an updated_did for the parent, even if all the files added are just duplicates?

cserf: You're right. I thought I implemented something to skip the insertion if all the files are duplicate, but it doesn't seem to be the case. Will do it.

lib/rucio/core/did.py
```diff
@@ -422,6 +422,7 @@ def __add_files_to_dataset(scope, name, files, account, rse_id, ignore_duplicate
     try:
         contents and session.bulk_insert_mappings(models.DataIdentifierAssociation, contents)
         session.flush()
+        return contents
```
bari12, line 426: Add `:return: to the method description.

bari12 (dismissed):
I think this is GTG. Just add the return description to the method.

lib/rucio/core/did.py
cserf: Done

lib/rucio/core/did.py
```diff
@@ -567,28 +569,31 @@ def attach_dids_to_dids(attachments, account, ignore_duplicate=False, session=No
                 raise exception.UnsupportedOperation("Data identifier '%(scope)s:%(name)s' is closed" % attachment)
 
             elif parent_did.did_type == DIDType.DATASET:
-                __add_files_to_dataset(scope=attachment['scope'], name=attachment['name'],
-                                       files=attachment['dids'], account=account,
-                                       ignore_duplicate=ignore_duplicate,
-                                       rse_id=attachment.get('rse_id'),
-                                       session=session)
+                cont = __add_files_to_dataset(scope=attachment['scope'], name=attachment['name'],
```
mlassnig, line 572: shouldn't this be `cont.append()` ?

cserf: Nope, see line 584 + removal  of duplicates line 595

mlassnig (approved):

PR #4613, opened 2021-05-07
Deletion: change constituent removal batch handling

rcarpa:
Instead of constructing the list of all constituents, which can
grow very big, delete them as soon as it reaches the batch size.


mlassnig (approved):
nice.

PR #4615, opened 2021-05-10
Deletion: fix query construction on is_archive dataset update. #4385

rcarpa:
- Only retrieve datasets for update if is_archive=True. The update
query already had a filter to only set is_archive as false if it was
True before the update, so this change shouldn't have much impact.
But it is probably a good idea to filter upstream when possible.

- Add a distinct() filter. If multiple files from the same dataset
are queried, the same dataset will be returned multiple times. This
results in a needlessly big update query.


mlassnig (approved):
LGTM

bari12:
Closes #4622

PR #4619, opened 2021-05-10
Hermes2 : Add a protection against empty transferred_at : Closes #4618

### LINKED ISSUE #4618: Hermes2: Add a protection against empty transferred_at
Motivation
----------
When someone cancels a transfer, `transferred_at` is not set which crashes hermes2 if it works on influx.

cserf:
Hermes2 : Add a protection against empty transferred_at : Closes #4618

### DIFF
--- a/lib/rucio/daemons/hermes/hermes2.py
+++ b/lib/rucio/daemons/hermes/hermes2.py
@@ -39,7 +39,6 @@ import socket
 import sys
 import threading
 import time
-import traceback
 from email.mime.text import MIMEText
 
 import requests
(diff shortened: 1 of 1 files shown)


cserf:
The fix is line 341 ;)

mlassnig (approved):
You and your spaces before ":"

:-p

bari12:
I think we should make a decision about the spaces before `:` once and for all. I don't even mind so much the spaces (Although I prefer not to have them), but mostly I care about the consistency/inconsistency through the code. It just makes the logs harder to read and with everything pushed to elastic, also harder to filter for.

Will add a discussion point for this to the next Dev meeting, so we setup a style guide for this.

bari12:
In the meanwhile I will merge, since it was like this before already anyway 👍 😄

Issue #4473, opened 2021-03-22
Refactor _list_replicas and correctly implement xroot prioritization for archives


rcarpa:
Motivation
----------
Xroot prioritization for constituent replicas was implemented in #2313, but the current implementation was rushed and lead to undesired side-effects. We tried to quickfix the problem in #4380, but the fix doesn't seem easy. In particular because we recursively call list_replicas on the archive in order to construct the constituent replicas. Some information needed for correct prioritization is lost during the return from this recursive call.  

Modification
------------
[_list_replicas](https://github.com/rucio/rucio/blob/55ddd574709092220f0ae70911e0964538863c46/lib/rucio/core/replica.py#L802) has grown to be quite complex. Add more functional tests to ensure lack of regressions; refactor the function; and ensure constituent replicas can be correctly prioritized.

rcarpa:
I'll start working on this one

PR #4620, opened 2021-05-11
Core & Internals: improve list_replica grouping by scope/name. #4473

rcarpa:
The for loop assumes that the replicas are returned in a sorted order
by (scope, name). While scope/name doesn't change, the replica is
appended to the file. When scope/name changes, the file is yielded and
a new file starts to be constructed. Because the same scope/name can
be returned by _list_replicas_for_datasets and _list_replicas_for_files,
and they are not correctly ordered, this logic is broken in some cases.

Fix the issue by ensuring that the two data sources are correctly
merged into an ordered sequence.

### DIFF
--- a/lib/rucio/core/replica.py
+++ b/lib/rucio/core/replica.py
@@ -41,6 +41,7 @@
 
 from __future__ import print_function
 
+import heapq
 import logging
 import random
 from collections import defaultdict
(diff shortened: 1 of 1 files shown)


mlassnig (commented):
This looks good in principle. Very nice with the heapq.

Through the diff it's difficult to see though if there were any more changes in the code blocks, as it looks like it was just blocks of code moved around. Is it the case?

rcarpa:
No, the big change inside the "for" is just de-identation due to the removal of the outer "for". I don't know why github has so much difficulties to show diffs for identation changes

bziemons (approved):

bziemons:
> Through the diff it's difficult to see though if there were any more changes in the code blocks, as it looks like it was just blocks of code moved around. Is it the case?



> No, the big change inside the "for" is just de-identation due to the removal of the outer "for". I don't know why github has so much difficulties to show diffs for identation changes

@mlassnig really good new feature "Hide whitespace changes": 
image

PR #4621, opened 2021-05-11
Metadata: Add an option to inherit the metadata from the parents : Closes #4515

### LINKED ISSUE #4515: Metadata: Add an option to inherit the metadata from the parents
Motivation
----------
For the collaborations using Rucio as a hierarchical namespace, it can be interesting if the files inherit the metadata of all their parents.

Modification
------------
Add an inherit option in the `get_metadata` method to propagate the metadata of the parents.

cserf:
Metadata: Add an option to inherit the metadata from the parents : Closes #4515

### DIFF
--- a/lib/rucio/api/did.py
+++ b/lib/rucio/api/did.py
@@ -15,7 +15,7 @@
 #
 # Authors:
 # - Vincent Garonne <vincent.garonne@cern.ch>, 2013-2017
-# - Cedric Serfon <cedric.serfon@cern.ch>, 2013-2020
+# - Cedric Serfon <cedric.serfon@cern.ch>, 2013-2021
 # - Ralph Vigne <ralph.vigne@cern.ch>, 2013
 # - Mario Lassnig <mario.lassnig@cern.ch>, 2013-2020
 # - Yun-Pin Sun <winter0128@gmail.com>, 2013
@@ -469,17 +469,18 @@ def get_metadata(scope, name, plugin='DID_COLUMN', vo='def'):
     return api_update_return_dict(d)
 
 
-def get_metadata_bulk(dids, vo='def', session=None):
+def get_metadata_bulk(dids, inherit=False, vo='def'):
     """
     Get metadata for a list of dids
-    :param dids: A list of dids.
-    :param session: The database session in use.
+    :param dids:               A list of dids.
+    :param inherit:            A boolean. If set to true, the metadata of the parent are concatenated.
+    :param vo:                 The VO to act on.
     """
 
     validate_schema(name='dids', obj=dids, vo=vo)
     for entry in dids:
         entry['scope'] = InternalScope(entry['scope'], vo=vo)
-    meta = did.get_metadata_bulk(dids)
+    meta = did.get_metadata_bulk(dids, inherit=inherit)
     for met in meta:
         yield api_update_return_dict(met)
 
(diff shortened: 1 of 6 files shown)


lib/rucio/api/did.py
```diff
@@ -469,17 +469,18 @@ def get_metadata(scope, name, plugin='DID_COLUMN', vo='def'):
...
+    :param dids:               A list of dids.
+    :param inherit:            A boolean. If set to true, the metadata of the parent are concatenated.
+    :param vo:                 The VO to act on.
     """
 
     validate_schema(name='dids', obj=dids, vo=vo)
     for entry in dids:
         entry['scope'] = InternalScope(entry['scope'], vo=vo)
-    meta = did.get_metadata_bulk(dids)
+    meta = did.get_metadata_bulk(dids, inherit=inherit)
```
mlassnig, line 483: is this missing the vo selection?

cserf: Nope, there is no vo in the signature of the core method. The vo filtering is done when the `InternalScope` is extracted at the previous line

cserf: The only tests that fail are the webpy ones, but webpy is not available anymore... I don't understand

bziemons:
@cserf could you please rebase this on latest master?

Edit: I think that will fix the failing tests

lib/rucio/core/did.py
```diff
@@ -1288,28 +1288,97 @@ def get_metadata(scope, name, plugin='DID_COLUMN', session=None):
...
+                for entry in parent_list:
+                    if entry[-1] == (child_scope, child_name):
+                        entry.append((scope, name))
+            dids = [{'scope': did[0], 'name': did[1]} for did in parents]
+            depth += 1
+        unique_dids = [{'scope': did[0], 'name': did[1]} for did in unique_dids]
+        meta_dict = {}
+        for did in unique_dids:
+            try:
+                meta = get_metadata(did['scope'], did['name'], plugin='JSON', session=session)
```
bari12, line 1355: Shouldn't this be `plugin='ALL`' here?

bari12 (commented):
I think it is good. Do I read it right or does the get_metadata_bulk method now fully support the metadata plugins? (It wasn't the case before, but both of these developments went on somewhat in parallel)

cserf:
@bziemons rebase didn't help

bziemons:
@cserf yeah, I was wrong. This doesn't work as a patch against master since you do need to edit the webpy endpoint. Only solution is to make an extra PR against (and originating from) `release-1.25` to edit the deleted webpy endpoint, if you want this to be merged into the current release.

cserf:
> @cserf yeah, I was wrong. This doesn't work as a patch against master since you do need to edit the webpy endpoint. Only solution is to make an extra PR against (and originating from) `release-1.25` to edit the deleted webpy endpoint, if you want this to be merged into the current release.

I don't get it. Why the rebase doesn't work ?

bziemons:
@bari12 we have sorted this out. You would need to cherry-pick/merge the commit into master and disregard the deleted webpy endpoint then.

lib/rucio/tests/test_did.py
```diff
@@ -1117,3 +1133,48 @@ def test_bulk_get_meta(self):
             list_dids.append({'scope': scope, 'name': cnt[idx]})
         list_meta = [_ for _ in self.did_client.get_metadata_bulk(list_dids)]
         assert len(list_meta) == 0
+
+    @pytest.mark.dirty
+    @pytest.mark.noparallel(reason='uses pre-defined RSE')
+    def test_bulk_get_meta_inheritance(self):
```
bziemons, line 1139: it would be great if you could create new tests with generated RSEs from e. g. `rse_factory.make_rse()` to avoid noparallel and dirty markers.

bziemons (changes requested):

Issue #4625, opened 2021-05-18
Conveyor submitter not fully consuming their partitions


bari12:
Motivation
----------
Recently we observe that some conveyor submitter threads do not fully consume their partitions and some requests stay behind. This has the noticeable effect that only a fraction of transfers for a replication rule (submitted at the same time) are processed, while the others stay in `QUEUED` state and are only picked up (sometimes hours) later. 

However, due to the partitions backlog being even below the submitters bulk size, these partitions should be submitted immediately, or latest a few minutes later.

Modification
------------
Under investigation.

rcarpa:
Can we consider this one as being fixed?

bari12:
Yes 👍 
We can re-open should it come up again.

PR #4628, opened 2021-05-18
Transfers: fix comparison with None for link_ranking. #4625

rcarpa:
Happens when the first found source is a multihop, so the
link ranking is None when returned from the database.

The current link_ranking variable should (tm) not be None,
because this conditional branch is only executed for single hop
transfers.

### DIFF
--- a/lib/rucio/core/transfer.py
+++ b/lib/rucio/core/transfer.py
@@ -1061,7 +1061,10 @@ def get_transfer_requests_and_source_replicas(total_workers=0, worker_number=0, 
                 if current_source_is_tape and new_source_is_tape:
                     # multiple Tape source replicas are not allowed in FTS3.
                     # Either keep the old source. Or substitute it with the new one.
-                    if ranking > transfers[req_id]['sources'][0][3] or (ranking == transfers[req_id]['sources'][0][3] and link_ranking < transfers[req_id]['sources'][0][4]):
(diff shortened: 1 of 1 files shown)


lib/rucio/core/transfer.py
```diff
@@ -1061,7 +1061,9 @@ def protocol(self, rse_id, scheme, operation):
                 if current_source_is_tape and new_source_is_tape:
                     # multiple Tape source replicas are not allowed in FTS3.
                     # Either keep the old source. Or substitute it with the new one.
-                    if ranking > transfers[req_id]['sources'][0][3] or (ranking == transfers[req_id]['sources'][0][3] and link_ranking < transfers[req_id]['sources'][0][4]):
+                    if ranking > transfers[req_id]['sources'][0][3] or \
+                            (ranking == transfers[req_id]['sources'][0][3] and (not transfers[req_id]['sources'][0][4]  # will be None if the previous transfer is multihop
```
bziemons: to make this easier to comprehend, add a variable like this:

```python
prev_is_multihop = transfers[req_id]['sources'][0][4] is None  # will be None if the previous transfer is multihop
```

rcarpa: Good idea. Done.

bziemons (approved):

PR #4631, opened 2021-05-18
Transfers: skip missing request on state update in submitter. #4625

rcarpa:
The submitter fails when trying to update the request state of missing
requests. I verified 5 requests manually on which the operation failed
and all of them where from the "Functional Tests" activity.
I assume some job creates and removes requests for testing. These
requests are queued in submitter. In the meantime they get deleted.
This crashes the submitter, which must restart all activities.

The docstring of the set_request_state function actually says that
missing requests will be ignored, but it's not what the function did.
It is used in multiple places in the code, some rely on the fact that
it will fail on missing requests. So fix the docstring and add a
version of this function which does actually skip missing requests.


bziemons (approved):
Urgh, wrong documentation :)
lgtm

PR #4632, opened 2021-05-18
Core & Internals: directly fetch constituents in list_replicas. #4473

rcarpa:
When listing archive constituent replicas, don't perform a recursive
call anymore to list the replicas of the archive. Thanks to that, we
are able to iterate over all the possible schemes of the archive file
and better prioritize the root protocol for downloading constituents.

### DIFF
--- a/lib/rucio/core/replica.py
+++ b/lib/rucio/core/replica.py
@@ -641,10 +641,14 @@ def _resolve_dids(dids, unavailable, ignore_availability, all_states, resolve_ar
     :param resolve_archives: When set to true, find archives which contain the replicas.
     :param session: The database session in use.
     """
-    did_clause, dataset_clause, file_clause, files, constituents = [], [], [], [], {}
+    did_clause, dataset_clause, file_clause, constituent_clause = [], [], [], []
+    # Accumulate all the dids which were requested explicitly (not via a container/dataset).
+    # If any replicas for these dids will be found latter, the associated did will be removed from the list,
+    # leaving, at the end, only the requested dids which didn't have any replicas at all.
+    files_wo_replica = []
     for did in [dict(tupleized) for tupleized in set(tuple(item.items()) for item in dids)]:
         if 'type' in did and did['type'] in (DIDType.FILE, DIDType.FILE.value) or 'did_type' in did and did['did_type'] in (DIDType.FILE, DIDType.FILE.value):  # pylint: disable=no-member
-            files.append({'scope': did['scope'], 'name': did['name']})
+            files_wo_replica.append({'scope': did['scope'], 'name': did['name']})
             file_clause.append(and_(models.RSEFileAssociation.scope == did['scope'],
                                     models.RSEFileAssociation.name == did['name']))
 
@@ -660,17 +664,11 @@ def _resolve_dids(dids, unavailable, ignore_availability, all_states, resolve_ar
                                                          .with_hint(models.DataIdentifier, "INDEX(DIDS DIDS_PK)", 'oracle')\
                                                          .filter(or_(*did_clause)):
             if resolve_archives and constituent:
-                # file is a constituent, resolve to parent archives if necessary
-                archive = session.query(models.ConstituentAssociation.scope,
-                                        models.ConstituentAssociation.name)\
-                                 .filter(models.ConstituentAssociation.child_scope == scope,
-                                         models.ConstituentAssociation.child_name == name)\
-                                 .with_hint(models.ConstituentAssociation, "INDEX(ARCHIVE_CONTENTS ARCH_CONTENTS_PK)", 'oracle')\
-                                 .all()
-                constituents['%s:%s' % (scope.internal, name)] = [{'scope': tmp[0], 'name': tmp[1]} for tmp in archive]
+                constituent_clause.append(and_(models.ConstituentAssociation.child_scope == scope,
+                                               models.ConstituentAssociation.child_name == name))
 
             if did_type == DIDType.FILE:
-                files.append({'scope': scope, 'name': name})
+                files_wo_replica.append({'scope': scope, 'name': name})
                 file_clause.append(and_(models.RSEFileAssociation.scope == scope,
                                         models.RSEFileAssociation.name == name))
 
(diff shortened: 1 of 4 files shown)


rcarpa:
The first commit is a dependency which was submitted in a separate PR, but not yet merged

lib/rucio/tests/test_archive.py
```diff
@@ -186,3 +186,63 @@ def test_archive_on_dataset_level(rse_factory, did_factory, root_account):
     assert not metadata['is_archive']
     metadata = get_metadata(**dataset2)
     assert not metadata['is_archive']
+
+
+def test_root_priority_is_highest(rse_factory, mock_scope, root_account, did_client, replica_client):
```
mlassnig, line 191: what happens where there's no ROOT protocol set on the RSE?

lib/rucio/tests/test_download.py
```diff
@@ -297,6 +298,37 @@ def test_download_from_archive_on_xrd(did_factory, download_client, did_client):
...
+    """
+    rse, _ = rse_factory.make_posix_rse()
+    scope = mock_scope.external
+    base_name = 'testDownloadClientExtract' + generate_uuid()
+    with TemporaryDirectory() as tmp_dir:
+        # Create a tar archive with a file inside and upload it
+        name = base_name + '.000'
+        data = '000'
+        adler32 = '01230091'
+        tar_name = base_name + '.tar.gz'
```
mlassnig, line 313: o_O I thought ROOT only handles ZIPs? @bari12 ?

lib/rucio/tests/test_replica.py
```diff
@@ -851,6 +853,34 @@ def test_client_access_denied_on_delete_replicas(rse_factory, mock_scope, replic
...
+    file, dataset, container = ({'scope': did['scope'].external, 'name': did['name']} for did in (file, dataset, container))
+
+    did_client.add_files_to_dataset(files=[file], **dataset)
+    did_client.add_datasets_to_container(dsns=[dataset], **container)
+
+    replicas = list(replica_client.list_replicas(dids=[file]))
+    assert len(replicas) == 1
+    assert not replicas[0]['rses']
+    assert not replicas[0]['pfns']
+    # TODO: fix listing dids without replicas from datasets and containers and uncomment the following 2 asserts
```
mlassnig, line 906: this is still from Rod's favourite ticket, right?

lib/rucio/core/replica.py
```diff
@@ -1233,6 +1217,18 @@ def _list_replicas(dataset_clause, file_clause, state_clause, show_pfns,
         yield file
         file = {}
 
+    for scope, name, bytes, md5, adler32 in _list_files_wo_replicas(files_wo_replica, session):
+        file = {
```
mlassnig: not necessary to instantiate a dict here? just stream out the values

lib/rucio/core/replica.py
```diff
@@ -864,13 +921,18 @@ def get_multi_cache_prefix(cache_site, filename, logger=logging.log):
...
                    updated_after, filters, ignore_availability,
                    session):
 
-    files = [dataset_clause and _list_replicas_for_datasets(dataset_clause, state_clause, rse_clause, ignore_availability, updated_after, session),
-             file_clause and _list_replicas_for_files(file_clause, state_clause, files, rse_clause, ignore_availability, updated_after, session)]
+    # iterator which merges multiple sorted replica sources into a combine sorted result without loading everything into the memory
+    replicas = heapq.merge(
+        _list_replicas_for_datasets(dataset_clause, state_clause, rse_clause, ignore_availability, updated_after, session),
+        _list_replicas_for_files(file_clause, state_clause, files_wo_replica, rse_clause, ignore_availability, updated_after, session),
+        _list_replicas_for_constituents(constituent_clause, state_clause, files_wo_replica, rse_clause, ignore_availability, updated_after, session),
```
mlassnig, line 970: this looks nice to me.

maybe worth to add some comment that `_files` and `_constituent` are very similar, and if one of the SQLAlchemy queries changes in one, that the other must be adapted as well? Otherwise we run the risk of having two queries with different outcomes.

lib/rucio/tests/test_download.py
bari12: But this is the client extract test, there it doesn't matter if it is zip or tar.gz
But in principle, yes, I believe root only supports zip

lib/rucio/tests/test_archive.py
rcarpa: It will work. This particular test even shows that. The test uses two RSEs: one with xroot and the second without. The best available protocol is selected according to normal rules for each replica. These rules are not changed here. The root replicas is given the lowest (best) priority. So they'll be sorted to be first in the list (if they exist). If no root replica exist at all, non-root ones will be used and client-extraction triggered.

lib/rucio/tests/test_replica.py
rcarpa: Hmm. Didn't know there is a ticket for that already. I added the test because the behavior of not showing empty replicas was not documented or tested anywhere; but @bari12 told me that it must behave like that.

lib/rucio/core/replica.py
rcarpa: It's even more than that: all 3 of these functions are very similar. There is an initial query, which is different, but then we append filters to it, which are all identical. The `_files`  has a slightly different structure, but I think it's only for historical reasons, not a technical one. It's maybe worth to extract common code (appending filters to initial query) into a separate function ? WDYT ?

lib/rucio/core/replica.py
rcarpa: you are right, but I did it on purpose like that. Just for readability. I don't have strong preferences, can yield directly if it is preferred.

bari12:
Rebase plz :-)

rcarpa:
Rebased

lib/rucio/core/replica.py
mlassnig: yeah, should be done. but can be a separate issue/PR

lib/rucio/core/replica.py
mlassnig: yeah, please, python is bad enough with memory management, no need to have extra allocations on our most often-repeated function :-)

rcarpa: done

mlassnig (approved):

Issue #4626, opened 2021-05-18
Optimize rucio download --nrandom


bari12:
Motivation
----------
Currently the `--nrandom` option of Rucio download works very inefficiently. It requests a `list_replicas` of all files in the dataset/container, and then picks a random amount (Most of the time 1) file out of the result.

Since some datasets/containers are very big, this essentially never succeeds, since you run into the request timeout of the `list_replicas` call on the server, with the client receiving a malformed xml/json.

Modification
------------
Two possible solutions, with both doing the random picking efficiently on the server side.
1. Add a `nrandom` option to `list_replicas` - with `list_replicas` just returning this specific amount of random replicas.
2. Add a random selection in `list_contents`; The client runs `list_contents` first, and then feeds `list_replicas` just with these files, instead of the entire dataset. This add the complication though that some of the contents might have no replicas.

PR #4640, opened 2021-05-25
Core & Internals, API, Clients: nrandom for list_replicas. #4626

rcarpa:
Introduce the nrandom filter for list_replicas. It is applied before
sending replicas to the client: all replicas are retrieved from the
database, processed; and then the filtering is applied. This is
sub-optimal, because a lot of processing is performed for replicas
which will be dropped afterwards, but is the only safe way to implement
it right now.

The reason is the following: some existing filters are only applied
in python processing logic. For example the "schemes" filter. Meaning
that "nrandom" can only be applied after these filters and cannot be
done during the SQL call.

This change still results in some optimisation, because less data will
be serialized to json and sent to clients.

### DIFF
--- a/lib/rucio/core/replica.py
+++ b/lib/rucio/core/replica.py
@@ -37,10 +37,12 @@
 # - Eric Vaandering <ewv@fnal.gov>, 2020-2021
 # - Benedikt Ziemons <benedikt.ziemons@cern.ch>, 2020
 # - Radu Carpa <radu.carpa@cern.ch>, 2021
+# - Gabriele Fronzé <sucre.91@hotmail.it>, 2021
 
 from __future__ import print_function
 
 import logging
+import random
 from collections import defaultdict
 from copy import deepcopy
 from curses.ascii import isprint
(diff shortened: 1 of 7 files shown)


rcarpa:
some release-patch-tests fail due to webpy removal

lib/rucio/core/replica.py
```diff
@@ -704,6 +706,41 @@ def _resolve_dids(dids, unavailable, ignore_availability, all_states, resolve_ar
...
+    """
+
+    if not nrandom:
+        # pass-through the data unchanged
+        yield from generator
+        return
+
+    # A "reservoir sampling" algorithm:
+    # Copy the N first files from the generator. After that, following element may be picked to substitute
+    # one of the previously selected element with a probability which decreases as the number of encountered elements grows.
```
bari12, line 721: I like it. Intuitively, when looking at it first, it seems that the first results have a much higher probability of selection, but in the end, as long as you touch all elements, it's uniform. 👍

bari12 (commented):
We should still introduce this into the `release-1.25` as well. Maybe base this on the release branch, I will cherry-pick it to `master` then.

rcarpa:
I don't believe there exists a way to update the destination branch of PR. Is  there any ?

bziemons:
> I don't believe there exists a way to update the destination branch of PR. Is there any ?

when you click edit on the PR you can edit the destination branch :)

rcarpa:
Thanks. I was sure I tried in the past and was not able to do it. Most probably I tried to update the source branch then.

rcarpa:
Updated the PR. It is split into two commits. The first one should apply to master unchanged.

bari12 (approved):

PR #4642, opened 2021-05-25
 Multi VO: Add long VO name mapping feature. Fix #4548 

### LINKED ISSUE #4548: Supporting longer VO names in multi-VO rucio?
Hi,

I would like to suggest adding a mapping of long VO names down to the three-letter VO tokens used by the database at the API level.

Motivation
----------
The motivation for this is to make the API easier for other projects to use. It also has the benefit of allowing users to put the full VO name in their local client config file (i.e. more user friendly too). One of my colleagues is working on a "multi-VO DIRAC" to "multi-VO Rucio" interface and this would make the API far easier to call from DIRAC.

Modification
------------
I've put together a proof-of-concept patch of how this could work:
https://github.com/sfayer/rucio/commit/5e2bd5ff389426bb713f066844136b52666fe2ad

It simply adds an optional VO mapping section to the rucio.cfg on the server and uses this to remap the VO name in the API functions. If a mapping isn't found (or the section doesn't exist), the vo field is passed through unchanged to keep this feature entirely optional and completely backwards compatible. This approach also avoids any database schema changes.

Would you please consider this as a possible addition to Rucio? I'd be happy to submit my code as a pull request or discuss other possible approaches to this.

Regards,
Simon

sfayer:
Hi,

This adds support for mapping long VO names to the internal three-character tags at client interfaces as discussed in ticket #4548. It also updates the multi VO tests to use the long names to ensure that all interfaces are properly covered.

Regards,
Simon

### DIFF
--- a/lib/rucio/tests/test_multi_vo.py
+++ b/lib/rucio/tests/test_multi_vo.py
@@ -20,6 +20,7 @@
 # - Mario Lassnig <mario.lassnig@cern.ch>, 2020
 # - Martin Barisits <martin.barisits@cern.ch>, 2020-2021
 # - Radu Carpa <radu.carpa@cern.ch>, 2021
+# - Simon Fayer <simon.fayer05@imperial.ac.uk>, 2021
 
 import os
 import unittest
@@ -27,7 +28,7 @@ from logging import getLogger
 from os import remove
 from random import choice
 from re import search
-from string import ascii_uppercase, ascii_lowercase
+from string import ascii_uppercase, ascii_lowercase, ascii_letters, digits
 from unittest.mock import patch
 from urllib.parse import urlparse, parse_qs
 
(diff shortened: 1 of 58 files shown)


bari12:
Hi @sfayer 
Sorry for the delay at the moment. I still think we would need many more unittests which actually test different commands with the VO mapping. Just to make sure that it is working across the stack.
@ijjorama can you please review the PR as well.

sfayer:
Hi @bari12,

The client command VO mappings are very thoroughly tested in the multi_vo test suite, this patch:
 * Sets the full VO names in the client config: https://github.com/rucio/rucio/blob/5aa81d53abd300dcb3e751053e3d97be010fadd1/etc/docker/test/extra/rucio_multi_vo_tst_postgres12.cfg#L20
 * Adds mappings to the config table in the test bootstrap: https://github.com/rucio/rucio/blob/5aa81d53abd300dcb3e751053e3d97be010fadd1/lib/rucio/tests/common_server.py#L34
 * The multi_vo test script then runs all of the usual unit tests with the two VO names set: https://github.com/rucio/rucio/blob/5aa81d53abd300dcb3e751053e3d97be010fadd1/tools/run_multi_vo_tests_docker.sh#L111

This should mean that every single client command that has an existing test, is tested with two full VOs names including the mappings: If there are any problems with it (missing or otherwise incorrect mapping), these tests would fail. I can't really see what extra tests I could add to these?

Regards,
Simon

bziemons (changes requested):
You are missing
- https://github.com/rucio/rucio/blob/02e86830f10c1bd7e5ea2a05d306706a09127bfe/lib/rucio/web/rest/flaskapi/v1/credentials.py#L80
- https://github.com/rucio/rucio/blob/02e86830f10c1bd7e5ea2a05d306706a09127bfe/lib/rucio/web/rest/flaskapi/v1/redirect.py#L74
- https://github.com/rucio/rucio/blob/02e86830f10c1bd7e5ea2a05d306706a09127bfe/lib/rucio/web/rest/flaskapi/v1/redirect.py#L173

(maybe this is a great occasion to add a flask common function like `extract_vo(request.headers)`)


I am assuming a client can now chose between using the short vo and the long vo name (since otherwise this change would break current configurations). Can you make a test in `lib/rucio/tests/test_multi_vo.py` using the short name via the API and the long name and check that they resolve to the same? Just so that this is painfully obvious.

lib/rucio/tests/test_multi_vo.py
```diff
@@ -663,12 +669,34 @@ def test_recover_vo_denied(self):
...
+            assert response.status_code == 200
+            token = str(response.headers.get('X-Rucio-Auth-Token'))
+            response = self.rest_client.get('/accounts/', headers=headers(auth(token)))
+            assert response.status_code == 200
+            accounts = [parse_response(a)['account'] for a in response.get_data(as_text=True).split('\n')[:-1]]
+            return sorted(accounts)
+        # The test VOs contain different account names
+        # We get all the account names with the long VO name and short VO name and check they are equal
+        accounts_long = get_vo_accounts(self.long_vo['vo'])
+        accounts_short = get_vo_accounts(self.vo['vo'])
```
bziemons, line 685: `assert len(accounts_short) > 0` ?

bziemons (dismissed):
Thanks for quickly addressing my comments :) @bari12 I think this is well-tested and good to go :+1:

bziemons:
I have taken the liberty to resolve the conflicts and update the file headers :)

bziemons (approved):

lib/rucio/tests/test_multi_vo.py
```diff
@@ -663,12 +668,35 @@ def test_recover_vo_denied(self):
         response = self.rest_client.post('/vos/' + self.vo['vo'] + '/recover', headers=headers(auth(token)), json=params)
         assert response.status_code == 401
 
+    def test_rest_vomap(self):
```
bziemons, line 672: This needs a noparallel mark, since the accounts in the vo can change from other tests in-between checking them. Happened on my end: https://github.com/bziemons/rucio/runs/2894657636?check_suite_focus=true#step:4:5341

bziemons (changes requested):

lib/rucio/tests/test_multi_vo.py
sfayer: Sure, I've added the noparallel mark to this test. Thanks for sorting out the conflicts and headers.

bziemons (approved):

Issue #4639, opened 2021-05-25
write_lan activity is not use on upload


bari12:
Motivation
----------
Recently it was identified that the `write_lan` activity is not used when uploading. This was supposedly fixed with #3626. My suspicion is that this is only the case in certain configurations though. (E.g. the one of TRIUMF)

https://aipanda403.cern.ch/data/jobs/2021-05-19/TRIUMF/5060545120.out

Shows an upload output of the pilot. The critical part is:

```
2021-05-19 07:57:42,369 | DEBUG    | Thread-5            | pilot.copytool.rucio             | upload                    | [{u'extended_attributes': None, u'hostname': u'xrootd.lcg.triumf.ca', u'prefix': u'//atlas/atlasdatadisk/rucio/', u'domains': {u'wan': {u'read': 2, u'write': 3, u'third_party_copy': 0, u'delete': 3}, u'lan': {u'read': 2, u'write': 3, u'delete': 3}}, u'scheme': u'root', u'port': 1094, u'impl': u'rucio.rse.protocols.gfal.Default'}, {u'extended_attributes': {u'space_token': u'ATLASDATADISK', u'web_service_path': u'/srm/managerv2?SFN='}, u'hostname': u'srm.triumf.ca', u'prefix': u'/atlas/atlasdatadisk/rucio/', u'domains': {u'wan': {u'read': 3, u'write': 2, u'third_party_copy': 2, u'delete': 2}, u'lan': {u'read': 3, u'write': 2, u'delete': 2}}, u'scheme': u'srm', u'port': 8443, u'impl': u'rucio.rse.protocols.gfal.Default'}, {u'extended_attributes': None, u'hostname': u'webdav-lan.lcg.triumf.ca', u'prefix': u'/atlas/atlasdatadisk/rucio/', u'domains': {u'wan': {u'read': 0, u'write': 0, u'third_party_copy': 0, u'delete': 0}, u'lan': {u'read': 1, u'write': 1, u'delete': 1}}, u'scheme': u'davs', u'port': 2880, u'impl': u'rucio.rse.protocols.gfal.Default'}]
2021-05-19 07:57:42,369 | INFO     | Thread-5            | pilot.copytool.rucio             | upload                    | Trying upload with davs to TRIUMF-LCG2_DATADISK
2021-05-19 07:57:42,369 | DEBUG    | Thread-5            | pilot.copytool.rucio             | upload                    | Processing upload with the domain: lan
[...]

PR #4643, opened 2021-05-26
Clients: propagate selected domain in uploadclient. #4639

rcarpa:
The upload client has some logic to select the 'lan' domain if the
client and rse 'site' are identical. This results into correctly
selecting the `scheme` which can be used for a lan transfer.
However, the domain is not enforced when the protocol object is
latter created. This will result in using the default "wan" domain
for the upload if a protocol with the same scheme is available on wan.


mlassnig (approved):
LGTM

PR #4653, opened 2021-06-01
Rest & API: always setup logger

rcarpa:
WSGI doesn't run applications via __main__, so the logger is not setup in production.


bziemons (changes requested):
Needs to be removed from `lib/rucio/web/rest/main.py` and `lib/rucio/web/ui/main.py` and also needs to be fixed in `lib/rucio/web/ui/flask/main.py`

rcarpa:
> Needs to be removed from `lib/rucio/web/rest/main.py` and `lib/rucio/web/ui/main.py` and also needs to be fixed in `lib/rucio/web/ui/flask/main.py`

Done. Thanks for the review.

bziemons (approved):

Issue #4661, opened 2021-06-07
replica sorting by GeoIP improvements


ivukotic:
Motivation
----------
Current replica sorting by distance relies on GeoIP info returned by client. 
Client finds its own IP by asking socket when connecting to 8.8.8.8. This returns local address when running in docker or behind NAT. Better approach is to ask specialized service for the client's IP. 
Another issue is that current distance calculation works only with IPv4 addresses and won't work for IPv6 only hosts. 
For places that need it, users might want to provide their lat, long and have it used in distance calculation.

Modification
------------
* Move from 8.8.8.8 to https://api64.ipify.org
* Add option for user to set environment vars: RUCIO_LATITUDE and RUCIO_LONGITUDE.

PR #4662, opened 2021-06-07
Replica sorting improvements Issue #4661

ivukotic:
resolves the issue.

### DIFF
--- a/lib/rucio/common/utils.py
+++ b/lib/rucio/common/utils.py
@@ -897,8 +897,9 @@ class Color:
 
 def detect_client_location():
     """
-    Client IP will be set on the server side (request.remote_addr)
-
+    Normally client IP will be set on the server side (request.remote_addr)
+    Here setting ip on the one seen by the host itself. There is no connection
+    to Google DNS servers.
     Try to determine the sitename automatically from common environment variables,
     in this order: SITE_NAME, ATLAS_SITE_NAME, OSG_SITE_NAME. If none of these exist
     use the fixed string 'ROAMING'.
@@ -906,7 +907,25 @@ def detect_client_location():
     If environment variables sets location, it uses it.
     """
 
-    ip = '0.0.0.0'
+    ip = None
+
+    try:
+        s = socket.socket(socket.AF_INET6, socket.SOCK_DGRAM)
+        s.connect(("2001:4860:4860:0:0:0:0:8888", 80))
+        ip = s.getsockname()[0]
+    except Exception:
+        pass
+
+    if not ip:
+        try:
+            s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+            s.connect(("8.8.8.8", 80))
+            ip = s.getsockname()[0]
+        except Exception:
+            pass
+
+    if not ip:
+        ip = '0.0.0.0'
 
     site = os.environ.get('SITE_NAME',
                           os.environ.get('ATLAS_SITE_NAME',
--- a/lib/rucio/common/utils.py
+++ b/lib/rucio/common/utils.py
@@ -897,27 +897,16 @@ class Color:
 
 def detect_client_location():
     """
-    Open a UDP socket to a machine on the internet, to get the local IPv4 and IPv6
-    addresses of the requesting client.
+    Client IP will be set on the server side (request.remote_addr) 
 
     Try to determine the sitename automatically from common environment variables,
     in this order: SITE_NAME, ATLAS_SITE_NAME, OSG_SITE_NAME. If none of these exist
     use the fixed string 'ROAMING'.
+
+    If environment variables sets location, it uses it. 
     """
 
     ip = '0.0.0.0'
-    try:
-        ip = requests.get('https://api64.ipify.org').text
-    except Exception:
-        pass
-
-    ip6 = '::'
-    try:
-        s = socket.socket(socket.AF_INET6, socket.SOCK_DGRAM)
-        s.connect(("2001:4860:4860:0:0:0:0:8888", 80))
-        ip6 = s.getsockname()[0]
-    except Exception:
-        pass
 
     site = os.environ.get('SITE_NAME',
                           os.environ.get('ATLAS_SITE_NAME',
@@ -937,7 +926,6 @@ def detect_client_location():
         latitude = longitude = 0
 
     return {'ip': ip,
-            'ip6': ip6,
             'fqdn': socket.getfqdn(),
             'site': site,
             'latitude': latitude,
--- a/lib/rucio/common/utils.py
+++ b/lib/rucio/common/utils.py
@@ -897,13 +897,13 @@ class Color:
 
 def detect_client_location():
     """
-    Client IP will be set on the server side (request.remote_addr) 
+    Client IP will be set on the server side (request.remote_addr)
 
     Try to determine the sitename automatically from common environment variables,
     in this order: SITE_NAME, ATLAS_SITE_NAME, OSG_SITE_NAME. If none of these exist
     use the fixed string 'ROAMING'.
 
-    If environment variables sets location, it uses it. 
+    If environment variables sets location, it uses it.
     """
 
     ip = '0.0.0.0'
--- a/lib/rucio/common/utils.py
+++ b/lib/rucio/common/utils.py
@@ -923,7 +923,7 @@ def detect_client_location():
             latitude = longitude = 0
             print('Client set latitude and longitude are not valid.')
     else:
-        latitude = longitude = 0
+        latitude = longitude = None
 
     return {'ip': ip,
             'fqdn': socket.getfqdn(),
--- a/lib/rucio/common/utils.py
+++ b/lib/rucio/common/utils.py
@@ -906,9 +906,7 @@ def detect_client_location():
 
     ip = '0.0.0.0'
     try:
-        s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
-        s.connect(("8.8.8.8", 80))
-        ip = s.getsockname()[0]
+        ip = requests.get('https://api64.ipify.org').text
     except Exception:
         pass
 
@@ -925,10 +923,24 @@ def detect_client_location():
                                          os.environ.get('OSG_SITE_NAME',
                                                         'ROAMING')))
 
+    latitude = os.environ.get('RUCIO_LATITUDE')
+    longitude = os.environ.get('RUCIO_LONGITUDE')
+    if latitude and longitude:
+        try:
+            latitude = float(latitude)
+            longitude = float(longitude)
+        except ValueError:
+            latitude = longitude = 0
+            print('Client set latitude and longitude are not valid.')
+    else:
+        latitude = longitude = 0
+
     return {'ip': ip,
             'ip6': ip6,
             'fqdn': socket.getfqdn(),
-            'site': site}
+            'site': site,
+            'latitude': latitude,
+            'longitude': longitude}
 
 
 def ssh_sign(private_key, message):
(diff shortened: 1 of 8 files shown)


lib/rucio/common/utils.py
```diff
@@ -906,9 +906,7 @@ def detect_client_location():
 
     ip = '0.0.0.0'
     try:
-        s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
-        s.connect(("8.8.8.8", 80))
-        ip = s.getsockname()[0]
+        ip = requests.get('https://api64.ipify.org').text
```
bziemons: I see multiple problems with this: returns ipv6 address when thats the default & relies on external service

lib/rucio/common/utils.py
```diff
@@ -925,10 +923,24 @@ def detect_client_location():
                                          os.environ.get('OSG_SITE_NAME',
                                                         'ROAMING')))
 
+    latitude = os.environ.get('RUCIO_LATITUDE')
+    longitude = os.environ.get('RUCIO_LONGITUDE')
```
bziemons, line 936: I wonder if it makes sense to add `rucio.cfg` settings for these

lib/rucio/common/utils.py
```diff
@@ -925,10 +923,24 @@ def detect_client_location():
...
+    else:
+        latitude = longitude = 0
+
     return {'ip': ip,
             'ip6': ip6,
             'fqdn': socket.getfqdn(),
-            'site': site}
+            'site': site,
+            'latitude': latitude,
+            'longitude': longitude}
```
bziemons, line 951: Not all usages of `detect_client_location` mention these in their documentation (like `rucio.client.replicaclient.ReplicaClient.list_replicas`) leading to false documentation

bziemons (changes requested):

lib/rucio/common/utils.py
ivukotic: 8.8.8.8 is also external service. 
I proposed using much nicer service that we manage (https://rucio.cern.workers.dev/) but Martin wants to avoid potentially political discussions.

concerning IPv6... if you look at actual calculation of the distance only ip is used. IPv6 is used nowhere as far as I can tell. I would have completely removed ipv6 but I don't know if someone planed to use it for something.

lib/rucio/common/utils.py
ivukotic: That was my proposal too, but Martin preferred this.

lib/rucio/common/utils.py
bziemons: > 8.8.8.8 is also external service.

which is completely irrelevant when using a DGRAM socket and "connecting" to just fetch the local ip. You could use any IP address that routes to the external network here, in fact even `0.0.0.1` works for me. I agree that any option seems hacky though.

> concerning IPv6... if you look at actual calculation of the distance only ip is used. IPv6 is used nowhere as far as I can tell. I would have completely removed ipv6 but I don't know if someone planed to use it for something.

good point

lib/rucio/common/utils.py
bari12: This was motivated a bit by the fact that there is a bit of a movement to use clients rucio-cfg less. Let's have a compromise and have both 😄

ivukotic: fixing this now.

lib/rucio/common/utils.py
bari12: You are right, `8.8.8.8` is already external anyway, I would propose to take this out entirely and just set the ip to the one the rucio REST interface sees. We then have the option to overwrite lat/long anyway, so no need to resolve the ip locally? Not sure if I am missing something where `detect_client_location` is called otherwise?

ivukotic: @bziemons problem is with socket method itself. It will give local IP if you are running in container or behind NAT.

bziemons: yes, but the server can see the client's public IP as Martin stated. I do not see a reason to have this :)

ivukotic: > You are right, `8.8.8.8` is already external anyway, I would propose to take this out entirely and just set the ip to the one the rucio REST interface sees. We then have the option to overwrite lat/long anyway, so no need to resolve the ip locally? Not sure if I am missing something where `detect_client_location` is called otherwise?

I am not sure what do you want to take out. Whole function? That we can't do as it also sets Site. Or you want to replace the address contacted with some Rucio IP address?

bari12: Only replace the ip part of the dict in the REST
E.g. https://github.com/rucio/rucio/blob/5db83564817e28207e2565f413a7857fbbba833f/lib/rucio/web/rest/flaskapi/v1/replicas.py#L292 already resolves the ip. If you replace in the `client_location` dictionary the IP there, I think that should be sufficient.

ivukotic: > yes, but the server can see the client's public IP as Martin stated. I do not see a reason to have this :)

I agree that Rucio server "sees" correct client's IP address. If you know how to propagate that info to here that would be great.

mlassnig: if you replace the socket.connect() with a requests.get() you will actually do a call to an outside service, we really should avoid this. the DGRAM socket connect is node-local only (since UDP has no "connection").

as a "better" alternative, I would completely move this to client-ip detection on the rucio server side.

ivukotic: I think we all agree that the best option is to use address that rucio already sees. Only question is how do we reorganize things. I did the simplest thing that works and requires no restructuring. And can be merged now :)

bari12: Have a look at the code link I posted above. There use the `client_ip` to overwrite the ip part in the `client_location` dictionary. It should be as easy as that.

ivukotic: @bari12 sorry I still don't get it... detect_client_location is used at three places (rucio/list_file_replicas, downloadclient and uploadclient), non of which has access to the request object?

ivukotic: this should be done now.

lib/rucio/common/utils.py
```diff
@@ -897,39 +897,39 @@ class Color:
...
                                                         'ROAMING')))
 
+    latitude = os.environ.get('RUCIO_LATITUDE')
+    longitude = os.environ.get('RUCIO_LONGITUDE')
+    if latitude and longitude:
+        try:
+            latitude = float(latitude)
+            longitude = float(longitude)
+        except ValueError:
+            latitude = longitude = 0
```
bari12, line 942: I would not set it to 0,0 - although unlikely, it's still a valid input. Just put `None` in the error/default case.

lib/rucio/web/rest/flaskapi/v1/replicas.py
```diff
@@ -298,6 +298,9 @@ def post(self):
                            'site': None}
         client_location.update(param_get(parameters, 'client_location', default={}))
 
+        # making sure IP address is not overwritten
+        client_location['ip'] = client_ip
```
bari12, line 302: There is a HTTP GET call as well for list_replicas in line 54. Needs to be done there too.

ivukotic: that one isn't overwritten?

lib/rucio/common/utils.py
ivukotic: that's what regular GeoIP would give. It is at least defined. But I changed it to None.

lib/rucio/web/rest/flaskapi/v1/replicas.py
bari12: No, you only did it for the POST call. I know it is quite confusing that there are two different ones, without going into to much details, I blame historic reasons 😄

lib/rucio/common/utils.py
```diff
@@ -897,39 +897,39 @@ class Color:
...
+    Client IP will be set on the server side (request.remote_addr)
 
     Try to determine the sitename automatically from common environment variables,
     in this order: SITE_NAME, ATLAS_SITE_NAME, OSG_SITE_NAME. If none of these exist
     use the fixed string 'ROAMING'.
+
+    If environment variables sets location, it uses it.
     """
 
     ip = '0.0.0.0'
```
bari12: Thinking more about this, I think we should leave the ip detection in the dictionary. (Without any outside connection, just reading it from the interface)
This method might be used (even in the future) by other methods, like traces, and they could use the local ip. (E.g. for a trace, you might want to know which node was it actually which created the data)
For the list_replicas, we will not use it and use the rest-exposed remote address (That you already changed)

bari12 (changes requested):

lib/rucio/common/utils.py
ivukotic: leave socket method? I think that's just opening a possibility for future issues and inconsistencies. 
Maybe we create three functions: one simply detecting site leaving everything else None, one detecting IP using socket where local ip is OK, one detecting using external service.

bari12: As @mlassnig  explained, since the `socket.SOCK_DGRAM` is used, I think it is fine, there is no outgoing connection for UDP. 
I would not break up the dictionary, as the method already needs to include fqdn, site, lat, long - it makes sense to leave ip and ip6 in. What we could do in the future is to have some options to the method do detect things differently.

ivukotic: I know there is no outgoing connection, but most people will be unaware of issues connected to the method. 
Also I would remove ipv6 from the dictionary. It creates false sense that in case you have ipv6 only node, things will work when in fact they wont. We leave only one ip. If it is ivp4 thats great, if that's ivp6 that's ok too.
I would not break the dictionary just have different functions that fill/overwrite its components...

mlassnig: pulling in @PalNilsson . I added this ipv6 field for the pilot, because he asked me to :-)

PalNilsson: Well, the pilot version of this function is a bit different;
[(pilot source)](https://github.com/PanDAWMS/pilot2/blob/29c4352edd9df14a480fa8b406c4a506e2313bc1/pilot/api/data.py#L372)
So if you remove the ip6 field from the dictionary, I (or better, you) will have to update the pilot code as well. From which version?

I wonder if I could just pull this function from the rucio API instead of duplicating it, which otherwise inevitably leads to situations like this. RUCIO_LAT/LONGITUDE don't appear to exist on the WNs though. I suppose you will have them set later..?

ivukotic: can someone please explain to how are these two functions related? Paul, what exactly pilot uses ip and ipv6 values it obtains?

bari12: As far as I can see Paul uses that "duplicated" `detect_client_location` to feed the `list_replica` calls the pilot makes, it is not used otherwise. To make use of RUCIO_LAT/LONG he would have to update that one as well. (Alternatively, you could also just import the one from `rucio.common.utils`)
For ip4/6 you actually don't need to change it, since for your usecase (list_replicas) the server will anyway use the incoming ip.

Coming back to `detect_client_location` itself:
I lean to merging this as well and just having `ip` (Precedence maybe ip6>ip4); Although public API, I don't think this is used by anyone, so I don't think it will break things. We should make it clear in the description how `ip` is generated and that it will simply be the device address, not location in a global sense.
As far as I can see this should work and will not break anything, as it is not used anywhere else.

ivukotic: So, is there anything you want me change?

bari12: Yes, bring the ip detection back to `detect_client_location`; Just make it one field `ip`, which gives either ip4 or ip6, if both available, ip6.

ivukotic: done now.

bziemons (approved):

Issue #2159, opened 2019-02-08
Add functionality in the WebUI to see that a certain transfer/request is WAITING


bari12:
Motivation
----------
Currently it is not clear why a certain request is REPLICATING, but did not issue a FTS transfer yet. We should expose that a request is in WATING (or even MISMATCH) state. What would also be good to maybe show the queue length, although this might be more difficult as it is depending on throttle configuration.

jwackito:
Hi @bari12. This change should be reflected where exactly? I mean, I can see the state of the rules in the  'Data Transfers/List my rules' menu. Am I looking at the right place?

bari12:
I think it would be best to get this in the rule view, when you open the locks panel
e.g. https://rucio-ui.cern.ch/rule?rule_id=14e9f8c820e44db88ba6911bc40eee9a#locks

Here for each file, in the RSEs column the lock status will be shown. Yellow for REPLICATING. However there are different request states for the REPLICATING lock which could be: WAITING, when the file is waiting in rucio to be released, Queued when it is still in the belly of rucio, but not submitted to FTS yet, or SUBMITTED, when it is actually within FTS. (At that point, in the FTS Monitoring column, you will be able to generate a FTS monitoring link)

I think we can perhaps expose this request state information next to the RSE name in the table.

jwackito:
Hi @bari12. Core changes are in a pull request, but review is requested. JavaScript part still giving some errors, I'm working on it, will pull request soon.

maany:
Hey @jwackito , could you point me to the PR. Also, are you still working on this?

PR #4663, opened 2021-06-08
[Core & internals]: Lock query to include request state WIP #2159

jwackito:
REVIEW REQUESTED!

### DIFF
--- a/lib/rucio/core/lock.py
+++ b/lib/rucio/core/lock.py
@@ -137,15 +137,25 @@ def get_replica_locks_for_rule_id(rule_id, session=None):
 
     locks = []
 
-    query = session.query(models.ReplicaLock).filter_by(rule_id=rule_id)
-
-    for row in query:
-        locks.append({'scope': row.scope,
-                      'name': row.name,
-                      'rse_id': row.rse_id,
-                      'rse': get_rse_name(rse_id=row.rse_id, session=session),
-                      'state': row.state,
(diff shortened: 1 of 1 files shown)


lib/rucio/core/lock.py
```diff
@@ -137,15 +137,25 @@ def get_replica_locks_for_rule_id(rule_id, session=None):
...
-    query = session.query(models.ReplicaLock).filter_by(rule_id=rule_id)
-
-    for row in query:
-        locks.append({'scope': row.scope,
-                      'name': row.name,
-                      'rse_id': row.rse_id,
-                      'rse': get_rse_name(rse_id=row.rse_id, session=session),
-                      'state': row.state,
-                      'rule_id': row.rule_id})
+    query = session.query(models.ReplicaLock, models.Request).join(
```
bari12, line 140: Hi JJ, I don't think it will work like this.
- The `get_replica_locks_for_rule_id` call should probably also return locks, if you don't have any requests, so you need to do a left outer join, otherwise it won't work.
- The join condition on rule_id, while in most cases this would work, in some it will not, because for some locks the request is actually from a different rule, so it won't match. you have to do the `ON` filter of the join on `replicaLock.scope=request.scope and replicalock.name=request.name`. But the filter condition can be on replicalock.rule_id, that is ok. 
Otherwise it is good, although a test wood be nice.

PR #4664, opened 2021-06-08
Release management: Move all requirements; Fix #4456

### LINKED ISSUE #4456: setup.py and requirements.txt overhaul
Motivation
----------
Current usages of the setup.py files and the `etc/pip-requires` files make it hard for new developers and extending the CI. Having a single requirements.txt will all dependencies and their version ranges (matching the supported dependency versions on target OSs like centos7) and letting the setup.py files decide which of these dependencies to ship for servers/clients and which are extra (optional) dependencies, seems like the best solution.

Modification
------------
A single `requirements.txt` list all the dependencies to develop with Rucio and they can be installed into a fresh virtual environment with `pip install -r requirements.txt`. The different `setup.py` files use a shared approach to determine which dependencies in what versions are extras [(1)] and hard requirements [(2)]. More information: https://packaging.python.org/discussions/install-requires-vs-requirements/

[(1)]: https://setuptools.readthedocs.io/en/latest/userguide/dependency_management.html?highlight=install_requires#optional-dependencies
[(2)]: https://setuptools.readthedocs.io/en/latest/userguide/dependency_management.html?highlight=install_requires#declaring-required-dependency

bziemons:
Moving all requirements with version information to the default location: `requirements.txt`.
The new `setuputil.py` will be the place to select client/server (and their extras) specific requirements.

~~Currently the integration tests fail. Adjustments need to be made to all container `Dockerfile`s to fix this. I will prepare a PR there.~~ Solved by https://github.com/rucio/containers/pull/137


requirements.txt
```diff
@@ -0,0 +1,67 @@
+# All dependencies needed to run rucio client (and server/daemons) should be defined here
```
bari12, line 1: Just a quick question, did you update any of the requirements here, or is this the state of the etc/pip* files?

setup_rucio.py
```diff
@@ -16,190 +16,92 @@
...
-
-    :param cmd: the git command.
-    :return: Output of the git command.
-    """
-    output = subprocess.Popen(["/bin/sh", "-c", cmd], stdout=subprocess.PIPE)
-    return output.communicate()[0].strip()
-
-
 if using_pip and using_git:
+    # FIXME: duplicate in tools/generate_version.py
```
bari12: I wonder if we still need this?
The vcversion file I usually create outside of the procedure (with the `tools/generate_version.py`) anyway. So that part doesn't seem useful anymore.

tools/install_venv.py
```diff
@@ -1,223 +0,0 @@
-# Copyright 2011-2020 CERN for the benefit of the ATLAS collaboration.
```
bari12, line 1: Note, this might still be mentioned in some documentations, need to remove it there.

bari12 (commented):
Looks good to me. Just two comments/questions, but otherwise I think this can be merged.

tools/install_venv.py
bziemons: will have a look

setup_rucio.py
bziemons: I leave that for you to decide. If you dont need it here, I can remove it now

requirements.txt
bziemons: I removed `setuptools` (since we don't need it for the actual code and it is obvious for the setup.py files), but otherwise only the extras were added.

setup_rucio.py
bari12: Let's remove it then, I think it's just clutter. Thanks :-)

tools/install_venv.py
bziemons: Actually, venv is not even mentioned in any documentation piece neither here nor in rucio-documentation – that we might want to change

bari12 (approved):
Very nice 🥳  Thanks

PR #4666, opened 2021-06-09
Add index to services column in messages table

cserf:
Add index to services column in messages table


bari12:
This is failing on oracle due to the size of the services column (2048) + event_Type (1024). I think we should consider decreasing the size of the services column, especially with the new model we have in mind.

bari12 (approved):

Issue #4624, opened 2021-05-14
Registering lfn2pfn algorithm in policy package causes import loop


jamesp-epcc:
My original intention was that policy packages would register any lfn2pfn algorithms when first loaded, by calling `rucio.rse.protocols.protocol.RSEDeterministicTranslation.register`. But this can cause a circular import (because it imports from `rucio.core.rse`, but `rucio.core.rse` indirectly causes `rucio.common.schema` to be imported, which in turn loads the policy package) and loading the policy package fails.
To fix this I propose having the policy package define an optional function that returns a dictionary of lfn2pfn algorithms that it wants to register, and have Rucio call this function if it exists, and add the result to the main registry. We should probably do the same for SURL functions as well. This should remove the need for policy packages to import Rucio modules when they first load, removing the potential for circular imports.

maxnoe:
Python actually has a mechanism for plugins supported in the standard library: entry points. This would allow the policy packages to just define the free functions, mark them in their project configuration as entry points for rucio and rucio could then load them.


For the example of lfn2pfn, the policy package would have a function:

# rucio_policy_foo.py
```python
def lfn2pfn(scope, name, rse, rse_attrs, protocol_attrs):
    return name
```

And would configure the entry point:
# pyproject.toml of rucio_policy_foo
```toml
[project.entry-points.rucio_fn2pfn]
foo = "rucio_policy_foo:lfn2pfn"
```

In the rucio core package, you would then get the lfn2pfn algorithm via the importlib.metadata std lib module:

```python
from importlib.metadata import entry_points

def get_lfn2pfn_algorithm(name):
    (lfn2pfn, ) = entry_points(group='rucio_lfn2pfn', name=name)
    return lfn2pfn  
```

PR #4667, opened 2021-06-10
Core & Internals: Make Rucio query policy package for surl and lfn2pfn algorithms #4624

jamesp-epcc:
This pull request changes the policy package code so that Rucio will query the policy package(s) for any lfn2pfn and surl algorithms that should be registered. This avoids the import loops that can happen if the policy package does the registration when it loads. I will submit another PR to update the documentation accordingly.

### DIFF
--- a/lib/rucio/common/utils.py
+++ b/lib/rucio/common/utils.py
@@ -579,7 +579,7 @@ def _register_policy_package_surl_algorithms():
                 _SURL_ALGORITHMS.update(module.get_surl_algorithms())
         except (NoOptionError, NoSectionError, ImportError):
             pass
-        
+
     from rucio.common import config
     from rucio.core.vo import list_vos
     try:
--- a/lib/rucio/common/utils.py
+++ b/lib/rucio/common/utils.py
@@ -571,6 +571,7 @@ register_surl_algorithm(construct_surl_BelleII, 'BelleII')
 
 def _register_policy_package_surl_algorithms():
     def try_importing_policy(vo=None):
+        import importlib
         try:
             package = config.config_get('policy', 'package' + ('' if not vo else '-' + vo['vo']))
             module = importlib.import_module(package)
@@ -581,7 +582,6 @@ def _register_policy_package_surl_algorithms():
         
     from rucio.common import config
     from rucio.core.vo import list_vos
-    import importlib
     try:
         multivo = config.config_get_bool('common', 'multi_vo')
     except (NoOptionError, NoSectionError):
--- a/lib/rucio/common/utils.py
+++ b/lib/rucio/common/utils.py
@@ -554,6 +554,7 @@ def construct_surl_BelleII(dsn, filename):
 
 _SURL_ALGORITHMS = {}
 _DEFAULT_SURL = 'DQ2'
+_loaded_policy_modules = False
 
 
 def register_surl_algorithm(surl_callable, name=None):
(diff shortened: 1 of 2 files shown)


lib/rucio/common/utils.py
```diff
@@ -107,6 +107,12 @@
 except ImportError:
     # Python 3
     import urllib.parse as urlparse
+try:
+    # Python 2
+    from ConfigParser import NoOptionError, NoSectionError
+except ImportError:
+    # Python 3
+    from configparser import NoOptionError, NoSectionError
```
bziemons: please use `from six.moves.configparser import NoOptionError, NoSectionError` since its more straightforward and pylint likes it more :)

lib/rucio/common/utils.py
```diff
@@ -618,9 +625,42 @@ def register_surl_algorithm(surl_callable, name=None):
...
+        # policy package per VO
+        vos = list_vos()
+        for vo in vos:
+            try:
+                package = config.config_get('policy', 'package-' + vo['vo'])
+                module = importlib.import_module(package)
+                if hasattr(module, 'get_surl_algorithms'):
+                    _SURL_ALGORITHMS.update(module.get_surl_algorithms())
+            except (NoOptionError, NoSectionError, ImportError):
+                pass
```
bziemons: try to extract same code like this to a function (probably sub-function of `_register_policy_package_surl_algorithms`), something like:

```py
def try_importing_policy(vo=None):
    try:
        package = config.config_get('policy', 'package' + ('' if not vo else '-' + vo['vo']))
        module = importlib.import_module(package)
        if hasattr(module, 'get_surl_algorithms'):
            _SURL_ALGORITHMS.update(module.get_surl_algorithms())
    except (NoOptionError, NoSectionError, ImportError):
        pass
  ```

also does `_SURL_ALGORITHMS` actually look correct after this? It looks like the different vo policies can override each other and a different order could result in a different `_SURL_ALGORITHMS` dictionary.

File: lib/rucio/rse/protocols/protocol.py [hunk 3]
```diff
@@ -242,6 +244,34 @@ def _module_init_(cls):
...
+                package = config.config_get('policy', 'package')
+                module = importlib.import_module(package)
+                if hasattr(module, 'get_lfn2pfn_algorithms'):
+                    RSEDeterministicTranslation._LFN2PFN_ALGORITHMS.update(module.get_lfn2pfn_algorithms())
+            except (NoOptionError, NoSectionError, ImportError):
+                pass
+        else:
+            # policy package per VO
+            vos = list_vos()
+            for vo in vos:
```
bziemons, line 269: similar as above, same code. Also similar issue with overriding the `RSEDeterministicTranslation._LFN2PFN_ALGORITHMS` dictionary?

lib/rucio/common/utils.py
jamesp-epcc: Done :)

lib/rucio/common/utils.py
jamesp-epcc: I have factored out the common code into a function as suggested. Yes, it is true that different VO policies can override each other. The documentation instructs policy package authors to ensure their function names are prefixed with the VO name to avoid conflicts. Maybe this should also be enforced by the code? (it wasn't possible with the old system, but it would be with this one).

lib/rucio/rse/protocols/protocol.py
jamesp-epcc: As above.

lib/rucio/common/utils.py
bziemons: > The documentation instructs policy package authors to ensure their function names are prefixed with the VO name to avoid conflicts.

that makes sense and it sounds good enough for this PR :+1:

It sounds like a good idea to enforce it (in another PR), thoughts @bari12 ?

bziemons (dismissed):

bari12 (approved):

Issue #4647, opened 2021-05-27
Add new index on messages.services


bari12:
In preparation for hermes2 design update, add a new index on services column.

PR #4673, opened 2021-06-14
[Database] rse_id added as index in replicas table

jwackito:
Added alembic schema index rse_id in the replicas table


bari12:
You will have to rebase this and adapt, since I just merged #4647 
In general, you also need to add this to the `models.py` and to the oracle schema file in etc/

The header in the file seems also outdated, but that might just be an issue with the template which we didn't update.

PR #4675, opened 2021-06-15
Testing: fix integration test timeouts #4674

### LINKED ISSUE #4674: timeouts in integration_tests
Motivation
----------
There is some flakiness in the integration tests as timeouts sometimes occur when certain tests(`test_tpc` and `test_conveyor`) are waiting for `<some_fts_activity>`.

This comes from the fact that our integration test for the master branch is currently failing after Martin merged a PR:
https://github.com/rucio/rucio/runs/2821434015?check_suite_focus=true
Interestingly, these tests ran without issues for the PR's checks :
https://github.com/rucio/rucio/pull/4505/checks?check_run_id=2746593221

When I started looking into this, test_tpc was failing. It was due to a timeout while polling fts to get a transfer_id for the newly created rule:
https://github.com/rucio/rucio/blob/master/lib/rucio/tests/test_tpc.py#L58

I re-ran the checks on the PR to confirm the suspicion, and this time it failed on `test_conveyor` instead. The root cause is a timeout in `__wait_for_replica_transfer`  function that waits for an fts transfer to complete.
https://github.com/rucio/rucio/blob/master/lib/rucio/tests/test_conveyor.py#L37

Modification
------------
After discussing with @bziemons and @rcarpa, we figured that the `test_conveyer` needs at least a 5-second timeout on a CERN OpenStack m2.large instance, while test_tpc needs at least 9 seconds. 

We have decided to increase the timeout in `__wait_for_replica_transfer` to 60s and change the `or` logic in test_tpc's `list_fts_transfer` timeout condition to a `nand` to ensure there's a max 60 seconds before timeout occurs.

If the timeouts occur again, after these fixes, we might look into the resources available on the runners for GH Actions and/or investigate if any previous submitter runs did not clean up itself correctly.

maany:
close #4674

### DIFF
--- a/lib/rucio/tests/test_tpc.py
+++ b/lib/rucio/tests/test_tpc.py
@@ -55,14 +55,14 @@ def check_url(pfn, hostname, path):
     assert path in pfn
 
 
-def list_fts_transfer(timeout=60, max_attempts=25):
+def list_fts_transfer(timeout=60, min_attempts=20):
     running_time = 0
     request_id = None
     request_status = None
     attempt = 1
 
     time_start = time.time()
-    while running_time < timeout and attempt <= max_attempts:
+    while running_time < timeout and attempt <= min_attempts:
         rcode, out = run_cmd_process("/usr/bin/python2 /usr/bin/fts-rest-transfer-list -v -s https://fts:8446")
         if "Request ID" in out:
             request_id = re.search("Request ID: (.*)", out).group(1)
--- a/lib/rucio/tests/test_tpc.py
+++ b/lib/rucio/tests/test_tpc.py
@@ -62,8 +62,8 @@ def list_fts_transfer(timeout=60, max_attempts=10):
     attempt = 1
 
     time_start = time.time()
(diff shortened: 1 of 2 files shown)


lib/rucio/tests/test_tpc.py
```diff
@@ -62,8 +62,8 @@ def list_fts_transfer(timeout=60, max_attempts=10):
     attempt = 1
 
     time_start = time.time()
-    while running_time < timeout or attempt < max_attempts:
-        rcode, out = run_cmd_process("/usr/bin/python2 /usr/bin/fts-rest-transfer-list -v -s https://fts:8446", timeout=30)
+    while not(running_time >= timeout and attempt > max_attempts):
```
bziemons: `running_time < timeout and attempt <= max_attempts` ? Also `max_attempts` => `min_attempts`

maany: you're right, it expands to the same logic. haha!

bziemons (approved):

Issue #4680, opened 2021-06-16
Upgrade urllib due to CVE-2021-33503


bari12:
Motivation
----------
https://github.com/advisories/GHSA-q2q7-5pp4-w6pg

Modification
------------
Upgrade urllib to at least `1.26.5`

bari12:
backport 1.20.16
backport 1.23.15

PR #4682, opened 2021-06-16
Release: urllib3 security upgrade #4680

bari12:


bziemons:
Needs the same `requests` update as #4683

PR #4687, opened 2021-06-17
[Database] Added rseid index on replicas table

jwackito:


bari12:
Forgot to tell you, you have to update https://github.com/rucio/rucio/blob/master/lib/rucio/alembicrevision.py too
:-)

lib/rucio/db/sqla/migrate_repo/versions/52153819589c_add_rse_id_to_replicas_table.py
```diff
@@ -0,0 +1,47 @@
...
+down_revision = '30fa38b6434e'
+
+
+def upgrade():
+    '''
+    Upgrade the database to this revision
+    '''
+
+    if context.get_context().dialect.name in ['oracle', 'mysql', 'postgresql']:
+        create_foreign_key('REPLICAS_RSE_ID_FK', 'replicas', 'rses', ['rse_id'], ['id'])
```
bari12: I suppose you do the FK changes for postgres? If so, we should probably isolate it to that.

bari12 (approved):

bari12:
Merged as feature.

PR #4688, opened 2021-06-17
Database: Fix oracle hint for list_replicas_for_files

bziemons:
The hint was on a field and the query was compiled without selecting the dialect.
Scanning through the code base, this seems to be the only occurrence with these problems.


rcarpa (commented):
thank you!

rcarpa (approved):

PR #4690, opened 2021-06-18
Clients: deprecate unavailable argument for list-replicas. Closes #4603

### LINKED ISSUE #4603: Further list_replicas changes
Motivation
----------
With #4353 the core `list_replicas` method was changed to, per default, filter replicas whose RSEs are read-disabled. This changes however the default behaviour from `rucio list-file-replicas` as this method should list replicas on read-disabled RSEs too.

We should do the following steps

1. in https://github.com/rucio/rucio/blob/57575a05ece137daac27bb0df28f0ca1a4faf560/lib/rucio/client/replicaclient.py#L121-L126
set `unavailable` per default as `True`. 
2. As `unavailable` is quite an ambiguous term, with 1.26 we should replace `unavailable` with `ignore_availability`
3. On the REST level we need to keep `unavailable`, for backwards-compatibility reasons, but should additionally add `ignore_availability` as well. This one takes then precedence over `unavailable` 
4. On an API and core level https://github.com/rucio/rucio/blob/57575a05ece137daac27bb0df28f0ca1a4faf560/lib/rucio/core/replica.py#L1238-L1244 the method parameters are ambiguous. Currently there is `ignore_availability`, `unavailable` and `all_states` which somewhat do similar functionality. `unavailable` should be removed entirely and only `ignore_availability` and `all_states ` should stay.
5. #3431 should be done, where we expand the `list-file-replicas` with actual state and RSE status information.

rcarpa:
unavailable has very similar behavior to all_states: it lists replicas
in following 3 states: Available/Unavailable/Copying. The argument
is confusing because it makes you think it ignores rse availability.
Remove the argument from clients, but leave it on the API side to keep
it compatible with older clients.

There are no references to `unavailable` in bin/rucio

### DIFF
--- a/lib/rucio/api/replica.py
+++ b/lib/rucio/api/replica.py
@@ -152,7 +152,7 @@ def list_replicas(dids, schemes=None, unavailable=False, request_id=None,
 
     :param dids: The list of data identifiers (DIDs).
     :param schemes: A list of schemes to filter the replicas. (e.g. file, http, ...)
-    :param unavailable: Also include unavailable replicas in the list.
+    :param unavailable: (deprecated) Also include unavailable replicas in the list.
     :param request_id: ID associated with the request for debugging.
     :param all_states: Return all replicas whatever state they are in. Adds an extra 'states' entry in the result dictionary.
     :param rse_expression: The RSE expression to restrict replicas on a set of RSEs.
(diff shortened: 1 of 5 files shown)


bziemons (dismissed):
I am guessing the breakage of the client API (removing a parameter) is as planned. I couldn't really read that from the mentioned issue. Rest looks good to me.

bari12 (approved):

PR #4691, opened 2021-06-18
Core & Internals: simplify list_replica response creation. Closes #4473

### LINKED ISSUE #4473: Refactor _list_replicas and correctly implement xroot prioritization for archives

rcarpa:
Extract pfn/rse sorting to a separate function and call it in the 2
needed places (each time before yielding the file).

Only resolve parents once per scope/name. As a side note, parents
are not resolved for archives, but this never worked. I don't see
any regression about it in #4473.

### DIFF
--- a/lib/rucio/core/replica.py
+++ b/lib/rucio/core/replica.py
@@ -956,6 +956,31 @@ def get_multi_cache_prefix(cache_site, filename, logger=logging.log):
     return ''
 
 
+def _sort_replica_file_pfns(file):
+    """
+    Helper function to be executed after analyzing the last replica of a file (same scope:name)
+
+    """
+    if not file.get('pfns'):
+        return
+
+    # set the total order for the priority
+    # --> exploit that L(AN) comes before W(AN) before Z(IP) alphabetically
+    # and use 1-indexing to be compatible with metalink
+    tmp = sorted([(file['pfns'][p]['domain'], file['pfns'][p]['priority'], p) for p in file['pfns']])
+    for i in range(0, len(tmp)):
+        file['pfns'][tmp[i][2]]['priority'] = i + 1
+
+    # also sort the pfns inside the rse structure
+    file['rses'] = {}
+    rse_pfns = sorted((file['pfns'][pfn]['rse_id'], file['pfns'][pfn]['priority'], pfn) for pfn in file['pfns'])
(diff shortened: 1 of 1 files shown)


lib/rucio/core/replica.py
```diff
@@ -1155,104 +1180,38 @@ def _list_replicas(dataset_clause, file_clause, state_clause, show_pfns,
...
-        tmp = sorted([(file['pfns'][p]['domain'], file['pfns'][p]['priority'], p) for p in file['pfns']])
-        for i in range(0, len(tmp)):
-            file['pfns'][tmp[i][2]]['priority'] = i + 1
-
-    if 'scope' in file and 'name' in file:
-        file['rses'] = {}
-
-        # don't forget to resolve parents for the last replica
-        if resolve_parents:
+            file['pfns'], file['rses'], file['states'] = {}, defaultdict(list), {}
```
mlassnig, line 1191: the `file['states']` was previously `file['states'] = {rse_id: str(state.name if state else state)}`, now it's just `{}`

on purpose?

rcarpa: it is done a couple of lines below.

lib/rucio/core/replica.py
```diff
@@ -1155,104 +1180,38 @@ def _list_replicas(dataset_clause, file_clause, state_clause, show_pfns,
...
+
+        if rse_id:
+            file['rses'][rse_id] += list(set([tmp_pfn[0] for tmp_pfn in pfns]))
+            file['states'][rse_id] = str(state.name if state else state)
```
rcarpa, line 1195: here

mlassnig (approved):

PR #4692, opened 2021-06-21
Undertaker : Force deletion of replicas : Closes #4689

### LINKED ISSUE #4689: Undertaker: Force deletion of replicas
Motivation
----------
There's a loophole in the undertaker workflow. If a DID collection has N rules and some of these rules are deleted without the `--purge-replicas` option, all the files replicas from these collection will stay with a non-Epoch tombstone. When someone erase the DID, the undertaker remove all the rules with `--purge-replicas` that get an Epoch tombstone, but nothing is done for the replicas from the previously expired rules that can stay around forever.

Modification
------------
Set Epoch tombstone to all the file replicas from the DID that have no locks.

cserf:
Undertaker : Force deletion of replicas : Closes #4689

### DIFF
--- a/lib/rucio/core/did.py
+++ b/lib/rucio/core/did.py
@@ -607,6 +607,7 @@ def delete_dids(dids, account, expire_rules=False, session=None, logger=logging.
     not_purge_replicas = []
     did_followed_clause = []
     metadata_to_delete = []
+    file_content_clause = []
 
     archive_dids = config_core.get('undertaker', 'archive_dids', default=False, session=session)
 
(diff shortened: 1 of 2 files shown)


lib/rucio/core/did.py
```diff
@@ -692,7 +694,19 @@ def delete_dids(dids, account, expire_rules=False, session=None, logger=logging.
         with record_timer_block('undertaker.parent_content'):
             for parent_did in session.query(models.DataIdentifierAssociation).filter(or_(*parent_content_clause)):
                 existing_parent_dids = True
-                detach_dids(scope=parent_did.scope, name=parent_did.name, dids=[{'scope': parent_did.child_scope, 'name': parent_did.child_name}], session=session)
+                detach_dids(scope=parent_did.scope, name=parent_did.name, dids=[{'scope': parent_did.child_scope, 'name': parent_did.child_type}], session=session)
+
+    # Set Epoch tombstone for the files replicas inside the did
+    if config_core.get('undertaker', 'remove_all_files', default=False, session=session) and file_content_clause:
```
bari12: remove_all_files I find not very clear as an option, I think we should find another name.

lib/rucio/core/did.py
```diff
@@ -692,7 +694,19 @@ def delete_dids(dids, account, expire_rules=False, session=None, logger=logging.
...
+        with record_timer_block('undertaker.file_content'):
+            file_replicas_clause = [and_(models.RSEFileAssociation.scope == cont['child_scope'], models.RSEFileAssociation.name == cont['child_name']) for cont in session.query(models.DataIdentifierAssociation).filter(or_(*file_content_clause))]
+            none_value = None  # Hack to get pep8 happy
+            for chunk in chunks(file_replicas_clause, 100):
```
bari12, line 704: Doesn't this ignore the case where the file is also in another dataset? In that case the file is a correct secondary replica?

bari12 (changes requested):

lib/rucio/core/did.py
cserf: Changed to `purge_all_replicas`

bari12 (approved):

PR #4694, opened 2021-06-21
Database: Update SQLAlchemy to 1.4

bziemons:
- Replace query_bunches with Result.partitions
- Query the table instead of a column when deleting/updating
- Add synchronize_session to nested update queries
- Add try_drop_constraint for oracle
- Fix deprecation warning from empty 'or'
- Fix delete_qos_policy
- Stop leaking database objects from core
- Fix redirect_auth_oidc

Depends on #4688

### DIFF
--- a/lib/rucio/api/subscription.py
+++ b/lib/rucio/api/subscription.py
@@ -25,14 +25,17 @@
 # - Patrick Austin <patrick.austin@stfc.ac.uk>, 2020
 # - James Perry <j.perry@epcc.ed.ac.uk>, 2020
 
+from collections import namedtuple
 from json import dumps, loads
-from sqlalchemy.util import KeyedTuple
 
 from rucio.api.permission import has_permission
 from rucio.common.exception import InvalidObject, AccessDenied
 from rucio.common.schema import validate_schema
 from rucio.common.types import InternalAccount, InternalScope
 from rucio.core import subscription
+
+
+SubscriptionRuleState = namedtuple('SubscriptionRuleState', ['account', 'name', 'state', 'count'])
 
 
 def add_subscription(name, account, filter, replication_rules, comments, lifetime, retroactive, dry_run, priority=None, issuer=None, vo='def'):
@@ -193,7 +196,7 @@ def list_subscription_rule_states(name=None, account=None, vo='def'):
     :param name: Name of the subscription
     :param account: Account identifier
     :param vo: The VO to act on.
-    :returns: List with tuple (account, name, state, count)
+    :returns: Sequence with SubscriptionRuleState named tuples (account, name, state, count)
     """
     if account is not None:
         account = InternalAccount(account, vo=vo)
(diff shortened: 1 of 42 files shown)


rcarpa (commented):
LGTM, but it mostly touches to code I've never seen, so I'm not very confident.

lib/rucio/api/subscription.py
```diff
@@ -35,6 +35,9 @@
 from rucio.core import subscription
 
 
+SubscriptionRuleState = namedtuple('SubscriptionRuleState', ['account', 'name', 'state', 'count'])
```
mlassnig, line 38: Is this only used in here? I thought we had something like this in common/ ?

lib/rucio/common/utils.py
```diff
@@ -128,28 +127,6 @@ def build_url(url, path=None, params=None, doseq=False):
     return complete_url
 
 
-def oidc_identity_string(sub, iss):
```
mlassnig, line 131: why is this removed? (the other one below i understand)

lib/rucio/db/sqla/types.py
```diff
@@ -46,8 +45,11 @@ class GUID(TypeDecorator):
     otherwise uses CHAR(32), storing as stringified hex values.
 
     """
+
     impl = CHAR
 
+    cache_ok = True
```
mlassnig, line 51: what and where is this used for?

lib/rucio/db/sqla/util.py
```diff
@@ -302,3 +302,17 @@ def json_implemented(session=None):
...
+def try_drop_constraint(constraint_name, table_name):
+    """
+    Tries to drop the given constrained and returns successfully if the
+    constraint already existed on Oracle databases.
+
+    :param constraint_name: the constraint's name
+    :param table_name: the table name where the constraint resides
+    """
+    try:
+        op.drop_constraint(constraint_name, table_name)
```
mlassnig, line 316: does this work without explicitly specifying the constraint type? previously this was required to add

lib/rucio/common/utils.py
bziemons: moved it to the oidc module, I dont think it belongs in client/server common code

lib/rucio/db/sqla/types.py
bziemons: That's a new SQLAlchemy feature: https://docs.sqlalchemy.org/en/14/core/custom_types.html#sqlalchemy.types.TypeDecorator.cache_ok

(not specifying it spammed warnings)

lib/rucio/db/sqla/util.py
bziemons: that is only necessary for mysql, according to the documentation https://alembic.sqlalchemy.org/en/latest/ops.html?highlight=drop_constraint#alembic.operations.Operations.drop_constraint (and confirmed by the tests)

lib/rucio/api/subscription.py
bziemons: can you point me to it? I had to replace the KeyedTuple, because that is no more

bziemons: can you point me to it? I had to replace the KeyedTuple, because that is no more

mlassnig: https://github.com/rucio/rucio/blob/master/lib/rucio/common/constants.py

--> e.g. it has the FTS States there

bziemons: yeah, this is only used here and does not break the API, so that you can still use

```py3
for row in list_subscription_rule_states():
    print(row.account, row.name, row.state, row.count)
```
or similar code. Contrary to `FTS_STATES`, this isn't constant so I would not put it into the constants module

mlassnig (approved):

bari12 (approved):
👍 I am a bit skeptical about the relative release specifier in the requirements. I agree that's what should be used, but I think we did have issues in the past even with patch release changes.

PR #4702, opened 2021-06-28
Clients: recursive upload of a folder into collections #4616

### LINKED ISSUE #4616: CLI/python Clients - Recursive Upload of an input folder
Motivation
----------
Currently is possible to upload the files contained in a folder. In order to facilitate the upload tasks we are interested in a recursive exploration of a given input folder. The goal is to replicate the input folder structure by creating sub-datasets then performing normal upload for the found files attaching them to the correct sub-datasets.

Modification
------------
Enhance the "folder" input management of the upload command in the CLI/python Clients.

davidpob99:
### Clients: recursive upload of a folder into collections. Closes #4616 
The changes allow to replicate the folder structure (recursively) into collections.

Add a new flag in rucio CLI client: `--recursive`. It needs at least one argument and it should be a folder.
It is not possible to create the folder structure into collections with a non-deterministic way.
A folder only can have either other folders inside or files, but not both of them:
- If it has folders, the root folder will be a container
- If it has files, the root folder will be a dataset
- If it is empty, it does not create anything

Add some tests.

### DIFF
--- a/lib/rucio/client/uploadclient.py
+++ b/lib/rucio/client/uploadclient.py
@@ -738,7 +738,6 @@ class UploadClient:
 
         :param: item        dictionary containing all descriptions of the files to upload
         """
-        logger = self.logger
         files = []
         datasets = []
         containers = []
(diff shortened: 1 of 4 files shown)


lib/rucio/client/uploadclient.py
```diff
@@ -720,3 +727,80 @@ def _send_trace(self, trace):
...
+                    datasets.append({'scope': scope, 'name': root2, 'rse': rse})
+                    logger(logging.DEBUG, 'Appended dataset with DID %s:%s' % (scope, path))
+                elif len(dirs) > 0:
+                    containers.append({'scope': scope, 'name': root2})
+                    logger(logging.DEBUG, 'Appended container with DID %s:%s' % (scope, path))
+                else:
+                    logger(logging.WARNING, 'The folder %s is empty, skipping' % root)
+            if len(dirs) > 0:
+                for directory in dirs:
+                    root_dir, dir_dir, fnames_dir = next(os.walk(os.path.join(root, directory)))
```
rcarpa: I don't understand why these additional calls to os.walk are needed. The initial one should walk recursively all subdirectories. Also, why is there any distinction in how the start directory is treated compared to its subdirectories. Can you please try to unify these code paths as much as possible?

lib/rucio/tests/test_bin_rucio.py
```diff
@@ -1525,3 +1526,116 @@ def test_list_account_usage(self):
         assert re.search('.*{0}.*{1}.*{2}.*{3}'.format(rse_exp, usage, global_limit, global_left), out) is not None
         self.account_client.set_local_account_limit(account, rse, -1)
         self.account_client.set_global_account_limit(account, rse_exp, -1)
+
+    def test_upload_recursive_ok(self):
```
rcarpa: Nice! 
would be a good bonus if you could adapt the `file_factory` fixture added by Mayank and used in test_upload.py to generate the folders and files as needed for your new tests; then use it here. We are gradually migrating to using pytest and fixtures to avoid code duplication in tests.

davidpob99: I have adapted the `file_factory` fixture but I am unable to use it inside `TestBinRucio` because it inherits `unittest.TestCase` and according to the [docs](https://pytest.org/en/latest/how-to/unittest.html): 

> unittest.TestCase methods cannot directly receive fixture arguments as implementing that is likely to inflict on the ability to run general unittest.TestCase test suites.

rcarpa: Hi. Thank you for doing this work. It will allow to easier migrate these tests to pytest when we'll do it. In the meantime you can just use the class directly like this https://github.com/rucio/rucio/blob/5d1085b03e376307fa33b94b52a25fc8f77fc9e0/lib/rucio/tests/conftest.py#L167 .

rcarpa (commented):
LGTM

bari12 (approved):

Issue #4511, opened 2021-03-31
[conveyor] multihop with duplicate requests in path may clog submitter


nsmith-:
Motivation
----------
In https://github.com/rucio/rucio/blob/740b825cac747280ca159089bca4635264c07fc1/lib/rucio/core/transfer.py#L1179 if a hop in a multihop path has a duplicate request, we need to fail the hop chain to let the duplicate request go through and try again later. However, this call to fail the request can never succeed since moving a request to `FAILED` state requires an `external_id` in https://github.com/rucio/rucio/blob/740b825cac747280ca159089bca4635264c07fc1/lib/rucio/core/request.py#L481-L485
Now imagine we have a big backlog of requests, where the duplicate request for the hop is also in queued state, and by chance the entire batch of requests currently processing has this error: now the conveyor will be stuck in a loop it cannot get out of since the same batch will continually be re-analyzed and duplicate request will never move forward.

The protection in `set_request_state` appears intended to force developers to recognize that there is an external request attached and it must also be canceled and cleaned up properly. However, we are trying to move the request state from `QUEUED` to `FAILED` and thus have no external request attached at this point.

Modification
------------
I propose we check in `set_request_state` whether `external_id` is null in the database, and if so it should be safe to allow the state transition. If not, we can raise the error as it currently does.

nsmith-:
Here's how I am proposing to patch our deployment: https://github.com/ericvaandering/rucio/pull/5/files

PR #4708, opened 2021-06-30
Implement a patch for rucio/rucio#4511

nsmith-:
This patch has been running well in production for CMS. Let me know what kind of tests would be appropriate.

cc @ericvaandering


rcarpa:
Thanks for submitting the PR. I think it's a good work-around until a real fix is implemented for the issue of duplicate requests.

@bari12 , please don't close the associated ticket when merging the PR. I think that the real root problem is trying to "queue_requests" which already exist in the database. We should avoid this as much as possible and I think there is more work to be done for #4511. I am delaying it until all the pending transfer work is merged.

bari12 (approved):

PR #4713, opened 2021-07-01
Rules: Allow repair when ignore_availability=True

### LINKED ISSUE #4152: Rule repair does not respect ignore_availability setting
Motivation
----------
If a rule was made with `ignore_availability=True` and a replica lock becomes stuck, any subsequent repair operation should consider `ignore_availability`. Currently it seems that RSE availability is always required before a repair can be submitted, per https://github.com/rucio/rucio/blob/51dd4f09c6541a10186b1e1360e38b6205ea5f28/lib/rucio/core/rule_grouping.py#L770

There may be some question if this is done on purpose for any reason.

Modification
------------
Probably just a change to https://github.com/rucio/rucio/blob/51dd4f09c6541a10186b1e1360e38b6205ea5f28/lib/rucio/core/rule_grouping.py#L770

nsmith-:
Fixes #4152

### DIFF
--- a/lib/rucio/tests/test_judge_repairer.py
+++ b/lib/rucio/tests/test_judge_repairer.py
@@ -23,7 +23,9 @@
 # - Patrick Austin <patrick.austin@stfc.ac.uk>, 2020
 # - Benedikt Ziemons <benedikt.ziemons@cern.ch>, 2020-2021
 # - Simon Fayer <simon.fayer05@imperial.ac.uk>, 2021
+# - Nick Smith <nick.smith@cern.ch>, 2021
 
+import itertools
 import unittest
 from hashlib import sha256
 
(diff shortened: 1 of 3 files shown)


lib/rucio/tests/test_judge_repairer.py
```diff
@@ -305,29 +307,44 @@ def test_to_repair_a_rule_with_only_1_rse_whose_site_is_blocklisted(self):
...
-        add_did(scope, dataset, DIDType.DATASET, self.jdoe)
-        attach_dids(scope, dataset, files, self.jdoe)
 
-        rule_id = add_rule(dids=[{'scope': scope, 'name': dataset}], account=self.jdoe, copies=1, rse_expression=rse, grouping='DATASET', weight=None, lifetime=None, locked=False, subscription_id=None, ignore_availability=True, activity='DebugJudge')[0]
-
-        assert(RuleState.STUCK == get_rule(rule_id)['state'])
-        rule_repairer(once=True)
-
-        # Stil assert STUCK because of ignore_availability:
-        assert(RuleState.STUCK == get_rule(rule_id)['state'])
```
nsmith-, line 324: This behavior is altered, so the old test case would fail. Based on our issue discussion, it sounds like this is OK.

nsmith-:
I think the test failure is intermittent?

bari12 (changes requested):
Sorry for the long delay @nsmith- 
PR is good, only one note
`lib/rucio/tests/test_judge_repairer.py:318:9: E306 expected 1 blank line before a nested definition, found 0`
Also please rebase

nsmith-:
Looks like everything except integration tests passed. Is this expected?

bari12:
Yes, they are unfortunately a bit flaky at the moment.

bari12 (approved):

nsmith-:
Why did this not make it into 1.26.4?

bari12:
It's part of 1.26.2, so should be in 1.26.4 too?

nsmith-:
Sorry you're right, I was going off github hints and didn't realize you cherry-pick all the patches off master

PR #4714, opened 2021-07-01
Transfers: archive timeout override option for submitter

### LINKED ISSUE #4641: Configurable archive-timeout override in submitter
Motivation
----------
After the implementation of the FTS component of https://github.com/rucio/rucio/issues/1797 just recently, CMS enabled this feature. In parallel, our _loadtest_ [daemon](https://github.com/dmwm/CMSRucio/blob/master/docker/CMSRucioClient/loadtest/loadtest.py), which periodically submits transfers to test RSE and link availability, now writes files to tape sites. Site administrators have mapped the loadtest destination directory in a special way to prevent actually writing these test files to tape and wasting tape and tape drive resources. However this poses a challenge now that the archiving feature is implemented, as the site storage will never report these files as being archived, as desired. So we need some mechanism to override the per-RSE `archive_timeout` setting in conveyor-submitter, and run a dedicated submitter with that setting disabled for the load test rules (which use a special activity).

Modification
------------
Add a configuration option `archive_timeout` to conveyor-submitter daemon
https://github.com/rucio/rucio/blob/c0576045445e054c86e5c3ac7005d570fca1649c/lib/rucio/daemons/conveyor/submitter.py#L78-L81
and forward its value to override the RSE-level value at
https://github.com/rucio/rucio/blob/740b825cac747280ca159089bca4635264c07fc1/lib/rucio/daemons/conveyor/common.py#L342

cc @ericvaandering

nsmith-:
Closes #4641

### DIFF
--- a/bin/rucio-conveyor-submitter
+++ b/bin/rucio-conveyor-submitter
@@ -21,6 +21,7 @@
 # - Hannes Hansen, <hannes.jakob.hansen@cern.ch>, 2018-2019
 # - Andrew Lister, <andrew.lister@stfc.ac.uk>, 2019
 # - Patrick Austin <patrick.austin@stfc.ac.uk>, 2020
+# - Nick Smith <nick.smith@cern.ch>, 2021
 #
 # PY3K COMPATIBLE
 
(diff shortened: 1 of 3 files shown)


rcarpa (approved):
I'm not sure if it's better to have a config value; or a command-line argument; other than that, LGTM.

If you feel like adding a test by inspiring from existing ones in test_conveyor_submitter.py, it will be a nice addition.

PR #4716, opened 2021-07-02
Release: Fix setup.py for webui

bari12:

### DIFF
--- a/setup_webui.py
+++ b/setup_webui.py
@@ -15,7 +15,7 @@
 #
 # Authors:
 # - Vincent Garonne <vincent.garonne@cern.ch>, 2015-2017
-# - Martin Barisits <martin.barisits@cern.ch>, 2016-2020
+# - Martin Barisits <martin.barisits@cern.ch>, 2016-2021
 # - Benedikt Ziemons <benedikt.ziemons@cern.ch>, 2021
 
 import sys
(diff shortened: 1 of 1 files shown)


setup_webui.py
```diff
@@ -34,7 +34,7 @@
     from setuputil import get_rucio_version
 
 name = 'rucio-webui'
-packages = ['rucio', 'rucio.web', 'rucio.web.ui', 'rucio.web.ui.common']
+packages = ['rucio', 'rucio.web', 'rucio.web.ui']
```
bziemons: `packages = ['rucio', 'rucio.web', 'rucio.web.ui', 'rucio.web.ui.flask', 'rucio.web.flask.common']` ?

bari12: Yes, you are right. Not sure why we didn't add this in the first place. Will add.

bziemons: probably because this setup.py is long forgotten :D

bziemons (approved):

setup_webui.py
bari12: Actually, `rucio.web.flask.common ` doesn't exist :-)

bziemons: I missed the `ui`, correct. Should've been `rucio.web.ui.flask.common`

PR #4722, opened 2021-07-07
Clients: add sleep-time option to daemons. Closes #3987

### LINKED ISSUE #3987: Some daemons CLI are missing a sleep-time option
Motivation
----------
Some daemons CLI are missing a sleep-time option. It needs to be added

davidpob99:
Clients: add sleep-time option to daemons. Closes #3987

### DIFF
--- a/lib/rucio/common/utils.py
+++ b/lib/rucio/common/utils.py
@@ -37,6 +37,7 @@
 # - Rahul Chauhan <omrahulchauhan@gmail.com>, 2021
 # - Radu Carpa <radu.carpa@cern.ch>, 2021
 # - Anil Panta <47672624+panta-123@users.noreply.github.com>, 2021
+# - David Población Criado <david.poblacion.criado@cern.ch>, 2021
 
 from __future__ import absolute_import, print_function
 
@@ -1372,6 +1373,15 @@ def setup_logger(module_name=None, logger_name=None, logger_level=None, verbose=
         add_handler(logger)
 
     return logger
+
+
+def daemon_sleep(start_time, sleep_time, graceful_stop):
+    """Sleeps a daemon the time provided by sleep_time"""
+    end_time = time.time()
+    time_diff = end_time - start_time
+    if time_diff < sleep_time:
+        logging.info('Sleeping for a while :  %s seconds', (sleep_time - time_diff))
+        graceful_stop.wait(sleep_time - time_diff)
 
 
 class retry:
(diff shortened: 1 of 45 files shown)


lib/rucio/daemons/judge/repairer.py
```diff
@@ -90,7 +91,11 @@ def rule_repairer(once=False):
 
             if not rules and not once:
                 logging.debug('rule_repairer[%s/%s] did not get any work (paused_rules=%s)' % (heartbeat['assign_thread'], heartbeat['nr_threads'], str(len(paused_rules))))
-                graceful_stop.wait(60)
+                end_time = time.time()
```
rcarpa: I think it's worth creating an utils function with this code snippet and call it in all places instead of copy-pasting this code block

davidpob99: Okay, so maybe I can create this function in `rucio.common.utils`:
```python
def sleep_time_daemons(start_time, sleep_time, graceful_stop):
    end_time = time.time()
    time_diff = end_time - start_time
    if time_diff < sleep_time:
        logging.info('Sleeping for a while :  %s seconds', (sleep_time - time_diff))
        graceful_stop.wait(sleep_time - time_diff)
```
And call it in each daemon:
```suggestion
                sleep_time_daemons(start, sleep_time, graceful_stop)
```

rcarpa: That's what I had in mind. Nothing very fancy, just to avoid copy-paste. Maybe a different naming ? maybe daemon_sleep() ? I don't know, but sleep_time_daemons is confusing because it makes me think that it will compute a time; not that it will sleep.

rcarpa (commented):
LGTM

lib/rucio/common/utils.py
```diff
@@ -1342,6 +1343,15 @@ def func(*args):
     return logger
 
 
+def daemon_sleep(start_time, sleep_time, graceful_stop):
```
bari12: It would be good to foresee the handthrough of the logger object there, so we can also do logging with the thread prefixes for the sleep log-message. (Have a look at other methods which pass through the logger object there)
Otherwise it looks good 👍

bari12 (changes requested):

bari12:
I'll wait until #4756 is done, then we can run this again for all the bin's

bari12:
Please rebase :-)

bari12 (approved):

PR #4723, opened 2021-07-07
Database: Update drop_everything method; Fix #4212

### LINKED ISSUE #4212: tools/reset_database.py should remove postgres types
Motivation
----------
Since the sqlalchemy update (#4133) I am getting errors that types already exist on postgres after resetting the database (with or without --drop-everything) in my docker-compose environment.

Modification
------------
Since the `tools/reset_database.py` script is meant to "destroy" the database and leave nothing behind, one option would be to remove all types found in the database.

bziemons:
Update drop_everything method according to the [wiki site](https://github.com/sqlalchemy/sqlalchemy/wiki/DropEverything).
Also drop schema when dropping everything (will be re-created anyway) and enum types for postgres. Fix #4212
Removing the echo parameter on `get_engine` since it wasn't working anyway. Use the echo config variable in the database section.

### DIFF
--- a/lib/rucio/db/sqla/util.py
+++ b/lib/rucio/db/sqla/util.py
@@ -30,15 +30,16 @@ from __future__ import print_function
 from datetime import datetime
 from hashlib import sha256
 from os import urandom
-from traceback import format_exc
 from typing import TYPE_CHECKING
 
+import sqlalchemy
 from alembic import command, op
 from alembic.config import Config
-from sqlalchemy import func
-from sqlalchemy.engine import reflection
+from sqlalchemy import func, inspect
+from sqlalchemy.dialects.postgresql.base import PGInspector
 from sqlalchemy.exc import IntegrityError, DatabaseError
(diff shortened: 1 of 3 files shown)


lib/rucio/db/sqla/util.py
```diff
@@ -79,63 +81,53 @@ def dump_schema():
...
+                    Table(tname, MetaData(), fk_constraint)
+                    drop_constraint_stmt = DropConstraint(fk_constraint)
+                    conn.execute(drop_constraint_stmt)
+
+        schema = config_get('database', 'schema', raise_exception=False)
+        if schema:
+            conn.execute(DropSchema(schema, cascade=True))
+
+        if engine.dialect.name == 'postgresql':
+            assert isinstance(inspector, PGInspector), 'expected a PGInspector'
```
rcarpa, line 128: Didn't know you can do like this in assert :) Nice to know

rcarpa (approved):

PR #4732, opened 2021-07-12
Feature 4731 multi vo certificate selection

thysk:
I have added additional features which as far as I can tell do not interfere with single VO functionality.
These additions detect when multi_vo is set to true in the /opt/rucio/etc/rucio.cfg
collect additional information when querying the DB for transfers (this is also the scope information to extract the VO that submitted the transfer)
Passes additional information to FTS3.py by adding VO information to the external_id and recovering that information to point to the correct VO cert if the information is placed within the rucio.cfg in a new section currently called [vo_certs] to give a centralised list of certs for each of the VOs this allows the submitter to function
ensuring the VO information is also available for the poller, or forbidden messages are returned when incorrect certificates without permissions are selected.

please do provide feedback and suggestions, I am keen to learn from this process and look forward to contributing more in the future


lib/rucio/transfertool/fts3.py
```diff
@@ -92,7 +92,14 @@ def __init__(self, external_host, token=None):
...
         """
-        usercert = config_get('conveyor', 'usercert', False, None)
+
+        self.multi_vo = config_get_bool('common', 'multi_vo', False, None)
+        if self.multi_vo:
+            # a default cert that can be used incase of no selection for each VO that has basic permissions for the public methods
+            usercert = config_get('conveyor', 'usercert', False, None)
+        else:
+            # a cert that is used for the single VO
+            usercert = config_get('conveyor', 'usercert', False, None)
```
rcarpa, line 99: I don't see any difference between if and else

lib/rucio/daemons/conveyor/poller.py
```diff
@@ -140,11 +140,22 @@ def poller(once=False, activities=None, sleep_time=60,
                 for transf in transfs:
                     if not transf['external_host'] in xfers_ids:
                         xfers_ids[transf['external_host']] = []
-                    xfers_ids[transf['external_host']].append((transf['external_id'], transf['request_id']))
+                    xfers_ids[transf['external_host']].append((transf['external_id'], transf['request_id'], transf['scope']))
```
rcarpa: I  would vote to switch using a namedtuple here. This will require some refactoring, but will make the code better.

lib/rucio/transfertool/fts3.py
```diff
@@ -151,52 +158,123 @@ def submit(self, files, job_params, timeout=None):
...
-        elif post_result and post_result.status_code == 409:
-            record_counter('transfertool.fts3.%s.submission.failure' % self.__extract_host(self.external_host), len(files))
-            labels = {'state': 'failure', 'host': self.__extract_host(self.external_host)}
-            SUBMISSION_COUNTER.labels(**labels).inc(len(files))
-            raise DuplicateFileTransferSubmission()
         else:
-            if expected_transfer_id:
-                transfer_id = expected_transfer_id
-                logging.warning("Failed to submit transfer to %s, will use expected transfer id %s, error: %s", self.external_host, transfer_id, post_result.text if post_result is not None else post_result)
+            # single VO mode
```
rcarpa, line 187: This code block is identical to the one in the `if` part. Please move it to a function and call whenever needed.

lib/rucio/transfertool/fts3.py
```diff
@@ -410,44 +487,88 @@ def bulk_query(self, transfer_ids, timeout=None):
...
-                jobs_response = jobs.json()
-                responses = self.__bulk_query_responses(jobs_response)
-            except ReadTimeout as error:
-                raise TransferToolTimeout(error)
-            except JSONDecodeError as error:
-                raise TransferToolWrongAnswer(error)
-            except Exception as error:
-                raise Exception("Failed to parse the job response: %s, error: %s" % (str(jobs), str(error)))
+        # multi-VO mode
+        if self.multi_vo:
```
rcarpa: same comment as previous: don't copy paste code; move the common code to a function and call it where needed

rcarpa (changes requested):

lib/rucio/transfertool/fts3.py
thysk: That is fair, I will remove the if/else

lib/rucio/transfertool/fts3.py
thysk: Thats what I thought when i copied and pasted, but there didnt seem to be similar methods, but will make this change

lib/rucio/transfertool/fts3.py
thysk: same as above

thysk:
Made changes to FTS3.py as suggested by @rcarpa, working on looking into namedtuple and how I would incorperate

lib/rucio/transfertool/fts3.py
thysk: I have removed the If/Else

lib/rucio/transfertool/fts3.py
thysk: Method abstracted

lib/rucio/transfertool/fts3.py
thysk: method abstracted

thysk (commented):
Incorperated the namedtuple suggestion, as well as fixed issues with poller not checking all transfers due to VO differences

lib/rucio/daemons/conveyor/poller.py
```diff
@@ -137,14 +137,26 @@ def poller(once=False, activities=None, sleep_time=60,
...
                     if not transf['external_host'] in xfers_ids:
                         xfers_ids[transf['external_host']] = []
-                    xfers_ids[transf['external_host']].append((transf['external_id'], transf['request_id']))
-
+                    xfers_ids[transf['external_host']].append(TransferParams(transf['external_id'], transf['request_id'], transf['scope']))
+                external_ids = []
                 for external_host in xfers_ids:
-                    external_ids = list({trf[0] for trf in xfers_ids[external_host]})
-                    request_ids = [trf[1] for trf in xfers_ids[external_host]]
+                    if TRANSFER_TOOL == 'fts3':
```
rcarpa: Why do we pass via a set in "else" but not in if? I would assume it's to remove duplicates. Aren't duplicates an issue for multivo case ? 

Also, something like this avoids retrieving the configuration value multiple times; avoids code repetition for the two else blocks, and makes it much easier to understand the difference between the if and else
```
                multi_vo_enabled = config_get_bool('common', 'multi_vo', False, None)
                for external_host in xfers_ids:
                    if TRANSFER_TOOL == 'fts3' and multi_vo_enabled:
                        external_ids = list({trf.external_id + '@' + trf.scope.vo 
                                             for trf in xfers_ids[external_host]})
                    else:
                        external_ids = list({trf.external_id 
                                             for trf in xfers_ids[external_host]})
````

File: lib/rucio/transfertool/fts3.py [hunk 6]
```diff
@@ -150,52 +151,45 @@ def submit(self, files, job_params, timeout=None):
...
+                params_str = json.dumps(params_dict, cls=APIEncoder)
+
+                post_result = None
+
+                vo_cert = config_get('vo_certs', vo, False, config_get('conveyor', 'usercert', False, None))
+
+                self.cert = (vo_cert, vo_cert)
+
+                transfer_id = self.__attempt_transfer(params_str, timeout, files, expected_transfer_id, post_result)
+                return transfer_id
```
rcarpa, line 184: This will stop submission after handling the first VO.

Also, now with the request code out of the way moved to a separate function, it feels like the vo handling code must be done upstream. Somewhere in bulk_group_transfers or in bulk_submit transfers. The 
```
self.cert = (vo_cert, vo_cert) 
```
part look like looking for troubles. Somebody re-using the same transfertool instance may not expect for certificates to have some random value.

rcarpa (changes requested):

lib/rucio/daemons/conveyor/poller.py
thysk: Those are some good points, I will look into making it a set instead
and the code exampe looks like it would work really well, thanks, I didnt think of that

lib/rucio/transfertool/fts3.py
thysk: So it would stop the submission, good catch! I will fix that now

As for the
`self.cert = (vo_cert, vo_cert) `
I was confused about that as well, but that seemed to be how it was handeled  for non-multiVO instances

```
if self.external_host.startswith('https://'):
            if self.token:
                self.cert = None
                self.verify = False
                self.headers['Authorization'] = 'Bearer ' + self.token
            else:
                self.cert = (usercert, usercert)
                self.verify = False
```

So was attempting to not cause issues with multiVO, or do you mean that is now called vo_cert? I suppose it can be renamed usercert like above

rcarpa: I mean that the object state is changed by a function which doesn't look (without reading it code) that it will change the object state. This may be handled correctly in the current state of code, but, in the future, somebody can decide to add an fts.query() after an fts.submit(). That code will pass the review because nobody would even think to look what submit() does inside. Will be deployed. And will fail on release (but only for people using multi-vo). That's why I say that it "feels" like looking for trouble, not that it is wrong right now.

thysk: Also I have done some more testing with my modified test containers (I altered the multi_vo_tests_docker.sh to initialise both VOs and and add files, containers, and datasets for each, and my testing does the submission for both VOs

I also have the look up there as its where the certificate is looked up currently is. so made the most sense. But I am happy to move it else where. but then would need to pass the cert down to the FTS

lib/rucio/daemons/conveyor/poller.py
```diff
@@ -137,14 +137,23 @@ def poller(once=False, activities=None, sleep_time=60,
...
-                    xfers_ids[transf['external_host']].append((transf['external_id'], transf['request_id']))
-
+                    xfers_ids[transf['external_host']].append(TransferParams(transf['external_id'], transf['request_id'], transf['scope']))
+                external_ids = []
+                multi_vo_enabled = config_get_bool('common', 'multi_vo', False, None)
                 for external_host in xfers_ids:
-                    external_ids = list({trf[0] for trf in xfers_ids[external_host]})
-                    request_ids = [trf[1] for trf in xfers_ids[external_host]]
+                    if TRANSFER_TOOL == 'fts3' and multi_vo_enabled:
+                        if config_get_bool('common', 'multi_vo', False, None):
```
rcarpa: this if shouldn't be here

thysk: You are absolutely right, my bad

lib/rucio/daemons/conveyor/poller.py
thysk: I have implemented the improvemnts you suggested

lib/rucio/daemons/conveyor/poller.py
thysk: Errant If statement removed

lib/rucio/transfertool/fts3.py
rcarpa: It may have worked because you did run the submitter with the default value for group_bulk = 1. If you increase this parameter, I don't think it will work. Because of that return

rcarpa: Also, would be nice to add an automatic test. You can start from one of the existing ones in test_conveyor_submitter.py . You can also check test_reaper.py for some examples of multi-vo testing

thysk: I have been thinking about your comment about VO handelling should be done in bulk_group_transfers
Which I think is a really good idea and stream lines a lot of the code if I can (as the current bulk process sorts by external_host) sort by VO prior to the sort by external_host, this will allow for a small change in the parsing of the grouped_jobs later in the submit, and submit all the transfers per external_host with the correct usercert

This of course is a different approach from what I have currently and I am coding and will be testing that version seperately but may be a better solution.

I also agree that I should look into writing some automatic testing for this portion of the code for multi-vo

Thanks very much for the comments and feedback

lib/rucio/daemons/conveyor/common.py
```diff
@@ -305,23 +309,37 @@ def bulk_group_transfers_for_fts(transfers, policy='rule', group_bulk=200, sourc
...
+        if multi_vo:
+            vo = transfer['account'].vo
+
         external_host = transfer['external_host']
         activity = t_file['activity']
 
-        if external_host not in grouped_transfers:
-            grouped_transfers[external_host] = {}
-            grouped_jobs[external_host] = []
+        if multi_vo:
```
rcarpa, line 331: I have a feeling that none of the changes in bulk_group_transfers bellow this line are needed. That adding 
```
            if vo:
                job_key = job_key + ',vo=%s' % vo 
````
after https://github.com/rucio/rucio/blob/e6940d402b168b601b92c7b3996bc7446d6d4cae/lib/rucio/daemons/conveyor/common.py#L390
is enough. Did you try it this way ?

**Review by rcarpa — CHANGES_REQUESTED** [review]

File: lib/rucio/transfertool/fts3.py [hunk 9]
```diff
@@ -356,44 +352,27 @@ def bulk_query(self, transfer_ids, timeout=None):
...
         """
 
-        jobs = None
-
         if not isinstance(transfer_ids, list):
             transfer_ids = [transfer_ids]
 
-        responses = {}
-        fts_session = requests.Session()
+        """# multi-VO mode
```
rcarpa: Don't leave commented code. If it's not needed, remove it

lib/rucio/daemons/conveyor/common.py
rcarpa: never mind, I'm wrong

lib/rucio/transfertool/fts3.py
thysk: Removed

lib/rucio/core/transfer.py
```diff
@@ -849,60 +850,61 @@ def update_transfer_state(external_host, transfer_id, state, session=None, logge
...
-                                               response['new_state'],
-                                               transfer_id=transfer_id,
-                                               src_rse_id=src_rse_id,
-                                               err_msg=err_msg,
-                                               session=session)
-
-                request_core.add_monitor_message(req, response, session=session)
-        else:
-            __set_transfer_state(external_host, transfer_id, state, session=session)
+    if state == RequestState.LOST:
```
rcarpa: Why this change is needed ? 
Are you still working on the PR? is it a draft?

thysk: I am not sure what happened here
This clearly doesnt need to change

thysk: I am still working on this PR yes, but not all my time  at the moment
And now I see my PR is failing on the Multi_VO tests but for GLOBUS, which I need to investigate

rcarpa: Then it's probably a good idea to convert it to a draft.

bari12:
Hi @Thysk is this still a draft?

thysk:
Hi @bari12, Yes, this is still in draft, With the major changes to conveyor core, looks like I will have to go back to the drawing board for this one.
When I get back looking at this after making our Rucio instance stable again I was going to ask @rcarpa for some advice on how to go about this if they have time

rcarpa:
@Thysk , I can easily find time for it. I can even take care of the rebase onto the new changes if the core functionality was already ready and it only requires adaptation to my changes. Give me a ping any time.

bari12:
Thanks @Thysk 
If it helps @rcarpa, you, and myself can have a call to sync up and align these changes a bit. 
Unfortunately the changes to the transfer machinery were a bit conflicting to your changes, but they were much needed.

bari12:
I would propose to close this in the meanwhile, if it is ok by you.

thysk:
Sure, I don't mind

PR #4735, opened 2021-07-14
Core: correct redefining of built-in vars. Closes #4466

### LINKED ISSUE #4466: Correct and Avoid redefining of built-in function
Motivation
----------
There are multiple redefinitions of built-in functions, such as:
1. type in client
2. filter in core

This is both confusing and a bad practice which renders the built-in functions useless.

Modification
------------
Come up with better alternatives. Maybe, request_type instead of type, filter_ (trailing underscore) when an alternative does not suit the requirements and avoid using built-in names or something else.

davidpob99:
Core: correct redefining of built-in vars. Closes #4466

If possible, I refactored the name of that vars to make them clearer (`type` -> `did_type`)

For the others I followed the [Style Guide for Python Code - Descriptive: Naming Styles](https://www.python.org/dev/peps/pep-0008/#descriptive-naming-styles), adding an underscore at the end of the name variables that redefined a built-in name (`filter` -> `filter_`).


bari12 (dismissed):
Will merge this as a feature for 1.27

bari12:
Please rebase

bari12 (approved):

PR #4736, opened 2021-07-15
Rebalancing : BB8 rewrite : Closes #4706

### LINKED ISSUE #4706: BB8 rewrite
Motivation
----------
BB8 has a lot of hardcoded values and is very ATLAS biased. Rewrite of some part of the code is needed

cserf:
Rebalancing : BB8 rewrite : Closes #4706

### DIFF
--- a/lib/rucio/daemons/bb8/common.py
+++ b/lib/rucio/daemons/bb8/common.py
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 # Copyright 2016-2021 CERN
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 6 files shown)


cserf:
WIP, but first comments are welcome

cserf:
The goal of this PR is manifold :
- It introduces the new BB8 daemon that replaces the scripts for nuclei and t2 background rebalancing
- It enforce that 2 BB8 instances on 2 set of RSEs that partially overlap cannot run at the same time
- The new daemon is VO-agnostic
- Better handling of already active rules
- Include tests to validate the new daemon behaviour

lib/rucio/daemons/bb8/common.py
```diff
@@ -229,7 +225,6 @@ def _list_rebalance_rule_candidates_dump(rse_id, mode=None):
 def list_rebalance_rule_candidates(rse_id, mode=None, session=None):
```
bari12, line 226: On this method, I wonder if we need to foresee more flexibility?

bari12 (commented):
Hi @cserf 
I quite like it, I think it goes into the right direction. I am wondering though if we can possibly foresee to run this without the necessity of the dumps? I think it would be worth a try.

bari12:
From my point of view this is ready to merge. @cserf ?
It's a bit difficult to judge without the input of other communities if it fits their use-case; I suppose this kind of input we only get when they run it. At least for ATLAS and Belle II this should be ok!?

bari12 (approved):

PR #4740, opened 2021-07-19
Traces: add validator for traces. Closes #1834

### LINKED ISSUE #1834: add validator for traces
Motivation
----------

The tracer server accepts any kind of json send to it and expects clients to send the information using the correct schema. Sometimes a bug in a client can lead to an invalid trace that then can break any of the systems using the traces (monitoring, Kronos, etc.)

Modification
------------

Add a validator that listens on the ActiveMQ topic and constantly checks that necessary fields are set correctly or set at all. If something doesn't match the schema send out notifications.

davidpob99:
Traces: add validator for traces. Closes #1834

Add a validator for the different schema types depending the `eventType` key of the JSON string

The schemes were inferred from the code of the different clients and the kronos daemon. If the schema is not validated, a warning is printed in the log.


mlassnig (approved):
it looks okay to me, but we really need to check if this works with our real-world traces, otherwise we might throw away events

PR #4743, opened 2021-07-22
Clients: add delete distance API. Closes #4742

### LINKED ISSUE #4742: Delete distance API
Although there is a core method
https://github.com/rucio/rucio/blob/a0823b3ec19ff1f98639361a14b628ca6c908651/lib/rucio/core/distance.py#L115
we have no client or REST API access to this method.

davidpob99:
Clients: add delete distance API. Closes #4742

### DIFF
--- a/bin/rucio-admin
+++ b/bin/rucio-admin
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
-# Copyright 2012-2020 CERN
+# Copyright 2012-2021 CERN
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


lib/rucio/api/rse.py
```diff
@@ -24,6 +24,7 @@
 # - Gabriele <sucre.91@hotmail.it>, 2019
 # - elichad <eli.chadwick.256@gmail.com>, 2020
 # - patrick-austin <patrick.austin@stfc.ac.uk>, 2020
+# - David Población Criado <david.poblacion.criado@cern.ch>, 2021
```
bari12, line 31: Please use `tools/add_header -i %s` script to add the header. Since you are using a special character in your name, we should make sure that the utf-8 encoding is mentioned at the top of the file (Which the add_header script will do)

bari12 (changes requested):

bin/rucio-admin
```diff
@@ -1,6 +1,5 @@
 #!/usr/bin/env python
-# -*- coding: utf-8 -*-
-# Copyright 2012-2020 CERN
+# Copyright 2012-2021 CERN
```
bziemons, line 3: why was the `# -*- coding: utf-8 -*-` part removed?

davidpob99: I ran the `tools/add_header` script with the `-i` flag (`tools/add_header -i bin/rucio-admin`) and it deleted that line

davidpob99: Related with #4756

bziemons: Thanks, that's the missing bit :)

bari12 (approved):

PR #4746, opened 2021-07-27
Core & Internals: Filtering engine for advanced metadata search; Fix #3750

### LINKED ISSUE #3750: Support DID metadata queries with inequality operators:

robbarnsley:
Refactoring of https://github.com/rucio/rucio/pull/4139

Note that filtering engine == inequality engine.

Some notable changes from 4139 and this PR:

- <filters> variable kept as single type as it progresses through client/rest/api workflow resulting in fewer changes required to these interfaces
- parsing of filter string moved to common/utils, as per previous workflow
- parsed <filters> formatted as "keyword.operator", complementing current syntax e.g. "length.gte"
- keyword sanity checking moved entirely to inside filter engine
- sqla queries generated by filter engine now compounded to single query
- != operator added with wildcard support (NOT LIKE)
- added logic to filter duplication of dids resulting from non-mutual result sets 
- fixed issue with "length" keyword backwards compatibility


robbarnsley:
Just to note, this PR only improves searching for `rucio list-dids`, not extended, and as such is only for the hardcoded DID columns.

If I can get this reviewed, I would be looking next to expand the functionality to include JSON metadata.

bin/rucio
```diff
@@ -2040,7 +2052,7 @@ You can filter by key/value, e.g.::
 
     list_parser.set_defaults(function=list_dids)
     list_parser.add_argument('--recursive', dest='recursive', action='store_true', default=False, help='List data identifiers recursively.')
-    list_parser.add_argument('--filter', dest='filter', action='store', help='Filter arguments in form `key=value,another_key>=next_value`. Valid keys are type and length. Use `type` to specify a DID type to search for. Valid values are `ALL`, `COLLECTION`, `CONTAINER`, `DATASET`, `FILE`. Use `length` to filter for an amount of corresponding Data IDentifiers. Valid operators are `=`, `>`, `>=`, `<`, `<=`.')  # noqa: E501
+    list_parser.add_argument('--filter', dest='filter', action='store', help='Filter arguments in form supported by the inequality-engine.')
```
mlassnig: Maybe instead just write what the form is?

lib/rucio/common/exception.py
```diff
@@ -1023,5 +1023,25 @@ class UnsupportedAccountName(RucioException):
...
         self.error_code = 97
+
+
+class DuplicateCriteriaInDIDFilter(RucioException):
+    """
+    Duplicate criteria found in DID filter.
+    """
+    def __init__(self, *args, **kwargs):
+        super(DuplicateCriteriaInDIDFilter, self).__init__(*args, **kwargs)
+        self._message = 'Duplicate criteria for key/operator in filter expression'
```
mlassnig: Maybe add the actual duplicate that was found in the expression?

lib/rucio/common/utils.py
```diff
@@ -1081,6 +1083,114 @@ def parse_did_filter_from_string(input_string):
...
+
+    # lookup table mapping operator opposites, used to reverse compound inequalities.
+    operator_opposites_LUT = {
+        'lt': 'gt',
+        'lte': 'gte'
+    }
+    operator_opposites_LUT.update({op2: op1 for op1, op2 in operator_opposites_LUT.items()})
+
+    filters = []
+    if input_string:
```
mlassnig, line 1148: I just have to accept that this does what it should :-D

lib/rucio/core/did.py
```diff
@@ -1418,7 +1422,7 @@ def set_status(scope, name, session=None, **kwargs):
 
 @stream_session
 def list_dids(scope, filters, did_type='collection', ignore_case=False, limit=None,
-              offset=None, long=False, recursive=False, session=None):
+              offset=None, long=False, recursive=False, session=None, ignore_dids=None):
```
mlassnig: we try to keep session as the last parameter for consistency

lib/rucio/core/did_meta_plugins/filter_engine.py
```diff
@@ -0,0 +1,319 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Authors:
+# - Gabriele Gaetano Fronzé <gfronze@cern.ch>, 2020
+# - Rob Barnsley <rob.barnsley@skao.int>, 2021
+#
+# PY3K COMPATIBLE
+
+from datetime import datetime
```
mlassnig: correctly sort the imports please

mlassnig (commented):
this is amazing.

just two tiny comments.

lib/rucio/core/did_meta_plugins/__init__.py
```diff
@@ -181,11 +181,8 @@ def list_dids(scope=None, filters=None, did_type='collection', ignore_case=False
     :returns: List of dids.
     """
     meta_handler_to_use = None
-
     # Ensure that a single handler manages all the keys of the query
     for key in filters:
-        if key == 'name':
```
bari12, line 187: Why is this removed?

bari12 (commented):
Hi @robbarnsley 
This is very nice work. Just one small comment, otherwise this can be added.

We have to think about the next steps though, to me this only gets it "half-way" there. To me the missing steps are:
- Filter engine should not be specific to the column way. Ideally it should prepare a filter which the individual plugin then interpret (or not, if they can't). - So the filter_engine itself should perhaps be sqlalchemy independent.
- Implement it for the column (done) and json plugin
- list_dids should actually be merged (and changed to) list_dids_extended, so that there is only one way of taking care of this. Maybe this can also be part of your next PR.

bin/rucio
robbarnsley: done.

lib/rucio/common/exception.py
robbarnsley: done.

lib/rucio/core/did.py
robbarnsley: changed.

lib/rucio/core/did_meta_plugins/filter_engine.py
robbarnsley: done.

lib/rucio/common/utils.py
robbarnsley: I appreciate your trust ...

bari12:
Still some conflicts and failing tests, please rebase. But in general I think this is GTG!

robbarnsley:
great, thanks Martin - it's on my list for this week.

robbarnsley:
Sorry took a bit of a while to get around to this, should be OK now.

bari12 (approved):

PR #4750, opened 2021-07-29
Clients: show did length estimation in downloadclient. Closes #3824

### LINKED ISSUE #3824: Incomplete download looks "complete" when restricting RSEs

rcarpa:
Some refactoring was needed to make it possible. see individual commits for details. 
I would keep the commits separated on merge instead of squashing them.

### DIFF
--- a/bin/rucio
+++ b/bin/rucio
@@ -1110,15 +1110,12 @@ def download(args):
 
     summary = {}
     for item in result:
-        scope = item.get('dataset_scope', item['scope'])
-        name = item.get('dataset_name', item['name'])
-        summary_key = '%s:%s' % (scope, name)
-
-        summary.setdefault(summary_key, {'DONE': 0, 'ALREADY_DONE': 0, '_total': 0})
-        summary[summary_key]['_total'] += 1
-        state = item['clientState'].upper()
-        if state in summary[summary_key]:
-            summary[summary_key][state] += 1
+        for did, did_stats in item.get('input_dids').items():
+            did_summary = summary.setdefault(did, {'length': did_stats.get('length'), 'DONE': 0, 'ALREADY_DONE': 0, '_total': 0})
+            did_summary['_total'] += 1
+            state = item['clientState'].upper()
+            if state in did_summary:
+                did_summary[state] += 1
 
     print('----------------------------------')
     print('Download summary')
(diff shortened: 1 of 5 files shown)


lib/rucio/tests/test_download.py
```diff
@@ -409,26 +512,45 @@ def test_trace_copy_out_and_checksum_validation(vo, rse_factory, did_factory, do
...
+
+    with TemporaryDirectory() as tmp_dir:
         nrandom = 2
-        result = download_client.download_dids([{'did': dataset_did_str, 'nrandom': nrandom, 'base_dir': tmp_dir}])
+        result = download_client.download_dids([{'did': dataset1_did_str, 'nrandom': nrandom, 'base_dir': tmp_dir}])
         assert len(result) == nrandom
 
 
-def test_download_blocklisted_replicas(rse_factory, did_factory, download_client):
+def test_download_blacklisted_replicas(rse_factory, did_factory, download_client):
```
mlassnig: blocklist was quite alright :-)

bin/rucio
```diff
@@ -1090,30 +1090,33 @@ def download(args):
...
+        print('DID %s' % summary_key)
+        length = did_summary['length']
+        ds_total = did_summary['_total']
+        downloaded_files = did_summary['DONE']
+        local_files = did_summary['ALREADY_DONE']
         not_downloaded_files = ds_total - downloaded_files - local_files
-        print('{0:40} {1:6d}'.format('Total files: ', ds_total))
+
+        if length:
+            print('{0:40} {1:6d}'.format('Total files (did): ', length))
```
mlassnig: I would uppercase `DID`

bin/rucio
```diff
@@ -1090,30 +1090,33 @@ def download(args):
...
+            print('{0:40} {1:6d}'.format('Total files (after filter):   ', ds_total))
```
mlassnig: shorter: `filtered`

rcarpa:
I cannot understand why that multi-vo tests fails after adding this function. But it's consistent :/

bari12:
@Thysk any idea what is going on with the Multi-VO tests here? It is a very strange behaviour @rcarpa is noticing. Basically the code he changed does not touch the failing test at all. But it is working on master...

thysk:
@bari12 It looks like its succeeding in the 1st VO, but then is failing in the second VO
But given its the same process each time not really sure why
If this something that is failing in other PRs as well?
As there are multiple tests after this that function perfectly well, and perform almost identical tasks
Is there a way to see the full logs of the tests? because after the assert that fails there is another assert and I would be curious to see if that had the correct number of files (though it looks from the bytes that no files have been added to the dataset)

Given that all the exact same processes are used in the next test (test_account_counter_judge_evaluate_detach)
In am not sure how it is only failing the one test. And only on this PR, as, as you said, the PR doesnt touch any of the code that runs in the tests (as far as I can tell)

bari12:
@rcarpa Please rebase and run again.
I don't think this is failing in other tests...

rcarpa:
I rebased and the problem persists :confused:

mlassnig (approved):

bari12:
Thanks @rcarpa 
I'll bring this in as a feature for 1.27 though, since I want to stabilise the 1.26 clients a bit at the moment and not introduce yet another change.

PR #4756, opened 2021-08-04
Fix add_header script for the encoding of bin files. Closes #4755

### LINKED ISSUE #4755: Fix add header script for bin files
Motivation
----------
The _tools/add_header_ script only add the encoding line (`# -*- coding: utf-8 -*-\n`) when the file is a `.py` but not for those in the `bin` folder which are Python files also.

Modification
------------
Modify the script in order to add the encoding line to the files stored in `bin` folder.

davidpob99:
Fix add_header script for the encoding of bin files. Closes #4755

### DIFF
--- a/tools/add_header
+++ b/tools/add_header
@@ -17,8 +17,10 @@
 # Authors:
 # - Vincent Garonne <vincent.garonne@cern.ch>, 2018
 # - Benedikt Ziemons <benedikt.ziemons@cern.ch>, 2020-2021
-# - Martin Barisits <martin.barisits@cern.ch>, 2020
+# - Martin Barisits <martin.barisits@cern.ch>, 2020-2021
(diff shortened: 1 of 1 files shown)


tools/add_header
```diff
@@ -172,7 +175,7 @@ def main(arguments):
 
             sys.exit(0)
 
-        if str(MyFile).endswith('.py'):
+        if str(MyFile).endswith('.py') or MyFile.split('/')[0] == "bin":
```
bziemons: I think this needs a more sophisticated check, making the path absolute and then checking the parent directory of `MyFile`. I sometimes run this script on absolute paths or from within the bin directory and different results would be weird in those cases.

bziemons (approved):

Issue #3416, opened 2020-03-27
support prometheus metrics in daemons and servers


bari12:
Motivation
----------
Currently we collect counters and timings in graphite, but more and more people rely on prometheus for their monitoring infrastructure. 

Allow to expose metrics in the prometheus scrape format.

PR #4764, opened 2021-08-09
Daemons: record counters to both prometheus and graphite #3416

rcarpa:
Introduce a MultiCounter class which takes care of registering
both statsd and prometheus Counters. Use it in all places where
both prometheus and statsd metrics where collected.

Update the record_counter function to also record both counters
as drop-in replacement. I didn't find any place where it would be
called with multiple counters, so remove this capability.

### DIFF
--- a/lib/rucio/core/monitor.py
+++ b/lib/rucio/core/monitor.py
@@ -1,18 +1,25 @@
-# Copyright European Organization for Nuclear Research (CERN)
+# -*- coding: utf-8 -*-
+# Copyright 2013-2021 CERN
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
-# You may not use this file except in compliance with the License.
+# you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
-# http://www.apache.org/licenses/LICENSE-2.0
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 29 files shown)


bari12 (dismissed):
LGTM, second pair of 👁️ maybe good though. @mlassnig

lib/rucio/core/monitor.py
```diff
@@ -35,27 +44,88 @@
...
-            if delta > 0:
-                CLIENT.incr(counter, delta)
-            else:
-                CLIENT.decr(counter, delta)
+
+    counter = COUNTERS.get(name)
+    if not counter:
+        COUNTERS[name] = counter = MultiCounter(statsd=name, labelnames=labels.keys() if labels else ())
+
+    if delta < 0:
```
mlassnig: do you mean here:

`delta = abs(delta)` ?

mlassnig (commented):
LGTM as well, we need to keep an eye on the plots though, it might be there's a typo in there that we didn't spot and our plots may start acting weird

lib/rucio/core/monitor.py
rcarpa: Absolutely correct. The code which is being substituted contained an "if delta > 0". Because of this, my dumb brain didn't realize it can be written differently :D

mlassnig: :-D

bari12 (approved):

PR #4765, opened 2021-08-09
Tests: wait for FTS transfer to actually finish

rcarpa:
Otherwise, we can poll the status while it is still active,
and the test fails

### DIFF
--- a/lib/rucio/tests/test_tpc.py
+++ b/lib/rucio/tests/test_tpc.py
@@ -31,6 +31,8 @@ from rucio.client.rseclient import RSEClient
 from rucio.client.ruleclient import RuleClient
 from rucio.common.utils import run_cmd_process
 
+MAX_POLL_WAIT_SECONDS = 60
+
 
 @pytest.fixture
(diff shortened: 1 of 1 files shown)


lib/rucio/tests/test_tpc.py
```diff
@@ -132,7 +134,13 @@ def test_tpc(containerized_rses, root_account, test_scope, did_factory, rse_clie
     assert fts_transfer_id is not None
     assert fts_transfer_status in ['SUBMITTED', 'ACTIVE']
```
maany: @rcarpa minor change, but this `assert` can be removed since you are checking this condition in the polling loop, and asserting this as a pre-condition would not really make a difference since `fts_transfer_status = None` before the loop begins ;)

bari12 (dismissed):

lib/rucio/tests/test_tpc.py
rcarpa: done

bari12 (approved):

Issue #4772, opened 2021-08-11
breaking tests following change of nomenclature


robbarnsley:
Motivation
----------

https://github.com/rucio/rucio/blob/a5e5e7d54b1d48ced74ec59673f9a9e27e9316b0/lib/rucio/client/rseclient.py#L679 needs changing in line with changes made in #4735  

Modification
------------

"type" -> "type_"

robbarnsley:
also "bytes" -> "bytes_": https://github.com/rucio/rucio/blob/a5e5e7d54b1d48ced74ec59673f9a9e27e9316b0/lib/rucio/tests/test_judge_repairer.py#L326

rcarpa:
Should be fixed by 

https://github.com/rucio/rucio/pull/4775

robbarnsley:
confirmed

PR #4774, opened 2021-08-12
Tests: don't import temp factories in test_bin_rucio.py

rcarpa:
temp factories cannot be used in python2.7 env because of
calling directly some core functions which are python3 only.


davidpob99:
Closes #4772

bziemons (approved):
lgtm

Issue #4782, opened 2021-08-16
Export all ports to localhost with specific bind ip


rak108:
Motivation
----------
As discussed during our call, this is an added security filter as suggested by @rcarpa (https://github.com/rak108/rucio/pull/1#discussion_r658746726)

Modification
------------
Prefixing all exports with 127.0.0.1: in the dockerfiles

PR #4783, opened 2021-08-16
Docker: Binding to specific IPs #4782

rak108:
This PR is in relation to issue #4782


rcarpa:
Love it. Hope it works with podman. @bziemons , any opinion ?

bziemons:
The title of the PR doesn't really explain what is actually happening, but yes: I prefer binding to 127.0.0.1 only and it works with podman. I wanted to make this change before, but it was decided not to do it, because some wanted to reach the services (rest api, web-ui, database, ...) from outside the own host... not sure if that is still a requirement though

rcarpa:
Weeeell, i'd say that those who want to do nasty things can maintain their own local config :)

mlassnig:
Well, at least we need it for the SSH container. For the rest, we should apply some good old democracy :-)

rak108:
@mlassnig ah alright, so should I close this PR?

mlassnig:
I'm voting for 127.0.0.1 on all of them :-)

bziemons:
I think it is decided then. There is no opposing vote ;)

mlassnig (approved):

bari12:
🥳

PR #4786, opened 2021-08-18
Core: config_get search in cfg and in config table. Closes #2630

### LINKED ISSUE #2630: rucio.cfg vs config table

davidpob99:
**Core: config_get search in cfg and in config table. Closes #2630**

The ticket description is outdated. The changes that this PR covers are:
- There is only one `config_get` (in `rucio.common`) method to retrieve all the configuration (both from config file and from config table).
- The method first checks in the configuration file (cfg) and, if the `(section, option)` pair is not found, looks in the configuration table only for server/daemon. Even if it is called from server/daemon, it is possible to avoid looking at the config table with the `check_config_table` argument in order to avoid circular imports and other issues such as trying to get config from the DB if the DB is not created yet (the section _database_ cannot be retrieved from config table).
- Created new method (`is_client`) in `rucio.common.utils` to check if it is called from a client or from a server/daemon. Code from: https://github.com/rucio/rucio/blob/900b555970e25d7e6b7137d4c725ebc9f0027a08/lib/rucio/rse/__init__.py#L33-L49
- Refactor all the calls to `rucio.core.get` to use the new method.

### DIFF
--- a/lib/rucio/common/config.py
+++ b/lib/rucio/common/config.py
@@ -33,7 +33,7 @@ import os
 import json
 import sys
 
-from rucio.common.exception import ConfigNotFound
+from rucio.common.exception import ConfigNotFound, DatabaseException
 
 try:
     import ConfigParser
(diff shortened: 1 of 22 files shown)


bari12 (dismissed):
I think this is good. @mlassnig 👀 please as well

mlassnig:
confirmed, looks good.

there's a conflict though.

mlassnig (approved):

bari12 (approved):

PR #4801, opened 2021-08-20
Clients: Fix enum python 2.7 issue; Fix #4800

### LINKED ISSUE #4800: enum dependency issue for python2
Motivation
----------
Since we do require the python3 `enum` package in https://github.com/rucio/rucio/blob/master/lib/rucio/common/utils.py which is used by the python 2.7 clients we need to add the backport of it to the py27 clients package.

Modification
------------
Add https://pypi.org/project/enum34/ to python27 dependencies for clients.

bari12:

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,6 +7,7 @@ tabulate>=0.8.0,<0.9.0                                      # Pretty-print tabul
 six>=1.12.0,<1.16.0                                         # Python 2 and 3 compatibility utilities
 jsonschema>=3.2.0                                           # For JSON schema validation (Policy modules)
 typing; python_version < '3.5'                              # typing package support for older Python versions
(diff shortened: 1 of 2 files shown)


bziemons (approved):

requirements.txt
```diff
@@ -7,6 +7,7 @@ tabulate>=0.8.0,<0.9.0                                      # Pretty-print tabul
 six>=1.12.0,<1.16.0                                         # Python 2 and 3 compatibility utilities
 jsonschema>=3.2.0                                           # For JSON schema validation (Policy modules)
 typing; python_version < '3.5'                              # typing package support for older Python versions
+enum34<=1.1.10; python_version < '3.5'                      # Enum package backport for python 2.7
```
bziemons, line 10: not a big deal and won't be a problem, but 100% correct would be `python_version < '3.4'`

Issue #3982, opened 2020-09-04
Extend get_dataset_locks API to support requests in bulk


amaltaro:
Motivation
----------
This is a similar feature request as the one made in: https://github.com/rucio/rucio/issues/2459

It would be extremely helpful to be able to provide a list of DIDs (dataset type) to the `get_dataset_locks` API, such that we could retrieve all the locks for a container with a single API call.

This is a very common use case for CMS, where provided a container name, we need to find out all the RSEs holding part of that container.

Modification
------------
Support a list of DIDs to be passed to the `get_dataset_locks`, such that the number of API calls can be drastically decreased for larger containers.

bari12:
That should be easy to do, similar to the #2459. @ericvaandering any chance anyone from the CMS team can take this issue?

nsmith-:
Should this API also be able to filter the locks by account?

ivmfnal:
How do you expect the resulting list to be structured ? Just plain list of all locks found, or by did ?

bari12:
Just a question, if it is always all datasetlocks for a given container, wouldn't it be better to have a feature where you can query by container? (Instead of resolving to all the datasets and then querying them in bulk?) - This can be in addition to a bulk query.

For the server it would probably be better to stream these through, so you don't have to build any dictionaries in memory. The did organization could then be done client side.

amaltaro:
> How do you expect the resulting list to be structured ? Just plain list of all locks found, or by did ?

Igor, having them organized by DID would be the best.

>Just a question, if it is always all datasetlocks for a given container, wouldn't it be better to have a feature where you can query by container? (Instead of resolving to all the datasets and then querying them in bulk?) - This can be in addition to a bulk query.

Honestly, it would be great to have a dataset_locks lookup for a given container. However, having the ability to provide a list of datasets is also needed. Shall we create another GH issue for the container based API?

bari12:
Yes, a second (related) issue would be good. 
This would also be more in line how, for example, `list_replicas` works, there you can also use a file/dataset/container did interchangeable and it will resolve to all replicas.

ivmfnal:
If we do both:

1. list locks for a list of datasets
2. list locks for datasets found in the container

Can in fact (2) be implemented using (1) at the web client level, hidden from the user as following ?

1. get list of datasets for the container
2. get list of locks for the list of datasets

amaltaro:
[...]

Issue #4807, opened 2021-08-24
Support API to return dataset locks for a given container name


amaltaro:
Motivation
----------
As an alternative to the feature requested in: https://github.com/rucio/rucio/issues/3982
we could have a new API, which would accept a container data type as input, and it would return a list with all the datasets and their corresponding locks.

This would provide a great performance improvement for client/server communication - as well as external applications - when containers contain thousands of datasets to be listed, reducing all of this to a single HTTP call.

Modification
------------
Implement an API that takes a container data type/name as input parameter, and returns as output a list with all the container datasets and their corresponding lock information. Similar to what is done for `get_dataset_locks`, provided a given dataset.

There are multiple ways to implement it, and I leave that up to the Rucio team, some ideas are:
* create a new API like get_container_locks
* or just allow the get_dataset_locks to have a container data type in the `name` parameter.

ivmfnal:
Waiting for the code review.. The PR will include changes for both #3982 and #4807

ivmfnal:
The PR is still under review...

PR #4808, opened 2021-08-26
Rules: extend get dataset locks api #3982 

ivmfnal:
Added LockClient.get_many_dataset_locks()

### DIFF
--- a/lib/rucio/client/lockclient.py
+++ b/lib/rucio/client/lockclient.py
@@ -22,7 +22,6 @@
 # - Andrew Lister <andrew.lister@stfc.ac.uk>, 2019
 # - Benedikt Ziemons <benedikt.ziemons@cern.ch>, 2021
 
-
 from requests.status_codes import codes
 from six.moves.urllib.parse import quote_plus
 
--- a/lib/rucio/client/lockclient.py
+++ b/lib/rucio/client/lockclient.py
@@ -22,6 +22,7 @@
 # - Andrew Lister <andrew.lister@stfc.ac.uk>, 2019
 # - Benedikt Ziemons <benedikt.ziemons@cern.ch>, 2021
 
+
 from requests.status_codes import codes
 from six.moves.urllib.parse import quote_plus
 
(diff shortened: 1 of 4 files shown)


lib/rucio/client/lockclient.py
```diff
@@ -61,6 +60,48 @@ def get_dataset_locks(self, scope, name):
                                                    status_code=result.status_code)
             raise exc_cls(exc_msg)
 
+    def get_locks_for_datasets(self, dataset_list):
+        """
+        Get a dataset locks of the specified list of datasets.
+
+        :param dataset_list: list dataset DIDs as dictionaries {"scope":..., "name":,...} or strings ["scope:name",...] to get locks for.
+        """
```
bari12, line 81: Add :returns: as well.

lib/rucio/client/lockclient.py
```diff
@@ -61,6 +60,48 @@ def get_dataset_locks(self, scope, name):
...
+
+        # convert did list to list of dictionaries
+
+        dids = []
+        for did in dataset_list:
```
bari12: Probably not a bad idea to put this to `common.utils` ;-)

lib/rucio/web/rest/flaskapi/v1/locks.py
```diff
@@ -92,6 +92,45 @@ def generate(vo):
...
+        did_type = request.args.get('did_type', default=None)
+        if did_type != 'dataset':
+            return 'Wrong did_type specified', 400
+
+        try:
+            def all_locks(dids, vo):
+                for did in dids:
+                    scope = did["scope"]
+                    name = did["name"]
+                    for lock in get_dataset_locks(scope, name, vo=vo):
```
bari12: I think it would be better to adapt the 
- `api.locks:get_dataset_locks`
- `core.locks:get_dataset_locks`
both to be bulk queries. (And adapt the single query to just use the bulk)
Within the core method, you can still loop it like this, but this gives us at least the option later to do something smarter in the core than just looping them. (E.g. bulk query to the db, filling temporary tables and join, etc.)

ivmfnal:
I am going to close this PR and combine issues [3982](https://github.com/rucio/rucio/issues/3982) and [4807](https://github.com/rucio/rucio/issues/4807) into single PR

PR #4816, opened 2021-09-01
Rucio download is not able to download files if the scope or name contains "/"

cserf:
Rucio download is not able to download files if the scope or name contains "/"

### DIFF
--- a/lib/rucio/client/downloadclient.py
+++ b/lib/rucio/client/downloadclient.py
@@ -236,7 +236,7 @@ class DownloadClient:
             item['scope'] = did_scope
             item['name'] = did_name
             item['sources'] = [{'pfn': pfn, 'rse': rse}]
-            dest_file_path = did_name
+            did_path_name = did_name
             if self.extract_scope_convention == 'belleii' and did_name.startswith('/'):
                 did_path_name = did_name[1:]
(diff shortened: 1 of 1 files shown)


lib/rucio/client/downloadclient.py
```diff
@@ -233,7 +236,10 @@ def download_pfns(self, items, num_threads=2, trace_custom_fields={}, traces_cop
             item['scope'] = did_scope
             item['name'] = did_name
             item['sources'] = [{'pfn': pfn, 'rse': rse}]
-            dest_file_path = os.path.join(dest_dir_path, did_name)
+            did_path_name = did_name
+            if self.extract_scope_convention == 'belleii' and did_name.startswith('/'):
```
mlassnig: extract_scope_convention can be None (as per config_get above), then this will equality check will fail

mlassnig: @bari12 maybe we should think of a way of abstracting this? having vo specific checks inside the generic download() is maybe not so good

mlassnig (commented):
two comments inside

bari12:
I'll merge it like this, but we need to think about reasonable ways to get this VO-specific code out of the code-base. 
For the server/daemon side, we have the policy modules, we should do something similar on client side as well.

bari12 (approved):

PR #4819, opened 2021-09-02
Metadata: Add an option to inherit the metadata from the parent : Closes #4515

### LINKED ISSUE #4515: Metadata: Add an option to inherit the metadata from the parents

cserf:
Metadata: Add an option to inherit the metadata from the parent : Closes #4515

### DIFF
--- a/lib/rucio/core/did.py
+++ b/lib/rucio/core/did.py
@@ -1317,26 +1317,95 @@ def get_metadata(scope, name, plugin='DID_COLUMN', session=None):
 
 
 @stream_session
-def get_metadata_bulk(dids, session=None):
+def list_parent_dids_bulk(dids, session=None):
     """
-    Get metadata for a list of dids
-    :param dids: A list of dids.
-    :param session: The database session in use.
+    List parent datasets and containers of a did.
+
+    :param dids:               A list of dids.
+    :param session:            The database session in use.
+    :returns:                  List of dids.
+    :rtype:                    Generator.
     """
     condition = []
     for did in dids:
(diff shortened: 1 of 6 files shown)


cserf:
This is the same code as https://github.com/rucio/rucio/pull/4621 that was mistakenly closed without being merged. The only change is the use of factories in the test to avoid noparallel and dirty markers.

lib/rucio/core/did.py
```diff
@@ -1303,28 +1303,97 @@ def get_metadata(scope, name, plugin='DID_COLUMN', session=None):
...
+                for entry in parent_list:
+                    if entry[-1] == (child_scope, child_name):
+                        entry.append((scope, name))
+            dids = [{'scope': did[0], 'name': did[1]} for did in parents]
+            depth += 1
+        unique_dids = [{'scope': did[0], 'name': did[1]} for did in unique_dids]
+        meta_dict = {}
+        for did in unique_dids:
+            try:
+                meta = get_metadata(did['scope'], did['name'], plugin='JSON', session=session)
```
bari12, line 1384: Just a note, here we need to remove the JSON force eventually.

bari12:
I think the PR is good. @cserf any idea about the failing tests?

mlassnig (approved):
LGTM

once of these days we need to sort out these fixed             `for did_chunk in query.yield_per(5):` things... these hardcoded bulk values are not good

PR #4822, opened 2021-09-06
Authentication: fix deletion of old tokens from DB. Closes #4815

### LINKED ISSUE #4815: Database: fix deletion of old tokens
Motivation
----------
The current deletion of old tokens does not work properly (deadlocks) and the hints do not run: https://github.com/rucio/rucio/blob/7c8f6ae1adfa0d41e533da572997bdfed6c555e1/lib/rucio/core/authentication.py#L139-L140

Modification
------------
Rewrite the queries to the new SQLAlchemy syntax, splitting each of them in a `select` and a `delete` statement. Add the following hint in the `delete` statement:
https://github.com/rucio/rucio/blob/7c8f6ae1adfa0d41e533da572997bdfed6c555e1/lib/rucio/db/sqla/models.py#L120

davidpob99:
**Authentication: fix deletion of old tokens from DB. Closes #4815**

### DIFF
--- a/lib/rucio/core/authentication.py
+++ b/lib/rucio/core/authentication.py
@@ -104,7 +104,7 @@ def get_auth_token_user_pass(account, username, password, appid, ip=None, sessio
     db_account = result['account']
 
     # remove expired tokens
-    __delete_expired_tokens_account(account=account)
+    __delete_expired_tokens_account(account=account, session=session)
 
     # create new rucio-auth-token for account
     tuid = generate_uuid()  # NOQA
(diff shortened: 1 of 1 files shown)


lib/rucio/core/authentication.py
```diff
@@ -102,10 +103,8 @@ def get_auth_token_user_pass(account, username, password, appid, ip=None, sessio
                                                                         account=account).first()
     db_account = result['account']
 
-    stmt = delete(models.Token).\
-        prefix_with('/*+ INDEX(TOKENS_ACCOUNT_EXPIRED_AT_IDX) */', dialect='oracle').\
-        where(and_(models.Token.expired_at < datetime.datetime.utcnow(), models.Token.account == account))
-    session.execute(stmt)
+    # remove expired tokens
+    __delete_expired_tokens_account(account=account)
```
mlassnig: These are missing the session parameter. How do we proceed with this?

mlassnig (changes requested):

bari12 (approved):

Issue #4751, opened 2021-07-30
Support for self-managed QoS (MAS)


bari12:
Motivation
----------
In order to support a short-term solution for self-managed QoS with the MAS system, several changes need to be done in Rucio. The general idea is that for a new replication rule, even though the data is already on storage, a `STAGEIN` request needs to be sent to FTS nevertheless, to potentially stage the data from tape to the buffer.

Modification
------------
- For storages with the RSE attribute `staging_required` staging requests need to be made for rules, whose replicas are already on storage.
- If the rules have long or infinite lifetime, the pin should be limited to a `maximum_pin_lifetime` attribute setting.
- Since the notification workflow of the rule should still be done correctly, the newly created replication rule will be in `REPLICATING` state. The usual state-workflow `REPLICATING` -> `OK` will be done for such rules
  - Need to formalize what happens if such staging requests fail. This should not impact already existing rules on storage.
  - Other rules which cover the same data on storage, will not be changed, thus they will stay in state `OK`. This somewhat breaks current rucio concepts, where replication rules are usually kept in sync.
  - Collection replica states will also not be updated.

PR #4824, opened 2021-09-07
Transfers: Support for self-managed QoS (MAS) #4751

maatthias:


lib/rucio/core/replica.py
```diff
@@ -2104,11 +2104,15 @@ def update_replicas_states(replicas, nowait=False, session=None):
...
         elif replica['state'] == ReplicaState.UNAVAILABLE:
-            rucio.core.lock.failed_transfer(scope=replica['scope'], name=replica['name'], rse_id=replica['rse_id'],
-                                            error_message=replica.get('error_message', None),
-                                            broken_rule_id=replica.get('broken_rule_id', None),
-                                            broken_message=replica.get('broken_message', None),
-                                            nowait=nowait, session=session)
+            # if qos rse don't set to UNAVAILABLE yet
+            staging_required = get_rse_attribute('staging_required', rse_id=replica['rse_id'], session=session)
+            maximum_pin_lifetime = get_rse_attribute('maximum_pin_lifetime', rse_id=replica['rse_id'], session=session)
+            if staging_required != [True] and not maximum_pin_lifetime:
```
mlassnig, line 2105: list of True? that looks suspicious

lib/rucio/core/rule_grouping.py
```diff
@@ -854,6 +854,26 @@ def __create_lock_and_replica(file, dataset, rule, rse_id, staging_area, availab
                                                         lifetime=lifetime,
                                                         session=session))
 
+    # for qos if staging_required then pin to maximum_pin_lifetime
+    staging_required = get_rse_attribute('staging_required', rse_id=rse_id, session=session)
+    maximum_pin_lifetime = get_rse_attribute('maximum_pin_lifetime', rse_id=rse_id, session=session)
+
+    if staging_required == [True] and maximum_pin_lifetime:
+        logger(logging.INFO, 'Destination RSE is type QoS')
```
mlassnig, line 862: add name of RSE to log

mlassnig (changes requested):

bari12 (changes requested):
Hi Matt, this is a good start, but it is not enough. We really need to work through the entire chain and think about what to expect in each case.

# Replica does not exist on MAS RSE yet:
- In this case a transfer should be scheduled (this is already working as it is ✅ )
- Lock transitions:
  - Need to make sure that the locks are transitioned to OK on success ✅ 
  - Need to make sure that the locks are transitioned to STUCK on failure ❌ 
    - Your changes in core/replica.py basically would block this for a FAILED transfer. This needs to be changed because this way even data transfered into the RSE would never FAIL.
- Notifications should be sent on SUCCES and FAILED transfers 
  - This is derived from the lock transitions. It will work on success ones ✅ , but not on failed ❌ 
- One special thing about this is that we still need to send a bring_online/lifetime to storage (Actually not sure if this works on a simple TPC request - please investigate) - if not possible, MAS need to figure out on it's own when it is allowed to downgrade the replica to the TAPE area (Probably something we need to discuss) ❓ 

# Replica already exists on MAS (We don't know if DISK or TAPE)
This case is a bit more tricky.

- Change in Rule grouping is good, we do a STAGEIN if the file is already on the RSE ✅ 
- Lock transitions:
  - Need to make sure that the locks are transitioned to OK on success ✅ 
  - Now the failure case is the complicated one and this is where this PR mostly will fail ❌ 
    - Other existing rules should not be transitioned to STUCK just because this one failed. The data is still safe on tape after all. 
    - This specific rule can be moved to STUCK (but others on the RSE should not, but `lock.failed_transfer` would cause this.
- Notifications
  - We should send a notification for the failure case as well, this PR (since the event is generated by `failed_transfer` would not. 

# Changes needed
- I think we need to move some code into `failed_transfer`, `successful_transfer`; It would help if you document what notifications, lock and rule states we expect in each different case. This blows up the state transitions, but there is no easy way around.
- We need much more testing for this. There needs to be a test for each of the two cases above, and at least for the success/failure case and then to check if the rules/locks/replicas are in the right state.

bari12:
Closing until re-submission.

Issue #4565, opened 2021-04-21
Expand automatic recovery of suspicious replicas


ChristophAmes:
Motivation
----------
The idea is to expand the daemon that deals with the recovery of suspicious replicas, so that less replicas need to be dealt with manually . The number of suspicious replicas on each RSE is checked. If an RSE has more than X (to be determined) suspicious replicas, then the RSE is considered problematic and (eventually) a Jira ticket will be sent to the RSE manager. If an RSE isn't considered problematic, then the replicas are dealt with individually, based to the file type. 

Modification
------------
Expanding  lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py

PR #4827, opened 2021-09-08
Recovery: Expanding the daemon suspicious_replica_recoverer #4565

ChristophAmes:
The idea is to expand the daemon "suspicious_replica_recoverer", so that it goes into more detail when dealing with suspicious files.

The first main change is that it can now deal with both suspicious replicas that have additional copies on other RSEs and replicas that are the last remaining copy, the latter being previously ignored. The last remaining copies are dealt with differently based on the file type / metadata of the replica:
Log files and user files are declared bad without further investigation. Replicas with the metadata " datatype='RAW' " are not declared bad, but are instead ignored (meaning that they will have to be dealt with manually).
The rest of the file types are currently being ignored, to make sure no mistakes are made. This can be easily changed by adding new entries to suspicious_replica_recoverer.json.

To be able to make these changes, the function "get_suspicious_replicas" from lib/rucio/core/replica.py had to be changed. The parameter "available_elsewhere" now has three options:
0: All suspicious replicas
1: Suspicious replicas that have at least one copy on another RSE
2: Suspicious replicas that are the last remaining copy

The second main change is that RSEs with more than a certain number of suspicious replicas (currently set to 5, but requires testing when the daemon is run in passive mode) are considered "problematic". This means that it is assumed that the high number of suspicious replicas is due to the RSE having a problem, as opposed to the replicas themselves having problems. In such a case, the suspicious replicas are declared temporarily unavailable.

### DIFF
--- a/bin/rucio-replica-recoverer
+++ b/bin/rucio-replica-recoverer
@@ -15,11 +15,14 @@
 # limitations under the License.
 #
 # Authors:
+# - Cedric Serfon, <cedric.serfon@cern.ch>, 2018
 # - Jaroslav Guenther <jaroslav.guenther@cern.ch>, 2019
 # - Martin Barisits <martin.barisits@cern.ch>, 2019
 # - Patrick Austin <patrick.austin@stfc.ac.uk>, 2020
 # - Benedikt Ziemons <benedikt.ziemons@cern.ch>, 2020
 # - David Población Criado <david.poblacion.criado@cern.ch>, 2021
+# - Christoph Ames <cames@cern.ch>, 2021
+# PY3K COMPATIBLE
 
 """
 Replica-Recoverer is a daemon that declares suspicious replicas as bad if they are found available on other RSE.
@@ -207,10 +210,11 @@ $$ from rucio.core.replica import get_suspicious_files
 $$ from datetime import datetime, timedelta
 $$ from_date = datetime.now() - timedelta(days=3)
 $$ from rucio.core.replica import list_bad_replicas_status
+$$ from rucio.common.constants import SuspiciousAvailability
 
-$$ get_suspicious_files('MOCK_SUSPICIOUS',younger_than=from_date, nattempts=10, is_suspicious=True, available_elsewhere=True)
+$$ get_suspicious_files('MOCK_SUSPICIOUS',younger_than=from_date, nattempts=10, is_suspicious=True, available_elsewhere=SuspiciousAvailability["EXIST_COPIES"].value)
 
->>> get_suspicious_files('MOCK_SUSPICIOUS',younger_than=from_date, nattempts=10, is_suspicious=True, available_elsewhere=True)
+>>> get_suspicious_files('MOCK_SUSPICIOUS',younger_than=from_date, nattempts=10, is_suspicious=True, available_elsewhere=SuspiciousAvailability["EXIST_COPIES"].value)
 []
 
 $$ list_bad_replicas_status(rse='MOCK_SUSPICIOUS', younger_than=from_date)
@@ -235,13 +239,12 @@ Note that attempting the use the ``--vos`` argument when in single-VO mode will 
   $ rucio-replica-recoverer --run-once --vos abc xyz
   2020-07-28 15:21:33,349 5488    WARNING Ignoring argument vos, this is only applicable in a multi-VO setup.
 ''')  # NOQA: E501
-    parser.add_argument("--nattempts", action="store", default=10, help='Minimum count of suspicious file replica appearance in bad_replicas table.Default value is 10.')
-    parser.add_argument("--younger-than", action="store", default=3, help='Consider all file replicas logged in bad_replicas table since speicified number of younger-than. Default value is 3.')
-    parser.add_argument("--rse-expression", action="store", default='MOCK', help='The RSE on which the recovery should be happening.')
+    parser.add_argument("--nattempts", action="store", default=10, help='Minimum count of suspicious file replica appearance in bad_replicas table. Default value is 10.')
+    parser.add_argument("--younger-than", action="store", default=3, help='Consider all file replicas logged in bad_replicas table since speicified number of younger-than days. Default value is 3.')
     parser.add_argument('--vos', nargs='+', type=str, help='Optional list of VOs to consider. Only used in multi-VO mode.')
     parser.add_argument("--run-once", action="store_true", default=False, help='One iteration only.')
-    parser.add_argument("--max-replicas-per-rse", action="store", default=100, help='The maximum number of suspicious replicas found on an RSE which may be declared bad. If a higher count is found, no action is be taken.')
-    parser.add_argument('--sleep-time', action="store", default=60, type=int, help='Concurrency control: thread sleep time after each chunk of work')
+    parser.add_argument("--limit-suspicious-files-on-rse", action="store", default=5, help='Maximum number of suspicious replicas on an RSE before that RSE is considered problematic and the suspicious replicas on that RSE are declared "TEMPORARY_UNAVAILABLE". Default value is 5.')
+    parser.add_argument('--sleep-time', action="store", default=300, type=int, help='Concurrency control: thread sleep time after each chunk of work')
     return parser
 
 
@@ -250,6 +253,6 @@ if __name__ == "__main__":
     PARSER = get_parser()
     ARGS = PARSER.parse_args()
     try:
-        run(once=ARGS.run_once, younger_than=ARGS.younger_than, nattempts=ARGS.nattempts, rse_expression=ARGS.rse_expression, vos=ARGS.vos, max_replicas_per_rse=ARGS.max_replicas_per_rse, sleep_time=ARGS.sleep_time)
+        run(once=ARGS.run_once, younger_than=ARGS.younger_than, nattempts=ARGS.nattempts, vos=ARGS.vos, limit_suspicious_files_on_rse=ARGS.limit_suspicious_files_on_rse)
     except KeyboardInterrupt:
         stop()
--- a/lib/rucio/core/replica.py
+++ b/lib/rucio/core/replica.py
@@ -40,6 +40,7 @@
 # - Gabriele Fronzé <sucre.91@hotmail.it>, 2021
 # - David Población Criado <david.poblacion.criado@cern.ch>, 2021
 # - Joel Dierkes <joel.dierkes@cern.ch>, 2021
+# - Christoph Ames <cames@cern.ch>, 2021
 
 from __future__ import print_function
 
@@ -73,6 +74,7 @@ from rucio.common import exception
 from rucio.common.config import config_get
 from rucio.common.types import InternalScope
 from rucio.common.utils import chunks, clean_surls, str_to_date, add_url_query
+from rucio.common.constants import SuspiciousAvailability
 from rucio.core.credential import get_signed_url
 from rucio.core import config as config_core
 from rucio.core.rse import get_rse, get_rse_name, get_rse_attribute, get_rse_vo, list_rses
@@ -3179,7 +3181,7 @@ def get_replicas_state(scope=None, name=None, session=None):
 
 
 @read_session
-def get_suspicious_files(rse_expression, filter_=None, **kwargs):
+def get_suspicious_files(rse_expression, available_elsewhere, filter_=None, logger=logging.log, younger_than=10, nattempts=0, session=None, exclude_states=['B', 'R', 'D'], is_suspicious=False):
     """
     Gets a list of replicas from bad_replicas table which are: declared more than <nattempts> times since <younger_than> date,
     present on the RSE specified by the <rse_expression> and do not have a state in <exclude_states> list.
@@ -3193,8 +3195,10 @@ def get_suspicious_files(rse_expression, filter_=None, **kwargs):
     :param: exclude_states: List of states which eliminates replicas from search result if any of the states in the list
                             was declared for a replica since younger_than date. Allowed values
                             = ['B', 'R', 'D', 'L', 'T', 'S'] (meaning 'BAD', 'RECOVERED', 'DELETED', 'LOST', 'TEMPORARY_UNAVAILABLE', 'SUSPICIOUS').
-    :param: available_elsewhere: If True, only replicas declared in addition as AVAILABLE on another RSE
-                                 than the one in the bad_replicas table will be taken into account. Default value = False.
+    :param: available_elsewhere: Default: SuspiciousAvailability["ALL"].value, all suspicious replicas are returned.
+                                 If SuspiciousAvailability["EXIST_COPIES"].value, only replicas that additionally have copies declared as AVAILABLE on at least one other RSE
+                                 than the one in the bad_replicas table will be taken into account.
+                                 If SuspiciousAvailability["LAST_COPY"].value, only replicas that do not have another copy declared as AVAILABLE on another RSE will be taken into account.
     :param: is_suspicious: If True, only replicas declared as SUSPICIOUS in bad replicas table will be taken into account. Default value = False.
     :param session: The database session in use. Default value = None.
 
@@ -3202,12 +3206,15 @@ def get_suspicious_files(rse_expression, filter_=None, **kwargs):
     [{'scope': scope, 'name': name, 'rse': rse, 'rse_id': rse_id, cnt': cnt, 'created_at': created_at}, ...]
     """
 
-    younger_than = kwargs.get("younger_than", datetime.now() - timedelta(days=10))
-    nattempts = kwargs.get("nattempts", 0)
-    session = kwargs.get("session", None)
-    exclude_states = kwargs.get("exclude_states", ['B', 'R', 'D'])
-    available_elsewhere = kwargs.get("available_elsewhere", False)
-    is_suspicious = kwargs.get("is_suspicious", False)
+    if available_elsewhere not in [SuspiciousAvailability["ALL"].value, SuspiciousAvailability["EXIST_COPIES"].value, SuspiciousAvailability["LAST_COPY"].value]:
+        logger(logging.WARNING, """ERROR, available_elsewhere must be set to one of the following:
+        SuspiciousAvailability["ALL"].value: (default) all suspicious replicas are returned
+        SuspiciousAvailability["EXIST_COPIES"].value: only replicas that additionally have copies declared as AVAILABLE on at least one other RSE are returned
+        SuspiciousAvailability["LAST_COPY"].value: only replicas that do not have another copy declared as AVAILABLE on another RSE are returned""")
+        raise exception.RucioException("""ERROR, available_elsewhere must be set to one of the following:
+        SuspiciousAvailability["ALL"].value: (default) all suspicious replicas are returned
+        SuspiciousAvailability["EXIST_COPIES"].value: only replicas that additionally have copies declared as AVAILABLE on at least one other RSE are returned
+        SuspiciousAvailability["LAST_COPY"].value: only replicas that do not have another copy declared as AVAILABLE on another RSE are returned""")
 
     # only for the 2 web api used parameters, checking value types and assigning the default values
     if not isinstance(nattempts, int):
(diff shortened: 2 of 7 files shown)


bin/rucio-replica-recoverer
```diff
@@ -235,13 +238,11 @@ Note that attempting the use the ``--vos`` argument when in single-VO mode will
...
+    parser.add_argument("--nattempts", action="store", default=10, help='Minimum count of suspicious file replica appearance in bad_replicas table. Default value is 10.')
     parser.add_argument("--younger-than", action="store", default=3, help='Consider all file replicas logged in bad_replicas table since speicified number of younger-than. Default value is 3.')
-    parser.add_argument("--rse-expression", action="store", default='MOCK', help='The RSE on which the recovery should be happening.')
     parser.add_argument('--vos', nargs='+', type=str, help='Optional list of VOs to consider. Only used in multi-VO mode.')
     parser.add_argument("--run-once", action="store_true", default=False, help='One iteration only.')
-    parser.add_argument("--max-replicas-per-rse", action="store", default=100, help='The maximum number of suspicious replicas found on an RSE which may be declared bad. If a higher count is found, no action is be taken.')
-    parser.add_argument('--sleep-time', action="store", default=60, type=int, help='Concurrency control: thread sleep time after each chunk of work')
```
cserf, line 244: Why did you remove `--sleep-time` ?  Can you put it back please

lib/rucio/core/replica.py
```diff
@@ -3168,9 +3171,19 @@ def get_suspicious_files(rse_expression, filter_=None, **kwargs):
     nattempts = kwargs.get("nattempts", 0)
     session = kwargs.get("session", None)
     exclude_states = kwargs.get("exclude_states", ['B', 'R', 'D'])
-    available_elsewhere = kwargs.get("available_elsewhere", False)
+    available_elsewhere = kwargs.get("available_elsewhere", 0)
     is_suspicious = kwargs.get("is_suspicious", False)
 
+    if available_elsewhere not in [0, 1, 2]:
+        print("""ERROR, available_elsewhere must be set to one of the following:
```
cserf: Use a logger instead (see https://github.com/rucio/rucio/blob/edbad7d45262eda611f19a69e1ab1cb726098764/lib/rucio/core/replica.py#L945)

lib/rucio/core/replica.py
```diff
@@ -3223,7 +3246,7 @@ def get_suspicious_files(rse_expression, filter_=None, **kwargs):
     # finally, the results are grouped by RSE, scope, name and required to have
     # at least 'nattempts' occurrences in the result of the query per replica
     query_result = query.group_by(models.RSEFileAssociation.rse_id, bad_replicas_alias.scope, bad_replicas_alias.name).having(func.count() > nattempts).all()
-    # print(query)
```
cserf, line 3264: Remove

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
```diff
@@ -40,30 +42,34 @@
...
 from rucio.common.types import InternalAccount
-from rucio.common.utils import daemon_sleep
 from rucio.core.heartbeat import live, die, sanity_check
 from rucio.core.monitor import record_counter
-from rucio.core.replica import list_replicas, declare_bad_file_replicas, get_suspicious_files
+from rucio.core.did import get_metadata
+from rucio.core.replica import list_replicas, get_suspicious_files, add_bad_pfns, declare_bad_file_replicas
 from rucio.core.rse_expression_parser import parse_expression
+
+# from rucio.core.rse import list_rses
```
cserf: Remove

lib/rucio/tests/test_replica_recoverer.py
```diff
@@ -169,7 +173,8 @@ def test_replica_recoverer(self):
 
         # Run replica recoverer once
         try:
-            run(once=True, younger_than=1, nattempts=2, rse_expression='MOCK_SUSPICIOUS')
+            run(once=True, younger_than=1, nattempts=2, limit_suspicious_files_on_rse=5)
+            print("Run")
```
cserf: Remove the print statement

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
```diff
@@ -145,75 +134,190 @@ def declare_suspicious_replicas_bad(once=False, younger_than=3, nattempts=10, rs
 
             start = time.time()
 
-            logging.info('replica_recoverer[%i/%i]: ready to query replicas at RSE %s,'
-                         + ' reported suspicious in the last %i days at least %i times which are available on other RSEs.',  # NOQA: W503
-                         worker_number, total_workers, rse_expression, younger_than, nattempts)
+            logging.info('replica_recoverer[%i/%i]: Ready to query replicas which were'
```
cserf: Use a logger (see in https://github.com/rucio/rucio/blob/edbad7d45262eda611f19a69e1ab1cb726098764/lib/rucio/daemons/reaper/reaper.py for instance)

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
```diff
@@ -103,23 +109,6 @@ def declare_suspicious_replicas_bad(once=False, younger_than=3, nattempts=10, rs
             vos = [v['vo'] for v in list_vos()]
         logging.info('replica_recoverer: This instance will work on VO%s: %s' % ('s' if len(vos) > 1 else '', ', '.join([v for v in vos])))
 
-    # Don't require a result from the expression at each VO, only raise if we can't get a result from any of them
-    rses = []
-    exceptions_raised = 0
-    for vo in vos:
-        try:
-            parsed_rses = parse_expression(expression=rse_expression, filter_={'vo': vo})
```
cserf, line 113: If `rse_expression` is not used anymore, maybe it can be removed from the method signature

lib/rucio/tests/test_replica_recoverer.py
```diff
@@ -169,7 +173,8 @@ def test_replica_recoverer(self):
 
         # Run replica recoverer once
         try:
-            run(once=True, younger_than=1, nattempts=2, rse_expression='MOCK_SUSPICIOUS')
+            run(once=True, younger_than=1, nattempts=2, limit_suspicious_files_on_rse=5)
```
cserf, line 210: I would add 1 or 2 other tests :
- 1 with RAW data and check that the daemon always ignore it
- 1 with datatype not specified in the json. For now it should ignore it
- 1 with a datatype specified and action `declare_bad`.

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
```diff
@@ -145,75 +134,190 @@ def declare_suspicious_replicas_bad(once=False, younger_than=3, nattempts=10, rs
...
-            logging.info('replica_recoverer[%i/%i]: ready to query replicas at RSE %s,'
-                         + ' reported suspicious in the last %i days at least %i times which are available on other RSEs.',  # NOQA: W503
-                         worker_number, total_workers, rse_expression, younger_than, nattempts)
+            logging.info('replica_recoverer[%i/%i]: Ready to query replicas which were'
+                         + ' reported as suspicious in the last %i days at least %i times.',
+                         worker_number, total_workers, younger_than, nattempts)
+
+            getfileskwargs_avail_elsewhere = {'younger_than': younger_than,
+                                              'nattempts': nattempts,
+                                              'exclude_states': ['B', 'R', 'D', 'L', 'T'],
```
cserf: I wonder if we should consider `T` i.e. temporary available in the list of states to exclude. To be discussed with Rod

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
```diff
@@ -145,75 +134,190 @@ def declare_suspicious_replicas_bad(once=False, younger_than=3, nattempts=10, rs
...
-                                            worker_number, total_workers, max_replicas_per_rse, str(len(surls_to_recover[vo][rse_id])), rse)
-                        else:
-                            declare_bad_file_replicas(pfns=surls_to_recover[vo][rse_id], reason='Suspicious. Automatic recovery.', issuer=InternalAccount('root', vo=vo), status=BadFilesStatus.BAD, session=None)
-                            logging.info('replica_recoverer[%i/%i]: finished declaring bad replicas on %s.', worker_number, total_workers, rse)
+                logging.info('replica_recoverer[%i/%i]: Start replica recovery for VO: %s', worker_number, total_workers, vo)
+                recoverable_replicas = {}
+                if vo not in recoverable_replicas:
+                    recoverable_replicas[vo] = {}
+
+                rse_list = sorted([rse for rse in parse_expression('enable_suspicious_file_recovery=true')], key=lambda k: k['rse'])
```
cserf: Need to add a VO filter since there is a loop on VO

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
```diff
@@ -145,75 +134,190 @@ def declare_suspicious_replicas_bad(once=False, younger_than=3, nattempts=10, rs
...
+
+                logging.debug("List of RSEs with enable_suspicious_file_recovery = True:")
+                for i in rse_list:
+                    logging.debug(i)
```
cserf: Put the prepend string. Done automatically when you use the logger
Comment valid for all the other calls to logging

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
```diff
@@ -145,75 +134,190 @@ def declare_suspicious_replicas_bad(once=False, younger_than=3, nattempts=10, rs
...
+                    rse_expr = rse['rse']
+                    cnt_surl_not_found = 0
+                    if rse_expr not in recoverable_replicas[vo]:
+                        recoverable_replicas[vo][rse_expr] = {}
+                    # Get a dictionary of the suspicious replicas on the RSE that have available copies on other RSEs
+                    suspicious_replicas_avail_elsewhere = get_suspicious_files(rse_expr, filter={'vo': vo}, **getfileskwargs_avail_elsewhere)
+                    # Get the suspicious replicas that are the last remaining copies
+                    suspicious_replicas_last_copy = get_suspicious_files(rse_expr, filter={'vo': vo}, **getfileskwargs_last_copy)
+
+                    logging.debug('\n')
```
cserf: Logs are sent to elasticSearch. Try to avoid putting empty lines

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
```diff
@@ -145,75 +134,190 @@ def declare_suspicious_replicas_bad(once=False, younger_than=3, nattempts=10, rs
...
+                                    logging.warning('replica_recoverer[%i/%i]: Skipping suspicious replica %s on %s, no surls were found.', worker_number, total_workers, rep_name, rse_expr)
+
+                    if suspicious_replicas_last_copy:
+                        for replica in suspicious_replicas_last_copy:
+                            if vo == replica['scope'].vo:
+                                scope = replica['scope']
+                                rep_name = replica['name']
+                                rse_id = replica['rse_id']
+                                surl_not_found = True
+                                for rep in list_replicas([{'scope': scope, 'name': rep_name}]):
```
cserf, line 231: I'm not sure I understand this check. Since it is the last replica, you should never find a valid replica.

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
```diff
@@ -145,75 +134,190 @@ def declare_suspicious_replicas_bad(once=False, younger_than=3, nattempts=10, rs
...
+                logging.info('\n\n\n\n\n\n\n\n')
+                logging.info('replica_recoverer[%i/%i]: All RSEs have been checked for suspicious replicas. Total time: %.2f seconds.', worker_number, total_workers, time.time() - start)
+                logging.info('replica_recoverer[%i/%i]: Begin check for problematic RSEs.\n\n\n', worker_number, total_workers)
+                time_start_check_probl = time.time()
+
+                # If an RSE has more than *limit_suspicious_files_on_rse* suspicious files, then there might be a problem with the RSE.
+                # The suspicious files are marked as temporarily unavailable.
+                list_problematic_rses = []
+                for rse_key in list(recoverable_replicas[vo].keys()):
+                    if len(recoverable_replicas[vo][rse_key].values()) > limit_suspicious_files_on_rse:
```
cserf, line 256: I wonder if we shouldn't put an upper limit too. If there are millions of files, we shouldn't probably declare them temporary unavailable, but just set the site to unavailable and notify the site.
But that's second order correction that can be added later

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
```diff
@@ -145,75 +134,190 @@ def declare_suspicious_replicas_bad(once=False, younger_than=3, nattempts=10, rs
...
+                            logging.info('replica_recoverer[%i/%i]: Scope: %s    Name: %s', worker_number, total_workers, rse_values['scope'], rse_values['name'])
+                        # Remove the RSE from the dictionary as it has been dealt with.
+                        del recoverable_replicas[vo][rse_key]
+
+                logging.info("\n\n")
+                logging.info("replica_recoverer[%i/%i]: Following RSEs were deemed problematic (total: %i)", worker_number, total_workers, len(list_problematic_rses))
+                for rse in list_problematic_rses:
+                    logging.info("replica_recoverer[%i/%i]: %s", worker_number, total_workers, rse)
+
+                json_file = open("/opt/rucio/etc/suspicious_replica_recoverer.json")
```
cserf: Just do it at the beginning of the loop on VOs.

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
```diff
@@ -145,75 +134,190 @@ def declare_suspicious_replicas_bad(once=False, younger_than=3, nattempts=10, rs
...
+                json_data = json.load(json_file)
```
cserf: Add a try except (if the json is badly formated it will crash). Also a validation of the json (check if mandatory fields are available can be useful)

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
```diff
@@ -145,75 +134,190 @@ def declare_suspicious_replicas_bad(once=False, younger_than=3, nattempts=10, rs
...
+                    # Get the rse_id by going to one of the suspicious replicas from that RSE and reading it from there
+                    rse_id = list(recoverable_replicas[vo][rse_key].values())[0]['rse_id']
+                    for replica_key in list(recoverable_replicas[vo][rse_key].keys()):
+                        if recoverable_replicas[vo][rse_key][replica_key]['available_elsewhere'] is True:
+                            # Replicas with other copies on at least one other RSE can safely be labeled as bad
+                            files_to_be_declared_bad.append(recoverable_replicas[vo][rse_key][replica_key]['surl'])
+                            # Remove replica from dictionary
+                            del recoverable_replicas[vo][rse_key][replica_key]
+                        elif recoverable_replicas[vo][rse_key][replica_key]['available_elsewhere'] is False:
+                            if (recoverable_replicas[vo][rse_key][replica_key]['name'].startswith("log.")) or (recoverable_replicas[vo][rse_key][replica_key]['name'].startswith("user")):
```
cserf, line 293: ATLAS specific policy. Will have to be modified if used by another VO. OK to keep like this for now

bin/rucio-replica-recoverer
```diff
@@ -235,13 +238,12 @@ Note that attempting the use the ``--vos`` argument when in single-VO mode will
...
+    parser.add_argument("--limit-suspicious-files-on-rse", action="store", default=5, help='Maximum number of suspicious replicas on an RSE before that RSE is considered problematic and the suspicious replicas on that RSE are declared "TEMPORARY_UNAVAILABLE". Default value is 5.')
+    parser.add_argument('--sleep-time', action="store", default=300, type=int, help='Concurrency control: thread sleep time after each chunk of work')
```
mlassnig, line 247: 60 to 300?

bin/rucio-replica-recoverer
```diff
@@ -208,9 +211,9 @@ $$ from datetime import datetime, timedelta
 $$ from_date = datetime.now() - timedelta(days=3)
 $$ from rucio.core.replica import list_bad_replicas_status
 
-$$ get_suspicious_files('MOCK_SUSPICIOUS',younger_than=from_date, nattempts=10, is_suspicious=True, available_elsewhere=True)
+$$ get_suspicious_files('MOCK_SUSPICIOUS',younger_than=from_date, nattempts=10, is_suspicious=True, available_elsewhere=1)
```
mlassnig: Can we change this? It's not really clear what the numbers mean

`
0: All suspicious replicas
1: Suspicious replicas that have at least one copy on another RSE
2: Suspicious replicas that are the last remaining copy
`

maybe better to add something to lib/rucio/common/constants?

SUSPICIOUS.ALL
SUSPICIOUS.EXISTING_REPLICAS
SUSPICIOUS.LAST_REPLICA

bin/rucio-replica-recoverer
mlassnig, line 247: 60 to 300?

ChristophAmes: The daemon generally takes up to five minutes to run, since it now runs over all RSEs at once, instead of going over them individually. It doesn't make sense to have sleep_time set to 60, as it will always take longer. I did however just notice that I forgot to include daemon_sleep, so I just created a new commit and pushed.

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
```diff
@@ -41,30 +42,34 @@
...
 from rucio.core.vo import list_vos
-from rucio.db.sqla.constants import BadFilesStatus
 from rucio.db.sqla.util import get_db_time
 
 GRACEFUL_STOP = threading.Event()
 
 
-def declare_suspicious_replicas_bad(once=False, younger_than=3, nattempts=10, rse_expression='MOCK', vos=None,
-                                    max_replicas_per_rse=100, sleep_time=60):
+# logging.basicConfig(filename='example.log', filemode='w', level=logging.DEBUG)
```
cserf: Remove

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
```diff
@@ -136,112 +128,237 @@ def declare_suspicious_replicas_bad(once=False, younger_than=3, nattempts=10, rs
...
+
             start = time.time()
 
-            logging.info('replica_recoverer[%i/%i]: ready to query replicas at RSE %s,'
-                         + ' reported suspicious in the last %i days at least %i times which are available on other RSEs.',  # NOQA: W503
-                         worker_number, total_workers, rse_expression, younger_than, nattempts)
+            try:
+                json_file = open("/opt/rucio/etc/suspicious_replica_recoverer.json")
+            except:
+                logger(logging.DEBUG, "An error occured whilst trying to open the json file.")
```
cserf: WARNING or ERROR, but not DEBUG

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
```diff
@@ -136,112 +128,237 @@ def declare_suspicious_replicas_bad(once=False, younger_than=3, nattempts=10, rs
...
+                break
+
+            json_data = json.load(json_file)
```
cserf: You should put a try except and catch `ValueError: No JSON object could be decoded`

cserf (dismissed):
Just a few cosmetic changes but I think it's good to go.

cserf (dismissed):

bin/rucio-replica-recoverer
bari12: This should be added with the constants then to the doc/code as well

lib/rucio/core/replica.py
```diff
@@ -3136,7 +3138,7 @@ def get_replicas_state(scope=None, name=None, session=None):
 
 
 @read_session
-def get_suspicious_files(rse_expression, filter_=None, **kwargs):
+def get_suspicious_files(rse_expression, filter_=None, logger=logging.log, **kwargs):
```
bari12: Any reason why `kwargs` is used here? We should really try to avoid this for core functions.

bari12 (commented):
Hi @ChristophAmes 
Very sorry for the late comments here. @cserf just asked me about this PR and I did not realise that it is still unmerged. Just two comments.

bari12:
Please also resolve the comments you already addressed.

lib/rucio/core/replica.py
ChristophAmes: The documentation needs a few changes, since the daemon has a different structure. I'll have to do that in the future.

bari12: Ok, for me it's fine to do this later, please create a ticket for it though and create a PR later on 😄

cserf (approved):

bari12:
Hi,
I am merging this like this to have it in the release, but there is a few things which should be addressed:
- At the core level, do not deal with the constant values anymore, here you should assume (and ensure) that it is the actual constants handed in to the core. (I know there is a few other parts in the code where this is not the case, but these are also not clean, and we should get away from that)
- In the daemon itself there is quite a few cases where you loop over replicas and then do individual `list_replicas` calls. That's not very efficient and these should be changed to bulk calls where possible.

Best,
Martin

PR #4829, opened 2021-09-08
Skip updating replica in uploadclient if register_after_upload is true

### LINKED ISSUE #4828: Skip updating replica in uploadclient if register_after_upload is true
Motivation
----------
In https://github.com/rucio/rucio/blob/b6092bcd78f8cfac079e5b3cc25a92b4163960ed/lib/rucio/client/uploadclient.py#L304-L308 there is no need to update the replica state if it is registered after upload, since it will already be registered with state `A`. Further, the permissions for updating a replica can be (are for CMS) more restrictive than for injection.

Modification
------------
Put L306-309 in an `else` block. Also, `update_replicas_states` raises an exception if it fails, so the handling is likely incorrect here.

nsmith-:
Closes #4828

### DIFF
--- a/lib/rucio/client/uploadclient.py
+++ b/lib/rucio/client/uploadclient.py
@@ -308,7 +308,7 @@ class UploadClient:
                         try:
                             self.client.update_replicas_states(rse, files=[replica_for_api])
                         except Exception as error:
-                            logger(logging.WARNING, 'Failed to update replica state')
+                            logger(logging.ERROR, 'Failed to update replica state for file {}'.format(basename))
                             logger(logging.DEBUG, 'Details: {}'.format(str(error)))
 
                 # add file to dataset if needed
--- a/lib/rucio/client/uploadclient.py
+++ b/lib/rucio/client/uploadclient.py
@@ -303,9 +303,13 @@ class UploadClient:
                 if not no_register:
                     if register_after_upload:
                         self._register_file(file, registered_dataset_dids)
-                    replica_for_api = self._convert_file_for_api(file)
-                    if not self.client.update_replicas_states(rse, files=[replica_for_api]):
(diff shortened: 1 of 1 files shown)


lib/rucio/client/uploadclient.py
```diff
@@ -303,9 +303,13 @@ def _pick_random_rse(rse_expression):
...
                     if register_after_upload:
                         self._register_file(file, registered_dataset_dids)
-                    replica_for_api = self._convert_file_for_api(file)
-                    if not self.client.update_replicas_states(rse, files=[replica_for_api]):
-                        logger(logging.WARNING, 'Failed to update replica state')
+                    else:
+                        replica_for_api = self._convert_file_for_api(file)
+                        try:
+                            self.client.update_replicas_states(rse, files=[replica_for_api])
+                        except Exception as error:
```
nsmith-, line 310: I'm not sure if this should be considered a fatal error or not. Some other places in this routine only emit warnings for what seem to me fatal errors.

mlassnig: Hmmmm, I concur, fatal seems the appropriate one. @bari12 opinion?

bari12: Yes, should be fatal.

mlassnig: @nsmith- will you change it to FATAL? or we keep it like this?

bari12: @nsmith- ping 🔔

nsmith-: ah sorry for the delay, I changed warning -> error. Let me know if something more needs to be done, its not clear to me why not all cases are errors, e.g. line 288 or 319 have only warning

bari12 (approved):

PR #4835, opened 2021-09-10
Core & Internals: merged changes for issues 3982 and 4807 #4807

ivmfnal:
combined PR for issues 3982 and 4807 since both affect same files


ivmfnal:
Re-staring this pull request

Issue #4830, opened 2021-09-09
Naming restrictions should be enforced for LFN2PFN and SURL algorithms in policy packages


jamesp-epcc:
The documentation already states that LFN2PFN and SURL algorithms registered in policy packages should have their names prefixed with the relevant VO name to avoid naming conflicts in multi-VO installations. This rule should be enforced by the code, as the new architecture makes it possible to do this.

PR #4842, opened 2021-09-13
Core & Internals: Enforce naming restrictions on algorithms registered from policy packages #4830

jamesp-epcc:
This PR ensures that algorithms registered by policy packages in multi-VO Rucio installations include the VO name at the start of the algorithm name, to avoid name clashes.

### DIFF
--- a/lib/rucio/common/exception.py
+++ b/lib/rucio/common/exception.py
@@ -1051,7 +1051,7 @@ class InvalidAlgorithmName(RucioException):
     """
     The given algorithm name is not valid for the VO.
     """
-    def __init__(self, *args, **kwargs):
+    def __init__(self, algorithm, vo, *args, **kwargs):
         super(InvalidAlgorithmName, self).__init__(*args, **kwargs)
-        self.message = 'The given algorithm name is not valid for this VO'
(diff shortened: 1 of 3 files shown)


lib/rucio/common/exception.py
```diff
@@ -1025,3 +1025,13 @@ def __init__(self, *args, **kwargs):
...
         self.error_code = 97
+
+
+class InvalidAlgorithmName(RucioException):
+    """
+    The given algorithm name is not valid for the VO.
+    """
+    def __init__(self, *args, **kwargs):
+        super(InvalidAlgorithmName, self).__init__(*args, **kwargs)
+        self.message = 'The given algorithm name is not valid for this VO'
```
mlassnig: add actual algorithmname to message. can the validation string also be added?

mlassnig (changes requested):

lib/rucio/common/exception.py
jamesp-epcc: I've added both to the exception message.

mlassnig:
There's a merge conflict that needs to be resolved.

jamesp-epcc:
> There's a merge conflict that needs to be resolved.

I've rebased with latest master.

bari12 (approved):

PR #4844, opened 2021-09-13
Core & Internals: merged changes #4807

ivmfnal:
merged all changes


lib/rucio/web/rest/flaskapi/v1/locks.py
```diff
@@ -76,18 +78,65 @@ def get(self, scope_name):
         :status 406: Not Acceptable.
         :returns: Line separated list of dictionary with lock information.
         """
+
         did_type = request.args.get('did_type', default=None)
-        if did_type != 'dataset':
+        if did_type not in ('dataset', 'container'):
```
mlassnig: don't we have a collection type that can be checked here? or is the string check required?

ivmfnal: What are you suggesting ?

lib/rucio/client/lockclient.py
```diff
@@ -47,15 +46,68 @@ def get_dataset_locks(self, scope, name):
...
             locks = self._load_json_data(result)
-            return locks
+            # reformat as a dictionary of lists indexed by dataset "scope:name"
+            return self._sort_locks_by_dataset(locks)
+        else:
+            exc_cls, exc_msg = self._get_exception(headers=result.headers,
+                                                   status_code=result.status_code)
+            raise exc_cls(exc_msg)
+
+    def get_locks_for_datasets(self, dataset_list):
```
bari12: Just food for thought:

I wonder if we should just name this `get_datasetlocks_for_dids`. The reason is, in general Rucio should be able to handle inputs as dids, no matter if they are files/datasets/containers (Files does not make sense in this case)
This case we could merge the `get_locks_for_datsets` and `get_container_locks` case. (I mean, you might even want to get dataset locks for multiple containers)

bari12 (commented):
Hi @ivmfnal 
I think this is good. I have only one comment (see below) which is more a conceptual, then implementation question and would mirror a bit the behaviour in other clients. Please have a look.

lib/rucio/client/lockclient.py
ivmfnal: Let me see if I understand this... Are you suggesting to have a method which will take a list of container or dataset DIDs and get all the locks for all the files found in the listed containers and datasets ?

lib/rucio/client/lockclient.py
```diff
@@ -61,6 +60,53 @@ def get_dataset_locks(self, scope, name):
...
+            #
+            # if a dataset appears in more than one container, then its locks may be duplicated here. we need to filter them
+            # make sure that for each dataset, combination (rule_id, rse_id) appears at most once
+            #
+            rule_rse = (lock["rule_id"], lock["rse_id"])
+            locks_by_dataset.setdefault(scope_name, {})[rule_rse] = lock
+        out = {scope_name: list(locks.values()) for scope_name, locks in locks_by_dataset.items()}
+        return out
+
+    def get_locks_for_dids(self, dids):
```
bari12: `get_dataset_locks_for_dids` instead?

lib/rucio/client/lockclient.py
```diff
@@ -61,6 +60,53 @@ def get_dataset_locks(self, scope, name):
...
+        """
+        Get list of locks for for all the files fund, recursively, in the listed datasets or containers.
```
bari12: fund -> found?

lib/rucio/web/rest/flaskapi/v1/locks.py
```diff
@@ -29,7 +29,10 @@
 from rucio.common.exception import RSENotFound
 from rucio.common.utils import render_json
 from rucio.web.rest.flaskapi.v1.common import check_accept_header_wrapper_flask, parse_scope_name, try_stream, \
-    request_auth_env, response_headers, generate_http_error_flask, ErrorHandlingMethodView
+    request_auth_env, response_headers, generate_http_error_flask, ErrorHandlingMethodView, json_parse
+from rucio.core.did import list_child_datasets, get_did
```
bari12: The REST layer should not directly interact with the core. We need to layer this
REST -> API -> CORE

lib/rucio/web/rest/flaskapi/v1/locks.py
```diff
@@ -76,18 +79,84 @@ def get(self, scope_name):
...
-            scope, name = parse_scope_name(scope_name, request.environ.get('vo'))
+            scope, name = parse_scope_name(scope_name, vo)
+            internal_scope = InternalScope(scope, vo=vo)
+
+            def generate(vo, did_type):
+                if did_type == "dataset":
+                    for lock in get_dataset_locks(scope, name, vo=vo):
+                        yield render_json(**lock) + '\n'
+                else:
+                    for dataset_info in list_child_datasets(internal_scope, name):
```
bari12: Ok I thought a little bit about this especially in regards to the recent discussion with DB operators.

We need to put this functionality to the core (See comment above). The problem, even if we just put this to the API and call the core like this from there is that we would create a new session for each call in these loops. This creates a lot of transactions on the DB. You need to move this functionality to the core where only one session is used.

Code should not change a lot, it just needs to be relocated.

lib/rucio/web/rest/flaskapi/v1/locks.py
```diff
@@ -76,18 +79,84 @@ def get(self, scope_name):
...
+                        except KeyError:
+                            raise ValueError("Unsupported or unknown DID type for %s:%s" % (scope, name))
+                    if did_type not in ("dataset", "container"):
+                        raise ValueError("Unsupported DID type '%s' for %s:%s" % (did_type, scope, name))
+                    if did_type == "dataset":
+                        for lock in get_dataset_locks(scope, name, vo=vo):
+                            lock["dataset_scope_name"] = f"{scope}:{name}"
+                            yield render_json(**lock) + '\n'
+                    else:
+                        for dataset_info in list_child_datasets(internal_scope, name):
```
bari12: Same here, see comment above.

bari12 (dismissed):
Hi @ivmfnal 
In principle this is correct, but the layering is wrong which might lead to some performance issues due to the number of transactions being opened. (DB Operators are already not happy about this) I added some suggestions to move this basically to the core. I think it should mostly be moving code around, the underlying code is correct.
Thanks

lib/rucio/core/lock.py
```diff
@@ -66,6 +69,41 @@ def get_dataset_locks(scope, name, session=None):
...
+    """
+    Get the dataset locks of a list of datasets or containers, recursively
+
+    :param dids:           List of dictionaries {"scope":scope(type:InternalScope), "name":name,
+                           "type":did type(DIDType.DATASET or DIDType.CONTAINER)}, "type" is optional
+    :param session:        The db session to use.
+    :return:               Generator of lock_info dicts, may contain duplicates
+    """
+
+    from rucio.core.did import list_child_datasets
```
bari12, line 83: At the top `rucio.core.did` is already imported. I think you can just address this as `rucio.core.did.list_child_datasets`

lib/rucio/core/lock.py
```diff
@@ -66,6 +69,41 @@ def get_dataset_locks(scope, name, session=None):
...
+
+    for did in dids:
+        scope = did["scope"]
+        assert isinstance(scope, InternalScope)
+        name = did["name"]
+        did_type = did.get("type")
+        if not did_type:
+            try:
+                did_info = rucio.core.did.get_did(scope, name)
```
bari12, line 92: You have to hand through the session here, otherwise it leaks.

bari12 (changes requested):
GTG, one small change, but otherwise this can be included.

PR #4859, opened 2021-09-21
Protocols: remove double / from prefixes in parse_pfns. Closes #4833

### LINKED ISSUE #4833: rucio-conveyor-finisher fails at a non-deterministic endpoint with root protocol
Motivation
----------
LCLS is using non-deterministic endpoints with root protocol. For root protocol, the endpoint prefix usually starts with two slashes (e.g. //psdm/rucio). However, rucio-conveyor-finisher runs into trouble because of this line:

https://github.com/rucio/rucio/blob/master/lib/rucio/rse/protocols/protocol.py#L444

At there, "path" is a sanitized one with '//' being replaced by a single '/', but prefix isn't. So the comparison will fail. This only happens for non-deterministic endpoints.

Modification
------------
In parse_pfns(), either not replacing '//' by single '/'. or do this replacement for both 'path' and 'prefix'.

rcarpa:

### DIFF
--- a/lib/rucio/rse/protocols/protocol.py
+++ b/lib/rucio/rse/protocols/protocol.py
@@ -450,12 +450,15 @@ class RSEProtocol(object):
             while '//' in parsed.path:
                 parsed = parsed._replace(path=parsed.path.replace('//', '/'))
             path = parsed.path
+            prefix = self.attributes['prefix']
+            while '//' in prefix:
+                prefix = prefix.replace('//', '/')
 
             # Protect against 'lazy' defined prefixes for RSEs in the repository
-            if not self.attributes['prefix'].startswith('/'):
-                self.attributes['prefix'] = '/' + self.attributes['prefix']
-            if not self.attributes['prefix'].endswith('/'):
-                self.attributes['prefix'] += '/'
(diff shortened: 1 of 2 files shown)


mlassnig (commented):
We need to guarantee that all PFNs look like this:

`protocol://hostname:port/path/to/file`

except (!!) for xrootd, here, the double slash at the beginning of the path is mandatory:

`root://hostname:port//path/to/file`

Is that ok with this patch?

rcarpa:
Multiple '/' are not preserved by this patch, but this function is responsible for parsing the pfn, not for generating one. Indeed, the patch can break things if some code uses this function to parse the pfn, then generates a new one without being careful about the `root` convention. IMHO, if this happens, we must fix that other incorrect code rather than having hacks everywhere for root.
However, if there is a push for it from your side,  I'm OK to add an additional protection here and return a prefix with double '/' for root in the parsed output.

rcarpa:
Just rebased, no changes. still waiting for an answer from Mario to see if I change the pr

mlassnig:
No, I agree. If a problem creeps up the incorrect code should be fixed.

mlassnig (approved):

PR #4862, opened 2021-09-22
Lifetime Model : Improvement for the submission ; Add client API ; Add tests : Closes #4843

### LINKED ISSUE #4843: Introduce option to enable/disable the submission of new exceptions
Motivation
----------
It sometimes happens that people submit exception to the lifetime model outside dedicated period. In that case, it should raise an exception. 

Modification
------------
Introduce a new option to enable/disable the submission of new exceptions.

cserf:
Lifetime Model :  Closes #4843

Improvement for the submission

Add client API

Add tests

### DIFF
--- a/lib/rucio/api/lifetime_exception.py
+++ b/lib/rucio/api/lifetime_exception.py
@@ -7,7 +7,7 @@
   http://www.apache.org/licenses/LICENSE-2.0
 
   Authors:
-  - Cedric Serfon, <cedric.serfon@cern.ch>, 2016-2017
+  - Cedric Serfon, <cedric.serfon@cern.ch>, 2016-2021
   - Andrew Lister, <andrew.lister@stfc.ac.uk>, 2019
 
   PY3K COMPATIBLE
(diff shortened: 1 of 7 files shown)


lib/rucio/api/lifetime_exception.py
```diff
@@ -50,9 +50,22 @@ def add_exception(dids, account, pattern, comments, expires_at, vo='def'):
     """
 
     account = InternalAccount(account, vo=vo)
-    for d in dids:
-        d['scope'] = InternalScope(d['scope'], vo=vo)
-    return lifetime_exception.add_exception(dids=dids, account=account, pattern=pattern, comments=comments, expires_at=expires_at)
+    for did in dids:
```
bari12, line 53: This should probably be moved, as a bulk, into the core. Otherwise we create a transaction per exception.

cserf: I don't get it. We already pass the list of dids to the core

bari12: You are right; I misread this. Please ignore 😄

bari12 (approved):

Issue #4882, opened 2021-10-04
LSST LFN2PFN algorithm


leo-ccin2p3:
Motivation
----------
The Rubin-LSST experiment is currently testing Rucio as file replica tool. The motivation for this LFN2PFN algo is to provide a specific management of LSST namespace in Rucio

Modification
------------
Adding a new LFN2PFN algo and declare it

PR #4883, opened 2021-10-04
core & internals: support LSST LFN2PFN algo #4882

leo-ccin2p3:

### DIFF
--- a/lib/rucio/rse/protocols/protocol.py
+++ b/lib/rucio/rse/protocols/protocol.py
@@ -226,9 +226,9 @@ class RSEDeterministicTranslation(object):
     def __lsst(scope, name, rse_id, rse_attrs, protocol_attrs):
         """
         LFN2PFN algorithm for LSST
-        
+
         Get any of the file replica and extract its relative path
-    
+
         :param scope: Scope of the LFN.
         :param name: File name of the LFN.
         :param rse_id: RSE for PFN (ignored)
@@ -242,7 +242,7 @@ class RSEDeterministicTranslation(object):
         try:
             from rucio.api import replica
             from rucio.api import rse
-            repl = list(replica.list_replicas(dids=[{'scope':scope,'name':name}]))[0]
+            repl = list(replica.list_replicas(dids=[{'scope': scope, 'name': name}]))[0]
             # get first key: a RSE
             RSE = list(repl['rses'].keys())[0]
             # The PFN on this RSE
--- a/lib/rucio/rse/protocols/protocol.py
+++ b/lib/rucio/rse/protocols/protocol.py
@@ -241,7 +241,7 @@ class RSEDeterministicTranslation(object):
         del rse
         del rse_attrs
         del protocol_attrs
-        return name.replace('__','/')
+        return name.replace('__', '/')
 
     @classmethod
     def _module_init_(cls):
--- a/lib/rucio/rse/protocols/protocol.py
+++ b/lib/rucio/rse/protocols/protocol.py
@@ -232,7 +232,7 @@ class RSEDeterministicTranslation(object):
 
         :param scope: Scope of the LFN (ignored)
         :param name: File name of the LFN.
-        :param rse_id: RSE for PFN (ignored)
+        :param rse: RSE for PFN (ignored)
         :param rse_attrs: RSE attributes for PFN (ignored)
         :param protocol_attrs: RSE protocol attributes for PFN (ignored)
         :returns: Path for use in the PFN generation.
--- a/lib/rucio/rse/protocols/protocol.py
+++ b/lib/rucio/rse/protocols/protocol.py
@@ -223,37 +223,25 @@ class RSEDeterministicTranslation(object):
         return '%s/%s/%s/%s' % (scope[0:7], scope[4:len(scope)], name.split('-')[0] + "-" + name.split('-')[1], name)
 
     @staticmethod
-    def __lsst(scope, name, rse_id, rse_attrs, protocol_attrs):
+    def __lsst(scope, name, rse, rse_attrs, protocol_attrs):
         """
-        LFN2PFN algorithm for LSST
+        LFN2PFN algorithm for Rubin-LSST in the ESCAPE project
 
-        Get any of the file replica and extract its relative path
+        Replace convention delimiter '__' by '/'
+        The Escape instance does use the 'generic' Rucio schema.
 
-        :param scope: Scope of the LFN.
+        :param scope: Scope of the LFN (ignored)
         :param name: File name of the LFN.
         :param rse_id: RSE for PFN (ignored)
         :param rse_attrs: RSE attributes for PFN (ignored)
         :param protocol_attrs: RSE protocol attributes for PFN (ignored)
         :returns: Path for use in the PFN generation.
         """
-        del rse_id
+        del scope
+        del rse
         del rse_attrs
         del protocol_attrs
-        try:
-            from rucio.api import replica
-            from rucio.api import rse
-            repl = list(replica.list_replicas(dids=[{'scope': scope, 'name': name}]))[0]
-            # get first key: a RSE
-            RSE = list(repl['rses'].keys())[0]
-            # The PFN on this RSE
-            orig_pfn = repl['rses'][RSE][0]
-            # The protocol of this RSE
-            rse_protocol = rse.get_rse(RSE)['protocols'][0]
-            # Extract the relative path
-            rse_relative_path = orig_pfn.split(rse_protocol['prefix'])[1]
-            return rse_relative_path
-        except IndexError:
-            return name
+        return name.replace('__','/')
 
     @classmethod
     def _module_init_(cls):
--- a/lib/rucio/rse/protocols/protocol.py
+++ b/lib/rucio/rse/protocols/protocol.py
@@ -30,6 +30,7 @@
 # - Luca Scotto Lavina, <scotto@lpnhe.in2p3.fr>, 2021
 # - Thomas Beermann, <thomas.beermann@cern.ch>, 2021
 # - James Perry <j.perry@epcc.ed.ac.uk>, 2021
+# - Lionel Schwarz <lionel.schwarz@in2p3.fr>, 2021
 #
 # PY3K COMPATIBLE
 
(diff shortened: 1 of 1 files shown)


lib/rucio/rse/protocols/protocol.py
```diff
@@ -221,6 +222,39 @@ def __xenon(scope, name, rse, rse_attrs, protocol_attrs):
...
+        :param rse_id: RSE for PFN (ignored)
+        :param rse_attrs: RSE attributes for PFN (ignored)
+        :param protocol_attrs: RSE protocol attributes for PFN (ignored)
+        :returns: Path for use in the PFN generation.
+        """
+        del rse_id
+        del rse_attrs
+        del protocol_attrs
+        try:
+            from rucio.api import replica
```
mlassnig: something is not right here. this function is potentially called thousands of times, so it shouldn't call list_replicas() or something similar. what exactly are you trying to achieve?

leo-ccin2p3: This code will be used for replication only. When replicating a file from a RSE to another, we want to make sure the file will keep the same relative path (this path will originally be specified when uploading on a non-deterministic RSE).
Example : the file **RSE-SOURCE/root_source/the/relative/path/to/the/file** will be replicated on RSE-DEST at **RSE-DEST/root_dest/the/relative/path/to/the/file**
So the algo on RSE-DEST needs to get a replica of the file to retrieve the relative path.
It seems input params on the LFN2PFN function are not sufficent to do this without calling replica.get_replicas().
If you think there is a simpler solution, please let me know

mlassnig: hmmm, isn't CMS doing it like this? @bari12 @ericvaandering mabey there's an easier way

ericvaandering: I don't 100% follow what you're doing, but CMS has two forms of doing this, one is a simplification of the other:

Form 1: PFN at site = RSE_SPECIFIC_PREFIX + '/' + LFN

Form 2: Pass LFN through a series of REGEX substitution processes to arrive at PFN (useful for changing the prefix of the PFN based on the LFN). 

In both cases, we actually discard (but do some checking) on the scope.

Both the RSE specific prefix and the regular expressions all come from RSE attributes (so we don't ignore those).

leo-ccin2p3: It's clear that we cannot do like CMS: we don't have enough information in the LFN itself to construct the PFN. If the algo can't call API methods like list_replicas(), we have a problem here.
Let me explain our use-case:
Say we have the DID **scope:lsst_a_262593_R43_S02_z.fits** that was uploaded in the first place on a RSE-SOURCE (non-deterministic) with the full path 
**datasets/desc/DC2/Run2.2i/sim/y1-wfd/00262593/lsst_a_262593_R43_S02_z.fits**. 
The PFN of the replica is then 
**RSE-SOURCE/root_source/datasets/desc/DC2/Run2.2i/sim/y1-wfd/00262593/lsst_a_262593_R43_S02_z.fits**
We want to replicate it on a RSE-DEST with the same relative path ie we need to build a PFN like **RSE-DEST/root_dest/datasets/desc/DC2/Run2.2i/sim/y1-wfd/00262593/lsst_a_262593_R43_S02_z.fits**

lib/rucio/rse/protocols/protocol.py
```diff
@@ -221,6 +222,39 @@ def __xenon(scope, name, rse, rse_attrs, protocol_attrs):
...
+    @staticmethod
+    def __lsst(scope, name, rse_id, rse_attrs, protocol_attrs):
+        """
+        LFN2PFN algorithm for LSST
+
+        Get any of the file replica and extract its relative path
+
+        :param scope: Scope of the LFN.
+        :param name: File name of the LFN.
+        :param rse_id: RSE for PFN (ignored)
```
bari12: rse, not rse_id.

lib/rucio/rse/protocols/protocol.py
bari12: This won't work like this and goes against the fundamental idea of the deterministic lfn2pfn algorithms. (We should probably put up a page to describe this better somewhere)
The deterministic lfn2pfn algorithms should be able to generate the pfn purely based on the scope and name of a file (+ information like hostname, prefixes etc.) - no other lookup must be necessary. So your `list_replicas` call here breaks this concept.

I think the usecase you are describing calls for a specific surl algorithm. https://github.com/rucio/rucio/blob/4ff378c848ccc88f3558500bf17004c79a02765e/lib/rucio/common/utils.py#L578
What the SURL algorithm does is to calculate the path for transfers to non-deterministic RSEs. Since it is non-deterministic, Rucio cannot use the lfn2pfn algorithm here, but because it is an async transfer, it also cannot wait for user input (Like you would do on a file registration on a non-deterministic RSE). Thus there are these surl algorithms which still can programatically calculate the pfn for non-deterministic RSEs, but in contrast to the lfn2pfn methods, they are allowed to use additional knowledge to do so.

mlassnig: Maybe I missing something, but why not make the DID look like this:

scope=`lsst` name=`/DC2/Run2.2i/sim/y1-wfd/00262593/lsst_a_262593_R43_S02_z.fits`

or

scope=`/DC2/Run2.2i/sim` name=`/y1-wfd/00262593/lsst_a_262593_R43_S02_z.fits`

? The whole point of this mapping is that you don't have to look it up in a database but that you can compute it offline.

leo-ccin2p3: > This won't work like this and goes against the fundamental idea of the deterministic lfn2pfn algorithms. (We should probably put up a page to describe this better somewhere) The deterministic lfn2pfn algorithms should be able to generate the pfn purely based on the scope and name of a file (+ information like hostname, prefixes etc.) - no other lookup must be necessary. So your `list_replicas` call here breaks this concept.
> 
OK I was not aware of this, I understand.

leo-ccin2p3: > Maybe I missing something, but why not make the DID look like this:
> 
> scope=`lsst` name=`/DC2/Run2.2i/sim/y1-wfd/00262593/lsst_a_262593_R43_S02_z.fits`
> 
> or
> 
> scope=`/DC2/Run2.2i/sim` name=`/y1-wfd/00262593/lsst_a_262593_R43_S02_z.fits`
> 
> ? The whole point of this mapping is that you don't have to look it up in a database but that you can compute it offline.

Good point. This means we have to build the appropriate DID.scope and DID.name when we first upload the file, but I guess it is not big issue.
I then have a try with this solution and using the 'identity' algo.
Thanks for your help and sorry for the newbie questions ;)

leo-ccin2p3: @mlassnig It seems the '/' is not allowed in the DID name (like in your proposal) by `rucio upload`:

```
$ rucio upload --rse IN2P3-CC-LSST-SOURCE --scope LSST_ESCAPE_SCHWARZ_test --name project.03 file
2021-10-05 17:07:29,960	INFO	Preparing upload for file .project
...
2021-10-05 17:07:32,448	INFO	Successfully uploaded file .project

$ rucio upload --rse IN2P3-CC-LSST-SOURCE --scope LSST_ESCAPE_SCHWARZ_test --name '/directory/project.03' file
2021-10-05 17:07:53,833	INFO	Preparing upload for file .project
2021-10-05 17:07:54,027	ERROR	An unknown exception occurred.
Details: no error information passed (http status code: 404)
```

any suggestions?

bari12: The schema of your Rucio instance must allow it, it's not allowed in the generic (standard) schema, and it seems not in the ESCAPE one either.

leo-ccin2p3: Digging a bit more in the Rucio API, I tried to register a file which already exists in the storage backend and give it a DID name with a '/' in it:

```
>>> from rucio.client.replicaclient import ReplicaClient
>>> rc = ReplicaClient()
>>> result = rc.add_replica(rse="IN2P3-CC-LSST-SOURCE",scope="LSST_ESCAPE_SCHWARZ_test",name="subdir/project",bytes=366,adler32="2b5b78fc")
  File "/usr/local/lib/python3.6/dist-packages/rucio/client/replicaclient.py", line 254, in add_replica
    return self.add_replicas(rse=rse, files=[dict])
  File "/usr/local/lib/python3.6/dist-packages/rucio/client/replicaclient.py", line 274, in add_replicas
    raise exc_cls(exc_msg)
rucio.common.exception.InvalidObject: Provided object does not match schema.
Details: Problem validating dids : 'subdir/project' does not match '^[A-Za-z0-9][A-Za-z0-9\\.\\-\\_]{1,250}$'

Failed validating 'pattern' in schema['items']['properties']['name']:
    {'description': 'Data Identifier name',
     'pattern': '^[A-Za-z0-9][A-Za-z0-9\\.\\-\\_]{1,250}$',
     'type': 'string'}

On instance[0]['name']:
    'subdir/project'
```

It seems the DID.name is limited to 250 chars and cannot contains chars other than alphanumeric and . - _

leo-ccin2p3: > The schema of your Rucio instance must allow it, it's not allowed in the generic (standard) schema, and it seems not in the ESCAPE one either.

good to know this can be customized. just by curiosity, how easy is it to change this. Is this documented in https://rucio.readthedocs.io/en/latest ?

mlassnig: looks like the default schema does not have it enabled. you should be able to use the schema from CMS

ericvaandering: Lines 70 to 84 here: https://github.com/rucio/rucio/blob/631e5d5daa81ab86f310fd046350e8212b91e1eb/lib/rucio/common/schema/cms.py#L70

leo-ccin2p3: If I understand correctly, 1 Rucio instance uses 1 particular schema. And in my case, I guess the ESCAPE instance uses the generic schema.

leo-ccin2p3: I just commited a version of the LSST algo which is compatible with the generic schema. Should fit with our tests on ESCAPE Rucio instance. Thanks

mlassnig (approved):

bari12:
👍 Please squash the commits next time (I will do it now)

Issue #4886, opened 2021-10-05
Dirac interface: make an extract_scope default algorithm variable


martynia:
Motivation
----------

This is an issue, where the `extract_scope` function has to use the same algorithm both on the client and the server side. This normally needs a config file on the client side to tell the function which algorithm to use. In Dirac, we would like to eliminate the config file on the client side, some attempts with managing exceptions have been already made.

The proposed solution below eliminates a need of a config file on a client and allows to change the default algorithm from atlas to something else. The new algorithm has to be present in Rucio utils and referenced in a server config file.

Modification
------------

The proposed solution is to change the signature of the function and modify one line within an `if` statement:

https://github.com/rucio/rucio/blob/41e1e966712a9209d56931dbf0ce61a3f9cb437e/lib/rucio/common/utils.py#L748

to read:

```python
def extract_scope(did, scopes=None, default_extract = _DEFAULT_EXTRACT):
    extract_scope_convention = config_get('common', 'extract_scope', False, None)
    if extract_scope_convention is None or extract_scope_convention not in _EXTRACT_SCOPE_ALGORITHMS:
        extract_scope_convention = default_extract
    return _EXTRACT_SCOPE_ALGORITHMS[extract_scope_convention](did=did, scopes=scopes)
```

All existing calls to `extract_scope` are unaffected.

PR #4889, opened 2021-10-05
Allow setting up default scope extraction algorithms programatically. Addresses #4886

martynia:
This change addresses the issue where a config file is needed on the client side. By allowing setting a different default algorithm programmatically we eliminate the need of a config file on the client side.


jamesp-epcc (approved):
This looks fine to me. I am working on this part of the code to add support for scope extraction algorithms in policy packages, but this change doesn't interfere with what I'm doing.

bari12:
Will be merged in next feature (1.27)

PR #4895, opened 2021-10-06
Messaging: Stop SSL reconnects

### LINKED ISSUE #2606: STOMP connections are not always cleanly closed
The STOMP specification (https://stomp.github.io/stomp-specification-1.2.html#DISCONNECT) indicates that connections should be closed using the `DISCONNECT` frame and a receipt.

It seems that there are 4 places where STOMP connections are created but not closed explicitly using `stomp.py`'s `disconnect()`: `lib/rucio/core/trace.py`, `lib/rucio/core/nongrid_trace.py` and `lib/rucio/daemons/hermes/hermes.py`.

FWIW, `disconnect()` is used elsewhere: `bin/rucio-cache-client`, `lib/rucio/daemons/cache/consumer.py`, `lib/rucio/daemons/conveyor/receiver.py` and `lib/rucio/daemons/tracer/kronos.py`.

ericvaandering:
This PR fixes the issue in #2606 which only affects CMS because we use SSL connections. It seems that providing the "timeout" parameter to stomp.py is actually signaling that the connection should ONLY last the length of the timeout. So no value is appropriate. 

The checks on connection status should be enough to get the connection recreated when it does drop

### DIFF
--- a/lib/rucio/daemons/hermes/hermes.py
+++ b/lib/rucio/daemons/hermes/hermes.py
@@ -218,7 +218,7 @@ def deliver_messages(once=False, brokers_resolved=None, thread=0, bulk=1000, del
                                      ssl_cert_file=config_get('messaging-hermes', 'ssl_cert_file'),
                                      vhost=vhost,
                                      keepalive=True,
-                                     timeout=broker_timeout)
+                                     )
 
         con.set_listener('rucio-hermes',
                          HermesListener(con.transport._Transport__host_and_ports[0]))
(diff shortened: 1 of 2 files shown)


bari12:
I am quite surprised that they handle timeouts like that. Is this confirmed? (Couldn't find much in their documentation)

ericvaandering:
Me too, but it's observationally what I see (and every example I see takes the default None value). I see it with non-SSL connections as well in our Kronos log where we are reconnecting every few minutes. Take a look at the ATLAS logs and see if you see the same.

bari12:
Ok, let's merge it like this and see 👍

bari12 (approved):

PR #4898, opened 2021-10-07
Reaper : Non optimal use of the limit parameter in list_and_mark_unlocked_replicas : Closes #4897

### LINKED ISSUE #4897: Non optimal use of the limit parameter in list_and_mark_unlocked_replicas
Motivation
----------
Non optimal use of the limit parameter in list_and_mark_unlocked_replicas

cserf:
Reaper : Non optimal use of the limit parameter in list_and_mark_unlocked_replicas : Closes #4897 

The purpose of this PR is to change the way the limit parameter is used by the method `list_and_mark_unlocked_replicas`
Currently the parameter is not used directly in the query but in the loop on the result of the query. If the limit parameter is smaller than the size of the chunk returned by the DB, some extra rows will be returned that won't be updated later. The change just move the parameter in the query itself, but for Oracle the limit + with_for_update do not work well together, therefore the introduction of the subquery.

### DIFF
--- a/lib/rucio/core/replica.py
+++ b/lib/rucio/core/replica.py
@@ -60,7 +60,7 @@ import requests
 from dogpile.cache import make_region
 from dogpile.cache.api import NO_VALUE
 from six import string_types
-from sqlalchemy import func, and_, or_, exists, not_, update, delete
+from sqlalchemy import func, and_, or_, tuple_, exists, not_, update, delete
 from sqlalchemy.exc import DatabaseError, IntegrityError
 from sqlalchemy.orm import aliased
 from sqlalchemy.orm.exc import FlushError, NoResultFound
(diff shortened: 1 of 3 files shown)


rcarpa:
Is it possible to have more context about the issue? What are we trying to solve; why; etc. Ideally, both in the commit message and the initial ticket ( #4897 ).

cserf:
I put some explanation. Let me know if it's clear

rcarpa:
> I put some explanation. Let me know if it's clear

perfect. Thank you

lib/rucio/core/replica.py
```diff
@@ -1985,24 +1985,53 @@ def list_and_mark_unlocked_replicas(limit, bytes_=None, rse_id=None, delay_secon
...
-        filter(case([(models.RSEFileAssociation.tombstone != none_value, models.RSEFileAssociation.rse_id), ]) == rse_id).\
-        filter(or_(models.RSEFileAssociation.state.in_((ReplicaState.AVAILABLE, ReplicaState.UNAVAILABLE, ReplicaState.BAD)),
-                   and_(models.RSEFileAssociation.state == ReplicaState.BEING_DELETED, models.RSEFileAssociation.updated_at < datetime.utcnow() - timedelta(seconds=delay_seconds)))).\
-        filter(~exists(select([1]).prefix_with("/*+ INDEX(SOURCES SOURCES_SC_NM_DST_IDX) */", dialect='oracle')
-                       .where(and_(models.RSEFileAssociation.scope == models.Source.scope,
-                                   models.RSEFileAssociation.name == models.Source.name,
-                                   models.RSEFileAssociation.rse_id == models.Source.rse_id)))).\
-        with_for_update(skip_locked=True).\
-        order_by(models.RSEFileAssociation.tombstone)
+                          models.RSEFileAssociation.rse_id).in_(sub_query)).\
```
bari12, line 2008: I am a bit worried if this scales well. The query is certainly more complicated; I don't have a better idea either though to get the limit in.
On a side-note: it would be good to move this to the new query syntax as well.

I think we should try to deploy this manually in INT/PROD and test a bit before merging upstream.

bari12:
Just to move forward on this:
1. Can we test it via a manual patch in production first? I am really not sure if it scales, it would be good to test it first before we expose it to the entire community.
2. It will need a rebase as well before the merge, bust see (1)

rcarpa:
How does the following manifest ? 
>  limit + with_for_update do not work well together

I see that a similar issue is marked as fixed https://github.com/sqlalchemy/sqlalchemy/issues/3741 in sqlalchemy, but didn't dig into it. So that sqlalchemy issue is not really fixed?

PR #4899, opened 2021-10-07
Report deletion events when the destination does not exist : Closes #4258, #4867

### LINKED ISSUE #4258: Report deletion events when the destination does not exist
Motivation
----------

Currently, when the Reaper fails to delete a file because the destination does not exist, no event is reported. This can be potentially useful information.

Modification
------------

Report a `deletion-notfound` event.

### LINKED ISSUE #4867: Deletion duration is not reported for unsuccesful attempts
Motivation
----------

Deletion events do not contain information on the duration of the operation when unsuccessful. This hinders identifying slow sites which can have a detrimental impact the the overall Reaper’s performance.

Modification
------------

Calculate and store the duration of unsuccessful deletions (depends on #4258). Report the duration in the log messages, like in the successful deletions. Take the opportunity to change the format and round to a single number past the decimal point.

cserf:
This PR addresses the 2 following related issues :
Report deletion events when the destination does not exist : Closes #4258
Deletion duration is not reported for unsuccesful attempts :  Closes #4867

### DIFF
--- a/lib/rucio/daemons/reaper/reaper.py
+++ b/lib/rucio/daemons/reaper/reaper.py
@@ -197,16 +197,22 @@ def delete_from_storage(replicas, prot, rse_info, staging_areas, auto_exclude_th
 
                 deletion_dict['duration'] = duration
                 add_message('deletion-done', deletion_dict)
-                logger(logging.INFO, 'Deletion SUCCESS of %s:%s as %s on %s in %s seconds', replica['scope'], replica['name'], replica['pfn'], rse_name, duration)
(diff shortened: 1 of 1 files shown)


dchristidis (approved):
Tried it on one Reaper instance; didn’t notice any issues. Thanks Cédric!

Someone should confirm that the event name `deletion-not-found` is appropriate.

bari12:
Yes, I am wondering if `deletion-not-found` will be correctly processed in the monitoring stack?
In any way, I think this is correct, we just should make sure that it is processed.

PR #4900, opened 2021-10-07
Client: Add activity and source-replica-expression args to move-rules Close #4855

### LINKED ISSUE #4855: request two extensions to move-rule
Motivation
----------
when draining an RSE, move-rule is used to move the data around. The add-rule command has several options that would be very helpful.  --activity to Tag the activity, --source-replica-expression to direct the input RSE for the transfer.

Modification
------------

add to extensions to move-rule command   --activity and --source-replica-expression

joeldierkes:
`move-rules` now can specify `--activity` and `--source-replica-expression`. The
respective new 'moved' rule will have these attributes set to these values.


rcarpa (commented):
LGTM,

mlassnig (approved):

PR #4902, opened 2021-10-12
Core: Add dataset lock exists after container test #1860

### LINKED ISSUE #1860: Datasetlocks not properly deleted when detaching datasets from containers
Motivation
----------
For rules on containers, when datasets are detached from the container, the respective DATASET_LOCK does not seem to be removed properly. There were some cases found where this was the case. Needs to be investigated and fixed.

Modification
------------
TBD

joeldierkes:
The entry for a dataset in `DATASET_LOCKS` is still set after the dataset got
detached from its parent. This should not be the case, since the dataset is not
linked anymore it shouldn't have a respective entry.

This bug does exist in production, however I was not able to reproduce it.


bari12:
Rebase please

bari12 (approved):

PR #4903, opened 2021-10-12
Patch documentation

joeldierkes:
## Documentation: Return a non-zero exit code for modified header

If a header in a python file is modified, the `tools/add_header` script with the
`dry-run` argument will now return a non-zero exit code. This means that
`tools/add_header` can be used as a pre-commit check or inside the CI/CD.


bziemons:
Hi! My understanding of dry runs is that they succeed (exit code 0), if everything would run fine. Instead of this change, I would suggest adding a `--check` option that exits with the proposed behavior

joeldierkes:
Good idea!

mlassnig (approved):

bziemons (approved):

Issue #4906, opened 2021-10-13
improve disk vs tape source prioritization


rcarpa:
Motivation
----------
a source which fails a transfer is de-prioritized rapidly. The goal is to avoid trying idefinetely a source which has issues, and to try other sources. Assuming 2 sources : a tape; and a disk one. The disk source will be tried first, if it fails, at next iteration we'll try the second (tape) source. This behavior is not perfect. A transitory issue on the disk source will result in mounting the tape.

Modification
------------
As a first step, just  ensure that the disk source is retried more than once before going to tape. 
A second improvement would be to  detect that type of disk source error. If it's a recoverable error, retry disk; If it's a non-recoverable error, useless to retry that source.

PR #4907, opened 2021-10-13
Transfers: reduce priority of tape sources. #4906

rcarpa:
Avoid going to tape sources immediately after failing a transfer from
disk. This is to avoid mounting the tape following a transitory issue
with a disk source.

### DIFF
--- a/lib/rucio/core/transfer.py
+++ b/lib/rucio/core/transfer.py
@@ -1227,12 +1227,15 @@ def __filter_unwanted_paths(candidate_paths: "Iterable[List[DirectTransferDefini
 def __sort_paths(candidate_paths: "Iterable[List[DirectTransferDefinition]]") -> "Generator[List[DirectTransferDefinition]]":
 
     def __transfer_order_key(transfer_path):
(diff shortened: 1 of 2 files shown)


lib/rucio/core/transfer.py
```diff
@@ -1227,12 +1227,15 @@ def __filter_unwanted_paths(candidate_paths: "Iterable[List[DirectTransferDefini
 def __sort_paths(candidate_paths: "Iterable[List[DirectTransferDefinition]]") -> "Generator[List[DirectTransferDefinition]]":
 
     def __transfer_order_key(transfer_path):
+        # Reduce the priority of the tape sources. If there are any disk sources,
+        # they must fail twice (1 penalty + 1 disk preferred over tape) before a tape will even be tried
+        source_ranking_penalty = 1 if transfer_path[0].src.rse.is_tape_or_staging_required() else 0
```
mlassnig, line 1232: it seems correct to me, but maybe it's better to have a separate function for this? just wondering

bari12 (approved):

PR #4911, opened 2021-10-18
Transfers: handle existing destinations and corrupted file overwrite. Closes #4447

### LINKED ISSUE #4447: Writing to tape: Destination file exists and overwrite is not enable
Motivation
----------

https://ggus.eu/index.php?mode=ticket_info&ticket_id=151016

> This pattern shows us a problem in the Rucio / FTS interaction when writing to TAPE.
> If FTS encounters a problem that would lead to data being successfully copied to tape, but the transfer marked as FAILED, Rucio will keep retrying it.
> 
> Rucio doesn't use overwrite when writing to TAPE (for good reason).
> FTS won't be able to copy the file again because it exists already (expected behavior).
> 
> But this leaves us in a bad spot where the file will be retried continuously.
> 
> FTS doesn't have "file-reuse" functionality. 
> E.g.: when attempting a transfer, check if the file is at the destination already.

rcarpa:


bari12:
Please rebase as well

bari12 (approved):

PR #4919, opened 2021-10-19
Protect the Evaluator from the Necromancer : Closes #4892

### LINKED ISSUE #4892: Protect the Evaulator from the Necromancer
Motivation
----------

The declaration of lost files is an important, but secondary workflow and should not affect deletions, which is a primary operation. The Reaper will stop if there’s a significant Evaluator backlog as a safety measure. The Necromancer directly contributes to this backlog, especially when there’s a large amount of files belonging to different large datasets.

Modification
------------

The Necromancer should pause if the Evaluator backlog passes some threshold. Note that it might be prudent to not rely solely on the number of updated DIDs, but also their size.

cserf:
Protect the Evaluator from the Necromancer : Closes #4892

Apply the same protection than for the reaper based on number of rows in updated_did table

### DIFF
--- a/lib/rucio/daemons/badreplicas/necromancer.py
+++ b/lib/rucio/daemons/badreplicas/necromancer.py
@@ -75,13 +75,12 @@ def necromancer(thread=0, bulk=5, once=False, sleep_time=60):
         logger = formatted_logger(logging.log, prefix + '%s')
 
         # Check if there is a Judge Evaluator backlog
-        # Currently use the same configuration as the reaper
         try:
-            max_evaluator_backlog_count = get('reaper', 'max_evaluator_backlog_count')
+            max_evaluator_backlog_count = get('necromancer', 'max_evaluator_backlog_count')
         except ConfigNotFound:
             max_evaluator_backlog_count = None
         try:
-            max_evaluator_backlog_duration = get('reaper', 'max_evaluator_backlog_duration')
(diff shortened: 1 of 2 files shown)


cserf:
First try with the same protection applied on the reaper. If it's not enough, will implement smarter throttling based on number of files in DIDs to reevaluate.

lib/rucio/daemons/badreplicas/necromancer.py
```diff
@@ -66,10 +68,40 @@ def necromancer(thread=0, bulk=5, once=False, sleep_time=60):
...
+        try:
+            max_evaluator_backlog_count = get('necromancer', 'max_evaluator_backlog_count')
+        except ConfigNotFound:
+            max_evaluator_backlog_count = None
+        try:
+            max_evaluator_backlog_duration = get('necromancer', 'max_evaluator_backlog_duration')
+        except ConfigNotFound:
+            max_evaluator_backlog_duration = None
+        if max_evaluator_backlog_count or max_evaluator_backlog_duration:
+            backlog = get_evaluation_backlog()
```
bari12: I wonder if we need to use different caching here (The duration that method caches might be too long for the necro)

cserf: Will add a `expiration_time` variable in `get_evaluation_backlog`

bari12 (approved):

PR #4921, opened 2021-10-19
Feature 4784 dump http parameters of the client request  combined with issue 4920

### LINKED ISSUE #4920: pass logger down to the Client
Motivation
----------
Currently, CLI option `--verbose` is used and visible by CLI layer only, but BaseClient uses another instance of logger, owned by the client.baseclient module. This prevents CLI option `--verbose` to control debug output produced by BaseClient.

Modification
------------
1. Add optional argument to the constructor of BaseClient and its subclasses to be used to pass a logger object created by the client side application 
2. Modify `rucio` CLI script to pass its logger to the Client object

ivmfnal:

### DIFF
--- a/lib/rucio/client/configclient.py
+++ b/lib/rucio/client/configclient.py
@@ -44,9 +44,6 @@ class ConfigClient(BaseClient):
 
     CONFIG_BASEURL = 'config'
 
-    def __init__(self, rucio_host=None, auth_host=None, account=None, ca_cert=None, auth_type=None, creds=None, timeout=600, user_agent='rucio-clients', vo=None):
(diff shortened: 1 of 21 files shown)


lib/rucio/client/configclient.py
```diff
@@ -44,9 +44,6 @@ class ConfigClient(BaseClient):
 
     CONFIG_BASEURL = 'config'
```
mlassnig, line 46: we have some places in the code where we instantiate the different clients directly, e.g., `c=client.AccountClient()`, not via the generic `c=client.Client()`. does this still work when the `__init__` is removed?

bari12: I think it should. The `__init__` was just to call the super() from `BaseClient` anyway, but since the class is derived from that, the `__init__` is taken from there as well. This looks ok.

bari12 (approved):

PR #4922, opened 2021-10-20
Reaper : Add an option to archive DID in the reaper workflow : Closes #4852

### LINKED ISSUE #4852: Add an option to archive DID in the reaper workflow
Motivation
----------
Currently deleted DIDs are not automatically archived, in particular for Oracle :
```
        if session.bind.dialect.name != 'oracle':
            rucio.core.did.insert_deleted_dids(chunk, session=session)
```

Modification
------------
Add an option to archive DID in the reaper workflow

cserf:
Reaper : Add an option to archive DID in the reaper workflow : Closes #4852

### DIFF
--- a/lib/rucio/core/did.py
+++ b/lib/rucio/core/did.py
@@ -612,7 +612,7 @@ def delete_dids(dids, account, expire_rules=False, session=None, logger=logging.
     metadata_to_delete = []
     file_content_clause = []
 
-    archive_dids = config_core.get('undertaker', 'archive_dids', default=False, session=session)
+    archive_dids = config_core.get('deletion', 'archive_dids', default=False, session=session)
 
     for did in dids:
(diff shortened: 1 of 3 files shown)


lib/rucio/core/did.py
```diff
@@ -637,7 +637,8 @@ def delete_dids(dids, account, expire_rules=False, session=None, logger=logging.
             not_purge_replicas.append((did['scope'], did['name']))
 
             # Archive content
-            # Disable for postgres
+        archive_content = config_core.get('deletion', 'archive_content', default=False, session=session)
+        if archive_content:
```
bari12, line 641: Aren't we disabling the archive content per default in this case? Is this what we want?

cserf: Well, if you check the code, that was already the case (bug)

bari12 (approved):

PR #4929, opened 2021-10-21
Core: Add update_rule boost option #4634

### LINKED ISSUE #4634: Add the ability to quicken the transition of STUCK rules to REPLICATING
Motivation
----------

Following a `rucio update-rule --stuck --cancel-requests __RULE__`, there might be a considerable delay (up to 8 hours; see [here](https://github.com/rucio/rucio/blob/98c87d1f943d92a8e3a121e772178e79aa9622c7/lib/rucio/core/rule_grouping.py#L777)) until the requests are recreated. Currently, one needs to manipulate the `updated_at` timestamps of the `STUCK` locks and the replication rule itself.

Modification
------------

Consider adding a new option in the client and the necessary implementation in the various layers to quicken the transition from `STUCK` to `REPLICATING`. This operation should be reserved to accounts with elevated privileges.

joeldierkes:
The timeout between the retries of the transitioning of a rule from `STUCK` to
`REPLICATING` are quite big. This commit introduces a feature `--boost-rule` to
allow almost instant transition between the states.

### DIFF
--- a/bin/rucio
+++ b/bin/rucio
@@ -47,7 +47,10 @@
 # - Rakshita Varadarajan <rakshitajps@gmail.com>, 2021
 # - Christoph Ames <christoph.ames@physik.uni-muenchen.de>, 2021
 # - James Perry <j.perry@epcc.ed.ac.uk>, 2021
+# - Petr Vokac <petr.vokac@fjfi.cvut.cz>, 2021
 # - David Población Criado <david.poblacion.criado@cern.ch>, 2021
+# - zoepap05 <90753392+zoepap05@users.noreply.github.com>, 2021
+# - KosKyr <90753277+KosKyr@users.noreply.github.com>, 2021
 # - Joel Dierkes <joel.dierkes@cern.ch>, 2021
 
 from __future__ import print_function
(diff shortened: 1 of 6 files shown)


bin/rucio
```diff
@@ -2372,6 +2377,7 @@ You can filter by account::
     update_rule_parser.add_argument('--cancel-requests', dest='cancel_requests', action='store_true', help='Cancel requests when setting rules to stuck.')
     update_rule_parser.add_argument('--priority', dest='priority', action='store', help='Priority of the requests of the rule.')
     update_rule_parser.add_argument('--child-rule-id', dest='child_rule_id', action='store', help='Child rule id of the rule.')
+    update_rule_parser.add_argument('--boost-rule', dest='boost_rule', action='store_true', help='')
```
mlassnig: help text :-)

lib/rucio/core/permission/atlas.py
```diff
@@ -525,11 +528,8 @@ def perm_update_rule(issuer, kwargs):
         return True
 
     # Only admin accounts can change account, state, priority of a rule
```
mlassnig: adapt comment :-)

mlassnig (approved):

Issue #3957, opened 2020-08-25
Use X.509 credentials specified in rucio.cfg with gfal operations


mkszuba:
Motivation
----------

(Note: for simplicity I shall discuss this for the _auth_type = x509_proxy_ scenario, that said the same applies to 'x509')

It would IMHO be very useful if the Rucio client set the X.509 credentials of gfal operations to those Rucio itself uses. As things stand now, 'rucio download' and 'rucio upload' end up looking for the proxy certificate in _/tmp/x509up_u${UID}_ regardless of what the value of _client_x509_proxy_ in rucio.cfg is - and unless gfal2 or its Python bindings feature some magic environment variables for this purpose, there doesn't seem to currently be a way of changing this behaviour. This can lead to somewhat confusing behaviour when interactions with the Rucio server itself work fine but things break as soon as a file transfer is to be executed.

Modification
------------

At a quick glance at the code, it seems this would be a simple matter of extending _rucio.rse.protocols.gfal.connect()_ to set additional option strings on the newly created gfal2 context:
 * for 'x509_proxy', X509/CERT
 * for 'x509', X509/CERT and X509/KEY

similarly to how it is already done with bearer tokens. I haven't tested this yet, though.

mlassnig:
Hi @mkszuba 

yes, you are absolutely right. By default gfal uses the X509_USER_PROXY, or alternatively the X509_USER_CERT & X509_USER_KEY environment variables. My suggestion would be to check if these variables exist, and if not use the values from the rucio.cfg

Would that be an ok approach?

mkszuba:
Makes sense to me - that way these variables will not suddenly stop working for people who do use them. Moreover, from what I can see in the code rucio-clients already support X509_USER_PROXY - albeit only for certain protocols (it would probably make sense to perform the relevant environment lookups only once, e.g. while populating _self.creds_ in baseclient).

bari12:
Needs to be added here https://github.com/rucio/rucio/blob/a106bb113b00803dd5b2f37a008fe211c75a313f/lib/rucio/rse/protocols/gfal.py#L189

PR #4932, opened 2021-10-22
Protocols: allow usage of rucio.cfg x509 proxy for gfal operations #3957

rcarpa:
It is only used if the gfal context wasn't already configured to
use other credentials. This is done to avoid breaking existing
configurations.


mlassnig (approved):
it looks correct to me

PR #4940, opened 2021-10-25
Client: Add the option to disable FilesDownload Exceptions Fix #4924

### LINKED ISSUE #4924: Add an option not to raise NoFilesDownloaded or NotAllFilesDownloaded in downloadclient
Motivation
----------
If there's a problem in the download client and some files cannot be downloaded, an exception `NoFilesDownloaded` or `NotAllFilesDownloaded` is raised. Because of this it is not possible to know which files were successfully downloaded. It should be possible to specify an option in all the download methods not to raise an exception (similar to what is done here : https://github.com/rucio/rucio/blob/master/lib/rucio/common/config.py#L56) and just to return the output dictionary with a `FAILED` status for those files.

joeldierkes:
If no or not all files are downloaded the `DownloadClient` raises an
exception. It is not possible to see which files are not downloaded or which
are. This commit introduces the opportunity to toggle the Exception.

### DIFF
--- a/bin/rucio
+++ b/bin/rucio
@@ -52,6 +52,7 @@
 # - zoepap05 <90753392+zoepap05@users.noreply.github.com>, 2021
 # - KosKyr <90753277+KosKyr@users.noreply.github.com>, 2021
 # - Joel Dierkes <joel.dierkes@cern.ch>, 2021
+# - jdierkes <joel.dierkes@cern.ch>, 2021
 
 from __future__ import print_function
 
@@ -1081,6 +1082,7 @@ def download(args):
         trace_pattern['taskid'] = args.trace_taskid
     if args.trace_usrdn:
         trace_pattern['usrdn'] = args.trace_usrdn
+    deactivate_file_download_exceptions = args.deactivate_file_download_exceptions if args.deactivate_file_download_exceptions is not None else False
 
     client = get_client(args)
     from rucio.client.downloadclient import DownloadClient
@@ -1134,11 +1136,11 @@ def download(args):
             items.append(item_defaults)
 
         if args.aria:
-            result = download_client.download_aria2c(items, trace_pattern)
+            result = download_client.download_aria2c(items, trace_pattern, deactivate_file_download_exceptions=deactivate_file_download_exceptions)
         elif args.metalink_file:
-            result = download_client.download_from_metalink_file(items[0], args.metalink_file)
+            result = download_client.download_from_metalink_file(items[0], args.metalink_file, deactivate_file_download_exceptions=deactivate_file_download_exceptions)
         else:
-            result = download_client.download_dids(items, args.ndownloader, trace_pattern)
+            result = download_client.download_dids(items, args.ndownloader, trace_pattern, deactivate_file_download_exceptions=deactivate_file_download_exceptions)
     else:
         if args.aria:
             logger.warning('Ignoring --aria option because --pfn option given')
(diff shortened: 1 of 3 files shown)


lib/rucio/tests/test_download.py
```diff
@@ -512,6 +513,14 @@ def test_trace_copy_out_and_checksum_validation(vo, rse_factory, did_factory, do
         assert len(traces) == 1 and traces[0]['clientState'] == 'DONE'
 
 
+def test_disable_no_files_download_error(vo, rse_factory, did_factory, download_client):
+    rse, _ = rse_factory.make_posix_rse()
+    with TemporaryDirectory() as tmp_dir:
+        res = download_client.download_dids([{'did': 'some:randomNonExistingDid', 'base_dir': tmp_dir}], raise_exception=False)
+        print(res)
```
rcarpa: Is this print here on purpose?

bin/rucio
```diff
@@ -1033,6 +1037,7 @@ def download(args):
         trace_pattern['taskid'] = args.trace_taskid
     if args.trace_usrdn:
         trace_pattern['usrdn'] = args.trace_usrdn
+    raise_exception = args.dont_raise_exception if args.dont_raise_exception is not None else True
```
rcarpa: raise_exception is much to generic for what it does; The name makes you think that all exceptions will be ignored; but it only acts on two particular exceptions; there are many others which can still be raised under current code. I vote for either making the option name more specific, or ensuring that all exceptions are handled silently.

lib/rucio/client/downloadclient.py
```diff
@@ -256,9 +259,9 @@ def download_pfns(self, items, num_threads=2, trace_custom_fields={}, traces_cop
         if num_files_in != num_files_out:
             raise RucioException('%d items were in the input queue but only %d are in the output queue' % (num_files_in, num_files_out))
```
rcarpa, line 260: this exception will be raised before reaching the check_output

bin/rucio
mlassnig: also, never use negative arguments :-)

--raise-exception, default=False

instead of

--dont-raise-exception, default=True

bin/rucio
```diff
@@ -2244,6 +2249,7 @@ You can filter by key/value, e.g.::
         selected_parser.add_argument('--filter', dest='filter', action='store', help='Filter files by key-value pairs like guid=2e2232aafac8324db452070304f8d745.')
         selected_parser.add_argument('--scope', dest='scope', action='store', help='Scope if you are using the filter option and no full DID.')
         selected_parser.add_argument('--metalink', dest='metalink_file', action='store', help='Path to a metalink file.')
+        selected_parser.add_argument('--dont-raise_exception', dest='dont_raise_exception', action='store_true', help='Does not raise NoFilesDownloaded or NotAllFilesDownloaded Exception.')
```
mlassnig: see above. also, no underscores ;-)

bin/rucio
joeldierkes: Do we want to disable the exceptions by default?

mlassnig: not sure.

for me, the important bit is to have the --argument not phrased inverse ;-)

joeldierkes: From my pov it should be enabled by default. I would suggest a `--deactivate-exceptions` flag.

lib/rucio/tests/test_download.py
joeldierkes: Yep, it's helpful for debugging. I added more context information to the output.

bari12 (approved):

PR #4941, opened 2021-10-25
Rules: Pause evaluating DIDs in case of Unique Constraint hit Fix #4939

### LINKED ISSUE #4939: Pause evaluating DIDs in case of Unique Constraint hit
Motivation
----------
The judge-evaluator is pausing the evaluation of dids (for a certain) time if certain error conditions are hit.
https://github.com/rucio/rucio/blob/2555f0a967bfa52c7486441cc704609800726151/lib/rucio/daemons/judge/evaluator.py#L129

Modification
------------
Also add the condition `cx_Oracle.IntegrityError: ORA-00001: unique constraint (ATLAS_RUCIO.REPLICAS_PK) violated` to this.

joeldierkes:
The judge-evaluator is pausing the evaluation of dids (for a certain) time if
certain error conditions are hit. A new case is introduced.

### DIFF
--- a/lib/rucio/daemons/judge/evaluator.py
+++ b/lib/rucio/daemons/judge/evaluator.py
@@ -25,6 +25,7 @@
 # - Benedikt Ziemons <benedikt.ziemons@cern.ch>, 2020
 # - David Población Criado <david.poblacion.criado@cern.ch>, 2021
 # - Radu Carpa <radu.carpa@cern.ch>, 2021
+# - Joel Dierkes <joel.dierkes@cern.ch>, 2021
 
 """
 Judge-Evaluator is a daemon to re-evaluate and execute replication rules.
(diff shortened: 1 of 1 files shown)


lib/rucio/daemons/judge/evaluator.py
```diff
@@ -126,7 +127,7 @@ def re_evaluator(once=False, sleep_time=30, did_limit=100):
                     except DataIdentifierNotFound:
                         delete_updated_did(id_=did.id)
                     except (DatabaseException, DatabaseError) as e:
-                        if match('.*ORA-00054.*', str(e.args[0])):
+                        if match('.*ORA-000(01|54).*', str(e.args[0])):
```
rcarpa, line 130: I remember this regexp match being in many places in many daemons. Is this place the only one which we want to modify ?

joeldierkes: The issue specifically points to the judge evaluator.
> The judge-evaluator is pausing the evaluation of dids (for a certain) time if certain error conditions are hit.

I don't know if there are any other daemons with a similar problem? @bari12

bari12: It's a bit specific to this one, since this is a unique-constraint hit. For example that won't show up at the judge-cleaner, since that one only removes things. I would leave this specific for the evaluator for now.

bari12 (approved):

PR #4948, opened 2021-10-26
Transfers: Add optional srm https compatibility Fix #4944

### LINKED ISSUE #4944: SRM to HTTPS optional compatibility
Motivation
----------
https://github.com/rucio/rucio/blob/882b662a24c75d515080606be635959566b17f14/lib/rucio/common/constants.py#L42
Defines a scheme_map between protocol compatibility in Rucio. We need an additional compatibility between `srm` and `https` (and vice-versa) which is only optionally (based on a config parameter) activated.

Modification
------------
Check if an option
```
[transfers]
srm_https_compatibility = True
```
is set and add it to the scheme_map.

joeldierkes:

### DIFF
--- a/etc/rucio.cfg.template
+++ b/etc/rucio.cfg.template
@@ -255,3 +255,6 @@ admin_issuer = wlcg
 
 [api]
 #endpoints = accountlimits, accounts, config, credentials, dids, export, heartbeats, identities, import, lifetime_exceptions, locks, meta, ping, redirect, replicas, requests, rses, rules, scopes, subscriptions
+
+[transfers]
+srm_https_compatibility = False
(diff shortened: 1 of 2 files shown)


joeldierkes:
`etc/rucio.cfg.template` had the execution flag set on the file, I reverted that (`chmod -x etc/rucio.cfg.template`).

etc/rucio.cfg.template
```diff
@@ -253,3 +253,6 @@ admin_issuer = wlcg
 
 [api]
 #endpoints = accountlimits, accounts, config, credentials, dids, export, heartbeats, identities, import, lifetime_exceptions, locks, meta, ping, redirect, replicas, requests, rses, rules, scopes, subscriptions
+
+[transfers]
```
mlassnig, line 259: maybe part of [conveyor]?

joeldierkes: The issue explicitly stated that it should be under `[transfer]`. `[conveyor]` makes sense to me. What do you think @bari12 ?

bari12: Should be transfers, since we want to get away from the code-names anyway.

bari12:
Rebase please; otherwise gtg

bari12 (approved):

PR #4953, opened 2021-10-26
Update cli_examples.rst

ivukotic:
that is not auth server we should be connecting but a regular rucio host


bari12:
Hi Ilija, there is no need to put this to the `old-doc` branch, that is really deprecated.
All that info should now go to rucio/documentation (I haven't checked if the relevant page exists there though)

Issue #4931, opened 2021-10-21
Rename command-line option --expiration-date


dchristidis:
Motivation
----------

From the documentation of `rucio-admin replicas declare-temporary-unavailable`:

```
  --expiration-date EXPIRATION_DATE
                        Timeout in hours when the replicas will become available again. Default 24
```

This option name is misleading as the argument is not in date format. This has lead to some case were people tried something like `--expiration-date 20210720`, which works, but the expiration date set in Rucio is in year 4327.

I think it’s preferable to maintain a relative timeout (like with rule lifetime) rather than enforcing some date format.

Modification
------------

1. Add a new option named something like `--expiration-timeout` or `--duration` and deprecate the current one, with the intention of removing it entirely in the next major Rucio release.
2. Use seconds instead of hours to match `update-rule --lifetime`.
3. Consider removing the default value (the user should set it explicitly).
3. Update the documentation.
4. Have an allowed range, maybe at least one day but less that a month? `0` should be allowed as an exception and would mean ‘revert the previous declaration’. This might have to be implemented server-side too (if so, a different issue is needed).

joeldierkes:
Some thoughts:
1. & 2. Explicitly add `--duration-in-hours` and `--duration-in-seconds` to be more specific. Users are used to hours and this could lead to confusion.
5. We should consider adding another CLI option like `--abort-previsous-declaration` rather than overloading some numbers. This would be more explicit and more user-intuitive.

dchristidis:
An interesting remark, but I think would rather have the CLI use seconds as a default and be able to process suffixes like `h`, `d`, `m` and `y` (this would be useful for `add-rule` and `update-rule` too).

joeldierkes:
I like the idea of doing it consistent and having more than just seconds. However, just defining our own norm could also be annoying for users. We should create an extra issue for that. I'll go with seconds now.

dchristidis:
This feature is rarely used and probably unknown to most; I don’t regard changing the unit (as long as it’s accompanied by changing the option name) as a concern.

dchristidis:
Given that item 5 remains to be discussed and implemented, I would suggest reopening the issue.

bari12:
@dchristidis or @joeldierkes please create a ticket for this separately. That range should be implemented on server side.

PR #4962, opened 2021-10-28
Core: `--duration` should be bounded #4956

### LINKED ISSUE #4956: `--duration` should be bounded
Motivation
----------
As referenced in #4931 the CLI option `--duration` of `rucio-admin replicas declare-temporary-unavailable` should rather have an allowed time-range than accepting every number. This avoids the mistake of submitting time-ranges that are too big (e.g. > 1000y).

Modification
------------
There will be a server-side check if the given time is between `(0s, 30days]`. If not, an explicit Exception is raised with a respective error message for the user.

joeldierkes:
As referenced in #4931 the CLI option `--duration` of `rucio-admin replicas
declare-temporary-unavailable` should rather have an allowed time-range than
accepting every number. This avoids the mistake of submitting time-ranges that
are too big (e.g. > 1000y).

### DIFF
--- a/bin/rucio-admin
+++ b/bin/rucio-admin
@@ -69,7 +69,8 @@ from rucio.common.exception import (AccountNotFound, DataIdentifierAlreadyExists
                                     DataIdentifierNotFound, InvalidObject, ReplicaNotFound,
                                     RSENotFound, RSEOperationNotSupported, InvalidRSEExpression,
                                     DuplicateContent, RuleNotFound, CannotAuthenticate,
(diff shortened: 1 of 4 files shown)


joeldierkes:
The test `test_client_add_temporary_unavailable_pfns` uses a duration of 0 to test temporary unavailable pfns. We should define properly if 0 should be included in the range or not.
From my POV it shouldn't, I don't see a point of having a bad_pfn expiring at the exact moment.

bari12 (approved):

PR #4968, opened 2021-11-02
Replicas: Sort `list_replicas` in non metalink case Fix #4923

### LINKED ISSUE #4923: Sorting of replicas with list_replicas in non metalink case
Motivation
----------

Currently replicas are only sorted (geoip, etc.) if the `list_replicas` request is made with `metalink=True`

https://github.com/rucio/rucio/blob/92474868cec7d64b83b46b451ecdc4a294ef2711/lib/rucio/web/rest/flaskapi/v1/replicas.py#L119

This is unfortunately a bit inconsistent, and also for the json return it should be sorted.

Modification
------------
The issue with the json return is that it is not a list, but a dictionary with `priority` values. These need to be updated by the function. (The historic `rses` field in the return is not sortable, only the `pfns` one is)

joeldierkes:
The non metalink case of `list_replicas` did not update the `priorities` values
of the listed replicas. This commit updates these values by using the already
sorted array of replicas.

### DIFF
--- a/lib/rucio/tests/test_replica_sorting.py
+++ b/lib/rucio/tests/test_replica_sorting.py
@@ -19,6 +19,8 @@
 # - Andrew Lister <andrew.lister@stfc.ac.uk>, 2019
 # - Patrick Austin <patrick.austin@stfc.ac.uk>, 2020
 # - Benedikt Ziemons <benedikt.ziemons@cern.ch>, 2020-2021
+# - Radu Carpa <radu.carpa@cern.ch>, 2021
+# - Joel Dierkes <joel.dierkes@cern.ch>, 2021
 
 import copy
 import json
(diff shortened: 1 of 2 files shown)


bziemons (commented):
Wouldn't it be better to mock the random sort function and check if it was called and return a predefined order to check against instead of relying on the random?

joeldierkes:
You're completely right. I had problems mocking the `sort_geoip` function. I changed it. 🙇

bziemons:
I know the mocks are sometimes a bit wonky and only work with certain imports :) I am glad it worked out

bziemons (dismissed):
lgtm

rcarpa (dismissed):
LGTM.

While not really needed for fixing this issue, I see a lot of similitudes with the list replicas further down below https://github.com/rucio/rucio/blob/ea32d215f50240796aefe6a78c2d003c619fd2be/lib/rucio/web/rest/flaskapi/v1/replicas.py#L375  
Would be nice if you'd try to extract common code into a separate function and use it in both places. 
Proof that code duplication is bad: I fixed the sorting issue for that case, and didn't know there is a second place to fix (the one fixed by this PR). 
Alternatively, open a new issue for somebody else (you can assign me) to take care of it

bari12 (approved):

PR #4971, opened 2021-11-02
Pre commit script

joeldierkes:
Add a verbose cli arument to the add_header script
---
The `add_header` script prints the generated header to the console per
default. However, most of the time this is not useful, since the change itself
is not important. The fact that the header changed is.

This commit disables the output of these unnecessary informations per default
and introduces the command line argument `--verbose` to display the entire
generated header when nessecary.

Add pre-commit file
---
Before creating a commit the changed files should be checked for thier
integrity. All linter errors should be fixed and the header should be generated.

This commit introduces the `pre-commit` config for the pre-commit git hook,
which automatically runs flake8 and the add_header script before a commit. This
prevents developers from submitting non-conform code changes. The config is used
by the python `pre-commit` program.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,18 @@
+repos:
+- repo: https://github.com/pre-commit/pre-commit-hooks
+  rev: v2.3.0
+  hooks:
+  - id: end-of-file-fixer
+  - id: trailing-whitespace
+- repo: https://github.com/PyCQA/flake8
+  rev: 4.0.1
+  hooks:
+  - id: flake8
+- repo: local
+  hooks:
+  - id: add-header
+    name: Run `add_header` script
+    description: Run the `add_header` script to verify the python file headers
+    entry: tools/add_header -i
+    language: system
+    types: [python]
--- a/tools/add_header
+++ b/tools/add_header
@@ -175,7 +175,7 @@ def main(arguments):
 
                 with open(MyFile, 'w') as modified:
                     modified.write(header + data)
-            else:
+            elif arguments.verbose:
                 print(header)
 
             sys.exit(0)
@@ -245,7 +245,7 @@ def main(arguments):
                         if count == 0 or not_comment_line:
                             modified.write(line)
 
-        elif not arguments.check:
+        elif not arguments.check or arguments.verbose:
             print(header)
     exit(exit_code)
 
@@ -261,6 +261,7 @@ if __name__ == '__main__':
                         help='Checks if the script would change anything (non-zero exit code equals change).')
     parser.add_argument('--disable-add-me', action='store_true',
                         help="Do not add the person who runs this command, except if there already is a contributing commit.")
+    parser.add_argument('--verbose', '-v', action='store_true', help='Verbose mode')
     parser.add_argument(dest='MyFiles', action='store', nargs='+', default=None, help='The files')
 
     main(parser.parse_args())


rcarpa (dismissed):
I'm strongly against a pre-commit hook. 
If we want to enforce such checks, I'd suggest to make it as an additional action in github. Which would actually be a very good idea.

Locally on my machine, I want to be able to have whatever commits I want locally and don't need a script to be run to tell me what I can and I cannot do in my personal repository.

joeldierkes:
I understand the argument, however the possibility to prevent simple errors (code format errors, header errors) in an early stage of the development is a chance we shouldn't miss. It would prevent GitHub runners from checking unnecessary mistakes, and the developer doesn't have to focus on unimportant things like running the `add_header` script himself.

I would suggest enabling it and let the developer decide if it should be disabled. There are plenty options to disable pre-commit checks (like [`git commit --no-verify`](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---no-verify) or simply not link the `pre-commit` script).

The integration of the `add_header` script into the workflows is a good idea, I'll create an issue for that.

rcarpa:
What you describe is not a job to be done in a pre-commit script. Keeping the upstream clean must be enforced at the moment when we merge to upstream. I do tens of commits locally for each one commit I propose for merge upstream. I guarantee that a script that makes my life more difficult in an operation which I do sometimes hundreds of times per day (rebase/rename/squash/etc a stack of many commits will easily end in 100 commits in a day), will be removed immediately from my machine. 
The result? upstream will not be kept cleaner, because I'll never run that script. 
If the goal is to make upstream cleaner, it must be a gitlab action;
if the goal is to make your life easier, it must be kept local on your machine;

joeldierkes:
> If the goal is to make upstream cleaner, it must be a gitlab action;
if the goal is to make your life easier, it must be kept local on your machine;

I'm on your site, but keep in mind that this PR merely adds the opportunity to use existing pre-commit hooks without writing some yourself. In my opinion some developers will benefit from that.

It won't force you to use them. There is just no better place to share them than in the `tools` folder of the project itself.

tools/configure_git.sh
```diff
@@ -17,8 +17,13 @@
...
-git remote add upstream https://github.com/rucio/rucio.git
+upstream_url="$(git config --get remote.upstream.url)"
+if [[ -z $upstream_url ]]
+then
+    git remote add upstream https://github.com/rucio/rucio.git
+fi
 
-cp tools/prepare-commit-msg .git/hooks/prepare-commit-msg
-chmod +x .git/hooks/prepare-commit-msg
+ln -sf ../../tools/pre-commit .git/hooks/pre-commit
```
rcarpa: What you say is not true. On this line, you enforce the hook on everybody who tries to setup their env by following the rucio documentation. This includes people who are not very technical. This can even result in additional support questions about why things don't work. Not everybody who will commit to rucio has extensive background in computer science.

joeldierkes: > What you say is not true.

What I wrote is true. Yes it enables the pre-commit script by default, but nobody prevents the developer from deactivating it (thus no-one is **forced** to use them).

The "additional support questions" you mentioned are neglectable improbable, also we can now reference this issue with the commented solutions on how to disable this feature.

We can deactivate it by default, but this does not answer the main question if we should include the script in the repo.

rcarpa:
I'm of the opinion that it bring your personal workflow into the public repository. But, if it's deactivated by default, I will not fight against it.

mlassnig:
Following both your points :-) We had some pre-commit hooks in the past, but it was a bit of a hassle for those who didn't like them, so we got rid of them.

IMHO it's fine to add the script, just keep it disabled by default and do something like `--enable-pre-commit` ?

bari12:
I thought about this a bit now and thinking back I think I also always disabled the hook (It was pylint + flake8 back then) since it was just too cumbersome.
I would
- Add it as an option like suggested
- Also add it to the GH action (We talked about that already)

joeldierkes:
GH action integration issue #4980

rcarpa:
If there was some git knowledge in the add_header script itself (rather than adding a separate hook), and it allowed to be run like 
```
./tools/add_headers --from-commit 0abc333d 
```
I would gladly use it :) If it ind all changed files and applies headers only on them

We can then add the a verify option
```
./tools/add_headers --from-commit 0abc333d  --verify 
```
and use it in actions.

maany (changes requested):
LGTM, I suggest we should update the [Contributors Guide](https://github.com/rucio/rucio/blob/master/CONTRIBUTING.rst#git-hooks) and specify that they need to 
```
pip install pre-commit
pre-commit install
```
to use these hooks ;)

joeldierkes:
Related to #5176, I'll create a new issue in the `rucio/documentation` project regarding the contribution documentation.

maany (dismissed):
Tested and verified ;) LGTM

maany:
@joeldierkes can you re-run the integration tests? thanks

joeldierkes:
I did it for now, however since I did not touch any files related to the failing job (if they were related to the conveyor we would have a big problem) I don't think it was really necessary... :sweat_smile:

bari12:
Needs to be rebased :-)

joeldierkes:
:+1:

Issue #4912, opened 2021-10-18
exception on missing external_host in submitter


rcarpa:
Motivation
----------

This case must be correctly handled in submitter

```
conveyor-submitter[3/40] : Failed to submit a job with error 'NoneType' object has no attribute 'replace'
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/rucio/daemons/conveyor/common.py", line 129, in submit_transfer
    timeout=timeout)
  File "/usr/local/lib/python3.6/site-packages/rucio/core/transfer.py", line 663, in submit_bulk_transfers
    transfer_id = FTS3Transfertool(external_host=external_host, token=transfer_token).submit(files=job_files, job_params=job_params, timeout=timeout)
  File "/usr/local/lib/python3.6/site-packages/rucio/transfertool/fts3.py", line 193, in submit
    record_counter('transfertool.fts3.%s.submission.failure' % self.__extract_host(self.external_host), len(files))
  File "/usr/local/lib/python3.6/site-packages/rucio/transfertool/fts3.py", line 607, in __extract_host
    return urlparse(external_host).hostname.replace('.', '_')
AttributeError: 'NoneType' object has no attribute 'replace'
```

PR #4978, opened 2021-11-05
Transfers: enable usage of Request.transfertool DB attribute. #4912

rcarpa:
Historically, submitter has two parameters related to transfer tools:
 - transfertool: defines which Transfertool to use for submission.
 - filter_transfertool: used to filter on Requests.transfertool
   attribute when retrieving requests from the database.
As implemented until now, if "filter_transfertool" was set,
"transfertool" had to have the exact same value, otherwise it could
result in unexpected behavior. Moreover, if "filter_transfertool"
wasn't set, the conveyor completely ignored the database value and
would try to submit the transfer to whichever tool was defined by
"transfertool".

Modify this behavior, if Request.transfertool is set in the database,
it is now a hard requirement and a submitter which cannot speak this
transfertool will ignore the request. The "transfertool" submitter
parameter can be now updated to define a list of tools in order of
preference. Transfertools evaluate in order if they are able to submit
the request. The first matching is used.

The transfertool objects were updated to support this use case
and many "if transfertool == bla" code paths were substituted by
polymorphic object method calls. FTS and globus specific code was
moved to corresponding Transfertool classes.


rcarpa:
This is a fix for the 1.26 line only. Associated PR for master: #4969

bari12 (approved):

PR #4987, opened 2021-11-09
Implement the `recalculate_rses_for_rule` function

### LINKED ISSUE #536: Possibility to re-evaluate a rule
Motivation
----------
For example, when decomissioning a site the RSE is already blacklisted, so all the rules could be "re-evaluated".

joeldierkes:

### DIFF
--- a/lib/rucio/core/rule.py
+++ b/lib/rucio/core/rule.py
@@ -1478,6 +1478,66 @@ def move_rule(rule_id, rse_expression, activity=None, source_replica_expression=
 
 
 @transactional_session
+def recalculate_rsese_for_rule(rule_id, session=None):
+    """
+    Update the associated rses for a rule.
+
+    :param rule_id:                    Rule to be moved.
+    :param session:                    The DB Session.
+    :raises:                           RuleNotFound, RuleReplaceFailed, InvalidRSEExpression, RSEWriteBlocked
+    """
+    try:
+        rule = session.query(models.ReplicationRule).filter_by(id=rule_id).one()
+
+        if rule.child_rule_id:
+            raise RuleReplaceFailed('The rule must not have a child rule.')
+
+        grouping = {RuleGrouping.ALL: 'ALL', RuleGrouping.NONE: 'NONE'}.get(rule.grouping, 'DATASET')
+
+        if rule.expires_at:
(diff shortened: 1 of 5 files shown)


lib/rucio/core/rule.py
```diff
@@ -3220,3 +3280,30 @@ def get_scratch_policy(account, rses, lifetime, session=None):
...
+                       .join(models.DataIdentifierAssociation,
+                             and_(
+                                 models.DataIdentifierAssociation.scope == models.ReplicationRule.scope,
+                                 models.DataIdentifierAssociation.name == models.ReplicationRule.name)) \
+                       .join(models.RSEFileAssociation,
+                             and_(
+                                 models.RSEFileAssociation.scope == models.DataIdentifierAssociation.child_scope,
+                                 models.RSEFileAssociation.name == models.DataIdentifierAssociation.child_name)) \
+                       .filter(models.RSEFileAssociation.rse_id == rse_id) \
+                       .distinct()
```
joeldierkes: I'm not sure if this could lead to performance issues, I would like to have the `filter` as early as possible, but the query must return the `ReplicationRule` entries.

lib/rucio/core/rule.py
```diff
@@ -1477,6 +1477,66 @@ def move_rule(rule_id, rse_expression, activity=None, source_replica_expression=
         raise RuleNotFound('No rule with the id %s found' % (rule_id))
 
 
+@transactional_session
+def recalculate_rsese_for_rule(rule_id, session=None):
```
mlassnig, line 1481: how about `evaluate_rses_for_rule` ?

lib/rucio/core/rule.py
```diff
@@ -1477,6 +1477,66 @@ def move_rule(rule_id, rse_expression, activity=None, source_replica_expression=
...
+                                   comment=rule.comments,
+                                   session=session)
+        except RSEWriteBlocked as e:
+            # Could not create new Rule
+            rule.copies = copies
+            session.flush()
+            raise e
+        session.flush()
+
+        update_rule(rule_id=rule_id, options={'lifetime': 0}, session=session)
```
mlassnig, line 1532: hardcoded lifetime?

PR #5003, opened 2021-11-12
REST & API: decorate some APIs with session managers. Closes #5002

### LINKED ISSUE #5002: decorate some APIs function with session handler and pass session
Motivation
----------
It is part of #4945,  add as separate issue to make it appear in changelog.

The session is destroyed each time a decorated session managed context is exited. This results in many rollbacks on database side.

Ensure that only one session is created per rest api call to api/rules. 
This is a test to verify our hypothesis and see if it's worth generalizing this solution.

rcarpa:
Only create the session object once per api call to rules enpoint
and pass it through. The goal is to test if the number of un-needed
rollbacks on the database side is decreased thanks to avoiding
sessionmaker.remove() calls.

### DIFF
--- a/lib/rucio/api/rule.py
+++ b/lib/rucio/api/rule.py
@@ -25,7 +25,7 @@
 # - Eli Chadwick <eli.chadwick@stfc.ac.uk>, 2020
 # - James Perry <j.perry@epcc.ed.ac.uk>, 2020
 # - Ian Johnson <ijjorama@gmail.com>, 2021
-# - Radu Carpa <radu.carpa@cern.ch>, 2021
+# - Radu Carpa <radu.carpa@cern.ch>, 2021-2022
 # - Joel Dierkes <joel.dierkes@cern.ch>, 2021-2022
 
 from rucio.api.permission import has_permission
@@ -35,19 +35,22 @@ from rucio.common.schema import validate_schema
 from rucio.common.types import InternalAccount, InternalScope
 from rucio.common.utils import api_update_return_dict
 from rucio.core import rule
+from rucio.db.sqla.session import read_session, stream_session, transactional_session
 
 
-def is_multi_vo():
+@read_session
+def is_multi_vo(session=None):
     """
     Check whether this instance is configured for multi-VO
     returns: Boolean True if running in multi-VO
     """
-    return config_get_bool('common', 'multi_vo', raise_exception=False, default=False)
+    return config_get_bool('common', 'multi_vo', raise_exception=False, default=False, session=session)
 
 
+@transactional_session
 def add_replication_rule(dids, copies, rse_expression, weight, lifetime, grouping, account, locked, subscription_id, source_replica_expression,
                          activity, notify, purge_replicas, ignore_availability, comment, ask_approval, asynchronous, delay_injection, priority,
-                         split_container, meta, issuer, vo='def'):
+                         split_container, meta, issuer, vo='def', session=None):
     """
     Adds a replication rule.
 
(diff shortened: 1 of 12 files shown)


lib/rucio/api/rule.py
```diff
@@ -160,102 +168,113 @@ def list_replication_rules(filters={}, vo='def'):
...
     Lists replication rule history..
 
     :param rule_id: The rule_id to list.
     :param issuer: The issuing account of this operation.
     :param vo: The VO of the issuer.
+    :param session: The database session in use.
     """
     kwargs = {'rule_id': rule_id}
-    if is_multi_vo() and not has_permission(issuer=issuer, vo=vo, action='access_rule_vo', kwargs=kwargs):
+    if is_multi_vo() and not has_permission(issuer=issuer, vo=vo, action='access_rule_vo', kwargs=kwargs, session=session):
```
bari12: `is_multi_vo` asks for `config_get_bool` which might query config table. Needs to be added there (Or we block it from querying the config table via the method option)

lib/rucio/api/rule.py
```diff
@@ -160,102 +168,113 @@ def list_replication_rules(filters={}, vo='def'):
...
     :param rule_id:     The rule_id to lock.
     :param options:     Options dictionary.
     :param issuer:      The issuing account of this operation
     :param vo:          The VO to act on.
+    :param session:     The database session in use.
     :raises:            RuleNotFound if no Rule can be found.
     """
     kwargs = {'rule_id': rule_id, 'options': options}
-    if is_multi_vo() and not has_permission(issuer=issuer, vo=vo, action='access_rule_vo', kwargs=kwargs):
+    if is_multi_vo() and not has_permission(issuer=issuer, vo=vo, action='access_rule_vo', kwargs=kwargs, session=session):
```
bari12: See above

lib/rucio/api/rule.py
```diff
@@ -264,36 +283,40 @@ def reduce_replication_rule(rule_id, copies, exclude_expression, issuer, vo='def
     :param exclude_expression:  RSE Expression of RSEs to exclude.
     :param issuer:              The issuing account of this operation
     :param vo:                  The VO to act on.
+    :param session:             The database session in use.
     :raises:                    RuleReplaceFailed, RuleNotFound
     """
     kwargs = {'rule_id': rule_id, 'copies': copies, 'exclude_expression': exclude_expression}
-    if is_multi_vo() and not has_permission(issuer=issuer, vo=vo, action='access_rule_vo', kwargs=kwargs):
+    if is_multi_vo() and not has_permission(issuer=issuer, vo=vo, action='access_rule_vo', kwargs=kwargs, session=session):
```
bari12: See above

lib/rucio/api/rule.py
```diff
@@ -264,36 +283,40 @@ def reduce_replication_rule(rule_id, copies, exclude_expression, issuer, vo='def
...
     Examine a replication rule.
 
     :param rule_id: The rule_id to get.
     :param issuer: The issuing account of this operation.
     :param vo: The VO of the issuer.
+    :param session: The database session in use.
     """
     kwargs = {'rule_id': rule_id}
-    if is_multi_vo() and not has_permission(issuer=issuer, vo=vo, action='access_rule_vo', kwargs=kwargs):
+    if is_multi_vo() and not has_permission(issuer=issuer, vo=vo, action='access_rule_vo', kwargs=kwargs, session=session):
```
bari12: See above

lib/rucio/api/rule.py
```diff
@@ -311,12 +334,13 @@ def move_replication_rule(rule_id, rse_expression, activity, source_replica_expr
         'activity': activity,
         'source_replica_expression': source_replica_expression
     }
-    if is_multi_vo() and not has_permission(issuer=issuer, vo=vo, action='access_rule_vo', kwargs=kwargs):
+    if is_multi_vo() and not has_permission(issuer=issuer, vo=vo, action='access_rule_vo', kwargs=kwargs, session=session):
```
bari12: See above

bari12 (changes requested):

bari12 (approved):

PR #5011, opened 2021-11-16
Transfers: enable per-vo certs for interacting with fts. Closes #4731 

### LINKED ISSUE #4731: Certificate selection for Multi_VO conveyor daemons
Motivation
----------
The conveyor daemon can only currently be pointed at a single certificate specified within the rucio.cfg, this means for multi_vo instances several daemons are required to run, each with a different VOs certificate present to allow submission and polling of the transfers. 
This however means each daemon is trying to submit (if not set to submit specific VOs) and poll all the VOs for transfers
which can lead to forbidden errors, wasted time and resources, and potentially negatively effecting monitoring.

Modification
------------
Passing VO information to the FTS3 level and extracting it when needed (if in multi_vo mode) and using that information to point to certificates specified in a new section of rucio.cfg [vo_certs] using the 3 char VO name which is easilly accessed from the scope information.

rcarpa:
To be able to add the test, I had to perform a couple of minor changes
in the functional code:
- support passing an empty "older_than" in poller. Otherwise,
utcnow() - timedelta(seconds=0) was filtering out the test request which
was just added to the database (only on oracle).
- make submission_builder_for_path a classmethod: to be able to subclass
from FTS3Transfertool while building correct object in child class.

### DIFF
--- a/lib/rucio/core/transfer.py
+++ b/lib/rucio/core/transfer.py
@@ -608,7 +608,7 @@ def is_recoverable_fts_overwrite_error(fts_status_dict):
     return False
 
 
-def bulk_query_transfers(request_host, transfers_by_eid, transfertool='fts3', timeout=None, logger=logging.log):
+def bulk_query_transfers(request_host, transfers_by_eid, transfertool='fts3', vo=None, timeout=None, logger=logging.log):
     """
     Query the status of a transfer.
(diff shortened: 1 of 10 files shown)


rcarpa:
This PR is based on top of #4997. Only the last 2 commits are new

bari12 (approved):

PR #5013, opened 2021-11-16
Patch: Fix list-file-replicas missing incompatibility Fix #5007

### LINKED ISSUE #5007: list-file-replicas rses and missing incompatibility
Motivation
----------
With client version `1.26.8.post1` if I use `rucio list-file-replicas DID --rse BLAH --missing` I get a warning
```
2021-11-12 11:48:25,296   WARNING The use of --rse is deprecated in favour of --rses and will be removed in 1.27
```
But if I change to using `--rses` I get the error:
```
Cannot use --missing without specifying a RSE
```

Modification
------------
Allow `--missing` with `--rses` I suppose?

joeldierkes:
The command line flag `--rses` did not properly replace the old cli flags. The
internal variable `args.rses` was not properly assigned and thus not initialized
if `--rse` was speciefed, where as the old internal variable `args.selected_rse`
was still used in some places.

This commit changes the internal value of `args.rses` to the appropriate value,
even if a deprecated flag is provided.

`args.rses` represents not a single RSE anymore, but a RSE expression. The
change now also fetches all RSEs, which are included in that RSE expression, and
list the file replicas for all of them if the `--rses` flag is provided.

### DIFF
--- a/bin/rucio
+++ b/bin/rucio
@@ -44,8 +44,6 @@
 # - Patrick Austin <patrick.austin@stfc.ac.uk>, 2020
 # - Radu Carpa <radu.carpa@cern.ch>, 2021
 # - Rahul Chauhan <omrahulchauhan@gmail.com>, 2021
-# - Gabriele Gaetano Fronze' <gabriele.fronze@to.infn.it>, 2020-2021
-# - Rob Barnsley <rob.barnsley@skao.int>, 2021
 # - Rakshita Varadarajan <rakshitajps@gmail.com>, 2021
 # - Christoph Ames <christoph.ames@physik.uni-muenchen.de>, 2021
 # - James Perry <j.perry@epcc.ed.ac.uk>, 2021
@@ -55,11 +53,13 @@
 # - KosKyr <90753277+KosKyr@users.noreply.github.com>, 2021
 # - Joel Dierkes <joel.dierkes@cern.ch>, 2021
 # - jdierkes <joel.dierkes@cern.ch>, 2021
+# - Rob Barnsley <robbarnsley@users.noreply.github.com>, 2021
 
 from __future__ import print_function
 
 import argparse
 import errno
+import itertools
 import logging
 import math
 import os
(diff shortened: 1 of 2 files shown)


bin/rucio
```diff
@@ -421,58 +430,62 @@ def list_file_replicas(args):
...
+            for replica, rse in itertools.product(replicas, rses):
+                if rse in list(replica['rses'].keys()) and replica['rses'][rse]:
+                    for pfn in replica['rses'][rse]:
+                        print(pfn)
         else:
-            if args.all_states:
-                header = ['SCOPE', 'NAME', 'FILESIZE', 'ADLER32', '(STATE) RSE: REPLICA']
-            else:
-                header = ['SCOPE', 'NAME', 'FILESIZE', 'ADLER32', 'RSE: REPLICA']
             for replica in replicas:
```
rcarpa, line 465: Is this a bug which is being fixed? If yes, would be nice to have a short sentence about it in the commit message.

bin/rucio
```diff
@@ -386,15 +387,30 @@ def list_file_replicas(args):
...
+        logger.error("Cannot use --rse and --rses together, please use --rses")
+        return FAILURE
+    elif args.rse_expression and args.rses:
+        # If the new cli arg and the old one are defined we don't know which one to take
+        logger.error("Cannot use --expression and --rses together, please use --rses")
+        return FAILURE
+    elif args.selected_rse:
+        option_deprecation_message("--rse", "--rses")
+        client.get_rse(args.selected_rse)    # Raise an exception if RSE does not exist
+        args.rses = args.selected_rse
```
rcarpa, line 403: I don't know if assigning to args can be problematic. Any reason for not using a local variable for the same purpose instead of assigning to args ?

rcarpa (commented):
LGTM. The comments are just nitpicking. However, I don't feel confident approving it, need a second opinion.

bin/rucio
joeldierkes: Just a refactoring to make the code more readable. It has an early exit now.
```python
if args.metalink:
    print(replicas[:-1])  # last character is newline, no need to print that
else:
    if args.missing:
    ...
```
to
```python
if args.metalink:
    print(replicas[:-1])  # last character is newline, no need to print that
    return SUCCESS

if args.missing:
...
```

bin/rucio
joeldierkes: https://docs.python.org/3/library/argparse.html#argparse.Namespace

`args` is a `Namespace` object, it shouldn't be a problem.
`args` is our central store for options, so it felt natural to add it there and don't create an extra variable.

From my pov the developer could get confused when there is another local variable (like `rses`). Someone could use `args.rses` instead of the properly initialized variable. So the `args.rses` should at least be deleted after the local variable is initialized, so why not use it in the first place.

bin/rucio
rcarpa: looks correct. I got confused by github diff. Sorry for the noise

bin/rucio
rcarpa: Good for me :)

bari12 (approved):

PR #5017, opened 2021-11-16
Core: allow `move_rule` to override all options Fix #4995

### LINKED ISSUE #4995: Allow overriding all possible options in move_rule
Motivation
----------

With #4855, it is now possible to override the activity and source replica expression when moving a rule. Client-side this looks great, but in the core the implementation could be improved (note that it’s still not possible to remove the source replica expression, which was one of the original intentions).

I would argue that there’s no need to have a fixed set of parameters. There could be a valid reason to override any of them (at the core for the needs of operations; there’s no need to expose it through the CLI yet).

Modification
------------

Could something like the following work?

```python
@transactional_session
def move_rule(rule_id: str, rse_expression: str, override: Dict[str, Any] = {}, session: Session = None):
    rule = session.query(models.ReplicationRule).filter_by(id=rule_id).one()

    …

    options = {dids=[{'scope': rule.scope, 'name': rule.name}],
               account=rule.account,
               …
    }

    for key in override:
        if key in ['dids']:
            raise Exception(f'Not allowed to override option {key!r}')
        elif key not in options:
            raise Exception(f'Non-valid override option {key!r}')
        else:
            options[key] = override[key]

    new_rule_id = add_rule(**options)

    …
```

joeldierkes:
Before the commit `move_rule` could only override two attributes of the new
rule. After the commit `move_rule` can override every field of the new rule with
the argument `override`.

### DIFF
--- a/bin/rucio
+++ b/bin/rucio
@@ -1443,10 +1443,16 @@ def move_rule(args):
     Update a rule.
     """
     client = get_client(args)
+
+    override = {}
+    if args.activity:
+        override['activity'] = args.activity
+    if args.source_replica_expression:
+        override['source_replica_expression'] = None if args.source_replica_expression.lower() == "none" else args.source_replica_expression
+
     print(client.move_replication_rule(rule_id=args.rule_id,
                                        rse_expression=args.rse_expression,
-                                       activity=args.activity,
-                                       source_replica_expression=args.source_replica_expression))
+                                       override=override))
     return SUCCESS
 
 
@@ -2524,8 +2530,8 @@ You can filter by account::
     move_rule_parser.set_defaults(function=move_rule)
     move_rule_parser.add_argument(dest='rule_id', action='store', help='Rule id')
     move_rule_parser.add_argument(dest='rse_expression', action='store', help='RSE expression of new rule')
-    move_rule_parser.add_argument('--activity', dest='activity', action='store', help='Update activity for moved rule')
-    move_rule_parser.add_argument('--source-replica-expression', dest='source_replica_expression', action='store', help='Update source-replica-expression for moved rule')
+    move_rule_parser.add_argument('--activity', dest='activity', action='store', help='Update activity for moved rule.')
+    move_rule_parser.add_argument('--source-replica-expression', dest='source_replica_expression', action='store', help='Update source-replica-expression for moved rule. Use "None" to remove the old value.')
 
     # The list-rses command
     list_rses_parser = subparsers.add_parser('list-rses', help='Show the list of all the registered Rucio Storage Elements (RSEs).')
--- a/lib/rucio/api/rule.py
+++ b/lib/rucio/api/rule.py
@@ -317,14 +317,13 @@ def examine_replication_rule(rule_id, issuer, vo='def', session=None):
 
 
 @transactional_session
-def move_replication_rule(rule_id, rse_expression, activity, source_replica_expression, issuer, vo='def', session=None):
+def move_replication_rule(rule_id, rse_expression, override, issuer, vo='def', session=None):
     """
     Move a replication rule to another RSE and, once done, delete the original one.
 
     :param rule_id:                    Rule to be moved.
     :param rse_expression:             RSE expression of the new rule.
-    :param activity:                   Activity of the new rule.
-    :param source_replica_expression:  Source-Replica-Expression of the new rule.
+    :param override:                   Configurations to update for the new rule.
     :param session:                    The DB Session.
     :param vo:                         The VO to act on.
     :raises:                           RuleNotFound, RuleReplaceFailed, InvalidRSEExpression, AccessDenied
@@ -332,16 +331,12 @@ def move_replication_rule(rule_id, rse_expression, activity, source_replica_expr
     kwargs = {
         'rule_id': rule_id,
         'rse_expression': rse_expression,
-        'activity': activity,
-        'source_replica_expression': source_replica_expression
+        'override': override,
     }
+
     if is_multi_vo(session=session) and not has_permission(issuer=issuer, vo=vo, action='access_rule_vo', kwargs=kwargs, session=session):
         raise AccessDenied('Account %s can not access rules at other VOs.' % (issuer))
     if not has_permission(issuer=issuer, vo=vo, action='move_rule', kwargs=kwargs, session=session):
         raise AccessDenied('Account %s can not move this replication rule.' % (issuer))
 
-    return rule.move_rule(rule_id=rule_id,
-                          rse_expression=rse_expression,
-                          activity=activity,
-                          source_replica_expression=source_replica_expression,
-                          session=session)
+    return rule.move_rule(**kwargs, session=session)
--- a/lib/rucio/core/rule.py
+++ b/lib/rucio/core/rule.py
@@ -16,7 +16,7 @@
 # Authors:
 # - Vincent Garonne <vincent.garonne@cern.ch>, 2012-2018
 # - Mario Lassnig <mario.lassnig@cern.ch>, 2013-2020
-# - Martin Barisits <martin.barisits@cern.ch>, 2013-2021
+# - Martin Barisits <martin.barisits@cern.ch>, 2013-2022
 # - Cedric Serfon <cedric.serfon@cern.ch>, 2014-2021
 # - David Cameron <david.cameron@cern.ch>, 2014
 # - Joaquín Bogado <jbogado@linti.unlp.edu.ar>, 2014-2018
@@ -52,6 +52,7 @@ from copy import deepcopy
 from datetime import datetime, timedelta
 from re import match
 from string import Template
+from typing import Dict, Any, Optional
 
 from dogpile.cache.api import NO_VALUE
 from six import string_types
@@ -1418,17 +1419,18 @@ def reduce_rule(rule_id, copies, exclude_expression=None, session=None):
 
 
 @transactional_session
-def move_rule(rule_id, rse_expression, activity=None, source_replica_expression=None, session=None):
+def move_rule(rule_id: str, rse_expression: str, override: Optional[Dict[str, Any]] = None, session=None):
     """
     Move a replication rule to another RSE and, once done, delete the original one.
 
     :param rule_id:                    Rule to be moved.
     :param rse_expression:             RSE expression of the new rule.
-    :param activity:                   Activity of the new rule.
-    :param source_replica_expression:  Source-Replica-Expression of the new rule.
+    :param override:                   Configurations to update for the new rule.
     :param session:                    The DB Session.
     :raises:                           RuleNotFound, RuleReplaceFailed, InvalidRSEExpression
     """
+    override = override or {}
+
     try:
         rule = session.query(models.ReplicationRule).filter_by(id=rule_id).one()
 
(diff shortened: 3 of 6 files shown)


lib/rucio/api/rule.py
```diff
@@ -318,5 +318,4 @@ def move_replication_rule(rule_id, rse_expression, activity, source_replica_expr
 
     return rule.move_rule(rule_id=rule_id,
                           rse_expression=rse_expression,
-                          activity=activity,
-                          source_replica_expression=source_replica_expression)
+                          override={'activity': activity, 'source_replica_expression': source_replica_expression})
```
dchristidis: I have the impression that this won’t work properly. In the CLI, `--activity` and `--source-replica-expression` are optional. I guess this means, when not specified, that the client will send `None`? If so, won’t this use of `override` actually replace the parameters of the original rule?

I think what needs to be done here is 1) populate `override` only when necessary and 2) have some special value which will actually translate to `None`. I propose doing in the client something similar to `update_rule`.

https://github.com/rucio/rucio/blob/d59c237f533f40116026dc9f347f4fc1297f1ff0/bin/rucio#L1419-L1420

lib/rucio/core/rule.py
```diff
@@ -1419,14 +1419,13 @@ def reduce_rule(rule_id, copies, exclude_expression=None, session=None):
 
 
 @transactional_session
-def move_rule(rule_id, rse_expression, activity=None, source_replica_expression=None, session=None):
+def move_rule(rule_id, rse_expression, override={}, session=None):
```
dchristidis: Please use this opportunity to add type hints.

dchristidis (changes requested):

lib/rucio/api/rule.py
joeldierkes: I think we don't have to replace the parameters at all when no `override` value is given, since the default values come from the original rule.

Do you confirm that @dchristidis?

dchristidis: Forgive me, I do not understand your question.

In my mind, two things should be possible from the CLI:

```
$ rucio move-rule RULE_ID RSE_EXPRESSION
$ rucio move-rule --source-replica-expression None RULE_ID RSE_EXPRESSION
```

I believe the latest changes fixed the first command, but not the second?

lib/rucio/core/rule.py
dchristidis: Thanks for indulging me. I believe `rule_id` is a string.

lib/rucio/api/rule.py
joeldierkes: Ah now i understand what you mean. You're right, the second one would not be possible right now. We have to change the command line argument itself since there is not difference between not specifying it and the value `None` right now. I'll have a look into that.

joeldierkes: Okay I though about it and would rather introduce new command line arguments to disable/remove the `source-replica-expression` and `activity`. (Something like `--without-source-replica-expression` or `--no-source-replica-expression` or `--set-source-replica-expression-none`)

If we rather want to overload the current one, we have to introduce a new `"None"` state and have to properly document it. The user has to know that `None` as value would do something different from not specifying it, and I think this could create some problems.

joeldierkes: Thoughts about that?

dchristidis: Your proposal is sensible. Personally, I have a preference for having a special value like `None` or an empty argument because I feel like it’s easier to remember, but I understand how it can be problematic.

Whatever we decide, I think it’s important to apply it consistently. Given that `rucio update-rule --lifetime None RULE_ID` has existed for some time, I think it’s worth discussing it during a DDM DevOps or Rucio meeting.

joeldierkes: Yes I'm on your side, it should be consistent! I'll implement it with a special `None` value.

Yeah we could ask that in a Rucio meeting, it should not be that important though. As you said consistency is the only important thing here.

bin/rucio
```diff
@@ -1459,10 +1460,16 @@ def move_rule(args):
     Update a rule.
     """
     client = get_client(args)
+
+    override = {}
+    if args.activity:
+        override['activity'] = None if args.activity.lower() == "none" else args.activity
+    if args.source_replica_expression:
+        override['source_replica_expression'] = None if args.source_replica_expression.lower() == "none" else args.source_replica_expression
```
joeldierkes: How should we document the `$ rucio move-rule --source-replica-expression None RULE_ID RSE_EXPRESSION` case? In the other cases, it's not documented at all.

lib/rucio/web/rest/flaskapi/v1/rules.py
```diff
@@ -307,13 +298,11 @@ def post(self, rule_id):
         parameters = json_parameters()
         rse_expression = param_get(parameters, 'rse_expression')
         rule_id = param_get(parameters, 'rule_id', default=rule_id)
-        activity = param_get(parameters, 'activity', default=None)
-        source_replica_expression = param_get(parameters, 'source_replica_expression', default=None)
+        override = param_get(parameters, 'override', default=None)
```
joeldierkes: The client can only change the `activity` and `source_replica_expression`. The rest api can override all other field too. Is that a problem?

joeldierkes: This alters the API, it's not back compatible. How should we handle it?

joeldierkes (commented):
Some questions from my side.

bin/rucio
dchristidis: May I suggest adding ‘Use 'None' to remove' or something similar here?
https://github.com/rucio/rucio/blob/4a896696917bee18b49c902b289d189ede697dfc/bin/rucio#L2488

lib/rucio/web/rest/flaskapi/v1/rules.py
dchristidis: I think this is good. Advanced users might want additional control (e.g. changing the `comment`), which may at some point come to the CLI.

dchristidis: I believe it’s necessary to accept all three, merge them into `override` if necessary. We should add a comment that this should be removed in the future.

joeldierkes: Okay yeah, I would do that too.

It should favor the override entry over the old ones if both are speified. I don't think it's necessary to throw an error here.

bin/rucio
```diff
@@ -1459,10 +1460,16 @@ def move_rule(args):
     Update a rule.
     """
     client = get_client(args)
+
+    override = {}
+    if args.activity:
+        override['activity'] = None if args.activity.lower() == "none" else args.activity
```
dchristidis: My apologies, I missed this part. It is not really possible to remove an activity, as the API will set it back to `User Subscriptions`. I think it might be preferable to not do this kind of special treatment of the argument.

https://github.com/rucio/rucio/blob/4a896696917bee18b49c902b289d189ede697dfc/lib/rucio/api/rule.py#L83-L84

joeldierkes: Oh good to know. I changed it 👍

lib/rucio/web/rest/flaskapi/v1/rules.py
```diff
@@ -307,13 +298,20 @@ def post(self, rule_id):
         parameters = json_parameters()
         rse_expression = param_get(parameters, 'rse_expression')
         rule_id = param_get(parameters, 'rule_id', default=rule_id)
+        override = param_get(parameters, 'override', default=None)
```
dchristidis: `None` or `{}`?

bin/rucio
dchristidis: Thanks. Don’t forget to revert the change in the documentation.

lib/rucio/web/rest/flaskapi/v1/rules.py
joeldierkes: Good catch

lib/rucio/api/rule.py
```diff
@@ -318,5 +316,4 @@ def move_replication_rule(rule_id, rse_expression, activity, source_replica_expr
 
     return rule.move_rule(rule_id=rule_id,
                           rse_expression=rse_expression,
-                          activity=activity,
-                          source_replica_expression=source_replica_expression)
+                          override=override if override else {})
```
dchristidis: Final comment, I promise! I believe there’s no need for this use of the ternary operator.

dchristidis: ~~In addition, given that `kwargs` is already built above, why not use that directly?~~

This appears to be some common pattern, so please ignore this comment.

joeldierkes: Uhhh, a `**kwargs`, love it! ❤️ I'd actually say this is the right place to overrule the consistency, since there is a gain in readability, and we can change it incrementally in the other places too.

PLEASE be nitpicky, that's what code-reviews are for. I'm not totally happy with the commit message either, if you have some suggestions.

dchristidis (dismissed):
All my comments have been addressed. Many thanks!

@bari12, before merging, I would like to draw your attention to two items:
1. I believe Joel did a wonderful job of maintaining backwards compatibility for older versions of the Rucio client, but you might want to double-check the implementation.
2. There’s a lot of boilerplate code needed to pass parameters through the various layers. This commit includes a reduction of that in the API layer (use of `kwargs` when calling the core function, see the last conversation in the review, which remains unresolved). I believe it to be a good idea, but maybe there’s some historical reason that I’m unaware of.

lib/rucio/core/rule.py
```diff
@@ -1424,14 +1425,13 @@ def reduce_rule(rule_id, copies, exclude_expression=None, session=None):
 
 
 @transactional_session
-def move_rule(rule_id, rse_expression, activity=None, source_replica_expression=None, session=None):
+def move_rule(rule_id: str, rse_expression: str, override: Dict[str, Any] = {}, session=None):
```
dchristidis: Following today’s meeting and the mention of pitfalls when using mutable defaults, is this a concern?

joeldierkes: Could you elaborate on that? The `override` option is never changed...

dchristidis: It’s not changed in this piece of code, but given the potential havoc it could cause in other parts of the Rucio source code (or, though highly unlikely, if this code is changed in the future) I think it’s a pattern that should be avoided completely. Pylint catches it as W0102.

(Context for anyone not present during the meeting: read [Python Mutable Defaults Are The Source of All Evil](https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil/))

joeldierkes: Ahh yes, just saw it in the chat. Good catch!
Yeah that should be an anti-pattern.

lib/rucio/core/rule.py
```diff
@@ -1424,17 +1425,18 @@ def reduce_rule(rule_id, copies, exclude_expression=None, session=None):
 
 
 @transactional_session
-def move_rule(rule_id, rse_expression, activity=None, source_replica_expression=None, session=None):
+def move_rule(rule_id: str, rse_expression: str, override: Dict[str, Any] = None, session=None):
```
dchristidis: I believe `Optional[Dict[str, Any]]` is more correct here.

joeldierkes: Yes, I am not being able to see the wood for trees...

dchristidis (dismissed):

joeldierkes:
Rebased and ready for merge. I requested a review again because I had to rebase some stuff. No changes in the actual functionality.

dchristidis (approved):

Issue #2216, opened 2019-02-19
Exception when calling core.replica.list_replicas() with a lengthy list of dids


dchristidis:
Motivation
----------

```
>>> from rucio.daemons.auditor import process_output
>>> process_output('…')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python2.7/site-packages/rucio/daemons/auditor/__init__.py", line 188, in process_output
    lost_pfns = [r['rses'][rse][0] for r in list_replicas(lost_replicas)
  File "/usr/lib/python2.7/site-packages/rucio/db/sqla/session.py", line 327, in new_funct
    raise DatabaseException(str(error))
DatabaseException: Database exception.
Details: (cx_Oracle.DatabaseError) ORA-00600: internal error code, arguments: [qcscbndv1], [65535], [], [], [], [], [], [], [], [], [], []
 [SQL: u'SELECT /*+ INDEX(DIDS DIDS_PK) */ atlas_rucio.dids.scope AS atlas_rucio_dids_scope, …
```

Note that this is not an ordinary case. The file used in this example has about 41k lost files.

Originally I was chunking on the Auditor’s side, but I was recommended against that.

Modification
------------

bari12:
How many files is this list long?

bari12:
Just following up on old tickets: The chunking should be done on the auditor side. The list_replicas method here is a core function, which should not need protection from other core/daemon calls - only on the API level. Thus please chunk this in the daemon itself.

PR #5050, opened 2021-11-26
Consistency check, Core & Internals: Patch #2216 

faluchet:
Prevent core exception from auditors list_replicas call, chunk both dark and lost files lists before any transaction

### DIFF
--- a/lib/rucio/daemons/auditor/__init__.py
+++ b/lib/rucio/daemons/auditor/__init__.py
@@ -192,7 +192,8 @@ def process_output(output, sanity_check=True, compress=True):
     for chunk in chunks(lost_replicas, 1000):
         lost_pfns.extend([r['rses'][rse_id][0] for r in list_replicas(chunk) if rse_id in r['rses']])
 
-    add_quarantined_replicas(rse_id=rse_id, replicas=dark_replicas)
+    for chunk in chunks(dark_replicas, 1000):
+        add_quarantined_replicas(rse_id=rse_id, replicas=chunk)
     logger.debug('Processed %d DARK files from "%s"', len(dark_replicas),
                  output)
     declare_bad_file_replicas(lost_pfns, reason='Reported by Auditor',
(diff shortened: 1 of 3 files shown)


lib/rucio/daemons/auditor/__init__.py
```diff
@@ -187,10 +188,15 @@ def process_output(output, sanity_check=True, compress=True):
 
     # While converting LOST replicas to PFNs, entries that do not
     # correspond to a replica registered in Rucio are silently dropped.
+    lost_pfns = []
+    for chunk in chunks(lost_replicas, 1000):
+        lost_pfns.extend([r['rses'][rse_id][0] for r in list_replicas(chunk) if rse_id in r['rses']])
```
dchristidis: I believe that this can be combined into a single statement, but I don’t have a strong preference.

lib/rucio/daemons/auditor/__init__.py
```diff
@@ -187,10 +188,15 @@ def process_output(output, sanity_check=True, compress=True):
...
+
     lost_pfns = [r['rses'][rse_id][0] for r in list_replicas(lost_replicas)
```
dchristidis: I believe this must be removed.

dchristidis (changes requested):

lib/rucio/daemons/auditor/__init__.py
faluchet: my bad, leftover now removed

lib/rucio/daemons/auditor/__init__.py
faluchet: would the following do?
`    lost_pfns = [r['rses'][rse_id][0] for chunk in chunks(lost_replicas, 1000) for r in list_replicas(chunk) if rse_id in r['rses']]
`
[177e03ad25007fa6a4bb527abe0c70cfe0a529e6](https://github.com/rucio/rucio/pull/5050/commits/177e03ad25007fa6a4bb527abe0c70cfe0a529e6)

dchristidis (approved):

dchristidis (commented):
Note: commit subjects are missing a reference to the issue.

rcarpa:
Look good. I only don't understand why the chunk size goes from 100 to 1000. Would be nice to have an explaination in the commit message.

bari12:
In general, please squash the commits together. If it is a big PR, you can leave them separate if it breaks up the development in different pieces, but for smaller things please squash them (I will squash it now)

PR #5068, opened 2021-12-02
Core : provide the possibility to specify PFNs or DIDs in declare_bad_file_replicas : Close #5033

### LINKED ISSUE #5033: Provide the possibility to specify PFNs or DIDs in declare_bad_file_replicas
Motivation
----------
Currently `declare_bad_file_replicas` only accept PFNs, but some collaboration prefer using DID, rse_id.

Modification
------------
Provide the possibility to specify PFNs or DIDs in declare_bad_file_replicas

### LINKED ISSUE #5059: declaring replica as suspicious in finisher removes sources from DB
Motivation
----------

As a result of this line executed under certain conditions in finisher for suspicious replicas: 
https://github.com/rucio/rucio/blob/289ce559ddfc9039d0fbab5af98544abb4256046/lib/rucio/core/replica.py#L347
and
https://github.com/rucio/rucio/blob/289ce559ddfc9039d0fbab5af98544abb4256046/lib/rucio/core/replica.py#L378

the sources are removed from the database. As a result, a suspicious best source for a request will never have it's ranking decreased. 
That source will be retried at each submission, even if other sources exist.

cserf:
Core : provide the possibility to specify PFNs or DIDs in declare_bad_file_replicas : Close #5033

Remove the source removal when a replicas is declared SUSPICIOUS : Closes #5059

### DIFF
--- a/lib/rucio/client/replicaclient.py
+++ b/lib/rucio/client/replicaclient.py
@@ -16,7 +16,7 @@
 # Authors:
 # - Vincent Garonne <vincent.garonne@cern.ch>, 2013-2018
 # - Mario Lassnig <mario.lassnig@cern.ch>, 2013-2021
-# - Cedric Serfon <cedric.serfon@cern.ch>, 2014-2018
+# - Cedric Serfon <cedric.serfon@cern.ch>, 2014-2021
 # - Ralph Vigne <ralph.vigne@cern.ch>, 2015
 # - Brian Bockelman <bbockelm@cse.unl.edu>, 2018
 # - Martin Barisits <martin.barisits@cern.ch>, 2018-2021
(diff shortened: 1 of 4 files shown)


lib/rucio/client/replicaclient.py
```diff
@@ -49,7 +49,7 @@ def declare_bad_file_replicas(self, pfns, reason):
         """
         Declare a list of bad replicas.
 
-        :param pfns: The list of PFNs.
```
bari12, line 52: What's the difference of this to the method below?

cserf: Before it was only a list of PFNs (string), now it can be either a list of PFNs (string) or a list of replicas {'scope': <scope>, 'name': <name>, 'rse_id': <rse_id>}
Or did I misunderstand your question ?

bari12: It's not exactly the same, but quite similar to the `declare_bad_did_replicas` below. I am just wondering, can we remove that one then?

cserf: `declare_bad_did_replicas` was introduced for CMS AFAIR. @ericvaandering should comment

bari12 (approved):

PR #5070, opened 2021-12-02
Replicas: Fix `expires_at` is None error Fix #5066

### LINKED ISSUE #5066: Error declaring bad file replicas
Motivation
----------

```
  File "/usr/local/lib/python3.6/site-packages/rucio/web/rest/flaskapi/v1/common.py", line 66, in dispatch_request
    return super(ErrorHandlingMethodView, self).dispatch_request(*args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/flask/views.py", line 163, in dispatch_request
    return meth(*args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/rucio/web/rest/flaskapi/v1/replicas.py", line 851, in post
    vo=request.environ.get('vo'),
  File "/usr/local/lib/python3.6/site-packages/rucio/api/replica.py", line 405, in add_bad_pfns
    if datetime.datetime.utcnow() <= expires_at and expires_at > datetime.datetime.utcnow() + datetime.timedelta(days=30):
TypeError: '<=' not supported between instances of 'datetime.datetime' and 'NoneType'
```

This appears to be a regression from #4962.

Modification
------------

joeldierkes:
The duration bound introduced in #4962 checks if the value of `expires_at` is in
a specific timerange. While `expires_at` is always set for calls from
`declare_temporary_unavailable_replicas`, calls from `declare_bad_file_replicas`
don't set this value. This is the desired behaivior, since they should be set
`BAD` indefinitely. However not setting `expires_at` throws an error.

This commit only checks the duration bound on `expires_at` if the argument is
set. If it is not set, the program executes as desired without throwing any
errors.

### DIFF
--- a/lib/rucio/api/replica.py
+++ b/lib/rucio/api/replica.py
@@ -402,7 +402,7 @@ def add_bad_pfns(pfns, issuer, state, reason=None, expires_at=None, vo='def'):
     if not permission.has_permission(issuer=issuer, vo=vo, action='add_bad_pfns', kwargs=kwargs):
         raise exception.AccessDenied('Account %s can not declare bad PFNs' % (issuer))
 
-    if datetime.datetime.utcnow() <= expires_at and expires_at > datetime.datetime.utcnow() + datetime.timedelta(days=30):
+    if expires_at and datetime.datetime.utcnow() <= expires_at and expires_at > datetime.datetime.utcnow() + datetime.timedelta(days=30):
         raise exception.InputValidationError('The given duration of %s days exceeds the maximum duration of 30 days.' % (expires_at - datetime.datetime.utcnow()).days)
 
     issuer = InternalAccount(issuer, vo=vo)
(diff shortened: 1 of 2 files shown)


lib/rucio/api/replica.py
```diff
@@ -402,7 +402,7 @@ def add_bad_pfns(pfns, issuer, state, reason=None, expires_at=None, vo='def'):
     if not permission.has_permission(issuer=issuer, vo=vo, action='add_bad_pfns', kwargs=kwargs):
         raise exception.AccessDenied('Account %s can not declare bad PFNs' % (issuer))
 
-    if datetime.datetime.utcnow() <= expires_at and expires_at > datetime.datetime.utcnow() + datetime.timedelta(days=30):
+    if expires_at and datetime.datetime.utcnow() <= expires_at and expires_at > datetime.datetime.utcnow() + datetime.timedelta(days=30):
```
cserf, line 405: It should apply only to TEMPORARY_UNAVAILABLE

joeldierkes: We will open an extra Issue for that, this is just the fix for the bug and does not introduce any new behavior.

bari12 (approved):

PR #5072, opened 2021-12-02
Replicas: Verify the value of `expires_at` Fix #5071

### LINKED ISSUE #5071: Verify the value of `expires_at` for bad/temporary unavailable replicas
Motivation
----------
When a replica is declared as `TEMPORARY UNAVAILABLE`, a value should be provided in `expires_at`. When it is declared as `BAD`, the argument should not be populated (`None`). While the CLI complies to these constraints, the user could provide wrong arguments in the API/Python client.

Modification
------------
Add all necessary constraints.

joeldierkes:
When a replica is declared as `TEMPORARY UNAVAILABLE`, a value should be
provided in `expires_at`. When it is declared as `BAD`, the argument should not
be populated (`None`). While the CLI complies to these constraints, the user
could provide wrong arguments in the API/Python client.

This commit throws an exception if these constraints are not met.

Needs to be applied after #5070 since the `expires_at` error needs to be resolved.

### DIFF
--- a/lib/rucio/core/replica.py
+++ b/lib/rucio/core/replica.py
@@ -3116,6 +3116,11 @@ def add_bad_pfns(pfns, account, state, reason=None, expires_at=None, session=Non
     else:
         rep_state = state
 
+    if rep_state == BadPFNStatus.TEMPORARY_UNAVAILABLE and expires_at is None:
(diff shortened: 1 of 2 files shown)


joeldierkes:
@cserf ping

bari12 (approved):

PR #5098, opened 2021-12-08
Clients: Add `expiration-date` for upload Fix #5069

### LINKED ISSUE #5069: option for expiry date instead of lifetime
As far as I can tell, there doesn't appear to be an explicit option in the upload client to set an expiry date in place of a lifetime. 

The majority of our workflows natively use an expiration date. It would be good to have an option for the former, with an exception raised if both lifetime and expiry are specified. I recognise that there may be some timezone implications with this.

joeldierkes:
The upload command takes the `--lifetime` argument, which specifies the lifetime
of the created rule in seconds. This commit also adds the `--expiraion-date`,
which specifies the timestamp on which the rule will be obsolet. They both can't
be used the same time and `--expiration-date` has to have a fixed format to
avoid ambiguousity.

### DIFF
--- a/bin/rucio
+++ b/bin/rucio
@@ -53,6 +53,7 @@
 # - KosKyr <90753277+KosKyr@users.noreply.github.com>, 2021
 # - Joel Dierkes <joel.dierkes@cern.ch>, 2021
 # - jdierkes <joel.dierkes@cern.ch>, 2021
+# - Igor Mandrichenko <ivm@fnal.gov>, 2021
 # - Rob Barnsley <robbarnsley@users.noreply.github.com>, 2021
 
 from __future__ import print_function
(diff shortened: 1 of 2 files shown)


mlassnig (dismissed):

bin/rucio
```diff
@@ -2284,6 +2295,7 @@ You can filter by key/value, e.g.::
     upload_parser.set_defaults(function=upload)
     upload_parser.add_argument('--rse', dest='rse', action='store', help='Rucio Storage Element (RSE) name.', required=True).completer = rse_completer
     upload_parser.add_argument('--lifetime', type=int, action='store', help='Lifetime of the rule in seconds.')
+    upload_parser.add_argument('--expiration-date', action='store', help='The date when the rule expires, format: <year>-<month>-<day>-<hour>:<minute>:<second>. E.g. 2022-10-20-20:00:00')
```
bari12: Add information that it must be UTC here.

bari12 (changes requested):

bari12 (approved):

PR #5103, opened 2021-12-09
common.utils: modularise hash algorithms #5079

### LINKED ISSUE #5079: Allow the use of checksum/hash algorithms other than Adler32 and MD5
Motivation
----------

Now that thanks to #3494 the use of either Adler32 or MD5 is governed by the schema in use for specific Rucio deployment, why not take it a step further and allow other algorithms to be used as well? Ideally this would be done in modular fashion so that instead of requiring Rucio developers to maintain support for each and every algorithm requested by the users, they could be added by users themselves and only optionally upstreamed.

The main motivation behind this suggestion is that while there are many file systems and storage solutions out there which now calculate data checksums, only a handful of them support algorithms currently supported by Rucio. With the latter extended to support more algorithms, one could in theory take advantage checksum information from the underlying storage - thus reducing CPU load.

Modification
------------

Off the top of my head:

1. This will almost certainly require a change of some of the API functions so that instead of _adler32_ and _md5_ keywords there is a single, unified one accepting many different algorithms - something along the lines of:

```
[
  { 'algorithm': 'foo', 'value': 0xdeadbeef },
  { 'algorithm': 'bar', 'value': 0x123456789abcdef },
  ...
]
```

2. Modify rucio.common.utils so that the list of supported algorithms is computed on the fly (e.g. by parsing available submodules containing implementations) instead of being hard-coded.

3. Also in rucio.common.utils, expose calls to loaded algorithms in well-behaved fashion so that other parts of Rucio code know how to call algorithms requested by the current schema.

4. Remove any non-schema-controlled references to adler32 and md5 elsewhere in the code.

mkszuba:
The first, and likely the easiest, part of implementing support for generic checksum/hash algorithms - implement a common package for the algorithms, complete with the ability to add new algorithms by dropping more files into the right directory. All four originally implemented algorithms have been put in the same file but I also created a proof-of-concept BLAKE2b implementation (using hashlib) in a separate file, which worked fine.

### DIFF
--- a/lib/rucio/common/hash_algorithms/__init__.py
+++ b/lib/rucio/common/hash_algorithms/__init__.py
@@ -1,0 +1,27 @@
+from abc import ABC, abstractmethod
+from importlib import import_module
+from pathlib import Path
+from pkgutil import iter_modules
+
+
+class RucioHashAlgorithm(ABC):
+    @staticmethod
+    @abstractmethod
+    def compute_on_file(file):
+        raise NotImplementedError("Abstract method called")
+
+
+package_dir = Path(__file__).resolve().parent
+for (_, module_name, _) in iter_modules([package_dir]):
+    module = import_module("%s.%s" % (__name__, module_name))
+    for attribute_name in dir(module):
+        attribute = getattr(module, attribute_name)
+
+        try:
+            if issubclass(attribute, RucioHashAlgorithm):
+                algorithm_name = attribute_name.lower()
+                # TODO: check uniqueness of algorithm names
+                globals()[algorithm_name] = attribute
+        except TypeError:
+            # We are only interested in classes here
+            pass
--- a/lib/rucio/common/hash_algorithms/__init__.py
+++ b/lib/rucio/common/hash_algorithms/__init__.py
@@ -11,6 +11,35 @@ class RucioHashAlgorithm(ABC):
         raise NotImplementedError("Abstract method called")
 
 
+GLOBALLY_SUPPORTED_CHECKSUMS = ["adler32", "md5"]
+PREFERRED_CHECKSUM = GLOBALLY_SUPPORTED_CHECKSUMS[0]
+CHECKSUM_ALGO_DICT = {}
+
+
+def is_checksum_valid(checksum_name):
+    """
+    A simple function to check wether a checksum algorithm is supported.
+    Relies on GLOBALLY_SUPPORTED_CHECKSUMS to allow for expandability.
+
+    :param checksum_name: The name of the checksum to be verified.
+    :returns: True if checksum_name is in GLOBALLY_SUPPORTED_CHECKSUMS list, False otherwise.
+    """
+
+    return checksum_name in GLOBALLY_SUPPORTED_CHECKSUMS
+
+
+def set_preferred_checksum(checksum_name):
+    """
+    A simple function to set the preferred checksum algorithm.
+    Unsupported algorithms are quietly ignored.
+
+    :param checksum_name: The name of the checksum algorithm to be made preferred.
+    """
+    if is_checksum_valid(checksum_name):
+        global PREFERRED_CHECKSUM
+        PREFERRED_CHECKSUM = checksum_name
+
+
 package_dir = Path(__file__).resolve().parent
 for (_, module_name, _) in iter_modules([package_dir]):
     module = import_module("%s.%s" % (__name__, module_name))
@@ -20,8 +49,10 @@ for (_, module_name, _) in iter_modules([package_dir]):
         try:
             if issubclass(attribute, RucioHashAlgorithm):
                 algorithm_name = attribute_name.lower()
+
                 # TODO: check uniqueness of algorithm names
                 globals()[algorithm_name] = attribute
+                CHECKSUM_ALGO_DICT[algorithm_name] = attribute.compute_on_file
         except TypeError:
             # We are only interested in classes here
             pass
--- a/lib/rucio/common/hash_algorithms/default.py
+++ b/lib/rucio/common/hash_algorithms/default.py
@@ -1,0 +1,112 @@
+import hashlib
+import io
+import mmap
+import zlib
+
+from functools import partial
+
+from . import RucioHashAlgorithm
+
+
+class Adler32(RucioHashAlgorithm):
+    @staticmethod
+    def compute_on_file(file):
+        """
+        An Adler-32 checksum is obtained by calculating two 16-bit checksums A and B
+        and concatenating their bits into a 32-bit integer. A is the sum of all bytes in the
+        stream plus one, and B is the sum of the individual values of A from each step.
+
+        :param file: file name
+        :returns: Hexified string, padded to 8 values.
+        :raises: Exception: if checksum calculation failed
+        """
+
+        # adler starting value is _not_ 0
+        adler = 1
+
+        can_mmap = False
+        try:
+            with open(file, "r+b") as f:
+                can_mmap = True
+        except:
+            pass
+
+        try:
+            # use mmap if possible
+            if can_mmap:
+                with open(file, "r+b") as f:
+                    m = mmap.mmap(f.fileno(), 0)
+                    # partial block reads at slightly increased buffer sizes
+                    for block in iter(partial(m.read, io.DEFAULT_BUFFER_SIZE * 8), b""):
+                        adler = zlib.adler32(block, adler)
+            else:
+                with open(file, "rb") as f:
+                    # partial block reads at slightly increased buffer sizes
+                    for block in iter(partial(f.read, io.DEFAULT_BUFFER_SIZE * 8), b""):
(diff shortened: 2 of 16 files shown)


mkszuba:
I think I have managed to fix the test failure in the `client_syntax` suite (turns out pylint has to be explicitly told about dynamic modules) as well as the python2.7 `client` failure (forgot to add `rucio.common.hash_algorithms` to the client package list - d'oh!). No idea what is going on with python3.8 tests though, as the error appears to be coming from somewhere within pkgutil rather than my changes.

mkszuba:
List `rucio.common.hash_algorithms` as an extension in `pylintrc` rather than on the pylint command line, maybe this will make pylint stop complaining about missing members. Furthermore, do not use f-strings in the new module - they are fine for all the Python3 versions currently supported by Rucio but not for Python2.7.

lib/rucio/common/hash_algorithms/__init__.py
```diff
@@ -0,0 +1,69 @@
...
+    @staticmethod
+    @abstractmethod
+    def compute_on_file(file):
+        raise NotImplementedError("Abstract method called")
+
+
+#
+# Moved from rucio.common.utils. Could be refactored a bit but for the sake
+# of backwards compatibility, leave them be for now.
+#
```
joeldierkes: nit: The comment style seems odd to me, are the empty comment lines needed?
The comment itself if also kind of strange, if there is something to do add a `TODO:` tag or open an issue, if that is not worth it I would suggest to delete the comment.

lib/rucio/common/hash_algorithms/__init__.py
```diff
@@ -0,0 +1,69 @@
...
+    return checksum_name in GLOBALLY_SUPPORTED_CHECKSUMS
+
+
+def set_preferred_checksum(checksum_name):
+    """
+    A simple function to check wether a checksum algorithm is supported.
+    Relies on GLOBALLY_SUPPORTED_CHECKSUMS to allow for expandability.
+
+    :param checksum_name: The name of the checksum to be verified.
+    :returns: True if checksum_name is in GLOBALLY_SUPPORTED_CHECKSUMS list, False otherwise.
```
joeldierkes: This `:returns:` commit should not be there.

lib/rucio/common/hash_algorithms/__init__.py
```diff
@@ -0,0 +1,69 @@
...
+    """
+    if is_checksum_valid(checksum_name):
+        global PREFERRED_CHECKSUM
+        PREFERRED_CHECKSUM = checksum_name
+
+
+#
+# End of backwards-compatibility section
+#
```
joeldierkes: This comment seems unnecessary, is it worth keeping it?

lib/rucio/common/hash_algorithms/default.py
```diff
@@ -0,0 +1,110 @@
+import hashlib
```
joeldierkes, line 1: `default` as filename does not make much sense, if they are the only algorithms. I would suggest `algorithms`.

lib/rucio/common/hash_algorithms/default.py
```diff
@@ -0,0 +1,110 @@
...
+                    for block in iter(partial(m.read, io.DEFAULT_BUFFER_SIZE * 8), b""):
+                        adler = zlib.adler32(block, adler)
+            else:
+                with open(file, "rb") as f:
+                    # partial block reads at slightly increased buffer sizes
+                    for block in iter(partial(f.read, io.DEFAULT_BUFFER_SIZE * 8), b""):
+                        adler = zlib.adler32(block, adler)
+
+        except Exception as e:
+            raise Exception(
```
joeldierkes, line 49: nit: Should we define our own Exception? e.g. `HashAlgorithmException`

pylintrc
```diff
@@ -3,7 +3,7 @@
 # A comma-separated list of package or module names from where C extensions may
 # be loaded. Extensions are loading into the active Python interpreter and may
 # run arbitrary code.
-extension-pkg-whitelist=
+extension-pkg-whitelist=rucio.common.hash_algorithms
```
joeldierkes, line 6: Which extensions are loaded here? I don't see them. :sweat_smile:

lib/rucio/common/hash_algorithms/default.py
```diff
@@ -0,0 +1,110 @@
...
+    @staticmethod
+    def compute_on_file(file):
+        """
+        An Adler-32 checksum is obtained by calculating two 16-bit checksums A and B
+        and concatenating their bits into a 32-bit integer. A is the sum of all bytes in the
+        stream plus one, and B is the sum of the individual values of A from each step.
+
+        :param file: file name
+        :returns: Hexified string, padded to 8 values.
+        """
```
joeldierkes, line 22: Please include a `:raises ...:` comment.

For reference:https://github.com/rucio/rucio/blob/c9ea1c2dcb2a64f9e2ccac49d021914b863a684e/lib/rucio/client/uploadclient.py#L124-L127

joeldierkes (changes requested):
The PR looks **very good**! I left some small comments.
Also, could you apply the following:
- Run the `add_header` script on the changed files (`tools/add_header -i FILE` should be fine)?
- We are trying to introduce static typing, could you add type annotations?

I will activate the Github actions on the next Review cycle to not clock the Pipeline. Feel free to activate it on your fork if you have not done it already. :+1:

joeldierkes:
Also, we use Issue labels as component name in the commit messages: 
`Valid component names are listed in the label list and are usually specified on the issue of the change.`
https://rucio.cern.ch/documentation/contributing#contributing

In your case this should be `Core & Internals`

lib/rucio/common/hash_algorithms/__init__.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+package_dir = Path(__file__).resolve().parent
```
bari12, line 43: Just to confirm: Here you want to load modules which are part of this directory?

I think it might be better to do this similar to policy packages and make these hash algorithms as part of the policy package. See https://github.com/rucio/rucio/blob/master/lib/rucio/common/schema/__init__.py for example.

lib/rucio/common/hash_algorithms/__init__.py
mkszuba: Fair enough, I'm not opposed to having this removed.

lib/rucio/common/hash_algorithms/__init__.py
mkszuba: Fine by me not to have it!

pylintrc
mkszuba: From what I could find on the Web this is the only way to convince pylint not to complain about symbols from dynamically loaded modules. Not that it works as things stand, CI runs clearly show pylint keeps complaining...

lib/rucio/common/hash_algorithms/__init__.py
mkszuba: > Just to confirm: Here you want to load modules which are part of this directory?

Yes, that's the idea.

mkszuba:
Thank you for your comments, @joeldierkes and @bari12 ! Unfortunately I left ESCAPE some time after my last set of changes and it is at the moment quite unlikely that I will be able to do more work on this PR in the near future, especially given that I have so far never had anything to do with Rucio policies and would have to learn more about them first. I'll leave the PR open for now in case either I do somehow manage to make time or someone else gets interested in picking up where I left off, that said if you'd rather close it go ahead.

lib/rucio/common/hash_algorithms/__init__.py
mkszuba: Fixed - but keep in mind this has been moved verbatim from rucio/common/utils.py.

lib/rucio/common/hash_algorithms/default.py
mkszuba: I'm not against the idea - but note that this particular code snippet has been moved verbatim from rucio/common/utils.py.

lib/rucio/common/hash_algorithms/default.py
mkszuba: Fixed - but that's another bit moved verbatim from rucio/common/utils.py.

mkszuba:
> Also, we use Issue labels as component name in the commit messages: `Valid component names are listed in the label list and are usually specified on the issue of the change.` https://rucio.cern.ch/documentation/contributing#contributing
> 
> In your case this should be `Core & Internals`

I do not think I am allowed to change labels on Rucio PRs.

lib/rucio/common/hash_algorithms/default.py
mkszuba: Leaving this be because this might have to be changed yet again in the course of making this a policy matter.

bari12:
Thanks a lot for all the work @mkszuba and all the best for the future. We will evaluate how we best carry on with this work, please leave the PR open in the meanwhile.

PR #5104, opened 2021-12-10
Migrating json metadata plugin to use filtering engine #3750 

### LINKED ISSUE #5144: Include new operator engine also to json based columns

robbarnsley:

### DIFF
--- a/bin/rucio
+++ b/bin/rucio
@@ -54,7 +54,7 @@
 # - Joel Dierkes <joel.dierkes@cern.ch>, 2021-2022
 # - jdierkes <joel.dierkes@cern.ch>, 2021
 # - Igor Mandrichenko <ivm@fnal.gov>, 2021
-# - Rob Barnsley <robbarnsley@users.noreply.github.com>, 2021
+# - Rob Barnsley <robbarnsley@users.noreply.github.com>, 2021-2022
 # - Carlosbg <bogomcar@gmail.com>, 2021
 
 from __future__ import print_function
@@ -619,7 +619,7 @@ def list_dids(args):
         table.append(['%s:%s' % (did['scope'], did['name']), did['did_type']])
 
     if args.short:
-        for did, dummy in table:
+        for did, _ in table:
             print(did)
     else:
         print(tabulate(table, tablefmt=tablefmt, headers=['SCOPE:NAME', '[DID TYPE]']))
@@ -638,18 +638,6 @@ def list_dids_extended(args):
     filters = {}
     type_ = 'collection'
     table = []
-    if args.filter:
-        try:
-            filters, type_ = parse_did_filter_from_string(args.filter)
-        except InvalidType as error:
-            logger.error(error)
-            return FAILURE
-        except ValueError as error:
-            logger.error(error)
-            return FAILURE
-        except Exception:
-            logger.error("Invalid Filter. Filter must be 'key=value', 'key>=value', 'key>value', 'key<=value', 'key<value'")
-            return FAILURE
 
     try:
         scope, name = get_scope(args.did[0], client)
@@ -658,6 +646,7 @@ def list_dids_extended(args):
     except InvalidObject:
         scope = args.did[0]
         name = '*'
+
     if scope not in client.list_scopes():
         logger.error('Scope not found')
         return FAILURE
@@ -665,19 +654,39 @@ def list_dids_extended(args):
     if args.recursive and '*' in name:
         logger.error('Option recursive cannot be used with wildcards')
         return FAILURE
-    elif ('name' in filters) and (name != '*'):
-        logger.error('You cannot use a wildcard query and a filter by name')
+    else:
+        if filters:
+            if ('name' in filters) and (name != '*'):
+                logger.error('Must have a wildcard in did name if filtering by name')
+                return FAILURE
+
+    try:
+        filters, type_ = parse_did_filter_from_string_fe(args.filter, name)
+    except InvalidType as error:
+        logger.error(error)
         return FAILURE
-    filters['name'] = name
+    except DuplicateCriteriaInDIDFilter as error:
+        logger.error(error)
+        return FAILURE
+    except DIDFilterSyntaxError as error:
+        logger.error(error)
+        return FAILURE
+    except ValueError as error:
+        logger.error(error)
+        return FAILURE
+    except Exception as e:
+        logger.error(e)
+        return FAILURE
 
     for did in client.list_dids_extended(scope, filters=filters, did_type=type_, long=True, recursive=args.recursive):
         table.append(['%s:%s' % (did['scope'], did['name']), did['did_type']])
 
     if args.short:
-        for did, dummy in table:
+        for did, _ in table:
             print(did)
     else:
         print(tabulate(table, tablefmt=tablefmt, headers=['SCOPE:NAME', '[DID TYPE]']))
+
     return SUCCESS
 
 
@@ -2150,7 +2159,7 @@ You can filter by key/value, e.g.::
 
     list_extended_parser.set_defaults(function=list_dids_extended)
     list_extended_parser.add_argument('--recursive', dest='recursive', action='store_true', default=False, help='List data identifiers recursively.')
-    list_extended_parser.add_argument('--filter', dest='filter', action='store', help='Filter arguments in form `key=value,another_key>=next_value`. Valid keys are type and length. Use `type` to specify a DID type to search for. Valid values are `ALL`, `COLLECTION`, `CONTAINER`, `DATASET`, `FILE`. Use `length` to filter for an amount of corresponding Data IDentifiers. Valid operators are `=`, `>`, `>=`, `<`, `<=`.')  # noqa: E501
+    list_extended_parser.add_argument('--filter', dest='filter', action='store', help='Single or logically combined filtering expression(s) either in the form <key><operator><value> or <value1><operator1><key><operator2><value2> (compound inequality). Operators must belong to the set of (<=, >=, ==, !=, >, <). The following conventions for combining expressions are used: ";" represents the logical OR operator; "," represents the logical AND operator.')  # noqa: E501
     list_extended_parser.add_argument('--short', dest='short', action='store_true', help='Just dump the list of DIDs.')
     list_extended_parser.add_argument(dest='did', nargs=1, action='store', default=None, help='Data IDentifier pattern')
 
--- a/lib/rucio/client/didclient.py
+++ b/lib/rucio/client/didclient.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Copyright 2013-2021 CERN
+# Copyright 2013-2022 CERN
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -17,7 +17,7 @@
 # - Vincent Garonne <vincent.garonne@cern.ch>, 2013-2018
 # - Ralph Vigne <ralph.vigne@cern.ch>, 2013-2015
 # - Mario Lassnig <mario.lassnig@cern.ch>, 2013-2020
-# - Martin Barisits <martin.barisits@cern.ch>, 2013-2020
+# - Martin Barisits <martin.barisits@cern.ch>, 2013-2021
 # - Yun-Pin Sun <winter0128@gmail.com>, 2013
 # - Thomas Beermann <thomas.beermann@cern.ch>, 2013
 # - Cedric Serfon <cedric.serfon@cern.ch>, 2014-2021
@@ -31,8 +31,9 @@
 # - Aristeidis Fkiaras <aristeidis.fkiaras@cern.ch>, 2020
 # - Alan Malta Rodrigues <alan.malta@cern.ch>, 2020
 # - Benedikt Ziemons <benedikt.ziemons@cern.ch>, 2020-2021
-# - Gabriele Gaetano Fronze' <gabriele.fronze@to.infn.it>, 2020-2021
-# - Rob Barnsley <rob.barnsley@skao.int>, 2021
+# - David Población Criado <david.poblacion.criado@cern.ch>, 2021
+# - Igor Mandrichenko <ivm@fnal.gov>, 2021
+# - Rob Barnsley <robbarnsley@users.noreply.github.com>, 2021-2022
 
 from __future__ import print_function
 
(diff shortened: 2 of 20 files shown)


bin/rucio
```diff
@@ -676,26 +664,47 @@ def list_dids_extended(args):
...
-    elif ('name' in filters) and (name != '*'):
-        logger.error('You cannot use a wildcard query and a filter by name')
+    else:
+        if filters:
+            if ('name' in filters) and (name != '*'):
+                logger.error('Must have a wildcard in did name if filtering by name')
+                return FAILURE
+
+    try:
+        filters, type_ = parse_did_filter_from_string_fe(args.filter, name)
```
joeldierkes, line 664: What if `--filter` is not provided? It's optional

lib/rucio/core/did_meta_plugins/__init__.py
```diff
@@ -16,186 +16,211 @@
...
     else:
-        for meta_handler in METADATA_HANDLERS:
-            if meta_handler.get_plugin_name().lower() == plugin.lower():
-                return meta_handler.get_metadata(scope, name, session=session)
-
-        raise NotImplementedError('Metadata plugin %s is not enabled on the server.' % plugin)
+        for metadata_plugin in METADATA_PLUGIN_MODULES:
+            if metadata_plugin.get_plugin_name().lower() == plugin.lower():
+                return metadata_plugin.get_metadata(scope, name, session=session)  
+    raise NotImplementedError('Metadata plugin "%s" is not enabled on the server.' % plugin)    #FIXME: this exception isn't returned correctly?
```
joeldierkes: I don't understand the `FIXME`, is it not correctly handled?

Also, please document that an exception is thrown (`:raises ...: ...`). It was forgotten before.

lib/rucio/core/did.py
```diff
@@ -1606,7 +1606,7 @@ def list_dids(scope, filters, did_type='collection', ignore_case=False, limit=No
 
 @read_session
 def list_dids_extended(scope, filters, did_type='collection', ignore_case=False, limit=None,
-                       offset=None, long=False, recursive=False, session=None):
+                       offset=None, long=False, recursive=False, ignore_dids=None, session=None):
```
joeldierkes, line 1609: A test would be nice, have a look at https://github.com/rucio/rucio/blob/c857f58149b09d785338c0abe02e3487b85f4cd9/lib/rucio/tests/test_did_meta_plugins.py

lib/rucio/client/didclient.py
```diff
@@ -94,28 +97,38 @@ def list_dids(self, scope, filters, did_type='collection', long=False, recursive
...
-        for k, v in list(filters.items()):
-            if k in ('created_before', 'created_after'):
-                payload[k] = date_to_str(v)
-            else:
-                payload[k] = v
-        payload['long'] = long
-        payload['type'] = did_type
-        payload['recursive'] = recursive
+        # stringify dates.
+        if isinstance(filters, dict):   # backwards compatability for filters as single {}
```
joeldierkes, line 115: A test would be good if `filters` is an array of dictionaries.

joeldierkes (changes requested):
The PR is way to big. It is a best practice to make commits rather small and descriptive with one objective. They are easier to review and less likely to introduce a bug. Feel free to open multiple PRs time after time and introduce features one after another.

Also, your commit messages are not descriptive at all. "intermediate commit" is not informative.

robbarnsley:
My mistake - this should've been a draft (hence the FIXMEs, unsquashed commits and zero tests). Marked it as such now.

Just to note, I know what best practice is - it is a little difficult for this feature because it spans multiple layers of the architecture. It would be difficult to compartmentalise.

lib/rucio/core/did_meta_plugins/__init__.py
```diff
@@ -16,186 +16,211 @@
...
+# Set default modules.
+#
+# - Hardcoded metadata is tied to column fields in the "dids" table. As such, types are restricted to column types and 
+#   additional fields cannot be created.
+# - Custom metadata refers to any plugin that has been written to interface with Rucio's metadata plugin system.
+# 
+# By default, the custom metadata plugin is set to the prepackaged json module. This utilises a separate table, 
+# "did_meta", allowing users to add custom metadata key/value pairs of any type.
+#
+HARDCODED_METADATA_PLUGIN_MODULE_PATH = "rucio.core.did_meta_plugins.did_column_meta.DidColumnMeta"
```
bari12: Maybe rename this, I know it doesn't come from you, but I was never a fan of the HARDCODED and FALLBACK naming.

bin/rucio
robbarnsley: `args.filter` defaults to None if not specified as an argument, it's then handled by the function

lib/rucio/core/did_meta_plugins/__init__.py
robbarnsley: How about 'base' and 'extra', or 'base' and 'custom'?

Side question: I presume it is critical to core functionality to have the metadata functions always look for metadata in the did columns first? Reason I ask is that, as it stands, there is no way to have a key with the same name as one in the did column store in a "custom" metadata store. 

It's also, in my opinion, not very transparent. If I was defining a list of metadata plugins in the `metadata -> plugins` section of the rucio cfg, i'd expect it to only use those plugins, and in the order I put them. That field would be better called something like `extra_plugins`.

bari12: We mostly did it this way to preserve the functionality of how Rucio was working before (--> column information is always queried first)
But I agree, this is not very intuitive/transparent. Best would probably be to just go to an ordered list, and metadata plugins are queried in that order. The rucio default should be to do column first, but it doesn't have to be like that for every community.

lib/rucio/core/did_meta_plugins/__init__.py
robbarnsley: fixme removed. added doc re: raises

lib/rucio/core/did.py
robbarnsley: added, but not here (to filter engine)

lib/rucio/core/did_meta_plugins/__init__.py
```diff
@@ -16,157 +16,194 @@
 # Authors:
 # - Aristeidis Fkiaras <aristeidis.fkiaras@cern.ch>, 2020
 # - Benedikt Ziemons <benedikt.ziemons@cern.ch>, 2020-2021
+# - David Población Criado <david.poblacion.criado@cern.ch>, 2021
+# - Rob Barnsley <robbarnsley@users.noreply.github.com>, 2021-2022
```
robbarnsley, line 20: A lot of the changes in this file are just housekeeping - changing variable names, adding comments and amending docstrings etc.

lib/rucio/core/did_meta_plugins/__init__.py
```diff
@@ -16,157 +16,194 @@
...
-        METADATA_HANDLERS.append(meta_handler_module)
+        metadata_plugin_module = getattr(importlib.import_module(base_module), base_class)()
+        METADATA_PLUGIN_MODULES.append(metadata_plugin_module)
     except ImportError:
-        raise PolicyPackageNotFound('Module ' + meta_module_path + ' not found')
+        raise exception.PolicyPackageNotFound('Module ' + meta_module_path + ' not found')
+
+# Set restricted character set for metadata in form character: reason
+#
+RESTRICTED_CHARACTERS = {
```
robbarnsley, line 76: Disallow certain characters in metadata. '.' is used as a delimiter between the key and operator.

lib/rucio/core/did_meta_plugins/did_column_meta.py
```diff
@@ -14,105 +14,44 @@
...
 from rucio.common import exception
-from rucio.common.utils import str_to_date
 from rucio.core import account_counter, rse_counter
 from rucio.core.did_meta_plugins.did_meta_plugin_interface import DidMetaPlugin
+from rucio.core.did_meta_plugins.filter_engine import FilterEngine
 from rucio.db.sqla import models
 from rucio.db.sqla.constants import DIDType
 from rucio.db.sqla.session import stream_session, read_session, transactional_session
 
-HARDCODED_KEYS = [
```
robbarnsley, line 53: Removed this in favour of getting the list directly from the model definition, with some exceptions.

lib/rucio/core/did_meta_plugins/did_column_meta.py
```diff
@@ -241,128 +180,141 @@ def list_dids(self, scope, filters, did_type='collection', ignore_case=False, li
...
+                models.DataIdentifier.scope,
+                models.DataIdentifier.name,
+                models.DataIdentifier.did_type,
+                models.DataIdentifier.bytes,
+                models.DataIdentifier.length
+            ], additional_filters=[
+                (models.DataIdentifier.scope, operator.eq, scope),
+                (models.DataIdentifier.suppressed, operator.ne, true())
+            ]
+        )
```
robbarnsley, line 230: just copying over the existing core `list_dids` function (which uses the filter engine).

lib/rucio/core/did_meta_plugins/filter_engine.py
```diff
@@ -235,72 +214,121 @@ def _translate_filters(self, model_class, strict_coerce=True):
...
+            except ValueError:
+                continue
+            else:
+                return value
+        try:
+            value = ast.literal_eval(value)     # will catch float, int and bool
+        except (ValueError, SyntaxError):
+            pass
+        return value
+
```
robbarnsley, line 250: When user inputs a metadata filter via CLI, need some way of attempting to cast the value of each key from string (important later when constructing the sqla filter expression).

lib/rucio/core/did_meta_plugins/filter_engine.py
```diff
@@ -235,72 +214,121 @@ def _translate_filters(self, model_class, strict_coerce=True):
...
+                        expression = or_(expression, key.is_(None))
+                elif json_column:                                                                   # -> this key filters on the content of a json column
+                    if session.bind.dialect.name == 'oracle':
+                        # Compiling for Oracle yields a UnsupportedCompilationError so fall back to support for equality only.
+                        # TODO
+                        if oper != operator.eq:
+                            raise exception.FilterEngineGenericError("Oracle implementation does not support this operator.")
+                        if isinstance(value, str) and any([char in value for char in ['*', '%']]):
+                            raise exception.FilterEngineGenericError("Oracle implementation does not support wildcards.")
+                        expression = text("json_exists(meta,'$?(@.{} == \"{}\")')".format(key, value))
```
robbarnsley, line 299: yet to implement this for Oracle - requires a more bespoke filter expression construction

lib/rucio/core/did_meta_plugins/filter_engine.py
```diff
@@ -317,3 +345,66 @@ def evaluate(self):
...
+                    return str(value)
+                elif isinstance(value, (str, date, datetime, timedelta)):
+                    return "'%s'" % str(value).replace("'", "''")
+                elif isinstance(value, list):
+                    return "'{%s}'" % (",".join([self.render_array_value(x, type_.item_type) for x in value]))
+                return super(LiteralCompiler, self).render_literal_value(value, type_)
+
+        return LiteralCompiler(dialect, statement).process(statement)
+
+    def print_filters(self):
```
robbarnsley, line 387: not new, moved in file

lib/rucio/core/did_meta_plugins/filter_engine.py
```diff
@@ -317,3 +345,66 @@ def evaluate(self):
...
+                if isinstance(value, sqlalchemy.orm.attributes.InstrumentedAttribute):
+                    value = and_group[2].key
+                elif isinstance(value, DIDType):
+                    value = and_group[2].name
+                filters = "{}{} {} {}".format(filters, key, oper, value)
+                if and_group != or_group[-1]:
+                    filters += ' AND '
+            if or_group != self._filters[-1]:
+                filters += ' OR\n'
+        return filters
```
robbarnsley: not new function, just moved

lib/rucio/tests/test_filter_engine.py
```diff
@@ -166,8 +186,43 @@ def test_OperatorsEqualNotEqual(self, session=None):
...
+            dids = []
+            q = FilterEngine('testkeyint1=1', model_class=models.DidMeta, strict_coerce=False).create_sqla_query(
+                additional_model_attributes=[
+                    models.DidMeta.scope,
+                    models.DidMeta.name
+                ],
+                json_column=models.DidMeta.meta)
+            dids += [did for did in q.yield_per(5)]
+            dids = set(dids)
+            self.assertEqual(list(map(lambda did: did.name in (did_name1, did_name2, did_name3), dids)).count(True), 1)
```
robbarnsley, line 208: equality supported by all sqla flavours, rest of the cases are not supported by oracle yet

robbarnsley:
I recognise this is a rather sizable PR, and there are a few housekeeping bits that could've probably been packaged as a separate request, however a lot of the changes here are all required for the JSON plugin to work with the filter engine.

From this point, the core functionality should now exist and future PRs will be smaller and more focused.

I've added a few comments that should hopefully guide a review.

lib/rucio/core/did_meta_plugins/json_meta.py
```diff
@@ -150,41 +139,77 @@ def delete_metadata(self, scope, name, key, session=None):
...
+        try:
+            for did in query.yield_per(5):                  # don't unpack this as it makes it dependent on query return order!
+                if long:
+                    did_full = "{}:{}".format(did.scope, did.name)
+                    if did_full not in ignore_dids:         # concatenating results of OR clauses may contain duplicate DIDs if query result sets not mutually exclusive.
+                        ignore_dids.add(did_full)
+                        yield {
+                            'scope': did.scope,
+                            'name': did.name,
+                            'did_type': 'Not available with {} plugin.'.format(self.get_plugin_name()),
```
bari12: I think it would be better to set these to `None`. Just to make sure this isn't accidentally processed otherwise.

lib/rucio/core/did_meta_plugins/json_meta.py
```diff
@@ -150,41 +139,77 @@ def delete_metadata(self, scope, name, key, session=None):
...
+            ], additional_filters=[
+                (models.DidMeta.scope, operator.eq, scope)
+            ],
+            json_column=models.DidMeta.meta
+        )
+
+        if limit:
+            query = query.limit(limit)
+        if recursive:
+            from rucio.core.did import list_content
```
bari12, line 169: This can't be loaded at top? Circular dependency?

bari12 (commented):
Hi @robbarnsley 
This is very nice work. Thanks a lot. 
I have two minor comments, otherwise I think this is GTG.

Maybe one recommendation for the future in terms of large PRs. It would help to separate things into different contained commits. Similar to the comments you put in the PR later.
E.g. one commit about just moving code around; One commit about housekeeping; one commit about adding a new feature
I know it's more work to do that, but makes the review much easier for large PRs like this.

lib/rucio/core/did_meta_plugins/json_meta.py
robbarnsley: Yep, no problem.

lib/rucio/core/did_meta_plugins/json_meta.py
robbarnsley: Not sure of the reasoning (someone else did this). Doesn't look circular but get a bunch of import errors when it's at the top.

robbarnsley:
> Maybe one recommendation for the future in terms of large PRs. It would help to separate things into different contained commits. Similar to the comments you put in the PR later.

Fair point, will keep this in mind.

bari12 (approved):

PR #5109, opened 2021-12-13
Add header workflow integration 2

### LINKED ISSUE #4980: Integrate the `add_header` script in the CI/CD
Motivation
----------
Since the PR #4903 the `add_header` script can check for changes and return a non-zero exit code if there would be any changes in the header comment. We can include the script in the workflow and check for inconsistent headers automatically.

Modification
------------
Run the `add_header` script with the `--dry-run` and `--check` option on changed python files in the workflow.

joeldierkes:
This commit adds a header check to the github actions.

As of right now, the headers get checked on every push. We could change that to every pull-request if required.

See individual commits for more information.

### DIFF
--- a/.github/workflows/autotest.yml
+++ b/.github/workflows/autotest.yml
@@ -5,6 +5,25 @@ on:
   - push
 
 jobs:
+  header-lint:
+    if: 'github.event_name == ''pull_request'''
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Check headers
+        shell: bash
+        run: |
+          git pull origin master
+          echo "Current git rev $(git rev-parse HEAD)"
+          echo "Common ancestor git rev $(git merge-base HEAD remotes/origin/master)"
+          echo "If a file header needs to be changed run 'tools/add_header -i FILE' from the rucio project directory."
+          FILES=$(git diff --name-status HEAD $(git merge-base HEAD remotes/origin/master) | grep -v '^A' | cut -f 2 | grep -E '\.py$|^bin/') || echo "No Files found"
+          if [[ -n $FILES ]]; then
+            echo "Files to check: $FILES"
+            tools/add_header --disable-add-me -cd $FILES
+          fi
   setup:
     runs-on: ubuntu-latest
     steps:
--- a/tools/add_header
+++ b/tools/add_header
@@ -95,6 +95,7 @@ def main(arguments):
     email = '<%s>' % email
     exit_code = 0
 
+    print("Checking %i files" % (len(arguments.MyFiles)))
     for MyFile in arguments.MyFiles:
         # Query log history
         cmd = '''git log --reverse --date=short --format='%s'  %s ''' % ('%aN,<%aE>,%ad', MyFile)
@@ -244,7 +245,7 @@ def main(arguments):
                         if count == 0 or not_comment_line:
                             modified.write(line)
 
-        else:
+        elif not arguments.check:
             print(header)
     exit(exit_code)
 


.github/workflows/autotest.yml
```diff
@@ -5,7 +5,16 @@ on:
   - push
 
 jobs:
+  add_header_test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
```
maany, line 14: I doubt if we need to pull history for all branches/tags for updating headers

joeldierkes: I don't have much experience with the GitHub `actions/checkout`. 😅 

We at least need the entire history before the commit, since the `add_header` script uses it.

bin/rucio-conveyor-fts-throttler
```diff
@@ -15,7 +15,11 @@
 # limitations under the License.
 #
 # Authors:
-# - Dilaksun Bavarajan, <dilaksun@hotmail.com>, 2019
+# - Dilaksun Bavarajan <dilaksun.bavarajan@cern.ch>, 2019
+# - Martin Barisits <martin.barisits@cern.ch>, 2019
+# - Benedikt Ziemons <benedikt.ziemons@cern.ch>, 2020
+# - Ocean Monjur <75680423+OCM-7898@users.noreply.github.com>, 2021
+# - jdierkes <joel.dierkes@cern.ch>, 2021
 # - Joel Dierkes <joel.dierkes@cern.ch>, 2021
```
maany: maybe stick with a single alias ;)

bin/rucio-hermes
```diff
@@ -20,6 +20,7 @@
 # - Hannes Hansen <hannes.jakob.hansen@cern.ch>, 2018
 # - Benedikt Ziemons <benedikt.ziemons@cern.ch>, 2020
 # - David Población Criado <david.poblacion.criado@cern.ch>, 2021
+# - Joel Dierkes <joel.dierkes@cern.ch>, 2021
```
maany: This PR has a lot of files that are changed and from a quick overview, I see that a lot of these changes include @joeldierkes as an author, like here. Which makes sense as the `add-header.py` is essentially looking at the commit history. I believe you can run the `add-header.py` and offset the git log to `HEAD~1` in order to keep the header consistent with the code contributors. ;)  Maybe it also brings down the number of files that need to be reviewed here significantly.

maany (commented):
Besides the comments I have already left, I noticed that some of the emails in the `add-header.py`:
https://github.com/rucio/rucio/blob/d59c237f533f40116026dc9f347f4fc1297f1ff0/tools/add_header#L52 
https://github.com/rucio/rucio/blob/d59c237f533f40116026dc9f347f4fc1297f1ff0/tools/add_header#L74
look weird/temporary. I guess not all commits by these authors come from the same email address. Are we sure that we should be using these email addresses? WDYT @rcarpa @bari12

bin/rucio-necromancer
```diff
@@ -22,6 +22,7 @@
 # - Benedikt Ziemons <benedikt.ziemons@cern.ch>, 2020
 # - Ocean Monjur <75680423+OCM-7898@users.noreply.github.com>, 2021
```
maany: This comment is not for you Joel. I wonder if we should be including author emails such as this ^ in the headers.

.github/workflows/autotest.yml
maany: Ah, I see. Correct me if I am wrong, this would mean that if I send a PR modifying, let's say, file a, b and c. The `add_header` script would still parse the entire rucio commit history and update all files, not just a,b and c?

bin/rucio-hermes
joeldierkes: Yeah my bad, I forgot to use the already implemented `--disable-add-me` argument.

rcarpa: Maybe it's better to have a "lazy" approach to enforcing header validation? Instead of verifying the whole project's headers, only run the verification on files which are modified by the Push ? 
It avoids this "fix everything commit"; will gradually improve the headers during small commits. 

Just an idea, how it's done righ now is ok for me too.

.github/workflows/autotest.yml
```diff
@@ -5,7 +5,16 @@ on:
...
+      - run: |
+          tools/add_header --disable-add-me -cd -- bin/* $(find . -type f -name '*.py')
```
rcarpa: the max length of a shell command is limited. This commands risks to fail with the growth of the project.

bin/rucio-hermes
joeldierkes: I tried that at first, but there are some limitations. I ran into the same problems as this Community post: https://github.community/t/can-i-process-only-changed-files-with-github-actions/137814/6 (under the __`//edit`__ part with the rewrite commits). At that moment, I decided to switch the strategy, since I didn't want to waste more time.

I talked with Mayank about it and we both decided, that this approach is fine, too.

rcarpa: Thank you for the link. As I said, I'm ok to do it this way.

joeldierkes: Okay perfect, I'll create a second version to Review then.

Also, Mayank suggested to only check the changes on a PR, and not on every push. I will implement that too.

bin/rucio-conveyor-fts-throttler
joeldierkes: Mentioned in #5114

.github/workflows/autotest.yml
joeldierkes: Yeah, totally forgot about that. We currently are at about 0.5% of the maximum line length. Changed it never the less.

bin/rucio-necromancer
joeldierkes: #5114

bari12:
Is it strictly needed to re-touch every file (422 changes) or can't we just get this up to date over time?

joeldierkes:
@maany and I discussed that, and the problem is that you have to get the diff between the PR and the master branch. I tried that first, but the commit SHA provided by GitHub Actions is not working properly when a commit is overwritten via a `push --force`. I also could not find a plugin suitable for that case.

I also don't know anyone who has experience with that particular issue...

joeldierkes:
Okay, after @bari12 gave some hints on how to get the changed files in GItHub Actions the implementation now only checks the headers of the changed python files.

.github/workflows/autotest.yml
```diff
@@ -5,7 +5,25 @@ on:
...
+          git pull origin master
+          echo "Current git rev $(git rev-parse HEAD)"
+          echo "Git rev $(git merge-base HEAD remotes/origin/master)"
+          echo "If a file header needs to be changed run "tools/add_header -i FILE" from the rucio project directory."
+          for file in $(git diff --name-status HEAD $(git merge-base HEAD remotes/origin/master) | grep -v '^A' | cut -f 2 | grep -E '\.py$|^bin/'); do
+            echo "Processing $file"
+            tools/add_header --disable-add-me -cd $file
+          done
   setup:
+    if: ${{ success() || cancelled() }}
```
maany: Umm, why is `success() || cancelled()`required? I assume that it is equivalent to `jobs.header-lint.status`, in which case, we can mention that explicitly ;) 

In the current setup, if `header-lint` job fails, the tests will not run at all. 

I guess for both `setup` and `release-patch-setup` jobs we can remove the `success() or failfure()` and `needs: header-lint` parts.

I'd probably let the test suite run even if the `header-lint` job fails. ;)

.github/workflows/autotest.yml
```diff
@@ -5,7 +5,25 @@ on:
...
 jobs:
+  header-lint:
+    if: 'github.event_name == ''pull_request'''
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - run: |
+          git pull origin master
```
maany, line 18: I guess this should not be required since you use `actions/checkout@v2` with a `fetch-depth: 0`. Could you double check this? In the containers repo, I was able to get the revision for the remotes/origin/master as follows, without the need to do a git pull:

https://github.com/rucio/containers/blob/b09dbd05b57ae3dc5774e09fa00dfb79529a6d9b/.github/workflows/docker-auto-build.yml#L102-L103


```
GIT_MASTER_LATEST_COMMIT_SHA=$(echo $(git ls-remote --heads origin | grep master  |  awk -F '\t' '{print $1}'))
```

.github/workflows/autotest.yml
```diff
@@ -5,7 +5,25 @@ on:
...
+          echo "Current git rev $(git rev-parse HEAD)"
+          echo "Git rev $(git merge-base HEAD remotes/origin/master)"
```
maany: Also, minor change, but let's rename this to 
```
echo "Common Ancestor git rev....
```

maany (changes requested):
Nice work @joeldierkes ! Awesome to see we don't need to parse all the files on every pull request. I've left some minor comments.  ;)

.github/workflows/autotest.yml
maany: So I'd assume, you would still get the same output from `$(git merge-base HEAD remotes/origin/master)` if you remove the `git pull`

.github/workflows/autotest.yml
joeldierkes: This got resolved, since the PR now only processes the changed files.

.github/workflows/autotest.yml
joeldierkes: I see a logical flaw in it:
The developer pushed some changes. If the headers are not correct, the tests will run nevertheless. However, since the tests are not correct, the developer has to update the headers and push the changes again. This means, that the tests will run twice with essentially the same code (just the headers changed). Do we want that?

In the current case, the tests early abort if the headers are not correct, this normally happens after 1 minute or less.

.github/workflows/autotest.yml
joeldierkes: We don't need the entire history for the changes, we need it for the `add_header` script.
https://github.com/rucio/rucio/blob/6d0d358e04f5431f0b9a98ae40f31af0ddff4833/tools/add_header#L100

I tried it without the git pull and then the script could not find the current master branch commit sha...

.github/workflows/autotest.yml
joeldierkes: Good idea! I'll change it as soon as we know how to deal with the other changes! 👍

rcarpa:
I really like the new version. Nice job!

.github/workflows/autotest.yml
maany: In the spirit of discovering all flaws with a PR early, we'd definitely want that ;) 

Diving into some details now,

There will be a new entry in every PR which would look like this:
```
Run all the autotests / header-lint (pull_request)
```
regardless of whether we choose to make `header-lint` a hard dependency on the following jobs (setup, test, ..). So, if the linting fails, the CI will not be in a 'green, ready to merge' state in either case.
I was thinking on the lines of exposing all the issues with a PR upfront i.e header lints and failing tests.

Let's consider the following cases:
1) headers are incorrect and tests are failing
If `header-lint` is not a hard dependency, the developer will be made aware of both the incorrect headers and the failing tests in the same GitHub Action run. You generally see all the jobs and success/failure statistics in the PR.
So the developer has a chance to fix both headers and failed tests in the next push. 
Otherwise, he/she would first fix the headers and pushes the code, then lets the tests run and discover the failing tests and then deal with the failing tests.

2) headers are correct and tests are failing
No problem here in either case i.e if header-lint is a hard dependency or not. 

3) headers are incorrect and tests are passing
In this case, the PR would show `Some checks were not successful` indicating that header-linting failed.

4) headers are correct and tests are passing
All good here in either case.

So I vote for keeping `header-lint` as an independent check for PRs, to allow discovery of test errors early in the PR pipeline.

Let me know what you think ;) 

---
Whether or not we choose to keep the dependency, the following line is redundant
```
 if: ${{ success() || cancelled() }}
```
since it is already accommodated by
```
needs: header-lint
```
Since debugging Github Actions can be a pain, I have simulated a simple example to demo this:
See the code change:
https://github.com/maany/rucio/commit/494db8fd824fc29db51b574ec9f92aa5844cceb2 
and the corresponding run:
https://github.com/maany/rucio/actions/runs/1707477530 

The run fails if `header-lint` fails despite dropping the `if` condition.


-----

.github/workflows/autotest.yml
maany: Interesting, let's keep it ;)

.github/workflows/autotest.yml
maany: Also, another minor one,  we should add a `name` field to this step, otherwise, the runs will display the `git pull origin master` as their name :P 

So, something like
```
- name: Check headers
  run: |
          git pull origin master
```
should be good to go :D

.github/workflows/autotest.yml
joeldierkes: Okay, I'll change it. For me, there is no strong reason to make it a dependency.

FYI The
```
 if: ${{ success() || cancelled() }}
```
is mandatory, since the `header-lint` step has the condition
```
if: 'github.event_name == ''pull_request'''
```
. So if the action run is not a PR (e.g. a normal commit) the `header-lint` gets canceled and thus the tests wouldn't have run too.

Since we don't want it to be a dependency any more it is not important.

maany: > ```
>  if: ${{ success() || cancelled() }}
> ```
> 
> 
>     
>   
> 
> is mandatory, since the `header-lint` step has the condition
> 
> 
> ```
> if: 'github.event_name == ''pull_request'''
> ```

Aaahh right! maketh sense ;)

.github/workflows/autotest.yml
joeldierkes: Changed

maany (approved):
LGTM

PR #5112, opened 2021-12-14
Documentation update for replicaclient.py

flgomezc:
I have added an example of replica format to set_tombstone


bari12:
Can you please re-submit this. The `old-doc` branch should not be used, instead make this submission just against `master`. I tried to change the base, but it has conflicts now. Thanks!

Issue #4918, opened 2021-10-19
Load policy packages in client


jamesp-epcc:
Motivation
----------
It would be useful to have the configurability provided by policy packages available in the Rucio client as well as on the server. Initially, this would allow the Belle II specific scope extraction code to be moved out of `downloadclient.py` and into a policy package. It is also required for the Rucio-DIRAC integration.

Modification
------------
The client should be modified to load selected modules from the policy package, if one is specified. (Other modules, such as permission modules, will not be relevant for the client). The policy package can be specified in the Rucio config file but we should also allow it to be specified in an environment variable as some users would prefer not to use a config file.

I also intend to make the registration of different types of algorithms more generic, as discussed with Martin. Currently policy packages can register SURL and LFN2PFN algorithms and we also want to add scope extraction algorithms. As much code as possible should be shared between these use cases. Rather than having a separate function to return a dictionary for each type of algorithm as at present, a single data structure could be made available by the policy package, containing all of the algorithms and their types.

PR #5118, opened 2021-12-17
Clients: allow policy packages on client #4918

jamesp-epcc:
This PR improves policy package support to allow policy packages to load on the client. It also makes algorithm registration in policy packages more generic and allows scope extraction algorithms to be registered from policy packages. To avoid the need for a config file, the policy package can be specified via an environment variable.

One of the reasons for this was to get rid of the Belle II specific code in `downloadclient.py`. I have made the removal of the trailing slash generic, since this is something that could potentially affect any experiment. However, I have left the Belle II specific scope extraction code for the moment, since removing it could break things for Belle II. Once Belle II have their scope extraction algorithm available in a policy package, it should be possible to remove this.

### DIFF
--- a/lib/rucio/client/downloadclient.py
+++ b/lib/rucio/client/downloadclient.py
@@ -242,7 +242,7 @@ class DownloadClient:
             item['name'] = did_name
             item['sources'] = [{'pfn': pfn, 'rse': rse}]
             did_path_name = did_name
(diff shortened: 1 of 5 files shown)


bari12:
I think this is good to go. 
@cserf can you please have a look if this is OK for Belle II, I spot some changes there but I think it should be fine.

bari12 (approved):

PR #5120, opened 2021-12-17
Consistency checks: add storage-consistency-actions #5113

### LINKED ISSUE #5113: Create CMS-style consistency checking daemon
Motivation
----------
It has been known for a while that the ATLAS-style storage consistency-checking approach does not work for the case of CMS storage. Proposed feature provides an alternative approach, based on remote scanning of sites (RSEs) over XRootD/AAA, comparing the results of those scans with the contents of the Rucio database, and identifying 'dark' and 'missing' files. 

Proposed daemon implements the 'actions' part of this process: deleting confirmed dark files, and re-transferring (copying) missing files from other existing replicas.

Modification
------------
A new Rucio daemon will be created, which will read the results of the scanner, and will interact with the existing Rucio daemons to request the remote deletion of the 'dark' files from the RSEs storage system, or the re-transfer of the missing files to the RSE.

Proposed name of the new daemon is - in compliance with the new naming conventions:
'storage-consistency-actions'

piperov:

### DIFF
--- a/bin/rucio-storage-consistency-actions
+++ b/bin/rucio-storage-consistency-actions
@@ -16,6 +16,7 @@ import signal
 
 from rucio.daemons.storage.consistency.actions import run, stop
 
+
 def get_parser():
     """
     Returns the argparse parser.
@@ -26,25 +27,24 @@ Run the daemon::
     ''')
     parser.add_argument("--run-once", action="store_true", default=False, help='One iteration only')
     parser.add_argument("--rses", nargs='+', default=[], type=str,
-        help='RSEs to check, specified as a RSE expression. Defaults to check all RSEs.')
+                        help='RSEs to check, specified as a RSE expression. Defaults to check all RSEs.')
     parser.add_argument('--sleep-time', action="store", default=60, type=int,
-        help='Concurrency control: thread sleep time (in seconds) after each chunk of work')
+                        help='Concurrency control: thread sleep time (in seconds) after each chunk of work')
     parser.add_argument('--dark-min-age', action="store", default=28, type=int,
-        help='Min. age (in days) of a file to be considered as DARK.')
+                        help='Min. age (in days) of a file to be considered as DARK.')
     parser.add_argument('--dark-threshold-percent', action="store", default=1.0,
-        type=float, help='Max. percentage of dark files at RSE, expressed as percents - e.g. 1.5 means 1.5%')
+                        type=float, help='Max. percentage of dark files at RSE, expressed as percents - e.g. 1.5 means 1.5%%')
     parser.add_argument('--miss-threshold-percent', action="store", default=1.0,
-        type=float, help='Max. percentage of missing files at RSE, expressed as percents - e.g. 1.5 means 1.5%')
+                        type=float, help='Max. percentage of missing files at RSE, expressed as percents - e.g. 1.5 means 1.5%%')
     parser.add_argument('--scanner-files-path', action="store",
-        default="/var/cache/consistency-dump", type=pathlib.Path,
-        help='The path where the CC scanner files are mounted.')
+                        default="/var/cache/consistency-dump", type=pathlib.Path,
+                        help='The path where the CC scanner files are mounted.')
     parser.add_argument("--threads", action="store", default=1, type=int,
-        help='Concurrency control: total number of threads on this process')
+                        help='Concurrency control: total number of threads on this process')
     parser.add_argument('-f', '--force-proceed', action="store_true", default=False,
-        help='Force CC actions, even if number of dark/missing files over thresholds')
+                        help='Force CC actions, even if number of dark/missing files over thresholds')
 
     return parser
-
 
 
 if __name__ == "__main__":
--- a/bin/rucio-storage-consistency-actions
+++ b/bin/rucio-storage-consistency-actions
@@ -21,21 +21,20 @@ def get_parser():
     Returns the argparse parser.
     """
     parser = argparse.ArgumentParser(description="The Consistency-Actions daemon is responsible for applying the corrective actions resulting from a consistency-check scan of an RSE.", epilog='''
-
 Run the daemon::
   $ rucio-storage-consistency-actions --run-once --rses T2_US_Purdue T2_US_Nebraska --dark-threshold-percent 2.0 --miss-threshold-percent 1.5 --scanner-files-path /tmp/consistency-dump --sleep-time 10
     ''')
     parser.add_argument("--run-once", action="store_true", default=False, help='One iteration only')
     parser.add_argument("--rses", nargs='+', default=[], type=str,
-        help='RSEs to check specified as a RSE expression, defaults to check (default: check all RSEs).')
+        help='RSEs to check, specified as a RSE expression. Defaults to check all RSEs.')
     parser.add_argument('--sleep-time', action="store", default=60, type=int,
-        help='Concurrency control: thread sleep time after each chunk of work')
+        help='Concurrency control: thread sleep time (in seconds) after each chunk of work')
     parser.add_argument('--dark-min-age', action="store", default=28, type=int,
-        help='Min. age of a file to be considered as DARK.')
+        help='Min. age (in days) of a file to be considered as DARK.')
     parser.add_argument('--dark-threshold-percent', action="store", default=1.0,
-        type=float, help='Max. percentage of dark files at RSE.')
+        type=float, help='Max. percentage of dark files at RSE, expressed as percents - e.g. 1.5 means 1.5%')
     parser.add_argument('--miss-threshold-percent', action="store", default=1.0,
-        type=float, help='Max. percentage of missing files at RSE.')
+        type=float, help='Max. percentage of missing files at RSE, expressed as percents - e.g. 1.5 means 1.5%')
     parser.add_argument('--scanner-files-path', action="store",
         default="/var/cache/consistency-dump", type=pathlib.Path,
         help='The path where the CC scanner files are mounted.')
--- a/bin/rucio-storage-consistency-actions
+++ b/bin/rucio-storage-consistency-actions
@@ -1,0 +1,53 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+#
+# Authors:
+# - Stefan Piperov <stefan.piperov@cern.ch>, 2021
+
+"""
+Storage-Consistency-Actions is a daemon to perform the deletion of dark files and re-transfer missing files at sites identified by the Storage-Consistency-Scanner.
+"""
+
+import time
+import argparse
+import pathlib
+import signal
+
+from rucio.core.rse import list_rses
+
+RSEs = [rse['rse'] for rse in list_rses({'availability_write': True})]
+
+from rucio.daemons.storage.consistency.actions import run, stop
+
+def get_parser():
+    """
+    Returns the argparse parser.
+    """
+    parser = argparse.ArgumentParser(description="The Consistency-Actions daemon is responsible for... .", epilog='''
+
+Run the daemon::
+  $ rucio-storage-consistency-actions --run-once --rses T2_US_Purdue T2_US_Nebraska --dark-threshold-percent 2.0 --miss-threshold-percent 1.5 --scanner-files-path /tmp/consistency-dump --sleep-time 10
+    ''')
+    parser.add_argument("--run-once", action="store_true", default=False, help='One iteration only')
+    parser.add_argument("--rses", nargs='+', default=[], type=str, help='RSEs to check specified as a RSE expression, defaults to check (default: check all RSEs).')
+    parser.add_argument('--sleep-time', action="store", default=60, type=int, help='Concurrency control: thread sleep time after each chunk of work')
+    parser.add_argument('--dark-min-age', action="store", default=28, type=int, help='Min. age of a file to be considered as DARK.')
+    parser.add_argument('--dark-threshold-percent', action="store", default=1.0, type=float, help='Max. percentage of dark files at RSE.')
+    parser.add_argument('--miss-threshold-percent', action="store", default=1.0, type=float, help='Max. percentage of missing files at RSE.')
+    parser.add_argument('--scanner-files-path', action="store", default="/var/cache/consistency-dump", type=pathlib.Path, help='The path where the CC scanner files are mounted.')
+    parser.add_argument("--threads", action="store", default=1, type=int, help='Concurrency control: total number of threads on this process')
+    parser.add_argument('-f', '--force-proceed', action="store_true", default=False, help='Force CC actions, even if number of dark/missing files over thresholds')
+
+    return parser
+
+
+
+if __name__ == "__main__":
+    signal.signal(signal.SIGTERM, stop)
+    parser = get_parser()
+    args = parser.parse_args()
+    try:
+        run(once=args.run_once, rses=args.rses, sleep_time=args.sleep_time, dark_min_age=args.dark_min_age, dark_threshold_percent=args.dark_threshold_percent, miss_threshold_percent=args.miss_threshold_percent, force_proceed=args.force_proceed, scanner_files_path=args.scanner_files_path,threads=args.threads)
+    except KeyboardInterrupt:
+        stop()
+
--- a/bin/rucio-storage-consistency-actions
+++ b/bin/rucio-storage-consistency-actions
@@ -5,9 +5,9 @@
 # - Stefan Piperov <stefan.piperov@cern.ch>, 2021
 
 """
-Storage-Consistency-Actions is a daemon to perform the
+Storage-Consistency-Actions is a daemon to trigger the
 deletion of dark files and re-transfer missing files at
-sites identified by the Storage-Consistency-Scanner.
+sites, as identified by the Storage-Consistency-Scanner.
 """
 
 import argparse
@@ -26,23 +26,22 @@ Run the daemon::
   $ rucio-storage-consistency-actions --run-once --rses T2_US_Purdue T2_US_Nebraska --dark-threshold-percent 2.0 --miss-threshold-percent 1.5 --scanner-files-path /tmp/consistency-dump --sleep-time 10
     ''')
     parser.add_argument("--run-once", action="store_true", default=False, help='One iteration only')
-    parser.add_argument("--rses", nargs='+', default=[], type=str,\
-        help='RSEs to check specified as a RSE expression,\
-        defaults to check (default: check all RSEs).')
-    parser.add_argument('--sleep-time', action="store", default=60, type=int,\
+    parser.add_argument("--rses", nargs='+', default=[], type=str,
+        help='RSEs to check specified as a RSE expression, defaults to check (default: check all RSEs).')
+    parser.add_argument('--sleep-time', action="store", default=60, type=int,
         help='Concurrency control: thread sleep time after each chunk of work')
-    parser.add_argument('--dark-min-age', action="store", default=28, type=int,\
+    parser.add_argument('--dark-min-age', action="store", default=28, type=int,
         help='Min. age of a file to be considered as DARK.')
-    parser.add_argument('--dark-threshold-percent', action="store", default=1.0,\
+    parser.add_argument('--dark-threshold-percent', action="store", default=1.0,
         type=float, help='Max. percentage of dark files at RSE.')
-    parser.add_argument('--miss-threshold-percent', action="store", default=1.0,\
+    parser.add_argument('--miss-threshold-percent', action="store", default=1.0,
         type=float, help='Max. percentage of missing files at RSE.')
-    parser.add_argument('--scanner-files-path', action="store",\
-        default="/var/cache/consistency-dump", type=pathlib.Path,\
+    parser.add_argument('--scanner-files-path', action="store",
+        default="/var/cache/consistency-dump", type=pathlib.Path,
         help='The path where the CC scanner files are mounted.')
-    parser.add_argument("--threads", action="store", default=1, type=int,\
+    parser.add_argument("--threads", action="store", default=1, type=int,
         help='Concurrency control: total number of threads on this process')
-    parser.add_argument('-f', '--force-proceed', action="store_true", default=False,\
+    parser.add_argument('-f', '--force-proceed', action="store_true", default=False,
         help='Force CC actions, even if number of dark/missing files over thresholds')
 
     return parser
@@ -54,9 +53,12 @@ if __name__ == "__main__":
     parser = get_parser()
     args = parser.parse_args()
     try:
-        run(once=args.run_once, rses=args.rses, sleep_time=args.sleep_time,\
-            dark_min_age=args.dark_min_age, dark_threshold_percent=args.dark_threshold_percent,\
-            miss_threshold_percent=args.miss_threshold_percent, force_proceed=args.force_proceed,\
-            scanner_files_path=args.scanner_files_path, threads=args.threads)
+        run(once=args.run_once, rses=args.rses, sleep_time=args.sleep_time,
+            default_dark_min_age=args.dark_min_age,
+            default_dark_threshold_percent=args.dark_threshold_percent,
+            default_miss_threshold_percent=args.miss_threshold_percent,
+            force_proceed=args.force_proceed,
+            default_scanner_files_path=args.scanner_files_path,
+            threads=args.threads)
     except KeyboardInterrupt:
         stop()
--- a/bin/rucio-storage-consistency-actions
+++ b/bin/rucio-storage-consistency-actions
@@ -1,8 +1,21 @@
 #!/usr/bin/env python3
 # -*- coding: utf-8 -*-
+# Copyright 2021-2022 CERN
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
 #
 # Authors:
-# - Stefan Piperov <stefan.piperov@cern.ch>, 2021
+# - Stefan Piperov <piperov@cern.ch>, 2021-2022
 
 """
 Storage-Consistency-Actions is a daemon to trigger the
--- a/bin/rucio-storage-consistency-actions
+++ b/bin/rucio-storage-consistency-actions
@@ -5,17 +5,14 @@
 # - Stefan Piperov <stefan.piperov@cern.ch>, 2021
 
 """
-Storage-Consistency-Actions is a daemon to perform the deletion of dark files and re-transfer missing files at sites identified by the Storage-Consistency-Scanner.
+Storage-Consistency-Actions is a daemon to perform the
+deletion of dark files and re-transfer missing files at
+sites identified by the Storage-Consistency-Scanner.
 """
 
-import time
 import argparse
 import pathlib
 import signal
-
-from rucio.core.rse import list_rses
-
-RSEs = [rse['rse'] for rse in list_rses({'availability_write': True})]
 
 from rucio.daemons.storage.consistency.actions import run, stop
 
@@ -23,20 +20,30 @@ def get_parser():
     """
     Returns the argparse parser.
     """
-    parser = argparse.ArgumentParser(description="The Consistency-Actions daemon is responsible for... .", epilog='''
+    parser = argparse.ArgumentParser(description="The Consistency-Actions daemon is responsible for applying the corrective actions resulting from a consistency-check scan of an RSE.", epilog='''
 
 Run the daemon::
   $ rucio-storage-consistency-actions --run-once --rses T2_US_Purdue T2_US_Nebraska --dark-threshold-percent 2.0 --miss-threshold-percent 1.5 --scanner-files-path /tmp/consistency-dump --sleep-time 10
     ''')
     parser.add_argument("--run-once", action="store_true", default=False, help='One iteration only')
-    parser.add_argument("--rses", nargs='+', default=[], type=str, help='RSEs to check specified as a RSE expression, defaults to check (default: check all RSEs).')
-    parser.add_argument('--sleep-time', action="store", default=60, type=int, help='Concurrency control: thread sleep time after each chunk of work')
-    parser.add_argument('--dark-min-age', action="store", default=28, type=int, help='Min. age of a file to be considered as DARK.')
-    parser.add_argument('--dark-threshold-percent', action="store", default=1.0, type=float, help='Max. percentage of dark files at RSE.')
-    parser.add_argument('--miss-threshold-percent', action="store", default=1.0, type=float, help='Max. percentage of missing files at RSE.')
-    parser.add_argument('--scanner-files-path', action="store", default="/var/cache/consistency-dump", type=pathlib.Path, help='The path where the CC scanner files are mounted.')
-    parser.add_argument("--threads", action="store", default=1, type=int, help='Concurrency control: total number of threads on this process')
-    parser.add_argument('-f', '--force-proceed', action="store_true", default=False, help='Force CC actions, even if number of dark/missing files over thresholds')
+    parser.add_argument("--rses", nargs='+', default=[], type=str,\
+        help='RSEs to check specified as a RSE expression,\
+        defaults to check (default: check all RSEs).')
+    parser.add_argument('--sleep-time', action="store", default=60, type=int,\
+        help='Concurrency control: thread sleep time after each chunk of work')
+    parser.add_argument('--dark-min-age', action="store", default=28, type=int,\
+        help='Min. age of a file to be considered as DARK.')
+    parser.add_argument('--dark-threshold-percent', action="store", default=1.0,\
+        type=float, help='Max. percentage of dark files at RSE.')
+    parser.add_argument('--miss-threshold-percent', action="store", default=1.0,\
+        type=float, help='Max. percentage of missing files at RSE.')
+    parser.add_argument('--scanner-files-path', action="store",\
+        default="/var/cache/consistency-dump", type=pathlib.Path,\
+        help='The path where the CC scanner files are mounted.')
+    parser.add_argument("--threads", action="store", default=1, type=int,\
+        help='Concurrency control: total number of threads on this process')
+    parser.add_argument('-f', '--force-proceed', action="store_true", default=False,\
+        help='Force CC actions, even if number of dark/missing files over thresholds')
 
     return parser
 
@@ -47,7 +54,9 @@ if __name__ == "__main__":
     parser = get_parser()
     args = parser.parse_args()
     try:
-        run(once=args.run_once, rses=args.rses, sleep_time=args.sleep_time, dark_min_age=args.dark_min_age, dark_threshold_percent=args.dark_threshold_percent, miss_threshold_percent=args.miss_threshold_percent, force_proceed=args.force_proceed, scanner_files_path=args.scanner_files_path,threads=args.threads)
+        run(once=args.run_once, rses=args.rses, sleep_time=args.sleep_time,\
+            dark_min_age=args.dark_min_age, dark_threshold_percent=args.dark_threshold_percent,\
+            miss_threshold_percent=args.miss_threshold_percent, force_proceed=args.force_proceed,\
+            scanner_files_path=args.scanner_files_path, threads=args.threads)
     except KeyboardInterrupt:
         stop()
-
--- a/bin/rucio-storage-consistency-actions
+++ b/bin/rucio-storage-consistency-actions
@@ -1,21 +1,8 @@
 #!/usr/bin/env python3
 # -*- coding: utf-8 -*-
-# Copyright 2021-2022 CERN
-#
-# Licensed under the Apache License, Version 2.0 (the "License");
-# you may not use this file except in compliance with the License.
-# You may obtain a copy of the License at
-#
-#    http://www.apache.org/licenses/LICENSE-2.0
-#
-# Unless required by applicable law or agreed to in writing, software
-# distributed under the License is distributed on an "AS IS" BASIS,
-# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-# See the License for the specific language governing permissions and
-# limitations under the License.
 #
 # Authors:
-# - Stefan Piperov <piperov@cern.ch>, 2021-2022
+# - Stefan Piperov <stefan.piperov@cern.ch>, 2021
 
 """
 Storage-Consistency-Actions is a daemon to trigger the
@@ -36,9 +23,11 @@ def get_parser():
     """
     parser = argparse.ArgumentParser(description="The Consistency-Actions daemon is responsible for applying the corrective actions resulting from a consistency-check scan of an RSE.", epilog='''
 Run the daemon::
-  $ rucio-storage-consistency-actions --run-once --rses T2_US_Purdue T2_US_Nebraska --dark-threshold-percent 2.0 --miss-threshold-percent 1.5 --scanner-files-path /tmp/consistency-dump --sleep-time 10
+  $ rucio-storage-consistency-actions --run-once --scope cms --rses T2_US_Purdue T2_US_Nebraska --dark-threshold-percent 2.0 --miss-threshold-percent 1.5 --scanner-files-path /tmp/consistency-dump --sleep-time 10
     ''')
     parser.add_argument("--run-once", action="store_true", default=False, help='One iteration only')
+    parser.add_argument("--scope", action="store", default=False, type=str,
+                        help='Scope of the input files produced by theiCC scanner - e.g.: "cms" ')
     parser.add_argument("--rses", nargs='+', default=[], type=str,
                         help='RSEs to check, specified as a RSE expression. Defaults to check all RSEs.')
     parser.add_argument('--sleep-time', action="store", default=60, type=int,
@@ -65,7 +54,7 @@ if __name__ == "__main__":
     parser = get_parser()
     args = parser.parse_args()
     try:
-        run(once=args.run_once, rses=args.rses, sleep_time=args.sleep_time,
+        run(once=args.run_once, scope=args.scope, rses=args.rses, sleep_time=args.sleep_time,
             default_dark_min_age=args.dark_min_age,
             default_dark_threshold_percent=args.dark_threshold_percent,
             default_miss_threshold_percent=args.miss_threshold_percent,
--- a/lib/rucio/daemons/storage/consistency/actions.py
+++ b/lib/rucio/daemons/storage/consistency/actions.py
@@ -1,0 +1,765 @@
+# -*- coding: utf-8 -*-
+# Copyright 2013-2020 CERN
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Authors:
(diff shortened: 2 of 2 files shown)


bin/rucio-storage-consistency-actions
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+#
+# Authors:
+# - Stefan Piperov <stefan.piperov@cern.ch>, 2021
+
+"""
+Storage-Consistency-Actions is a daemon to perform the
+deletion of dark files and re-transfer missing files at
```
ericvaandering, line 22: I'd say "trigger for deletion and re-transfer" since that's more accurate

bin/rucio-storage-consistency-actions
```diff
@@ -0,0 +1,62 @@
...
+    """
+    Returns the argparse parser.
+    """
+    parser = argparse.ArgumentParser(description="The Consistency-Actions daemon is responsible for applying the corrective actions resulting from a consistency-check scan of an RSE.", epilog='''
+
+Run the daemon::
+  $ rucio-storage-consistency-actions --run-once --rses T2_US_Purdue T2_US_Nebraska --dark-threshold-percent 2.0 --miss-threshold-percent 1.5 --scanner-files-path /tmp/consistency-dump --sleep-time 10
+    ''')
+    parser.add_argument("--run-once", action="store_true", default=False, help='One iteration only')
+    parser.add_argument("--rses", nargs='+', default=[], type=str,\
```
ericvaandering: \ continuation marks are not needed when there is an open parenthesis and are no longer recommended.

bin/rucio-storage-consistency-actions
```diff
@@ -0,0 +1,62 @@
...
+
+
+
+if __name__ == "__main__":
+    signal.signal(signal.SIGTERM, stop)
+    parser = get_parser()
+    args = parser.parse_args()
+    try:
+        run(once=args.run_once, rses=args.rses, sleep_time=args.sleep_time,\
+            dark_min_age=args.dark_min_age, dark_threshold_percent=args.dark_threshold_percent,\
```
ericvaandering: For all these min/max arguments, they will eventually be defaults which can be overridden by RSE specific settings. So name them defaults now.

lib/rucio/daemons/storage/consistency/actions.py
```diff
@@ -0,0 +1,771 @@
...
+        raise exception.RucioException(error.args)
+
+    return unknown_replicas
+### NOTE: declare_bad_file_replicas will be used directly from core/replica.py
+### when handling of DID is added there
+##########################################################################
+
+
+##############################
+### This is Igor's Stats class
```
ericvaandering: Why not break this out and put it into rucio/lib/rucio/common/? Add @ivmfnal as an author.

lib/rucio/daemons/storage/consistency/actions.py
```diff
@@ -0,0 +1,771 @@
...
+def list_cc_scanned_rses(Path):
+    files = glob.glob(f"{Path}/*_stats.json")
+    rses = set()
+    for path in files:
+        fn = path.rsplit("/", 1)[-1]
+        rse, timestamp, typ, ext = parse_filename(fn)
+        rses.add(rse)
+    return sorted(list(rses))
+
+def list_runs_by_age(Path, rse, reffile):
```
ericvaandering: Maybe too much for this round, but all these functions taking Path as a first argument (why is Path upper case?) are a good candidate for a class

lib/rucio/daemons/storage/consistency/actions.py
```diff
@@ -0,0 +1,771 @@
...
+
+###
+#   SAFEGUARD
+#   If a large fraction (larger than 'maxdarkfraction') of the files at a site
+#   are reported as 'dark', do NOT proceed with the deletion.
+#   Instead, put a warning in the _stats.json file, so that an operator can have a look.
+###
+
+# Get the number of files recorded by the scanner
+                print("latest_run", latest_run)
```
ericvaandering: Remove the print statements or use the logging modules common to Rucio

lib/rucio/daemons/storage/consistency/actions.py
```diff
@@ -0,0 +1,771 @@
...
+##########################################################################
+
+graceful_stop = threading.Event()
+
+
+##########################################################################
+### NOTE: declare_bad_file_replicas will be used directly from core/replica.py
+### when handling of DID is added there
+@transactional_session
+def declare_bad_file_replicas(dids, rse_id, reason, issuer,\
```
ericvaandering: Again, here and every where no need for continuation marks

lib/rucio/daemons/storage/consistency/actions.py
```diff
@@ -0,0 +1,771 @@
...
+    if "cc_miss" in stats:
+        if "status" in stats['cc_miss']:
+            cc_miss_status = stats['cc_miss']['status']
+    if cc_dark_status == 'done' or cc_miss_status == 'done':
+        return True
+    else:
+        return False
+
+
+def deckard(rse, dark_min_age, dark_threshold_percent, miss_threshold_percent,\
```
ericvaandering: This is a very long function. It should be split into parts so each can be understood on their own

lib/rucio/daemons/storage/consistency/actions.py
```diff
@@ -0,0 +1,771 @@
...
+
+
+def run(once=False, rses=None, sleep_time=60, dark_min_age=28, dark_threshold_percent=1.0,\
+  miss_threshold_percent=1.0, force_proceed=False, scanner_files_path="/var/cache/consistency-dump",\
+  threads=1):
+    """
+    Starts up the Consistency-Actions.
+    """
+    print("\nNow inside run(once...)\n")
+    if rses == []:
```
ericvaandering, line 725: Take a look at the reaper code. It allows RSEs to be specified as an RSE expression.

ericvaandering (changes requested):

bin/rucio-storage-consistency-actions
piperov: Agree - will change.

bin/rucio-storage-consistency-actions
piperov: pylint somehow managed to convince me to put them in.
Will remove.

bin/rucio-storage-consistency-actions
piperov: Agree - will change.

lib/rucio/daemons/storage/consistency/actions.py
piperov: The capital P is a left-over from the deckard code. Will correct it.
Grouping them into a class is a good idea for future development/improvement.

lib/rucio/daemons/storage/consistency/actions.py
piperov: Will do!

lib/rucio/daemons/storage/consistency/actions.py
piperov: Agree - will change.

lib/rucio/daemons/storage/consistency/actions.py
piperov: Discussed this with Igor. He agrees to break this out in the future, because it might be useful in other places, but for the moment we agreed to keep it in my pull request, and just add him as author.
I'll do that.

When he breaks it out one day, my code can start importing it from rucio/lib/rucio/common/ to avoid duplication.

lib/rucio/daemons/storage/consistency/actions.py
piperov: Split out the 'dark files processing' and 'missing replicas processing' in separate functions.

lib/rucio/daemons/storage/consistency/actions.py
piperov: The purpose of this statement is just to handle the case of daemon operating normally, vs operating in a single-RSE mode (interactively). So, no more cases need to be distinguished here. I believe it's OK as it is.

lib/rucio/daemons/storage/consistency/actions.py
piperov: Done.
The only print statements that remain are the ones from Igor's 'Stats' class. These can stay in for now.

bin/rucio-storage-consistency-actions
```diff
@@ -0,0 +1,64 @@
...
+    ''')
+    parser.add_argument("--run-once", action="store_true", default=False, help='One iteration only')
+    parser.add_argument("--rses", nargs='+', default=[], type=str,
+        help='RSEs to check specified as a RSE expression, defaults to check (default: check all RSEs).')
+    parser.add_argument('--sleep-time', action="store", default=60, type=int,
+        help='Concurrency control: thread sleep time after each chunk of work')
+    parser.add_argument('--dark-min-age', action="store", default=28, type=int,
+        help='Min. age of a file to be considered as DARK.')
+    parser.add_argument('--dark-threshold-percent', action="store", default=1.0,
+        type=float, help='Max. percentage of dark files at RSE.')
```
mlassnig: can you add a comment that this value has to be between 0..1 ?

bin/rucio-storage-consistency-actions
```diff
@@ -0,0 +1,64 @@
...
+Run the daemon::
+  $ rucio-storage-consistency-actions --run-once --rses T2_US_Purdue T2_US_Nebraska --dark-threshold-percent 2.0 --miss-threshold-percent 1.5 --scanner-files-path /tmp/consistency-dump --sleep-time 10
+    ''')
+    parser.add_argument("--run-once", action="store_true", default=False, help='One iteration only')
+    parser.add_argument("--rses", nargs='+', default=[], type=str,
+        help='RSEs to check specified as a RSE expression, defaults to check (default: check all RSEs).')
+    parser.add_argument('--sleep-time', action="store", default=60, type=int,
+        help='Concurrency control: thread sleep time after each chunk of work')
+    parser.add_argument('--dark-min-age', action="store", default=28, type=int,
+        help='Min. age of a file to be considered as DARK.')
```
mlassnig: can you add a comment which unit to use? seconds, days, etc...

bin/rucio-storage-consistency-actions
```diff
@@ -0,0 +1,64 @@
...
+    parser.add_argument('--dark-threshold-percent', action="store", default=1.0,
+        type=float, help='Max. percentage of dark files at RSE.')
+    parser.add_argument('--miss-threshold-percent', action="store", default=1.0,
+        type=float, help='Max. percentage of missing files at RSE.')
```
mlassnig: same

lib/rucio/daemons/storage/consistency/actions.py
```diff
@@ -0,0 +1,805 @@
...
+# - Stefan Piperov <stefan.piperov@cern.ch>, 2020-2021
+# - Igor Mandrichenko <ivm@fnal.gov>, 2020-2021
+
+"""
+Storage-Consistency-Actions is a daemon to delete dark files, and re-subscribe the missing ones, identified previously in a Storage-Consistency-Scanner run.
+"""
+
+import traceback
+
+import logging
```
mlassnig, line 26: please sort imports alphabetically

lib/rucio/daemons/storage/consistency/actions.py
```diff
@@ -0,0 +1,805 @@
...
+import sys 
+import glob
+import json
+import time
+import gzip
+import re
+import csv
+
+
+from rucio.core.rse import list_rses, get_rse_id
```
mlassnig: same here and below

lib/rucio/daemons/storage/consistency/actions.py
```diff
@@ -0,0 +1,805 @@
...
+from rucio.common import exception
+from rucio.common.logging import formatted_logger, setup_logging
+from rucio.common.exception import DatabaseException, UnsupportedOperation, RuleNotFound
+from rucio.core.heartbeat import live, die, sanity_check
+from rucio.db.sqla.util import get_db_time
+
+
+
+##########################################################################
+### NOTE: these are needed by local version of declare_bad_file_replicas()
```
mlassnig: can you add a `#TODO` tag

lib/rucio/daemons/storage/consistency/actions.py
```diff
@@ -0,0 +1,805 @@
...
+
+graceful_stop = threading.Event()
+
+
+##########################################################################
+### NOTE: declare_bad_file_replicas will be used directly from core/replica.py
+### when handling of DID is added there
+@transactional_session
+def declare_bad_file_replicas(dids, rse_id, reason, issuer,
+     status=BadFilesStatus.BAD, scheme='srm', session=None):
```
mlassnig: something seems off with the indentation here? (or maybe just the github webclient?)

mlassnig: srm default? shouldn't be something different?

lib/rucio/daemons/storage/consistency/actions.py
```diff
@@ -0,0 +1,805 @@
...
+    """
+    unknown_replicas = []
+    replicas = []
+    if True:
+        for did in dids:
+            scope = InternalScope(did['scope'], vo=issuer.vo)
+            name = did['name']
+            __exists, scope, name, already_declared, size =\
+               __exists_replicas(rse_id, scope, name, path=None, session=session)
+            if __exists and ((str(status) == str(BadFilesStatus.BAD) and not\
```
mlassnig: can you format this a bit differently so it's easier to read? more linebreaks, vertical alignment, etc., thanks

mlassnig: you use `str(status)` and compare with typecast `str` of a type. that doesn't look correct to me? i see it also below in the code.

lib/rucio/daemons/storage/consistency/actions.py
```diff
@@ -0,0 +1,805 @@
...
+### NOTE: declare_bad_file_replicas will be used directly from core/replica.py
+### when handling of DID is added there
+##########################################################################
+
+
+##############################
+### This is Igor's Stats class
+class Stats(object):
+
+    def __init__(self, path):
```
mlassnig, line 124: are these all needed? they seem rather straightforward? are the default class accessors not enough?

lib/rucio/daemons/storage/consistency/actions.py
```diff
@@ -0,0 +1,805 @@
...
+                stats = json.loads(f.read())
+        if stats_key:
+            stats[stats_key] = my_stats
+        else:
+            stats.update(my_stats)
+        open(stats_file, "w").write(json.dumps(stats))
+
+
+
+def cmp2dark(new_list="T2_US_Purdue_2021_06_18_02_28_D.list",
```
mlassnig: these default parameters look like they come from a unittest?

lib/rucio/daemons/storage/consistency/actions.py
```diff
@@ -0,0 +1,805 @@
...
+        stats[stats_key] = my_stats
+###############################
+### This was Igor's Stats class
+###############################
+
+
+####################
+### from Deckard
+####################
+def parse_filename(fn):
```
mlassnig, line 210: all these util classes should maybe go into a separate python module?

lib/rucio/daemons/storage/consistency/actions.py
```diff
@@ -0,0 +1,805 @@
...
+# This run was already processed
+            logger(logging.INFO, '\n Nothing to do here')
+
+    else:
+# No scans outputs are available for this RSE
+        logger(logging.INFO, '\n No scans available for this RSE')
+
+
+
+def deckard_loop(rses, dark_min_age, dark_threshold_percent, miss_threshold_percent,
```
mlassnig: not entirely sure what deckard means? (blade runner reference? :-)

lib/rucio/daemons/storage/consistency/actions.py
mlassnig: this is for the whole module, btw

lib/rucio/daemons/storage/consistency/actions.py
piperov: Yes, this was already suggested by Eric V., and agreed to do be done at a later stage. Probably at the stage when the 'Stats' code gets factored out as a separate class.

lib/rucio/daemons/storage/consistency/actions.py
piperov: Agree - will do!

lib/rucio/daemons/storage/consistency/actions.py
piperov: Agree - will do!

bin/rucio-storage-consistency-actions
piperov: The parameter is named 'percent', so it should be self-explanatory. It is supposed to be between 0 and 100%
I'll add a comment if you feel it's necessary though.

bin/rucio-storage-consistency-actions
piperov: Agreed - will do! 
(good point!)

bin/rucio-storage-consistency-actions
piperov: The names imply percentages, but I'll add a comment if that's not clear enough.

bin/rucio-storage-consistency-actions
mlassnig: yeah, I understand, but it has to be given as a float, so would be good to say that it's float 0...1, and not integer 0..100.

lib/rucio/daemons/storage/consistency/actions.py
piperov: You are right - those defaults are not needed in the production version.
Removed.

lib/rucio/daemons/storage/consistency/actions.py
piperov: This class will be re-implemented in the future, when the Stats class gets its own implementation under rucio/lib/rucio/common/, as Eric V. suggested in the previous comments. I'd rather leave this for those times.

lib/rucio/daemons/storage/consistency/actions.py
piperov: Done

lib/rucio/daemons/storage/consistency/actions.py
piperov: This class will be re-implemented in the future, when the Stats class gets its own implementation under rucio/lib/rucio/common/, as Eric V. suggested in the previous comments.
So, for the moment we can leave this code as-is. Maybe add a "TO-DO" tag?

lib/rucio/daemons/storage/consistency/actions.py
piperov: `deckard` is the name of the non-daemon version of the code that we developed in the last year or so, before deciding to turn this into a proper daemon. I decided to keep the function's name, as it reminds the readers of the code where it all comes from.

And - yes, it is a Blade runner reference (invented by Eric when many Rucio daemons had similar names ;)

lib/rucio/daemons/storage/consistency/actions.py
ericvaandering: It's not a type, but a member of the Enum class. It does work (here's the code: https://github.com/python/cpython/blob/4a57fa296b92125e41220ecd201eb2e432b79fb0/Lib/enum.py#L749) 

But maybe BadFileStatus.Bad.name should be compared to status.name? 

Somehow IIRC (it was a long time ago) it didn't work to compare the two objects directly

lib/rucio/daemons/storage/consistency/actions.py
piperov: Done

lib/rucio/daemons/storage/consistency/actions.py
piperov: This is just additional indentation for the continuation of the previous line.

ericvaandering: The usual standard is to align "status" with "dids"

ericvaandering: scheme is completely unused. It can be removed entirely if we're planning on replacing it with the https://github.com/rucio/rucio/pull/5068 version. I assume it's their to retain API compatibility with the old method?

So either remove it or just default it to None so people stop looking. :-)

bin/rucio-storage-consistency-actions
```diff
@@ -0,0 +1,64 @@
...
+import signal
+
+from rucio.daemons.storage.consistency.actions import run, stop
+
+def get_parser():
+    """
+    Returns the argparse parser.
+    """
+    parser = argparse.ArgumentParser(description="The Consistency-Actions daemon is responsible for applying the corrective actions resulting from a consistency-check scan of an RSE.", epilog='''
+
```
joeldierkes: Do we need the blank line in the beginning of the epilog?

lib/rucio/daemons/storage/consistency/actions.py
```diff
@@ -0,0 +1,805 @@
...
+    :param rse_id: The RSE id.
+    :param reason: The reason of the loss.
+    :param issuer: The issuer account.
+    :param status: Either BAD or SUSPICIOUS.
+    :param scheme: The scheme of the PFNs.
+    :param session: The database session in use.
+    """
+    unknown_replicas = []
+    replicas = []
+    if True:
```
joeldierkes: This seems a bit strange, please remove your debug statements

lib/rucio/daemons/storage/consistency/actions.py
joeldierkes: https://github.com/rucio/rucio/runs/4928763917?check_suite_focus=true
Flake8 has some ideas on what you should change. Have a look at it :)

lib/rucio/daemons/storage/consistency/actions.py
```diff
@@ -0,0 +1,805 @@
...
+        for did in dids:
+            scope = InternalScope(did['scope'], vo=issuer.vo)
+            name = did['name']
+            __exists, scope, name, already_declared, size =\
```
joeldierkes: I don't know if a double underscore in variables outside of classes is pythonish... (It just seems to affect the polymorphism https://www.datacamp.com/community/tutorials/role-underscore-python)

joeldierkes: Is it intended to override `scope` and `name` here? If they are just the same, think about using a `_` to denote that the values returned by the function are not used, and the old ones are kept. If they change, consider changing the variable names and not just override them.
Maybe ```exists, _, _, already_declared, size = ...```

lib/rucio/daemons/storage/consistency/actions.py
```diff
@@ -0,0 +1,805 @@
...
+                session.query(models.Source).filter_by(scope=scope, name=name,
+                  rse_id=rse_id).delete(synchronize_session=False)
+            else:
+                if already_declared:
+                    unknown_replicas.append('%s:%s %s' % (did['scope'], did['name'],
+                     'Already declared'))
+                else:
+                    unknown_replicas.append('%s:%s %s' % (did['scope'], did['name'],
+                     'Unknown replica'))
+        if str(status) == str(BadFilesStatus.BAD):
```
joeldierkes: Is it required to compare the string values here? Can status be something else than the `BadFileStatus` enum?

lib/rucio/daemons/storage/consistency/actions.py
```diff
@@ -0,0 +1,805 @@
...
+from rucio.common.utils import daemon_sleep
+from rucio.common import exception
+from rucio.common.logging import formatted_logger, setup_logging
+from rucio.common.exception import DatabaseException, UnsupportedOperation, RuleNotFound
+from rucio.core.heartbeat import live, die, sanity_check
+from rucio.db.sqla.util import get_db_time
+
+
+
+##########################################################################
```
joeldierkes: From my POV, all these horizontal comment lines don't add value to the source code. They hinder the reading flow, and if you want to divide your code with these, it is an indicator for too little functions or that the file should be split into multiple files. We are not using them in rucio...

lib/rucio/daemons/storage/consistency/actions.py
```diff
@@ -0,0 +1,805 @@
...
+    except IntegrityError as error:
+        raise exception.RucioException(error.args)
+    except DatabaseError as error:
+        raise exception.RucioException(error.args)
+    except FlushError as error:
+        raise exception.RucioException(error.args)
+
+    return unknown_replicas
+### NOTE: declare_bad_file_replicas will be used directly from core/replica.py
+### when handling of DID is added there
```
joeldierkes: Duplicated in line 73-74.
Also, I have trouble understanding the 'Note'. I don't really know  what you mean by that.
Also, the comments beginning with "TODO" were added to quickly find open tasks in a project. I have never seen 'NOTE' before, from my POV it is just a regular comment. Thus, the prefix is redundant? Or is there a special meaning behind it?

lib/rucio/daemons/storage/consistency/actions.py
```diff
@@ -0,0 +1,805 @@
...
+##########################################################################
+
+
+##############################
+### This is Igor's Stats class
```
joeldierkes: Cool for Igor, consider adding doc comments to the class though.
Consider using the :author: key there instead of the unformatted comment line.

lib/rucio/daemons/storage/consistency/actions.py
```diff
@@ -0,0 +1,805 @@
...
+class Stats(object):
+
+    def __init__(self, path):
+        self.Path = path
+        self.Data = {}
```
joeldierkes: Variables in python should be lower case

lib/rucio/daemons/storage/consistency/actions.py
```diff
@@ -0,0 +1,805 @@
...
+    def update(self, data):
+        self.Data.update(data)
+        self.save()
+
+    def save(self):
+        try:
+            with open(self.Path, "r") as f:
+                data = f.read()
+        except:
+            #traceback.print_exc()
```
joeldierkes: Consider removing this comment

lib/rucio/daemons/storage/consistency/actions.py
```diff
@@ -0,0 +1,805 @@
...
+            data = ""
+        data = json.loads(data or "{}")
+        data.update(self.Data)
+        open(self.Path, "w").write(json.dumps(data, indent=4))
+
+
+def write_stats(my_stats, stats_file, stats_key=None):
+    if stats_file:
```
joeldierkes, line 154: Consider throwing an error here if `stats_file` is not defined in contrast to doing nothing.

lib/rucio/daemons/storage/consistency/actions.py
```diff
@@ -0,0 +1,805 @@
...
+        stats = {}
+        if os.path.isfile(stats_file):
+            with open(stats_file, "r") as f:
+                stats = json.loads(f.read())
+        if stats_key:
+            stats[stats_key] = my_stats
+        else:
+            stats.update(my_stats)
+        open(stats_file, "w").write(json.dumps(stats))
```
joeldierkes, line 163: The function is essentially doing two things:
If no `stats_key` is defined, then the entire `my_stats` dict gets added to the `stats` dict and written back to the file.
If the `stats_key` is defined, then the `my_stats` dict gets assigned to a property of the `stats` dict and everything is written back.

Consider breaking the logic into two functions and marking them accordingly. The name `write_stats` does not represent this complex behaivior at all.

joeldierkes: Also, where is this function used?

lib/rucio/daemons/storage/consistency/actions.py
joeldierkes: Consider using the sphinx documentation style we use in rucio

lib/rucio/daemons/storage/consistency/actions.py
```diff
@@ -0,0 +1,805 @@
...
+    if stats_file:
+        my_stats.update({
+            "elapsed": t1-t0,
+            "end_time": t1,
+            "status": "done"
+        })
+        stats[stats_key] = my_stats
+###############################
+### This was Igor's Stats class
+###############################
```
joeldierkes: This was a class and two functions. **REALLY** consider refactoring that into common or in a file on its own.

lib/rucio/daemons/storage/consistency/actions.py
```diff
@@ -0,0 +1,805 @@
...
+
+
+####################
+### from Deckard
+####################
```
joeldierkes: Consider adding a more unique identifier, I thought Deckard is a person for 20 minutes straight.
Also, still the comment style.

lib/rucio/daemons/storage/consistency/actions.py
joeldierkes, line 210: Parse filename into what?
Maybe create a new class for the file identifiers. This way the returned information is bagged up into one variable instead of four when the function is called.

lib/rucio/daemons/storage/consistency/actions.py
```diff
@@ -0,0 +1,805 @@
...
+    if "cc_miss" in stats:
+        if "status" in stats['cc_miss']:
+            cc_miss_status = stats['cc_miss']['status']
+    if cc_dark_status == 'done' or cc_miss_status == 'done':
+        return True
+    else:
+        return False
+
+
+def process_dark_files(path, rse, latest_run, maxdarkfraction, max_files_at_site, oldenough_run):
```
joeldierkes: Inconsistent use of `_`.

lib/rucio/daemons/storage/consistency/actions.py
```diff
@@ -0,0 +1,805 @@
...
+        "confirmed_dark_files": 0,
+        "x-check_run": oldenough_run,
+        "status": "started"
+    }
+    stats[stats_key] = cc_stats
+
+# Compare the two lists, and take only the dark files that are in both
+    latest_dark = re.sub('_stats.json$', '_D.list', latest_run)
+    oldenough_dark = re.sub('_stats.json$', '_D.list', oldenough_run)
+    logger(logging.INFO,'\n latest_dark = %s' % latest_dark)
```
joeldierkes: Are the new line characters needed? This makes the log output harder to read.

lib/rucio/daemons/storage/consistency/actions.py
```diff
@@ -0,0 +1,805 @@
...
+    logger(logging.INFO,'\n oldenough_dark = %s' % oldenough_dark)
+    confirmed_dark = "%s_DeletionList.csv" % latest_run
+    confirmed_dark = re.sub('_stats.json$', '_DeletionList.csv', latest_run)
```
joeldierkes: You override the variable you introduced one line earlier.

lib/rucio/daemons/storage/consistency/actions.py
```diff
@@ -0,0 +1,805 @@
...
+        logger(logging.INFO, '\n Can proceed with dark files deletion')
+
+# Then, do the real deletion (code from DeleteReplicas.py)
+        issuer = InternalAccount('root')
+        with open(confirmed_dark, 'r') as csvfile:
+            reader = csv.reader(csvfile)
+            dark_replicas = []
+            #for rse, scope, name, reason in reader:
+            scope = "cms"
+            reason = "deleteing dark file"
```
joeldierkes: Typo

lib/rucio/daemons/storage/consistency/actions.py
```diff
@@ -0,0 +1,805 @@
...
+                paths = parse_pfns(attributes, urls, operation='delete')
+                replicas = [{'scope': Intscope, 'rse_id': rse_id, 'name': name,
+                  'path': paths[url]['path']+paths[url]['name']}]
+                add_quarantined_replicas(rse_id, replicas, session=None)
+                deleted_files += 1
+                labels = {'rse': rse}
+                record_counter('storage.consistency.actions_dark_files_deleted_counter',
+                  delta=1, labels=labels)
+
+        #Update the stats
```
joeldierkes: Unnecessary comment

lib/rucio/daemons/storage/consistency/actions.py
```diff
@@ -0,0 +1,805 @@
...
+
+        cc_stats.update({
+            "end_time": t1,
+            "initial_dark_files": dark_files,
+            "confirmed_dark_files": 0,
+            "status": "ABORTED",
+            "aborted_reason": "%3.2f%% dark" % darkperc,
+        })
+        stats[stats_key] = cc_stats
+        record_gauge('storage.consistency.actions_dark_files_deleted', 0, labels=labels)
```
joeldierkes: A lot of this lines are duplicated. Consider refactoring it

lib/rucio/daemons/storage/consistency/actions.py
```diff
@@ -0,0 +1,805 @@
...
+
+#####################################
+#   Done with Dark Files processing
+#####################################
+
+
+def process_miss_files(path, rse, latest_run, maxmissfraction, max_files_at_site, oldenough_run):
```
joeldierkes: I don't see the gain in abbreviating missing

lib/rucio/daemons/storage/consistency/actions.py
joeldierkes: I like the idea of referencing the code origin, however you should use references that are clear to readers. Maybe give more context, if @mlassnig does not understand that with his experience in Rucio and the field, how should a normal mortal person be able to understand it then?

lib/rucio/daemons/storage/consistency/actions.py
```diff
@@ -0,0 +1,805 @@
...
+
+
+def run(once=False, rses=None, sleep_time=60, default_dark_min_age=28, default_dark_threshold_percent=1.0,
+  default_miss_threshold_percent=1.0, force_proceed=False, default_scanner_files_path="/var/cache/consistency-dump",
+  threads=1):
+    """
+    Starts up the Consistency-Actions.
+    """
+
+# Setup logging
```
joeldierkes: redundant

lib/rucio/daemons/storage/consistency/actions.py
```diff
@@ -0,0 +1,805 @@
...
+
+    prefix = 'storage-consistency-actions (run())'
+    logger = formatted_logger(logging.log, prefix + '%s')
+
+    logger(logging.INFO, '\n Now inside run(once...)\n')
+
+    dark_min_age = default_dark_min_age
+    dark_threshold_percent = default_dark_threshold_percent
+    miss_threshold_percent = default_miss_threshold_percent
+    scanner_files_path = default_scanner_files_path
```
joeldierkes, line 723: Redundant, there is no reason to create new variables.
Consider replacing the `default_` variables with their significant other

lib/rucio/daemons/storage/consistency/actions.py
```diff
@@ -0,0 +1,805 @@
...
+    else:
+        logging.info('Consistency Actions starting %s threads' % str(threads))
+        threads = [threading.Thread(target=actions_loop, kwargs={'once': once, 'rses': rses, 'sleep_time': sleep_time,
+         'dark_min_age': dark_min_age, 'dark_threshold_percent': dark_threshold_percent,
+         'miss_threshold_percent': miss_threshold_percent, 'force_proceed': force_proceed,
+         'scanner_files_path': scanner_files_path}) for i in range(0, threads)]
+        logger(logging.INFO, '\n Threads: %d' % len(threads))
+        [t.start() for t in threads]
+        # Interruptible joins require a timeout.
+        while threads[0].is_alive():
```
joeldierkes, line 762: What if `thread[0]` is done but `thread[1]` is not?

lib/rucio/daemons/storage/consistency/actions.py
```diff
@@ -0,0 +1,805 @@
...
+            [t.join(timeout=3.14) for t in threads]
```
joeldierkes, line 763: Is the timeout too much? 3 Seconds seems kinda long. (name of your sextape lol)

joeldierkes (changes requested):

lib/rucio/daemons/storage/consistency/actions.py
piperov: I don't understand this comment. Can you elaborate, please?

lib/rucio/daemons/storage/consistency/actions.py
piperov: I can change that. I was using it consistently in my code as the log is MORE readable for me that way. But if that makes it more difficult for everybody else, I'll change it.

lib/rucio/daemons/storage/consistency/actions.py
joeldierkes: You have the variable  `max_files_at_site`, which delimits the words with a `_`. At the same time, you have the variable `maxdarkfraction`, without the word delimiter. I would suggest to use the word delimiter consistently, e.g.:
- `max_dark_fraction`
- `old_enough_run`

bin/rucio-storage-consistency-actions
piperov: No, we don't 
Removed.

lib/rucio/daemons/storage/consistency/actions.py
piperov: Done.

lib/rucio/daemons/storage/consistency/actions.py
piperov: This whole block of code is going away soon, as we are switching to the modified/improved declare_bad_file_replicas()
- see: https://github.com/rucio/rucio/pull/5068

So, I wouldn't put too much beautification efforts int it at this point.

lib/rucio/daemons/storage/consistency/actions.py
piperov: Again - this code was temporarily borrowed from the production Rucio code, and will be replaced with the new version in PR5068:  https://github.com/rucio/rucio/pull/5068
So - we can leave this in unchanged for the moment.

lib/rucio/daemons/storage/consistency/actions.py
piperov: Agree.
Removed.

lib/rucio/daemons/storage/consistency/actions.py
piperov: I've added an explanation as to why this is done. 
And used a # TODO codetag, so that this gets fixed in the future.

lib/rucio/daemons/storage/consistency/actions.py
piperov: I am not an expert in multi-threading. This part of the code is modeled after the judge-cleaner daemon (daemons/judge/cleaner.py) Same principles apply for the consistency-actions daemon.

lib/rucio/daemons/storage/consistency/actions.py
piperov: Same as above.

lib/rucio/daemons/storage/consistency/actions.py
piperov: Done.

piperov: Replaced with 'None'

lib/rucio/daemons/storage/consistency/actions.py
piperov: As I responded in other comments, this whole piece is being removed, once the modified declare_bad_file_replicas() from PR5068 is accepted.
https://github.com/rucio/rucio/pull/5068

lib/rucio/daemons/storage/consistency/actions.py
piperov: 'miss' and 'dark' are both 4 characters, so some sort of symmetry between the two functions is implied/maintained.
This is a consistent approach used throughout the code.

piperov (commented):
All tests are passing now successfully.

rcarpa:
The diff doesn't seem right at all

lib/rucio/daemons/storage/consistency/actions.py
piperov: Done.

lib/rucio/daemons/storage/consistency/actions.py
piperov: NOTEs replaced with FIXME tags, following https://www.python.org/dev/peps/pep-0350

lib/rucio/daemons/storage/consistency/actions.py
piperov: Cannot insert Igor as author, because it breaks the header.
tools/add_header -i %filename% removes potential authors who have not made a git commit.
Leaving just for future reference.

lib/rucio/daemons/storage/consistency/actions.py
piperov: Corrected.

lib/rucio/daemons/storage/consistency/actions.py
piperov: Done.

lib/rucio/daemons/storage/consistency/actions.py
piperov: This part of the code is handled by Igor, and will be corrected when the class is split out in the near future.
Marking it with TODO: tag for now.

lib/rucio/daemons/storage/consistency/actions.py
piperov: This part of the code is handled by Igor, and will be corrected when the class is split out in the near future.
Marking it with TODO: tag for now.

lib/rucio/daemons/storage/consistency/actions.py
piperov: This was a reference to the first implementation of the daemon. 
Naming convention is now changed, so the reference can be removed altogether.
Done.

lib/rucio/daemons/storage/consistency/actions.py
piperov: This part of the code is handled by Igor, and will be corrected when the class is split out in the near future.

lib/rucio/daemons/storage/consistency/actions.py
piperov: The logic here is common with the Consistency Scanner, so it can be changed in the future in both places.
Adding a  TODO: tag for that purpose.

lib/rucio/daemons/storage/consistency/actions.py
piperov: Done.

lib/rucio/daemons/storage/consistency/actions.py
piperov: Done.

lib/rucio/daemons/storage/consistency/actions.py
piperov: That was an overlooked replacement. Thanks for noticing!
Fixed.

lib/rucio/daemons/storage/consistency/actions.py
piperov: Fixed.

lib/rucio/daemons/storage/consistency/actions.py
piperov: Removed.

lib/rucio/daemons/storage/consistency/actions.py
piperov: This can be addressed in the final version of the Stats class, when Igor contributes it separately to Rucio.
Placing a TODO: tag here to remind us of that fact.

piperov:
I'm not sure I understand this comment.
Which diff doesn't look right?



---
"Covido, ergo Zoom"

bari12 (commented):
Hi Stefan,

I went through the code also with the help of @joeldierkes and @dchristidis. It's much better now, but there is still some open points. I'll go to the functional ones first, then there are others which are more cosmetic.

- When looking at the `process_dark_files` method, it really looks like you are dealing with lfns at this point (or in general). This works for CMS, but I don't think it will (easily) work for anyone else. Especially on deterministic storage. For example if there is garbage data on the storage and you assume the path is a lfn, this will not translate to a removable pfn.
- There is a few hardcoded scopes `cms` in there. These methods would have to be made scope independent.
- The `declare_bad_file_replicas` method in there should be part of the core. There should be a strict separation between daemons and core methods. Core methods being the one accessing the database. Can't you use the already existing core methods for that?
- The `threads = 1` overwrite is not great. If it only works with one thread (which is fine) remove the option entirely (In that case I think we can even remove the `threading` module)

The more cosmetic ones:
- The external files should be documented somewhere (Structure etc.). I understand that this is part of a component stack, but it needs to be documented somewhere.
 
There is a few more but this we can even do in a separate PR, I think it's more important to address the functional comments.

Cheers,
Martin

bari12 (approved):
Thanks for including all the comments Stefan! 👍

Issue #5024, opened 2021-11-18
Extend list_dataset_replicas_bulk to use the --deep flag


germanfgv:
Motivation
----------
Currently, It is only possible to query dataset replicas info with the `--deep` flag  using `list_dataset_replicas`. This method only supports a single dataset per call. It would be very helpful to be able to query info from several datasets at once, similar to what's done with `list_dataset_replicas_bulk` (added in #3736). Unfortunately, this method currently lacks support for the `deep` flag. 

This feature was initially though to be unnecessary (see [this discussion](https://github.com/rucio/rucio/issues/2459#issuecomment-481268727)), but the recent issues with the synchronization of table `collection_replicas`  in CMS Rucio have made it clear that we would benefit from the ability using the deep flag in bulk.

Related to issue [#2459](https://github.com/rucio/rucio/issues/2459)

@bari12 @ivmfnal 

Modification
------------

Add support for the `--deep` flag in the list_dataset_replicas_bulk API, from the rucio-clients package.

ivmfnal:
Currently, there are 2 ways to query replicas in datasets:


1. Query multiple datasets and return basic info about each replica found in these datasets (file scope, path, rse_id, etc...)
2. Query for replicas in single dataset with the option to return extended info in addition to the basic. The query to get extended info seems to be quite complicated and probably is expensive.

So as of now, it's either (1) get little information from a fast query for many datasets or (2) get detailed info using expensive query for single dataset.

If I understand the request correctly, the objective is to give the user the ability to combine the two - to be able to run the expensive query on multiple datasets.

Is this what we want ?

germanfgv:
The concern is not so much about the "amount" of information, is about reliability. For example, take the following commands using Rucio CLI:
```
1. rucio list-dataset-replicas <scope>:<dataset_name>
2. rucio list-dataset-replicas --deep <scope>:<dataset_name>
```
Both options return the same information, but the option `2` is more reliable as it is not affected by sync issues. This is, of course, with the cost of being more expensive. The same thing happens when using Python API's [`list_dataset_replicas`](https://rucio.readthedocs.io/en/latest/api/replica.html#rucio.client.replicaclient.ReplicaClient.list_dataset_replicas) with or without `deep=True`.

[...]

PR #5123, opened 2021-12-21
Replicas: implemented and tested #5024

ivmfnal:

### DIFF
--- a/lib/rucio/core/replica.py
+++ b/lib/rucio/core/replica.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Copyright 2013-2021 CERN
+# Copyright 2013-2022 CERN
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


lib/rucio/core/replica.py
```diff
@@ -2678,43 +2678,49 @@ def list_dataset_replicas(scope, name, deep=False, session=None):
...
+                                      models.CollectionReplica.created_at,
+                                      models.CollectionReplica.updated_at,
+                                      models.CollectionReplica.accessed_at) \
+                    .filter(models.CollectionReplica.did_type == DIDType.DATASET) \
+                    .filter(models.CollectionReplica.rse_id == models.RSE.id) \
+                    .filter(or_(*chunk)) \
+                    .filter(models.RSE.deleted == false())
+                for row in query:
+                    yield row._asdict()
+        except NoResultFound:
```
bari12: I am not sure if this part would ever be raised? Did you test this? If yes, there is definitely a difference then if you run it with `deep=True` - not sure if this would be raised in that case. So the method acts a bit different for empty results; Would be worth to check.

ivmfnal: I do not think these are my changes. I only changed indentation for this fragment

bari12: Yes I understand, sorry should have phrased this more clearly. I just think that due to this part there is a difference if there are no results between `deep=True` and `deep=False`, which is a bit inconsistent. So I am wondering if it is not better to remove this except completely. (Not sure if it is raised at all in the first place.

ivmfnal: Do you mean that if there are no results, then it should simply yield nothing in both cases (deep=True and deep=False), instead of generating the NotFound exception ?

I would agree to that actually. I think "no results" is a valid result in this case. But I wonder if maybe at some place, someone expects this exception ?

bari12: Exactly it should just yield nothing.
I suspect that `NoResultFound` is not even raised in this case actually, but I think we can relatively safely remove that exception.

Issue #4961, opened 2021-10-28
Make dirac server plugin multi-VO capable


martynia:
Motivation
----------
The current version of Dirac-Rucio interface on the server side does not work in a multi-VO Rucio environment

Modification
------------
We need to enhance all 3 levels of the code (`flaskapi/v1/dirac.py`, `core/dirac.py`, `api/dirac.py`) to handle VO specific requests.

To achieve this we need to get the VO name from `requests` headers and pass in down the invocation line. Multi-VO rucio makes a VO-aware decision on the API level, not in the core. This is unfortunately not true in the current Rucio-interface.

Let's look at this step-by-step. The first VO-dependent call can be found in api/dirac.py here:

https://github.com/rucio/rucio/blob/a9ea3b359d221e157830cb1961d4a2488ada32f8/lib/rucio/api/dirac.py#L52

The change is simple, adding the `vo` parameter to the call solves the problem, but not for very long ...

Another problematic call is here:

https://github.com/rucio/rucio/blob/92474868cec7d64b83b46b451ecdc4a294ef2711/lib/rucio/core/dirac.py#L68

Scopes are VO-dependent, so one could import an API counterpart of this function, rather than the core one and get the scopes for a VO we want. This is not very elegant, since the core should not call any API functions, it should rather be another way around. One could copy the filtering code from the  API, this would be code repetition. Not ideal either.

After having done any of these modifications we get a bit further:

`[Wed Oct 27 16:36:05.410228 2021] [wsgi:error] [pid 30268] [remote 146.179.232.106:34688]   File "/opt/rucio/rucio/lib/rucio/core/did.py", line 134, in add_did

[Wed Oct 27 16:36:05.410269 2021] [wsgi:error] [pid 30268] [remote 146.179.232.106:34688] rucio.common.exception.ScopeNotFound: Scope does not exist.
[Wed Oct 27 16:36:05.410276 2021] [wsgi:error] [pid 30268] [remote 146.179.232.106:34688] Details: Scope not found!
[Wed Oct 27 16:36:05.410291 2021] [wsgi:error] [pid 30268] [remote 146.179.232.106:34688]`

This is expected (_scope not found_), because once in the core we try to get dids and replicas ignoring the VO, i.e. using the core versions of the methods, rather that their API counterparts.

[...]

PR #5125, opened 2021-12-22
Make Dirac server plugin Multi-VO capable

martynia:
The changes in 3 dirac.py files  (core, api,  flaskapi/v1) make Dirac-Rucio multi-VO capable.
Adresses #4961

To have a scope properly extracted by the server the value of  `SCOPE_NAME_REGEXP` has to be set: 
```
SCOPE_NAME_REGEXP = '/([^/]*)(/.*)' 
```

In addition the server has to be configured to select a Dirac scope algorithm (as defined in `rucio/common/utils.py` in this PR).
The client will work without a config file.

### DIFF
--- a/lib/rucio/api/dirac.py
+++ b/lib/rucio/api/dirac.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Copyright 2020-2021 CERN
+# Copyright 2020-2022 CERN
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -18,6 +18,8 @@
 # - Eli Chadwick <eli.chadwick@stfc.ac.uk>, 2020
 # - Martin Barisits <martin.barisits@cern.ch>, 2020-2021
 # - Benedikt Ziemons <benedikt.ziemons@cern.ch>, 2020
+# - martynia <janusz.martyniak@googlemail.com>, 2021
+# - Janusz Martyniak <janusz.martyniak@googlemail.com>, 2022
 
 from __future__ import print_function
 
@@ -30,7 +32,7 @@ from rucio.common.exception import AccessDenied
 from rucio.common.utils import extract_scope
 
 
-def add_files(lfns, issuer, ignore_availability):
+def add_files(lfns, issuer, ignore_availability, vo='def'):
     """
     Bulk add files :
     - Create the file and replica.
@@ -40,8 +42,10 @@ def add_files(lfns, issuer, ignore_availability):
     :param lfns: List of lfn (dictionary {'lfn': <lfn>, 'rse': <rse>, 'bytes': <bytes>, 'adler32': <adler32>, 'guid': <guid>, 'pfn': <pfn>}
     :param issuer: The issuer account.
     :param ignore_availability: A boolean to ignore blocked sites.
+    :param vo: The VO to act on.
+
     """
-    scopes = list_scopes()
+    scopes = list_scopes(vo=vo)
     dids = []
     rses = {}
     for lfn in lfns:
@@ -49,7 +53,7 @@ def add_files(lfns, issuer, ignore_availability):
         dids.append({'scope': scope, 'name': name})
         rse = lfn['rse']
         if rse not in rses:
-            rse_id = get_rse_id(rse=rse)
+            rse_id = get_rse_id(rse=rse, vo=vo)
             rses[rse] = rse_id
         lfn['rse_id'] = rses[rse]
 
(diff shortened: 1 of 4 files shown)


martynia:
I fixed a problem with a `filter` argument (it should be `filter_`) in a separate branch it it passes now. The problem which remains is one when testing the conveyor:

=========================== short test summary info ============================
3559
FAILED test_conveyor.py::test_multihop_intermediate_replica_lifecycle[caches_mock0-core_config_mock0]
3560
====== 1 failed, 14 passed, 1 skipped, 28 warnings in 1005.65s (0:16:45) =======
3561
Error: Process completed with exit code 1.

Not sure yet what it is related to.

joeldierkes:
Hello @martynia,

if you did not find any bugs in your code or did not change any code related to the conveyor (including core) and the `test_conveyor.py::test_multihop_intermediate_replica_lifecycle` test is still failing, rerun it again. The `test_conveyor` has some complicated logic, and there seems to be a raise condition which is related to the test.

martynia:
Hello @joeldierkes,

I haven't done any changes related to the conveyor, I merged in my own changes, it triggered the tests. Let's see what happens.

martynia:
All tests are passing now.

cserf (dismissed):
LGTM

lib/rucio/api/dirac.py
```diff
@@ -40,31 +42,33 @@ def add_files(lfns, issuer, ignore_availability):
...
         lfn['rse_id'] = rses[rse]
 
     # Check if the issuer can add dids and use skip_availabitlity
     for rse in rses:
         rse_id = rses[rse]
         kwargs = {'rse': rse, 'rse_id': rse_id}
-        if not has_permission(issuer=issuer, action='add_replicas', kwargs=kwargs):
-            raise AccessDenied('Account %s can not add file replicas on %s' % (issuer, rse))
-        if not has_permission(issuer=issuer, action='skip_availability_check', kwargs=kwargs):
+        if not has_permission(issuer=issuer, action='add_replicas', kwargs=kwargs, vo=vo):
```
bari12, line 64: I think it would be better to add the `vo` to the `kwargs` (above) instead.

lib/rucio/core/dirac.py
```diff
@@ -51,7 +54,7 @@ def _exists(scope, name, session=None):
 
 
 @transactional_session
-def add_files(lfns, account, ignore_availability, session=None):
+def add_files(lfns, account, ignore_availability, session=None, vo='def'):
```
bari12: `session` should stay the last argument

lib/rucio/core/dirac.py
```diff
@@ -61,18 +64,20 @@ def add_files(lfns, account, ignore_availability, session=None):
...
     :param ignore_availability: A boolean to ignore blocklisted sites.
-    :session: The session used
+    :param session: The session used
+    :param vo: The VO to act on
     """
     attachments = []
     # The list of scopes is necessary for the extract_scope
-    scopes = list_scopes(session=session)
+    filter_ = {'scope': InternalScope(scope='*', vo=vo)}
+    scopes = list_scopes(session=session, filter_=filter_)
```
bari12: Put session last for readability

lib/rucio/web/rest/flaskapi/v1/dirac.py
```diff
@@ -55,8 +58,11 @@ def post(self):
         lfns = param_get(parameters, 'lfns')
         ignore_availability = param_get(parameters, 'ignore_availability', default=False)
 
+        LOGGER = logging.getLogger('Dirac.addFiles')
+        LOGGER.setLevel(logging.INFO)
```
bari12: Why this logger?

bari12 (changes requested):
Hi Janusz, some minor comments and questions. But looks good to me too.

lib/rucio/api/dirac.py
martynia: Hi,
It would probably be better, but it would require making a change in `has_permission` itself, the bit of code I haven't touched (it already has a `vo` keyword argument). 
The change would be something like this:
```python
def has_permission(issuer, action, kwargs):

    # at the bottom get the VO
    vo = kwargs.get('vo', 'def') 
    issuer = InternalAccount(issuer, vo=vo)

```
But I'm not sure if this does not break existing applications, if someone relies on an existence of the `vo` keyword argument in the signature above.

lib/rucio/web/rest/flaskapi/v1/dirac.py
martynia: Oh, sorry, this is a left-over form some testing. Removed.

lib/rucio/core/dirac.py
martynia: This is a relatively new code, so assuming no-one is using purely positional arguments here. Will make the change.

lib/rucio/api/dirac.py
thysk: Looking at the has_permission, may other places (104 different calls) all call this function and specify the vo separately 
often as - `if not has_permission(issuer=issuer, vo=vo, action='various actions', kwargs=kwargs):`

Is it not then beyond the scope of this PR?

bari12: You are right, I made a mistake here by directly looking at the core `has_permission` method, which does not have the `vo` parameter. But `api.has_permission` has it, which is the correct one. 👍 
You can ignore my comment, consider it resolved 😄

bari12 (approved):

PR #5127, opened 2022-01-05
Hermes 2 evolution : Closes #4610

### LINKED ISSUE #4610: Hermes 2 evolution
During the discussion in the [development meeting on 2021-04-29](https://indico.cern.ch/event/1020696/) it was decided to evolve hermes2 in a similar fashion to the reaper daemon and finally fully replace hermes1.

For this transition to work the following steps have been identified:

1. Currently the hermes2 daemon is exposed to starvation issue. If one of the destination services is unavailable, the queue cannot be continuously processed and a backlog will rise. To solve this, we propose to expand the messages table with a `service` column and on message creation, directly create duplicates (one for each service) into the table. The hermes2 daemon then processes each service queue independently, thus other queues will not starve if one service is down.
2. Documentation needs to be expanded to properly explain how to setup a monitoring infrastructure based on hermes2.
3. Tests are needed
4. Remove hermes1 and replace by hermes2

cserf:
Hermes 2 evolution : Closes #4610

- Treat each service individually
- Add tests

### DIFF
--- a/lib/rucio/core/message.py
+++ b/lib/rucio/core/message.py
@@ -31,6 +31,8 @@ from rucio.db.sqla.session import transactional_session
 
 REGION = make_region_memcached(expiration_time=900)
 
+SUPPORTED_SERVICES = ['influx', 'elastic', 'email', 'activemq']
+
 
 @transactional_session
 def add_message(event_type, payload, session=None):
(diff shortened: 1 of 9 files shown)


lib/rucio/tests/test_hermes.py
```diff
@@ -8,34 +8,127 @@
...
+                              be sure to check your area for personal belongings.
 
+                              Thank you, and have a very safe, and productive day.'''})
 
-class TestHermes(object):
-    ''' Test the messaging deamon. '''
+    messages = retrieve_messages(50, old_mode=False)
+    service_dict = {'influx': 0, 'elastic': 0, 'email': 0, 'activemq': 0}
+    for message in messages:
+        print(message)
```
bari12, line 96: Remove print?

lib/rucio/core/message.py
```diff
@@ -64,33 +66,44 @@ def add_message(event_type, payload, session=None):
         try:
             services_list = config_get('hermes', 'services_list')
         except (NoOptionError, NoSectionError, RuntimeError):
-            services_list = None
+            services_list = 'influx,activemq,elastic,email'
```
bari12, line 54: No caching needed around common.config_get, it's already cached within.
Wouldn't it be better to make the default only `activemq,email` - to mirror hermes1 behaviour?

bari12 (commented):
In general I think it is ok, some smaller comments.
But there is a discrepancy I think we should get rid of which is the one between
`event_type` and `service`.

For example `email` is both a service and an event_type. I think it might be better to redesign parts of the code there. 
One option could be to change the `add_message` (or other parts of the code) and add a `service` parameter to it. Per default, all messages go to all services, if not one specifically is selected (email).

bari12:
Hi @cserf for the 🥶 on Wednesday I think it would still be good to get these two changes in. I don't think we can right away switch hermes1 with hermes2 then, but probably only very small changes are needed then.

cserf:
Closing the PR to debug tests

PR #5136, opened 2022-01-11
Fix id/id_ and some PEP-8 issues

ericvaandering:

### DIFF
--- a/lib/rucio/common/stomp_utils.py
+++ b/lib/rucio/common/stomp_utils.py
@@ -1,0 +1,58 @@
+# -*- coding: utf-8 -*-
+# Copyright 2022 CERN
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


bari12:
Hi @ericvaandering 
If you are still developing this, please close the PR and submit it once done. I know it's just a draft, but each commit submission takes runner slots away from the CI which delays everything else. You can just let the GH actions run in your private repository (Just push to your origin) and you get exactly the same test runs as here. Thanks

ericvaandering:
Closing to be kind to the CI

ericvaandering:
OK, this is now working and I'd like to get it merged. As you can see from the comments I lifted a very useful function from kronos.py so I'd propose to refactor this. It seems to me lib/common/stomp_utils.py or something with a similar name would be a good place to put this function. Thoughts?

bari12:
Making the stomp handling a common module is a good idea. We would have to include the different workflows where stomp is used though (Mostly hermes and kronos, and then all the minor workflows like the cache-consumer)

ericvaandering:
I’d propose to add the simple refactoring to this PR and then open an issue for Hermes.  I don’t know how different it is.

ericvaandering:
@bari12 This passes all the tests (after the rebase) and runs. Can we review and merge? (I can squash one final commit if everyone signs off, I don't want to trigger a new round of tests just for that).

bari12 (approved):

PR #5147, opened 2022-01-12
Transfers: get bring_online as int from config. Closes #5146

### LINKED ISSUE #5146: `bring_online` parameter is submitted as a string, instead of int, to FTS
Motivation
----------
https://github.com/rucio/rucio/blob/e2ed93b6d315091cf4960f76b4253ce8fad27a6c/lib/rucio/daemons/conveyor/submitter.py#L107
Loads the value as a string. Later on in the workflow this is submitted to FTS as a string in the dictionary, which the new FTS3 REST interface on fts3-pilot does not like.

Modification
------------
This should be cast to int, or better loaded with `config_get_int`. Also applies to the stager.

rcarpa:

### DIFF
--- a/lib/rucio/daemons/conveyor/submitter.py
+++ b/lib/rucio/daemons/conveyor/submitter.py
@@ -48,7 +48,7 @@ from six.moves.configparser import NoOptionError
 
 import rucio.db.sqla.util
 from rucio.common import exception
-from rucio.common.config import config_get, config_get_bool
+from rucio.common.config import config_get, config_get_bool, config_get_int
 from rucio.common.logging import setup_logging
 from rucio.common.schema import get_schema_value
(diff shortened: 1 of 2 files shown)


lib/rucio/daemons/conveyor/submitter.py
```diff
@@ -104,7 +104,7 @@ def submitter(once=False, rses=None, partition_wait_time=10,
         timeout = None
 
     try:
-        bring_online = config_get('conveyor', 'bring_online')
+        bring_online = config_get_int('conveyor', 'bring_online')
```
bari12, line 107: I think the same change needs to be made in the stager as well.

rcarpa: Good catch !

bari12 (approved):

PR #5158, opened 2022-01-19
Monitoring & Logging: prometheus metrics for servers. Closes #3416

### LINKED ISSUE #3416: support prometheus metrics in daemons and servers

rcarpa:
Mostly rely on prometheus_client library's functions for the job,
but had to do some small tweaks:
- The library has two type of supported value classes: MutexValue,
which is intended to be used in single-process (possible multi-threaded)
env; and MultiProcessValue, which is not thread-safe, but is backed by
a mmap-ed file, so can be shared between processes. As we use the apache
MPM module, which is both multi-processes and multi-threaded, had to
implement an additional value type and use it in servers.
- It can happen that mmap-ed *.db files are not correctly cleaned
up if an apache subprocess ends unexpectedly. Try to cleanup old files
when the endpoint is queried.

The prometheus_client library uses 'prometheus_multiproc_dir' env
variable to decide if it must switch to multiprocess mode. This is
backed soo deep into all the corners of the library, that it's not
reasonable to try to configure the behavior via a rucio config value.
As a result, to activate prometheus metrics in rucio servers, some
would have to set this counter-intuitive env variable.
Thankfully, if the env var is not set, the message output by
'generate_latest()' is quite clear about what has to be done.

Add a new /metrics endpoint, which can either be used as any other
route on the same port, or on a completely separate port via a
separate WSGI application.

### DIFF
--- a/lib/rucio/core/monitor.py
+++ b/lib/rucio/core/monitor.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Copyright 2013-2021 CERN
+# Copyright 2013-2022 CERN
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


bari12:
@maany @mlassnig review please 😄

mlassnig:
> @maany @mlassnig review please smile

I mean.. it looks okay to me code-wise, have to trust @rcarpa here that this is needed :-)

bari12:
Then you know which button to click 😆

mlassnig (approved):

mlassnig:
> Then you know which button to click laughing

image

PR #5184, opened 2022-01-26
Extensions: Remove the forecast extension

### LINKED ISSUE #5172: Remove forecast extension and numpy as a dependency
Motivation
----------
The forecast extension is not used anymore, thus it should be removed from the codebase along with the numpy dependency.

Modification
------------
- Remove `lib/rucio/extensions/forecast.py` and references to it in the code.
- Remove numpy as a dependency

joeldierkes:
The forecast extension is not used anymore. This commit deprecates it. It also
deletes numpy as dependency, since it was only used by the forecast extension.


bari12 (approved):

amaltaro:
Hello, I might have missed it in the release notes. In case I didn't, then I would like to suggest to make a clear reference to these (python) client API changes in the future - in the release notes - especially when it's about removing functionality.

I understand that Rucio provides gazillions of client APIs and it's easy to miss something, but I thought it was worth to mention it. Thanks!

bari12:
Hi @amaltaro 
It was somewhat included in the 1.28 release notes
https://github.com/rucio/rucio/releases/tag/1.28.0

But I agree that it is not very clear and should probably be included as an important change.

Apologies, will add it next time.

Best,
Martin

PR #5202, opened 2022-01-28
 Dark reaper improvements due to DB load : Closes #5192

### LINKED ISSUE #5192: Dark reaper improvements due to DB load
Motivation
----------
The dark-reaper is the top-query on the ATLAS database, causing a significant fraction of the DB load, although it is a rather unimportant service. There is just too many executions of the cycles and lots of retries.

This is related to #4744 

Modification
------------
1. `rucio.core.quarantined_replica.list_rses` is a very ambigous name. Rename to `list_rses_with_quarantined_replicas`
2. The `list_quarantined_replicas` command itself removes replicas which also exist in the `replicas` table. But this causes the RSE to always show up in each dark-reaper loop, although it is never doing anything. `list_quarantined_replicas` should directly remove `quarantined_replicas` from the table if they have a replica in the `replicas` table. No need to treat them in the `dark-reaper`
3. Sleep times can be increased to 5 minutes, at least.

cserf:
Dark reaper improvements due to DB load : Closes #5192

Change `rucio.core.quarantined_replica.list_rses` to `rucio.core.quarantined_replica.list_rses_with_quarantined_replicas`

Remove "real replicas" from the list of quarantine replicas

Increase sleep time to 300 seconds

### DIFF
--- a/lib/rucio/core/quarantined_replica.py
+++ b/lib/rucio/core/quarantined_replica.py
@@ -23,14 +23,16 @@
 # - Benedikt Ziemons <benedikt.ziemons@cern.ch>, 2020
 # - Rahul Chauhan <omrahulchauhan@gmail.com>, 2021
 # - Martin Barisits <martin.barisits@cern.ch>, 2021
+# - Cedric Serfon <cedric.serfon@cern.ch>, 2022
 
 import datetime
 
-from sqlalchemy import and_, or_, exists, not_
-from sqlalchemy.sql.expression import select, false
+from sqlalchemy import and_, or_
+from sqlalchemy.sql.expression import false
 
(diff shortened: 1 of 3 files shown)


bari12:
Thanks! merging this with test-errors, since I want it in the `1.27.4` release.
Please run the `add_header` script next time for the header test to pass 😄

bari12 (approved):

joeldierkes:
This is the bug in the `add_header` script, resolved in #5199 . The headers should be fine, the PR was just not rebased.

bari12:
Doesn't the script also check the 
```
# Copyright 2016-2021 CERN
```
line? Which was not updated in this case.

joeldierkes:
You're right, it would've been updated by the script.

PR #5204, opened 2022-01-28
Transfers: delete fts-throttler. Closes #5203

### LINKED ISSUE #5203: delete conveyor-fts-throttler executable
Motivation
----------
conveyor-fts-throttler (not to be confused with conveyor-throttler), is an executable which is assumed to be dead code. It's not tested at all, and a recent discovery (#4979) showed that it doesn't do anything when called. 

Modification
------------
Remove the code from rucio

rcarpa:
It looks very much like dead code.


joeldierkes (approved):
lgtm

bari12:
This goes into `1.28` 
I don't think anyone uses it anyway though.

PR #5206, opened 2022-02-01
Deletion: rework list_and_mark_unlocked_replicas. Closes #4897

### LINKED ISSUE #4897: Non optimal use of the limit parameter in list_and_mark_unlocked_replicas

rcarpa:
Rework the workflow for getting deletion replicas. Refer to individual commits for details.

### DIFF
--- a/lib/rucio/core/replica.py
+++ b/lib/rucio/core/replica.py
@@ -2049,17 +2049,18 @@ def list_and_mark_unlocked_replicas(limit, bytes_=None, rse_id=None, delay_secon
             filter(and_(models.RSEFileAssociation.scope == scope, models.RSEFileAssociation.name == name, models.RSEFileAssociation.rse_id != rse_id)).one()
 
         if replica_cnt[0] > 1:
+            if tombstone != OBSOLETE and only_delete_obsolete:
+                break
+
+            if needed_space is not None and total_bytes > needed_space:
+                break
+
             if state != ReplicaState.UNAVAILABLE:
-                if tombstone != OBSOLETE:
-                    if only_delete_obsolete:
-                        break
-                    if needed_space is not None and total_bytes > needed_space:
-                        break
                 total_bytes += bytes_
 
-                total_files += 1
-                if total_files > limit:
-                    break
+            total_files += 1
+            if total_files > limit:
+                break
 
             rows.append({'scope': scope, 'name': name, 'path': path,
                          'bytes': bytes_, 'tombstone': tombstone,
@@ -2075,12 +2076,14 @@ def list_and_mark_unlocked_replicas(limit, bytes_=None, rse_id=None, delay_secon
                             models.Request.name == name)).one()
 
             if request_cnt[0] == 0:
-                if tombstone != OBSOLETE:
-                    if only_delete_obsolete:
-                        break
-                    if needed_space is not None and total_bytes > needed_space:
-                        break
-                total_bytes += bytes_
+                if tombstone != OBSOLETE and only_delete_obsolete:
+                    break
+
+                if needed_space is not None and total_bytes > needed_space:
+                    break
+
+                if state != ReplicaState.UNAVAILABLE:
+                    total_bytes += bytes_
 
                 total_files += 1
                 if total_files > limit:
--- a/lib/rucio/core/replica.py
+++ b/lib/rucio/core/replica.py
@@ -44,6 +44,7 @@
 
 import heapq
 import logging
+import math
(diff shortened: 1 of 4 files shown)


bari12:
- I agree with the changes in the first commit. I am not entirely sure why it was done differently before. @cserf ?
- The second commit looks ok, but the whole `(2 * limit), math.ceil(1.25 * limit)` doesn't feel very good. I don't have a better idea and perhaps this is just not doable due to the reasons you explained in the ticket. I think it's still an improvement though.
@cserf ?

lib/rucio/core/replica.py
```diff
@@ -2018,6 +2022,146 @@ def list_and_mark_unlocked_replicas(limit, bytes_=None, rse_id=None, delay_secon
...
+    ).where(
+        or_(models.RSEFileAssociation.state.in_((ReplicaState.AVAILABLE, ReplicaState.UNAVAILABLE, ReplicaState.BAD)),
+            and_(models.RSEFileAssociation.state == ReplicaState.BEING_DELETED, models.RSEFileAssociation.updated_at < datetime.utcnow() - timedelta(seconds=delay_seconds)))
+    ).outerjoin(
+        models.Source,
+        and_(models.RSEFileAssociation.scope == models.Source.scope,
+             models.RSEFileAssociation.name == models.Source.name,
+             models.RSEFileAssociation.rse_id == models.Source.rse_id)
+    ).where(
+        models.Source.scope.is_(None)  # Only try to delete replicas if they are not used as sources in any transfers
```
cserf, line 2060: This is what we had in the past, but had to relax it because it can completely block the deletion of some files, e.g. : you have a file on RSE A and RSE B and a request to transfer to RSE C that is unavailable. In that case there is a row in the source table for RSE A and B, and there is no way to delete at least one of these copies until RSE C is back online.

lib/rucio/core/replica.py
```diff
@@ -2018,6 +2022,146 @@ def list_and_mark_unlocked_replicas(limit, bytes_=None, rse_id=None, delay_secon
...
+    ).order_by(
+        models.RSEFileAssociation.tombstone
+    ).with_for_update(
+        skip_locked=True,
+        # oracle: we must specify a column, not a table; however, it doesn't matter which column, the lock is put on the whole row
```
cserf, line 2065: Oh. Didn't know about this. Good to know

lib/rucio/core/replica.py
```diff
@@ -2018,6 +2022,146 @@ def list_and_mark_unlocked_replicas(limit, bytes_=None, rse_id=None, delay_secon
...
+        # postgresql/mysql: sqlalchemy driver automatically converts it to a table name
+        # sqlite: this is completely ignored
+        of=models.RSEFileAssociation.scope,
+    )
+
+    for chunk in chunks(session.execute(stmt).yield_per(2 * limit), math.ceil(1.25 * limit)):
```
cserf, line 2071: Magic numbers ;) ?

cserf (commented):
The logic looks good, now we need to evaluate at scale

lib/rucio/core/replica.py
rcarpa: I'm not sure what you mean by "relax". To my understanding this is just literal translation of the current behavior : https://github.com/rucio/rucio/blob/e7f1f44e00f44349a146b8e9e93ab942996896da/lib/rucio/core/replica.py#L2044 

Did I make a mistake and changed the current behavior in some way I didn't realize?

lib/rucio/core/replica.py
rcarpa: Yes, but probably not any more magic than the 1000 in the existing code: 
https://github.com/rucio/rucio/blob/e7f1f44e00f44349a146b8e9e93ab942996896da/lib/rucio/core/replica.py#L2055

lib/rucio/core/replica.py
bari12: This check was re-introduced by us last year, we did have to relax it before, that's what Cedric mentions.
But I think it is fine now, this PR does not change that behaviour.

bari12:
Please rebase, unless there are any more comments, it's gtg

rcarpa:
I rebased; but I also added one additional commit. I thought this change was already included in this PR, but I only pushed it to one of my personal branches. Please review the last commit too.

bari12 (approved):

Issue #4793, opened 2021-08-20
Optimize database interactions


bari12:
Motivation
----------

Identify common queries where optimisations can be done, evaluate the optimisation and implement them.

- executeMany queries
- Temporary tables + JOINs, instead of many OR-chunked queries
- etc.

- [x] #5130
- [x] #5441

PR #5214, opened 2022-02-02
Core & Internals: avoid calling session.remove() after commit. #4793

rcarpa:
It will save us an empty rollback on each successful transactional
session.


rcarpa:
This change has absolutely no impact on the database.

PR #5217, opened 2022-02-04
Clients: Add client cert and key env variables #5216

### LINKED ISSUE #5216: user environment variables to override client_cert and client_key of rucio.cfg
Motivation
----------
Having the **environment variable** CLIENT_CERT and CLIENT_KEY evaluated by rucio clients (overriding defaults in a system-wide config file's _client_cert_ and _client_key_) would allow implementing of system-wide defaults (e.g. in /etc/profile) pointing to each user's individual certificate stored in his/her home directory. 

Rucio version is: 1.27.3

Modification
------------
Have those two variables checked (probably in baseclient.py), and set _client_cert_ and _client_key_ accordingly.

joeldierkes:
If the user authenticates via the X509 protocol, a valid certification and key
file has to be provided. Currently, the path to these files can only be set via
the config file.

This commit introduces two environment variables to set the path of the
respective files. If these variables are not provided, the ones specified in the
config will be used.

### DIFF
--- a/lib/rucio/client/baseclient.py
+++ b/lib/rucio/client/baseclient.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Copyright 2012-2021 CERN
+# Copyright 2012-2022 CERN
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


bari12 (approved):

bari12:
Merging as a patch

PR #5239, opened 2022-02-15
Consistency checks: Stop relying on AGIS to get RSEs information, use…

### LINKED ISSUE #5226: Drop references to the AGIS-API endpoint to fix the auditor
Motivation
----------
There are several references to the atlas-agis-api.cern.ch endpoint in the codebase.
For example, the auditor daemon relies on it to fetch information on the RSEs and and to download the site storage reports.
With the shutdown of AGIS this does not work anymore and so the auditor, the way we currently deploy it, has become ineffective. 

Modification
------------
We could either replace the AGIS endpoint with [CRIC](https://atlas-cric.cern.ch/api/atlas/ddmendpoint/query/?json) or, better, not rely on external services at all and expolit the rucio.client.rseclient classes to expose the same information to the auditor. The latter has been tested and it showed to work.

faluchet:
… core API instead - Fix #5226

### DIFF
--- a/lib/rucio/common/dumper/__init__.py
+++ b/lib/rucio/common/dumper/__init__.py
@@ -28,7 +28,6 @@
 import contextlib
 import datetime
 import gzip
-import json
 import logging
 import os
 import re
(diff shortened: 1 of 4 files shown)


dchristidis (changes requested):
```python
from rucio.client.rseclient import RSEClient
```

Forgive me, I do not believe this to be the right approach. No Rucio daemon should rely on the Client. Instead, it should use the Core directly.

dchristidis (approved):

PR #5259, opened 2022-02-18
 Set the itsdangerous module version prior to 2.1.0 #5258

### LINKED ISSUE #5258: Flask fails after itsdangerous module’s update
Motivation
----------
As described in https://github.com/pallets/flask/issues/4455, the `itsdangerous` package version 2.1.0 breaks Flask 1.1.2.

Modification
------------
Bump the Flask version to 1.1.4 as proposed by the flask issue. This is just a minor version bump and will fix the dependency version mismatch.

joeldierkes:
The new `itsdangerous` package version 2.1.0 is incompatible with the Flask
version 1.1.2.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -33,6 +33,7 @@ geoip2==4.1.0                                               # GeoIP2 API (for IP
 oauth2client==4.1.3                                         # OAuth 2.0 client library
 retrying==1.3.3                                             # general-purpose retrying library to simplify the task of adding retry behavior to just about anything
 redis==3.5.3                                                # Python client for Redis key-value store
+itsdangerous<2.1.0                                          # Flask dependency, broken support since 2.1.0 https://github.com/pallets/flask/issues/4455
 Flask==1.1.2                                                # Python web framework
 oic==1.2.1                                                  # for authentication via OpenID Connect protocol
 prometheus_client==0.9.0                                    # Python client for the Prometheus monitoring system
--- a/setuputil.py
+++ b/setuputil.py
@@ -96,6 +96,7 @@ server_requirements_table = {
         'oauth2client',
         'retrying',
         'redis',
+        'itsdangerous',
         'flask',
         'oic',
         'prometheus_client',


requirements.txt
```diff
@@ -34,6 +34,7 @@ oauth2client==4.1.3                                         # OAuth 2.0 client l
 retrying==1.3.3                                             # general-purpose retrying library to simplify the task of adding retry behavior to just about anything
 redis==3.5.3                                                # Python client for Redis key-value store
 Flask==1.1.2                                                # Python web framework
+itsdangerous==2.0.1                                         # Flask dependency, broken support after 2.0.1 https://github.com/pallets/flask/issues/4455
```
rcarpa: I thought we agreed to just bump Flask to 1.1.4. Did I missunderstand something?

joeldierkes: Did that, same error as before. https://github.com/rucio/rucio/actions/runs/1864534929

rcarpa: The error seems different : ``` ImportError: cannot import name 'soft_unicode' from 'markupsafe' ```

joeldierkes: Oh, yes, you are right. https://github.com/pallets/flask/issues/4456 is the tracking issue.

Most comment suggest to not use an unsupported Flask version...
However, this would mean that we have to bump Flask by one major release, do we want to do that now?
From my POV, freezing the faulting dependency makes more sense. The pipeline would be intact afterwards and we can upgrade the Flask version some when else, when we have time to test it.

joeldierkes: What are your thoughts on that @rcarpa @maany ?

bari12: Better freeze the sub-dependency in that case. I wouldn't want to do a major dependency upgrade within the release line.

maany: Also, let's list `itsdangerours` ( <= instead of ==) before `Flask`. So, something like:
```
itsdangerous<=2.0.1 
Flask==1.1.2                                                
```

joeldierkes:
```
itsdangerous<2.1.0 
Flask==1.1.2  
```
seems okay too

rcarpa (dismissed):

requirements.txt
bari12: I think you will have to add it
https://github.com/rucio/rucio/blob/2930272ff394b28ed9f4d81e1b9f898d8f418951/setuputil.py#L81-L102
too?

joeldierkes: Yes that makes sense

maany (approved):
LGTM

PR #5248, opened 2022-02-16
Metadata: add external metadata provider mongo #5143

robbarnsley:

PR #5261, opened 2022-02-18
enhancement: requirements for mongodb to be included as an external metadata provider in CI pipeline #5260

### LINKED ISSUE #5260: add pymongo to requirements & mongodb container to CI pipeline
For #5143 to pass unit tests, a mongo container needs to be added to the CI pipeline. pymongo also needs to be added as a requirement.

robbarnsley:
adding mongodb service to integration test suite

### DIFF
--- a/etc/docker/dev/docker-compose-storage-externalmetadata.yml
+++ b/etc/docker/dev/docker-compose-storage-externalmetadata.yml
@@ -1,0 +1,123 @@
+version: "3.5"
+services:
+  rucio:
+    image: docker.io/rucio/rucio-dev:py3
+    extra_hosts:
+      - "ruciodb:127.0.0.1"
+      - "graphite:127.0.0.1"
+      - "fts:127.0.0.1"
+      - "ftsdb:127.0.0.1"
+      - "xrd1:127.0.0.1"
+      - "xrd2:127.0.0.1"
+      - "xrd3:127.0.0.1"
+      - "xrd4:127.0.0.1"
+      - "minio:127.0.0.1"
+      - "activemq:127.0.0.1"
+      - "ssh1:127.0.0.1"
+      - "mongo:127.0.0.1"
+    ports:
+      - "127.0.0.1:8443:443"
+      - "127.0.0.1:5432:5432"
+      - "127.0.0.1:8080:80"
+      - "127.0.0.1:8446:8446"
+      - "127.0.0.1:8449:8449"
+      - "127.0.0.1:3306:3306"
+      - "127.0.0.1:1094:1094"
+      - "127.0.0.1:1095:1095"
+      - "127.0.0.1:1096:1096"
+      - "127.0.0.1:1097:1097"
+      - "127.0.0.1:9000:9000"
+      - "127.0.0.1:61613:61613"
+      - "127.0.0.1:2222:22"
+      - "127.0.0.1:27017:27017"
+    volumes:
+      - ../../../tools:/opt/rucio/tools:Z
+      - ../../../bin:/opt/rucio/bin:Z
+      - ../../../lib:/opt/rucio/lib:Z
+    environment:
+      - X509_USER_CERT=/opt/rucio/etc/usercert.pem
+      - X509_USER_KEY=/opt/rucio/etc/userkey.pem
+      - RDBMS=postgres11
+  ruciodb:
+    image: docker.io/postgres:11
+    network_mode: "service:rucio"
+    environment:
+      - POSTGRES_USER=rucio
+      - POSTGRES_DB=rucio
+      - POSTGRES_PASSWORD=secret
+    command: ["-c", "fsync=off","-c", "synchronous_commit=off","-c", "full_page_writes=off"]
+  graphite:
+    image: docker.io/graphiteapp/graphite-statsd
+    network_mode: "service:rucio"
+  fts:
+    image: docker.io/rucio/fts
+    network_mode: "service:rucio"
+  ftsdb:
+    image: docker.io/mysql:5
+    network_mode: "service:rucio"
+    environment:
+      - MYSQL_USER=fts
+      - MYSQL_PASSWORD=fts
+      - MYSQL_ROOT_PASSWORD=fts
+      - MYSQL_DATABASE=fts
+  xrd1:
+    image: docker.io/rucio/xrootd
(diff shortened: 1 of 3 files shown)


robbarnsley:
This is a bit of a guess based on a cursory look of how the CI pipeline works. If there's a different, preferred way of doing this, let me know.

maany:
Hey @robbarnsley, thanks for starting to work on this. I'm happy to point you to the right files that would require further changes. :) 

For starters, would you like to add MongoDB to the autotest pipeline or the integration test pipeline, or both?

robbarnsley:
Thanks @maany. 

I'll be honest, i'm not entirely sure. I am working on #5248 & need a mongodb container to perform unit tests against. I suspect it's just autotests I need?

maany:
Hey @robbarnsley, yes the autotests is way to go here :) 
Here's the description of the workflow and the files you'll have to modify along the way to get things working.

Step 1: Configure the autotest build matrix
-------
You'll have to make some changes to [matrix.yml](https://github.com/rucio/rucio/blob/master/etc/docker/test/matrix.yml#L39)

There's 2 options here:
### Option 1: create a new test suite 
This is useful if you only want to run a subset of tests against mongodb.
Let's call the new suite `metadata` for this example
 https://github.com/rucio/rucio/blob/f5541e54ad2c8998f45067daa46667434bc78bae/etc/docker/test/matrix.yml#L26
In this case, you'd add a new suite under the configuration of the suite. The new suite would probably look like
```
suites:
  - id: metadata
    RDBMS:
         - mongodb4
```
### Option 2: Add mongodb4 as a RDBMS to the `all` suite
This is useful if you want to run all the tests against mongodb. In this case you'll add `mongodb` to the following field:
https://github.com/rucio/rucio/blob/f5541e54ad2c8998f45067daa46667434bc78bae/etc/docker/test/matrix.yml#L39-L44

Step 2: Start your mongodb container
----------------------------
You'll have to modify the [`before_script`](https://github.com/rucio/rucio/blob/master/tools/test/before_script.sh#L59-L77) to actually start the `mongodb4` container. Take a look at some of the examples to come up with your version https://github.com/rucio/rucio/blob/f5541e54ad2c8998f45067daa46667434bc78bae/tools/test/before_script.sh#L59-L77

Here, you'll also configure `rucio.cfg` that should be used in your test configuration.
**NOTE** I am not sure how you'll be using the `mongodb` container, but in any case please also use the `rucio.cfg` make sure that a db session is made available to the test environment. Take a look at:
https://github.com/rucio/rucio/blob/f5541e54ad2c8998f45067daa46667434bc78bae/lib/rucio/db/sqla/session.py#L170-L199

**NOTE 2**Also, if required ( in case the `session.py` is not how you wish to configure mongodb), we can discuss how to create a new fixture in pytest so that the tests you write could connect to the database with a fixture instead.

Step 3: Configure your test-suite
---------
If you created a new test-suite in Step 1 or if you need to do some additional configuration before the tests are run, you can handle your `metadata` test suite or the additional configuration in the `test.sh` file:
https://github.com/rucio/rucio/blob/f5541e54ad2c8998f45067daa46667434bc78bae/tools/test/test.sh#L59-L64

I think that should be enough for the autotests to run your mongodb container. For more information or any specific details, please ping me here or on Slack. We can also get on a call ;)

etc/docker/dev/docker-compose-storage-externalmetadata.yml
```diff
@@ -0,0 +1,123 @@
...
+      - "fts:127.0.0.1"
+      - "ftsdb:127.0.0.1"
+      - "xrd1:127.0.0.1"
+      - "xrd2:127.0.0.1"
+      - "xrd3:127.0.0.1"
+      - "xrd4:127.0.0.1"
+      - "minio:127.0.0.1"
+      - "activemq:127.0.0.1"
+      - "ssh1:127.0.0.1"
+      - "mongo:127.0.0.1"
```
maany, line 17: Cool, we'd need this file in either case ( autotest/integration-tests)

tools/test/before_script.sh
```diff
@@ -148,6 +148,10 @@ elif [ $RDBMS == "sqlite" ]; then
     RESTART_HTTPD=1
 fi
 
+# external metadata providers
+docker $CONTAINER_RUNTIME_ARGS run -d $CONTAINER_RUN_ARGS docker.io/mongo:5.0
```
maany: This would always start a new mongodb container, please take a look at the comment I've left on this PR on making it specific to particular test cases ;)

maany (dismissed):
Good start. Let's edit some more files that I've mentioned in the PR comments and we should be close to getting this working :)

robbarnsley:
Hi @maany, thank you for the help.

Could I just clarify: this would be an additional database for metadata storage only - it's not meant as a replacement tech for the "core" rucio database. I think that's why I originally put it always starting a new mongodb container in `before_script`. It would need to be made available to all core database flavours.

Correct me if i'm wrong, but i'm not sure about some of the pieces here,  - it's not an RDBMS for example, and consequently there's no interface in sqla for it (it's nosql).

maany:
Hey @robbarnsley, understood, and thanks for the clarifications. Can we get on a call to discuss how to proceed ( would be easier)? Let's continue on Slack to set it up.

> Correct me if i'm wrong, but i'm not sure about some of the pieces here, - it's not an RDBMS for example, and consequently there's no interface in sqla for it (it's nosql).

Indeed, there is no sqla adapter for mongodb, which would also defeat the purpose of no-sql database.

I was not clear in this part sorry. I meant, in addition to starting the container, another step would be to configure the PyMongo connector somewhere in the rucio codebase. If you need to do it for tests only to begin with, we could create a new fixture that exposes the PyMongo connection ;). Just an idea.

maany:
Hey @robbarnsley , regarding your question on how to make tests run only in `integration-test` suite; 
You can decorate the test with `@skip_rse_tests_with_accounts` to make them run only in the `integration-test` environment.
This actually made me think of a better way to do this and I will implement it sometime later to make it more intuitive :)

robbarnsley:
May need some pointers as to why the release-patch-setup test is failing.

bari12:
Hi @robbarnsley 
This needs a rebase 😄

bari12 (approved):

PR #5262, opened 2022-02-18
traces:  add sfo2eos trace schema #5255

### LINKED ISSUE #5255: add schema for sfo2eos traces
Motivation
----------
Similar to #5156, the incoming traces of `"eventType": "sfo2eos"`, do not have a schema to be validated against. This leads to NoneType errors, as seen in #5156.

Modification
------------
1. add a schema for sfo2eos eventTypes.
2. create a generic schema that new eventTypes can be validated against
3. automatically open/update a github issue when new event types are detected or store it in a database

As discussed with @bari12 , for new eventTypes, we'll display a warning and hide the dataset

maany:
1. handle `sfo2eos` eventTypes
2. capture unsupported `eventTypes` and display just the error instead of the whole stacktraces.

### DIFF
--- a/lib/rucio/core/trace.py
+++ b/lib/rucio/core/trace.py
@@ -360,9 +360,12 @@ def validate_schema(obj):
         if obj and 'eventType' in obj:
             event_type = SCHEMAS.get(obj['eventType'].lower())
             if not event_type:
-                raise TypeError(f"Trace schema for eventType {obj['eventType']} is not currently supported.")
+                validation_error = ValidationError(message=f"Trace schema for eventType {obj['eventType']} is not currently supported.")
+                validation_error.cause = "SCHEMA_NOT_FOUND"
+                raise validation_error
             validate(obj, SCHEMAS.get(obj['eventType'].lower()), format_checker=FORMAT_CHECKER)
     except ValidationError as error:
-        raise InvalidObject(error)
-    except TypeError as error:
-        LOGGER.error(error)
+        if error.cause is "SCHEMA_NOT_FOUND":
+            LOGGER.error(error)
+        else:
+            raise InvalidObject(error)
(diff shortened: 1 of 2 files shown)


lib/rucio/core/trace.py
```diff
@@ -329,12 +352,17 @@ def validate_schema(obj):
...
-    :raises: ValidationError
+    :raises: InvalidObject
     """
     obj = json.loads(obj)
 
     try:
         if obj and 'eventType' in obj:
+            event_type = SCHEMAS.get(obj['eventType'].lower())
+            if not event_type:
+                raise TypeError(f"Trace schema for eventType {obj['eventType']} is not currently supported.")
```
bari12: I am wondering why we should differentiate here and raise the `TypeError` and not instead also go for the `InvalidObject`, which will then be handled here
https://github.com/rucio/rucio/blob/c9ea1c2dcb2a64f9e2ccac49d021914b863a684e/lib/rucio/core/trace.py#L296

In the end, it's just a schema validation error, just that in this case the schema does not exists, while in other cases it does not match attributes.

mlassnig: TypeError should only be used for Python types IMHO, ValidationError makes more sense here.

maany: @bari12, since the idea is to reduce log size and just print an error or warning instead of the stack traces when a schema is missing in 
https://github.com/rucio/rucio/blob/c9ea1c2dcb2a64f9e2ccac49d021914b863a684e/lib/rucio/core/trace.py#L227-L241
I raise and handle the `TypeError` as the `ValidationError` currently just raises an InvalidObject error and prints the stack trace, which defeats the purpose of reducing the logs

maany:
Closing and continuing in #5267 #5265

PR #5267, opened 2022-02-21
Traces: bulk updates for schemas #5265

### LINKED ISSUE #5255: add schema for sfo2eos traces

### LINKED ISSUE #5265: bulk updates for trace schemas
Motivation
----------

### url
Incoming traces do not have `url` set, which leads to the following error:
```
Failed validating 'type' in schema['properties']['url']:
   {'type': 'string'}

On instance['url']:
    None
```
### clientState
There's also new values that `clientState` can take up
1. SERVICE_ERROR
2. CP_TIMEOUT
3. COPY_ERROR
4. STAGEIN_ATTEMPT_FAILED
5. SourceNotFound
6. MISSINGOUTPUTFILE

### remoteSite
remoteSite currently validates against the regex:
```
{'description': 'RSE name',
'pattern': '^([A-Z0-9]+([_-][A-Z0-9]+)*)$',
'type': 'string'}
```
but incoming traces have the value set to `unknown`

### scope
for `eventType: put_sm_a` 
```
Details: 'scope' is a required property
```
or
regex pattern fails as `SCOPE_LENGTH` > 25

Modification
------------
1. add `unspecified` as a default value for url
1. Add `SERVICE_ERROR` `CP_TIMEOUT` `COPY_ERROR` `STAGEIN_ATTEMPT_FAILED` `SourceNotFound` `MISSINGOUTPUTFILE` to clientState enum
1. Change remoteSite schema to `{"type": "string"}`
1. Remove scope as a required property from `put_sm_a`, 
2. Bump `SCOPE_LENGTH` to 1000 characters

maany:
Close #5265

### DIFF
--- a/lib/rucio/common/schema/generic.py
+++ b/lib/rucio/common/schema/generic.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Copyright 2017-2021 CERN
+# Copyright 2017-2022 CERN
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -23,6 +23,7 @@
 # - Eli Chadwick <eli.chadwick@stfc.ac.uk>, 2020
 # - Radu Carpa <radu.carpa@cern.ch>, 2021
 # - David Población Criado <david.poblacion.criado@cern.ch>, 2021
+# - Mayank Sharma <imptodefeat@gmail.com>, 2022
 
 from jsonschema import validate, ValidationError
 
@@ -197,7 +198,9 @@ CLIENT_STATE = {
     "description": "Client state",
     "type": "string",
     "enum": ['DONE', 'FAILED', 'PROCESSING', 'ALREADY_DONE', 'FILE_NOT_FOUND', 'FOUND_IN_PCACHE', 'DOWNLOAD_ATTEMPT',
-             'FAIL_VALIDATE', 'FOUND_ROOT', 'ServiceUnavailable']
+             'FAIL_VALIDATE', 'FOUND_ROOT', 'ServiceUnavailable', 'SERVICE_ERROR', 'CP_TIMEOUT', 'COPY_ERROR',
+             'STAGEIN_ATTEMPT_FAILED', 'SourceNotFound', 'MISSINGOUTPUTFILE', 'MD_MISMATCH', 'CHECKSUMCALCULATIONFAILURE',
+             'MISSINGINPUT', 'MISSING_INPUT']
 }
 
 RULE = {"description": "Replication rule",
--- a/lib/rucio/core/trace.py
+++ b/lib/rucio/core/trace.py
@@ -39,7 +39,7 @@ from jsonschema import validate, ValidationError, draft7_format_checker
 
 from rucio.common.config import config_get, config_get_int
 from rucio.common.exception import InvalidObject
-from rucio.common.schema.generic import SCOPE, RSE, UUID, TIME_ENTRY, IPv4orIPv6, CLIENT_STATE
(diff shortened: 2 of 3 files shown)


lib/rucio/common/schema/generic.py
```diff
@@ -54,7 +55,7 @@
                      "T0 Export", "T0 Tape", "Upload/Download (Job)",
                      "Upload/Download (User)", "User Subscriptions", "Data Challenge"]}
 
-SCOPE_LENGTH = 25
+SCOPE_LENGTH = 25 * 40
```
bari12: Why?
This would certainly be a problem to change, since the models.py uses this value the instantiate the length of the database columns. So you increased this column size by a factor of 40 by this :-)

bari12 (changes requested):

lib/rucio/common/schema/generic.py
maany: I definitely went way overboard with the new size. But we need to bump this number a bit. Here's an example that fails (117 characters).
```
[Mon Feb 21 12:57:15.876101 2022] [wsgi:error] [pid 32:tid 140136349914880] [remote 10.100.143.192:58938] Failed validating 'pattern' in schema['properties']['scope']:
[Mon Feb 21 12:57:15.876103 2022] [wsgi:error] [pid 32:tid 140136349914880] [remote 10.100.143.192:58938]     {'description': 'Scope name',
[Mon Feb 21 12:57:15.876106 2022] [wsgi:error] [pid 32:tid 140136349914880] [remote 10.100.143.192:58938]      'pattern': '^[a-zA-Z_\\\\-.0-9]{1,25}$',
[Mon Feb 21 12:57:15.876108 2022] [wsgi:error] [pid 32:tid 140136349914880] [remote 10.100.143.192:58938]      'type': 'string'}
[Mon Feb 21 12:57:15.876111 2022] [wsgi:error] [pid 32:tid 140136349914880] [remote 10.100.143.192:58938]
[Mon Feb 21 12:57:15.876113 2022] [wsgi:error] [pid 32:tid 140136349914880] [remote 10.100.143.192:58938] On instance['scope']:
[Mon Feb 21 12:57:15.876117 2022] [wsgi:error] [pid 32:tid 140136349914880] [remote 10.100.143.192:58938]     'valid1.361107.PowhegPythia8EvtGen_AZNLOCTEQ6L1_Zmumu.merge.NTUP_PHYSVAL.e5112_s3214_d1731_r13252_p4889_tid27512986_00'
```

Shall I change `SCOPE_LENGTH` to 120?

bari12: This is not a valid scope, I think we need to separate the schema scope_length from the traces scope length.
In tis example, this is just wrong and the scope for that should be `valid1`, but that's not something we can enforce now.

bari12: This is not a valid scope, I think we need to separate the schema scope_length from the traces scope length.
In tis example, this is just wrong and the scope for that should be `valid1`, but that's not something we can enforce now.

maany: Aye, I can revert this tomorrow i.e. set SCOPE_LENGTH back to 25 and check if the logs blow up again.

bari12 (approved):

PR #5269, opened 2022-02-21
Transfers: disentangle fts3 code from poller/receiver. Closes #4873. Closes #4576

### LINKED ISSUE #4576: Allow for additional fields from transfer response on the poller daemon for messages
Motivation
----------
Currently `request.set_request_state` and `request.add_monitor_message` only allow for static list of fields from the transfer response to add to message.

Modification
------------
Allow passing of other fields from transfer response message.

### LINKED ISSUE #4873: Disentangle fts3 specific code from conveyor poller and receiver
Motivation
----------
Make most of poller and receiver generic and interface it with the transfertool which will be in charge of message reception.

rcarpa:
Refactor poller and receiver with the goal to get rid of fts-specifics in code paths which should be transfertool-agnostic.

Refer to individual commits for more details.

### DIFF
--- a/lib/rucio/daemons/conveyor/common.py
+++ b/lib/rucio/daemons/conveyor/common.py
@@ -407,7 +407,7 @@ def __create_missing_replicas_and_requests(
                initial_request_id, transfer_path[0].src, transfer_path[-1].dst)
         try:
             for request_id in created_requests:
-                set_request_state(request_id=request_id, new_state=RequestState.FAILED,
+                set_request_state(request_id=request_id, state=RequestState.FAILED,
                                   err_msg="Cancelled hop in multi-hop", session=session)
         except UnsupportedOperation:
             logger(logging.ERROR, '%s: Multihop : Cannot cancel all the parent requests : %s', initial_request_id, str(created_requests))
@@ -435,7 +435,7 @@ def submit_transfer(transfertool_obj, transfers, job_params, submitter='submitte
             return
         except Exception:
             logger(logging.ERROR, 'Failed to prepare requests %s state to SUBMITTING. Mark it SUBMISSION_FAILED and abort submission.' % [str(t.rws) for t in transfers], exc_info=True)
-            set_request_state(request_id=transfer.rws.request_id, new_state=RequestState.SUBMISSION_FAILED)
+            set_request_state(request_id=transfer.rws.request_id, state=RequestState.SUBMISSION_FAILED)
             return
 
     try:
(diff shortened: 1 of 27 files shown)


lib/rucio/daemons/conveyor/common.py
```diff
@@ -157,6 +165,254 @@ def run_conveyor_daemon(once, graceful_stop, executable, logger_prefix, partitio
                     activity_next_exe_time[activity] = time.time() + 1
 
 
+@transactional_session
+def next_transfers_to_submit(total_workers=0, worker_number=0, partition_hash_var=None, limit=None, activity=None, older_than=None, rses=None, schemes=None,
```
bari12, line 171: One guiding principle (Although we broke it several times) was always to keep database-connected methods (e.g. everything with a session annotator) in the core. Is it needed to have this in conveyor/common? (Circular dependency?) I am not saying this is wrong or it should be changed, mostly wondering if we should try to "hold" this in core?

bari12 (commented):
Otherwise this looks GTG for me, just one comment.

lib/rucio/daemons/conveyor/common.py
rcarpa: yes, this function calls replica_core.create_replicas  ; because of that, transfer_core cannot be used in rules_core without removing access to replicas from transfers_core. A obvious solution would be to add some other "core" file and move this function there. "core/conveyor.py" ? :D  "core/transfer_execution.py" ? Or I can try to see about splitting core/rules or core/replicas into two files to avoid the circular import; which may be a better solution, but I don't know how big of a rabbit hole i'll be getting myself into.

bari12: No, let's leave it like this then. I don't find it easier to split these core files. Do I have to check core/transfer.py or core/transfer_execution.py now? 😄 Then I might as well just check conveyor/common.py. 👍

rcarpa: I didn't know about this desire to put database-connected methods in core. I'll keep it in my head and, if I find an opportunity, I'll try to get rid of such accesses from outside the core in the future.

bari12: It's not a rule written in stone, but that was always a bit the "idea" behind it. (Maybe we should write this down somewhere 😆 )

bari12 (approved):

bari12:
Merged as feature for 1.28

PR #5270, opened 2022-02-22
feature: Extend list_dataset_replicas_bulk to use the --deep flag #5024

ivmfnal:

### DIFF
--- a/lib/rucio/api/replica.py
+++ b/lib/rucio/api/replica.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Copyright 2013-2021 CERN
+# Copyright 2013-2022 CERN
 #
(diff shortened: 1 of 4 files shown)


bari12:
@ivmfnal Please use some more descriptive Pull request titles and commit names. I know that it is written in the ticket too, but it should give somebody reading the commit log an understanding of what was changed, without having to go to the ticket itself.
It is actually not a bad practice to go beyond that, and add some explanation text to the commit itself as well, which makes reading commit logs and debugging much easier. But we don't require this in Rucio (yet).

bari12:
Ping on that, This needs a rebase and if the tests continue to fail, please fix them.
Also please use a more descriptive PR title. 😄

PR #5272, opened 2022-02-22
Added test_judge_dataset_grouping_all()

ivmfnal:
Added test_judge_dataset_grouping_all() to test the case when a new dataset is added to an existing container with a rule with grouping=ALL

Also made some cosmetic changes to keep pylint happy

### DIFF
--- a/lib/rucio/tests/test_judge_evaluator.py
+++ b/lib/rucio/tests/test_judge_evaluator.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Copyright 2014-2021 CERN
+# Copyright 2014-2022 CERN
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


lib/rucio/tests/test_judge_evaluator.py
```diff
@@ -143,11 +146,57 @@ def test_judge_add_dataset_to_container(self):
...
+        attach_dids(scope, dataset1, files, self.jdoe)
+
+        # attach the dataset to the container
+        attach_dids(scope, parent_container, [{'scope': scope, 'name': dataset1}], self.jdoe)
+
+        # add rule to copy everything in this container to T1, use grouping=ALL
+        add_rule(dids=[{'scope': scope, 'name': parent_container}],
+                 account=self.jdoe,
+                 copies=1, rse_expression=self.T1, grouping='ALL',
+                 weight=None, lifetime=None, locked=False, subscription_id=None
```
nsmith-, line 177: As currently written, I think the test will be non-deterministic as there is a 1 in 3 chance the RSESelector will select the same RSE as previously used. Perhaps a way to make it more likely to fail is to assign a weight here such that it almost certainly goes to a specific T1 RSEs, say `MOCK1`, then update the weight after L180 to be almost sure that `MOCK3` would be chosen when the re-evaluator runs next unless `preferred_rse_ids` is properly set to `MOCK1`.

bari12: I don't think so. The rule here will always use the `rse1` since the data is already there. Since the new dataset is on `rse4` this should always be fine.

nsmith-: Yes for this line it will use `rse1`, but the issue I am trying to trigger happens during the RSE selection process in L192 where the re-evaluator runs now with the new child `dataset2`. There we hope that the selection process puts locks on `dataset2` at `rse1` but in practice what I observe in production is it may sometimes put locks at `rse3`. In this test, if the grouping is not respected it still might select `rse1` by chance. Hence why I was advocating a weight in the rule.

bari12: I see what you mean. 👍 
Currently there is two things going wrong here, which Igor already identified.
1. The re-evaluation of a dataset/container being added to a container at all. This might not be triggered.
2. The correct identification of the RSEs already selected.
This test will correctly test for 1, but not necessarily for 2.

PR #5274, opened 2022-02-22
Select containers based on release brach for Integration Tests (1.26) Close #5256

### LINKED ISSUE #5256: Use LTS base containers for integration tests in release branches.
Motivation
----------

Modification
------------

maany:


joeldierkes (dismissed):
LGTM

bari12 (approved):

PR #5275, opened 2022-02-22
Select containers based on release brach for Integration Tests (1.27) Close #5256

### LINKED ISSUE #5256: Use LTS base containers for integration tests in release branches.

maany:


maany:
@bari12 I guess this will automatically make it to the 1.27 branch on the next release. If so, we can close this PR as there's already one for pushing these changes to the master branch.

maany:
Closing as this would automatically be part of 1.27 when the next release on master happens

PR #5276, opened 2022-02-22
Select containers based on release brach for Integration Tests (master) 

### LINKED ISSUE #5256: Use LTS base containers for integration tests in release branches.

maany:

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -35,10 +35,45 @@ jobs:
         uses: actions/checkout@v2
         with:
           repository: rucio/containers
+          fetch-depth: 0
       - uses: actions/checkout@v2
         name: Checkout rucio source
         with:
           path: dev/rucio
+          fetch-depth: 0
+      - name: Select tag for rucio containers
+        shell: bash
+        run: | 
+          # Change to cloned rucio/rucio repo
+          cd $GITHUB_WORKSPACE/dev/rucio
+          
+          # Get current branch and corresponding latest tag in rucio/rucio repo
+          BRANCH=$(git rev-parse --abbrev-ref HEAD)
+          OWNER="${{ github.repository_owner }}"
+          
+          if [ $OWNER != 'rucio' ]; then
+            echo "The workflow is running in user ${OWNER}'s fork. Fetching branches and tags from rucio/rucio instead."
+            git remote add rucio https://github.com/rucio/rucio
+            git fetch rucio --tags
+          fi
+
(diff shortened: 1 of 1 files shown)


joeldierkes (dismissed):
LGTM

maany:
Converting to draft to see if a simpler mechanism based on branch name's can be used to switch the containers repository to the correct branch.

maany:
So turns out we cannot use the solution based on branch names since the integration tests are run for `pull_requests`  whose source branch is almost never `master`  or `release-1.26-LTS`  etc.

When a `pull_request`  is made, GitHub actions use the branch from the author's fork.

For example, in this [run](https://github.com/maany/rucio/runs/5317878071?check_suite_focus=true), the `rucio/containers#master` branch will be used since `patch-5256-LTS_containers_integration_tests` does not exist in `rucio/containers` repo.   Here's the output from the run:
```
Checking if `rucio/containers` has branch `patch-5256-LTS_containers_integration_tests` 
Switching rucio/containers to branch master
Already on 'master'
Your branch is up to date with 'origin/master'.
```

Therefore, I think the solution I currently have is probably the one we should go ahead with.

.github/workflows/integration_tests.yml
```diff
@@ -35,10 +35,45 @@ jobs:
...
+        shell: bash
+        run: | 
+          # Change to cloned rucio/rucio repo
+          cd $GITHUB_WORKSPACE/dev/rucio
+          
+          # Get current branch and corresponding latest tag in rucio/rucio repo
+          BRANCH=$(git rev-parse --abbrev-ref HEAD)
+          OWNER="${{ github.repository_owner }}"
+          
+          if [ $OWNER != 'rucio' ]; then
```
maany, line 54: Added a check for `OWNER`, so that integration test runs in forked repositories can still use the correct `rucio/containers` tags. Forked repositories do not contain all the branches and hence the commit history available on `rucio/rucio`, therefore the selected `rucio/containers` can be from a different release line. For example, I did not have 1.26.x tags/branches in my fork, so the containers were picked up from the 1.23.x release line as that existed in my fork.

maany: Another note to keep in mind, in case we run into any issues in the future:
The current codebase will go through the commit history shown by the following command:

```
git log --oneline --decorate=short
```

The first tag it finds in the commit history (now->past commits), is used to fetch the correct dev container for integration tests.

bari12 (approved):

PR #5282, opened 2022-02-23
Transfers: store transfer paths in a new database table. Closes #5177

### LINKED ISSUE #5177: multihop handled badly on multiple rules
Motivation
----------

```
SRC -> DATADISK -> DST1
               |-> DST2
```
Assume a tape RSE `SRC` which has an unique copy of a file. Two requests (via same rule, or via different rules) for the file to be copied to rses `DST1` and `DST2`. A multihop via `DATADISK` is required to execute both these transfers. 

One transfer will be successfully scheduled. The second will fail continuously because of  
https://github.com/rucio/rucio/blob/80f8c5a96e06abb80e3e6c854b89298f16be6270/lib/rucio/core/transfer.py#L1567
In a case of a stage-in from tape, the first request may take long enough for the rule to be marked as stalled on the second request. 

Modification
------------

Find a way to handle this issue

rcarpa:
Until now, a multihop transfer didn't have a real database state.
Each hop of the multihop was a separate transfer request with almost
no relationship between individual requests (only a loosely coupled
json attribute). This behavior is source of multiple issues:
 - we have recurrent problems with intermediate hops living their
own life (their lifetime should be linked to the initial request
and they shouldn't be retried on their own).
 - multiple multi-hops sharing a common sub-path are not possible
and result in permanent failures of one of the transfers until the
second finishes.
 - it doesn't allow implementation of cross-tranfertool multihops

For these particular reasons, we introduce a table storing the
relationship between independent requests of a path. I change the
workflows to rely on joins with this table:
- (1) submitter only retrieves from database requests which don't have
any pending previous hops.
- (2) submitter is capable to detect that a sub-path of the current
multi-hop transfer already exists in the database. It only creates
the link with the existing sub-path and aborts submission. Submission
automatically resumes once the common sub-path is transferred thanks to
behavior mentioned in (1)
- (3) when an intermediate hop fails, the poller/receiver will also
fail the next requests in all affected paths.

### DIFF
--- a/lib/rucio/db/sqla/models.py
+++ b/lib/rucio/db/sqla/models.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Copyright 2015-2021 CERN
+# Copyright 2015-2022 CERN
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -31,6 +31,7 @@
 # - Matt Snyder <msnyder@bnl.gov>, 2021
 # - joaquinbogado <joaquinbogado@gmail.com>, 2021
 # - Ilija Vukotic <ivukotic@uchicago.edu>, 2021
+# - Radu Carpa <radu.carpa@cern.ch>, 2022
 
 import datetime
 import sys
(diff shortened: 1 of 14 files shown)


rcarpa:
This PR is based on top of #5269. Only the last commit is to be reviewed.

lib/rucio/db/sqla/models.py
```diff
@@ -1310,6 +1311,19 @@ class Request(BASE, ModelBase):
                    Index('REQUESTS_TYP_STA_TRA_ACT_IDX', 'request_type', 'state', 'transfertool', 'activity'))
 
 
+class TransferHop(BASE, ModelBase):
```
bari12, line 1314: Add this to https://github.com/rucio/rucio/blob/master/etc/sql/oracle/schema.sql as well

bari12 (changes requested):

lib/rucio/db/sqla/models.py
rcarpa: Done

bari12:
Merged as feature for `1.28`

bari12 (approved):

PR #5293, opened 2022-02-28
 Temp tables to delete replicas. #4793. Closes #5166

### LINKED ISSUE #5166: incorrect update of Container is_archive in reaper
Motivation
----------
Collection dids have an is_archive database field which should be true when at least one File DID from the collection is an archive. 
Otherwise it's failse. The reaper [has a clause](https://github.com/rucio/rucio/blob/d19db9eeb09440ae97477840fff57112f9b25620/lib/rucio/core/replica.py#L1974) which updates this field to False if the last archive file is removed. However, this clause doesn't correctly handle the recursive update of Containers after a dataset is updated to false.

Modification
------------

Fix this bug

rcarpa:
Use temporary tables to implement delete_replicas. 

I start by simply copying the existing delete_replicas code 
and using a configuration switch to activate the new behavior.

This PR is split into multiple commits for easier review. Each
commit migrates a small subset of existing queries to using
temp tables. One of the commits also fixes an existing bug.

### DIFF
--- a/lib/rucio/core/replica.py
+++ b/lib/rucio/core/replica.py
@@ -59,7 +59,7 @@ from traceback import format_exc
 import requests
 from dogpile.cache.api import NO_VALUE
 from six import string_types
-from sqlalchemy import func, and_, or_, exists, not_, update, delete, Column
+from sqlalchemy import func, and_, or_, exists, not_, update, delete, insert, Column
 from sqlalchemy.exc import DatabaseError, IntegrityError
 from sqlalchemy.orm import aliased
 from sqlalchemy.orm.exc import FlushError, NoResultFound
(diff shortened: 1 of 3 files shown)


lib/rucio/core/replica.py
```diff
@@ -92,6 +92,10 @@
 REGION = make_region_memcached(expiration_time=60)
 
 
+ScopeName = namedtuple('ScopeName', ['scope', 'name'])
+Association = namedtuple('Association', ['scope', 'name', 'child_scope', 'child_name'])
```
bari12, line 96: Should we move these to somewhere common?

bari12 (approved):
I think it's good, but hard to review. I mean I did not double check the `__cleanup_after_replica_deletion` compared to the previous one. I just assume the steps are correct 👍 
From the changes done with temp tables it looks really promising. Curious how it works out.

lib/rucio/core/replica.py
rcarpa: Yes, in the future, but I wouldn't put them in a common place until we start needing them in multiple places.

PR #5306, opened 2022-03-02
Transfers: rework heartbeat handling. Closes #5252

### LINKED ISSUE #5252: rework conveyor heartbeat handling
Motivation
----------
Right now heartbeats are updated in the database at each iteration (each activity for multi-activity daemons). This forces us to use a big "older_than" value to avoid the race condition when a big bulk size and slow access to transfertool makes hearbeats expire. 

At the same time, this puts strain on database with un-needed heartbeat updates when the daemons are lightly loaded.

Modification
----------
An idea for improvement is to perform a hearbeat_handler.live() more frequently (each submission, for example), but modify live() to only perform a database update if enough time had passed from last update. The frequency of updates and older_than should probably defined as a function of sleep_time

rcarpa:
Right now heartbeats are updated in the database at each iteration
(each activity for multi-activity daemons). This forces us to use
a big "older_than" value to avoid the race condition when a big bulk
size and slow access to transfertool makes hearbeats expire.
At the same time, this puts strain on database with un-needed heartbeat
updates when the daemons are lightly loaded.

Rework this behavior to perform hearbeat_handler.live() more frequently
(each submission, for example), but modify live() to only perform a
database update if enough time has passed from last update.

### DIFF
--- a/lib/rucio/daemons/conveyor/common.py
+++ b/lib/rucio/daemons/conveyor/common.py
@@ -79,8 +79,16 @@ class HeartbeatHandler:
     Simple contextmanager which sets a heartbeat and associated logger on entry and cleans up the heartbeat on exit.
     """
 
-    def __init__(self, executable, logger_prefix=None):
+    def __init__(self, executable, renewal_interval, logger_prefix=None):
+        """
+        :param executable: the executable name which will be set in heartbeats
+        :param renewal_interval: the interval at which the heartbeat will be renewed in the database.
+        Calls to live() in-between intervals will re-use the locally cached heartbeat.
+        :param logger_prefix: the prefix to be prepended to all log messages
+        """
         self.executable = executable
+        self.renewal_interval = renewal_interval
+        self.older_than = renewal_interval * 10 if renewal_interval and renewal_interval > 0 else None  # 10 was chosen without any particular reason
         self.logger_prefix = logger_prefix or executable
 
         self.hostname = socket.getfqdn()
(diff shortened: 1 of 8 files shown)


lib/rucio/daemons/conveyor/common.py
```diff
@@ -78,8 +78,10 @@ class HeartbeatHandler:
     Simple contextmanager which sets a heartbeat and associated logger on entry and cleans up the heartbeat on exit.
     """
 
-    def __init__(self, executable, logger_prefix=None):
+    def __init__(self, executable, renewal_interval, logger_prefix=None):
```
bari12, line 82: It's very nit-picky, but I would add some documentation for the parameters. Especially what the renewal_interval does.

lib/rucio/daemons/conveyor/common.py
```diff
@@ -100,26 +103,28 @@ def __exit__(self, exc_type, exc_val, exc_tb):
             if self.logger:
                 self.logger(logging.INFO, 'Heartbeat cleaned up')
 
-    def live(self, older_than=None):
-        if older_than:
-            self.last_heart_beat = heartbeat.live(self.executable, self.hostname, self.pid, self.hb_thread, older_than=older_than)
-        else:
-            self.last_heart_beat = heartbeat.live(self.executable, self.hostname, self.pid, self.hb_thread)
+    def live(self):
```
bari12, line 113: I would also document the returns part here.

bari12 (commented):
I wonder if we should move the `HeartbeatHandler` right away to a more common place? Core.heartbeat? It clearly seems like something we should use in other daemons too.

bari12 (approved):

Issue #5303, opened 2022-03-01
Reduce default memcached expiration lifetime


bari12:
Motivation
----------
Rucio uses different cache lifetimes for different data in Rucio, however, the default expiration lifetime is set to 60 minutes for most things, which is quite long for many changes to trickle through the system.

We can probably reduce the default lifetime easily to 10-20m without having much of a noticeable effect on the infrastructure.

Modification
------------
Reduce default cache lifetime to 20m.

PR #5309, opened 2022-03-02
Core & Internals: allow external cache for auth tokens. Closes #5297

### LINKED ISSUE #5297: allow use of memcached for the auth tokens
Motivation
----------

All other caches were moved from in-memory python dicts to memcached some time ago https://github.com/rucio/rucio/pull/5162 . The auth tokens were still kept in memory due to security concerns. Especially for people who may use a separate memcached instance and would be surprised that such information is now leacked. 

Modification
------------
Allow setting auth tokens in memcached, but this should be configurable and default to false.

rcarpa:
Allow setting auth tokens in memcached, but make it configurable and
default this behavior to disabled.

Also reduce the duration of token, for https://github.com/rucio/rucio/issues/5303

### DIFF
--- a/lib/rucio/core/authentication.py
+++ b/lib/rucio/core/authentication.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Copyright 2012-2021 CERN
+# Copyright 2012-2022 CERN
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


mlassnig (dismissed):
LGTM

bari12:
> LGTM

Please approve it then as well 😄

bari12 (approved):

lib/rucio/core/authentication.py
```diff
@@ -59,12 +62,10 @@ def generate_key(token, session=None):
     return generate_key
 
 
-TOKENREGION = make_region(
-    function_key_generator=token_key_generator
-).configure(
-    'dogpile.cache.memory',
-    expiration_time=3600
-)
+if config_get_bool('cache', 'use_external_cache_for_auth_tokens', default=False):
```
bari12, line 65: Just one comment, please add this also to the documentation
http://rucio.cern.ch/documentation/configuration_parameters

PR #5315, opened 2022-03-03
Core & Internals: use third_party_copy_read/write. Closes #3254

### LINKED ISSUE #3254: Switch protocol selection to third_party_copy read/write

rcarpa:
There is a desire to have different configurations for an RSE
when used as source or destination in a transfer. The old
third_party_copy doesn't allow that, so the _read/_write was
introduced to the database at some moment, but were ignored
by rucio code.

Switch to using these setting by default. Add a database migration
to copy the values from the existing field. Also ensure that
setting third_party_copy only when creating a protocol results in
the population of all 3 fields.


bari12:
This needs a rebase due to the merge of the hops PR
@dchristidis please have a look as well.

lib/rucio/db/sqla/migrate_repo/versions/fe1a65b176c9_set_third_party_copy_read_and_write_.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def upgrade():
+    '''
+    Upgrade the database to this revision
+    '''
+    if context.get_context().dialect.name in ['oracle', 'mysql', 'postgresql']:
+        schema = context.get_context().version_table_schema if context.get_context().version_table_schema else ''
+        schema_prefix = schema + '.' if schema else ''
+        execute('UPDATE ' + schema_prefix + 'rse_protocols SET third_party_copy_read=third_party_copy WHERE third_party_copy_read is NULL')
```
bari12, line 36: Do we technically need this? Since the functionality above is backwards compatible? Thus for the people who still just sync in `third_party_copy` it would still be working?

rcarpa: As implemented right now, third_party_copy isn't used anymore anyway. everything relies on _read/_write counterparts. When somebody sets third_party_copy using the cli/api, its value is copied into _read/_write, which are then used. So, yes, in the current form of the pull request, this is necessary. But I'm open to suggestions to make it differently

bari12: I think the way you currently have it is probably better. I am wondering, if  we should just remove the `third_party_copy` column then entirely. People can still sync in a `third_party_copy` value, which will just set both `read/write` or the `read` and `write` value individually.

rcarpa: I propose to do it at next major release. Just to ensure in the meantime that none of corner cases was forgotten.

rcarpa: Having a possibility for easy roll-back seems important to me. Removing the column breaks such possibility

bari12 (approved):

PR #5322, opened 2022-03-04
Client : Add add-exception command in the CLI : Closes #5292

### LINKED ISSUE #5292: Add add-exception command in the CLI
Motivation
----------
A CLI command to add a new exception is missing and need to be added

cserf:
Client : Add add-exception command in the CLI : Closes #5292

### DIFF
--- a/bin/rucio
+++ b/bin/rucio
@@ -20,7 +20,7 @@
 # - Martin Barisits <martin.barisits@cern.ch>, 2012-2022
 # - Thomas Beermann <thomas.beermann@cern.ch>, 2012-2021
 # - Yun-Pin Sun <winter0128@gmail.com>, 2012-2013
-# - Cedric Serfon <cedric.serfon@cern.ch>, 2013-2020
+# - Cedric Serfon <cedric.serfon@cern.ch>, 2013-2022
 # - Ralph Vigne <ralph.vigne@cern.ch>, 2013
 # - David Cameron <david.cameron@cern.ch>, 2014-2021
 # - Tomáš Kouba <tomas.kouba@cern.ch>, 2014
@@ -1460,7 +1460,6 @@ def info_rule(args):
     if args.estimate_ttc:
         logger.error('"--estimate-ttc" is deprecated!')
         return FAILURE
-
     client = get_client(args)
     if args.examine:
         analysis = client.examine_replication_rule(rule_id=args.rule_id)
@@ -1503,7 +1502,6 @@ def info_rule(args):
         print("Notification:               %s" % rule['notification'])
         print("End of life:                %s" % rule['eol_at'])
         print("Child Rule Id:              %s" % rule['child_rule_id'])
-
     return SUCCESS
 
 
@@ -1793,6 +1791,91 @@ def list_datasets_rse(args):
         print('----------')
         for dsn in dsns:
             print(dsn)
+    return SUCCESS
+
+
+@exception_handler
+def add_lifetime_exception(args):
+    """
+    %(prog)s add_lifetime_exception [options] <field1=value1 field2=value2 ...>
+
+    Declare a lifetime model exception.
+
+    """
+    client = get_client(args)
+    if not args.reason:
+        logger.error('reason for the extension is mandatory')
+        return FAILURE
+    reason = args.reason
+    if not args.expiration:
+        logger.error('expiration is mandatory')
+        return FAILURE
+    try:
+        expiration = datetime.strptime(args.expiration, "%Y-%m-%d")
+    except Exception as err:
+        logger.error(err)
+        return FAILURE
+
+    if not args.inputfile:
+        logger.error('inputfile is mandatory')
+        return FAILURE
+    with open(args.inputfile) as infile:
+        dids = list(filter(None, [line.strip() for line in infile]))
+
+    dids_list = []
+    containers = []
+    datasets = []
+    error_types = ['Total DIDs', 'DID not submitted because it is a file', 'DID not submitted because it is a container', 'DID not submitted because it is not part of the lifetime campaign', 'DID successfully submitted']
+    for did in dids:
+        scope, name = get_scope(did, client)
+        dids_list.append({'scope': scope, 'name': name})
+    summary = {0: len(dids_list), 1: 0, 2: 0, 3: 0, 4: 0}
+    for meta in client.get_metadata_bulk(dids_list):
+        scope, name = meta['scope'], meta['name']
+        dids_list.remove({'scope': scope, 'name': name})
+        if meta['did_type'] == 'FILE':
+            logger.warning('%s:%s is a file. Will be ignored' % (scope, name))
+            summary[1] += 1
+        elif meta['did_type'] == 'CONTAINER':
+            logger.warning('%s:%s is a container. It needs to be resolved' % (scope, name))
+            containers.append({'scope': scope, 'name': name})
+            summary[2] += 1
+        elif not meta['eol_at']:
+            logger.warning('%s:%s is not affected by the lifetime model' % (scope, name))
+            summary[3] += 1
+        else:
+            logger.info('%s:%s will be declared' % (scope, name))
+            datasets.append({'scope': scope, 'name': name})
+            summary[4] += 1
+
+    for did in dids_list:
+        scope = did['scope']
+        name = did['name']
+        logger.warning('%s:%s does not exist' % (scope, name))
+
+    if containers:
+        logger.error('One or more DIDs are containers. They must be resolved into a list of datasets to request exception. Full list below')
+        for container in containers:
+            logger.error('%s:%s' % (container['scope'], container['name']))
+        logger.error('Please take action. Submission aborted')
+        return FAILURE
+
+    if not datasets:
+        logger.error('Nothing to submit')
+        return SUCCESS
+    try:
+        client.add_exception(dids=datasets, account=client.account, pattern='', comments=reason, expires_at=expiration)
+    except UnsupportedOperation as err:
+        logger.error(err)
+        return FAILURE
+    except Exception:
+        logger.error('Failure to submit exception. Please retry')
+        logger.debug(traceback.format_exc())
+        return FAILURE
+
+    logger.info('Exception successfully submitted. Summary below')
+    for cnt, error in enumerate(error_types):
+        print('{0:100} {1:6d}'.format(error, len(summary[cnt])))
     return SUCCESS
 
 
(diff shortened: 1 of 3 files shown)


bin/rucio
```diff
@@ -1796,6 +1795,97 @@ def list_datasets_rse(args):
...
+    if not args.reason:
+        logger.error('reason for the extension is mandatory')
+        return FAILURE
+    reason = args.reason
+    if not args.expiration:
+        logger.error('expiration is mandatory')
+        return FAILURE
+    expiration = args.expiration
+    try:
+        expiration = isoparse(expiration)
```
joeldierkes: nit: re-assigned a variable with a different type. Maybe `expiration = isoparse(args.expiration)` and delete line 1814?

bin/rucio
```diff
@@ -1796,6 +1795,97 @@ def list_datasets_rse(args):
...
+    except ValueError as err:
+        logger.error(err)
+        return FAILURE
+
+    if args.inputfile:
+        with open(args.inputfile) as infile:
+            dids = list(filter(None, [line.strip() for line in infile]))
+    else:
+        dids = args.listdids
```
joeldierkes: Where is this `listdids` coming from?

bin/rucio
```diff
@@ -1796,6 +1795,97 @@ def list_datasets_rse(args):
...
+        else:
+            if not meta['eol_at']:
+                logger.warning('%s:%s is not affected by the lifetime model' % (scope, name))
+                dids_list.remove({'scope': scope, 'name': name})
+                summary[3].append({'scope': scope, 'name': name})
+            else:
+                logger.info('%s:%s will be declared' % (scope, name))
+                datasets.append({'scope': scope, 'name': name})
+                dids_list.remove({'scope': scope, 'name': name})
+                summary[4].append({'scope': scope, 'name': name})
```
joeldierkes: This is not readable, confusing and needs to be refactored. Specifically:
- All values in `containers` and `datasets` are duplicated in `summary[2]` and `summary[4]`.
- The `logger.waning`s like `logger.warning('%s:%s is a file. Will be ignored' % (scope, name))` do not provide any additional information, since the summary already displays this information.
- `dids_list.remove({'scope': scope, 'name': name})` happens in every case. This should be before the `if`.
- The `else` branch with the two cases is harder than necessary. You could use an `elif ...: ... else`.

From my POV, it is unnecessary to delete the entries in `did_list`. Just create a new list and append all items which do not exist.

bin/rucio
```diff
@@ -1796,6 +1795,97 @@ def list_datasets_rse(args):
     return SUCCESS
 
 
+@exception_handler
+def add_lifetime_exception(args):
+    """
+    %(prog)s show [options] <field1=value1 field2=value2 ...>
```
joeldierkes: Needs to be edited.

bin/rucio
```diff
@@ -1796,6 +1795,97 @@ def list_datasets_rse(args):
...
+            logger.debug(traceback.format_exc())
+            return FAILURE
+        logger.info('Exception successfully submitted. Summary below')
+        for cnt, error in enumerate(error_types):
+            print('{0:100} {1:6d}'.format(error, len(summary[cnt])))
+        return SUCCESS
+
+    else:
+        logger.error('Nothing to submit')
+        return SUCCESS
```
joeldierkes: This creates an unnecessary indentation for `if datasets:`.
I would suggest an early exit at line 1868:
```pyhton
it not datasets:
  logger.error('Nothing to submit')
  return SUCCESS

# This is the case in which datasets contains items.
...
```

joeldierkes (changes requested):
Maybe a test in `test_bin_rucio.py` is appropriate?

bin/rucio
cserf: Done

bin/rucio
cserf: Removed. I wanted to add the possibility to provide a list of dids directly in the command line, but actually that's not needed.

bin/rucio
cserf: :sweat_smile: Will improve it

bin/rucio
```diff
@@ -1796,6 +1795,91 @@ def list_datasets_rse(args):
     return SUCCESS
 
 
+@exception_handler
+def add_lifetime_exception(args):
+    """
+    %(prog)s add_lifetime_exception [options] <field1=value1 field2=value2 ...>
+
+    Declare a list of bad replicas.
```
dchristidis: Needs to be corrected.

bin/rucio
```diff
@@ -1796,6 +1795,91 @@ def list_datasets_rse(args):
...
+
+    """
+    client = get_client(args)
+    if not args.reason:
```
dchristidis, line 1806: Forgive me, I do not understand the need to check whether the arguments were supplied here. Why would not use `required=True`? For example:
```
add_lifetime_exception_parser.add_argument('--reason', action='store', help='The reason for the extension.', required=True)
```

bin/rucio
```diff
@@ -74,6 +74,7 @@ import traceback
 import unittest
 import uuid
 from datetime import datetime
+from dateutil.parser import isoparse
```
dchristidis: Is the required version available on ALRB?

bin/rucio
```diff
@@ -1796,6 +1795,91 @@ def list_datasets_rse(args):
...
+    client = get_client(args)
+    if not args.reason:
+        logger.error('reason for the extension is mandatory')
+        return FAILURE
+    reason = args.reason
+    if not args.expiration:
+        logger.error('expiration is mandatory')
+        return FAILURE
+    try:
+        expiration = isoparse(args.expiration)
```
dchristidis: I would have `argparse` check for this automatically (see below).

bin/rucio
```diff
@@ -1796,6 +1795,91 @@ def list_datasets_rse(args):
...
+        dids = list(filter(None, [line.strip() for line in infile]))
+
+    dids_list = []
+    containers = []
+    datasets = []
+    error_types = ['Total DIDs', 'DID not submitted because it is a file', 'DID not submitted because it is a container', 'DID not submitted because it is not part of the lifetime campaign', 'DID successfully submitted']
+    for did in dids:
+        scope, name = get_scope(did, client)
+        dids_list.append({'scope': scope, 'name': name})
+    summary = {0: len(dids_list), 1: 0, 2: 0, 3: 0, 4: 0}
```
dchristidis, line 1832: A `collections.namedtuple` feels more appropriate here.

bin/rucio
```diff
@@ -1796,6 +1795,91 @@ def list_datasets_rse(args):
...
+    for meta in client.get_metadata_bulk(dids_list):
+        scope, name = meta['scope'], meta['name']
+        dids_list.remove({'scope': scope, 'name': name})
+        if meta['did_type'] == 'FILE':
+            logger.warning('%s:%s is a file. Will be ignored' % (scope, name))
```
dchristidis, line 1837: Defer interpolation for logging messages, see: https://docs.python.org/3/howto/logging.html#optimization
So, this should have been:
```
logger.warning('%s:%s is a file. Will be ignored', scope, name)
````

File: bin/rucio [hunk 12]
```diff
@@ -2480,6 +2564,21 @@ can be found in ' + Color.BOLD + 'http://rucio.cern.ch/documentation/rse_express
     touch_parser.add_argument(dest='dids', nargs='+', action='store', help='List of space separated data identifiers.')
     touch_parser.add_argument('--rse', dest='rse', action='store', help="The RSE of the DIDs that are touched.").completer = rse_completer
 
+    # The add-lifetime-exception command
+    add_lifetime_exception_parser = subparsers.add_parser('add-lifetime-exception', help='Add an exception to the lifetime model.',
+                                                          formatter_class=argparse.RawDescriptionHelpFormatter, epilog='''Usage example
+"""""""""""""
+::
+
+    $ rucio add-lifetime-exception --inputfile myfile.txt --reason "Needed for my analysis" --expiration 2015-10-30
```
dchristidis, line 2572: I’m afraid that ‘needed for my analysis’ will be used literally. :smile:
I think we need a better example in the documentation which encourages the user to better express the need for the exception.

bin/rucio
```diff
@@ -2480,6 +2564,21 @@ can be found in ' + Color.BOLD + 'http://rucio.cern.ch/documentation/rse_express
...
+
+    ''')
+
+    add_lifetime_exception_parser.set_defaults(function=add_lifetime_exception)
+    add_lifetime_exception_parser.add_argument('--inputfile', action='store', help='File where the list of datasets requested to be extended are located.')
+    add_lifetime_exception_parser.add_argument('--reason', action='store', help='The reason for the extension.')
+    add_lifetime_exception_parser.add_argument('--expiration', action='store', help='The expiration date format YYYY-MM-DD')
```
dchristidis: Use
```
add_lifetime_exception_parser.add_argument('--expiration', action='store', type=isoparse, help='The expiration date format YYYY-MM-DD')
```
to automatically check and convert the command-line argument.

dchristidis (commented):
Unfortunately, I did not have the time to check the implementation. If my comments can be addressed in time, it would be good. Otherwise, because *we need this* for the next application of the Lifetime Model, I would prefer to merge as is and improve later.

bin/rucio
dchristidis: :point_up_2: This is the only point that *should* be addressed.

bin/rucio
```diff
@@ -1796,6 +1795,91 @@ def list_datasets_rse(args):
...
+            summary[3] += 1
+        else:
+            logger.info('%s:%s will be declared' % (scope, name))
+            datasets.append({'scope': scope, 'name': name})
+            summary[4] += 1
+
+    for did in dids_list:
+        scope = did['scope']
+        name = did['name']
+        logger.warning('%s:%s does not exist' % (scope, name))
```
joeldierkes, line 1854: I also just noticed, that this code is never reached.

bin/rucio
```diff
@@ -1796,6 +1795,91 @@ def list_datasets_rse(args):
...
+
+    """
+    client = get_client(args)
+    if not args.reason:
+        logger.error('reason for the extension is mandatory')
+        return FAILURE
+    reason = args.reason
+    if not args.expiration:
+        logger.error('expiration is mandatory')
+        return FAILURE
```
joeldierkes, line 1812: Same here

bin/rucio
joeldierkes, line 1832: The different lists also do not seem clean. You have redundant information: `summary[2] == len(containers)', `summary[4] == len(datasets)`. The `namedtuple` seems like a nice solution, maybe add the lists there and use `len` instead of a counter.

bin/rucio
```diff
@@ -1796,6 +1795,91 @@ def list_datasets_rse(args):
...
+    containers = []
+    datasets = []
+    error_types = ['Total DIDs', 'DID not submitted because it is a file', 'DID not submitted because it is a container', 'DID not submitted because it is not part of the lifetime campaign', 'DID successfully submitted']
+    for did in dids:
+        scope, name = get_scope(did, client)
+        dids_list.append({'scope': scope, 'name': name})
+    summary = {0: len(dids_list), 1: 0, 2: 0, 3: 0, 4: 0}
+    for meta in client.get_metadata_bulk(dids_list):
+        scope, name = meta['scope'], meta['name']
+        dids_list.remove({'scope': scope, 'name': name})
```
joeldierkes, line 1835: I haven't noticed before, but since this is run every time, the list will be empty in the end. You could just remove this line.

joeldierkes (changes requested):

bin/rucio
joeldierkes: Okay, this comment could be wrong. I'm not sure if `client.get_metadata_bulk` always returns a list with the same length as the input list. If not, it should be safe to ignore.

bin/rucio
```diff
@@ -2480,6 +2563,21 @@ can be found in ' + Color.BOLD + 'http://rucio.cern.ch/documentation/rse_express
...
+"""""""""""""
+::
+
+    $ rucio add-lifetime-exception --inputfile myfile.txt --reason "Needed for my analysis" --expiration 2015-10-30
+
+    ''')
+
+    add_lifetime_exception_parser.set_defaults(function=add_lifetime_exception)
+    add_lifetime_exception_parser.add_argument('--inputfile', action='store', help='File where the list of datasets requested to be extended are located.', required=True)
+    add_lifetime_exception_parser.add_argument('--reason', action='store', help='The reason for the extension.'), required=True
```
bari12: Syntax error ⚠️

bari12 (approved):
Will be included in 1.27.9 as discussed, some cosmetic changes still needed afterwards

PR #5324, opened 2022-03-04
Update docker versions; Fix #4198

### LINKED ISSUE #4198: Improve docker image performance

mlassnig:

### DIFF
--- a/etc/docker/test/matrix.yml
+++ b/etc/docker/test/matrix.yml
@@ -40,9 +40,9 @@ suites:
       - oracle
       - mysql8
       - postgres9
-      - postgres12
+      - postgres14
       - sqlite
   - id: multi_vo
-    RDBMS: postgres12
+    RDBMS: postgres14
 image_identifier:
   - autotest
--- a/etc/docker/dev/docker-compose-storage-alldb.yml
+++ b/etc/docker/dev/docker-compose-storage-alldb.yml
@@ -1,6 +1,6 @@
 services:
   rucio:
-    image: docker.io/rucio/rucio-dev:py3
+    image: docker.io/rucio/rucio-dev
     extra_hosts:
       - "ruciomy8:127.0.0.1"
       - "oracle:127.0.0.1"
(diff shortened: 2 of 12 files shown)


maany:
LGTM! It is important to merge this quickly as it contains changes required in `rucio/rucio` for fixing `test-conveyor` and `test-tpc` since the `fts-rest-client` was updated in https://github.com/rucio/containers/pull/182

maany (commented):
LGTM, just the header in  `lib/rucio/tests/test_tpc.py` needs to be fixed. ;)

.github/workflows/integration_tests.yml
```diff
@@ -41,39 +41,6 @@ jobs:
...
-
-          cd $GITHUB_WORKSPACE
-
-          # Check if rucio/containers has a corresponding $GIT_REF 
-          if [ $(git tag -l "$GIT_REF") ]; then
-              git checkout tags/$GIT_REF
-          else
-              echo "Tag $GIT_REF not found in rucio/containers. Integration test containers will be built against the master branch instead."
-              git checkout master
-          fi
```
joeldierkes: Why is this deleted?
As far as I know, this was included (#5256) since changes made to the rucio/containers repo were only used when a new container was released. However, we want to use these changes in the tests immediately.

maany: It needs to be patched. I will send a separate PR later today.  ;)

etc/docker/test/matrix.yml
```diff
@@ -40,9 +40,9 @@ suites:
       - oracle
       - mysql8
       - postgres9
-      - postgres12
+      - postgres14
```
bari12, line 43: I think this is good, I am wondering though about the `postgres9` one. In principle this is not supported anymore since end of last year. I wonder if we should just test against postgres14 and remove postgres9 entirely.
Raises a bit also the question of our compatibilities we mention in http://rucio.cern.ch/documentation/requirements

mlassnig: We only have the old one because psql9 is the only available one in CentOS7. psql10 is default in CentOS8, however psql12 is also available. I think it's fine to just have psql14.

bari12: As discussed, let's remove everything but postgres14

bari12:
Rebase please

bari12 (approved):

PR #5325, opened 2022-03-04
redirect_uri now takes into account paths #5220

### LINKED ISSUE #5220: oidc_redirect URL is incorrect for routes with a path
The URL sent to the user for authentication with the IdP:

https://github.com/rucio/rucio/blob/d18fb0183a1456957994a94fb981d1386e11b386/lib/rucio/core/oidc.py#L310-L311

assumes that the server sits at the root, i.e. `oidc_redirect` is found at `https://<netloc>/auth/oidc_redirect`. This is not correct if there is any path after the network location. 

In our case, the server, and consequently `oidc_redirect`, is routed via an nginx proxy to `https://<netloc>/rucio/auth/oidc_redirect`, so the above doesn't work.

I have hacked this in our own deployment to:

```python
auth_url = build_url('https://{}{}'.format(auth_server.netloc, auth_server.path.split('/auth/')[0]),
    path='auth/oidc_redirect', params=access_msg)
```

which works (for us), but I haven't thought about possible edge cases or whether there is a more sensible way of constructing this URL.

robbarnsley:
Mangled the branch, easier to just recreate the request.

### DIFF
--- a/lib/rucio/core/oidc.py
+++ b/lib/rucio/core/oidc.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Copyright 2019-2021 CERN
+# Copyright 2019-2022 CERN
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


lib/rucio/core/oidc.py
```diff
@@ -18,6 +18,29 @@
 # - Patrick Austin <patrick.austin@stfc.ac.uk>, 2020
 # - Benedikt Ziemons <benedikt.ziemons@cern.ch>, 2021
 # - Radu Carpa <radu.carpa@cern.ch>, 2021
+# - Rob Barnsley <rob.barnsley@skao.int>, 2022
+
+# -*- coding: utf-8 -*-
```
bari12: Header appears twice 😄

bari12 (changes requested):
Otherwise I think this is GTG

lib/rucio/core/oidc.py
robbarnsley: no idea how that happened! i've had that before with the add_header script

maany: I think the option `-i` is required to modify the header in place instead of adding a new one on the top of the file. I've been bitten by this once or twice :P

bari12 (approved):

Issue #4992, opened 2021-11-10
Fix broken Rest API docs view


joeldierkes:
Motivation
----------
The client and rest API documentation improperly. The Markdown syntax seems not to be supported properly. This makes it harder for the user to find necessary information.

We want to use the [`OpenApi`](https://www.openapis.org/) specification, since it's widely used and common understandable.

Modification
------------
Update the view of the client and rest API to display the information correctly. For this, use [`apispec`](https://github.com/marshmallow-code/apispec). `apispec` automatically generates the `OpenApi` specification from the REST API source code.

Components to update (Branch created, PR created):
- [x] [x] accountlimits
- [x] [x] accounts
- [x] [x] archives
- [x] [ ] auth
- [x] [x] config
- [x] [x] credentials
- [x] [x] dids
- [x] [x] dirac
- [x] [x] export
- [x] [x] heartbeats
- [x] [x] identities
- [x] [x] import
- [x] [x] lifetime_exceptions
- [x] [x] locks
- [x] [x] meta
- [x] [x] nongrid_traces
- [x] [x] ping
- [x] [x] redirect
- [x] [x] replicas
- [x] [x] requests
- [x] [x] rses
- [x] [x] rules
- [x] [x] scopes
- [x] [x] subscriptions
- [x] [ ] tmp_dids
- [x] [ ] traces
- [x] [ ] vos

PR #5333, opened 2022-03-08
REST & API: Change the `nongrid traces` api docs to the OpenApi forma…

joeldierkes:
…t #4992

The Api documentation is obsolete and not displayed correctly. For this sake we
introduce the OpenApi format for our docs. It is way stricter and a commonly
used format.


bari12:
Note to myself and @maany 
Eventually, once we have a better schema for the traces, it should be added here too.

bari12 (approved):

PR #5336, opened 2022-03-08
Dirac : Provide the possibility to set a lifetime on the default rule…

### LINKED ISSUE #5164: Dirac : Provide the possibility to set a lifetime on the default rule
Motivation
----------
When `add_files` is called, a rule is created on the dataset. This rule has infinite lifetime. We would like to have the possibility to set a finite lifetime based on the scope. Moreover, it should be possible to reset the lifetime each time the method is called

cserf:
Dirac : Provide the possibility to set a lifetime on the default rule : Closes #5164

### DIFF
--- a/lib/rucio/core/dirac.py
+++ b/lib/rucio/core/dirac.py
@@ -14,25 +14,30 @@
 # limitations under the License.
 #
 # Authors:
-# - Cedric Serfon <cedric.serfon@cern.ch>, 2020
+# - Cedric Serfon <cedric.serfon@cern.ch>, 2020-2022
 # - Eli Chadwick <eli.chadwick@stfc.ac.uk>, 2020
 # - Martin Barisits <martin.barisits@cern.ch>, 2020-2021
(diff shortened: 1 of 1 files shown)


lib/rucio/core/dirac.py
```diff
@@ -67,12 +72,23 @@ def add_files(lfns, account, ignore_availability, vo='def', session=None):
...
     """
+    rule_extension_list = []
     attachments = []
     # The list of scopes is necessary for the extract_scope
     filter_ = {'scope': InternalScope(scope='*', vo=vo)}
     scopes = list_scopes(filter_=filter_, session=session)
     scopes = [scope.external for scope in scopes]
     exist_lfn = []
+    try:
+        lifetime_dict = config_get(section='dirac', option='lifetime', session=session)
```
bari12, line 83: Please add this also to http://rucio.cern.ch/documentation/configuration_parameters
So it's documented there as well.

bari12 (approved):

PR #5337, opened 2022-03-08
Always print the integration test logs #5253

### LINKED ISSUE #5253: Print the integration test server logs in the actions
Motivation
----------
In #5249 the server logs were needed to track down an issue. We should print them by default to have more insight in case of a bug.

Modification
------------
Print the server logs when the integration tests are running in the actions. They should be printed after the test's run.

joeldierkes:
If any job before the `Print server error logs` fails, the job itself will not
run. Since these are the cases in which the logs can provide additional
information, they should be printed.

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -145,6 +145,7 @@ jobs:
             exit 1
           fi
       - name: Print server error logs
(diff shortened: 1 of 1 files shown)


maany (approved):
good catch! ;) LGTM

PR #5340, opened 2022-03-09
Change mechanism to update bad replica states to synchronous mode : Closes #5268

### LINKED ISSUE #5268: Necromancer update-history procedure is wasteful when there’s a backlog
Motivation
----------

With a period of one hour (hard-coded), Necromancer will run a procedure to transition the bad replicas from state `BAD` to either `DELETED`, `LOST` or `RECOVERED`. As far as I can tell, this does not affect the replica or DID availability and, as such, should be non-critical.

If there’s a significant backlog and the Necromancer-Evaluator protection is enabled, it could be that most or a very significant amount of bad replicas have yet to be processed. In that case, running the update-history procedure is wasteful.

Modification
------------

Consider delaying running the procedure when the back-off check on the Evaluator backlog is triggered.

cserf:
Change mechanism to update bad replica states to synchronous mode : Closes #5268

Previously, the update of the bad replicas after recovery was done in an asynchronous way in the necromancer

This patch moves the update in the finisher, undertaker and reaper to make it synchronous

### DIFF
--- a/lib/rucio/core/replica.py
+++ b/lib/rucio/core/replica.py
@@ -268,7 +268,7 @@ def list_bad_replicas_status(state=BadFilesStatus.BAD, rse_id=None, younger_than
 
 
 @read_session
-def list_bad_replicas_history(limit=10000, thread=None, total_threads=None, session=None):
+def list_bad_replicas_history(limit=10000, thread=None, total_threads=None, session=None):   # Deprecated method. To be removed
     """
     List the bad file replicas history. Method only used by necromancer
 
@@ -292,7 +292,7 @@ def list_bad_replicas_history(limit=10000, thread=None, total_threads=None, sess
 
 
 @transactional_session
-def update_bad_replicas_history(dids, rse_id, session=None):
+def update_bad_replicas_history(dids, rse_id, session=None):   # Deprecated method. To be removed
     """
     Update the bad file replicas history. Method only used by necromancer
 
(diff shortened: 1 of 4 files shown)


lib/rucio/core/replica.py
```diff
@@ -288,7 +288,7 @@ def list_bad_replicas_history(limit=10000, thread=None, total_threads=None, sess
 
 
 @transactional_session
-def update_bad_replicas_history(dids, rse_id, session=None):
+def update_bad_replicas_history(dids, rse_id, session=None):   # Deprecated method. To be removed
```
bari12: Can't these be removed right away?

bari12: Can't these be removed right away?

lib/rucio/core/replica.py
```diff
@@ -264,7 +264,7 @@ def list_bad_replicas_status(state=BadFilesStatus.BAD, rse_id=None, younger_than
 
 
 @read_session
-def list_bad_replicas_history(limit=10000, thread=None, total_threads=None, session=None):
+def list_bad_replicas_history(limit=10000, thread=None, total_threads=None, session=None):   # Deprecated method. To be removed
```
bari12: Can't these be removed right away?

bari12: Can't these be removed right away?

bari12 (commented):
Two comments, seems ok to me.
@rcarpa please have a look as well, this might need to be added to the temp table path as well?

rcarpa:
> Two comments, seems ok to me. @rcarpa please have a look as well, this might need to be added to the temp table path as well?

Yes, the query must also be added to the temporary tables code path. Shouldn't be difficult though. 

Also, +1 for removing the function which are not used anymore. The history remains in git, you can always find the old one if needed later.

lib/rucio/core/replica.py
cserf: Done

cserf: Done

lib/rucio/core/replica.py
cserf: Done

bari12 (approved):

PR #5350, opened 2022-03-09
Add nightly build, remove py37,38 

### LINKED ISSUE #5349: Remove py37 and py38 test workflows from autotests
Motivation
----------
The load on CI runners should be reduced to keep the run time for autotests, integration tests, votests in check.

Modification
------------
1. Remove all tests for py37 and py38 other than postgres RDBMS
2. Add a nightly trigger that runs all the build matrix permutations

maany:
Close #5349
p.s. Sending this PR now, so we can check if the scheduled workflow ran at 3 a.m. tonight ;)

### DIFF
--- a/.github/workflows/autotest.yml
+++ b/.github/workflows/autotest.yml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   header-lint:
-    if: ${{ 'github.event_name == ''pull_request''' || 'github.event_name == ''schedule''' }}
+    if: 'github.event_name == ''pull_request'''
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/autotest.yml
+++ b/.github/workflows/autotest.yml
@@ -1,12 +1,14 @@
 name: Run all the autotests
 
 on:
-  - pull_request
-  - push
+  pull_request:
+  push:
+  schedule:
+    - cron: '0 3 * * *'
 
 jobs:
   header-lint:
-    if: 'github.event_name == ''pull_request'''
+    if: ${{ 'github.event_name == ''pull_request''' || 'github.event_name == ''schedule''' }}
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


.github/workflows/autotest.yml
```diff
@@ -1,12 +1,14 @@
...
-  - push
+  pull_request:
+  push:
+  schedule:
+    - cron: '0 3 * * *'
 
 jobs:
   header-lint:
-    if: 'github.event_name == ''pull_request'''
+    if: ${{ 'github.event_name == ''pull_request''' || 'github.event_name == ''schedule''' }}
```
joeldierkes: We create a diff between the current branch and the master branch to get the changed files. In the case of a scheduled run this would not be possible, since the master branch is the current branch.

I would only run this job on every pull request.

etc/docker/test/matrix.yml
```diff
@@ -20,8 +20,6 @@ python:
         - client
         - client_syntax
   - "3.6"
-  - "3.7"
-  - "3.8"
   - "3.9"
```
joeldierkes, line 25: What about the 3.10?

etc/docker/test/matrix_nightly.yml
```diff
@@ -0,0 +1,48 @@
...
+    allow:
+      python: "2.7"
+  - id: client
+    RDBMS: sqlite
+  - id: all
+    RDBMS:
+      - oracle
+      - mysql8
+      - postgres9
+      - postgres12
```
joeldierkes: Do we want to support `postgres14` in the nightly builds? @bari12

.github/workflows/autotest.yml
```diff
@@ -1,12 +1,14 @@
 name: Run all the autotests
 
 on:
-  - pull_request
-  - push
+  pull_request:
+  push:
+  schedule:
+    - cron: '0 3 * * *'
```
joeldierkes, line 7: Monday till Friday should be enough (or shifted by one if you want to rather run the tests on Saturday morning then Monday morning).
`0 3 * * 1-5`

joeldierkes (changes requested):
Nice work @maany ! 😁

etc/docker/test/matrix.yml
maany: 3.10 will need more work ;) New dockerfile's etc. That would be in another PR

.github/workflows/autotest.yml
maany: IMO, nightly/daily builds should happen every day regardless of weekends.

etc/docker/test/matrix.yml
maany: Aah btw, I just realized I need to make some more changes in the matrix_parser.py in order to enable pu37,py38 builds just for Postgres as db.

etc/docker/test/matrix_nightly.yml
mlassnig: my patch sets the dev default to psql14, so should be added here as well

bari12: As discussed in marios PR https://github.com/rucio/rucio/pull/5324#discussion_r820775921
Let's just do postgres14, remove 9 and 12.

.github/workflows/autotest.yml
bari12: I would also do it every day (for now). We are doing already a big reduction of tests with removing a large part of the jobs for the PR tests. So even for environmental considerations, this is a net positive step. If we see that it is really not needed, we can still remove.

bari12 (approved):

PR #5352, opened 2022-03-11
Consistency checks: Drop unmaintened bz2file module, use plain bz2 in…

### LINKED ISSUE #5348: Remove/change `bz2file` dependency
Motivation
----------
`bz2file` has not been updated since 2014. Switch to a different dependency if needed, or remove entirely from the codebase.

faluchet:
…stead - Fix #5348

### DIFF
--- a/setuputil.py
+++ b/setuputil.py
@@ -17,6 +17,7 @@
 # - Benedikt Ziemons <benedikt.ziemons@cern.ch>, 2021
 # - Martin Barisits <martin.barisits@cern.ch>, 2021-2022
 # - Joel Dierkes <joel.dierkes@cern.ch>, 2022
+# - Fabio Luchetti <fabio.luchetti@cern.ch>, 2022
 
 from __future__ import print_function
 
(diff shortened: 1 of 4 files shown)


setuputil.py
```diff
@@ -51,7 +52,7 @@
     'sftp': ['pysftp'],
     # 'gfal2': ['gfal2-python'],
     'dumper': [
-        'bz2file',
```
bari12, line 54: You can remove `bz2` here, since it is a python standard library.

setuputil.py
```diff
@@ -81,7 +82,7 @@
     'install_requires': clients_requirements_table['install_requires'] + [
         'argcomplete',
         'boto',
-        'bz2file',
```
bari12, line 84: You can remove `bz2` here, since it is a python standard library.

bari12 (changes requested):
Looks good 👍 
One small change request, you can remove `bz2` from the setuputil (You already did from the requirements file anyway) - It's not needed (Since it is a default lib)

bari12 (approved):

PR #5357, opened 2022-03-11
Client : Bug in the upload client for cloud resources : Closes #5354

### LINKED ISSUE #5354: Bug in the upload client for cloud resources
Motivation
----------
Upload fails because of use of wrong signed URL to check that the replica exist

cserf:
Client : Bug in the upload client for cloud resources : Closes #5354

### DIFF
--- a/lib/rucio/client/uploadclient.py
+++ b/lib/rucio/client/uploadclient.py
@@ -608,6 +608,7 @@ class UploadClient:
 
         # Getting pfn
         pfn = None
+        readpfn = None
         try:
             pfn = list(protocol_write.lfns2pfns(make_valid_did(lfn)).values())[0]
             logger(logging.DEBUG, 'The PFN created from the LFN: {}'.format(pfn))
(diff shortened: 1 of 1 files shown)


lib/rucio/client/uploadclient.py
```diff
@@ -621,23 +622,24 @@ def _upload_item(self, rse_settings, rse_attributes, lfn, source_dir=None, domai
         # Auth. mostly for object stores
         if sign_service:
             pfn = self.client.get_signed_url(rse_settings['rse'], sign_service, 'write', pfn)       # NOQA pylint: disable=undefined-variable
+            protocol_read = self._create_protocol(rse_settings, 'read', domain=domain, impl=impl)
+            readpfn = self.client.get_signed_url(rse_settings['rse'], sign_service, 'read', pfn)    # NOQA pylint: disable=undefined-variable
```
bari12: I think it would be better to rename this to `signed_read_pfn` and the tmp below to `signed_read_pfn_tmp`. Just to make it absolutely clear in the code, that this is not a normal readpfn which might be needed otherwise, but it is really just for the signed urls.

bari12, line 626: I think it would be better to rename this to `signed_read_pfn` and the tmp below to `signed_read_pfn_tmp`. Just to make it absolutely clear in the code, that this is not a normal readpfn which might be needed otherwise, but it is really just for the signed urls.

mlassnig (approved):

PR #5359, opened 2022-03-12
Rules: Makes the update-rule locked field insensitve #5356

### LINKED ISSUE #5356: rucio update-rule --locked is case sensitive
Motivation
----------

It should accept not only `False`, but also `false`

KarthikSundar2002:
Changes the if-else condition of update_rule checking to accept all cases

### DIFF
--- a/bin/rucio
+++ b/bin/rucio
@@ -56,6 +56,7 @@
 # - Igor Mandrichenko <ivm@fnal.gov>, 2021
 # - Rob Barnsley <robbarnsley@users.noreply.github.com>, 2021
 # - Carlosbg <bogomcar@gmail.com>, 2021
+# - KarthikSundar2002 <karthiksundar30092002@gmail.com>, 2022
 
 from __future__ import print_function
 
@@ -1403,10 +1404,13 @@ def update_rule(args):
     if args.lifetime:
         options['lifetime'] = None if args.lifetime.lower() == "none" else int(args.lifetime)
     if args.locked:
-        if args.locked == "True":
+        if args.locked.title() == "True":
             options['locked'] = True
-        elif args.locked == "False":
+        elif args.locked.title() == "False":
             options['locked'] = False
+        else:
+            logger.error('Locked must be True or False')
+            return FAILURE
     if args.comment:
         options['comment'] = args.comment
     if args.rule_account:


joeldierkes:
We automatically generate the headers fot the python files. Since you made a commit, you have to be listed in the list of contributors as well.
You can run `tools/add-header -i bin/rucio` to fix the header automatically, then amend the new commit you created. This will fix the failing test.

KarthikSundar2002:
Yea , amended my last commit to pass the header test

joeldierkes:
I think you forgot to change to commit :sweat_smile:

bin/rucio
```diff
@@ -1403,9 +1403,9 @@ def update_rule(args):
     if args.lifetime:
         options['lifetime'] = None if args.lifetime.lower() == "none" else int(args.lifetime)
     if args.locked:
-        if args.locked == "True":
+        if args.locked.title() == "True":
             options['locked'] = True
-        elif args.locked == "False":
+        elif args.locked.title() == "False":
```
joeldierkes, line 1409: I also noticed a possible improvement in the code, we should abort on an invalid input instead of silently continuing:
```python
if args.locked:
  if args.locked.lower() not in ("true", "false"):
    logger.error('--locked should be either "true" or "false"')
    return FAILURE
  options['locked'] = args.locked.lower() == "true"
```

joeldierkes (changes requested):

KarthikSundar2002:
I have added the necessary changes and amended my last commit to do so, I hope it works now :)

bin/rucio
```diff
@@ -1403,10 +1404,14 @@ def update_rule(args):
...
             options['locked'] = False
+        else:
+            logger.error('Locked must be True or False')
+            return FAILURE
+        options['locked'] = True
```
joeldierkes: It will work if you delete this line. This line would override everything with `True`.

joeldierkes (changes requested):

bin/rucio
KarthikSundar2002: Yea, my bad :( ...Removed it now

joeldierkes (approved):
lgtm :+1:

Issue #5300, opened 2022-03-01
Changing log messages in suspicious_replica_recoverer, fixing sleep_time issue and adding passive mode


ChristophAmes:
- Changing the log messages to make it easier to search for specific actions

- Fixing an issue where the sleep_time parameter isn't being propagated, as mentioned here: https://github.com/rucio/rucio/issues/5285

PR #5365, opened 2022-03-15
Recovery: Change log messages, fix sleep_time propagation, add passive_mode #5300

ChristophAmes:

### DIFF
--- a/bin/rucio-replica-recoverer
+++ b/bin/rucio-replica-recoverer
@@ -235,7 +235,7 @@ Note that attempting the use the ``--vos`` argument when in single-VO mode will 
     parser.add_argument("--run-once", action="store_true", default=False, help='One iteration only.')
     parser.add_argument("--limit-suspicious-files-on-rse", action="store", default=5, help='Maximum number of suspicious replicas on an RSE before that RSE is considered problematic and the suspicious replicas on that RSE are declared "TEMPORARY_UNAVAILABLE". Default value is 5.')
     parser.add_argument('--sleep-time', action="store", default=3600, type=int, help='Concurrency control: thread sleep time after each chunk of work')
-    parser.add_argument('--passive-mode', action="store", default=False, help='Mainly for testing purposes. If True, the daemon will be run without taking any actions on any files. The log file will be produced normally. By default False.')
+    parser.add_argument('--active-mode', action="store_true", default=False, help='If NOT specified, the daemon will run without taking any actions on any files. In either case, the log file will be produced normally.')
     return parser
 
 
@@ -244,6 +244,6 @@ if __name__ == "__main__":
     PARSER = get_parser()
     ARGS = PARSER.parse_args()
     try:
-        run(once=ARGS.run_once, younger_than=ARGS.younger_than, nattempts=ARGS.nattempts, vos=ARGS.vos, limit_suspicious_files_on_rse=ARGS.limit_suspicious_files_on_rse, sleep_time=ARGS.sleep_time, passive_mode=ARGS.passive_mode)
+        run(once=ARGS.run_once, younger_than=ARGS.younger_than, nattempts=ARGS.nattempts, vos=ARGS.vos, limit_suspicious_files_on_rse=ARGS.limit_suspicious_files_on_rse, sleep_time=ARGS.sleep_time, active_mode=ARGS.active_mode)
     except KeyboardInterrupt:
         stop()
--- a/bin/rucio-replica-recoverer
+++ b/bin/rucio-replica-recoverer
@@ -234,7 +234,8 @@ Note that attempting the use the ``--vos`` argument when in single-VO mode will 
     parser.add_argument('--vos', nargs='+', type=str, help='Optional list of VOs to consider. Only used in multi-VO mode.')
     parser.add_argument("--run-once", action="store_true", default=False, help='One iteration only.')
     parser.add_argument("--limit-suspicious-files-on-rse", action="store", default=5, help='Maximum number of suspicious replicas on an RSE before that RSE is considered problematic and the suspicious replicas on that RSE are declared "TEMPORARY_UNAVAILABLE". Default value is 5.')
-    parser.add_argument('--sleep-time', action="store", default=300, type=int, help='Concurrency control: thread sleep time after each chunk of work')
+    parser.add_argument('--sleep-time', action="store", default=3600, type=int, help='Concurrency control: thread sleep time after each chunk of work')
+    parser.add_argument('--passive-mode', action="store", default=False, help='Mainly for testing purposes. If True, the daemon will be run without taking any actions on any files. The log file will be produced normally. By default False.')
     return parser
 
 
@@ -243,6 +244,6 @@ if __name__ == "__main__":
     PARSER = get_parser()
     ARGS = PARSER.parse_args()
     try:
-        run(once=ARGS.run_once, younger_than=ARGS.younger_than, nattempts=ARGS.nattempts, vos=ARGS.vos, limit_suspicious_files_on_rse=ARGS.limit_suspicious_files_on_rse)
+        run(once=ARGS.run_once, younger_than=ARGS.younger_than, nattempts=ARGS.nattempts, vos=ARGS.vos, limit_suspicious_files_on_rse=ARGS.limit_suspicious_files_on_rse, sleep_time=ARGS.sleep_time, passive_mode=ARGS.passive_mode)
     except KeyboardInterrupt:
         stop()
--- a/lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
+++ b/lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
@@ -55,7 +55,7 @@ from rucio.db.sqla.util import get_db_time
 GRACEFUL_STOP = threading.Event()
 
 
-def declare_suspicious_replicas_bad(once=False, younger_than=3, nattempts=10, vos=None, limit_suspicious_files_on_rse=5, sleep_time=3600, passive_mode=True):
+def declare_suspicious_replicas_bad(once=False, younger_than=3, nattempts=10, vos=None, limit_suspicious_files_on_rse=5, sleep_time=3600, active_mode=False):
     """
     Main loop to check for available replicas which are labeled as suspicious.
 
@@ -247,13 +247,13 @@ def declare_suspicious_replicas_bad(once=False, younger_than=3, nattempts=10, vo
                         for replica_value in recoverable_replicas[vo][rse_key].values():
                             surls_list.append(replica_value['surl'])
 
-                        if not passive_mode:
+                        if active_mode:
                             add_bad_pfns(pfns=surls_list, account=InternalAccount('root', vo=vo), state='TEMPORARY_UNAVAILABLE', expires_at=datetime.utcnow() + timedelta(days=3))
 
                         logger(logging.INFO, "%s is problematic (more than %s suspicious replicas). Send a Jira ticket for the RSE (to be implemented).", rse_key, limit_suspicious_files_on_rse)
                         logger(logging.INFO, "The following files on %s have been marked as TEMPORARILY UNAVAILABLE:", rse_key)
                         for replica_values in recoverable_replicas[vo][rse_key].values():
-                            logger(logging.INFO, 'Temporarily unavailable: RSE: %s    Scope: %s    Name: %s    Pfn: %s', rse_key, replica_values['scope'], replica_values['name'], replica_values['surl'])
+                            logger(logging.INFO, 'Temporarily unavailable: RSE: %s    Scope: %s    Name: %s    PFN: %s', rse_key, replica_values['scope'], replica_values['name'], replica_values['surl'])
                         # Remove the RSE from the dictionary as it has been dealt with.
                         del recoverable_replicas[vo][rse_key]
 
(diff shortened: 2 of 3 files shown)


bin/rucio-replica-recoverer
```diff
@@ -244,7 +242,8 @@ Note that attempting the use the ``--vos`` argument when in single-VO mode will
     parser.add_argument('--vos', nargs='+', type=str, help='Optional list of VOs to consider. Only used in multi-VO mode.')
     parser.add_argument("--run-once", action="store_true", default=False, help='One iteration only.')
     parser.add_argument("--limit-suspicious-files-on-rse", action="store", default=5, help='Maximum number of suspicious replicas on an RSE before that RSE is considered problematic and the suspicious replicas on that RSE are declared "TEMPORARY_UNAVAILABLE". Default value is 5.')
-    parser.add_argument('--sleep-time', action="store", default=300, type=int, help='Concurrency control: thread sleep time after each chunk of work')
+    parser.add_argument('--sleep-time', action="store", default=3600, type=int, help='Concurrency control: thread sleep time after each chunk of work')
```
joeldierkes, line 237: I don't know the requirements. Is 1h maybe too much?

bin/rucio-replica-recoverer
```diff
@@ -244,7 +242,8 @@ Note that attempting the use the ``--vos`` argument when in single-VO mode will
...
+    parser.add_argument('--passive-mode', action="store", default=False, help='Mainly for testing purposes. If True, the daemon will be run without taking any actions on any files. The log file will be produced normally. By default False.')
```
joeldierkes: - `action="store_true"` is more appropriate here. It sets the value of the variable to `True` if specified, `False` otherwise. If you use `store`, the user has to provide a value: `--passive-mode True`.
- The default case can be omitted then.
- From my POV, the text should be edited to something like: ~Mainly for testing purposes.~ If ~True~ specified, the ... . ~By default False.~
https://docs.python.org/3/library/argparse.html#action

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
```diff
@@ -68,7 +69,7 @@
 GRACEFUL_STOP = threading.Event()
 
 
-def declare_suspicious_replicas_bad(once=False, younger_than=3, nattempts=10, vos=None, limit_suspicious_files_on_rse=5, sleep_time=300):
+def declare_suspicious_replicas_bad(once=False, younger_than=3, nattempts=10, vos=None, limit_suspicious_files_on_rse=5, sleep_time=3600, passive_mode=True):
```
joeldierkes: What should be the default case? In the argument parser it's `False`, here `True`. This should be consistent.

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
```diff
@@ -260,12 +261,13 @@ def declare_suspicious_replicas_bad(once=False, younger_than=3, nattempts=10, vo
                         for replica_value in recoverable_replicas[vo][rse_key].values():
                             surls_list.append(replica_value['surl'])
 
-                        add_bad_pfns(pfns=surls_list, account=InternalAccount('root', vo=vo), state='TEMPORARY_UNAVAILABLE', expires_at=datetime.utcnow() + timedelta(days=3))
+                        if not passive_mode:
+                            add_bad_pfns(pfns=surls_list, account=InternalAccount('root', vo=vo), state='TEMPORARY_UNAVAILABLE', expires_at=datetime.utcnow() + timedelta(days=3))
 
                         logger(logging.INFO, "%s is problematic (more than %s suspicious replicas). Send a Jira ticket for the RSE (to be implemented).", rse_key, limit_suspicious_files_on_rse)
                         logger(logging.INFO, "The following files on %s have been marked as TEMPORARILY UNAVAILABLE:", rse_key)
```
joeldierkes, line 254: Do we still need this line, the `TEMPORARILY UNAVAILABLE` value is now included in the other log message.

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
```diff
@@ -285,48 +287,50 @@ def declare_suspicious_replicas_bad(once=False, younger_than=3, nattempts=10, vo
                     # Get the rse_id by going to one of the suspicious replicas from that RSE and reading it from there
                     rse_id = list(recoverable_replicas[vo][rse_key].values())[0]['rse_id']
                     for replica_key in list(recoverable_replicas[vo][rse_key].keys()):
+                        file_metadata = get_metadata(recoverable_replicas[vo][rse_key][replica_key]["scope"], recoverable_replicas[vo][rse_key][replica_key]["name"])
+                        recoverable_replicas[vo][rse_key][replica_key]["datatype"] = str(file_metadata["datatype"])
                         if recoverable_replicas[vo][rse_key][replica_key]['available_elsewhere'] is True:
                             # Replicas with other copies on at least one other RSE can safely be labeled as bad
-                            files_to_be_declared_bad.append(recoverable_replicas[vo][rse_key][replica_key]['surl'])
+                            files_to_be_declared_bad.append(recoverable_replicas[vo][rse_key][replica_key])
```
joeldierkes, line 280: Was this a bug?

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
```diff
@@ -285,48 +287,50 @@ def declare_suspicious_replicas_bad(once=False, younger_than=3, nattempts=10, vo
...
                                         else:
                                             logger(logging.WARNING, "RSE: %s, replica name %s, surl %s: Match for the metadata 'datatype' (%s) of replica found in json file, but no match for 'action' (%s)",
                                                    rse_key, replica_key, recoverable_replicas[vo][rse_key][replica_key]['surl'], i["datatype"], i["action"])
                                         break
                                     else:
                                         # If no policy has be set, default to ignoring the file (no action taken).
-                                        files_to_be_ignored.append(recoverable_replicas[vo][rse_key][replica_key]['surl'])
+                                        files_to_be_ignored.append(recoverable_replicas[vo][rse_key][replica_key])
 
                     logger(logging.INFO, '(%s) Remaining replicas (pfns) that will be ignored:', rse_key)
```
joeldierkes, line 308: Do we need this line anymore?

joeldierkes (changes requested):
Good PR, just some general stuff.

bin/rucio-replica-recoverer
ChristophAmes: We changed it to be 1 hour, because the number of executions per run is relatively high. We will have to make some adjustments to the default value of `--limit-suspicious-files-on-rse` to compensate for this change at some time in the future.

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
ChristophAmes: The default case should be passive_mode=True, so that no accidents happen. Unfortunately, we currently only have the option to test the suspicious replicas recoverer on production servers, so it is very important that no actions are taken by mistake.

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
ChristophAmes: No, but we are changing the log messages for files that are ignored or declared bad (line 324, 327) so that we have more information in the log file about which replicas are being affected. For this change to work, we need more information than just the surl of each file.

joeldierkes: Okay, seems reasonable. This is also used in the `pfns`, which will be passed to the `declare_bad_file_replicas`. I think this should be fine.

joeldierkes:
Also, please split your commit into multiple commits. A commit should only do one thing!
There are a lot of blogs and websites out there explaining how to form an appropriate commit, and especially how not to:
- http://who-t.blogspot.com/2009/12/on-commit-messages.html
  ```
  # How not to do it
  Two changes in one patch. Something like "Fixed bug 2345 and renamed all foo to
  bar". Unless bug 2345 required the renaming, fixes whould be split it up into
  multiple patches. Others may have to take one of those bug fixes and apply it to
  a stable branch but not the other one. Picking bad patches apart into useful
  chunks is one of the most time-consuming and frustrating things I've done since
  it doesn't actually add any value to the project.
  ```
- https://www.git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines
- https://cbea.ms/git-commit/

cserf:
I 

> Also, please split your commit into multiple commits. A commit should only do one thing! There are a lot of blogs and websites out there explaining how to form an appropriate commit, and especially how not to:
> 
>     * http://who-t.blogspot.com/2009/12/on-commit-messages.html
>       ```
>       # How not to do it
>       Two changes in one patch. Something like "Fixed bug 2345 and renamed all foo to
>       bar". Unless bug 2345 required the renaming, fixes whould be split it up into
>       multiple patches. Others may have to take one of those bug fixes and apply it to
>       a stable branch but not the other one. Picking bad patches apart into useful
>       chunks is one of the most time-consuming and frustrating things I've done since
>       it doesn't actually add any value to the project.
>       ```
> 
>     * https://www.git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_commit_guidelines
> 
>     * https://cbea.ms/git-commit/

But https://rucio.cern.ch/documentation/contributing/
"Watch the pull request for comments and reviews. For any pull requests update, please try to squash/amend your commits to avoid “in-between” commits." :smile:

joeldierkes:
Yes, that is the case for in-between commits with messages like:
- Fix syntax
- Fix
- bla bla bla
- ...
Which we don't want in the history. There should not be a problem in splitting your commits into single work-packages.

This was also more like a "nice to have" and "best practice" comment. It is always a good idea to do that.

bin/rucio-replica-recoverer
joeldierkes: Okay

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
joeldierkes: We should activate it by default then and provide a way to disable it in the client? Something like: `--disable-passive-mode` or `--active-mode`?

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
ChristophAmes: Yes, this is so that we can keep track of which files are being ignored by referencing them in the log file.

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
ChristophAmes: This line assigns the replicas the "temporarily unavailable" state, it doesn't add them to the log file.

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
ChristophAmes: Alright, I've changed it to be "--disable-passive-mode", as I think this is a bit clearer.

bin/rucio-replica-recoverer
ChristophAmes: I'm going to change this to `--disable-passive-mode`, `action=store_false` and `default=True`, so that the passive mode is run by default. I'm going to kept the argument `passive_mode=ARGS.disable_passive_mode` in `run`, so that we can avoid having to implement a double-negative (something like `disable_passive_mode=False`) in `suspicious_replica_recoverer.py`.

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
ChristophAmes: Yes, this line is where the files are declared temporarily unavailable. It doesn't include anything in the log messages.

cserf (dismissed):
Looks good to me

bin/rucio-replica-recoverer
```diff
@@ -244,7 +242,8 @@ Note that attempting the use the ``--vos`` argument when in single-VO mode will
     parser.add_argument('--vos', nargs='+', type=str, help='Optional list of VOs to consider. Only used in multi-VO mode.')
     parser.add_argument("--run-once", action="store_true", default=False, help='One iteration only.')
     parser.add_argument("--limit-suspicious-files-on-rse", action="store", default=5, help='Maximum number of suspicious replicas on an RSE before that RSE is considered problematic and the suspicious replicas on that RSE are declared "TEMPORARY_UNAVAILABLE". Default value is 5.')
-    parser.add_argument('--sleep-time', action="store", default=300, type=int, help='Concurrency control: thread sleep time after each chunk of work')
+    parser.add_argument('--sleep-time', action="store", default=3600, type=int, help='Concurrency control: thread sleep time after each chunk of work')
+    parser.add_argument('--disable-passive-mode', action="store_false", default=True, help='If NOT specified, the daemon will run without taking any actions on any files. In either case, the log file will be produced normally.')
```
mlassnig: aarg, this confuses my brain :-) please no negative flags.

either `--passive-mode` default false

or `--active-mode` default true

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
```diff
@@ -260,12 +261,13 @@ def declare_suspicious_replicas_bad(once=False, younger_than=3, nattempts=10, vo
...
+                        if not passive_mode:
+                            add_bad_pfns(pfns=surls_list, account=InternalAccount('root', vo=vo), state='TEMPORARY_UNAVAILABLE', expires_at=datetime.utcnow() + timedelta(days=3))
 
                         logger(logging.INFO, "%s is problematic (more than %s suspicious replicas). Send a Jira ticket for the RSE (to be implemented).", rse_key, limit_suspicious_files_on_rse)
                         logger(logging.INFO, "The following files on %s have been marked as TEMPORARILY UNAVAILABLE:", rse_key)
-                        for rse_values in recoverable_replicas[vo][rse_key].values():
-                            logger(logging.INFO, 'Scope: %s    Name: %s', rse_values['scope'], rse_values['name'])
+                        for replica_values in recoverable_replicas[vo][rse_key].values():
+                            logger(logging.INFO, 'Temporarily unavailable: RSE: %s    Scope: %s    Name: %s    Pfn: %s', rse_key, replica_values['scope'], replica_values['name'], replica_values['surl'])
```
mlassnig: Nitpicking: `Pfn` -> `PFN`  (also a few times below)

ChristophAmes: Done

bin/rucio-replica-recoverer
ChristophAmes: I've opted for `--active-mode`.

joeldierkes (approved):
lgtm

bari12:
Merging now, just one comment: 
I think it would be better to reverse the `--active-mode` logic. 
If I run a daemon I would have the expectation that it does things. So the passive mode should be an opt-in from my point of view (Maybe `--passive-mode` or `--dry-run`)
We can address this in a later PR, but I think this would also be more in line with other daemons.

Issue #5384, opened 2022-03-24
Change the add header job base branch


joeldierkes:
Motivation
----------
The add header job always uses the `master` branch as base branch. In cases like the LTS releases, this does not make sense. We should use the target branch of the pull request.

It is ok to assume that there will be a target branch, since the add header job only runs on pull requests.

Modification
------------
Change the target branch to the GitHub provided environment variable `GITHUB_BASE_REF`. Also create a PR for the 1.27-LTS release line.

PR #5385, opened 2022-03-24
Use a generic base branch for the add header job #5384

joeldierkes:
The add header job runs on every pull request. It finds the changes python files
in comparison to the master branch. If the pull request is made for a branch
which is not the master (e.g. LTS branches), the changes won't be found.

This commit uses the environment variable provided by GitHub to select the
target branch. This allows us to idetify all changed files correctly.


maany (approved):
LGTM

Issue #4046, opened 2020-10-01
Test suite for importing python modules


bziemons:
Motivation
----------
There were several issues with importing modules that automatically load configuration values or importing modules that imported server dependencies on the clients.

Modification
------------
Create a test suite that outputs warnings for all files being imported that load configuration values so this general issue can be followed up on.
Also add a test to this suite that imports all client modules without any server or test pip dependencies, so that errors are thrown when unavailable dependencies are loaded.

Replaces `lib/rucio/tests/test_module_import.py`

maany:
Hey @bziemons, when you have some time, could you loop me into this?

LukasMoll:
Hi, I want to work on this.

Quesion: for the files that import configuration values, are those the files mentioned inside before_script.sh (for example alembic_sqlite.ini)?

@bziemons

bziemons:
@LukasMoll Specifically, I meant the use of this module: https://github.com/rucio/rucio/blob/c44effdbc4598db7b35e6b4df38e2c055acaa14b/lib/rucio/common/config.py

Using `config_get` without default value in the global scope (on import) will unfortunately crash Rucio (if the configuration file is absent or incomplete) without a good description of the error and should be avoided. Therefore, I thought a warning in a test suite would be a good start to catch these problems. Even with a default value set, to a new user/developer it is sometimes unclear what Rucio does and tests might not reach certain code.

LukasMoll:
@bziemons Thank you.

Here's an overview of what I'm thinking

I check in the cfg directories wether there is a file if not: warning, if: check wether all the fields are there, if not warning(references: [config1 template](https://github.com/rucio/rucio/blob/c44effdbc4598db7b35e6b4df38e2c055acaa14b/etc/rse-accounts.cfg.template), [config2 template](https://github.com/rucio/rucio/blob/c44effdbc4598db7b35e6b4df38e2c055acaa14b/etc/rucio.cfg.template)).

For the second part: scrape all imports from clients folder and try to import them, if not: error.

Does this sound good?

bziemons:
[...]

PR #5426, opened 2022-04-03
testing: Test suite for importing python modules #4046

LukasMoll:
**Tests all imports of python files inside client directory.**
Prints out all failed imports
Throws an error if one or more imports failed.
(I can add assert)

**Tests config and credentials based on templates.**
Prints a warning if it's unable to open the file, unable to read the file format or not able to compare.
Not sure how I can make this more detailed would love some input. I was thinking about individually checking components.
For example:
- for a URL check if it's actually a URL.
- for a password check if it's long enough.
- give a warning to values that should probably stay default.
etc.

I'm looking forward to your feedback.
@bziemons

### DIFF
--- a/lib/rucio/common/config.py
+++ b/lib/rucio/common/config.py
@@ -349,12 +349,15 @@ def get_lfn2pfn_algorithm_default():
     return default_lfn2pfn
 
 
-def get_rse_credentials(path_to_credentials_file=None):
+def get_rse_credentials(path_to_credentials_file=None, template=False):
     """ Returns credentials for RSEs. """
 
     path = ''
     if path_to_credentials_file:  # Use specific file for this connect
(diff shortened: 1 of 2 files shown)


bziemons:
hi, because there is no movement here: The tests fail with e. g.

```
_____________________ TestModuleImport.test_client_modules _____________________
[gw2] linux -- Python 3.6.8 /usr/bin/python
test_module_import.py:56: in test_client_modules
    result = module_find(path)
test_module_import.py:98: in module_find
    for fname in files:
E   TypeError: 'NoneType' object is not iterable
```

also as I stated in https://github.com/rucio/rucio/issues/4046#issuecomment-1081031017 I am pretty sure this does not report the mentioned lines, but I cannot see the result.

Issue #5435, opened 2022-04-04
Removal of Python 2.7 dependencies from the code


bari12:
Motivation
----------
With Rucio `1.29` we will drop all Python 2.7 dependencies and testing.

PR #5461, opened 2022-04-07
Fix headers with new `add_header` script

joeldierkes:
Fixes the file headers with the new add_header script. Should be added after #5456 or as the only one


bari12:
You are putting these `#!/usr/bin/env python3` shebang lines instead of the `#!/usr/bin/env python` ones. I was wondering if we should rather do this in a later step. If we want to cherry-pick this PR still to the `release-1.28` branch, then we need to keep some bins still on python(2).

joeldierkes:
Yes, good idea. I'll change the script so that the same shebag line as present will be used.

joeldierkes:
Done

joeldierkes:
I can open a second PR for the master with only `python3` support if you'd like. We could also do that later.

bari12 (approved):

bari12:
The second PR we can do when we do #5435

PR #5481, opened 2022-04-13
CLI can’t handle input files with more than 1000 datasets : Closes #5479

### LINKED ISSUE #5479: CLI can’t handle input files with more than 1000 datasets
Motivation
----------

https://github.com/rucio/rucio/blob/47e93cfbe5887071c70de4ba815c1bbdddfac2ce/bin/rucio#L1808

```
  ERROR: Provided object does not match schema. Details: Problem validating dids : [...] is too long
  Failed validating 'maxItems' in schema: [...]
```

This call has a limit of 1000 DIDs on the server side.

Modification
------------

The client should chunk it, if necessary.

cserf:
CLI can’t handle input files with more than 1000 datasets : Closes #5479

### DIFF
--- a/bin/rucio
+++ b/bin/rucio
@@ -30,6 +30,7 @@ import time
 import traceback
 import unittest
 import uuid
+from copy import deepcopy
 from datetime import datetime
 from functools import wraps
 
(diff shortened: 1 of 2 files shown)


bin/rucio
```diff
@@ -1805,23 +1806,26 @@ def add_lifetime_exception(args):
...
-            containers.append({'scope': scope, 'name': name})
-            summary[2] += 1
-        elif not meta['eol_at']:
-            logger.warning('%s:%s is not affected by the lifetime model' % (scope, name))
-            summary[3] += 1
-        else:
-            logger.info('%s:%s will be declared' % (scope, name))
-            datasets.append({'scope': scope, 'name': name})
-            summary[4] += 1
+    chunk_limit = 500  # Server should be able to accept 1000
```
joeldierkes, line 1809: Can't we take the `maxItems` entry (or some fraction of it) from one schema (e.g. atlas)?
https://github.com/rucio/rucio/blob/47e93cfbe5887071c70de4ba815c1bbdddfac2ce/lib/rucio/common/schema/atlas.py#L259

This reduces the number of hard-coded constants in our code by one.

bari12: The problem is that the schema is not necessarily part of the client, so this information might not be available to the client.

bari12 (approved):

Issue #4455, opened 2021-03-18
mypy static code analysis


bziemons:
Motivation
----------
Since there were many foreseeable problems with Python code and wrong type usages, I suggested adding mypy as a static code analysis tool.

For a quick introduction, see [mypy.pdf](https://github.com/rucio/rucio/files/6165303/mypy.pdf)

Modification
------------
Add mypy check script in tools/ and add CI step to automatically check type usages for PRs and commits.

bziemons:
@maany @joeldierkes I made a simple first script to run mypy on the client sources here: https://github.com/bziemons/rucio/blob/5408a5e32c657a3c46355949ec0502a7c22824d7/tools/check_mypy.sh (in case you haven't seen it yet)
Please message me on slack if you have any general questions. I assume you will take this task over, but I'll be happy to help out where I can.

maany:
Hey @bziemons , awesome. I will try to plug this in the autotest/integration-test workflows. I just need to check that mypy raises just warnings about the violations, and not hard errors, to begin with ;)

bari12:
I am closing this in regards to what is needed for 1.29 
However, this doesn't mean the mypi story in general is done, particularly we need:
- A guide of how we expect developers to use types in Rucio
- Activate the actual type-checking
- A sprint to actually improve the current code-base
But especially the second and third I see beyond 1.29. Please create tickets for these @maany and @joeldierkes

PR #5483, opened 2022-04-14
Feature 4455 mypy static code analysis

joeldierkes:
## Check the python type annotation count in the CI #4455

Since we want to introduce python annotations in our repository, we need a way
to check if new PRs include type annotations.

Since the repository contains a lot of code, we can not add type annotations in
a big PR. We should rather ensure that PRs annotate their new code and gradually
annotate old code.

This commit compared the number of missing type annotations before the PR,
i.e. on the most recent common ancestor commit with the rucio master, and after
the PR. The CI job fails if the number of missing type annotations increases,
since this means that new unannotated code was introduced. The re-viewers are
responsible for checking the integrity of the type-annotations for now.

## Add the `Checkout ancestor commit` action #4455

Getting the commit from which a branch is branched off of is helpful. The
progress of an implementation in a branch can be measured by comparing some
generated numbers before the feature branch and after. This enables us to do CI
tests based on differences in the committed code to the initial one.

This commit introduces the `Checkout common ancestor` action, which checks out
the most recent ancestor of the current working HEAD and a specified reference.

This implementation can copy files from the edited branch to the most recent
ancestor. This is necessary in case the file got modified and the modified one
is needed in the ancestor commit.

## Add the `count_missing_python_annotations` script #4455

With rucio 1.29 python2 got deprecated. This deprecation enables us to introduce
new features used by python3. One of these features is type annotations on the
code itself.

Type hints allow us to run a static type checker like `mypy` and analyze the
code for inconsistencies. Having a strongly typed codebase enables us to avoid
type inconsistency bugs, and it makes the code more readable, thus potentially
shorten the time to add new features.

The plan is to integrate a job in the CI/CD, which reports an error if the
number of missing type annotations is increased by a PR. This is the result of
newly written code, which is not annotated. This forces the user to annotate
their (or potentially some other) code.

This commit introduces the `count_missing_python_annotations_rucio` functions,
which uses `flake8` and the flake8 plugin `flake8-annotations` to count the
number of missing type annotations.

`mypy` enables us to check the type annotations for inconsistencies. However, we
can not use `mypy` to count the number of missing type annotations, since it
caps the number of reported errors to 200. We can use `mypy` in a later step to
 check the annotations for consistency.

Certain directories are excluded from the type checking:
 
- The test directory: Type checking the tests is not our goal right now. It
  decreases the time to write test code while providing very little additive.

- The db migration directory: The db migration should focus on the database, not
  the implementation. We could enable this in the future, but it has no priority
  right now.

- The tools directory: Scripts should be written quickly and should not
  necessarily require typed annotations. This can also be changed in the future,
  to enable type-save scripts.

- The bin directory: A strategy is required here on how to handle type
  annotations properly.

### DIFF
--- a/.github/actions/checkout_ancestor_commit/action.yaml
+++ b/.github/actions/checkout_ancestor_commit/action.yaml
@@ -1,0 +1,63 @@
+name: "Checkout Ancestor Commit"
+description: >-
+  Checks out the most recent common ancestor of two commits.
+
+  This action clones all necessary repositories, no prior clone required.
+
+inputs:
+  ref:
+    description: >-
+      The reference for which the ancestor commit, with the current checked out
+      commit, should be found.
+    required: false
+    default: master
+  repository:
+    description: >-
+      `repository_owner/repository_name` of the repository containing the
+      reference commit. For example, rucio/rucio.
+    default: rucio/rucio
+  files_to_copy:
+    description: >-
+      Necessary files to copy (e.g. scripts, configurations, ...) to run code in
+      the older version of the project. This is for the case an ancestor commit
+      gets checked out wich does not contain the required files.
+
+      New line character as delimiter of items.
+    required: false
+
+runs:
+  using: "composite"
+  steps:
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+    - name: Checkout Ancestor Commit
+      shell: bash
+      env:
+        COMMON_ANCESTOR_REPO: upstream
+      run: |
+        echo "Fetch the repository of the common ancestor branch."
+        git remote add $COMMON_ANCESTOR_REPO https://github.com/${{ inputs.repository }}.git
+
+        echo "Fetch the branch of the common ancestor."
+        git fetch $COMMON_ANCESTOR_REPO ${{ inputs.ref }}
+
+        FORK_POINT=$(git merge-base $COMMON_ANCESTOR_REPO/${{ inputs.ref }} HEAD)
+        echo "The fork point of the current branch with the master is $FORK_POINT"
+
+        echo "Backup all specified files since they could be changed."
+        IFS=$'\n'
+        FILES_TO_COPY=$'.github/actions/checkout_ancestor_commit/action.yaml\n${{ inputs.files_to_copy }}'
+        for f in $FILES_TO_COPY; do
+          mkdir -p ../tmp/$(dirname $f)
+          cp $f ../tmp/$f
+        done
+
+        echo "Checking out the fork point"
+        git checkout $FORK_POINT
+
+        echo "Restore backed up files."
+        for f in $FILES_TO_COPY; do
+          mkdir -p $(dirname $f)
+          cp ../tmp/$f $f
+        done
--- a/.github/workflows/autotest.yml
+++ b/.github/workflows/autotest.yml
@@ -19,6 +19,52 @@ jobs:
         shell: bash
         run: |
           python3 tools/add_header --dry-run --disable-progress-bar
+  python_annotations:
+    name: Check Python Type Annotations
+    runs-on: ubuntu-latest
+    env:
+      UPDATED_PYTHON_TYPE_ANNOTATIONS_REPORT_FILE: ../updated_type_annotation_report.txt
+      INITIAL_PYTHON_TYPE_ANNOTATIONS_REPORT_FILE: ../initial_type_annotation_report.txt
+    steps:
+      - uses: actions/checkout@v3
+      - name: Count current number of missing type annotations
+        id: count_updated
+        shell: bash
+        run: |
+          pip install flake8 flake8-annotations
+          source tools/count_missing_type_annotations_utils.sh
+          create_missing_python_type_annotations_report ${{ env.UPDATED_PYTHON_TYPE_ANNOTATIONS_REPORT_FILE }}
+          echo "The current number of missing python type annotations: $(wc -l < ${{ env.UPDATED_PYTHON_TYPE_ANNOTATIONS_REPORT_FILE }})"
+      - name: Checkout ancestor commit with rucio/master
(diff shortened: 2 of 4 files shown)


bziemons:
Generally looks like a good first step. I wish the actual log output would be more present, but that may be a good follow-up (thinking of a PR comment which is automatically updated, as I saw in other projects; example here: <https://github.com/containers/podman/pull/13913#issuecomment-1102303671>).

I think it would make sense to add `flake8-annotations` to the `dev_requirements` in `setuputil.py` and the requirements, so I can run `tools/count_missing_annotations.sh` without additional installs.

Last, I usually use the phrase "type annotations" to be more specific over "annotations". Developers coming from other backgrounds might associate something else without a closer look.

bziemons (dismissed):

joeldierkes:
Thanks for the feedback.

- **Automatically created PR comment**
  We can discuss that in the future, however we would need to include a summary of all CI tests which run, not just the type annotations.

- **Adding the flake8-annotations to the `dev_requirements`**
  I thought about this, however, it would make sense to do this once we reduced the number of missing type annotations to a reasonable number. The flake8 would contain a lot of noise, which could harm the workflow of the developer.

- **Annotations vs. Type Annotations**
  Yes you're right, annotations are general. I'll have a look into that and fix it in appropriate places. :+1:

.github/workflows/autotest.yml
```diff
@@ -19,6 +19,52 @@ jobs:
...
         run: |
           python3 tools/add_header --dry-run --disable-progress-bar
+  python_annotations:
+    name: Check Python Annotations
+    runs-on: ubuntu-latest
+    env:
+      UPDATED_PYTHON_ANNOTATIONS_REPORT_FILE: ../updated_annotation_report.txt
+      INITIAL_PYTHON_ANNOTATIONS_REPORT_FILE: ../initial_annotation_report.txt
+    steps:
+      - uses: actions/checkout@v2
```
maany: Let's move to chekcout@v3?

.github/workflows/autotest.yml
```diff
@@ -19,6 +19,52 @@ jobs:
...
+      - name: Count current number of missing annotations
+        id: count_updated
+        shell: bash
+        run: |
+          pip install flake8 flake8-annotations
```
maany, line 34: I would move this to `tools/count_missing_annotations.py`. This way if a dev runs the script locally, it would not fail ;)

.github/actions/checkout_ancestor_commit/action.yaml
```diff
@@ -0,0 +1,63 @@
+name: "Checkout Ancestor Commit"
+description: >-
+  Checksout the most recent common ancestor of two commits.
```
maany: minor typo: `Checks out`

.github/actions/checkout_ancestor_commit/action.yaml
```diff
@@ -0,0 +1,63 @@
...
+inputs:
+  ref:
+    description: >-
+      The reference for which the ancestor commit, with the current checked out
+      commit, should be found.
+    required: false
+    default: master
+  repository:
+    description: >-
+      Repository name with owner of the reference commit. For example,
```
maany: Really nitpicking here but `repository_owner/repository_name` of the repository containing the reference commit :P

maany (changes requested):
Minor changes but good work @joeldierkes.

.github/workflows/autotest.yml
joeldierkes: Okay, I changed the behavior a bit:

The `count_missing_annotations.py` script proceeds normally if all required packages are installed.
If there is a package missing, it installs it temporarily. The package will be removed once the scripts ends (e.g. error, normal exit, abort, ...). This way, users can run the script without enabling this feature of flake8 forever.

I would leave the installation of `flake8-annotations` at this point, so it can be used as a "cached" version and does not have to be installed every time the script runs.

maany (dismissed):
Nice work @joeldierkes. Shall we also add a `threshold` value for the missing type annotations such that when the number of missing type annotations is below the threshold, the CI will print a message like "Please enable mypi based type annotation checking for PRs"?

Otherwise, I guess in a year we might forget about it.

joeldierkes:
I think an open issue should be enough. If the CI is not failing, nobody looks into it. I don't want to let it fail, since I expect this to be an ongoing task over years. We might add `mypy` earlier.

We can discuss this with @bari12 and open another PR if we need one. :)

.github/workflows/autotest.yml
```diff
@@ -19,6 +19,52 @@ jobs:
...
+      - name: Compare numbers of missing type annotations
+        shell: bash
+        run: |
+          INITIAL_NUMBER_OF_MISSING_TYPE_ANNOTATIONS=$(wc -l < ${{ env.INITIAL_PYTHON_TYPE_ANNOTATIONS_REPORT_FILE }})
+          UPDATED_NUMBER_OF_MISSING_TYPE_ANNOTATIONS=$(wc -l < ${{ env.UPDATED_PYTHON_TYPE_ANNOTATIONS_REPORT_FILE }})
+
+          echo "The initial number of missing type annotations is: $INITIAL_NUMBER_OF_MISSING_TYPE_ANNOTATIONS"
+          echo "The updated number of missing type annotations is: $UPDATED_NUMBER_OF_MISSING_TYPE_ANNOTATIONS"
+          if [ $UPDATED_NUMBER_OF_MISSING_TYPE_ANNOTATIONS -gt $INITIAL_NUMBER_OF_MISSING_TYPE_ANNOTATIONS ]; then
+            echo "The number of python type annotations should never decrease! This way we ensure that new functions are type annotated."
```
bari12: Shouldn't this say `increase`?

joeldierkes: Good catch!

bari12 (approved):

PR #5490, opened 2022-04-19
Database: Fix alembic offline mode migration logic #5488

### LINKED ISSUE #5488: Fix alembic offline mode migration logic
Motivation
----------
Currently, in order to make an offline schema upgrade with alembic, one needs to specify the URL of the database just for the purpose of detecting the DB dialect.
https://github.com/rucio/rucio/blob/47e93cfbe5887071c70de4ba815c1bbdddfac2ce/lib/rucio/db/sqla/migrate_repo/env.py#L56

Furthermore, the migration script does not consider any starting revision from the configuration, this will end up producing all revisions unless the user specifies `alembic upgrade <CURRENT_REVISION>:head --sql` in the CLI.

Modification
------------

- Make the DB url configuration optional, and assume a `dialect` configuration option if the URL is not given. It makes sense to give the option to do the offline schema upgrade without having access to the DB in the working environment.
- Take into account the `starting_rev` configuration, in order to produce the upgrades from that starting point.

rizart:
This PR is a proof of concept for issue #5488.


bari12:
LGTM @mlassnig any comment?

etc/alembic_offline.ini.template
```diff
@@ -0,0 +1,78 @@
+# Copyright 2014-2019 CERN for the benefit of the ATLAS collaboration.
```
bari12: Cosmetic, but please change this header to the current one. :smile

bari12 (changes requested):
This can go, please just change the header.

etc/alembic_offline.ini.template
rizart: Should I also remove the authors as in the new header style?

bari12: Yes 😄

etc/alembic.ini.template
```diff
@@ -1,4 +1,5 @@
-# Copyright 2014-2019 CERN for the benefit of the ATLAS collaboration.
+# -*- coding: utf-8 -*-
+# Copyright CERN since 2014
```
mlassnig: The unified header for everything is now:

`Copyright European Organization for Nuclear Research (CERN) since 2012`

mlassnig (dismissed):

etc/alembic.ini.template
rizart: Thanks Mario, will fix! So, we put `2012` even for code/files written after that date?

mlassnig: Yes, because this covers the whole project.

bari12 (approved):

PR #5493, opened 2022-04-21
Monitoring & Logging: reduce gfal2 log level on daemon side

### LINKED ISSUE #5492: reduce gfal2 log verbosity
Motivation
----------

We have messages like 
```
Using bearer token for HTTPS request authorization
```
and 
```
[gfal_module_load] plugin /usr/lib64/gfal2-plugins//libgfal_plugin_xrootd.so loaded with success
```
spamming our logs millions of times per day in reaper

I didn't find any useful information in gfal logs on INFO level. Whatever is really useful is already extracted by rucio and printed in rucio log messages. So probably good idea to move to warning in gfal2 logs.

rcarpa:

### DIFF
--- a/lib/rucio/rse/protocols/gfal.py
+++ b/lib/rucio/rse/protocols/gfal.py
@@ -181,7 +181,10 @@ class Default(protocol.RSEProtocol):
         """
         self.logger(logging.DEBUG, 'connecting to storage')
 
-        gfal2.set_verbose(gfal2.verbose_level.verbose)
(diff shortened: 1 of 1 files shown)


maany:
LGTM, thanks @rcarpa ! Do you think we can already add this as a patch and see what log reduction we get?

rcarpa:
Already done. But I think it only impacts daemons. Not sure how many gfal connections we can have in servers. Server-side here is more "in contrast with client-side" than meaning "rucio servers"

bari12 (approved):

PR #5496, opened 2022-04-22
Feature 5455 rewrite the add header script

### LINKED ISSUE #5455: Rewrite the `add_header` script
Motivation
----------
The `add_header` script mentions all developers for the file in a list. This can lead to confusing changes in a PR, where a lot of additional changes are made. 

Modification
------------
Do not list all developers in the header. This reduces the header length itself, as well as unnecessary changes in the PR.

joeldierkes:

### DIFF
--- a/tools/add_header
+++ b/tools/add_header
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 # -*- coding: utf-8 -*-
-# Copyright CERN since 2018
+# Copyright European Organization for Nuclear Research (CERN) since 2012
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -30,7 +30,6 @@ the new header have to be provided too.
 import argparse
 import logging
 import os
-import re
 import sys
 from abc import ABC, abstractmethod
 from enum import Enum, auto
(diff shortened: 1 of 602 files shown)


tools/add_header
```diff
@@ -283,6 +283,75 @@ class PythonHeaderTemplate(HeaderTemplate):
...
+        return os.path.splitext(file_path)[1] == ".sh" or is_file_using_shebag
+
+    def __init__(self, file_path: str):
+        super().__init__(file_path)
+
+    def get_header(self) -> str:
+        file_creation_year = get_git_file_creation_year(self.file_path)
+
+        shebag_string = BashHeaderTemplate._get_file_shebag(self.file_path) or ""
+        copyright_string = "# Copyright CERN" \
```
mlassnig: I would still suggest to use the full name like this:

European Organization for Nuclear Research (CERN)

instead of just the abbreviation

CERN


Also, file_creation_year will be wrong on some of them due to the way some of the files were moved around in git history rewrites. Wouldn't it be easier just to use the overall project duration and simply say "since 2012" everywhere?

joeldierkes: 1. Yeah we can change that. I'd create a separate PR for this, since we also need to change this in the python files.
2. https://github.com/rucio/rucio/blob/47e93cfbe5887071c70de4ba815c1bbdddfac2ce/tools/add_header#L60 The `--follow` option in git should resolve this. This is why the year changes on some files. However we can set it fixed to 2012.

bari12 (approved):

Issue #5500, opened 2022-04-28
Deprecation message in Paramiko


joeldierkes:
Motivation
----------
Paramiko outputs a deprecation message due to an outdated security algorithm. We do not depend on that algorithm.
https://github.com/paramiko/paramiko/pull/2039

Modification
------------
The paramiko team is aware of the problem. They opened a PR to fix it.

joeldierkes:
Proposed solution: Set the `cryptography` version fixed to the last working one. The only relevant change in the new version 37.0.0 is the deprecation message, which fails some of our test cases. As soon as `paramiko` does not use the deprecated feature anymore we can dump up the version.

bari12:
Backport 1.26.13

PR #5504, opened 2022-04-28
Core & Internals: ignore some cryptography deprecation warning. #5500

rcarpa:
Paramiko.transport imports cryptography.Blowfish, which is deprecated.
Also, python 2 and 3.6 are deprecated in cryptography.
This makes our tests fail due to a warning.
Catch that warning and ignore it with the filter.

### DIFF
--- a/lib/rucio/common/extra.py
+++ b/lib/rucio/common/extra.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 
 import importlib
+import warnings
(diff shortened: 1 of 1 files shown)


joeldierkes:
We have to add this changes to the LTS releases as well.

joeldierkes (approved):
LGTM. Good to go as soon as the CI passes. :+1:

PR #5505, opened 2022-04-28
Core & Internals: ignore cryptography deprecation warnings. Closes #5500

### LINKED ISSUE #5500: Deprecation message in Paramiko

rcarpa:
Paramiko.transport imports cryptography.Blowfish, which is deprecated.
Also, python 2 and 3.6 are deprecated in cryptography.
This makes our tests fail due to a warning.
Catch that warning and ignore it with the filter.

### DIFF
--- a/lib/rucio/common/extra.py
+++ b/lib/rucio/common/extra.py
@@ -17,6 +17,7 @@
 # - Benedikt Ziemons <benedikt.ziemons@cern.ch>, 2021
 
 import importlib
+import warnings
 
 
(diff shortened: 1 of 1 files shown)


joeldierkes (approved):
Same as #5504

Issue #5443, opened 2022-04-05
Introduce heartbeat handler to all daemons


bari12:
Motivation
----------
The `HeartbeatHandler` which is used in the conveyor daemons offers several advantages, especially in terms of database interactions. The goal is to introduce this heartbeat handler to all daemons.

- [x] @rcarpa move the heartbeat handler to a common location
- [x] @rcarpa write a short information for component leads of do's & dont's for the handler https://github.com/rcarpa/rucio/pull/1
- [x] all component leads then introduce the handler to their daemons
  - [x] Atropos @dchristidis 
  - [x] Automatix @cserf 
  - [x] Minos, Minos_temporary_replicas, Necromancer @cserf 
  - [x] BB8 @cserf 
  - [x] Hermes @mlassnig 
  - [x] Hermes2 @cserf
  - [x] oauthmanager @rizart 
  - [x] replicarecoverer @cserf @ChristophAmes 
  - [x] kronos @maany 
  - [x] transmogrifier @cserf 
  - [x] undertaker @bari12

PR #5506, opened 2022-04-28
Deletion: refactor reaper. #5443. Closes #5442. Closes #5198

### LINKED ISSUE #5198: Reaper can loop up to 1000 times when called on a single RSE
Motivation
----------

Due to this formula,
https://github.com/rucio/rucio/blob/382571bf8d591d1e7596684f1841ecc772a79ed3/lib/rucio/daemons/reaper/reaper.py#L512
reaper is looping 1000 times when called on a single RSE. While this is the worst case scenario, similar race conditions are possible when many RSEs exist, but only one/<a couple> of them need data to be deleted. 

It's has low impact. Thanks to the "pause rses" protection, this isn't very problematic and only results in nuisances in tests where 1000 log messages pollute the test output. Still, it's worth fixing this race condition whenever we have time.

### LINKED ISSUE #5442: Re-factor/cleanup deletion workflow
Motivation
----------

Modification
------------

rcarpa:


rcarpa:
PR is based on top of #5495, please review that one first.

bari12 (approved):

PR #5510, opened 2022-04-28
Add Belle II tests : Closes #4864

### LINKED ISSUE #4864: Belle II : Add Belle II tests
Motivation
----------
In some recent release, there was a regression affecting RSE expression for RSE following the Belle II convention. In addition, there is no test for Dirac methods.

Modification
------------
Will introduce some Belle II specific tests to cover the issues mentioned above

cserf:
Add Belle II tests : Closes #4864

- Basic tests on B2 schema and add_file method in dirac
- Introduce new `did_name_generator` to generate DIDs valid for multiple VOs
- Fix r`se_name_generator` to match B2 schema

### DIFF
--- a/etc/docker/test/matrix_policy_package_tests.yml
+++ b/etc/docker/test/matrix_policy_package_tests.yml
@@ -1,6 +1,97 @@
+atlas:
+  installation_cmd: ""
+  config_overrides:
+    permission: atlas
+    extract_scope:  atlas
+    schema: atlas
+    lfn2pfn_algorithm_default: "hash"
+  tests:
+    allow:
+      - rucio_tests
+    deny:
+      - rucio_tests/test_bad_replica.py
+      - rucio_tests/test_rule.py
+      - rucio_tests/test_replica.py
+      - rucio_tests/test_account.py
+      - rucio_tests/test_rse_selector.py
+      - rucio_tests/test_s3.py
+      - rucio_tests/test_judge_injector.py
+      - rucio_tests/test_qos.py
+      - rucio_tests/test_did_meta_plugins.py
+      - rucio_tests/test_throttler.py
+      - rucio_tests/test_abacus_account.py
+      - rucio_tests/test_api_external_representation.py
+      - rucio_tests/test_authentication.py
+      - rucio_tests/test_bb8.py
+      - rucio_tests/test_rse_protocol_s3.py
+      - rucio_tests/test_oauthmanager.py
+      - rucio_tests/test_rse_lfn2path.py
+      - rucio_tests/test_oidc.py
+      - rucio_tests/test_pfns.py
+      - rucio_tests/test_curl.py
+      - rucio_tests/test_abacus_rse.py
+      - rucio_tests/test_boolean.py
+      - rucio_tests/test_meta.py
+      - rucio_tests/test_dumper_data_model.py
+      - rucio_tests/test_clients.py
+      - rucio_tests/test_did.py
+      - rucio_tests/test_dirac.py
+      - rucio_tests/test_permission.py
+      - rucio_tests/test_judge_evaluator.py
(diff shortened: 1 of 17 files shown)


etc/docker/test/matrix_policy_package_tests.yml
```diff
@@ -1,46 +1,28 @@
-atlas:
+belleii:
```
maany, line 92: @bari12 do we want to replace atlas vo-tests with belleii?

lib/rucio/common/utils.py
```diff
@@ -783,7 +783,7 @@ def extract_scope(did, scopes=None, default_extract=_DEFAULT_EXTRACT):
     if not _loaded_policy_package_scope_algorithms:
         register_policy_package_algorithms('scope', _EXTRACT_SCOPE_ALGORITHMS)
         _loaded_policy_package_scope_algorithms = True
-    extract_scope_convention = config_get('common', 'extract_scope', False, None)
+    extract_scope_convention = config_get('common', 'extract_scope', False, None) or config_get('policy', 'extract_scope', False, None)
```
maany, line 786: @cserf in this case `extract_scope` in `common` gets a priority over the `extract_scope` in `policy` section. Should it be the other way around?

etc/docker/test/matrix_policy_package_tests.yml
bari12: Hmm not really, @cserf what's the motivation here?

lib/rucio/common/utils.py
cserf: This is for backward compatibility. I'll remove the common one later

etc/docker/test/matrix_policy_package_tests.yml
cserf: I thought that all the autotests were already using ATLAS policy

lib/rucio/tests/common.py
```diff
@@ -64,12 +66,32 @@ def scope_name_generator():
...
+    return 'mock_' + str(uuid())
+
+
 def rse_name_generator(size=10):
     """ Generate random RSE name.
 
     :returns: A random RSE name
     """
-    return 'MOCK_' + ''.join(choice(ascii_uppercase) for x in range(size))
+    return 'MOCK-' + ''.join(choice(ascii_uppercase) for x in range(size))
```
maany, line 94: does the '-' work for both atlas and belleii ?

tools/bootstrap_tests.py
```diff
@@ -92,8 +122,11 @@
...
-    for i in additional_test_accounts:
+    for account in additional_test_accounts:
         try:
-            c.add_identity(account='root', identity=i[0], authtype=i[1], email=i[2])
+            client.add_identity(account='root', identity=account[0], authtype=account[1], email=account[2])
         except Exception:
-            print('Already added: ', i)
+            print('Already added: ', account)
+
+    if os.getenv('POLICY') == 'belleii':
```
maany, line 131: Minor comment, but just for consistency I think the `config_get` should work as the `rucio.cfg` would already contain the policy package info IMO. Could you check @cserf. Thanks

lib/rucio/tests/common.py
cserf: Yes

tools/bootstrap_tests.py
cserf: Will check

cserf: I checked with the config_get but it fails

maany: Okie thanks @cserf , I will add it to my list of todo and modify this code later on :)

maany (approved):
Looks good. Thanks @cserf

Issue #5346, opened 2022-03-09
Remove legacy S3 protocols and libraries


bari12:
Motivation
----------
In the Rucio dependencies we use both `boto` (Clients) and `boto3` (Server/Daemons)
`boto` has de-facto been deprecated in favour of `boto3`. With the drop of python2 for Rucio 1.29, we can finally fully remove `boto` in favour of `boto3`.

Modification
------------
Remove `boto` and replace with `boto3`.

jamesp-epcc:
One option would be to remove the protocol that uses boto altogether. The new method using the signed urls uses boto3 on the server and doesn't require any S3-specific libraries on the client. Though we should check if anyone is still using the s3boto protocol before doing that. (There is also `s3.py` which uses the S3 library instead of boto but otherwise it looks as if it basically does the same thing).

bari12:
If nobody uses it, I would also be in favour of removing and simplify things.

jamesp-epcc:
I've started putting together a PR to remove this legacy code. I wasn't sure what to do about `light_reaper.py`. It doesn't use the S3 libraries, but it does manually construct a `s3://` URL. Is this still the right thing for it to do?

mlassnig:
Any kind of s3:// URL seems wrong to me.

rizart:
Not sure if this helps, but FTS expects `https://` when you provide an already signed URL, as is the case now for Rucio. This supports additionally Mario's point.

jamesp-epcc:
I agree, we should probably not be using `s3://` URLs now. In the case of the light reaper, it appears it used to call `rsemgr.lfns2pfns` to generate the URL, but at some point was changed to construct only S3 URLs instead:

https://github.com/rucio/rucio/blob/5e92ee1e6d323d7e2f9eeb6e97ecb4f6fc4c3abe/lib/rucio/daemons/reaper/light_reaper.py#L108-L111

This seems strange to me. I have no experience with the light reaper so I'm not sure what the correct fix is.

rizart:
Please also pay attention to [this](https://docs.google.com/document/d/1ofV5iIgpCKCnANNuqIavN_6FvvMTE6QPtQKK_n1P3SE/edit#heading=h.bvzk0w47bfs7) behavior when it comes to the connection of signed URLS and GFAL/FTS.

jamesp-epcc:
[...]

PR #5512, opened 2022-04-29
Protocols: remove legacy S3 protocols and libraries #5346

jamesp-epcc:
This PR removes the old S3 protocols that accessed objectstores via s3cmd and boto. The newer code that uses URL signing and https protocol should be used instead. This should not affect replications via FTS, which will still use `s3://` URLs when required.


mlassnig (approved):

bari12:
Merged against `1.29`

aportelli:
Hi all, I am working on a project for using Rucio with lattice QCD data and cloud storage. We have a developpment project in progress and we are using `rucio.rse.protocols.s3boto`. Could I please ask you if there is an alternative for S3 with the new version coming?

@jamesp-epcc FYI I am working at the University of Edinburgh this is a project for DiRAC, happy to get in touch offline if you have time!

jamesp-epcc:
@aportelli Apologies for late reply, been away on leave and only just seen this. Yes, there is an alternative. The new way to use S3 is to set Rucio up to use signed URLs and then use https to access the RSEs. The documentation for setting this up needs to be improved. If you need any help, please feel free to get in touch with me at `j.perry@epcc.ed.ac.uk`.

PR #5519, opened 2022-05-03
Feature 5491 rucio admin rse update shows no error for unsupported options

### LINKED ISSUE #5491: rucio-admin rse update shows no error for unsupported options
Motivation
----------

e.g. `rucio-admin rse update --setting foo --value bar --rse RSE`

Modification
------------

1. Use argparse’s choices in the client.
2. Raise an exception in the core.

joeldierkes:
Core & Internals: Throw an exception on wrong RSE update option #5491

The user can only change specific values of an RSE. If they tried changing an
option which does not exist, the server and therefore the client simply did not
change anything and continued as normal. We want to throw an exception in this
case, to notify the user that there was an invalid input.

This commit changes the RSE update function to throw an exception if the caller
tries to change a non-existing property.

### DIFF
--- a/lib/rucio/core/rse.py
+++ b/lib/rucio/core/rse.py
@@ -41,7 +41,7 @@ from rucio.db.sqla.constants import (RSEType, ReplicaState)
 from rucio.db.sqla.session import read_session, transactional_session, stream_session
 
 if TYPE_CHECKING:
-    from typing import Dict, Optional, Sequence
+    from typing import Any, Dict, Optional, Sequence
     from sqlalchemy.orm import Session
 
 REGION = make_region_memcached(expiration_time=900)
@@ -1395,7 +1395,7 @@ MUTABLE_RSE_PROPERTIES = {
 
 
 @transactional_session
-def update_rse(rse_id, parameters, session=None):
+def update_rse(rse_id: str, parameters: 'Dict[str, Any]', session=None):
     """
     Update RSE properties like availability or name.
 
--- a/lib/rucio/core/rse.py
+++ b/lib/rucio/core/rse.py
@@ -1374,6 +1374,26 @@ def del_protocols(rse_id, scheme, hostname=None, port=None, session=None):
                     i += 1
 
 
+MUTABLE_RSE_PROPERTIES = {
+    'name',
+    'availability_read',
+    'availability_write',
+    'availability_delete',
+    'latitude',
+    'longitude',
+    'time_zone',
+    'rse_type',
+    'volatile',
+    'deterministic',
+    'region_code',
+    'country_name',
+    'city',
+    'staging_area',
+    'qos_class',
+    'availability'
+}
+
+
 @transactional_session
 def update_rse(rse_id, parameters, session=None):
     """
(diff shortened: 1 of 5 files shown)


joeldierkes:
I changed the proposed modification from the issue a bit:
Instead of adding an `argarse choice` to the client, I send the information about the wrong option from the server and display that in the client. These checks should always happen on the backend side. We also avoid code duplication.
General thoughs on that?

joeldierkes:
I changed the proposed modification from the issue a bit:
Instead of adding an `argarse choice` to the client, I send the information about the wrong option from the server and display that in the client. These checks should always happen on the backend side. We also avoid code duplication.

joeldierkes:
I changed the proposed modification from the issue a bit:
Instead of adding an `argarse choice` to the client, I send the information about the wrong option from the server and display that in the client. These checks should always happen on the backend side. We also avoid code duplication.

lib/rucio/core/rse.py
```diff
@@ -1229,7 +1249,12 @@ def update_rse(rse_id, parameters, session=None):
     :param session: The database session in use.
 
     :raises RSENotFound: If RSE is not found.
+    :raises InputValidationError: If a parameter does not exist. Nothing will be added then.
     """
+    for key in parameters.keys():
+        if key not in MUTABLE_RSE_PROPERTIES:
+            raise exception.InputValidationError(f"The key '{key}' does not exist for RSE properties.")
```
dchristidis, line 1411: I don’t know if the following is better or worse, but I’ve been recently seeing it a lot, so just putting it for your consideration.
```python
if any(key not in MUTABLE_RSE_PROPERTIES for key in parameters.keys()):
    raise exception.InputValidationError(f"The key '{key}' does not exist for RSE properties.")
```

joeldierkes: I like the idea and changed it. However, the variable `key` is used in the for loop to create a more descriptive error message.

joeldierkes: We could write something like
```python
    for key in parameters.keys():
        raise exception.InputValidationError(f"The key '{key}' does not exist for RSE properties.") if key not in MUTABLE_RSE_PROPERTIES
```
However, I don't like this syntax. It "hides" the condition and makes the line unnecessary longer, while just saving one line.

dchristidis: Darn, you’re very right, I completely missed this implication. Apologies for the noise.

lib/rucio/core/rse.py
```diff
@@ -1219,8 +1219,28 @@ def del_protocols(rse_id, scheme, hostname=None, port=None, session=None):
                     i += 1
 
 
+MUTABLE_RSE_PROPERTIES = {
```
bari12, line 1377: @dchristidis @cserf 
I am wondering a bit about the `rse_type`, `deterministic` and `staging_area` attribute. Especially the first two if this is something we should allow to be changed? Opinions?

dchristidis: Yes, given how many people have the necessary privileges to do this in ATLAS, the idea that it can silently happen is not comforting. But I don’t have a better proposal.

bari12: That's more of a permission issue in ATLAS though. I am more wondering if these should be mutable at all, or if we just say that in these cases people need to make a new RSE. Especially for `rse_type`. Myself I am also a bit conflicted on it. I can see reasons where this makes sense, but especially the `deterministic` setting is dangerous, since this could corrupt the catalog.

joeldierkes: Any updates on this?

joeldierkes: Also keep in mind, that the `import_rses` uses the `update_rse` function. This functionality is used in the `rucio-admin` to import rses and update existing ones...

bari12: I thought we discussed this in our followup 😄 Keep it how it works right now.

bari12 (approved):

PR #5524, opened 2022-05-05
Clean-up unneeded RSEs + Start introducing rse_factory in tests : Closes #5499

### LINKED ISSUE #5499: Cleanup unneeded RSEs in sync_rses.py
Motivation
----------
In `sync_rses.py` a lot of RSEs are defined but not used anywhere. They need to be cleaned. Moreover, some of the RSEs do not match the schema from some VOs. Naming needs to be fixed.
Longer term is to move everything to `rse_factory`

cserf:
Tests : Cleanup unneeded RSEs in sync_rses.py #5499 

- Remove unneeded RSEs
- Start moving to `rse_factory` in a few tests `test_dids`, `test_subscriptions`, etc.

### DIFF
--- a/etc/rse_repository.json
+++ b/etc/rse_repository.json
@@ -302,60 +302,6 @@
     },
     "storage_usage_tool": "rucio.rse.protocols.posix.Default.getSpace"
   },
-  "MOCK_SUSPICIOUS":{
-    "backend_type": "POSIX",
-    "deterministic": true,
-    "protocols": {
-      "default": "file",
-      "supported": {
-        "file": {
-          "prefix": "/tmp/rucio_rse1/",
-          "impl": "rucio.rse.protocols.posix.Default",
-          "extended_attributes": {"web_service_path": "/srm/managerv2?SFN=", "space_token": "ATLASDATADISK1"},
-          "domains": {
-            "lan": {
-              "read": 1,
-              "write": 1,
-              "delete": 1
-            },
-            "wan": {
-              "read": 1,
-              "write": 1,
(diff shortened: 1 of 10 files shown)


cserf:
Need rebase when https://github.com/rucio/rucio/pull/5525 is merged

rcarpa (dismissed):

bari12:
@cserf I think there are some tests related to this PR failing

maany:
@bari12 I need to look why the integration tests are failing ( and where are they exactly creating the mock-rse's from)

cserf:
All tests are fixed now, except the integration-tests

rcarpa:
That integration test workflow is very complex. I would bet it uses an old version of rse_repository.json from the master or container rather than using the one from the current commit.

rcarpa:
Probably this one : https://github.com/rucio/containers/blob/master/dev/rse_repository.json

cserf:
I see. Would it be possible to update the json file ? If we remove the RSEs before this PR is merged, all the other tests will fail, so maybe what can be done is just add the new `SSH-DISK` to begin with, then when the PR is merged, the RSEs can be removed and we ask people to rebase their open PR.

rcarpa:
yes, probably duplicating the RSE is the best solution. I'm not 100% sure the issue comes from there howether.

cserf:
Rebase

cserf:
Once all the tests succeed, this PR needs to be merged asap, otherwise the integration tests will fail for the other PRs.

bari12 (approved):

PR #5525, opened 2022-05-06
Tests: migrate test_did.py to pytest

rcarpa:
No functional rework at all. Only get rid of unittest.testcase
and use equivalent fixtures.


joeldierkes (approved):
LGTM!

PR #5530, opened 2022-05-06
Hermes 2 evolution : Closes #4610

### LINKED ISSUE #4610: Hermes 2 evolution

cserf:
Recreating PR after closing the first one : https://github.com/rucio/rucio/pull/5127

### DIFF
--- a/etc/docker/dev/docker-compose.yml
+++ b/etc/docker/dev/docker-compose.yml
@@ -1,13 +1,21 @@
+version: "3"
 services:
   rucio:
     image: docker.io/rucio/rucio-dev
     extra_hosts:
       - "ruciodb:127.0.0.1"
       - "graphite:127.0.0.1"
+      - "influxdb:127.0.0.1"
+      - "elasticsearch:127.0.0.1"
+      - "activemq:127.0.0.1"
     ports:
       - "127.0.0.1:8443:443"
       - "127.0.0.1:5432:5432"
       - "127.0.0.1:8080:80"
+      - "127.0.0.1:8086:8086"
+      - "127.0.0.1:9200:9200"
+      - "127.0.0.1:9300:9300"
+      - "127.0.0.1:61613:61613"
     volumes:
       - ../../../tools:/opt/rucio/tools:Z
       - ../../../bin:/opt/rucio/bin:Z
@@ -27,3 +35,31 @@ services:
   graphite:
     image: docker.io/graphiteapp/graphite-statsd
     network_mode: "service:rucio"
+  influxdb:
+    image: influxdb:latest
+    network_mode: "service:rucio"
+    environment:
+      - DOCKER_INFLUXDB_INIT_MODE=setup
+      - DOCKER_INFLUXDB_INIT_USERNAME=myusername
+      - DOCKER_INFLUXDB_INIT_PASSWORD=passwordpasswordpassword
+      - DOCKER_INFLUXDB_INIT_ORG=rucio
+      - DOCKER_INFLUXDB_INIT_BUCKET=rucio
+      - DOCKER_INFLUXDB_INIT_ADMIN_TOKEN=mytoken
+  elasticsearch:
+    image: docker.io/elasticsearch:7.4.0
+    network_mode: "service:rucio"
+    environment:
+      - discovery.type=single-node
+  activemq:
+    image: docker.io/webcenter/activemq:latest
+    network_mode: "service:rucio"
+    environment:
+      - ACTIVEMQ_CONFIG_NAME=activemq
+      - ACTIVEMQ_CONFIG_DEFAULTACCOUNT=false
+      - ACTIVEMQ_USERS_fts=supersecret
+      - ACTIVEMQ_GROUPS_writes=fts
+      - ACTIVEMQ_USERS_receiver=supersecret
+      - ACTIVEMQ_GROUPS_reads=receiver
+      - ACTIVEMQ_CONFIG_SCHEDULERENABLED=true
+      - ACTIVEMQ_USERS_hermes=supersecret
+      - ACTIVEMQ_CONFIG_QUEUES_events=/queue/events'
--- a/lib/rucio/common/config.py
+++ b/lib/rucio/common/config.py
@@ -268,13 +268,19 @@ def config_get_list(section, option, raise_exception=True, default=None, check_c
     from rucio.common.utils import is_client
     client_mode = is_client()
     try:
-        return get_config().get(section, option)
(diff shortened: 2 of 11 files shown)


lib/rucio/core/message.py
```diff
@@ -43,39 +40,45 @@ def add_message(event_type, payload, session=None):
...
-        raise InvalidObject('Invalid JSON for payload: %(e)s' % locals())
-
-    if len(payload) > 4000:
-        new_message = Message(event_type=event_type, payload='nolimit', payload_nolimit=payload, services=services_list)
-    else:
-        new_message = Message(event_type=event_type, payload=payload, services=services_list)
+    except TypeError as err:  # noqa: F841
+        raise InvalidObject('Invalid JSON for payload: %(err)s' % locals())
+
+    services = services_list.split(',')
```
joeldierkes: We already have a `config_get_int`, `config_get_bool`, ... function in the `common/config.py`. This here is the best opportunity to introduce a `config_get_list`.

lib/rucio/core/message.py
```diff
@@ -43,39 +40,45 @@ def add_message(event_type, payload, session=None):
...
+    for service in services:
+        service.strip(' ')
```
joeldierkes: This should be useless then.

lib/rucio/core/message.py
```diff
@@ -19,17 +19,14 @@
...
-from rucio.common.cache import make_region_memcached
 from rucio.common.config import config_get
 from rucio.common.exception import InvalidObject, RucioException
 from rucio.common.utils import APIEncoder
 from rucio.db.sqla import filter_thread_work
 from rucio.db.sqla.models import Message, MessageHistory
 from rucio.db.sqla.session import transactional_session
 
-REGION = make_region_memcached(expiration_time=900)
+SUPPORTED_SERVICES = ['influx', 'elastic', 'email', 'activemq']
```
joeldierkes: This is a well-defined list, let's use an Enum here. This has several advantages:
- We can test it.
- We don't have to pass strings around, it acts as a "Type"
- We can implement custom function such as `from_string`, which automatically test for corrupted inputs.

lib/rucio/core/message.py
```diff
@@ -43,39 +40,45 @@ def add_message(event_type, payload, session=None):
...
+    for service in services:
+        service.strip(' ')
+        if service not in SUPPORTED_SERVICES:
+            raise RucioException('Unsupported service type (%s)  for messages' % service)
+        if event_type == 'email' and service != 'email':
+            continue
+        if service == 'email' and event_type != 'email':
+            continue
+
+        if len(payload) > 4000:
```
joeldierkes: Hard coded constant.

joeldierkes (changes requested):

cserf:
Last commit is to move to new heartbeat mechanism https://github.com/rucio/rucio/issues/5443

lib/rucio/core/message.py
cserf: Enum was implemented. Can you please have a look ?

joeldierkes: :+1:

joeldierkes (dismissed):
lgtm

etc/docker/dev/docker-compose.yml
```diff
@@ -1,13 +1,21 @@
+version: "3"
 services:
   rucio:
     image: docker.io/rucio/rucio-dev
     extra_hosts:
       - "ruciodb:127.0.0.1"
       - "graphite:127.0.0.1"
+      - "influxdb:127.0.0.1"
```
mlassnig, line 8: oof, you want this in the default dev environment?

lib/rucio/common/constants.py
```diff
@@ -59,17 +59,33 @@
...
     AVAILABLE = 'A'
     UNAVAILABLE = 'U'
     COPYING = 'C'
     BEING_DELETED = 'B'
     BAD = 'D'
     TEMPORARY_UNAVAILABLE = 'T'
+
+
+@enum.unique
+class HermesSupportedServices(str, enum.Enum):
```
mlassnig: Maybe just `HermesService` (singular, and shorter) ?

etc/docker/dev/docker-compose.yml
cserf: Not so big :
```
docker images | grep influx
influxdb                                        latest              2639b070d3b3   4 weeks ago    341MB
```
At least, it's much smaller than Rucio image :sweat_smile:

mlassnig: ah, that's not what i meant :-) when you start up Elastic&Co it eats away the RAM

lib/rucio/common/constants.py
cserf: Done

etc/docker/dev/docker-compose.yml
bari12: @mlassnig @cserf is this really such an issue? Seems ok to me in terms of mem consumption?

lib/rucio/common/config.py
```diff
@@ -237,6 +237,72 @@ def config_get_bool(section, option, raise_exception=True, default=None, check_c
                 raise err_
 
 
+def config_get_list(section, option, raise_exception=True, default=None, check_config_table=True, session=None,
```
bari12, line 244: Please rebase this and make it similar to how @joeldierkes adapted the config_get methods. It's much simpler now and should be easy.

bari12 (commented):
I think this is GTG, @cserf please just add the one change (config)

etc/docker/dev/docker-compose.yml
cserf: I don't think it's an issue. The tests dont't crash on my machine nor in github

bari12 (approved):

PR #5535, opened 2022-05-09
Rules: only re-evaluate dataset size, but on all databases.  Closes #5531 

### LINKED ISSUE #5531: dataset length/bytes only re-evaluated on oracle
Motivation
----------

For some reason, dataset size is only re-evaluated on oracle: 
https://github.com/rucio/rucio/blob/b10d485fd31555d17ac1e4143380a3b09ab83a4a/lib/rucio/core/rule.py#L1487 

Modification
------------
Remove oracle constrain

rcarpa:
The evaluation of the container size gives completely wrong results,
because DataIdentifierAssociation bytes is not updated for collections.
Moreover, for containers, it will always compute the number of
children in the container, not the number of files, which is probably
not what we want.

So make it explicit that we don't re-evaluate containers.

Also, remove the constraint of doing this evaluation only on oracle,
which makes no sense.

### DIFF
--- a/lib/rucio/core/rule.py
+++ b/lib/rucio/core/rule.py
@@ -35,7 +35,7 @@ from six import string_types
 from sqlalchemy.exc import IntegrityError, StatementError
 from sqlalchemy.orm.exc import NoResultFound
 from sqlalchemy.sql import func
-from sqlalchemy.sql.expression import and_, or_, text, true, null, tuple_, false
+from sqlalchemy.sql.expression import and_, or_, text, true, null, tuple_, false, select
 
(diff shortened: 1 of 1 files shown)


rcarpa:
@cserf mentioned offline that they don't care at all about dataset and container length/bytes being correct. So this should rather be a behind a configuration option.

bari12:
This looks good to me. I think what @cserf mentioned is more about the container sizes. So #5559 which probably should be an opt-out. Per default I think people would expect this to be calculated.

rcarpa:
This needs more thinking and work. Clossing for now

Issue #5250, opened 2022-02-17
rucio get doesn't show summary anymore


cserf:
Motivation
----------
CLI : rucio get doesn't show summary anymore

bari12:
Adding some more info from DAST
```
2022-01-26 18:26:01,532 INFO    Processing 1 item(s) for input
2022-01-26 18:26:02,546 INFO    No preferred protocol impl in rucio.cfg: No section: 'download'
2022-01-26 18:26:02,546 INFO    No preferred protocol impl in rucio.cfg: No section: 'download'
2022-01-26 18:26:02,546 INFO    No preferred protocol impl in rucio.cfg: No section: 'download'
2022-01-26 18:26:02,546 INFO    No preferred protocol impl in rucio.cfg: No section: 'download'
2022-01-26 18:26:02,546 INFO    No preferred protocol impl in rucio.cfg: No section: 'download'
2022-01-26 18:26:02,546 INFO    No preferred protocol impl in rucio.cfg: No section: 'download'
2022-01-26 18:26:02,546 INFO    No preferred protocol impl in rucio.cfg: No section: 'download'
2022-01-26 18:26:02,547 INFO    No preferred protocol impl in rucio.cfg: No section: 'download'
2022-01-26 18:26:02,547 INFO    No preferred protocol impl in rucio.cfg: No section: 'download'
2022-01-26 18:26:02,547 INFO    No preferred protocol impl in rucio.cfg: No section: 'download'
2022-01-26 18:26:02,547 INFO    No preferred protocol impl in rucio.cfg: No section: 'download'
2022-01-26 18:26:02,553 INFO    Using 5 threads to download 11 files
2022-01-26 18:26:02,553 INFO    Thread 0/5: Preparing download of user.XXX:user.XXX.27765327._000001.out-skim.root
2022-01-26 18:26:13,056 INFO    Thread 1/5: Preparing download of user.XXX:user.XXX.27765327._000002.out-skim.root
2022-01-26 18:26:13,057 INFO    Thread 2/5: Preparing download of user.XXX:user.XXX.27765327._000003.out-skim.root
2022-01-26 18:26:15,036 INFO    Thread 3/5: Preparing download of user.XXX:user.XXX.27765327._000004.out-skim.root
2022-01-26 18:26:37,719 INFO    Thread 0/5: File exists already locally: user.XXX:user.XXX.27765327._000001.out-skim.root
[...]

PR #5536, opened 2022-05-10
Clients: Change the download client state to an enum #5250

joeldierkes:
The client state uses strings to store the state. This has the problem that the
evalutation, if a client state is wrong, is done at runtime. Also, it uses the
exception name as state. This is inconsistent, because all available states
should be well defined.

This commit uses an enum for the download client state. This enables checks at
runtime, since all usable states are defined. The values of the enum ensure the
serialization to strings.

It also changes the usage from an exception name as state to the `FAILING`
state. The exception itself is added in the body already.

### DIFF
--- a/lib/rucio/client/downloadclient.py
+++ b/lib/rucio/client/downloadclient.py
@@ -16,6 +16,7 @@
 from __future__ import division
 
 import copy
+import enum
 import itertools
 import logging
 import os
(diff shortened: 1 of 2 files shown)


mlassnig (approved):
LGTM, but there might be some downstream effects of this on the trace handling. please keep kronos daemon in mind when dealing with traces.

Issue #5247, opened 2022-02-16
Migration of SQLAlchemy queries to new query syntax (1.4 -> 2.0)


bari12:
Motivation
----------
With SQLAlchemy 2.0 the current [sqlalchemy.orm.Session](https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.Session) query syntax will become legacy and the new core query syntax should be used [sqlalchemy.sql.expression.select](https://docs.sqlalchemy.org/en/20/core/selectable.html#sqlalchemy.sql.expression.Select). This is already available in version 1.4.

Modification
------------
Over time, we need to migrate all queries to this new query syntax. This needs to be done module by module, with new PR only accepting the new query syntax.

yuyiguo:
@bari12 
Which rucio modules are good ones to start with?

bari12:
Most of the SQLAlchemy interaction will be in the core

https://github.com/rucio/rucio/tree/master/lib/rucio/core

Best start with the smaller ones: account, distance, identity, ...

yuyiguo:
@bari12 

@ericvaandering and I tried to generated `RemovedIn20Warning`. We enabled cms rucio int server with 
```
SetEnv PYTHONWARNINGS always::DeprecationWarning
SetEnv SQLALCHEMY_WARN_20 1
```
https://github.com/yuyiguo/containers/blob/set-python-envs/server/rucio.conf.j2#L96
https://github.com/yuyiguo/CMSKubernetes/blob/set-python-envs/docker/rucio-server/Dockerfile#L17 

I ran below commands:
```
rucio list-dataset-replicas cms:/BTag/Run2012A-v1/RAW#00e8f24c-993d-11e1-9605-003048caaace --deep
rucio list-rules "cms:/BTag/Run2012A-v1/RAW#00e8f24c-993d-11e1-9605-003048caaace"
```
In the server logs, I found:
```
k logs server-rucio-server-ff68c4d5b-7xp4d -c rucio-server | grep yuyi

[...]

PR #5539, opened 2022-05-11
Core & Internals: core/did queries in sqlalchemy2 syntax. #5247 

rcarpa:


bari12 (approved):

bari12:
I had to revert both commits due to #5593 
I haven't looked into the exact problem there, but it is confirmed that it is these changes causing the regression.

PR #5542, opened 2022-05-11
Protocols: Create posix.Symlink to create links on download; Fix #5319

### LINKED ISSUE #5319: posix protocol modifications
Motivation
----------

in ATLAS StoRM sites can can read directly from the file system (probably write too) but the path resolution is done via storm server. 

Some sites setting up cephFS would like to read directly from the FS but there is no possibility at the moment.  

Modification
------------

Modify posix protocol to allow direct access to the fs. At the moment it copies the files but the best thing would be to have a symlink like storm sites but without needing to get the information from a storage server.

this would simplify the setup at sites removing the need of extra hardware to handle the read and possibly write to the file system.

snafus:
Subclass of the Default POXIS implementation to allow a symlink to be made to the remote location
when issuing at 'get' of the remote file.

Only the get() method is overridden to provide this functionality. Other methods use the default implementation,
which involve using the remote file directly.

### DIFF
--- a/lib/rucio/rse/protocols/posix.py
+++ b/lib/rucio/rse/protocols/posix.py
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import logging
 import os
 import os.path
 import shutil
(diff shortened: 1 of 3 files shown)


lib/rucio/rse/protocols/posix.py
```diff
@@ -16,7 +16,9 @@
 import os
 import os.path
 import shutil
+import logging
```
mlassnig: please sort imports alphabetically

mlassnig (commented):
tiny comment. and you might want to add a unittest?

snafus:
Hi @mlassnig; Sorry for the delay, and the problems above. The code was refactored a little to allow (python 2) tests to pass and retested. James

lib/rucio/rse/protocols/posix.py
```diff
@@ -210,3 +211,46 @@ def stat(self, pfn):
...
+        if not os.lstat(dest):
+            # problem in creating the symlink
+            self.logger(logging.ERROR, 'Symlink {} could not be created'.format(dest))
+            raise exception.DestinationNotAccessible()
+        if not os.path.exists(dest):
+            # could not find the file following the symlink
+            self.logger(logging.ERROR, 'Symlink {} appears to be a broken link to {}'
+                        .format(dest, path))
+            if os.lstat(dest) and os.path.islink(dest):
+                call(['rm', '-rf', dest])
```
mlassnig: this can't be done with shutil.rmtree() ?

calling rm -rf seems so ... :scream:

lib/rucio/rse/protocols/posix.py
```diff
@@ -210,3 +211,46 @@ def stat(self, pfn):
...
+            :raises SourceNotFound: if the source file was not found on the referred storage.
+         """
+        try:
+            path = self.pfn2path(pfn)
+            os.symlink(path, dest)
+            self.logger(logging.DEBUG,
+                        'Symlink {} created for {} from {}'
+                        .format(dest, path, pfn))
+        except Exception as e:
+            print("JW,", e)
```
mlassnig: remove?

snafus: Argh - thanks. I removed the print statement (unless you were suggesting the general exception?).

lib/rucio/rse/protocols/posix.py
snafus: I was attempting consistency with (in the Default implementation): https://github.com/rucio/rucio/blob/2c9d77f203fbe6e6ec78b415ccfe457d206b0519/lib/rucio/rse/protocols/posix.py#L77 

shutil.rmtree() does not work with symlinks (afaik).  
os.unlink(dest) might be the simplest option here, (and I've put in this commit)? 

I could replace the 'rm' in the Default class implementation with an shutil.rmtree, if you wanted here?

lib/rucio/rse/protocols/posix.py
snafus: Removed also the generic exception, which should propagate up higher.

mlassnig (approved):

PR #5551, opened 2022-05-13
Deletion: add support for oidc in reaper. Closes #5538 

### LINKED ISSUE #5538: investigate/fix oidc authentication in reaper
Motivation
----------
It was once implemented for reaper, but doesn't seem to work anymore. We assume the code was not migrated in reaper1->2 transition

rcarpa:
This is a very basic support, mostly replicated from how conveyor
does the oidc authentication.

At this stage, I have no way to test it, so ... hope it works... ,
we should work on adding end-to-end authentication tests.

### DIFF
--- a/lib/rucio/daemons/reaper/reaper.py
+++ b/lib/rucio/daemons/reaper/reaper.py
@@ -42,11 +42,13 @@ from rucio.common.exception import (DatabaseException, RSENotFound,
                                     RSEAccessDenied, ResourceTemporaryUnavailable, SourceNotFound,
                                     VONotFound)
 from rucio.common.logging import setup_logging
+from rucio.common.types import InternalAccount
 from rucio.common.utils import chunks
 from rucio.core import monitor
 from rucio.core.credential import get_signed_url
 from rucio.core.heartbeat import list_payload_counts
 from rucio.core.message import add_message
+from rucio.core.oidc import get_token_for_account_operation
 from rucio.core.replica import list_and_mark_unlocked_replicas, list_and_mark_unlocked_replicas_no_temp_table, delete_replicas
 from rucio.core.rse import list_rses, RseData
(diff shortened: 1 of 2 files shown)


rizart:
@rcarpa this might actually work indeed :D 

I have the impression though, that it would be more clear to pass `admin=False` in:
```
get_token_for_account_operation(account, req_audience=oidc_audience, req_scope=oidc_scope, admin=True)
```
It makes sense to aim for this clause:
https://github.com/rucio/rucio/blob/1326078774bb1bd77361788e3dc9fc513f5fc42d/lib/rucio/core/oidc.py#L580-L581

and this can now be achieved with your changes only if we configure `reaper_oidc_account` to point to the Rucio "IAM admin client" account*.

Btw your `admin=True` would still work if (*) is configured as such, but if not configured, then Rucio will try to find an existing token for `reaper_oidc_account`, and if that does not exist, it will fail.

So I think that somehow `admin=False` reduces the complexity and leaves less options for misconfiguration.

rcarpa:
`admin` being True or False has absolutely no influence on reaching and executing this block : https://github.com/rucio/rucio/blob/2a41a64899f41b356660fff80a7530765b5d59e0/lib/rucio/core/oidc.py#L581 

It only influences which of the next 2 blocks is executed if  get_token_for_adminacc is false

rizart:
I agree, I guess my point is that since it does not affect the functionality we need for the reaper, it makes more sense to leave the default `admin=False`. But okay, this will be probably refactored anyway in the future when the token logic will be refined, so no big deal. Kudos for the fast patch!

rcarpa:
I see your point about the default. I'm not sure what is the correct option though. Using user tokens in reaper makes however no sense, so better be explicit about requiring admin token and avoid ever getting into the user token code under any circumstances.

bari12 (approved):

Issue #5422, opened 2022-03-31
Calculation of container `length` and `events`


bari12:
Motivation
----------
For datasets the `length` and `events` field is updated via the `re_evaluate_did` workflow whenever a file is attached/detached from the dataset. However, this information is not communicated up the tree to parent containers (and possibly their parents).

This is not easily do-able in the same workflow, as it opens the risk of row contention.

It raises the question if, for containers, this information should ever be filled in the database, or always kept `NULL`?

Modification
------------
For containers, this information should be calculated dynamically, not necessarily down to the file level, but to the dataset level, since the dataset information is trustworthy. It also needs to be done for closed containers, since the information can potentially still change if the containing datasets are not closed yet.

rcarpa:
The problem seems even more complex. A lot of code relies on DataIdentifierAssociation length/bytes and those are never updated. For Datasets it's not a problem. For Containers it is.

PR #5559, opened 2022-05-16
Dynamically compute container size #5422

rcarpa:
Apply it in two cases: 
- For downloads, we are interested in an exact statistics, so perform the iteration up to the file. 
- When calling "stats" on the container. In this case, only compute it using data from datasets. 

Also add support for temporary tables in get_did when dynamically computing the size.

### DIFF
--- a/lib/rucio/api/did.py
+++ b/lib/rucio/api/did.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 
 from copy import deepcopy
+from typing import TYPE_CHECKING
 
 import rucio.api.permission
 from rucio.common.constants import RESERVED_KEYS
@@ -25,6 +26,10 @@ from rucio.core import did, naming_convention, meta as meta_core
 from rucio.core.rse import get_rse_id
 from rucio.db.sqla.constants import DIDType
 from rucio.db.sqla.session import read_session, stream_session, transactional_session
+
+if TYPE_CHECKING:
+    from typing import Any, Dict, Optional
+    from sqlalchemy.orm import Session
 
 
 @stream_session
@@ -399,13 +404,15 @@ def scope_list(scope, name=None, recursive=False, vo='def', session=None):
 
 
 @read_session
-def get_did(scope, name, dynamic=False, vo='def', session=None):
+def get_did(scope: str, name: str, dynamic_depth: "Optional[DIDType]" = None, vo: str = 'def', session: "Optional[Session]" = None) -> "Dict[str, Any]":
     """
     Retrieve a single data did.
 
     :param scope: The scope name.
     :param name: The data identifier name.
-    :param dynamic:  Dynamically resolve the bytes and length of the did
+    :param dynamic_depth: the DID type to use as source for estimation of this DIDs length/bytes.
+    If set to None, or to a value which doesn't make sense (ex: requesting depth = CONTAINER for a did of type DATASET)
+    will not compute the size dynamically.
     :param vo: The VO to act on.
     :return did: Dictionary containing {'name', 'scope', 'type'}, Exception otherwise
     :param session: The database session in use.
@@ -413,7 +420,7 @@ def get_did(scope, name, dynamic=False, vo='def', session=None):
 
     scope = InternalScope(scope, vo=vo)
 
-    d = did.get_did(scope=scope, name=name, dynamic=dynamic, session=session)
+    d = did.get_did(scope=scope, name=name, dynamic_depth=dynamic_depth, session=session)
     return api_update_return_dict(d, session=session)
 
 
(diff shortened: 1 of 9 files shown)


lib/rucio/web/rest/flaskapi/v1/dids.py
```diff
@@ -575,9 +576,15 @@ def get(self, scope_name):
           schema:
             type: string
           style: simple
+        - name: dynamic_depth
+          in: query:
```
joeldierkes: `in: query`, without the second `:`

lib/rucio/api/did.py
```diff
@@ -401,21 +401,23 @@ def scope_list(scope, name=None, recursive=False, vo='def', session=None):
 
 
 @read_session
-def get_did(scope, name, dynamic=False, vo='def', session=None):
+def get_did(scope, name, dynamic_depth=None, vo='def', session=None):
```
joeldierkes: Typehints?

```python
def get_did(scope: str, name: str, dynamic_depth: Optional[DIDType] = None, vo: str = 'def', session: Optional[Session] = None) -> Dict[str, Any]:
```

Same with every other function signature.

lib/rucio/client/didclient.py
```diff
@@ -400,19 +400,24 @@ def list_files(self, scope, name, long=None):
...
         :param name: The data identifier name.
-        :param dynamic: Calculate sizes dynamically when True
+        :param dynamic_depth: The DID type as string ('FILE'/'DATASET') at which to stop the dynamic
+        length/bytes calculation. If not set, the size will not be computed dynamically.
+        :param dynamic: (Deprecated) same as dynamic_depth = 'FILE'
         """
 
         path = '/'.join([self.DIDS_BASEURL, quote_plus(scope), quote_plus(name)])
+        params = {}
         if dynamic:
```
joeldierkes: The deprecated `dynamic` can override the `dynamic_depth`. From my POV the deprecated one should be the second option to the newer one.

lib/rucio/web/rest/flaskapi/v1/dids.py
```diff
@@ -575,9 +576,15 @@ def get(self, scope_name):
...
+          description: The DID type at which to stop the dynamic length/size estimation
+          schema:
+            type: string
+            enum: ["FILE", "DATASET"]
         - name: dynamic
           in: query
-          description: Flag to dynamically calculate size for open DIDs.
+          description: (Deprecated) Same as dynamic_depth = "FILE"
```
joeldierkes: This is actually a premiere event for us. The `OpenApi` specification has a `deprecated` attribute (https://swagger.io/specification/#parameter-object).
```text
description: Same as dynamic_depth = "FILE".
deprecated: true
```

lib/rucio/web/rest/flaskapi/v1/dids.py
```diff
@@ -654,8 +661,13 @@ def get(self, scope_name):
         """
         try:
             scope, name = parse_scope_name(scope_name, request.environ.get('vo'))
-            dynamic = 'dynamic' in request.args
-            did = get_did(scope=scope, name=name, dynamic=dynamic, vo=request.environ.get('vo'))
+            dynamic_depth = None
+            if 'dynamic_depth' in request.args:
+                orig = request.args['dynamic_depth'].upper()
+                dynamic_depth = DIDType.DATASET if orig == 'DATASET' else DIDType.FILE
```
joeldierkes: This has the side effect that every invalid value gets translated to a `DIDType.File`. (e.g. `request.args['dynamic_depth'] = "Bratwurst"`). Do we want that?

```text
If set to None, or to a value which doesn't make sense (ex: requesting depth = CONTAINER for a did of type DATASET) will not compute the size dynamically.
```

lib/rucio/core/did.py
```diff
@@ -2052,52 +2073,98 @@ def create_did_sample(input_scope, input_name, output_scope, output_name, accoun
 
 
 @transactional_session
-def __resolve_bytes_length_events_did(scope: str, name: str, session: Session) -> Tuple[int, int, int]:
+def __resolve_bytes_length_events_did(
+        did: models.DataIdentifier,
+        dynamic_depth: "DIDType" = DIDType.FILE,
```
joeldierkes, line 2093: `dynamic_depth` default value is sometimes `None`, here `DIDType.File`. We should maybe stick to one.

joeldierkes (dismissed):
Good job, small changes requested! :+1:

lib/rucio/core/did.py
rcarpa: I don't think this is problematic. It's a private function and its name clearly states that we will resolve the size of whatever it's passed in.

bari12 (approved):

PR #5567, opened 2022-05-17
Metadata: Add additional operators in metadata filter engine for Oracle #5503 

### LINKED ISSUE #4512: Using plugin ALL or JSON returns DataIdentifierNotFound if json metadata is not set
Motivation
----------
Using plugin ALL or JSON returns DataIdentifierNotFound if json metadata is not set, e.g.:
```
    meta = get_metadata(did['scope'], did['name'], plugin='ALL', session=session)
  File "/usr/lib/python2.7/site-packages/retrying.py", line 49, in wrapped_f
    return Retrying(*dargs, **dkw).call(f, *args, **kw)
  File "/usr/lib/python2.7/site-packages/retrying.py", line 206, in call
    return attempt.get(self._wrap_exception)
  File "/usr/lib/python2.7/site-packages/retrying.py", line 247, in get
    six.reraise(self.value[0], self.value[1], self.value[2])
  File "/usr/lib/python2.7/site-packages/retrying.py", line 200, in call
    attempt = Attempt(fn(*args, **kwargs), attempt_number, False)
  File "/usr/lib/python2.7/site-packages/rucio/db/sqla/session.py", line 309, in new_funct
  File "/usr/lib/python2.7/site-packages/rucio/core/did.py", line 1255, in get_metadata
    return did_meta_plugins.get_metadata(scope, name, plugin=plugin, session=session)
  File "/usr/lib/python2.7/site-packages/rucio/core/did_meta_plugins/__init__.py", line 74, in get_metadata
    metadata = meta_handler.get_metadata(scope, name, session=session)
  File "/usr/lib/python2.7/site-packages/retrying.py", line 49, in wrapped_f
    return Retrying(*dargs, **dkw).call(f, *args, **kw)
  File "/usr/lib/python2.7/site-packages/retrying.py", line 206, in call
    return attempt.get(self._wrap_exception)
  File "/usr/lib/python2.7/site-packages/retrying.py", line 247, in get
    six.reraise(self.value[0], self.value[1], self.value[2])
  File "/usr/lib/python2.7/site-packages/retrying.py", line 200, in call
    attempt = Attempt(fn(*args, **kwargs), attempt_number, False)
  File "/usr/lib/python2.7/site-packages/rucio/db/sqla/session.py", line 309, in new_funct
  File "/usr/lib/python2.7/site-packages/rucio/core/did_meta_plugins/json_meta.py", line 73, in get_metadata
    raise exception.DataIdentifierNotFound("No generic metadata found for '%(scope)s:%(name)s'" % locals())
rucio.common.exception.DataIdentifierNotFound: Data identifier not found
```

Modification
------------
In case of ALL, the exception should be caught so that at least the column metadata are reported. Regarding the exception raised for the JSON alone, shouldn't we create a new exception ?

### LINKED ISSUE #5503: Add additional operators in metadata filter engine for Oracle
Motivation
----------
The filter engine for metadata search currently supports only the equality operator when Oracle is the backend DB.

Modification
------------
Implement the additional operators.

rizart:
This PR: 
- Adds support for all metadata operations (except the `not like` one which is not supported) when using Oracle.
- Introduces a change into how to manage cases when querying JSON metadata for nonexistent keys. Up until now, when querying for inequality, e.g. `sample_key != 7`, the engine would return even DIDs that don't have the `sample_key` in their json meta dictionary. This PR assumes all keys to be present, even for inequality comparisons, in order to evaluate the clause and return the DID.
- Fixes #4512 

Implementation references: [1] [2]

[1] https://docs.oracle.com/en/database/oracle/oracle-database/19/adjsn/json-path-expressions.html
[2] https://docs.oracle.com/en/database/oracle/oracle-database/19/adjsn/condition-JSON_EXISTS.html

### DIFF
--- a/lib/rucio/core/did_meta_plugins/filter_engine.py
+++ b/lib/rucio/core/did_meta_plugins/filter_engine.py
@@ -397,8 +397,6 @@ class FilterEngine:
                                     expression = oper(json_column[key].as_string(), value)
                             except Exception as e:
                                 raise exception.FilterEngineGenericError(e)
-                            if oper == operator.ne:                                                 # set .ne operator to include NULLs.
-                                expression = or_(expression, json_column[key].is_(None))
                 else:
                     raise exception.FilterEngineGenericError("Requested filter on key without model attribute, but [json_column] not set.")
 
--- a/lib/rucio/core/did_meta_plugins/filter_engine.py
+++ b/lib/rucio/core/did_meta_plugins/filter_engine.py
@@ -38,6 +38,16 @@ OPERATORS_CONVERSION_LUT = {
     "": operator.eq
 }
 
+# lookup table converting pythonic operators to oracle operators
+ORACLE_OP_MAP = {
+    operator.eq: "==",
+    operator.ne: "<>",
+    operator.gt: ">",
+    operator.lt: "<",
+    operator.ge: ">=",
+    operator.le: "<="
+}
+
 # understood date formats.
 VALID_DATE_FORMATS = (
     '%Y-%m-%d %H:%M:%S',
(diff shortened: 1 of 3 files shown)


bari12:
@robbarnsley please review this as well

lib/rucio/core/did_meta_plugins/filter_engine.py
```diff
@@ -344,13 +354,24 @@ def create_sqla_query(self, session=None, additional_model_attributes=[], additi
...
-                            raise exception.FilterEngineGenericError("Oracle implementation does not support this operator.")
-                        if isinstance(value, str) and any([char in value for char in ['*', '%']]):
-                            raise exception.FilterEngineGenericError("Oracle implementation does not support wildcards.")
-                        expression = text("json_exists(meta,'$?(@.{} == \"{}\")')".format(key, value))
+                        if isinstance(value, str) and any([char in value for char in ['*', '%']]):  # wildcards
+                            if value in ('*', '%', u'*', u'%'):                                     # match wildcard exactly == no filtering on key
+                                continue
+                            else:                                                                   # partial match with wildcard == like || notlike
+                                if oper == operator.eq:
+                                    expression = text("json_exists({},'$?(@.{} like \"{}\")')".format(json_column.key, key, value.replace('*', '%')))
```
robbarnsley, line 362: i'm just going to assume this Oracle syntax does what I think it's doing... 😅

lib/rucio/core/did_meta_plugins/filter_engine.py
```diff
@@ -344,13 +354,24 @@ def create_sqla_query(self, session=None, additional_model_attributes=[], additi
...
+                                elif oper == operator.ne:
+                                    raise exception.FilterEngineGenericError("Oracle implementation does not support this operator.")
+                        else:
+                            try:
+                                if isinstance(value, (bool)):                                       # bool must be checked first (as bool subclass of int)
+                                    expression = text("json_exists({},'$?(@.{}.boolean() {} \"{}\")')".format(json_column.key, key, ORACLE_OP_MAP[oper], value))
+                                elif isinstance(value, (int, float)):
+                                    expression = text("json_exists({},'$?(@.{} {} {})')".format(json_column.key, key, ORACLE_OP_MAP[oper], value))
+                                else:
+                                    expression = text("json_exists({},'$?(@.{} {} \"{}\")')".format(json_column.key, key, ORACLE_OP_MAP[oper], value))
```
robbarnsley, line 372: likewise!

lib/rucio/core/did_meta_plugins/filter_engine.py
mlassnig: holy moly! :-D

rizart: Yeah :p the tests look fine, and I also tested it in our ESCAPE instance.
One thing maybe to consider, is the differences of `like`, `like_regex` and `eq_regex` as described in [1].
I think the `like` expression matches best what we want in this case, but please feel free to take a look and let me know if you think some of the others would work better (based on their description).

lib/rucio/core/did_meta_plugins/filter_engine.py
rizart: Here, same for the testing. What is maybe additional functionality here is this `.boolean()` cast which allows values to be any combination of characters that indicate boolean (e.g. `True`, `true`, `tRue`, `tRuE`, `fAlse`, `False`, `false`, etc.)

robbarnsley:
> * Introduces a change into how to manage cases when querying JSON metadata for nonexistent keys. Up until now, when querying for inequality, e.g. `sample_key != 7`, the engine would return even DIDs that don't have the `sample_key` in their json meta dictionary. This PR assumes all keys to be present, even for inequality comparisons, in order to evaluate the clause and return the DID.

Just to add to this - more so I don't forget - this PR will mean that "not equals" will work slightly differently for the column metadata and JSON. I think it makes sense the way it is now, i.e. for JSON metadata, as @rizart points out, if the key doesn't exist, the entry is excluded from the results, but for column metadata, if a clause is included that filters on a column with a NULL value, the row is included in the result set. Consider what you'd expect from [this](http://sqlfiddle.com/#!9/5d71f0/1) as an example.

robbarnsley (approved):

bari12:
👍 I am merging this in as a feature with 1.29 though

bari12:
👍 I am merging this in as a feature with 1.29 though

PR #5569, opened 2022-05-18
Heartbeat handler for transmogrifier : Closes #5568

### LINKED ISSUE #5568: Heartbeat handler for transmogrifier
Motivation
----------
See https://github.com/rucio/rucio/issues/5443

### LINKED ISSUE #5585: Transmogrifier does fails to create rules if the comments field is too long
Motivation
----------

The Transmogrifier may copy over the `comments` field of a subscription into the created rules.

https://github.com/rucio/rucio/blob/86390dfa9e6a99d82a315a2733ba06f6687c4bb4/lib/rucio/daemons/transmogrifier/transmogrifier.py#L289-L291

However, these two columns are defined differently in the database: the former is `String(4000)` while the latter is `String(255)`.  This can result in database errors like `(cx_Oracle.DatabaseError) ORA-12899: value too large for column "ATLAS_RUCIO"."RULES"."COMMENTS" (actual: 267, maximum: 255)`.

Modification
------------

I would suggest that, if necessary, the Transmogrifier should automatically truncated it.

cserf:
Heartbeat handler for transmogrifier : Closes #5568

Implement new heartbeat mechanism

Reformat the transmogrifier code

Transmogrifier does fails to create rules if the comments field is too long : Closes #5585

### DIFF
--- a/lib/rucio/daemons/transmogrifier/transmogrifier.py
+++ b/lib/rucio/daemons/transmogrifier/transmogrifier.py
@@ -315,6 +315,7 @@ def run_once(heartbeat_handler: "HeartbeatHandler", bulk: int, **_kwargs) -> boo
     identifiers = []
     #  Loop over all the new dids
     for did in dids:
+        _, _, logger = heartbeat_handler.live()
         did_success = True
         if did["did_type"] == str(DIDType.DATASET) or did["did_type"] == str(
             DIDType.CONTAINER
--- a/lib/rucio/daemons/transmogrifier/transmogrifier.py
+++ b/lib/rucio/daemons/transmogrifier/transmogrifier.py
@@ -26,11 +26,23 @@ from traceback import format_exception
 
 import rucio.db.sqla.util
 from rucio.db.sqla.constants import DIDType, SubscriptionState
-from rucio.common.exception import (DatabaseException, DataIdentifierNotFound, InvalidReplicationRule, DuplicateRule,
-                                    RSEWriteBlocked, InvalidRSEExpression, InsufficientTargetRSEs,
-                                    InsufficientAccountLimit, InputValidationError, RSEOverQuota,
-                                    ReplicationRuleCreationTemporaryFailed, InvalidRuleWeight,
-                                    StagingAreaRuleRequiresLifetime, SubscriptionWrongParameter, SubscriptionNotFound)
+from rucio.common.exception import (
+    DatabaseException,
+    DataIdentifierNotFound,
+    InvalidReplicationRule,
+    DuplicateRule,
+    RSEWriteBlocked,
+    InvalidRSEExpression,
+    InsufficientTargetRSEs,
+    InsufficientAccountLimit,
+    InputValidationError,
+    RSEOverQuota,
+    ReplicationRuleCreationTemporaryFailed,
+    InvalidRuleWeight,
+    StagingAreaRuleRequiresLifetime,
+    SubscriptionWrongParameter,
+    SubscriptionNotFound,
+)
 from rucio.common.logging import setup_logging
 from rucio.common.schema import validate_schema
 from rucio.common.utils import chunks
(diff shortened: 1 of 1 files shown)


rcarpa (dismissed):
LGTM for the hearbeat handler perspective. 

I saw that it removes a couple of monitoring counters, but they seem to have wrong names, so probably not a big problem. 

Why "must_sleep = False" on exceptions? 

I hate the new formatting. I never liked black formatting (at least keep single-quotes? black has an argument switch for that) :)

lib/rucio/daemons/transmogrifier/transmogrifier.py
```diff
@@ -188,269 +221,463 @@ def transmogrifier(bulk=5, once=False, sleep_time=60):
...
+        )
+        must_sleep = False
+        return must_sleep
+
+    results = {}
+    start_time = time.time()
+    blocklisted_rse_id = [rse["id"] for rse in list_rses({"availability_write": False})]
+    identifiers = []
+    #  Loop over all the new dids
+    for did in dids:
```
rcarpa, line 317: How long can it take to iterate over all dids ? maybe add heartbeat_handler.live() inside the loop ? Hearbeats have a default duration of 10*sleep_time now ... which is probably 10 minutes

cserf:
For some exceptions, we want to restart immediately without sleep

Regarding the `black` formatting, I also don't like everything, but at least it allows to have something consistent and to make the code more readable. I wanted to try on one or two files and ask for opinions.

rcarpa:
Black struggles in this case because there are too many levels of identations inside the function, so it re-fortmats in wired ways just to keep the line length under control. Splitting the transmografier code into functions is probably much better for readability. IMHO

lib/rucio/daemons/transmogrifier/transmogrifier.py
cserf: I'm not sure it is necessary and even if it's a good idea.
Usually, the loop is pretty fast, e.g. :
```
2022-05-23 08:13:02,001 root    1       INFO    transmogrifier[3/13] : It took 1.240969 seconds to process 87 DIDs
2022-05-23 08:13:06,377 root    1       INFO    transmogrifier[4/13] : It took 6.169517 seconds to process 87 DIDs
2022-05-23 08:14:02,394 root    1       INFO    transmogrifier[4/13] : It took 0.878970 seconds to process 156 DIDs
2022-05-23 08:14:02,981 root    1       INFO    transmogrifier[3/13] : It took 1.352124 seconds to process 181 DIDs
2022-05-23 08:14:02,984 root    1       INFO    transmogrifier[2/13] : It took 1.619641 seconds to process 181 DIDs
2022-05-23 08:14:07,596 root    1       INFO    transmogrifier[1/13] : It took 7.792062 seconds to process 170 DIDs
2022-05-23 08:15:01,006 root    1       INFO    transmogrifier[1/13] : It took 0.445081 seconds to process 174 DIDs
2022-05-23 08:15:01,737 root    1       INFO    transmogrifier[4/13] : It took 0.657582 seconds to process 183 DIDs
2022-05-23 08:15:02,201 root    1       INFO    transmogrifier[2/13] : It took 1.030301 seconds to process 167 DIDs
2022-05-23 08:15:06,871 root    1       INFO    transmogrifier[3/13] : It took 5.895660 seconds to process 175 DIDs
```
If we add a hearbeat live for each did, it will increase by 2 orders of magnitude. IMHO, it's a bit too much.

bari12: With the heartbeat_handler it should be fine. This will only fire off an actual DB interaction if the last heartbeat is quite some time ago. So you can actually call this method quite often without any impact.

cserf: Done

rcarpa (approved):
I still don't like the new formatting. But OK for the rest

PR #5579, opened 2022-05-19
Fix various bugs in BB8 : Closes #5571

### LINKED ISSUE #5571: Various bugs affecting BB8
Motivation
----------
Various bugs affecting BB8 were identified

cserf:
Fix various bugs in BB8 : Closes #5571

Add `allowed_scopes` option

Move to new heartbeat handler

Black formatting

### DIFF
--- a/bin/rucio-bb8
+++ b/bin/rucio-bb8
@@ -29,8 +29,8 @@ def get_parser():
     Returns the argparse parser.
     """
     parser = argparse.ArgumentParser(description='The BB8 daemon is responsible for rebalancing data between RSEs.')
-    parser.add_argument('rse', action='store', help='RSE to rebalance. Can be either a RSE or RSE expression.')
-    parser.add_argument('bytes', action='store', type=int, help='Number of bytes to rebalance')
+    parser.add_argument('--rse', action='store', help='RSE to rebalance. Can be either a RSE or RSE expression.')
+    parser.add_argument('--bytes', action='store', type=int, help='Number of bytes to expected to be rebalanced. It is a goal without guarantees')
     parser.add_argument("--run-once", action="store_true", default=False, help='One iteration only')
     parser.add_argument('--sleep-time', action="store", default=30, type=int, help='Concurrency control: thread sleep time after each chunk of work')
     parser.add_argument("--threads", action="store", default=1, type=int, help='Concurrency control: total number of threads for this process')
--- a/bin/rucio-bb8
+++ b/bin/rucio-bb8
@@ -20,6 +20,7 @@ BB8 is a daemon to rebalance data.
 
 import argparse
 
+from rucio.daemons.bb8.bb8 import run
 from rucio.daemons.bb8.common import rebalance_rse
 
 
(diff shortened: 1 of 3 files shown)


bin/rucio-bb8
```diff
@@ -28,8 +29,11 @@ def get_parser():
     Returns the argparse parser.
     """
     parser = argparse.ArgumentParser(description='The BB8 daemon is responsible for rebalancing data between RSEs.')
-    parser.add_argument('rse', action='store', help='RSE to rebalance')
+    parser.add_argument('rse', action='store', help='RSE to rebalance. Can be either a RSE or RSE expression.')
```
mlassnig: should be `--rse` ?

bin/rucio-bb8
```diff
@@ -28,8 +29,11 @@ def get_parser():
...
     parser.add_argument('bytes', action='store', type=int, help='Number of bytes to rebalance')
```
mlassnig: should be `--bytes` ? also, not clear if this a goal without guarantees, or a minimum that will be used ?

lib/rucio/daemons/bb8/common.py
```diff
@@ -112,28 +132,46 @@ def __dump_url(rse_id, logger=logging.log):
...
-        dump_dates = [dump_date.strftime('%d-%m-%Y')]
+        dump_dates = [dump_date.strftime("%d-%m-%Y")]
 
     # getting structure (template) of url location of a dump
-    url_template_str = config_get('bb8', 'dump_url_template', raise_exception=False, default='http://rucio-analytix.cern.ch:8080/LOCKS/GetFileFromHDFS?date=${date}&rse=${rse}')
+    url_template_str = config_get(
+        "bb8",
+        "dump_url_template",
+        raise_exception=False,
+        default="http://rucio-analytix.cern.ch:8080/LOCKS/GetFileFromHDFS?date=${date}&rse=${rse}",
```
mlassnig: is this supposed to be the real URL? because then it's wrong. :-)

cserf: :man_shrugging: That's the one that was used previously

bin/rucio-bb8
cserf: Done

bin/rucio-bb8
cserf: Done

lib/rucio/daemons/bb8/common.py
cserf: Remove default

lib/rucio/daemons/bb8/common.py
```diff
@@ -221,100 +264,239 @@ def list_rebalance_rule_candidates(rse_id, mode=None, session=None):
...
+    query = (
+        session.query(
+            models.DatasetLock.scope,
+            models.DatasetLock.name,
+            models.ReplicationRule.id,
+            models.ReplicationRule.rse_expression,
+            models.ReplicationRule.subscription_id,
+            models.DataIdentifier.bytes,
+            models.DataIdentifier.length,
+            case(
```
rcarpa, line 420: this new formatting makes me want to tear my eyes off

rcarpa (approved):

lib/rucio/daemons/bb8/common.py
mlassnig: I agree at least in this case.... usually I don't mind so much but this introduces way too much verticality. Much easier to understand what's going on if you don't have to scroll.

cserf: That's much better on my 4:3 display :trollface:

bari12: +1 I find that one also extremely unreadable 😢

PR #5580, opened 2022-05-20
Authentication & Authorisation: Further OIDC and SSO optimization and documentation

### LINKED ISSUE #5031: Multiple daemons are impacted when OIDC is configured and the OIDC clients cannot be loaded properly
Motivation
----------
When configuring Rucio to support OIDC, if the OIDC clients cannot be loaded properly for whatever reason, an exception is thrown.
https://github.com/rucio/rucio/blob/86b27bcb485f92759eae25b93b9efb9f7e4103c0/lib/rucio/core/oidc.py#L118
The problem is that multiple files and daemons eventually import functions from oidc.py and thus this piece of code is executed even when OIDC functionality is not really needed right away.
The daemons that are impacted by this behavior include:
- oauthmanager
- conveyor-throttler
- conveyor-stager
- conveyor-submitter

Modification
------------
Initialize the OIDC clients to be loaded when OIDC tokens are actually going to be used, move the `__get_rucio_oidc_clients` function call in a new function that shall be called when the clients are first needed.

### LINKED ISSUE #5437: Further OIDC and SSO optimization and documentation
Motivation
----------

Modification
------------

### LINKED ISSUE #5689: OIDC providers not shown on login selection page
Motivation
----------
The authentication issuers provided in idpsecrets are not shown on the login selection page because the template for that part has not been updated to support the flask templating engine.

Modification
------------
Adapt the template to support flask.

rizart:
This PR:
* Patches the WebUI - OIDC integration
* Enables support for the CERN SSO
* Fixes bug for token refresh functionality
* Fixes bug for token exchange functionality
   Functionality is restored, nevertheless you may see an error which shouldn't be there in the conveyor logs, related issue here: https://github.com/OpenIDC/pyoidc/issues/821
* Introduces additional logging into the OIDC code & cleanup
* Adds partial type annotation into the OIDC code
* When implementing the authorisation code flow, assumes that the parsed `issuer` URL (only `netloc` kept) from the `OAuthRequest` redirect message is a substring of the `issuer` URL as it is specified in the `idpsecrets.json` file. This replaces the previous logic where those two had to be identical. What this allows in practice is having issuers with a path in their URLs (e.g. `https://auth.cern.ch/auth/realms/cern`). Please note that in any case it is assumed that the `issuer` endpoint and the `authorization_endpoint` have the same `netloc`. Direct effect of this is also that different IdPs with the same netloc in the `issuer` URL are not supported.
* Fixes #5031


rizart:
Just identified some weird session cookie behavior, please do not merge yet.

EDIT: https://github.com/rucio/rucio/pull/5580/commits/cffd1797458c40d3580b6828175a01a644b0a6d4 fixes the issue, nevertheless, it needs some reviewing.

bari12:
GTG for me, please also update the refresh_token size in the oracle schema.sql 
[etc/sql/](https://github.com/rucio/rucio/blob/master/etc/sql/oracle/schema.sql)

rizart:
https://github.com/rucio/rucio/pull/5580/commits/5e38f74eb541351611f7c34f55f4e3e71c13b2c6 broke the Oracle tests for some reason :/

joeldierkes:
https://github.com/rucio/rucio/pull/5650 has the same issues in the CI, even though the code shouldn't influence the setup of the container. This seems to be a problem in master. It only affects patch PRs.

joeldierkes:
#5658 is the reference.

rizart:
Thanks @joeldierkes, I thought it was just my (innocent) change that broke this.

bari12 (dismissed):

bari12 (approved):

PR #5582, opened 2022-05-20
Transfers: pass scope to construct_surl. Closes #4836

### LINKED ISSUE #4836: Passing scope to construct_surl()
Motivation
----------

currently lib/rucio/core/transfer.py only passes dataset name and file name to construct_surl() for non-deterministic. Can we also pass the scope to construct_surl()? 

Modification
------------

rcarpa:

### DIFF
--- a/lib/rucio/common/utils.py
+++ b/lib/rucio/common/utils.py
@@ -501,7 +501,7 @@ def my_key_generator(namespace, fn, **kw):
     return generate_key
 
 
-def construct_surl_DQ2(dsn, filename):
+def construct_surl_DQ2(dsn: str, scope: str, filename: str) -> str:
     """
     Defines relative SURL for new replicas. This method
     contains DQ2 convention. To be used for non-deterministic sites.
@@ -540,7 +540,7 @@ def construct_surl_DQ2(dsn, filename):
         return '/%s/%s/%s/%s/%s' % (project, dataset_type, tag, stripped_dsn, filename)
 
 
-def construct_surl_T0(dsn, filename):
+def construct_surl_T0(dsn: str, scope: str, filename: str) -> str:
     """
     Defines relative SURL for new replicas. This method
     contains Tier0 convention. To be used for non-deterministic sites.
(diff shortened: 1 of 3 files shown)


lib/rucio/common/utils.py
```diff
@@ -503,7 +503,7 @@ def generate_key(*arg, **kw):
     return generate_key
 
 
-def construct_surl_DQ2(dsn, filename):
+def construct_surl_DQ2(dsn, scope, filename):
```
joeldierkes: I will create a PR to remove this function since it is not used anymore #5587 . If you want to add this before your vacation, just change it. :+1:

joeldierkes (dismissed):
This PR is the best example for the Missing Type Annotations. 4 new parameters in function definitions, 4 new Missing Type Annotations. :mechanical_arm:
Please add Type Annotations. :wink:

joeldierkes:
EDIT: NVM sav the issue

lib/rucio/common/utils.py
bari12: This is actually still used, so don't remove 😄

joeldierkes: Whoops okay. I did not forget to include it in #5587, it was completely on purpose then! :sweat_smile:

rcarpa: I knew before the merge of that PR that I'll quickly become crazily annoyed by these type annotations ...

joeldierkes: Yes, we all will be! haha
Sorry not sorry :raising_hand_man:

lib/rucio/common/utils.py
```diff
@@ -595,7 +595,7 @@ def register_surl_algorithm(surl_callable, name=None):
 register_surl_algorithm(construct_surl_BelleII, 'BelleII')
 
 
-def construct_surl(dsn, filename, naming_convention=None):
+def construct_surl(dsn, scope, filename, naming_convention=None):
```
joeldierkes: I thought about this a bit and it is not obvious _why_ the scope gets passed here. It does not seem to be used, so someone reading the code (me) could get confused and think, that this is some legacy unused parameter. I'd suggest a comment to clarify, that this parameter may be used in custom implémentations.

rcarpa: What do you mean ? it's being passed to a function a couple of lines below

rcarpa: Done

bari12 (approved):

Issue #5335, opened 2022-03-08
receiver incorrectly handles multi-source transfers


rcarpa:
Motivation
----------

In case of multi-source transfers, receiver analyzes the message even if the job didn't yet finish:
https://github.com/rucio/rucio/blob/8acc9c9b873eef1d7e3b081821665ee03b95f4a0/lib/rucio/daemons/conveyor/receiver.py#L93

If the the message concerns one failed source, this results in a transition of the transfer to a failed state :  
https://github.com/rucio/rucio/blob/8acc9c9b873eef1d7e3b081821665ee03b95f4a0/lib/rucio/transfertool/fts3.py#L523

One may be miss-leaded, me included until now, that the `t_final_transfer_state` field is related to the job_state, but it  doesn't seem to be the case on FTS side. It depends on the file_state and the job state may not yet be final: 
https://gitlab.cern.ch/fts/fts3/-/blob/a34b78420e7f06aebca4f468e39312189e9e1454/src/url-copy/LegacyReporter.cpp#L214

This means that the receiver may transition the file to a failed state even if the job is not yet completed.

bari12:
Backport 1.26.13

PR #5590, opened 2022-05-20
Transfers: always skip ACTIVE transfers in receiver. #5335

rcarpa:

### DIFF
--- a/lib/rucio/daemons/conveyor/receiver.py
+++ b/lib/rucio/daemons/conveyor/receiver.py
@@ -87,9 +87,7 @@ class Receiver(object):
            and 'issuer' in msg['job_metadata'].keys() \
            and str(msg['job_metadata']['issuer']) == str('rucio'):
 
-            if 'job_state' in msg.keys() and (
-                    str(msg['job_state']) != str('ACTIVE')
-                    or str(msg['job_state']) == str('ACTIVE') and 'job_m_replica' in msg.keys() and (str(msg['job_m_replica']).lower() == str('true'))):
+            if 'job_state' in msg.keys() and str(msg['job_state']) != str('ACTIVE'):
 
                 response = {'new_state': None,
                             'transfer_id': msg.get('tr_id').split("__")[-1],


bari12:
Any idea why the test-cases don't run here?

joeldierkes:
Could be a problem with GitHub again (like we had some weeks ago), however, there is no incident in the GitHub Status page.
We changed nothing in the 1.26-LTS pipeline, so I don't think this is on our side.
Maybe try to update the branch or create another PR?

lib/rucio/daemons/conveyor/receiver.py
```diff
@@ -87,9 +87,7 @@ def on_message(self, frame):
            and 'issuer' in msg['job_metadata'].keys() \
            and str(msg['job_metadata']['issuer']) == str('rucio'):
 
-            if 'job_state' in msg.keys() and (
-                    str(msg['job_state']) != str('ACTIVE')
-                    or str(msg['job_state']) == str('ACTIVE') and 'job_m_replica' in msg.keys() and (str(msg['job_m_replica']).lower() == str('true'))):
+            if 'job_state' in msg.keys() and str(msg['job_state']) != str('ACTIVE'):
```
joeldierkes, line 90: This seems oddly written, I'd suggest:

```python
if 'job_state' in msg and msg['job_state'] != 'ACTIVE':
```

- The python docu suggst to use `key in dict` to check if a key is contained in a dictionary (http://docs.python.org/3/tutorial/datastructures.html#dictionaries)
- `'ACTIVE'` is a string already, the cast is redundant
- `msg['job_state']` should always be a string if I'm not mistaken. It makes sense to throw an error if that is not the case so we can investigate that. The cast to a string and comparison afterwards would not throw an error...
  We maybe have to test if the value is `NONE` as a special case, but I'd rather test that explicitly

joeldierkes (changes requested):

lib/rucio/daemons/conveyor/receiver.py
bari12: It should still work as it is. @rcarpa can check this on return, I need to merge it in the release now.

PR #5598, opened 2022-05-24
Documentation: Change the file header of `.py.mako` files #5597

### LINKED ISSUE #5597: Fix header in script.py.mako
Motivation
----------
`rucio/db/sqla/migrate_repo/script.py.mako` is used as template to create alembic revision but contain an old header file that needs to be updated.

joeldierkes:
`.py.mako` files are template files used to generate python files. The
file headers should be treated the same as python ones.


cserf (approved):
LGTM

PR #5602, opened 2022-05-25
Transfers: support for qos #4751

maatthias:
Adding support for QoS: manage pinning replica lifetimes on storage by RSE.

### DIFF
--- a/lib/rucio/core/rule_grouping.py
+++ b/lib/rucio/core/rule_grouping.py
@@ -26,7 +26,7 @@ from rucio.core import account_counter, rse_counter, request as request_core
 import rucio.core.did
 import rucio.core.lock
 import rucio.core.replica
-from rucio.core.rse import get_rse
+from rucio.core.rse import get_rse, get_rse_attribute, get_rse_name
 from rucio.db.sqla import models
 from rucio.db.sqla.constants import LockState, RuleGrouping, ReplicaState, RequestType, DIDType, OBSOLETE
 from rucio.db.sqla.session import transactional_session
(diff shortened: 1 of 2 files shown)


lib/rucio/core/rule_grouping.py
```diff
@@ -842,6 +842,27 @@ def __create_lock_and_replica(file, dataset, rule, rse_id, staging_area, availab
                                                         lifetime=lifetime,
                                                         session=session))
 
+    # If QoS type RSE then set pin to RSE attribute maximum_pin_lifetime
```
bari12: Replace `QoS` with `staging_required`

lib/rucio/core/rule_grouping.py
```diff
@@ -842,6 +842,27 @@ def __create_lock_and_replica(file, dataset, rule, rse_id, staging_area, availab
...
+    staging_required = next(iter(get_rse_attribute('staging_required', rse_id=rse_id, session=session)), False)
+    maximum_pin_lifetime = next(iter(get_rse_attribute('maximum_pin_lifetime', rse_id=rse_id, session=session)), None)
+
+    if staging_required and maximum_pin_lifetime:
```
bari12, line 849: I don't think this is correct, because you are not differentiating between the replica already existing or not. If it does not exist, I think this is correct, if it does, a `STAGEIN` request seems to be more appropriate to me. Due to some luck, this code might work, but I don't think it is conceptually not correct.

lib/rucio/tests/test_conveyor.py
```diff
@@ -738,6 +739,124 @@ def test_stager(rse_factory, did_factory, root_account, replica_client):
...
+    for rse_id in all_rses:
+        rse_core.add_rse_attribute(rse_id, 'fts', TEST_FTS_HOST)
+
+    did = did_factory.upload_test_file(src_rse)
+    replica = replica_core.get_replica(rse_id=src_rse_id, **did)
+
+    replica_client.add_replicas(rse=dst_rse, files=[{'scope': did['scope'].external, 'name': did['name'], 'state': 'C',
+                                                     'bytes': replica['bytes'], 'adler32': replica['adler32'], 'md5': replica['md5']}])
+
+    rule_core.add_rule(dids=[did], account=root_account, copies=1, rse_expression=dst_rse, grouping='ALL', weight=None, lifetime=None, locked=False, subscription_id=None)
```
bari12: I think that test is a bit incomplete.

What you additionally want to test here is that when you create the rule that it is in `REPLICATING` state and the lock corresponding to the replica is `REPLICATING` too. 

Then after the `__wait_for_replica_transfer` call, the lock and rule should be `OK`.

The test for `ReplicaState.AVAILABLE` in line 776 is a bit pointless here, because this replica will always be AVAILABLE.

lib/rucio/tests/test_conveyor.py
```diff
@@ -738,6 +739,124 @@ def test_stager(rse_factory, did_factory, root_account, replica_client):
...
+
+    submitter(once=True, rses=[{'id': rse_id} for rse_id in all_rses], partition_wait_time=None, transfertype='single', filter_transfertool=None)
+
+    replica = __wait_for_replica_transfer(dst_rse_id=dst_rse_id, max_wait_seconds=2 * MAX_POLL_WAIT_SECONDS, **did)
+    assert replica['state'] == ReplicaState.AVAILABLE
+
+
+# @skip_rse_tests_with_accounts
+@pytest.mark.noparallel(reason="runs submitter; poller and finisher")
+def test_qos_transfer_with_existing_disk_replica_and_rule(rse_factory, did_factory, root_account, replica_client):
```
bari12: I don't entirely understand what you are testing here 😄 
The `previous_rse` seems useless, why do you need it?

What we want to test here would be something like the following:
- Replica AVAILABLE on `src_rse` (Somewhere else in the world, NOT the staging_required RSE)
- Create rule1 for `dst_rse` (The staging_required RSE) with long/infinite lifetime (It's an archival rule to put the data on tape)
- Wait until it is complete and the replica is AVAILABLE on `dst_rse` (At this point we say the replica is save on tape)
- Create a rule2 for `dst_rse` (Also the staging_required RSE) to stage the data back into disk (E.g. Panda needs to process this data and wants it on disk)
  - Now make sure that rule1 is still in state `OK` and rule2 is in state `REPLICATING` and their corresponding locks are in the same states. Replica state still `AVAILABLE` 
- Wait for data to be staged in
  - Now rule1 and rule2 should be `OK`

The same test we need in some other versions where you make the stagein of rule2 fail. Now rule2 should be `STUCK` while rule1 is still `OK` while the replica stays in `AVAILABLE` (Normal rucio logic would be that both rule1 and rule2 are in state `STUCK` since they point to the same replica, but here we want the workflow different)

Just to be clear, that test will not work at all with the changes you made until now. But best have the test-case now and then construct the code so you pass the test.

bari12 (changes requested):

PR #5604, opened 2022-05-26
Add the possibility to archive metadata for deleted DIDs : Closes #5484

### LINKED ISSUE #5484: Add the possibility to archive metadata for deleted DIDs
Motivation
----------
Currently when a DID is deleted, it is archived in the `deleted_did` table. The metadata stored in the `dids` are therefore archived, but not the one stored in json. We should have the possibility to choose to archive metadata once a DID is deleted.

cserf:
Add the possibility to archive metadata for deleted DIDs : Closes #5484

This PR introduces the possibility to archive the metadata even when the DID is deleted. It is configurable via an `archive_metadata` option in the `deletion` section. The archival is currently only implemented for the json metadata and for `delete_dids` (`undertaker` workflow). The `reaper` workflow (`delete_replicas`) is a bit more complicated to implement and would require some discussion.


bari12 (commented):
- Rebase please
- Also add table to schema.sql in oracle folder
- I wonder a bit about the `get_metadata_archived` method. Just in the plugin it seems a bit useless, shouldn't we add this also to the interface and later on to the clients etc.?

cserf:
Rebased.
About defining the table in `etc/sql/oracle/schema.sql`, is it needed ? I think this file is only used by ATLAS. Moreover `did_meta` table is not defined in this file.

bari12:
Unfortunately, this needs to be updated further, since the handling of the `session` argument is now done differently...

robbarnsley:
Thanks for this @cserf - we need this functionality!

Could I ask what the status of this is now? What still needs to be done? 

https://github.com/rucio/rucio/blob/169f17a420a575c6a4a3626fd5ceddc3ec79ed39/lib/rucio/core/replica.py#L2082

Do I understand correctly that this PR doesn't connect the delete/archive functionality within the plugins to the actual did deletion call?

bari12:
Continued in #6727

PR #5606, opened 2022-05-31
Replicas: Fix misleading documentation in ReplicaClient.update_replic…

### LINKED ISSUE #5487: Misleading documentation for replicaclient.update_replicas_states()
Motivation
----------

https://github.com/rucio/rucio/blob/47e93cfbe5887071c70de4ba815c1bbdddfac2ce/lib/rucio/client/replicaclient.py#L278-L284

This appears to be a mix between `ReplicaState` and `BadFilesStatus`.

Note that the schemas are also inconsistent. For example:

https://github.com/rucio/rucio/blob/47e93cfbe5887071c70de4ba815c1bbdddfac2ce/lib/rucio/common/schema/atlas.py#L98-L100

Modification
------------

joeldierkes:
…as_state() #5487

The documentation about the available `ReplicaStates` is wrong. It is
mixed with the `BadFilesStatus`. This seems to be a simple mistake, the
`BadFilesStatus` is not used anywhere.

It got introduced in c24ac10ed (Merge pull request #5552 from
rcarpa/patch-5523-circular_reference_containers, 2022-05-30)

### DIFF
--- a/lib/rucio/client/replicaclient.py
+++ b/lib/rucio/client/replicaclient.py
@@ -275,13 +275,12 @@ class ReplicaClient(BaseClient):
         :param files: The list of files. This is a list of DIDs like :
             [{'scope': <scope1>, 'name': <name1>, 'state': <state1>}, {'scope': <scope2>, 'name': <name2>, 'state': <state2>}, ...],
             where a state value can be either of:
-            'A' (available)
-            'S' (suspicious)
-            'U' (unavailable)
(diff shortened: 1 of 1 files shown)


dchristidis (approved):
This looks fine, but I am wondering whether we need some sort of disclaimer to be careful when using this. For example, setting the replica state to `BAD` or `TEMPORARY_UNAVAILABLE` will probably not do what one expects.

PR #5610, opened 2022-06-01
Testing: Drop the type annotations check for client code #4455

joeldierkes:
The type annotations got introduced because we dropped support for
Python2. While this is true on the server side, we still support Python2
on the client one. Since these type annotations are not part of the
Python2 syntax, we can't introduce them yet.

There are two possibilities:

- Use the Python2 compatible type annotations. These are comments with
  the pre-fix "type: ". While the code would be type-annotated, this has
  some major drawbacks:

  - The syntax should be changed back to the Python3 one, to have a
    consistent style.

  - The comments are ugly and delude the code.

- Don't introduce type annotations in the client code till we drop
  support for Python2. This leaves us without type annotated code for a
  short while.

Option two seems reasonable better, since the support will be dropped in
a short amount of time. Once the Python2 support is dropped, we can
activate the missing type annotations count job again for the client.

### DIFF
--- a/tools/count_missing_type_annotations_utils.sh
+++ b/tools/count_missing_type_annotations_utils.sh
@@ -46,7 +46,7 @@ create_missing_python_type_annotations_report() {
     # This does not include tests, tools, database and bin files.
     # :param $1: The name of the output file.
 
-    flake8 lib/rucio --exclude tools,lib/rucio/tests,lib/rucio/db --output-file $1 --select ANN || true
+    flake8 lib/rucio --exclude tools,lib/rucio/tests,lib/rucio/db,lib/rucio/client,lib/rucio/common,lib/rucio/rse,bin --output-file $1 --select ANN || true
 }
 
 


joeldierkes:
I _think_ this is the only code shared with the client, please tell me if I'm wrong.

tools/count_missing_type_annotations_utils.sh
```diff
@@ -46,7 +46,7 @@ create_missing_python_type_annotations_report() {
     # This does not include tests, tools, database and bin files.
     # :param $1: The name of the output file.
 
-    flake8 lib/rucio --exclude tools,lib/rucio/tests,lib/rucio/db --output-file $1 --select ANN || true
+    flake8 lib/rucio --exclude tools,lib/rucio/tests,lib/rucio/db,lib/rucio/client,lib/rucio/common --output-file $1 --select ANN || true
```
maany: https://github.com/rucio/rucio/blob/730ea648608292cff6cd69afb10113090a5064f2/setup_rucio_client.py#L38
I see a `lib/rucio` in the setup script :O. Let's discuss tomorrow

joeldierkes: I added the `lib/rucio/rse` folder as well. Yeah, let's talk about the `lib/rucio` one.

joeldierkes: Summary after the talk: Although the `rucio` module is listed as dependency of the client, not everything is used (e.g. `core`). Since not imported code does not get interpreted, it is fine to add type annotations to these unused parts. Our tests would capture all type annotated places, which is used by the client code.

bari12 (approved):

Issue #5612, opened 2022-06-01
Fix get_did_from_pfns for non-ATLAS policy


cserf:
Motivation
----------
`get_did_from_pfns` is currently only working using ATLAS naming scheme (hash). It needs to be changed to handle other conventions

bari12:
Use algorithms defined in #6817

PR #5613, opened 2022-06-02
Fix unit tests for non-ATLAS policies : Closes #5605

### LINKED ISSUE #5605: Fix VO specific tests
Motivation
----------
A  lot of tests are still disabled for VO-specific tests because of wrong naming convention for DIDs or RSEs. They need to be fixed

cserf:
Fix unit tests for non-ATLAS policies : Closes #5605

This PR starts fixing the unit tests that are failing for non-ATLAS policies (in particular belle II). Since many tests need a fix, only a few tests are fixed here (listed below):
```
 lib/rucio/tests/test_abacus_account.py
 lib/rucio/tests/test_abacus_collection_replica.py
 lib/rucio/tests/test_abacus_rse.py
 lib/rucio/tests/test_account.py
 lib/rucio/tests/test_api_external_representation.py
 lib/rucio/tests/test_did.py
 lib/rucio/tests/test_dirac.py
 lib/rucio/tests/test_permission.py
 lib/rucio/tests/test_reaper.py
 lib/rucio/tests/test_replica.py
 lib/rucio/tests/test_rse.py
 lib/rucio/tests/test_rule.py
 lib/rucio/tests/test_subscription.py
 lib/rucio/tests/test_undertaker.py
```

Some tests cannot be fixed because some of the features that they are testing only works for ATLAS. In that case they were disabled and an issue was open to fix the feature to make it work with other policies (e.g. https://github.com/rucio/rucio/issues/5612)


lib/rucio/tests/common.py
```diff
@@ -66,13 +67,15 @@ def scope_name_generator():
...
-def did_name_generator(did_type='file'):
+def did_name_generator(did_type='file', name_prefix='', name_suffix='', cnt=0):
     """ Generate random did name.
 
     :returns: A random did name
     """
     if os.getenv('POLICY') == 'belleii':
-        path = '/belle'
+        path = '/belle/mock' + name_prefix
+        if not path[-1] != '/':
```
maany: umm, double negation?

lib/rucio/tests/common.py
```diff
@@ -66,13 +67,15 @@ def scope_name_generator():
     return 'mock_' + str(uuid()).lower()[:16]
 
 
-def did_name_generator(did_type='file'):
+def did_name_generator(did_type='file', name_prefix='', name_suffix='', cnt=0):
```
maany: @cserf could you also update the docstring, thanks  :)

lib/rucio/tests/temp_factories.py
```diff
@@ -274,8 +275,9 @@ def _random_did(self, scope, name_prefix, name_suffix=''):
         scope = self._sanitize_or_set_scope(scope)
         if not name_prefix:
             name_prefix = 'lfn'
-        name = '%s_%s_%s%s' % (name_prefix, self.base_uuid, len(self.created_dids), name_suffix)
+        name = did_name_generator(did_type=name_prefix, name_prefix='%s_%s' % (name_prefix, self.base_uuid), name_suffix=name_suffix, cnt=len(self.created_dids))
         did = {'scope': scope, 'name': name}
+        print(did)
```
maany: post-debug cleanup? :P

lib/rucio/tests/temp_factories.py
```diff
@@ -285,6 +287,7 @@ def random_did(self, scope=None, name_prefix=None, name_suffix=''):
 
     def make_dataset(self, scope=None):
         did = self._random_did(scope=scope, name_prefix='dataset')
+        print(did)
```
maany: same as above

lib/rucio/tests/common.py
```diff
@@ -81,9 +84,10 @@ def did_name_generator(did_type='file'):
             return path
         path += '/file_%s' % str(uuid())
         return path
-    if did_type in ['file', 'dataset', 'container']:
-        return '%s_%s' % (did_type, str(uuid()))
-    return 'mock_' + str(uuid())
+    if name_prefix:
+        name = '%s_%s%s' % (name_prefix, cnt, name_suffix)
+        return name
+    return '%s_%s' % (did_type, str(uuid()))
```
maany, line 97: Could we add a path parameter to this method such as:
```
def  did_name_generator(path=None, did_type='file', name_prefix='', name_suffix='', cnt=0):
 if path is not None:
      return os.path.basename(path)
```
That way, we could also have some connection between a file_path and the file_name. Might save someone some debugging time in the future :P

lib/rucio/tests/temp_factories.py
```diff
@@ -318,8 +322,9 @@ def upload_test_dataset(self, rse_name, scope=None, size=2, nb_files=2):
         self.created_dids.append(did)
         items = list()
         for _ in range(0, nb_files):
+            # TODO : Call did_name_generator
             path = file_generator(size=size)
-            name = os.path.basename(path)
```
maany, line 321: please see the comment above regarding adding the os.path.basename(path) inside the did_name_generator method :)

maany (changes requested):
Thanks @cserf !! It's awesome to see the tests moving to the factories. I have left some comments, let me know what you think 👍

lib/rucio/tests/temp_factories.py
cserf: Done

lib/rucio/tests/common.py
```diff
@@ -66,24 +67,36 @@ def scope_name_generator():
...
+        path += 'cont_%s' % str(uuid())
         if did_type == 'container':
             return path
-        path += '/dataset_%s' % str(uuid())
+        path += 'dataset_%s' % str(uuid())
         if did_type == 'dataset':
             return path
-        path += '/file_%s' % str(uuid())
+        path += 'file_%s' % str(uuid())
         return path
```
joeldierkes: Handling paths by string manipulation is error-prone and hard to read, how about using the `os.path` module?

This should translate to something like:
```Python
if path is not None:
    return path

container_path = os.path.join("/belle", "mock", name_prefix, 'cont_%s' % str(uuid()))
if did_type == 'container':
    return container_path

dataset_path = os.path.join(container_path, 'dataset_%s' % str(uuid()))
if did_type == 'dataset':
    return dataset_path

file_path = os.path.join(dataset_path, 'dataset_%s' % str(uuid()))
return file_path
```

Or something like that?

joeldierkes (dismissed):
Just a small nit, else gtg

lib/rucio/tests/common.py
cserf: Good suggestion. Implemented.

cserf:
Up

maany (approved):

PR #5621, opened 2022-06-07
Improvements of automatix daemon : Closes #5616, #5617

### LINKED ISSUE #5616: Use uploadclient in automatix
Motivation
----------
A specific `upload` method was implemented in `automatix` before `uploadclient` was developed. It is not needed anymore, and `uploadclient` should be used in the daemon now.

### LINKED ISSUE #5617: Heartbeat handler for automatix daemon
Motivation
----------
Heartbeat handler for automatix daemon (see https://github.com/rucio/rucio/issues/5562)

cserf:
Improvements of automatix daemon :
- Implementation of heartbeat handler : Closes #5617 
- Use uploadclient instead of upload method previously implemented in automatix : Closes #5616 
- Add unit tests

### DIFF
--- a/lib/rucio/daemons/automatix/automatix.py
+++ b/lib/rucio/daemons/automatix/automatix.py
@@ -13,34 +13,35 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import functools
 import logging
 import random
-import socket
 import tempfile
 import threading
 from configparser import NoOptionError, NoSectionError
 from datetime import datetime
 from json import load
-from os import remove, rmdir, getpid
-from time import sleep, time
+from os import remove, rmdir
+import time
+from typing import TYPE_CHECKING
 
 import rucio.db.sqla.util
 from rucio.client import Client
 from rucio.client.uploadclient import UploadClient
 from rucio.common import exception
-from rucio.common.config import config_get
-from rucio.common.logging import formatted_logger, setup_logging
+from rucio.common.config import config_get, config_get_int, config_get_bool
+from rucio.common.logging import setup_logging
 from rucio.common.types import InternalScope
-from rucio.common.utils import daemon_sleep
 from rucio.common.utils import execute, generate_uuid
-from rucio.core import monitor, heartbeat
+from rucio.core import monitor
 from rucio.core.scope import list_scopes
-
-SUCCESS = 0
-FAILURE = 1
+from rucio.daemons.common import run_daemon
 
 
-GRACEFUL_STOP = threading.Event()
+if TYPE_CHECKING:
+    from rucio.daemons.common import HeartbeatHandler
+
+graceful_stop = threading.Event()
 
 
 def get_data_distribution(inputfile: str):
@@ -49,7 +50,7 @@ def get_data_distribution(inputfile: str):
     probabilities = {}
     probability = 0
     for key in data:
-        probability += data[key]['probability']
+        probability += data[key]["probability"]
         probabilities[key] = probability
     for key in probabilities:
         probabilities[key] = float(probabilities[key]) / probability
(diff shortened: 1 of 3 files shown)


rcarpa:
This is a nightmare to review. Please don't mix multiple unrelated changes in the same commit anymore. Even if I spend an hour trying to understand every single change, I'm still sure I'll miss a bug if there is any. 

Moving code around should ideally go into one commit. And adding new functionality to another commit. And, even if the previous ones are mixed, applying code formater should _definitely_! go into a separate commit.

lib/rucio/daemons/automatix/automatix.py
```diff
@@ -210,58 +139,61 @@ def automatix(sites, inputfile, sleep_time, account, worker_number=1, total_work
         logger(logging.INFO, 'Getting data distribution')
         probabilities, data = get_data_distribution(inputfile)
         logger(logging.DEBUG, 'Probabilities %s', probabilities)
-        totretries = 3
```
rcarpa, line 211: why do we remove the retry mechanism? There is no explaination about it in the commit message.

lib/rucio/daemons/automatix/automatix.py
```diff
@@ -280,9 +212,15 @@ def run(total_workers=1, once=False, inputfile=None, sleep_time=-1):
         raise exception.DatabaseException('Database was not updated, daemon won\'t start')
 
     try:
-        sites = [s.strip() for s in config_get('automatix', 'sites').split(',')]
+        rses = [s.strip() for s in config_get('automatix', 'rses').split(',')]  # TODO use config_get_list
```
rcarpa: I don't understand why this change is bundled into the commit "Use uploadclient in automatix" . how is it related to the uploadclient usage ? Why do we need this two-level 'rses', then 'sites' config parsing? Is it a new functionality to prefer the "rses" naming over "sites". If yes, why doesn't it come in a separate PR (or, at least, commit) ?

lib/rucio/daemons/automatix/automatix.py
```diff
@@ -68,210 +69,216 @@ def choose_element(probabilities: dict, data: str) -> float:
...
-        if once is True:
-            logger(logging.INFO, 'Run with once mode. Exiting')
-            break
-        tottime = time() - starttime
-        if status:
-            logger(logging.INFO, 'It took %s seconds to upload one dataset on %s', str(tottime), str(rses))
-            daemon_sleep(start_time=starttime, sleep_time=sleep_time, graceful_stop=GRACEFUL_STOP, logger=logger)
+def run_once(heartbeat_handler: "HeartbeatHandler", inputfile: str, **_kwargs) -> bool:
+
+    _, _, logger = heartbeat_handler.live()
```
rcarpa, line 134: Why move this code from run() to run_once() in a commit called "Heartbeat handler for automatix daemon". If this is needed for this commit, it must be explained in the commit message. If it's unrelated, it must be in a different commit.

lib/rucio/daemons/automatix/automatix.py
```diff
@@ -68,210 +69,216 @@ def choose_element(probabilities: dict, data: str) -> float:
...
+            remove(physical_fname)
+        rmdir(tmpdir)
+    tottime = time.time() - starttime
+    logger(
+        logging.INFO,
+        "It took %s seconds to upload one dataset on %s",
+        str(tottime),
+        str(rses),
+    )
+    return True
```
rcarpa, line 248: feels like it will always sleep now, which wasn't the case before. Am I wrong?

lib/rucio/tests/test_automatix.py
```diff
@@ -81,3 +84,5 @@ def test_automatix(vo, root_account, rse_factory):
         assert status
         for file_ in files:
             assert file_dict["%s:%s" % (file_["scope"], file_["name"])] is True
+    config_remove_option("automatix", "rses")
```
rcarpa, line 87: did you try using the `file_config_mock` fixture for this ? If the automatix test doesn't interact using the API with the rest server, it should be a better solution than marking the test noparallel.

cserf: There are 2 issues here :
- Some options depends on variables (e.g. rse coming from rse_factory) that are generated by the test itself
- Other options are conditional, based on environment variable (POLICY) and I don't think that the `file_config_mock` allows to do it

lib/rucio/daemons/automatix/automatix.py
cserf: I checked during the last month and this retry mechanism was never used. That's why it was removed

lib/rucio/daemons/automatix/automatix.py
cserf: Yes, that's new functionality that should have gone into an independent commit

lib/rucio/daemons/automatix/automatix.py
cserf: Main reason is that the config was resolved once at the start of the daemon, but then never refreshed

lib/rucio/daemons/automatix/automatix.py
```diff
@@ -13,134 +13,52 @@
...
 import tempfile
 import threading
 from configparser import NoOptionError, NoSectionError
 from datetime import datetime
 from json import load
-from math import exp
-from os import remove, rmdir, stat, getpid
-from time import sleep, time
+from os import remove, rmdir
+import time
```
bari12: Move import up

bari12 (commented):
Other than Radus comments, One minor change + rebase please.

bari12 (approved):

PR #5623, opened 2022-06-07
Delete where file does not exist is OK

### LINKED ISSUE #5622: Check before deleting with GFAL
Motivation
----------
This is a patch CMS has been running with for years. I'd like to see it merged if possible. Basically just check that a file exists before trying to delete it.

Modification
------------

See PR

ericvaandering:
Fix #5622 check that a file exists before trying to delete it with GFAL.

CMS has been running with this for years, it's time to try to get it into the release. @nsmith- is the original author

### DIFF
--- a/lib/rucio/rse/protocols/gfal.py
+++ b/lib/rucio/rse/protocols/gfal.py
@@ -491,9 +491,12 @@ class Default(protocol.RSEProtocol):
 
         try:
             for path in paths:
-                ret = ctx.unlink(str(path))
-                if ret:
-                    return ret
+                if self.__gfal2_exist(path) == 0:
+                    ret = ctx.unlink(str(path))
+                    if ret:
+                        return ret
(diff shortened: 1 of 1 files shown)


mlassnig (approved):

lib/rucio/rse/protocols/gfal.py
```diff
@@ -491,9 +491,12 @@ def __gfal2_rm(self, paths):
 
         try:
             for path in paths:
-                ret = ctx.unlink(str(path))
-                if ret:
-                    return ret
+                if self.__gfal2_exist(path) == 0:
```
nsmith-, line 494: Do we know why this use of `path` is not wrapped with `str`?

nsmith-: Ah because it is a function in the same class, nvm

PR #5627, opened 2022-06-08
Generate a keyboard interrupt and thus die 

ericvaandering:
if kronos cannot find the message brokers

This is in production and working for CMS. Otherwise with 1.28 we have kronos just hanging forever


rcarpa (commented):
Is it complex to "gracefully" recover from the issue ? 

Like retrying connections for example. 

If yes, did you try just setting graceful_stop and letting the normal termination path to be used? This would correctly cleanup hearbeats, while your code probably doesn't.

bari12:
Ping @ericvaandering ?

PR #5641, opened 2022-06-13
Monitoring & Logging: fix transfer_link sent by receiver. Closes #5640

### LINKED ISSUE #5640: Incorrect FTS transfer_link on messages sent by receiver
Motivation
----------

The transfer_link is generated here: 
https://github.com/rucio/rucio/blob/07fb2776928b7b030d570bd1f911d2ce5c43f30a/lib/rucio/transfertool/fts3.py#L396

Which uses external_host as passed into the constructor: 
https://github.com/rucio/rucio/blob/876904d4dab98386cd6e6ea6513d63ab03adae1f/lib/rucio/daemons/conveyor/receiver.py#L83

This field comes from an FTS config file parameter, which may not be initialized the same way as rucio expects it to be. As a result, rucio will generate monitoring links which don't point to a correct fts fqdn. 

Modification
------------
Under normal operation, when we get a notification from FTS about a transfer completion, we know about the transfer and we retrieve it from the database. It doesn't cost us anything to use the request.external_host field from the database, which has the correct format. 

We can ignore the ocurances when it's not the case. It means that we got a notification from FTS about a transfer submitted by rucio, but about which rucio doesn't know anything.... we would have bigger problems to investigate in such case than a wrong monitoring link :D

rcarpa:
Under normal operation, when we get a notification from FTS about a
transfer completion, we know about the transfer and retrieve it
from the database. It doesn't cost us anything to use the
request.external_host field from the db table, which has the correct
format.

We can ignore the occurances when it's not the case. It means that we
got a notification from FTS about a transfer submitted by rucio, but
about which rucio doesn't know anything.... we would have bigger
problems to investigate in such case than a wrong monitoring link.

### DIFF
--- a/lib/rucio/tests/test_conveyor.py
+++ b/lib/rucio/tests/test_conveyor.py
@@ -480,6 +480,7 @@ def test_multisource_receiver(vo, did_factory, replica_client, root_account, met
         assert msg_done['payload']['datatype'] == 'RAW'
         assert msg_done['payload']['datasetScope'] == dataset['scope'].external
         assert msg_done['payload']['dataset'] == dataset['name']
+        assert msg_done['payload']['transfer_link'].startswith('https://fts:8449/')
     finally:
         receiver_graceful_stop.set()
         receiver_thread.join(timeout=5)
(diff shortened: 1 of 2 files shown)


rcarpa:
Requires #5645 to be merged first dependency

rcarpa:
I added a test and then had to fight because the refactorings in tests was not python 2 compatible. 

It should be good now, but it is built on top of #5645 now. So skip that commit on review

bari12 (dismissed):

bari12 (approved):

PR #5645, opened 2022-06-14
Tests: support mocking Message table

rcarpa:
It allows to more easily verify the content of messages generated
during a test.

Re-use the same mechanism which was used for mocking the config table.
Move the common code to a generic function allowing to create any
in-memory table and use that code in both places.

Had to re-work the registration in the sessionmaker: it is now
possible to have both a Message and a Config table mock used in
parallel and this was not supported by previous code.

### DIFF
--- a/lib/rucio/tests/conftest.py
+++ b/lib/rucio/tests/conftest.py
@@ -261,6 +261,77 @@ def __get_fixture_param(request):
     return fixture_param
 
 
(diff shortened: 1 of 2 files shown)


rcarpa:
Had to fight with python 2 :D should be ok now

bari12 (approved):

PR #5646, opened 2022-06-14
Monitoring & Logging: additional monitoring fields. Closes #5521. Closes #4557

### LINKED ISSUE #4557: Add DID data type to transfer and deletion events
Motivation
----------

This will improve our monitoring capabilities, allowing for better understanding of data flows.

Modification
------------

Enrich the transfer and deletion events with the DID data type. It’s fine if it’s unset or some communities do not use it at all. We should carefully evaluate the performance cost.

### LINKED ISSUE #5521: Propagate dataset metadata to transfer
Motivation
----------

```
Dear rucio devs,

I am making RAW data to DAOD_PHYS plot (https://its.cern.ch/jira/browse/ADCMONITOR-507). There, I need to plot duration of RAW/AOD data transfer to Tier-1s. This is doable using data in rucio is sending to the monit infrastructure, namely data.dataset variable. But it seems that T0 export data do not have that field defined.

example:

https://rucio-ui.cern.ch/did?scope=data17_13TeV&name=AOD.28873750._000372.pool.root.1

is part of data17_13TeV:data17_13TeV.00340030.calibration_BphysPEB.merge.AOD.r13576_p5087_tid28873750_00

but in the monit data, there is no data.dataset variable filled:

https://es-monit-st.cern.ch/kibana/app/kibana#/discover?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:'2022-05-02T00:00:00.000Z',to:'2022-05-02T23:59:59.999Z'))&_a=(columns:!(_source),filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!f,index:d2660fe0-d0ba-11ea-a34a-f59641c8979f,key:data.activity,negate:!f,params:(query:'T0%20Export'),type:phrase,value:'T0%20Export'),query:(match:(data.activity:(query:'T0%20Export',type:phrase)))),('$state':(store:appState),meta:(alias:!n,disabled:!f,index:d2660fe0-d0ba-11ea-a34a-f59641c8979f,key:data.name,negate:!f,params:(query:AOD.28873750._000372.pool.root.1),type:phrase,value:AOD.28873750._000372.pool.root.1),query:(match:(data.name:(query:AOD.28873750._000372.pool.root.1,type:phrase))))),index:d2660fe0-d0ba-11ea-a34a-f59641c8979f,interval:auto,query:(language:kuery,query:''),sort:!(metadata.timestamp,desc))

Would it be possible to send dataset info for all files?


thanks

Michal 
```

rcarpa:

### DIFF
--- a/lib/rucio/core/request.py
+++ b/lib/rucio/core/request.py
@@ -1554,6 +1554,8 @@ def add_monitor_message(new_state, request, additional_fields, session=None):
                'protocol': None,
                'scope': request['scope'],
                'name': request['name'],
+               'dataset': None,
+               'datasetScope': None,
                'src-type': None,
                'src-rse': request.get('source_rse', None),
                'src-url': None,
@@ -1577,6 +1579,12 @@ def add_monitor_message(new_state, request, additional_fields, session=None):
 
     if message['started_at'] and message['transferred_at']:
         message['duration'] = (message['transferred_at'] - message['started_at']).seconds
+    ds_scope = request['attributes'].get('ds_scope')
+    if not message['datasetScope'] and ds_scope:
+        message['datasetScope'] = ds_scope
+    ds_name = request['attributes'].get('ds_name')
+    if not message['dataset'] and ds_name:
+        message['dataset'] = ds_name
     if not message.get('protocol'):
         dst_url = message['dst-url']
         if dst_url and ':' in dst_url:
--- a/lib/rucio/core/request.py
+++ b/lib/rucio/core/request.py
@@ -1533,6 +1533,14 @@ def add_monitor_message(new_state, request, additional_fields, session=None):
         transfer_status = 'transfer-%s' % new_state.name
     transfer_status = transfer_status.lower()
 
+    stmt = select(
+        models.DataIdentifier.datatype
+    ).where(
+        models.DataIdentifier.scope == request['scope'],
+        models.DataIdentifier.name == request['name'],
+    )
+    datatype = session.execute(stmt).scalar_one_or_none()
+
     # Start by filling up fields from database request or with defaults.
     message = {'activity': request.get('activity', None),
                'request-id': request['id'],
@@ -1561,7 +1569,8 @@ def add_monitor_message(new_state, request, additional_fields, session=None):
                'started_at': request.get('started_at', None),
                'transferred_at': request.get('transferred_at', None),
                'tool-id': 'rucio-conveyor',
-               'account': request.get('account', None)}
+               'account': request.get('account', None),
+               'datatype': datatype}
 
     # Add (or override) existing fields
     message.update(additional_fields)
--- a/lib/rucio/core/transfer.py
+++ b/lib/rucio/core/transfer.py
@@ -24,7 +24,7 @@ from typing import TYPE_CHECKING
 
 from dogpile.cache import make_region
 from dogpile.cache.api import NoValue
-from sqlalchemy import update
+from sqlalchemy import select, update
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.sql.expression import false
 
@@ -439,6 +439,14 @@ def set_transfers_state(transfers, state, submitted_at, external_host, external_
             if rowcount == 0:
                 raise RucioException("%s: failed to set transfer state: request doesn't exist or is not in SUBMITTING state" % rws)
 
+            stmt = select(
+                models.DataIdentifier.datatype
+            ).where(
+                models.DataIdentifier.scope == rws.scope,
+                models.DataIdentifier.name == rws.name,
+            )
+            datatype = session.execute(stmt).scalar_one_or_none()
+
             msg = {'request-id': rws.request_id,
                    'request-type': rws.request_type,
                    'scope': rws.scope.external,
@@ -455,7 +463,8 @@ def set_transfers_state(transfers, state, submitted_at, external_host, external_
                    'checksum-adler': rws.adler32,
                    'external-id': external_id,
                    'external-host': external_host,
-                   'queued_at': str(submitted_at)}
+                   'queued_at': str(submitted_at),
+                   'datatype': datatype}
             if rws.scope.vo != 'def':
                 msg['vo'] = rws.scope.vo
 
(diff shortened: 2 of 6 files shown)


rcarpa:
The first commit is from #5645

lib/rucio/core/transfer.py
```diff
@@ -441,10 +441,20 @@ def set_transfers_state(transfers, state, submitted_at, external_host, external_
...
+                models.DataIdentifier.name == rws.name,
+            )
+            datatype = session.execute(stmt).scalar_one_or_none()
+
             msg = {'request-id': rws.request_id,
                    'request-type': rws.request_type,
                    'scope': rws.scope.external,
                    'name': rws.name,
+                   'dataset': None,
+                   'datasetScope': None,
```
mlassnig, line 455: should these two be `dataset-name` and `dataset-scope` instead?

mlassnig (commented):
so, this looks correct to me semantically.

the question is, does the additional join impact the performance noticeable. it should not, as it's a PK lookup, but who knows. we should at least keep an eye on reaper performance

@bari12 second pair of eyes pls

lib/rucio/core/transfer.py
rcarpa: If we follow the ~~technical debt~~ convention used in other parts of rucio, it must the way I put it. See here for example: https://github.com/rucio/rucio/blob/2a41a64899f41b356660fff80a7530765b5d59e0/lib/rucio/core/trace.py#L174

lib/rucio/core/request.py
```diff
@@ -1534,6 +1534,14 @@ def add_monitor_message(new_state, request, additional_fields, session=None):
...
     transfer_status = transfer_status.lower()
 
+    stmt = select(
+        models.DataIdentifier.datatype
+    ).where(
+        models.DataIdentifier.scope == request['scope'],
+        models.DataIdentifier.name == request['name'],
+    )
+    datatype = session.execute(stmt).scalar_one_or_none()
+
```
joeldierkes, line 1543: How about a `get_datatype_from_did_table` function? This function here is quite sophisticated/a monster already.

lib/rucio/core/transfer.py
```diff
@@ -441,6 +441,14 @@ def set_transfers_state(transfers, state, submitted_at, external_host, external_
...
                 raise RucioException("%s: failed to set transfer state: request doesn't exist or is not in SUBMITTING state" % rws)
 
+            stmt = select(
+                models.DataIdentifier.datatype
+            ).where(
+                models.DataIdentifier.scope == rws.scope,
+                models.DataIdentifier.name == rws.name,
+            )
+            datatype = session.execute(stmt).scalar_one_or_none()
+
```
joeldierkes, line 449: Same as above

lib/rucio/tests/conftest.py
```diff
@@ -261,6 +261,77 @@ def __get_fixture_param(request):
     return fixture_param
 
 
+def __create_in_memory_db_table(name, *columns, **kwargs):
```
joeldierkes: Double underscore is used to avoid Name Mangling (https://github.com/rucio/rucio/pull/5673#discussion_r907254297). This should be a single one.

joeldierkes (changes requested):

lib/rucio/core/transfer.py
mlassnig: ouch

lib/rucio/tests/conftest.py
rcarpa: This one was part of a commit from another PR which is already merged.

lib/rucio/core/request.py
rcarpa: I'm not sure this is a good idea. I cannot put `get_datatype_from_did_table` into a common place (dids.py) for reasons I explain in the commit message. The function is not complicated at all (don't know why you say it's a monster). I don't think it's justified in this particular case to move the query to a separate function which nobody else will use except this code.

joeldierkes: While this function "only" creates a message for hermes, it does this with a lot of logic
- It creates the `transfer-status` based on the request type and the new_state name
  - It makes the `transfer-status` lower case
- It selects the data-type (or None) from the DIDs table, based on the defined primary-key
- It constructs the message with 31 fields, some of them are None for any reason apperently
- It then updates the default message values with some additional fields, or adds them if they dont exist
- It then sets some of the values, which are by default None, to a value based on logic which is implemented with 7 if statements and one for loop
- And it adds the message in the end

I think we mix several use-cases together:
- Creating and adding a message:
  - These are two different things, so why don't we have a `create_hermes_message` function?
- Creating the `transfer_status`:
  - This has nothing to do with the message creation, just with the adding a message. So why is there no `get_hermes_transfer_status`?

From my POV, the `add_monitor_message` should look like:
```python
def get_hermes_transfer_status(new_state: str, request_type: ... = None) -> str:
    if request_type:
        transfer_status = '%s-%s' % (request_type.name, new_state)
    else:
        transfer_status = 'transfer-%s' % new_state
    return transfer_status.lower()


def create_hermes_message(...):
    pass


@read_session
def add_monitor_message(new_state, request, additional_fields, session=None):
    transfer_status = get_hermes_transfer_status(new_state, request['request_type'])
    message = create_hermes_message(request, additional_fields, session=session)  # No new_state

    add_message(transfer_status, message, session=session)
```

And now we can test the creation of the transfer_status, the creation of the message and the add message independently. If there is an error in the message, I only look into `create_hermes_message`, not in `get_hermes_transfer_status`. In the current _petit_ monster, this would've cluttered my mind too.

To the `get_datatype_from_did_table`: You can add the function in this file and another one in the other file. It is not about code duplication (in this case), but readability. If I have to change a field in the message not related to the `datatype`, I don't want to read the code to get the `datatype`. A function with a good name therefore allows me to skip this part, and look into it if it's important.

rcarpa: The "improvement" you propose is very subjective and I don't agree that it makes things better. I consider it complicates the flow of understanding the function. This is a very simple function without loops and without complex state changes which does only one job: correctly formatting a "message" dictionary.

joeldierkes: Ok, let us keep it as it is.

joeldierkes (approved):

PR #5648, opened 2022-06-16
Feature 5435 removal of python 2 7 related code

joeldierkes:
This PR removes the old Python2 related code.

### DIFF
--- a/lib/rucio/client/baseclient.py
+++ b/lib/rucio/client/baseclient.py
@@ -586,12 +586,7 @@ class BaseClient(object):
                 print("Copy paste the code from the browser to the terminal and press enter:")
                 count = 0
                 while count < 3:
-                    # Python3 default
-                    get_input = input
-                    # if Python version <= 2.7 use raw_input
-                    if sys.version_info[:2] <= (2, 7):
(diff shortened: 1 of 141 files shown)


lib/rucio/client/baseclient.py
```diff
@@ -588,9 +586,6 @@ def __get_token_OIDC(self):
                 while count < 3:
                     # Python3 default
                     get_input = input
-                    # if Python version <= 2.7 use raw_input
-                    if sys.version_info[:2] <= (2, 7):
-                        get_input = raw_input  # noqa: F821 pylint: disable=undefined-variable
                     fetchcode = get_input()
```
rcarpa: we don't need to assign input to a different variable anymore

rcarpa (dismissed):

lib/rucio/client/baseclient.py
joeldierkes: Good catch! :+1:

mlassnig (approved):
:man_cook: oh yeah!

maany (approved):
Changes related to tests/ci are good to go

bari12 (approved):
Very nice 👍 Thanks!

PR #5649, opened 2022-06-16
added postgres_json extension and unit tests. fixed a few typos in mongo plugin

### LINKED ISSUE #5629: Add external postgres metadata interface
Metadata will be expected to be stored in jsonb field.

robbarnsley:
Metadata plugin for interacting with an external postgres database (JSONB).

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -168,5 +168,7 @@ jobs:
         run: docker exec -t dev_rucio_1 tools/pytest.sh -v --tb=short test_download.py::test_download_from_archive_on_xrd
       - name: Test external metadata plugin (mongo)
         run: docker exec -t dev_rucio_1 tools/pytest.sh -v --tb=short test_did_meta_plugins.py::TestDidMetaMongo
+      - name: Test external metadata plugin (postgres)
+        run: docker exec -t dev_rucio_1 tools/pytest.sh -v --tb=short test_did_meta_plugins.py::TestDidMetaExternalPostgresJSON
       - name: Stop containers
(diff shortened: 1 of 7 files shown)


robbarnsley:
Are type annotations compulsory for this PR to be merged? IMHO, I think it's a mistake to make this mandatory, and I am quite reluctant, although if it's necessary I will do it.

I don't really have the time to go back and refactor old code to include this, and having a mixture of annotated functions and non-annotated functions in the same module will be significantly more confusing to read.

I think it would be wise to exercise some discretion about this rather than have a blanket policy. This code is largely self-contained; it's not an interface.

bari12:
Hi Rob, sorry I missed the comment. In principle they are mandatory, since one of our goals is to static type check the entire code. However, your concerns have also been raised and discussed by others, so we decided not to enforce them for e.g. tests or internal methods. 
In large, the motivation/objective however should be to add them to as much as possible (even internal methods) since the static type checking will work better if this is the case.

I think the code is good, I only have some concerns with the name `PostgresJSONDidMeta`. I think this might be confusing for people who actually run rucio on postgres. They might then think they HAVE to activate this plugin instead of the JSONDIDMeta one. I wonder if this should rather be called `ExternalPostgresJSONDidMeta` or something like that.

robbarnsley:
I agree, that would be less ambiguous. Changed.

bari12 (dismissed):

bari12:
The integration test seems to be failing.

robbarnsley:
Anything more you need done to this @bari12 ?

bari12 (approved):

PR #5650, opened 2022-06-17
Replicas: Fix variable name bug in add_bad_dids #5642

### LINKED ISSUE #5642: Declare a list of bad replicas bug
Motivation
----------
The `add_bad_dids` function in `core` creates a temporary list of dids from the input, which should be used to determine the replicas. This list (falsely labeled as `replicas_list`) https://github.com/rucio/rucio/blob/21c93bd7d02dbc70bc3127ad77fb1a1981b83058/lib/rucio/core/replica.py#L352 however is never filled. Instead, the `list_replicas` function tries to be filled https://github.com/rucio/rucio/blob/21c93bd7d02dbc70bc3127ad77fb1a1981b83058/lib/rucio/core/replica.py#L357

This should throw an error.

The bug got introduced in https://github.com/rucio/rucio/commit/f432881425d62aa35979a0b726c2c2ea512abcd2, the `add_bad_dids` function is not tested.

Modification
------------
Resolve the bug, add tests.

joeldierkes:
The temporary list of replicas to mark as _BAD_ is not references
correctly. Instead, a function is referenced. This produces an
`AttributeError: 'function' object has no attribute 'append'` exception.

### DIFF
--- a/lib/rucio/core/replica.py
+++ b/lib/rucio/core/replica.py
@@ -354,7 +354,7 @@ def add_bad_dids(dids, rse_id, reason, issuer, state=BadFilesStatus.BAD, session
     for did in dids:
         scope = InternalScope(did['scope'], vo=issuer.vo)
         name = did['name']
-        list_replicas.append((scope, name, None))
+        replicas_list.append((scope, name, None))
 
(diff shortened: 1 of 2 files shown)


joeldierkes:
This one is good to go. The issue in the pipeline comes from https://github.com/rucio/rucio/issues/5658, I'm still investigating.

rcarpa (approved):

PR #5653, opened 2022-06-20
Feature #5591 create rules for suspicious replicas with only one declaration

### LINKED ISSUE #5591: Create rules for suspicious replicas with only one declaration
Motivation
----------
Then suspicious replica recoverer only handles replicas that have been declared suspicious at least ten times within the last three days; however, many replicas that have to be dealt with manually are declared less than ten times. This will be the first change that starts to deal with such replicas, starting with replicas that have only been declared once.

Modification
------------
It isn't possible to handle replicas that only have one trace, since there is too little information to go off of. Therefore, a rule for each such replica will be created, which will create multiple traces. This will cause the replica to be picked up by the suspicious replica recoverer, if the replica (or the RSE it is on) does indeed have a problem.

ChristophAmes:
In addition to what is stated below, Python annotations have been added to `suspicious_file_recoverer.py`.

Motivation

Then suspicious replica recoverer only handles replicas that have been declared suspicious at least ten times within the last three days; however, many replicas that have to be dealt with manually are declared less than ten times. This will be the first change that starts to deal with such replicas, starting with replicas that have only been declared once.

Modification

It isn't possible to handle replicas that only have one trace, since there is too little information to go off of. Therefore, a rule for each such replica will be created, which will create multiple traces. This will cause the replica to be picked up by the suspicious replica recoverer, if the replica (or the RSE it is on) does indeed have a problem.

### DIFF
--- a/lib/rucio/core/replica.py
+++ b/lib/rucio/core/replica.py
@@ -3813,7 +3813,7 @@ def get_replicas_state(scope=None, name=None, session=None):
 
 
 @read_session
-def get_suspicious_files(rse_expression, available_elsewhere, filter_=None, logger=logging.log, younger_than=10, nattempts=0, session=None, exclude_states=['B', 'R', 'D'], is_suspicious=False):
+def get_suspicious_files(rse_expression, available_elsewhere, filter_=None, logger=logging.log, younger_than=10, nattempts=0, nattempts_exact=False, session=None, exclude_states=['B', 'R', 'D'], is_suspicious=False):
     """
     Gets a list of replicas from bad_replicas table which are: declared more than <nattempts> times since <younger_than> date,
     present on the RSE specified by the <rse_expression> and do not have a state in <exclude_states> list.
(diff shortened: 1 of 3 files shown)


lib/rucio/core/replica.py
```diff
@@ -3908,8 +3909,13 @@ def get_suspicious_files(rse_expression, available_elsewhere, filter_=None, logg
     query = query.filter(not_(other_states_present))
 
     # finally, the results are grouped by RSE, scope, name and required to have
-    # at least 'nattempts' occurrences in the result of the query per replica
-    query_result = query.group_by(models.RSEFileAssociation.rse_id, bad_replicas_alias.scope, bad_replicas_alias.name).having(func.count() > nattempts).all()
+    # at least 'nattempts' occurrences in the result of the query per replica.
+    # If nattempts_exact, then only replicas are required to have exactly
+    # 'nattempts' occurences.
+    if nattempts_exact:
+        query_result = query.group_by(models.RSEFileAssociation.rse_id, bad_replicas_alias.scope, bad_replicas_alias.name).having(func.count() == nattempts).all()
```
mlassnig, line 3916: I'm not understanding something here. Wouldn't it be easier to just test for `>=` always? Is there a different way to handle `==` and `>` later on that I'm not seeing?

ChristophAmes: The idea is that we want to create rules for replicas with exactly `nattempts==1`. Whether or not this should be applied to replicas with `nattempts>=1` as well is still up for discussion.

mlassnig: :+1:

cserf:
Looks good to me, but a few comments:
- The check on the availability will have to be changed when Joel has implemented the new availability model : https://github.com/rucio/rucio/issues/5664
- The threshold at 1 should be OK for the suspicious files reported by consistency checks or by failed download. But for failed transfers between 2 RSEs, there are 3 attempts in a relatively short timescale. These ones won't probably be selected to create the extra rules, but that should be fine because another 3 attempts are retried after a few hours.

cserf (approved):

Issue #5464, opened 2022-04-08
Cannot update `continent` attribute of RSE


joeldierkes:
Motivation
----------
The `continent` attribute of an RSE can not be changed. This is inconsistent, since all other place related attributes (`city`, `region_code`, ...) can be changed.

https://github.com/rucio/rucio/blob/3d6a672da0401c57b5ea4dfe2037ca83a7bfbadd/lib/rucio/core/rse.py#L1280

Modification
------------
Allow the update of the `continent` attribute.

bari12:
Side-comment: Maybe re-design RSE attributes vs. rse settings (RSE Settings should only be a very limited list of keys)

PR #5654, opened 2022-06-20
Core & Internals: Change unnecessary RSE settings to RSE attributes #5464

joeldierkes:
The RSE database table is convoluted with unnecessary entries. Columns
like `continent` do not need to be defined here, they should rather be a
RSE attribute.

This PR does not change the exposed interface.

I added the properties I thought were important, maybe we should add more? (e.g. `ISP`?)

### DIFF
--- a/lib/rucio/core/rse.py
+++ b/lib/rucio/core/rse.py
@@ -1396,6 +1396,7 @@ MUTABLE_RSE_PROPERTIES = {
     'city',
     'staging_area',
     'qos_class',
+    'continent',
 }
 
 
--- a/lib/rucio/core/rse.py
+++ b/lib/rucio/core/rse.py
@@ -24,7 +24,7 @@ import sqlalchemy
 import sqlalchemy.orm
 from dogpile.cache.api import NO_VALUE
 from sqlalchemy.exc import DatabaseError, IntegrityError, OperationalError
-from sqlalchemy.orm import aliased
+from sqlalchemy.orm import aliased, Session
 from sqlalchemy.orm.exc import FlushError
 from sqlalchemy.sql.expression import or_, false, func, case
 
@@ -39,7 +39,6 @@ from rucio.db.sqla.constants import (RSEType, ReplicaState)
 from rucio.db.sqla.session import read_session, transactional_session, stream_session
 
 from typing import Any, Dict, Optional, Sequence
-from sqlalchemy.orm import Session
 
 REGION = make_region_memcached(expiration_time=900)
 
@@ -158,6 +157,9 @@ class RseData:
                 rse_data.limits = limits
 
 
+RSE_SETTINGS = ["continent", "city", "region_code", "country_name", "time_zone", "ISP", "ASN"]
+
+
 @transactional_session
 def add_rse(rse, vo='def', deterministic=True, volatile=False, city=None, region_code=None, country_name=None, continent=None, time_zone=None,
             ISP=None, staging_area=False, rse_type=RSEType.DISK, longitude=None, latitude=None, ASN=None, availability_read: Optional[bool] = None,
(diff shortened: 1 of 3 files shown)


joeldierkes:
Since this is my first PR where I migrate data in the database, please review extra careful! :sweat_smile:

lib/rucio/core/rse.py
```diff
@@ -161,6 +158,9 @@ def bulk_load(rse_datas: "Sequence[RseData]", load_name=False, load_columns=Fals
                 rse_data.limits = limits
 
 
+RSE_SETTINGS = ["continent", "city", "region_code", "country_name", "time_zone"]
+
+
 @transactional_session
 def add_rse(rse, vo='def', deterministic=True, volatile=False, city=None, region_code=None, country_name=None, continent=None, time_zone=None,
```
mlassnig, line 164: I think we can actually update the signature as well. It needs to be fixed here as well though: https://github.com/rucio/probes/blob/master/atlas/check_sync_rses_with_cric

lib/rucio/core/rse.py
```diff
@@ -188,9 +188,8 @@ def add_rse(rse, vo='def', deterministic=True, volatile=False, city=None, region
     if isinstance(rse_type, string_types):
         rse_type = RSEType(rse_type)
 
-    new_rse = models.RSE(rse=rse, vo=vo, deterministic=deterministic, volatile=volatile, city=city,
-                         region_code=region_code, country_name=country_name,
-                         continent=continent, time_zone=time_zone, staging_area=staging_area, ISP=ISP, availability=availability,
+    new_rse = models.RSE(rse=rse, vo=vo, deterministic=deterministic, volatile=volatile,
+                         staging_area=staging_area, ISP=ISP, availability=availability,
```
mlassnig: don't think ISP is in any way useful

lib/rucio/core/rse.py
```diff
@@ -1400,7 +1423,7 @@ def update_rse(rse_id: str, parameters: 'Dict[str, Any]', session=None):
     Update RSE properties like availability or name.
 
     :param rse_id: the id of the new rse.
-    :param  parameters: A dictionnary with property (name, read, write, delete as keys).
+    :param parameters: A dictionnary with property (name, read, write, delete as keys).
```
mlassnig: dictionnary -> dictionary

lib/rucio/db/sqla/migrate_repo/versions/2190e703eb6e_move_rse_settings_to_rse_attributes.py
```diff
@@ -0,0 +1,113 @@
...
+)
+
+
+def upgrade():
+    '''
+    Upgrade the database to this revision
+    '''
+
+    if context.get_context().dialect.name in ['oracle', 'mysql', 'postgresql']:
+        conn = get_bind()
```
mlassnig, line 62: probably needs a transaction

lib/rucio/db/sqla/models.py
```diff
@@ -752,11 +752,6 @@ class RSE(BASE, SoftModelBase):
                       default=False)
     staging_area = Column(Boolean(name='RSE_STAGING_AREA_CHK', create_constraint=True),
                           default=False)
-    city = Column(String(255))
-    region_code = Column(String(2))
-    country_name = Column(String(255))
-    continent = Column(String(2))
-    time_zone = Column(String(255))
     ISP = Column(String(255))
```
mlassnig: i think these two (ISP and ASN) can go as well

lib/rucio/core/rse.py
joeldierkes: This will involve a bigger change (Rest/Client Api with deprecation, other places where `add_rse(` is called), I'd like to do that in a follow-up PR once the data is migrated.

lib/rucio/db/sqla/models.py
joeldierkes: Erased from existence

lib/rucio/db/sqla/migrate_repo/versions/2190e703eb6e_move_rse_settings_to_rse_attributes.py
joeldierkes: Same as https://github.com/rucio/rucio/pull/5665#discussion_r907260030

joeldierkes:
Version two of this PR contains some changes:

Same as #5664, the columns do not get removed from the database for backwards compatibility. The `add_rse`, `update_rse` and `{get,list}_rses` use the new behavior.

This PR therefore needs a follow-up: remove the database columns and backwards compatibility. I'd include the signature change there, since this involves more changes. In this PR I want to focus on the data migration/new behavior.

lib/rucio/db/sqla/migrate_repo/versions/2190e703eb6e_move_rse_settings_to_rse_attributes.py
```diff
@@ -0,0 +1,111 @@
...
+            rse_table = sa.sql.table(
+                "rses",
+                sa.Column("id", GUID()),
+                sa.Column(setting, setting_datatype),
+                sa.Column("created_at", sa.DateTime),
+                sa.Column("updated_at", sa.DateTime),
+                schema=SCHEMA
+            )
+
+            rses = conn.execute(rse_table.select().where(sa.sql.column(setting) != None)).fetchall()  # noqa: E711
```
bari12: As discussed on Slack already, I am not sure this works. 
One key principle is that you can also do a `alembic upgrade head --sql` which generates you all the SQL statements (to verify and execute by hand). But if I do this with this revision, it fails.

joeldierkes: Yes, I changed it as discussed.

The sql query used to downgrade the db is using a subquery since oracle does not support the update query with multiple tables.

bari12 (approved):

PR #5655, opened 2022-06-21
Core & Internals: reduce verbosity of cache access errors

rcarpa:
Otherwise we print the stack trace all the time when we run the
rucio client.

Also, check if it's a client running and completely disable the cache
in those cases. is_client() is not 100% accurate, as it will still
return False when the rucio client is run from a machine configured
to run a server/daemon. For example, in the dev containers the
logging messages will still be printed, but should be less scary now.

### DIFF
--- a/lib/rucio/common/cache.py
+++ b/lib/rucio/common/cache.py
@@ -18,17 +18,24 @@ import logging
 from dogpile.cache import make_region
 
 from rucio.common.config import config_get
+from rucio.common.utils import is_client
 
 CACHE_URL = config_get('cache', 'url', False, '127.0.0.1:11211', check_config_table=False)
 
 ENABLE_CACHING = True
 _mc_client = None
 try:
-    import pymemcache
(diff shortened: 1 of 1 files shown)


lib/rucio/common/cache.py
```diff
@@ -18,17 +18,24 @@
...
 
 ENABLE_CACHING = True
 _mc_client = None
 try:
-    import pymemcache
-    _mc_client = pymemcache.Client(CACHE_URL, connect_timeout=1, timeout=1)
-    _mc_client.version()
-except (IOError, ImportError):
-    logging.warning("Cannot access memcached at {}. Caching will be disabled".format(CACHE_URL), exc_info=True)
+    if is_client():
```
mlassnig, line 28: is this check necessary as you set ENABLE_CACHING to true above anyway?

mlassnig: sorry, forget, i'm stupid and don't understand basic logic

mlassnig (approved):

PR #5657, opened 2022-06-21
Database: Add poolclass configuration option for database engine #5391

### LINKED ISSUE #5391: Add poolclass configuration options for database engine
Motivation
----------

In case there's a db connection pooler sitting between SQLAlchemy and the Database (e.g. pgpool), it makes sense to default SQLAlchemy to NullPool. 

Modification
------------
Extend [database] config option for poolclass.

akselaase:
Adds the configuration option `poolclass` to the database section of `rucio.cfg`, which takes one of the values `queuepool` (the default) or `nullpool`. Setting the option to `nullpool` disables connection pooling in SQLAlchemy, which is useful if pooling is already performed elsewhere. Unsetting the option or setting it to `queuepool` retains the previous pooling behaviour.

I'll squash the commits before merging.

Fixes #5391

### DIFF
--- a/lib/rucio/db/sqla/session.py
+++ b/lib/rucio/db/sqla/session.py
@@ -28,9 +28,10 @@ from sqlalchemy import create_engine, event
 from sqlalchemy.exc import DatabaseError, DisconnectionError, OperationalError, TimeoutError
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.orm import sessionmaker, scoped_session
+from sqlalchemy.pool import QueuePool, SingletonThreadPool, NullPool
 
 from rucio.common.config import config_get
-from rucio.common.exception import RucioException, DatabaseException
+from rucio.common.exception import RucioException, DatabaseException, InputValidationError
 from rucio.common.extra import import_extras
 
 EXTRA_MODULES = import_extras(['MySQLdb', 'pymysql'])
@@ -158,6 +159,29 @@ def my_on_connect(dbapi_con, connection_record):
     dbapi_con.action = caller
 
 
+def _get_engine_poolclass(poolclass):
+    """Resolve the correct SQLAlchemy Pool type to use from the
+    poolclass config option.
+
+    :param poolclass: User-selected pool class from config file.
+    :returns: The corresponding SQLAlchemy Pool class.
+    :raises InputValidationError: if config value doesn't correspond to an SQLAlchemy Pool class.
+    """
+
+    SQLA_CONFIG_POOLCLASS_MAPPING = {
+        'queuepool': QueuePool,
+        'singletonthreadpool': SingletonThreadPool,
+        'nullpool': NullPool,
+    }
+
+    poolclass = poolclass.lower()
+
+    if poolclass not in SQLA_CONFIG_POOLCLASS_MAPPING:
+        raise InputValidationError('Unknown poolclass: %s' % poolclass)
+
+    return SQLA_CONFIG_POOLCLASS_MAPPING[poolclass]
+
+
 def get_engine():
     """ Creates a engine to a specific database.
         :returns: engine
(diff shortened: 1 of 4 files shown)


mlassnig:
That was quick! :eyes:

akselaase:
It was a... good first issue.

akselaase:
Note that setting `poolclass = nullpool` conflicts with other options like `pool_size`. Should this be documented somewhere?

mlassnig:
Maybe add this to the rucio.cfg.template file?

mlassnig:
And there's of course https://github.com/rucio/documentation/ :-)

akselaase:
Added the `singletonthreadpool` choice as well, since this is the default pooling for the SQLite engine. Two more pooling mechanisms supported by SQLAlchemy could be added as well, `StaticPool` and `AssertionPool`, but I'm not sure if they are desired.

Also removed type annotations according to [Rucio Type Annotation Guide](https://codimd.web.cern.ch/6-SU3cTpQSWRK6FHkM7mAA?view).

I've submitted a PR for documentation at https://github.com/rucio/documentation/pull/139.

I'm not sure what I would add to `rucio.cfg.template`. Setting the option will incur a change of behavior, so best to just leave it like before since this is a patch PR maybe? I could add a commented line demonstrating the usage though.

lib/rucio/db/sqla/session.py
```diff
@@ -158,6 +159,21 @@ def my_on_connect(dbapi_con, connection_record):
     dbapi_con.action = caller
 
 
+def _get_engine_poolclass(poolclass):
+    """Resolve the correct SQLAlchemy Pool type to use from the
+        poolclass config option.
```
joeldierkes: The second line should align with the double colons. The parameters should be documented, to be consistent with the other docs.

```python
"""Resolve the...
poolclass

:param poolclass: ...
:returns: ...
:raises ValueError: ...
"""
```

lib/rucio/db/sqla/session.py
```diff
@@ -158,6 +159,21 @@ def my_on_connect(dbapi_con, connection_record):
...
+    """
+    try:
+        resolved_cls = {
+            'queuepool': QueuePool,
+            'singletonthreadpool': SingletonThreadPool,
+            'nullpool': NullPool,
+        }[poolclass]
```
joeldierkes: The dict is a constant, we should extract it. This also enables us to use an early return instead of the try ... catch:
```python
if poolclass not in SQLALCHEMY_POOL_TYPES:
    raise raise ValueError('Unknown poolclass: %s' % poolclass)

return SQLALCHEMY_POOL_TYPES[poolclass]
```

That's a personal choice, thoughts about that?

joeldierkes: Also, how about calling `lower` on the poolclass variable. In case, someone uses the class name instead of the key defined by us.

lib/rucio/db/sqla/session.py
```diff
@@ -158,6 +159,21 @@ def my_on_connect(dbapi_con, connection_record):
...
+        return resolved_cls
+    except KeyError:
+        raise ValueError('Unknown poolclass: %s' % poolclass)
```
joeldierkes: We define our own Exception types in `lib/rucio/common/exception.py`. `InputValidationError` could work here.

joeldierkes (changes requested):
GJ, small nits.

akselaase:
@joeldierkes Rebased and squashed commits.

joeldierkes (dismissed):
lgtm!

mlassnig (approved):

joeldierkes (approved):

bari12:
Although rather smaller change, I will merge this as a feature, thus it only becomes available with 1.29 (Which incidentally is now anyway)

Issue #5660, opened 2022-06-22
Multi-vo tests run httpd server with non-multi-vo policy package


rcarpa:
Motivation
----------
httpd is started before the multi-vo suite is initialized, so it runs with the default configuration used by non-multi-vo tests. 

When migrating did.py to using temporary tables, test_dids.test_list_recursive started to fail because 
`get_schema_value('SCOPE_LENGTH'))` returns 25 and the temporary table is created with wrong scope length. The multi-vo scope length is expected to be 29.

PR #5661, opened 2022-06-22
Testing: use correct httpd rucio configuration in multi-vo tests. #5660

rcarpa:
This doesn't fully fix the linked issue, because the multi-vo tests are run with to configurations and the second iteration
will still have an incorrect config. 

However, it improves the situation and fixes the test failures for my current did rework


maany (approved):
Thanks Radu! LGTM, can we open a followup issue to look into the multi-vo tests. :)

joeldierkes (approved):

Issue #5658, opened 2022-06-21
Oracle container breaks GitHub Action patch pipeline


joeldierkes:
Motivation
----------
The `release-patch-test (..., oracle)` tests fail to create the necessary Docker containers. This results in an error
```
release-patch-test (fedora33, 3.9, autotest, all, oracle) 
```
which breaks the pipeline.

Modification
------------
Resolve the error and fix the pipeline.

joeldierkes:
This might now be the oracle container, I'm investigating.

PR #5663, opened 2022-06-23
Testing: Fix oracle script shebag #5658

joeldierkes:
The `oracle_setup.sh` script uses `bash` as shell. While this comes with
more features than sh, it seems to be incompatible with `sh` only
containers. The oracle container used in the GitHub Actions patch jobs
seem to be affected by this.


maany:
Looks good. Nice find @joeldierkes

maany (approved):

PR #5665, opened 2022-06-23
Feature 5664 split the rse availability column into its respective values

joeldierkes:
This PR splits the RSE availability column.

### DIFF
--- a/lib/rucio/core/replica.py
+++ b/lib/rucio/core/replica.py
@@ -763,7 +763,7 @@ def _list_replicas_for_datasets(dataset_clause, state_clause, rse_clause, ignore
                  models.DataIdentifierAssociation.child_name)
 
     if not ignore_availability:
-        replica_query = replica_query.filter(models.RSE.availability.in_((4, 5, 6, 7)))
+        replica_query = replica_query.filter(models.RSE.availability_read == true())
 
     if state_clause is not None:
         replica_query = replica_query.filter(and_(state_clause))
@@ -812,7 +812,7 @@ def _list_replicas_for_constituents(constituent_clause, state_clause, files_wo_r
                  models.ConstituentAssociation.child_name)
 
     if not ignore_availability:
-        constituent_query = constituent_query.filter(models.RSE.availability.in_((4, 5, 6, 7)))
+        constituent_query = constituent_query.filter(models.RSE.availability_read == true())
 
     if state_clause is not None:
         constituent_query = constituent_query.filter(and_(state_clause))
@@ -846,7 +846,7 @@ def _list_replicas_for_files(file_clause, state_clause, files_wo_replica, rse_cl
         ]
 
         if not ignore_availability:
-            filters.append(models.RSE.availability.in_((4, 5, 6, 7)))
+            filters.append(models.RSE.availability_read == true())
 
         if state_clause is not None:
             filters.append(state_clause)
@@ -1450,7 +1450,7 @@ def add_replicas(rse_id, files, account, ignore_availability=True,
     if replica_rse.volatile is True:
         raise exception.UnsupportedOperation('Cannot add replicas on volatile RSE %s ' % (replica_rse.rse))
 
-    if not (replica_rse.availability & 2) and not ignore_availability:
+    if not (replica_rse.availability_write) and not ignore_availability:
         raise exception.ResourceTemporaryUnavailable('%s is temporary unavailable for writing' % replica_rse.rse)
 
     for file in files:
@@ -1551,7 +1551,7 @@ def __delete_replicas(rse_id, files, ignore_availability=True, session=None):
 
     replica_rse = get_rse(rse_id=rse_id, session=session)
 
-    if not (replica_rse.availability & 1) and not ignore_availability:
+    if not replica_rse.availability_delete and not ignore_availability:
         raise exception.ResourceTemporaryUnavailable('%s is temporary unavailable'
                                                      'for deleting' % replica_rse.rse)
 
@@ -2076,7 +2076,7 @@ def __delete_replicas_without_temp_tables(rse_id, files, ignore_availability=Tru
     """
     replica_rse = get_rse(rse_id=rse_id, session=session)
 
-    if not (replica_rse.availability & 1) and not ignore_availability:
+    if not replica_rse.availability_delete and not ignore_availability:
         raise exception.ResourceTemporaryUnavailable('%s is temporary unavailable'
                                                      'for deleting' % replica_rse.rse)
 
--- a/lib/rucio/core/rse.py
+++ b/lib/rucio/core/rse.py
@@ -19,7 +19,6 @@ import logging
 import traceback
 from io import StringIO
 from re import match
-from typing import TYPE_CHECKING
 
 import sqlalchemy
 import sqlalchemy.orm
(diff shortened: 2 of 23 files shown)


lib/rucio/core/rse.py
```diff
@@ -1414,26 +1400,25 @@ def update_rse(rse_id: str, parameters: 'Dict[str, Any]', session=None):
...
     except sqlalchemy.orm.exc.NoResultFound:
         raise exception.RSENotFound('RSE with ID \'%s\' cannot be found' % rse_id)
-    availability = 0
     rse = query.rse
-    for column in query:
-        if column[0] == 'availability':
-            availability = column[1] or availability
+
+    # What do these lines I removed mean????????
+    ###########################################################################################################
```
joeldierkes: Some closure on what this piece did would be cool!

lib/rucio/core/rse.py
```diff
@@ -1414,26 +1400,25 @@ def update_rse(rse_id: str, parameters: 'Dict[str, Any]', session=None):
         query = session.query(models.RSE).filter_by(id=rse_id).one()
     except sqlalchemy.orm.exc.NoResultFound:
         raise exception.RSENotFound('RSE with ID \'%s\' cannot be found' % rse_id)
-    availability = 0
     rse = query.rse
-    for column in query:
-        if column[0] == 'availability':
-            availability = column[1] or availability
+
```
mlassnig, line 1404: this looks like some ancient way to find a column. not entirely sure :confused:

lib/rucio/db/sqla/migrate_repo/versions/1677d4d803c8_split_rse_availability_into_multiple.py
```diff
@@ -0,0 +1,127 @@
...
+    schema=SCHEMA,
+)
+
+
+def upgrade():
+    """
+    Upgrade the database to this revision
+    """
+
+    conn = get_bind()
```
mlassnig: I think this is missing some transaction?

rcarpa:
As somebody who did quite a bit of C programming and unix administration, I find the arguments for the change lacking :D (but I'm not against this PR :)  )

mlassnig:
> As somebody who did quite a bit of C programming and unix administration, I find the arguments for the change lacking :D (but I'm not against this PR :) )

I love the POSIX permissions as well :-) My main request for this was that I continously forget the numbers, and my old brain can deal better with explicit permissions :-p

lib/rucio/core/replica.py
```diff
@@ -763,7 +763,8 @@ def _list_replicas_for_datasets(dataset_clause, state_clause, rse_clause, ignore
                  models.DataIdentifierAssociation.child_name)
 
     if not ignore_availability:
-        replica_query = replica_query.filter(models.RSE.availability.in_((4, 5, 6, 7)))
+        true_value = True
+        replica_query = replica_query.filter(models.RSE.availability_read == true_value)
```
rcarpa: I would rather use the `models.RSE.availability_read.is_(True)` or `models.RSE.availability_read == true()`, (assuming the import `from sqlalchemy.sql.expression import true`)

lib/rucio/core/rse.py
```diff
@@ -476,28 +483,12 @@ def list_rses(filters={}, session=None):
                     query = query.filter(getattr(models.RSE, k) == RSEType[v])
                 else:
                     query = query.filter(getattr(models.RSE, k) == v)
-            elif k in ['availability_read', 'availability_write', 'availability_delete']:
```
rcarpa, line 479: This has the side effect that `filters["availability_read"] = None` is now translated into a query `models.RSE.availability_read is NULL". This is probably not problematic, but is a corner case to be kept in mind.

lib/rucio/db/sqla/migrate_repo/versions/1677d4d803c8_split_rse_availability_into_multiple.py
```diff
@@ -0,0 +1,127 @@
...
+                    {
+                        "availability_read": sa.bindparam("availability_read"),
+                        "availability_write": sa.bindparam("availability_write"),
+                        "availability_delete": sa.bindparam("availability_delete"),
+                    }
+                ),
+                transformed_rses,
+            )
+
+        drop_column("rses", "availability", schema=SCHEMA)
```
rcarpa: This will break the possibility to run 1.28.<whatever> in parallel with 1.29 during the testing phase. We should probably ensure that both versions can be run in parallel for a while. The old column can be dropped in the next big release.

rcarpa (commented):
I'm not against the change, but, personally, I'd would have just added 3 constants in constants.py:
```
AVAILABILITY_READ = ...
AVAILABILITY_WRITE = ...
AVAILABILITY_DELETE = ...
```
imported them everywhere and used them instead of hardcoded numbers:
```
if rse.availability == AVAILABILITY_READ | AVAILABILITY_WRITE | AVAILABILITY_DELETE
```
It has the same benefices of making the things more lisible, and I personally find it more concise and easier to read

lib/rucio/core/rse.py
joeldierkes: I think so too, just wanted to make sure.

lib/rucio/db/sqla/migrate_repo/versions/1677d4d803c8_split_rse_availability_into_multiple.py
joeldierkes: This `get_bind` lets us use the alembic connection to the database. https://alembic.sqlalchemy.org/en/latest/ops.html#alembic.operations.Operations.get_bind

https://alembic.sqlalchemy.org/en/latest/ops.html#alembic.operations.Operations.execute
This describes the usage, I think it should be fine.

lib/rucio/core/replica.py
joeldierkes: Yeah, I like it!

lib/rucio/core/rse.py
joeldierkes: I think we had the same problem with `filters["availability"]`. Yeah we should keep this in mind.

lib/rucio/db/sqla/migrate_repo/versions/1677d4d803c8_split_rse_availability_into_multiple.py
joeldierkes: Yes, changed. This means that this PR needs a follow up to remove the `availability` column.

joeldierkes:
As discussed with @rcarpa about his Review, the usage of these flags would not solve the problem in the database to easily filter for certain availabilities (except if you're a hardcore Linux user, which unfortunately not enough people are :( ).

Introducing the `availability_read` and friend in the database and flags in the code would involve to always cast the availability when the database is accessed, since we don't have a proper abstraction for the access. This introduces a lot of space for bugs, e.g. if the cast is forgotten.

PS:
The enum [(Int)Flag](https://docs.python.org/3/library/enum.html#flag) provides some nice interfaces, compared to using constants.

lib/rucio/core/replica.py
joeldierkes:
```python
models.RSE.availability_read.is_(True)
```
seems to be problematic with Oracle, it produces the SQL query part: `system.rses.availability_read IS 1`, which is not recognized by Oracle.

mlassnig (approved):
judging from a "user" point of view, this now does exactly what i want. thanks!

rcarpa (approved):

PR #5666, opened 2022-06-24
Feature 5583 make subscriptions page real time

### LINKED ISSUE #5583: Make subscriptions page real-time
Motivation
----------

https://github.com/rucio/rucio/blob/86390dfa9e6a99d82a315a2733ba06f6687c4bb4/lib/rucio/web/ui/static/rucio.js#L1687

The Web UI appears to depend on dumps to populate the information on the subscriptions page. This can cause problems if the production of these dumps is disrupted or never configured in the first place. In addition, there seems to be some time-related calculation that is not timezone-agnostic.

Modification
------------

Evaluate the effect of fetching the information from the database in real-time and consider switching to it.

nimishbongale:
Fixes #5583 

1. Removes dependency on file reading to update subscriptions
2. Hits `subscriptions/<account>/Rules/State` directly to fetch subscriptions
3. Updates retry logic

### DIFF
--- a/lib/rucio/web/ui/static/rucio.js
+++ b/lib/rucio/web/ui/static/rucio.js
@@ -1684,7 +1684,7 @@ RucioClient.prototype.send_trace = function(options) {
 /* --- rucio dumps methods --- */
 
 /* list all subscriptions from dumps */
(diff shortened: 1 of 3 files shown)


cserf:
LGTM

cserf (approved):

PR #5669, opened 2022-06-24
Core: Fix RSE `supported_checksums` attribute parsing; Fix #5155

### LINKED ISSUE #5155: Per-RSE checksum specification misbehaving for MD5-only
Motivation
----------
I'm trying to use an admittedly ancient gridftp server which reports the MD5 value instead of Adler32.  I thought the solution/workaround was to use:
```
rucio-admin rse set-attribute --rse LIGO-CIT --key supported_checksums --value md5
```
However, I notice that this actually completely disables all FTS checksums.

Modification
------------
If i check the supported checksums with `get_rse_supported_checksums_from_attributes`, the result is:
```
['m']
```
And `checksum_validation_strategy` returns:
```
('destination', ['m'])
```
but I believe the correct result should instead be:
```
('both', ['md5'])
```
As far as I can tell, it looks like the problem is in [`rse.py#L641`](https://github.com/rucio/rucio/blob/master/lib/rucio/core/rse.py#L641):
```
checksum_attribute[0].split(',')
```
I'm not sure what useful purpose that `[0]` serves?

akselaase:
The function `rse.py::parse_checksum_support_attribute` misbehaved when
being called with arguments of type `str` instead of the assumed
`list[str]`.

This commit adds explicit type hints to functions related to parsing the
`supported_checksums` attribute, and fixes errors at call sites passing
wrong types.

### DIFF
--- a/lib/rucio/core/rse.py
+++ b/lib/rucio/core/rse.py
@@ -19,7 +19,7 @@ import logging
 import traceback
 from io import StringIO
 from re import match
-from typing import Any, Dict, Optional, Sequence, Union
+from typing import Any, Dict, List, Optional, Sequence, Union
 
 import sqlalchemy
 from dogpile.cache.api import NO_VALUE
@@ -762,15 +762,16 @@ def get_rse_attributes(rse_id, session=None):
     return result
 
 
-def get_rse_supported_checksums_from_attributes(rse_attributes):
+def get_rse_supported_checksums_from_attributes(rse_attributes: Dict[str, Any]) -> List[str]:
     """
     Parse the RSE attribute defining the checksum supported by the RSE
     :param rse_attributes: attributes retrieved using list_rse_attributes
+    :returns: A list of the names of supported checksums indicated by the specified attributes.
     """
-    return parse_checksum_support_attribute(rse_attributes.get(CHECKSUM_KEY))
+    return parse_checksum_support_attribute(rse_attributes.get(CHECKSUM_KEY, ''))
 
 
-def parse_checksum_support_attribute(checksum_attribute):
+def parse_checksum_support_attribute(checksum_attribute: str) -> List[str]:
     """
     Parse the checksum support RSE attribute.
     :param checksum_attribute: The value of the RSE attribute storing the checksum value
@@ -782,10 +783,12 @@ def parse_checksum_support_attribute(checksum_attribute):
 
     if not checksum_attribute:
         return GLOBALLY_SUPPORTED_CHECKSUMS
+
+    supported_checksum_list = [c.strip() for c in checksum_attribute.split(',') if c.strip()]
+
+    if 'none' in supported_checksum_list:
+        return []
     else:
-        supported_checksum_list = [c.strip() for c in checksum_attribute.split(',') if c.strip()]
-        if 'none' in supported_checksum_list:
-            return []
         return supported_checksum_list
 
 
--- a/lib/rucio/tests/test_rse.py
+++ b/lib/rucio/tests/test_rse.py
@@ -25,12 +25,14 @@ from rucio.common.exception import (Duplicate, RSENotFound, RSEProtocolNotSuppor
                                     InvalidObject, ResourceTemporaryUnavailable,
                                     RSEAttributeNotFound, RSEOperationNotSupported,
                                     InputValidationError)
-from rucio.common.utils import generate_uuid
+from rucio.common.utils import generate_uuid, GLOBALLY_SUPPORTED_CHECKSUMS, CHECKSUM_KEY
 from rucio.core.rse import (add_rse, get_rse_id, del_rse, restore_rse, list_rses,
                             rse_exists, add_rse_attribute, list_rse_attributes,
                             set_rse_transfer_limits, get_rse_transfer_limits,
                             delete_rse_transfer_limits, get_rse_protocols,
                             del_rse_attribute, get_rse_attribute, get_rse, rse_is_empty,
+                            parse_checksum_support_attribute,
+                            get_rse_supported_checksums_from_attributes,
                             update_rse)
 from rucio.db.sqla import session, models
 from rucio.db.sqla.constants import RSEType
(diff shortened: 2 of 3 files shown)


lib/rucio/core/rse.py
```diff
@@ -749,14 +759,14 @@ def parse_checksum_support_attribute(checksum_attribute):
...
-        supported_checksum_list = checksum_attribute[0].split(',')
+        supported_checksum_list = checksum_attribute.split(',')
         if 'none' in supported_checksum_list:
             return []
         return supported_checksum_list
 
 
 @read_session
-def get_rse_is_checksum_supported(checksum_name, rse_id=None, session=None):
+def get_rse_is_checksum_supported(checksum_name: str, rse_id: str = None, session=None):
```
joeldierkes: `rse_id` is `Optional[str]`

lib/rucio/core/rse.py
```diff
@@ -721,22 +718,35 @@ def get_rse_attributes(rse_id, session=None):
 
 
 @read_session
-def get_rse_supported_checksums(rse_id, session=None):
+def get_rse_supported_checksums(rse_id: str, session=None):
     """
     Retrieve from the DB and parse the RSE attribute defining the checksum supported by the RSE
```
joeldierkes: `:returns:`, `:params:` comments

lib/rucio/core/rse.py
```diff
@@ -721,22 +718,35 @@ def get_rse_attributes(rse_id, session=None):
...
+        raise RuntimeError(message)
+
+    return parse_checksum_support_attribute(first_value)
+
+
+def get_rse_supported_checksums_from_attributes(rse_attributes: Dict[str, Any]):
     """
     Parse the RSE attribute defining the checksum supported by the RSE
     :param rse_attributes: attributes retrieved using list_rse_attributes
     """
```
joeldierkes: `:returns:` missing

lib/rucio/core/rse.py
```diff
@@ -721,22 +718,35 @@ def get_rse_attributes(rse_id, session=None):
...
+
+    else:
+        message = 'Multiple supported_checksum attributes for RSE: "%s"' % values
+        logging.error(message)
+        raise RuntimeError(message)
+
+    return parse_checksum_support_attribute(first_value)
+
+
+def get_rse_supported_checksums_from_attributes(rse_attributes: Dict[str, Any]):
```
joeldierkes: Return type annotation missing

lib/rucio/core/rse.py
```diff
@@ -749,14 +759,14 @@ def parse_checksum_support_attribute(checksum_attribute):
     if not checksum_attribute:
         return GLOBALLY_SUPPORTED_CHECKSUMS
     else:
-        supported_checksum_list = checksum_attribute[0].split(',')
+        supported_checksum_list = checksum_attribute.split(',')
         if 'none' in supported_checksum_list:
             return []
         return supported_checksum_list
```
joeldierkes: nit: This `else` indentation is unnecessary, since there is an early return.

lib/rucio/tests/test_rse.py
```diff
@@ -218,6 +221,65 @@ def test_empty_rse(self):
...
         assert not rse_is_empty(rse_id=rse_id)
 
+    def test_parse_checksum_support_attribute(self):
+        assert parse_checksum_support_attribute('') == GLOBALLY_SUPPORTED_CHECKSUMS
+        assert parse_checksum_support_attribute('none') == []
+        assert parse_checksum_support_attribute('none,md5') == []
+        assert parse_checksum_support_attribute('md5') == ['md5']
+        assert parse_checksum_support_attribute('md5,adler32') == ['md5', 'adler32']
+
+    @pytest.mark.dirty
```
joeldierkes: How about a `del_rse(rse_id)` instead of the dirty marker?

lib/rucio/core/rse.py
```diff
@@ -19,14 +19,14 @@
 import traceback
 from io import StringIO
 from re import match
-from typing import TYPE_CHECKING
+from typing import Any, List, Dict, Optional, Sequence
```
joeldierkes: Alphabetical sorting?

joeldierkes (dismissed):
This is a good Draft PR. :wink:

lib/rucio/core/rse.py
akselaase: I actually think this should be a required argument (so remove the default `None` value). You have `is_checksum_valid` for the RSE-independent case.

lib/rucio/core/rse.py
akselaase: I haven't added them since they're inferred from the return of `parse_checksum_support_attribute`, would you prefer them being explicitly added everywhere?

joeldierkes: For the moment I'd suggest to rather be descriptive, so yes. We can talk about it later.

akselaase:
Currently waiting on #5673 and maybe a PR from @joeldierkes to fix caching behaviour before updating this.

lib/rucio/core/rse.py
```diff
@@ -782,13 +795,32 @@ def parse_checksum_support_attribute(checksum_attribute):
...
+        return []
     else:
-        supported_checksum_list = [c.strip() for c in checksum_attribute.split(',') if c.strip()]
-        if 'none' in supported_checksum_list:
-            return []
         return supported_checksum_list
 
 
+@read_session
+def get_rse_is_checksum_supported(checksum_name: str, rse_id: str = None, session=None) -> bool:
```
rcarpa: this function isn't used anywhere. Why add it ? Do you intend to use it later ?

akselaase: Leftovers from before the rebase, removed.

lib/rucio/core/rse.py
```diff
@@ -782,10 +783,12 @@ def parse_checksum_support_attribute(checksum_attribute):
 
     if not checksum_attribute:
         return GLOBALLY_SUPPORTED_CHECKSUMS
+
+    supported_checksum_list = checksum_attribute.split(',')
```
rcarpa: Why are you getting rid of strip() ?

rcarpa (approved):

Issue #5671, opened 2022-06-27
Refactor the `get_rse_attribute` function


joeldierkes:
Motivation
----------
The `get_rse_attribute` function is hard to read and does some unexpected things for a "get" function:
https://github.com/rucio/rucio/blob/5f200954836884f7d3d2d6a60be31f9078519764/lib/rucio/core/rse.py#L664-L697

In particular:
- The return value is `List`, for a get (a single) attribute function!
- The `value` parameter is messed up.
- The control flow is strange (if the key is cached do not return it immediately, rather don't run the database-fetch and return it afterwards).

Modification
------------
Refactor the `get_rse_attribute` function, so it does what we expect from it.

PR #5672, opened 2022-06-27
Core: Refactor the get_rse_attribute method #5671

joeldierkes:
The `get_rse_attribute` method does some things which are unexpected
by a `get` function:

- The return type is `List[Union[bool,str]]`: The function returns a
  list with one element (since it filters for `rse_id` and `key`, both
  construct the primary key of the `RSEAttrAssociation` table) if an
  attribute is found, or an empty list if no attribute is found. It may
  also return a list of all attributes for all rses if `rse_id` is not
  specified or a list with just multiple items with the value `value` if
  `key` and `value` is specified.

- The `value` parameter: This parameter lets the user select all values
  for a specific key, which has the value `value`. This returns a list
  containing the value `value` as often as the `key` - `value` pair
  exists in the table, regardless of the rse. This functionality is
  gladly never used in our code-base.

The following are inconsistencies too:

- The cache: The cache does not conform with non-existing values for
  `rse_id` and `value`.

- The `use_cache` parameter: This parameter toggles an execution branch
  depending on a boolean value. This makes the function more complex
  then it has to be.

The new `get_rse_attribute` function has some improvements:

- It does what it says, it returns **one** rse attribute for a rse. The
  return value `None` clearly communicates that a value does not exist.

- The function is split into two functions: one for a cached and one for
  an un-cached access. This removes the `use_cache` parameter in the
  function.

- The `value` parameter got removed completely. No explanation needed,
  what did it do there in the first place?

- The `rse_id` and `key` are both mandatory now. It makes little sense
  to "get" the rse attribute for a specific key without a rse. This
  case should be handled in a `list_rse_attributes` function.

- The `rse_id` now comes before the `key` parameter. This is logical,
  since we want to get an attribute of a rse and not a rse of an
  attribute.

### DIFF
--- a/lib/rucio/core/replica.py
+++ b/lib/rucio/core/replica.py
@@ -1097,11 +1097,9 @@ def _list_replicas(dataset_clause, file_clause, state_clause, show_pfns,
 
                     # do we need to sign the URLs?
                     if sign_urls and protocol.attributes['scheme'] == 'https':
-                        service = get_rse_attribute('sign_url',
-                                                    rse_id=rse_id,
-                                                    session=session)
-                        if service and isinstance(service, list):
-                            pfn = get_signed_url(rse_id=rse_id, service=service[0], operation='read', url=pfn, lifetime=signature_lifetime)
+                        service = get_rse_attribute(rse_id, 'sign_url', session=session)
+                        if service:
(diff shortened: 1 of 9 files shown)


joeldierkes:
Closed in favor of #5673

PR #5673, opened 2022-06-27
Core: Refactor the get_rse_attribute method #5671

### LINKED ISSUE #5671: Refactor the `get_rse_attribute` function

joeldierkes:
The `get_rse_attribute` method does some things which are unexpected by
a `get` function:

- The return type is `List[Union[bool,str]]`: The function returns a
  list with one element (since it filters for `rse_id` and `key`, both
  construct the primary key of the `RSEAttrAssociation` table) if an
  attribute is found, or an empty list if no attribute is found. It may
  also return a list of all attributes for all rses if `rse_id` is not
  specified or a list with just multiple items with the value `value` if
  `key` and `value` is specified.

- The `value` parameter: This parameter lets the user select all values
  for a specific key, which has the value `value`. This returns a list
  containing the value `value` as often as the `key` - `value` pair
  exists in the table, regardless of the rse. This functionality is
  gladly never used in our code-base.

The following are inconsistencies too:

- The cache: The cache does not conform with non-existing values for
  `rse_id` and `value`.

- The `use_cache` parameter: This parameter toggles an execution branch
  depending on a boolean value. This makes the function more complex
  than it has to be.

The new `get_rse_attribute` function has some improvements:

- It does what it says, it returns **one** rse attribute for a rse. The
  return value `None` clearly communicates that a value does not exist.

- The `value` parameter got removed completely. No explanation needed,
  what did it do there in the first place?

- The `rse_id` and `key` are both mandatory now. It makes little sense
  to "get" the rse attribute for a specific key without a rse. This case
  should be handled in a `list_rse_attributes` function.

- The `rse_id` now comes before the `key` parameter. This is logical,
  since we want to get an attribute of a rse and not a rse of an
  attribute.

The `use_cache` parameter is still present, so the function is
consistent with the rest of the code base.

### DIFF
--- a/lib/rucio/core/replica.py
+++ b/lib/rucio/core/replica.py
@@ -1096,11 +1096,9 @@ def _list_replicas(dataset_clause, file_clause, state_clause, show_pfns,
 
                     # do we need to sign the URLs?
                     if sign_urls and protocol.attributes['scheme'] == 'https':
-                        service = get_rse_attribute('sign_url',
-                                                    rse_id=rse_id,
-                                                    session=session)
-                        if service and isinstance(service, list):
-                            pfn = get_signed_url(rse_id=rse_id, service=service[0], operation='read', url=pfn, lifetime=signature_lifetime)
+                        service = get_rse_attribute(rse_id, 'sign_url', session=session)
+                        if service:
+                            pfn = get_signed_url(rse_id=rse_id, service=service, operation='read', url=pfn, lifetime=signature_lifetime)
 
                     # server side root proxy handling if location is set.
                     # supports root and http destinations
@@ -1109,11 +1107,7 @@ def _list_replicas(dataset_clause, file_clause, state_clause, show_pfns,
                     if domain == 'wan' and protocol.attributes['scheme'] in ['root', 'http', 'https'] and client_location:
 
                         if 'site' in client_location and client_location['site']:
-                            # is the RSE site-configured?
-                            rse_site_attr = get_rse_attribute('site', rse_id, session=session)
-                            replica_site = ['']
-                            if isinstance(rse_site_attr, list) and rse_site_attr:
-                                replica_site = rse_site_attr[0]
+                            replica_site = get_rse_attribute(rse_id, 'site', session=session)
 
                             # does it match with the client? if not, it's an outgoing connection
                             # therefore the internal proxy must be prepended
@@ -1388,7 +1382,7 @@ def __bulk_add_replicas(rse_id, files, account, session=None):
         filter(or_(*condition))
     available_replicas = [dict([(column, getattr(row, column)) for column in row._fields]) for row in query]
 
-    default_tombstone_delay = next(iter(get_rse_attribute('tombstone_delay', rse_id=rse_id, session=session)), None)
+    default_tombstone_delay = get_rse_attribute(rse_id, 'tombstone_delay', session=session)
     default_tombstone = tombstone_from_delay(default_tombstone_delay)
 
     new_replicas = []
--- a/lib/rucio/core/request.py
+++ b/lib/rucio/core/request.py
@@ -1823,15 +1823,13 @@ def __throttler_request_state(activity, source_rse_id, dest_rse_id, session: "Op
 
 
 def get_supported_transfertools(rse_id: str, session=None) -> "Set[str]":
-    transfertool_attr = get_rse_attribute('transfertool', rse_id=rse_id, session=session)
-    if transfertool_attr:
-        result = set()
-        for attr in transfertool_attr:
-            if attr:
-                assert type(attr) == str
-                # split attribute values by comma
-                for transfertool in filter(bool, map(str.strip, attr.split(sep=','))):
-                    result.add(transfertool)
-        if result:
-            return result
-    return {'fts3', 'globus'}
+    transfertool_attr = get_rse_attribute(rse_id, 'transfertool', session=session)
+
+    if not transfertool_attr:
+        return {'fts3', 'globus'}
+
+    result = set()
+    for transfertool in transfertool_attr.split(sep=','):
+        if transfertool.strip():
+            result.add(transfertool)
+    return result
--- a/lib/rucio/core/rse.py
+++ b/lib/rucio/core/rse.py
@@ -19,9 +19,9 @@ import logging
 import traceback
 from io import StringIO
 from re import match
+from typing import Any, Dict, Optional, Sequence, Union
 
 import sqlalchemy
-import sqlalchemy.orm
 from dogpile.cache.api import NO_VALUE
 from sqlalchemy.exc import DatabaseError, IntegrityError, OperationalError
 from sqlalchemy.orm import aliased, Session
@@ -38,7 +38,6 @@ from rucio.db.sqla import models
 from rucio.db.sqla.constants import (RSEType, ReplicaState)
 from rucio.db.sqla.session import read_session, transactional_session, stream_session
 
-from typing import Any, Dict, Optional, Sequence
 
 REGION = make_region_memcached(expiration_time=900)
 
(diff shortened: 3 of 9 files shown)


lib/rucio/core/rse.py
```diff
@@ -40,9 +39,8 @@
 from rucio.db.sqla.constants import (RSEType, ReplicaState)
 from rucio.db.sqla.session import read_session, transactional_session, stream_session
 
-if TYPE_CHECKING:
-    from typing import Any, Dict, Optional, Sequence
-    from sqlalchemy.orm import Session
+from typing import Any, Dict, Optional, Sequence, Union
+from sqlalchemy.orm import Session
```
akselaase: These imports aren't sorted correctly. `typing` is a standard module and `sqlalchemy.orm` is already imported earlier.

lib/rucio/core/rse.py
```diff
@@ -662,39 +660,51 @@ def get_rses_with_attribute_value(key, value, lookup_key, vo='def', session=None
...
 
-    :returns: A list with RSE attribute values for a Key.
+    :returns: The value for the rse attribute, None if it does not exist.
     """
 
-    result = NO_VALUE
-    if use_cache:
-        result = REGION.get('%s-%s-%s' % (key, rse_id, value))
-    if result is NO_VALUE:
+    value = session.query(models.RSEAttrAssociation.value).filter_by(rse_id=rse_id, key=key).first()
```
akselaase: [`Query.one_or_none`](https://docs.sqlalchemy.org/en/14/orm/query.html#sqlalchemy.orm.Query.one_or_none) could be useful here instead of `Query.first`, depends on how you'd like to handle exceptions. This is supposed to be a primary key after all.

lib/rucio/core/rse.py
```diff
@@ -662,39 +660,51 @@ def get_rses_with_attribute_value(key, value, lookup_key, vo='def', session=None
...
+    cache_key = f'rse_attributes_{rse_id}_{key}'
+    value = REGION.get(cache_key)
+
+    if value is not NO_VALUE:
+        return value
+
+    value = get_rse_attribute_without_cache(rse_id, key, session=session)
+
+    if value is None:
+        return None
```
akselaase: Should `None` values also be cached?

lib/rucio/core/rse.py
```diff
@@ -725,7 +735,7 @@ def get_rse_supported_checksums(rse_id, session=None):
     """
     Retrieve from the DB and parse the RSE attribute defining the checksum supported by the RSE
     """
-    return parse_checksum_support_attribute(get_rse_attribute(key=CHECKSUM_KEY, rse_id=rse_id, session=session))
+    return parse_checksum_support_attribute([get_rse_attribute(rse_id, CHECKSUM_KEY, session=session)])
```
akselaase: Conflicts with https://github.com/rucio/rucio/pull/5669, but I can rebase that afterwards and fix this weird list-wrapping.

lib/rucio/core/rse.py
```diff
@@ -1126,9 +1135,9 @@ def get_rse_protocols(rse_id, schemes=None, session=None):
             'qos_class': _rse.qos_class,
             'rse': _rse.rse,
             'rse_type': _rse.rse_type.name,
-            'sign_url': sign_url[0] if sign_url else None,
+            'sign_url': sign_url if sign_url is not None else None,
```
akselaase: This can be reduced to just `sign_url`.

lib/rucio/core/request.py
```diff
@@ -1807,15 +1807,13 @@ def __throttler_request_state(activity, source_rse_id, dest_rse_id, session: "Op
...
-    return {'fts3', 'globus'}
+    transfertool_attr = get_rse_attribute(rse_id, 'transfertool', session=session)
+
+    if not transfertool_attr:
+        return {'fts3', 'globus'}
+
+    result = set()
+    # split attribute values by comma
+    for transfertool in filter(bool, map(str.strip, transfertool_attr.split(sep=','))):
+        result.add(transfertool)
```
akselaase: Could this be rewritten as this for readability?
```python3
result = set()
for transfertool in transfertool_attr.split(sep=','):
    if transfertool.strip():
        result.add(transfertool)
```

lib/rucio/core/replica.py
```diff
@@ -1111,10 +1109,10 @@ def _list_replicas(dataset_clause, file_clause, state_clause, show_pfns,
 
                         if 'site' in client_location and client_location['site']:
                             # is the RSE site-configured?
-                            rse_site_attr = get_rse_attribute('site', rse_id, session=session)
+                            rse_site_attr = get_rse_attribute(rse_id, 'site', session=session)
                             replica_site = ['']
```
akselaase: `replica_site` is defined as a list here, but reassigned as a string just below. Probably should be initialized to `None`?

akselaase (changes requested):

lib/rucio/core/rse.py
joeldierkes: I like it. Let's not handle the exception since it's a primary key.

lib/rucio/core/rse.py
joeldierkes: I don't think so. None indicates, that a value is missing. If it gets set, the value should be accessible immediately.

We could talk about accessing the cache also in `add_rse_attribute` and `del_rse_attribute` (and friends). This, however, would add more logic to the cache...

lib/rucio/core/rse.py
joeldierkes: Let's handle conflicts once they arrive. The one who gets merged first has priority.

akselaase (dismissed):

lib/rucio/core/rse.py
```diff
@@ -662,39 +658,51 @@ def get_rses_with_attribute_value(key, value, lookup_key, vo='def', session=None
 
 
 @read_session
-def get_rse_attribute(key, rse_id=None, value=None, use_cache=True, session=None):
+def get_rse_attribute_without_cache(rse_id: str, key: str, session: Optional[Session] = None) -> Optional[Union[str, bool]]:
```
mlassnig: I would suggest not to expose this function. This is purely for the unit tests and should not be used by anyone else.
Maybe prefix with `__` ?

akselaase: Should be prefixed with a single underscore in that case: https://docs.python.org/3/tutorial/classes.html#private-variables

mlassnig: ah, yes.

joeldierkes: This is a public function. It is called when the `rse_attribute` should be looked up directly, without the cache.

rcarpa (dismissed):
Overal I like the change very much. I wouldn't change the "use cache" behavior though. It's not the only place in rucio where we have such a "use_cache" argument. IMHO, i'd keep this behavior unchanged in this particular PR and (maybe) concentrate on this issue everywhere in another PR.

lib/rucio/core/rse.py
```diff
@@ -662,39 +658,51 @@ def get_rses_with_attribute_value(key, value, lookup_key, vo='def', session=None
...
+
+    if value is not NO_VALUE:
+        return value
+
+    value = get_rse_attribute_without_cache(rse_id, key, session=session)
+
+    if value is None:
+        return None
+
+    REGION.set(cache_key, value)
```
rcarpa: Does it mean the missing attribute will never be cached ? seems like a possibility for performance issues.

joeldierkes: Yes, I'm thinking of the use case: add an RSE Attribute if it does not exist. This would set the cache to None for some time. In that time, the new RSE attribute can not be used.

joeldierkes: I will change the behavior in favor of performance, `None` values will be cached for now. The use case does not seem to appear that often.

lib/rucio/core/rse.py
joeldierkes: Okay, after a talk with @rcarpa I decided to keep the `use_cache` parameter for consistency. I also changed the `get_rse_attribute_without_cache` to a private function.

lib/rucio/tests/test_rse.py
```diff
@@ -1479,3 +1480,24 @@ def test_rse_update_unsupported_option(self):
...
+def test_get_rse_attribute(rse_factory):
+    _, rse_id = rse_factory.make_mock_rse()
+
+    assert get_rse_attribute(rse_id, "test") is None
+
+    add_rse_attribute(rse_id, "test", "test")
+    assert get_rse_attribute(rse_id, "test", use_cache=False) == "test"
+
+    if ENABLE_CACHING:
+        assert get_rse_attribute(rse_id, "test") is None
```
akselaase: Is this assertion correct? You just set the "test" key to the value "test", wouldn't you expect to get the value "test" back as well?

joeldierkes: The expected behavior is different, the current one asserts that. I changed it to make it obvious that we want to change it in the future.

mlassnig (commented):
just commenting for `_list_replicas()`: the change there looks good

rcarpa (dismissed):

lib/rucio/tests/test_rse.py
```diff
@@ -1479,3 +1479,25 @@ def test_rse_update_unsupported_option(self):
         with pytest.raises(InputValidationError):
             update_rse(rse_id, parameters={'city': 'Not Berlin', 'non_existing_option': 3})
         assert get_rse(rse_id)['city'] == 'Berlin'
+
+
+@pytest.mark.parametrize("use_cache", [
+    False,
+    pytest.param(True, marks=pytest.mark.xfail(reason='FIXME: Calling functions which cachange the rse attribute should invalidate the cache.')),
```
rcarpa: small typo

rcarpa (dismissed):

akselaase (approved):

PR #5674, opened 2022-06-27
transfertool: Fix FTS3 `checksum_validation_strategy`; Closes #5670

### LINKED ISSUE #5670: Logic error in FTS3 transfertool checksum validation strategy
Motivation
----------
The function `checksum_validation_strategy` in `rucio/transfertool/fts3.py` determines which checksumming strategy to apply to a transfer. This function has some errors, for instance comparing the variables `source_supported_checksums` and `dest_supported_checksums` with `['none']`, instead of the correct value `[]`.

Another error is when the source supports checksumming, but the destination does not, which implies there are no common checksum algorithms. In this case case the checksum validation strategy is still selected as `destination`, and not `source`, despite the fact that the destination doesn't have any supported algorithms. This causes checksumming to be skipped on the source, even when it should have been performed.

Relevant source for context:

```python3
def checksum_validation_strategy(src_attributes, dst_attributes, logger):
    source_supported_checksums = get_rse_supported_checksums_from_attributes(src_attributes)
    dest_supported_checksums = get_rse_supported_checksums_from_attributes(dst_attributes)
    common_checksum_names = set(source_supported_checksums).intersection(dest_supported_checksums)

    ...

    if len(common_checksum_names) == 0:
        logger(logging.INFO, 'No common checksum method. Verifying destination only.')
        verify_checksum = 'destination'

    ...
```

Modification
------------
Determine the correct expected behaviour and rewrite `checksum_validation_strategy`.

akselaase:
The FTS3 transfertool wrapper incorrectly determined which checksumming algorithms to apply based on RSE attributes. In one case where the source RSE supported checksums, but the destination RSE did not, the determined strategy would be to perform no checksumming at all, instead of selecting the source as checksum target.

This PR simplifies the logic in this function considerably and adds test to verfiy behaviour.

### DIFF
--- a/lib/rucio/tests/transfertool/__init__.py
+++ b/lib/rucio/tests/transfertool/__init__.py
@@ -1,0 +1,14 @@
+# -*- coding: utf-8 -*-
+# Copyright European Organization for Nuclear Research (CERN) since 2012
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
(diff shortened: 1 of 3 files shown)


rcarpa:
@akselaase , nice catch for the bugs and very good initiative to fix them.

Unfortunately, when performing the review of your PR I realized just how broken is everything related to checksums and that your PR is far from being enough to fix them. I propose you to close this PR and I'll open another one which incorporates both your fix and many additional fixes I had to add to make it work.

rcarpa:
I submitted https://github.com/rucio/rucio/pull/5678, which re-uses most of your work, and fixes some other bugs. Closing this one.

Issue #5251, opened 2022-02-17
Adherence to rule grouping dependent on order of ATTACH


ericvaandering:
Motivation
----------
CMS is seeing that rules which use "CONTAINER" grouping are not always grouped at the same RSE. In fact, we have rules made at different times on the same data and the two classes behave differently.

This is from @nsmith- . 

If I look at the wmcore_output rules instead (using the rse expression (tier=2|tier=1)&cms_type=real&rse_type=DISK) I see only one out of 9000 rules is split: 12cc4f5e0dfb4809922498b474be16c3. I suspect a straggling file was added somehow after MSOutput made the rule, not sure if that's possible?

Since the US MiniAOD subscription-based rules are created as soon as the first file appears in the dataset, and the wmcore_output rules are only made once the dataset is complete (modulo the one exception) this means the culprit is probably in the daemon that updates rules when the DID is modified.
 
The relevant lines are
https://github.com/rucio/rucio/blob/88984a4dbc9d8be4e254f61545c7066e6c67de56/lib/rucio/core/rule.py#L2617-L2629
Looking at this it seems if a dataset is attached to a container and then a file is attached to a dataset, then the code will consider previous files, but if the file is attached to the dataset and then the dataset is attached to the container, this will not run and preferred_rse_ids will be left empty. Will try to construct an example to test.

Modification
------------

piperov:
This could be related to (or is the root cause of) the observed problems, reported recently by CMS users, of datasets not getting automatically transferred to RSEs, when being added to already existing/subscribed containers covered by active replication rules.

The problem is reproducibly manifested in the following scenario:

1. User creates a container with `rucio add-container `
2. User attaches a (number of) datasets with `rucio attach`
3. User creates a replication rule with `rucio add-rule --lifetime 2592000 --ask-approval ... RSE`
4. Request is approved.
4. Container gets replicated at the RSE
5. User attaches additional datasets to the container with `rucio attach`
6. New datasets do NOT get replicated at the RSE.

Expectation in last step is that the newly attached datasets automatically get replicated at the RSE.

ericvaandering:
@piperov In point #2 I presume you mean Containers, not datasets, in the Rucio terminology.

piperov:
@ericvaandering Correct. I use 'Datasets' in the CMS sense here. Proper Rucio term would be 'Container'.

bari12:
@piperov can you create a test-case for this please? In the end to me this just looks like
[...]

PR #5676, opened 2022-06-27
Rules: Added test_judge_dataset_grouping_all, mark it as XFAIL for no…

ivmfnal:
…w; Force parent rules re-evaluation if a dataset or a container is added to the parent container #5251

### DIFF
--- a/lib/rucio/tests/test_judge_evaluator.py
+++ b/lib/rucio/tests/test_judge_evaluator.py
@@ -138,6 +138,53 @@ class TestJudgeEvaluator(unittest.TestCase):
         dataset_locks = [lock for lock in get_dataset_locks(scope=scope, name=dataset)]
         assert(len(dataset_locks) == 2)
 
+    @pytest.mark.xfail(reason="This is a test for a known bug. See issue 5251")
+    @pytest.mark.noparallel(reason="uses mock scope and predefined RSEs; runs judge evaluator")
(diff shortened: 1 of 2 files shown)


mlassnig (dismissed):

lib/rucio/tests/test_judge_evaluator.py
```diff
@@ -138,6 +138,53 @@ def test_judge_add_dataset_to_container(self):
         dataset_locks = [lock for lock in get_dataset_locks(scope=scope, name=dataset)]
         assert(len(dataset_locks) == 2)
 
+    @pytest.mark.xfail(reason="This is a test for a known bug. See issue 5251")
```
rcarpa, line 141: Isn't your change from this same commit in dids.py fixing the bug ? Why is it still marked as xfail?

ivmfnal: no

bari12:
I know you don't want to hear this @ivmfnal but please rebase this 😄 There is a slight change in the code with the temporary tables and I didn't want to manually resolve this on merge myself, since it should be tested if there are any differences. Thanks

lib/rucio/tests/test_judge_evaluator.py
rcarpa: What's the point of adding a test then if it's not used? Also, I would have hoped for a longer explanation than "no"

PR #5678, opened 2022-06-29
Transfers: fix checksum validation in FTS. Closes #5670 

### LINKED ISSUE #5670: Logic error in FTS3 transfertool checksum validation strategy

rcarpa:
This PR build on top of the initial work by @akselaase in #5674. During the review I discovered multiple additional bugs, which are fixed in the additional commits.

### DIFF
--- a/lib/rucio/core/rse.py
+++ b/lib/rucio/core/rse.py
@@ -32,7 +32,7 @@ import rucio.core.account_counter
 from rucio.common import exception, utils
 from rucio.common.cache import make_region_memcached
 from rucio.common.config import get_lfn2pfn_algorithm_default
-from rucio.common.utils import CHECKSUM_KEY, Availability, is_checksum_valid, GLOBALLY_SUPPORTED_CHECKSUMS
+from rucio.common.utils import CHECKSUM_KEY, GLOBALLY_SUPPORTED_CHECKSUMS, Availability
 from rucio.core.rse_counter import add_counter, get_counter
 from rucio.db.sqla import models
 from rucio.db.sqla.constants import (RSEType, ReplicaState)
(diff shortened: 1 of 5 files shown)


lib/rucio/core/rse.py
```diff
@@ -749,29 +741,12 @@ def parse_checksum_support_attribute(checksum_attribute):
     if not checksum_attribute:
         return GLOBALLY_SUPPORTED_CHECKSUMS
     else:
-        supported_checksum_list = checksum_attribute[0].split(',')
+        supported_checksum_list = [c for c in checksum_attribute.split(',') if c]
```
mlassnig: maybe add `.strip()` ?

rcarpa: done

bari12:
Rebase please

bari12 (approved):

bari12:
Thanks @rcarpa and @akselaase 👍

PR #5683, opened 2022-06-29
Deletion: fixed oidc_account getting retrieved as bool #5682

### LINKED ISSUE #5682: oidc_account retrieved from config as bool in reaper

robbarnsley:
changed from config_get_bool to config_get, and default to root (since this account is the default account that has the IAM admin client sub)

### DIFF
--- a/lib/rucio/daemons/reaper/reaper.py
+++ b/lib/rucio/daemons/reaper/reaper.py
@@ -408,7 +408,7 @@ def reaper(rses, include_rses, exclude_rses, vos=None, chunk_size=100, once=Fals
     """
 
     executable = 'reaper'
-    oidc_account = config_get_bool('reaper', 'oidc_account', False, '')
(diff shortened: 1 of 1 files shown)


mlassnig (approved):

robbarnsley:
Just looking at this again now i'm changing poller also - i've slightly changed the behaviour here. Since i've set a valid default account, if the RSE supports OIDC, this:

https://github.com/rucio/rucio/blob/69a8540502b612e67c798c70b5221f52f57e1b07/lib/rucio/daemons/reaper/reaper.py#L612-L617

will now always attempt to proceed using tokens if the RSE supports it. Since `get_token_for_account_operation` is wrapped in a try/except, it will presumably fallback to X.509 (or whatever) if it fails, but it's still changed the order of preference for authentication somewhat.

The default should probably be `None` if we only want token authentication to go ahead when the `oidc_account` parameter has been explicitly set in the config.

Let me know if this is an issue.

Issue #5685, opened 2022-06-29
Introduce a heartbeat handler to kronos


maany:
As described in  #5443, refactor kronos to use @rcarpa's heartbeat handler

PR #5687, opened 2022-06-30
daemons: add heartbeat handler to kronos #5685

### LINKED ISSUE #5684: Introduce heartbeat handler to Kronos
Add a heartbeat handler to kronos as described in #5443

maany:


lib/rucio/daemons/tracer/kronos.py
```diff
@@ -341,87 +360,62 @@ def kronos_file(thread=0, dataset_queue=None, sleep_time=60):
...
-                    conn.connect()
-                conn.subscribe(destination=config_get('tracer-kronos', 'queue'), ack='client-individual', id=subscription_id, headers={'activemq.prefetchSize': prefetch_size})
-
-        tottime = time() - start_time
-        if tottime < sleep_time:
-            logger(logging.INFO, 'Will sleep for %s seconds' % (sleep_time - tottime))
-            sleep(sleep_time - tottime)
-
-    logger(logging.INFO, 'graceful stop requested')
+    conns = get_stomp_brokers(
```
rcarpa: you cannot assign, this way it will not return the values to the calling context. Either update the array in place, or (probably better) use a dict `return_values` and assign the conns to it. Something like return_values['conns'] = get_stomp....

rcarpa (approved):

PR #5690, opened 2022-07-01
webui: fix missing OIDC providers on login page #5689

tbeerman:


rizart:
FYI, this is already fixed in https://github.com/rucio/rucio/pull/5580.

PR #5693, opened 2022-07-01
Make the dogpile cache decorator usable

### LINKED ISSUE #5692: Make the dogpile cache decorator usable
Motivation
----------

`dogpile.cache` provides a decorator as an easy way to cache function results. This decorator would help us to reduce the complexity of functions and the number of code lines, while still being descriptive and easy to understand.

The default decorator only caches on all input arguments. This, however, won't let us cache database results, since we use the `..._session` decorator to provide us with a database session. This session is part of the signature, which would be included in the cache key.

We need a way to not include the session argument (or any arbitrary one) in the cache key.

The default key generator used by dogpile does not include keyword arguments in the generation. If a function only needs to construct the cache key based on the positional ones, the default function can be used.

Modification
------------

The syntax would be:

```python
@REGION.cache_on_arguments(function_key_generator=ignore_arguments_key_generator(["session"]))
@session
def get_rse_attribute(rse_id, key, *, session=None):
    ...
```

The `ignore_arguments_key_generator` provides away to select keyword arguments, which would not be included in the cache key. This would let us retrieve the same cache entries, even if the session changes.

joeldierkes:
See commit messages

### DIFF
--- a/lib/rucio/common/cache.py
+++ b/lib/rucio/common/cache.py
@@ -13,11 +13,20 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 from __future__ import absolute_import
+import inspect
 
 import logging
-from dogpile.cache import make_region
+from contextlib import contextmanager
+from datetime import timedelta
+from functools import wraps
+from typing import Callable, Optional, Union
+
+from dogpile.cache import CacheRegion, make_region
+from dogpile.cache import util as dogpile_util
+from dogpile.cache.backends.null import NullBackend
 
 from rucio.common.config import config_get
+from rucio.common.exception import InputValidationError
 from rucio.common.utils import is_client
 
 CACHE_URL = config_get('cache', 'url', False, '127.0.0.1:11211', check_config_table=False)
@@ -65,3 +74,110 @@ def make_region_memcached(expiration_time, function_key_generator=None):
         region.configure('dogpile.cache.null')
 
     return region
+
+
+def ignore_arguments_key_generator(namespace, fn: Callable, to_str: Callable[..., str] = str) -> Callable[..., str]:
+    """
+    This function is used to construct a
+    :py:func:`dogpile.cache.FunctionKeyGenerator` function. This function can be
+    used to cache function results based on all positional keys.
+
+    :param namespace: The namespace of the cache.
+    :param fn: The wrapped function.
+    :param to_str: The function to use to generate a string from the provided value.
+    :returns: A :py:func:`dogpile.cache.FunctionKeyGenerator` implementation.
+    """
+
+    dogpile_cache_key_generator = dogpile_util.function_key_generator(namespace, fn, to_str=to_str)
+
+    @wraps(fn)
+    def generate_key(*args, **ignored_kwargs):
+        """
+        Construct a cache-key wihtout the keyword arguments.
+        """
+        return dogpile_cache_key_generator(*args)
+
+    return generate_key
+
+
+"""
+The global :ref:class:dogpile.cache.region.CacheRegion: instance used to cache
+items.
+"""
+_REGION = make_region_memcached(
+    expiration_time=600,
+    function_key_generator=ignore_arguments_key_generator,
+)
+
+
+def cache(expiration_time_in_s: Optional[Union[float, timedelta]] = 600):
+    """
+    A function that provides a decorator for easy cache access. It creates a
+    :py:class:`dogpile.cache.region.CacheRegion` with the respective arguments.
+
+    Example::
+
+        @cache(600)
+        @session
+        def get_rse_attribute(rse_id, key, *, session=None):
+            ...
+
+    This example wraps the `get_rse_attribute` into a caching layer. The
+    expiration time of the items is 600s.
+
+    The decorator provides the same functions like the `cache_on_arguments`
+    dogpile cache decorator [1]. If, for example, a function wants to invalidate
+    a cache entry, it can call::
+
+        get_rse_attribute.invalidate(rse_id, key)
+
+    Important: Insert the `*` into the function signature before the keyword
+    arguments. If this is not done, the keyword arguments get wrongly translated
+    to positional ones, which creates a wrong cache key. Together with
+    pymemcache, that produces the error::
+
+        pymemcache.exceptions.MemcacheIllegalInputError: Key contains whitespace
+
+    If used with `{read,...}_session`, the `cache` decorator should be the outer
+    one. This prevent the creation of the session in case of a cache-hit.
+
+    [1] https://dogpilecache.sqlalchemy.org/en/latest/usage.html#using-a-region
+
+    :param expiration_time_in_s: The expiration time for the cached elements.
+    :returns: A decorator function to cache function results.
+    """
+
+    if callable(expiration_time_in_s):
+        raise InputValidationError("The decorator `cache` should be called with parenthesis, not without.")
+
+    def positional_arguments_with_default_check(fn):
+        def inner(*args, **kwargs):
+            # The input here gets forwardef to dogpile, the first argument being
+            # the function to decorate.
+            if inspect.getfullargspec(args[0]).defaults is not None:
+                raise InputValidationError("The `cache()` decorator only works with keyword-only arguments, it should not be called with positional arguments with defaults.")
+
+            return fn(*args, **kwargs)
+        return inner
+
+    result = positional_arguments_with_default_check(_REGION.cache_on_arguments(expiration_time=expiration_time_in_s))
+    return result
+
+
+@contextmanager
+def disable_cache() -> CacheRegion:
+    """
+    A context manager to disable the caching. After the context, the old caching
+    behavior will be restored, and all previously cached elements are present
+    again.
+
+    :returns: The global dogpile CacheRegion (with the NullBackend) used by
+              Rucio.
+    """
+    global _REGION
+    old_backend = _REGION.backend
+    _REGION.backend = NullBackend({})
+    try:
+        yield _REGION
+    finally:
+        _REGION.backend = old_backend
--- a/lib/rucio/db/sqla/session.py
+++ b/lib/rucio/db/sqla/session.py
@@ -17,7 +17,7 @@ import os
 import sys
 
 from functools import wraps
-from inspect import isgeneratorfunction
+from inspect import isgeneratorfunction, signature
 from retrying import retry
 from threading import Lock
 from os.path import basename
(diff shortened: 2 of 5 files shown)


lib/rucio/db/sqla/session.py
```diff
@@ -323,6 +323,7 @@ def new_funct(*args, **kwargs):
         except Exception:
             raise
     new_funct.__doc__ = function.__doc__
+    new_funct.__signature__ = signature(function)
```
akselaase, line 326: Is it possible to do this without `signature`? One idea I have is to use generic types like this, which maybe provides better IDE support:

```python3
from typing import TypeVar, Callable, cast

_F = TypeVar('_F', bound=Callable)

def read_session(function: _F) -> _F:
    ...
    def new_funct(...):
        ...
    return cast(_F, new_funct)
```

Edit: Saw the test for argspec further down, I'm guessing `dogpile` explicitly wants it this way then?

lib/rucio/common/cache.py
```diff
@@ -65,3 +68,66 @@ def make_region_memcached(expiration_time, function_key_generator=None):
...
+    generate cache-keys which ignores the specified keyword arguments.
+
+    This function can be used together with
+    :py:func:`dogpile.cache.region.CacheRegion.cache_on_arguments` to cache
+    function results based on some (not all) arguments::
+
+        @REGION.cache_on_arguments(function_key_generator=ignore_arguments_key_generator(["session"]))
+        @session
+        def get_rse_attribute(rse_id, key, *, session=None):
+            ...
```
akselaase: I feel like this is too bothersome to write. In this case I'd be in favor of another wrapper to simplify usage:

```python3
def cache(*, ignore: str | Iterable[str] = ''):
    ignored_arg_names: tuple[str, ...]
    if isinstance(ignore, str):
        ignored_arg_names = tuple(ignore.split(','))
    else:
        ignored_arg_names = tuple(ignore)
    return partial(
        REGION.cache_on_arguments, 
        function_key_generator=ignore_arguments_key_generator(ignored_arg_names))
```

Which can then be used as 

```python3
@cache(ignore='session')
def get_rse_attribute(..., session = None):
    ...
```

(note not tested, hopefully got number of partials and wrapper layers right)

lib/rucio/tests/test_cache.py
```diff
@@ -0,0 +1,153 @@
...
+
+no_ignored_arguments_function_key_generator = ignore_arguments_key_generator([])
+
+
+class TestBasicCacheKeyGenerator:
+    @pytest.mark.parametrize(
+        "left,str_comp_func,right",
+        [
+            ([], str.__eq__, []),
+            ([None], str.__eq__, [None]),
```
akselaase: Check out the `operator` module.

akselaase (commented):
Made some initial suggestions, can do more in depth when it's closer to ready.

lib/rucio/db/sqla/session.py
joeldierkes: Yes, dogpile is very explicit about that... xD
https://github.com/sqlalchemy/dogpile.cache/blob/1de93aab14c1274f20c1f44f8adff3b143c864f6/dogpile/util/compat.py#L33-L34

lib/rucio/tests/test_cache.py
joeldierkes: Change in functionality, this part got removed

lib/rucio/common/cache.py
```diff
@@ -65,3 +72,93 @@ def make_region_memcached(expiration_time, function_key_generator=None):
...
+    again.
+
+    :returns: The global dogpile CacheRegion (with the NullBackend) used by
+              Rucio.
+    """
+    global REGION
+    old_backend = REGION.backend
+    REGION.backend = NullBackend({})
+
+    yield REGION
```
akselaase: Wrap in `try/finally`

lib/rucio/tests/conftest.py
```diff
@@ -401,6 +403,9 @@ def file_config_mock(request):
 @pytest.fixture
 def caches_mock(request):
     """
+    Warning: This cache behavior is deprecated. We strongly encurage you to use
```
akselaase, line 406: Could be even clearer with a `warnings.warn(...)`, but that might be too noisy as well.

lib/rucio/common/cache.py
```diff
@@ -65,3 +72,93 @@ def make_region_memcached(expiration_time, function_key_generator=None):
...
+    pymemcache, that produces the error::
+
+        pymemcache.exceptions.MemcacheIllegalInputError: Key contains whitespace
+
+    [1] https://dogpilecache.sqlalchemy.org/en/latest/usage.html#using-a-region
+
+    :param expiration_time_in_s: The expiration time for the cached elements.
+    :returns: A decorator function to cache function results.
+    """
+
```
akselaase, line 149: Todo: check if the first argument is a callable, which implies the user wrote `@cache` instead of `@cache()`, and throw and exception if so.

lib/rucio/common/cache.py
```diff
@@ -65,3 +72,93 @@ def make_region_memcached(expiration_time, function_key_generator=None):
...
+        return dogpile_cache_key_generator(*args)
+
+    return generate_key
+
+
+"""
+The global :ref:class:dogpile.cache.region.CacheRegion: instance used to cache
+items.
+"""
+REGION = make_region_memcached(
```
akselaase: Mark private with a leading underscore.

lib/rucio/common/cache.py
```diff
@@ -65,3 +72,93 @@ def make_region_memcached(expiration_time, function_key_generator=None):
...
+    a cache entry, it can call::
+
+        get_rse_attribute.invalidate(rse_id, key)
+
+    Important: Insert the `*` into the function signature before the keyword
+    arguments. If this is not done, the keyword arguments get wrongly translated
+    to positional ones, which creates a wrong cache key. Together with
+    pymemcache, that produces the error::
+
+        pymemcache.exceptions.MemcacheIllegalInputError: Key contains whitespace
```
akselaase, line 139: Possible to fail-fast here and give the user a better error message?

lib/rucio/common/cache.py
```diff
@@ -65,3 +72,93 @@ def make_region_memcached(expiration_time, function_key_generator=None):
...
+items.
+"""
+REGION = make_region_memcached(
+    expiration_time=600,
+    function_key_generator=ignore_arguments_key_generator,
+)
+
+
+def cache(expiration_time_in_s: Optional[Union[float, timedelta]] = 600):
+    """
```
akselaase, line 114: Document that `@cache` should be the outermost decorator (wrt. `@read_session` for instance).

lib/rucio/tests/test_cache.py
```diff
@@ -0,0 +1,103 @@
...
+        assert cached_func_1(1, session="test") == 1
+
+    @pytest.mark.skipif(not ENABLE_CACHING, reason="The results are different if the cache is enabled.")
+    def test_invalidate(self):
+        @cache(600)
+        def cached_func_2(x, *, y=0):
+            return x + y
+
+        # Clean-up from previous runs
+        cached_func_2.invalidate(1)
```
akselaase: Is this necessary?

lib/rucio/tests/test_cache.py
```diff
@@ -0,0 +1,103 @@
...
+
+        assert cached_func_3() == 0
+
+        time.sleep(0.05)
+        assert cached_func_3() == 0
+
+        time.sleep(0.06)
+        assert cached_func_3() == 1
+
+    @pytest.mark.skipif(not ENABLE_CACHING, reason="The results are different if the cache is enabled.")
```
akselaase: "if the cache is disabled." I guess is the right wording here.

lib/rucio/tests/test_cache.py
```diff
@@ -0,0 +1,103 @@
...
+        with disable_cache():
+            assert cached_func_4() == 1
+            assert cached_func_4() == 2
+
+        assert cached_func_4() == 0
+
+    def test_disable_cache_fixture(self, disable_cache_fixture):
+        number_called = count(0)
+
+        @cache(1)
```
akselaase, line 93: Looking forward to the day the test fails because the computer is overloaded and suspends the test for a second before continuing.

lib/rucio/tests/test_cache.py
```diff
@@ -0,0 +1,103 @@
...
+        def cached_func_3():
+            return next(number_called)
+
+        # Clean-up from previous runs
+        cached_func_3.invalidate()
+
+        assert cached_func_3() == 0
+
+        time.sleep(0.05)
+        assert cached_func_3() == 0
```
akselaase, line 64: Same here, you're optimistic about performance! Honestly don't know of a better way to do this though than waiting longer, which also sucks...

akselaase (changes requested):
Good job, you'll get there eventually!

lib/rucio/tests/test_cache.py
joeldierkes: Same here, maybe a separate pytest mark (`uses_sleep`, `tests_via_time`, `this_sucks`, ...) to signal the dependence on the execution time?

lib/rucio/tests/test_cache.py
joeldierkes: Yes, running this multiple times will not work. There is no convenient way to clear the entire cache ... yet.

lib/rucio/tests/conftest.py
joeldierkes: I think that's too noisy as well...

lib/rucio/tests/test_cache.py
joeldierkes: `!!`

lib/rucio/common/cache.py
```diff
@@ -65,3 +74,110 @@ def make_region_memcached(expiration_time, function_key_generator=None):
...
+
+
+"""
+The global :ref:class:dogpile.cache.region.CacheRegion: instance used to cache
+items.
+"""
+_REGION = make_region_memcached(
+    expiration_time=600,
+    function_key_generator=ignore_arguments_key_generator,
+)
```
akselaase, line 110: Should this really be created at import time? Other options are to create it on first use, or to make a dedicated `setup_cache` function to initialize (then you could for instance use parameters from `rucio.cfg`).

rcarpa (commented):
LGTM.

The PR doesn't solve any problem which bothers me personally, but doesn't complicate my life either. So i'm completely neutral.

PR #5694, opened 2022-07-01
Transfers: added oidc_account to poller #5457

### LINKED ISSUE #5457: conveyor-poller does not attempt to use tokens for authentication with FTS
We're not able to get conveyor-poller to use tokens to poll FTS. Submitter works fine.

Had a cursory look at the code, and can't seem to find where [oidc_account ](https://github.com/rucio/rucio/blob/master/lib/rucio/transfertool/fts3.py#L634) is set in FTS3TransferTool for the polling flow. For submitter, it seems like it's set [here ](https://github.com/rucio/rucio/blob/release-1.28/lib/rucio/transfertool/fts3.py#L714)(called by [next_transfers_to_submit](https://github.com/rucio/rucio/blob/release-1.28/lib/rucio/daemons/conveyor/submitter.py#L65)).

Without this variable being set, polling falls back to x509 authentication with FTS. This is obviously not compatible with our goal to have an end-to-end token-only workflow.

robbarnsley:

### DIFF
--- a/lib/rucio/daemons/conveyor/poller.py
+++ b/lib/rucio/daemons/conveyor/poller.py
@@ -34,6 +34,7 @@ from sqlalchemy.exc import DatabaseError
 import rucio.db.sqla.util
 from rucio.common.config import config_get, config_get_bool
 from rucio.common.exception import DatabaseException, TransferToolTimeout, TransferToolWrongAnswer
+from rucio.common.types import InternalAccount
 from rucio.common.logging import setup_logging
 from rucio.common.utils import dict_chunks
 from rucio.core import transfer as transfer_core, request as request_core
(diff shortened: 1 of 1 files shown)


lib/rucio/daemons/conveyor/poller.py
```diff
@@ -95,7 +96,9 @@ def run_once(fts_bulk, db_bulk, older_than, activity_shares, multi_vo, timeout,
                 if TRANSFER_TOOL == 'globus':
                     transfertool_obj = GlobusTransferTool(external_host=None)
                 else:
-                    transfertool_obj = FTS3Transfertool(external_host=external_host, vo=vo)
+                    if oidc_account:
+                        account = InternalAccount(oidc_account, vo=vo)
+                    transfertool_obj = FTS3Transfertool(external_host=external_host, vo=vo, oidc_account=account)
```
rcarpa, line 102: the `account` variable may be accessed before assignment. The tests will fail for this PR

rcarpa: Other than the comment above, looks good to me

robbarnsley:
As I mentioned in #5649, I really think it would be prudent to be more judicious about type annotations.

I'm now failing checks for adding an argument to an existing function that doesn't have annotations. I don't really know much about this function, or what it accepts apart from the argument i've added, so i'd have to trawl the code to make sure I fully understood it before I felt comfortable adding annotations for the rest of the arguments.

bari12:
I explained this a bit in my other comment, but I get your point for this too; I would recommend to just add the annotation for the newly added parameter. The test will still fail, but we can ignore it in that case.

rcarpa (approved):

PR #5701, opened 2022-07-05
Core & Internals: Introduce a heartbeat handler to oauthmanager #5695

### LINKED ISSUE #5695: Introduce a heartbeat handler to oauthmanager
Motivation
----------
Part of #5443.

Modification
------------
Introduce a heartbeat handler to oauthmanager.

rizart:


lib/rucio/daemons/oauthmanager/oauthmanager.py
```diff
@@ -38,140 +38,139 @@
...
     """
-    if sleep_time == OAuthManager.__defaults__[3] and loop_rate != OAuthManager.__defaults__[1]:
-        sleep_time = loop_rate
 
-    executable = 'oauth-manager'
+    run_daemon(
+        once=once,
+        graceful_stop=graceful_stop,
+        executable='oauth-manager',
+        logger_prefix='OAuthManager',
```
rcarpa: unless there is a particular reason to have it different, I would just set `executable` = `logger_prefix` = 'oauth-manager'. Especially, I see that the logger prefix is `oauth_manager` in the code, so I don't understand why 'OAuthManager' here?

lib/rucio/daemons/oauthmanager/oauthmanager.py
```diff
@@ -183,22 +182,21 @@ def run(once=False, threads=1, loop_rate=300, max_rows=100, sleep_time=300):
     sanity_check(executable='OAuthManager', hostname=socket.gethostname())
```
rcarpa: this doesn't seem right. executable should be the same everywhere. Also, run_daemon actually does a sanity_check, so I would just remove this line.

lib/rucio/daemons/oauthmanager/oauthmanager.py
```diff
@@ -38,140 +38,139 @@
...
-            elif match('.*ORA-03135.*', str(err.args[0])):
-                logger(logging.WARNING, traceback.format_exc())
-                record_counter('oauth_manager.exceptions.{exception}', labels={'exception': err.__class__.__name__})
-            else:
-                logger(logging.CRITICAL, traceback.format_exc())
-                record_counter('oauth_manager.exceptions.{exception}', labels={'exception': err.__class__.__name__})
-        except Exception as err:
-            logger(logging.CRITICAL, traceback.format_exc())
-            record_counter('oauth_manager.exceptions.{exception}', labels={'exception': err.__class__.__name__})
+    graceful_stop.wait(1)
```
rcarpa: this is not needed anymore. Remove this line. It is done by run_daemon

lib/rucio/daemons/oauthmanager/oauthmanager.py
```diff
@@ -38,140 +38,139 @@
...
-                logger(logging.CRITICAL, traceback.format_exc())
-                record_counter('oauth_manager.exceptions.{exception}', labels={'exception': err.__class__.__name__})
-        except Exception as err:
+    try:
+        # ACCESS TOKEN REFRESH - better to run first (in case some of the refreshed tokens needed deletion after this step)
+
+        # make a heartbeat
+        worker_number, total_workers, logger = heartbeat_handler.live()
+
+        logger(logging.INFO, '----- START ----- ACCESS TOKEN REFRESH ----- ')
```
rcarpa, line 94: This is unrelated to this PR. So you can leave it unchanged, but splitting one log message into two separate lines like that is not a good idea. It may improve the readability of logs for people who look at bare logs, but it makes things worse for everybody who uses a proper log collection pipeline and then searches for logs in elastic-search for example.

lib/rucio/daemons/oauthmanager/oauthmanager.py
```diff
@@ -38,140 +38,139 @@
...
         else:
-            daemon_sleep(start_time=start, sleep_time=sleep_time, graceful_stop=GRACEFUL_STOP)
+            logger(logging.CRITICAL, traceback.format_exc())
+            record_counter('oauth_manager.exceptions.{exception}', labels={'exception': err.__class__.__name__})
 
-    die(executable=executable, hostname=socket.gethostname(), pid=os.getpid(), thread=threading.current_thread())
-    logger(logging.INFO, 'graceful stop done')
+    tottime = time.time() - start
+    logger(logging.INFO, 'took %f seconds to delete %i tokens, %i session parameters and refreshed %i tokens', tottime, ndeleted, ndeletedreq, nrefreshed)
+    record_timer(name='oauth_manager.duration', time=1000 * tottime)
```
rcarpa: record_timer doesn't exist anymore. Please rebase the PR to use the changed behavior from https://github.com/rucio/rucio/pull/5718.

rcarpa (changes requested):

lib/rucio/daemons/oauthmanager/oauthmanager.py
rizart: No reason, just kept the convention that was already in place, will change.

lib/rucio/daemons/oauthmanager/oauthmanager.py
rizart: What you say is true, this again I left untouched, I would probably assign this logging refactoring to a separate issue/PR.

rcarpa (approved):

Issue #5700, opened 2022-07-05
WebDAV protocol XML parser doesn't handle XML namespaces correctly


akselaase:
Motivation
----------
The XML response parser in the WebDAV implementation does not properly handle resolution of different XML namespaces. 

[RFC 4918](http://www.webdav.org/specs/rfc4918.html) specifies that the response to a PROPFIND request will use tags in the `DAV:` XML namespace, which is usually shortened to just `D:` by setting `xmlns:D="DAV:"` on the `D:multistatus` tag. The XML parsing in `webdav.py` depends on this `D:` alias, which causes it to mishandle responses from for instance the Apache DAV module which often uses `lp1:` instead by setting `xmlns:lp1="DAV:"`.

Issue occurred when working on #5308 due to trying to test with an Apache WebDAV server.

Modification
------------
Use Python's `xml.etree.elementtree` to properly resolve namespaces and traverse the DAV response, instead of comparing raw string values.

Issue #5308, opened 2022-03-02
webdav.py Default.stat() module: Match error when finding size of current file


cfenell:
Motivation
----------
The stat method tries to find the reported size of the current file in the list returned by the PROPFIND request in this way:

            parser.feed(result.text)
            for file_name in parser.sizes:
                if '%s%s' % (self.server, file_name) == path:
                    dict_['size'] = parser.sizes[file_name]
            parser.close()
            return dict_

However, for me  self_server ends with PREFIX and file_name starts with PREFIX, so the match will always fail unless one PREFIX is stripped. As an example this should match:
     if "https://store1.eiscat3d.se:2880/home/tester/home/tester/root.EISCAT/dd/f0/file2.bin.rucio.upload"== "https://store1.eiscat3d.se:2880/home/tester/root.EISCAT/dd/f0/file2.bin.rucio.upload"

Modification
------------
self.server should probably be only https://hostname:port without the /prefix/ part.

PR #5706, opened 2022-07-06
Protocols: Migrate WebDAV XML parsing to Python's ElementTree

akselaase:
As described in #5700, the parser for the WebDAV server responses had issues with namespace handling. This caused Rucio to misinterpret responses from for instance the Apache DAV module. Changing the implementation to use Python's `xml.etree.ElementTree` fixes this issue.

This PR also includes a fix for #5308, in which Rucio doesn't properly handle DAV servers with path prefixes.

### DIFF
--- a/lib/rucio/rse/protocols/webdav.py
+++ b/lib/rucio/rse/protocols/webdav.py
@@ -15,9 +15,10 @@
 
 import os
 import sys
+from typing import Optional, Tuple
+from dataclasses import dataclass
 
 import xml.etree.ElementTree as ET
-from xml.parsers import expat
 
 import requests
 from requests.adapters import HTTPAdapter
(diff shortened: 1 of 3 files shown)


setuputil.py
```diff
@@ -18,6 +18,12 @@
 
 from pkg_resources import safe_name, parse_requirements
 
+_extra_client_requirements = []
+_extra_server_requirements = []
+
+if sys.version_info[:2] == (3, 6):
+    _extra_client_requirements.append('dataclasses')
+    _extra_server_requirements.append('dataclasses')
```
akselaase: TODO: rename client->common and don't append to server list.

joeldierkes: Info: We use markers in the requirements file to include packages in specific environments: https://github.com/rucio/rucio/commit/c301af509ba7eed5e34b644a68db39fcc2fa7553 : e.g. the mock package that got removed.

This work because we parse the requirements file with the `pkg_resources` package. [https://setuptools.pypa.io/en/latest/pkg_resources.html](https://setuptools.pypa.io/en/latest/pkg_resources.html#requirements-parsing)

As far as I can tell, you don't need to introduce the different lists. All elements in the install_requires of the client are also in the server.
```python
server_requirements_table = {
    'install_requires': clients_requirements_table['install_requires'] + [
```

lib/rucio/rse/protocols/webdav.py
```diff
@@ -89,56 +90,63 @@ def __len__(self):
...
-            self.dict[self.href] = data
-        if self.size:
-            self.sizes[self.href] = data
+        xml_size = node.find('./{DAV:}propstat/{DAV:}prop/{DAV:}getcontentlength')
+        if xml_size is None:
+            size = None
+        else:
+            size = int(xml_size.text)
+
+        return _PropfindFile(href=href, size=size)
```
joeldierkes: How about using a `classmethod`?

lib/rucio/rse/protocols/webdav.py
```diff
@@ -89,56 +90,63 @@ def __len__(self):
...
+
+
+@dataclass(frozen=True)
+class _PropfindResponse:
+    """Contains all the files from a PROPFIND response."""
+
+    files: Tuple[_PropfindFile]
+
+    @staticmethod
```
joeldierkes: +1

lib/rucio/rse/protocols/webdav.py
```diff
@@ -483,12 +496,28 @@ def stat(self, path):
...
+
+                if file.size is None:
+                    continue
+
+                dict_['filesize'] = file.size
+                break
+
+            else:
+                raise exception.ServiceUnavailable("WebDAV response didn't include content length for requested path.")
+
```
joeldierkes: As far as I see it this does not has to be inside the outer try?

lib/rucio/rse/protocols/webdav.py
```diff
@@ -89,56 +90,63 @@ def __len__(self):
...
 
 
-class Parser:
+@dataclass(frozen=True)
+class _PropfindFile:
+    """Contains the properties of one file from a PROPFIND response."""
 
-    """ Parser to parse XML output for PROPFIND ."""
+    href: Optional[str]
+    size: Optional[int]
```
joeldierkes: Is there a use case where we recieve one of `href`, `size`, but not the other?

From my perspective these are tighly coulped, I'd rather see a `None` there opposed to `_PropfindFile(None, 22)` or `_PropfindFile(path, None)`, which both does not make sense?

At least `href` should not be Optional...

requirements.txt
```diff
@@ -4,6 +4,7 @@ urllib3>=1.24.2,<=1.26.8                                    # HTTP library with
 dogpile.cache>=1.1.2,<=1.1.5                                # Caching API plugins (1.1.2 is the first version to support pymemcache)
 tabulate~=0.8.0                                             # Pretty-print tabular data
 jsonschema~=3.2.0                                           # For JSON schema validation (Policy modules)
+dataclasses==0.8.0; python_version == '3.6'                  # For dataclass support in 3.6
```
joeldierkes: Comment alignment, plus the comment itself is useless. I am not in favor of documenting it because everyhting else is when it adds not value.

joeldierkes (changes requested):
It's an _interesting_ PR...

Small nits.

I'd also propose to split the commits into two:
- Introducing dataclasses
- Things related to the webdav stuff

That way we can revert the webdav stuff in case of an error, without removing the precious dataclasses.

lib/rucio/rse/protocols/webdav.py
akselaase: It's just because I can't guarantee what the WebDAV server responds with. I agree that `href` could be made mandatory (it is in the RFC at least), so I'll change it to throw an exception if it is missing from the response.

joeldierkes (approved):

bari12 (approved):
👍 Looks good. I will only include it in 1.30 though, due to the added dependency.

Issue #5688, opened 2022-07-01
fix prometheus timer recording


rcarpa:
Motivation
----------
To record timers in prometheus, we use [Histograms](https://prometheus.io/docs/concepts/metric_types/#histogram), which are basically a set of counters with pre-defined buckets. In order for them to be useful, the buckets must be correctly set. The promtheus library uses some values by default https://github.com/prometheus/client_python/blob/5a5261dd45d65914b5e3d8225b94d6e0578882f3/prometheus_client/metrics.py#L544 

The problem is that rucio is not consistent and registers sometimes milisseconds in the timer. So all values above 10ms are collected into the "infinite" bucket, making the histogram useless for us. 

Modification
------------
We must go timer by timer and ensure that it registers correct values into it (seconds?). Alternatively, we can set non-default buckets on desired timers.

PR #5718, opened 2022-07-08
Ensure consistent units in Prometheus timer monitoring 

akselaase:
This PR aims to make the timer monitoring consistent with regards to recording elapsed seconds vs. milliseconds (#5688). A new `Stopwatch` class is introduced which handles the actual timing, and a new `Timer` class utilizes this stopwatch to replace `record_timer_block` and simplify usage of `record_timer`.

All previous uses of `record_timer` and `record_timer_block` are then checked and refactored if needed to ensure consistent units.

The PR also disables the flake8 warning `ANN101: missing type annotation for self in method`. The `self` argument is not usually typed except in [special cases](https://peps.python.org/pep-0484/#annotating-instance-and-class-methods), so this warning mostly produces noise.

### DIFF
--- a/lib/rucio/core/monitor.py
+++ b/lib/rucio/core/monitor.py
@@ -25,6 +25,7 @@ import time
 from abc import abstractmethod
 from datetime import datetime, timedelta
 from pathlib import Path
+from typing import Dict, List, Optional, Tuple, Union
 from retrying import retry
 from threading import Lock
 
@@ -238,12 +239,12 @@ def record_gauge(name, value, labels=None):
         gauge.set(value)
 
 
-def record_timer(name, time, labels=None):
+def record_timer(name: str, time: float, labels: Optional[Dict] = None):
     """
      Log timing information for a single stat (in miliseconds)
 
     :param name: The name of the stat to be updated.
-    :param time: The time to log.
+    :param time: The time (in seconds) to log.
     :param labels: labels used to parametrize the metric
     """
     timing = TIMINGS.get(name)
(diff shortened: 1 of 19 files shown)


lib/rucio/core/oidc.py
```diff
@@ -298,7 +298,7 @@ def get_auth_oidc(account: str, session=None, **kwargs) -> str:
             return None
 
     try:
-        start = time.time()
+        timer = Stopwatch()
```
rcarpa: This is probably just a "variable name" issue, but having a "stopwatch" with the name "timer" and having a "record" method on a stopwatch confuses my brain. 

i would rather have:
```
stopwatch = Stopwatch()
record_timer('....', stopwatch) 
```
or try to find a different name for the stopwatch. 

Alternatively, I didn't think it through very well, but a full-featured "Timer?/TimerHandler?/TimerRecoreder?/TimerManager?" (pick the name) class may also be interesting. One which has a higher level interface and allows to substitute both record_timer and record_timer_block. And also implements the stopwatch internally. One class to rule them (timers) all

lib/rucio/core/monitor.py
```diff
@@ -256,12 +257,56 @@ def record_timer(name, time, labels=None):
...
+    @property
+    def elapsed(self) -> float:
+        """Returns the total number of elapsed seconds."""
+        if self._t_end is None:
+            return self._now() - self._t_start
+        else:
+            return self._t_end - self._t_start
+
+
 class record_timer_block(object):
```
rcarpa: I just looked it up and `record_timer_block` seems to never be called with a "divisor". Maybe it's time to simplify it and get rid of the complex divisor logic in it ?

rcarpa (dismissed):
it's already an improvement over what we have. The other comments are just nitpicking. So good for me

lib/rucio/core/monitor.py
akselaase: It's used with a divisor some places, but never with multiple stats at once, so I'll simplify that.

rcarpa: Where is it used with divisor? I only saw a couple of tests and no actual application code using it

akselaase: You were right, the only usages were the ones I introduced in this PR. Removed and simplified now.

lib/rucio/core/oidc.py
akselaase: Tried another approach now with a single unified `Timer` class (and extracted the stopwatch-functionality into a pure `Stopwatch` class in `common/`).

`Timer` works both with specifying the statistic name on construction and when calling `record`. That might be confusing, but is the only way AFAIK to get it to work both inline and as a context manager.

rcarpa (dismissed):

lib/rucio/core/monitor.py
```diff
@@ -238,58 +249,122 @@ def record_gauge(name, value, labels=None):
...
     :param labels: labels used to parametrize the metric
+    :param buckets: Optional iterable of histogram bucket separators.
     """
-    timing = TIMINGS.get(name)
-    if not timing:
-        TIMINGS[name] = timing = MultiTiming(statsd=name, labelnames=labels.keys() if labels else ())
+    if name not in HISTOGRAMS:
+        # TODO: shouldn't the dictionary key really be (name, labels.keys(), buckets)?
+        # do we need the dictionary at all, or could we re-construct this class each time its needed?
+        HISTOGRAMS[name] = MultiTiming(statsd=name, labelnames=labels.keys() if labels else (), buckets=buckets)
```
akselaase: I'm not quite sure why the `MultiTiming`s are cached in this dictionary, but to ensure correctness the cache key should probably be extended to include the other attributes as well, right?

rcarpa: Prometheus doesn't allow duplicate metrics for the same name in a collector. The only reason to have this dictionary is because you can have multiple threads executing the same record_timer() and trying to create the same metric multiple time. 

Not sure how/if it will work when provided different labelnames, but it's not necessarily something we need (and it would complicate things on prometheus side anyway). In principle, we shouldn't have the same metric name recorded from two differents parts of the code. 

If you have good ideas on how to improve this workflow, don't hesitate.

akselaase: Ah okay, and it's not like the buckets will change from call to call either. Thanks for the explanation :+1:

rcarpa (approved):

Issue #5722, opened 2022-07-11
Refactor the query result to dict functionality


joeldierkes:
Motivation
----------
Multiple constructs like this
```python
dic = {}
for column in item.__table__.columns:
    dic[column.name] = getattr(res, column.name)
return dic
```
exist in the code base. They convert the query result `item` to a dictionary. This has some advantages, e.g. not altering attributes of the item `item` in the database, which is done by default for a result object.

Modification
------------
Refactor all occurences of these constructs and use a function. This prevents mistakes and let's us test the functionality.

Explicitly calling the `dict` function on the item works as well, however, sqlalchemy adds it's own attributes too (for now `_sa_instance_state`). To prevent adding these and possibly other keys in the future, we will iterate over the expected items by the model (like in the example).

PR #5723, opened 2022-07-11
Database: Refactor the query result to dict functionality #5722

joeldierkes:
Sqlalchemy query results integrate some functionalities to communicate
changes to the database (e.g. update fields when they are updated in the
code). While this is desired and helpful, it is not always needed. If a
dict is passed to an external dependency or returned as write only, the
communication with the database is not needed.

Converting a query result to a dict is not trivial. Calling the `dict`
function on the result returns not just the desired fields, but also
some custom ones from `sqlalchemy` (for now `_sa_instance_state`, maybe
more in the future). To only get the desired ones, iterate over the
model fields of the result and only append them to the resulting dict.

### DIFF
--- a/lib/rucio/db/sqla/util.py
+++ b/lib/rucio/db/sqla/util.py
@@ -17,7 +17,7 @@ import logging
 from datetime import datetime
 from hashlib import sha256
 from os import urandom
-from typing import TYPE_CHECKING
+from typing import Any, Dict, Iterable, Union
 
 import sqlalchemy
 from alembic import command, op
@@ -25,8 +25,10 @@ from alembic.config import Config
 from dogpile.cache.api import NoValue
 from sqlalchemy import func, inspect, Column, PrimaryKeyConstraint
 from sqlalchemy.dialects.postgresql.base import PGInspector
+from sqlalchemy.engine import Inspector, Result
 from sqlalchemy.exc import IntegrityError, DatabaseError
 from sqlalchemy.ext.declarative import declarative_base
+from sqlalchemy.orm import Session, Query
 from sqlalchemy.schema import CreateSchema, MetaData, Table, CreateTable, DropTable, ForeignKeyConstraint, DropConstraint
 from sqlalchemy.sql.ddl import DropSchema
 from sqlalchemy.sql.expression import select, text
@@ -37,16 +39,10 @@ from rucio.common.config import config_get, config_get_list
 from rucio.common.schema import get_schema_value
 from rucio.common.types import InternalAccount
 from rucio.common.utils import generate_uuid
-from rucio.core.account_counter import create_counters_for_new_account
 from rucio.db.sqla import models
 from rucio.db.sqla.constants import AccountStatus, AccountType, IdentityType
 from rucio.db.sqla.session import get_engine, get_session, get_dump_engine
 from rucio.db.sqla.types import InternalScopeString, String
-
-if TYPE_CHECKING:
-    from typing import Optional, Union  # noqa: F401
-    from sqlalchemy.orm import Session  # noqa: F401
-    from sqlalchemy.engine import Inspector  # noqa: F401
 
 REGION = make_region_memcached(expiration_time=600, memcached_expire_time=3660)
 
@@ -202,6 +198,9 @@ def create_root_account(create_counters=True):
 
     # Account counters
     if create_counters:
+        # Importing rucio core creates a cyclic dependency issue. We should only
+        # import it when we really need it.
+        from rucio.core.account_counter import create_counters_for_new_account
         create_counters_for_new_account(account=account.account, session=s)
 
     # Apply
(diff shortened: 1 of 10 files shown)


lib/rucio/tests/test_db.py
```diff
@@ -36,3 +38,50 @@ def test_config_poolclass():
 
     with pytest.raises(InputValidationError, match='Unknown poolclass: unknown'):
         _get_engine_poolclass('unknown')
+
+
+def test_result_to_dict_contains_all_keys(db_session, rse_factory):
```
rcarpa, line 43: I always was an opponent of these low-level tests of the internal functions. Their usefulness is discutable because, on each simplest modification of the code, you have to change the tests too. And this ends with introduction of bugs both in tests an in code without anybody noticing. 

Moreover, in this particular case, you rely on `models.SomeModel.__table__.columns` on both sides of the assertion equality. This is almost the same as running `assert some_function() == some_function()`, which is useless.  

I would argue that we don't need these test at all. But I'll not fight at all for removing them if others consider them useful.

joeldierkes: Yes and no. While I see your arguments, we do test more that that. In that specific case:
- The use cases, e.g.: an empty query should return nothing, a query that returns one element should return one element. This is also a form of documentation.
- The signature of the function. This function can be seen as an interface in utils (multiple other functions call that), we therefore should test it.

And I rather have a faililng test case than not noticing, that one of the use case chages. I'd argue that the rest of our code base is not tested enough to notize _some_ changes. Some other thoughs?

lib/rucio/db/sqla/util.py
```diff
@@ -514,3 +513,31 @@ def temp_table_mngr(session: "Session") -> TempTableManager:
...
+
+    `dict(res)` would be correct too, however, sqlalchemy adds it's own keys
+    (for now `_sa_instance_state`). To not add unnecessary keys, we iterate over
+    the expected ones and add them to the result.
+
+    :param res: One result of a sqlalchemy query.
+    :returns: A dicitonary with the items of the res parameter.
+    """
+    dic = {}
+    for column in res.__table__.columns:
```
rcarpa: did you check if https://docs.sqlalchemy.org/en/14/core/connections.html#sqlalchemy.engine.Row._asdict 

can be doing exactly what we want ?

joeldierkes: Yes, if we can trust StackOverflow there are multiple methods of doing this. The `_asdict` only works on queries with selected fields (again, source is StackOverflow): https://stackoverflow.com/a/21029850. We want a generic solution for all queries...

This is the reason why I'm introcuding the funtion and don't just use an already provided method...

rcarpa: Just went deeper into it and you are right.

lib/rucio/tests/test_db.py
rcarpa: My opinion is that we must test that `did_client.get_did()` contains whatever result we want to contain. The fact that it uses your new function behind the hood automatically means your function is also tested. If tomorrow I refactor the code and use something else and not your function to generate the result for our clients, the result must not change and I must not write/change the tests for that. But it's ok to keep the test. This is a very "IMHO" comment.

rcarpa:
I just discovered this (I didn't know it existed until today :D ) : 

https://github.com/rucio/rucio/blob/46dc0d03dccdd8d676e2c3af64146cea968d3fa8/lib/rucio/db/sqla/models.py#L271

maybe worth integrating your change into this function an use it ? it's basically an as_dict() function which is present on all our models via inheritance. It's almost never used so slightly changing it's behavior and only returning the columns is probably a good idea? What do you think ?

joeldierkes:
While we have these in our models, it is not present on arbitrary queries. What about `session.query(models.RSE.id)`. Where do you want to get the information about the available fields from? What if we make a join and have more than one model?

rcarpa:
> While we have these in our models, it is not present on arbitrary queries. What about `session.query(models.RSE.id)`. Where do you want to get the information about the available fields from? What if we make a join and have more than one model?

Is this a problem we actually have and actually want to solve? to convert arbitrary queries to an output without any additional logic and return it to an external caller ?

akselaase (approved):

joeldierkes:
Yes, `get_rses_with_attribute` and `get_replica` for example.

Right now they use the same logic (which is used in other places too). This should be defined in one place then, so we don't repeat ourselfes.

rcarpa:
> Yes, `get_rses_with_attribute` and `get_replica` for example.
> 
> Right now they use the same logic (which is used in other places too). This should be defined in one place then, so we don't repeat ourselfes.

One queries `models.RSE`, the other `models.RSEFileAssociation`. I don't see how these two are examples to confirm that we need to solve the problem you mentioned earlier.

bari12:
New PR coming with limited scope

PR #5735, opened 2022-07-14
Documentation: Rename no_doc parameter #4992

joeldierkes:
Flask differentiates between calls with and without a trailing slash in the
url. We have some Rest Api Endpoints (e.g. `/<account>/usage/`), where it is
important to differentiate. To not include these calls in the docs twice, the
`no_doc` parameter was introduced. It is `True` when the docs should not be
generated, `False` otherwise.

Since the parameter contains a negation in the word itself, it is hard to
understand what behavior is meant when the parameter is set. Renaming the
variable without the negation makes the logic more understandable.


akselaase (approved):
Looks good to me.

rcarpa (approved):

PR #5747, opened 2022-07-18
Replicas: refactor list replicas PFN generation

rcarpa:


rcarpa:
The first commit is common with https://github.com/rucio/rucio/pull/5712

mlassnig (commented):
it looks correct to me, will study it in more detail....

bari12 (approved):
Looks good to me too!

PR #5749, opened 2022-07-19
Tests: don't run transmogrifier in parallel. Closes #5698

### LINKED ISSUE #5698: transmogrifier test flacky
Motivation
----------
I had frequently this kind of random failure recently : https://github.com/rcarpa/rucio/runs/7183421993?check_suite_focus=true

I'm not 100% sure, but I think I only saw this happen for postgresql database. Which means it should have issues with being run in parallel with some other test. We only run parallel tests on postgresql

rcarpa:
It is the usual issue with heartbeats and partitioning the work.


rcarpa:
PR for the 1.29 branch

maany (approved):

PR #5750, opened 2022-07-19
Add the Pyright type checker

### LINKED ISSUE #5727: Enable static type checking
Motivation
----------
Building on #4455, now that Python 2.7-support is over, enabling a static type checker is easier. This additional tooling will help catch new programmer errors and probably reveal pre-existing bugs. With a proper configuration, the type checker will hopefully not be too noisy, but still provide valuable feedback. Integrating the tooling into the CI pipeline in a similar manner to the `flake8-annotations` check could help catch new errors without complaining too much about old errors.

Modification
------------
The first step is to enable a static type checker with a suitable configuration. There will probably be a lot of typing errors and missing annotations, which can be gradually reviewed and resolved through follow-up PRs. 

The current `flake8-annotations` CI check will still be left in place, since ensuring proper coverage of annotations is vital to also get useful results from the static type checker.

Some pointers for converting an existing codebase with MyPy are provided [here](https://mypy.readthedocs.io/en/stable/existing_code.html).

akselaase:
This PR introduces tooling for the Pyright Python type checker, usable both as a local tool during development, and also integrated into the CI pipeline.

The CI check works in a similar way to the current `flake8-annotations` check, in that it compares the number of typing errors before and after applying an incoming pull request. However, the existing check only counts the number of missing annotations, which does not help in ensuring correctness of the annotations. For instance, [this parameter](https://github.com/rucio/rucio/blob/58efd21b5e21182df80bef3dbe8befa636e440b8/lib/rucio/core/transfer.py#L581) should probably not be a `List[str]`, but a `str`, which Pyright rightly notices in the function body.

The CI check prevents new PRs from increasing the error count. The check also attempts to intelligently recognize where the new errors were introduced, to help the developer find their mistakes more easily. However, this could cause a disruption in other developers' workflow, so maybe some guidelines for how strictly we adhere to the CI result should be developed. Pyright's config can also be modified to be more relaxed if it proves to be too noisy. 

Pyright was chosen due to the relative ease of configuring it and processing its output compared to MyPy. Other issues like MyPy limiting the amount of output in case of too many errors also led to the PR focusing on Pyright (though this issue could maybe be resolved by passing the `--soft-error-limit N` flag to MyPy).

Another benefit of Pyright (that could just as well be present in other type checkers - I didn't look) is the ability to integrate it into your IDE of choice through the Language Server Protocol. I know at least VSCode has seamless integration when using the [Pylance](https://marketplace.visualstudio.com/items?itemName=ms-python.vscode-pylance) extension, which greatly improves the developer experience by highlighting errors immediately, without having to wait for the CI check to run.

As a follow-up to this PR, I want to gradually improve type annotations and correctness throughout the codebase. Pyright reports numerous issues, some of which could be bugs that need fixing.

Closes #5727

### DIFF
--- a/.github/actions/checkout_ancestor_commit/action.yaml
+++ b/.github/actions/checkout_ancestor_commit/action.yaml
@@ -16,9 +16,9 @@ inputs:
       `repository_owner/repository_name` of the repository containing the
       reference commit. For example, rucio/rucio.
     default: rucio/rucio
-  files_to_copy:
+  paths_to_copy:
     description: >-
-      Necessary files to copy (e.g. scripts, configurations, ...) to run code in
+      Necessary paths to copy (e.g. scripts, configurations, ...) to run code in
       the older version of the project. This is for the case an ancestor commit
       gets checked out wich does not contain the required files.
 
@@ -45,19 +45,19 @@ runs:
         FORK_POINT=$(git merge-base $COMMON_ANCESTOR_REPO/${{ inputs.ref }} HEAD)
         echo "The fork point of the current branch with the master is $FORK_POINT"
 
-        echo "Backup all specified files since they could be changed."
+        echo "Backup all specified paths since they could be changed."
         IFS=$'\n'
-        FILES_TO_COPY=$'.github/actions/checkout_ancestor_commit/action.yaml\n${{ inputs.files_to_copy }}'
-        for f in $FILES_TO_COPY; do
+        PATHS_TO_COPY=$'.github/actions/checkout_ancestor_commit/action.yaml\n${{ inputs.paths_to_copy }}'
+        for f in $PATHS_TO_COPY; do
           mkdir -p ../tmp/$(dirname $f)
-          cp $f ../tmp/$f
+          cp --recursive $f ../tmp/$f
         done
 
         echo "Checking out the fork point"
         git checkout $FORK_POINT
 
-        echo "Restore backed up files."
-        for f in $FILES_TO_COPY; do
+        echo "Restore backed up paths."
+        for f in $PATHS_TO_COPY; do
           mkdir -p $(dirname $f)
-          cp ../tmp/$f $f
+          cp --recursive ../tmp/$f $f
         done
--- a/.github/workflows/autotest.yml
+++ b/.github/workflows/autotest.yml
@@ -39,7 +39,7 @@ jobs:
       - name: Checkout ancestor commit with rucio/master
         uses: ./.github/actions/checkout_ancestor_commit
         with:
-          files_to_copy: |
+          paths_to_copy: |
             tools/count_missing_type_annotations.sh
             tools/count_missing_type_annotations_utils.sh
       - name: Count initial number of missing type annotations
--- a/.github/workflows/autotest.yml
+++ b/.github/workflows/autotest.yml
@@ -71,6 +71,40 @@ jobs:
             echo "[2] https://codimd.web.cern.ch/6-SU3cTpQSWRK6FHkM7mAA#"
             exit 1
           fi
+  python_pyright:
+    name: Python type check (Pyright)
+    runs-on: ubuntu-latest
+    env:
+      PYRIGHT_CURRENT_REPORT: ../pyright_current_report.json
+      PYRIGHT_ANCESTOR_REPORT: ../pyright_ancestor_report.json
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install dependencies
+        run: |
+          sudo apt-get install -y npm
+          npm install --global pyright
+      - name: Make pyright report of current commit
+        run: |
+          tools/run_pyright.sh generate ${{ env.PYRIGHT_CURRENT_REPORT }}
+      - name: Show the current Pyright report
+        run: cat ${{ env.PYRIGHT_CURRENT_REPORT }}
+      - name: Checkout ancestor commit with rucio/master
+        uses: ./.github/actions/checkout_ancestor_commit
+        with:
+          paths_to_copy: |
+            pyrightconfig.json
+            tools/run_pyright.sh
+            tools/run_pyright/
+      - name: Make pyright report of ancestor commit
+        run: |
+          tools/run_pyright.sh generate ${{ env.PYRIGHT_ANCESTOR_REPORT }}
+      - name: Show the ancestor Pyright report
+        run: cat ${{ env.PYRIGHT_ANCESTOR_REPORT }}
+      - name: Compare reports
+        run: |
+          tools/run_pyright.sh compare --Werror \
+            ${{ env.PYRIGHT_ANCESTOR_REPORT }} \
+            ${{ env.PYRIGHT_CURRENT_REPORT }}
   setup:
     if: github.repository_owner == 'rucio' || github.event_name != 'schedule'
     runs-on: ubuntu-latest
--- a/tools/run_pyright/__main__.py
+++ b/tools/run_pyright/__main__.py
@@ -1,0 +1,45 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+# Copyright European Organization for Nuclear Research (CERN) since 2012
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import sys
+from argparse import ArgumentParser
+
+from . import generate
+from . import compare
+
+
+def parse_arguments():
+    parser = ArgumentParser('tools/run_pyright.sh', description="Generate and compare Pyright typing reports.")
+    subparser = parser.add_subparsers(dest='command')
+    subparser.required = True
+
+    parse_generate = subparser.add_parser('generate')
+    generate.setup_parser(parse_generate)
+
+    parse_compare = subparser.add_parser('compare')
+    compare.setup_parser(parse_compare)
+
+    return parser.parse_args()
+
+
+def main():
+    args = parse_arguments()
+    exit_code = args.func(args)
+    sys.exit(exit_code)
+
+
+if __name__ == '__main__':
+    main()
--- a/tools/run_pyright/compare.py
+++ b/tools/run_pyright/compare.py
@@ -1,0 +1,97 @@
+# -*- coding: utf-8 -*-
+# Copyright European Organization for Nuclear Research (CERN) since 2012
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from argparse import ArgumentParser, Namespace
+from collections import Counter
+from pathlib import Path
+from typing import List, Tuple
+
+from .models import Report, ReportDiagnostic, ReportDiagnosticWithoutRange, Severity
+from .utils import group_by, load_json
+
+
+def setup_parser(parser: ArgumentParser) -> None:
+    parser.description = """
+    Compares two Pyright reports and outputs newly introduced warnings and errors.
+
+    If new errors are introduced (or warnings with --Werror) the exit code is 2,
+    otherwise 0.
+    """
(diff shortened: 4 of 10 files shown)


tools/run_pyright/__main__.py
```diff
@@ -0,0 +1,54 @@
...
+
+sys.stdout = codecs.getwriter('utf-8')(sys.stdout.buffer)
+sys.stderr = codecs.getwriter('utf-8')(sys.stderr.buffer)
+
+
+def parse_arguments():
+    parser = ArgumentParser(description="Runs the pyright type checker.")
+    subparser = parser.add_subparsers(dest='command')
+    subparser.required = True
+    func: Callable[[Namespace], int]
```
joeldierkes: I don't see the value of this line, `func` is always initialized.

tools/run_pyright/compare.py
```diff
@@ -0,0 +1,93 @@
...
+from typing import Callable, List, Tuple
+
+from .models import Report, ReportDiagnostic, ReportDiagnosticWithoutRange
+from .utils import group_by, load_json
+
+
+def setup_parser(parser: ArgumentParser) -> Callable[[Namespace], int]:
+    parser.add_argument('old', type=Path, help='First report of comparison.')
+    parser.add_argument('new', type=Path, help='Second report of comparison.')
+    parser.add_argument('--werror', action='store_true', help='Treat warnings as errors.')
```
joeldierkes: I only know this argument from gcc, there it is `--Werror`...

tools/run_pyright/compare.py
```diff
@@ -0,0 +1,93 @@
...
+def compare(args: Namespace) -> int:
+    """Compares two reports to find new warnings and errors."""
+    old_report = Report.from_dict(load_json(args.old))
+    new_report = Report.from_dict(load_json(args.new))
+    new_errors = _compare_reports(new_report, old_report)
+
+    print_regressions(new_errors, new_report)
+
+    any_errors = any(err.severity == 'error' for err, _ in new_errors)
+    any_warnings = any(err.severity == 'warning' for err, _ in new_errors)
```
joeldierkes: `serverity` could be an enum. Thoughts?

.github/workflows/autotest.yml
```diff
@@ -71,6 +71,41 @@ jobs:
...
+      - name: Install dependencies
+        shell: bash
+        run: |
+          apt-get install -yy npm
+          npm install --global pyright
+      - name: Make pyright report of current commit
+        shell: bash
+        run: |
+          tools/run_pyright.sh generate \
+            ${{ env.UPDATED_PYRIGHT_REPORT_FILE }}
```
joeldierkes: I want to see the output of the GitHub Actions job. Should we treat warnings here as errors? (`--werror`)

tools/run_pyright/generate.py
```diff
@@ -0,0 +1,79 @@
...
+
+    save_json(args.out, reportdict)
+
+    report = Report.from_dict(reportdict)
+
+    num_errors = report.summary.num_errors
+    if args.werror:
+        num_errors += report.summary.num_warnings
+
+    return 1 if num_errors > 0 else 0
```
joeldierkes: While I like return codes, I am not completely sure if this one is right. The doc comment says `Generate a pyright-report and save it at the specified path.`, which implies for me that a non-zero return code results in a not created pyright-report, not that pyright found some errors. Thoughts?

tools/run_pyright/generate.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+PATHS = (
+    'lib/',
+)
+
+
+def setup_parser(parser: ArgumentParser) -> Callable[[Namespace], int]:
+    parser.add_argument('out', type=Path, help='Store Pyright report at this path.')
+    parser.add_argument('--strip', choices=['line', 'character', 'range'], help='Strip certain attributes from the output.')
```
joeldierkes: An explanation of why we need that would be nice, not necessarily in the git history in that case ;-)

tools/run_pyright/models.py
```diff
@@ -0,0 +1,90 @@
...
+        return cls(
+            severity=obj['severity'],
+            file=obj['file'],
+            rule=obj['rule'],
+            message=obj['message'],
+            range_start_line=obj['range']['start']['line'],
+            range_start_char=obj['range']['start']['character'],
+            range_end_line=obj['range']['end']['line'],
+            range_end_char=obj['range']['end']['character'],
+        )   # type: ignore
```
joeldierkes: Number of spaces before comment

tools/run_pyright/utils.py
```diff
@@ -0,0 +1,46 @@
...
+        result.setdefault(k, []).append(elem)
+    return result
+
+
+def load_json(path: Path) -> ReportDict:
+    with open(path, 'r') as f:
+        return json.load(f)
+
+
+def save_json(out: Union[Path, IO[str]], data: Dict[str, Any]) -> None:
```
joeldierkes: `save_json` is called once with a `Path`. Is the recursive call and the split of the parameter type better compared to:
```python
def save_json(path: Path, data: Dict[str, Any]) -> None:
    with open(path, 'w') as f:
        json.dump(data, f, indent=4)
        f.write('\n')
```

.github/workflows/autotest.yml
```diff
@@ -71,6 +71,41 @@ jobs:
             echo "[2] https://codimd.web.cern.ch/6-SU3cTpQSWRK6FHkM7mAA#"
             exit 1
           fi
+  python_pyright:
+    name: Check Pyright
```
joeldierkes: Maybe a more descriptive name than the package, like `Check Types`/`Check Static Typing`?

tools/run_pyright/generate.py
```diff
@@ -0,0 +1,79 @@
...
+    if args.werror:
+        num_errors += report.summary.num_warnings
+
+    return 1 if num_errors > 0 else 0
+
+
+def _run_pyright() -> ReportDict:
+    """Runs the pyright type-checker and returns its output as json."""
+    cmdline = ['pyright', '--outputjson', *PATHS]
+    process = subprocess.run(cmdline, stdout=subprocess.PIPE)
```
joeldierkes: The error message in case `pyright` is not installed is not descriptive:
```
...
  File "/usr/lib/python3.10/subprocess.py", line 1845, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'pyright'
```
I'd suggest to either give more information on how to resolve that, or maybe use `npx`?

joeldierkes (changes requested):
While I like `mypy`, I don't mind using some other tool. In this case some more information would be nice:
- Why don't we use mypy?
- Why pyright? There are other tools like facebooks [`pyre-ceck`](https://github.com/facebook/pyre-check) and googles [`pytype`](https://github.com/google/pytype).

Very good PR, this helps a lot! Small nits/Questions.

tools/run_pyright/__main__.py
akselaase: It's to type it properly. Ensures that errors show up if any of the `{generate,compare,...}.setup_parser()` calls doesn't return the correct type.

.github/workflows/autotest.yml
akselaase: How about "Python type check (Pyright)"?

tools/run_pyright/generate.py
akselaase: At the moment unused, but I initially used it to produce reports that could be diffed without getting too much noise from changing line numbers. Unsure whether to leave it there or not, thoughts? If so, will explain it better.

.github/workflows/autotest.yml
akselaase: Both reports are now shown in a separate step in the workflow.

.github/workflows/autotest.yml
```diff
@@ -71,6 +71,45 @@ jobs:
...
+          tools/run_pyright.sh generate \
+            ${{ env.PYRIGHT_ANCESTOR_REPORT }} \
+            || true
+      - name: Show the ancestor Pyright report
+        run: cat ${{ env.PYRIGHT_ANCESTOR_REPORT }}
+      - name: Compare reports
+        run: |
+          tools/run_pyright.sh compare \
+            ${{ env.PYRIGHT_ANCESTOR_REPORT }} \
+            ${{ env.PYRIGHT_CURRENT_REPORT }}
```
akselaase: To decide: With or without `--Werror`?

.github/actions/checkout_ancestor_commit/action.yaml
```diff
@@ -53,6 +59,12 @@ runs:
           cp $f ../tmp/$f
         done
 
+        DIRECTORIES_TO_COPY=$'.github/actions/checkout_ancestor_commit/action.yaml\n${{ inputs.directories_to_copy }}'
+        for f in $DIRECTORIES_TO_COPY; do
+          mkdir -p ../tmp/$(dirname $f)
+          cp -r $f ../tmp/$f
+        done
+
```
akselaase: Could also just rename `files_to_copy` to `paths_to_copy` and have it support both files and directories by just adding `-r` to the `cp` command, instead of having separate loops for files and directories.

tools/run_pyright/__main__.py
```diff
@@ -0,0 +1,54 @@
...
+from typing import Callable
+
+from . import generate
+from . import compare
+
+
+sys.stdout = codecs.getwriter('utf-8')(sys.stdout.buffer)
+sys.stderr = codecs.getwriter('utf-8')(sys.stderr.buffer)
+
+
```
akselaase: Not needed when were setting `PYTHONIOENCODING=utf8` I believe, todo: test.

tools/run_pyright/generate.py
```diff
@@ -0,0 +1,95 @@
...
+              ' This could be due to pyright not being installed on your system,'
+              ' in which case it may be installed using `npm install --global pyright`.\n'
+              'Additional details:', ex,
+              file=sys.stderr)
+        sys.exit(1)
+    except OSError as ex:
+        print('Unknown error running pyright.\n'
+              'Additional details:', ex,
+              file=sys.stderr)
+        raise
```
akselaase: This doesn't really provide anything useful.

.github/actions/checkout_ancestor_commit/action.yaml
joeldierkes: That's the cleaner alternative, go for it.
Please create a proper commit with a proper commit message for that (including the intention).

You can force push the branch and amend/edit your changes. The comments on GitHub are still present in the view (if the line is not deleted) and you have revisions for each review.

tools/run_pyright/generate.py
akselaase: Agreed to remove since it is unused.

akselaase:
Demo of output can be seen in the "Compare reports" step here: https://github.com/rucio/rucio/runs/7469292269?check_suite_focus=true

tools/run_pyright/compare.py
```diff
@@ -0,0 +1,105 @@
...
+    """Prepends `prefix` to each line in `text` except the first."""
+    def _inner():
+        first = True
+        for line in text.splitlines(keepends=True):
+            if first:
+                yield line
+            else:
+                yield prefix + line
+            first = False
+    return ''.join(_inner())
```
joeldierkes: nit: maybe we can utilize the `textwrap.indent` function for that?
Something like:
```python
lines = text.splitlines(keepends=True)
return lines[0] + textwrap.indent(''.join(lines[1:]), prefix, lambda_: True)
```

joeldierkes (dismissed):
Good PR!

nit: The second commit message is not wrapped.
nit: The phrase `Pyright would have a hard time without any annotations` in the commit message is rather generic. Maybe it should be explained here that Pyright is more of a "static" type checker and requires some type annotations. While other type checkers are more dynamic and better at type inference, they did not provide the needed functionalities and integration into tooling. (or are just painfully slow).

tools/run_pyright/compare.py
```diff
@@ -0,0 +1,97 @@
...
+    diff = new_count
+    diff.subtract(old_count)
+
+    new_diagnostics = [(err, count) for err, count in diff.most_common() if count > 0]
+    return new_diagnostics
+
+
+def _indent(text: str, prefix: str) -> str:
+    """Prepends `prefix` to each line in `text` except the first."""
+    return text.replace('\n', '\n' + prefix)
```
joeldierkes, line 73: :joy:

akselaase:
Slightly edited the commit message and simplified the indentation function. This dismissed the review, is this desired behaviour? It can be configured as shown here: https://github.com/isaacs/github/issues/1157#issuecomment-450409941.

joeldierkes:
I don't know if I'n in favor of that. This use case hardly ever appears, and to resolve it you just have to re-request a new review.

Since we work on an open-source project, I'd rather have explicit approval than someone who changes more than the requested changes.

joeldierkes (dismissed):

akselaase:
To make sure this works in the development container, https://github.com/rucio/containers/pull/197 should also be merged. I can prepare a separate PR for documentation.

.github/workflows/autotest.yml
akselaase: Added `--Werror`. Better to get the reviewer to decide in case something is questionable.

bari12:
Needs to be rebased due to an issue which existed on master yesterday.

joeldierkes:
I think this is good to go :+1:

bari12 (approved):

PR #5754, opened 2022-07-21
Transfers: ensure transfer and stagein logic for qos and tests #4751

maatthias:


lib/rucio/core/lock.py
```diff
@@ -398,7 +398,14 @@ def failed_transfer(scope, name, rse_id, error_message=None, broken_rule_id=None
     :param session:         The database session in use.
     """
 
-    locks = session.query(models.ReplicaLock).with_for_update(nowait=nowait).filter_by(scope=scope, name=name, rse_id=rse_id)
+    staging_required = next(iter(get_rse_attribute('staging_required', rse_id=rse_id, session=session)), False)
```
bari12: This `next(iter(..)` doesn't seem necessary to me. If the attribute does not exist, you will get None back and the if block below will evaluate to False?

lib/rucio/core/lock.py
```diff
@@ -398,7 +398,14 @@ def failed_transfer(scope, name, rse_id, error_message=None, broken_rule_id=None
...
+    if staging_required:
+        rse_name = get_rse_name(rse_id=rse_id, session=session)
+        logger(logging.INFO, f'Destination RSE {rse_name} is type staging_required so do not update other OK replica locks.')
```
bari12: This should be DEBUG

lib/rucio/core/rule_grouping.py
```diff
@@ -842,6 +842,14 @@ def __create_lock_and_replica(file, dataset, rule, rse_id, staging_area, availab
                                                         lifetime=lifetime,
                                                         session=session))
 
+    # If staging_required type RSE then set pin to RSE attribute maximum_pin_lifetime
+    staging_required = next(iter(get_rse_attribute('staging_required', rse_id=rse_id, session=session)), False)
+    maximum_pin_lifetime = next(iter(get_rse_attribute('maximum_pin_lifetime', rse_id=rse_id, session=session)), None)
```
bari12: Per default this would already reply `None`

lib/rucio/core/rule_grouping.py
```diff
@@ -842,6 +842,14 @@ def __create_lock_and_replica(file, dataset, rule, rse_id, staging_area, availab
                                                         lifetime=lifetime,
                                                         session=session))
 
+    # If staging_required type RSE then set pin to RSE attribute maximum_pin_lifetime
+    staging_required = next(iter(get_rse_attribute('staging_required', rse_id=rse_id, session=session)), False)
```
bari12: Again, the `None` reply should be enough for the if block below?

lib/rucio/core/rule_grouping.py
```diff
@@ -842,6 +842,14 @@ def __create_lock_and_replica(file, dataset, rule, rse_id, staging_area, availab
...
+    maximum_pin_lifetime = next(iter(get_rse_attribute('maximum_pin_lifetime', rse_id=rse_id, session=session)), None)
+
+    if staging_required:
+        rse_name = get_rse_name(rse_id=rse_id, session=session)
+        logger(logging.INFO, f'Destination RSE {rse_name} is type staging_required and if replica exists create STAGEIN request.')
```
bari12: DEBUG

lib/rucio/core/rule_grouping.py
```diff
@@ -855,12 +863,27 @@ def __create_lock_and_replica(file, dataset, rule, rse_id, staging_area, availab
...
+                                                            scope=file['scope'],
+                                                            name=file['name'],
+                                                            rule=rule,
+                                                            lock=new_lock,
+                                                            bytes_=file['bytes'],
+                                                            md5=file['md5'],
+                                                            adler32=file['adler32'],
+                                                            ds_scope=dataset['scope'],
+                                                            ds_name=dataset['name'],
+                                                            lifetime=maximum_pin_lifetime if staging_required else None,
```
bari12: This should probably be the the rule_lifetime, unless the rule_lifetime is bigger than maximum_pin_lifetime, in which case maximum_pin_lifetime is used.
The `else None` part is also not really possible in this case :-)

lib/rucio/core/rule_grouping.py
```diff
@@ -893,6 +916,7 @@ def __create_lock_and_replica(file, dataset, rule, rse_id, staging_area, availab
                                                                 adler32=file['adler32'],
                                                                 ds_scope=dataset['scope'],
                                                                 ds_name=dataset['name'],
+                                                                lifetime=maximum_pin_lifetime if staging_required else None,
```
bari12: This is correct (The comment with the lifetime limit still holds though)
I am wondering though if it works. E.g. For a transfer from an RSE, somewhere in the world, to the MAS RSE, will the PIN lifetime by communicated? Or will this just be communicated for a STAGEIN?

lib/rucio/core/rule_grouping.py
```diff
@@ -954,6 +978,7 @@ def __create_lock_and_replica(file, dataset, rule, rse_id, staging_area, availab
                                                             adler32=file['adler32'],
                                                             ds_scope=dataset['scope'],
                                                             ds_name=dataset['name'],
+                                                            lifetime=maximum_pin_lifetime if staging_required else None,
```
bari12: Also see comment with lifetime limit from above

lib/rucio/tests/test_conveyor.py
```diff
@@ -757,6 +759,291 @@ def test_stager(rse_factory, did_factory, root_account, replica_client):
     assert replica['state'] == ReplicaState.AVAILABLE
 
 
+@skip_rse_tests_with_accounts
+@pytest.mark.noparallel(reason="runs submitter; poller and finisher")
+def test_transfer_to_mas_new_replica(rse_factory, did_factory, root_account):
+    """
+    Test qos: transfer from tape to disk
```
bari12: This is actually from DISK to TAPE
I would add some more description, something like. Test QoS transfer from some RSE in the world to MAS

lib/rucio/tests/test_conveyor.py
```diff
@@ -757,6 +759,291 @@ def test_stager(rse_factory, did_factory, root_account, replica_client):
...
+
+    # mock a failed transfer
+    # re-query request to get external_id
+    request = request_core.get_request(request_id=request['id'])
+    request_core.set_request_state(request_id=request['id'], state=RequestState.FAILED, external_id=request['external_id'])
+    finisher(once=True, partition_wait_time=None)
+    lock_core.failed_transfer(scope=did['scope'], name=did['name'], rse_id=dst_rse_id)
+
+    assert rule_core.get_rule(rule_id)['state'] == RuleState.STUCK
+    assert lock_core.get_replica_locks_for_rule_id(rule_id=rule_id)[0]['state'] == LockState.STUCK
```
bari12, line 847: Also test the replica state here (UNAVAILABLE)

lib/rucio/tests/test_conveyor.py
```diff
@@ -757,6 +759,291 @@ def test_stager(rse_factory, did_factory, root_account, replica_client):
...
+        assert filtered_lock['state'] == LockState.OK
+    for rse_id in all_rses:
+        assert replica_core.get_replica(rse_id=rse_id, **did)['state'] == ReplicaState.AVAILABLE
+
+
+@skip_rse_tests_with_accounts
+@pytest.mark.noparallel(reason="runs submitter; poller and finisher")
+def test_failed_transfer_to_mas_new_replica(rse_factory, did_factory, root_account):
+    """
+    Test a qos failed transfer from tape to disk
```
bari12: Also from DISK to TAPE here?

lib/rucio/tests/test_conveyor.py
```diff
@@ -757,6 +759,291 @@ def test_stager(rse_factory, did_factory, root_account, replica_client):
...
+        assert replica_core.get_replica(rse_id=rse_id, **did)['state'] == ReplicaState.AVAILABLE
+
+    # now test a second rule, different account
+    set_local_account_limit(InternalAccount('jdoe'), dst_rse_id, bytes_=-1)
+    rule2_id = rule_core.add_rule(dids=[did], account=InternalAccount('jdoe'), copies=1, rse_expression=dst_rse, grouping='ALL', weight=None, lifetime=-1, locked=False, subscription_id=None, source_replica_expression=dst_rse)[0]
+    request = request_core.get_request_by_did(rse_id=dst_rse_id, **did)
+
+    assert request['request_type'] == RequestType.STAGEIN
+
+    submitter(once=True, rses=[{'id': dst_rse_id}], partition_wait_time=None, transfertool='mock', transfertype='single', filter_transfertool=None)
```
bari12: Technically you have to call the `stager` here. `submitter` only handles TRANSFER requests, stager handles STAGEIN. But it doesn't do anything anyway here since you have to mock everything. (Just for the sake of completeness)

bari12 (commented):
Some minor comments, but in general this looks good to me. I will do a second pass tomorrow thinking about some of the cases in case I forgot something, but in general I think this can be used like this.

lib/rucio/core/lock.py
maatthias: It looks like a change was merged for `get_rse_attribute` last week that now returns `None` if the key doesn't exist.  Before that I needed the above to handle that case.  (in other words Joel fixed `get_rse_attribute`). ;)
https://github.com/rucio/rucio/commit/0225fac1b6d9ca2c326256db827cecb24013e10a

joeldierkes: https://github.com/rucio/rucio/commit/826f095d06191cc5526225268deab22d0a16fd25 Just for clearance, that was the idea! ;-)

bari12 (approved):

Issue #5573, opened 2022-05-19
implement sending logs directly to ES


rcarpa:
Motivation
----------
In servers, we write logs to a file. Then have filebeat read that file, merge lines together, and send it to ES. This is counter-productive. First of all, we can send formatted ES messages directly from our logger. Second, filebeat does a bad job of merging stacktraces into one message, making it completely impossible for an operator to find meaningful information in ES

maany:
I could take this over @rcarpa if you prefer to spend effort on other things ;)

rcarpa:
Feel free to pick it up if you want :)

rcarpa:
Just to inform that I'm working on it

PR #5760, opened 2022-07-26
Monitoring & Logging: implement rucio log formatter. #5573

rcarpa:
The formatter should be a drop-in replacement to the python builtin
formatter, with two additional additions:
- it can output directly to json
- it can include data from the flask 'request' object into the format

When the logger writes to a json format, it tries to respect the
Elastic Common Schema (ECS) specification, but without getting too
strict about it.

When the format string contains a dot-separated "path" starting with
`http.request.`, the rucio formatter will try to extract the given
path from the flask `request` object.

Both these changes should allow to simplify the log ingestion pipeline.
- Json simplifies injection, removing the need to have complex parsing
of log lines and tokenization + merging of multiple lines into one log.
- Request data, like the client ip, can now be collected from within
the wsgi application instead of parsing httpd logs to extract it.

### DIFF
--- a/lib/rucio/common/logging.py
+++ b/lib/rucio/common/logging.py
@@ -341,18 +341,21 @@ class RucioFormatter(logging.Formatter):
                 return '{}'
 
 
+def rucio_log_formatter():
+    config_logformat = config_get('common', 'logformat', raise_exception=False, default='%(asctime)s\t%(name)s\t%(process)d\t%(levelname)s\t%(message)s')
+    output_json = config_get_bool('common', 'logjson', default=False)
+    return RucioFormatter(fmt=config_logformat, output_json=output_json)
+
+
 def setup_logging(application=None):
     """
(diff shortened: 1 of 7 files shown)


lib/rucio/common/logging.py
```diff
@@ -12,25 +12,355 @@
...
+
+    from typing import Any, Callable, Dict, Generator, Optional, Sequence, Tuple
+
+
+# Mapping from ECS field paths
+# https://www.elastic.co/guide/en/ecs-logging/overview/current/intro.html#_field_mapping
+# https://www.elastic.co/guide/en/ecs/8.5/ecs-field-reference.html
+# to python log record attributes:
+# https://docs.python.org/3/library/logging.html#logrecord-attributes
+BUILTIN_FIELDS = (
```
mlassnig, line 40: eew, these are really ugly... but these are the official names. :-(

mlassnig (approved):

PR #5769, opened 2022-08-01
Transfers: gracefully skip missing sources in poller/receiver

rcarpa:
The only difference is that the "except" block will not be triggered
and the stack-trace will not be included in the log message.

On the couple of cases I investigated, the missing sources were due to
the poller working on a request which is handled by receiver+finisher
in the meantime.

### DIFF
--- a/lib/rucio/core/request.py
+++ b/lib/rucio/core/request.py
@@ -1835,6 +1835,7 @@ def get_source_rse(request_id, src_url, session=None, logger=logging.log):
             return None, None
 
         sources = get_sources(request_id, session=session)
+        sources = sources or []
         for source in sources:
             if source['url'] == src_url:
                 src_rse_id = source['rse_id']


lib/rucio/core/request.py
```diff
@@ -1835,6 +1835,7 @@ def get_source_rse(request_id, src_url, session=None, logger=logging.log):
             return None, None
 
         sources = get_sources(request_id, session=session)
```
mlassnig, line 1837: why not 

`sources = get_sources(request_id, session=session) or []` ?

rcarpa: are we paying by line of code ? like in the good old "C compilation by line of code" days ? :) 

In reality, I initially did it like that. After that I changed because I considered it's much more explicit this way and less likely to be missed by somebody reading the code.

mlassnig: haha :-D

och, I don't mind much I was just interested

mlassnig (approved):

Issue #5768, opened 2022-07-29
Type annotations for `rucio.client`


akselaase:
Motivation
----------
Type annotations help prevent programmer errors and serve as documentation for users of the `rucio.client` package. Implementing annotations for all Rucio clients would therefore improve the experience of using the package.

Modification
------------
Add type annotations to the following files under `lib/rucio/client`.
- [ ] accountclient.py
- [ ] accountlimitclient.py
- [ ] baseclient.py
- [ ] client.py
- [ ] configclient.py
- [ ] credentialclient.py
- [ ] didclient.py
- [ ] diracclient.py
- [ ] downloadclient.py
- [ ] exportclient.py
- [ ] fileclient.py
- [ ] importclient.py
- [ ] lifetimeclient.py
- [ ] lockclient.py
- [ ] metaclient.py
- [ ] pingclient.py
- [ ] replicaclient.py
- [ ] requestclient.py
- [ ] rseclient.py
- [ ] ruleclient.py
- [ ] scopeclient.py
- [ ] subscriptionclient.py
- [ ] touchclient.py
- [ ] uploadclient.py

PR #5770, opened 2022-08-01
Clients: Type annotations for `ReplicaClient`; #5768

akselaase:
This PR adds type annotations to some function definitions in `ReplicaClient`. Typed dictionaries and Literal types are created in the `rucio.common.types` module, so that the same types can be reused both client- and server-side.

The PR is not ready to be merged yet, but to just serve as a starting point for discussions around how to implement type annotations for the Rucio clients.


lib/rucio/client/replicaclient.py
```diff
@@ -95,12 +99,16 @@ def get_did_from_pfns(self, pfns, rse=None):
...
+    def list_replicas(self, dids: List[DidDict], schemes: Optional[List[str]] = None,
+                      ignore_availability: bool = True, all_states: bool = False, metalink: bool = False,
+                      rse_expression: Optional[str] = None,
+                      client_location: Optional[ClientLocationDict] = None,
+                      sort: Optional[ReplicaSortLiteral] = None,
+                      domain: Literal['all', 'wan', 'lan', None] = None,
+                      signature_lifetime: Optional[int] = None, nrandom: Optional[int] = None,
+                      resolve_archives: bool = True, resolve_parents: bool = False,
+                      updated_after: Union[datetime, str, None] = None
+                      ) -> List[ReplicaDict]:
```
akselaase: I could improve formatting a bit if desired.

lib/rucio/common/types/dicts/typed_dict.py
```diff
@@ -0,0 +1,23 @@
...
+# limitations under the License.
+
+import sys
+
+__all__ = ['TypedDict', 'is_typeddict']
+
+if sys.version_info >= (3, 8):
+    from typing import TypedDict, is_typeddict
+else:
+    from typing_extensions import TypedDict, is_typeddict
```
akselaase: The reason for this file is to serve as a single source for the `TypedDict` class. An alternative to this is to just unconditionally import from `typing_extensions` everywhere we use it.

lib/rucio/common/types/literals/literal.py
```diff
@@ -0,0 +1,23 @@
...
+# limitations under the License.
+
+import sys
+
+__all__ = ['Literal']
+
+if sys.version_info >= (3, 8):
+    from typing import Literal
+else:
+    from typing_extensions import Literal
```
akselaase: Same as for `TypedDict`.

lib/rucio/common/types/literals/replica_sort.py
```diff
@@ -0,0 +1,18 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from .literal import Literal
+
+ReplicaSortLiteral = Literal['geoip', 'closeness', 'dynamic', 'ranking', 'random']
```
akselaase: Do we want to define `Literal`s like this? Ideally we'd define an `Enum`, but I want to avoid runtime changes and just focus on the annotations for now.

lib/rucio/common/types/dicts/replica.py
```diff
@@ -0,0 +1,44 @@
...
+
+class ReplicaDict(TypedDict):
+    scope: str
+    name: str
+    bytes: int
+    md5: str
+    adler32: str
+    rses: Dict[str, List[str]]
+    pfns: Dict[str, Dict[str, ReplicaPFNDict]]
+    states: Dict[str, str]
```
akselaase: I'm not sure about naming for these classes. Function parameters that use `BadReplicaDict` are usually called `pfns`, but is there a more general term that includes both PFN and a RSE-specific replica? `Replica` is also a somewhat unclear term whether it refers to a specific file on a single RSE or the replication of a file across multiple RSEs, any suggestions for differentiating the names? 

The goal of these classes is to re-use them across different functions. A problem I can imagine appearing is if functions take slightly different versions of the same dict (maybe some keys are optional), which would require duplicating the dicts for the different functions. I'm not sure if it will be a problem though.

akselaase: See `ReplicaClient.declare_bad_file_replicas` for instance.

lib/rucio/common/types/dicts/typed_dict.py
akselaase: Pyright seem to be confused when `TypedDict` is imported from `.typed_dict` instead of from `typing_extensions`, so I will remove this forwarding and use `typing_extensions` everywhere.

lib/rucio/common/types/literals/literal.py
akselaase: Same import confusion, will remove this file.

lib/rucio/common/types/dicts/replica.py
akselaase: It's becoming quite the overlap between some slightly-different dicts now. Some potential solutions:
- Use subclassing where possible, for instance every dict with `scope: str` and `name: str` could inherit from `DidDict`
- Define TypedDict privately just before the corresponding function definitions if they are only used once. Not ideal since using TypedDicts often mean annotating variable assignments explicitly (otherwise just `Dict` is assumed), which means that they have to be public. Could still be done in-line with the function definitions, but it is also a bit messy. 
- Ignore it. It is not a new problem with having this many different dictionaries anyway, they have always been there, just not this explicit.

lib/rucio/common/constants.py
```diff
@@ -71,16 +72,6 @@ class SuspiciousAvailability(enum.Enum):
     LAST_COPY = 2
 
 
-class ReplicaState(enum.Enum):
```
bari12, line 74: Is this just not used anywhere?

lib/rucio/common/types/dicts/replica.py
bari12: I am leaning to just ignore it for now. The root of evil here is that we give the same name to actual different things. So this would have to be "fixed" in the first place, but I am not even sure if it is fix-worthy. In the scope of type checking I would ignore it and just have different dicts.

bari12:
This is very nice @akselaase I have nothing to add much. Just one question (above)

lib/rucio/common/constants.py
akselaase: It's used a lot of places, just moved to `rucio.common.types.enums` (and imported from there into `constants.py` so I don't have to change imports all over the codebase).

akselaase:
Rebased on master, had some conflicts with #5773 which were resolved.

Issue #5652, opened 2022-06-20
Proper versioning for policy packages


jamesp-epcc:
A recent change to the permissions code in Rucio core broke the API for policy packages. To avoid this happening again, I propose to enforce a proper versioning scheme for the policy packages. The packages already contain a `SUPPORTED_VERSION` field but currently this is never used by Rucio.

My plan is to have Rucio check the supported version field when loading a policy package. If there have been API changes between the version of Rucio supported by the policy package and the version actually running on the server, an error will be raised so that the mismatch can be easily detected.

PR #5777, opened 2022-08-04
Core & Internals: enforce policy package supported version #5652

jamesp-epcc:
This pull request checks the supported version field of the policy package and throws an exception if it is not compatible with the Rucio version.

I am not sure where would be best to store the `first_compatible_version` value. It will need to be updated every time the interface between Rucio and the policy package changes (for example, when the session argument was added to the permissions code recently). I don't know what is the best way to ensure this doesn't get forgotten, but I am open to suggestions.

### DIFF
--- a/lib/rucio/common/utils.py
+++ b/lib/rucio/common/utils.py
@@ -47,7 +47,7 @@ from configparser import NoOptionError, NoSectionError
 
 from rucio.common.config import config_get, config_has_section
 from rucio.common.exception import MissingModuleException, InvalidType, InputValidationError, MetalinkJsonParsingError, RucioException, \
-    DuplicateCriteriaInDIDFilter, DIDFilterSyntaxError, InvalidAlgorithmName
+    DuplicateCriteriaInDIDFilter, DIDFilterSyntaxError, InvalidAlgorithmName, PolicyPackageVersionError
 
 from rucio.common.extra import import_extras
 from rucio.common.types import InternalAccount, InternalScope
@@ -1772,6 +1772,7 @@ def register_policy_package_algorithms(algorithm_type, dictionary):
                 package = os.environ[env_name]
             else:
                 package = config.config_get('policy', 'package' + ('' if not vo else '-' + vo))
+            check_policy_package_version(package)
             module = importlib.import_module(package)
             if hasattr(module, 'get_algorithms'):
                 all_algorithms = module.get_algorithms()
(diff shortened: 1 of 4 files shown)


lib/rucio/common/utils.py
```diff
@@ -1824,6 +1825,31 @@ def try_importing_policy(algorithm_type, dictionary, vo=None):
...
+        module = importlib.import_module(package)
+    except ImportError:
+        # package not found. Will be picked up elsewhere
+        return
+    if not hasattr(module, 'SUPPORTED_VERSION'):
+        # package is not versioned
+        return
+    first_compatible_version = '1.28.0'
+    supported_version = module.SUPPORTED_VERSION
+    if supported_version < first_compatible_version:
```
mlassnig: Does this really do the right thing? :exploding_head: Or does it only work accidentally because it converts the version string into a float?

jamesp-epcc: It should do string comparison by default, rather than converting to float automatically. I guess the question is whether the version strings are always in the correct form for the string ordering to work as intended? It seems like they should be, and all the ones I tested worked correctly, but there may be unusual cases I haven't thought of.

bari12: Sorry for my late reply here.
I don't think this works entirely correct, at least if you take `rcN` and `.postN` versions into account.
I am wondering in general though what the comparison here should be and I think it basically should be equivalence of the major version.

Because this code would allow if the POLICY has `SUPPORTED_VERSION=1.29` but the rucio version is already `1.30` it would not fail. Which seems wrong?

jamesp-epcc: My idea was that we could update `first_compatible_version` only when there is a change that breaks the interface used by the policy package, so a policy package would be compatible with a range of Rucio versions. But maybe that is adding too much complexity. A simple test for version equality might be better, and people can update their policy package `SUPPORTED_VERSION` field when required.

bari12: I think this will miss an important case. New Rucio version comes out (`1.30`) which has some important API change. Thus for unmaintained policies, this will just assume it is working, while it is not.
I think it would be best to implement just a list of compatible versions. (Major versions)
If we really want to make this "elegant" we can implement something like [pip version specifiers](https://peps.python.org/pep-0440/#version-specifiers)
Where you could give a `compatible_version = ['~=1.28', '~=1.30']`etc.

jamesp-epcc:
I have updated this to allow the policy package to contain a list of supported versions, as suggested by Martin. My main concern is whether my method for parsing out the major version from the version string is going to work in all cases.

bari12 (approved):

PR #5782, opened 2022-08-05
Transfers: fix marking multi-source as failed. Closes #5780

### LINKED ISSUE #5780: Failed multi-source transfers not transitioned to failed state
Motivation
----------

this condition is incomplete https://github.com/rucio/rucio/blob/5804712c7eb187ffb1fcfc3bbf49d89c02c0aeff/lib/rucio/transfertool/fts3.py#L576

rcarpa:
The issue is that none of the conditions were matched for a failed
multi-source transfer. So it was never transitioned. The bug was
introduced in 1.29


mlassnig (approved):
oh my, that was confusing :-D

Issue #5779, opened 2022-08-04
Speed up list_replicas() implementation


ivmfnal:
Motivation
----------
DUNE is planning to use ReplicaClient.list_replics() method at the scale of ~1000 files. Currently, the method is too slow to be practical. It hakes from 2.5 to 3 seconds for 100 files. The method needs to be faster to be useful.

Modification
------------
Currently, core.replica.list_replicas() calls _resolve_dids(), which is issuing a query per requested DID, unless the DID type is known and is "file". Unfortunately, when I try to add "did_type":"file" to my list of DIDs on the client side, the server rejects my request because "did_type" is not accepted by the request schema:

```
Traceback (most recent call last):
  File "list_replicas.py", line 7, in <module>
    out = list(client.list_replicas([{"scope":f["namespace"], "name":f["name"], "did_type":"file"} for f in files], 
  File "/home/ivm/miniconda3/lib/python3.8/site-packages/rucio/client/replicaclient.py", line 196, in list_replicas
    raise exc_cls(exc_msg)
rucio.common.exception.InvalidObject: Provided object does not match schema.
Details: Problem validating r_dids : Additional properties are not allowed ('did_type' was unexpected)
...

On instance[0]:
    {'did_type': 'file',
     'name': 'dc4_np02tde_407021731_455_59_cb.test',
     'scope': 'dc4-vd-coldbox-top'}
```

There can be 2 modifications:
1. Accept "did_type" on the server side
2. Optimize _resolve_dids() so that it issues single query to resolve DID types instead of one query per DID at https://github.com/rucio/rucio/blob/8e32fd1527116640e0ed719cb526bb5ff4303a52/lib/rucio/core/replica.py#L649

mlassnig:
So, I'm trying to understand what's going on here.

Here's an example from the ATLAS instance:

```
$ time -p rucio list-file-replicas mc15_13TeV:mc15_13TeV.422003.ParticleGunEvtGen_single_B0_flatPt10to1000.simul.HITS.e4566_s2726_tid06828093_00 | wc -l
1300
real 5.43
user 0.97
sys 0.27
```


So roughly 6 seconds to query a dataset with 999 files (this is first resolved into its constituent file DIDs, then replicas are resolved) into 1300 replicas. Interestingly, the time stays rather constant, with a tendency to have a larger baseline cost. I tried for a couple differently-sized datasets (from 1-2 files, up to 10k files), and it's always noticeably under 10 seconds, but never less than 2 seconds.

[...]

PR #5783, opened 2022-08-06
Replicas: list_replicas() optimization #5779

ivmfnal:
Optimize list_replicas():
1. Do not call get_rse_name() twice for each replica in the request. Use RSE names returned by core.list_replicas.
2. Document passing "type":"FILE" as a way to bypass DID resolution

### DIFF
--- a/lib/rucio/api/replica.py
+++ b/lib/rucio/api/replica.py
@@ -200,20 +200,9 @@ def list_replicas(dids, schemes=None, unavailable=False, request_id=None,
                                      client_location=client_location, domain=domain,
                                      sign_urls=sign_urls, signature_lifetime=signature_lifetime,
                                      resolve_archives=resolve_archives, resolve_parents=resolve_parents,
-                                     nrandom=nrandom, updated_after=updated_after, session=session)
+                                     nrandom=nrandom, updated_after=updated_after, by_rse_name=True, session=session)
 
     for rep in replicas:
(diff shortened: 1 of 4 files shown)


lib/rucio/api/replica.py
```diff
@@ -204,15 +204,10 @@ def list_replicas(dids, schemes=None, unavailable=False, request_id=None,
...
-            old_dict = rep.get(k, None)
+        for k in ('rses', 'states'):
+            old_dict = rep.get(k)
             if old_dict is not None:
-                new_dict = {}
-                for rse_id in old_dict:
-                    rse = get_rse_name(rse_id=rse_id, session=session) if rse_id is not None else None
-                    new_dict[rse] = old_dict[rse_id]
-                rep[k] = new_dict
+                rep[k] = rep.pop(k + "_by_name", None)
```
rcarpa: This pattern: using the result dictionary as a storage for temporary data which is then removed from the dict is extremely confusing and prone to introduction of bugs in the future. I just spent many months getting rid of this behavior in conveyor and fixing numerous bugs linked to it.

I would rather pass an additional function parameter to core.replica.list_replicas(). For example: `group_rses_by_name = True/False` and directly return the data in the correct format. My IMHO...

ivmfnal: Thanks Radu. Done.

PR #5785, opened 2022-08-08
Fix usages of `datetime.now()` instead of `datetime.utcnow()`

akselaase:
According to the [Rucio Documentation](https://rucio.cern.ch/documentation/developing_with_rucio/#date-format) the server should treat all naive datetime objects as UTC.  Unfortunately, `datetime.now()` was used extensively, which returns the server's local time instead of UTC time. This could for instance cause rules to expire some hours too late or early.

This commit replaces many calls to `datetime.now` with calls to `datetime.utcnow`. Some calls to `datetime.now` are left behind, these are in most cases used for measuring elapsed time, though some are also left where they concern file modification times, which are in the local time zone.

The PR contains three commits, the last two of which are somewhat unrelated changes which could be 1) squashed into the main commit or 2) moved to separate PRs if desired.

### DIFF
--- a/lib/rucio/common/utils.py
+++ b/lib/rucio/common/utils.py
@@ -27,6 +27,7 @@ import mmap
 import os
 import os.path
 import re
+import signal
 import socket
 import subprocess
 import tempfile
@@ -1349,21 +1350,19 @@ def run_cmd_process(cmd, timeout=3600):
     :return: stdout xor stderr, and errorcode
     """
 
-    time_start = datetime.datetime.now().second
     process = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=True, preexec_fn=os.setsid, universal_newlines=True)
 
-    running_time = 0
-    while process.poll() is None and running_time < timeout:
-        time_now = datetime.datetime.now().second
-        running_time = int(time_now - time_start)
-        time.sleep(3)
-    if process.poll() is None:
-        process.terminate()
-        time.sleep(3)
-    if process.poll() is None:
-        process.kill()
+    try:
+        stdout, stderr = process.communicate(timeout=timeout)
+    except subprocess.TimeoutExpired:
+        try:
+            # Kill the whole process group since we're using shell=True.
+            os.killpg(os.getpgid(process.pid), signal.SIGTERM)
+            stdout, stderr = process.communicate(timeout=3)
+        except subprocess.TimeoutExpired:
+            os.killpg(os.getpgid(process.pid), signal.SIGKILL)
+            stdout, stderr = process.communicate()
 
-    stdout, stderr = process.communicate()
     if not stderr:
         stderr = ''
     if not stdout:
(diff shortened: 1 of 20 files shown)


lib/rucio/core/replica.py
```diff
@@ -3855,7 +3855,7 @@ def get_suspicious_files(rse_expression, available_elsewhere, filter_=None, logg
     if not isinstance(nattempts, int):
         nattempts = 0
     if not isinstance(younger_than, datetime):
-        younger_than = datetime.now() - timedelta(days=10)
+        younger_than = datetime.utcnow() - timedelta(days=younger_than)
```
mlassnig: @cserf @dchristidis :eyes: pls

lib/rucio/common/utils.py
```diff
@@ -1347,21 +1348,19 @@ def run_cmd_process(cmd, timeout=3600):
...
-        running_time = int(time_now - time_start)
-        time.sleep(3)
-    if process.poll() is None:
-        process.terminate()
-        time.sleep(3)
-    if process.poll() is None:
-        process.kill()
-
-    stdout, stderr = process.communicate()
+    try:
```
mlassnig, line 1355: nice!

joeldierkes:
After reading the `datetime.now` and `datetime.utcnow` documentation, we should change this PR a bit:

The [`datetime.now` documentation](https://docs.python.org/3/library/datetime.html#datetime.datetime.now) clearly states, that this approach is favorable over `datetime.utcnow`:
> This function is preferred over [today()](https://docs.python.org/3/library/datetime.html#datetime.datetime.today) and [utcnow()](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow).

The disadvantage of `datetime.utcnow` is that it sets the timezone info of the resulting datetime object to `None`, which we probably don't want. It suggests to use `datetime.now(timezone.utc)`, which sets the timezone info correctly.

I'll change that in a bit.

lib/rucio/core/replica.py
joeldierkes: Ping

akselaase:
That is Python's recommendations, but the whole Rucio codebase assumes that naive datetime objects (tz=None) are in UTC, and the point of this PR is to fix places where that assumption/convention is broken. I'm not sure introducing tz-aware datetimes is a good idea without knowing how for instance sqlalchemy treats those objects.

joeldierkes:
Okay, then let's not change it. I rebased the commits (sorry for overriding a branch in your repo @akselaase, that somehow feels dirty haha) and resolved the merge problems.

joeldierkes (dismissed):

lib/rucio/core/replica.py
dchristidis: This isn’t anything I’ve worked on, but I can provide my thoughts.

I would recommend leaving `timedelta(days=10)` as is. Obviously, it’s not right, but I would argue that this function needs to be reviewed in a separate issue. Personally, I regard the fact that it can be either an `int` or a `datetime.dateime` object as a potential source of headaches.

joeldierkes: Ok, I created #5847 and #5848 and removed it from this PR

bari12 (approved):

PR #5789, opened 2022-08-09
Migrate hermes to new heartbeat handler; Fix #5557

### LINKED ISSUE #5557: Heartbeat handler for Hermes
Motivation
----------
https://github.com/rucio/rucio/issues/5443

mlassnig:


tools/bootstrap_tests.py
```diff
@@ -148,6 +148,6 @@ def create_influxdb_database():
     if os.getenv('POLICY') == 'belleii':
         belleii_bootstrap(client)
 
-    response = create_influxdb_database()
-    if response.status_code != 201:
-        print('Failed to create rucio database in influxDB : %s' % response.text)
+    #response = create_influxdb_database()
```
rcarpa: I see what you did here :)

lib/rucio/daemons/hermes/hermes.py
```diff
@@ -202,159 +205,121 @@ def deliver_messages(once=False, brokers_resolved=None, thread=0, bulk=1000, del
...
+
+        delete_messages(to_delete)
+        logger(logging.INFO, 'submitted %i messages',
+               len(to_delete))
+
+    if len(messages) < bulk:
+        logger(logging.INFO, "Only %d messages, which is less than the bulk %d, will sleep"
+               % (len(messages), bulk))
+        must_sleep = True
+        return must_sleep
```
rcarpa: same here

lib/rucio/daemons/hermes/hermes.py
```diff
@@ -26,109 +27,95 @@
...
-
-    logger(logging.DEBUG, 'graceful stop done')
+        delete_messages(to_delete)
+        logger(logging.INFO, 'submitted %i messages', len(to_delete))
 
+    if len(messages) < bulk:
+        logger(logging.INFO, "Only %d messages, which is less than the bulk %d, will sleep"
+               % (len(messages), bulk))
+        must_sleep = True
+        return must_sleep
```
rcarpa: return False. There is a difference between None and False. On None it will sleep

tools/bootstrap_tests.py
mlassnig: Creative problems require creative solutions :-)

lib/rucio/daemons/hermes/hermes.py
mlassnig: then I don't understand the significance of `must_sleep = True` :-D

rcarpa: True = True
False = False
None = [True](https://github.com/rucio/rucio/blob/51ca4ed4bb3d3458bedb3b05e67ae023811a0139/lib/rucio/daemons/common.py#L143)

But I'm both hands for changing all daemons which rely on this implicit behavior to make it explicit and remove this "None" behavior

rcarpa: I think you missunderstood me. 
I was speaking about returning "False" when it must not sleep. rather than relying on the implicit "return None" at the end of the function 
```
    must_sleep = False
    if len(messages) < bulk:
        must_sleep = True
    return must_sleep
```

lib/rucio/daemons/hermes/hermes.py
```diff
@@ -17,145 +17,168 @@
...
+        logger(logging.INFO, "Only %d messages, which is less than the bulk %d, will sleep"
+               % (len(messages), bulk))
+        must_sleep = True
 
-    logger(logging.DEBUG, 'graceful stop requested')
-
-    die(executable, hostname, pid, heartbeat_thread)
-
-    logger(logging.DEBUG, 'graceful stop done')
+    return must_sleep
```
rcarpa, line 126: this variable may not be initialized before access

rcarpa (approved):

Issue #5696, opened 2022-07-04
Tests failing when using `docker-compose-storage.yml`


akselaase:
Motivation
----------
Due to the additions in https://github.com/rucio/rucio/pull/5530, `./tools/run_tests_docker.sh -ir` is now failing when using the `docker-compose-storage.yml` template due to a missing InfluxDB service.

The PR introduced a dependency on InfluxDB in `bootstrap_tests.py:151`, which crashes on line 69 when InfluxDB is not available. The required InfluxDB dependency (together with ElasticSearch and ActiveMQ) was added to `docker-compose.yml` and `docker-compose-storage-externalmetadata.yml`, but left out the other compose files.

The required services are also started in `test/before_script.sh` to make them work in the CI pipeline, but with some inconsistencies with regards to using the `:latest` tag or specific `:6.4.2`. 

There is also a question whether the additional services should be present in the base `docker-compose.yml` file at all. InfluxDB, ElasticSearch and ActiveMQ take up an additional 445MB + 859MB + 422MB = 1.7GB storage space, and uses about the same amount of RAM on my computer when running, which is possibly unnecessary.

Modification
------------
Temporary workaround for local development is to wrap the call to `create_influx_database` in a `try/except` clause, but @maany will hopefully figure out a better long-term solution.

mlassnig:
IMHO, extra databases for monitoring, such as Influx, should only be used when called by the `-monit` yaml in the dev environment. Maybe there's a nice way to flag this in the script, e.g., through an environment variable.

PR #5794, opened 2022-08-10
Tests: completely disable influx db bootstrap. #5696

rcarpa:
Every single person trying rucio in our docker containers gets
failures to bootstrap tests. Disable this until it's fixed.


rcarpa:
seems like we need to do something more than just disable. @cserf , can you please help with that, it becomes a real burden to explain everybody that it's "normal" to get exceptions on bootstrap. Maybe just disable all offending tests for now until the linked ticket is fixed ?

PR #5801, opened 2022-08-12
Testing: fix bootstrap_tests, remote_dbs test_suites, influx elastic test containers #5696

maany:
Your highness @bziemons, @rcarpa, please find for your noble consideration, this PR that addresses the following:

- The test suite 'all' was split into `remote_dbs` and `sqlite`. The required changes have been propogated to matrix_nightly.yaml and test.sh to initialize the test environment

- A 'services' statement has been introduced in the test matrix to specify additional services required by the test suites.

- A new `service` called `influxdb_elastic` has been added to matrix.yml and matrix_nightly.yaml

- A new `skipif` condition is defined to execute a test only when `influxdb_elastic` services are present

- test_hermes will only execute if elastic search and influxdb are present in the runtime environment on ports 8086 and 9200 respectively

- bootstrap_tests.py will only initialize influxdb if an influxdb instance is reachable on port 8086. This should fix the connection errors in cases where influxdb is not available.

- remove test suite `all` from `test_rucio_bin` #5712
- disable sqlite for centos7 in matrix.yml and matrix_nightly.yml #5696

### DIFF
--- a/etc/docker/test/matrix.yml
+++ b/etc/docker/test/matrix.yml
@@ -4,6 +4,9 @@ dists:
       python:
         - "3.6"
         - "3.7"
+    deny:
+      suites:
+        - sqlite
   - id: fedora32
     allow:
       python:
--- a/etc/docker/test/matrix.yml
+++ b/etc/docker/test/matrix.yml
@@ -47,6 +47,8 @@ suites:
       - oracle
       - mysql8
       - postgres14
+    services:
+      - influxdb_elastic
   - id: sqlite
     RDBMS:
       - sqlite
--- a/lib/rucio/tests/common.py
+++ b/lib/rucio/tests/common.py
@@ -20,6 +20,7 @@ import os
 import tempfile
 from random import choice
 from string import ascii_uppercase
+import requests
 
 import pytest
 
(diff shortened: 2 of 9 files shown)


rcarpa:
If the mysql test fails, then tests are working correctly now :D i'm already working on fixing the mysql issue

etc/docker/test/matrix.yml
```diff
@@ -4,6 +4,9 @@ dists:
       python:
         - "3.6"
         - "3.7"
+    deny:
+      suites:
+        - sqlite
   - id: fedora32
     allow:
       python:
```
rcarpa, line 12: I think the 'py37py38' tests bellow are also broken

lib/rucio/tests/common.py
```diff
@@ -36,6 +37,31 @@
...
+    try:
+        response = requests.get('http://localhost:9200/')
+        if response.status_code == 200:
+            return True
+    except requests.exceptions.ConnectionError:
+        print('Elasticsearch is not running at localhost:9200')
+        return False
+
+
+skip_missing_elasticsearch_influxdb_in_env = pytest.mark.skipif(not (is_influxdb_available() and is_elasticsearch_available()), reason='influxdb is not available')
```
maany, line 62: @cserf you can use this as a decorator, to mark tests that should run only when elastic and influxdb are present in the test environment ;)

bari12 (approved):

PR #5803, opened 2022-08-12
Transfer queued message does not include the correct services : Closes #5772

### LINKED ISSUE #5772: `transfer-queued` message does not include the correct services
Motivation
----------
With all messages handled by the new hermes2 logic, messages get attached to the services defined in the configuration. This is then always handled when calling `add_message`. However, the `queue_requests` workflow does not call `add_message` but instead bulk inserts the messages to the database, not respecting the services logic.

https://github.com/rucio/rucio/blob/a5687ef8469e62e6466c8c26deaf306b3c0d7c9a/lib/rucio/core/request.py#L323-L324

Modification
------------
Either `add_message` needs to be called here, or better create a bulk logic for `add_message` which can be called here to have the advantage of the bulk inserts.

cserf:
This patch introduce a new bulk `add_messages` method to replace the `bulk_insert_mappings` into `Messages` used in `queue_requests`. Closes #5772
In addition 2 extra tests are added to check that handling of jumbo payload is done properly and that submission to non supported services will break.

### DIFF
--- a/lib/rucio/tests/test_message.py
+++ b/lib/rucio/tests/test_message.py
@@ -16,7 +16,8 @@
 import pytest
 
 from rucio.common.exception import InvalidObject
-from rucio.core.message import add_message, retrieve_messages, delete_messages, truncate_messages
+from rucio.common.utils import generate_uuid
+from rucio.core.message import add_message, add_messages, retrieve_messages, delete_messages, truncate_messages
 
 
 @pytest.mark.noparallel(reason='fails when run in parallel')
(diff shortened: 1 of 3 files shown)


lib/rucio/tests/test_message.py
```diff
@@ -48,35 +56,92 @@ def test_add_message(core_config_mock, caches_mock):
...
+                        Message.services)\
+        .filter_by(event_type=event_type)\
+        .first()
+    assert msg.payload == 'nolimit'
+    assert msg.payload_nolimit == json.dumps(dict_long_payload)
+    messages = retrieve_messages(40)
+    assert messages[0]['payload'] == dict_long_payload
+
+
+@pytest.mark.noparallel(reason='fails when run in parallel')
```
rcarpa, line 133: why would it fail when run in parallel ?

rcarpa: ah, truncate_messages() ... maybe use the message_mock fixture?

rcarpa (approved):

PR #5807, opened 2022-08-16
Testing: Improve manually running autotests; #4047

### LINKED ISSUE #4047: Improve the usability of manually running the CI tests
Motivation
----------
Make `tools/run_autotests.sh` more usable and useful without CI by honoring exit signals and clean up started conatiners.

Modification
------------
Make `tools/test/run_tests.py` and probably `build_images.py` honor signals like `SIGTERM` and `SIGINT` and properly cleanup, also on docker (requested by @mlassnig ;) ). This includes moving container creation from `tools/test/before_script.sh` to python for ownership.

bziemons:
- Testing: Move tools/run_autotest.sh to donkeyrider.py
- Testing: Cleanup additional containers on termination
- Testing: Enable filtering of test cases
- Testing: Remove --cache-from argument as it works differently nowadays.
  With docker, this needs --build-arg BUILDKIT_INLINE_CACHE=1 and with podman it only works with images built with --cache-to beforehand.

Example usage: `tools/run_autotests.sh --build -s remote_dbs --db postgres14 -d centos7 -f test_replica.py::TestReplicaCore`

I will also update the documentation on https://github.com/rucio/documentation/blob/main/docs/contributing.md

Fixes #4047 

For Your Eminence, @maany to review.

### DIFF
--- a/etc/docker/test/matrix.yml
+++ b/etc/docker/test/matrix.yml
@@ -16,14 +16,7 @@ dists:
       python:
         - "3.9"
 python:
-  - id: "3.6"
-    allow:
-      suites:
-        - syntax
-        - docs
-        - client_syntax
-        - remote_dbs
-        - multi_vo
+  - "3.6"
   - id: "3.7"
     allow:
       suites:
--- a/tools/run_tests_docker.sh
+++ b/tools/run_tests_docker.sh
@@ -27,7 +27,7 @@ function usage {
   echo '  -t    Verbose output from pytest'
   echo '  -a    Skip alembic downgrade/upgrade test'
   echo '  -x    exit instantly on first error or failed test'
-  echo '  -p    Indicate policy package tests'
+  echo '  -p    Indicate test selection by TESTS environment variable'
   exit
 }
 
@@ -37,11 +37,11 @@ do
     h) usage;;
     i) init_only="true";;
     r) activate_rse="true";;
-    s) special="true";;
+    s) special="true";selection="true";TESTS="test_dirac.py";;
     t) trace="true";;
     a) noalembic="true";;
     x) exitfirst="-x";;
-    p) votest="true";;
+    p) selection="true";;
   esac
 done
 export RUCIO_HOME=/opt/etc/test
@@ -138,31 +138,22 @@ if test ${init_only}; then
     exit
 fi
 
-if test ${special}; then
-    echo 'Using the special config and only running test_dirac'
-    tools/pytest.sh -v --tb=short ${exitfirst:-} test_dirac.py
+PYTEST_SH_ARGS="${exitfirst:-}"
+
+if test ${trace}; then
+    echo 'Running tests in verbose mode'
+    PYTEST_SH_ARGS="-vvv $PYTEST_SH_ARGS"
 else
-    if test ${trace}; then
-        echo 'Running tests in verbose mode'
-        if test ${votest}; then
-            echo "Running the following votests for ${POLICY}"
-            TESTS=$(echo "${@:2}")
-            echo $TESTS | tr " " "\n"
-            tools/pytest.sh -vvv ${exitfirst:-} $TESTS
-        else
-            tools/pytest.sh -vvv ${exitfirst:-}
-        fi
-    else
-        echo 'Running tests'
-        if test ${votest}; then
-            echo "Running the following votests for ${POLICY}"
-            TESTS=$(echo "${@:2}")
-            echo $TESTS | tr " " "\n"
-            tools/pytest.sh -v --tb=short ${exitfirst:-} $TESTS
-        else
-            tools/pytest.sh -v --tb=short ${exitfirst:-}
-        fi
-    fi
+    echo 'Running tests'
+    PYTEST_SH_ARGS="-v --tb=short $PYTEST_SH_ARGS"
+fi
+
+if test ${selection}; then
+    echo "Running the following tests:"
+    echo $TESTS | tr " " "\n"
+    tools/pytest.sh $PYTEST_SH_ARGS $TESTS
+else
+    tools/pytest.sh $PYTEST_SH_ARGS
 fi
 
 exit $?
--- a/tools/test/donkeyrider.py
+++ b/tools/test/donkeyrider.py
@@ -20,17 +20,19 @@ import shutil
 import signal
 import sys
 
-stop = False
+# make sure the following imports work
+sys.path.append(os.path.abspath(os.path.dirname(__file__)))
+import build_images
+from matrix_parser import parse_matrix
+from run_tests import run_tests
 
 
 def shutdown(signum, frame):
-    global stop
-    stop = True
     print("Caught", signal.Signals(signum).name + ". Stopping..", file=sys.stderr)
     sys.exit(1)
 
 
-def main():
+def check_container_runtime():
     if 'USE_PODMAN' not in os.environ or os.environ['USE_PODMAN'] == "":
         docker_executable = shutil.which("docker")
         podman_executable = shutil.which("podman")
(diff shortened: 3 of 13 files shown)


etc/docker/test/matrix.yml
```diff
@@ -16,14 +16,7 @@ dists:
...
 python:
-  - id: "3.6"
-    allow:
-      suites:
-        - syntax
-        - docs
-        - client_syntax
-        - remote_dbs
-        - multi_vo
+  - "3.6"
```
maany, line 19:
```
- id: 3.6
```

bziemons: doesn't matter, if the entry is not a dict, it will be interpreted as the id: https://github.com/rucio/rucio/blob/2dfb00ef4e032867170aa5b02abb4272e594a9dd/tools/test/matrix_parser.py#L33-L34

I just reverted https://github.com/rucio/rucio/commit/9b5d43ca0a288bdbd3dd97573d5a082464af4081#diff-3273b199b4da4a866dddefe08c4fc04f3d69243f9472b929669ac5cc0364f608R16-R23

tools/run_tests_docker.sh
```diff
@@ -37,11 +37,11 @@ do
     h) usage;;
     i) init_only="true";;
     r) activate_rse="true";;
-    s) special="true";;
+    s) selection="true";TESTS="test_dirac.py";;
```
bziemons: Needs special=true as well

tools/test/donkeyrider.py
```diff
@@ -0,0 +1,133 @@
...
+        elif docker_executable:
+            os.environ['USE_PODMAN'] = "0"
+        elif podman_executable:
+            os.environ['USE_PODMAN'] = "1"
+        else:
+            print("No compatible container executable (podman/docker) found! Exiting..", file=sys.stderr)
+            sys.exit(1)
+
+    project_dir = os.path.abspath(os.path.join(os.path.abspath(os.path.dirname(__file__)), "../.."))
+    print("Detected project directory", project_dir, file=sys.stderr)
```
joeldierkes: Writing to stderr as debug does not make a lot of sense from my POV. How about using `logging`?

tools/test/donkeyrider.py
```diff
@@ -0,0 +1,133 @@
...
+    parser.add_argument('-d', '--dist', dest='dist', type=str, help='select distribution')
+    parser.add_argument('--py', '--python', dest='python', type=str, help='select Python version')
+    parser.add_argument('-s', '--suite', dest='suite', type=str, help='select test suite')
+    parser.add_argument('--db', '--database', dest='database', type=str, help='select RDBMS')
+    args = parser.parse_args()
+    if args.test_matrix_file is None:
+        matrix_file = os.path.join(project_dir, "etc/docker/test/matrix.yml")
+    else:
+        matrix_file = args.test_matrix_file
+    print("Using test matrix from file", matrix_file, file=sys.stderr)
```
joeldierkes: How about setting `os.path.join(project_dir, "etc/docker/test/matrix.yml")` as default for `args.test_matrix_file`. That way we can get rid of the `if` for good and don't need to think about that.

tools/test/donkeyrider.py
```diff
@@ -0,0 +1,133 @@
...
+
+    # make sure the following imports work
```
joeldierkes: Explaining comments have some flaws: They most likely get forgotten when the code is refactored and indicate, that the code is not self explaining.

How about creating functions instead of the comments and blocks under them:
```python
def main():
    check_imports()
    args = parse_args()
    matrix = parse_martrix(args)
    images = build_images(matrix)
    run_tests(matrix, images, args.filter_tests)
```

Now the code clearly communicates the steps it takes, and you don't have to look into the `parse_args` code if `parse_martrix` has a problem.

tools/test/donkeyrider.py
```diff
@@ -0,0 +1,133 @@
...
+    sys.path.append(os.path.abspath(os.path.dirname(__file__)))
+
+    # matrix parsing
+
+    from matrix_parser import parse_matrix
```
joeldierkes: What is the advantage of importing the code in each place than before everything else?

tools/test/donkeyrider.py
```diff
@@ -0,0 +1,133 @@
...
+import shutil
+import signal
+import sys
+
+stop = False
+
+
+def shutdown(signum, frame):
+    global stop
+    stop = True
```
joeldierkes: The `stop` variable is used nowhere else, what is the point of it?

tools/test/donkeyrider.py
```diff
@@ -0,0 +1,133 @@
...
+
+    from run_tests import run_tests
+
+    run_tests(test_matrix, images, args.filter_tests)
+
+
+if __name__ == "__main__":
+    signal.signal(signal.SIGINT, shutdown)
+    signal.signal(signal.SIGTERM, shutdown)
+    main()
```
joeldierkes: The name `donkeyrider` is very creative, and I like it, however it is even more un-descriptive. I would not expect someone new to the project asking themself: what is this and can I eat it?

How about `run_test_pipeline` or `run_test_suite` or ...

tools/test/donkeyrider.py
bziemons: Using stderr as logging is very common in posix environments. I don't think it is necessary to use logging for a few lines of logging inbetween. This is a developer tool after all.

tools/test/donkeyrider.py
bziemons: sure, will do.

tools/test/donkeyrider.py
bziemons: will have another look at this part of the code. I get your point.

tools/test/donkeyrider.py
bziemons: it was just a routine to port each of the different tools that weren't possible to run directly in Python. I will move the imports to the top, as it is usually done.

tools/test/donkeyrider.py
bziemons: leftover. Will remove

tools/test/donkeyrider.py
bziemons: As a developer tool I think it is fine. I was thinking the same that it would be obscure, but somehow we sticked to it. I don't think it is necessary to change the name

tools/test/donkeyrider.py
bziemons: Just realized I need the `sys.path.append(os.path.abspath(os.path.dirname(__file__)))`: That is why I don't like putting these imports to the imports on top of the file. It basically means that every other Python module just importing this module will modify the global sys.path. Running code on import is bad practice imo.

tools/test/donkeyrider.py
bziemons: I have boiled down the main method to something more understandable. Of course it doesn't come close to your example, since there are more variables created and used in different parts of the code.

joeldierkes (dismissed):

tools/run_tests_docker.sh
```diff
@@ -138,31 +138,22 @@ if test ${init_only}; then
...
-        fi
-    fi
+    echo 'Running tests'
+    PYTEST_SH_ARGS="-v --tb=short $PYTEST_SH_ARGS"
+fi
+
+if test ${selection}; then
+    echo "Running the following tests:"
+    echo $TESTS | tr " " "\n"
+    tools/pytest.sh $PYTEST_SH_ARGS $TESTS
```
maany, line 154: Quick question @bziemons: The TESTS env var should contain 1) the full path to the test or 2) path relative to the tests dir?

bziemons: relative path, afaik (see the example in the PR post). It is how pytest interprets it and pytest is run on the test directory.

bziemons:
Rebased on master and added commit https://github.com/rucio/rucio/pull/5807/commits/91c7474705b2e115e4063db7344a7ab06ad73e40

bari12:
LGTM! @bziemons please rebase it once more.
@maany any more comments?

bari12 (approved):

maany (approved):
Looks good! @bziemons thanks for this :D

PR #5812, opened 2022-08-17
Lifetime Model: Migrate to SQLAlchemy 2.0 syntax #5247

dchristidis:
Also take the opportunity to add some type hints.

### DIFF
--- a/lib/rucio/core/lifetime_exception.py
+++ b/lib/rucio/core/lifetime_exception.py
@@ -17,7 +17,7 @@ from re import match
 from datetime import datetime, timedelta
 from configparser import NoSectionError
 
-from sqlalchemy import or_
+from sqlalchemy import or_, select, update
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm.exc import NoResultFound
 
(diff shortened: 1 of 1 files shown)


lib/rucio/core/lifetime_exception.py
```diff
@@ -207,28 +232,42 @@ def __add_exception(dids, account, pattern, comments, expires_at, estimated_volu
...
+    if state not in ALLOWED_STATES:
         raise UnsupportedOperation
 
+    query = update(
+        models.LifetimeExceptions
+    ).where(
+        models.LifetimeExceptions.id == exception_id
+    ).values(
+        state=state,
+        updated_at=datetime.utcnow()
```
rcarpa, line 257: `synchronize_session=False` was lost in translation

rcarpa: `.execution_options(synchronize_session=False)`

dchristidis: Previously there was a `SELECT` then an `UPDATE`. Is it still necessary?

rcarpa: Yes, it is needed for the update. It's just to tell sqlalchemy to not update its python objects mapping to the database. I wonder why we don't dissabe this behavior all together. With so many queries using synchronize_session=False in our code, that mapping should be borderline useless anyway

dchristidis: I’m afraid I do not understand. :(

dchristidis: I believe that our in-person chat resulted in our agreement that this change is warranted. Please let me know in case there’s more to do.

rcarpa (approved):

PR #5818, opened 2022-08-19
Add some logging and monitoring for the FTS tape overwrite/checking use case

### LINKED ISSUE #5817: Add logging and monitoring for repair/overwrite of existing files on tape
Motivation
----------

Rucio and FTS added a feature to compare checksums and file sizes of files on tape to declare these replicas good and the requests done if everything matches. However, there was no feedback on if this was working. 

Modification
------------

Add a couple of log messages for this and add three counters to track it.

ericvaandering:
Fixes #5817

### DIFF
--- a/lib/rucio/transfertool/fts3.py
+++ b/lib/rucio/transfertool/fts3.py
@@ -19,7 +19,7 @@ import functools
 import logging
 import traceback
 import uuid
-from typing import Any, Callable, Tuple, TYPE_CHECKING
+from typing import Any, Callable, Dict, Tuple, TYPE_CHECKING
 from urllib.parse import urlparse
 
 import requests
@@ -66,7 +66,6 @@ BULK_QUERY_COUNTER = MultiCounter(prom='rucio_transfertool_fts3_bulk_query', sta
                                   documentation='Number of bulk queries', labelnames=('state', 'host'))
 QUERY_DETAILS_COUNTER = MultiCounter(prom='rucio_transfertool_fts3_query_details', statsd='transfertool.fts3.{host}.query_details.{state}',
                                      documentation='Number of detailed status queries', labelnames=('state', 'host'))
-
 
 ALLOW_USER_OIDC_TOKENS = config_get_bool('conveyor', 'allow_user_oidc_tokens', False, False)
 REQUEST_OIDC_SCOPE = config_get('conveyor', 'request_oidc_scope', False, 'fts:submit-transfer')
@@ -446,7 +445,8 @@ class Fts3TransferStatusReport(TransferStatusReport):
         return False
 
     @classmethod
-    def _is_recoverable_fts_overwrite_error(cls, request, reason, file_metadata):
+    def _is_recoverable_fts_overwrite_error(cls, request: Dict[str, Any], reason: str,
+                                            file_metadata: Dict[str, Any]) -> bool:
         """
         Verify the special case when FTS cannot copy a file because destination exists and overwrite is disabled,
         but the destination file is actually correct.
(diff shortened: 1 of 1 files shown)


ericvaandering:
@rcarpa, please review

lib/rucio/transfertool/fts3.py
```diff
@@ -66,6 +66,18 @@
                                   documentation='Number of bulk queries', labelnames=('state', 'host'))
 QUERY_DETAILS_COUNTER = MultiCounter(prom='rucio_transfertool_fts3_query_details', statsd='transfertool.fts3.{host}.query_details.{state}',
                                      documentation='Number of detailed status queries', labelnames=('state', 'host'))
+OVERWRITE_CHECK_COUNTER = MultiCounter(prom='rucio_transfertool_fts3_overwrite_check',
```
rcarpa: I would use record_counter() instead of MultiCounter directly. 
In any case, the `statsd` attribute is wrong, because it doesn't have the placeholders for the labels to be filled correctly

rcarpa: Something like `transfertool.fts3.overwrite.check.{rsetype}.{rse}.{reason}'

ericvaandering: Ok. Thanks. Do we have a plan for deprecating statsd? I was hoping I could just say "statsd=None" and not have to worry about it. :-)

I saw some with "record_counter()" but assumed those were older than these. But I'll adjust. Thanks.

ericvaandering:
If this looks OK let me know and I'll squash the commits before asking for it to end up in a release.

lib/rucio/transfertool/fts3.py
```diff
@@ -460,12 +460,18 @@ def _is_recoverable_fts_overwrite_error(cls, request, reason, file_metadata):
...
-                if dst_file.get('file_on_tape'):
-                    return True
-                elif dst_type == 'DISK':
+                if dst_type == 'DISK' or dst_file.get('file_on_tape'):
+                    record_counter('transfertool.fts3.overwrite.ok.{rsetype}.{rse}.{reason}',
+                                   labels={'rse': file_metadata["dst_rse"], 'rsetype': dst_type, 'reason': reason})
                     return True
+
+        record_counter('transfertool.fts3.overwrite.fail.{rsetype}.{rse}.{reason}',
+                       labels={'rse': file_metadata["dst_rse"], 'rsetype': dst_type, 'reason': reason})
```
rcarpa: having RSE and reason in labels seems like a potential for an explosion of number of metrics. Neither prometheus nor statstd will like it. Are you sure about recording the "reason" ?  I don't see it's usefulness. For "ok" it will always be 'Destination file exists and overwrite is not enabled'. 
Maybe move this "fail" metric inside  `if 'Destination file exists and overwrite is not enabled'` ... this way you'll always record metrics for this particular reason.

lib/rucio/transfertool/fts3.py
```diff
@@ -576,8 +584,16 @@ def initialize(self, session, logger=logging.log):
...
                     file_state == FTS_STATE.FAILED and not self._multi_sources:  # for multi-source transfers we must wait for the job to be in a final state
                 if self._is_recoverable_fts_overwrite_error(self.request(session), reason, self._file_metadata):
+                    file_meta = self._file_metadata
+                    self.logger(logging.DEBUG,
+                                f'Recovered existing file: {request_id} to {file_meta["dst_rse"]} '
+                                f'({self.external_host} {self._transfer_id})')
                     new_state = RequestState.DONE
                 else:
+                    file_meta = self._file_metadata
+                    self.logger(logging.INFO,
```
rcarpa: This will be printed for every single FAILED transfer in atlas and other communities. And we already have 20million logs for poller and receiver per day :/  Even in CMS, I think you are not interested in this log for most transfers.

I would "assume" that you want to print this log message inside the if, before `record_counter`, here : https://github.com/rucio/rucio/pull/5818/files#diff-10fdb713ecac24728050341f83e17dbe7cc3f1f5a317823860677917579caea2R472

lib/rucio/transfertool/fts3.py
```diff
@@ -576,8 +584,16 @@ def initialize(self, session, logger=logging.log):
             elif file_state == FTS_STATE.FAILED and job_state == FTS_STATE.FAILED or \
                     file_state == FTS_STATE.FAILED and not self._multi_sources:  # for multi-source transfers we must wait for the job to be in a final state
                 if self._is_recoverable_fts_overwrite_error(self.request(session), reason, self._file_metadata):
+                    file_meta = self._file_metadata
+                    self.logger(logging.DEBUG,
```
rcarpa: I'm not sure this log is needed at all, even in debug. I don't think it bring additional information compared to existing log messages. But I will not fight strongly against it.

rcarpa (dismissed):

rcarpa (dismissed):

ericvaandering:
Let me squash these, then @bari12 can merge when the next set of tests finish.

ericvaandering:
@rcarpa want to give your approval to the new version

rcarpa (approved):

PR #5837, opened 2022-08-25
make account header optional in auth server userpass endpoint

### LINKED ISSUE #5833: Make X-RUCIO-ACCOUNT header optional in auth server
Motivation
----------
In the rucio server, `/auth/userpass` endpoint has a hard requirement for the `account` header.
https://github.com/rucio/rucio/blob/5c665be7891e76a2c31d3d50d5f51ca38fecfa4c/lib/rucio/web/rest/flaskapi/v1/auth.py#L110-L111

This complicates the login workflow for the new webui. The login page asks the user for the identity i.e. `username' and `password` but if the user does not specify the account, the login will fail.

In order to make the account field optional,  this endpoint must be modified

Modification
------------

- if the header is present, use that account
- if the header is not present, list the accounts associated with the identity.
    - if 0 accounts are found, raise appropriate error
    - if a single account is found, use that
    - if multiple accounts are found
        - if default account is specified, use that
        - otherwise, return a list of accounts that the UI can render and ask the person to pick one explicitly
 
This must be done in a way that other endpoints can re-use the logic

maany:
Fix #5833

### DIFF
--- a/lib/rucio/web/rest/flaskapi/v1/auth.py
+++ b/lib/rucio/web/rest/flaskapi/v1/auth.py
@@ -22,7 +22,6 @@ from typing import TYPE_CHECKING
 
 from flask import Flask, Blueprint, request, Response, redirect, render_template
 from urllib.parse import urlparse
-from lib.rucio.api.identity import get_default_account
 from werkzeug.datastructures import Headers
 
 from rucio.api.authentication import get_auth_token_user_pass, get_auth_token_gss, get_auth_token_x509, \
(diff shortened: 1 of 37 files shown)


lib/rucio/web/rest/flaskapi/v1/auth.py
```diff
@@ -107,17 +112,30 @@ def get(self):
         appid = request.headers.get('X-Rucio-AppID', default='unknown')
         ip = request.headers.get('X-Forwarded-For', default=request.remote_addr)
 
-        if not account or not username or not password:
+        if not username or not password:
             return generate_http_error_flask(401, CannotAuthenticate.__name__, 'Cannot authenticate without passing all required arguments', headers=headers)
 
+        if not account:
```
bari12, line 125: I think we still need to check the validity of username & password. This way, if you do not give an `account` you can access the list of accounts the username has access to, without having a valid password. This is an issue.

The second problem I see here is also MultiVO deployments. The `list_accounts_for_identity` method as well as `get_default_account` would need to be made Multi VO compatible. @Thysk any input?

bari12 (changes requested):

maany:
Messed up the rebase, closing this PR and sending a new one

PR #5848, opened 2022-08-30
Core: Fix usage of unused parameter `younger_than` in `core/replica.py` #5847

### LINKED ISSUE #5847: Type of the `younger_than` parameter
Motivation
----------
As @akselaase found out (#5785), the `younger_than` parameter in `get_suspicious_files` is unused if it is an integer. This integer however represents a valid value and should be regared as well. 
https://github.com/rucio/rucio/blob/f4f8f3948c31a50a82e81f9aa0e0deb61f5d7d58/lib/rucio/core/replica.py#L3977

As @dchristidis pointed out, having two different types for the same parameter might be confusing and could lead to some errors. We should change that as well and only allow one type. 

Modification
------------
Use the `younger_than` parameter in all cases.

Change the type of the parameter. We can either make it a:
- `int`
  - Not the most ideal way, since we have to implicitly assume the unit (e.g. seconds or days) or include it in the parameter name.
  - This also limits us to one scale, try representing 1h in days when the type is an integer.
- `float`
  - Same as the first point above
- `date(time)`
  - This seems to be optimal, the unit is integrated in the parameter, and we could refine it if needed.
- `timedelta`
  - This seems to be optimal as well, see the previous point.

The question is how we interpret `younger_than`, is it a fixed time like "I want all suspicious replicas younger than the first of July 2022", or is it a relative point in time like "I want all suspicious replicas younger than a day"...

joeldierkes:
The `younger_than` parameter was not used in case it is not a datetime.

### DIFF
--- a/lib/rucio/core/replica.py
+++ b/lib/rucio/core/replica.py
@@ -3974,7 +3974,7 @@ def get_suspicious_files(rse_expression, available_elsewhere, filter_=None, logg
     if not isinstance(nattempts, int):
         nattempts = 0
     if not isinstance(younger_than, datetime):
-        younger_than = datetime.now() - timedelta(days=10)
+        younger_than = datetime.now() - timedelta(days=younger_than if younger_than is not None else 10)
 
     # assembling exclude_states_clause
     exclude_states_clause = []


mlassnig (commented):
yes, this looks more correct to me :-)

dchristidis (commented):
I’m not sure this change is wise. In the API layer:

https://github.com/rucio/rucio/blob/f4f8f3948c31a50a82e81f9aa0e0deb61f5d7d58/lib/rucio/api/replica.py#L475

So, calling that function without specifying a `younger_than` parameter will result in an exception.

joeldierkes:
Yes, I am on your side. The PR regards the `None` case now, however we should have a general concept about the `younger_than` parameter...

joeldierkes:
This PR is only a temporary fix for a problem no one encountered/noticed so far. Since we don't know the consequences for some other parts of our code and external ones, that depend on this, we won't change it not. #5847 is the tracking issue for the wanted solution (to use only one type for the `younger_than` parameter).

PR #5857, opened 2022-09-05
Authentication: Make account header optional in auth server

### LINKED ISSUE #5833: Make X-RUCIO-ACCOUNT header optional in auth server

maany:
Fix #5833. Implements changes requested in PR #5837

### DIFF
--- a/lib/rucio/api/identity.py
+++ b/lib/rucio/api/identity.py
@@ -17,6 +17,7 @@
 Interface for identity abstraction layer
 """
 
+import typing
 from rucio.api import permission
 from rucio.common import exception
 from rucio.common.types import InternalAccount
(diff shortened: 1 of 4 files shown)


maany:
@bari12 any feedback? :)

bari12 (dismissed):
Sorry the long delay. This looks good to me.
@mlassnig please have a look as well

mlassnig (dismissed):
The whole thing looks incredibly hacky to me, but I can't find an obvious auth issue with it, so approving :-)

bari12 (approved):

Issue #5392, opened 2022-03-25
Add web, client and UI methods to declare multiple replicas as dark and missing


ivmfnal:
Motivation
----------
Consistency Checking procedures produce lists of missing and "dark" replicas:

- Missing is the replica, which exist in the RSE according to Rucio database, but in fact does not
- Dark replica is a file replica found in an RSE, which should not be there, according to Rucio
   - the file may be unknown to Rucio
   - the file may be known, but the replica should not be found in the RSE at the path where it is found

There should be a way to request Rucio to perform actions to correct the inconsistencies:

- In case of a "dark" replica, the replica should "quarantined" so that the Reaper can initiate its removal
- In case of missing replica, the replica in the Rucio database should be marked "bad" so that it can be re-delivered

Modification
------------
- Add Web, Client and UI methods to declare multiple replicas as "dark" or missing. 
- Make sure appropriate safeguards are implemented to prevent the server overload in case the list is large
- Send statistics using Prometheus

ivmfnal:
PR: https://github.com/rucio/rucio/pull/5428

ivmfnal:
Waiting for PR https://github.com/rucio/rucio/pull/5798 to be merged in

ivmfnal:
Apparently using api.declare_bad_file_replicas() method is not correct way to declare missing replicas. The underlying core.declare_bad_file_replicas() checks whether the replica is already known to Necromancer as bad by looking it up in the bad_replicas table. In cases when the replica is declared as bad in bad_replicas and is active in replicas table, Necromancer can not do anything with it.

> According to Dimitrios: Such an inconsistency is a terminal state that cannot be resolved on its own. I’m afraid manually updating the database might be the only way.


The recommendation is to update the replica state to bad in replicas table ignoring the presence of the bad replica in the bad_replicas table.

The proposal is to add "force" flag to api., core.declare_bad_file_replicas() and all the way to the client. This flag would be used by the Consistency Enforcement when declaring missing replicas and make core.declare_bad_file_replicas() bypass the check of the replica state in the bad_replicas.

If that does not help, then the next step would be to have core.declare_bad_file_replicas() to remove the replica record from the bad_replicas thus resetting the Necromancer workflow for the replica.

ivmfnal:
[...]

PR #5864, opened 2022-09-11
Replicas: Added force flag to declare_bad_file_replicas #5392

ivmfnal:
Implemented the "force" flag to ignore the replica status in the bad_replicas table
Chunk large requests to declare bad replicas and quarantine dark replicas by 1000 files
Avoid unneeded calls to translate RSE id to RSE name

### DIFF
--- a/lib/rucio/api/replica.py
+++ b/lib/rucio/api/replica.py
@@ -24,7 +24,7 @@ from rucio.core.rse import get_rse_id, get_rse_name
 from rucio.common import exception
 from rucio.common.schema import validate_schema
 from rucio.common.types import InternalAccount, InternalScope
-from rucio.common.utils import api_update_return_dict
+from rucio.common.utils import api_update_return_dict, invert_dict
 from rucio.common.constants import SuspiciousAvailability
 
 
@@ -64,14 +64,16 @@ def list_bad_replicas_status(state=BadFilesStatus.BAD, rse=None, younger_than=No
 
 
 @transactional_session
-def declare_bad_file_replicas(replicas, reason, issuer, vo='def', session=None):
+def declare_bad_file_replicas(replicas, reason, issuer, vo='def', force=False, session=None):
     """
     Declare a list of bad replicas.
 
     :param replicas: Either a list of PFNs (string) or a list of replicas {'scope': <scope>, 'name': <name>, 'rse_id': <rse_id> or "rse": <rse_name>}.
+            If both rse_id and rse are present, rse will be ignored.
     :param reason: The reason of the loss.
     :param issuer: The issuer account.
     :param vo: The VO to act on.
+    :param force: boolean, ignore existing replica status in the bad_replicas table. Default: False
     :param session: The database session in use.
     :returns: Dictionary {rse_name -> [list of replicas failed to declare]}
     """
@@ -80,7 +82,7 @@ def declare_bad_file_replicas(replicas, reason, issuer, vo='def', session=None):
         return {}
 
     kwargs = {}
-    rse_map = {}
+    rse_map = {}                # RSE name -> RSE id
     if not permission.has_permission(issuer=issuer, vo=vo, action='declare_bad_file_replicas', kwargs=kwargs, session=session):
         raise exception.AccessDenied('Account %s can not declare bad replicas' % (issuer))
 
@@ -107,20 +109,24 @@ def declare_bad_file_replicas(replicas, reason, issuer, vo='def', session=None):
                 "rse_id": rse_id,
                 "name": r["name"]
             })
-    undeclared = replica.declare_bad_file_replicas(replicas_lst, reason=reason, issuer=issuer, status=BadFilesStatus.BAD, session=session)
+
+    rse_id_to_name = invert_dict(rse_map)   # RSE id -> RSE name
+
+    undeclared = replica.declare_bad_file_replicas(replicas_lst, reason=reason, issuer=issuer, status=BadFilesStatus.BAD,
+                                                   force=force, session=session)
     out = {}
     for rse_id, ulist in undeclared.items():
         if ulist:
             rse_name = None
             if rse_id == 'unknown':
                 rse_name = 'unknown'
+            elif rse_id in rse_id_to_name:
+                rse_name = rse_id_to_name[rse_id]
             else:
                 try:
                     uuid.UUID(rse_id)
                     rse_name = get_rse_name(rse_id=rse_id, session=session)
-                except exception.RSENotFound:
-                    rse_name = str(rse_id)
-                except ValueError:
+                except (ValueError, exception.RSENotFound):
                     rse_name = str(rse_id)
             if rse_name:
                 out[rse_name] = out.get(rse_name, []) + ulist
--- a/lib/rucio/client/replicaclient.py
+++ b/lib/rucio/client/replicaclient.py
@@ -21,13 +21,14 @@ from urllib.parse import quote_plus
 
 from rucio.client.baseclient import BaseClient
 from rucio.client.baseclient import choice
-from rucio.common.utils import build_url, render_json
+from rucio.common.utils import build_url, render_json, chunks
 
 
 class ReplicaClient(BaseClient):
     """Replica client class for working with replicas"""
 
     REPLICAS_BASEURL = 'replicas'
+    REPLICAS_CHUNK_SIZE = 1000
 
     def quarantine_replicas(self, replicas, rse=None, rse_id=None):
         """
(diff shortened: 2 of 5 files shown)


lib/rucio/client/replicaclient.py
```diff
@@ -41,46 +41,43 @@ def quarantine_replicas(self, replicas, rse=None, rse_id=None):
...
-            exc_cls, exc_msg = self._get_exception(headers=r.headers, status_code=r.status_code, data=r.content)
-            raise exc_cls(exc_msg)
+        chunk_size = 1000
+        for i in range(0, len(replicas), chunk_size):
+            chunk = replicas[i:i + chunk_size]
+            data = {'rse': rse, 'rse_id': rse_id, 'replicas': chunk}
+            r = self._send_request(url, headers=headers, type_='POST', data=dumps(data))
+            if r.status_code // 100 != 2:
+                exc_cls, exc_msg = self._get_exception(headers=r.headers, status_code=r.status_code, data=r.content)
+                raise exc_cls(exc_msg)
```
joeldierkes: We already have the `chunk` function for this defined in `utils`:
https://github.com/rucio/rucio/blob/e231b3b580bf196e58bce145dfe0c4565d795ba4/lib/rucio/common/utils.py#L459-L472
Example usage:
https://github.com/rucio/rucio/blob/e231b3b580bf196e58bce145dfe0c4565d795ba4/bin/rucio#L1811

lib/rucio/client/replicaclient.py
```diff
@@ -41,46 +41,43 @@ def quarantine_replicas(self, replicas, rse=None, rse_id=None):
...
             raise ValueError("Either RSE name or RSE id must be specified, but not both")
 
-        data = {'rse': rse, 'rse_id': rse_id, 'replicas': replicas}
         url = build_url(self.host, path='/'.join([self.REPLICAS_BASEURL, 'quarantine']))
         headers = {}
-        r = self._send_request(url, headers=headers, type_='POST', data=dumps(data))
-        if r.status_code // 100 != 2:
-            exc_cls, exc_msg = self._get_exception(headers=r.headers, status_code=r.status_code, data=r.content)
-            raise exc_cls(exc_msg)
+        chunk_size = 1000
```
joeldierkes: How should we handle constants here? Let's follow `pep-8`: https://peps.python.org/pep-0008/#constants

lib/rucio/client/replicaclient.py
joeldierkes: Also, as far as I can tell, this code has nothing to do with the force flag. If there is an error with this, and we have to revert this commit, the changes for the force flag would be reverted as well. The best way to solve it would be to create a separate commit with these changes, then explain the reason for the change in the commit message. Have there been any errors/performance issues so far?

lib/rucio/client/replicaclient.py
```diff
@@ -41,46 +41,43 @@ def quarantine_replicas(self, replicas, rse=None, rse_id=None):
...
-        data = {'reason': reason, 'replicas': replicas}
-        url = build_url(self.host, path='/'.join([self.REPLICAS_BASEURL, 'bad']))
-        headers = {}
-        r = self._send_request(url, headers=headers, type_='POST', data=dumps(data))
-        if r.status_code == codes.created:
-            return loads(r.text)
-        exc_cls, exc_msg = self._get_exception(headers=r.headers, status_code=r.status_code, data=r.content)
-        raise exc_cls(exc_msg)
 
-    def declare_bad_did_replicas(self, rse, dids, reason):
```
joeldierkes, line 68: Nit: Same here, the removal of `declare_bad_did_replicas` has nothing to do with the force flag. Maybe an extra commit?

I am a strong supporter of the removal of unused methods, but we should think about the backwards compatibility here. The endpoint in the Rest Api exists, so there might be some external devs using this function. I don't think we should remove it. @cserf ?

lib/rucio/client/replicaclient.py
```diff
@@ -41,46 +41,43 @@ def quarantine_replicas(self, replicas, rse=None, rse_id=None):
...
+        url = build_url(self.host, path='/'.join([self.REPLICAS_BASEURL, 'bad']))
         headers = {}
-        r = self._send_request(url, headers=headers, type_='POST', data=dumps(data))
-        if r.status_code == codes.created:
-            return loads(r.text)
-        exc_cls, exc_msg = self._get_exception(headers=r.headers, status_code=r.status_code, data=r.content)
-        raise exc_cls(exc_msg)
+        chunk_size = 1000
+        for i in range(0, len(replicas), chunk_size):
+            chunk = replicas[i:i + chunk_size]
```
joeldierkes: Same `chunk` function as as above

lib/rucio/web/rest/flaskapi/v1/replicas.py
```diff
@@ -812,6 +812,9 @@ def post(self):
             application/json:
               schema:
                 type: object
+                required:
+                  - replicas
+                  - reason
```
joeldierkes: This `required` is technically not correct, since default values are present. I could call the endpoint without the parameters and everything is fine.

Either we keep this and remove the default values/add an exception if none are given, or they should not be declared as required...

joeldierkes (changes requested):

lib/rucio/client/replicaclient.py
ivmfnal: That's right, chunking has nothing to do with adding the flag, but I thought I would do it to avoid inter-dependencies between patches. So I would like to keep the chunking here.

lib/rucio/client/replicaclient.py
ivmfnal: Sorry, I have no idea how I managed to remove that method. I put it back.

ivmfnal: Thanks for catching

lib/rucio/web/rest/flaskapi/v1/replicas.py
ivmfnal: removed the "required", mostly to avoid creating some backward incompatibility. Although I can not imagine why someone would call this method without providing a list of replicas

lib/rucio/api/replica.py
```diff
@@ -102,18 +104,22 @@ def declare_bad_file_replicas(replicas, reason, issuer, vo='def', session=None):
...
+    undeclared = replica.declare_bad_file_replicas(replicas_lst, reason=reason, issuer=issuer, status=BadFilesStatus.BAD,
+                                                   force=force, session=session)
     out = {}
     for rse_id, ulist in undeclared.items():
         if ulist:
             rse_name = None
             if rse_id == 'unknown':
                 rse_name = 'unknown'
+            elif rse_id in rse_id_to_name:
+                rse_name = rse_id_to_name[rse_id]
```
joeldierkes, line 124: After a quick talk with @faluchet, we think that the `rse_id_to_name` could be redundant.

How about this:
utils.py:
```python
def invert_dict(d):
  return {value: key for (key, value) in d.items()}
```

api/replica.py
```python
elif rse_id in rse_map.values():
  rse_name = invert_dict(rse_map)[rse_id]
```

This way we don't store information redundant. The performance added is negligible small.

joeldierkes (dismissed):

lib/rucio/api/replica.py
ivmfnal: I do not see invert_dict in utils...
Are you proposing to invert the dictionary every time the mapping from rse_id to name is needed ?
I think I will disagree with you.

lib/rucio/core/replica.py
```diff
@@ -303,10 +304,15 @@ def __declare_bad_file_replicas(pfns, rse_id, reason, issuer, status=BadFilesSta
...
-        if __exists and ((status == BadFilesStatus.BAD and not already_declared) or status == BadFilesStatus.SUSPICIOUS):
+        if __exists and \
+                (
+                    status == BadFilesStatus.BAD and (force or not already_declared)
+                    or status == BadFilesStatus.SUSPICIOUS
+                ):
             declared_replicas.append({'scope': scope, 'name': name, 'rse_id': rse_id, 'state': ReplicaState.BAD})
-            new_bad_replica = models.BadReplicas(scope=scope, name=name, rse_id=rse_id, reason=reason, state=status, account=issuer, bytes=size)
-            new_bad_replica.save(session=session, flush=False)
+            if not already_declared:
```
cserf: There's a change in the logic here. Before, we always inserted a new row if `status == BadFilesStatus.SUSPICIOUS`. The more a replica got SUSPICIOUS counts, the more likely it was bad. This is used by the `replica-recover` to apply automatic recovery policy. Can you please keep the old behaviour ?

cserf (changes requested):
Thanks Joel for already providing some of the first comments I had in mind ;)
On my side, I noticed a change in the code logic for the suspicious files. Can you please keep the old logic ?

lib/rucio/api/replica.py
joeldierkes: `inverted_dict` does not exist at this point, however it's a simple one-liner and adds a lot of value. I'd propose to create it.

`rse_map` is not changed in the loop, and the inverted dict is not needed before it. The inverted dictionary can be created before it, thus no, it does not have to be created every time. Just once.

```python
    rse_id_to_name = invert_dict(rse_map)
    for rse_id, ulist in undeclared.items():
        if ulist:
            rse_name = None
            if rse_id == 'unknown':
                rse_name = 'unknown'
            elif rse_id in rse_id_to_name:
                rse_name = rse_id_to_name[rse_id]
```

The part I don't get is this new line:
```diff
                 except exception.RSENotFound:
                     rse_name = str(rse_id)
                 except ValueError:
                     rse_name = str(rse_id)
+                rse_id_to_name[rse_id] = rse_name
             if rse_name:
                 out[rse_name] = out.get(rse_name, []) + ulist
```
`rse_id` is unique, it's the key of the dictionary. Setting it is unnecessary from my POV.

About the timing point: Don't optimize code for performance that's not clearly needed. However, optimize it for readability. Using two dictionaries instead of one is not readable at all from my POV.

lib/rucio/core/replica.py
ivmfnal: Thanks Cedric. Now I understand that logic. I think I restored it.

lib/rucio/api/replica.py
ivmfnal: I thought originally you proposed to inverse the dictionary in the loop where it is used (not where it is created) and I do not think that was a good idea.

Now it seems you are proposing to invert it once, before the loop where it is needed ? Please clarify.

That new line is how the mapping from rse_id to rse name is filled. Or now it is my turn to be missing something ?

ivmfnal: And speaking of performance, I know that at least in the use case I am working on it is clearly needed.

joeldierkes: Yes, in the beginning I did propose to add it into the loop. However, I realized that it can be added before it.

> Now it seems you are proposing to invert it once, before the loop where it is needed ?

Yes.

> That new line is how the mapping from rse_id to rse name is filled.

I think this line (line 131 of the `api/replcia.py` **after** your change) is useless. You set the value for the `rse_id` key in the `rse_id_to_name` dictionary, but since the `rse_id` is unique (it is the key in the `undeclared` dictionary, the one which items we iterate over) it will never be read. (Each `rse_id` will get processed exactly _once_ in the for loop, thus writing it there makes no sense.)
We only read `rse_id_to_name` in the for loop, thus we can generate it before, from the `rse_map` variable.

faluchet: Thanks @joeldierkes  for bringing the conversation forward.

@ivmfnal  thanks for the edits so far, just two notes:

a) From a user perspective it is not clear what happens when a caller sets both `rse_id` and `rse`, possibly mismatching.
L103 reveals that `rse_id` takes always priority, but unless one inspects the code it is definitely not clear. Is this the intended behavior, or perhaps the call should fail? Would you mind clarify, at least with a comment on the function documentation?

b) Also, I find inconsistent how the `rse_id`s returned from `core.declare_bad_file_replicas` are checked against exceptions in L125, while the `rse` coming as parameter from the function caller are not in L105. Typically, it is desirable to early sanitize inputs, instead of their by-products. Actually, I am not even sure that the L125 try-except block would ever trigger. Please can you address that?


Other than that, it looks like accommodating both a list of strings (pfns) and a list of dictionaries (replicas metadata) as parameters of the same function made this part of the code a bit convoluted, but this is not the place to address it.

JM2C

lib/rucio/api/replica.py
```diff
@@ -107,20 +108,24 @@ def declare_bad_file_replicas(replicas, reason, issuer, vo='def', session=None):
                 "rse_id": rse_id,
                 "name": r["name"]
             })
-    undeclared = replica.declare_bad_file_replicas(replicas_lst, reason=reason, issuer=issuer, status=BadFilesStatus.BAD, session=session)
+
```
faluchet, line 112: From a user perspective it is not clear what happens when a caller sets both `rse_id` and `rse`, possibly mismatching.
L103 reveals that `rse_id` takes always priority, but unless one inspects the code it is definitely not clear. Is this the intended behavior, or perhaps the call should fail? Would you mind clarify, at least with a comment on the function documentation?

lib/rucio/api/replica.py
```diff
@@ -107,20 +108,24 @@ def declare_bad_file_replicas(replicas, reason, issuer, vo='def', session=None):
...
+            elif rse_id in rse_id_to_name:
+                rse_name = rse_id_to_name[rse_id]
             else:
                 try:
                     uuid.UUID(rse_id)
                     rse_name = get_rse_name(rse_id=rse_id, session=session)
-                except exception.RSENotFound:
-                    rse_name = str(rse_id)
-                except ValueError:
+                except (ValueError, exception.RSENotFound):
```
faluchet, line 129: I do find inconsistent how the `rse_id`s returned from `core.declare_bad_file_replicas` are checked against exceptions, while the `rse` coming as parameter from the function caller are not in L105. Typically, it is desirable to early sanitize inputs, instead of their by-products. Actually, I am not even sure that the L125 try-except block would ever trigger. Please can you address that?

faluchet (dismissed):

cserf (approved):
LGTM

PR #5868, opened 2022-09-13
Documentation: Document the Api Key in the Rest Api Docs #4992

joeldierkes:


joeldierkes:
The Authentication header is displayed for every endpoint, it is collapsed by default.
It is also automatically added to the examples.
Screenshot Rest Api with Authentification

maany (approved):

joeldierkes:
This PR also solves, by coincidence, the current error in the pipeline from https://github.com/Redocly/redocly-cli/issues/846. The `security` field should actively be set in the OpenApi Spec file, if it is not set an error is thrown. Since this PR sets this field, the rule is satisfied.

PR #5870, opened 2022-09-13
Testing: Enable replica tests for ATLAS

bari12:


joeldierkes:
The `Every API should have security defined on the root level or for each operation.` in the `test (*, docs, *)` jobs are not related to this PR. #5868 resolves them.

maany (approved):

PR #5871, opened 2022-09-13
REST & API: added pfns key as a synonym Fix #5867

### LINKED ISSUE #5867: Bad replicas declaration silently fails with server on 1.29.2 and client <1.29.2
Motivation
----------
This commit modified the REST API for declaring bad replicas :  
https://github.com/rucio/rucio/commit/335bc1deb7623f363da95ca6709c9a8de83f2d21
It replaces the `pfns` parameter by `replicas`. The client API is properly modified, so if one uses `1.29.2` on the server and client everything is fine. But using a client prior `1.29.2` with a server on `1.29.2` will silently fail the declaration of bad files (which did happen to ATLAS). Would it be possible to still use `pfns` if `replicas` is not send ?

Modification
------------

ivmfnal:

### DIFF
--- a/lib/rucio/web/rest/flaskapi/v1/replicas.py
+++ b/lib/rucio/web/rest/flaskapi/v1/replicas.py
@@ -818,6 +818,12 @@ class BadReplicas(ErrorHandlingMethodView):
                     type: array
                     items:
                       type: string
+                  pfns:
+                    deprecated: true
+                    description: The list of pfns, for backward compatibility with older versions of the ReplicaClient
+                    type: array
+                    items:
+                      type: string
                   reason:
                     description: The reason for the declaration.
                     type: string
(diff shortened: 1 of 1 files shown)


lib/rucio/web/rest/flaskapi/v1/replicas.py
```diff
@@ -818,6 +818,11 @@ def post(self):
                     type: array
                     items:
                       type: string
+                  pfns:
+                    description: The list of pfns, for backward compatibility with older versions of the ReplicaClient
+                    type: array
+                    items:
+                      type: string
```
joeldierkes: Nit: Adding the `deprecated: true` attribute marks this field as deprecated and thus communicates, that it should no longer be used.
```yaml
pfns:
  deprecated: true
  description: ...
```

joeldierkes (dismissed):
LGTM :+1: 

The error in the pipeline comes from a recent change in the linter we use for the OpenApi Spec file (https://github.com/Redocly/redocly-cli/issues/846). This is resolved in #5868. Sorry for the noise.

ivmfnal:
I added "deprecated: true" but I am still seeing the same errors.
I can not say any of this makes any sense to me, TBH

joeldierkes (approved):

PR #5872, opened 2022-09-13
move db/sqla to sqlAlchemy 2.0 and turned future=True.

yuyiguo:
unit tests' results: 
863 passed, 373 skipped, 7 xfailed, 1 xpassed, 77 warnings in 1491.63s (0:24:51)


joeldierkes:
The `Every API should have security defined on the root level or for each operation.` in the `test (*, docs, *)` jobs are not related to this PR. #5868 resolves them.

yuyiguo:
HI @bari12 ,
Can you look into the error from the unit tests? https://github.com/rucio/rucio/actions/runs/3086014413/jobs/4989944947#step:4:3219
Read the source I got: The test creates a file, uploads it to rucio, then downloads it. 
I did not see the file removed before downloading anywhere in the code.  So `File exists already locally` makes sense. However, `assert re.search(search, err) is not None` failed because it cannot `successfully downloaded` message.

lib/rucio/db/sqla/session.py
```diff
@@ -303,21 +306,26 @@ def new_funct(*args, **kwargs):
...
-                raise DatabaseException(str(error))
-            except:
-                session.rollback()  # pylint: disable=maybe-no-member
-                raise
-            finally:
-                session.remove()
+            # using sqlAlchemy 2.0 format
+            # the scoped_session vs context manager problem.
+            # see https://github.com/sqlalchemy/sqlalchemy/issues/6519
+            Session = get_session()
```
rcarpa: Upper case is usually used in Rucio to designate a Class name. This is a frequent convention in general in python. Here, `Session` is an object, so probably should use a lower case naming. I think it would be better to respect the convention. Maybe name it `session_scoped = get_session()` ? This also has the advantage of really emphasizing on the difference between these two objects. Otherwise, a newcomer will be confused by having both `Session` and `session` variables inside this function.  I, for sure, was very confused at the beginning.

lib/rucio/db/sqla/session.py
```diff
@@ -303,21 +306,26 @@ def new_funct(*args, **kwargs):
...
+            with Session() as session:
```
rcarpa: I don't think we must use a context manager here.  The code below already executes `Session.remove()`, which does a `session.close()`. Using a context manager results in calling `session.close()` twice. I assume this may result in useless empty rollbacks

so this code will become 
```
session = session_scoped()
session.begin()
try:
.....
finally:
    session_scoped.remove()
```

lib/rucio/db/sqla/util.py
```diff
@@ -128,12 +133,15 @@ def drop_everything():
 def create_base_vo():
     """ Creates the base VO """
 
-    s = get_session()
+    session = get_session()
```
rcarpa: Above, you used the name `Session` for  this variable, here it is `session`. Better to use the same convention everywhere. Name it "session_scoped" everywhere ?

lib/rucio/db/sqla/util.py
```diff
@@ -128,12 +133,15 @@ def drop_everything():
...
     vo = models.VO(vo='def', description='Default base VO', email='N/A')
-
-    s.add_all([vo])
-    s.commit()
+    # the scoped_session vs context manager problem.
+    # see https://github.com/sqlalchemy/sqlalchemy/issues/6519
+    with session() as s:
+        with s.begin():
+            s.add_all([vo])
+            s.commit()
```
rcarpa: when doing `session.begin()`, it creates an explicit transaction which is automatically committed if needed at the end of the context. No need to call commit explicitly.

lib/rucio/db/sqla/session.py
```diff
@@ -303,21 +306,26 @@ def new_funct(*args, **kwargs):
...
-                session.rollback()  # pylint: disable=maybe-no-member
-                raise DatabaseException(str(error))
-            except:
-                session.rollback()  # pylint: disable=maybe-no-member
-                raise
-            finally:
-                session.remove()
+            # using sqlAlchemy 2.0 format
+            # the scoped_session vs context manager problem.
+            # see https://github.com/sqlalchemy/sqlalchemy/issues/6519
```
rcarpa: I don't understand how this github thread is relevant and why it is linked in multiple places of this pull request. I personally don't see any problem at all. Just somebody trying to incorrectly use the sqlalchemy objects. Sqlalchemy can be very confusing with all the Session vs SessionMaker vs scoped_session thing, but I wouldn't put this github link here. It has potential to make people think that there is some bug/issue in sqlalchemy for which we implemented a workaround here. And it's not the case

lib/rucio/db/sqla/util.py
```diff
@@ -201,21 +209,24 @@ def create_root_account(create_counters=True):
...
-            s.add(identity)
-            s.commit()
-        except IntegrityError:
-            # Identities may already be in the DB when running multi-VO conversion
-            s.rollback()
-    s.add(account)
-    s.commit()
-    s.add_all([iaa1, iaa2, iaa3, iaa4])
-    s.commit()
+    # sqlAlchemy 2.0 format
```
rcarpa: If the goal is to migrate everything to the 'sqlalchemy 2.0' format, this format becomes the new norm. So no need to put this comment at each code block

lib/rucio/db/sqla/util.py
```diff
@@ -233,11 +244,13 @@ def get_db_time():
         else:
             query = select([func.current_date()])
 
-        for now, in s.execute(query):
-            if storage_date_format:
-                return datetime.strptime(now, storage_date_format)
-            return now
-
+        # sqlAlchemy 2.0 format
+        with s() as session:
```
rcarpa: again, please use the same convention everywhere. I already saw 'Session', 'session' and 's' to designate a scoped_session. I do completely agree that these things were incorrectly named at the first place, but this is a good opportunity to make things right and avoid confusion in the future.

lib/rucio/db/sqla/session.py
yuyiguo: Good point, fixed.

lib/rucio/db/sqla/session.py
yuyiguo: I would remove below 
finally:
    session_scoped.remove()

lib/rucio/db/sqla/util.py
yuyiguo: yes, will use "session_scoped" for all the scoped sessions.

lib/rucio/db/sqla/util.py
yuyiguo: You are right. I missed the old code.

lib/rucio/db/sqla/session.py
yuyiguo: It is a "problem" in the sense of using a context manager for the `scoped session`, but the problem cannot be fixed in SQLalchemy. Because of this, one cannot write code as below
```
with Session.begin() as session:
    session.add(Test(value="foo"))
```
instead,  one has to do 
```
with Session() as session:
     with session.begin():
         session.add(Test(value="foo"))
```
It took me some time to find out this, so I thought it will help people when they want to be more succinct.  I may just leave one link instead of in all py files.

lib/rucio/db/sqla/util.py
yuyiguo: I put it there to remind myself. removed.

lib/rucio/db/sqla/util.py
yuyiguo: I will make an overall consistent check and change the code accordingly.

lib/rucio/db/sqla/session.py
yuyiguo: Acctually, if you look at the SQLAlchemy official example, I was doing the same with `Session `.
https://docs.sqlalchemy.org/en/14/orm/contextual.html
```
>>> from sqlalchemy.orm import scoped_session
>>> from sqlalchemy.orm import sessionmaker
>>> session_factory = sessionmaker(bind=some_engine)
>>> Session = scoped_session(session_factory) 
>>> some_session = Session()
```

lib/rucio/db/sqla/session.py
rcarpa: It's not a "problem", because nobody advertised the possibility to use "for Session.begin() as session" for this purpose. It is clearly written in the documentation that scoped_session instances work as a proxies. So Session.begin() is equivalent to Session().begin() and returns a transaction, not a session. This is different from sessionmaker.begin() (https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.sessionmaker.begin). While I understand your confusion, linking this issue in comments will confuse even more whoever reads the code in the future.

lib/rucio/db/sqla/session.py
rcarpa: I don't necessarily agree with this naming convention. But I'm OK to use it as long as this convention is respected everywhere in rucio where scoped_session is used

yuyiguo: Looked into the get_session() function. The lowercase notation is used. I made using session_scoped for all if possible.

lib/rucio/db/sqla/session.py
yuyiguo: The link removed.

lib/rucio/db/sqla/util.py
yuyiguo: Done

yuyiguo:
@rcarpa ,
I updated the code based on your review. Please review my changes.

lib/rucio/db/sqla/util.py
```diff
@@ -201,46 +204,48 @@ def create_root_account(create_counters=True):
...
+        session = session_scoped()
+            session.begin():
+            for now, in session.execute(query):
+                if storage_date_format:
+                    return datetime.strptime(now, storage_date_format)
+                return now
     finally:
-        s.remove()
-
+        session_scope.remove()
```
rcarpa: session_scoped (missed the 'd' at the end)

lib/rucio/db/sqla/util.py
```diff
@@ -201,46 +204,48 @@ def create_root_account(create_counters=True):
...
         else:
             query = select([func.current_date()])
 
-        for now, in s.execute(query):
-            if storage_date_format:
-                return datetime.strptime(now, storage_date_format)
-            return now
-
+        session = session_scoped()
+            session.begin():
```
rcarpa: this line doesn't look right

rcarpa (changes requested):

lib/rucio/db/sqla/util.py
yuyiguo: Thanks for catching the typo.

lib/rucio/db/sqla/util.py
yuyiguo: `session.begin() `

rcarpa (dismissed):

yuyiguo:
@mlassnig
Hi Mario,
I am going to be off for almost two weeks starting this Friday afternoon. I hope that I can get the PR completed before I am off.  Please review it ASAP.

mlassnig (dismissed):
It looks correct to me. If there really is a problem with the new session handling we will only see it in production anyway when we have competing requests on the Oracle instance.

yuyiguo:
@mlassnig,
Thanks, Mario for your quick response.

yuyiguo:
@rcarpa @mlassnig 
Hi,

I did commit again because of the conflict with the master branch.  The changes are at 
https://github.com/rucio/rucio/pull/5872/files#diff-32882e2964a57c2774f2daf224f26c75a7c2a846dc998bac2f3ef8841a05578dR203
The below part had to be removed as they are no longer in the master.
```
209         if create_counters:
210             create_counters_for_new_account(account=account.account, session=s)
```
The PR passed End to End integration tests and VO-specific tests. Some of the autotests failed, but I did not find them related to this PR.  It said, "Merging can be performed automatically with 1 approving review." 
can any of you review it again?
Thanks!

yuyiguo:
@rcarpa @mlassnig 
Any suggestions for next?

rcarpa (dismissed):

bari12:
Thanks @yuyiguo 
Just one syntax test to fix
```
lib/rucio/db/sqla/util.py:202:1: W293 blank line contains whitespace
```

yuyiguo:
@bari12 
I fixed 
`lib/rucio/db/sqla/util.py:202:1: W293 blank line contains whitespace`. 
And rebased with the origin/master, however, I got the below error: 

```
Auto-merging lib/rucio/db/sqla/util.py
CONFLICT (content): Merge conflict in lib/rucio/db/sqla/util.py
STDERR:
error: could not apply 3a93ce34d... moved db/sqla to sqlAlchemy 2.0 and turned future=True;
```
because it 
`could not be cherry-picked onto rucio/rucio:release-1.29-LTS` .
Why was it on release-1.29-LTS, not master?

There was no conflict while I rebased on the HEAD of repo.

bari12:
I think you are referring to the `release-patch-setup` test. You can ignore this one.
Since the PR is not marked specifically as a feature (Your branch name does not start with `feature-` like explained in the [contribution guide](https://rucio.cern.ch/documentation/contributing/#3-create-a-local-working-branch) the CI assumes it is a patch and must also be tested against the ongoing release line (`1.29`)
But your changes are meant as a feature and never to be tested against the ongoing release line (just against master), so you can ignore this.

bari12 (approved):

PR #5874, opened 2022-09-14
REST & API: JSON docs in try_stream; #5859

### LINKED ISSUE #5859: Investigate streaming errors
Motivation
----------
We repeatedly have issues where results are streamed (`application/x-json-stream` via REST) to the clients, but due to the size of the request it runs into HTTPD/Database timeout and the stream gets killed. However, in some cases the received result on client side still looks like a proper result, although it is incomplete.

#3659 is an example of this issue. 

I believe that the root of these issues is that we are not properly handling exceptions during the stream.

https://github.com/rucio/rucio/blob/9b666b66989f819abc856a76f010fa6be4945666/lib/rucio/web/rest/flaskapi/v1/locks.py#L112-L119

Thus the exception would happen within the Response object generated by `try_stream`. This behaviour needs to be analysed.

bziemons:
Make all streaming APIs returning JSON documents with try_stream return an incomplete JSON document per streamed result. All JSON documents are completed when the follow-up arrived or the stream ends. This makes e.g., list_replicas JSON streaming in line with metalink streaming, which also returns an incomplete document when errors occurred.
Fixes #5859

### DIFF
--- a/lib/rucio/web/rest/flaskapi/v1/common.py
+++ b/lib/rucio/web/rest/flaskapi/v1/common.py
@@ -17,25 +17,24 @@ import itertools
 import json
 import logging
 import re
+import typing
+from configparser import NoOptionError, NoSectionError
 from functools import wraps
 from time import time
 from typing import TYPE_CHECKING
 
 import flask
-import typing
 from flask.views import MethodView
 from werkzeug.datastructures import Headers
 from werkzeug.exceptions import HTTPException
 from werkzeug.wrappers import Request, Response
 
 from rucio.api.authentication import validate_auth_token
+from rucio.common import config
 from rucio.common.exception import DatabaseException, RucioException, CannotAuthenticate, UnsupportedRequestedContentType
 from rucio.common.schema import get_schema_value
 from rucio.common.utils import generate_uuid, render_json
 from rucio.core.vo import map_vo
-from rucio.common import config
-from configparser import NoOptionError, NoSectionError
-
 
 if TYPE_CHECKING:
     from typing import Optional, Union, Dict, Sequence, Tuple, Callable, Any, List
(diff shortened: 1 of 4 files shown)


lib/rucio/web/rest/flaskapi/v1/common.py
```diff
@@ -192,7 +191,43 @@ def parse_scope_name(scope_name, vo):
...
+                yield part[:document_end_index]
+            last_document_end = part[document_end_index:]
+        else:
+            if last_document_end:
+                yield last_document_end + part
+            else:
+                yield part
+
+    if last_document_end:
+        yield last_document_end
```
joeldierkes, line 227: Okay, I have a **hard** time understanding this. My first question is:
`is_json_doc` is set to a boolean value if the current string (without whitespaces) ends with a `}`. If it does (`is_json_doc` is `True`) the  `document_end_index` is set to the index of this position. Doesn't that mean that `last_document_end` is only whitespaces? Thus, why do we need it? The whitespaces after a json document should be fairly insignificant...

joeldierkes: This seems like a suitable solution, but why are we doing it this complicated?
Wouldn't it be enough to return an incorrect string once the error is thrown? Something like:
```python
def _cut_json_stream(generator: typing.Iterator[typing.AnyStr]) -> typing.Iterator[typing.Union[str, typing.AnyStr]]:
    """
    Generator for cutting off JSON document endings until
    the next element from the generator arrived.
    """
    try:
        yield from generator
    except Exception as e:
        yield str(e) + "-"
        raise e
```

The coresponding test would still recieve the two elements, but then an error:
```python
def test_try_stream_error_after_two_items(flask_application):
    first_object = {'some': 'object'}
    second_object = {'another': 'object'}

    def generate_with_error():
        yield json.dumps(first_object) + '\n'
        yield json.dumps(second_object) + '\n'
        raise RuntimeError('Error for testing')

    with flask_application.app_context():
        with flask_application.test_request_context('/test_try_stream_error_after_two_items'):
            response_bytes: typing.List[bytes] = []
            with pytest.raises(RuntimeError, match='Error for testing'):
                _try_stream_to_bytes(flask_application, response_bytes, generate_with_error())
            response_docs = _cut_at_newlines(response_bytes)
            assert len(response_docs) == 3

            first_object_response = json.loads(response_docs[0])
            assert first_object_response == first_object
            first_object_response = json.loads(response_docs[1])
            assert first_object_response == second_object
            with pytest.raises(json.JSONDecodeError):
                print("Cropped JSON response from third object:", repr(response_docs[2]))
                json.loads(response_docs[2])
```

And the test output:
```
b'{"some": "object"}\n'
b'{"another": "object"}\n'
b'Error for testing-'
Cropped JSON response from third object: b'Error for testing-'
```

bziemons: Simply because we do not know who is closing the connection. The method I implemented also works with any reverse proxies or underlying network issues closing the HTTP connection.

joeldierkes: Ah ok, I understand now. Thanks! :sweat_smile:

bziemons:
Rebased on master

bari12:
@maany did you try this with the new webui finally? I remember you had some concerns.

maany:
Not yet, more info soon. For ben to be looped in, the goal is to check if the new WebUI can parse the response stream into valid JSON objects that can be rendered as soon as they are received.

bziemons:
?

bari12:
Sorry Ben, forgot to post a comment.
We tested this a bit and could somewhat make it work, but the risk is still that it doesn't work for everyone. There is another solution (Basically transparent end-of-stream markers, which new clients will understand and old clients ignore) which we are going to try out and see if this works and then judge for which one to go.
I will re-open this, but put it in draft mode in the meanwhile.

bari12:
Hi Ben,
Thanks a lot for trying this out, I am closing it now though in favour of a different solution described in the ticket. (Just to be clear, we did not implement that one yet either)

PR #5876, opened 2022-09-15
Lifetime Model: Make use of heartbeat handler

dchristidis:
The patch is deceptively large due to a changing of indentation level. It is also paired with a lot of low-hanging cosmetic changes, but the behaviour should be identical.

### DIFF
--- a/bin/rucio-atropos
+++ b/bin/rucio-atropos
@@ -25,7 +25,7 @@ import signal
 from rucio.daemons.atropos.atropos import run, stop
 
 
-def get_parser():
+def get_parser() -> argparse.ArgumentParser:
     """
     Returns the argparse parser.
     """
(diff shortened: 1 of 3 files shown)


dchristidis:
https://github.com/rucio/rucio/actions/runs/3061233015/jobs/4940746927#step:9:11

Do I understand correctly that this requires `HeartbeatHandler.live()` to have type hints?

joeldierkes:
The type checker is complaining about the logger. It is set to `None` and the type checker does not seem to notice that it will be set the first time `live` is called, thus the `"""Object of type "None" cannot be called"""` message (which is fine).

Yes, setting the return type of `live` to something like `Typing.Tuple[int, int, Callable]` _could_ work.

rcarpa (approved):

PR #5888, opened 2022-09-21
Remove the local file in `test_download_dataset` before download #5887

### LINKED ISSUE #5887: Infrequent failing `test_download_dataset` test
Motivation
----------
The `test_download_dataset` is failing sometimes with the error:
```
[gw1] linux -- Python 3.9.9 /usr/bin/python
test_bin_rucio.py:1048: in test_download_dataset
    assert re.search(search, err) is not None
E   AssertionError: assert None is not None
E    +  where None = <function search at 0x7f252a347ca0>('file_OQAYTFTGUO successfully downloaded',
```

As @yuyiguo pointed out in https://github.com/rucio/rucio/pull/5872#issuecomment-1252920038, the test does not download the dataset because it is already available locally. There seems to be no pattern on when the test is failing and when not.

```
2022-09-19 22:32:02,995	INFO	Preparing download of data13_hip:file_OQAYTFTGUO
2022-09-19 22:32:02,995	INFO	File exists already locally: data13_hip:file_OQAYTFTGUO
```

Modification
------------
Remove the file before the download, so it does not exist locally and thus needs to be downloaded.

joeldierkes:
The local created file in `test_download_dataset`, which gets uploaded, is not
removed before the download. This results in a failing test, since it is cached
and thus does not need to be downloaded.

### DIFF
--- a/lib/rucio/tests/test_bin_rucio.py
+++ b/lib/rucio/tests/test_bin_rucio.py
@@ -1039,6 +1039,9 @@ class TestBinRucio:
         print(self.marker + cmd)
         exitcode, out, err = execute(cmd)
         print(out, err)
+
+        os.remove(tmp_file1)
+
         # download dataset
         cmd = 'rucio -v download --dir /tmp {0}'.format(tmp_dataset)  # triming '/tmp/' from filename
         print(self.marker + cmd)


yuyiguo (commented):
Should we check if the file is there or not, before removing it? Because sometimes the test does pass like Martin restarted the test today.

joeldierkes:
The problem I have with that is that the file should not be deleted in the first place. Uploading a file does not delete it to my knowledge. I think the test fails because due to some internal behavior the file is recognized sometimes, and sometimes not (which is a very unsatisfying statement, I know).

I think it would be beneficial to know if the file gets deleted and thus can not be deleted anymore, since this should not happen in the first place.

joeldierkes:
@yuyiguo @bari12 does that make sense?

yuyiguo:
How about below? 
```
try:
os.remove(file)
except:
pass
```

joeldierkes:
> I think it would be beneficial to know if the file gets deleted and thus can not be deleted anymore, since this should not happen in the first place.

Okay, this sentence is probably a bit ambiguous.
I meant to say that we want to know if the file does not exist anymore, since we don't _explicitly_ delete it. I think it would not be beneficial to silent the error.

yuyiguo:
I got you now. It is depending on what the test is for. The current fixing will give the failing test to someone whose PR may not relate to the failed issue at all.  But it is ok if the test will eventually be fixed after we know why the file sometimes "disappeared".

bari12 (approved):

Issue #5169, opened 2022-01-20
Remove all instances of mutable objects (e.g. empty lists or dictionaries) as default parameter values


dchristidis:
Motivation
----------

Read [Python Mutable Defaults Are The Source of All Evil](https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil/). The likelihood of this anti-pattern is low, but given the magnitude, it might be worth removing all such instances.

Modification
------------

Identify all instances with `pylint --disable=all --enable=W0102 lib/rucio/`. Use `None` as default value, but *add type hints* and initialise the parameter in the body. For example:

https://github.com/rucio/rucio/blob/d19db9eeb09440ae97477840fff57112f9b25620/lib/rucio/core/rule.py#L735

Should become:

```python
@stream_session
def list_rules(filters: Optional[Dict] = None, session: Session = None) -> Iterable[Dict]:
    …
    filters = filters or {}
    …
```

chinmaym07:
Hey Can I work on this ?

bari12:
Hi @chinmaym07 yes, please go ahead. I would suggest to go module by module, starting with the ones in https://github.com/rucio/rucio/tree/master/lib/rucio/core

PR #5898, opened 2022-09-29
Introduce heartbeat handler to suspicious_replica_recoverer

ChristophAmes:
As described in https://github.com/rucio/rucio/issues/5443
An example is given in https://github.com/rcarpa/rucio/pull/1

### DIFF
--- a/lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
+++ b/lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
@@ -21,41 +21,34 @@ Consequently, automatic replica recovery is triggered via necromancer daemon, wh
 """
 
 import logging
-import os
-import socket
 import threading
 import time
-import traceback
 from datetime import datetime, timedelta
-from re import match
 from sys import argv
-from typing import List
+from typing import List, Any, Optional
 
+import functools
 import json
 
-from sqlalchemy.exc import DatabaseError
-
 import rucio.db.sqla.util
+from rucio.db.sqla.util import get_db_time
 from rucio.common.config import config_get_bool
 from rucio.common.exception import DatabaseException, VONotFound
-from rucio.common.logging import formatted_logger, setup_logging
+from rucio.common.logging import setup_logging
 from rucio.common.types import InternalAccount
-from rucio.common.utils import daemon_sleep
 from rucio.common.constants import SuspiciousAvailability
-from rucio.core.heartbeat import live, die, sanity_check
-from rucio.core.monitor import record_counter
 from rucio.core.did import get_metadata
 from rucio.core.replica import list_replicas, get_suspicious_files, add_bad_pfns, declare_bad_file_replicas, get_suspicious_reason
-from rucio.core.rse_expression_parser import parse_expression
 from rucio.core.rule import add_rule
-
 from rucio.core.vo import list_vos
-from rucio.db.sqla.util import get_db_time
+from rucio.core.rse_expression_parser import parse_expression
+from rucio.daemons.common import run_daemon
+
 
 GRACEFUL_STOP = threading.Event()
 
 
-def declare_suspicious_replicas_bad(once: bool = False, younger_than: int = 3, nattempts: int = 10, vos: List[str] = None, limit_suspicious_files_on_rse: int = 5, sleep_time: int = 3600, active_mode: bool = False):
+def declare_suspicious_replicas_bad(once: bool = False, younger_than: int = 3, nattempts: int = 10, vos: Optional[List[str]] = None, limit_suspicious_files_on_rse: int = 5, sleep_time: int = 3600, active_mode: bool = False) -> None:
     """
     Main loop to check for available replicas which are labeled as suspicious.
 
@@ -71,7 +64,7 @@ def declare_suspicious_replicas_bad(once: bool = False, younger_than: int = 3, n
     :param nattempts: The minimum number of appearances in the bad_replica DB table
                       in order to appear in the resulting list of replicas for recovery.
     :param vos: VOs on which to look for RSEs. Only used in multi-VO mode.
-                If None, we either use all VOs if run from "def",
+                If empty, we either use all VOs if run from "def",
     :param limit_suspicious_files_on_rse: Maximum number of suspicious replicas on an RSE before that RSE
                                           is considered problematic and the suspicious replicas on that RSE
                                           are labeled as 'TEMPORARY_UNAVAILABLE'.
@@ -80,13 +73,31 @@ def declare_suspicious_replicas_bad(once: bool = False, younger_than: int = 3, n
     :returns: None
     """
 
+    run_daemon(
+        once=once,
+        graceful_stop=GRACEFUL_STOP,
+        executable=argv[0],
+        logger_prefix='replica_recoverer',
+        partition_wait_time=1,
+        sleep_time=sleep_time,
+        run_once_fnc=functools.partial(
+            run_once,
+            younger_than=younger_than,
+            nattempts=nattempts,
+            vos=vos,
+            limit_suspicious_files_on_rse=limit_suspicious_files_on_rse,
+            active_mode=active_mode
+        ),
+    )
+
+
+def run_once(heartbeat_handler: Any, younger_than: int, nattempts: int, vos: Optional[List[str]], limit_suspicious_files_on_rse: int, active_mode: int, **_kwargs) -> bool:
     # assembling the worker name identifier ('executable') including the rses from <rse_expression>
     # in order to have the possibility to detect a start of a second instance with the same set of RSES
 
-    executable = argv[0]
+    worker_number, total_workers, logger = heartbeat_handler.live()
 
-    prepend_str = 'replica_recoverer: '
-    logger = formatted_logger(logging.log, prepend_str + '%s')
+    vos = vos or []
 
     multi_vo = config_get_bool('common', 'multi_vo', raise_exception=False, default=False)
     if not multi_vo:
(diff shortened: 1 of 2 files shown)


lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
```diff
@@ -21,41 +21,34 @@
...
 from rucio.core.vo import list_vos
-from rucio.db.sqla.util import get_db_time
+from rucio.core.rse_expression_parser import parse_expression
+from rucio.daemons.common import run_daemon
+
 
 GRACEFUL_STOP = threading.Event()
 
 
-def declare_suspicious_replicas_bad(once: bool = False, younger_than: int = 3, nattempts: int = 10, vos: List[str] = None, limit_suspicious_files_on_rse: int = 5, sleep_time: int = 3600, active_mode: bool = False):
```
dchristidis, line 58: See #5169.

dchristidis (changes requested):

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
```diff
@@ -80,13 +73,32 @@ def declare_suspicious_replicas_bad(once: bool = False, younger_than: int = 3, n
     :returns: None
     """
 
+    run_daemon(
+        once=once,
+        graceful_stop=GRACEFUL_STOP,
+        executable=argv[0],
+        logger_prefix='replica_recoverer: ',
```
dchristidis: I would suggest dropping the `: ` suffix. Refer to the format in https://github.com/rucio/rucio/blob/771a80183a60aff917031d8b182f574d35f325e1/lib/rucio/daemons/common.py#L81

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
```diff
@@ -447,15 +425,14 @@ def run(once=False, younger_than=3, nattempts=10, vos=None, limit_suspicious_fil
...
     if isinstance(db_time, datetime):
         if db_time - client_time > max_offset or client_time - db_time > max_offset:
-            logger(logging.CRITICAL, 'Offset between client and db time too big. Stopping Suspicious-Replica-Recoverer.')
+            # logger(logging.CRITICAL, 'Offset between client and db time too big. Stopping Suspicious-Replica-Recoverer.')
+            logging.critical('Offset between client and db time too big. Stopping Suspicious-Replica-Recoverer.')
             return
 
-    sanity_check(executable='rucio-replica-recoverer', hostname=socket.gethostname())
-
     if once:
```
dchristidis: I have the impression that this is unnecessary and can be handled by `run_daemon()` instead.

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
```diff
@@ -80,13 +73,32 @@ def declare_suspicious_replicas_bad(once: bool = False, younger_than: int = 3, n
...
     # assembling the worker name identifier ('executable') including the rses from <rse_expression>
     # in order to have the possibility to detect a start of a second instance with the same set of RSES
 
-    executable = argv[0]
+    worker_number, total_workers, logger = heartbeat_handler.live()
 
-    prepend_str = 'replica_recoverer: '
-    logger = formatted_logger(logging.log, prepend_str + '%s')
+    # wait a moment in case all workers started at the same time
+    GRACEFUL_STOP.wait(1)
```
dchristidis: I have the impression that this is unnecessary and should be left to `partition_wait_time` instead, see https://github.com/rucio/rucio/blob/771a80183a60aff917031d8b182f574d35f325e1/lib/rucio/daemons/common.py#L112

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
```diff
@@ -103,342 +115,308 @@ def declare_suspicious_replicas_bad(once: bool = False, younger_than: int = 3, n
             vos = [v['vo'] for v in list_vos()]
         logger(logging.INFO, 'replica_recoverer: This instance will work on VO%s: %s' % ('s' if len(vos) > 1 else '', ', '.join([v for v in vos])))
```
dchristidis: The `logger` object is already configured with the appropriate prefix.

dchristidis (changes requested):

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
dchristidis: Sorry, I now realise my comment can be confusing. I’m referring to:

https://github.com/rucio/rucio/blob/76f3e26df77f822f071c640aab9e422e4b77baee/lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py#L431-L448

Is there a benefit to checking `once` in this part of the code? `run_daemon` can also ensure that there is only one iteration.

ChristophAmes: Having `once` at this position made it easier to test the daemon, as the thread would cause the daemon to continue to loop after the first run. I'm realising now that `once` no longer needs to be a parameter in `declare_suspicious_replicas_bad`, so I shall remove it from there.

dchristidis: I’m not sure I understand what you mean. The flow is
```
run()
↳ declare_suspicious_replicas_bad()
  ↳ run_daemon()
    ↳ run_once()
````
If `run_daemon()` has `once=True`, then there should be a single execution of `run_once()`.

Comment (ChristophAmes) [reply lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py]: Ah, I misunderstood something. Then yes, the `if once ... else` can be removed, as it is no longer required.

File: lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py [hunk 6]
```diff
@@ -67,11 +60,11 @@ def declare_suspicious_replicas_bad(once: bool = False, younger_than: int = 3, n
     :param once: If True, the loop is run just once, otherwise the daemon continues looping until stopped.
     :param younger_than: The number of days since which bad_replicas table will be searched
                          for finding replicas declared 'SUSPICIOUS' at a specific RSE ('rse_expression'),
-                         but 'AVAILABLE' on other RSE(s).
+                         but 'AVAILABLE' on other RSE(s).
```
dchristidis: Forgive me for being pedantic, but please remove this trailing whitespace.

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
```diff
@@ -447,29 +424,25 @@ def run(once=False, younger_than=3, nattempts=10, vos=None, limit_suspicious_fil
     max_offset = timedelta(hours=1, seconds=10)
     if isinstance(db_time, datetime):
         if db_time - client_time > max_offset or client_time - db_time > max_offset:
-            logger(logging.CRITICAL, 'Offset between client and db time too big. Stopping Suspicious-Replica-Recoverer.')
+            # logger(logging.CRITICAL, 'Offset between client and db time too big. Stopping Suspicious-Replica-Recoverer.')
```
dchristidis: Is this commented line necessary?

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
```diff
@@ -101,344 +112,310 @@ def declare_suspicious_replicas_bad(once: bool = False, younger_than: int = 3, n
...
+        logger(logging.INFO, 'Total time for VO "%s": %s seconds', vo, time_passed)
 
+    time_passed = time.time() - start
+    logger(logging.INFO, 'Total time: %s seconds', time_passed)
+    must_sleep = True
+    return must_sleep
 
-def run(once=False, younger_than=3, nattempts=10, vos=None, limit_suspicious_files_on_rse=5, sleep_time=3600, active_mode=False):
+
+def run(once: bool = False, younger_than: int = 3, nattempts: int = 10, vos: List[str] = None, limit_suspicious_files_on_rse: int = 5, sleep_time: int = 3600, active_mode: bool = False):
```
dchristidis: Please add type hints for the return type (`-> None`).

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
```diff
@@ -80,13 +73,31 @@ def declare_suspicious_replicas_bad(once: bool = False, younger_than: int = 3, n
...
+            younger_than=younger_than,
+            nattempts=nattempts,
+            vos=vos,
+            limit_suspicious_files_on_rse=limit_suspicious_files_on_rse,
+            active_mode=active_mode
+        ),
+    )
+
+
+def run_once(heartbeat_handler: Any, younger_than: int, nattempts: int, vos: Optional[List[str]], limit_suspicious_files_on_rse: int, active_mode: int, **_kwargs):
```
dchristidis: The return type is either `None` or `bool`. Could you converge to one and add a hint?

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
```diff
@@ -447,29 +424,25 @@ def run(once=False, younger_than=3, nattempts=10, vos=None, limit_suspicious_fil
...
-            t.join(timeout=3.14)
+    logging.info('Suspicious file replicas recovery starting 1 worker.')
+    t = threading.Thread(target=declare_suspicious_replicas_bad,
+                            kwargs={'once': once,
+                                    'younger_than': younger_than,
+                                    'nattempts': nattempts,
+                                    'vos': vos,
+                                    'limit_suspicious_files_on_rse': limit_suspicious_files_on_rse,
+                                    'sleep_time': sleep_time,
+                                    'active_mode': active_mode})
```
dchristidis: CI doesn’t like this indentation.

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
dchristidis: Sorry, maybe I wasn’t clear enough. You have two return statements in `run_once()`, each with a different type:
https://github.com/rucio/rucio/blob/d189dd1df70546576b690135db08cf8ba7959d80/lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py#L127
https://github.com/rucio/rucio/blob/d189dd1df70546576b690135db08cf8ba7959d80/lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py#L411

dchristidis (approved):
Thank you for addressing all my my concerns. The remaining areas of improvement I have identified are part of the original code and unrelated to the introduction of the heartbeat handler. Maybe @cserf or @rcarpa might want to have a look as well.

PR #5905, opened 2022-10-05
Don't merge. This is for debugging Message queues.

yuyiguo:
comment out message delete after reading.


bari12:
If it's not to merge, why do you submit it upstream? 😄

PR #5908, opened 2022-10-06
Transfers: rework `transfertool` RSE attribute usage. Closes #5825

### LINKED ISSUE #5825: get rid of the `transfertool` RSE attribute usage
Motivation
----------

We have multiple hacks in rucio related to handling missing `transfertool` attribute on RSEs. Sometimes it's treated as "any transfertool". Sometimes as "no transfertool". Configuring conveyor in a multi-transfertool env starts to become a mine field. 

to allow a certain transfertool on an RSE, two distinct attributes have to be set: 1) the `transfertool` attribute which could be a list (ex: globus,fts3); and 2) the transfertool-specific configuration attribute (ex: fts=<list_of_servers>, globus_endpoint_id=<some_uuid>). 

The 'transfertool' attribute is usually redundant. So it was cumbersome to set it for each RSE, and it was decided that not setting it at all means 'all transfertools are allowed'. However, some queries required this attribute to be set if the filter_transfertool configuration option was set in submitter. 

The proposition is to get rid of the `transfertool` attribute

rcarpa:
Until this commit, to allow a certain transfertool on an RSE, two distinct attributes had to be set: 1) the `transfertool` attribute which could be a list (ex: globus,fts3); and 2) the transfertool- specific configuration attribute (ex: fts=<list_of_servers>, globus_endpoint_id=<some_uuid>).
The 'transfertool' attribute is usually redundant. So it was cumbersome to set it for each RSE, and it was decided that not setting it at all means 'all transfertools are allowed'.
However, some queries required this attribute to be set if the filter_transfertool configuration option was set in submitter.

To get rid of this confusing behavior, we decided to get rid of the 'transfertool' attribute usage in conveyor all together. Now, each transfertool defines a list of 'required' rse attributes. If all those attributes are set, it is considered that the corresponding transfertool is enabled on the RSE. To enable FTS it's now enough to set the `fts` attribute to a non-empty value. Similarly, globus requires `globus_endpoint_id` to be set.

### DIFF
--- a/lib/rucio/core/request.py
+++ b/lib/rucio/core/request.py
@@ -20,10 +20,10 @@ import traceback
 from collections import namedtuple
 from typing import TYPE_CHECKING
 
-from sqlalchemy import and_, or_, func, update, select, delete
+from sqlalchemy import and_, or_, update, select, delete, exists
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm import aliased
-from sqlalchemy.sql.expression import asc, true, false, null
+from sqlalchemy.sql.expression import asc, true, false, null, func
 
 from rucio.common.config import config_get_bool
 from rucio.common.exception import RequestNotFound, RucioException, UnsupportedOperation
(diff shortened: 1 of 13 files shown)


bari12:
Probably http://rucio.cern.ch/documentation/transfers-submitter needs a slight update due to this.

bari12 (approved):

PR #5913, opened 2022-10-10
Client: Server SSL Expiry added better error message. Fix: #3866

ThePhisch:
added a new exception to handle serverside ssl expiry with a better error message

### DIFF
--- a/lib/rucio/client/baseclient.py
+++ b/lib/rucio/client/baseclient.py
@@ -29,7 +29,7 @@ from tempfile import mkstemp
 
 from dogpile.cache import make_region
 from requests import Session, Response
-from requests.exceptions import ConnectionError
+from requests.exceptions import ConnectionError, SSLError
 from requests.status_codes import codes
 from configparser import NoOptionError, NoSectionError
 from urllib.parse import urlparse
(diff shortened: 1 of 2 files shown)


lib/rucio/client/baseclient.py
```diff
@@ -390,6 +390,7 @@ def _send_request(self, url, headers=None, type_='GET', data=None, params=None,
             self.logger.debug("Request data (length=%d): [%s]" % (len(data), text))
 
         result = None
+        SSLvalid: bool = True
```
mlassnig, line 393: I would flip it around, i.e. start from a level of non-trust

lib/rucio/common/exception.py
```diff
@@ -1059,3 +1059,13 @@ def __init__(self, *args, **kwargs):
...
         self.error_code = 102
+
+
+class ServerSSLCertificateExpiredException(RucioException):
+    """
+    Server SSL Certificate has expired, auth is no longer possible
+    """
+    def __init__(self, *args, **kwargs):
+        super(ServerSSLCertificateExpiredException, self).__init__(*args, **kwargs)
+        self._message = 'Server SSL Certificate has expired, auth is no longer possible'
```
mlassnig, line 1070: if easy to do, add timestamps when the cert expired?

bari12: @ThePhisch ?

ThePhisch: it should do that

ThePhisch:
this is an old PR, refer to #5919

PR #5917, opened 2022-10-10
Web: enforce auth by default Fix #5822

### LINKED ISSUE #5822: Enforce tokens by default in Flask
Motivation
----------

Currently we explicitly require

`bp.before_request(request_auth_env)`

for each call that requires authentication.

We should flip this around and make that the default, and instead introduce something like

`bp.before_request(no_request_auth_env_necessary)`

on the few where it makes sense.

This provides hardening against accidental issues like #5820 [5820](https://github.com/rucio/rucio/issues/5820)

ThePhisch:
Made a wrapper class for the Blueprint that enables authentication.
went through the files that called a blueprint and changed the code to use the new class.

### DIFF
--- a/lib/rucio/web/ui/flask/bp.py
+++ b/lib/rucio/web/ui/flask/bp.py
@@ -18,7 +18,7 @@ from flask import Blueprint, request, render_template, make_response
 
 from rucio.api.authentication import get_auth_token_x509
 from rucio.common.config import config_get, config_get_bool
-from rucio.web.rest.flaskapi.v1.common import generate_http_error_flask
+from rucio.web.rest.flaskapi.v1.common import generate_http_error_flask, request_auth_env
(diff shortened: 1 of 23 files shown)


rcarpa (commented):
Just nitpicking, but I would call it "AuthenticatedBlueprint", because it enforces authentication, not authorization. Authorization is done later, by the "permissions".

ThePhisch:
@rcarpa done

lib/rucio/web/ui/flask/bp.py
```diff
@@ -142,6 +142,18 @@ def view_maker(template, title):
     return lambda: authenticate(template=template, title=title)
 
 
+class AuthenticatedBlueprint(Blueprint):
+    """
+    Enforce authorisation of blueprints by default
+    returns a normal blueprint otherwise
+    """
+
+    def __init__(self, enforce_auth: bool = True, *args, **kwargs) -> None:
```
bari12: I think the `enforce_auth` option should be taken out. When the AuthenticatedBlueprint is used, it always must enforce auth. Otherwise the standard blueprint should be used.

bari12 (dismissed):

bari12 (approved):

Issue #5080, opened 2021-12-06
server should respond with a HTTP 503 in case of ‘Details: QueuePool limit of size X overflow X reached, connection timed out, timeout X’ 


dchristidis:
Motivation
----------

Such errors can happen when one of the backends is exceptionally busy. The BaseClient has a retry mechanism, which should protect against this, but only acts on certain HTTP errors.

https://github.com/rucio/rucio/blob/226a9f8efb0dfe09b52607cd2333e184ab88d105/lib/rucio/client/baseclient.py#L85

Modification
------------

Check if a HTTP 500 is returned when this error happens and, if so, replace with a 503 (service unavailable) instead.

bari12:
We could catch for the Rucio `DatabaseException` in the REST and raise this to the client. Right now it is not caught, and per default raises a `500`.
So this would have to be added here
https://github.com/rucio/rucio/blob/740b825cac747280ca159089bca4635264c07fc1/lib/rucio/web/rest/flaskapi/v1/common.py#L61-L84
by checking for `DatabaseException` and possibly also `QueuePool` in the error. I wonder though if we accelerate a Session pool issue due to the retries, but in general it's probably still the right thing to do.

sayang14:
Hi @bari12 !, Can I have a try with this issue if someone is not currently assigned yet?

bari12:
Yes sure! Please go ahead 👍

sayang14:
Thanks a lot!, I will take it from here

sayang14:
> by checking for `DatabaseException` and possibly also `QueuePool` in the error. I wonder though if we accelerate a Session pool issue due to the retries, but in general it's probably still the right thing to do.

Hi @bari12 , can you help me with this issue by telling about it in details especially the last line you said relating to DatabaseException and QueuePool if some error is not being caught? 
Also I would like to know if this is the function (https://github.com/rucio/rucio/blob/740b825cac747280ca159089bca4635264c07fc1/lib/rucio/web/rest/flaskapi/v1/common.py#L61-L84) for raising the default which you are talking about? Thanks!

bari12:
Yes this is the correct location. Maybe have a chat with @joeldierkes , he wanted to include some of the common errors (DidNotFound, RuleNotFound, etc.) in this handler anyway.

sayang14:
Hi @joeldierkes , can you please help me with this issue by telling about this in details? Thanks!

PR #5918, opened 2022-10-10
Web: Return HTTP 503 for DatabaseException Fix: #5080

ThePhisch:
added a further exception (databaseexception) to be handled within ErrorHandlingMethodView:dispatch_request. No automated tests added. I tested the functionality by manually raising a databaseexception and reading the output.

### DIFF
--- a/lib/rucio/web/rest/flaskapi/v1/common.py
+++ b/lib/rucio/web/rest/flaskapi/v1/common.py
@@ -29,7 +29,7 @@ from werkzeug.exceptions import HTTPException
 from werkzeug.wrappers import Request, Response
 
 from rucio.api.authentication import validate_auth_token
-from rucio.common.exception import RucioException, CannotAuthenticate, UnsupportedRequestedContentType
+from rucio.common.exception import DatabaseException, RucioException, CannotAuthenticate, UnsupportedRequestedContentType
 from rucio.common.schema import get_schema_value
 from rucio.common.utils import generate_uuid, render_json
 from rucio.core.vo import map_vo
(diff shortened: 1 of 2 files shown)


lib/rucio/web/rest/flaskapi/v1/common.py
```diff
@@ -91,6 +91,17 @@ def dispatch_request(self, *args, **kwargs):
...
         except HTTPException:
             raise
+        except DatabaseException as error:
+            msg = f'DatabaseException in {self.__class__.__module__} {self.__class__.__name__} {flask.request.method}'
+            # logged, because this should be the __exception__
+            logging.debug(msg, exc_info=True)
+            return generate_http_error_flask(
+                status_code=503,
+                exc=error.__class__.__name__,
+                exc_msg='The Rucio database is overloaded. Please try again in a few minutes.',
```
bari12: Let's be a bit more specific here.
Check for the `QueuePool` in the database exception and only then raise the 503. Otherwise we are raising exceptions which are not necessarily load related. I would also change the `exc_msg` to something more generic. `A database exception occurred. Please try again in a moment` or something like that.

bari12 (changes requested):

lib/rucio/web/rest/flaskapi/v1/common.py
mlassnig: While I agree on the first, the message for QueuePool problem should mention that it's overloaded. Maybe something like "There are currently too many requests for the Rucio servers to handle. Please try again in a few minutes."

bari12 (approved):

PR #5919, opened 2022-10-12
Client: SSL Expiry Error Message and Warnings Fix: #3866

ThePhisch:
Added functionality to warn the user of SSL certs that are either expired _or about to_ expire.
Some functions are derived from `cryptography.x509` which by default raises a DeprecationWarning in Python 3.6, these warnings have been disabled.


ThePhisch:
i made a new pull request so that my branchname actually fulfills the requirements in the contributing guide (as pointed out by @dchristidis )

.gitignore
```diff
@@ -65,3 +65,6 @@ tools/test.file.1G
 
 # autotest logs/data
 /.autotest
+
+# VS Code
+.vscode/
```
rcarpa: Don't add this to repository-wide .gitignore. It is specific to your personal dev env, so the correct place for it is in ~/.gitignore on your machine.

ThePhisch: true. fixed it

lib/rucio/client/baseclient.py
```diff
@@ -432,8 +448,18 @@ def _send_request(self, url, headers=None, type_='GET', data=None, params=None,
...
         if result is None:
-            raise ServerConnectionException
+            if not SSLvalid:
+                raise ServerSSLCertificateExpiredException(date_to_str(ssl_expirydate))
+            else:
+                raise ServerConnectionException
+
+        # check if ssl expires soon and if client warning is requested
+        if ssl_expirydate < datetime.now() + timedelta(weeks=self.sslexpiry_warnlimit_weeks) and self.sslexpiry_warnclient:
+            self.logger.warning(f'SSL Certificate will expire in less than {self.sslexpiry_warnlimit_weeks} weeks on {date_to_str(ssl_expirydate)}.')
```
mlassnig: This should state which certificate (client or server one... :-)

mlassnig (approved):

PR #5920, opened 2022-10-12
Core & Internals: lazy account_usage counter creation. Closes #5914

### LINKED ISSUE #5914: switch to lazy-filling of account_usage counters
Motivation
----------
Right now, whenever we create an account or an rse, we pre-fill the account_usage table with empty counters. Meaning that this table always contains the cartesian product of accounts and rses tables. Most of those rows containing useless empty data.

For example, in ATLAS: 

```
SELECT count(*) FROM ACCOUNT_USAGE au WHERE au.FILES != 0
COUNT(*)|
--------+
   18493|

SELECT count(*) FROM ACCOUNT_USAGE au WHERE au.FILES = 0
COUNT(*)|
--------+
 3974991|
```

This also means the creation of a new RSE and/or ACCOUNT takes much more time than actually needed. This is especially visible in tests, where we regularly create temporary RSEs and accounts.

Modification
------------
Rework workflow to only put counters in the `account_usage` table when the value is non-empty. A missing row will correspond to an empty counter.

rcarpa:
Until this commit, whenever an RSE or an account was created, the
account_usage table was pre-filled with empty counters to maintain
the cartesian product between the account and rse tables.
This means that most counters are empty. In atlas, only 0.5% of
them has a non-empty value.

This commit re-works the workflow to not fill the table with empty
counters and consider missing rows as a counter with value `0`.

The primary sponsor for this change are the unit tests, where we create
and teardown test accounts and RSEs quite frequently.

### DIFF
--- a/lib/rucio/core/account.py
+++ b/lib/rucio/core/account.py
@@ -57,8 +57,6 @@ def add_account(account, type_, email, session=None):
         new_account.save(session=session)
     except IntegrityError:
         raise exception.Duplicate('Account ID \'%s\' already exists!' % account)
-    # Create the account counters for this account
-    rucio.core.account_counter.create_counters_for_new_account(account=account, session=session)
 
 
 @read_session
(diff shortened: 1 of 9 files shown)


rcarpa:
this PR is based on top of #5915 . Only the last commit is the actual change.

bari12:
Please rebase. I had to adapt the other PR slightly when merging to fix conflicts.

bari12 (approved):

PR #5921, opened 2022-10-13
Transfers: rework throttler core/database interface. Closes #5805 

### LINKED ISSUE #5805: rework conveyor-throttler
Motivation
----------
Conveyor throttler is functional and pretty well tested. 
However, there are many improvements which must be done: 

- [ ] there is no user-facing cli/api to configure it (one has to use core function directly to set the limits)
- [ ] the sql queries are extremely complex. Investigate and fix them if necessary
- [x] #5901

rcarpa:
This commit completely breaks the compatibility with limits previously
set in the database.

Limits are now in their separate table. And apply to an rse_expression.
All rses resolving to that expression are now subject to a common limit.
For performance reasons, the rse expression is converted to rows
in a new rse_transfer_limits table, which now is just a join-table
between `transfer_limits` and `rses` tables.

The core interface will now automatically fill this join-table whenever
a certain limit is set or updated. The re-evaluation of rse expressions
is also done by the throttler at the start of each run.


rcarpa:
This pr is based on top of  #5902 Only the last commit is new change

bari12:
Rebase please
This might need a second PR against the release branch as well I fear.

bari12 (commented):
Looks all good to me, but please also add the schema change this to the oracle file under etc/sql/...

etc/sql/oracle/schema.sql
```diff
@@ -1426,27 +1426,44 @@ CONSTRAINT DISTANCES_DEST_RSES_FK FOREIGN KEY(DEST_RSE_ID) REFERENCES RSES (ID),
...
+    CONSTRAINT TRANSFER_LIMITS_RSE_EXPRESSION_NN CHECK (RSE_EXPRESSION IS NOT NULL),
+    CONSTRAINT TRANSFER_LIMITS_CREATED_NN CHECK (CREATED_AT IS NOT NULL),
+    CONSTRAINT TRANSFER_LIMITS_UPDATED_NN CHECK (UPDATED_AT IS NOT NULL)
+) ORGANIZATION INDEX COMPRESS 1;
+
+CREATE INDEX TRANSFER_LIMITS_SELECTORS_IDX ON TRANSFER_LIMITS (RSE_EXPRESSION, ACTIVITY) COMPRESS 1;
 
+-- 49.2 ) ========================================= RSE_TRANSFER_LIMITS table =========================================
+
+CREATE TABLE RSE_TRANSFER_LIMITS
```
bari12, line 1455: I thought this one is being dropped? In that case you can just take it out of the schema file.

rcarpa: It is not being completely dropped, it is re-created with a different schema

bari12 (approved):

PR #5922, opened 2022-10-13
Core & Internals: get rid of the retrying dependency. Closes #5344

### LINKED ISSUE #5344: Remove `retrying` dependency
Motivation
----------
The `retrying` dependency has not been updated since 2014.

Modification
------------
Remove and either switch to Rucio-internal code, such as `rucio.common.utils.retry()` or include a different, up-to-date dependency.

rcarpa:
It doesn't install anymore on python 3.10.

Implement a very minimalist decorator which re-implements only the parts of that library which were used by rucio.

### DIFF
--- a/lib/rucio/common/utils.py
+++ b/lib/rucio/common/utils.py
@@ -36,8 +36,9 @@ import time
 import zlib
 from collections import OrderedDict
 from enum import Enum
-from functools import partial
+from functools import partial, wraps
 from uuid import uuid4 as uuid
+from typing import TYPE_CHECKING
 from xml.etree import ElementTree
 
 import requests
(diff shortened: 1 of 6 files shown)


lib/rucio/common/utils.py
```diff
@@ -1896,3 +1896,28 @@ def integer(self):
         delete_value = (self.delete or self.delete is None) * 1
 
         return read_value + write_value + delete_value
+
+
+def retrying(retry_on_exception, wait_fixed, stop_max_attempt_number):
```
bari12: I know this is already a bit older, so sorry for that. But please add type annotations here 😄

rcarpa: :trollface:  But it passed all the tests with a green check mark, meaning it's good, isn't it ? :trollface: :grin:

bari12: :trollface:

bari12 (approved):

PR #5929, opened 2022-10-18
Protocols: Add support for S3 virtual hosted buckets #5451 

### LINKED ISSUE #5451: Add support for S3 virtual hosted buckets
Motivation
----------

Modification
------------

https://docs.google.com/document/d/1ofV5iIgpCKCnANNuqIavN_6FvvMTE6QPtQKK_n1P3SE

rizart:
This PR adds support for S3 virtual-hosted buckets.
Some actual testing with an S3 resource is needed to validate the changes.

Many thanks to @rcarpa for helping me in part of this PR.

Refs:
https://fts3-docs.web.cern.ch/fts3-docs/fts-rest/docs/bulk.html#parameters
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/core/session.html#boto3.session.Session.client
https://botocore.amazonaws.com/v1/documentation/api/latest/reference/config.html
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.generate_presigned_url
https://boto3.amazonaws.com/v1/documentation/api/latest/guide/s3-presigned-urls.html

P.S.

This PR does not contribute to the necessary `https -> s3s` protocol translation that is needed for FTS.
My understanding is that this functionality is already present in the following methods:
https://github.com/rucio/rucio/blob/408ec623d1cfe5f2e7f6c9e085a00f886234eae7/lib/rucio/core/transfer.py#L165
https://github.com/rucio/rucio/blob/408ec623d1cfe5f2e7f6c9e085a00f886234eae7/lib/rucio/core/transfer.py#L191

### DIFF
--- a/lib/rucio/core/credential.py
+++ b/lib/rucio/core/credential.py
@@ -128,12 +128,14 @@ def get_signed_url(rse_id, service, operation, url, lifetime=600):
                 raise UnsupportedOperation('Not a valid Path-Style S3 URL')
             bucket = pathcomponents[1]
             key = '/'.join(pathcomponents[2:])
-        else:
+        elif s3_url_style == "virtual":
             hostcomponents = host.split('.')
             bucket = hostcomponents[0]
             if len(pathcomponents) < 2:
                 raise UnsupportedOperation('Not a valid Virtual-Style S3 URL')
             key = '/'.join(pathcomponents[1:])
+        else:
+            raise UnsupportedOperation('Not a valid RSE S3 URL style (allowed values: path|virtual)')
 
         # remove port number from host if present
         colon = host.find(':')
--- a/lib/rucio/core/credential.py
+++ b/lib/rucio/core/credential.py
@@ -110,9 +110,9 @@ def get_signed_url(rse_id, service, operation, url, lifetime=600):
 
     elif service == 's3':
 
-        # get RSE S3 URL style (path or virtual)
+        # get RSE S3 URL style (path or host)
         # path-style: https://s3.region-code.amazonaws.com/bucket-name/key-name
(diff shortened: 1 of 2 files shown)


lib/rucio/core/credential.py
```diff
@@ -107,14 +108,33 @@ def get_signed_url(rse_id, service, operation, url, lifetime=600):
                                                                                  signature)
 
     elif service == 's3':
+
+        # get RSE S3 URL style (path or virtual)
+        # path-style: https://s3.region-code.amazonaws.com/bucket-name/key-name
```
mlassnig, line 114: The code is good, but I wonder if we should find a better name for the "virtual" variant. Wouldn't "host" be more expressive? As in, in one case the bucket name is in the path and in the other case it's in the host. "virtual" by itself doesn't mean much IMHO. WDYT?

rizart: Yeah, `host` would make indeed more sense. The reason `virtual` was put there is because that's what the library expects here:
https://github.com/rucio/rucio/pull/5929/files#diff-65d1a04f8f23b2b7967130d235c39b9be0a294a929dcd9b18a6ee8adafd2128dR173
thus the code becomes less verbose.

We can proceed as you see fit, no strong opinions on this.

mlassnig: Interesting.... I would prefer `host`, yes, purely for user/operational reasons. But maybe someone else has a stronger opinion? :-D

rizart:
@mlassnig this should be ready.

mlassnig (approved):
nice!

bari12:
Awesome, thanks a lot @rizart 🥳

PR #5933, opened 2022-10-19
Subscription : Add the possibility to reevaluate subscription when a dataset is closed : Closes #5926

### LINKED ISSUE #5926: Add the possibility to reevaluate subscription when a dataset is closed
Motivation
----------
Currently subscriptions are evaluated at the creation of a DID. We might want to delay the subscription evaluation at the time the container or dataset is closed. To achieve this, we can add the possibility to set the `is_open` flag to True when a collection is closed

cserf:
Subscription : Add the possibility to reevaluate subscription when a dataset is closed : Closes #5926

This introduces the possibility to automatically set the `is_new flag` to True when a collection (dataset or container) is closed.

### DIFF
--- a/lib/rucio/core/did.py
+++ b/lib/rucio/core/did.py
@@ -2734,6 +2734,7 @@ def set_status(scope, name, session=None, **kwargs):
     :param kwargs:  Keyword arguments of the form status_name=value.
     """
     statuses = ['open', ]
+    reevaluate_dids_at_close = config_get_bool('subscriptions', 'reevaluate_dids_at_close', raise_exception=False, default=False, session=session)
 
     update_stmt = update(
         models.DataIdentifier
@@ -2784,6 +2785,10 @@ def set_status(scope, name, session=None, **kwargs):
                     message['vo'] = scope.vo
 
                 add_message('CLOSE', message, session=session)
+                if reevaluate_dids_at_close:
+                    set_new_dids(dids=[{'scope': scope, 'name': name}],
+                                 new_flag=True,
+                                 session=session)
 
             else:
                 # Set status to open only for privileged accounts
--- a/lib/rucio/tests/test_did.py
+++ b/lib/rucio/tests/test_did.py
@@ -29,7 +29,8 @@ from rucio.common.types import InternalAccount, InternalScope
 from rucio.common.utils import generate_uuid
 from rucio.core.did import (list_dids, add_did, delete_dids, get_did_atime, touch_dids, attach_dids, detach_dids,
                             get_metadata, set_metadata, get_did, get_did_access_cnt, add_did_to_followed,
-                            get_users_following_did, remove_did_from_followed, set_status)
+                            get_users_following_did, remove_did_from_followed, set_status, list_new_dids,
+                            set_new_dids)
 from rucio.core.replica import add_replica, get_replica
 from rucio.db.sqla.constants import DIDType
 from rucio.tests.common import rse_name_generator, scope_name_generator, did_name_generator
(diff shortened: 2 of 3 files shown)


lib/rucio/core/did.py
```diff
@@ -2734,6 +2734,7 @@ def set_status(scope, name, session=None, **kwargs):
     :param kwargs:  Keyword arguments of the form status_name=value.
     """
     statuses = ['open', ]
+    reevaluate_close_did = config_get_bool('core', 'reevaluate_close_did', raise_exception=False, default=False, session=session)
```
dchristidis: I am wondering whether `reevaluate_close_did` has room for improvement. Maybe `reevaluate_closed_dids`, `reevaluate_dids_at_close` or `reevaluate_dids_at_closure`? I don’t really like any of them.

cserf: Of course, none of them are good. They are all snake case, it should be camel case :smile: !!
Seriously, let's just decide on one. My preferred one is `reevaluate_dids_at_close`

mlassnig: `...at_close` implies that it only happens when the DID is closed.

`reevaluate_closed_dids` is what you mean, right?

dchristidis: I’m not sure I can agree with you, Mario. The re-evaluation indeed is supposed to happen once: at the time the dataset is closed. Which is correct. In turn, `reevaluate_closed_dids` implies a continuous operation.

mlassnig: aha! then i misunderstood and `reevaluate_dids_at_close` is indeed the right one

cserf: `reevaluate_dids_at_close` . Is that your final answer ? You can still phone a friend to check if it's the correct answer

mlassnig: Then I'd like to phone  @tbeerman . He's good in naming things.

dchristidis (changes requested):
Please combine the first and third commits.

dchristidis (dismissed):

lib/rucio/core/did.py
```diff
@@ -2734,6 +2734,7 @@ def set_status(scope, name, session=None, **kwargs):
     :param kwargs:  Keyword arguments of the form status_name=value.
     """
     statuses = ['open', ]
+    reevaluate_close_did = config_get_bool('core', 'reevaluate_dids_at_close', raise_exception=False, default=False, session=session)
```
bari12: Do we really need this? I think less configuration is better, so if really needed, I would make this per default `True`, but perhaps we can live without it?

cserf: Not sure it should be `True` by default. Different communities will have different expectations. I don't know how many communities rely on subscriptions, but if we always reevaluate it can cause unwanted replications.

bari12: Does it?
I thought one of the things we need to make sure is that this doesn't happen. Thus the transmogrifier checks to not create rules twice. If this is the case, I would not offer this option (Or at least make it `True` by default).
If this is not the case, then we should look into this. But assuming things work correctly, this shouldn't lead to any side-effects?

cserf: One can have also something like : A dataset is evaluated, but after some time one wants to decrease the number of copies before it's closed. Then when the dataset is closed, a new copy is made. Also, in the case of `split_rule` if a RSE is initially chosen that matches the RSE expression, but shortly after some RSE attribute is changed and then this RSE does not match the RSE expression anymore. In that case, the transmogrifier will create another replica on a new RSE.

bari12: Yes you are right, I think in this case there is no way to detect it. Ok, let's make it opt-in then. 
I would suggest a different category though and adding it to https://rucio.cern.ch/documentation/configuration_parameters as well.

I would suggest
```
[subscriptions]
reevaluate_dids_at_close
```
So same name, just different category.

cserf: Done

lib/rucio/core/did.py
```diff
@@ -2734,6 +2734,7 @@ def set_status(scope, name, session=None, **kwargs):
     :param kwargs:  Keyword arguments of the form status_name=value.
     """
     statuses = ['open', ]
+    reevaluate_close_did = config_get_bool('subscriptions', 'reevaluate_dids_at_close', raise_exception=False, default=False, session=session)
```
dchristidis: This is nitpicking, but wouldn’t it be better if the variable name matches the configuration?

cserf: What do you mean ?

dchristidis: I’m suggesting to rename `reevaluate_close_did` (the local variable) to `reevaluate_dids_at_close`.

cserf: Sigh...

dchristidis: I understand. Apologies for not catching this earlier.

dchristidis (approved):

PR #5938, opened 2022-10-21
traces: patch dataset updates #5936

### LINKED ISSUE #5936: Access information from traces are not propagated to datasets
Motivation
----------

It has come to our attention that, while files have proper access times, datasets do not.

https://github.com/rucio/rucio/blob/58a5ff3bff3bdff2b7251167522dbfe4c0b444f8/lib/rucio/daemons/tracer/kronos.py#L410-L416

Effectively, `__update_datasets()` is not run during the daemon loop; it’s only run when the daemon exits. Most likely, there isn’t enough time to process all the datasets. This is a regression from #5687.

Modification
------------

Replace this obsolete and non-functional wait mechanism with the `run_daemon()`’s `sleep_time`.

maany:
Fix #5936 

```
dataset_wait = config_get_int('tracer-kronos', 'dataset_wait')
    start = datetime.now()
    if (datetime.now() - start).seconds > dataset_wait:
        __update_datasets(dataset_queue=dataset_queue, logger=logger)
        start = datetime.now()
```
Since it never takes more than 10s to go from line 2 -> line 3 in the above code, the datasets never got updated since the heartbeat handler for kronos was introduced in https://github.com/rucio/rucio/pull/5687. What a miss!
 
I've gotten rid of the additional logic to check time intervals between dataset updates. The datasets should get updated every time a heartbeat is sent. I have not tested the changes, but in theory, this should resolve the issue.

If this looks good to the reviewers, we can patch integration/production clusters already to avoid the issue in the future. Updating the last 4 months of dataset a_time's is a different story altogether.

### DIFF
--- a/lib/rucio/daemons/tracer/kronos.py
+++ b/lib/rucio/daemons/tracer/kronos.py
@@ -389,7 +389,8 @@ def run_once_kronos_file(heartbeat_handler: HeartbeatHandler, return_values: Dic
             conn.subscribe(destination=config_get('tracer-kronos', 'queue'), ack='client-individual', id=subscription_id, headers={'activemq.prefetchSize': prefetch_size})
 
 
-def kronos_dataset(once: bool = False, dataset_queue: Queue = None, sleep_time: int = 60):
+def kronos_dataset(dataset_queue: Queue, once: bool = False, sleep_time: int = 60):
+    return_values = {'heartbeat_handler': HeartbeatHandler("kronos-dataset", 10)}
     run_daemon(
         once=once,
         graceful_stop=graceful_stop,
@@ -400,23 +401,21 @@ def kronos_dataset(once: bool = False, dataset_queue: Queue = None, sleep_time: 
         run_once_fnc=functools.partial(
             run_once_kronos_dataset,
             dataset_queue=dataset_queue,
+            return_values=return_values,
         )
     )
 
-    # once again for the backlog
-    __update_datasets(dataset_queue)
+    # once again for potential backlog
+    run_once_kronos_dataset(dataset_queue=dataset_queue, return_values=return_values, heartbeat_handler=return_values['heartbeat_handler'], sleep_time=sleep_time)
 
 
-def run_once_kronos_dataset(heartbeat_handler: HeartbeatHandler, dataset_queue: Queue, **kwargs):
+def run_once_kronos_dataset(dataset_queue: Queue, return_values: dict, heartbeat_handler: HeartbeatHandler, **kwargs):
+    if heartbeat_handler is None:
+        if "heartbeat_handler" not in return_values.keys():
+            return_values["heartbeat_handler"] = HeartbeatHandler("kronos-dataset", 10)
+        heartbeat_handler = return_values["heartbeat_handler"]
+
     _, _, logger = heartbeat_handler.live()
-    dataset_wait = config_get_int('tracer-kronos', 'dataset_wait')
-    start = datetime.now()
-    if (datetime.now() - start).seconds > dataset_wait:
-        __update_datasets(dataset_queue=dataset_queue, logger=logger)
-        start = datetime.now()
-
-
-def __update_datasets(dataset_queue, logger=logging.log):
     len_ds = dataset_queue.qsize()
     datasets = {}
     dslocks = {}


lib/rucio/daemons/tracer/kronos.py
```diff
@@ -496,6 +492,11 @@ def run(once=False, threads=1, sleep_time_datasets=60, sleep_time_files=60):
     if rucio.db.sqla.util.is_old_db():
         raise DatabaseException('Database was not updated, daemon won\'t start')
 
+    try:
+        dataset_wait = config_get_int('tracer-kronos', 'dataset_wait')
```
rcarpa: I don't think dataset_wait is intended to be used as sleep time.

lib/rucio/daemons/tracer/kronos.py
```diff
@@ -409,11 +409,7 @@ def kronos_dataset(once: bool = False, dataset_queue: Queue = None, sleep_time:
 
 def run_once_kronos_dataset(heartbeat_handler: HeartbeatHandler, dataset_queue: Queue, **kwargs):
     _, _, logger = heartbeat_handler.live()
-    dataset_wait = config_get_int('tracer-kronos', 'dataset_wait')
```
rcarpa, line 412: To me it looks like the correct fix would be to move 
```
    start = datetime.now()
```
between lines 392 and 293 above and pass it to run_once_kronos_dataset as an argument

rcarpa (changes requested):

lib/rucio/daemons/tracer/kronos.py
dchristidis: To me, it looks like `dataset_wait` is a remnant from before we had sleep times. I can’t see a reason why it’s needed at all. Refer to commit https://github.com/rucio/rucio/commit/d26d4239bc04c92f7a06e6d1b27c65778b9961fa.

rcarpa: For me it's a mechanism to avoid running kronos.dataset all together during a small time window after the start of the daemon. Probably to avoid issues during releases because of heartbeat rebalancing. It's more an equivalent of `partition_wait_time` than of `sleep`

maany: I'll leave it as it used to be for now. But it does seem a bit redundant since the `run_daemon` method inserts the `sleep_time` between the subsequent runs of the `run_once` method. If we want to control the `partition_wait_time`, maybe we should rename the config parameter in rucio.cfg as partition_wait_time or explicitly set `patiotion_wait_time=dataset_wait` at some place.

dchristidis: I find myself disagreeing with this course of action.

The original intent of `dataset_wait` should be clear. I fail to see an argument for keeping it any longer. `run_daemons()`’s `partition_wait_time` and `sleep_time` can replace it, with the former having a hard-coded value (as we do everywhere else) and the latter being set through `--sleep-time-datasets`.

rcarpa (dismissed):

dchristidis (changes requested):
Please squash the three commits. I’m also wondering whether `run_once_kronos_dataset()` and `__update_datasets()` should be merged (in a subsequent commit).

lib/rucio/daemons/tracer/kronos.py
```diff
@@ -403,17 +403,13 @@ def kronos_dataset(once: bool = False, dataset_queue: Queue = None, sleep_time:
...
 
 
 def run_once_kronos_dataset(heartbeat_handler: HeartbeatHandler, dataset_queue: Queue, **kwargs):
     _, _, logger = heartbeat_handler.live()
-    dataset_wait = config_get_int('tracer-kronos', 'dataset_wait')
-    start = datetime.now()
-    if (datetime.now() - start).seconds > dataset_wait:
-        __update_datasets(dataset_queue=dataset_queue, logger=logger)
-        start = datetime.now()
+    __update_datasets(dataset_queue=dataset_queue, logger=logger)
```
rcarpa: is there any point to have this two-line function ? merge __update_Datasets into this one?

lib/rucio/daemons/tracer/kronos.py
```diff
@@ -403,17 +403,13 @@ def kronos_dataset(once: bool = False, dataset_queue: Queue = None, sleep_time:
         )
     )
 
-    # once again for the backlog
+    # once again for the any backlog
```
dchristidis: Doesn’t seem like an improvement. Maybe remove the ‘the’ or replace ‘any’ with ‘potential’? (or keep it as it was, of course)

lib/rucio/daemons/tracer/kronos.py
maany: if I merge the two methods, I make the dataset update logic dependent on `heartbeat_handler`. So, if we want to execute dataset updates outside of the `run_daemon` method (before the daemon dies), we would need to provide the `heartbeat_handler` outside of the `run_daemon` wrapper.

maany: I could return the `heartbeat_handler` from the merged method and re-use when it is called again manually. Might look hacky though

maany: I have a better way to do this. Sending the updated PR ;)

rcarpa: I see now. But how does this protect us from any race conditions ? nothing is stopping the AMQ consumer from continuing to consume the queue while we call the second _update_datasets. So, should we call it a 3rd time, a 4th, ... ?

maany: I agreel. it does not, so I won't defend this. ;)

rcarpa (dismissed):

lib/rucio/daemons/tracer/kronos.py
rcarpa: I looked though the code and it was probably useful, because that second call is done just after the "sleep". So it reduces the number of unhandled messages. Sorry for the confusion. Probably a good idea to keep the second call for now, but we must find a better solution for all these AMQ consumer daemons .

maany: Yeah sure, but doing it a 3rd, or 4rth time would reduce the number of unhandled messages even further :P. I will add it back to the code

lib/rucio/daemons/tracer/kronos.py
```diff
@@ -403,20 +403,13 @@ def kronos_dataset(once: bool = False, dataset_queue: Queue = None, sleep_time:
...
-    _, _, logger = heartbeat_handler.live()
-    dataset_wait = config_get_int('tracer-kronos', 'dataset_wait')
-    start = datetime.now()
-    if (datetime.now() - start).seconds > dataset_wait:
-        __update_datasets(dataset_queue=dataset_queue, logger=logger)
-        start = datetime.now()
-
-
-def __update_datasets(dataset_queue, logger=logging.log):
+def run_once_kronos_dataset(dataset_queue: Queue, heartbeat_handler: Union[HeartbeatHandler, None] = None, **kwargs):
```
dchristidis: Please use `Optional` instead of `Union`.

lib/rucio/daemons/tracer/kronos.py
```diff
@@ -403,20 +403,13 @@ def kronos_dataset(once: bool = False, dataset_queue: Queue = None, sleep_time:
...
+    if heartbeat_handler is not None:
+        _, _, logger = heartbeat_handler.live()
```
dchristidis: Wouldn’t the lack of `logger` cause a `NameError` exception below?

dchristidis (changes requested):

lib/rucio/daemons/tracer/kronos.py
```diff
@@ -389,7 +389,8 @@ def run_once_kronos_file(heartbeat_handler: HeartbeatHandler, return_values: Dic
             conn.subscribe(destination=config_get('tracer-kronos', 'queue'), ack='client-individual', id=subscription_id, headers={'activemq.prefetchSize': prefetch_size})
 
 
-def kronos_dataset(once: bool = False, dataset_queue: Queue = None, sleep_time: int = 60):
+def kronos_dataset(dataset_queue: Queue, once: bool = False, sleep_time: int = 60):
+    return_values = {'heartbeat_handler': HeartbeatHandler("kronos-dataset", 10)}
```
rcarpa, line 393: This is so dirty that it hurts :D

rcarpa (approved):

lib/rucio/daemons/tracer/kronos.py
maany: It really does man, but following the discussions above I have limited options. 
I propose that we can worry about reducing the size of the unhandled AMQ messages later. In the future, we can maybe tweak what the daemons should do before exiting. For eg, we could create a new callback method to the run_daemon method for this. 
If we agree for now, then I can happily get rid of the custom HeartbeatHandler and return_values dict.
What do you think?

rcarpa: Sorry that I wasn't clear enough. I did approve the PR, so it's good for me to merge it this way. It just hurts to see it :grinning:

dchristidis (approved):
I am not qualified to review the most recent changes, but I will gladly approve everything else. Thanks for addressing all of my concerns.

PR #5939, opened 2022-10-24
Core: Changed and added to identity functionality; Fix #2881

### LINKED ISSUE #2881: Security update for identity recreation
Motivation
----------

After recreating an Identity using the following steps the RUCIO allows the token creation using 

Creating the test identity **fernando**

`rucio-admin identity add --account carlos --type USERPASS --id fernando --email cgamboa@bnl.gov --password secret
Added new identity to account: fernando-carlos`

Listing identities for account **carlos**
`[root@2b8957951db8 rucio]# rucio-admin account list-identities carlos
Identity: fernando,	type: USERPASS`

Deleting the Identity recently created:

`[root@2b8957951db8 rucio]# rucio-admin identity delete --account carlos --type USERPASS --id fernando
Deleted identity: fernando`

Creating same Identity with a different password 

`rucio-admin identity add --account carlos --type USERPASS --id fernando --email cgamboa@bnl.gov --password othersecret`

I can still connect using the old password the one set when the identity was created. Token is created for the this password.

The token is not created for the **identity** that was deleted and then reused/recreated with a different password.

We are using RUCIO 1.20.2

Modification
------------

ThePhisch:
might need to reconsider the permissions for some of the new commands.

### DIFF
--- a/bin/rucio-admin
+++ b/bin/rucio-admin
@@ -439,12 +439,58 @@ def identity_delete(args):
     """
     %(prog)s delete [options] <field1=value1 field2=value2 ...>
 
-    Revoke an identity's access to an account.
+    Current effect: Revoke an identity's access to an account.
+    Obsolete!!
+    identity_detach will take over this role
+    identity_newdelete will take over this name but have a different effect
 
     """
     client = get_client(args)
     client.del_identity(args.account, args.identity, authtype=args.authtype)
     print('Deleted identity: %s' % args.identity)
+    return SUCCESS
+
+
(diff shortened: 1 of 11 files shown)


bari12:
Hi @ThePhisch this was in draft since forever and not much happened here (Sorry for that)
To move forward I propose to close this and re-submit it in two parts. One which purely tackles the issue described (which is about the token purge) and then secondly the other changes you are proposing as a different feature.
Due to the age, this will also need quite a bit of rebasing.

PR #5943, opened 2022-10-26
Feature 5930 unsetting the child rule

ThePhisch:
* putting in `child_rule_id` as `None` in `update_rule` will unset a previously set child rule.
* wrote tests (`test_bin_rucio` and `test_rule`)

* also included: a small edit solving the `execute` function that is mainly used in `test_bin_rucio`. originally, this function had two different return signatures, depending on the value of a keyword-argument `blocking`, which defaulted to `True`. This duality caused the linter/type checker to raise a warning each time the execute-function was called and unpacked into the three elements of the tuple it would normally return (ie when blocking is True). since the function is called frequently during the testing process, a file such as `test_bin_rucio` might have >200 warnings. the solution was to keep the `execute`-function only in the default form (returning a tuple of three elements) and (since the other form was only used once in the entire codebase) replace the function call with `blocking=False` with the code block within the `execute` function that would have been called otherwise.

### DIFF
--- a/lib/rucio/api/rule.py
+++ b/lib/rucio/api/rule.py
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import TYPE_CHECKING
+from typing import Any, Dict, TYPE_CHECKING
 
 from rucio.api.permission import has_permission
 from rucio.common.config import config_get_bool
@@ -233,7 +233,7 @@ def delete_replication_rule(rule_id, purge_replicas, issuer, vo='def', *, sessio
 
 
 @transactional_session
-def update_replication_rule(rule_id, options, issuer, vo='def', *, session: "Session"):
+def update_replication_rule(rule_id: str, options: Dict[str, Any], issuer: str, vo: str = 'def', *, session: "Session"):
     """
     Update lock state of a replication rule.
 
@@ -251,11 +251,11 @@ def update_replication_rule(rule_id, options, issuer, vo='def', *, session: "Ses
         if not has_permission(issuer=issuer, vo=vo, action='approve_rule', kwargs=kwargs, session=session):
             raise AccessDenied('Account %s can not approve/deny this replication rule.' % (issuer))
 
-        issuer = InternalAccount(issuer, vo=vo)
+        issuer_ia = InternalAccount(issuer, vo=vo)
         if options['approve']:
-            rule.approve_rule(rule_id=rule_id, approver=issuer, session=session)
+            rule.approve_rule(rule_id=rule_id, approver=issuer_ia, session=session)
         else:
-            rule.deny_rule(rule_id=rule_id, approver=issuer, reason=options.get('comment', None), session=session)
+            rule.deny_rule(rule_id=rule_id, approver=issuer_ia, reason=options.get('comment', None), session=session)
     else:
         if not has_permission(issuer=issuer, vo=vo, action='update_rule', kwargs=kwargs, session=session):
             raise AccessDenied('Account %s can not update this replication rule.' % (issuer))
--- a/lib/rucio/client/ruleclient.py
+++ b/lib/rucio/client/ruleclient.py
@@ -17,6 +17,7 @@ from json import dumps, loads
 
 from requests.status_codes import codes
 from urllib.parse import quote_plus
+from typing import Dict, List, Any, Optional, Union
 
 from rucio.client.baseclient import BaseClient
 from rucio.client.baseclient import choice
@@ -29,31 +30,50 @@ class RuleClient(BaseClient):
 
     RULE_BASEURL = 'rules'
 
-    def add_replication_rule(self, dids, copies, rse_expression, weight=None, lifetime=None, grouping='DATASET', account=None,
-                             locked=False, source_replica_expression=None, activity=None, notify='N', purge_replicas=False,
-                             ignore_availability=False, comment=None, ask_approval=False, asynchronous=False, delay_injection=None,
-                             priority=3, meta=None):
+    def add_replication_rule(
+        self,
+        dids: List[str],
+        copies: int,
+        rse_expression: str,
+        priority: int = 3,
+        lifetime: Optional[int] = None,
+        grouping: str = 'DATASET',
+        notify: str = 'N',
+        source_replica_expression: Optional[str] = None,
+        activity: Optional[str] = None,
+        account: Optional[str] = None,
+        meta: Optional[str] = None,
+        ignore_availability: bool = False,
+        purge_replicas: bool = False,
+        ask_approval: bool = False,
+        asynchronous: bool = False,
+        locked: bool = False,
+        delay_injection=None,
+        comment=None,
+        weight=None,
+    ):
         """
         :param dids:                       The data identifier set.
         :param copies:                     The number of replicas.
         :param rse_expression:             Boolean string expression to give the list of RSEs.
-        :param weight:                     If the weighting option of the replication rule is used, the choice of RSEs takes their weight into account.
+        :param priority:                   Priority of the transfers.
         :param lifetime:                   The lifetime of the replication rules (in seconds).
         :param grouping:                   ALL -  All files will be replicated to the same RSE.
                                            DATASET - All files in the same dataset will be replicated to the same RSE.
                                            NONE - Files will be completely spread over all allowed RSEs without any grouping considerations at all.
-        :param account:                    The account owning the rule.
-        :param locked:                     If the rule is locked, it cannot be deleted.
+        :param notify:                     Notification setting for the rule (Y, N, C).
         :param source_replica_expression:  RSE Expression for RSEs to be considered for source replicas.
         :param activity:                   Transfer Activity to be passed to FTS.
-        :param notify:                     Notification setting for the rule (Y, N, C).
-        :param purge_replicas:             When the rule gets deleted purge the associated replicas immediately.
+        :param account:                    The account owning the rule.
+        :param meta:                       Metadata, as dictionary.
         :param ignore_availability:        Option to ignore the availability of RSEs.
+        :param purge_replicas:             When the rule gets deleted purge the associated replicas immediately.
         :param ask_approval:               Ask for approval of this replication rule.
         :param asynchronous:               Create rule asynchronously by judge-injector.
-        :param priority:                   Priority of the transfers.
+        :param locked:                     If the rule is locked, it cannot be deleted.
+        :param delay_injection:
         :param comment:                    Comment about the rule.
-        :param meta:                       Metadata, as dictionary.
+        :param weight:                     If the weighting option of the replication rule is used, the choice of RSEs takes their weight into account.
         """
         path = self.RULE_BASEURL + '/'
         url = build_url(choice(self.list_hosts), path=path)
@@ -70,7 +90,9 @@ class RuleClient(BaseClient):
         exc_cls, exc_msg = self._get_exception(headers=r.headers, status_code=r.status_code, data=r.content)
         raise exc_cls(exc_msg)
 
-    def delete_replication_rule(self, rule_id, purge_replicas=None):
+    def delete_replication_rule(
+        self, rule_id: str, purge_replicas: Optional[bool] = None
+    ):
         """
         Deletes a replication rule and all associated locks.
 
@@ -91,7 +113,7 @@ class RuleClient(BaseClient):
         exc_cls, exc_msg = self._get_exception(headers=r.headers, status_code=r.status_code, data=r.content)
         raise exc_cls(exc_msg)
 
-    def get_replication_rule(self, rule_id):
+    def get_replication_rule(self, rule_id: str):
         """
         Get a replication rule.
 
@@ -107,7 +129,7 @@ class RuleClient(BaseClient):
             exc_cls, exc_msg = self._get_exception(headers=r.headers, status_code=r.status_code, data=r.content)
             raise exc_cls(exc_msg)
 
-    def update_replication_rule(self, rule_id, options):
+    def update_replication_rule(self, rule_id: str, options: Dict[str, Any]):
         """
         :param rule_id:   The id of the rule to be retrieved.
         :param options:   Options dictionary.
@@ -122,7 +144,9 @@ class RuleClient(BaseClient):
         exc_cls, exc_msg = self._get_exception(headers=r.headers, status_code=r.status_code, data=r.content)
         raise exc_cls(exc_msg)
 
-    def reduce_replication_rule(self, rule_id, copies, exclude_expression=None):
+    def reduce_replication_rule(
+        self, rule_id: str, copies: int, exclude_expression=None
+    ):
         """
         :param rule_id:             Rule to be reduced.
         :param copies:              Number of copies of the new rule.
@@ -139,7 +163,9 @@ class RuleClient(BaseClient):
         exc_cls, exc_msg = self._get_exception(headers=r.headers, status_code=r.status_code, data=r.content)
         raise exc_cls(exc_msg)
 
-    def move_replication_rule(self, rule_id, rse_expression, override):
+    def move_replication_rule(
+        self, rule_id: str, rse_expression: str, override
+    ):
         """
         Move a replication rule to another RSE and, once done, delete the original one.
 
@@ -162,7 +188,7 @@ class RuleClient(BaseClient):
         exc_cls, exc_msg = self._get_exception(headers=r.headers, status_code=r.status_code, data=r.content)
         raise exc_cls(exc_msg)
 
-    def approve_replication_rule(self, rule_id):
+    def approve_replication_rule(self, rule_id: str):
         """
         :param rule_id:             Rule to be approved.
         :raises:                    RuleNotFound
@@ -177,7 +203,7 @@ class RuleClient(BaseClient):
         exc_cls, exc_msg = self._get_exception(headers=r.headers, status_code=r.status_code, data=r.content)
         raise exc_cls(exc_msg)
 
-    def deny_replication_rule(self, rule_id):
+    def deny_replication_rule(self, rule_id: str):
         """
         :param rule_id:             Rule to be denied.
         :raises:                    RuleNotFound
@@ -192,7 +218,9 @@ class RuleClient(BaseClient):
         exc_cls, exc_msg = self._get_exception(headers=r.headers, status_code=r.status_code, data=r.content)
         raise exc_cls(exc_msg)
 
-    def list_replication_rule_full_history(self, scope, name):
+    def list_replication_rule_full_history(
+            self, scope: Union[str, bytes], name: Union[str, bytes]
+    ):
         """
         List the rule history of a DID.
 
@@ -207,7 +235,7 @@ class RuleClient(BaseClient):
         exc_cls, exc_msg = self._get_exception(r.headers, r.status_code)
         raise exc_cls(exc_msg)
 
-    def examine_replication_rule(self, rule_id):
+    def examine_replication_rule(self, rule_id: str):
         """
         Examine a replication rule for errors during transfer.
 
@@ -222,7 +250,7 @@ class RuleClient(BaseClient):
         exc_cls, exc_msg = self._get_exception(r.headers, r.status_code)
         raise exc_cls(exc_msg)
 
-    def list_replica_locks(self, rule_id):
+    def list_replica_locks(self, rule_id: str):
         """
         List details of all replica locks for a rule.
 
--- a/lib/rucio/core/rule.py
+++ b/lib/rucio/core/rule.py
@@ -1175,7 +1175,7 @@ def get_rule(rule_id, *, session: "Session"):
 
 
 @transactional_session
-def update_rule(rule_id, options, *, session: "Session"):
+def update_rule(rule_id: str, options: Dict[str, Any], *, session: "Session") -> None:
     """
     Update a rules options.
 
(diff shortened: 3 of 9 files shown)


dchristidis (commented):
I’m guessing that one of the two commits shouldn’t be part of this PR?

ThePhisch:
> 

no both are intended to be in this PR. I considered not opening an issue for the first one but I decided to do so anyway because I changed the functionality of a frequently-used function in order to provide better type hints.

dchristidis (commented):
I think I prefer the approach used in `move-rule`, where `None` is not case-sensitive and is not passed as a string to the API. See commit d440f9353a.

ThePhisch:
I reworked my commit to mirror d440f93 as requested

lib/rucio/api/rule.py
```diff
@@ -228,7 +230,7 @@ def delete_replication_rule(rule_id, purge_replicas, issuer, vo='def', session=N
 
 
 @transactional_session
-def update_replication_rule(rule_id, options, issuer, vo='def', session=None):
+def update_replication_rule(rule_id: str, options: Dict[str, Any], issuer: str, vo='def', session=None):
```
dchristidis: Why avoid type hints for `vo` and `session`?

lib/rucio/client/ruleclient.py
```diff
@@ -107,7 +108,7 @@ def get_replication_rule(self, rule_id):
             exc_cls, exc_msg = self._get_exception(headers=r.headers, status_code=r.status_code, data=r.content)
             raise exc_cls(exc_msg)
 
-    def update_replication_rule(self, rule_id, options):
+    def update_replication_rule(self, rule_id, options: Dict[str, Any]):
```
dchristidis: Ditto for `rule_id`.

lib/rucio/core/rule.py
```diff
@@ -1180,14 +1180,18 @@ def update_rule(rule_id, options, session=None):
     :raises:            RuleNotFound if no Rule can be found, InputValidationError if invalid option is used, ScratchDiskLifetimeConflict if wrong ScratchDiskLifetime is used.
     """
 
-    valid_options = ['comment', 'locked', 'lifetime', 'account', 'state', 'activity', 'source_replica_expression', 'cancel_requests', 'priority', 'child_rule_id', 'eol_at', 'meta', 'purge_replicas', 'boost_rule']
+    valid_options = ['comment', 'locked', 'lifetime', 'account', 'state',
+                     'activity', 'source_replica_expression', 'cancel_requests',
+                     'priority', 'child_rule_id', 'eol_at', 'meta',
+                     'purge_replicas', 'boost_rule']
```
dchristidis, line 1192: Isn’t this purely a cosmetic change? I find that bundling it in a commit with functional changes can be confusing.

lib/rucio/core/rule.py
```diff
@@ -1285,11 +1289,24 @@ def update_rule(rule_id, options, session=None):
 
             elif key == 'child_rule_id':
                 # Check if the child rule has the same scope/name as the parent rule
-                child_rule = session.query(models.ReplicationRule).filter_by(id=options[key]).one()
-                if rule.scope != child_rule.scope or rule.name != child_rule.name:
-                    raise InputValidationError('Parent and child rule must be set on the same dataset.')
-                if child_rule.state != RuleState.OK:
-                    rule.child_rule_id = options[key]
+                child_id: Union[str, None] = options[key]
```
dchristidis: This is what `Optional` is for.

lib/rucio/core/rule.py
```diff
@@ -1285,11 +1289,24 @@ def update_rule(rule_id, options, session=None):
...
+                if child_id is None:
+                    # dissolve exisiting relationship
+                    if not rule.child_rule_id:
+                        raise InputValidationError('Cannot detach child when '
+                                                   'no such relationship exists')
```
dchristidis: I’m starting to feel strongly against breaking string with errors message for the sake of maintaining a maximum line length. It hinders searching for them with `grep`. I’m not convinced that it is worthwhile.

lib/rucio/web/rest/flaskapi/v1/rules.py
```diff
@@ -117,6 +118,9 @@ def put(self, rule_id):
                       child_rule_id:
                         description: The child rule. Parent and child rule must be on the same dataset.
                         type: string
+                      unset_child:
+                        description: Whether to unset the child rule.
+                        type: boolean
```
dchristidis: This appears to be a leftover from your previous work.

lib/rucio/web/rest/flaskapi/v1/rules.py
```diff
@@ -150,14 +154,16 @@ def put(self, rule_id):
             description: No rule found for the given id
        """
         parameters = json_parameters()
-        options = param_get(parameters, 'options')
+        options: Dict[str, Any] = param_get(parameters, 'options', default={})
+        print(options)
```
dchristidis: Ditto.

dchristidis (changes requested):

lib/rucio/api/rule.py
ThePhisch: in the case of vo, mypy can infer the type from the default value given. i consider it to be redundant information to add a `str` tye-hint.
in the case of `session`, Radu is working on this already https://github.com/rcarpa/rucio/commit/94bcc0940451545628e0e53f5feb324d7911c665

lib/rucio/client/ruleclient.py
ThePhisch: will fix this.

lib/rucio/core/rule.py
ThePhisch: it is a purely cosmetic change, i can roll it back if you feel strongly about that.

lib/rucio/web/rest/flaskapi/v1/rules.py
ThePhisch: true, will fix this

lib/rucio/web/rest/flaskapi/v1/rules.py
ThePhisch: i added type hints to the options dict and a default value, i dont think that that has to go. on the other hand, i will remove the `print` statement.

lib/rucio/api/rule.py
dchristidis: For type hints in parameters, I find that ‘explicit is better than implicit’, as instructed by _The Zen of Python_. But I won’t push you.

lib/rucio/core/rule.py
dchristidis: No, I do not object, but I would advise having a separate commit for this. Same for the addition of type hints (especially in unrelated code).

lib/rucio/api/rule.py
ThePhisch: you are right. consider it done

lib/rucio/core/rule.py
```diff
@@ -1187,7 +1187,8 @@ def update_rule(rule_id, options, session=None):
             raise InputValidationError('%s is not a valid option to set.' % key)
 
     try:
-        rule = session.query(models.ReplicationRule).filter_by(id=rule_id).one()
+        rule: models.ReplicationRule = session.query(models.ReplicationRule).\
+            filter_by(id=rule_id).one()
```
dchristidis: Based on what we discussed, this shouldn’t be part of this commit.

lib/rucio/core/rule.py
```diff
@@ -1285,14 +1286,24 @@ def update_rule(rule_id, options, session=None):
...
-                if rule.scope != child_rule.scope or rule.name != child_rule.name:
-                    raise InputValidationError('Parent and child rule must be set on the same dataset.')
-
-                if rule.id == options[key]:
-                    raise InputValidationError('Self-referencing parent/child-relationship.')
-                if child_rule.state != RuleState.OK:
-                    rule.child_rule_id = options[key]
+                child_id: Optional[str] = options[key]
+                if child_id is None:
+                    # dissolve exisiting relationship
```
dchristidis: I find this comment redundant.

lib/rucio/core/rule.py
```diff
@@ -1285,14 +1286,24 @@ def update_rule(rule_id, options, session=None):
...
+                    if not rule.child_rule_id:
+                        raise InputValidationError('Cannot detach child when no such relationship exists')
+                    # dissolve relationship
+                    rule.child_rule_id = None  # type: ignore
+                    # remove expiration date
+                    rule.expires_at = None  # type: ignore
+                else:
+                    child_rule = (session.query(models.ReplicationRule).
+                                  filter_by(id=child_id).one())
```
dchristidis: This would look better like so:
```python
                    child_rule = (session.query(models.ReplicationRule)
                                         .filter_by(id=child_id)
                                         .one())
```

dchristidis (changes requested):
The commit message of e90284f6a7bcd531118a99d1a2d4f7b8115d7ecc is no longer valid and should be rewritten. Please provide some context as to *why* you’re doing what you’re doing.

When adding type hints, please try to address all parameters and the return type. It’s good that you’re doing it for `rule_id`, but I think it’s better to do the entire function (especially since some appear to be trivial to complete.

lib/rucio/core/rule.py
dchristidis: Bonus points if you use the SQLAlchemy 2.0 syntax instead! Refer to #5247 if you’re willing, but note that it would probably have to go into its own commit.

lib/rucio/web/rest/flaskapi/v1/rules.py
```diff
@@ -150,14 +150,15 @@ def put(self, rule_id):
             description: No rule found for the given id
        """
         parameters = json_parameters()
-        options = param_get(parameters, 'options')
+        options = param_get(parameters, 'options', default={})
```
dchristidis: Doesn’t this require some special mention?

lib/rucio/api/rule.py
```diff
@@ -246,11 +248,11 @@ def update_replication_rule(rule_id, options, issuer, vo='def', session=None):
...
             raise AccessDenied('Account %s can not approve/deny this replication rule.' % (issuer))
 
-        issuer = InternalAccount(issuer, vo=vo)
+        issuer_ia = InternalAccount(issuer, vo=vo)
         if options['approve']:
-            rule.approve_rule(rule_id=rule_id, approver=issuer, session=session)
+            rule.approve_rule(rule_id=rule_id, approver=issuer_ia, session=session)
         else:
-            rule.deny_rule(rule_id=rule_id, approver=issuer, reason=options.get('comment', None), session=session)
+            rule.deny_rule(rule_id=rule_id, approver=issuer_ia, reason=options.get('comment', None), session=session)
```
dchristidis, line 258: Why is this better?

lib/rucio/core/rule.py
```diff
@@ -1190,7 +1191,13 @@ def update_rule(rule_id: str, options: Dict[str, Any], session=None) -> None:
             raise InputValidationError('%s is not a valid option to set.' % key)
 
     try:
-        rule = session.query(models.ReplicationRule).filter_by(id=rule_id).one()
+        query = select(
+            models.ReplicationRule
+        ).where(
+            models.ReplicationRule.id == rule_id
+        )
+        rule, = session.execute(query).one()
```
dchristidis, line 1199: Use `scalar_one()` instead.

lib/rucio/core/rule.py
```diff
@@ -1208,62 +1215,162 @@ def update_rule(rule_id: str, options: Dict[str, Any], session=None) -> None:
...
-                for lock in session.query(models.ReplicaLock).filter_by(rule_id=rule.id, state=LockState.REPLICATING).all():
-                    transfers_to_cancel = request_core.cancel_request_did(scope=lock.scope, name=lock.name, dest_rse_id=lock.rse_id, session=session)
+                query = select(
+                    models.ReplicaLock
+                ).where(
+                    models.ReplicaLock.rule_id == rule.id,
+                    models.ReplicaLock.state == LockState.REPLICATING
+                )
+                replicalocks = session.execute(query).all()
+                for lock, in replicalocks:
```
dchristidis: Use `.scalars().all()` (and maybe merge into one line?)

lib/rucio/core/rule.py
```diff
@@ -1208,62 +1215,162 @@ def update_rule(rule_id: str, options: Dict[str, Any], session=None) -> None:
...
-                locks = session.query(models.ReplicaLock).filter_by(rule_id=rule.id).all()
+                query = select(
+                    models.ReplicaLock
+                ).where(
+                    models.ReplicaLock.rule_id == rule.id
+                )
+                locks = session.execute(query).all()
                 counter_rses = {}
-                for lock in locks:
+                for lock, in locks:
```
dchristidis: Ditto.

lib/rucio/core/rule.py
```diff
@@ -1208,62 +1215,162 @@ def update_rule(rule_id: str, options: Dict[str, Any], session=None) -> None:
...
+                        query = select(
+                            models.ReplicaLock
+                        ).where(
+                            models.ReplicaLock.scope == lock.scope,
+                            models.ReplicaLock.name == lock.name,
+                            models.ReplicaLock.rse_id == lock.rse_id,
+                            models.ReplicaLock.state == LockState.REPLICATING
+                        )
+                        locks2 = session.execute(query).all()
+                        for lock2, in locks2:
```
dchristidis: Ditto.

lib/rucio/core/rule.py
```diff
@@ -1208,62 +1215,162 @@ def update_rule(rule_id: str, options: Dict[str, Any], session=None) -> None:
...
-                        replica = session.query(models.RSEFileAssociation).filter(
+                        query = select(
+                            models.RSEFileAssociation
+                        ).where(
                             models.RSEFileAssociation.scope == lock.scope,
                             models.RSEFileAssociation.name == lock.name,
-                            models.RSEFileAssociation.rse_id == lock.rse_id).one()
+                            models.RSEFileAssociation.rse_id == lock.rse_id
+                        )
+                        replica, = session.execute(query).one()
```
dchristidis: Ditto.

lib/rucio/core/rule.py
dchristidis: Use `scalar_one()` instead.

lib/rucio/core/rule.py
```diff
@@ -1208,62 +1215,162 @@ def update_rule(rule_id: str, options: Dict[str, Any], session=None) -> None:
...
+                for locktype in (models.ReplicaLock, models.DatasetLock):
+                    query = select(
+                        locktype
+                    ).where(
+                        locktype.rule_id == rule.id
+                    )
+
+                    locks = session.execute(query).all()
+                    for lock, in locks:
+                        lock.update({'account': options['account']})
```
dchristidis: Not sure I like this.

lib/rucio/core/rule.py
```diff
@@ -1208,62 +1215,162 @@ def update_rule(rule_id: str, options: Dict[str, Any], session=None) -> None:
...
                     rule_ids_to_stuck = set()
-                    for lock in session.query(models.ReplicaLock).filter_by(rule_id=rule.id, state=LockState.REPLICATING).all():
+                    query = select(
+                        models.ReplicaLock
+                    ).where(
+                        models.ReplicaLock.rule_id == rule.id,
+                        models.ReplicaLock.state == LockState.REPLICATING
+                    )
+                    locks = session.execute(query).all()
+                    for lock, in locks:
```
dchristidis: Ditto.

dchristidis (changes requested):
Your last commit combines functional and aesthetic changes, which makes it more difficult to review than it should be.

> The commit message of https://github.com/rucio/rucio/commit/e90284f6a7bcd531118a99d1a2d4f7b8115d7ecc is no longer valid and should be rewritten. Please provide some context as to why you’re doing what you’re doing.

This remains to be addressed, but for commit 4caa9dbe79746ac541ad6f4b4dbd3327834a94b1.

lib/rucio/api/rule.py
ThePhisch: because issuer is an argument of type string. Overwriting it with an InternalAccount causes linting errors.

lib/rucio/web/rest/flaskapi/v1/rules.py
ThePhisch: i decided to take it out. I want the error for a missing parameter to be raised as early as possible, not later.

lib/rucio/tests/test_rule.py
```diff
@@ -952,6 +952,39 @@ def test_update_rule_child_rule(self, mock_scope, did_factory, jdoe_account):
...
+        assert get_rule(parent_id)['expires_at']
+
+        # detach this child
+        update_rule(parent_id, {'child_rule_id': None})
+        assert not get_rule(parent_id)['child_rule_id']
+        assert not get_rule(parent_id)['expires_at']
+
+        # test detaching when there is no more attached child
+        with pytest.raises(InputValidationError):
+            update_rule(parent_id, {'unset_child': True})
```
dchristidis: No bueno!

lib/rucio/web/rest/flaskapi/v1/rules.py
dchristidis: Did you actually take it out?

lib/rucio/client/ruleclient.py
```diff
@@ -192,7 +218,9 @@ def deny_replication_rule(self, rule_id):
         exc_cls, exc_msg = self._get_exception(headers=r.headers, status_code=r.status_code, data=r.content)
         raise exc_cls(exc_msg)
 
-    def list_replication_rule_full_history(self, scope, name):
+    def list_replication_rule_full_history(
+            self, scope: Union[str, bytes], name: Union[str, bytes]
```
dchristidis, line 222: Why `bytes`?

lib/rucio/core/rule.py
```diff
@@ -1231,15 +1251,26 @@ def update_rule(rule_id: str, options: Dict[str, Any], *, session: "Session") ->
...
-                session.query(models.DatasetLock).filter_by(rule_id=rule.id).update({'account': options['account']})
+                for locktype in (models.ReplicaLock, models.DatasetLock):
+                    query = update(
+                        locktype
+                    ).where(
+                        locktype.rule_id == rule.id
+                    ).values(
+                        account=options['account']
+                    )
+
```
dchristidis: Is this query executed?

lib/rucio/core/rule.py
```diff
@@ -1251,23 +1282,61 @@ def update_rule(rule_id: str, options: Dict[str, Any], *, session: "Session") ->
...
+                        for lock2 in session.execute(query).scalars().all():
                             lock2.state = LockState.STUCK
                             rule_ids_to_stuck.add(lock2.rule_id)
-                        transfers_to_cancel = request_core.cancel_request_did(scope=lock.scope, name=lock.name, dest_rse_id=lock.rse_id, session=session)
+                        transfers_to_cancel = request_core.cancel_request_did(
+                            scope=lock.scope,
+                            name=lock.name,
+                            dest_rse_id=lock.rse_id,
+                            session=session
+                        )
```
dchristidis, line 1344: Isn’t this purely decorative?

lib/rucio/core/rule.py
```diff
@@ -1251,23 +1282,61 @@ def update_rule(rule_id: str, options: Dict[str, Any], *, session: "Session") ->
...
+                        replication_rule.update({'state': RuleState.STUCK})
+
+                        # update dataset lock state
+                        query = select(
+                            models.DatasetLock
+                        ).where(
+                            models.DatasetLock.rule_id == rid
+                        )
+                        dataset_lock = session.execute(query).scalar_one()
+                        dataset_lock.update({'state': LockState.STUCK})
```
dchristidis: Why SELECT then UPDATE?

lib/rucio/core/rule.py
```diff
@@ -1276,8 +1345,22 @@ def update_rule(rule_id: str, options: Dict[str, Any], *, session: "Session") ->
...
+                        replica_lock = session.execute(query).scalar_one()
+                        replica_lock.update({'state': LockState.STUCK})
+
+                        query = select(
+                            models.DatasetLock
+                        ).where(
+                            models.DatasetLock.rule_id == rule.id
+                        )
+                        dataset_lock = session.execute(query).scalar_one()
+                        dataset_lock.update({'state': LockState.STUCK})
```
dchristidis: Ditto.

dchristidis (changes requested):
Almost there!

lib/rucio/client/ruleclient.py
ThePhisch: scope and name are used as arguments for the `quote_plus` function, which takes `str` or `bytes`. This was the only upper bound to the types applicable to scope and name I could find.

lib/rucio/core/rule.py
ThePhisch: no

ThePhisch: it is now

lib/rucio/core/rule.py
ThePhisch: to make sure that exactly one is updated. but this assumption is false, since multiple ReplicationRules/DatasetLocks can be filtered and updated

lib/rucio/core/rule.py
ThePhisch: yup

dchristidis (dismissed):
No further comments. Thanks for bearing with me.

Unfortunately, a rebase on the latest master is necessary.

dchristidis (dismissed):

dchristidis (approved):

PR #5944, opened 2022-10-26
Core: Prevent Self-Referencing Parent-Child-Relations; Fix #5931

### LINKED ISSUE #5931: Do not allow the child rule to be the same as the parent
Motivation
----------

An action such as `$ rucio update-rule --child-rule-id 3d34489b97bd4cdb9e2a0b0509da1bfa 3d34489b97bd4cdb9e2a0b0509da1bfa` is currently possible. Especially due to #5930, it means that one can create unintended situations that might require direct database manipulation in order to fix.

Modification
------------

Add some protection against this.

ThePhisch:
raises an InputValidationError when child id and parent id match.
much of the test that i have written for this commit can be merged with the test written in `test_bin_rucio.py` of PR #5943

### DIFF
--- a/lib/rucio/core/rule.py
+++ b/lib/rucio/core/rule.py
@@ -1288,6 +1288,9 @@ def update_rule(rule_id, options, session=None):
                 child_rule = session.query(models.ReplicationRule).filter_by(id=options[key]).one()
                 if rule.scope != child_rule.scope or rule.name != child_rule.name:
                     raise InputValidationError('Parent and child rule must be set on the same dataset.')
+
+                if rule.id == options[key]:
+                    raise InputValidationError('Self-referencing parent/child-relationship.')
                 if child_rule.state != RuleState.OK:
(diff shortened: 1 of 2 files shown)


dchristidis (commented):
The implementation solves the original issue, which is self-referencing parent-child relationships. However, the wording suggests that it will prevent all circular relationships, which I don’t think it can (e.g. with two rules `A` ↔ `B`).

ThePhisch:
True. While self-referencing is circular, it is only a subset of circular relationships. Perhaps this should be a new issue, albeit more difficult to tackle.
_edit_ I changed the title to reflect this

dchristidis:
I’m afraid that I still believe that the commit message and the included error message can be misleading. Editing the PR title does not change that.

ThePhisch:
changed the commit message and the error message

dchristidis (approved):

PR #5961, opened 2022-11-02
RSE: Set transfer_parameters to create_parent True; Fix #2973

### LINKED ISSUE #2973: Use GFAL’s create_parent transfer parameter
Motivation
----------

The implementation here can be simplified:
https://github.com/rucio/rucio/blob/master/lib/rucio/rse/protocols/gfal.py#L398

*This should not be done until GFAL 2.16.4 reaches EPEL.*

Modification
------------

Do not make a separate call to GFAL to create the destination path. Use the `create_parent` transfer parameter instead.

ThePhisch:


ThePhisch:
this PR is not up to standards (yet)

PR #5988, opened 2022-11-22
Deletion: do not silently ignore errors in Globus. Closes #5987

### LINKED ISSUE #5987: globus RSE protocol silently ignores errors
This will potentially result in dark data. 

To avoid this, just throw an exception which will be picked by reaper and skip removing the replicas

rcarpa:
Otherwise it can result in dark data on any error.

To avoid this, just throw an exception which will be picked by reaper and skip removing the replicas

### DIFF
--- a/lib/rucio/rse/protocols/globus.py
+++ b/lib/rucio/rse/protocols/globus.py
@@ -198,13 +198,14 @@ class GlobusRSEProtocol(RSEProtocol):
             try:
                 delete_response = send_delete_task(endpoint_id=self.globus_endpoint_id, path=path, logger=self.logger)
             except TransferAPIError as err:
-                print(err)
+                self.logger(logging.WARNING, str(err))
+                raise exception.RucioException(err)
         else:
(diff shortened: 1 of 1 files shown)


lib/rucio/rse/protocols/globus.py
```diff
@@ -198,13 +198,14 @@ def delete(self, path):
...
+                self.logger(logging.WARNING, str(err))
+                raise exception.RucioException(err)
         else:
-            print('No rse attribute found for globus endpoint id.')
+            raise exception.RucioException('No rse attribute found for globus endpoint id.')
 
         if delete_response['code'] != 'Accepted':
-            print('delete_task not accepted by Globus')
-            print('delete_response: %s' % delete_response)
+            self.logger(logging.DEBUG, 'delete_response: %s' % delete_response)
```
dchristidis, line 207: May I suggest deferring the log message interpolation (i.e. `s/ % /, /`)? Also below.

bari12 (approved):

PR #5991, opened 2022-11-23
WebUI: Rectify broken r2d2 list_rules hlink. Fix #5970

### LINKED ISSUE #5970: r2d2 list_rules link is broken, gives 404
Motivation
----------
In the r2d2 request rule page (and maybe elsewhere too?), a suggestion is given to the user by presenting a hlink which should list his rules, as in: `https://rucio-ui.cern.ch/list_rules?account=myaccount`. 
See [request_rule.js#L989](https://github.com/rucio/rucio/blob/master/lib/rucio/web/ui/static/request_rule.js#L989
) 

Unfortunately , turns out that the link is broken (returns err. 404)

Modification
------------
I'm not sure on whether/how this was changed at some point, but I guess the intended hlink would be (and should change to):
`https://rucio-ui.cern.ch/r2d2?account=myaccount`

faluchet:


maany (approved):
LGTM, no other links in webui seem to be broken with the same error ;)

PR #5999, opened 2022-11-23
Remove 'network_mode: "service:rucio"' from docker-compose-*

### LINKED ISSUE #5983: remove podman-specific settings in docker-compose*.yml
### Description

adding new images to the docker-compose files in /etc/docker/dev is very tedious, since the current setup makes heavy use of largely undocumented and likely unnecessary.

### Motivation

i am trying and failing to add a webdav server to the docker-composes. using a pure docker-compose approach, i have already managed to create a functioning docker-compose file. In addition, refactoring the current docker-compose yamls to a format that is more commonly used has the advantage of being easier to maintain and faster to understand for new developers.

This change would not necessarily break compatibility with podman. The issue lies with podman-compose.

### Change

refactor the etc/docker/dev/docker-compose*.yml files to use settings more natural to docker-compose (and which are well documented). Specifically, getting rid of `network_mode` and `external_hosts` would simplify the

### LINKED ISSUE #5996: `network_mode` in docker-compose-*.ymls'
### Description

The current networking strategy in `docker-compose-*.yml` files is the following:
1. Create a link from `xrd`, `ssh`, `postgres` etc. containers to `rucio-dev` container by specifying `network_link: "service:rucio"` in the `xrd`, `ssh`, `postgres`, etc service specification: 
https://github.com/rucio/rucio/blob/ad28dbd8eb2d99cfd4cd37b983064d21386e9e23/etc/docker/dev/docker-compose-storage.yml#L105
2. Create a backlink from `rucio-dev` container to the `xrd`, `ssh`, `postgres`, etc. by specifying the `extra_hosts` section in the `rucio-dev` service specification
https://github.com/rucio/rucio/blob/ad28dbd8eb2d99cfd4cd37b983064d21386e9e23/etc/docker/dev/docker-compose-storage.yml#L4-L11
3. Expose services like  `xrd`, `ssh`, `postgres`, `activemq` to the host machine by specifying port mappings from the `rucio-dev` service.
https://github.com/rucio/rucio/blob/ad28dbd8eb2d99cfd4cd37b983064d21386e9e23/etc/docker/dev/docker-compose-storage.yml#L16-L22

Essentially we have a star network topology with the `rucio-dev` container acting as the network hub
image

** The dashed lines represent the backlinks from  rucio-dev container to other service containers created via `extra_host` 
section.
** The solid lines represent links from rucio_dev container to other service containers defined via `network_modde: "service:rucio"`
** The dotted lines represent port forwarding from All the services.

I have the following 3 concerns with this design:

1. It is not clear from the `docker-compose-*.yml` what ports are exposed by what service since it appears that all ports are forwarded from the rucio-dev container. For example, figuring out which container is actually responsible for running the service port "61613" is ambiguous 

2. **Conflicting ports**: It is currently not possible to have two independent containers to run services on the same port.  For example, the `webdav` container that @ThePhisch is working on exposes the httpd on port 80, which is already taken up by the rucio server running in the rucio-dev container. Ideally, it should be possible to refer to the rucio-server in the docker network as `rucio-dev:80` and the webdav RSE as `webdav:80`.

3. It is not possible to ping containers directly. For example, the `xrd1` container cannot ping `xrd2` container directly. 
```
✘⚡ root@rucio-devmaany  ~/rucio# docker exec -it dev_xrd1_1 bash
[root@ec0b85bf99ff /]# ping dev_xrd2_1
ping: dev_xrd2_1: Name or service not known
[root@ec0b85bf99ff /]# ping dev_rucio_1
PING dev_rucio_1 (172.21.0.2) 56(84) bytes of data.
64 bytes from ec0b85bf99ff (172.21.0.2): icmp_seq=1 ttl=64 time=0.040 ms
64 bytes from ec0b85bf99ff (172.21.0.2): icmp_seq=2 ttl=64 time=0.054 ms
```
I speculate that this behavior could potentially play a role in TPC testing, forcing the files to go via the rucio-server as that is the only way to route traffic between containers?

### Motivation

_No response_

### Change

Switch to a mesh-network topology and expose ports to the host from the containers running that particular service.

image

This would address all 3 points of concern mentioned in the issue description

maany:
p.s. Ports are forwarded from hosts only to the containers mentioned in the [Setting up Rucio Demo Environment](http://rucio.cern.ch/documentation/setting_up_demo) page in rucio documentation. The rest of the containers should resolve each-other internally via the docker/podman bridge network.

Fix #5996 
Fix #5983

### DIFF
--- a/etc/docker/dev/docker-compose-storage-alldb.yml
+++ b/etc/docker/dev/docker-compose-storage-alldb.yml
@@ -1,36 +1,8 @@
 services:
   rucio:
     image: docker.io/rucio/rucio-dev
-    extra_hosts:
-      - "ruciomy8:127.0.0.1"
-      - "oracle:127.0.0.1"
-      - "ruciodb:127.0.0.1"
-      - "graphite:127.0.0.1"
-      - "fts:127.0.0.1"
-      - "ftsdb:127.0.0.1"
-      - "xrd1:127.0.0.1"
-      - "xrd2:127.0.0.1"
-      - "xrd3:127.0.0.1"
-      - "xrd4:127.0.0.1"
-      - "minio:127.0.0.1"
-      - "activemq:127.0.0.1"
-      - "ssh1:127.0.0.1"
     ports:
-      - "127.0.0.1:8443:443"
-      - "127.0.0.1:3308:3308"
-      - "127.0.0.1:1521:1521"
-      - "127.0.0.1:5432:5432"
-      - "127.0.0.1:8080:80"
-      - "127.0.0.1:8446:8446"
-      - "127.0.0.1:8449:8449"
-      - "127.0.0.1:3306:3306"
-      - "127.0.0.1:1094:1094"
-      - "127.0.0.1:1095:1095"
-      - "127.0.0.1:1096:1096"
-      - "127.0.0.1:1097:1097"
-      - "127.0.0.1:9000:9000"
-      - "127.0.0.1:61613:61613"
-      - "127.0.0.1:2222:22"
+      - "8443:443"
     volumes:
       - ../../../tools:/opt/rucio/tools:Z
       - ../../../bin:/opt/rucio/bin:Z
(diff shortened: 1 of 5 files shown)


maany:
Duplicate of #5998

PR #6000, opened 2022-11-24
Core & Internals: Cleanup deprecated workarounds; #4045

### LINKED ISSUE #4045: Unicode support along all Core/API/REST API
Motivation
----------
When working on porting the REST framework tests to Flask #4033 I encountered problems when handing Python 2 `unicode` type text. While we want to remove Python 2 compatibility in 1.24, these layers of compatibility can be removed. Left to research is how these functions work with Python 3 unicode native `str` type strings. I have already fixed an issue in #3995 but I think there is more to find and fix.

Modification
------------
Research API functions and called functions for unicode support and remove Python 2 compatibility code (for the server only).

bziemons:
Fixes #4045

### DIFF
--- a/lib/rucio/__init__.py
+++ b/lib/rucio/__init__.py
@@ -14,9 +14,5 @@
 # limitations under the License.
 
 import gettext
-import sys
 
-if sys.version_info < (3,):
-    gettext.install('rucio', unicode=1)  # pylint: disable=unexpected-keyword-arg
-else:
-    gettext.install('rucio')
+gettext.install('rucio')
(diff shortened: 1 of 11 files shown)


bziemons:
I haven't included a fix for the following code, because I do not know whether CMS is using this code with older Python versions.

https://github.com/rucio/rucio/blob/2a41a64899f41b356660fff80a7530765b5d59e0/lib/rucio/core/permission/cms.py#L25-L30

bari12 (approved):
Took some time, but gtg :-)

Issue #5984, opened 2022-11-21
Update Github Actions: `::set-output` is set to deprecate in May 2023


ThePhisch:
### Description

At the moment, we make use of `::set-output...`-notation to redirect output within the Github Actions (see https://github.com/rucio/rucio/blob/8288e2106f33b39930d16691905a6a73d576cf1d/.github/workflows/integration_tests.yml#L18 for an example). This already raises a deprecation warning, and support will be deactivated in May 2023. It seems prudent to change to the recommended [method](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/).

### Motivation

The Github Actions might begin failing in May 2023 if we don't take action eventually.

maany:
Nice find @ThePhisch ;)

PR #6004, opened 2022-11-28
Tests: Changed syntax of outputting to modern form in test yamls; Fix…

ThePhisch:
… #5984

Formerly used 
`echo "::set-output name={name}::{value}"`
now uses
`echo "{name}={value}" >> $GITHUB_OUTPUT`.
This is done since Github is deprecating the old syntax. Accessing the values outputted remains the same.


maany (dismissed):
LGTM, CI runs as expected. The `cherry-picking` for release patch tests fail for `autotest.yml` fails, @ThePhisch rebase maybe?

Issue #5928, opened 2022-10-18
Allow for file recovery policies based on scope


ChristophAmes:
Currently, file recovery policies can be set for individual file types. This will be expanded to also allow for policies to be set for an entire scope. How file type and scope policies shall interact with one another is still to be decided.

This change will be applied to the suspicious replica recoverer.

PR #6019, opened 2022-12-09
Recovery: Allow for file recovery policies based on scope #5928

ChristophAmes:
Currently, file recovery policies can be set for individual file types. This will be expanded to also allow for policies to be set for an entire scope. Policies based on file types take priority over policies based on scopes.

This change will be applied to the suspicious replica recoverer.

### DIFF
--- a/etc/suspicious_replica_recoverer.json
+++ b/etc/suspicious_replica_recoverer.json
@@ -1,10 +1,14 @@
 [
     {
-        "datatype": "RAW",
-        "action": "ignore"
+        "action": "declare bad",
+        "datatype": ["HITS"],
+        "scope": [],
+        "scope_wildcard": []
     },
     {
-        "datatype": "testtypedeclarebad",
-        "action": "declare bad"
+        "action": "ignore",
+        "datatype": ["RAW"],
+        "scope": [],
+        "scope_wildcard": []
     }
 ]
--- a/lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
+++ b/lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
@@ -26,6 +26,7 @@ import time
 from datetime import datetime, timedelta
 from sys import argv
 from typing import List, Any, Optional
+import re
 
 import functools
 import json
@@ -92,8 +93,6 @@ def declare_suspicious_replicas_bad(once: bool = False, younger_than: int = 3, n
 
 
 def run_once(heartbeat_handler: Any, younger_than: int, nattempts: int, vos: Optional[List[str]], limit_suspicious_files_on_rse: int, active_mode: int, **_kwargs) -> bool:
-    # assembling the worker name identifier ('executable') including the rses from <rse_expression>
-    # in order to have the possibility to detect a start of a second instance with the same set of RSES
 
     worker_number, total_workers, logger = heartbeat_handler.live()
 
@@ -114,10 +113,6 @@ def run_once(heartbeat_handler: Any, younger_than: int, nattempts: int, vos: Opt
             vos = [v['vo'] for v in list_vos()]
         logger(logging.INFO, 'This instance will work on VO%s: %s' % ('s' if len(vos) > 1 else '', ', '.join([v for v in vos])))
 
-    # make an initial heartbeat - expected only one replica-recoverer thread on one node
-    # heartbeat mechanism is used in this daemon only for information purposes
-    # (due to expected low load, the actual DB query does not filter the result based on worker number)
-
     start = time.time()
 
     try:
@@ -131,10 +126,11 @@ def run_once(heartbeat_handler: Any, younger_than: int, nattempts: int, vos: Opt
         json_data = json.load(json_file)
     except ValueError:
         logger(logging.WARNING, "No JSON object could be decoded.")
+        return True
 
     # Checking that the json file is formatedd properly.
     for i, entry in enumerate(json_data):
-        if "datatype" not in entry or "action" not in entry:
+        if ("datatype" not in entry or "scope" not in entry or "action" not in entry):
             logger(logging.ERROR, 'Entry %s in the json file is incomplete (missing either "datatype" or "action").', i)
             break
 
@@ -197,7 +193,7 @@ def run_once(heartbeat_handler: Any, younger_than: int, nattempts: int, vos: Opt
 
             # RSEs that aren't available shouldn't have suspicious replicas showing up. Skip to next RSE.
             if not rse['availability_read'] and ((len(suspicious_replicas_avail_elsewhere) > 0) or (len(suspicious_replicas_last_copy) > 0)):
-                logger(logging.WARNING, "%s is not available (availability_read: %s), yet it has suspicious replicas. Please investigate. \n", rse_expr, rse['availability_read'])
+                logger(logging.WARNING, "%s is not available (availability: %s), yet it has suspicious replicas. Please investigate. \n", rse_expr, rse['availability_read'])
                 continue
 
             if suspicious_replicas_avail_elsewhere:
@@ -338,7 +334,9 @@ def run_once(heartbeat_handler: Any, younger_than: int, nattempts: int, vos: Opt
             # Get the rse_id by going to one of the suspicious replicas from that RSE and reading it from there
             rse_id = list(recoverable_replicas[vo][rse_key].values())[0]['rse_id']
             for replica_key in list(recoverable_replicas[vo][rse_key].keys()):
-                file_metadata = get_metadata(recoverable_replicas[vo][rse_key][replica_key]["scope"], recoverable_replicas[vo][rse_key][replica_key]["name"])
+                file_scope = recoverable_replicas[vo][rse_key][replica_key]["scope"]
+                file_name = recoverable_replicas[vo][rse_key][replica_key]["name"]
+                file_metadata = get_metadata(file_scope, file_name)
                 recoverable_replicas[vo][rse_key][replica_key]["datatype"] = str(file_metadata["datatype"])
                 if recoverable_replicas[vo][rse_key][replica_key]['available_elsewhere'] is True:
                     # Replicas with other copies on at least one other RSE can safely be labeled as bad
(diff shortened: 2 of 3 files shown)


lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
```diff
@@ -131,10 +126,11 @@ def run_once(heartbeat_handler: Any, younger_than: int, nattempts: int, vos: Opt
         json_data = json.load(json_file)
     except ValueError:
         logger(logging.WARNING, "No JSON object could be decoded.")
+        return True
 
     # Checking that the json file is formatedd properly.
     for i, entry in enumerate(json_data):
-        if "datatype" not in entry or "action" not in entry:
+        if (("datatype" or "scope") not in entry or "action" not in entry):
```
cserf: Are you sure about the test ?
Shouldn't it be : 
```python
if ("datatype" not in entry or "scope" not in entry or "action" not in entry):
```

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
```diff
@@ -196,8 +192,8 @@ def run_once(heartbeat_handler: Any, younger_than: int, nattempts: int, vos: Opt
                 logger(logging.DEBUG, '%s', i)
 
             # RSEs that aren't available shouldn't have suspicious replicas showing up. Skip to next RSE.
-            if not rse['availability_read'] and ((len(suspicious_replicas_avail_elsewhere) > 0) or (len(suspicious_replicas_last_copy) > 0)):
-                logger(logging.WARNING, "%s is not available (availability_read: %s), yet it has suspicious replicas. Please investigate. \n", rse_expr, rse['availability_read'])
+            if (rse['availability'] not in {4, 5, 6, 7}) and ((len(suspicious_replicas_avail_elsewhere) > 0) or (len(suspicious_replicas_last_copy) > 0)):
```
cserf: You should keep the code using `rse['availability_read']` that now replace the old one based on `rse['availability']` + mask

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
```diff
@@ -338,49 +334,101 @@ def run_once(heartbeat_handler: Any, younger_than: int, nattempts: int, vos: Opt
...
                         del recoverable_replicas[vo][rse_key][replica_key]
                     else:
                         # Deal with replicas based on their metadata.
                         if file_metadata["datatype"] is None:  # "None" type has no function "split()"
                             files_to_be_ignored.append(recoverable_replicas[vo][rse_key][replica_key])
+                            logger(logging.WARNING, "RSE: %s, replica name %s, surl %s: Replica does not have a data type associated with it. No action will be taken.",
+                                   rse_key, replica_key, recoverable_replicas[vo][rse_key][replica_key]['surl'])
                             continue
+
+                        file_metadata_datatype = file_metadata["datatype"].split("_")[-1]
```
cserf: In case of `DAOD_PHYS` it would select `PHYS` which is probably not what you wanted to do. Morever `_` in datatype name is very ATLAS specific. I would just keep `datatype` as it is, but rely on regular expression

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
```diff
@@ -338,49 +334,101 @@ def run_once(heartbeat_handler: Any, younger_than: int, nattempts: int, vos: Opt
...
+                        file_metadata_scope = str(file_metadata["scope"])
+                        in_file_metadata_datatype = False
+                        in_file_scope = False
+                        in_scope_wildcard = False
+                        action = ""
                         for i in json_data:
```
cserf, line 365: i is already used as a variable in line 132. In addition shouldn't it be `entry` instead of `json_data`

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
```diff
@@ -338,49 +334,101 @@ def run_once(heartbeat_handler: Any, younger_than: int, nattempts: int, vos: Opt
...
-                                            break
-                                else:
-                                    logger(logging.WARNING, "RSE: %s, replica name %s, surl %s: Match for the metadata 'datatype' (%s) of replica found in json file, but no match for 'action' (%s)",
-                                           rse_key, replica_key, recoverable_replicas[vo][rse_key][replica_key]['surl'], i["datatype"], i["action"])
-                                break
-                            else:
-                                # If no policy has been set, default to ignoring the file (no action taken).
+                            # Check if a policy has been set for the specific datatype or scope
+                            # or if there are any wildcard policies regarding the scope.
+                            # Datatype action has priority over scope action.
```
cserf, line 368: I wonder if it should be this way. Let's rediscuss later

etc/suspicious_replica_recoverer.json
```diff
@@ -1,10 +1,20 @@
...
-        "action": "declare bad"
+        "datatype_action": "ignore",
+        "datatype": ["RAW"]
+    },
+    {
+        "scope_action": "declare bad",
+        "scope": [],
+        "scope_wildcard": []
+    },
+    {
```
cserf, line 8: Isn't it possible to have combined policy for scope and datatype, e.g. : 
```json
    {
        "action": "declare bad",
        "scope": ["mc.*", "data12_8TeV"],
        "datatype": ["DAOD_PHYS.*", "DAOD_SUSY1"]
    },
```

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
```diff
@@ -338,49 +334,101 @@ def run_once(heartbeat_handler: Any, younger_than: int, nattempts: int, vos: Opt
...
+                            # Specific scope has priority over wildcard scope.
+                            if "datatype" in i:
+                                if file_metadata_datatype in i["datatype"]:
+                                    try:
+                                        action = i["datatype_action"]
```
cserf: Why not simply action ?

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
```diff
@@ -338,49 +334,101 @@ def run_once(heartbeat_handler: Any, younger_than: int, nattempts: int, vos: Opt
...
+                                        in_file_metadata_datatype = True
+                                        break
+                                    except:
+                                        logger(logging.WARNING, "RSE: %s, replica name %s, surl %s: Datatype %s found in the policy library, but an error occured while retrieving the policy assigned to it. No action will be taken.",
+                                               rse_key, replica_key, recoverable_replicas[vo][rse_key][replica_key]['surl'], file_metadata_datatype)
+                                        break
+                            if ("scope" in i) and not in_file_metadata_datatype:
```
cserf: The action should be taken only if `in_file_metadata_datatype` if `datatype` is defined in the policy and `in_file_scope` if `scope` is defined in the policy. Therefore I would remove `and not in_file_metadata_datatype`

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
```diff
@@ -338,49 +334,101 @@ def run_once(heartbeat_handler: Any, younger_than: int, nattempts: int, vos: Opt
...
+                                # print("file_metadata_scope in i[scope]:", file_metadata_scope in i["scope"])
+                                if file_metadata_scope in i["scope"]:
+                                    try:
+                                        action = i["scope_action"]
+                                        in_file_scope = True
+                                    except:
+                                        logger(logging.WARNING, "RSE: %s, replica name %s, surl %s: Scope %s found in the policy library, but an error occured while retrieving the policy assigned to it. No action will be taken.",
+                                               rse_key, replica_key, recoverable_replicas[vo][rse_key][replica_key]['surl'], file_scope)
+                                        break
+                            if ("scope_wildcard" in i) and not (in_file_metadata_datatype or in_file_scope):
```
cserf: Merge `scope_wildcard` with scope `scope

cserf (changes requested):

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
ChristophAmes: The loop in line 132 closes in line 135. The main loop for the logic starts at line 149.

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
ChristophAmes: Do you mean why bother with the "try"?

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
ChristophAmes: The "not" clause is what makes sure that a datatype action has priority over a scope action. I'll remain the variable to make it a bit clearer.

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
ChristophAmes: The idea is that if a policy is set for a scope wildcard, but a single scope affected by that wildcard should have a different policy, then the policy for that single scope will take priority over the wildcard policy for that one scope.

cserf (dismissed):
LGTM. Still missing a `datatype_wildcard` but it can go into another patch.

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
```diff
@@ -197,7 +193,7 @@ def run_once(heartbeat_handler: Any, younger_than: int, nattempts: int, vos: Opt
 
             # RSEs that aren't available shouldn't have suspicious replicas showing up. Skip to next RSE.
             if not rse['availability_read'] and ((len(suspicious_replicas_avail_elsewhere) > 0) or (len(suspicious_replicas_last_copy) > 0)):
-                logger(logging.WARNING, "%s is not available (availability_read: %s), yet it has suspicious replicas. Please investigate. \n", rse_expr, rse['availability_read'])
+                logger(logging.WARNING, "%s is not available (availability: %s), yet it has suspicious replicas. Please investigate. \n", rse_expr, rse['availability'])
```
bari12: Why did you change this back to `availability`? This will be removed this release so we would have to change it back again then.

ChristophAmes: I'm not sure what happened there - it still says `availability_read` on my end. I've fixed it.

bari12 (approved):

bari12:
I only merged this against `master` and thus the 1.31 release in march.
For the cherry-pick to the release branch, there is lots of conflicts which I won't fix manually. If this is needed during `1.30.*` still, please submit a second PR with this change against `release-1.30` branch.

PR #6023, opened 2022-12-14
Deletion: Use the proper availability #6022

### LINKED ISSUE #6022: Wrong availability being used for deletion
### Description

https://github.com/rucio/rucio/blob/0c04740b4a15b6a37dcc8430c66473aff97226e8/lib/rucio/daemons/reaper/reaper.py#L515-L518

Using the separate and well-named columns is fairly recent. The bug must have existed since the use of the single-column availability containing bit fields.

### Steps to reproduce

Have an RSE that is available for deletion but not for writing:

```
Settings:
=========
  availability: 5
  availability_delete: True
  availability_read: True
  availability_write: False
…
```

Then notice `RSE … is blocklisted for delete` messages in the logs.

### Rucio Version

1.30.0

### Additional Information

Credits @faluchet for noticing the issue.

dchristidis:


cserf (approved):
Good catch

Issue #5968, opened 2022-11-05
modification to Rucio QOS workflow


dougbenjamin:
Motivation
----------
According Joao Lopes (FTS development team) -

"When the `--copy-pin-lifetime` is set, FTS will do a Bring Online operation at the source before starting the transfer. The `---copy-pin-lifetime` value is sent to the source storage endpoint for the lifetime of the replica on disk. This is explained in the FTS documentation [1].

In FTS there is now way to specify the pintime at the destination (only at the source).

The errors you are seeing happen because FTS is trying to do a Bring Online operation in a disk only endpoint (which is not supported)."

With Rucio QOS the setting of the copy pin life causes FTS to issue a bring online on source disk site. (which obviously fails).

 In the use case that file X is not at RSE B (QOS RSE - Disk w/tapes).

Is it possible have Rucio QOS   workflow modified to transfer a file X from source RSE A to destination RSE B (QOS RSE). After successful transfer - issue copy-pin-lifetime (ie set the lifetime of the pin) at destination RSE B. 

Modification
------------

PR #6029, opened 2022-12-15
Transfers: modification to Rucio QoS workflow #5968

maatthias:
Removing copy_pin_lifetime from the transfers_to_create dictionaries.


bari12 (changes requested):
You need to remove it in L1020 as well.

maatthias:
In testing this a bit more I think we need to keep L1020.  This handles the case when there is no replica at destination.  Which as I understand we need when it's a transfer and not a stagein.  What do you think?

bari12:
I don't think it will work, because the `copy_pin_lifetime` will be enforced at the source, it's always enforced on the source. And the source might very well be a disk...

maatthias:
Yes I see your point but it also renders the tests useless (they will always fail) as they're currently written.  To rewrite to test a pass we are effectively forcing the copy_pin_lifetime pass, the request to be DONE and the rule OK and not really testing any transitions.  Which isn't testing anything :)  I will remove the unit tests, make this change and do a push.

bari12:
Hmm, I don't think I follow/agree 😄 
You are right that `test_transfer_to_mas_new_replica` and `test_failed_transfer_to_mas_new_replica` doesn't test anything which is not already tested. (Since we can't set a pin at destination)
But the other two `test_transfer_to_mas_existing_replica` and `test_failed_transfers_to_mas_existing_replica` are correct, these test the stagein from the tape.

maatthias:
Yes indeed you are right.  Those tests are valid when we remove the initial check that the request has the copy_pin_lifetime (they do not because the replica doesn't exist yet.)  The second check is the test we want.  I just pushed the changes.

maatthias:
@bari12 can this be merged?  That `git cherry pick` seems unrelated to the code change and typing requirements, yes?

maatthias:
@bari12 I ran into merge conflicts squashing the commits here.  Perhaps it was all of the fork syncs going back a few months.  Pushed a clean branch and new [pull request](https://github.com/rucio/rucio/pull/6079)  If that's OK I'll close this one.

Issue #6031, opened 2022-12-19
WebDAV Tests (for the RSE Protocol)


ThePhisch:
### Description

Some RSE protocols (such as xrootd, ssh, rclone, rsync) have testing suites that are integrated into the Github actions workflow. On the other hand, officially supported protocols such as WebDAV or GFAL2 (see issue #5962) do not work in this context. This is an oversight that should be fixed.

### Motivation

This ties into issue #5962 because when rewriting GFAL2 tests, it would make sense to move from the deprecated SRM protocol to a more modern one. WebDAV lends itself to this use, i.e. this would mean that the the GFAL2 tests are implemented via WebDAV. Any experience gathered implementing tests in WebDAV will be helpful further down the line with GFAL2.

### Change

The WebDAV testing suite should either be reworked or entirely rewritten (i.e. to use the preferred pytest framework). In addition, the testing infrastructure required for integration into the Github actions as part of [`.github/workflows/integration_tests.yml`](https://github.com/rucio/rucio/blob/c9545766c92b7c279cc1345fa656950c8f71557c/.github/workflows/integration_tests.yml) needs to be created:

- a Docker container implementing a WebDAV server
- starting the Docker container within the appropriate docker-compose
- initialisation of this container as an RSE within [`tools/docker_activate_rses.sh`](https://github.com/rucio/rucio/blob/c9545766c92b7c279cc1345fa656950c8f71557c/tools/docker_activate_rses.sh)
- integrating the test within Github actions integration tests

Issue #5962, opened 2022-11-02
port gfal2 testing suite to python3


ThePhisch:
Motivation
----------
the testing file `tests/test_rse_protocol_gfal2.py` has not been ported to Python3 yet. It is currently set to be skipped in the tests.

Modification
------------
Port the file.

PR #6032, opened 2022-12-20
GFAL2 and WebDAV testing

ThePhisch:
This PR covers issues #5962 and #6031 and implements tests for the GFAL2 and WebDAV RSE protocols. In addition to writing the test files, this means the following:

- Creating the RSE that rucio connects to
- initialising it within docker compose (in doing so, issue #5983 was solved)
- Integrating the tests into Github Actions (integration tests)

In addition, a docker image was created that runs a WebDAV server which rucio can connect to. It is this docker image that is referred to within the docker compose specifications. Of course, SSL certificates have been generated and baked into the image.

A final commit also handles #6041.


etc/docker/dev/docker-compose-storage-externalmetadata.yml
```diff
@@ -1,4 +1,5 @@
-version: "3.5"
+version: '3.4'
```
mlassnig: going down a version number?

lib/rucio/tests/test_rse_protocol_gfal2_new.py
```diff
@@ -0,0 +1,161 @@
+# -*- coding: utf-8 -*-
```
mlassnig: looks good, but maybe use a more descriptive name than `..._new` ? this suggests that the old file can go away?

tools/docker_activate_rses.sh
```diff
@@ -89,6 +92,7 @@ rucio-admin account set-limits root XRD2 -1
 rucio-admin account set-limits root XRD3 -1
 rucio-admin account set-limits root XRD4 -1
 rucio-admin account set-limits root SSH1 -1
+rucio-admin account set-limits root WEB1 -1
```
mlassnig, line 110: we should also connect the distances between them

etc/docker/dev/docker-compose-storage-externalmetadata.yml
ThePhisch: Fixed. I renamed the final commit (which concerns the `docker-compose`s) because it no longer fixes issue #5983, it only adds the `web1` container.

lib/rucio/tests/test_rse_protocol_gfal2_new.py
ThePhisch: yes, so i have been told. done.

mlassnig (commented):
@ThePhisch @bari12 this is good to go. the only thing we need to decide is if we're ok with the try image being hosted by thephish's docker account, or if it needs to be made available on the rucio docker account?

ThePhisch:
> @ThePhisch @bari12 this is good to go. the only thing we need to decide is if we're ok with the try image being hosted by thephish's docker account, or if it needs to be made available on the rucio docker account?

I thought that the image will end up hosted on the rucio docker account

bari12:
Yes, should definitely be the rucio docker account. So better add it to rucio/containers.

bari12:
Push 😄

Issue #487, opened 2018-01-18
Possibility to give reason when denying a rule


bari12:
Motivation
----------
Approvers want to give a reason when denying a replication rule.

Modification
------------
This change needs two modifications:
- [x] WebUI: A website where an approver can type in a short reason when denying a replication rule. This can use the same REST call to the server, just a comment is supplied as well.
- [x] Core: Change of the deny method in core and change of the email templates, to deliver the reason to the user.

PR #6033, opened 2022-12-20
Clients: allow reason parameter for denying; Fix #6020

### LINKED ISSUE #6020: Allow reason in ruleclient when denying a rule
### Description

Currently [ruleclient.py](https://github.com/rucio/rucio/blob/98b18092263e2fe9dddffc30798a94d7a04a075b/lib/rucio/client/ruleclient.py) does not allow to give a reason in `deny_replication_rule()`.

The basic functionality was implemented in #487, though, and is also available in the web interface.

It would be useful if the ruleclient could use this functionality.

### Motivation

Currently, when using the rule client to deny rules, the users will simply get `Reason: None`. If operators want to give a reason, they either need to write a mail manually or use the web interface. The latter is not really an option if you want to "bulk-deny" tens or hundreds of rules. Simply bulk-denying with, e.g., `Reason: No lifetime set` would be great.

### Change

From looking at #487, it should be sufficient to add a new parameter to the function below and in the options to the web request add `'comment': reason`.
https://github.com/rucio/rucio/blob/98b18092263e2fe9dddffc30798a94d7a04a075b/lib/rucio/client/ruleclient.py#L180-L193

If you think this would be a reasonable change, I would be happy to try and implement it. (Although I need to figure out a reasonable test setup/scenario.)

jolange:
When "bulk-denying" rules, one can specify a reason (e.g. `Reason: No lifetime set`. The relevant functionality was already implemented in #487, but the client function lacked a parameter so far.

I tested this "in production" simply by creating a subclass of the client.

Fixes #6020

### DIFF
--- a/lib/rucio/client/ruleclient.py
+++ b/lib/rucio/client/ruleclient.py
@@ -203,15 +203,19 @@ class RuleClient(BaseClient):
         exc_cls, exc_msg = self._get_exception(headers=r.headers, status_code=r.status_code, data=r.content)
         raise exc_cls(exc_msg)
 
-    def deny_replication_rule(self, rule_id: str):
+    def deny_replication_rule(self, rule_id: str, reason: Optional[str] = None):
         """
         :param rule_id:             Rule to be denied.
+        :param reason:              Reason for denying the rule.
         :raises:                    RuleNotFound
         """
 
         path = self.RULE_BASEURL + '/' + rule_id
         url = build_url(choice(self.list_hosts), path=path)
-        data = dumps({'options': {'approve': False}})
+        options: Dict[str, Union[bool, str]] = {'approve': False}
+        if reason:
+            options['comment'] = reason
+        data = dumps({'options': options})
         r = self._send_request(url, type_='PUT', data=data)
         if r.status_code == codes.ok:
             return True


dchristidis (approved):
Note that this should have been [a patch instead of a feature](https://rucio.cern.ch/documentation/contributing/#what-should-i-know-before-i-get-started), but otherwise looks good!

jolange:
> Note that this should have been [a patch instead of a feature](https://rucio.cern.ch/documentation/contributing/#what-should-i-know-before-i-get-started), but otherwise looks good!

Thanks, I actually read that part but was still unsure -- sorry for the wrong choice. Should I somehow correct this or just leave it this way?

dchristidis:
> Should I somehow correct this or just leave it this way?

I believe no correction is necessary. Branches starting with `feature-` have a slightly different test coverage, but it shouldn’t matter for a change like this. Martin will correct me when merging it in case I’m mistaken.

dchristidis (changes requested):
Please rebase on the latest master.

On second though, do you think it would be preferable if the default value was an empty string as opposed to `None`? I would also recommend adding a type hint.

jolange:
I just rebased on master and adapted to the new type-hinting style.

I personally prefer having `None` as the default with the `Optional` type hint. Do you have a reason to prefer an empty string? In any case, I will adapt to whatever the maintainers here like better ;-)

dchristidis:
> Do you have a reason to prefer an empty string?

No, I do not have a strong logical argument, more of a personal preference. If one looks only at the function signature without its body, then `reason: Optional[str] = None` implies (at least in my mind) that anything other than `None` will be propagated to the API. However, due to line 216, an empty string is treated the same as `None`.

Of course, a counter argument is that `reason: str = ''` implies that it will be always propagated. Hence, I’m not actually asking you to change this, but merely providing some food for thought.

dchristidis:
```
 Found 1 new problems in /home/runner/work/rucio/rucio/lib/rucio/client/ruleclient.py
  - 1 errors with message """Argument of type "str" cannot be assigned to parameter "__value" of type "bool" in function "__setitem__"
                               "str" is incompatible with "bool"""".
    Candidates: line 217
```

I think you need to add type hints to `options` in order to keep pyright happy.

lib/rucio/client/ruleclient.py
```diff
@@ -203,15 +203,19 @@ def approve_replication_rule(self, rule_id: str):
...
         """
         :param rule_id:             Rule to be denied.
+        :param reason:              Reason for denying the rule.
         :raises:                    RuleNotFound
         """
 
         path = self.RULE_BASEURL + '/' + rule_id
         url = build_url(choice(self.list_hosts), path=path)
-        data = dumps({'options': {'approve': False}})
+        options: dict[str, Union[bool, str]] = {'approve': False}
```
dchristidis: Rucio is still officially supporting Python 3.6. Consequently, you can’t use `dict` here and will have to use `typing.Dict` instead.

dchristidis (changes requested):

lib/rucio/client/ruleclient.py
jolange: Thanks! I can squash the commits to meaningful ones, once the CI is happy.

dchristidis: Yes, please do. A single commit is expected for this PR.

jolange: Alright, squashed (and rebased again), now that the CI succeeds.

dchristidis (approved):

bari12:
Thanks! 🥳

PR #6034, opened 2022-12-20
RSE: Make use of GFAL create_parent; Fix #2973

### LINKED ISSUE #2973: Use GFAL’s create_parent transfer parameter

ThePhisch:
Replace code which creates the parent directories manually (recursively) by setting the params.create_parent = True flag within __gfal2_copy


mlassnig (approved):
it looks correct to me.. but... if this breaks we have a big problem :-)

PR #6037, opened 2023-01-09
Feature 6006 support storage not supporting multirange byte requests

ThePhisch:
If an RSE has the RSE attribute `simulate_multirange: <int>`, then `&#multirange=false&nconnections=10` is appended to the PFN as described in the issue.
Added unit tests.
Also worked on decorative changes (type hints and reformatting, mainly converting different forms of building strings into f-strings)

### DIFF
--- a/lib/rucio/core/credential.py
+++ b/lib/rucio/core/credential.py
@@ -37,7 +37,7 @@ REGION = make_region_memcached(expiration_time=900)
 METRICS = MetricManager(module=__name__)
 
 
-def get_signed_url(rse_id, service, operation, url, lifetime=600):
+def get_signed_url(rse_id: str, service: str, operation: str, url: str, lifetime=600) -> str:
     """
     Get a signed URL for a particular service and operation.
 
@@ -69,7 +69,6 @@ def get_signed_url(rse_id, service, operation, url, lifetime=600):
             except:
                 raise UnsupportedOperation('Lifetime must be convertible to numeric.')
 
-    signed_url = None
     if service == 'gcs':
         if not CREDS_GCS:
             CREDS_GCS = Credentials.from_service_account_file(config_get('credentials', 'gcs',
@@ -102,11 +101,11 @@ def get_signed_url(rse_id, service, operation, url, lifetime=600):
         signature = urlencode({'': base64.b64encode(CREDS_GCS.sign_bytes(to_sign))})[1:]
 
         # assemble final signed URL
-        signed_url = 'https://%s%s?GoogleAccessId=%s&Expires=%s&Signature=%s' % (host,
-                                                                                 path,
-                                                                                 CREDS_GCS.service_account_email,
-                                                                                 lifetime,
-                                                                                 signature)
+        signed_url = (
+            f'https://{host}{path}'
+            f'?GoogleAccessId={CREDS_GCS.service_account_email}'
+            f'&Expires={lifetime}&Signature={signature}'
+        )
 
     elif service == 's3':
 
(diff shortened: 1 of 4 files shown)


lib/rucio/core/credential.py
```diff
@@ -144,11 +143,20 @@ def get_signed_url(rse_id, service, operation, url, lifetime=600):
...
-            signed_url = s3.generate_presigned_url(s3op, Params={'Bucket': bucket, 'Key': key}, ExpiresIn=lifetime)
-
-    elif service == 'swift':
+            s3 = boto3.client(
+                's3', endpoint_url=f'https://{host}:{port}',
+                aws_access_key_id=access_key, aws_secret_access_key=secret_key,
+                config=Config(
+                    signature_version=signature_version,
+                    region_name=region_name
+                )
```
mlassnig: I'm not so much in favour of all these linebreak changes... especially ones that result in empty line parentheses.

lib/rucio/core/replica.py
```diff
@@ -804,16 +804,19 @@ def _build_list_replicas_pfn(
         sign_urls: bool,
         signature_lifetime: int,
         client_location: "Dict[str, Any]",
+        *,
```
mlassnig, line 808: what's this needed for?

lib/rucio/core/replica.py
```diff
@@ -855,7 +860,12 @@ def _build_list_replicas_pfn(
                             pass  # ATLAS HACK
                         else:
                             # don't forget to mangle gfal-style davs URL into generic https URL
-                            pfn = 'root://' + root_proxy_internal + '//' + pfn.replace('davs://', 'https://')
+                            pfn = f'root://{root_proxy_internal}//{pfn.replace("davs://", "https://")}'
+
+    simulate_multirange = get_rse_attribute(rse_id, 'simulate_multirange')
```
mlassnig, line 862: there needs to be a fallback if simulate_multirange is not a number (and if it's a number it must be larger than 0)

lib/rucio/core/replica.py
```diff
@@ -804,16 +804,19 @@ def _build_list_replicas_pfn(
...
 ) -> str:
     """
     Generate the PFN for the given scope/name on the rse.
     If needed, sign the PFN url
     If relevant, add the server-side root proxy to te pfn url
     """
-    pfn = list(protocol.lfns2pfns(lfns={'scope': scope.external,
-                                        'name': name,
-                                        'path': path}).values())[0]
+    pfn: str = list(
```
mlassnig: I have a strong preference on the original style here. This new style makes it much more difficult to read.

lib/rucio/tests/temp_factories.py
```diff
@@ -68,7 +72,7 @@ def _cleanup_rses(self):
             # So running test in parallel results in some tests failing on foreign key errors.
             rse_core.del_rse(rse_id)
 
-    def _make_rse(self, scheme, protocol_impl, parameters=None, add_rse_kwargs=None):
+    def _make_rse(self, scheme, protocol_impl, parameters=None, add_rse_kwargs=None) -> "Tuple[str, str]":
```
mlassnig, line 75: Does this work as expected? Tuple in a quotation marks? Am I missing something obvious?

lib/rucio/web/ui/static/subscriptions.js
```diff
@@ -14,15 +14,10 @@
 display_data = function(data, chosen_account, date, hour, minutes) {
     var tmp = {};
     $.each(data, function(index, value) {
```
mlassnig: this should not be part of this PR

lib/rucio/tests/temp_factories.py
ThePhisch: relevant [PEP 3107](https://peps.python.org/pep-3107/)
more [info](https://www.geeksforgeeks.org/function-annotations-python/)

lib/rucio/core/replica.py
ThePhisch: the asterisk signifies the end of positional arguments and states that every argument that follows is a keyword argument. decorators such as `read_session` or `stream_session` attach to this. since `_build_list_replicas_pfn` is not decorated, the asterisk is not necessary.

I added the asterisk for consistency with functions such as [`_list_files_wo_replicas`](https://github.com/rucio/rucio/blob/6f07c57c0c588c4e6ea03b40e977877be2a59b5b/lib/rucio/core/replica.py#L680).

lib/rucio/core/replica.py
ThePhisch: done. includes changing the corresponding tests

mlassnig (dismissed):

bari12:
Please resolve conflict

mlassnig (approved):

PR #6038, opened 2023-01-10
Feature 5979 handle exception replication rule on file

ThePhisch:
Fixed issue and added unit tests: 

The issue was that core/rule.py::add_rule is designed to be applied on
DIDs with the boolean flag 'open'. When applyied to files, the call to
the did dict with key 'open' is still performed, leading to a keyerror.

By performing a check whether the key exists first (and setting the
value to None otherwise), the keyerror is avoided.

This issue did not show up in any tests so far because the code that
causes the error is only run when the --ask-approval flag is passed and
the file has rule_approvers assigned.

This was followed by typing and decorative changes in the functions edited in the process of working on this branch.

### DIFF
--- a/lib/rucio/core/rule.py
+++ b/lib/rucio/core/rule.py
@@ -67,6 +67,7 @@ from rucio.db.sqla.session import read_session, transactional_session, stream_se
 
 if TYPE_CHECKING:
     from sqlalchemy.orm import Session
+    from typing import List, Tuple
 
 
 REGION = make_region_memcached(expiration_time=900)
@@ -3224,7 +3225,7 @@ def __create_rule_approval_email(rule, *, session: "Session"):
                                          'did_type': rule.did_type,
                                          'length': '0' if did['length'] is None else str(did['length']),
                                          'bytes': '0' if did['bytes'] is None else sizefmt(did['bytes']),
-                                         'closed': not did['open'],
+                                         'open': did.get('open', 'Not Applicable'),
                                          'complete_rses': ', '.join(rses),
                                          'approvers': ','.join([r[0] for r in recipents]),
                                          'approver': recipent[1],
@@ -3248,7 +3249,7 @@ def __create_recipents_list(rse_expression, filter_=None, *, session: "Session")
     :param session:         The database session in use.
     """
 
-    recipents = []  # (eMail, account)
+    recipents: "List[Tuple]" = []  # (eMail, account)
 
     # APPROVERS-LIST
     # If there are accounts in the approvers-list of any of the RSEs only these should be used
--- a/lib/rucio/core/rule.py
+++ b/lib/rucio/core/rule.py
@@ -2247,7 +2247,7 @@ def approve_rule(rule_id, approver=None, notify_approvers=True, *, session: "Ses
                 text = template.safe_substitute({'rule_id': str(rule.id),
                                                  'approver': approver})
                 vo = rule.account.vo
-                recipients = __create_recipients_list(rse_expression=rule.rse_expression, filter_={'vo': vo}, session=session)
+                recipients = _create_recipients_list(rse_expression=rule.rse_expression, filter_={'vo': vo}, session=session)
                 for recipient in recipients:
                     add_message(event_type='email',
                                 payload={'body': text,
@@ -2306,7 +2306,7 @@ def deny_rule(rule_id, approver=None, reason=None, *, session: "Session"):
                                                    'approver': approver,
                                                    'reason': reason})
             vo = rule.account.vo
-            recipients = __create_recipients_list(rse_expression=rule.rse_expression, filter_={'vo': vo}, session=session)
+            recipients = _create_recipients_list(rse_expression=rule.rse_expression, filter_={'vo': vo}, session=session)
             for recipient in recipients:
                 add_message(event_type='email',
                             payload={'body': email_body,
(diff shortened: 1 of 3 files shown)


dchristidis (commented):
I think there’s an opportunity here to potentially correct an unfortunate decision of the past. The e-mail template has a ‘closed’ field whereas in the database and everywhere else we have ’open’ instead. I can see no good reason for this inconsistency.

Commit 9e4fb5f66879bd19c644b4aaaffd57d8f8b2c399 is lacking motivation. However, as someone with a lot of involvement in operations, I don’t foresee any disruption from changing it.

lib/rucio/core/rule.py
```diff
@@ -3212,25 +3213,29 @@ def __create_rule_approval_email(rule, *, session: "Session"):
...
+                'bytes': '0' if did['bytes'] is None else sizefmt(did['bytes']),
+                'open': did.get('open', "Not Applicable"),
+                'complete_rses': ', '.join(rses),
+                'approvers': ','.join([r[0] for r in recipents]),
+                'approver': recipent[1],
+                'target_rse': target_rse,
+                'free_space': free_space,
+                'free_space_after': free_space_after
+            }
+        )
```
dchristidis: This is the part where I suggest separating functional from aesthetic changes. This one-line change is completely overshadowed by everything around it.

ThePhisch: :sweat_smile:

dchristidis (changes requested):
Am I missing it, or is the change to `rule_approval_request.tmpl` not included in the PR?

Then there’s the fact that in the first commit you add type hints to the `recipents` variable, then in the next commit, together with the rename, you change those type hints again?

Also, I see some inconsistent use of single versus double quotes (e.g. `did.get('open', "Not Applicable"),`). I am not asking you to change it. However, until we come up with an explicit guideline, my suggestion is to stick to whatever is used around the lines you’re touching.

Finally, I would also like to suggest some changes to the message of the first commit:
- applyied → applied
- did → \`did\`
- keyerror → KeyError
- key-val → key-value
- rucio → Rucio
- ant → and

lib/rucio/core/rule.py
```diff
@@ -3171,19 +3171,28 @@ def __delete_lock_and_update_replica(lock, purge_replicas=False, nowait=False, *
...
+def __create_rule_approval_email(rule: "models.ReplicationRule", *, session: "Session"):
     """
     Create the rule notification email.
 
     :param rule:      The rule object.
     :param session:   The database session in use.
     """
 
-    with open('%s/rule_approval_request.tmpl' % config_get('common', 'mailtemplatedir'), 'r') as templatefile:
+    with open(f'{config_get("common", "mailtemplatedir")}/rule_approval_request.tmpl', 'r') as templatefile:
```
dchristidis: I think using `os.path.join` would have been a further improvement.

lib/rucio/core/rule.py
```diff
@@ -3261,47 +3274,57 @@ def __create_recipents_list(rse_expression, filter_=None, *, session: "Session")
...
             rse_attr = list_rse_attributes(rse_id=rse['id'], session=session)
             if rse_attr.get('type', '') in ('LOCALGROUPDISK', 'LOCALGROUPTAPE'):
-                accounts = session.query(models.AccountAttrAssociation.account).filter_by(key='country-%s' % rse_attr.get('country', ''),
-                                                                                          value='admin').all()
+                accounts = session.query(
+                    models.AccountAttrAssociation.account
+                ).filter_by(
+                    key=f'country-{rse_attr.get("country", "")}',
+                    value='admin'
+                ).all()
```
dchristidis: Can I tempt you into rewriting this and the other query in the SQLAlchemy 2.0 syntax? They should be trivial. Please do so in a separate commit.

dchristidis (changes requested):

lib/rucio/core/rule.py
```diff
@@ -3260,47 +3278,62 @@ def __create_recipents_list(rse_expression, filter_=None, *, session: "Session")
                 try:
                     email = get_account(account=account, session=session).email
                     if email:
-                        recipents.append((email, account))
+                        recipients.append((email, account))
                 except Exception:
                     pass
 
+    model = models.AccountAttrAssociation
```
dchristidis: I have my reservations about this. Is your motivation to reduce the line length? It comes at the cost of readability.

lib/rucio/core/rule.py
```diff
@@ -3260,47 +3278,62 @@ def __create_recipents_list(rse_expression, filter_=None, *, session: "Session")
...
             if rse_attr.get('type', '') in ('LOCALGROUPDISK', 'LOCALGROUPTAPE'):
-                accounts = session.query(models.AccountAttrAssociation.account).filter_by(key='country-%s' % rse_attr.get('country', ''),
-                                                                                          value='admin').all()
-                for account in accounts:
+
+                query = select(
+                    model.account
+                ).where(
+                    model.account.key == f'country-{rse_attr.get("country", "")}',
+                    model.account.value == 'admin'
```
dchristidis: This (and similarly below) is not correct.

ThePhisch:
As per our discussion, I will write tests that cover https://github.com/rucio/rucio/blob/8be3f9d1910ec0a08f75fef18b16267ca3d9b770/lib/rucio/core/rule.py#L3268-L3295 

In particular, these tests shall test the database queries made.

lib/rucio/core/rule.py
```diff
@@ -3297,9 +3297,9 @@ def __create_recipients_list(rse_expression: str, filter_=None, *, session: "Ses
 
                 for account in session.execute(query).scalars().all():
                     try:
-                        email = get_account(account=account[0], session=session).email
+                        email = get_account(account=account, session=session).email
                         if email:
-                            recipients.append((email, account[0]))
+                            recipients.append((email, account))
```
dchristidis: Please move these to the previous commit. Same below.

dchristidis (changes requested):

dchristidis (approved):

PR #6040, opened 2023-01-11
Client/Web: Strings passed via JSON; Fix #5608

### LINKED ISSUE #5608: create_did_sample broken for DIDs with "/"
Motivation
----------
`create_did_sample` is broken for DIDs with "/" (problem at the REST layer), so cannot be used for policies with `SCOPE_NAME_REGEXP = '/([^/]*)(?=/)(.*)'` (Belle II, CMS, Icecube, etc.)

ThePhisch:
In the didclient, information (such as scope and name of DIDs) is passed through the URL. This has led to errors in
client/didclient.py::DIDClient::create_did_sample when attempting to pass scope- and name-strings containing slashes. To circumvent this, the data is now dumped into JSON and passed to the server in this form.

### DIFF
--- a/lib/rucio/client/didclient.py
+++ b/lib/rucio/client/didclient.py
@@ -663,9 +663,16 @@ class DIDClient(BaseClient):
         :param account: The account.
         :param nbfiles: The number of files to register in the output dataset.
         """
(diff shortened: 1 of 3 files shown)


ThePhisch:
the pull request is not yet fully formed, since i forgot to edit the UI. will close PR now and reopen later

PR #6042, opened 2023-01-13
Conveyor: Add archive metadata to destination URLs; Fix #2706

### LINKED ISSUE #2706: Storage Classes and Co-location hints
Motivation
----------

To help tape systems with colocation, we will pass two metadata hints per file:

`storage_class`:`<classname>` as a single value,

and `group_hints`:`['<mostimportant>', .... , '<leastimportant>']` as an ordered list of string values.

mlassnig:


lib/rucio/core/transfer.py
```diff
@@ -189,19 +189,37 @@ def __rewrite_source_url(source_url, source_sign_url, dest_sign_url, source_sche
         return source_url
 
     @staticmethod
-    def __rewrite_dest_url(dest_url, dest_sign_url):
+    def __rewrite_dest_url(dest_url: str, rewrite_url: str, values: "Optional[Union[Dict, None]]" = None) -> str:
```
rcarpa, line 192: "Optional[Dict]" should be enough

lib/rucio/core/transfer.py
```diff
@@ -189,19 +189,37 @@ def __rewrite_source_url(source_url, source_sign_url, dest_sign_url, source_sche
...
             dest_url = re.sub('https', 'gclouds', dest_url)
-        elif dest_sign_url == 's3':
+        elif rewrite_url == 's3':
             dest_url = re.sub('davs', 's3s', dest_url)
             dest_url = re.sub('https', 's3s', dest_url)
 
+        # Archival colocation hints
+        if rewrite_url == 'archive_metadata' and values is not None:
+            # Consider already existing URL parameters
+            if '?' in dest_url:
```
rcarpa, line 208: Wouldn't it be easier to use the `add_url_query` function from common/utils to add the query parameters ?

rcarpa (changes requested):
There should be changes on the protocol side to correctly handle such urls in parse_pfn

bari12:
Closing this based on the discussion we had today in the Rucio meeting

PR #6050, opened 2023-01-19
Patch 5608 slash name issue for create did sample

ThePhisch:
`dids.py::Sample` of the Flask-API does not work when scope or name of the dataset/collection to be sampled from contains slashes `/`. These parameters are passed from client to server via the URL, and slashes are not URL-safe. In order to change this, the server backend has been changed to accept data in the form of JSON. The old server endpoint remains, albeit under a different name.

The WebUI, which is used to generate such samples, has been adapted so as to make the correct API call using Ajax. The changes to the WebUI have been manually tested.

In order to manually test the WebUI, a script was written to populate the database with entries that are necessary to perform the test.

### DIFF
--- a/lib/rucio/client/didclient.py
+++ b/lib/rucio/client/didclient.py
@@ -663,9 +663,16 @@ class DIDClient(BaseClient):
         :param account: The account.
         :param nbfiles: The number of files to register in the output dataset.
         """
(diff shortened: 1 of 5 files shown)


mlassnig (commented):
looks good to me but there's some problems with the autotests for the documentation?

cserf (approved):
LGTM. The failing test is just missing annotation and can be ignored

PR #6053, opened 2023-01-23
auth: update x509 endpoint for browser clients #6048

### LINKED ISSUE #6048: Extend /auth/x509 endpoint for browser-based clients
### Description

The existing `/auth/x509`  endpoint must be extended to work together with the new webui for converting `x509 certs -> rucio token`. 

In particular, it needs to return an HTTP 206 response with `X-RUCIO-AUTH-ACCOUNTS` header with a list of possible accounts in situations where the following conditions are all met: 
1. There was no account specified in the incoming request
2. The provided x509 identity was valid
3. The provided x509 identity was mapped to multiple accounts
4. There was no default account specified

After a valid authentication, it should return the `X-RUCIO-AUTH-ACCOUNT` header with the account that was ultimately used for the authentication process.

Additionally, the Apache config for rucio server in `rucio/containers` must be modified to explicitly force client certificate requests for the `/auth/x509` endpoint. This will display a pop-up in the browser for selecting a user certificate if a request is made from a browser-based client

### Motivation
The old web-ui completely bypassed this endpoint and implemented its own x509 authentication endpoint for which it required direct database access.

See the comment I have left below, which contains the old issue description, for details on how it was done in the old webui.

Since the new webui is a pure REST'ful client, the current apache config and rucio server endpoint need some tweaking to help avoid the workarounds done in the previous webui

### Change

The entire scope of work is:
-  Add a HTTP 206 response to the exiting `/auth/x509` endpoint, return more descriptive errors when x509 authentication fails ( this repo) 
- Modify httpd config to enable `SSL_CLIENT_VERIFY` for `/auth/x509` endpoint ( https://github.com/rucio/containers/issues/216 )
- Send request from the webui to '/auth/x509' endpoint (https://github.com/rucio/webui/issues/116)

For this issue, the new endpoint must handle MultiVO use-cases and situations where a x509 identity may be mapped to multiple accounts.

maany:
Fix #6048 ( solves the auth server aspect of the issue )

The new webui requires the x509 endpoint to be updated to support browser clients. This commit updates the endpoint to support both browser clients and CLI clients.

The /auth/x509 endpoint can return HTTP 206 Partial Content to indicate that multiple accounts were found for the given x509 identity and can work with the new WebUI to pick an account.

The /auth/x509 endpoint can return more detailed error messages to the webui (identity does not exist, identity not mapped to an account, ..).

### DIFF
--- a/lib/rucio/tests/common.py
+++ b/lib/rucio/tests/common.py
@@ -18,11 +18,11 @@ import contextlib
 import itertools
 import json
 import os
+from random import choice, choices
+import requests
+from string import ascii_uppercase, ascii_letters, digits
 import tempfile
 from typing import Optional
-from random import choice, choices
-from string import ascii_uppercase, ascii_letters, digits
-import requests
 
 import pytest
 
@@ -44,10 +44,7 @@ def is_influxdb_available() -> bool:
     """Return True if influxdb is available, else return False."""
     try:
         response = requests.get('http://localhost:8086/ping')
-        if response.status_code == 204:
-            return True
-        else:
-            return False
+        return response.status_code == 204
     except requests.exceptions.ConnectionError:
         print('InfluxDB is not running at localhost:8086')
         return False
@@ -57,10 +54,7 @@ def is_elasticsearch_available() -> bool:
     """Return True if elasticsearch is available, else return False."""
     try:
         response = requests.get('http://localhost:9200/')
-        if response.status_code == 200:
-            return True
-        else:
-            return False
+        return response.status_code == 200
     except requests.exceptions.ConnectionError:
         print('Elasticsearch is not running at localhost:9200')
         return False
--- a/lib/rucio/tests/common.py
+++ b/lib/rucio/tests/common.py
@@ -19,8 +19,9 @@ import itertools
 import json
 import os
 import tempfile
-from random import choice
-from string import ascii_uppercase
+from typing import Optional
+from random import choice, choices
+from string import ascii_uppercase, ascii_letters, digits
 import requests
 
 import pytest
@@ -39,23 +40,27 @@ skip_non_belleii = pytest.mark.skipif(not ('POLICY' in os.environ and os.environ
                                       reason="specific belleii tests")
 
 
-def is_influxdb_available():
+def is_influxdb_available() -> bool:
     """Return True if influxdb is available, else return False."""
     try:
         response = requests.get('http://localhost:8086/ping')
         if response.status_code == 204:
             return True
+        else:
+            return False
     except requests.exceptions.ConnectionError:
         print('InfluxDB is not running at localhost:8086')
         return False
 
 
-def is_elasticsearch_available():
+def is_elasticsearch_available() -> bool:
     """Return True if elasticsearch is available, else return False."""
     try:
         response = requests.get('http://localhost:9200/')
         if response.status_code == 200:
             return True
+        else:
+            return False
     except requests.exceptions.ConnectionError:
         print('Elasticsearch is not running at localhost:9200')
         return False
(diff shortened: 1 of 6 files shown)


lib/rucio/web/rest/flaskapi/v1/auth.py
```diff
@@ -943,14 +943,14 @@ def get_headers(self) -> "Optional[HeadersType]":
...
-        headers['Access-Control-Expose-Headers'] = 'X-Rucio-Auth-Token'
+        headers['Access-Control-Expose-Headers'] = 'X-Rucio-Auth-Token, X-Rucio-Auth-Account, X-Rucio-Auth-Accounts'
         return headers
 
     def options(self):
         """
         ---
-        summary: x509 Allow cross-site scripting
-        description: x509 Allow cross-site scripting. Explicit for Authentication.
+        summary: x509_web Allow cross-site scripting
```
bari12: Isn't this just staying `x509`?

lib/rucio/web/rest/flaskapi/v1/auth.py
```diff
@@ -943,14 +943,14 @@ def get_headers(self) -> "Optional[HeadersType]":
...
+        description: x509_web Allow cross-site scripting. Explicit for Authentication.
```
bari12: Same

lib/rucio/web/rest/flaskapi/v1/auth.py
```diff
@@ -984,8 +984,8 @@ def options(self):
     def get(self):
         """
         ---
-        summary: x509
-        description: Authenticate a Rucio account via x509.
+        summary: x509_web
```
bari12: Also just `x509`?

lib/rucio/web/rest/flaskapi/v1/auth.py
```diff
@@ -984,8 +984,8 @@ def options(self):
...
+        description: Authenticate a Rucio account via x509. Works only for web browsers.
```
bari12: Scratch?

bari12 (commented):
Please update the initial ticket description, this is not current anymore (Since we do not create a new endpoint)

I think this is correct, I wonder though if the workflow should be different. We do more database queries this way since you always do `verify_identity` first, and then later on a `get_auth_token_x509`. I wonder if this should be done the other way around. Try `get_auth_token_x509` first, which will fail if no account is set and no default account is defined. and then go for the block you added basically.

lib/rucio/tests/common.py
```diff
@@ -18,8 +18,9 @@
 import json
 import os
 import tempfile
-from random import choice
-from string import ascii_uppercase
+from typing import Optional
```
mlassnig: t follows after r, s in the alphabet :-p

lib/rucio/tests/common.py
```diff
@@ -38,23 +39,27 @@
                                       reason="specific belleii tests")
 
 
-def is_influxdb_available():
+def is_influxdb_available() -> bool:
     """Return True if influxdb is available, else return False."""
     try:
         response = requests.get('http://localhost:8086/ping')
         if response.status_code == 204:
```
mlassnig: all these can probably be rewritten shorter like:

`return response.status_code == 204`

maany:
@bari12 Converting to a draft PR and I will discuss this tomorrow in person during our meeting. I have rewritten big chunks of the [webui](https://github.com/maany/rucio-webui-nextjs), which enable a more secure x509 workflow for browser clients. Essentially the rucio-token is never stored or made available to the browser unencrypted and is only available in an encrypted cookie. If we go this route, which is more secure, it might require reworking this PR a bit.

bari12 (approved):

Issue #6055, opened 2023-01-25
Undertaker Handling of Expired Containers


ThePhisch:
The Undertaker is a daemon that deals with expired DIDs. The handling of expired datasets is reasonably covered by tests in [`test_undertaker.py`](https://github.com/rucio/rucio/blob/ee6e349ae3688ec1abfcda297d319bcaf1050e94/lib/rucio/tests/test_undertaker.py) while tests for file DIDs are in progress (see #6054 ). The question remains how the Undertaker shall handle expired collections. This has not been relevant so far because ATLAS does not permit the deletion of collections (only of datasets and files).

I can suggest two ways of handling expired containers but would welcome further suggestions:

1. Behaviour similar to `rmdir`, meaning that the collection may only be erased if it is completely empty. Might also include a sort of `--force` option, which would then lead to behaviour described in (2)
2. Recursively detach all collections and files contained in the expired container and tombstone all of their replicas (in the same way file replicas are deleted when their dataset expires). This also means that exceptions need to be made for collections and files protected by any other rules.

Upon reaching a decision, it will be necessary to cast the expected behaviour into test cases and then edit the codebase in order to pass the tests.

ThePhisch:
As per our conversation @bari12 the handling of expired containers will be the following:

1. Detach all direct children from self.
2. Detach self from any parents.
3. Remove rules from self. _(does this include locked rules?)_

bari12:
For locked ones, it should probably FAIL. I think that's the policy we do right now for datasets as well.

bari12:
Thank you for raising this. We have reviewed this issue as part of our regular issue review.

We see value in this request, but we do not have the capacity to work on it within the next twelve months. 

In line with our [issue management guideline](https://rucio.cern/documentation/developer/issue_management), we are labelling it `wishlist` and closing it. This does not mean it is forgotten: wishlist items are revisited regularly during component reviews and roadmap planning, and this issue can be re-opened and added back to the pipeline as soon as interest and capacity align.

PR #6063, opened 2023-02-02
Patch 5154 Undertaker Test

ThePhisch:
Extension of the Undertaker test suite and changes to the underlying rucio codebase in order to ensure that the tests pass. Includes the following testcases:

`TestUndertaker::test_undertaker` in which files belonging to expired datasets are tombstoned and the datasets deleted. The image only shows one dataset with three files, the actual test has many more datasets and larger amounts of files attached to each. The green diamond denotes a rule protecting the dataset.
test_undertaker

`test_file_protected_by_rule` in which a file (F1) belonging to an expired dataset is protected by an external second rule (R2). The protected file is detached from the expired dataset (dotted line). The unprotected file (F2) is also detached and its replicas are tombstoned. The F1 replicas protected by R2 remain untouched, whereas the F1 replicas indirectly protected by R1 via the dataset are tombstoned.
test_file_protected_by_rule

`test_file_protected_by_dataset`: similarly, F1 belongs to a nonexpired dataset D2 as well as the expired dataset D1.
test_file_protected_by_dataset

`test_file_dids_in_dataset` covers the situation in which an expired file DID (F3) is attached to a nonexpired dataset. The expected outcome is that F3 is detached and its replicas are tombstoned.
test_file_dids_in_dataset

`test_file_did_deletion` checks whether a file F that is _not_ attached to any dataset has its replicas tombstoned. The rule R is optional in the actual test (the test runs once with and once without the rule).
test_file_did_deletion

In summary, this PR ensures that the undertaker fulfills its mission of handling expired DIDs properly. In general, an expired DID has all its replicas (be they directly associated because the expired DID is a file or indirectly because the expired DID is a collection) receive the epoch-tombstone (a special datetime marker set in 1970). In addition, any attached DIDs are detached. The behaviour becomes more complicated when additional rules and attachments (to a parent) are at play. Additional rules and parents protect any non-expired DIDs which would otherwise be affected by a parent being expired.

The case for containers is currently being argued in #6055


lib/rucio/core/did.py
```diff
@@ -1650,6 +1666,20 @@ def _delete_dids_wo_temp_tables(
...
+        query = update(
+            models.RSEFileAssociation
+        ).filter(
+            models.RSEFileAssociation.scope == scope,
+            models.RSEFileAssociation.name == name
+        ).values(
+            tombstone=OBSOLETE
+        )
+        session.execute(query)
+        pass
```
mlassnig: unnecessary pass?

lib/rucio/core/did.py
```diff
@@ -1290,14 +1290,28 @@ def _delete_dids(
...
+                query = update(
+                    models.RSEFileAssociation
+                ).filter(
+                    models.RSEFileAssociation.scope == didscope,
+                    models.RSEFileAssociation.name == didname
+                ).values(
+                    tombstone=OBSOLETE
+                )
+                session.execute(query)
+                continue
```
mlassnig, line 1315: unnecessary continue? or is it just to skip the logger?

lib/rucio/tests/test_undertaker.py
```diff
@@ -15,22 +15,33 @@
 
 from datetime import datetime, timedelta
 from logging import getLogger
+from collections import namedtuple
```
mlassnig: sort alphabetically

lib/rucio/tests/test_undertaker.py
```diff
@@ -15,22 +15,33 @@
...
+from rucio.db.sqla.constants import OBSOLETE
 from rucio.tests.common import rse_name_generator, did_name_generator
+from rucio.common.exception import DataIdentifierNotFound
+
+from typing import TYPE_CHECKING
+if TYPE_CHECKING:
+    from rucio.tests.temp_factories import TemporaryRSEFactory, TemporaryDidFactory
 
 LOG = getLogger(__name__)
+Rse = namedtuple('Rse', ['name', 'id'])
```
mlassnig: `Rse`  capitalisation looks... very wrong

lib/rucio/core/did.py
ThePhisch: yup, removed it

lib/rucio/core/did.py
ThePhisch: necessary continue: this `if` block is reached when no rule could be queried for the given scopename. Some calls in the following code (i.e. `rucio.core.rule.delete_rule(rule_id=rule_id, ...`) will lead to errors when run with `rule_id = None`.

The point you made with the logger is a good one, I added logger calls.

lib/rucio/tests/test_undertaker.py
ThePhisch: true. Changed it to `RSE_namedtuple`

lib/rucio/core/rule.py
```diff
@@ -889,6 +889,32 @@ def delete_rule(rule_id, purge_replicas=None, soft=False, delete_parent=False, n
         transfers_to_delete = []  # [{'scope': , 'name':, 'rse_id':}]
         account_counter_decreases = {}  # {'rse_id': [file_size, file_size, file_size]}
 
+        if locks.first() is None:
```
bari12, line 892: Why is this entire part needed?

lib/rucio/core/did.py
```diff
@@ -1290,14 +1290,29 @@ def _delete_dids(
             models.ReplicationRule.rse_expression,
             models.ReplicationRule.locks_ok_cnt,
             models.ReplicationRule.locks_replicating_cnt,
-            models.ReplicationRule.locks_stuck_cnt
-        ).join_from(
+            models.ReplicationRule.locks_stuck_cnt,
+            temp_table.scope,
+            temp_table.name
+        ).outerjoin_from(
```
bari12, line 1296: Why is this an outer join now?

bari12 (commented):
I think the tests are good, but I don't fully follow some of the did/rule changes. @cserf please have a look on this as well.

PR #6065, opened 2023-02-02
Replicas: handle listing mixed nonarchive/archive replicas. Closes #5746

### LINKED ISSUE #5746: error-prone xroot prioritization when listing mixed archive/non-archive replicas
Motivation
----------
_list_replicas() has some logic to prioritize xroot for archives. This logic relies on appending the 'root' scheme to the list of accepted schemes when building the protocol for an rse. This operation, of building the protocol, is however only done once per RSE and then the protocol list is re-used by the same list-replicas call.

If, in the same list-replica call, we fetch archive and non-archive replicas, the previous behavior can result in the 'root' scheme being considered for non-archive replicas while it shouldn't. Respectively, if a non-archive replica is the first to be analyzed on an RSE, we will not prioritize xroot for the archives

rcarpa:
Keep a separate list of protocols in cache if it's an archive or non-archive replica.

### DIFF
--- a/lib/rucio/core/replica.py
+++ b/lib/rucio/core/replica.py
@@ -12,7 +12,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
+import copy
 import heapq
 import logging
 import math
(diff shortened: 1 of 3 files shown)


lib/rucio/core/replica.py
```diff
@@ -919,32 +922,30 @@ def _list_replicas(replicas, show_pfns, schemes, files_wo_replica, client_locati
                 continue
 
             # It's the first time we see this RSE, initialize the protocols needed for PFN generation
-            if rse_id not in protocols_by_rse_id:
+            protocols = protocols_cache.get(rse_id, {}).get(is_archive)
```
mlassnig, line 925: I don't understand this particular piece. It calls .get() for is_archive which is a bool. what am I missing?

rcarpa: it comes together with the change a couple of lines below : 
```
protocols_cache[rse_id][is_archive] = protocols
```

mlassnig (approved):
nice

PR #6076, opened 2023-02-08
Tests/Rule `move_rule` with invalid Source Replica Expression; Fix #4845

### LINKED ISSUE #4845: Misleading error message when moving a rule with a no-longer-valid source replica expression
Motivation
----------

```
$ rucio move-rule E3D2B68CD15949769DE889A37474534C SWT2_CPB_DATADISK
2021-09-15 20:48:24,395 ERROR   Provided RSE expression is considered invalid.
Details: RSE Expression resulted in an empty set.
$ rucio update-rule --source-replica-expression None E3D2B68CD15949769DE889A37474534C
Updated Rule
$ rucio move-rule E3D2B68CD15949769DE889A37474534C SWT2_CPB_DATADISK
2a1035284c19482ba850719bb8a9fec7

```

Looking at the rule:
```
$ rucio rule-info E3D2B68CD15949769DE889A37474534C | grep ^Source
Source replica expression:  UKI-SCOTGRID-GLASGOW_DATADISK
```

This RSE has been decommissioned. In that respect, the error message is appropriate, but one assumes that it’s the destination RSE expression that is at fault.

Modification
------------

Improve the error message or, better yet, consider making the appropriate correction automatically.

ThePhisch:
Currently, a misleading error message is raised to the user upon attempting to move a rule which has an invalid Source Replica Expression. An invalid Source Replica Expression is an RSE expression which evaluates to an empty set, which can occur due to a variety of reasons.

The misleading error message informs the user that the provided `RSE expression is considered invalid`, this is misleading, since the error does not stem from the RSE expression provided to the `move_rule` command but from the Source Replica Expression attribute of the rule prior to moving/updating it.

The solution is to introduce a new exception
`InvalidSourceReplicaExpression` and raise it within `core/rule.py::add_rule`, should the attempt to parse the Source Replica Expression with `parse_expression` fail.

This commit also includes a test added to `test_rule.py::TestCore` which simulates two ways in which the Source Replica Expression might become invalid:
1) By deleting the only RSE the Source Replica Expression points to.
2) By matching the RSE to the Source Replica Expression via an RSE attribute but then deleting
   this RSE attribute from the RSE.
In both cases the Source Replica Expression will evaluate to an empty set.

As per the conversation with @dchristidis, no further action (i.e. further than raising a more specific error message) is taken. The two options would be:
1. Server-Side correction (replacing the Source Replica Expression by `None`) as is currently the common workaround.
2. Client-Side correction (upon receiving the specific error message, first run `update-rule --source-replica-expression None ...` and then continue).

In the end, both options would lead to rucio client/server taking action that does not fully correspond to user input.

### DIFF
--- a/lib/rucio/tests/test_rule.py
+++ b/lib/rucio/tests/test_rule.py
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from collections import namedtuple
 import json
 import random
 import string
@@ -26,7 +27,7 @@ from rucio.client.ruleclient import RuleClient
 from rucio.common.config import config_get_bool
 from rucio.common.exception import (RucioException, RuleNotFound, AccessDenied, InsufficientAccountLimit, DuplicateRule, RSEWriteBlocked,
                                     RSEOverQuota, RuleReplaceFailed, ManualRuleApprovalBlocked, InputValidationError,
-                                    UnsupportedOperation, InvalidValueForKey)
+                                    UnsupportedOperation, InvalidValueForKey, InvalidSourceReplicaExpression)
 from rucio.common.policy import get_policy
 from rucio.common.schema import get_schema_value
 from rucio.common.types import InternalAccount, InternalScope
@@ -37,7 +38,7 @@ from rucio.core.did import add_did, attach_dids, set_status
 from rucio.core.lock import get_replica_locks, get_dataset_locks, successful_transfer
 from rucio.core.replica import add_replica, get_replica
 from rucio.core.request import get_request_by_did
-from rucio.core.rse import add_rse_attribute, add_rse, update_rse, get_rse_id, del_rse_attribute, set_rse_limits
+from rucio.core.rse import add_rse_attribute, add_rse, update_rse, get_rse_id, del_rse, del_rse_attribute, set_rse_limits
 from rucio.core.rse_counter import get_counter as get_rse_counter
 from rucio.core.rule import add_rule, get_rule, delete_rule, add_rules, update_rule, reduce_rule, move_rule, list_rules
 from rucio.core.scope import add_scope
@@ -53,8 +54,10 @@ from rucio.tests.common_server import get_vo
 from typing import TYPE_CHECKING
 if TYPE_CHECKING:
     from typing import Tuple
+    from rucio.tests.temp_factories import TemporaryDidFactory, TemporaryRSEFactory
 
 LOG = getLogger(__name__)
+RSE_namedtuple = namedtuple('RSE_namedtuple', ['name', 'id'])
 
 
 def create_files(nrfiles, scope, rse_id, bytes_=1):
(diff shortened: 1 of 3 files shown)


dchristidis:
I have my doubts that introducing the term SRE is prudent.

ThePhisch:
Updated the PR to make use of an appropriate pytest fixture for caching. Also changed mentions of *SRE* to the typed out *Source Replica Expression*.

lib/rucio/tests/test_rule.py
```diff
@@ -153,6 +156,74 @@ def setup_class(request, vo, root_account, jdoe_account, rse_factory):
...
+        delete_rse: bool,
+        did_factory: "TemporaryDidFactory",
+        rse_factory: "TemporaryRSEFactory",
+        root_account: "InternalAccount"
+    ):
+        """
+        REPLICATION RULE (CORE): Move a rule using an invalid source replica expression.
+
+        Parametrised so that there are two different ways for the source replica
+        expression (SRE) to become invalid:
```
dchristidis: > Also changed mentions of SRE to the typed out Source Replica Expression.

Did you?

dchristidis (changes requested):
> The misleading error message informs the user that the provided `RSE
expression is considered invalid`, this is misleading, since the error
does not stem from the RSE expression provided to the `move_rule`
command but from the Source Replica Expression attribute of the rule
prior to moving/updating it.

This needs some rewording. Note that the invalid source replica expression doesn’t necessarily come from the original rule (e.g. `rucio move-rule --source-replica-expression INVALID XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX CERN-PROD_DATADISK`).

lib/rucio/tests/test_rule.py
ThePhisch: I did so in the commit message, but forgot to do so in the source code itself.

ThePhisch: Changed reference to SREs within the docstring, but did not change the variable name `SRE` in the test code.

ThePhisch:
> > The misleading error message informs the user that the provided `RSE expression is considered invalid`, this is misleading, since the error
> > does not stem from the RSE expression provided to the `move_rule`
> > command but from the Source Replica Expression attribute of the rule
> > prior to moving/updating it.
> 
> This needs some rewording. Note that the invalid source replica expression doesn’t necessarily come from the original rule (e.g. `rucio move-rule --source-replica-expression INVALID XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX CERN-PROD_DATADISK`).

done

lib/rucio/tests/test_rule.py
```diff
@@ -153,6 +156,75 @@ def setup_class(request, vo, root_account, jdoe_account, rse_factory):
...
+        """
+
+        from dogpile.cache.api import NO_VALUE
+        from hashlib import sha256
+
+        # create RSEs A, B, C (structured into namedtuples)
+        rseA, rseB, rseC = (
+            RSE_namedtuple(*rse_factory.make_mock_rse()) for _ in range(3)
+        )
+        rsekey, rseval = tag_generator(), tag_generator()
```
dchristidis, line 192: Why combine two assignments into a single line when there’s no unpacking?

lib/rucio/tests/test_rule.py
```diff
@@ -153,6 +156,75 @@ def setup_class(request, vo, root_account, jdoe_account, rse_factory):
...
+            "copies": 1,
+            "rse_expression": rseB.name,
+            "source_replica_expression": source_replica_expression
+        }
+        ruledict = add_rules([dataset], [rule])
+        rule_id = ruledict[(dataset["scope"], dataset["name"])][0]
+
+        # delete RSE-A (or otherwise unmatching from source replica expr)
+        if delete_rse:
+            from rucio.core.rse import del_rse
```
dchristidis: Why is this import hidden here?

dchristidis (changes requested):

lib/rucio/tests/test_rule.py
ThePhisch: I feel like it is more legible this way. It also shifts the focus away from the setup of the test (which is written in a much more concise manner than when using the components that actually are under scrutiny).

lib/rucio/tests/test_rule.py
ThePhisch: I use a pytest fixture to parametrise the test. Only one of the two testcases produced this way requires this function to be imported, so I "hid" it there.

lib/rucio/tests/test_rule.py
dchristidis: Though I don’t quite see it the same way, your motivation is logical and I have little reason to insist.

lib/rucio/tests/test_rule.py
dchristidis: I’m afraid I don’t find argument convincing. :smile:

ThePhisch: As per our conversation, I moved the import statement. Seeing as this type of import shows up in other parts of the codebase (and in particular among the test suites), one might open up a low-importance issue to move the import statements to the head of the file (provided there are no other reasons to keep them where they currently are).

ThePhisch: I opened #6090

dchristidis (approved):

Issue #5893, opened 2022-09-26
Support `ignore_availability` in preparer


rcarpa:
Motivation
----------
There are two possible behaviors: 
- either requests are ignored when fetched from the database: https://github.com/rucio/rucio/blob/9a5d044d9f5760a224dba9feb14b71bfdf213252/lib/rucio/core/request.py#L420 
- or `ignore_availability` is set and they are marked NO_SOURCES: https://github.com/rucio/rucio/blob/9a5d044d9f5760a224dba9feb14b71bfdf213252/lib/rucio/core/transfer.py#L863 

Submitter has the ability to switch between these two modes using a CLI switch. Preparer is hard-coded into `ignore_availability = False`

Modification
------------
Implement this switch from the submitter:

https://github.com/rucio/rucio/blob/ac95ef3fabb8e1fcabddf1075cc19ae06db44a34/bin/rucio-conveyor-submitter#L108

in the preparer: 
https://github.com/rucio/rucio/blob/2a41a64899f41b356660fff80a7530765b5d59e0/bin/rucio-conveyor-preparer#L35

And pass it through the call stack to this function:
https://github.com/rucio/rucio/blob/e3c77683184063424fdcf9ecfda7d37cb4641c4f/lib/rucio/daemons/conveyor/preparer.py#L128

Add a test for the two behaviors. A example test (which will have to be simplified and adapted for this issue):
https://github.com/rucio/rucio/blob/2d81b39b8560f376f412fcba62a622a6db748780/lib/rucio/tests/test_conveyor.py#L611

koolgax99:
Hii @rcarpa  I am Nihar, we just discussed on Mattermost. Can you assign this issue to me?

PR #6085, opened 2023-02-16
Added `ignore_availability` in preparer

koolgax99:
Added `ignore_availability` feature in `preparer`

Solving the Issue #5893


rcarpa:
The commits you are reverting must not be reverted. Please drop those commits. Ideally: also rebase/squash all your commits into one

PR #6086, opened 2023-02-16
Testing: add a rucioclient container to docker files. documentation#177

rcarpa:
Add a container which is identical to the rucio container, but does nothing at all. It's just a placeholder.

This commit is a pre-requirement for usage of the intellij docker-compose remote execution. When configuring intellij remote interpreter, the user has to define the container in which the commands will be executed. However, when you actually execute the commands via the intellij UI, that container is destroyed and re-created while overriding the entry-point with intellij wrappers. As a result, we cannot use the "rucio" container for that, because the httpd server will be killed.

While I don't like adding such IDE-specific hacks upstream, I don't see a better solution if we want to create an official document about how to use intellij remote execution. I maintained a local override patch since forever for that, but we cannot require other developers to do the same thing.

### DIFF
--- a/etc/docker/dev/docker-compose-storage-alldb.yml
+++ b/etc/docker/dev/docker-compose-storage-alldb.yml
@@ -1,5 +1,16 @@
 version: "3"
 services:
+  rucioclient:
+    image: docker.io/rucio/rucio-dev
+    command: ["sleep", "infinity"]
+    volumes:
+      - ../../../tools:/opt/rucio/tools:Z
+      - ../../../bin:/opt/rucio/bin:Z
+      - ../../../lib:/opt/rucio/lib:Z
+    environment:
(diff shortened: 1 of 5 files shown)


maany:
Could you add this container as an extended configuration that lives in a separate docker-compose file as discussed here: https://docs.docker.com/compose/extends/ 
This would mean that if we have to start the dev environment with this new container, we would run something like
```
docker-compose -f docker-compose-storage.yml -f docker-compose-intellij-patch.yaml
```

rcarpa:
While intellij automates most of the things for you, it doesn't have the flexibility to support such behavior.

maany (approved):
Looks good. We would have to follow up with a vscode env + pycharm env + rucio-dev container in the next steps.

Issue #6075, opened 2023-02-08
Add parameters to daemons’ stop() function


dchristidis:
### Description

Per [the documentation](https://docs.python.org/3/library/signal.html?highlight=signal%20signal#signal.signal), the signal handler is expected to have two parameters. Their absence may cause a somewhat-benign unhandled exception.

```
Traceback (most recent call last):
  File "/usr/local/bin/rucio-conveyor-preparer", line 26, in main
  File "/usr/lib64/python3.6/threading.py", line 1072, in _wait_for_tstate_lock
    thr.join(timeout=3.14)
  File "/usr/local/bin/rucio-conveyor-preparer", line 37, in <module>
    run(once=args.run_once, threads=args.threads, sleep_time=args.sleep_time, bulk=args.bulk)
  File "/usr/lib64/python3.6/threading.py", line 1060, in join
    elif lock.acquire(block, timeout):
TypeError: stop() takes 0 positional arguments but 2 were given
    self._wait_for_tstate_lock(timeout=max(timeout, 0))
  File "/usr/local/lib/python3.6/site-packages/rucio/daemons/conveyor/preparer.py", line 70, in run
    main(args=parser.parse_args())
```

### Steps to reproduce

The above-provided traceback was noticed on the ATLAS instance. It is unknown whether it is always reproducible when shutting down daemons.

shalearkane:
Hey, if this issue is not assigned to anyone, can I work on this?

shalearkane:
I failed to reproduce the issue. I followed the instructions [here](https://github.com/rucio/rucio/tree/master/etc/docker/dev) and I ran the `run_daemons` command in the `dev-rucio-1` container. I tried to terminate the command using `^C` multiple times however I was unable to reproduce it. Could it be because in the ATLAS instance, it is using a different kill signal?

dchristidis:
`^C` sends a `SIGINT`, not a `SIGTERM`. Here’s a minimal example demonstrating the issue.

```console
$ cat foo.py
#!/usr/bin/env python3

import signal
import time

def stop():
    print('stop')

if __name__ == '__main__':
    signal.signal(signal.SIGTERM, stop)
    try:
        while True:
            time.sleep(60)
    except KeyboardInterrupt:
        stop()
$ python3 foo.py 
^Cstop
$ python3 foo.py &
[1] 168059
$ kill 168059
Traceback (most recent call last):
  File "/var/home/dchristidis/foo.py", line 13, in <module>
[...]

PR #6087, opened 2023-02-16
Add params and type hints to stop() #6075

shalearkane:
Fixes issue #6075 

### Overview
https://github.com/rucio/rucio/blob/f9d52b7c7afbc1cfa41076f1d24f85299aa06cb8/lib/rucio/daemons/conveyor/preparer.py#L42-L47 should accept parameters `signalnum` and `handler` before being used as as signal handler in https://github.com/rucio/rucio/blob/f9d52b7c7afbc1cfa41076f1d24f85299aa06cb8/bin/rucio-conveyor-preparer#L23-L28 according to the [python docs](https://docs.python.org/3/library/signal.html#signal.signal). However the same function is being called without parameters in Line 28 https://github.com/rucio/rucio/blob/f9d52b7c7afbc1cfa41076f1d24f85299aa06cb8/bin/rucio-conveyor-preparer#L28 So I added `*args` to the function parameters to resolve the issue. 

### Alternative Solution
The other possible solution is to add the parameters:
```python
def stop(signalnum, handler):
```
However, then for the function call in Line 28,
https://github.com/rucio/rucio/blob/f9d52b7c7afbc1cfa41076f1d24f85299aa06cb8/bin/rucio-conveyor-preparer#L28 I will have to pass those arguments to `stop()`.

### DIFF
--- a/lib/rucio/daemons/atropos/atropos.py
+++ b/lib/rucio/daemons/atropos/atropos.py
@@ -36,6 +36,9 @@ from rucio.daemons.common import run_daemon
 from rucio.db.sqla.constants import LifetimeExceptionsState
 
 if TYPE_CHECKING:
+    from types import FrameType
+    from typing import Optional
+
     from rucio.daemons.common import HeartbeatHandler
 
 GRACEFUL_STOP = threading.Event()
@@ -233,7 +236,7 @@ def run(
         thread_list = [t.join(timeout=3.14) for t in thread_list if t and t.is_alive()]
 
 
-def stop(signum=None, frame=None):
+def stop(signum: "Optional[int]" = None, frame: "Optional[FrameType]" = None) -> None:
     """
     Graceful exit.
     """
--- a/lib/rucio/daemons/c3po/c3po.py
+++ b/lib/rucio/daemons/c3po/c3po.py
@@ -21,8 +21,10 @@ import logging
 from datetime import datetime
 from hashlib import md5
 from json import dumps
+from queue import Queue
 from threading import Event, Thread
 from time import sleep
+from typing import TYPE_CHECKING
 from uuid import uuid4
 
 from requests import post
@@ -39,7 +41,9 @@ from rucio.daemons.c3po.collectors.free_space import FreeSpaceCollector
 from rucio.daemons.c3po.collectors.jedi_did import JediDIDCollector
 from rucio.daemons.c3po.collectors.workload import WorkloadCollector
 
-from queue import Queue
+if TYPE_CHECKING:
+    from types import FrameType
+    from typing import Optional
 
 GRACEFUL_STOP = Event()
 
@@ -249,7 +253,7 @@ def place_replica(once=False,
         logging.critical(e)
 
 
-def stop(signum=None, frame=None):
+def stop(signum: "Optional[int]" = None, frame: "Optional[FrameType]" = None) -> None:
     """
     Graceful exit.
     """
--- a/lib/rucio/daemons/hermes/hermes.py
+++ b/lib/rucio/daemons/hermes/hermes.py
@@ -18,8 +18,8 @@
 """
 
 import calendar
-import functools
 import datetime
+import functools
 import json
 import logging
 import random
@@ -29,9 +29,9 @@ import socket
 import sys
 import threading
 import time
-from typing import TYPE_CHECKING
 from configparser import NoOptionError, NoSectionError
 from email.mime.text import MIMEText
+from typing import TYPE_CHECKING
 
 import requests
 import stomp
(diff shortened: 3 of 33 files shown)


dchristidis (changes requested):
I think your choosing to use variadic arguments is reasonable.

This change should be applied to all daemons and not just the Conveyor Preparer. In addition, I think it’s a good opportunity to add some type hints, given that they are trivial.

shalearkane:
On looking into other stop functions, I saw that they had parameters `signum` and `frame` which defaulted to `None`. So I changed the variadic argument to the same and added type hints.

lib/rucio/daemons/hermes/hermes.py
```diff
@@ -337,7 +338,7 @@ def run_once_messages(heartbeat_handler: HeartbeatHandler,
     return must_sleep
 
 
-def stop(signum: int = None, frame: str = None) -> None:
```
shalearkane: Here the type of `frame` is set to `str`. But according to this [discussion](https://github.com/python/typing/discussions/1042), it should be `FrameType`. Also the syntax of the type-hint is wrong.

dchristidis: Which is why I found your choice of `*args` a nice compromise. The parameters aren’t used in the body of the function, so there is little to gain by being overly pedantic.

shalearkane:
I have changed the arguments of the stop functions of other daemons to `*args`. I added type-hint `None` to the return type. Should I add type hints to other functions as well?

dchristidis:
I was surprised to see that so many `stop()` functions have some parameters already. I’m now not as confident about the best approach and would welcome someone else to comment. They way I see it, there are two alternatives:

 1. `def stop(*args: Any) -> None:` It’s simple and, at least in my eyes, somewhat suggests that the arguments are ignored.
 2. `def stop(signum: Optional[int] = None, frame: Optional[FrameType] = None) -> None:` More complex, but communicates well that this function is a signal handler.

bari12:
I prefer the second;

dchristidis (changes requested):
Be sure to import `Optional` and `FrameType`. Frequently we do it like so:

```python

if TYPE_CHECKING:
    from types import FrameType
    from typing import Optional
```

And then place the type hints inside strings.

Please note that the commits should be squashed together. Finally, the commit message has some room for improvement.

shalearkane:
I have imported the Types and I have fixed imports in several files and squashed the commit. Please review the commit message. Should I make it more descriptive?

dchristidis (changes requested):
> And then place the type hints inside strings.

You missed this part. Type hints are evaluated and will lead to `NameError` exceptions if the necessary classes are not imported at runtime.

As for the commit message:
 1. It should follow the format `<component>: <change_message> #<issue number>`. You did well, but since you are touching multiple components, you should omit the first part (as opposed to prefixing the message with ‘bug’).
 2. The rest of the message can be misleading. ‘add default arguments’ suggests that the parameters already exist, which is not the case. As such, ‘add parameters’ is more appropriate.
 3. The body of the message is empty. You should use it to document *why* you are making this change. Explain how this is the proper signature for signal handlers and how deviating from it can lead to a `TypeError`.

shalearkane:
I have updated my patch against the latest master branch and amended my commit message accordingly. Please review my changes :smile:

dchristidis (changes requested):
Please review the failed tests. It would appear that there are some missing imports.

For example:
`lib/rucio/daemons/conveyor/preparer.py:42:49: F821 undefined name 'FrameType'`

Similarly, in some places you moved the import statement under a conditional block (i.e. `if TYPE_CHECKING`) for things that are already in use.

Finally, you’ll need to rebase your branch on the latest master.

lib/rucio/daemons/atropos/atropos.py
```diff
@@ -37,6 +37,8 @@
 
 if TYPE_CHECKING:
     from rucio.daemons.common import HeartbeatHandler
+    from types import FrameType
+    from typing import Optional
```
dchristidis: Please sort as you’ve done elsewhere (e.g. in `automatix.py`).

lib/rucio/daemons/c3po/c3po.py
```diff
@@ -24,6 +24,7 @@
 from threading import Event, Thread
 from time import sleep
 from uuid import uuid4
+from typing import TYPE_CHECKING
```
dchristidis: `typing` should come before `uuid`.

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
```diff
@@ -27,6 +27,7 @@
 from sys import argv
 from typing import List, Any, Optional
 import re
+from typing import TYPE_CHECKING
```
dchristidis: There’s already a `from typing …` above.

lib/rucio/daemons/hermes/hermes.py
```diff
@@ -50,7 +50,9 @@
 from rucio.daemons.common import run_daemon
 
 if TYPE_CHECKING:
-    from typing import Callable, List, Dict
+    from typing import Callable, List, Dict, Optional
+    from types import FrameType
```
dchristidis: `types` should come before `typing`.

dchristidis (changes requested):
This is likely to be my last request for changes. Please adjust some of the imports as noted in the comments below.

In addition, I would like to ask you to adjust the commit message. Please ensure that there is a blank line between the subject and the rest of the message, then move ‘Fixes #6075’ to the end of the subject.

shalearkane:
Sir, could you please review the changes? I have made all the requested changes and have rebased the code against master branch.

dchristidis (approved):

PR #6093, opened 2023-02-20
Transfers: read conveyor scheme config into a list. Fix #6092

### LINKED ISSUE #6092: submitter doesn't use the configured scheme list
```
[conveyor]
scheme = gsiftp,root,https,davs
```
will not be used by the submitter at all. It is passed through the call stack all the way. But ignored at the last moment: 
https://github.com/rucio/rucio/blob/da626e61657370a252b01847a89bcadfddaf5e69/lib/rucio/core/transfer.py#L883

rcarpa:


bari12:
Any update here?

PR #6097, opened 2023-02-24
If we get a full URL, don't try to build one

### LINKED ISSUE #6096: CMS LFN to PFN translation gives double hosts
### Description

CMS has a situation where the PFN becomes completely mangled. The reason for this seems to be that the CMS TFC which takes care of this in complicated cases can actually change the host name (e.g. at FNAL depending on the LFN, it can go to different file systems on different gateways). 

I have a patch for this which is that if the protocol gives back a URL instead of a path, Rucio will just accept it and not build a URL from the path.

### Steps to reproduce

See above

### Rucio Version

_No response_

### Additional Information

_No response_

ericvaandering:
If the underlying protocol path translation gives a full URL, then just pass it along.

### DIFF
--- a/lib/rucio/rse/protocols/gfal.py
+++ b/lib/rucio/rse/protocols/gfal.py
@@ -83,7 +83,10 @@ class Default(protocol.RSEProtocol):
                 path = lfn['path'] if 'path' in lfn and lfn['path'] else self._get_path(scope=scope, name=name)
                 if self.attributes['scheme'] != 'root' and path.startswith('/'):  # do not modify path if it is root
                     path = path[1:]
-                pfns['%s:%s' % (scope, name)] = ''.join([self.attributes['scheme'], '://', hostname, ':', str(self.attributes['port']), web_service_path, prefix, path])
+                if re.match(r'^\w+://', path):  # This is already a URL
+                    pfns['%s:%s' % (scope, name)] = path
+                else:
+                    pfns['%s:%s' % (scope, name)] = ''.join([self.attributes['scheme'], '://', hostname, ':', str(self.attributes['port']), web_service_path, prefix, path])
 
         return pfns
 


ericvaandering:
Please check and review. More info in #6096

bari12:
Hi Eric, 
Sorry for the long silence here. I am wondering if this will introduce any issues with cascading URLs. xroot caches for example. Probably not, but it is still a risky change.
@mlassnig @dchristidis any opinions?

ericvaandering:
For us it is solving a case of cascading URLs. :-)

bari12:
I don't think your change would break anything for others, since you wouldn't build a cascading url like this anyway (like there wouldn't be a did part in the first part of a cascading URL) - But I am still wondering if this might have any side effects I am not seeing. 

(Sorry for the close, misclicked)

dchristidis:
I’m afraid I’m not able to review this. We have at least two other locations where we do some form of URL manipulation (e.g. [replica.py](https://github.com/rucio/rucio/blob/b0f7064bc4b07f5be704d656bacdd96e7a03426b/lib/rucio/core/replica.py#L798-L877), [transfer.py](https://github.com/rucio/rucio/blob/b0f7064bc4b07f5be704d656bacdd96e7a03426b/lib/rucio/core/transfer.py#L165-L205)). I can’t say with certainty that 1) it won’t have any unintended consequences and 2) that the proposed patch is an appropriate solution.

It doesn’t help that the information provided in both the issue and the pull request is, at least for someone like me who is unfamiliar with this part of the code, lacking.

ericvaandering:
I'll explain better when I get some time.

ericvaandering:
Sorry it took so long for me to get to/remember this.

@dchristidis @bari12 let me explain more fully.

CMS uses the TFC (trivial file catalog) to translate LFNs to PFNs. LFNs are the name part of the DID in Rucio for us.

At this point, most sites just attach a prefix to the LFN to make the PFN and all is well since there is also a single hostname. Here we just use the standard Rucio code for the translation.

A few sites have been able to get clever over the years and one site (Fermilab let's say) put's not only different prefixes on different paths but different hosts too (hosting them on completely different filesystem technologies). For these, we put our old TFC regex's in the extended attributes of the plugin.

For instance here is FNAL's in a slightly more readable format than we actually use in the Rucio settings:
```
            "rules": [
               {  "lfn": "/+store/temp/user/(.*)",
                  "pfn": "root://cmseos.fnal.gov//eos/uscms/store/temp/user/$1"
               },
               {  "lfn": "/+store/(.*)",
                  "pfn": "root://cmsdcadisk.fnal.gov//dcache/uscmsdisk/store/$1"
               }

``` 
and `hostname` for the RSE is set to cmsdcadisk.FNAL.gov

Now, what happens is that if the LFN is /store/temp/user/Eric/myfile.txt first the URI root://cmseos.fnal.gov//eos/uscms/store/temp/user/Eric/myfile.txt is generated. But then, the combination of the LFN to PFN and Rucio tried to get fancy and decides that because the result doesn't have ` cmsdcadisk.FNAL.gov` in it, that that too has to be added. So you end up with something like 
`root://cmsdcadisk.fnal.gov/root:/cmseos.fnal.gov/eos/uscms/store/temp/user/Eric/myfile.txt' as the URI for the file.

So instead, this patch says "if I get a fully formed URI from the layer below, I'm just going to trust it and not try to rearrange it"

ericvaandering:
@dchristidis @bari12 thoughts?

dchristidis:
You’ll have to excuse me, I do not feel qualified to comment on this any further.

bari12:
Thanks for the explanation Eric. Two things:
- This TFC manipulation of URLs business gets us into a strange situation. We are really pushing boundaries here where concepts potentially clash, or are already clashing. For this different hostname workflows, protocols should be used (I understand that this doesn't really match to the TFC usecase). The further we go with this, the more we will run into walls of concepts clashing. And especially we might make other communities run into walls due to these changes.
- I also understand that the TFC decision was made way before Rucio, so it's not something we can not just "ignore". I will merge this, from my understanding this should be transparent to everyone else, though I am not 100% sure about it. I would just implore you (and CMS) that we do not push these "possibilities" with TFC further and further. 

Will merge this for `32` only though, due the potential implications.

bari12 (approved):

ericvaandering:
For what it's worth, I don't think this is a new thing from CMS, just a part of the namespace that Rucio now needs access too. 

I guess the fundamental conflict is stated this way: An RSE may have multiple hostnames supplying its services.

PR #6098, opened 2023-02-24
trace: added option to submit trace to elastic. Fix #5756

### LINKED ISSUE #5756: tracer.Kronos daemon to send trace to Elasticsearch.
Motivation
----------
To get aggregated info/analysis on trace, it would be great that kronos send the trace to elasticsearch.
This will make things easy as we don't have to rely on anyother external implementation to do this..

Modification
------------
Can be something like Hermes2 does: https://github.com/rucio/rucio/blob/master/lib/rucio/daemons/hermes/hermes2.py#L355

panta-123:
Currently, trace is sent to activeMQ and cosumed by Kronos daemon to updata metadata of dids. Since we can sent extra information in trace , we want to see aggregated information of trace and infer something out of it.
To do this, we want to store trace into ElasticSearch, which this PR does.
Sending trace to elastic is optional via rucio config as:

```
[trace]
...
...
elastic_url = http://elasticsearch:9200
elastic_index = trace
to_elastic = True   
```
If to_elastic is False no trace is sent to elastic.

Closes: #5756

### DIFF
--- a/etc/docker/test/extra/rucio_multi_vo_ts2_postgres14.cfg
+++ b/etc/docker/test/extra/rucio_multi_vo_ts2_postgres14.cfg
@@ -95,7 +95,7 @@ username = username
 password = password
 topic = /topic/rucio.tracer
 to_elastic = True
-elastic_url = http://elasticsearch:9200
+elastic_url = http://elastic:9200
 elastic_index = trace
 
 [tracer-kronos]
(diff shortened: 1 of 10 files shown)


cserf:
The logic looks good. Would be good to add a test (have a look at https://github.com/rucio/rucio/blob/master/lib/rucio/tests/test_hermes.py#L146 for instance)
Some flake8 errors. Please correct.

bari12:
What do we do with this? @panta-123 @cserf 
I think the functionality is useful, but by now the PR is massively out of sync? Should we close and re-submit?

PR #6111, opened 2023-03-02
Bug when the transmogrifier uses wildcard copies : Closes #6105

### LINKED ISSUE #6105: Bug when the transmogrifier use wildcard copies
### Description

When `*` is used for `copies`, the number of copies is revolved to match the number of RSEs returned by parsing the RSE expression, then the transmogrifier loop over all the RSEs and create the rules. If some of these RSE is not available, the transmogrifier fails and won't create the rules for the remaining RSE.

### Steps to reproduce

Create a subscription with `'copies': '*'` and a certain RSE expression. Set one of the RSE from this RSE expression to unavailable.

### Rucio Version

_No response_

### Additional Information

_No response_

cserf:
Bug when the transmogrifier uses wildcard copies : Closes #6105

### DIFF
--- a/lib/rucio/daemons/transmogrifier/transmogrifier.py
+++ b/lib/rucio/daemons/transmogrifier/transmogrifier.py
@@ -124,7 +124,7 @@ def __split_rule_select_rses(
     weight: int,
     rse_expression: str,
     copies: int,
-    blocklisted_rse_id: list,
+    blocklisted_rse_ids: list,
     logger: LoggerFunction,
 ) -> tuple[list, bool, bool]:
     """
@@ -138,7 +138,7 @@ def __split_rule_select_rses(
     :param weight: The weight of the rule.
     :param rse_expression: The RSE expression of the rule.
     :param copies: The number of copies.
-    :param blocklisted_rse_id: The list of blocklisted_rse_id.
+    :param blocklisted_rse_ids: The list of blocklisted rse_ids.
     :param logger: The logger.
     :return: A tuple with list selected_rses, and 2 booleans create_rule, wont_reevaluate.
     """
@@ -172,7 +172,7 @@ def __split_rule_select_rses(
                 copies=copies,
                 size=0,
                 preferred_rses=preferred_rses,
-                blocklist=blocklisted_rse_id,
+                blocklist=blocklisted_rse_ids,
             )
             wont_reevaluate = True
             break
@@ -184,26 +184,22 @@ def __split_rule_select_rses(
         ) as error:
             logger(
                 logging.WARNING,
-                'Problem getting RSEs for subscription "%s" for account %s : %s. %s'
-                % (
-                    subscription_name,
-                    account,
-                    str(error),
-                    'Try including blocklisted sites' if attempt == 0 else 'Skipping rule creation.'
-                ),
+                'Problem getting RSEs for subscription "%s" for account %s : %s. %s',
+                subscription_name,
+                account,
+                str(error),
+                'Try including blocklisted sites' if attempt == 0 else 'Skipping rule creation.'
             )
             # Now including the blocklisted sites
-            blocklisted_rse_id = []
+            blocklisted_rse_ids = []
             METRICS.counter(name="addnewrule.errortype.{exception}").labels(exception=str(error.__class__.__name__)).inc()
             wont_reevaluate = True
         except Exception as error:
             logger(
                 logging.ERROR,
-                "Problem resolving RSE expression %s : %s"
-                % (
-                    rse_expression,
-                    str(error),
-                )
+                "Problem resolving RSE expression %s : %s",
+                rse_expression,
+                str(error),
             )
     if len(preferred_rses) - len(preferred_unmatched) >= copies:
         create_rule = False
@@ -242,7 +238,7 @@ def get_subscriptions(logger: LoggerFunction = logging.log) -> list[dict]:
                     break
                 if rule.get("copies") == "*":
                     rule["copies"] = len(list_rses_from_expression)
-                    rule["wild_card"] = True
+                    rule["wildcard"] = True
                     overwrite_rules = True
             if skip_sub:
                 continue
@@ -273,7 +269,7 @@ def get_subscriptions(logger: LoggerFunction = logging.log) -> list[dict]:
             subscriptions.extend(sub_dict[priority])
         logger(logging.INFO, "%i active subscriptions", len(subscriptions))
     except SubscriptionNotFound as error:
-        logger(logging.WARNING, "No subscriptions defined: %s" % (str(error)))
+        logger(logging.WARNING, "No subscriptions defined: %s", (str(error)))
         return []
     except TypeError as error:
         logger(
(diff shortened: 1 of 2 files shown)


lib/rucio/daemons/transmogrifier/transmogrifier.py
```diff
@@ -477,7 +478,8 @@ def run_once(heartbeat_handler: "HeartbeatHandler", bulk: int, **_kwargs) -> boo
 
     worker_number, total_workers, logger = heartbeat_handler.live()
     stopwatch = Stopwatch()
-    blocklisted_rse_id = [rse["id"] for rse in list_rses({"availability_write": False})]
+    block_listed = {rse['rse']: rse["id"] for rse in list_rses({"availability_write": False})}
+    blocklisted_rse_id = list(block_listed.values())
```
dchristidis: I’m having trouble understanding this change. Could you explain your reasoning?

lib/rucio/tests/test_subscription.py
```diff
@@ -799,3 +801,46 @@ def test_run_transmogrifier_chained_subscription_exclude_site_algo(self, rse_fac
...
+                'activity': activity}
+
+        subid = rucio_client.add_subscription(name=subscription_name,
+                                              account=root_account.external,
+                                              filter_={'scope': [tmp_scope.external, ], 'pattern': '%s.*' % dsn_prefix, 'split_rule': True, 'did_type': ['DATASET', ]},
+                                              replication_rules=[rule],
+                                              lifetime=None,
+                                              retroactive=False,
+                                              dry_run=False,
+                                              comments='Ni ! Ni!',
```
dchristidis: Is this absolutely necessary? :smile:

lib/rucio/daemons/transmogrifier/transmogrifier.py
```diff
@@ -585,20 +587,23 @@ def run_once(heartbeat_handler: "HeartbeatHandler", bulk: int, **_kwargs) -> boo
...
                             )
+                            weight = selected_rses[rse].get("weight", None)
+                        logger(
+                            logging.INFO,
+                            "Will insert one rule for %s:%s on %s"
+                            % (did["scope"], did["name"], rse),
+                        )
+                        if rse in block_listed and rule_dict.get("wild_card"):
+                            logger(logging.INFO, "RSE %s is unavailable and wild_card number of copies is used. Skipping rule creation" % rse)
+                            continue
```
dchristidis: I see how this works when `copies == '*'`, but I have the impression that we should come up with a policy that works for other as well.

For example, if we define a subscription with `{"rse_expression": "CERN-PROD_DATADISK|BNL-OSG2_DATADISK", "copies": 2, "split_rule": true}`, and one of the RSEs is excluded for writing, how do we want the Transmogrifier to handle it? I believe I see a few scenarios:

 1. Either create zero on one rules and raise an exception (non-deterministic, depends on the order of the RSEs). I have the impression that this is what happens now?
 2. Create exactly one rule (like what was just implemented for `copies == '*'`).
 3. Create exactly two rules, the one on the excluded RSE with `ignore_availability=True`.

lib/rucio/tests/test_subscription.py
cserf: Ni !

lib/rucio/daemons/transmogrifier/transmogrifier.py
cserf: I agree we should decide of a policy. But depending on the type of data we might want to have different behaviour, e.g. for tests datasets, that's fine to have 2., whereas if you have some datasets that are really needed on every RSEs we may want 3.
One possibility is to define a default policy, but provide a way to bypass it. If we say 2. is the default, we can add a new parameter in the subscription (e.g. `force_replication` that will force the creation of the rule on non available RSEs).
We also should define what to do if number of copies is bigger than the total list of RSEs (i.e. including the blocklisted ones) in the RSE expression..

lib/rucio/daemons/transmogrifier/transmogrifier.py
cserf: I need to have a list of RSEs (not RSE_ID) blocklisted for line 603

dchristidis: I see, but I confess that I still find it confusing. It would have been easier to understand if named `blocklisted_rses` and `blocklisted_rse_ids` and if both were lists.

lib/rucio/daemons/transmogrifier/transmogrifier.py
dchristidis: Agreed, but I would rather have `ignore_availability` instead of `force_replication` (the more terms we can reuse, the better). I think a default (i.e. not configurable) policy as described in 2 is a reasonable compromise.

If the number of copies is greater than the number of RSEs, then I think some sort of WARNING-level message and falling back to `'*'` should suffice.

lib/rucio/daemons/transmogrifier/transmogrifier.py
dchristidis: I’m not convinced that the new variable names are superior to the ones I proposed.

lib/rucio/daemons/transmogrifier/transmogrifier.py
```diff
@@ -245,6 +245,7 @@ def get_subscriptions(logger: Callable = logging.log) -> list[dict]:
                     break
                 if rule.get("copies") == "*":
                     rule["copies"] = len(list_rses_from_expression)
+                    rule["wild_card"] = True
```
dchristidis: I believe that in computing ‘wildcard’ is a single word.

lib/rucio/daemons/transmogrifier/transmogrifier.py
```diff
@@ -529,6 +531,7 @@ def run_once(heartbeat_handler: "HeartbeatHandler", bulk: int, **_kwargs) -> boo
                     #  Get all the rule and subscription parameters
                     rule_dict = __get_rule_dict(rule_dict, subscription)
                     weight = rule_dict.get("weight", None)
+                    ignore_availability = rule_dict.get("ignore_availability", None)
```
dchristidis: Since it’s used below in `add_rule()`, the default should be `False`, not `None`. I’m surprised that our type checking didn’t catch this.

lib/rucio/daemons/transmogrifier/transmogrifier.py
```diff
@@ -589,20 +592,26 @@ def run_once(heartbeat_handler: "HeartbeatHandler", bulk: int, **_kwargs) -> boo
...
+                            source_replica_expression = selected_rses[rse].get(
+                                "source_replica_expression",
+                                None,
                             )
+                            weight = selected_rses[rse].get("weight", None)
+                        logger(
+                            logging.INFO,
+                            "Will insert one rule for %s:%s on %s"
+                            % (did["scope"], did["name"], rse),
+                        )
```
dchristidis: Please pass the variables as arguments to the `logger()` call instead of using string interpolation (see https://docs.astral.sh/ruff/rules/logging-percent-format/ for the justification).

dchristidis (changes requested):
Please also address the errors from Ruff.

bari12 (approved):

PR #6129, opened 2023-03-09
Core & Internals: activate temp tables by default. Closes #6084

### LINKED ISSUE #6084: switch ON temporary tables workflows by default
Just change the default value of the config in rucio.cfg from "False" to "True". This way, temporary tables will be used by default, but can be disabled if somebody encounters problems.

rcarpa:


rcarpa:
Builds on top of https://github.com/rucio/rucio/pull/6126 ; only the last commit is different

bari12 (approved):

PR #6131, opened 2023-03-10
Propagate the timeout parameter to gfal : Closes #6125

### LINKED ISSUE #6125: Propagate the timeout parameter to gfal
### Description

When using the downloadclient, one can specify a timeout. This timeout is used by the watchdog to stop unresponsive transfers. But if the timeout is set to a large value, the download might first hits the gfal internal timeout. It would be better to have the same timeout for both.

### Motivation

_No response_

### Change

_No response_

cserf:
Propagate the timeout parameter to gfal : Closes #6125

### DIFF
--- a/lib/rucio/rse/protocols/gfal.py
+++ b/lib/rucio/rse/protocols/gfal.py
@@ -440,11 +440,8 @@ class Default(protocol.RSEProtocol):
             ctx.set_opt_integer("HTTP PLUGIN", "OPERATION_TIMEOUT", transfer_timeout)
             ctx.set_opt_integer("SRM PLUGIN", "OPERATION_TIMEOUT", transfer_timeout)
             ctx.set_opt_integer("GRIDFTP PLUGIN", "OPERATION_TIMEOUT", transfer_timeout)
-            params = ctx.transfer_parameters()
-            params.timeout = int(transfer_timeout)
-            watchdog = Timer(params.timeout + 60, self.__gfal2_cancel)
-        else:
-            params = ctx.transfer_parameters()
+            watchdog = Timer(int(transfer_timeout) + 60, self.__gfal2_cancel)
+        params = ctx.transfer_parameters()
         if src_spacetoken:
             params.src_spacetoken = str(src_spacetoken)
         if dest_spacetoken:
@@ -458,6 +455,7 @@ class Default(protocol.RSEProtocol):
 
         try:
             if transfer_timeout:
+                params.timeout = int(transfer_timeout)
                 watchdog.start()
             ret = ctx.filecopy(params, str(src), str(dest))
             if transfer_timeout:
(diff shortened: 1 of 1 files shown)


lib/rucio/rse/protocols/gfal.py
```diff
@@ -436,14 +436,19 @@ def __gfal2_copy(self, src, dest, src_spacetoken=None, dest_spacetoken=None, tra
         :raises RucioException: if it failed to copy the file.
         """
         ctx = self.__ctx
-        params = ctx.transfer_parameters()
```
dchristidis, line 439: What is your motivation for moving this line?

cserf: Just to set the context parameters before

dchristidis: Forgive me, I still do not understand.

It seems to me that there might be some unexpected quirks or hidden side-effects in the GFAL interface that one cannot anticipate just by reading the source code alone. I think we must document them *somewhere* (e.g. in the source code, the commit message, or the issue), otherwise future developments may silently introduce a regression.

Essentially, I have two questions:

```python
            params.timeout = int(transfer_timeout)
```

Why is this not sufficient? Why is it necessary to set GFAL plug-in timeouts?

```python
            ctx.set_opt_integer("HTTP PLUGIN", "OPERATION_TIMEOUT", transfer_timeout)
            ctx.set_opt_integer("SRM PLUGIN", "OPERATION_TIMEOUT", transfer_timeout)
            ctx.set_opt_integer("GRIDFTP PLUGIN", "OPERATION_TIMEOUT", transfer_timeout)
            params = ctx.transfer_parameters()
            params.timeout = int(transfer_timeout)
```

Why is the order important? Why does one *have* to first set the plug-in timeouts, then the transfer timeout? If it is not, then I would have found it preferable to merely introduce the context configuration after line 444.

bari12: The first one I wondered about too when I first looked at this issue. My (pure speculation based) interpretation was that there is a difference in gfal between a timeout and an OPERATION_TIMEOUT, maybe the timeout gets only triggered when data is already flowing, and the OPERATION_TIMEOUT happens already during the connection-phase.

The second one, I agree, I would move this back up, simply to also get rid of this rather ugly else block :-)

If you can do it before the afternoon, that would be great for the release.

dchristidis: With the recent changes, I think the implementation is much clearer. My original concern still stands: there is insufficient information as to why this change is necessary. In addition, the should be a single commit, not two.

Purely in the interest of including this fix in today’s release, I think it’s fine to mark this conversation as resolved.

dchristidis (approved):

PR #6132, opened 2023-03-14
deprecated list-dids-extended. Closes #5448

### LINKED ISSUE #5448: Merge list_dids and list_dids_extended methods
Motivation
----------
The `list_dids` and `list_dids_extended` methods should basically do the same thing, they were just functionally separated in early days of json columns. This change should functionally be enforced through all layers of the Architecture (CLI, Clients, REST, API, Core)

Modification
------------

robbarnsley:
removed list-dids-extended and SearchExtended functions in REST, api and core layers.
added deprecation warnings in binary and client layer.

### DIFF
--- a/lib/rucio/core/did.py
+++ b/lib/rucio/core/did.py
@@ -14,7 +14,6 @@
 # limitations under the License.
 
 import logging
-import operator
 import random
 from datetime import datetime, timedelta
 from enum import Enum
@@ -37,7 +36,6 @@ from rucio.core import did_meta_plugins, config as config_core
 from rucio.core.message import add_message
 from rucio.core.monitor import MetricManager
 from rucio.core.naming_convention import validate_name
-from rucio.core.did_meta_plugins.filter_engine import FilterEngine
 from rucio.db.sqla import models, filter_thread_work
 from rucio.db.sqla.constants import DIDType, DIDReEvaluation, DIDAvailability, RuleState, BadFilesStatus
 from rucio.db.sqla.session import read_session, transactional_session, stream_session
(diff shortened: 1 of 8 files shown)


lib/rucio/core/did.py
```diff
@@ -2845,109 +2843,12 @@ def set_status(scope, name, *, session: "Session", **kwargs):
                 rucio.core.rule.generate_rule_notifications(rule=rule, session=session)
 
 
-@stream_session
+@read_session
```
bari12, line 2846: I think this should still be a `stream_session`, simply because list_dids in the meta_plugin is also a `stream_session`.

lib/rucio/tests/test_did_meta_plugins.py
```diff
@@ -455,145 +455,6 @@ def test_get_metadata(self, mock_scope, did_client, db_session):
             assert all_metadata['key2'] == value2
             assert all_metadata['project'] == "data12_14TeV"
 
-    @pytest.mark.dirty
-    @pytest.mark.noparallel(reason='fails when run in parallel')
-    def test_list_dids_extended(self, did_client, db_session):
```
bari12, line 460: Couldn't this test just be re-written to use `list_dids` instead of deleting it?

bari12 (commented):
I have two very minor comments; Otherwise this is good.

lib/rucio/core/did.py
robbarnsley: If I remember correctly, the only reason I changed this was because it gave me an error telling me to do so .

lib/rucio/tests/test_did_meta_plugins.py
robbarnsley: yes it could, my mistake - I hadn't realised these were tests for the client

lib/rucio/core/did.py
robbarnsley: `Details: stream_session decorator should be used only with generator. Use read_session instead.` 

looks like there's a mixture of returns and yields in the list_dids chain i.e. from core->did.py through core->did_meta_plugins->\_\_init\_\_.py to the corresponding plugin module.

I'll leave it as-is for now (as it doesn't work otherwise)?

bari12 (approved):

PR #6136, opened 2023-03-16
Authentication & Authorisation: Add get_auth_token_ssh check code. Fix #6119

### LINKED ISSUE #6119: SSH Auth Fails for Users Without Admin Rights
### Description

The authentication API method `get_auth_token_ssh()` [checks ](https://github.com/rucio/rucio/blob/85492437ace81e6957a713d3d8895309703ded36/lib/rucio/api/authentication.py#L219) for a permission, `get_auth_token_ssh`, which is not present in the `perm` dictionary [defined](https://github.com/rucio/rucio/blob/release-1.29-LTS/lib/rucio/core/permission/generic.py#L25) in `core/permission/generic.py`. The result of this is that the default permission check function is [called](https://github.com/rucio/rucio/blob/85492437ace81e6957a713d3d8895309703ded36/lib/rucio/core/permission/generic.py#L123), and only users with admin rights are granted access. 

### Steps to reproduce

- Create an account
- Add an SSH pubkey to that account
- Add both `get_auth_token_ssh: 1` and `get_ssh_challenge_token: 1` attributes to account
- Try to `rucio whoami` using SSH authentication
    - Fail
- Add admin role
    - Success

### Rucio Version

Client/Server : 1.29.10/1.29.10

### Additional Information

_No response_

bjwhite-fnal:


dchristidis (commented):
Though I have no reservation against this proposed fix, I can’t help but wonder: why do we need two permissions for the same flow? Is there any scenario where the Rucio administrators might want to allow one but deny the other?

Looking at commit 6871028 where this issue was introduced, there is no argument against the proposed fix being what the author omitted, but what I’m questioning is the original design.

bari12:
I kind of agree, I would just use `perm_get_auth_token_ssh` for both the challenge and auth token.

lib/rucio/core/permission/generic.py
```diff
@@ -1037,9 +1038,9 @@ def perm_update_lifetime_exceptions(issuer, kwargs, *, session: "Optional[Sessio
     return _is_root(issuer) or has_account_attribute(account=issuer, key='admin', session=session)
 
 
-def perm_get_ssh_challenge_token(issuer, kwargs, *, session: "Optional[Session]" = None):
+def perm_get_auth_token_ssh(issuer: "InternalAccount", kwargs: dict, *, session: "Optional[Session]" = None) -> bool:
     """
-    Checks if an account can request a challenge token.
+    Checks if an account can request an ssh token.
```
dchristidis, line 1043: Please capitalise ‘SSH’.

dchristidis (dismissed):
Well done. I hope you won’t mind my being pedantic, but please prefix the commit message with a component (the shortened ‘Authentication:’ or ‘Auth:’ should be enough). The I guess the already included colon would be better replaced with a semicolon.

bjwhite-fnal:
I don't think I asked for this, but when this does eventually get merged could we please have it backported to 1.29? Thanks!

bari12 (approved):

bari12:
@bjwhite-fnal This is merged, but the cherry-picking to the 1.29 release branch is a bit cumbersome. Can you submit a second PR specifically for the release-1.29-LTS branch with these changes. You can skip all the type checking changes, since they are not necessary for 1.29. Thanks!

PR #6160, opened 2023-04-05
Tests: explicitly mount certificates into containers

rcarpa:
I discovered recently that these certificates are hardcoded all over the rucio repositories. This includes, at least, the rucio/containers; and the rucio/k8s-tutorial; repositories. Most things only worked correctly because these certificates where bundled into the containers on build time.

We even went as far as bundling the development CA into some production images. We dodged the huge security issue bullet only because it was placed (into the correct directory) with and incorrect name, so apache ignored it.

This PR re-generates the certificates and explicitly mounts them inside docker containers. This will allow us to follow up with a commit into the rucio/containers repository to not bundle development certificates into production images anymore.


etc/docker/dev/docker-compose-storage-alldb.yml
```diff
@@ -4,6 +4,14 @@ services:
     image: docker.io/rucio/rucio-dev
     command: ["sleep", "infinity"]
     volumes:
+      - ../../certs/rucio_ca.pem:/etc/grid-security/certificates/5fca1cb1.0:z
```
mlassnig, line 7: will the hash stay the same?

rcarpa: yes, I was surprised too. It probably only depends on the common name

mlassnig (commented):
This is OK, if the hash stays the same.

mlassnig (approved):

rcarpa:
Please merge https://github.com/rucio/containers/pull/243 after merging this one

PR #6161, opened 2023-04-05
Fixing heartbeat return

ivukotic:
Heartbeat endpoint was not working correctly for either get or post. 

Get was trying to return table rows that are not json serializable while post endpoint was not returning anything if post was correct (so it returned 500).

This fixes both issues.

### DIFF
--- a/lib/rucio/core/heartbeat.py
+++ b/lib/rucio/core/heartbeat.py
@@ -226,7 +226,7 @@ def list_heartbeats(*, session: "Session"):
     for r in range(len(result)):
         o = result[r]
         json_result.append({
-            "readable": o[0],
+            "executable": o[0],
             "hostname": o[1],
             "pid": o[2],
             "thread_name": o[3],
--- a/lib/rucio/core/heartbeat.py
+++ b/lib/rucio/core/heartbeat.py
@@ -133,15 +133,15 @@ def live(executable, hostname, pid, thread=None, older_than=600, hash_executable
     query = session.query(Heartbeats.hostname,
                           Heartbeats.pid,
                           Heartbeats.thread_id)\
-                   .with_hint(Heartbeats, "index(HEARTBEATS HEARTBEATS_PK)", 'oracle')\
-                   .filter(Heartbeats.executable == hash_executable)\
-                   .filter(Heartbeats.updated_at >= datetime.datetime.utcnow() - datetime.timedelta(seconds=older_than))\
-                   .group_by(Heartbeats.hostname,
-                             Heartbeats.pid,
-                             Heartbeats.thread_id)\
-                   .order_by(Heartbeats.hostname,
-                             Heartbeats.pid,
-                             Heartbeats.thread_id)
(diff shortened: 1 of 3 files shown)


lib/rucio/core/heartbeat.py
```diff
@@ -133,15 +133,15 @@ def live(executable, hostname, pid, thread=None, older_than=600, hash_executable
...
-                   .with_hint(Heartbeats, "index(HEARTBEATS HEARTBEATS_PK)", 'oracle')\
-                   .filter(Heartbeats.executable == hash_executable)\
-                   .filter(Heartbeats.updated_at >= datetime.datetime.utcnow() - datetime.timedelta(seconds=older_than))\
-                   .group_by(Heartbeats.hostname,
-                             Heartbeats.pid,
-                             Heartbeats.thread_id)\
-                   .order_by(Heartbeats.hostname,
-                             Heartbeats.pid,
-                             Heartbeats.thread_id)
+        .with_hint(Heartbeats, "index(HEARTBEATS HEARTBEATS_PK)", 'oracle')\
```
bari12, line 136: Why change the indentation here?

lib/rucio/core/heartbeat.py
```diff
@@ -217,11 +217,24 @@ def list_heartbeats(*, session: "Session"):
...
-                                                          Heartbeats.thread_name)
-
-    return query.all()
+                          Heartbeats.created_at,
+                          Heartbeats.payload).order_by(Heartbeats.readable,
+                                                       Heartbeats.hostname,
+                                                       Heartbeats.thread_name)
+    result = query.all()
+    json_result = []
+    for r in range(len(result)):
```
bari12, line 226: Better do something like https://github.com/rucio/rucio/blob/b0f7064bc4b07f5be704d656bacdd96e7a03426b/lib/rucio/core/rule.py#L1173-L1776 which stays generic in case we add columns at some point in the future.

lib/rucio/core/heartbeat.py
```diff
@@ -241,10 +254,10 @@ def list_payload_counts(executable, older_than=600, hash_executable=None, *, ses
         hash_executable = calc_hash(executable)
     query = session.query(Heartbeats.payload,
                           func.count(Heartbeats.payload))\
-                   .filter(Heartbeats.executable == hash_executable)\
-                   .filter(Heartbeats.updated_at >= datetime.datetime.utcnow() - datetime.timedelta(seconds=older_than))\
-                   .group_by(Heartbeats.payload)\
-                   .order_by(Heartbeats.payload)
+        .filter(Heartbeats.executable == hash_executable)\
```
bari12, line 257: Why the indentation change?

bari12 (changes requested):
Similar to the other PR, please just do one commit

ivukotic:
will close this one and replace it with https://github.com/rucio/rucio/pull/6176.
it is rather annoying that when I save in vs code, it reformats long rows differently. It would be nice if required formatting is in .vscode...

Issue #6107, opened 2023-03-01
finalize migration to unified config table + config file


rcarpa:
We still have a couple of places in the code where we access the config table directly instead of going through wrappers which allow both using the config table and config file. 

For example: 
https://github.com/rucio/rucio/blob/ac95ef3fabb8e1fcabddf1075cc19ae06db44a34/lib/rucio/core/topology.py#L68 
But there should be other places like that. 

Change the code to allow these configurations to both be set in the config table ; and in config file.

Issue #6164, opened 2023-04-06
get rid of the `use_multihop` configuration option


rcarpa:
This is to be done in a major rucio release. 

The configuration option is redundant since https://github.com/rucio/rucio/issues/6163 . To disable multihop, it's enough to set it to an empty or invalid rse expression.

PR #6165, opened 2023-04-06
Transfers: allow configuring intermediate hop RSE expression. Fix #6163

### LINKED ISSUE #6163: allow configuring intermediate hop RSE expression
RSEs allowed to be used as an intermediate hop are selected using a hard-coded RSE expression:
https://github.com/rucio/rucio/blob/ac95ef3fabb8e1fcabddf1075cc19ae06db44a34/lib/rucio/core/topology.py#L73

We would like to allow this expression to be configurable. In particular, this would allow to have a dynamic list of RSEs, which depends on their utilisation. For example: `available_for_multihop=true&freespace>500|BNLHPC_DATADISK` 

This would also allow people to activate multihop globally with one configuration value : `*` for the RSE expression. Instead of having to set an attribute on each RSE

rcarpa:
This will allow to have a dynamic list of RSEs, which depends on their utilisation. For example: `available_for_multihop=true&freespace>500`.

It will also allow people to activate multihop globally with one configuration value: `*`.

The commit also fixes one configuration retrieval related to #6107.

### DIFF
--- a/lib/rucio/core/topology.py
+++ b/lib/rucio/core/topology.py
@@ -23,10 +23,9 @@ from sqlalchemy import select, false
 
 from rucio.common.utils import PriorityQueue
 from rucio.common.cache import make_region_memcached
-from rucio.common.config import config_get_int
+from rucio.common.config import config_get_int, config_get
 from rucio.common.exception import NoDistance, RSEProtocolNotSupported, InvalidRSEExpression
 from rucio.core.rse import RseCollection, list_rses
-from rucio.core.config import get as core_config_get
 from rucio.core.rse_expression_parser import parse_expression
 from rucio.db.sqla import models
 from rucio.db.sqla.session import read_session, transactional_session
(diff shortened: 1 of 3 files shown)


lib/rucio/core/topology.py
```diff
@@ -65,12 +64,13 @@ def __init__(
...
+        include_multihop = config_get('transfers', 'use_multihop', default=False, expiration_time=600, session=session)
+        multihop_rse_expression = config_get('transfers', 'multihop_rse_expression', default='available_for_multihop=true', expiration_time=600, session=session)
 
         multihop_rses = set()
         if include_multihop:
             try:
-                multihop_rses = {rse['id'] for rse in parse_expression('available_for_multihop=true', session=session)}
+                multihop_rses = {rse['id'] for rse in parse_expression(multihop_rse_expression, session=session)}
             except InvalidRSEExpression:
                 pass
```
dchristidis, line 75: If `include_multihop` is `True` and the RSE expression evaluates to an empty set, wouldn’t it be prudent to have a warning-level log entry?

rcarpa: no. But I don't intend to keep that configuration : https://github.com/rucio/rucio/issues/6164 it's redundant

dchristidis: Even so, while it doesn’t have to be placed here specifically, I would find it highly useful to have it *somewhere*. For example, when the Preparer starts or as part of `run_once()`, log something like ‘Multi-hop is not available’?

This comes from a common issue: Rucio allows operators to make bad configuration decisions. Maybe it shouldn’t stop you from doing so, but at least it should warn you.

dchristidis (dismissed):

lib/rucio/core/topology.py
rcarpa: Done

lib/rucio/core/topology.py
```diff
@@ -65,14 +64,17 @@ def __init__(
...
+            if multihop_rse_expression and multihop_rse_expression.strip() and not multihop_rses:
```
dchristidis, line 76: Was this latest change necessitated by the type hints?

rcarpa: yes :smiling_face_with_tear:

dchristidis: How unfortunate.

dchristidis (approved):

bari12 (approved):

Issue #6159, opened 2023-04-04
Make Virtual Placement return only complete dataset replicas


dchristidis:
### Description

https://github.com/rucio/rucio/blob/a86e74e9a3ee45bf8ffd2c040dd1e2eb4e28ea17/lib/rucio/core/replica.py#L3542-L3551

This checks that the dataset exists on a disk RSE which has a `root://` protocol for reading. It should also check that the dataset is fully replicated.

Reported by R. Walker.

bari12:
@ivukotic are you available to implement this change?

ivukotic:
I could do it if I would know how to ask if the replica is complete... can you please let me know?

bari12:
Should be an easy change in principle. In the loop above, `list_dataset_replicas` should give you all the necessary information already, including a `state` entry (`AVAILABLE`/`UNAVAILABLE`)

ivukotic:
ah OK. I will try to do it today.

ivukotic:
solved by: https://github.com/rucio/rucio/pull/6162

PR #6169, opened 2023-04-06
Patch 6159 checks for available replicas

ivukotic:
VP replicas should be delivered only when there is an available replica that is not on the TAPE.
This change just adds a condition that the replica is AVAILABLE.

### DIFF
--- a/lib/rucio/core/replica.py
+++ b/lib/rucio/core/replica.py
@@ -3133,15 +3133,15 @@ def get_and_lock_file_replicas_for_dataset(scope, name, nowait=False, restrict_r
                                           models.DataIdentifierAssociation.md5,
                                           models.DataIdentifierAssociation.adler32,
                                           models.RSEFileAssociation)\
-                                   .with_hint(models.DataIdentifierAssociation,
-                                              "INDEX_RS_ASC(CONTENTS CONTENTS_PK) NO_INDEX_FFS(CONTENTS CONTENTS_PK)",
-                                              'oracle')\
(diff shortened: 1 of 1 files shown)


dchristidis (changes requested):
Thanks for bearing with me. This is better, but note quite there yet.

* There are three commits, changing about 30 lines. This should be a single commit, only adding two lines.
* The commit message format is proper, but the message itself could be improved (e.g. `Replicas: skip unavailable datasets for VP; Fix #6159`).

ivukotic:
> Thanks for bearing with me. This is better, but note quite there yet.
> 
> * There are three commits, changing about 30 lines. This should be a single commit, only adding two lines.
> * The commit message format is proper, but the message itself could be improved (e.g. `Replicas: skip unavailable datasets for VP; Fix #6159`).

now done as requested

dchristidis (changes requested):
This PR now has four commits.

ivukotic:
is that a problem? we just do squash merge and it will be one...

dchristidis:
Yes, it’s a problem (otherwise I would not have asked!) We use the ‘rebase and merge’ strategy, which would mean that all four commits go into the history.

ivukotic:
so what would you like me to do? close this PR and create a new one?

dchristidis:
No, a new PR won’t be necessary. You can rebase and fix the issue locally and then force-push.

ivukotic:
this got very messed up. will close this one and create a new one

ivukotic:
Hi,

Rebase went horribly wrong, so I deleted that PR and created a new one:
https://github.com/rucio/rucio/pull/6174

Ilija

PR #6171, opened 2023-04-11
Transfers: apply hop penalty from the node, not towards. Fix #6170

### LINKED ISSUE #6170: incorrect distance logged in conveyor logs
All distances are shifted by 10 (the hop_penalty). This has absolutely no impact on the final outcome, because absolutely all paths get an additional +10. But this is confusing for people who try to investigate problems and see that numbers don't match.

rcarpa:
This way it will not be applied on the first hop, while still correctly
using the RSE attribute for multi-hop transfers. 

This is a slight behavioral change. Under some combinations of short
distances, more single hops will be preferred over multihops. But it's 
unlikely that anybody will complain about this behavioral change. 
Moreover, it's more in line with the definition of "hop_penalty".

### DIFF
--- a/lib/rucio/core/topology.py
+++ b/lib/rucio/core/topology.py
@@ -161,10 +161,12 @@ class Topology:
                 if adjacent_node not in remaining_sources and adjacent_node not in self.multihop_rses:
                     continue
 
-                try:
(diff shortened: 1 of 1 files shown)


dchristidis (commented):
Please note that the commit message does not reference the correct issue.

rcarpa:
> Please note that the commit message does not reference the correct issue.

thank you, fixed.

dchristidis (approved):

PR #6178, opened 2023-04-13
Traces: better handle stomp connections. Closes #6177

### LINKED ISSUE #6177: bad activemq connection handling in kronos
`--sleep-time-files` in Kronos defines the interval at which to re-create a new activemq handler. But also the timeout of that connection. 

https://github.com/rucio/rucio/blob/b0f7064bc4b07f5be704d656bacdd96e7a03426b/lib/rucio/daemons/tracer/kronos.py#L377

As a consequence, Kronos is re-connecting to the brokers each `--sleep-time-files` seconds, while the currently active connections are force-closed, sometimes resulting in errors putting back the messages into the queue and most probably lost messages. 

This, obviously, is not the correct way of handing a connection. It's possible that the only problem is related to the value of timeout being too small

rcarpa:

### DIFF
--- a/lib/rucio/daemons/tracer/kronos.py
+++ b/lib/rucio/daemons/tracer/kronos.py
@@ -25,14 +25,13 @@ import functools
 from json import loads as jloads, dumps as jdumps
 from threading import Event, Thread
 from time import time
-from typing import Dict
 
 import rucio.db.sqla.util
 from rucio.daemons.common import HeartbeatHandler, run_daemon
-from rucio.common.config import config_get, config_get_bool, config_get_int
+from rucio.common.config import config_get, config_get_bool, config_get_int, config_get_list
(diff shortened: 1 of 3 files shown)


bari12 (approved):

lib/rucio/daemons/tracer/kronos.py
```diff
@@ -367,10 +361,11 @@ def run_once_kronos_file(heartbeat_handler: HeartbeatHandler, return_values: Dic
         ssl_key_file=ssl_key_file,
         ssl_cert_file=ssl_cert_file,
         timeout=sleep_time,
```
yuyiguo, line 363: Can you explain why timeout=sleep_time? In cms case, the sleep_time is 120 seconds. Why the connection should be timed out in such a short time?   We have messages to be consumed all the time, should the connection last for hours?

Issue #6141, opened 2023-03-20
Rucio client lists replicas in wrong order


jamesp-epcc:
### Description

When using the `--sort` option with `rucio list-file-replicas`, the replicas are sorted correctly on the server, but the client does not preserve this ordering when printing the list of replicas. I think this is because the sorted replicas are in the `pfns` member of the JSON returned from the server, but the client iterates based on the `rses` member, which is not sorted. To fix this, the client should respect the ordering in the `pfns` member when displaying the list of replicas.

### Steps to reproduce

1. Create several replicas of a file.
2. Run `rucio list-file-replicas` and pass the `--sort` option to sort the replicas according to some criteria (it doesn't matter which as this affects all sorting algorithms).
3. Observe that the list printed by the client is not in the right order. (Inspecting the actual JSON returned from the server will confirm that this is a client problem).

### Rucio Version

1.30.0

dsgeorge:
With rucio 1.31.5 I am still seeing replicas returned in the wrong order by the python client used in the ATLAS pilot. Is this expected to be fixed?

Issue #6114, opened 2023-03-03
Customisable replica sorter algorithm


jamesp-epcc:
### Description

I plan to add a new replica sorter algorithm that is similar to the current geo IP algorithm, but can download a custom (VO-specific) graph of RSEs, sites and distances, and use this graph to sort replicas in order of preference.

### Motivation

For some VOs (such as DUNE), the current geo IP replica sorter does not map well onto their storage infrastructure, and replica selection would be more optimal with a custom algorithm that takes into account the specifics of the network and storage being used. Rather than write a DUNE-specific solution, I plan to add a customisable algorithm that can also be used by other experiments.

### Change

Propose a format for a VO-specific description file containing all the required information.
Add a new replica sorter algorithm that downloads this file from a configured location and uses it to sort replicas.

PR #6184, opened 2023-04-17
Replicas: Implement custom replica sorter algorithm #6114

jamesp-epcc:
This PR implements a replica sorter algorithm that uses a custom data table fetched from a configurable URL to provide distances between sites.

(At present the sorted output when using the command line client may not be correct due to bug #6141, but that is a separate issue).

### DIFF
--- a/lib/rucio/core/replica_sorter.py
+++ b/lib/rucio/core/replica_sorter.py
@@ -22,7 +22,7 @@ from datetime import datetime, timedelta
 from math import asin, cos, radians, sin, sqrt
 from pathlib import Path
 from tempfile import TemporaryDirectory, TemporaryFile
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Union
 from urllib.parse import urlparse
 
 import geoip2.database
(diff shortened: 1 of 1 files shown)


lib/rucio/core/replica_sorter.py
```diff
@@ -158,6 +158,59 @@ def __get_distance(se1, client_location, ignore_error):
...
+            custom_distance_download_time = datetime.now()
+            custom_distance_table = new_custom_distance_table
+        else:
+            raise Exception('Cannot download custom distance table: %s, Code: %s, Error: %s' % (download_url,
+                                                                                                result.status_code,
+                                                                                                result.text))
+
+
+def __get_distance_custom(rse: Union[tuple, str], client_location: dict) -> float:
+    # get RSE name out of tuple if necessary
```
bari12, line 219: I think this should be re-written to use the memcache REGION which is also used above. It is better than to rely on these module-loaded dictionaries.

lib/rucio/core/replica_sorter.py
```diff
@@ -158,6 +158,59 @@ def __get_distance(se1, client_location, ignore_error):
     return cache_val
 
 
+custom_distance_table = {}
+custom_distance_download_time = None
+
+
+def __download_custom_distance_table() -> None:
```
bari12, line 161: Please add the usual method description etc. I know this is an internal method, but it would be helpful to also explain the format of the downloaded table, possible values, etc.

lib/rucio/core/replica_sorter.py
```diff
@@ -158,6 +158,59 @@ def __get_distance(se1, client_location, ignore_error):
...
+    global custom_distance_table, custom_distance_download_time
```
bari12: See comment below with memcache, maybe you can do this without having to use `global`?

lib/rucio/core/replica_sorter.py
```diff
@@ -158,6 +158,59 @@ def __get_distance(se1, client_location, ignore_error):
...
+                new_custom_distance_table[site][rse] = distance
+            custom_distance_download_time = datetime.now()
+            custom_distance_table = new_custom_distance_table
+        else:
+            raise Exception('Cannot download custom distance table: %s, Code: %s, Error: %s' % (download_url,
+                                                                                                result.status_code,
+                                                                                                result.text))
+
+
+def __get_distance_custom(rse: Union[tuple, str], client_location: dict) -> float:
```
bari12, line 212: Also add method description

lib/rucio/core/replica_sorter.py
```diff
@@ -204,6 +257,8 @@ def sort_replicas(dictreplica: "Dict", client_location: "Dict", selection: "Opti
     # all sorts must be stable to preserve the priority (the Python standard sorting functions always are stable)
     if selection == 'geoip':
         replicas = sort_geoip(dictreplica, client_location, ignore_error=True)
+    elif selection == 'custom':
```
bari12: Maybe `custom_table`? `custom` alone seems a bit ambiguous.

bari12 (changes requested):

lib/rucio/core/replica_sorter.py
jamesp-epcc: I've modified the code to store the data in the memcache REGION.

lib/rucio/core/replica_sorter.py
jamesp-epcc: I have added the method description including the table format.

lib/rucio/core/replica_sorter.py
jamesp-epcc: I'm now using memcache instead of `global`

lib/rucio/core/replica_sorter.py
jamesp-epcc: Done

lib/rucio/core/replica_sorter.py
jamesp-epcc: I've renamed it to `custom_table`.

bari12 (approved):

PR #6195, opened 2023-04-25
Replace all legacy-formatted X509 identities #6182

### LINKED ISSUE #6182: Replace all instances of X509 identities presented in the legacy format
### Description

Since release 1.30, Rucio fully supports both the legacy and the RFC-based format for X509 identities. This is configured using the `LegacyDNStringFormat` option of `mod_ssl`, see [this](https://github.com/rucio/rucio/issues/5842#issuecomment-1232840554) comment for some additional details. By default, it is disabled, meaning that X509 identities must be stored in the RFC format.

However, all of the examples, testing and documentation still use the legacy format. This can be a source of confusion.

### Motivation

Make the RFC-based format primary, without removing support for the legacy one.

### Change

Replace all instances of X509 identities presented in the legacy format with the RFC-based one. Add a section in the documentation explaining the two formats, what Rucio supports, and how to choose between the two.

dchristidis:


maany (approved):
Looks good

PR #6191, opened 2023-04-19
Lifetime Model: Deduplicate input file : Closes #6146

cserf:

PR #6198, opened 2023-04-28
Lifetime Model : Resolve containers to datasets : Close #6147

### LINKED ISSUE #6147: Convert containers in Lifetime Model exceptions to datasets
### Description

Currently, if the input file provided by the user includes containers, the Rucio client will show a warning and ignore them.

### Motivation

CREM decided that it would be more helpful for users to evaluate it to a list of datasets at the time the exception is being submitted.

### Change

 1. Show a warning that the container is being replaced by datasets (it must clearly state that it works on the current content only).
 2. Treat those datasets as if they were already included in the input file, in place of the container.

cserf:
Lifetime Model : Resolve containers to datasets : Close #6147

### DIFF
--- a/bin/rucio
+++ b/bin/rucio
@@ -243,6 +243,19 @@ def get_client(args):
     return client
 
 
+def __resolve_containers_to_datasets(scope, name, client):
+    """
+    Helper function to resolve a container into its dataset content.
+    """
+    datasets = []
+    for did in client.list_content(scope, name):
+        if did['type'] == 'DATASET':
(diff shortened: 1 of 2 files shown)


dchristidis (changes requested):
Seems to include changes for #6191.

cserf:
> Seems to include changes for #6191.

Yes, the same change was needed. Since #6191 was merged, I rebased. Should be OK now

bari12 (approved):

PR #6202, opened 2023-05-03
Deletion : The Reaper is not deleting obsolete replicas in a minimally-configured Rucio instance : Closes #6148

### LINKED ISSUE #6148: The Reaper is not deleting obsolete replicas in a minimally-configured Rucio instance
### Description

If the RSE is minimally-configure (i.e. no additional RSE attributes beyond the identity attribute and no RSE usage counters beyond `rucio`), then the Reaper won’t do anything with this RSE, not even delete the obsolete replicas.

### Steps to reproduce

 1. Create a new RSE, in an environment that does not create additional RSE attributes or RSE usage counters.
 2. Take the necessary steps so that there is one or more deletable replicas with a Unix epoch tombstone.
 3. Run the Reaper.

One will only see `Nothing to delete on RSE` messages in the logs. The only way to overcome this is to set the `greedyDeletion` RSE attribute to `True`.

### Rucio Version

_No response_

### Additional Information

https://github.com/rucio/rucio/blob/bc71856931bf8384fb6715b1ae977a32ef30a0ac/lib/rucio/daemons/reaper/reaper.py#L362-L371

The issue is that, if `source_for_used_space` and `source_for_total_space` point to an RSE usage counter that does not exists and it’s the same for `obsolete`, then this function will return `(0, False)`.

Either the Reaper should be able to work correctly without `obsolete` or its calculation should be implemented in the Abacus daemon so that it’s available in a default Rucio installation.

cserf:
Rewrite/simplification of the `__check_rse_usage` so that the reaper is able to run on `OBSOLETE` replicas even if the `rse_usage` is not properly set.

### DIFF
--- a/lib/rucio/daemons/reaper/reaper.py
+++ b/lib/rucio/daemons/reaper/reaper.py
@@ -330,64 +330,46 @@ def __check_rse_usage(rse: RseData, greedy: bool = False, logger: "Callable[...,
     :returns: needed_free_space, only_delete_obsolete.
     """
 
+    needed_free_space = 0
     # First of all check if greedy mode is enabled for this RSE
     if greedy:
         return 1000000000000, False
 
     rse.ensure_loaded(load_limits=True, load_usage=True, load_attributes=True)
+    available_sources = [key['source'] for key in rse.usage]
 
     # Get RSE limits
     min_free_space = rse.limits.get('MinFreeSpace', 0)
 
-    # Check from which sources to get used and total spaces
-    # Default is storage
+    # Check from which sources to get used and total spaces (default storage)
+    # If specified sources do not exist, only delete obsolete
     source_for_total_space = rse.attributes.get('source_for_total_space', 'storage')
+    if source_for_total_space not in available_sources:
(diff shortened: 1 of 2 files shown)


lib/rucio/daemons/reaper/reaper.py
```diff
@@ -326,65 +326,47 @@ def __check_rse_usage(rse: RseData, greedy: bool = False, logger: "Callable[...,
     :returns: needed_free_space, only_delete_obsolete.
     """
 
+    needed_free_space = 0
     # First of all check if greedy mode is enabled for this RSE
     if greedy:
         return 1000000000000, False
 
     rse.ensure_loaded(load_limits=True, load_usage=True, load_attributes=True)
+    available_sources = [key['source'] for key in rse.usage]
```
rcarpa:
```
available_sources= {key['source']: key for key in rse.usage}
```
and, bellow, 
```
total = available_sources[source_for_total_space]
used = available_sources[source_for_used_space]
```
seems better

rcarpa (dismissed):

bari12 (approved):

PR #6208, opened 2023-05-15
Transfers: internal refactoring in the conveyor daemons. Closes #6207

### LINKED ISSUE #6190: New dispatcher logic for conveyors
Most rucio daemons poll the database for rows on which to perform some work. 
For example, the conveyor submitter will poll regularly for requests currently the "Queued" state; and transitions them to "Submitted" state. On its turn, the conveyor-poller will then work on requests in "Submitted" state to transition them to next daemon. 

Due to this design decision, the database is used as a work queue by the daemons. When multiple daemons of the same type run on the database, they use a hash function when selecting rows to avoid having two daemons working on the same row.  This has the advantage of only requiring the database to do all rucio's work. There are also disadvantages: 
- the fastest time rucio can take to transfer a file start to finish depends on the polling rate of multiple daemons. 
- it's difficult to have intelligent grouping techniques (example : multiple transfers of the same rule will be submitted randomly by different submitters)
- each instance of the daemon MUST take the work from the database, which is not necessarily desired. For example: if we assume that it takes a lot of time to perform the action on a row (reaper is slow deleting files on some RSE), it's not possible to scale out the part of reaper which deletes files without adding more consumers of the database work queue. 

The goal of this ticket is to re-work the internals of the conveyor daemons to dissociate the part which gets the rows from the database; from the part which performs the work on those rows. 
Longer-term this may allow adding alternative ways to input data to workers. For example, using a queuing system

### LINKED ISSUE #6207: internal refactoring in the conveyor
In preparation for #6189 and #6190, perform a refactoring of the internal conveyor logic to have a more graph-like representation of the topology and to allow stream-like handling of transfers in submitter.

rcarpa:

### DIFF
--- a/bin/rucio-conveyor-submitter
+++ b/bin/rucio-conveyor-submitter
@@ -107,8 +107,8 @@ Note that attempting the use the ``--vos`` argument when in single-VO mode will 
                         help='If True, will also try to submit transfers having blocklisted RSEs as sources')
     parser.add_argument('--sleep-time', action="store", default=600, type=int,
                         help='Seconds to sleep if few requests')
-    parser.add_argument('--max-sources', action="store", default=4, type=int,
-                        help='Maximum source replicas per FTS job')
+    parser.add_argument('--max-sources', action="store", default=5, type=int,
+                        help='Maximum source replicas per multi-source FTS job')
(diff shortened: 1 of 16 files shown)


rcarpa:
This PR builds on top of https://github.com/rucio/rucio/pull/6206, so the first 3 commits are to be reviewed there.

bari12 (approved):
💥

PR #6227, opened 2023-05-26
Tests: implement non-global noparallel marks 

rcarpa:
It is now possible to have "noparallel groups". Tests with the same
group will never be run in parallel, but two different groups can be
run on two different nodes. If no groups are provided to
@pytest.mark.noparallel, the old behavior is maintained: that test
cannot be run concurrently at all.


bari12:
Please rebase

bari12 (dismissed):

bari12:
Looks good to me. Maybe @maany have a second look as well, since quite a few things in tests are changing.

pyrightconfig.json
```diff
@@ -3,8 +3,7 @@
         "lib"
     ],
     "exclude": [
-        "lib/rucio/db",
-        "lib/rucio/tests",
```
maany, line 7: Do we want to run pyright on tests?

etc/docker/dev/README.rst
```diff
@@ -54,9 +54,9 @@ To verify that everything is in order, you can now either run the full unit test
 Alternatively, you can bootstrap the test environment once with the `-i` option and then selectively or repeatedly run test case modules, test case groups, or even single test cases, for example::
 
     tools/run_tests_docker.sh -i
-    tools/pytest.sh test_replica.py
-    tools/pytest.sh -vvv test_replica.py::TestReplicaCore
-    tools/pytest.sh -vvv --full-trace test_replica.py::TestReplicaCore::test_delete_replicas_from_datasets
+    tools/pytest.sh tests/test_replica.py
```
maany, line 57: Just a slight note, for the donkeyrider, we will have to probably specify the values for `--filter-tests` in the tests/<filter> format ;) 
https://github.com/rucio/rucio/blob/81bacae53dbe2486bcff87d55feb0dc65daa7ac0/tools/test/donkeyrider.py#L126C9-L126C9 

Not sure what would be the correct place to document it. Maybe the `donkeyrider.py` itself?

pyrightconfig.json
rcarpa: It only runs inside "lib". See a couple of lines above

etc/docker/dev/README.rst
rcarpa: I'm not sure I understand correctly the issue. Is it just about the paths in filter-tests changing ?

maany (dismissed):
I've reviewed the changes to the rucio pytest plugin and the new test groups. They look good to go.

bari12 (approved):

PR #6241, opened 2023-06-30
Rules: Trigger rule evaluation for attaches to containers; Fix #6234

### LINKED ISSUE #6234: Attaching a collection to a container does not trigger the re-evaluation of its rules
### Description

This appears to be a regression from #4611. It seems to have been partially fixed in #5676 but only for the implementation without temporary tables.

One thing that needs to be investigated is why the related test is successful.

On the operations side, it might be necessary to manually trigger the re-evaluation of all rules on containers.

### Steps to reproduce

1. Attach a dataset to a container that has a rule in `OK` state.
2. Watch the rule remain unaffected.

### Rucio Version

_No response_

### Additional Information

_No response_

bari12:


rcarpa:
Is it clear why the associated test didn't detect the issue? Is there something to fix in the test?

Issue #6225, opened 2023-05-25
Typos in docstrings for Python API of DIDClient


vaustrup:
### Description

There are several typos in the documentation of methods in the DIDClient, which leads to the documentation at [https://rucio.cern.ch/documentation/client_api/didclient](https://rucio.cern.ch/documentation/client_api/didclient) to not be rendered correctly.

Examples are:

`:paran type:` -> `:param type:`
`:meta:` -> `:param meta:`
`:rules:` -> `:param rules:`

### Motivation

The documentation is not rendered correctly currently, with several parameters not appearing for certain methods.

PR #6250, opened 2023-07-05
Documentation: editing params for get_data and add_did in didclient.py #6225

Kwaizer:


lib/rucio/client/didclient.py
```diff
@@ -395,6 +395,7 @@ def get_metadata(self, scope, name, plugin='DID_COLUMN'):
 
         :param scope: The scope name.
         :param name: The data identifier name.
+        :param plugin: 'need description'
```
bari12: Replace by: `Backend Metadata plugin the Rucio server should use to query data`

bari12 (approved):
Please squash commits into one commit next time. I will do it now on merge.

PR #6253, opened 2023-07-05
Testing: Add type checking for rules

bari12:
The type checking doesn't run through due to SQLAlchemy thinking every element of an object is a column. This needs to be corrected later on.


rdimaio:
New PR for reference: https://github.com/rucio/rucio/pull/6516

Issue #6090, opened 2023-02-17
Import Statements Should Be at Top of File


ThePhisch:
As per my conversation with @dchristidis , import statements should be found at the top of the file (and not elsewhere, such as in function definitions). Moving forward, it would make sense to change the location of these import statements.

(Quite obviously, this is issue is of low importance)

Examples:

https://github.com/rucio/rucio/blob/f9d52b7c7afbc1cfa41076f1d24f85299aa06cb8/lib/rucio/tests/test_preparer.py#L145-L150
https://github.com/rucio/rucio/blob/f9d52b7c7afbc1cfa41076f1d24f85299aa06cb8/lib/rucio/tests/test_replica.py#L1187-L1188

PR #6255, opened 2023-07-06
Enhancement: Relocated imports #6090

Kwaizer:
Relocated specified imports on top of the files

### DIFF
--- a/lib/rucio/tests/test_preparer.py
+++ b/lib/rucio/tests/test_preparer.py
@@ -144,13 +144,15 @@ def test_preparer_for_request_without_source(db_session, mock_request_no_source)
 
 
 @pytest.mark.noparallel(reason='uses preparer')
-def test_preparer_without_and_with_mat(db_session, source_rse, dest_rse, mock_request):
+@pytest.mark.parametrize("caches_mock", [{"caches_to_mock": [
+    'rucio.core.rse.REGION'
+]}], indirect=True)
+def test_preparer_without_and_with_mat(db_session, source_rse, dest_rse, mock_request, caches_mock):
     add_rse_attribute(source_rse['id'], 'fts', 'a')
     add_rse_attribute(dest_rse['id'], 'globus_endpoint_id', 'b')
 
-    from rucio.core.rse import REGION
-
-    REGION.invalidate()
+    [cache_region] = caches_mock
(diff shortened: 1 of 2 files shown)


lib/rucio/tests/test_preparer.py
```diff
@@ -20,6 +20,7 @@
 from rucio.core.replica import add_replicas
 from rucio.core.request import list_transfer_requests_and_source_replicas, set_transfer_limit, list_transfer_limits
 from rucio.core.transfer import get_supported_transfertools
+from rucio.core.rse import REGION
```
rcarpa: no, I don't like this one at all. The name 'REGION' is used all over rucio for the "local cache instance". Importing the cache instance from another file without changing it's name is prone to generate confusion.

A correct way to improve this test would be to use the caches_mock fixture. And example: 
https://github.com/rucio/rucio/blob/e2f5808eb265856edba5aa1bf69f66bc2843fdb9/lib/rucio/tests/test_reaper.py#L201

rcarpa (changes requested):

lib/rucio/tests/test_preparer.py
Kwaizer: I'm a bit confused, I should rename it like `from rucio.core.rse import REGION as REGION_test_preparer` for example or delete this import at all and write 
```
[cache_region] = caches_mock
cache_region.invalidate()
``` 
in `test_preparer_without_and_with_mat` function in `test_preparer.py`?

rcarpa: Substitute:
```
@pytest.mark.noparallel(reason='uses preparer')
def test_preparer_without_and_with_mat(db_session, source_rse, dest_rse, mock_request):
    ...
    REGION.invalidate()
```

with:
```
@pytest.mark.noparallel(reason='uses preparer')    
@pytest.mark.parametrize("caches_mock", [{"caches_to_mock": [
    'rucio.core.rse.REGION'
]}], indirect=True)
def test_preparer_without_and_with_mat(db_session, source_rse, dest_rse, mock_request, caches_mock):
    ...
    [cache_region] = caches_mock
    cache_region.invalidate()
```

This way, the test will not pollute the cache for other tests.

rcarpa (approved):

Issue #5578, opened 2022-05-19
Reduce missing type annotations


maany:
Motivation
----------
This issue is targeted to be a sprint, probably at a rucio hackathon, for reducing the number of missing type annotations. This is a follow-up of #4455 

Modification
------------

PR #6252, opened 2023-07-05
Dependencies: prepare for the 32.0.0 release. Closes #6251

rcarpa:

PR #6256, opened 2023-07-06
Core & Internals: type annotate sqlalchemy models. #5578 

rcarpa:

### DIFF
--- a/lib/rucio/core/transfer.py
+++ b/lib/rucio/core/transfer.py
@@ -1096,7 +1096,7 @@ def prepare_transfers(
             logger(logging.WARNING, '%s: all available sources were filtered', rws)
             continue
 
-        update_dict = {
+        update_dict: Dict[Any, Any] = {
             models.Request.state: _throttler_request_state(
                 activity=rws.activity,
                 source_rse=selected_source.rse,
--- a/lib/rucio/core/did.py
+++ b/lib/rucio/core/did.py
@@ -297,7 +297,7 @@ def attach_dids_to_dids(
 
             if not first_iteration:
                 session.query(children_temp_table).delete()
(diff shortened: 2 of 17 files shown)


rcarpa:
The PR is built on top of https://github.com/rucio/rucio/pull/6252 and https://github.com/rucio/rucio/pull/6271

bari12 (approved):

lib/rucio/core/transfer.py
```diff
@@ -1096,7 +1096,7 @@ def prepare_transfers(
             logger(logging.WARNING, '%s: all available sources were filtered', rws)
             continue
 
-        update_dict = {
+        update_dict: Dict[Any, Any] = {
```
dchristidis, line 1099: How did this not raise an exception? `Dict` and `Any` are not imported under normal execution.

rcarpa: This is scary ; but I just debugged it and that code IS executed

dchristidis: I can change it in #6277 if it proves necessary since this line is modified, but I am baffled.

dchristidis: It would appear that this is working perfectly fine, though we can’t explain why. :man_shrugging:

rcarpa: Would still be nice if you fix it in #6277

dchristidis: From [PEP-526](https://peps.python.org/pep-0526/#runtime-effects-of-type-annotations):

> Annotating a local variable will cause the interpreter to treat it as a local, even if it was never assigned to. Annotations for local variables will not be evaluated:
> 
> ```
> def f():
>     x: NonexistentName  # No error.
> ```

Issue #5977, opened 2022-11-15
Webdav Protocol `stat` does not return data as specified


ThePhisch:
Motivation
----------
The docstring for `rse/protocols/webdav.py::Default::stat` states that a dict with two keys (filesize and adler32) are returned. This matches the implementations of `stat` in the SSH and xrootd protocols (although these protocols do not explicitly say that it is adler32 but any one of `GLOBALLY_SUPPORTED_CHECKSUMS`. At the moment, the implementation of webdav only returns the filesize.

Modification
------------
Add an entry to the dictionary containing the checksum.

mohitsatr:
hi @kwaizer, have you solved this one ?

Kwaizer:
hi @mohitsatr, I'm waiting for a review.

PR #6257, opened 2023-07-10
Protocols: Adding an entry to the dictionary #5977

Kwaizer:


lib/rucio/rse/protocols/webdav.py
```diff
@@ -516,6 +516,7 @@ def stat(self, path):
             if file.size is None:
                 continue
 
+            dict_['adler32'] = file.href
```
bari12: href doesn't seem to be the adler32 checksum though? That seems to "conflict" with how href is used above?

lib/rucio/rse/protocols/webdav.py
```diff
@@ -516,6 +517,7 @@ def stat(self, path):
             if file.size is None:
                 continue
 
+            dict_['adler32'] = adler32(file)
```
bari12: This won't work. This is a remote file, you can't (locally) adler32 checksum it. 

The idea of the issue is to really to check if adler32 is mentioned in the propfind, which I doubt. In that case we just remove it from the documented line in 488.

Kwaizer: So I didn't found adler32 too and I removed its mentioning in docstring. 

It is what I was expected to do or I can make other changes here?

bari12 (approved):

Issue #6091, opened 2023-02-20
Improve error reporting during upload when reusing a deleted DID


dchristidis:
### Description

What the operators see:
```
Database exception.
Details: (raised as a result of Query-invoked autoflush; consider using a session.no_autoflush block if this flush is occurring prematurely)
(cx_Oracle.DatabaseError) ORA-20101: Primary key constraint DELETED_DIDS_PK violated
ORA-06512: at "ATLAS_RUCIO.CHECK_DID_UNIQUENESS", line 12
ORA-04088: error during execution of trigger 'ATLAS_RUCIO.CHECK_DID_UNIQUENESS'
[SQL: INSERT INTO atlas_rucio.dids (scope, name, account, did_type, is_open, complete, availability, bytes, length, md5, adler32, expired_at, deleted_at, events, guid, project, datatype, run_number, stream_name, prod_step, version, campaign, task_id, panda_id, lumiblocknr, provenance, phys_group, accessed_at, closed_at, eol_at, is_archive, constituent, access_cnt, created_at, updated_at) VALUES (:scope, :name, :account, :did_type, :is_open, :complete, :availability, :bytes, :length, :md5, :adler32, :expired_at, :deleted_at, :events, :guid, :project, :datatype, :run_number, :stream_name, :prod_step, :version, :campaign, :task_id, :panda_id, :lumiblocknr, :provenance, :phys_group, :accessed_at, :closed_at, :eol_at, :is_archive, :constituent, :access_cnt, :created_at, :updated_at)]
[...]

PR #6258, opened 2023-07-10
Clients: Intoduced new constraint #6091

Kwaizer:


lib/rucio/web/rest/flaskapi/v1/common.py
```diff
@@ -103,6 +103,15 @@ def dispatch_request(self, *args, **kwargs):
                              'servers to handle. Please try again in a few minutes.'),
                     headers=headers
                 )
+            elif 'DELETED_DIDS_PK violated' in str(error):
```
bari12: In principle this would work, but this is the wrong location to put it. The reason why the `QueuePool` error is put in the `ErrorHandlingMethodView` here is that it can happen in ANY REST call. But in this case, the `DELETED_DIDS_PK violated` only happens during registration of a did. So it should only be put into that specific endpoint. As a second point: It would be nice to put the scope/name of the DID in the error message

bari12 (changes requested):

bari12:
Your original intuition to put it in the rest endpoint was right, it was just the wrong one :-) But moving it to the client side is not good, this should really happen server side. (in the right rest endpoint)

Kwaizer:
I put error handling of this violation in `post` method of `dids.py` .
So while uploading and reusing a deleted DID error handling will happen on the server side in more precise rest endpoint.

lib/rucio/web/rest/flaskapi/v1/dids.py
```diff
@@ -656,6 +656,13 @@ def post(self, scope_name):
             return generate_http_error_flask(409, error)
         except AccessDenied as error:
             return generate_http_error_flask(401, error)
+        except DatabaseException as error:
+            if 'DELETED_DIDS_PK violated' in str(error):
+                return generate_http_error_flask(
+                    status_code=406,
+                    exc=error.__class__.__name__,
+                    exc_msg=str('A deleted DID {} with scope {} is reused'.format(name, scope))
+                )
```
bari12, line 665: If the exception does not match that string, you still need to raise it, otherwise this will continue to returning the 201.

bari12 (changes requested):

bari12 (approved):

Issue #6259, opened 2023-07-11
Allow for tape-to-tape transfer for QoS RSE


maatthias:
Rucio enforces stagein requests to be tape-to-disk.  Allow the case for tape-to-tape for QoS RSE to submit a pin lifetime on a stagein request.

PR #6260, opened 2023-07-11
Transfers: Allow for tape-to-tape transfer for QoS RSE #6259

maatthias:


maatthias:
The tape-to-tape blocker was solved with a check for the rse attribute `staging_required` in the class `StageinTransferDefinition` and bypassing the `RucioException`.  This could also have been done by a separate class and definition (something like `StageinQoSTransferDefinition`).

lib/rucio/core/transfer.py
```diff
@@ -895,6 +897,12 @@ def build_transfer_paths(
             candidate_sources = [rws.requested_source] if rws.requested_source in filtered_sources else []
 
         if rws.request_type == RequestType.STAGEIN:
+            # if staging_required RSE make sure source is included
+            if rws.dest_rse.attributes.get('staging_required', None):
+                rse_data = RseData(id_=rws.dest_rse.id)
```
rcarpa: I don't like the manipulation of the list of sources so late in the pipeline. It is prone to introduce bugs which are very difficult to track. Why isn't rucio configured to have the correct RSE returned as a source from the database without this hack ?

maatthias: Hi Radu- the problem that this is trying to solve is that Rucio enforces for stagein request types that the source RSE be tape and destination be disk.  However fts supports file staging by requesting a transfer where destination matches source.  Do you think it better (as I suggest above) to create a new (transfer) class to make this clearer?

rcarpa: I understand what is desired. And that is covered by the code above: 
```
if not destination.rse.attributes.get('staging_required', None):
```

I don't understand why we need the following code: 
```
            if rws.dest_rse.attributes.get('staging_required', None):
                rse_data = RseData(id_=rws.dest_rse.id)
                rse_data.ensure_loaded(load_name=True, load_info=True, load_attributes=True, session=session)
                candidate_sources.append(RequestSource(rse_data))
```

maatthias: if the source is not set here to be same as destination the request is evaluated to state `NO_SOURCES`.  And even specifying `--source-replica-expression` at time of rule creation has no effect here.

rcarpa: That's my whole point: if the source is not retrieved correctly from the database by the [list_transfer_requests_and_source_replicas](https://github.com/rucio/rucio/blob/5fa3f825d85721cc897a81aef39cbfc82144926d/lib/rucio/core/request.py#L346), then the configuration should be updated for it to be returned correctly. Or a fix must be done in that function to return it correctly. We should try to avoid adding a hack at the end of the processing pipeline to fake a source which doesn't exist.

maatthias: OK looking at that now.  `list_transfer_requests_and_source_replicas` in core request.

rcarpa: Can you please provide a high-level description of what you are trying to achieve. I'm completely lost. I see there are many pieces of code being added to rules and transfers over multiple releases, but I have a hard time understanding the big picture: "what is desired"?

Some of the code I mention seems strange. For example the if/else block being identical here: 
https://github.com/rucio/rucio/blob/0fedc5b6c1f02e8fdb873f2961c846a63498a716/lib/rucio/core/rule_grouping.py#L919

Others don't integrate into how things already work. For example, the stager requires a `staging_buffer` attribute to work:
https://github.com/rucio/rucio/blob/12e0d7d4808db450d06ff724d8704ca38c3d391c/lib/rucio/core/transfer.py#L873
so I don't quite get how your intend to use rucio. 

I try to understand the big picture.

maatthias: The ultimate goal here is to send a `copy_pin_lifetime` value to FTS when the destination RSE has an attribute of `staging_required` = True.  The `copy_pin_lifetime` comes from the destination RSE attribute `maximum_pin_lifetime`.  This means that we need a request type of `STAGEIN` with identical source and destination values for FTS.

rcarpa: Then there is probably a bug here: https://github.com/rucio/rucio/blob/0fedc5b6c1f02e8fdb873f2961c846a63498a716/lib/rucio/core/rule_grouping.py#L919 one of the if/else branches must create a STAGEIN request. Isn't it? 

Then list_transfer_requests_and_source_replicas will have to be adapted to work differently for STAGEIN requests generate transfers with same source and destination.

maatthias: It's actually not a bug as this should be a transfer (transfer from somewhere else to a tape RSE if somehow replica `UNAVAILABLE`).  However that else branch is indeed obsolete and I'll remove it.

maatthias: In debugging this some more I see that the request source is actually correct through the pipeline until [here ](https://github.com/rucio/rucio/blob/67746549a7ad0ff1976e3da08d74233a3b274512/lib/rucio/core/transfer.py#L866C46-L866C46).  I do agree that we should handle this earlier but it seems to me moving this special case handling to `list_transfer_requests_and_source_replicas` is effectively the same "trick" but just earlier and doing it in the db query.  But I'm still looking at this and of course appreciate the input @rcarpa

maatthias: Can this be merged?

maatthias: It looks like the autotests failed for python 3.10.  Do I need to rebase/re-push or known issue?

maatthias: @rcarpa Are you OK with the changes?

rcarpa: Hi @maatthias, sorry for the delay, I was on vacation. Please rebase. There where some changes in that field and you'll have to substitute `RseData(id_=dest_rse_id)` with `rse_collection[dest_rse_id]` to avoid undesired behaviors. 
I don't quite like the implementation, but I'll merge it this way for now. Just beware that I intend to re-work this when I'll have time. I'll need help from you to define correctly the user requirements.

maatthias: Hi Radu vacation here as well.  I've rebased, made your suggested change and re-pushed.  Happy to help with anything you need later.

rcarpa (approved):

Issue #5906, opened 2022-10-05
Allow to declare suspicious replicas by RSE and LFN


nsmith-:
Motivation
----------
In #5392 `declare_bad_file_replicas` was modified to allow either a list of PFNs (string) or a list of replicas {'scope': <scope>, 'name': <name>, 'rse_id': <rse_id> or "rse": <rse_name>} as inputs. The `declare_suspicious_file_replicas` API does not have the same feature.

Modification
------------
`declare_suspicious_file_replicas` should be modified to have the same feature.

PR #6261, opened 2023-07-14
Enhancement: Modified declare_suspicious_file_replicas #5906

Kwaizer:


bari12 (changes requested):
In this change you only considered one part of the change, which is the force flag. But not the other part, which is that the input can either be pfns or list of dids. This should be considered :-)

bari12:
@Kwaizer please rebase this

bari12:

Issue #6233, opened 2023-06-22
Increase Clarity of Listing Subscriptions for Users Without Any


bjwhite-fnal:
### Description

Currently when you try to ask Rucio if a user has Subscriptions and they have none, a `SubscriptionNotFound` exception is intentionally raised.

This has been very confusing for myself and other users as we learn to use Subscriptions. It would be much clearer if this just indicated that there were no subscriptions for this user, without the scary looking traceback.

```
(rucio-clients) [bjwhite@GALACTICA:~/rucio-clients]$ rucio-admin subscription list --account root
Subscription not found.
Details: Subscription for account 'root' named 'None' not found
Rucio exited with an unexpected/unknown error, please provide the traceback below to the developers.
Traceback (most recent call last):
  File "/home/bjwhite/rucio-clients/bin/rucio-admin", line 97, in new_funct
    return function(*args, **kwargs)
  File "/home/bjwhite/rucio-clients/bin/rucio-admin", line 924, in list_subscriptions
    subs = client.list_subscriptions(name=args.name, account=account)
  File "/home/bjwhite/rucio-clients/lib64/python3.9/site-packages/rucio/client/subscriptionclient.py", line 99, in list_subscriptions
    raise exc_cls(exc_msg)
rucio.common.exception.SubscriptionNotFound: Subscription not found.
Details: Subscription for account 'root' named 'None' not found
```

bari12:
This needs to be adressed in the [subscription](https://github.com/rucio/rucio/blob/3c693f0a2d48e45e0e0775a96fac28671335574b/lib/rucio/core/subscription.py#L207) core itself.

PR #6264, opened 2023-07-17
Subscription: Adding check for subscription #6233

Kwaizer:


bari12:
I don't see how this PR fixes the problem? :-) Maybe we should discuss on Chat

PR #6270, opened 2023-07-19
Ivukotic/issue4935

ivukotic:
Adds a function to return only heartbeats for a given executable.
Adds a class implementing weighted randezvous hashing.
Changes prefix determination from VP to Rucio.


bari12:
Closing this now since the test cases do not work; Please re-submit once fixed

PR #6277, opened 2023-07-21
Remove deprecated constructs from the typing module #6276

### LINKED ISSUE #6276: Remove deprecated constructs from the `typing` module
### Description

[PEP 585](https://peps.python.org/pep-0585/) describes this in detail. Since we bumped the minimum required Python version to 3.9, we can start removing them.

### Motivation

At some point in the future, we might be required to remove them (see [Deprecation Timeline of Major Features](https://docs.python.org/3/library/typing.html#deprecation-timeline-of-major-features)). Doing so earlier means less work.

### Change

For example, replace instances of `typing.List[str]` with `list[str]`.

dchristidis:


dchristidis:
I’ve checked the issues reported by Pyright and I believe them to be pre-existing and not introduced by this PR.

rcarpa (dismissed):

bari12 (approved):

PR #6290, opened 2023-07-31
Bin: Fix type error in argparse; Fix #6289

### LINKED ISSUE #6289: bin/rucio transfer-timeout default value TypeError
### Description

When building the docs, the following error occurs during `tools/run_in_docker/generate_rest_api_docs.sh`:
```Generating the Executable Help Docs...
  File "rucio/bin/rucio", line 2521, in <module>
    oparser = get_parser()
  File "rucio/bin/rucio", line 2308, in get_parser
    selected_parser.add_argument('--transfer-timeout', dest='transfer_timeout', type=float, action='store', default=config_get_float('download', 'transfer_timeout', False, None), help='Transfer timeout (in seconds). Default: computed dynamically from --transfer-speed-timeout. If set to any value >= 0, --transfer-speed-timeout is ignored.')  # NOQA: E501
  File "/usr/local/lib/python3.8/site-packages/rucio/common/config.py", line 197, in config_get_float
    return float(config_get(
TypeError: float() argument must be a string or a number, not 'NoneType'
```
The error on [line 2308](https://github.com/rucio/rucio/blob/43aabc439fa89cc6c3669dc791ea802768e4f2d8/bin/rucio#L2308) is caused when the function `float()` is applied to the default supplied value `None`. Note that the `type` argument in `argparse.ArgumentParser` is a function that is applied to the input (i.e. a preprocessor, and *not* a type checker).

Some background: A new addition to `bin/rucio` (on said [line 2308](https://github.com/rucio/rucio/blob/43aabc439fa89cc6c3669dc791ea802768e4f2d8/bin/rucio#L2308)) uses `config_get_float(section: str, option: str, raise_excepition: bool, default: _T, *) -> Union[float, _T]` instead of the corresponding `config_get`. Some confusion came up because it was expected that a `config_get_float` function could not return a `None`. (note: these functions are heavily overloaded, it is difficult to the matching overload). 

### Steps to reproduce

1. Run `tools/build_documentation.sh` of the [Rucio Documentation](https://github.com/rucio/documentation) repo.

### Rucio Version

_No response_

### Additional Information

_No response_

ThePhisch:
in `argparse.ArgumentParser`, the `type` argument allows for type conversions, it is a function that is run on the input value. The issue was that the input could be `string` or `None`, but the function `float()` can only deal with `string` (or real numbers).

Fix: Changed the function to also handle `None` correctly.


rcarpa:
As discusses offline, this is not at all the reason of the failure. Something is setup incorrectly in the documentation build environment. In particular, I think the reason is python3.8. From the github action log: 
```
Default: computed dynamically from --transfer-speed-timeout. If set to any value >= 0, --transfer-speed-timeout is ignored.')  # NOQA: E501
  File "/usr/local/lib/python3.8/site-packages/rucio/common/config.py", line 197, in config_get_float
    return float(config_get(
```
Also, the line 197 of `rucio/common/config.py` contains a completely different code now. 
Something is really broken in that action

ThePhisch:
> As discusses offline, this is not at all the reason of the failure. Something is setup incorrectly in the documentation build environment. In particular, I think the reason is python3.8. From the github action log:
> 
> ```
> Default: computed dynamically from --transfer-speed-timeout. If set to any value >= 0, --transfer-speed-timeout is ignored.')  # NOQA: E501
>   File "/usr/local/lib/python3.8/site-packages/rucio/common/config.py", line 197, in config_get_float
>     return float(config_get(
> ```
> 
> Also, the line 197 of `rucio/common/config.py` contains a completely different code now. Something is really broken in that action

you are right, and this is why i am closing this issue. i forgot to do it earlier.

PR #6296, opened 2023-08-02
Transfers: fix conveyor queries for fetching requests

rcarpa:
We want requests which where last processed more than N seconds ago.


dchristidis (approved):

dchristidis (commented):
Maybe needs `<=`?

Issue #5978, opened 2022-11-16
Add size information to list_rules()


dchristidis:
Motivation
----------

When a user has run out of quota, their first action is to list their replication rules. However, that offers no information on their sizes. It would be useful to do so, so that they may start by deleted the most voluminous ones.

Modification
------------

- Modify the core, API, REST, and client API layers to accept an additional option which would join the `rules` and the `dids` tables and include the `dids.bytes` column in the returned results. This should work well for rules on individual files and datasets. For containers, it might be preferable to return `None`.
- Consider making the use of this option the default on the Web UI.
- Make this option available through a command-line argument in the CLI.

PR #6297, opened 2023-08-03
Rules: Size information is returned when listing replication rules #5978

Kwaizer:
Modified `core` to accept the `size` info and return it while listing replication rules

### DIFF
--- a/bin/rucio
+++ b/bin/rucio
@@ -1478,6 +1478,7 @@ def list_rules(args):
                   '%s[%d/%d/%d]' % (rule['state'], rule['locks_ok_cnt'], rule['locks_replicating_cnt'], rule['locks_stuck_cnt']),
                   rule['rse_expression'],
                   rule['copies'],
+                  sizefmt(rule['bytes'], args.human) if rule['bytes'] is not None else 'N/A',
                   rule['expires_at'],
                   rule['created_at'],
                   sep=',')
@@ -1490,9 +1491,10 @@ def list_rules(args):
                           '%s[%d/%d/%d]' % (rule['state'], rule['locks_ok_cnt'], rule['locks_replicating_cnt'], rule['locks_stuck_cnt']),
                           rule['rse_expression'],
                           rule['copies'],
+                          sizefmt(rule['bytes'], args.human) if rule['bytes'] is not None else 'N/A',
                           rule['expires_at'],
                           rule['created_at']])
-        print(tabulate(table, tablefmt='simple', headers=['ID', 'ACCOUNT', 'SCOPE:NAME', 'STATE[OK/REPL/STUCK]', 'RSE_EXPRESSION', 'COPIES', 'EXPIRES (UTC)', 'CREATED (UTC)'], disable_numparse=True))
+        print(tabulate(table, tablefmt='simple', headers=['ID', 'ACCOUNT', 'SCOPE:NAME', 'STATE[OK/REPL/STUCK]', 'RSE_EXPRESSION', 'COPIES', 'SIZE', 'EXPIRES (UTC)', 'CREATED (UTC)'], disable_numparse=True))
     return SUCCESS
 
 
--- a/lib/rucio/core/rule.py
+++ b/lib/rucio/core/rule.py
@@ -714,26 +714,26 @@ def list_rules(filters={}, *, session: "Session"):
     :raises:        RucioException
     """
 
-    query = session.query(models.ReplicationRule)
+    stmt = select(models.ReplicationRule)
     if filters:
         for (key, value) in filters.items():
             if key in ['account', 'scope']:
                 if '*' in value.internal:
                     value = value.internal.replace('*', '%')
-                    query = query.filter(getattr(models.ReplicationRule, key).like(value))
+                    stmt = stmt.where(getattr(models.ReplicationRule, key).like(value))
                     continue
                 # else fall through
             elif key == 'created_before':
-                query = query.filter(models.ReplicationRule.created_at <= str_to_date(value))
+                stmt = stmt.where(models.ReplicationRule.created_at <= str_to_date(value))
                 continue
             elif key == 'created_after':
-                query = query.filter(models.ReplicationRule.created_at >= str_to_date(value))
+                stmt = stmt.where(models.ReplicationRule.created_at >= str_to_date(value))
                 continue
             elif key == 'updated_before':
-                query = query.filter(models.ReplicationRule.updated_at <= str_to_date(value))
+                stmt = stmt.where(models.ReplicationRule.updated_at <= str_to_date(value))
                 continue
             elif key == 'updated_after':
-                query = query.filter(models.ReplicationRule.updated_at >= str_to_date(value))
+                stmt = stmt.where(models.ReplicationRule.updated_at >= str_to_date(value))
                 continue
             elif key == 'state':
                 if isinstance(value, str):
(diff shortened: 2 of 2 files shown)


lib/rucio/core/rule.py
```diff
@@ -28,6 +28,7 @@
 from sqlalchemy import select, update
 from sqlalchemy.exc import IntegrityError, StatementError
 from sqlalchemy.orm.exc import NoResultFound
+from sqlalchemy.orm import joinedload
```
dchristidis: Unused.

lib/rucio/core/rule.py
```diff
@@ -714,7 +715,10 @@ def list_rules(filters={}, *, session: "Session"):
     :raises:        RucioException
     """
 
-    query = session.query(models.ReplicationRule)
+    query = session.query(models.ReplicationRule).join(models.DataIdentifier,
+                                                       (models.ReplicationRule.scope == models.DataIdentifier.scope) &
+                                                       (models.ReplicationRule.name == models.DataIdentifier.name)
+                                                       ).with_entities(models.ReplicationRule, models.DataIdentifier.bytes)
```
dchristidis: 1. Is there a reason for `with_entities()`? Would it not work if you added `models.DataIdentifier.bytes` as a second argument to `query()`?
 2. Please prefer the `and_()` function to the `&` operator.
 3. Since you’re touching this area, would you like to try to first migrate it to the SQLAlchemy 2.0 syntax (as its own commit, referencing #5247)? I believe that it will prove trivial.

lib/rucio/core/rule.py
```diff
@@ -750,10 +754,11 @@ def list_rules(filters={}, *, session: "Session"):
             query = query.filter(getattr(models.ReplicationRule, key) == value)
 
     try:
-        for rule in query.yield_per(5):
+        for rule, data_identifier_bytes in query.yield_per(5):
             d = {}
             for column in rule.__table__.columns:
                 d[column.name] = getattr(rule, column.name)
```
dchristidis: Please take the opportunity to replace with `to_dict()` (in a separate commit).

bin/rucio
```diff
@@ -1490,9 +1491,10 @@ def list_rules(args):
                           '%s[%d/%d/%d]' % (rule['state'], rule['locks_ok_cnt'], rule['locks_replicating_cnt'], rule['locks_stuck_cnt']),
                           rule['rse_expression'],
                           rule['copies'],
+                          rule['size'],
```
dchristidis: I am concerned about the output of `list-rules` being already very long in terms of width. Also, understanding the magnitude from bytes takes effort. I think it will be preferable to replace this with `sizefmt(rule['size']) if rule['size'] is not None else 'N/A'` (and adapt the header accordingly).

lib/rucio/core/rule.py
```diff
@@ -750,10 +754,11 @@ def list_rules(filters={}, *, session: "Session"):
...
+            d['size'] = data_identifier_bytes
```
dchristidis: I wonder whether `bytes` is better than `size` because it’s more consistent with all other attributes that represent volume.

dchristidis (changes requested):

lib/rucio/core/rule.py
Kwaizer: 1. Initially I was considering using the second argument but it evoked several `TypeError`s in tests. But I'll try to change it with two arguments again
3 .So it should be redone like: 
```
query = select(models.ReplicationRule, models.DataIdentifier.bytes).join(
models.DataIdentifier, and_(
(models.ReplicationRule.scope == models.DataIdentifier.scope),
(models.ReplicationRule.name == models.DataIdentifier.name))
```

right?

lib/rucio/core/rule.py
Kwaizer: So I need to make another commit to this pr and then just squash all the commits?

lib/rucio/core/rule.py
dchristidis: Yes, like so. You can remove the parentheses around the comparisons. I hope I’m not misleading you, but a quick attempt of my own appeared to be successful.

lib/rucio/core/rule.py
dchristidis: I always recommend changes that are supposed to maintain exact functionality to be isolated and merged as separate commits (and not squashed). This aids the review process, but also future debugging. Consequently, I find it reasonable for this PR to be made out of three commits: the SQLAlchemy 2.0 migration, this replacement with `to_dict()`, and finally the changes that close the issue related to this PR.

lib/rucio/core/rule.py
dchristidis: Please also replace all instances of the `filter()` method with `where()`.

We’ve adopted a different styling convention for SQL queries which improves their readability. Consequently, I would recommend formatting it like so:
```python
stmt = select(
    models.ReplicationRule,
    models.DataIdentifier.bytes
).join(
    models.DataIdentifier,
    and_(
        models.ReplicationRule.scope == models.DataIdentifier.scope,
        models.ReplicationRule.name == models.DataIdentifier.name
    )
)
```

lib/rucio/core/rule.py
dchristidis: You’ll need to take some time to rewrite your commit history in order to achieve what I described above.

Kwaizer: Sure, no problem

dchristidis: Despite the latest changes, the commit history has room for improvement.

 1. The subject of the first commit is `Rules: SQLAlchemy 2.0 migration #5978`, yet it modifies `bin/rucio` and changes the structure return by `list_rules()`. Also, it introduces some code formatted unconventionally, that is then improved by the third commit.
 2. The second commit is perfect.
 3. The third commit will have to be adjusted, once the first commit is fixed.

In short, the end result (meaning the three commits combined) is perfect. However, your individual commits are introducing some mistakes (e.g. `size` instead of `bytes`) and fixing them afterwards whereas you should be introducing the final iteration of your code immediately.

Let me know in case I’m not being clear or you require assistance.

Kwaizer: I rewrote commit history. Please, have a look, it seems to be complete: 

I deleted and edited changes in my local branch to the point it had only SQLAlchemy 2.0 migration, pushed it, added `to_dict()` - pushed another commit 
and then made final commit. 

In total I had 6 commits in PR and squashed 3 of them. Such operations don't seem to be very effective. Could I have changed commit history easier and faster?

dchristidis: Yes, this is exactly what I was hoping for. Thanks!

`git rebase --interactive` allows all sort of rewrites of your commit history (e.g. edit, merge, split, reorder, and drop commits) in a systematic way. In this particular case, I would have expected something along the lines of:

 1. Start the rebase at the base of your branch, `edit` the first and third commit and `pick` the second commit.
 2. Edit the two files as appropriately, then amend the commit. Alternatively, undo the commit with `git reset`, edit, then commit anew.
 3. Continue the rebase. The second commit should be applied automatically.
 4. The third commit will pose a conflict, due to the changes applied on the first commit. Simply edit (and reintroduce the changes to `bin/rucio` which were removed in the first commit) until you’re happy with the end result.
 5. Finish the rebase.

It might be worth looking up for some specific examples first, but once gotten used to, it will become second nature.

Feel free to mark this conversation as resolved after you’re read my response.

dchristidis (approved):

lib/rucio/core/rule.py
Kwaizer: Such approach seems to be more convenient. Thanks! I'll keep that in mind.

bin/rucio
```diff
@@ -1478,6 +1478,7 @@ def list_rules(args):
                   '%s[%d/%d/%d]' % (rule['state'], rule['locks_ok_cnt'], rule['locks_replicating_cnt'], rule['locks_stuck_cnt']),
                   rule['rse_expression'],
                   rule['copies'],
+                  sizefmt(rule['bytes']) if rule['bytes'] is not None else 'N/A',
```
dchristidis: Forgive me, I forgot about something.

https://github.com/rucio/rucio/blob/a73004e1e6ecc5815c2c11a4c80e641be0b97930/bin/rucio#L1861

For completeness, you need to pass `args.human` as the second argument to `sizefmt()`. This also applies to the other call too.

dchristidis (changes requested):

dchristidis (approved):

Issue #6188, opened 2023-04-18
Incorporate IAM-based testing for token workflows


dchristidis:
### Description

The upcoming development for token-based workflows requires the ability to run an identity provider in our development and testing environment. As a first goal, we need to investigate how well we can integrate IAM in them (ideally at the same level as FTS).

esilvaju:
Talking with @vokac, he recommended to talk with CERN IAM team about this topic, since they should have ATLAS test IAM instance. He raise with them via e-mail and we continue from there (I'm in CC).

vokac:
We discussed that it'll be easier to use existing token issuer for Rucio token development testbed, because it'll make storage configuration easier / identical to the configuration that we need anyway for ATLAS / WLCG tokens. First we considered ATLAS IAM instance with special scope policies defined for Rucio testbed IAM client allowed to get `storage.*:/atlastestdisk/` scopes in the token, but for practical reasons it'll be easier to use similarly configured WLCG IAM. I'll ask one storage provider for each SE implementation (EOS, dCache, StoRM, XRootD) to configure access for WLCG tokens.

bari12:
This isn't really possible, nor a good idea to do. We need this containerised for our own isolated development environments plus for the CI environment on GH. I do not want to connect any of these to an actual instance. This will be a pain to do, requires to have actual credentials used in GH etc. This also must be independent of ATLAS, since we need to be able to run this for other (possibly non-WLCG) communities.

That being said, of course we will need to run tests like @vokac suggested, but this will not come from the CI, nor from the dev environment mentioned here.

esilvaju:
I created a [repository](https://github.com/esilvaju/Rucio-IAM) with docker-compose file and instructions that makes possible to spin up an IDp.

The next step is to integrate this containers to the Rucio test environment.

PR #6299, opened 2023-08-03
Feature: IAM docker compose and assets files #6188

esilvaju:
Related to https://github.com/rucio/rucio/issues/6188

### DIFF
--- a/etc/docker/dev/docker-compose-storage-iam.yml
+++ b/etc/docker/dev/docker-compose-storage-iam.yml
@@ -1,0 +1,200 @@
+version: "3"
+services:
+  rucioclient:
+    image: docker.io/rucio/rucio-dev:latest-alma9
+    command: ["sleep", "infinity"]
+    volumes:
+      - ../../certs/rucio_ca.pem:/etc/grid-security/certificates/5fca1cb1.0:z
+      - ../../certs/hostcert_rucio.pem:/etc/grid-security/hostcert.pem:z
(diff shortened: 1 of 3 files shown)


etc/docker/dev/docker-compose-storage-iam.yml
```diff
@@ -0,0 +1,200 @@
...
+  db-iam:
+    image: mariadb:10.11
+    environment:
+      - TZ=Europe/Paris
+      - MYSQL_ROOT_PASSWORD=supersecret
+      - MYSQL_USER=iam
+      - MYSQL_PASSWORD=secret
+      - MYSQL_DATABASE=iam_db
+    ports:
+      - "3307:3306"
```
rcarpa: please bind ports to localhost. See how it is done above for other containers.

etc/docker/dev/docker-compose-storage-iam.yml
```diff
@@ -0,0 +1,200 @@
...
+  nginx-iam:
+    image: nginx
+    dns_search: cern.ch
+    environment:
+      TZ: Europe/Paris
+      NGINX_HOST: iam
+      NGINX_PORT: 443
+    ports:
+      - "443:443"
```
rcarpa: Same as above. But, also, don't use the privileged port on the host. Re-assign it to some less-probably to be used port.

bari12 (approved):

PR #6301, opened 2023-08-07
Transfers: Do not skip requests failed during submission #6300

### LINKED ISSUE #6300: The Conveyor is not transitioning some requests in failed states
### Description

Requests with state `SUBMITTING`, `SUBMISSION_FAILED` and `LOST` are affected. This appears to be a regression from 21513ab.

https://github.com/rucio/rucio/blob/7ccac3dcf98af3ccd383478c52c3e7fca2752527/lib/rucio/core/request.py#L563-L577

When the Finisher is fetching the requests to work on with `list_and_mark_transfer_requests_and_source_replicas()`, it is also updating them. Though not explicitly done, this also updates the `updated_at` column (as part of `ModelBase`). However, the check below in the Finisher means that all recently-updated requests in the above-mentioned states are silently skipped.

https://github.com/rucio/rucio/blob/7ccac3dcf98af3ccd383478c52c3e7fca2752527/lib/rucio/daemons/conveyor/finisher.py#L293-L295

As such, they end up in an endless cycle which prevents their resubmission.

For the states in `SUBMISSION_FAILED` and `LOST`, there does not seem to be a valid motivation for this: no other daemon will work on them. For `SUBMITTING`, the concern is valid, but such requests are already filtered-out.

https://github.com/rucio/rucio/blob/7ccac3dcf98af3ccd383478c52c3e7fca2752527/lib/rucio/daemons/conveyor/finisher.py#L92-L97

Consequently, we can suppose that this `continue` statement is a leftover from when such a check was not implemented at the time the Finisher was fetching its work. By removing it, we might see some requests being retried quicker than before, but we can expect the Repairer to eventually throttle them.

### Steps to reproduce

 1. Have a request reach one of the above-mentioned states.
 2. Observe that it is never resubmitted.

### Rucio Version

32.0.0rc2

### Additional Information

_No response_

dchristidis:
Following 21513abef, requests are now implicitly updated by `list_and_mark_transfer_requests_and_source_replicas()` at the time they are fetched for processing. However, there is a check in the Finisher to silently skip recently-updated requests. As such, requests failed during submission end up in an endless cycle which prevents their resubmission.

There does not seem to be a valid motivation for this. Requests in states `SUBMISSION_FAILED` and `LOST` are not processed by any other daemon. Requests in state `SUBMITTING` that are recently updated are already skipped in `_fetch_requests()`.

This commit removes that check.

### DIFF
--- a/lib/rucio/daemons/conveyor/finisher.py
+++ b/lib/rucio/daemons/conveyor/finisher.py
@@ -290,9 +290,6 @@ def _finish_requests(reqs, suspicious_patterns, retry_protocol_mismatches, logge
 
             # All other failures
             elif req['state'] in failed_during_submission or req['state'] in failed_no_submission_attempts:
-                if req['state'] in failed_during_submission and req['updated_at'] > (datetime.datetime.utcnow() - FAILED_DURING_SUBMISSION_DELAY):
-                    # To prevent race conditions
-                    continue
                 try:
                     if request_core.should_retry_request(req, retry_protocol_mismatches):
                         new_req = request_core.requeue_and_archive(req, source_ranking_update=False, retry_protocol_mismatches=retry_protocol_mismatches, logger=logger)


dchristidis:
The failing/skipped `release-patch-test` suite can be safely ignored.

bari12 (approved):

PR #6302, opened 2023-08-07
Testing: Remove obsolete command-line argument

dchristidis:
Directory `lib/rucio/tests/` was moved to `tests/` in f61d2ab28.

Related to #6227.


dchristidis:
Actually, the `lib/rucio/tests/` directory remains, but I would argue that its content should not be excluded from the tests.

bari12:
I'll merge it like this, but I would agree that the tests should be pylinted as well.

bari12 (approved):

PR #6304, opened 2023-08-07
Testing: Replace `which` with `command -v` #6303

### LINKED ISSUE #6303: `which` executable is not available in the testing environment
### Description

As seen in the CI output:

```
./tools/test/install_script.sh: line 19: which: command not found
./tools/test/install_script.sh: line 19: which: command not found
```

### Steps to reproduce

 1. Open a PR.

### Rucio Version

_No response_

### Additional Information

_No response_

dchristidis:
The `which` utility is an external utility that has to be installed as an RPM package. We could include it during the container creation or we could use the equivalent `command -v` builtin from the Bash (the default shell).

### DIFF
--- a/tools/test/install_script.sh
+++ b/tools/test/install_script.sh
@@ -16,7 +16,7 @@
 
 set -eo pipefail
 
-echo "* Using $(which python) $(python --version 2>&1) and $(which pip) $(pip --version 2>&1)"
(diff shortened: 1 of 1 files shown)


dchristidis:
Is there anything I’m expected to do here? I would have thought it non-controversial.

bari12 (approved):

Issue #5972, opened 2022-11-09
rse_settings dictionary datatype


ThePhisch:
Motivation
----------
at the moment, the `rsemanager` takes most of its settings from a dictionary, often called `rse_settings`, eg in rsemanager.py::create_protocol. this dictionary is unnecessarily complex and hence obscure. it is also not easy to type-hint

Modification
------------
it would be minimally invasive to type-hint the dictionary, which is possible starting from python 3.8 -- but we currently still want to support py3.6

i would suggest replacing it with a NamedTuple (added in py2.6 or py3.0, respectively) (with the benefit of being immutable, easy to type-hint, and more transparent), or possibly even a custom dataclass.

PR #6305, opened 2023-08-08
Core & Internals: rse_settings dictionary type-hinting #5972

Kwaizer:
As supporting of py3.6 was dropped type-hinting has become possible, so I considered using it for editing `rse_settings` dict parameter of `rsemanager` functions

### DIFF
--- a/lib/rucio/common/types.py
+++ b/lib/rucio/common/types.py
@@ -13,6 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from typing import Any, Optional, TypedDict, Union
+
+
 class InternalType(object):
     '''
     Base for Internal representations of string types
@@ -99,3 +102,57 @@ class InternalScope(InternalType):
     '''
     def __init__(self, scope, vo='def', fromExternal=True):
         super(InternalScope, self).__init__(value=scope, vo=vo, fromExternal=fromExternal)
+
+
+class RSEDomainLANDict(TypedDict):
+    read: Optional[int]
+    write: Optional[int]
+    delete: Optional[int]
+
+
+class RSEDomainWANDict(TypedDict):
+    read: Optional[int]
+    write: Optional[int]
+    delete: Optional[int]
+    third_party_copy_read: Optional[int]
+    third_party_copy_write: Optional[int]
+
+
+class RSEDomainsDict(TypedDict):
+    lan: RSEDomainLANDict
+    wan: RSEDomainWANDict
+
+
+class RSEProtocolDict(TypedDict):
+    auth_token: Optional[str]  # FIXME: typing.NotRequired
+    hostname: str
+    scheme: str
+    port: int
+    prefix: str
+    impl: str
+    domains: RSEDomainsDict
+    extended_attributes: Optional[Union[str, dict[str, Any]]]
+
+
+class RSESettingsDict(TypedDict):
+    availability_delete: bool
+    availability_read: bool
+    availability_write: bool
+    credentials: Optional[dict[str, Any]]
+    lfn2pfn_algorithm: str
+    qos_class: Optional[str]
+    staging_area: bool
+    rse_type: str
+    sign_url: Optional[str]
+    read_protocol: int
+    write_protocol: int
+    delete_protocol: int
+    third_party_copy_read_protocol: int
+    third_party_copy_write_protocol: int
+    id: str
+    rse: str
+    volatile: bool
+    verify_checksum: bool
+    deterministic: bool
+    domain: list[str]
+    protocols: list[RSEProtocolDict]
--- a/lib/rucio/rse/rsemanager.py
+++ b/lib/rucio/rse/rsemanager.py
@@ -20,13 +20,14 @@ from time import sleep
 from urllib.parse import urlparse
 
 from rucio.common import exception, utils, constants
+from rucio.common import types
 from rucio.common.config import config_get_int
 from rucio.common.constraints import STRING_TYPES
 from rucio.common.logging import formatted_logger
 from rucio.common.utils import make_valid_did, GLOBALLY_SUPPORTED_CHECKSUMS
 
 
-def get_rse_info(rse=None, vo='def', rse_id=None, session=None):
+def get_rse_info(rse=None, vo='def', rse_id=None, session=None) -> types.RSESettingsDict:
     """
         Returns all protocol related RSE attributes.
         Call with either rse and vo, or (in server mode) rse_id
@@ -66,7 +67,7 @@ def get_rse_info(rse=None, vo='def', rse_id=None, session=None):
     return rse_info
 
 
-def _get_possible_protocols(rse_settings, operation, scheme=None, domain=None, impl=None):
+def _get_possible_protocols(rse_settings: types.RSESettingsDict, operation, scheme=None, domain=None, impl=None):
     """
     Filter the list of available protocols or provided by the supported ones.
 
@@ -116,7 +117,7 @@ def _get_possible_protocols(rse_settings, operation, scheme=None, domain=None, i
     return [c for c in candidates if c not in tbr]
 
 
-def get_protocols_ordered(rse_settings, operation, scheme=None, domain='wan', impl=None):
+def get_protocols_ordered(rse_settings: types.RSESettingsDict, operation, scheme=None, domain='wan', impl=None):
     if operation not in utils.rse_supported_protocol_operations():
         raise exception.RSEOperationNotSupported('Operation %s is not supported' % operation)
 
@@ -128,7 +129,7 @@ def get_protocols_ordered(rse_settings, operation, scheme=None, domain='wan', im
     return candidates
 
 
-def select_protocol(rse_settings, operation, scheme=None, domain='wan'):
+def select_protocol(rse_settings: types.RSESettingsDict, operation, scheme=None, domain='wan'):
     if operation not in utils.rse_supported_protocol_operations():
         raise exception.RSEOperationNotSupported('Operation %s is not supported' % operation)
 
@@ -141,7 +142,7 @@ def select_protocol(rse_settings, operation, scheme=None, domain='wan'):
     return min(candidates, key=lambda k: k['domains'][domain][operation])
 
 
-def create_protocol(rse_settings, operation, scheme=None, domain='wan', auth_token=None, protocol_attr=None, logger=logging.log, impl=None):
+def create_protocol(rse_settings: types.RSESettingsDict, operation, scheme=None, domain='wan', auth_token=None, protocol_attr=None, logger=logging.log, impl=None):
     """
     Instantiates the protocol defined for the given operation.
 
@@ -191,7 +192,7 @@ def create_protocol(rse_settings, operation, scheme=None, domain='wan', auth_tok
     return protocol
 
 
-def lfns2pfns(rse_settings, lfns, operation='write', scheme=None, domain='wan', auth_token=None, logger=logging.log, impl=None):
+def lfns2pfns(rse_settings: types.RSESettingsDict, lfns, operation='write', scheme=None, domain='wan', auth_token=None, logger=logging.log, impl=None):
     """
         Convert the lfn to a pfn
 
@@ -209,7 +210,7 @@ def lfns2pfns(rse_settings, lfns, operation='write', scheme=None, domain='wan', 
     return create_protocol(rse_settings, operation, scheme, domain, auth_token=auth_token, logger=logger, impl=impl).lfns2pfns(lfns)
 
 
-def parse_pfns(rse_settings, pfns, operation='read', domain='wan', auth_token=None):
+def parse_pfns(rse_settings: types.RSESettingsDict, pfns, operation='read', domain='wan', auth_token=None):
     """
         Checks if a PFN is feasible for a given RSE. If so it splits the pfn in its various components.
 
@@ -231,7 +232,7 @@ def parse_pfns(rse_settings, pfns, operation='read', domain='wan', auth_token=No
     return create_protocol(rse_settings, operation, urlparse(pfns[0]).scheme, domain, auth_token=auth_token).parse_pfns(pfns)
 
 
-def exists(rse_settings, files, domain='wan', scheme=None, impl=None, auth_token=None, vo='def', logger=logging.log):
+def exists(rse_settings: types.RSESettingsDict, files, domain='wan', scheme=None, impl=None, auth_token=None, vo='def', logger=logging.log):
     """
         Checks if a file is present at the connected storage.
         Providing a list indicates the bulk mode.
@@ -292,7 +293,7 @@ def exists(rse_settings, files, domain='wan', scheme=None, impl=None, auth_token
     return [gs, ret]
 
 
-def upload(rse_settings, lfns, domain='wan', source_dir=None, force_pfn=None, force_scheme=None, transfer_timeout=None, delete_existing=False, sign_service=None, auth_token=None, vo='def', logger=logging.log, impl=None):
+def upload(rse_settings: types.RSESettingsDict, lfns, domain='wan', source_dir=None, force_pfn=None, force_scheme=None, transfer_timeout=None, delete_existing=False, sign_service=None, auth_token=None, vo='def', logger=logging.log, impl=None):
     """
         Uploads a file to the connected storage.
         Providing a list indicates the bulk mode.
@@ -491,7 +492,7 @@ def upload(rse_settings, lfns, domain='wan', source_dir=None, force_pfn=None, fo
     return {0: gs, 1: ret, 'success': gs, 'pfn': pfn}
 
 
-def delete(rse_settings, lfns, domain='wan', auth_token=None, logger=logging.log, impl=None):
+def delete(rse_settings: types.RSESettingsDict, lfns, domain='wan', auth_token=None, logger=logging.log, impl=None):
     """
         Delete a file from the connected storage.
         Providing a list indicates the bulk mode.
@@ -534,7 +535,7 @@ def delete(rse_settings, lfns, domain='wan', auth_token=None, logger=logging.log
     return [gs, ret]
 
 
-def rename(rse_settings, files, domain='wan', auth_token=None, logger=logging.log, impl=None):
+def rename(rse_settings: types.RSESettingsDict, files, domain='wan', auth_token=None, logger=logging.log, impl=None):
     """
         Rename files stored on the connected storage.
         Providing a list indicates the bulk mode.
@@ -610,7 +611,7 @@ def rename(rse_settings, files, domain='wan', auth_token=None, logger=logging.lo
     return [gs, ret]
 
 
-def get_space_usage(rse_settings, scheme=None, domain='wan', auth_token=None, logger=logging.log, impl=None):
+def get_space_usage(rse_settings: types.RSESettingsDict, scheme=None, domain='wan', auth_token=None, logger=logging.log, impl=None):
     """
         Get RSE space usage information.
 
--- a/lib/rucio/rse/rsemanager.py
+++ b/lib/rucio/rse/rsemanager.py
@@ -40,20 +40,20 @@ def get_rse_info(rse=None, vo='def', rse_id=None, session=None):
                     id                ...     an internal identifier
                     rse               ...     the name of the RSE as string
                     type              ...     the storage type odf the RSE e.g. DISK
-                    volatile          ...     boolean indictaing if the RSE is volatile
+                    volatile          ...     boolean indicating if the RSE is volatile
                     verify_checksum   ...     boolean indicating whether RSE supports requests for checksums
-                    deteministic      ...     boolean indicating of the nameing of the files follows the defined determinism
-                    domain            ...     indictaing the domain that should be assumed for transfers. Values are 'ALL', 'LAN', or 'WAN'
-                    protocols         ...     all supported protocol in form of a list of dict objects with the followig structure
+                    deterministic      ...     boolean indicating of the naming of the files follows the defined determinism
+                    domain            ...     indicating the domain that should be assumed for transfers. Values are 'ALL', 'LAN', or 'WAN'
+                    protocols         ...     all supported protocol in form of a list of dict objects with the following structure
                     - scheme              ...     protocol scheme e.g. http, srm, ...
                     - hostname            ...     hostname of the site
                     - prefix              ...     path to the folder where the files are stored
                     - port                ...     port used for this protocol
                     - impl                ...     naming the python class of the protocol implementation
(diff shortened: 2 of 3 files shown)


lib/rucio/rse/rsemanager.py
```diff
@@ -17,6 +17,7 @@
 import logging
 import random
 from time import sleep
+from typing import Dict, List, Union
```
dchristidis: Please replace `typing.Dict` with `dict` and `typing.List` with `list`. Refer to #6276.

dchristidis (changes requested):

lib/rucio/rse/rsemanager.py
Kwaizer: Replaced `typing.Dict` and `typing.List`, also corrected some typos. Is everything all right? Do these commits need to be squashed?

dchristidis: Yes, a single commit is preferable (unless you wish to split the spelling fixes in a separate commit).

Kwaizer: I can try to make a separate commit. It will be good for practice

Kwaizer: I have made changes in commit history.

Also Python type check fails in `_get_possible_protocols` due to `bool` type when `protocols` type-hinted as `list[dict[str, Union[str, int, dict[str, int]]]]` (where `bool` is not present) is called.

Please, can you say am I something missing or it is not a fault on my side?

dchristidis: Note that in the first commit you’re importing `typing.Union` (line 20) and breaking the definition of `upload()` (lines 296–297). These don’t match the commit description ‘spelling fixes’ and should be move to the second commit.

Also note the improper break before a comma in line 70.

As for you question, simple type type hints can only get us so far. Suppose you have a dictionary structure of:
```python
d = {
    'a_bool': True,
    'a_list': ['a', 'b']
}
```
If you type hint it as `dict[str, Union[bool, list]`, you aren’t specifying that `a_bool` is expected to be a `bool` and `a_list` is expected to be a `list[str]`. Hence, if somewhere below you do something like `d['a_list'].append('c')`, it will raise an error because `d['a_list']` is hinted as `Union[bool, list]`, so it *may* be a `bool`.

To be frank, I think it’s actually better to use `dict[str, Any]`. It creates a placeholder for future work: using `typing.TypedDict` to unequivocally define the structure and expected content of dictionaries. We’ve discussed it multiple times, but haven’t started working on it.

Kwaizer: Thank you for pointing out issues in first commit!

I will change type-hinting to `dict[str, Any]` in second commit. Don't know if type-hinting with `typing.TypedDict` can be done right in parameters of a function but I think it could look something like this: 
```
Protocols = TypedDict('Protocols',
                      {'scheme': str,
                       'hostname': str,
                       'prefix': str,
                       'port': int,
                       'impl': str,
                       'extended_attributes': dict[str, str],
                       'domains': dict[str, int]
                       })

Rse_settings = TypedDict('Rse_settings',
                         {'id': str,
                          'rse': str,
                          'type': str,
                          'volatile': bool,
                          'verify_checksum': bool,
                          'deterministic': bool,
                          'domain': str,
                          'protocols': list[Protocols]
                          })


def _get_possible_protocols(rse_settings: Rse_settings, operation, scheme=None, domain=None, impl=None):
```
or implementation of such syntax will be discussed later for whole project (like migration to SQLAlchemy 2.0)?

dchristidis: This is my personal opinion, but I believe it to be a structured task that will involve:

 1. List all functions that have dictionaries as parameters or return values.
 2. Identify the minimal set of types that are common across the entire Rucio project.
 3. Define them as `TypedDicts` is a common file like `lib/rucio/common/types.py`.
 4. Use them, refactoring the code as necessary (some arguments can be multiple, incompatible types, which makes type checking cumbersome).

One potential hurdle is types where the existence of elements is optional. [Python 3.11](https://docs.python.org/3/whatsnew/3.11.html#whatsnew311-pep655) offers support for them, but it will be quite some time before we can raising the minimum required version.

Kwaizer: Python versions are gradually dropping in the project, so I don't think it will take much time.

dchristidis: I’ve changed my mind, sorry! I believe it would be a waste to not make use of the work you’ve already done. Here are some guidelines on how to proceed:

 1. Definitions should go to `lib/rucio/common/types.py`.
 2. They should use camel case and have the `Dict` suffix.
 3. They should be defined as classes, inheriting from `typing.TypedDict`.
 4. In case there’s any key whose presence is optional, then that line should end with a comment `# FIXME: typing.NotRequired`.

I would like you to review the code at:

https://github.com/rucio/rucio/blob/a73004e1e6ecc5815c2c11a4c80e641be0b97930/lib/rucio/core/rse.py#L1376-L1450

and then define `RSEDomainWANDict`, `RSEDomainLANDict`, `RSEDomainsDict`, `RSEProtocolDict` and `RSESettingsDict`. Then comes the difficult part: you need to start from that function and add hints when necessary, while also reviewing whether the structure is changed.

Does that interest you? I will be happy to assist along the way.

Kwaizer: Yeah, I will be happy to implement this! 

So to start with, I need to define in `lib/rucio/common/types.py` classes 
`RSEDomainWANDict`, 
`RSEDomainLANDict`, 
`RSEDomainsDict`, 
`RSEProtocolDict` and 
`RSESettingsDict`, 
all with inheritance from `typing.TypedDict`.

Then if key inside one of these classes is optional I write something like:
```
class RSEDomainsDict(typing.TypedDict):
    name: str # FIXME: typing.NotRequired
```
After that I start type-hinting when necessary in `def _format_get_rse_protocols()` using early defined classes for hints while also checking if the dictionaries I have defined can represent all of the possible data structures that can be returned by the function. 

Am I understanding you correctly?

dchristidis: Thank you for the enthusiasm. Yes, that sounds about right. For `_format_get_rse_protocols()` you’ll only need to replace the return type with `RSESettingsDict`, then do the same for `get_rse_protocols()`, then follow until you eventually get back to the RSE Manager.

But start with the definitions. You might want to switch this pull request to a draft for the time being.

Kwaizer: I created custom classes and type-hinted all the way to rsemanager. 

Marked `prefix`  and `extended_attributes` of `RSEProtocolDict` with `#FIXME: typing.NotRequired`.  

Also it seems `_format_get_rse_protocols` needs to be type-hinted more precisely to have compatible return value type (mypy shows it is expected to return dict[str, Any]).

dchristidis: Very well. I’ll have a closer look tomorrow, but already I noticed something.

`typing.NotRequired` should be used when a key may not exist in the dictionary. I have the impression that `prefix` always exists, is a string , but may be an empty string. `extended_attributes` is more complex. It always exists too, but its value may be `None`, `str` or some dictionary whose types are unclear.

At least that’s my understanding from reading `_format_get_rse_protocols()`. Have you found signs to the contrary?

Kwaizer: Oh, I thought that `Optional[]` and `nullable=True` in `models.RSEProtocols` could signify that key may not be passed to the dictionary. I'll check presence of these keys more carefully in this case.

lib/rucio/common/types.py
```diff
@@ -12,7 +12,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
+from typing import TypedDict
```
dchristidis: `lib/rucio/common/types.py:16:1: E302 expected 2 blank lines, found 0`

lib/rucio/common/types.py
```diff
@@ -99,3 +99,43 @@ class InternalScope(InternalType):
...
+    write: int
+    delete: int
+
+
+class RSEDomainWANDict(TypedDict):
+    read: int
+    write: int
+    delete: int
+    third_party_copy_read: int
+    third_party_copy_write: int
```
dchristidis: All protocol priorities should be `Optional[int]` because `None` is a possible value (it’s not yet reflected in the models).

lib/rucio/common/types.py
```diff
@@ -99,3 +99,43 @@ class InternalScope(InternalType):
...
+
+
+class RSEDomainsDict(TypedDict):
+    lan: dict[str, RSEDomainLANDict]
+    wan: dict[str, RSEDomainWANDict]
```
dchristidis: Not exactly. `lan: dict[str, RSEDomainLANDict]` means the value must resemble:
```json
{
    "foo": {
        "read": 0,
        "write": 0,
        "delete": 0
    }
}
```
to be valid. You want `lan: RSEDomainLANDict` instead (similarly for many others).

lib/rucio/common/types.py
```diff
@@ -99,3 +99,43 @@ class InternalScope(InternalType):
...
+
+
+class RSEProtocolDict(TypedDict):
+    hostname: str
+    scheme: str
+    port: int
+    prefix: str  # FIXME: typing.NotRequired
+    impl: str
+    domains: dict[str, RSEDomainsDict]
+    extended_attributes: dict[str, str]  # FIXME: typing.NotRequired
```
dchristidis: As mentioned in a comment above, I believe it’s `prefix: str` and `extended_attributes: Optional[Union[str, dict[str, Any]]]`, but I haven’t checked thoroughly.

lib/rucio/rse/rsemanager.py
```diff
@@ -17,16 +17,18 @@
 import logging
 import random
 from time import sleep
+from typing import Any
```
dchristidis: `lib/rucio/rse/rsemanager.py:20:1: F401 'typing.Any' imported but unused`

lib/rucio/common/types.py
```diff
@@ -99,3 +99,43 @@ class InternalScope(InternalType):
...
+
+class RSESettingsDict(TypedDict):
+    id: str
+    rse: str
+    type: str
+    volatile: bool
+    verify_checksum: bool
+    deterministic: bool
+    domain: str
+    protocols: list[dict[str, RSEProtocolDict]]
```
dchristidis: What about the other keys set below?

https://github.com/rucio/rucio/blob/c4f4eb9902fe0be6ff52d22357f763769c8e9860/lib/rucio/core/rse.py#L1399-L1417

dchristidis (changes requested):

lib/rucio/common/types.py
Kwaizer: I think it's my fault: `get_rse_info` returns them to `rse_settings` parameter, so they are required

lib/rucio/common/types.py
Kwaizer: Just figured it out. I didn't include these keys because they weren't obtained by `rse_settings` parameter but they are still returned by `_format_get_rse_protocols `. I think they are the ones that should be marked with ` # FIXME: typing.NotRequired`. Thank you for pointing it out.

dchristidis: > I think they are the ones that should be marked with ` # FIXME: typing.NotRequired`.

Why? Is there a code path which could result in the structure not containing those keys? Consider the following example:

```python
from typing import NotRequired, TypedDict

class FooDict(TypedDict):
    foo: str
    bar: NotRequired[str]

def foo(add_bar: bool = True) -> FooDict:
    d = {'foo': 'foo'}
    if add_bar:
        d['bar'] = 'bar'
    return d
```

Here, it makes sense to define `bar` as `NotRequired` because a `foo(add_bar=False)` call will result in its absence from the returned dictionary. Is it the case for the ones in `RSESettingsDict`?

lib/rucio/common/types.py
```diff
@@ -13,6 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from typing import TypedDict, Union, Optional, Any
```
dchristidis: Please sort alphabetically.

lib/rucio/common/types.py
```diff
@@ -99,3 +102,59 @@ class InternalScope(InternalType):
...
+
+class RSEProtocolDict(TypedDict):
+    auth_token: Optional[str]  # FIXME: typing.NotRequired
+    hostname: Optional[str]
+    scheme: Optional[str]
+    port: Optional[int]
+    prefix: Optional[str]
+    impl: Optional[str]
+    domains: RSEDomainsDict
+    extended_attributes: Optional[Union[str, dict[str, Any]]]
```
dchristidis: I think you’ve overused `Optional`. None of `hostname`, `scheme`, `port` or `impl` may be allowed to be `None`.

dchristidis (changes requested):

lib/rucio/common/types.py
Kwaizer: Oh, if it should be used in such cases then I don't see where `NotRequired` could be possibly implemented:

- `rse_attributes` isn't always used as an argument for function but the default value `None` will be returned anyway and if-else statements will give values to `lfn2pfn_algorithm`, `verify_checksum` and `sign_url` in `_format_get_rse_protocols`.
- It seems that `get_rse_protocols` and `get_rse_info`  also do not change structure of returned dictionaries.
- Then `get_rse_info` passes `rse_settings` as an argument to parameters of `rsemanager` functions.

Could I missed something?

lib/rucio/common/types.py
dchristidis: In addition, could you point me to where `auth_token` is coming from?

lib/rucio/common/types.py
dchristidis: But if `NotRequired` is not necessary, then that is a good thing! I mentioned it as a solution to a possible situation, as describe in my example above.

lib/rucio/common/types.py
Kwaizer: Sure, 
https://github.com/rucio/rucio/blob/7f0d229ac0b3bc7dec12c6e158bea2b82d414a3b/lib/rucio/rse/rsemanager.py#L189
this line of code made one of the github checks fail so I decided it is a good idea to include `auth_token`.

dchristidis: Thank you. This is one of those unfortunate cases where the dictionary structure changes as it’s passed between layers. I think this is a prime case for `typing.NotRequired`, as you had before.

lib/rucio/common/types.py
```diff
@@ -99,3 +102,58 @@ class InternalScope(InternalType):
...
+    third_party_copy_read_protocol: int
+    third_party_copy_write_protocol: int
+    id: str
+    rse: str
+    type: str
+    volatile: bool
+    verify_checksum: bool
+    deterministic: bool
+    domain: str
+    protocols: list[RSEProtocolDict]
```
dchristidis: Consider this real-life example:

```python
>>> pprint(get_rse_protocols('a8cbe2f23cc44cac8417df73d259d392'))
{'availability_delete': True,
 'availability_read': True,
 'availability_write': True,
 'credentials': None,
 'delete_protocol': 1,
 'deterministic': False,
 'domain': ['lan', 'wan'],
 'id': 'a8cbe2f23cc44cac8417df73d259d392',
 'lfn2pfn_algorithm': 'hash',
 'protocols': [{'domains': {'lan': {'delete': 1, 'read': 1, 'write': 1},
                            'wan': {'delete': 1,
                                    'read': 1,
                                    'third_party_copy_read': 1,
                                    'third_party_copy_write': 1,
                                    'write': 1}},
                'extended_attributes': {'space_token': 'ATLASDATATAPE',
                                        'web_service_path': '/srm/managerv2?SFN='},
                'hostname': 'ccsrm.in2p3.fr',
                'impl': 'rucio.rse.protocols.gfal.Default',
                'port': 8443,
                'prefix': '/pnfs/in2p3.fr/data/atlas/atlasdatatape/',
                'scheme': 'srm'}],
 'qos_class': None,
 'read_protocol': 1,
 'rse': 'IN2P3-CC_DATATAPE',
 'rse_type': 'TAPE',
 'sign_url': None,
 'staging_area': False,
 'third_party_copy_read_protocol': 1,
 'third_party_copy_write_protocol': 1,
 'verify_checksum': True,
 'volatile': False,
 'write_protocol': 1}
 ```

I can see some deviations from the hints you’ve provided:

 * `availability_*` may not be `None`.
 * `domain` is a `list[str]`.
 * `type` does not exist. It’s there with `get_rse` but it’s an `RSEType`, not an `str`. However, that structure looks *very* different.
```python
>>> pprint(get_rse('a8cbe2f23cc44cac8417df73d259d392'))
{'ASN': None,
 'ISP': None,
 'availability': 7,
 'availability_delete': True,
 'availability_read': True,
 'availability_write': True,
 'city': None,
 'continent': None,
 'country_name': None,
 'created_at': datetime.datetime(2014, 3, 17, 10, 22, 27),
 'deleted': False,
 'deleted_at': None,
 'deterministic': False,
 'id': 'a8cbe2f23cc44cac8417df73d259d392',
 'latitude': 45.7825,
 'longitude': 4.8655,
 'qos_class': None,
 'region_code': None,
 'rse': 'IN2P3-CC_DATATAPE',
 'rse_type': <RSEType.TAPE: 'TAPE'>,
 'staging_area': False,
 'time_zone': None,
 'type': <RSEType.TAPE: 'TAPE'>,
 'updated_at': datetime.datetime(2023, 6, 27, 16, 0, 7),
 'vo': 'def',
 'volatile': False}
```

We might need two another class to do this properly for now, but I wonder whether they can be merged in the future.

Kwaizer: Docstring in `rsemanager` confused me: probably it should be `rse_type` here
https://github.com/rucio/rucio/blob/7f0d229ac0b3bc7dec12c6e158bea2b82d414a3b/lib/rucio/rse/rsemanager.py#L39-L42
So we need to create class for `get_rse` and `RSEType` class?
The last one will represent enum. But I wonder why `rse_type` and `type` of `get_rse` prints `<EnumName.EnumMember: 'Value'>` and `get_rse_protocols` prints `'Value'`. This possibly brings some difficulties in creating more universal RSEType class.

dchristidis: > So we need to create class for get_rse and RSEType class?

No, I would suggest sticking to `get_rse_protocols()` for this PR. You’re doing good work!

dchristidis: Forgive me, I might have confused you. For this PR, `rse_type` is a `str`, as seen in:
https://github.com/rucio/rucio/blob/7f0d229ac0b3bc7dec12c6e158bea2b82d414a3b/lib/rucio/core/rse.py#L1410
It’s for the other structure returned by `get_rse()` where it’s an instance of `RSEType`. And what I meant was that we might want to examine the differences and see whether they should converge at some point in the future.

Kwaizer: Thank you! So I made corrections, deleted `type`, type-hinted `rse_type` with `RSEType` imported from `rucio.db.sqla.constants` and marked it with `# FIXME: Introduce new TypedDict class for RSEType` that shows need for future changes when type-hinting for `get_rse` will be implemented.

Or it should be done some other way?

Kwaizer: Ok, then I'll fix it now

> Forgive me, I might have confused you. For this PR, `rse_type` is a `str`, as seen in
> 
> https://github.com/rucio/rucio/blob/7f0d229ac0b3bc7dec12c6e158bea2b82d414a3b/lib/rucio/core/rse.py#L1410
> .

Kwaizer: Any other fixes need to be done or this PR should stay as a draft for now?

lib/rucio/common/types.py
```diff
@@ -13,6 +13,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from rucio.db.sqla.constants import RSEType
```
dchristidis: Unused.

lib/rucio/common/types.py
```diff
@@ -99,3 +103,57 @@ class InternalScope(InternalType):
...
+    impl: str
+    domains: RSEDomainsDict
+    extended_attributes: Optional[Union[str, dict[str, Any]]]
+
+
+class RSESettingsDict(TypedDict):
+    availability_delete: bool
+    availability_read: bool
+    availability_write: bool
+    credentials: Optional[str]
```
dchristidis: No, I do not think the possible type to be a `str`. It would appear to be a `dict[str, Any]` whose structure can vary significantly. Could you investigate and let me know if I am correct?

dchristidis (changes requested):
Almost there! I would also ask that you merge the two commits (there’s little benefit in adding hints `dict[str, Any]` only to replace them immediately after.

lib/rucio/common/types.py
Kwaizer: You're right credentials seem to be a dictionary:
https://github.com/rucio/rucio/blob/7f0d229ac0b3bc7dec12c6e158bea2b82d414a3b/lib/rucio/rse/protocols/webdav.py#L170-L172

dchristidis (approved):
Looking good, thanks!

One thing I was surprised to see is that Pyright finds `RSESettingsDict` to be incompatible with `Mapping[str, Any]`. I have no suspicion as to why, but this isn’t something that should be addressed in this PR.
```
Found 1 new problems in /home/runner/work/rucio/rucio/lib/rucio/daemons/conveyor/submitter.py
  - 1 errors with message """Argument of type "list[RSESettingsDict] | None" cannot be assigned to parameter "rses" of type "list[Mapping[str, Any]] | None" in function "submitter"""".
  ```

Kwaizer:
I can only say for now that mypy checking points out that `_format_get_rse_protocols` needs to be type-hinted more precisely to have compatible return value type (mypy shows it is expected to return dict[str, Any] instead of RSESettingsDict). Maybe this potentially can cause Pyright check failing.
https://github.com/rucio/rucio/blob/7f0d229ac0b3bc7dec12c6e158bea2b82d414a3b/lib/rucio/core/rse.py#L1370-L1373
> Looking good, thanks!
> 
> One thing I was surprised to see is that Pyright finds `RSESettingsDict` to be incompatible with `Mapping[str, Any]`. I have no suspicion as to why, but this isn’t something that should be addressed in this PR.
> 
> ```
> Found 1 new problems in /home/runner/work/rucio/rucio/lib/rucio/daemons/conveyor/submitter.py
>   - 1 errors with message """Argument of type "list[RSESettingsDict] | None" cannot be assigned to parameter "rses" of type "list[Mapping[str, Any]] | None" in function "submitter"""".
> ```

bari12 (approved):

Issue #6278, opened 2023-07-21
register-after-upload doesn't transition existing replicas to Available


rcarpa:
When this option is set to `rucio upload`, this code is executed: 
https://github.com/rucio/rucio/blob/11cec3f66281013808076007c04e6099fb9f570b/lib/rucio/client/uploadclient.py#L290

but, `_register_file` doesn't do anything if the replica exists: 
https://github.com/rucio/rucio/blob/11cec3f66281013808076007c04e6099fb9f570b/lib/rucio/client/uploadclient.py#L408

Even if it's not in the available state. We should transition the replica to available if it's not already in this state.

PR #6315, opened 2023-08-16
Clients: Transition replicas to Available #6278

Kwaizer:
Transitioning the replica to available if it's not already in this state.


bari12:

PR #6306, opened 2023-08-08
Authentication: Migrate to SQLAlchemy 2.0 syntax #5247

dchristidis:
Inspired by #5247, but also fixes #5838.

PR #6319, opened 2023-08-21
Authentication: Migrate to SQLAlchemy 2.0 syntax #5247

### LINKED ISSUE #5838: sqlalchemy deprecation warning in oauthmanager
Motivation
----------

```

test_oauthmanager.py::TestOAuthManager::test_oauthmanager
  /opt/rucio/lib/rucio/core/oidc.py:937: SADeprecationWarning: Object <sqlalchemy.orm.query.Query object at 0x7f546c7fb710> should not be used directly in a SQL statement context, such as passing to methods such as session.execute().  This usage will be disallowed in a future release.  Please use Core select() / update() / delete() etc. with Session.execute() and other statement execution methods.
    for items in session.execute(query).partitions(10):

test_oauthmanager.py::TestOAuthManager::test_oauthmanager
  /opt/rucio/lib/rucio/core/authentication.py:361: SADeprecationWarning: Object <sqlalchemy.orm.query.Query object at 0x7f546c7bed30> should not be used directly in a SQL statement context, such as passing to methods such as session.execute().  This usage will be disallowed in a future release.  Please use Core select() / update() / delete() etc. with Session.execute() and other statement execution methods.
    for items in session.execute(query).partitions(10):

test_oauthmanager.py::TestOAuthManager::test_oauthmanager
  /opt/rucio/lib/rucio/core/oidc.py:1059: SADeprecationWarning: Object <sqlalchemy.orm.query.Query object at 0x7f546c7e0ba8> should not be used directly in a SQL statement context, such as passing to methods such as session.execute().  This usage will be disallowed in a future release.  Please use Core select() / update() / delete() etc. with Session.execute() and other statement execution methods.
    for items in session.execute(query).partitions(10):
```

dchristidis:
Continuation from #6306. Inspired by #5247, but also fixes #5838.

### DIFF
--- a/lib/rucio/core/account.py
+++ b/lib/rucio/core/account.py
@@ -19,7 +19,7 @@ from re import match
 from traceback import format_exc
 from typing import TYPE_CHECKING
 
-from sqlalchemy import and_
+from sqlalchemy import select, and_
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm import exc
 
@@ -64,17 +64,20 @@ def add_account(account, type_, email, *, session: "Session"):
 
 @read_session
 def account_exists(account, *, session: "Session"):
-    """ Checks to see if account exists. This procedure does not check it's status.
+    """ Checks to see if account exists and is active.
 
     :param account: Name of the account.
     :param session: the database session in use.
 
     :returns: True if found, otherwise false.
     """
-
-    query = session.query(models.Account).filter_by(account=account, status=AccountStatus.ACTIVE)
-
-    return True if query.first() else False
+    query = select(
+        models.Account
+    ).where(
+        models.Account.account == account,
+        models.Account.status == AccountStatus.ACTIVE
+    )
+    return session.execute(query).scalar() is not None
 
 
 @read_session
@@ -86,10 +89,13 @@ def get_account(account, *, session: "Session"):
 
     :returns: a dict with all information for the account.
     """
+    query = select(
+        models.Account
+    ).where(
+        models.Account.account == account
+    )
 
-    query = session.query(models.Account).filter_by(account=account)
-
-    result = query.first()
+    result = session.execute(query).scalar()
     if result is None:
         raise exception.AccountNotFound('Account with ID \'%s\' cannot be found' % account)
     return result
@@ -102,9 +108,14 @@ def del_account(account, *, session: "Session"):
     :param account: the account name.
     :param session: the database session in use.
     """
-    query = session.query(models.Account).filter_by(account=account).filter_by(status=AccountStatus.ACTIVE)
+    query = select(
+        models.Account
+    ).where(
+        models.Account.account == account,
+        models.Account.status == AccountStatus.ACTIVE
+    )
     try:
-        account = query.one()
+        account = session.execute(query).scalar_one()
     except exc.NoResultFound:
         raise exception.AccountNotFound('Account with ID \'%s\' cannot be found' % account)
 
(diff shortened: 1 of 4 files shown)


lib/rucio/core/account.py
```diff
@@ -183,22 +220,31 @@ def list_identities(account, *, session: "Session"):
...
-    query = session.query(models.IdentityAccountAssociation, models.Identity)\
-                   .join(models.Identity, and_(models.Identity.identity == models.IdentityAccountAssociation.identity,
-                                               models.Identity.identity_type == models.IdentityAccountAssociation.identity_type))\
-                   .filter(models.IdentityAccountAssociation.account == account)
-    for identity in query:
-        identity_list.append({'type': identity[0].identity_type, 'identity': identity[0].identity, 'email': identity[1].email})
-
-    return identity_list
+    query = select(
+        models.IdentityAccountAssociation.identity_type.label('type'),
```
rcarpa, line 235: didn't know it's possible to do that to rely on row._asdict() later. Nice.

lib/rucio/core/account.py
```diff
@@ -236,9 +288,13 @@ def has_account_attribute(account, key, *, session: "Session"):
 
     :returns: True or False
     """
-    if session.query(models.AccountAttrAssociation.value).filter_by(account=account, key=key).first():
-        return True
-    return False
+    query = select(
+        models.AccountAttrAssociation
```
rcarpa: This changes the behavior in terms that it will return `True` on an attribute with value `None`. I don't know if this problematic or if that was an undesired behavior before.

lib/rucio/core/authentication.py
```diff
@@ -385,23 +411,43 @@ def delete_expired_tokens(total_workers, worker_number, limit=1000, *, session:
     # get expired tokens
     try:
         # delete all expired tokens except tokens which have refresh token that is still valid
-        query = session.query(models.Token.token).filter(and_(models.Token.expired_at <= datetime.datetime.utcnow()))\
-                                                 .filter(or_(models.Token.refresh_expired_at.__eq__(None),
-                                                             models.Token.refresh_expired_at <= datetime.datetime.utcnow()))\
-                                                 .order_by(models.Token.expired_at)
+        query = select(
```
rcarpa, line 414: I would have hoped that there is an easier way to perform a simple deletion :thinking: but it's OK for me to keep it this way

lib/rucio/core/authentication.py
```diff
@@ -497,14 +542,22 @@ def __delete_expired_tokens_account(account, *, session: "Session"):
...
-                    models.Token.account == account)) \
-        .with_for_update(skip_locked=True)
-    tokens = session.execute(stmt_select).scalars().all()
-
-    for t in chunks(tokens, 100):
-        stmt_delete = delete(models.Token) \
-            .where(models.Token.token.in_(t)) \
-            .prefix_with("/*+ INDEX(TOKENS_ACCOUNT_EXPIRED_AT_IDX) */")
-        session.execute(stmt_delete)
+    select_query = select(
```
rcarpa, line 545: Why not just perform a delete query directly ? Is this really so much easier on the database to explicitly lock the rows 100 at a time rather than leaving it to do however it desires ? But this is a comment for another change. No need to bundle it in this "sqlalchemy 2 syntax migration"

rcarpa (dismissed):

lib/rucio/core/account.py
dchristidis: I don’t remember exactly what my thinking was. I might have not realised the change in behaviour. The documentation string suggests that `has_account_attribute()` should only care about the existence of the attribute and not its value, but I will make it as before because this PR is supposed to maintain exact behaviour. Thanks for bringing this up.

lib/rucio/core/authentication.py
dchristidis: Both to this and the previous comment, I assumed that the necessity for two queries (`SELECT` then `DELETE` as opposed to just `DELETE`) was to apply `FOR UPDATE SKIP LOCKED`. I opted for maintaining this implementation, without fully understanding the motivation behind it.

rcarpa: Ah, yes, skip locked should have been the reason. This makes sense. Thanks for explaining

rcarpa (approved):

PR #6329, opened 2023-09-05
Core & Internals: Resolving improper use of NoResultFound #6309

### LINKED ISSUE #6309: Improper use of `NoResultFound`
### Description

We need to review all uses of `NoResultFound` because there’s at least one that is misleading (it will never be raised):

https://github.com/rucio/rucio/blob/9d432930de0a754c0a65ce3ed19e65051326c3b3/lib/rucio/core/did.py#L1649-L1671

### Steps to reproduce

```python
>>> from rucio.common.types import InternalScope
>>> from rucio.core.did import list_content
>>> list(list_content(scope=InternalScope('foo'), name='bar'))
[]
```

### Rucio Version

_No response_

### Additional Information

_No response_

Kwaizer:
Initial idea of this PR was introducing `yielded` flag (like in`lib/rucio/core/replica.py`) which will indicate if any result was yielded at all and raise exception if not. Such implementation wasn't fully successful for all uses in `lib/rucio/core/did.py` due to failed tests so I considered removing unused constraints here for now.

### DIFF
--- a/lib/rucio/core/account.py
+++ b/lib/rucio/core/account.py
@@ -337,10 +337,9 @@ def get_usage_history(rse_id, account, *, session: "Session"):
 
     result = []
     AccountUsageHistory = models.AccountUsageHistory
-    try:
-        query = session.query(AccountUsageHistory).filter_by(rse_id=rse_id, account=account).order_by(AccountUsageHistory.updated_at)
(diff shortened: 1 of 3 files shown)


bari12:

PR #6332, opened 2023-09-15
Core & Internals: respect add_dids permissions on replica creation. Closes #6058

### LINKED ISSUE #6058: rucio upload does not check that the user has permission to add DIDs
### Description

When a user uploads new files to Rucio using the `rucio upload` command, new DIDs are created in the Rucio database for the files. However, this is done using an `add_replicas` call, which adds the DIDs directly to the database table (in `core/replica.py:__bulk_add_new_file_dids`). Because `add_did`/`add_dids` is never called, the permission check for `add_did` is also never called. This is a problem for DUNE as we are using a custom `add_did` permission check to prevent users from adding files that are not present in DUNE's metadata catalogue.

### Steps to reproduce

1. Modify permissions so that the current user is not allowed to add new DIDs (or change the permission code so that the `add_did` and `add_dids` checks always fail).
2. Perform a `rucio upload` to add a new file.
3. Observe that the upload succeeds and a new DID is created.

### Rucio Version

1.29

### Additional Information

_No response_

rcarpa:


bari12:
Closing, followed in #6686
Thanks Radu 🙏

PR #6340, opened 2023-09-28
Pass rse_ids to the permissions module

ivmfnal:
… rse_id(s) to the permissions module

### DIFF
--- a/lib/rucio/api/replica.py
+++ b/lib/rucio/api/replica.py
@@ -22,8 +22,9 @@ from rucio.common import exception
 from rucio.common.constants import SuspiciousAvailability
 from rucio.common.schema import validate_schema
 from rucio.common.types import InternalAccount, InternalScope
-from rucio.common.utils import api_update_return_dict, invert_dict
+from rucio.common.utils import api_update_return_dict
 from rucio.core import replica
+from rucio.core.replica import get_pfn_to_rse
 from rucio.core.rse import get_rse_id, get_rse_name
 from rucio.db.sqla.constants import BadFilesStatus
 from rucio.db.sqla.session import read_session, stream_session, transactional_session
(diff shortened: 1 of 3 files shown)


lib/rucio/api/replica.py
```diff
@@ -85,36 +86,54 @@ def declare_bad_file_replicas(replicas, reason, issuer, vo='def', force=False, *
...
+                                                  kwargs={"rse_id": rse_id},
+                                                  session=session)
+        except:
+            permitted = True
+
+        if not permitted:
+            raise exception.AccessDenied('Account %s can not declare bad replicas in RSE %s'
+                                         % (issuer, rse_id_to_name.get(rse_id, rse_id)))
 
     undeclared = replica.declare_bad_file_replicas(replicas_lst, reason=reason, issuer=issuer, status=BadFilesStatus.BAD,
```
dynamic-entropy, line 135: Hi @ivmfnal 

Now, that the replica_lst is always a dict with lfns and rse_id, we may get rid of dead code that deals with replicas being specified as `pfns` in replica core, and often does not work for other experiments, e.g. __declare_bad_file_replicas for cms when pfns are specified.
My comment makes an assumption that `get_pfn_to_rse` is experiment-independent.

https://github.com/rucio/rucio/blob/6eac5dcc30cb6dac427cbc5070d46c170ae7fedf/lib/rucio/core/replica.py#L273-L288

ivmfnal: this code may be "dead" for CMS, but maybe it is still used by other users.

ivmfnal: The replica_lst is not always a list of dictionaries. If the original list was a list of PFNs, then the list is not modified and passed to the `core.decalre_bad_file_replicas()` as is.

lib/rucio/api/replica.py
```diff
@@ -85,36 +86,54 @@ def declare_bad_file_replicas(replicas, reason, issuer, vo='def', force=False, *
...
-                "scope": scope,
+                "scope": InternalScope(r['scope'], vo=vo),
                 "rse_id": rse_id,
                 "name": r["name"]
             })
+            rse_ids_to_check.add(rse_id)
 
-    rse_id_to_name = invert_dict(rse_map)   # RSE id -> RSE name
+    for rse_id in rse_ids_to_check:
+        try:
```
bari12: Why do we need the try/except block here?

bari12 (commented):
Just a small comment.
Also please add the issue number to the commit title, so the PR can be linked to the issue.

lib/rucio/api/replica.py
ivmfnal: to handle exceptions caused by potentially buggy implementation of the permission function.

Perhaps it would be better to assume that if the permissions function raises a exception, the answer is "no the client does not have the permission"

dynamic-entropy: @bari12 Just to take the opportunity here. In general, we shall modify the interface of the permission module function to be able to raise exceptions or at least pass a message that can be relayed to the client, instead of "`user` does not have permission for `function`". 
As our permission functions become more complex, in certain situations, we would like to return a specific message on why the permission was denied. What do you think?

bari12: @dynamic-entropy I think that makes sense, but we would have to adapt it coherently, also having the policy modules in mind.
@ivmfnal Ok, it's just a bit "out-of-sync" since we do not do this anywhere else with permissions, I am not against it, but the way it is done it might rather hide things going wrong in the permission module.

ivmfnal: I think I should revert to the original behavior - not try to catch exceptions raised by the permissions function and let them propagate upwards so that they can be handled as they are - with the type and whatever messages are attached.

Issue #6337, opened 2023-09-25
Adjust suspicious replica recoverer creation of rules.


ChristophAmes:
Currently, ten rules are created for one day for replicas that are declared suspicious only once. This allows for those files to be handled by the recoverer during its next run. This lead to an edge case where a few thousand rules were created at once, so some changes will be applied to prevent this from happening again.

PR #6349, opened 2023-10-12
Recovery: Adjust suspicious replica recoverer creation of rules. #6337

ChristophAmes:
Currently, ten rules are created for one day for replicas that are declared suspicious only once. This allows for those files to be handled by the recoverer during its next run. This lead to an edge case where a few thousand rules were created at once, so some changes will be applied to prevent this from happening again.

### DIFF
--- a/etc/suspicious_replica_recoverer.json
+++ b/etc/suspicious_replica_recoverer.json
@@ -9,4 +9,9 @@
         "datatype": ["RAW"],
         "scope": []
     }
+    {
+        "action": "dry run",
+        "datatype": [],
+        "scope": ["mc.*"]
+    }
 ]
--- a/lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
+++ b/lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
@@ -56,7 +56,32 @@ GRACEFUL_STOP = threading.Event()
 DAEMON_NAME = 'suspicious-replica-recoverer'
 
 
-def declare_suspicious_replicas_bad(once: bool = False, younger_than: int = 3, nattempts: int = 10, vos: Optional[list[str]] = None, limit_suspicious_files_on_rse: int = 5, json_file_name: str = "/opt/rucio/etc/suspicious_replica_recoverer.json", sleep_time: int = 3600, active_mode: bool = False) -> None:
+def check_suspicious_policy(policy: dict[str, str], file_metadata_datatype: str, file_metadata_scope: str) -> str:
+    match_scope = False
+    match_datatype = False
+    action = ""
+
+    if not policy.get("scope", []):
+        match_scope = True
+    for scope in policy.get("scope", []):
+        if re.match(scope, file_metadata_scope):
+            match_scope = True
+            break
+
+    if not policy.get("datatype", []):
+        match_datatype = True
+    for datatype in policy.get("datatype", []):
+        if re.match(datatype, file_metadata_datatype):
+            match_datatype = True
+            break
+
+    if match_scope and match_datatype:
+        action = policy["action"]
+
+    return action
+
+
+def declare_suspicious_replicas_bad(once: bool = False, younger_than: int = 5, nattempts: int = 5, vos: Optional[list[str]] = None, limit_suspicious_files_on_rse: int = 5, json_file_name: str = "/opt/rucio/etc/suspicious_replica_recoverer.json", sleep_time: int = 3600, active_mode: bool = False) -> None:
     """
     Main loop to check for available replicas which are labeled as suspicious.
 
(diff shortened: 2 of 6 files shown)


etc/suspicious_replica_recoverer.json
```diff
@@ -9,4 +9,9 @@
         "datatype": ["RAW"],
         "scope": []
     }
+    {
+        "action": "dry run",
+        "datatype": [],
+        "scope": ["MC"]
```
cserf: Should be : 
```
       "scope": ["mc.*"]
```

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
```diff
@@ -282,20 +282,80 @@ def run_once(heartbeat_handler: Any, younger_than: int, nattempts: int, vos: Opt
...
+            if not replicas_nattempts_1[vo][rse_key]:
+                # This is needed for testing purposes.
+                continue
+            files_to_be_declared_bad_nattempts_1 = []
             dids_nattempts_1 = []
-            for replica_values in replicas_nattempts_1[vo][rse_key].values():
-                dids = {'scope': replica_values['scope'], 'name': replica_values['name'], 'rse': rse_key}
-                dids_nattempts_1.append(dids)
+            # Get the rse_id by going to one of the suspicious replicas from that RSE and reading it from there
+            rse_id = list(replicas_nattempts_1[vo][rse_key].values())[0]['rse_id']
```
cserf, line 320: Why do you need rse_id here ?

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
```diff
@@ -282,20 +282,80 @@ def run_once(heartbeat_handler: Any, younger_than: int, nattempts: int, vos: Opt
...
+            for replica_key in replicas_nattempts_1[vo][rse_key].keys():
+                from_auditor = False
+                file_scope = replicas_nattempts_1[vo][rse_key][replica_key]["scope"]
+                file_name = replicas_nattempts_1[vo][rse_key][replica_key]["name"]
+                file_metadata = get_metadata(file_scope, file_name)
```
cserf, line 325: Why not defining `rse_id = replicas_nattempts_1[vo][rse_key][replica_key]["rse_id"]` here ?

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
```diff
@@ -282,20 +282,80 @@ def run_once(heartbeat_handler: Any, younger_than: int, nattempts: int, vos: Opt
...
+                replicas_nattempts_1[vo][rse_key][replica_key]["datatype"] = str(file_metadata["datatype"])
+
+                # Auditor
+                suspicious_reason = get_suspicious_reason(replicas_nattempts_1[vo][rse_key][replica_key]["rse_id"], file_scope, file_name, nattempts)
```
cserf, line 329: The `get_suspicious_reason` returns something even if the suspicious replica was recovered meanwhile. Some protection needs to be added (check if there's a bad replica entry after the most recent suspicious entry)

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
```diff
@@ -282,20 +282,80 @@ def run_once(heartbeat_handler: Any, younger_than: int, nattempts: int, vos: Opt
...
+                for reason in suspicious_reason:
+                    if "auditor" in reason["reason"].lower():
+                        from_auditor = True
+                        files_to_be_declared_bad_nattempts_1.append(recoverable_replicas[vo][rse_key][replica_key])
+                        break
+
+                # Bad
+                if not from_auditor:
+                    if (file_name.startswith("log.")) or (file_name.startswith("user")):
```
cserf, line 338: Will only work for ATLAS. In the future, this should be moved as a policy

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
```diff
@@ -335,93 +395,110 @@ def run_once(heartbeat_handler: Any, younger_than: int, nattempts: int, vos: Opt
...
-                        action = ""
-                        if file_metadata_datatype:
-                            # Some files don't have a datatype. They should be ignored.
-                            for policy in json_data:
-                                match_scope = False
-                                match_datatype = False
+                suspicious_reason = get_suspicious_reason(recoverable_replicas[vo][rse_key][replica_key]["rse_id"],
+                                                          file_scope,
+                                                          file_name,
+                                                          nattempts)
```
cserf, line 431: Same comment as above

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
```diff
@@ -335,93 +395,110 @@ def run_once(heartbeat_handler: Any, younger_than: int, nattempts: int, vos: Opt
...
-                                for scope in policy.get("scope", []):
-                                    if re.match(scope, file_metadata_scope):
+                if not from_auditor:
+                    if recoverable_replicas[vo][rse_key][replica_key]['available_elsewhere'] is True:
+                        # Replicas with other copies on at least one other RSE can safely be labeled as bad
+                        files_to_be_declared_bad.append(recoverable_replicas[vo][rse_key][replica_key])
+                        # Remove replica from dictionary
+                        del recoverable_replicas[vo][rse_key][replica_key]
+                    elif recoverable_replicas[vo][rse_key][replica_key]['available_elsewhere'] is False:
+                        if (file_name.startswith("log.")) or (file_name.startswith("user")):
```
cserf, line 446: Same comment as above

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
```diff
@@ -335,93 +395,110 @@ def run_once(heartbeat_handler: Any, younger_than: int, nattempts: int, vos: Opt
...
+                            if file_metadata["datatype"] is None:  # "None" type has no function "split()"
+                                files_to_be_ignored.append(recoverable_replicas[vo][rse_key][replica_key])
+                                logger(logging.WARNING, "RSE: %s, replica name %s, surl %s: Replica does not have a data type associated with it. No action will be taken.",
+                                       rse_key, replica_key, recoverable_replicas[vo][rse_key][replica_key]['surl'])
+                                continue
+
+                            file_metadata_datatype = str(file_metadata["datatype"])
+                            file_metadata_scope = str(file_metadata["scope"])
+                            action = ""
+                            if file_metadata_datatype:
```
cserf, line 461: The following lines are code duplication from line 325 and after. It can be factorized in a common method

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
```diff
@@ -335,93 +395,110 @@ def run_once(heartbeat_handler: Any, younger_than: int, nattempts: int, vos: Opt
...
+                                    # Monitoring purposes: Will look like a file has been declared bad, even though no
+                                    # actions will be taken.
+                                    files_dry_run_monitoring.append(recoverable_replicas[vo][rse_key][replica_key])
+                                elif action == "ignore":
+                                    files_to_be_ignored.append(recoverable_replicas[vo][rse_key][replica_key])
+                                elif action == "declare bad":
+                                    suspicious_reason = get_suspicious_reason(recoverable_replicas[vo][rse_key][replica_key]["rse_id"],
+                                                                              file_scope,
+                                                                              file_name,
+                                                                              nattempts)
```
cserf, line 483: Same comment as above for `get_suspicious_reason`

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
ChristophAmes: I think it was previously used for more; now it's just used for the log message.

lib/rucio/daemons/replicarecoverer/suspicious_replica_recoverer.py
ChristophAmes: Since all the replicas in the loop are from the same RSE, the ID is same, so I defined it outside of the loop.

cserf (approved):

Issue #6339, opened 2023-09-28
declare_bad_file_replicas does not pass rse_id to the policy module


ivmfnal:
### Description

Currently, the `api.declare_bad_file_replicas()` does not pass `rse_id` to the `permission.has_permission()` function:
   
   https://github.com/ivmfnal/rucio/blob/fc3973a83b9d658fba5ef9f645f5816a594d4586/lib/rucio/api/replica.py#L88C1-L92C1

CMS needs rse_id to be passed to the permissions module to be able to determine whether the user has permissions to declare bad replicas in the RSE

### Steps to reproduce

I think this is obvious from the code.
Too difficult to reproduce.

PR #6350, opened 2023-10-15
Policies: pass RSE ids to has_permissionfrom api.declare_bad_file_replicas() #6339

### LINKED ISSUE #587: Site admins cannot declare bad replicas

ivmfnal:

### DIFF
--- a/lib/rucio/api/replica.py
+++ b/lib/rucio/api/replica.py
@@ -14,7 +14,6 @@
 # limitations under the License.
 
 import datetime
-import uuid
 from typing import TYPE_CHECKING
 
 from rucio.api import permission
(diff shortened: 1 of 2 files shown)


lib/rucio/api/replica.py
```diff
@@ -109,14 +110,27 @@ def declare_bad_file_replicas(replicas, reason, issuer, vo='def', force=False, *
...
 
-    undeclared = replica.declare_bad_file_replicas(replicas_lst, reason=reason, issuer=issuer, status=BadFilesStatus.BAD,
+    for rse_id in rse_ids_to_check:
+        if not permission.has_permission(issuer=issuer, vo=vo, action='declare_bad_file_replicas',
+                                         kwargs={"rse_id": rse_id},
+                                         session=session):
+            raise exception.AccessDenied('Account %s can not declare bad replicas in RSE %s' %
+                                         (issuer, rse_id_to_name.get(rse_id, rse_id)))
+
+    # if not permission.has_permission(issuer=issuer, vo=vo, action='declare_bad_file_replicas', kwargs={}, session=session):
```
bari12: I think the two comments can be removed.

bari12 (commented):
LGTM
Just two small comments.
In the commit message, reference the issue as #6339 (Not just "6339") - this way GH will link them.

tests/test_bad_replica.py
```diff
@@ -249,11 +249,6 @@ def test_client_add_list_bad_replicas(rse_factory, replica_client, did_client):
                     nbbadrep += 1
     assert len(replicas) == nbbadrep
 
-    # InvalidType is raised if list_rep contains a mixture of replicas and PFNs
```
bari12, line 252: On a second look, why was this test removed? It still seems to be raised correctly in the API?

ivmfnal: If you are talking about test_bad_replica.py lines 253-256, somehow I was not able to make the `with pytest.raises(InvalidType)` catch and mask the exception. So I decided to simply remove that part of the test.

bari12: This should be fixed in the code then. This is still a valid test and points to a problem in the code.

bari12 (approved):

PR #6362, opened 2023-11-03
Consistency checks: Make auditor work on objectstores #2263

### LINKED ISSUE #2263: Add support for object storages in Auditor
Motivation
----------

Currently, the Auditor expects to be able to list directory entries in order to identify the latest dump. This is not possible on object storages.

Modification
------------

Investigate and implement an alternative, such as:
 1. Try the last _n_ days, based on the pattern
 2. Use another file as an index
 3. Use a static location and obtain the date from its metadata

jamesp-epcc:
This PR allows the auditor daemon to download dump files from objectstores. It does this by trying the last 30 dates until it finds a dump file, avoiding having to perform a directory listing as is done on other storage types. The advantage of this method is that it allows site admins to manage the dump files in the same way as for other storage types, while avoiding introducing dependencies on objectstore client libraries in the auditor. Tested on AWS S3, but there is nothing specific to the objectstore type and this method should work on any objectstore that uses URL signing.

I also modified `download_rse_dump` slightly so that it uses `None` to indicate that the latest dump should be downloaded rather than the string `'latest'`. This is because Pyright was giving errors due to thinking the `strftime` method was going to be called on a string (even though that was never possible).

### DIFF
--- a/lib/rucio/daemons/auditor/srmdumps.py
+++ b/lib/rucio/daemons/auditor/srmdumps.py
@@ -26,12 +26,16 @@ import gfal2
 import requests
 
 from rucio.common.config import get_config_dirs
-from rucio.common.dumper import DUMPS_CACHE_DIR, ddmendpoint_url, gfal_download_to_file, http_download_to_file, temp_file
+from rucio.common.dumper import DUMPS_CACHE_DIR, HTTPDownloadFailed, ddmendpoint_url, gfal_download_to_file, http_download_to_file, temp_file
+from rucio.core.credential import get_signed_url
+from rucio.core.rse import get_rse_id, list_rse_attributes
 
 CHUNK_SIZE = 10485760
 
 __DUMPERCONFIGDIRS = (os.path.join(confdir, 'auditor') for confdir in get_config_dirs())
 __DUMPERCONFIGDIRS = list(filter(os.path.exists, __DUMPERCONFIGDIRS))
+
+OBJECTSTORE_NUM_TRIES = 30
 
 
 class Parser(ConfigParser.RawConfigParser):
@@ -211,7 +215,7 @@ def parse_configuration(conf_dirs=__DUMPERCONFIGDIRS):
     return configuration
 
 
-def download_rse_dump(rse, configuration, date='latest', destdir=DUMPS_CACHE_DIR):
+def download_rse_dump(rse, configuration, date=None, destdir=DUMPS_CACHE_DIR):
     '''
     Downloads the dump for the given ddmendpoint. If this endpoint does not
     follow the standarized method to publish the dumps it should have an
@@ -221,7 +225,7 @@ def download_rse_dump(rse, configuration, date='latest', destdir=DUMPS_CACHE_DIR
 
     `configuration` is a RawConfigParser subclass.
 
-    `date` is a datetime instance with the date of the desired dump or 'latest'
+    `date` is a datetime instance with the date of the desired dump or None
     to download the lastest available dump.
 
     `destdir` is the directory where the dump will be saved (the final component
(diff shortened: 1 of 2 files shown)


jamesp-epcc:
I am not sure why this failed the end to end integration tests. All the checks passed on my repo and the failure is in the conveyor, which I haven't touched, so I'm wondering if this is just a random transient failure.

faluchet:
Hi,

while I agree with the general implementation idea, I'd like a few changes to be implemented:

- Necessary changes: the existence of the `sign_url` attribute alone should not mean that an RSE is an object store, nor the vice-versa. There can very well be object store RSEs that does not require url signature and function without this attribute (in ATLAS there are) . On the other hand, I'm afraid that setting a dummy value for such RSEs would also be problematic, as it may break the logic in other parts of the code. For one, `get_signed_url` called in #L271 would need to be adapted. 

Instead of chaining changes elsewhere, I think is better to keep the meanings of `sign_url` and "the RSE is an object store" decoupled, and make the if in #L248 depend on something else. I have two alternatives: either we use a key-value map from the already available configuration argument (see #L220 in the function signature) , for example using the `rse` key; either we use a new attribute `is_object_store` (I'm usually reluctant to the proliferation of custom attributes, but this feels like a good fit). Opinion?


- Minor changes: the use of the `success` variable is redundant, and I am not fond of the way it switches to True before even making a download attempt, only to be reset to False upon failure. We can just assign 0 to `tries` to terminate the loop upon a successful download, while decreasing its value and the date by one in case of an error on the current date/attempt.

jamesp-epcc:
Hi, thanks for reviewing. I'm happy to make those changes. If no-one objects I will use a new attribute to determine whether the RSE is an object store or not instead of using the `sign_url` attribute.

faluchet (approved):

lib/rucio/daemons/auditor/srmdumps.py
```diff
@@ -234,29 +238,63 @@ def download_rse_dump(rse, configuration, date='latest', destdir=DUMPS_CACHE_DIR
...
+                date.strftime('%d-%m-%Y'),
+                hashlib.sha1(url.encode()).hexdigest()
+            )
+            filename = re.sub(r'\W', '-', filename)
+            path = os.path.join(destdir, filename)
+            if not os.path.exists(path):
+                logger.debug('Trying to download: "%s"', url)
+                if 'sign_url' in rse_attr:
+                    url = get_signed_url(rse_id, rse_attr['sign_url'], 'read', url)
+                try:
```
faluchet, line 268: On a second look, another note on #L270-L276

The call to download(url, f) may throw different exceptions, i.e. gfal2.Error in https://github.com/rucio/rucio/blob/0f674128c48eef880eb9362bb818de3fbb6a7154/lib/rucio/common/dumper/__init__.py#L294 

Also, binary=True in L271 won't always work
```
  File "/usr/local/lib/python3.6/site-packages/rucio/common/dumper/__init__.py", line 297, in gfal_download_to_file
    file_.write(chunk)
TypeError: a bytes-like object is required, not 'str'
```

faluchet (changes requested):

jamesp-epcc:
Well spotted - I initially thought that GFAL (as opposed to HTTP) couldn't be called for objectstores, but in fact it can, so I've amended the code to catch the GFAL exception as well.

I am not sure if my solution to the binary vs. text issue is the right one. When I download a log file from S3 via requests, the encoding is not defined and `iter_content` returns bytes rather than strings. However other parts of Rucio call `http_download_to_file` with a file open in text mode and this apparently works. I've added some code to handle the case where the encoding is not set, and to tell it explicitly to decode, but I am not sure if this is the best solution.

jamesp-epcc:
OK, this is breaking the tests because (I think) the mock response object doesn't have all the features of the real one. I will rethink how to solve this.

jamesp-epcc:
I've tried to make the download function work for every case without having to modify unrelated code such as the tests, but I'm open to suggestions for better ways to do that. The log downloader tells `http_download_to_file` to always decode the content to a string, which fixes the S3 case, but by default the old behaviour remains.

faluchet:
Hi James, apologies for the later reply.

I appears that the ambiguity on the bytes vs. text encoding applies to any of the `[protocol(url)]['download']` calls, and it is not specific to the http case indeed. I reproduced it with the `gfal_download`, with the un-handled exception raised being `UnicodeDecodeError` , but this may better be tracked on a separate issue perhaps. However, I fail to see how the default behaviour for `http_download_to_file` will not change with  this implementation.

Why the need for the `force_decode` argument?  Can you try without it, but leaving the `if response.encoding is None` check?

Other than that, I'm OK with the PR.

faluchet:
Oh, I get it. I think the simpler way out is to issue a new `requests.get()` with something like:

```
try:
    if session is None:
        response = requests.get(url, stream=True)
    else:
        response = session.get(url)
    
    if response.status_code != 200:
        logging.error(
            'Retrieving %s returned %d status code',
            url,
            response.status_code,
        )
        raise HTTPDownloadFailed('Error downloading ' + url, response.status_code)

        for chunk in response.iter_content(CHUNK_SIZE):
            file_.write(chunk)

except UnicodeDecodeError:

    if session is None:
        response = requests.get(url, stream=True)
    else:
        response = session.get(url)
    
    if response.status_code != 200:
        logging.error(
            'Retrieving %s returned %d status code',
            url,
            response.status_code,
        )
        raise HTTPDownloadFailed('Error downloading ' + url, response.status_code)

        if response.encoding is None:
            response.encoding = 'utf-8'

        for chunk in response.iter_content(CHUNK_SIZE,decode_unicode=True):
            file_.write(chunk)
```

which I'm not particularly fond of. What troubles me is that we have a nice wrapper in`http_download` https://github.com/jamesp-epcc/rucio/blob/9605f0ebdfbb4935c57acf93a5a0072286982319/lib/rucio/common/dumper/__init__.py#L276 where logically the exception would be handled nicely and would allow to possibly pass the `decode_unicode=True` argument, but no function seems to make use of it.

jamesp-epcc:
Thanks for the review and sorry it's taken me a while to get back to this. I'm happy to do it the way you suggest in your most recent comment - I think it should work for every case, and it would avoid the need for the `force_decode` argument, containing the solution within the download function. It could be made a bit shorter/cleaner by factoring most of the code into a function that gets called twice, I think.

faluchet (approved):

faluchet:
Hi James (I missed myself the notification on this),

it looks better now, thanks for the contributions, let's ship it!

Issue #6292, opened 2023-08-01
SUBSCRIPTIONS_HISTORY integrity error


dchristidis:
### Description

```
File "/usr/local/lib/python3.9/site-packages/rucio/daemons/common.py", line 216, in _generator
    result = run_once_fnc(heartbeat_handler=heartbeat_handler, activity=activity)
  File "/usr/local/lib/python3.9/site-packages/rucio/daemons/transmogrifier/transmogrifier.py", line 699, in run_once
    update_subscription(
  File "/usr/local/lib/python3.9/site-packages/rucio/db/sqla/session.py", line 437, in new_funct
    raise DatabaseException(str(error))
```
```
Database exception.
Details: (cx_Oracle.IntegrityError) ORA-00001: unique constraint (ATLAS_RUCIO.SUBSCRIPTIONS_HISTORY_PK) violated
[SQL: INSERT INTO atlas_rucio.subscriptions_history (id, name, filter, replication_rules, policyid, state, last_processed, account, lifetime, comments, retroactive, expired_at, created_at, updated_at) VALUES (:id, :name, :filter, :replication_rules, :policyid, :state, :last_processed, :account, :lifetime, :comments, :retroactive, :expired_at, :created_at, :updated_at)]
```

In the Transmogrifier, the partitioning is done based on DIDs (`rucio.core.did.list_new_dids()`). However, it could be that two different DIDs, handled by two different threads, match the same subscription. It would appear that there’s a part of code that is not thread-safe:

https://github.com/rucio/rucio/blob/43aabc439fa89cc6c3669dc791ea802768e4f2d8/lib/rucio/daemons/transmogrifier/transmogrifier.py#L698-L703

The first thread to call `update_subscription()` has already inserted a row in `subscriptions_history`, so the second one, if it does the same at the exact same second, it will lead to the error.

This error appeared with the RC release, but it is unknown if it was introduced in it.

### Steps to reproduce

This error appears consistently under the running ATLAS instance. The conditions are maybe too complex to reproduce outside of it (e.g. subscription configuration, number of processes and threads, new DIDs).

### Rucio Version

32.0.0rc1

cserf:
There's something fishy
```
grep -c SUBSCRIPTIONS_HISTORY_PK lib/rucio/db/sqla/models.py 
0
```
The Primary Key for the `subscription_history` table is 
```python
class SubscriptionHistory(BASE, ModelBase):
    """Represents a subscription history"""
[...]

PR #6364, opened 2023-11-09
Subscription history updates throw unknown database errors : Closes #6351

### LINKED ISSUE #6351: Subscription history updates throw unknown database errors
### Description

- [ ] Keep history config expects a boolean value but wrongly uses config_get instead of config_get_bool, https://github.com/rucio/rucio/blob/6a5ccae83bb0abed78998fb04a20dabb1fdf2ee4/lib/rucio/core/subscription.py#L149C24-L149C34

- [ ] A subscription update results in a database exception when keep_history is set
```bash
rucio-admin subscription update --lifetime 0 .....

Database exception.
Details: An unknown Database Exception has occurred.
Rucio exited with an unexpected/unknown error, please provide the traceback below to the developers.
  File "/cvmfs/cms.cern.ch/rucio/x86_64/rhel9/py3/current/bin/rucio-admin", line 97, in new_funct
  File "/cvmfs/cms.cern.ch/rucio/x86_64/rhel9/py3/current/bin/rucio-admin", line 947, in update_subscription
    client.update_subscription(name=args.name, account=account, filter_=json.loads(args.filter), replication_rules=json.loads(args.replication_rules),
  File "/cvmfs/cms.cern.ch/rucio/x86_64/rhel9/py3/current/lib/python3.9/site-packages/rucio/client/subscriptionclient.py", line 144, in update_subscription
    raise exc_cls(exc_msg)
rucio.common.exception.DatabaseException: Database exception.
Details: An unknown Database Exception has occurred.
```

- [ ] The config does not immediately update within the servers and instead the cache value is used for the first attempt after the config change. So, your command will work or not (based on config change) n number of times where n=number of rucio servers.

### Steps to reproduce

1. Set the keep_history option under the subscriptions section of the config. 
` rucio-admin config set --section subscriptions --option keep_history --value True`

2. Try updating any subscription
```python3
client.update_subscription(
     name='MyTestSubscription',
     account = 'rchauhan',
     replication_rules = [{
         'rse_expression':'T2_CH_CERN',
         'copies':1,
         'lifetime': 864000,
         'comment':'test subscription rule',
         'activity': "User Subscriptions"
         }],
     comments = 'Test subscription updated oct 16',
     filter_= {
         'pattern':'txt',
         'did_type':'CONTAINER',
         'scope':['cms']
         },
     lifetime=30
     )
```

### Rucio Version

_No response_

### Additional Information

_No response_

cserf:
This patch tries to address the problem reported in #6351 (eventhough I was not able to reproduce the bug) : 
-  `config_get_bool` is now used to read `keep_history` option
-  Fix a bug when a new subscription is added (subscription_id set to null)
- Introduce a new exception in case there's a problem updating the history

### DIFF
--- a/lib/rucio/core/subscription.py
+++ b/lib/rucio/core/subscription.py
@@ -99,20 +99,21 @@ def add_subscription(name: str,
                                            lifetime=date_lifetime,
                                            retroactive=retroactive,
                                            policyid=priority, comments=comments)
-    if keep_history:
-        subscription_history = SubscriptionHistory(id=new_subscription.id,
-                                                   name=new_subscription.name,
-                                                   filter=new_subscription.filter,
-                                                   account=new_subscription.account,
-                                                   replication_rules=new_subscription.replication_rules,
-                                                   state=new_subscription.state,
-                                                   lifetime=new_subscription.lifetime,
-                                                   retroactive=new_subscription.retroactive,
-                                                   policyid=new_subscription.policyid,
-                                                   comments=new_subscription.comments)
     try:
         new_subscription.save(session=session)
         if keep_history:
+            subscription_history = SubscriptionHistory(id=new_subscription.id,
+                                                       name=new_subscription.name,
+                                                       filter=new_subscription.filter,
+                                                       account=new_subscription.account,
+                                                       replication_rules=new_subscription.replication_rules,
+                                                       state=new_subscription.state,
+                                                       lifetime=new_subscription.lifetime,
+                                                       retroactive=new_subscription.retroactive,
+                                                       policyid=new_subscription.policyid,
+                                                       created_at=new_subscription.created_at,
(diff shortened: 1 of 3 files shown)


lib/rucio/core/subscription.py
```diff
@@ -78,7 +79,7 @@ def add_subscription(name: str,
     :returns:                  The subscriptionid
     """
     try:
-        keep_history = config_get('subscriptions', 'keep_history')
+        keep_history = config_get_bool('subscriptions', 'keep_history')
```
rcarpa, line 82: All the try/except block can be substituted with 
```
config_get_bool('subscriptions', 'keep_history', raise_exception=False, default=False)
```

bari12:
This needs to be fixed differently. Discussed with @cserf in #6292

Issue #6183, opened 2023-04-17
Design and implement a token cache module


dchristidis:
### Description

With the move from X509 proxy certificates to down-scoped audience-restricted tokens, it is of great importance that they are reused as much as possible. This means that the components that can request new tokens must also cache them.

These components are:

* Reaper, Dark Reaper for deletion,
* Conveyor Submitter for third-party-copy transfers, and
* Servers for user upload and download (this is a longer-term goal).

It makes sense that the cache module is shared among these components. The backend is most likely to be Memcached (an in-memory key-value store already used in Rucio). If so, then the values will be the tokens themselves and what remains is to design an abstraction for the keys. This is likely to be some serialisation of the following fields:

* Operation (e.g. read, stage, write, delete). This itself will need to be an abstraction around the scopes of the supported token profiles (initially only WLCG).
* RSE. Eventually this will translate to a specific audience (i.e. hostname), but RSEs are much, much more convenient to handle. One potential issue that needs to be considered are RSEs that use different hostnames for different protocols.
* Path. Most likely, this will be relative to the RSE’s prefix (note again that this could differ across different protocols). The choice of the path depends on the configuration (from RSE-level down to file-level), but the cache should not be aware of this.

One other potential issue that has not been investigated yet is the expiration time of the entries.

PR #6380, opened 2023-11-20
Auth: Introduce a token cache #6183

dchristidis:
We’ve decided to place this code in `oidc.py` for now as that’s where all other token-related functions exist. It should be expected that some restructuring will happen in the future.


dchristidis:
Closes #6183.

(Surprisingly, I cannot seem to link this PR to the issue.)

bari12 (approved):

Issue #6187, opened 2023-04-18
Design and implement a token request module


dchristidis:
### Description

Issue #6183 describes how to cache down-scoped audience-restricted tokens. This issue describes how to produce them in the first place.

The components which will request tokens are identical, but the interface and level of abstraction around different identity providers (e.g. IAM, CILogon) and token profiles (e.g. WLCG, EGI) remain to be investigated.

PR #6389, opened 2023-11-27
Auth: Introduce a new token request procedure #6187 

dchristidis:
I made all attempts to ensure that the existing token implementation is not affect. To that end, I decided to reuse the configuration and also mimicked the code style (e.g. global variables).

This PR is downsized compare to the original goals described in the issue. Instead of having a high-level interface which can handle multiple plugins, it introduces a mid-level layer which can handle the OAuth flow, but does not decide itself the content of the claims. This should be considered and intermediate step that offers the necessary support for the Data Challenge 2024 and allows us to gain some experience.

### DIFF
--- a/lib/rucio/core/oidc.py
+++ b/lib/rucio/core/oidc.py
@@ -188,10 +188,6 @@ def __initialize_oidc_clients() -> None:
         pass
 
 
-# try loading OIDC clients uppon module import
-__initialize_oidc_clients()
-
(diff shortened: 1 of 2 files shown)


dchristidis:
Again, I’m unsure why I cannot find link the issue from the Development pane, but this closes #6187.

bari12 (approved):

Issue #6316, opened 2023-08-17
Capital `I` in `Id` in REST API for Subscriptions


ThePhisch:
### Description

See https://github.com/rucio/rucio/blob/c4f4eb9902fe0be6ff52d22357f763769c8e9860/lib/rucio/web/rest/flaskapi/v1/subscriptions.py#L625

there is a capital `I` in `Id` which seems unintentional.

### Steps to reproduce

attempt a `GET` to `https://rucio.cern.ch/subscriptions/Id/{subscription_id}` and it will work, `GET` from `https://rucio.cern.ch/subscriptions/id/{subscription_id}` (note the lower-case `i` in `id`) will not.

### Rucio Version

v32

mohitsatr:
Hi @ThePhisch , are you working on this one.

ThePhisch:
> Hi @ThePhisch , are you working on this one.

no, you can go ahead and fix it

mohitsatr:
Sorry for not being cleared in pre-edit version
I was asking whether `I`  in `ID` is just a typo and need to be changed to `i` or do we want it to return response no matter Capitalised `https://rucio.cern.ch/subscriptions/Id/{subscription_id}` or lowercased `https://rucio.cern.ch/subscriptions/id/{subscription_id}`

For example ,  `Rules`  in Below Url is also capitalised .
https://github.com/rucio/rucio/blob/c4f4eb9902fe0be6ff52d22357f763769c8e9860/lib/rucio/web/rest/flaskapi/v1/subscriptions.py#L625-L628

/

voetberg:
`subscriptions` has the only endpoints with a capitalized urls, so that's almost certainly a mistake. 

As for handling the error, case-insensitive urls are not recommended by w3c, so I think we should follow that follow that convention. (We could set up redirects, but I think that's way more trouble than it's worth).

PR #6393, opened 2023-11-28
REST & API: uniform case for urls in flask subscription endpoint Fix: #6316

voetberg:
Addressing #6316 

As stated in the comment there - this is almost certainly a minor mistake. Alt solution would be setting up a automatic redirect for variants of urls with mixed case - which would be a larger undertaking and a separate issue.

### DIFF
--- a/lib/rucio/web/rest/flaskapi/v1/subscriptions.py
+++ b/lib/rucio/web/rest/flaskapi/v1/subscriptions.py
@@ -618,17 +618,25 @@ def blueprint():
     bp = AuthenticatedBlueprint('subscriptions', __name__, url_prefix='/subscriptions')
 
     subscription_id_view = SubscriptionId.as_view('subscription_id')
-    bp.add_url_rule('/Id/<subscription_id>', view_func=subscription_id_view, methods=['get', ])
+    bp.add_url_rule('/id/<subscription_id>', view_func=subscription_id_view, methods=['get', ])
     states_view = States.as_view('states')
-    bp.add_url_rule('/<account>/<name>/Rules/States', view_func=states_view, methods=['get', ])
-    bp.add_url_rule('/<account>/Rules/States', view_func=states_view, methods=['get', ])
+    bp.add_url_rule('/<account>/<name>/rules/states', view_func=states_view, methods=['get', ])
+    bp.add_url_rule('/<account>/rules/states', view_func=states_view, methods=['get', ])
     rules_view = Rules.as_view('rules')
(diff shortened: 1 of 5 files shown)


bari12 (changes requested):
I think this will break backwards compatibility. Because you might still have old clients (pre your change) going against the upper-case endpoints. And in this case, the server will reply a 404. 

We probably have to add rules for both, the lower-case endpoint and the previous uppercase endpoint and take it out once the `33` clients fall out of the legacy window. (Which will be the Rucio `38` release)

lib/rucio/web/rest/flaskapi/v1/subscriptions.py
```diff
@@ -622,17 +622,23 @@ def blueprint():
...
-    bp.add_url_rule('/<account>/<name>/Rules', view_func=rules_view, methods=['get', ])
+    bp.add_url_rule('/<account>/<name>/rules', view_func=rules_view, methods=['get', ])
     subscription_view = Subscription.as_view('subscription')
     bp.add_url_rule('/<account>/<name>', view_func=subscription_view, methods=['get', 'post', 'put'])
     bp.add_url_rule('/<account>', view_func=subscription_view, methods=['get', ])
     bp.add_url_rule('/', view_func=subscription_view, methods=['get', ])
     subscription_name_view = SubscriptionName.as_view('subscription_name')
+    bp.add_url_rule('/name/<name>', view_func=subscription_name_view, methods=['get', ])
+
+    # Legacy url support
```
bari12, line 634: Please add a comment:
`TODO: Take out with Rucio 38` :-)

lib/rucio/web/rest/flaskapi/v1/subscriptions.py
```diff
@@ -622,17 +622,24 @@ def blueprint():
...
+    # TODO: Take out with Rucio 38
+    bp.add_url_rule('/Id/<subscription_id>', view_func=subscription_id_view, methods=['get', ])
+    bp.add_url_rule('/<account>/<name>/Rules/States', view_func=states_view, methods=['get', ])
+    bp.add_url_rule('/<account>/<name>/Rules', view_func=rules_view, methods=['get', ])
```
bari12, line 639: I think one legacy rule is missing
`/<account>/Rules/States`

bari12 (changes requested):
One last change 😄

voetberg:
Test failure due to infamous download_pfns - otherwise everything in the clear :o

bari12:
Please rebase and let's run the tests again.

bari12 (approved):

PR #6403, opened 2023-12-04
Transfers: Add FailureRate SourceRankingStrategy

JChenster:
We add a SourceRankingStrategy that makes source selection dynamic with respect to failure rate. We look up files done/failed in the past hour in the TransferStats schema. We assign RSEs with lower failure rates higher priority.

### DIFF
--- a/lib/rucio/core/transfer.py
+++ b/lib/rucio/core/transfer.py
@@ -1071,6 +1071,44 @@ class PreferSingleHop(PathDistance):
         return int(len(path) > 1)
 
 
+class FailureRate(SourceRankingStrategy):
+    """
+    A source ranking strategy that ranks source nodes based on their failure rates for the past hour. Failure rate is
+    calculated by dividing files failed by files attempted.
+    """
+    class _FailureRateStat:
+        def __init__(self) -> None:
+            self.files_done = 0
+            self.files_failed = 0
+
+        def incorporate_stat(self, stat: "Mapping[str, int]") -> None:
+            self.files_done += stat['files_done']
+            self.files_failed += stat['files_failed']
+
+        def get_failure_rate(self) -> int:
+            files_attempted = self.files_done + self.files_failed
+
+            # If no files have been sent yet, return failure rate as 0
+            if files_attempted == 0:
+                return 0
+
+            return int((self.files_failed / files_attempted) * 10000)
+
+    def __init__(self, stats_manager: "request_core.TransferStatsManager") -> None:
+        super().__init__()
+        self.source_stats = {}
+
+        for stat in stats_manager.load_totals(
+            datetime.datetime.utcnow() - datetime.timedelta(hours=1),
+            by_activity=False
+        ):
+            self.source_stats.setdefault(stat['src_rse_id'], self._FailureRateStat()).incorporate_stat(stat)
+
+    def apply(self, ctx: RequestRankingContext, source: RequestSource) -> "Optional[int | _SkipSource]":
+        failure_rate = cast(FailureRate, ctx.strategy).source_stats.get(source.rse.id, self._FailureRateStat()).get_failure_rate()
+        return failure_rate
+
+
 class SkipSchemeMissmatch(PathDistance):
     filter_only = True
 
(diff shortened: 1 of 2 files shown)


lib/rucio/core/transfer.py
```diff
@@ -1071,6 +1071,69 @@ def apply(self, ctx: RequestRankingContext, source: RequestSource) -> "Optional[
...
+
+        for src in rws.sources:
+            failure_rate_for_rws[src.rse.id] = (
+                self.source_stats.get(src.rse.id, self._FailureRateStat()).get_failure_rate()
+            )
+
+        return FailureRate._RankingContext(self, rws, failure_rate_for_rws)
+
+    def apply(self, ctx: RequestRankingContext, source: RequestSource) -> "Optional[int | _SkipSource]":
+        failure_rate = cast(FailureRate._RankingContext, ctx).failure_rate_for_rws.get(source.rse.id)
```
rcarpa: access the data from `cast(ctx.strategy, FailureRate).source_stats.get(source.rse.id)` directly. No need to just copy data from one dict to the other in `for_request`. If some additional computation was needed, it would be justified to implement that custom logic in `for_request`, but in this case I don't see need for complicating the program flow. You could the completely get rid of the `FailureRate._RankingContext` sub-class. And remove the override implementation of `for_request`

lib/rucio/core/transfer.py
```diff
@@ -1156,6 +1222,7 @@ def build_transfer_paths(
         PreferDiskOverTape.external_name,
         PathDistance.external_name,
         PreferSingleHop.external_name,
+        FailureRate.external_name,
```
rcarpa: Don't add this strategy to the default ones. We must not change the default behavior. Users would be surprised to see the behavior of the program change so radically after an upgrade.

lib/rucio/core/transfer.py
```diff
@@ -1071,6 +1071,69 @@ def apply(self, ctx: RequestRankingContext, source: RequestSource) -> "Optional[
         return int(len(path) > 1)
 
 
+class FailureRate(SourceRankingStrategy):
```
rcarpa, line 1074: Add a docstring with a high level description/documentation of this strategy.

rcarpa (changes requested):

rcarpa:
Also, add the type hints. And fix the pyright problems (see the results of the tests pipeline run). I recommend to convert the cost to int (for example, transform it from ratio to percentage by multiplying by 100 and taking the integer part of it)

lib/rucio/core/transfer.py
```diff
@@ -1071,6 +1071,44 @@ def apply(self, ctx: RequestRankingContext, source: RequestSource) -> "Optional[
...
+        self.source_stats = {}
+
+        for stat in stats_manager.load_totals(
+            datetime.datetime.utcnow() - datetime.timedelta(hours=1),
+            by_activity=False
+        ):
+            self.source_stats.setdefault(stat['src_rse_id'], self._FailureRateStat()).incorporate_stat(stat)
+
+    def apply(self, ctx: RequestRankingContext, source: RequestSource) -> "Optional[int | _SkipSource]":
+        failure_rate = cast(FailureRate, ctx.strategy).source_stats.get(source.rse.id).get_failure_rate()
```
rcarpa: get() will return None if the element doesn't exist. This has to be handled gracefully . 
It's even shown as a mistake by the pipeline: https://github.com/rucio/rucio/actions/runs/7096607130/job/19322721914

lib/rucio/core/transfer.py
```diff
@@ -1071,6 +1071,44 @@ def apply(self, ctx: RequestRankingContext, source: RequestSource) -> "Optional[
         return int(len(path) > 1)
 
 
+class FailureRate(SourceRankingStrategy):
+    """
+    A source ranking strategy that ranks source nodes based on their failure rates for the past hour. Failure rate is
+    calculated by dividing files failed by files attempted.
+    """
+    class _FailureRateStat:
+        def __init__(self):
```
rcarpa: You are still missing some type annotations (the return values of the two __init__ . The pipeline is too strinct to my liking on this subject, but we have to live with it :)  https://github.com/rucio/rucio/actions/runs/7096607130/job/19322721542

rcarpa (changes requested):

rcarpa (approved):

PR #6415, opened 2023-12-12
Transfers: Add Dynamic TransferWaitTime SourceRankingStrategy

JChenster:
We add a dynamic TransferWaitTime SourceRankingStrategy to estimate for each source node, how long a new transfer has to wait to begin by approximating how long it will take to process the current queue. Source nodes with smaller estimated wait times are preferred. We introduce randomization so that we are able to distribute requests across multiple sources if their estimated wait times are similar and so that even sources that we think have the longest wait time will still get picked a small amount of the time. We choose below average wait sources 90% of the time and above average wait sources 10% of the time, and proportionally pick randomly based on wait time.

NOTE: We intend to roll out a Metrics tracking scheme to be able to collect metrics on the effectiveness of such a dynamic source selection strategy such as this one. This will come in a later pull request.

### DIFF
--- a/lib/rucio/core/transfer.py
+++ b/lib/rucio/core/transfer.py
@@ -16,6 +16,7 @@
 import datetime
 import logging
 import operator
+import random
 import re
 import sys
 import time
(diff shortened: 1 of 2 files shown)


lib/rucio/core/transfer.py
```diff
@@ -1063,6 +1064,88 @@ def apply(self, ctx: RequestRankingContext, source: RequestSource) -> "Optional[
...
+            logger: "LoggerFunction" = logging.log,
+            session: "Session"
+    ) -> "RequestRankingContext":
+        wait_by_rse = {}
+
+        # For each source, estimate the average queue waiting time
+        for src in sources:
+            src_files_queued = 0
+            src_bytes_queued = 0
+            metrics = request_core.get_request_metrics(src_rse_id=src.rse.id, session=session)
```
rcarpa: This call to get_request_metrics hides a very costly operation. Especially if done in a loop. The database table is not indexed by src_rse_id. So it will result in a full table scan. Currently we are approaching a million rows in it.

lib/rucio/core/transfer.py
```diff
@@ -1063,6 +1064,88 @@ def apply(self, ctx: RequestRankingContext, source: RequestSource) -> "Optional[
...
+
+            for payload in metrics.values():
+                src_files_queued += payload.get('files', {}).get('queued-total', 0)
+                src_bytes_queued += payload.get('bytes', {}).get('queued-total', 0)
+
+            wait_by_rse[src.rse] = src_files_queued * self.FILE_OVERHEAD_MS + src_bytes_queued * self.PER_BYTE_MS
+
+        costs = defaultdict(lambda: 0)
```
rcarpa: defaultdict(int) or float seems more appropriate. Even if i'm not sure it will actually be faster on python interpreter side.

tests/test_transfer.py
```diff
@@ -247,6 +249,98 @@ def test_disk_vs_tape_with_custom_strategy(rse_factory, root_account, mock_scope
...
+    dst_rse_name, dst_rse_id = rse_factory.make_posix_rse()
+    all_rses = [shortest_wait_rse_id, short_wait_rse_id, long_wait_rse_id, longest_wait_rse_id, dst_rse_id]
+    add_distance(shortest_wait_rse_id, dst_rse_id, distance=40)
+    add_distance(short_wait_rse_id, dst_rse_id, distance=30)
+    add_distance(long_wait_rse_id, dst_rse_id, distance=20)
+    add_distance(longest_wait_rse_id, dst_rse_id, distance=10)
+
+    # Add mock data about existing source queues
+    db_session = get_session()
+
```
rcarpa: just decorate the function with the transactional_session decorator. Which will take care of creating the session and commiting it.

tests/test_transfer.py
```diff
@@ -247,6 +249,98 @@ def test_disk_vs_tape_with_custom_strategy(rse_factory, root_account, mock_scope
...
+        print("shortest_wait_rse_picks =", picks_by_rse_name[shortest_wait_rse_name])
+        print("short_wait_rse_picks =", picks_by_rse_name[short_wait_rse_name])
+        print("long_wait_rse_picks =", picks_by_rse_name[long_wait_rse_name])
+        print("longest_wait_rse_picks =", picks_by_rse_name[longest_wait_rse_name])
+
+        # We can model probability of picking a source using a binomial distribution.
+        # Standard deviation is sqrt(num_requests * .1 * .9). +- 3 standard deviation is 99.7% confidence interval
+        stdev = (num_requests * .1 * .9) ** 0.5
+        for rse_name, expected_prob in expected_prob_by_rse_name.items():
```
rcarpa, line 336: 99.7 % still means that 3 out of 1000 test runs can fail without there being any bug. This is a lot. considering that we run something like 20 tests in each run of the pipeline. Having tests randomly failing for no reason is not good. Even if the probability is low. Developers will start thinking they did something wrong in their code and can spend hours on a wild chase for an non-existing bug.

rcarpa (commented):
I didn't read through the strategy and cannot comment about the mathematics behind it. Only some comments related to high-level issues of the code.

tests/test_transfer.py
JChenster: I have changed it to +- 4 standard deviations. With 4 standard deviations, confidence interval is 99.9936% accurate (~1/15,625). I have also added a comment notifying folks that in the unlikely situation that it fails to simply re-run. Hopefully this is acceptable or is a randomized unit test always frowned upon?

For our test case, the st dev is 3 so +- 4 means +- 12 from the expected value. This is still meaningful to test because for example if expected value is 30, then confidence interval is [18, 42] and when expected value is 60, confidence interval is [48, 72] which do not overlap.

lib/rucio/core/transfer.py
JChenster: I realized that `get_request_metrics` also scans the `TransferStatsManager` table which we don't need so I migrated over to using `get_request_stats` which only scans the `Request` table. To our understanding, this table only contains information on the currently active requests. To clarify: is there 1M rows of active requests at a given time?

rcarpa: It still isn't indexed by the request source as of today. But we were discussing about adding that index for a long time. So may be worth doing it.

bari12:
Dear @JChenster 
Sorry for taking that long. Although I don't know of anyone who actually wants to push for it in production, I think it is still a good example of a more advanced strategy and would therefore merge it. Can you please rebase.

PR #6416, opened 2023-12-12
enhancement: reset stale waiting requests Fix #6404

### LINKED ISSUE #6404: Throttled requests don't react to source changes
The preparer picks a source out of available candidates and then applies throttling rules (which also include the rules configured from the source rse). However, it can happen that the situation changes over time. For example, a replica becomes available on a better source while the request is blocked in a Waiting state within the throttler. It's not unheard of requests staying in waiting state for multiple days or even weeks. 

To handle this case and trigger source re-evaluation. I propose to transition the requests back to the Preparing state if they spent too much time in a waiting state.

MatthewQiu-5:
Added functionality at the end of the conveyor-throttler main loop that will reset all requests that have been in the waiting state for >1 day back to the preparing state and clear their source_rse_id. Also added additional checks to the integration test of the preparer/throttler workflow to validate this scenario and ensure that requests properly get reset based on the last_processed_at timestamp. This allows for requests that have been throttled for a long time to account for new sources and changes that have occurred while they have been throttled.

### DIFF
--- a/lib/rucio/core/request.py
+++ b/lib/rucio/core/request.py
@@ -2817,3 +2817,39 @@ def list_requests_history(src_rse_ids, dst_rse_ids, states=None, offset=None, li
         stmt = stmt.limit(limit)
     for request in session.execute(stmt).yield_per(500).scalars():
         yield request
+
+
+@transactional_session
+def reset_stale_waiting_requests(time_limit: Optional[datetime.timedelta] = datetime.timedelta(days=1), *, session: "Session") -> None:
+    """
+    Clear source_rse_id for requests that have been in the waiting state for > time_limit amount of time and
+    transition back to preparing state (default time limit = 1 day).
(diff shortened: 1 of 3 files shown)


lib/rucio/core/request.py
```diff
@@ -2817,3 +2817,39 @@ def list_requests_history(src_rse_ids, dst_rse_ids, states=None, offset=None, li
         stmt = stmt.limit(limit)
     for request in session.execute(stmt).yield_per(500).scalars():
         yield request
+
+
+@transactional_session
+def reset_stale_waiting_requests(day_limit: Optional[int] = 1, *, session: "Session") -> None:
```
rcarpa: just pass a datetime.timedelta directly as an argument

rcarpa (changes requested):
Nice work. Only a minor change to be done and it will be good for me.

rcarpa (approved):

Issue #5012, opened 2021-11-16
Dynamic distance between RSEs


rcarpa:
Motivation
----------

For now, the shortest path computation and source selection is completely static. This can result in overloaded RSEs while other sources are completely unused. 

Modification
------------

We would like to make the decisions more dynamic and dependent on the load

rcarpa:
As first step, maybe use the backlog request queue length as a weight in sorting

rodwalker:
Hi,
I think the numbers on
https://atlasdistributedcomputing-live.web.cern.ch/ATLASDistributedComputing-live/#Satellites
come from some huge json, ultimately from rucio - so the information is there.
http://atlas-adc-netmetrics-lb.cern.ch/metrics/latest.json

The CERN-PROD queued input 30k and estimated transfer time of 19hrs matches somewhat with what I see in FTS
https://fts3-atlas.cern.ch:8449/fts3/ftsmon/#/?vo=&source_se=davs:%2F%2Fxrootd.echo.stfc.ac.uk&dest_se=davs:%2F%2Feosatlas.cern.ch&time_window=2

The #active is limited by the outbound total for RAL.

I think almost anything will be better than static.

Cheers,
Rod.

rcarpa:
I performed the internal work in rucio to allow taking into consideration the queues and the historical transfer data for source selection. By historical data, I mean the information available in the "Transfer Plots" section of the ATLAS "DDM Transfers" dashboard. 

At this point I'm quite blocked because there is no consensus even within the team if we should be more intelligent in source selection. 
We must define a strategy. Implement it. And test it in ATLAS. I don't see any other way forward. This means accepting the risk of actually doing things worse.

rodwalker:
So we are talking about taking the source selection away from FTS and giving it a single source replica? I think it can take an ordered list too, to allow some failover. If this new behaviour could be triggered by some attribute to the rule, e.g. priority, then maybe it could be tested without breaking what we have. Or could do like the pilot and 1% of  transfers use the dev schema(and are tagged as such). So I am in favour of just trying it.

It would be nice to have some metric, other than "Rod is moaning a bit". Something like the mean transfer time for files with multiple disk replicas.
After excluding non-functional or very slow connections, the main criteria will be the queued transfers. This would include file size as well as number, since we move a lot of 20GB files now. Time to transfer is Nf*Tf + Nmb/mbpersec and we order sources by this. Tf ~10s, is the overhead for transferring even a very small file.
[...]

PR #6420, opened 2023-12-13
Transfers: add average transfer wait time metric #5012

davidcheng02:
We keep track of the running average wait time for all completed transfers sent to update_transfer_state. This metric aims to measure the effectiveness of load distribution from different source node selection strategies. Lower average wait time generally signifies a more effective strategy. The average is based on a tumbling window of 6 hours.

### DIFF
--- a/lib/rucio/core/request.py
+++ b/lib/rucio/core/request.py
@@ -1371,6 +1371,9 @@ class TransferStatsManager:
             state: RequestState,
             file_size: int,
             *,
+            submitted_at: Optional[datetime.datetime] = None,
+            started_at: Optional[datetime.datetime] = None,
+            transferred_at: Optional[datetime.datetime] = None,
             session: "Optional[Session]" = None
     ) -> None:
         """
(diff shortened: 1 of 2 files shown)


lib/rucio/core/request.py
```diff
@@ -1390,6 +1396,20 @@ def observe(
                 if state == RequestState.DONE:
                     record.files_done += 1
                     record.bytes_done += file_size
+
+                    if submitted_at is not None and started_at is not None:
+                        wait_time = (started_at - submitted_at).total_seconds()
+
+                        # avg_wait_time averages across a tumbling window, so reset avg every few hours
+                        hours_window = 6
+                        if self.avg_wait_time_start_timestamp < (now - datetime.timedelta(hours=hours_window)):
```
davidcheng02: We decided to use tumbling window for calculating our running average. While sliding window may make more sense, it requires storing all the transfers in the past X hours, and additional computation is required to remove old transfers and adding in new transfers. Therefore, tumbling window is the most feasible.

lib/rucio/core/request.py
```diff
@@ -1390,6 +1396,20 @@ def observe(
...
+                            self.avg_wait_time = 0
+                            self.transfers_done = 0
+                            self.avg_wait_time_start_timestamp = now
+
+                        self.avg_wait_time = ((self.avg_wait_time * self.transfers_done) + wait_time) / (self.transfers_done + 1)
+                        self.transfers_done += 1
+                        METRICS.gauge('avg_wait_time').set(self.avg_wait_time)
```
davidcheng02: We do not add our gauge metric with any filters (such as by src id or activity), since this gauge will be a global value corresponding to the overall average wait time for an entire topology. We imagine that on prometheus, this gauge will show up as a line graph or time plot, where the x-axis is the timestamp when the gauge metric was pulled, and y-axis is the avg_wait_time.

rcarpa:
As mentioned off-line, you should use a prometheus Histogram (It's badly named "Timing"/"timer" in our code for historical reasons linked to usage of statsd in the past). The usage of histograms is a bit counter-intuitive and tricky. Mostly because you have to correctly set the "buckets", so you'll have to learn how to use it. The advantage is that we don't need to compute a moving average in the code. We can record raw values. If the buckets are correctly chosen, we'll be able to use prometheus integrated query engine to answer questions like "how many requests which ended in the last 24 hours took more than a hour to complete" or "what is the duration of transfers if we ignore the highest 5 percentile" .. etc.

In this case buckets will have to be set in a way which cover both the shortest transfers and the longest ones. For example: 
`10s 30s 1m 5m 10m 20m 40m 1h30m 3h 6h 12h 1d 2d 3d 4d 5d 6d 7d 10d 14d 1m infinity`. You have quite sparse granularity at higher ranges, but it should be "good enough"

rcarpa (dismissed):

rcarpa:
The test failure at https://github.com/rucio/rucio/blob/436ecc9005934e1e4332cd2718d160bdbf5105d2/tests/test_conveyor.py#L287 may actually be a sign that the new code raises an exception. This has to be double-checked .

rcarpa (approved):

Issue #6419, opened 2023-12-13
Throttled sources should react to source changes


NukeWolf:
### Description

The preparer picks a source out of available candidates and then applies throttling rules (which also include the rules configured from the source rse). However, it can happen that the situation changes over time. For example, a replica becomes available on a better source while the request is blocked in a Waiting state within the throttler. 

To handle this case and trigger source re-evaluation. I propose to transition the requests back to the Preparing state when requests with a potential replica get completed

PR #6421, opened 2023-12-14
Transfers: reset throttled requests when replica become available #6419

NukeWolf:
When requests finish through the finisher and new replicas become available, replica names are sent through a message queue through ActiveMQ to the throttler. The throttler will proceed to receive the message and reset the state of any affect requests with the same DID back to PREPARING.

### DIFF
--- a/etc/docker/test/extra/rucio_default.cfg
+++ b/etc/docker/test/extra/rucio_default.cfg
@@ -54,6 +54,7 @@ transfertool = fts3
 ftshosts = https://fts:8446
 cacert = /etc/grid-security/certificates/5fca1cb1.0
 usercert = /opt/rucio/etc/usercertkey.pem
+use_preparer = true
 
 [messaging-fts3]
 port = 61613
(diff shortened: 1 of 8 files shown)


rcarpa (commented):
I'm in favor of the idea, but I don't want it to be an example of yet another copy-pasting of the messaging code from other components. Unfortunately, we had a bad track record of doing it in the past, and now we have this messaging code duplicated in `cache consummer`, `receiver`, `hermes`, Before even starting to review the actual functional change from this PR (related to communication between finisher and throttler), I think we need a cleanup of the current situation. Otherwise we are only making it worse. 

we need to factorize all this common code into a helper module or class, maybe into `common/stomp_utils.py`, and re-use it everywhere we need instead of duplicating the code once again. 

If you are willing and have time to work on it, this is a very tedious task and prone to breaking stuff, but will be waaaay more useful to the overall rucio project than the feature proposed in this PR: https://github.com/rucio/rucio/issues/6423

bari12:
Hi @NukeWolf 
While this is in principle a good idea, I think this adds a layer of complication I don't really want to add to Rucio at this point in time. We would have to do this in a more organised push, but I don't think we really need it at this point in time and it will make things more complicated and complex than easier. Thanks for the good work, and sorry for the bad news; Closing this now.

Issue #6382, opened 2023-11-20
Refactor policy package algorithm code


jamesp-epcc:
### Description

[PR 6215](https://github.com/rucio/rucio/pull/6215) introduced a new class for registering algorithms from policy packages. This is a cleaner way of doing it than the method previously used elsewhere in the code so it would be good to refactor the code to use this class for all access to policy package algorithms (namely for SURL, scope extraction and LFN2PFN algorithms).

However would it make sense to move the class from `core/plugins.py` to `common/plugins.py` to preserve the layering of the code and avoid calling `core` from `common`? Two of the other uses of policy package algorithms are in `common/utils.py`.

PR #6426, opened 2023-12-18
Policies: refactor policy package code to use new plugin code #6382

jamesp-epcc:
This PR refactors the policy package algorithm code to use the new class introduced as part of the rule auto-approval work. All of the algorithm types (SURL, LFN2PFN and scope extraction) now use this base class, resulting in less duplication and hopefully cleaner code.

I have moved `plugins.py` from `core` to `common` to preserve the layering of the code, because the scope extraction algorithms are used client side.

### DIFF
--- a/lib/rucio/common/plugins.py
+++ b/lib/rucio/common/plugins.py
@@ -20,8 +20,6 @@ from typing import Any, Callable, Dict, TypeVar, Type
 
 from rucio.common import config
 from rucio.common.exception import InvalidAlgorithmName
-from rucio.common.utils import check_policy_package_version
-from rucio.core.vo import list_vos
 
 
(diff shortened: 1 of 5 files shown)


bari12 (dismissed):

bari12:
@jamesp-epcc can you please rebase this and solve the conflict.

jamesp-epcc:
I've rebased and fixed the conflict. Also had to modify `fts3_plugins` as it was trying to import the plugins module from `core`.

jamesp-epcc:
I see the integration tests failed, but I can't see an obvious connection with my changes. I'm wondering if this is one of the failures mentioned at the meeting yesterday.

bari12 (approved):

Issue #6423, opened 2023-12-14
factorize duplicate messaging code into a common module or class


rcarpa:
The current situation of activemq message handling is a huge mess in rucio. Every daemon which needed to perform communication via  a message queue copy-pasted the code from another daemon. It's time to cleanup it. We need to factorize the code and re-use things as much as posisble.

bari12:
Thank you for raising this. We have reviewed this issue as part of our regular issue review.

We see value in this request, but we do not have the capacity to work on it within the next twelve months. 

In line with our [issue management guideline](https://rucio.cern/documentation/developer/issue_management), we are labelling it `wishlist` and closing it. This does not mean it is forgotten: wishlist items are revisited regularly during component reviews and roadmap planning, and this issue can be re-opened and added back to the pipeline as soon as interest and capacity align.

PR #6427, opened 2023-12-18
Common : refactored stomp connection code #6421

NukeWolf:
Refactored and moved configuration code and connection code into common stomp_lib


bari12:
Dear @NukeWolf 
Apologies for the slow reply here. I will close this PR now. It was not in vain though since it motivated parts of the #6423 issue where this work will continue! THanks a lot

PR #6443, opened 2024-01-08
Tests: initial token infrastructure

rcarpa:
Install and configure two oidc providers: indigo iam and keycloak. For this, had to rework the container initialization to wait for the database to be ready and to initialize it with the test configuration.

Configure xrood in the XRD2 container to allow token authentication: this required a separate entrypoint and mounting multiple configuration files into the container.

Add two Webdav Containers. Only WEB0 is configured for token support, the second one relies on certificates.


rcarpa:
This PR depends on https://github.com/rucio/containers/pull/280  and https://github.com/rucio/rucio/pull/6440

rcarpa:
It also relies on some more advanced features of docker-compose which I'm not sure are supported by podman-compose

bari12 (approved):
LGTM

Issue #6398, opened 2023-11-30
Metadata for tape co-location and transfer prority


voetberg:
### Description

Provide a way to include metadata in a transfer, including priority based on activity and a tree structure to group like data. 

### Motivation

Priority - Ensures important data is moved up on the buffer and will be written in a timely manner (and not get tossed out in case of overflow) 
Metadata - Allows FTS to co-locate like data and increase read efficiency of tapes

![image](https://github.com/rucio/rucio/assets/27649232/ba7a34ec-7b0f-4288-a73c-a1785662e3ac)
[No metadata for collocation (left), vs a perfect world of collocation (right).] 

### Priority 
- Use experiment activities [already existing], and map those to a priority (scale from 0-100) [an addition]. 
- Pass the priority level to the FTS Transfer Tool in the form of: 
```
archive_metadata = {
   "scheduling_hints":{
         "priority": "<priority:int>"
}
} 
```
- Subclass `PolicyPackageAlgorithms`
![image](https://github.com/rucio/rucio/assets/27649232/080f596b-7a74-44da-b608-040147d65119)

### Metadata 
- Create an abstract class (subclass of `PolicyPackageAlgorithms`) that allows each experiment to determine their own tree structure and how they are applying colocation hints. 
- Pass hints to FTS in the form of: 
```
archive_metadata = {
   "collocation_hints": {
      "<level:int>": <associated label of experiment, project, dataset, etc>
}
```
![image](https://github.com/rucio/rucio/assets/27649232/bf8dffa8-1aea-4b0e-ab1c-b6be1a717dad)

### Housekeeping 
- Need to ensure none of the archive metadata json's exceed the fts metadata size limit (4kb?) 
- Integrate this into `conveyor` daemon. 
- Dummy collocation hints (testing) 
- Default priorities, add priorities to existing activities. 
- Docs [for the general usage, and creating a custom plugin collocation hint method]

PR #6445, opened 2024-01-10
Transfers: metadata for collocation on tape #6398

voetberg:
Constructs a plugin class that adds a field to the transfer metadata, utilized by fts3 for placing like with like when storing on tape

Plugin is a subclass of `PolicyPackageAlgorithm`, so any experiment-specific plugin someone writes needs to be registered as a policy under the name "fts3_plugins" to be picked up during the initialization of the fts3 transfer tool. 

Additionally, to use a specifically plugin, it needs to be included in the rucio.config as the following: 
```
[transfers] 
plugins=<A list of the names of plugins being used, as defined during the plugin registration>
```

Included pre-build is a priority field based on activity, which produces an append to the transfer job submission in the form: 
```
{"archive_metadata": {"scheduling_hints": {"priority": "20"}}
```
Where priority is assigned based on activity, defined by a table set in the config. Default priority is 20. 
```

[transfers]
plugins=activity

[tape priority]
<activity_1>=40 
<activity_2>=100
```

### DIFF
--- a/lib/rucio/common/utils.py
+++ b/lib/rucio/common/utils.py
@@ -2180,3 +2180,16 @@ def retrying(
                 time.sleep(wait_fixed / 1000.0)
         return _wrapper
     return _decorator
+
+
+def deep_merge_dict(source: dict, destination: dict) -> dict:
+    """Merge two dictionaries together recurively"""
+    for key, value in source.items():
+        if isinstance(value, dict):
+            # get node or create one
+            node = destination.setdefault(key, {})
+            deep_merge_dict(value, node)
+        else:
+            destination[key] = value
+
+    return destination
--- a/lib/rucio/common/utils.py
+++ b/lib/rucio/common/utils.py
@@ -40,7 +40,7 @@ from enum import Enum
 from functools import partial, wraps
 from io import StringIO
 from itertools import zip_longest
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Callable, Optional
 from urllib.parse import urlparse, urlencode, quote, parse_qsl, urlunparse
 from uuid import uuid4 as uuid
 from xml.etree import ElementTree
@@ -65,9 +65,7 @@ if EXTRA_MODULES['paramiko']:
         EXTRA_MODULES['paramiko'] = False
 
 if TYPE_CHECKING:
-    from collections.abc import Callable
-    from typing import TypeVar, Optional
-
+    from typing import TypeVar
     T = TypeVar('T')
 
 
@@ -2183,7 +2181,7 @@ def retrying(
 
 
 def deep_merge_dict(source: dict, destination: dict) -> dict:
-    """Merge two dictionaries together recurively"""
+    """Merge two dictionaries together recursively"""
     for key, value in source.items():
         if isinstance(value, dict):
             # get node or create one
--- a/lib/rucio/transfertool/fts3.py
+++ b/lib/rucio/transfertool/fts3.py
@@ -31,11 +31,11 @@ from requests.adapters import ReadTimeout
 from requests.packages.urllib3 import disable_warnings  # pylint: disable=import-error
 
 from rucio.common.cache import make_region_memcached
-from rucio.common.config import config_get, config_get_bool, config_get_int
+from rucio.common.config import config_get, config_get_bool, config_get_int, config_get_list
 from rucio.common.constants import FTS_JOB_TYPE, FTS_STATE, FTS_COMPLETE_STATE
 from rucio.common.exception import TransferToolTimeout, TransferToolWrongAnswer, DuplicateFileTransferSubmission
 from rucio.common.stopwatch import Stopwatch
-from rucio.common.utils import APIEncoder, chunks, PREFERRED_CHECKSUM
+from rucio.common.utils import APIEncoder, chunks, PREFERRED_CHECKSUM, deep_merge_dict
 from rucio.core.monitor import MetricManager
 from rucio.core.oidc import request_token
 from rucio.core.request import get_source_rse, get_transfer_error
@@ -43,6 +43,7 @@ from rucio.core.rse import (determine_audience_for_rse, determine_scope_for_rse,
                             get_rse_supported_checksums_from_attributes)
 from rucio.db.sqla.constants import RequestState
 from rucio.transfertool.transfertool import Transfertool, TransferToolBuilder, TransferStatusReport
+from rucio.transfertool.fts3_plugins import FTS3MetadataPlugin
 
 if TYPE_CHECKING:
     from rucio.core.request import DirectTransfer
@@ -812,6 +813,9 @@ class FTS3Transfertool(Transfertool):
         self.default_lifetime = default_lifetime
         self.archive_timeout_override = archive_timeout_override
 
+        plugins = config_get_list("transfers", "plugins", False, "[]")
+        self.plugins = [FTS3MetadataPlugin(plugin.strip(" ")) for plugin in plugins]
+
         self.token = None
         if oidc_support:
             fts_hostname = urlparse(external_host).hostname
@@ -950,6 +954,11 @@ class FTS3Transfertool(Transfertool):
             activity_id = self.scitags_activity_ids.get(rws.activity)
             if isinstance(activity_id, int):
                 t_file['scitag'] = self.scitags_exp_id << 6 | activity_id
+
+        for plugin in self.plugins:
+            plugin_hints = plugin.hints(t_file['metadata'])
+
+            t_file = deep_merge_dict(source=plugin_hints, destination=t_file)
 
         return t_file
 
--- a/lib/rucio/transfertool/fts3.py
+++ b/lib/rucio/transfertool/fts3.py
@@ -22,7 +22,7 @@ import uuid
 from collections.abc import Callable
 from configparser import NoOptionError, NoSectionError
 from json import loads
-from typing import Any, Optional, TYPE_CHECKING
+from typing import TYPE_CHECKING, Any, Optional
 from urllib.parse import urlparse
 
 import requests
@@ -43,11 +43,12 @@ from rucio.core.rse import (determine_audience_for_rse, determine_scope_for_rse,
                             get_rse_supported_checksums_from_attributes)
 from rucio.db.sqla.constants import RequestState
 from rucio.transfertool.transfertool import Transfertool, TransferToolBuilder, TransferStatusReport
-from rucio.transfertool.fts3_plugins import FTS3MetadataPlugin
+from rucio.transfertool.fts3_plugins import FTS3TapeMetadataPlugin
 
 if TYPE_CHECKING:
     from rucio.core.request import DirectTransfer
     from rucio.core.rse import RseData
+    from sqlalchemy.orm import Session
 
 logging.getLogger("requests").setLevel(logging.CRITICAL)
 disable_warnings()
@@ -309,7 +310,12 @@ def _use_tokens(transfer_hop: "DirectTransfer"):
     return True
 
 
-def build_job_params(transfer_path, bring_online, default_lifetime, archive_timeout_override, max_time_in_queue, logger):
+def build_job_params(
+        transfer_path: list["DirectTransfer"],
+        bring_online: Optional[int] = None,
+        default_lifetime: Optional[int] = None,
+        archive_timeout_override: Optional[int] = None,
+        max_time_in_queue: Optional[dict] = None, logger: Callable = logging.log) -> dict[str, Any]:
     """
     Prepare the job parameters which will be passed to FTS transfertool
     """
@@ -403,8 +409,16 @@ def build_job_params(transfer_path, bring_online, default_lifetime, archive_time
     return job_params
 
 
-def bulk_group_transfers(transfer_paths, policy='rule', group_bulk=200, source_strategy=None, max_time_in_queue=None,
-                         logger=logging.log, archive_timeout_override=None, bring_online=None, default_lifetime=None):
+def bulk_group_transfers(
+        transfer_paths: list[list["DirectTransfer"]],
+        policy: str = 'rule',
+        group_bulk: int = 200,
+        source_strategy: Optional[str] = None,
+        max_time_in_queue: Optional[dict] = None,
+        logger: Callable = logging.log,
+        archive_timeout_override: Optional[int] = None,
+        bring_online: Optional[int] = None,
+        default_lifetime: Optional[int] = None) -> list[dict[str, Any]]:
     """
     Group transfers in bulk based on certain criterias
 
@@ -495,7 +509,7 @@ class Fts3TransferStatusReport(TransferStatusReport):
         'attributes',
     ]
 
-    def __init__(self, external_host, request_id, request=None):
+    def __init__(self, external_host: str, request_id: str, request: Optional[dict] = None):
         super().__init__(request_id, request=request)
         self.external_host = external_host
 
@@ -524,10 +538,10 @@ class Fts3TransferStatusReport(TransferStatusReport):
         return f'Transfer {self._transfer_id} of {self._file_metadata["scope"]}:{self._file_metadata["name"]} ' \
                f'{self._file_metadata["src_rse"]} --({self._file_metadata["request_id"]})-> {self._file_metadata["dst_rse"]}'
 
-    def initialize(self, session, logger=logging.log):
+    def initialize(self, session: "Session", logger: Callable = logging.log) -> None:
         raise NotImplementedError(f"{self.__class__.__name__} is abstract and shouldn't be used directly")
 
-    def get_monitor_msg_fields(self, session, logger=logging.log):
+    def get_monitor_msg_fields(self, session: "Session", logger: Callable = logging.log) -> dict[str, Any]:
         self.ensure_initialized(session, logger)
         fields = {
             'transfer_link': self._transfer_link(),
@@ -546,7 +560,7 @@ class Fts3TransferStatusReport(TransferStatusReport):
     def _transfer_link(self):
         return '%s/fts3/ftsmon/#/job/%s' % (self._fts_address.replace('8446', '8449'), self._transfer_id)
 
-    def _find_attribute_updates(self, request, new_state, reason, overwrite_corrupted_files):
+    def _find_attribute_updates(self, request: dict, new_state: RequestState, reason: str, overwrite_corrupted_files: Optional[bool] = None) -> Optional[dict[str, Any]]:
         attributes = None
         if new_state == RequestState.FAILED and 'Destination file exists and overwrite is not enabled' in (reason or ''):
             dst_file = self._file_metadata.get('dst_file', {})
@@ -556,7 +570,7 @@ class Fts3TransferStatusReport(TransferStatusReport):
                     attributes['overwrite'] = True
         return attributes
 
-    def _find_used_source_rse(self, session, logger):
+    def _find_used_source_rse(self, session: "Session", logger: Callable) -> tuple[Optional[str], Optional[str]]:
         """
         For multi-source transfers, FTS has a choice between multiple sources.
         Find which of the possible sources FTS actually used for the transfer.
@@ -574,7 +588,7 @@ class Fts3TransferStatusReport(TransferStatusReport):
         return meta_rse_name, meta_rse_id
 
     @staticmethod
-    def _dst_file_set_and_file_corrupted(request, dst_file):
+    def _dst_file_set_and_file_corrupted(request: dict, dst_file: dict) -> bool:
         """
         Returns True if the `dst_file` dict returned by fts was filled and its content allows to
         affirm that the file is corrupted.
@@ -587,7 +601,7 @@ class Fts3TransferStatusReport(TransferStatusReport):
         return False
 
     @staticmethod
-    def _dst_file_set_and_file_correct(request, dst_file):
+    def _dst_file_set_and_file_correct(request: dict, dst_file: dict) -> bool:
         """
         Returns True if the `dst_file` dict returned by fts was filled and its content allows to
         affirm that the file is correct.
@@ -644,7 +658,7 @@ class FTS3CompletionMessageTransferStatusReport(Fts3TransferStatusReport):
         self._src_url = fts_message.get('src_url', None)
         self._dst_url = fts_message.get('dst_url', None)
 
-    def initialize(self, session, logger=logging.log):
+    def initialize(self, session: "Session", logger: Callable = logging.log) -> None:
 
         fts_message = self.fts_message
         request_id = self.request_id
@@ -716,7 +730,7 @@ class FTS3ApiTransferStatusReport(Fts3TransferStatusReport):
         self._dst_url = file_response.get('dest_surl', None)
         self.logger = logging.log
 
-    def initialize(self, session, logger=logging.log):
+    def initialize(self, session: "Session", logger=logging.log) -> None:
 
         self.logger = logger
         job_response = self.job_response
@@ -813,8 +827,8 @@ class FTS3Transfertool(Transfertool):
         self.default_lifetime = default_lifetime
         self.archive_timeout_override = archive_timeout_override
 
-        plugins = config_get_list("transfers", "plugins", False, "[]")
-        self.plugins = [FTS3MetadataPlugin(plugin.strip(" ")) for plugin in plugins]
+        tape_plugins = config_get_list("transfers", "fts3tape_metadata_plugins", False, "[]")
+        self.tape_metadata_plugins = [FTS3TapeMetadataPlugin(plugin.strip(" ")) for plugin in tape_plugins]
 
         self.token = None
         if oidc_support:
@@ -955,7 +969,7 @@ class FTS3Transfertool(Transfertool):
             if isinstance(activity_id, int):
                 t_file['scitag'] = self.scitags_exp_id << 6 | activity_id
 
-        for plugin in self.plugins:
+        for plugin in self.tape_metadata_plugins:
             plugin_hints = plugin.hints(t_file['metadata'])
 
             t_file = deep_merge_dict(source=plugin_hints, destination=t_file)
@@ -1478,7 +1492,7 @@ class FTS3Transfertool(Transfertool):
                                                                                                                                    job_response['http_message'] if 'http_message' in job_response else None))
         return responses
 
-    def __query_details(self, transfer_id):
+    def __query_details(self, transfer_id: Optional[dict[str, Any]]) -> Optional[str]:
         """
         Query the detailed status of a transfer in FTS3 via JSON.
 
--- a/lib/rucio/transfertool/fts3_plugins.py
+++ b/lib/rucio/transfertool/fts3_plugins.py
@@ -12,37 +12,35 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
 import logging
-from typing import Callable, Optional
 import json
 import sys
+from configparser import NoSectionError
+from typing import Any, Callable, Optional
 
 from rucio.core.plugins import PolicyPackageAlgorithms
 from rucio.common.config import config_get_int, config_get_items
-from configparser import NoSectionError
-
 from rucio.common.exception import InvalidRequest
 
 
-class FTS3MetadataPlugin(PolicyPackageAlgorithms):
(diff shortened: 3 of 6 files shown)


lib/rucio/transfertool/fts3.py
```diff
@@ -797,6 +798,9 @@ def __init__(self, external_host, oidc_account=None, oidc_support: bool = False,
         self.default_lifetime = default_lifetime
         self.archive_timeout_override = archive_timeout_override
 
+        plugins = config_get("transfers", "plugins", False, "[]").strip("[]").split(",")
```
rcarpa: You probably want to use config_get_list here

lib/rucio/transfertool/fts3.py
```diff
@@ -936,6 +940,21 @@ def _file_from_transfer(self, transfer, job_params):
...
+        for plugin in self.plugins:
+            plugin_hints = PolicyPackageAlgorithms._get_one_algorithm(
+                "fts3_plugins", plugin
+            )(t_file["metadata"])
+            self.logger(logging.INFO, f"{plugin} added to transfer")
+
+            # Update the t_file
+            # Support up to 2 levels of further updates
+
+            for key in plugin_hints.keys():
```
rcarpa: directly iterate keys and values using items().
Ideally, implement a `deep_merge_dicts` function somewhere in rucio/coomon/utils and call it here. It will make the code easier to read rather than try to understand what exactly is happening here.

lib/rucio/transfertool/fts3_plugins.py
```diff
@@ -0,0 +1,121 @@
...
+from rucio.core.plugins import PolicyPackageAlgorithms
+from rucio.common.config import config_get, config_get_int
+
+from rucio.common.exception import InvalidRequest
+
+
+class FTS3Plugin(PolicyPackageAlgorithms):
+    """
+    Generic plugin for fts3
+    Default plugin, appends an empty dict to the job params
```
rcarpa: I don't see any job params being updated. You meant file params?

lib/rucio/transfertool/fts3.py
```diff
@@ -936,6 +940,21 @@ def _file_from_transfer(self, transfer, job_params):
             if isinstance(activity_id, int):
                 t_file['scitag'] = self.scitags_exp_id << 6 | activity_id
 
+        for plugin in self.plugins:
+            plugin_hints = PolicyPackageAlgorithms._get_one_algorithm(
```
rcarpa: Why are we calling a private method here ? Necessary work must be done to make it public if it should be called here.

lib/rucio/transfertool/fts3_plugins.py
```diff
@@ -0,0 +1,121 @@
...
+    def _query_session_config(self, activity: str) -> dict[str, str]:
+        """Access the prority for an activity set in config
+
+        Args:
+            activity (str): did activity
+
+        Returns:
+            dict: {'prority': 'Activity Transfer Prority'}
+        """
+        prority = config_get(
```
rcarpa: If we start calling config_get functions for each transfer, the database session must be passed through. This will require a refactoring of the whole transfertool interface.

lib/rucio/transfertool/fts3_plugins.py
```diff
@@ -0,0 +1,121 @@
...
+        return self.set_in_hints(*hint_kwargs)
+
+
+class FTS3SchedulingHints(FTS3Plugin):
+    """Query the config file to determine the priority of a transfer based on activity"""
+
+    def __init__(self):
+        super().__init__()
+
+        self.default_prority = config_get(
```
rcarpa: Isn't it an int?

tests/test_fts3_transfer_tools.py
```diff
@@ -0,0 +1,397 @@
...
+
+    # Produce a new did
+    mock_did = did_factory.random_file_did()
+
+    # Make the transfer path
+    transfer_path = _make_transfer_path(mock_did, rse_factory, root_account)
+
+    activity = transfer_path[0].rws.activity
+    _make_config_section()
+    config_set(section="tape_prority", option=activity, value="100")
```
rcarpa: Setting config values in tests like that can impact all other tests running in parallel. Check file_config_mock and core_config_mock fixtures.

tests/test_fts3_transfer_tools.py
```diff
@@ -0,0 +1,397 @@
...
+    config_set(
+        section="transfers",
+        option="plugins",
+        value="[FTS3SchedulingHints]",
+    )
+    FTS3SchedulingHints()
+
+    # Mock Transfer Tool
+    fts3_tool = FTS3Transfertool(TEST_FTS_HOST)
+    assert "FTS3SchedulingHints" in fts3_tool.plugins
```
rcarpa: I, personally, am a  very strong opponent of tests which verify the status of internal data structures. If we rename self.plugins to self.metadata_plugins later, we have to also update tests. Any simple refactoring will require updating the tests. This way, if you introduce bugs, you end also updating the test to fit the new bugs. Making the test useless. I advise to update some tests in `test_conveyor.py` by setting the correct configuration values using file_config_mock and core_config_mock fixtures. Following by checking that FTS (not rucio!) does what it's intended to do.

rcarpa (dismissed):

tests/test_fts3_transfer_tools.py
voetberg: Ah, didn't know about those. Changing it now.

tests/test_fts3_transfer_tools.py
voetberg: Honestly it's just how I was taught to do tests and sort of did it uncritically, but your logic makes sense. They also end up being redundant, so, removed. I was struggling with figuring out how to verify fts is correctly using the metadata, but I will check out `test_conveyor`, thanks!

lib/rucio/transfertool/fts3.py
voetberg: Yep, totally right there. I made some changes to avoid this.

lib/rucio/transfertool/fts3.py
voetberg: Didn't know that existed! Useful, thanks.

tests/test_fts3_transfer_tools.py
voetberg: So I did start adding conveyor tests, but I consistently get an error `root:fts3.py:1015 Could not submit transfer to https://fts:8446 - HTTPSConnectionPool(host='fts', port=8446): Max retries exceeded with url: /jobs (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x7f0e9c2d36d0>: Failed to establish a new connection: [Errno -2] Name or service not known'))`
Is there is a known work-around or known way to properly connect to that fts server that I'm missing?

rcarpa: How are you running the tests ? You need to run them from within the docker-compose env: https://rucio.github.io/documentation/operator/setting_up_demo/#using-the-environment-including-storage 
(also see the previous section on how to run the tests). If you want to use it from a GUI and have a paid intellij licence, you can use the quite neat integration (only tested on linux): https://rucio.github.io/documentation/developer/setting_up_intellij_dev_env; or (a bit more complex to set-up, but free) vscode alternative: https://rucio.github.io/documentation/developer/setting_up_vscode_dev_env

voetberg: Oh, yeah I've been using the vscode integration, no worries. Never mind though, looks like I was just mis-reading some of the error messages.

lib/rucio/transfertool/fts3_plugins.py
```diff
@@ -0,0 +1,152 @@
...
+import sys
+
+from rucio.core.plugins import PolicyPackageAlgorithms
+from rucio.common.config import config_get_int, config_get_items
+from configparser import NoSectionError
+
+from rucio.common.exception import InvalidRequest
+
+
+class FTS3MetadataPlugin(PolicyPackageAlgorithms):
```
bari12: Let's rename this to `FTS3TapeMetadataPlugin`, I think this is more accurate, since it is not meant for all kinds of metadata.

lib/rucio/transfertool/fts3.py
```diff
@@ -797,6 +799,9 @@ def __init__(self, external_host, oidc_account=None, oidc_support: bool = False,
         self.default_lifetime = default_lifetime
         self.archive_timeout_override = archive_timeout_override
 
+        plugins = config_get_list("transfers", "plugins", False, "[]")
+        self.plugins = [FTS3MetadataPlugin(plugin.strip(" ")) for plugin in plugins]
```
bari12: I wonder if we would ever have the case that this would be multiple different plugins.

lib/rucio/transfertool/fts3_plugins.py
```diff
@@ -0,0 +1,152 @@
...
+    """
+    Add a "archive_metadata" field to a file's transfer parameters.
+    Plugins are registered during initialization and called during a transfer with FTS3
+    """
+
+    ALGORITHM_NAME = "fts3_plugins"
```
bari12: Let's also go for `fts3_tape_metadata_plugins` here

lib/rucio/transfertool/fts3_plugins.py
```diff
@@ -0,0 +1,152 @@
...
+        """
+        return {"collocation_hints": collocation_func(*hints)}
+
+    def _test_collocation(self, *hint: dict) -> dict:
+        return {"0": "", "1": "", "2": "", "3": ""}
+
+    def _default(self, *hints: dict) -> dict:
+        return {}
+
+    def _cms_collocation(self, *hints: dict) -> None:
```
bari12: Remove, I don't see why this would need to be mentioned in the abstract class. Should only be in the CMS specific implementation.

lib/rucio/transfertool/fts3_plugins.py
```diff
@@ -0,0 +1,152 @@
...
+        :param init_func: Initialization requirements for the plugin, defaults to None
+        :type init_func: Optional[Callable], optional
+        """
+        super()._register(cls.ALGORITHM_NAME, algorithm_dict={name: func})
+        if init_func is not None:
+            super()._register(cls.HINTS_NAME, algorithm_dict={name: init_func})
+
+    def _init_activity_hints(self) -> None:
+        try:
+            self.prority_table = dict(config_get_items("tape_priority"))
```
bari12, line 82: Maybe also some more descriptive name here

bari12 (changes requested):
This looks very good and is almost gtg 👍 
I added some very minor comments.
@rdimaio can you please have a look if there is anything to change in respect to the typing.

bari12:
Oh and please rebase :-)

lib/rucio/common/utils.py
```diff
@@ -1944,3 +1944,16 @@ def _wrapper(*args, **kwargs):
                 time.sleep(wait_fixed / 1000.0)
         return _wrapper
     return _decorator
+
+
+def deep_merge_dict(source: dict, destination: dict) -> dict:
+    """Merge two dictionaries together recurively"""
```
rdimaio: typo here, "recurively" -> "recursively"

lib/rucio/transfertool/fts3.py
```diff
@@ -295,7 +297,7 @@ def _pick_fts_checksum(
     return checksum_to_use
 
 
-def build_job_params(transfer_path, bring_online, default_lifetime, archive_timeout_override, max_time_in_queue, logger):
+def build_job_params(transfer_path: list, bring_online: Optional[int], default_lifetime: Optional[int], archive_timeout_override: Optional[int], max_time_in_queue: Optional[dict], logger: Callable) -> dict[str, Any]:
```
rdimaio: `transfer_path: list[DirectTransfer]` might be more accurate (I'm assuming that's the right type based on these two instances: https://github.com/rucio/rucio/blob/48de26679d1fb405745e1cc02963a9faf520c5bb/lib/rucio/transfertool/fts3.py#L247 and https://github.com/rucio/rucio/blob/48de26679d1fb405745e1cc02963a9faf520c5bb/lib/rucio/core/transfer.py#L281)

lib/rucio/transfertool/fts3.py
```diff
@@ -388,8 +390,8 @@ def build_job_params(transfer_path, bring_online, default_lifetime, archive_time
     return job_params
 
 
-def bulk_group_transfers(transfer_paths, policy='rule', group_bulk=200, source_strategy=None, max_time_in_queue=None,
-                         logger=logging.log, archive_timeout_override=None, bring_online=None, default_lifetime=None):
+def bulk_group_transfers(transfer_paths: list, policy: str = 'rule', group_bulk: int = 200, source_strategy: Optional[str] = None, max_time_in_queue: Optional[dict] = None,
```
rdimaio: `transfer_paths: list[list[DirectTransfer]]` might be more accurate (I'm assuming that's the right type based on these two instances: https://github.com/rucio/rucio/blob/48de26679d1fb405745e1cc02963a9faf520c5bb/lib/rucio/transfertool/fts3.py#L247 and https://github.com/rucio/rucio/blob/48de26679d1fb405745e1cc02963a9faf520c5bb/lib/rucio/core/transfer.py#L281)

lib/rucio/transfertool/fts3.py
rdimaio: For the `Optional` arguments, it might be worth adding `None` as default value (unless it's expected that this function will always have those parameters passed in, even when they're `None`).

lib/rucio/transfertool/fts3.py
```diff
@@ -388,8 +390,8 @@ def build_job_params(transfer_path, bring_online, default_lifetime, archive_time
...
+                         logger: Callable = logging.log, archive_timeout_override: Optional[int] = None, bring_online: Optional[int] = None, default_lifetime: Optional[int] = None):
```
rdimaio: missing return type `list[dict[str, any]]`

As a consideration - it might be worth making `{'transfers': ..., 'job_params': ...}` into a custom `TypedDict` for typing purposes (but I wouldn't necessarily do this if this type of dict is not being used anywhere other than here)

lib/rucio/transfertool/fts3.py
```diff
@@ -509,10 +511,10 @@ def __str__(self):
         return f'Transfer {self._transfer_id} of {self._file_metadata["scope"]}:{self._file_metadata["name"]} ' \
                f'{self._file_metadata["src_rse"]} --({self._file_metadata["request_id"]})-> {self._file_metadata["dst_rse"]}'
 
-    def initialize(self, session, logger=logging.log):
+    def initialize(self, session: "Session", logger: Callable = logging.log):
```
rdimaio: please add `-> None` for the return type

rdimaio: Side-note: you can use the [`@abstractmethod` decorator](https://docs.python.org/3/library/abc.html#abc.abstractmethod) to mark this method as abstract, in order to enforce its implementation in child classes

lib/rucio/transfertool/fts3.py
```diff
@@ -509,10 +511,10 @@ def __str__(self):
...
         raise NotImplementedError(f"{self.__class__.__name__} is abstract and shouldn't be used directly")
 
-    def get_monitor_msg_fields(self, session, logger=logging.log):
+    def get_monitor_msg_fields(self, session: "Session", logger: Callable = logging.log):
```
rdimaio: return type: `dict[str, Any]`

lib/rucio/transfertool/fts3.py
```diff
@@ -531,7 +533,7 @@ def get_monitor_msg_fields(self, session, logger=logging.log):
     def _transfer_link(self):
         return '%s/fts3/ftsmon/#/job/%s' % (self._fts_address.replace('8446', '8449'), self._transfer_id)
 
-    def _find_attribute_updates(self, request, new_state, reason, overwrite_corrupted_files):
+    def _find_attribute_updates(self, request: dict, new_state: RequestState, reason: str, overwrite_corrupted_files: Optional[bool]):
```
rdimaio: I would change `overwrite_corrupted_files: Optional[bool]` to `overwrite_corrupted_files: bool = False`

return type: `Optional[dict[str, Any]])`

lib/rucio/transfertool/fts3.py
```diff
@@ -541,7 +543,7 @@ def _find_attribute_updates(self, request, new_state, reason, overwrite_corrupte
                     attributes['overwrite'] = True
         return attributes
 
-    def _find_used_source_rse(self, session, logger):
+    def _find_used_source_rse(self, session: "Session", logger: Callable):
```
rdimaio: return type: `tuple[Optional[str], Optional[str]]`

lib/rucio/transfertool/fts3.py
```diff
@@ -559,7 +561,7 @@ def _find_used_source_rse(self, session, logger):
         return meta_rse_name, meta_rse_id
 
     @staticmethod
-    def _dst_file_set_and_file_corrupted(request, dst_file):
+    def _dst_file_set_and_file_corrupted(request: dict, dst_file: dict):
```
rdimaio: return type: `bool`

lib/rucio/transfertool/fts3.py
```diff
@@ -572,7 +574,7 @@ def _dst_file_set_and_file_corrupted(request, dst_file):
         return False
 
     @staticmethod
-    def _dst_file_set_and_file_correct(request, dst_file):
+    def _dst_file_set_and_file_correct(request: dict, dst_file: dict):
```
rdimaio: return type: `bool`

lib/rucio/transfertool/fts3.py
```diff
@@ -629,7 +631,7 @@ def __init__(self, external_host, request_id, fts_message):
         self._src_url = fts_message.get('src_url', None)
         self._dst_url = fts_message.get('dst_url', None)
 
-    def initialize(self, session, logger=logging.log):
+    def initialize(self, session: "Session", logger: Callable = logging.log):
```
rdimaio: return type: `None`

lib/rucio/transfertool/fts3.py
```diff
@@ -1454,7 +1464,7 @@ def __bulk_query_responses(self, jobs_response, requests_by_eid):
                                                                                                                                    job_response['http_message'] if 'http_message' in job_response else None))
         return responses
 
-    def __query_details(self, transfer_id):
+    def __query_details(self, transfer_id: str):
```
rdimaio: return type: `Optional[dict[str, Any]]`

lib/rucio/transfertool/fts3.py
voetberg: `fts3tape_metadata_plugins` would probably be better for future proofing (and legibility)

lib/rucio/transfertool/fts3_plugins.py
```diff
@@ -0,0 +1,152 @@
...
+
+        self.set_in_hints = self._get_one_algorithm(self.ALGORITHM_NAME, name=policy_algorithm)
+
+    @classmethod
+    def register(cls, name: str, func: Callable, init_func: Optional[Callable] = None) -> None:
+        """
+        Register a fts3 transfer plugin
+
+        :param name: name to register under
+        :type name: str
```
rdimaio: I think the `type` notes in the docstring might be unnecessary given that the types are already mentioned in the function arguments

lib/rucio/transfertool/fts3_plugins.py
```diff
@@ -0,0 +1,152 @@
...
+        if init_func is not None:
+            super()._register(cls.HINTS_NAME, algorithm_dict={name: init_func})
+
+    def _init_activity_hints(self) -> None:
+        try:
+            self.prority_table = dict(config_get_items("tape_priority"))
+        except NoSectionError:
+            self.prority_table = {}
+
+    def _activity_hints(self, activity_kwargs: dict[str, str], default_prority: str = '20') -> dict[str, dict]:
```
rdimaio, line 86: return type `dict[str, dict[str, str]]` would be more accurate

lib/rucio/transfertool/fts3_plugins.py
```diff
@@ -0,0 +1,152 @@
...
+        :param collocation_func: Function that defines the collocation rules
+        :type collocation_func: Callable
+        :param hints: kwargs utilized by the collocation rules
+        :type hints: dict
+        :return: Collocation hints produced by the collocation_func, wrapped
+        :rtype: dict
+        """
+        return {"collocation_hints": collocation_func(*hints)}
+
+    def _test_collocation(self, *hint: dict) -> dict:
```
rdimaio: a more accurate return type: `dict[str, str]`

lib/rucio/transfertool/fts3_plugins.py
```diff
@@ -0,0 +1,152 @@
...
+        return {"0": "", "1": "", "2": "", "3": ""}
+
+    def _default(self, *hints: dict) -> dict:
```
rdimaio, line 112: if this is meant to be implemented in the child classes, it might be worth using the [`@abstractmethod` decorator](https://docs.python.org/3/library/abc.html#abc.abstractmethod)

lib/rucio/transfertool/fts3_plugins.py
```diff
@@ -0,0 +1,152 @@
...
+            assert sys.getsizeof(hints_json) < self.transfer_limit
+
+        except TypeError as e:
+            raise InvalidRequest("Request malformed, cannot encode to JSON", e)
+        except AssertionError as e:
+            raise InvalidRequest(
+                f"Request too large, decrease to less than {self.transfer_limit}", e
+            )
+
+    def hints(self, hint_kwargs: dict) -> dict:
```
rdimaio: return type `dict[str, Any]` would be more accurate

tests/test_transfer_plugins.py
```diff
@@ -0,0 +1,271 @@
...
+        logger=logging.log,
+    )
+
+    # Get the job params used for each transfer
+    transfer_params = transfer_path[0]
+    transfer_params.rws.activity = "Not A Real Activity"
+
+    job_params = fts3_tool._file_from_transfer(transfer_params, job_params)
+
+    # Exstract the hints
```
rdimaio: typo: "Exstract" -> "Extract"

tests/test_transfer_plugins.py
```diff
@@ -0,0 +1,271 @@
...
+        bring_online=None,
+        default_lifetime=None,
+        archive_timeout_override=None,
+        max_time_in_queue=None,
+        logger=logging.log,
+    )
+    # Get the job params used for each transfer
+    job_params = fts3_tool._file_from_transfer(transfer_path[0], job_params)
+
+    # Exstract the hints
```
rdimaio: Typo: "Exstract" -> "Extract"

lib/rucio/transfertool/fts3.py
voetberg: Shouldn't that be `typing.List[DirectTransfer]`? (I was just reviewing the typing guide and noticed....)

tests/test_transfer_plugins.py
voetberg: 😅

lib/rucio/transfertool/fts3.py
voetberg: That seems to be right, according to my ide at least

lib/rucio/transfertool/fts3.py
rdimaio: I assume the typing guide was written before Python 3.9, when they [added support for type hinting using standard collections](https://docs.python.org/3/whatsnew/3.9.html#type-hinting-generics-in-standard-collections); @bari12 could you confirm?

(source: https://stackoverflow.com/questions/39458193/using-list-tuple-etc-from-typing-vs-directly-referring-type-as-list-tuple-etc)

dchristidis: Yes, the guide is unfortunately out of date. See also #6276.

bari12: Correct

lib/rucio/transfertool/fts3.py
bari12: What I meant here is: Would it ever happen that a community loads multiple plugins? I was operating under the assumption it would be 0 or 1.

bari12:
The multi_vo tests seem to be failing now. Please investigate

voetberg:
> The multi_vo tests seem to be failing now. Please investigate

According to the logs they look like they're failing before they even get to the tests (Looks like the docker-compose hits an error with getting the activemq image.)

As for the integration test - Seems unrelated? [The version on fork passed without problem. ](https://github.com/voetberg/rucio/actions/runs/8098623093/job/22132464803)

rdimaio:
Looking into the autotest failures (these two for reference, [1](https://github.com/rucio/rucio/actions/runs/8098624010/job/22185879591?pr=6445) and [2](https://github.com/voetberg/rucio/actions/runs/8143245286/job/22254672967))

1. `test.sh` starts: `docker exec c37ff6ff8585cb8d_rucio_1 ./tools/test/test.sh`
2. gets to `Waiting for httpd`
3. Exits: `** Process '['docker', 'exec', 'c37ff6ff8585cb8d_rucio_1', './tools/test/test.sh']' exited with code 23 {'DIST': 'alma9', 'PYTHON': '3.9', 'IMAGE_IDENTIFIER': 'autotest', 'SUITE': 'multi_vo', 'RDBMS': 'postgres14', 'RUCIO_HOME': '/opt/rucio/etc/multi_vo/tst', 'IMAGE': 'ghcr.io/voetberg/rucio/rucio-autotest:alma9-python3.9'}`

For `multi_vo`, the `wait_for_httpd` line is here: https://github.com/rucio/rucio/blob/master/tools/test/test.sh#L115, followed by
```
httpd -k restart
tools/run_multi_vo_tests_docker.sh
```

`tools/run_multi_vo_tests_docker.sh` echoes `Clearing memcache` as initial message: https://github.com/rucio/rucio/blob/master/tools/run_multi_vo_tests_docker.sh#L39 

so it seems like the`test.sh` fails before entering `tools/run_multi_vo_tests_docker.sh`.

Here's an example of this `multi_vo` test for another PR: https://github.com/rucio/rucio/actions/runs/7755635332/job/21151415317#step:4:84 - here `tools/run_multi_vo_tests_docker.sh` is entered successfully

rdimaio:
There are only two commands running between `echo 'Waiting for httpd'` and `tools/run_multi_vo_tests_docker.sh`:

1. `curl --output /dev/null --silent --retry 15 --retry-all-errors --retry-delay 1 -k https://localhost/ping`
2. `httpd -k restart`

`httpd` has no exit code `23`: https://nightlies.apache.org/httpd/trunk/doxygen/group__HTTP__Status.html

`curl` has exit code `23`: https://curl.se/docs/manpage.html and https://everything.curl.dev/cmdline/exitcode
> Write error. Curl could not write data to a local filesystem or similar. curl receives data chunk by chunk from the network and it stores it like at (or writes it to stdout), one piece at a time. If that write action gets an error, this is the exit status.

So the issue is occurring here: https://github.com/rucio/rucio/blob/master/tools/test/test.sh#L34

voetberg:
That makes sense, but I suppose the question is - why is this only an issue on this PR that doesn't (or shouldn't rather), modify that at all? I'll admit I can't think of a reason, especially when it fails consistently on both my fork and this PR running the canonical version of the tests.

rdimaio:
I pulled your branch and pushed it to my upstream fork, just to see if maybe it was some weird issue with the fork itself - I get the same failure on my fork: https://github.com/rdimaio/rucio/actions/runs/8144295087/job/22258118979

I made a commit to remove `--output /dev/null --silent` (commit for reference: https://github.com/rdimaio/rucio/commit/79e109cb716f943607be853ebd56c2c92153701e), here's the test run: https://github.com/rdimaio/rucio/actions/runs/8144460703/job/22258608666 - it works here, so the issue is definitely with outputting the curl output to `/dev/null`. I have no idea why this is only affecting this branch, but that would be a quick fix to resolve it.

Side note - the test logic involved here (e.g. the `curl` command) is related to the token infrastructure changes (https://github.com/rucio/rucio/blame/master/tools/test/test.sh#L115), so it might be a good idea to get someone that worked on that to share their opinion on this (@bari12 tagging you in case you know anyone other than Radu who could have some insight here)

voetberg:
Thanks @rdimaio ! I was about to put in a PR for the same. Glad you were able to replicate it on your fork - I was loosing my mind. 

The only thing I can think that causes is maybe running out of disk on github's ends, as the mutli-vo tests are the last things run and is a larger commit? Might be nonsense but besides for a permission error that's one of the few things that can cause that `23` error code.

bari12 (approved):

PR #6446, opened 2024-01-10
Transfers: get rid of legacy_sources

rcarpa:
This code was inherited all the way from before the conveyor refactoring, when we used tuples as a workaround to pass information through the code. Now we pass directly structured objects which can be accessed as needed.


dchristidis (changes requested):
Would you mind adding hints for the return types of the two `source_url()` methods?

dchristidis (approved):

PR #6448, opened 2024-01-10
Docker & Kubernetes: Specify amd64 platform on images with no arm64 support #6447

### LINKED ISSUE #6447: arm64 architecture and Docker containers
At the moment, at least some of the Rucio containers (e.g. https://hub.docker.com/r/rucio/fts) are only built on linux/amd64. When pulling these on arm64 machines, they error out with `no matching manifest for linux/arm64/v8 in the manifest list entries`.

This can be fixed manually:
- on a container-by-container basis by specifying `--platform linux/amd64` in the `docker pull` command
- on a machine-wide basis by setting an env variable `export DOCKER_DEFAULT_PLATFORM=linux/amd64`

This could be specified in the documentation, but for ease of development (e.g. for developers on ARM-based architecture like Apple Silicon) it might be worth addressing this on our end. 

The simplest fix is to modify the Docker compose file(s) to specify `platform: linux/amd64` under each of the involved containers. We could also [build multi-platform images](https://docs.docker.com/build/building/multi-platform/) to add direct support for arm64, but there are still some containers used by Rucio that do not support arm64 (e.g. `docker.io/elasticsearch:7.4.0`:  https://hub.docker.com/_/elasticsearch/tags?page=1&name=7.4.0).

rdimaio:
Setting `platform: linux/amd64` for images without `arm64` support in order to support the Docker compose file in ARM-based environments (fixes #6447)

### DIFF
--- a/etc/docker/dev/docker-compose.yml
+++ b/etc/docker/dev/docker-compose.yml
@@ -72,8 +72,7 @@ services:
       - DOCKER_INFLUXDB_INIT_BUCKET=rucio
       - DOCKER_INFLUXDB_INIT_ADMIN_TOKEN=mytoken
   elasticsearch:
-    image: docker.io/elasticsearch:7.4.0
-    platform: linux/amd64
+    image: docker.io/elasticsearch:7.8.0
     environment:
       - discovery.type=single-node
   activemq:
--- a/etc/docker/dev/docker-compose.yml
+++ b/etc/docker/dev/docker-compose.yml
@@ -29,6 +29,7 @@ services:
       - RDBMS
   rucio:
     image: docker.io/rucio/rucio-dev:latest-alma9
+    platform: linux/amd64
     entrypoint: ["/rucio_source/etc/docker/dev/rucio/entrypoint.sh"]
     command: ["httpd","-D","FOREGROUND"]
     volumes:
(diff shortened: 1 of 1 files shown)


maany:
Although I appreciate the call to arms(64) for our brothers at Apple, however, we should not force everyone to use `arm64` images, just because we have M1's.

I would rather mention the required change in the `rucio/documentation` repo as a note for developer's using Apple Silicone chips.

Could you send a PR in the documentation repo instead and close this one?

rdimaio:
> Although I appreciate the call to arms(64) for our brothers at Apple, however, we should not force everyone to use `arm64` images, just because we have M1's.
> 
> I would rather mention the required change in the `rucio/documentation` repo as a note for developer's using Apple Silicone chips.
> 
> Could you send a PR in the documentation repo instead and close this one?

It's the other way around - with this change, `arm64` machines will pull `amd64` images, instead of just erroring out because `arm64` is not in the manifest list entries. This change only affects ARM-based machines

rdimaio:
Tests failed for initial commit, seems to be related to the `elasticsearch` container. Bumped it up to the nearest minor release that supports arm64 (7.8.0): https://hub.docker.com/_/elasticsearch/tags?page=1&name=7.8.0 and removed the `platform: linux/amd64` line

rcarpa (dismissed):

rcarpa:
Please rebase rather than merge the branches.

rcarpa (approved):

Issue #5522, opened 2022-05-04
Resolve Warnings in tests


joeldierkes:
Motivation
----------
Rucio is throwing Warnings in the tests. We should resolve them.
https://github.com/rucio/rucio/runs/6283942231?check_suite_focus=true#step:4:3116
```
=============================== warnings summary ===============================
../core/did_meta_plugins/filter_engine.py:362
  /opt/rucio/lib/rucio/core/did_meta_plugins/filter_engine.py:362: DeprecationWarning: invalid escape sequence \_
    expression = json_column[key].as_string().notlike(value.replace('*', '%').replace('_', '\_'), escape='\\')  # NOQA: W605

../core/did_meta_plugins/filter_engine.py:360
  /opt/rucio/lib/rucio/core/did_meta_plugins/filter_engine.py:360: DeprecationWarning: invalid escape sequence \_
    expression = json_column[key].as_string().like(value.replace('*', '%').replace('_', '\_'), escape='\\')     # NOQA: W605

../core/did_meta_plugins/filter_engine.py:340
  /opt/rucio/lib/rucio/core/did_meta_plugins/filter_engine.py:340: DeprecationWarning: invalid escape sequence \_
    expression = key.notlike(value.replace('*', '%').replace('_', '\_'), escape='\\')  # NOQA: W605

../core/did_meta_plugins/filter_engine.py:338
  /opt/rucio/lib/rucio/core/did_meta_plugins/filter_engine.py:338: DeprecationWarning: invalid escape sequence \_
    expression = key.like(value.replace('*', '%').replace('_', '\_'), escape='\\')     # NOQA: W605
...
```

Modification
------------
Just resolve the Warnings.

PR #6450, opened 2024-01-11
Testing: fix test warnings

rdimaio:
Issue: #5522

Fixing test warning:
```
/opt/venv/lib64/python3.9/site-packages/_pytest/python.py:198: PytestReturnNotNoneWarning: Expected None, but tests/test_import_export.py::test_active returned True, which will be an error in a future version of pytest.  Did you mean to use `assert` instead of `return`?
```

### DIFF
--- a/tests/test_import_export.py
+++ b/tests/test_import_export.py
@@ -81,11 +81,6 @@ def reset_rses():
     db_session.commit()
 
 
-def test_active():
-    db_session = session.get_session()
-    assert db_session.bind.dialect.name != 'sqlite'
-
-
 @pytest.fixture
 def importer_example_data(vo):
     if not config_has_section('importer'):
--- a/tests/test_import_export.py
+++ b/tests/test_import_export.py
@@ -83,9 +83,7 @@ def reset_rses():
 
 def test_active():
     db_session = session.get_session()
-    if db_session.bind.dialect.name == 'sqlite':
-        return False
-    return True
+    assert db_session.bind.dialect.name != 'sqlite'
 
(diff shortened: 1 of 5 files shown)


rdimaio:
Regarding https://github.com/rucio/rucio/pull/6450/commits/aaf52e578117b81f00faa920711c1f45c872260d - `test_active` fails the autocheck that runs on `sqlite`: https://github.com/rucio/rucio/actions/runs/7487468726/job/20380554506?pr=6450.

It's a bit unclear what the purpose of `test_active` is; @maany mentioned that this file was discussed in the past, will check to see what to do next

tests/test_import_export.py
```diff
@@ -83,9 +83,7 @@ def reset_rses():
 
 def test_active():
```
maany: My recommendation: remove the test altogether. It does not do anything useful. On the contrary, it tries to do things the testing framework was designed for i.e. selectively run/skip tests based on the environment.

Looking back at the commit where this  test was added, it looks more like a debug util for the developer working on the issue, that slipped through the cracks.
 
https://github.com/rucio/rucio/commit/e6bc1b3d6ba3c5f53ba85d425b12cf5ef4b28c1

maany (changes requested):
Remove the test altogether

tests/test_import_export.py
rdimaio: thanks for checking :) I'll go ahead and take it out

maany (approved):
LGTM

PR #6453, opened 2024-01-12
bug: Fix prepare-commit-msg to work on both GNU and BSD sed

### LINKED ISSUE #6452: prepare-commit-msg: `sed` in-place errors out on BSD `sed`
### Description

The `prepare-commit-msg` script (https://github.com/rucio/rucio/blob/master/tools/prepare-commit-msg) works with GNU `sed`, but does not work with BSD `sed` (also used on macOS), causing the error:

```
sed: 1: ".git/COMMIT_EDITMSG": invalid command code .
```

### Steps to reproduce

1. BSD-based system
2. Run `git commit...`

### Rucio Version

_No response_

### Additional Information

Will push a PR for this now

rdimaio:
fixes #6452

From https://unix.stackexchange.com/questions/401905/bsd-sed-vs-gnu-sed-and-i:
>the BSD build of sed mandates the extension for the backup file with -i, rather than it being optional, as in GNU sed.

Fixing the prepare-commit-msg script so that it works for both BSD and GNU based `sed`

See also:
- https://stackoverflow.com/questions/5694228/sed-in-place-flag-that-works-both-on-mac-bsd-and-linux
- https://unix.stackexchange.com/questions/92895/how-can-i-achieve-portability-with-sed-i-in-place-editing

### DIFF
--- a/tools/prepare-commit-msg
+++ b/tools/prepare-commit-msg
@@ -23,7 +23,7 @@ else
     if echo $EDITMSG | grep "# Please enter the commit message for your changes.*"
     then
 	echo "No commit message given, preparing template for editing."
-	sed -i "1s/^/\#\<component\>: \<message\>; Fix \#$NUMBER\n/" $1
+	sed -i.bak "1s/^/\#\<component\>: \<message\>; Fix \#$NUMBER\n/" $1 && rm $1.bak
     else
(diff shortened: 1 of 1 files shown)


rcarpa (approved):
Seems that we found a second person who uses these scripts :grin:

Issue #6457, opened 2024-01-15
Add authentication to hermes when submitting to ElasticSearch


thysk:
### Description

Currently there is no way for hermes to submit to an elasticsearch that has authentication needed

### Motivation

Allows hermes to interact with a wider range of ElasticSearch deployments

### Change

Adding the auth parameters to requests.post in submit_to_elastic without enforcing authentication being needed for all instances

Would require adding RUCIO_CFG_HERMES_ELASTICSEARCH_USERNAME and RUCIO_CFG_HERMES_ELASTICSEARCH_PASSWORD to the documentation
Or should this be in MESSAGING_HERMES?

PR #6459, opened 2024-01-15
hermes: add auth to http request for hermes against ES #6457

thysk:
Add auth to hermes when it is attempting to send messages to elastic search via the elastic messages

### DIFF
--- a/lib/rucio/daemons/hermes/hermes.py
+++ b/lib/rucio/daemons/hermes/hermes.py
@@ -48,6 +48,7 @@ from rucio.common.logging import setup_logging
 from rucio.core.message import delete_messages, retrieve_messages
 from rucio.core.monitor import MetricManager
 from rucio.daemons.common import run_daemon
+from requests.auth import HTTPBasicAuth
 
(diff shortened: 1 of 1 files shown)


lib/rucio/daemons/hermes/hermes.py
```diff
@@ -365,10 +366,18 @@ def submit_to_elastic(messages: list[dict], endpoint: str, logger: "Callable") -
     :returns:                  HTTP status code. 200 and 204 OK. Rest is failure.
     """
     text = ""
+    elastic_username = config_get("hermes", "elasticsearch_username",
```
rcarpa: one of the the configuration options used below is `elastic_endpoint`. Should we keep the same naming? `elastic_username` ?

thysk: yes, honestly I thought I had named it that (and it what I have it called on my internal tests)! will make change now

rcarpa (approved):

bari12:
I am merging this as an enhancement in the `33` release line (so not waiting for `34`)

PR #6462, opened 2024-01-15
Transfers: Fix SAWarnings for coerced subqueries #5522

rdimaio:
Fixing these warnings: `SAWarning: Coercing Subquery object into a select() for use in IN(); please pass a select() construct explicitly`


rdimaio:
I'll close this PR for now; I discussed these queries with Radu for a bit and they're a bit risky to modify as they must be versatile for different DB systems, I don't think it's worth risking issues just to fix some warnings

PR #6480, opened 2024-01-22
Transfers: add BitTorrent transfertool. Close #6479. Close #6092

### LINKED ISSUE #6092: submitter doesn't use the configured scheme list

### LINKED ISSUE #6479: implement support for transfers via BitTorrent
Rucio already supports two 3rd party copy drivers: globus and FTS. However, both of these strongly rely on non-standart/wlcg-specific extensions to execute transfers between source and destination RSEs.

What if we could just install an of-the-shelf bittorrent client on each RSE and use its administrative interface to move the desired data. In theory, it would even be possible to perform "true" multi-source transfers this way.

rcarpa:
While most changes could be included as a patch, the commit related to correctly handling the scheme is breaking for everybody who set this config value (it was ignored in the past by the submitter, it's not ignored anymore).


bari12 (approved):

bari12:
🥳

PR #6485, opened 2024-01-24
Create BaseDaemon and refactor Undertaker to inherit from BaseDaemon

rdimaio:
Initial commit where BaseDaemon is created. As an initial test, Undertaker inherits from BaseDaemon. If tests pass, other daemons can also be refactored to inherit from BaseDaemon.


rdimaio:
closing for now to use actions on fork instead of PR

Issue #6478, opened 2024-01-22
Consider refactoring daemons into classes with a base `Daemon` class


rdimaio:
Looking at the Rucio daemons (https://github.com/rucio/rucio/tree/37bb888fe6c0af8cb74cea7f36e087468185cbb8/lib/rucio/daemons), there's a few functions that are in common to many daemons. For example, `run`, `stop` and `run_once` are used in:
- undertaker: https://github.com/rucio/rucio/blob/37bb888fe6c0af8cb74cea7f36e087468185cbb8/lib/rucio/daemons/undertaker/undertaker.py
- transmogrifier: https://github.com/rucio/rucio/blob/37bb888fe6c0af8cb74cea7f36e087468185cbb8/lib/rucio/daemons/transmogrifier/transmogrifier.py
- reaper: https://github.com/rucio/rucio/blob/37bb888fe6c0af8cb74cea7f36e087468185cbb8/lib/rucio/daemons/reaper/reaper.py

It might make sense to abstract these functions into a parent class `BaseDaemon` to avoid replication where possible.

bari12:
Thank you for raising this. We have reviewed this issue as part of our regular issue review.

We see value in this request, but we do not have the capacity to work on it within the next twelve months.

In line with our [issue management guideline](https://rucio.cern/documentation/developer/issue_management), we are labelling it `wishlist` and closing it. This does not mean it is forgotten: wishlist items are revisited regularly during component reviews and roadmap planning, and this issue can be re-opened and added back to the pipeline as soon as interest and capacity align.

Issue #6488, opened 2024-01-26
Add an automatic formatter to the pre-commit hooks


rdimaio:
Right now, the pre-commit script (https://github.com/rucio/rucio/blob/master/.pre-commit-config.yaml) runs `flake8` as linter, warning of style issues. It might be worth also adding an automatic formatter like `black` (https://github.com/psf/black) to fix style issues automatically. 

- Using them together: https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html#flake8

We could also consider using `isort` to sort the imports (https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html#isort)

bari12:
We have talked about this in the past, but could not agree.
Using a formatter now would somewhat complicate git history, but I don't think this is the primary issue, mostly it was that people just did not agree to use one specific formatter.
What we fell back on is that we will just enforce PEP8 rules, and on top of that people can somewhat freely choose how to format their contributions, as long as it is PEP8 compliant.
Re-formatting large parts of the code, without any functional changes, is generally frowned upon though.

rdimaio:
Closing - discussed offline, we might discuss this at the retreat

Issue #1924, opened 2018-12-19
Daemon migration to python3


hahahannes:
Motivation
----------
Make the daemons python3 compatible.

Modification
------------

PR #6489, opened 2024-01-26
Create base Daemon class and refactor Undertaker, Reaper, Abacus Account and Judge Repairer daemons to inherit from base class

rdimaio:
Initial commit where base class Daemon is created, and Undertaker inherits from Daemon. If approved, I will refactor the other daemons in the same way. Tested in local fork.

(note: the changes in `common.py` outside of the main `Daemon` class are due to the `black` formatter adjusting style to meet PEP8)

Initial commit for #6478

### DIFF
--- a/bin/rucio-undertaker
+++ b/bin/rucio-undertaker
@@ -21,7 +21,7 @@ Undertaker is a daemon to manage expired did.
 import argparse
 import signal
 
-from rucio.daemons.undertaker.undertaker import run, stop
+from rucio.daemons.undertaker.undertaker import Undertaker
 
 
 def get_parser():
@@ -66,12 +66,11 @@ Check if the DID exists::
 
 
 if __name__ == "__main__":
-
-    signal.signal(signal.SIGTERM, stop)
     parser = get_parser()
     args = parser.parse_args()
+    undertaker = Undertaker(total_workers=args.total_workers, chunk_size=args.chunk_size, once=args.run_once, sleep_time=args.sleep_time)
+    signal.signal(signal.SIGTERM, undertaker.stop)
     try:
-        run(total_workers=args.total_workers, chunk_size=args.chunk_size, once=args.run_once,
-            sleep_time=args.sleep_time)
+        undertaker.run()
     except KeyboardInterrupt:
-        stop()
+        undertaker.stop()
--- a/bin/rucio-undertaker
+++ b/bin/rucio-undertaker
@@ -70,6 +70,7 @@ if __name__ == "__main__":
     args = parser.parse_args()
     undertaker = Undertaker(total_workers=args.total_workers, chunk_size=args.chunk_size, once=args.run_once, sleep_time=args.sleep_time)
     signal.signal(signal.SIGTERM, undertaker.stop)
+    signal.signal(signal.SIGINT, undertaker.stop)
     try:
         undertaker.run()
     except KeyboardInterrupt:
--- a/lib/rucio/daemons/common.py
+++ b/lib/rucio/daemons/common.py
@@ -19,18 +19,120 @@ import logging
 import os
 import queue
 import socket
+import signal
 import threading
 import time
 from collections.abc import Callable, Generator, Iterator, Sequence
 from typing import Any, Generic, Optional, TypeVar, Union
+from types import TracebackType, FrameType
+from abc import ABC, abstractmethod
 
-from rucio.common.logging import formatted_logger
+import rucio.db.sqla.util
+from rucio.common.logging import formatted_logger, setup_logging
 from rucio.common.utils import PriorityQueue
 from rucio.core import heartbeat as heartbeat_core
 from rucio.core.monitor import MetricManager
+from rucio.common.exception import DatabaseException
 
-T = TypeVar('T')
+T = TypeVar("T")
 METRICS = MetricManager(module=__name__)
+
+
+class Daemon(ABC):
+    """
+    Base daemon class
+    """
+
+    def __init__(
+        self,
+        once: bool = False,
+        total_workers: int = 1,
+        sleep_time: int = 60,
+        partition_wait_time: int = 1,
+        daemon_name: str = "undefined_daemon",
+    ) -> None:
+        """
+        :param once: Only execute one iteration.
+        :param total_workers: Number of total workers.
+        :param sleep_time: Sleep time between daemon iterations.
+        :param partition_wait_time: Time to wait for database partition rebalancing before starting the actual daemon loop.
+        :param daemon_name: Name of daemon that is constructed.
+        """
+        self.once = once
+        self.total_workers = total_workers
+        self.sleep_time = sleep_time
+        self.partition_wait_time = partition_wait_time
+        self.daemon_name = daemon_name
+        self.graceful_stop = threading.Event()
+        signal.signal(signal.SIGTERM, self.stop)
+        setup_logging(process_name=daemon_name)
+
+    @staticmethod
+    def _pre_run_checks() -> None:
+        """
+        Checks to run before daemon execution
+        """
+        if rucio.db.sqla.util.is_old_db():
+            raise DatabaseException("Database was not updated, daemon won't start")
+
+    @abstractmethod
+    def _run_once(self, heartbeat_handler: "HeartbeatHandler", **_kwargs) -> None:
+        """
+        Daemon-specific logic (to be defined in child classes) for a single iteration
+        """
+        pass
+
+    def _call_daemon(self, activities: Optional[list[str]] = None) -> None:
+        """
+        Run the daemon loop and call the function _run_once at each iteration
+        """
+
+        run_once_fnc = functools.partial(self._run_once)
+
+        daemon = db_workqueue(
+            once=self.once,
+            graceful_stop=self.graceful_stop,
+            executable=self.daemon_name,
+            partition_wait_time=self.partition_wait_time,
+            sleep_time=self.sleep_time,
+            activities=activities,
+        )(run_once_fnc)
+
+        for _ in daemon():
+            pass
+
+    def run(self) -> None:
+        """
+        Run the daemon.
+        """
+        self._pre_run_checks()
+
+        if self.once:
+            self._call_daemon()
+        else:
+            logging.info("main: starting threads")
+            threads = [
+                threading.Thread(target=self._call_daemon)
+                for _ in range(0, self.total_workers)
+            ]
+            [t.start() for t in threads]
+            logging.info("main: waiting for interrupts")
+
+            # Interruptible joins require a timeout.
+            while threads[0].is_alive():
+                [t.join(timeout=3.14) for t in threads]
+
+    def stop(
+        self, signum: "Optional[int]" = None, frame: "Optional[FrameType]" = None
+    ) -> None:
+        """
+        Graceful exit the daemon. Used as handler for SIGTERM.
+        The unused parameters are needed for the method to be accepted as handler for signal.signal().
+
+        :param signum: signal number
+        :param frame: stack frame
+        """
+        self.graceful_stop.set()
 
 
 class HeartbeatHandler:
@@ -38,7 +140,7 @@ class HeartbeatHandler:
     Simple contextmanager which sets a heartbeat and associated logger on entry and cleans up the heartbeat on exit.
     """
 
-    def __init__(self, executable: str, renewal_interval: int):
+    def __init__(self, executable: str, renewal_interval: int) -> None:
         """
         :param executable: the executable name which will be set in heartbeats
         :param renewal_interval: the interval at which the heartbeat will be renewed in the database.
@@ -47,7 +149,9 @@ class HeartbeatHandler:
         self.executable = executable
         self._hash_executable = None
         self.renewal_interval = renewal_interval
-        self.older_than = renewal_interval * 10 if renewal_interval and renewal_interval > 0 else None  # 10 was chosen without any particular reason
+        self.older_than = (
+            renewal_interval * 10 if renewal_interval and renewal_interval > 0 else None
+        )  # 10 was chosen without any particular reason
 
         self.hostname = socket.getfqdn()
         self.pid = os.getpid()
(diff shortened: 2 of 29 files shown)


rcarpa (dismissed):

rdimaio:
In https://github.com/rucio/rucio/pull/6489/commits/29a5666a6ccb39332cf4178117217b8b29c55ebf, Reaper is also refactored to follow the base Daemon class. Tests pass on fork branch: https://github.com/rdimaio/rucio/commit/29a5666a6ccb39332cf4178117217b8b29c55ebf

lib/rucio/daemons/common.py
```diff
@@ -23,22 +23,133 @@
 import time
 from collections.abc import Callable, Generator, Iterator, Sequence
 from typing import Any, Generic, Optional, TypeVar, Union
+from types import TracebackType, FrameType
+from abc import ABC, abstractmethod
```
dchristidis: Please sort modules and imported functions/classes alphabetically. This may apply elsewhere as well.

lib/rucio/daemons/common.py
```diff
@@ -23,22 +23,133 @@
...
+        while thread_list:
+            thread_list = [
+                thread.join(timeout=3.14)
+                for thread in thread_list
+                if thread and thread.is_alive()
+            ]
+
+    def stop(
+        self, signum: "Optional[int]" = None, frame: "Optional[FrameType]" = None
+    ) -> None:
```
dchristidis: I see no need for the hints to be quoted.

lib/rucio/daemons/common.py
```diff
@@ -23,22 +23,133 @@
...
+        :param sleep_time: Sleep time between daemon iterations.
+        :param partition_wait_time: Time to wait for database partition rebalancing before starting the actual daemon loop.
+        :param daemon_name: Name of daemon that is constructed.
+        """
+        self.once = once
+        self.total_workers = total_workers
+        self.sleep_time = sleep_time
+        self.partition_wait_time = partition_wait_time
+        self.daemon_name = daemon_name
+        self.graceful_stop = threading.Event()
```
dchristidis: I wonder if it’s useful to use `typing.Final` here.

lib/rucio/daemons/common.py
```diff
@@ -23,22 +23,133 @@
...
+        for _ in daemon():
+            pass
+
+    def run(self) -> None:
+        """
+        Run the daemon.
+        """
+        self._pre_run_checks()
+
+        logging.info(f"{self.daemon_name}: starting threads")
```
dchristidis: It is not recommended to do string interpolation outside and prior to a `logging.log` call. See: https://docs.python.org/3/howto/logging.html#optimization

This may apply elsewhere as well.

lib/rucio/daemons/common.py
```diff
@@ -23,22 +23,133 @@
...
+        thread_list = [
+            threading.Thread(target=self._call_daemon)
+            for _ in range(0, self.total_workers)
+        ]
+        [t.start() for t in thread_list]
```
dchristidis: I never particularly liked this. Pylint should complain with `expression-not-assigned`. Is it maybe a good opportunity to split into two lines?

lib/rucio/daemons/common.py
```diff
@@ -23,22 +23,133 @@
...
+    def run(self) -> None:
+        """
+        Run the daemon.
+        """
+        self._pre_run_checks()
+
+        logging.info(f"{self.daemon_name}: starting threads")
+        thread_list = [
+            threading.Thread(target=self._call_daemon)
+            for _ in range(0, self.total_workers)
```
dchristidis: Is there any usefulness to specifying `start=0`, which is already the default?

lib/rucio/daemons/common.py
```diff
@@ -23,22 +23,133 @@
...
+        if not self.once:
+            logging.info(f"{self.daemon_name}: waiting for interrupts")
+
+        # Interruptible joins require a timeout.
+        while thread_list:
+            thread_list = [
+                thread.join(timeout=3.14)
+                for thread in thread_list
+                if thread and thread.is_alive()
+            ]
```
dchristidis: Firstly, I think placing the comment above the block makes it unclear. Better put it together with line 129.

But more importantly, do you understand this loop? `threading.Thread.join()` always returns `None`, according to its documentation. So under what circumstances could we have multiple iterations?

lib/rucio/daemons/common.py
```diff
@@ -47,7 +158,9 @@ def __init__(self, executable: str, renewal_interval: int):
         self.executable = executable
         self._hash_executable = None
         self.renewal_interval = renewal_interval
-        self.older_than = renewal_interval * 10 if renewal_interval and renewal_interval > 0 else None  # 10 was chosen without any particular reason
+        self.older_than = (
+            renewal_interval * 10 if renewal_interval and renewal_interval > 0 else None
+        )  # 10 was chosen without any particular reason
```
dchristidis: Why is `if renewal_interval` necessary here? The hint is `int`.

lib/rucio/daemons/common.py
```diff
@@ -58,59 +171,90 @@ def __init__(self, executable: str, renewal_interval: int):
...
-                or not self.last_heart_beat \
-                or self.last_time < datetime.datetime.now() - datetime.timedelta(seconds=self.renewal_interval) \
-                or self.last_payload != payload:
+        if (
+            force_renew
+            or not self.last_time
+            or not self.last_heart_beat
+            or self.last_time
+            < datetime.datetime.now()
+            - datetime.timedelta(seconds=self.renewal_interval)
```
dchristidis: I don’t think that I can agree that this improves readability.

lib/rucio/daemons/common.py
```diff
@@ -23,22 +23,133 @@
...
+
+        :param heartbeat_handler: Handler to set and manage the heartbeat for this execution.
+
+        :returns: Tuple of (must_sleep, ret_value).
+                  must_sleep: set to True to signal to the calling context that it must sleep before next execution, False otherwise
+                  ret_value: Daemon-specific return value
+        """
+        pass
+
+    def _call_daemon(self, activities: Optional[list[str]] = None) -> None:
```
dchristidis, line 92: I wonder if we should take the opportunity to rethink some of the chosen names. There’s ‘pre-run’, ‘run’, ‘run once’, ‘call daemon’. Do they feel intuitive? Had I not known, I would have guessed that `_call_daemon()` is the outer function, `_pre_run_checks()` is executed before `run()`, and that `run()` and `_run_once()` are alternatives.

dchristidis (changes requested):
My initial impression was positive; I think the proposed design has a lot of merit. But I would not recommend its inclusion in its current state.

What hinders the review is the mix of functional and aesthetical changes in the same commit. Things like folding long lines, changing indentation levels, and adding type hints to existing functions, are all very welcome, but bundling them together with the actual behavioural changes for PRs as large as this should be discouraged.

Then there’s the change of existing strings to use double instead of single quotes. The Rucio project does not have an official guideline, so the recommendation is to follow the existing pattern when modifying code and using one’s preference when introducing big blocks of new code. But I’m not too enthusiastic about trying to conform an entire existing file. What if the next developer that works on it thinks that single quotes are better? It can also be an annoyance when going through the history of a file.

I stopped halfway through `lib/rucio/daemons/common.py`, without looking too close at the rest. I would be very happy to continue the review, but if ends up being up to me, I think the PR may need to go through multiple iterations before I can approve it.

PS: Afterwords I saw that:

> (note: the changes in common.py outside of the main Daemon class are due to the black formatter adjusting style to meet PEP8)

Black is highly opinionated. Its use induced negative reactions in the past. I would be careful.

rdimaio:
> Things like folding long lines, changing indentation levels, and adding type hints to existing functions, are all very welcome, but bundling them together with the actual behavioural changes for PRs as large as this should be discouraged.

I agree on all points except for type hinting. On the formatting side, I recognize I made a mistake in using a formatter, as it makes it harder to spot the functional changes; I'm happy to do whatever you think it's best to do here, e.g. if you want me to amend the commit so that the formatting changes are removed I'm okay with doing that, but if you have other ideas let me know. I will be careful about this for future PRs.

For the type hinting to existing functions: the functions whose type hints have been changed/added in this PR are related to the daemons that have been refactored (e.g. `parse_expression` is used in Reaper). I added type hints to those functions while refactoring the daemons in order to better understand the logic in the daemons themselves. I understand that these type hints are not part of the "refactoring" itself, but I personally don't see an issue in having them within the same PR, but of course this is just my opinion so I'm okay with separating them if you prefer.

lib/rucio/daemons/common.py
rdimaio: To avoid doing this kind of work manually (which would be a waste of time), would it make sense to enforce something like `isort` as part of the pre-commit hook? https://pycqa.github.io/isort/

dchristidis:
It’s the scale of the PR that makes this matter to me. If it were a few-line change in a single function, that also happened to add type hints, then I would not object to a single commit. This, however, is over one thousand lines and spans multiple files. Hence why I tend to request that behavioural and non-behavioural changes are separated (and type hints are part of the latter). Both are welcome, of course! But as a reviewer, it aids me to know when a commit is not supposed to make any effective changes.

So yes, if Martin expects the final say from me, then I need to ask you to rewrite the commits as described.

lib/rucio/daemons/common.py
dchristidis: That’s an interesting proposal. I would recommend opening a separate issue to discuss this and to consider applying it across the entire codebase beforehand.

rdimaio: I had opened this some weeks ago https://github.com/rucio/rucio/issues/6488 where I mentioned `isort` as well - I will bring it up in the retreat and see what the overall thoughts are. For now, I will make this change manually

lib/rucio/daemons/common.py
rdimaio: Good idea, it's definitely appropriate here, will add it now

lib/rucio/daemons/common.py
rdimaio: I am okay with changing the names if there are any suggestions, but I could not think of anything else at the moment. There are some things that I do not like, e.g. I think `once` and consequently `run_once` are not clear enough, but those names currently match some of the names from the daemon's user's perspective (e.g. the `--run-once` flag) so I didn't want to change them

lib/rucio/daemons/common.py
rdimaio: I think the reference recommends not to interpolate strings when that involves running a potentially expensive function, and accessing an instance attribute would likely not be expensive. 

Is there any alternative to doing this if I want to log the daemon name? Alternatively we can just log "daemon: starting threads" without specifying the daemon name, but that might be confusing if there are multiple daemons running at the same time

lib/rucio/daemons/common.py
rdimaio: That's how it was written in `Undertaker` and I didn't notice it while refactoring. Removed `0`

lib/rucio/daemons/common.py
rdimaio: This whole code block was copied from Reaper:

```
    # Interruptible joins require a timeout.
    while threads_list:
        threads_list = [thread.join(timeout=3.14) for thread in threads_list if thread and thread.is_alive()]
```

In general, every daemon involved in the refactoring had a loop of this kind, e.g. in Undertaker:

```
       # Interruptible joins require a timeout.
        while threads[0].is_alive():
            [t.join(timeout=3.14) for t in threads]
```

Based on the fact that `threading.Thread.join()` returns `None`, then I think the version of the loop present in Undertaker is the one that's actually correct, given that it loops as long as the first thread is alive, right?

lib/rucio/daemons/common.py
rdimaio: That's the way it was written in the refactored daemons, e.g. in Judge Repairer:

```
def stop(signum: "Optional[int]" = None, frame: "Optional[FrameType]" = None) -> None:
    """
    Graceful exit.
    """

    graceful_stop.set()
```

I agree that the quotes are not necessary here, removed.

lib/rucio/daemons/common.py
dchristidis: You don’t have to avoid interpolation altogether, just *defer* it to `logging`. See Pylint’s `logging-fstring-interpolation`. The solution is:

```python
logging.info("%s: starting threads", daemon_name)
```

rdimaio:
> It’s the scale of the PR that makes this matter to me. If it were a few-line change in a single function, that also happened to add type hints, then I would not object to a single commit. This, however, is over one thousand lines and spans multiple files. Hence why I tend to request that behavioural and non-behavioural changes are separated (and type hints are part of the latter). Both are welcome, of course! But as a reviewer, it aids me to know when a commit is not supposed to make any effective changes.
> 
> So yes, if Martin expects the final say from me, then I need to ask you to rewrite the commits as described.

I've rebased the PR to remove all unnecessary non-behavioural changes

lib/rucio/daemons/common.py
rdimaio: Thanks, amended changes to use this instead (also changed other places where I was logging with fstring)

lib/rucio/daemons/common.py
dchristidis: I’m afraid I don’t know. That’s why I’m genuinely asking, because you’re (rightfully) copying existing implementations, and maybe they copied it over from somewhere else, but I wonder if anyone really understands them.

So, since the long-term goal of your effort is to implement this exactly once, I am taking the opportunity to try to understand exactly what it is supposed to do.

dchristidis: The more I look at the existing implementations, the more perplexed I am. Won’t the loop in the Reaper exit after one iteration, while the threads are still running?

```python
>>> import threading
>>> import time
>>> def worker():
...     while True:
...         time.sleep(60)
... 
>>> threads_list = [threading.Thread(target=worker) for _ in range(2)]
>>> for thread in threads_list:
...     thread.start()
... 
>>> while threads_list:
...     print(threads_list)
...     threads_list = [thread.join(timeout=0.1) for thread in threads_list if thread and thread.is_alive()]
... 
[<Thread(Thread-1 (worker), started 140304535520960)>, <Thread(Thread-2 (worker), started 140304525035200)>]
[None, None]
>>> # Huh?
```

dchristidis: Following my own little tests, I am starting to question the need for a timeout and whether the try-except structure we have have in the daemons for keyboard interrupts is working how it is supposed to.

I am also looking for reproducible example as to why the following won’t do:

```python
for thread in threads:
    thread.join()
```

dchristidis: I’m getting different results in Python 3.6 and 3.9. :shrug: I think this further drives the point that we implemented a solution in Python 2, based on the situation at the time, and kept it as it is without checking if it still makes sense.

rdimaio: Thanks for looking into this! I'm not sure what's the best way forward here - the different results in different Python versions are a bit worrying (though AFAIK Rucio requires Python 3.9 to run, so at least from the user's perspective it will be consistent), but it might also be one of those things that "if it works now, better not touch it" (even though it doesn't really work).

It's definitely a carryover from Python 2 in some way, I found [this commit](https://github.com/rucio/rucio/commit/9fa89245fe468b3ce9b67cfa9dc06bac4c1b77e6) under [this issue](https://github.com/rucio/rucio/issues/1924) for Python 3 compatibility with daemons, and that logic was left untouched

dchristidis: As far as I can tell, the implementation seen in the Undertaker has been there since the very beginning (over a decade ago, see b896d4c64bec067f2f3a48971ac4fa46229a5b99). The justification is lost in time.

Here’s a modern test case:

```python
#!/usr/bin/env python3

import os
import signal
import threading
import time
from types import FrameType
from typing import Optional

def worker() -> None:
    while not GRACEFUL_STOP.is_set():
        print('worker active')
        time.sleep(1)
    print('worker terminating')
    time.sleep(5)
    print('worker terminated')


def stop(signum: Optional[int] = None, frame: Optional[FrameType] = None) -> None:
    GRACEFUL_STOP.set()
    if signum == signal.SIGINT:
        if hasattr(stop, 'received_sigint'):
            exit(1)
        else:
            stop.received_sigint = True

if __name__ == '__main__':
    if 'CAPTURE_SIGINT' in os.environ:
        signal.signal(signal.SIGINT, stop)
    threads = [threading.Thread(target=worker) for _ in range(1)]
    for thread in threads:
        thread.start()
    for thread in threads:
        thread.join()
    print('main thread terminated')
```

Here’s a sample run with Python 3.9:

```console
$ python3 test.py 
worker active
worker active
^CTraceback (most recent call last):
  File "/afs/cern.ch/user/d/dichrist/test.py", line 37, in <module>
    thread.join()
  File "/usr/lib64/python3.9/threading.py", line 1060, in join
    self._wait_for_tstate_lock()
  File "/usr/lib64/python3.9/threading.py", line 1080, in _wait_for_tstate_lock
    if lock.acquire(block, timeout):
KeyboardInterrupt

$ CAPTURE_SIGINT=TRUE python3 test.py 
worker active
worker active
^Cworker terminating
worker terminated
main thread terminated
$ CAPTURE_SIGINT=TRUE python3 test.py 
worker active
worker active
^C^C$ 
```

It all comes down to this:

 1. If we don’t care about graceful shutdown with `SIGINT`, then the simple implementation works.
 2. If we care about it, then we can easily adjust the `stop()` function which is already used for `SIGTERM` to exit gracefully upon the first keyboard interrupt and immediately upon the second.

@bari12 @mlassnig Thoughts?

bari12: I would adjust this. We should take this as an opportunity to renew the code here and don't carry outdated (not to say wrong) choices further in the future.

dchristidis: But the question remains: should we implement Ctrl+C leading to a graceful stop?

lib/rucio/daemons/common.py
```diff
@@ -18,27 +18,143 @@
...
+        """
+        Graceful exit the daemon. Used as handler for SIGTERM.
+        The unused parameters are needed for the method to be accepted as handler for signal.signal().
+
+        :param signum: signal number
+        :param frame: stack frame
+        """
+        self.graceful_stop.set()
+        logging.info("%s: terminating", self.daemon_name)
+        if signum == signal.SIGINT or signum == signal.SIGTERM:
```
rdimaio, line 145: should be sigint only, not sigterm as well

bin/rucio-undertaker
```diff
@@ -66,12 +66,12 @@ Check if the DID exists::
...
+    undertaker = Undertaker(total_workers=args.total_workers, chunk_size=args.chunk_size, once=args.run_once, sleep_time=args.sleep_time)
+    signal.signal(signal.SIGTERM, undertaker.stop)
+    signal.signal(signal.SIGINT, undertaker.stop)
     try:
-        run(total_workers=args.total_workers, chunk_size=args.chunk_size, once=args.run_once,
-            sleep_time=args.sleep_time)
+        undertaker.run()
     except KeyboardInterrupt:
-        stop()
+        undertaker.stop()
```
rdimaio, line 77: this try except is not needed

lib/rucio/daemons/common.py
```diff
@@ -18,27 +18,143 @@
...
+                executable=self.daemon_name,
+                partition_wait_time=self.partition_wait_time,
+                sleep_time=self.sleep_time,
+                activities=activities,
+            )(run_once_fnc)
+
+            for _ in daemon():
+                pass
+
+    def run(self) -> None:
```
rdimaio, line 112: could be renamed to start

lib/rucio/daemons/common.py
rdimaio: worth noting: the daemons in their current version use `SIGTERM`, not `SIGINT`, e.g. https://github.com/rucio/rucio/blob/c8316a595af9ef6336a131ead947414f32e7a8ca/bin/rucio-auditor#L86

dchristidis: The Auditor is not a good example.

`SIGTERM` is what the OS/supervisor/scheduler is using to terminated a process under its control. It is supposed to send `SIGKILL` after a timeout.

`SIGINT` is what Ctrl+C sends. It only makes sense when one starts a daemon process from their terminal, which is unlikely, especially in a production environment.

The daemons that are (supposedly) handling `SIGNIT` are doing so via the `KeyboardInterrupt` exception.

rdimaio: I linked to Auditor as a random example, but it was part of this search https://github.com/search?q=repo%3Arucio%2Frucio%20sigterm&type=code where it seems like most daemons use `SIGTERM`. I agree that for the purpose of gracefully stopping manually we should use `SIGINT`, I just wanted to bring this up for reference

rdimaio:
Closing this; we can keep the https://github.com/rucio/rucio/issues/6478 issue open, and work on this again later down the line, as it's not urgent now

Issue #5224, opened 2022-02-10
Update/Re-design `core.meta` module


bari12:
Motivation
----------
Currently the `core.meta.py` module makes it look like it deals with metadata directly, while this is not true, it is about metadata conventions and enforcing these. To make this more clear and avoid confusion, parts of the module (module name, method names, descriptions) should be updated. I would suggest to name everything as `meta_conventions`. 

Modification
------------
- Change the nodule name, method names, descriptions
- Check that the new metadata handling with the different plugins is actually compatible to this

voetberg:
Would `metadata_interface` make more sense? While it isn't the metadata itself, it does control interactions with metadata

bari12:
This would be a bit confusing with the actual metadata interface (in core/did_meta_plugins). I think I would call it somewhat like `metadata_conventions`?

voetberg:
For the rest api - it advisable to make a legacy endpoint that responds to `/meta/` and a new one for `/meta_conventions/`, or just leave it as `meta`?

bari12:
I think a new one + legacy, and we deprecate the legacy one over time.

PR #6490, opened 2024-01-26
 Metadata: Rename core.meta to core.meta_conventions #5224

voetberg:
Renaming `meta` to `meta_conventions` - updates the core and internal api calls to following the naming scheme. For the web api, a legacy endpoint is included with a warning it will be depreciated.

### DIFF
--- a/lib/rucio/api/did.py
+++ b/lib/rucio/api/did.py
@@ -22,8 +22,9 @@ from rucio.common.exception import RucioException
 from rucio.common.schema import validate_schema
 from rucio.common.types import InternalAccount, InternalScope
 from rucio.common.utils import api_update_return_dict
-from rucio.core import did, naming_convention, meta as meta_core
+from rucio.core import did, naming_convention
 from rucio.core.rse import get_rse_id
+from rucio.core import meta_conventions as meta_convention_core
 from rucio.db.sqla.constants import DIDType
 from rucio.db.sqla.session import read_session, stream_session, transactional_session
 
@@ -117,7 +118,7 @@ def add_did(scope, name, did_type, issuer, account=None, statuses={}, meta={}, r
                 raise rucio.common.exception.InvalidObject("Provided metadata %s doesn't match the naming convention: %s != %s" % (k, meta[k], extra_meta[k]))
 
         # Validate metadata
-        meta_core.validate_meta(meta=meta, did_type=DIDType[did_type.upper()], session=session)
+        meta_convention_core.validate_meta(meta=meta, did_type=DIDType[did_type.upper()], session=session)
 
     return did.add_did(scope=scope, name=name, did_type=DIDType[did_type.upper()], account=account or issuer,
                        statuses=statuses, meta=meta, rules=rules, lifetime=lifetime,
--- a/lib/rucio/client/client.py
+++ b/lib/rucio/client/client.py
@@ -27,7 +27,7 @@ from rucio.client.exportclient import ExportClient
 from rucio.client.importclient import ImportClient
 from rucio.client.lifetimeclient import LifetimeClient
 from rucio.client.lockclient import LockClient
-from rucio.client.metaclient import MetaClient
+from rucio.client.metaconventionsclient import MetaConventionClient
 from rucio.client.pingclient import PingClient
 from rucio.client.replicaclient import ReplicaClient
 from rucio.client.requestclient import RequestClient
@@ -40,7 +40,7 @@ from rucio.client.touchclient import TouchClient
 
 class Client(AccountClient,
              AccountLimitClient,
-             MetaClient,
+             MetaConventionClient,
              PingClient,
              ReplicaClient,
              RequestClient,
--- a/lib/rucio/client/metaconventionsclient.py
+++ b/lib/rucio/client/metaconventionsclient.py
@@ -21,17 +21,19 @@ from requests.status_codes import codes
 from rucio.client.baseclient import BaseClient
 from rucio.client.baseclient import choice
 from rucio.common.utils import build_url
+from rucio.db.sqla.constants import KeyType
+from typing import Union, Optional
 
 
-class MetaClient(BaseClient):
+class MetaConventionClient(BaseClient):
 
-    """Meta client class for working with data identifier attributes"""
+    """Metadata client class for working with data identifier attributes"""
 
-    META_BASEURL = 'meta'
+    META_BASEURL = 'meta_conventions'
 
-    def add_key(self, key, key_type, value_type=None, value_regexp=None):
+    def add_key(self, key: str, key_type: Union[KeyType, str], value_type: Optional[str] = None, value_regexp: Optional[str] = None) -> Optional[bool]:
         """
-        Sends the request to add a new key.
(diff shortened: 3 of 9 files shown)


lib/rucio/api/did.py
```diff
@@ -15,14 +15,15 @@
 
 from copy import deepcopy
 from typing import TYPE_CHECKING
+from rucio.core import meta_conventions as meta_convention_core
```
bari12: Move below to the other `rucio.core` imports

lib/rucio/client/client.py
```diff
@@ -40,7 +40,7 @@
 
 class Client(AccountClient,
              AccountLimitClient,
-             MetaClient,
+             MetConventionsClient,
```
bari12: `MetaConventionClient` :-)

lib/rucio/client/metaconventionsclient.py
```diff
@@ -23,7 +23,7 @@
 from rucio.common.utils import build_url
 
 
-class MetaClient(BaseClient):
+class MetConventionsClient(BaseClient):
```
bari12: Please update the documentation descriptions here as well.

lib/rucio/core/meta_conventions.py
```diff
@@ -32,9 +32,9 @@
 
 
 @transactional_session
-def add_key(key, key_type, value_type=None, value_regexp=None, *, session: "Session"):
+def add_key(key: str, key_type: Union[KeyType, str], value_type: Optional[str] = None, value_regexp: Optional[str] = None, *, session: "Session"):
     """
-    Adds a new allowed key.
+    Add an allowed key for DID metadata.
```
bari12: I am really wondering if we should rename the models below as well.

Thus from `DIDKey` to `DIDMetaConventionKey` or similar
The question is though, if we should als rename the database tables for that. Probably we should

lib/rucio/core/meta_conventions.py
```diff
@@ -81,9 +81,9 @@ def add_key(key, key_type, value_type=None, value_regexp=None, *, session: "Sess
 
 
 @transactional_session
-def del_key(key, *, session: "Session"):
+def del_key(key: str, *, session: "Session"):
     """
-    Deletes a key.
+    Delete a key in the DID Metadata table.
```
bari12: Delete a key in the DID Metadata Convention table

lib/rucio/core/meta_conventions.py
```diff
@@ -94,7 +94,7 @@ def del_key(key, *, session: "Session"):
 @read_session
 def list_keys(*, session: "Session"):
     """
-    Lists all keys.
+    Lists all keys for DID Metadata.
```
bari12: Same. Really be careful so people do not confuse the conventions with actual metadata

bari12 (changes requested):

lib/rucio/core/meta_conventions.py
voetberg: I agree, that makes sense to me. It was difficult to figure out wording for some of the documentation even trying to differentiate a metadata key and a metadata conventions key.

voetberg: Something of note while I'm playing around with the tables - after trying to convert the queries in meta_conventions to sqlalchemy 2.0 (re #5247) I noticed a lot of queries that here break when you do this because of foreign key constraints. (e.g.: `(psycopg2.errors.ForeignKeyViolation) update or delete on table "did_keys" violates foreign key constraint "DID_MAP_KEYS_FK" on table "did_key_map"` is raised for any interaction that involves both of the tables, specifically for updating the value of an existing key). 

Is there an additional statements that need to be added to the sqla2.0 queries, is there something running amuck in the model definitions, or are the existing tests just doing unintended things?

bari12: I don't think so, this really shouldn't be the case if it worked in sqlalchemy 1.0 syntax. Do you have a specific example?

voetberg: Yes, for example swapping 
```
session.query(models.DIDMetaConventionsConstraints.value).\
                filter_by(key=key).\
                filter_by(value=meta[key]).\
                one()

```

for 
```
            statement = select(
                models.DIDMetaConventionsConstraints.value
            ).where(
                and_(models.DIDMetaConventionsConstraints.value == meta[key], models.DIDMetaConventionsConstraints.key == key)
            )
            session.execute(statement).one()

```

Causes this, as does swapping out the `del_key` functionality: 
```
   session.query(models.DIDMetaConventionsKey).filter(key == key).delete()
```
-> 
```
     statement = delete(
             models.DIDMetaConventionsKey
         ).where(
             models.DIDMetaConventionsKey.key == key
         )
     session.execute(statement)
```

voetberg: Nevermind, figured it out. I hadn't been re-syncing the db tables between running tests so things got confused.

lib/rucio/core/meta_conventions.py
```diff
@@ -77,46 +77,52 @@ def add_key(key, key_type, value_type=None, value_regexp=None, *, session: "Sess
...
     :param session: The database session in use.
 
     :returns: A list containing all keys.
     """
     key_list = []
-    query = session.query(models.DIDKey)
+    statement = select(models.DIDMetaConventionsKey.key)
+    query = session.execute(statement)
     for row in query:
-        key_list.append(row.key)
```
bari12, line 106: Use `.scalars()` here. so you don't have to do `row[0]`

lib/rucio/core/meta_conventions.py
```diff
@@ -137,54 +143,64 @@ def add_value(key, value, *, session: "Session"):
...
         raise RucioException(error.args)
 
-    k = session.query(models.DIDKey).filter_by(key=key).one()
+    statement = select(
+        models.DIDMetaConventionsKey.value_regexp,
+        models.DIDMetaConventionsKey.value_type
+    ).where(
+        models.DIDMetaConventionsKey.key == key
+    )
+    query = session.execute(statement).one()
```
bari12: Also use `.scalars()` here, the `query[0]`, `query[1]` etc is not very nice :-)
See https://github.com/rucio/rucio/blob/2c5b291f53caebf85259010f11791ed7f497572f/lib/rucio/core/rse.py#L1115

lib/rucio/core/meta_conventions.py
```diff
@@ -137,54 +143,64 @@ def add_value(key, value, *, session: "Session"):
...
+    Lists all allowed values for a DID key (all values for a key in DID Metadata Conventions).
 
     :param key: the name for the key.
     :param session: The database session in use.
 
     :returns: A list containing all values.
     """
     value_list = []
-    query = session.query(models.DIDKeyValueAssociation).filter_by(key=key)
+    statement = select(models.DIDMetaConventionsConstraints.value).where(models.DIDMetaConventionsConstraints.key == key)
```
bari12, line 174: Use `.scalars()`

bari12 (changes requested):

bari12 (approved):

Issue #6057, opened 2023-01-27
Continue migration to SQLAlchemy 2.0 syntax


bari12:
### Description

This is an epic ticket to track the migration of SQLAlchemy 2.0 ORM syntax.

### Motivation

https://docs.sqlalchemy.org/en/20/changelog/migration_20.html

### Change

- [ ] Create an overview of which Rucio modules are correctly migrated and which ones still need changes (and possible how many)
- [ ] Plan how/who to implement the changes

rcarpa:
Maybe just upgrade to sqlalchemy2 and fix whatever breaks ? Some deprecated constructs like the "query" api are only deprecated on paper and will be probably supported at least till sqlalchemy3 :D we don't "need" to migrate all queries to the new syntax, but the fixes from sqlalchemy2 and type checking support would be welcomed asap

rcarpa:
I just submitted a PR which updates many queries to a format which supports both sqlalchemy1.4 and 2.0: https://github.com/rucio/rucio/pull/6060 .

yuyiguo:
Rucio modules already in SQLAlchemy 2.0 syntax:

1. lib/rucio/db
2. lib/rucio/core (partialiy)
3. lib/rucio/daemons/bb8


Rucio modules need to migrate to SQLAlchemy 2.0 syntax:
lib/rucio/tests
lib/rucio/daemons
lib/rucio/api

These were  roughly estimated based the " import sqlalchemy" . We will need details before the assignment

bari12:
Hi Yuyi, a bit more details would be good, since this is still quite large categories. Is there some kind of simple test we can just run to see which code is not 2.0 ready?

yuyiguo:
Hi @bari12 
Based on the [document](https://docs.sqlalchemy.org/en/20/changelog/migration_20.html) , there are seven steps to migrate to 2.0. Let's go through these.

1. Migration to 2.0 Step One - Python 3 only (Python 3.7 minimum for 2.0 compatibility)
    done
2. Migration to 2.0 Step Two - Turn on RemovedIn20Warnings
    I tried and did not find a way to do so because we are running in a server or daemon.
3. Migration to 2.0 Step Three - Resolve all RemovedIn20Warnings
    We cannot do it because we cannot run step 2.
4.  Migration to 2.0 Step Four - Use the future flag on Engine
     We did it in [5872](https://github.com/rucio/rucio/pull/5872/files).
5. Migration to 2.0 Step Five - Use the future flag on Session
[...]

PR #6491, opened 2024-01-29
Core & Internals: core/replica.py sqla20 syntax #6057

erlingstaff:
- Refactored db queries in core/replica.py to use SQLAlchemy 2.0 syntax.
- Added new util function in db/sqla/__init__.py: 'filter_thread_work_where' which simplifies 'filter_thread_work' and uses new sqla2.0 syntax

This PR is a small part of the bigger initiative of migrating the entire codebase to use SQLAlchemy 2.0 syntax (#6057). This PR is submitted for only core/replica.py as there are many LOC changes in this file, subsequent PRs will likely contain several files.


lib/rucio/db/sqla/__init__.py
```diff
@@ -36,3 +39,21 @@ def filter_thread_work(session, query, total_threads, thread_id, hash_variable=N
             else:
                 query = query.filter(text('mod(abs((\'x\'||md5(%s::text))::bit(32)::bigint), %s) = %s' % (hash_variable, total_threads, thread_id)))
     return query
+
+
+def filter_thread_work_where(session: "Session", stmt: "Select", total_threads: Optional[int], thread_id: Optional[int], hash_variable=None):
```
rcarpa: Duplicating this function is not needed. where is a synonym to filter. Both for the query API and the statement api.

lib/rucio/core/replica.py
```diff
@@ -160,27 +161,32 @@ def __exist_replicas(rse_id, replicas, *, session: "Session"):
...
+                    models.RSEFileAssociation.path,
+                    models.RSEFileAssociation.scope,
+                    models.RSEFileAssociation.name,
+                    models.RSEFileAssociation.rse_id,
+                    models.RSEFileAssociation.bytes,
+                    func.max(case((models.BadReplicas.state == BadFilesStatus.SUSPICIOUS, 0),
+                                  (models.BadReplicas.state == BadFilesStatus.BAD, 1),
+                                  else_=0))
+                ).with_hint(models.RSEFileAssociation, "INDEX(REPLICAS REPLICAS_PATH_IDX", 'oracle').\
                     outerjoin(models.BadReplicas,
```
rcarpa: Please update the formatting to respect the same formatting rules everywhere. For example: 
```
).with_hint(
  models.RSEFileAssociation, "INDEX(REPLICAS REPLICAS_PATH_IDX", 'oracle'
).outerjoin(
  models.BadReplicas,
```

lib/rucio/core/replica.py
```diff
@@ -520,18 +542,24 @@ def get_bad_replicas_backlog(*, session: "Session"):
 
     :returns: a list of dictionary {'rse_id': cnt_bad_replicas}.
     """
-    query = session.query(func.count(models.RSEFileAssociation.rse_id), models.RSEFileAssociation.rse_id). \
-        with_hint(models.RSEFileAssociation, 'INDEX(DIDS DIDS_PK) USE_NL(DIDS) INDEX_RS_ASC(REPLICAS ("REPLICAS"."STATE"))', 'oracle'). \
-        filter(models.RSEFileAssociation.state == ReplicaState.BAD)
+    stmt = select(func.count(models.BadReplicas.scope), models.BadReplicas.rse_id).\
```
rcarpa: reformat the code to the new formatting style.

lib/rucio/core/replica.py
```diff
@@ -1486,9 +1537,15 @@ def __bulk_add_replicas(rse_id, files, account, *, session: "Session"):
...
-        filter(or_(*condition))
+    stmt = select(
+        models.RSEFileAssociation.scope,
+        models.RSEFileAssociation.name,
+        models.RSEFileAssociation.rse_id,
+    ).where(
+        or_(*condition)
+    ).with_hint(models.RSEFileAssociation, "INDEX(REPLICAS REPLICAS_PK)", 'oracle')
+
+    query = session.execute(stmt)
```
rcarpa: it's not a `query` anymore, it's a `result`.

lib/rucio/core/replica.py
```diff
@@ -1978,11 +2041,13 @@ def __cleanup_after_replica_deletion(scope_name_temp_table, scope_name_temp_tabl
...
-                                  models.DataIdentifier.name,
-                                  models.DataIdentifier.did_type). \
-                with_hint(models.DataIdentifier, "INDEX(DIDS DIDS_PK)", 'oracle'). \
-                filter(or_(*chunk))
+            stmt = select(
+                models.DataIdentifier.scope,
+                models.DataIdentifier.name,
+                models.DataIdentifier.did_type,
+            ).with_hint(models.DataIdentifier, "INDEX(DIDS DIDS_PK)", 'oracle').\
+                where(or_(*chunk))
```
rcarpa: fix the formatting

rcarpa (changes requested):
There are many formatting issues. I didn't' even come close to listing all of them. This Refactoring should make full use of the opportunity to fix the formatting of all queries to respect the same convention as in other places which were migrated to the new sqlalchemy 2 format.

lib/rucio/db/sqla/__init__.py
erlingstaff: Removed this in new commit

lib/rucio/core/replica.py
erlingstaff: Formatting should be following new conventions in the new commit now

lib/rucio/core/replica.py
erlingstaff: Renamed this and one other occurrence of this error in new commit

lib/rucio/core/replica.py
```diff
@@ -468,18 +492,25 @@ def get_pfn_to_rse(pfns, vo='def', *, session: "Session"):
...
+        models.RSEProtocols.prefix
+    ).join(
+        models.RSE, models.RSE.id == models.RSEProtocols.rse_id
+    ).where(
+        and_(or_(*se_condition),
+             models.RSEProtocols.scheme == scheme)
+    ).where(
+        and_(models.RSE.deleted == false(),
+             models.RSE.staging_area == false())
+    )
```
rdimaio: I think this could also be written under a single `where` statement, but it would functionally be the same so I don't think it's necessary to make a change

lib/rucio/core/replica.py
```diff
@@ -520,18 +551,27 @@ def get_bad_replicas_backlog(*, session: "Session"):
...
-        with_hint(models.RSEFileAssociation, 'INDEX(DIDS DIDS_PK) USE_NL(DIDS) INDEX_RS_ASC(REPLICAS ("REPLICAS"."STATE"))', 'oracle'). \
-        filter(models.RSEFileAssociation.state == ReplicaState.BAD)
+    stmt = select(
+        func.count(models.BadReplicas.scope),
+        models.BadReplicas.rse_id
+    ).with_hint(
+        models.RSEFileAssociation, 'INDEX(DIDS DIDS_PK) USE_NL(DIDS) INDEX_RS_ASC(REPLICAS ("REPLICAS"."STATE"))', 'oracle'
+    ).where(
+        models.BadReplicas.state == BadFilesStatus.BAD
+    )
```
rdimaio: The new query here seems so be slightly different compared to the old query (the queried columns are different and the `where` condition is different) - is this intended?

lib/rucio/core/replica.py
```diff
@@ -520,18 +551,27 @@ def get_bad_replicas_backlog(*, session: "Session"):
...
-        group_by(models.RSEFileAssociation.rse_id)
+    stmt = stmt.join(
+        models.DataIdentifier,
+        and_(models.DataIdentifier.scope == models.BadReplicas.scope,
+             models.DataIdentifier.name == models.BadReplicas.name)
+    ).where(
+        models.DataIdentifier.availability != DIDAvailability.LOST
+    ).group_by(
+        models.BadReplicas.rse_id
+    )
```
rdimaio: the `group_by` field is different in the new query - is that intended?

lib/rucio/core/replica.py
```diff
@@ -2456,15 +2620,19 @@ def get_and_lock_file_replicas(scope, name, nowait=False, restrict_rses=None, *,
...
+    stmt = select(
+        models.RSEFileAssociation
+    ).where(
+        and_(models.RSEFileAssociation.scope == scope,
+             models.RSEFileAssociation.name == name,
+             models.RSEFileAssociation.state != ReplicaState.BEING_DELETED)
+    )
+    if restrict_rses is not None and len(restrict_rses) < 10:
+        rse_clause = [models.RSEFileAssociation.rse_id == rse_id for rse_id in restrict_rses]
+        stmt = stmt.where(or_(*rse_clause))
```
rdimaio: I would still put this under `if rse_clause:` to keep it functionally the same as before

lib/rucio/core/replica.py
```diff
@@ -3253,9 +3589,33 @@ def bulk_add_bad_replicas(replicas, account, state=BadFilesStatus.TEMPORARY_UNAV
...
-            query = session.query(models.BadReplicas).filter_by(scope=replica['scope'], name=replica['name'], rse_id=replica['rse_id'], state=state)
-            if query.count():
-                query.update({'state': BadFilesStatus.TEMPORARY_UNAVAILABLE, 'updated_at': datetime.utcnow(), 'account': account, 'reason': reason, 'expires_at': expires_at}, synchronize_session=False)
+            stmt = select(
+                models.BadReplicas
+            ).where(
+                and_(models.BadReplicas.scope == replica['scope'],
+                     models.BadReplicas.name == replica['name'],
+                     models.BadReplicas.rse_id == replica['rse_id'],
+                     models.BadReplicas.state == BadFilesStatus.TEMPORARY_UNAVAILABLE)
```
rdimaio: Would be best to use `models.BadReplicas.state == state` here to be consistent with the previous version of the query

lib/rucio/core/replica.py
```diff
@@ -3253,9 +3589,33 @@ def bulk_add_bad_replicas(replicas, account, state=BadFilesStatus.TEMPORARY_UNAV
...
+            )
+            if session.execute(stmt).scalar():
```
rdimaio: Could use [`one_or_none`](https://docs.sqlalchemy.org/en/14/orm/query.html#sqlalchemy.orm.Query.one_or_none) here

lib/rucio/core/replica.py
```diff
@@ -3607,16 +4025,19 @@ def get_RSEcoverage_of_dataset(scope, name, *, session: "Session"):
...
-                              models.DataIdentifierAssociation.child_name == models.RSEFileAssociation.name,
-                              models.DataIdentifierAssociation.scope == scope,
-                              models.DataIdentifierAssociation.name == name,
-                              models.RSEFileAssociation.state != ReplicaState.BEING_DELETED,
-                              ))
-
-    query = query.group_by(models.RSEFileAssociation.rse_id)
+    stmt = select(
+        models.RSEFileAssociation.rse_id,
+        func.sum(models.RSEFileAssociation.bytes)
```
rdimaio: this is different from the previous query, where `models.DataIdentifierAssociation.bytes` was used instead of `models.RSEFileAssociation.bytes`

erlingstaff:
Closed, moved to https://github.com/rucio/rucio/pull/6591

PR #6493, opened 2024-02-01
Core & Internals: core/rule.py sqla20 syntax #6057

erlingstaff:
* Migrated the db queries in core/rule.py to SQLAlchemy 2.0 syntax, as per #6057  

PR submitted for only rule.py as this file contains ~1000 additions/deletions.

### DIFF
--- a/lib/rucio/core/rule.py
+++ b/lib/rucio/core/rule.py
@@ -25,7 +25,7 @@ from string import Template
 from typing import TYPE_CHECKING, Any, Callable, Literal, Optional, Type, TypeVar, Union
 
 from dogpile.cache.api import NoValue
-from sqlalchemy import select, update
+from sqlalchemy import delete, desc, select, update
 from sqlalchemy.exc import IntegrityError, StatementError
 from sqlalchemy.exc import NoResultFound  # https://pydoc.dev/sqlalchemy/latest/sqlalchemy.exc.NoResultFound.html
 from sqlalchemy.sql import func
@@ -260,8 +260,13 @@ def add_rule(
             # 3. Get the did
             with METRICS.timer('add_rule.get_did'):
                 try:
-                    did = session.query(models.DataIdentifier).filter(models.DataIdentifier.scope == elem['scope'],
-                                                                      models.DataIdentifier.name == elem['name']).one()
+                    stmt = select(
+                        models.DataIdentifier
+                    ).where(
+                        and_(models.DataIdentifier.scope == elem['scope'],
+                             models.DataIdentifier.name == elem['name'])
+                    )
+                    did = session.execute(stmt).scalar_one()
                 except NoResultFound as exc:
                     raise DataIdentifierNotFound('Data identifier %s:%s is not valid.' % (elem['scope'], elem['name'])) from exc
                 except TypeError as error:
@@ -270,23 +275,44 @@ def add_rule(
             # 3.1 If the did is a constituent, relay the rule to the archive
             if did.did_type == DIDType.FILE and did.constituent:
                 # Check if a single replica of this DID exists; Do not put rule on file if there are only replicas on TAPE
-                replica_cnt = session.query(models.RSEFileAssociation).join(models.RSE, models.RSEFileAssociation.rse_id == models.RSE.id)\
-                    .filter(models.RSEFileAssociation.scope == did.scope,
-                            models.RSEFileAssociation.name == did.name,
-                            models.RSEFileAssociation.state == ReplicaState.AVAILABLE,
-                            models.RSE.rse_type != RSEType.TAPE).count()
+                stmt = select(
+                    func.count()
+                ).select_from(
+                    models.RSEFileAssociation
+                ).join(
+                    models.RSE,
+                    models.RSEFileAssociation.rse_id == models.RSE.id
+                ).where(
+                    and_(models.RSEFileAssociation.scope == did.scope,
+                         models.RSEFileAssociation.name == did.name,
+                         models.RSEFileAssociation.state == ReplicaState.AVAILABLE,
+                         models.RSE.rse_type != RSEType.TAPE)
+                )
+                replica_cnt = session.execute(stmt).scalar()
+
                 if replica_cnt == 0:  # Put the rule on the archive
-                    archive = session.query(models.ConstituentAssociation).join(models.RSEFileAssociation,
-                                                                                and_(models.ConstituentAssociation.scope == models.RSEFileAssociation.scope,
-                                                                                     models.ConstituentAssociation.name == models.RSEFileAssociation.name))\
-                        .filter(models.ConstituentAssociation.child_scope == did.scope,
-                                models.ConstituentAssociation.child_name == did.name).first()
+                    stmt = select(
+                        models.ConstituentAssociation
+                    ).join(
+                        models.RSEFileAssociation,
+                        and_(models.ConstituentAssociation.scope == models.RSEFileAssociation.scope,
+                             models.ConstituentAssociation.name == models.RSEFileAssociation.name)
+                    ).where(
+                        and_(models.ConstituentAssociation.child_scope == did.scope,
+                             models.ConstituentAssociation.child_name == did.name)
+                    )
+                    archive = session.execute(stmt).scalars().first()
                     if archive is not None:
                         elem['name'] = archive.name
                         elem['scope'] = archive.scope
                         try:
-                            did = session.query(models.DataIdentifier).filter(models.DataIdentifier.scope == elem['scope'],
-                                                                              models.DataIdentifier.name == elem['name']).one()
+                            stmt = select(
+                                models.DataIdentifier
+                            ).where(
+                                and_(models.DataIdentifier.scope == elem['scope'],
+                                     models.DataIdentifier.name == elem['name'])
+                            )
+                            did = session.execute(stmt).scalar_one()
                         except NoResultFound as exc:
                             raise DataIdentifierNotFound('Data identifier %s:%s is not valid.' % (elem['scope'], elem['name'])) from exc
                         except TypeError as error:
@@ -386,11 +412,25 @@ def add_rule(
                 new_rule.state = RuleState.STUCK
                 new_rule.error = 'MissingSourceReplica'
                 if new_rule.grouping != RuleGrouping.NONE:
-                    session.query(models.DatasetLock).filter_by(rule_id=new_rule.id).update({'state': LockState.STUCK})
+                    stmt = update(
+                        models.DatasetLock
+                    ).where(
+                        models.DatasetLock.rule_id == new_rule.id
+                    ).values({
+                        models.DatasetLock.state: LockState.STUCK
+                    })
+                    session.execute(stmt)
             elif new_rule.locks_replicating_cnt == 0:
                 new_rule.state = RuleState.OK
                 if new_rule.grouping != RuleGrouping.NONE:
-                    session.query(models.DatasetLock).filter_by(rule_id=new_rule.id).update({'state': LockState.OK})
+                    stmt = update(
+                        models.DatasetLock
+                    ).where(
+                        models.DatasetLock.rule_id == new_rule.id
+                    ).values({
+                        models.DatasetLock.state: LockState.OK
+                    })
+                    session.execute(stmt)
                     session.flush()
                 if new_rule.notification == RuleNotification.YES:
                     generate_email_for_rule_ok_notification(rule=new_rule, session=session)
@@ -398,7 +438,14 @@ def add_rule(
             else:
                 new_rule.state = RuleState.REPLICATING
                 if new_rule.grouping != RuleGrouping.NONE:
-                    session.query(models.DatasetLock).filter_by(rule_id=new_rule.id).update({'state': LockState.REPLICATING})
+                    stmt = update(
+                        models.DatasetLock
+                    ).where(
+                        models.DatasetLock.rule_id == new_rule.id
+                    ).values({
+                        models.DatasetLock.state: LockState.REPLICATING
+                    })
+                    session.execute(stmt)
 
             # Add rule to History
             insert_rule_history(rule=new_rule, recent=True, longterm=True, session=session)
@@ -457,9 +504,13 @@ def add_rules(
             # 2. Get the did
             with METRICS.timer('add_rules.get_did'):
                 try:
-                    did = session.query(models.DataIdentifier).filter(
-                        models.DataIdentifier.scope == elem['scope'],
-                        models.DataIdentifier.name == elem['name']).one()
+                    stmt = select(
+                        models.DataIdentifier
+                    ).where(
+                        and_(models.DataIdentifier.scope == elem['scope'],
+                             models.DataIdentifier.name == elem['name'])
+                    )
+                    did = session.execute(stmt).scalar_one()
                 except NoResultFound as exc:
                     raise DataIdentifierNotFound('Data identifier %s:%s is not valid.' % (elem['scope'], elem['name'])) from exc
                 except TypeError as error:
@@ -467,23 +518,43 @@ def add_rules(
 
             # 2.1 If the did is a constituent, relay the rule to the archive
             if did.did_type == DIDType.FILE and did.constituent:  # Check if a single replica of this DID exists
-                replica_cnt = session.query(models.RSEFileAssociation).join(models.RSE, models.RSEFileAssociation.rse_id == models.RSE.id)\
-                    .filter(models.RSEFileAssociation.scope == did.scope,
-                            models.RSEFileAssociation.name == did.name,
-                            models.RSEFileAssociation.state == ReplicaState.AVAILABLE,
-                            models.RSE.rse_type != RSEType.TAPE).count()
+                stmt = select(
+                    func.count()
+                ).select_from(
+                    models.RSEFileAssociation
+                ).join(
+                    models.RSE,
+                    models.RSEFileAssociation.rse_id == models.RSE.id
+                ).where(
+                    and_(models.RSEFileAssociation.scope == did.scope,
+                         models.RSEFileAssociation.name == did.name,
+                         models.RSEFileAssociation.state == ReplicaState.AVAILABLE,
+                         models.RSE.rse_type != RSEType.TAPE)
+                )
+                replica_cnt = session.execute(stmt).scalar()
                 if replica_cnt == 0:  # Put the rule on the archive
-                    archive = session.query(models.ConstituentAssociation).join(models.RSEFileAssociation,
-                                                                                and_(models.ConstituentAssociation.scope == models.RSEFileAssociation.scope,
-                                                                                     models.ConstituentAssociation.name == models.RSEFileAssociation.name))\
-                        .filter(models.ConstituentAssociation.child_scope == did.scope,
-                                models.ConstituentAssociation.child_name == did.name).first()
+                    stmt = select(
+                        models.ConstituentAssociation
+                    ).join(
+                        models.RSEFileAssociation,
+                        and_(models.ConstituentAssociation.scope == models.RSEFileAssociation.scope,
+                             models.ConstituentAssociation.name == models.RSEFileAssociation.name)
+                    ).where(
+                        and_(models.ConstituentAssociation.child_scope == did.scope,
+                             models.ConstituentAssociation.child_name == did.name)
+                    )
+                    archive = session.execute(stmt).scalars().first()
                     if archive is not None:
                         elem['name'] = archive.name
                         elem['scope'] = archive.scope
                         try:
-                            did = session.query(models.DataIdentifier).filter(models.DataIdentifier.scope == elem['scope'],
-                                                                              models.DataIdentifier.name == elem['name']).one()
+                            stmt = select(
+                                models.DataIdentifier
+                            ).where(
+                                and_(models.DataIdentifier.scope == elem['scope'],
+                                     models.DataIdentifier.name == elem['name'])
+                            )
+                            did = session.execute(stmt).scalar_one()
                         except NoResultFound as exc:
                             raise DataIdentifierNotFound('Data identifier %s:%s is not valid.' % (elem['scope'], elem['name'])) from exc
                         except TypeError as error:
@@ -653,11 +724,25 @@ def add_rules(
                         new_rule.state = RuleState.STUCK
                         new_rule.error = 'MissingSourceReplica'
                         if new_rule.grouping != RuleGrouping.NONE:
-                            session.query(models.DatasetLock).filter_by(rule_id=new_rule.id).update({'state': LockState.STUCK})
+                            stmt = update(
+                                models.DatasetLock
+                            ).where(
+                                models.DatasetLock.rule_id == new_rule.id
+                            ).values({
+                                models.DatasetLock.state: LockState.STUCK
+                            })
+                            session.execute(stmt)
                     elif new_rule.locks_replicating_cnt == 0:
                         new_rule.state = RuleState.OK
                         if new_rule.grouping != RuleGrouping.NONE:
-                            session.query(models.DatasetLock).filter_by(rule_id=new_rule.id).update({'state': LockState.OK})
+                            stmt = update(
+                                models.DatasetLock
+                            ).where(
+                                models.DatasetLock.rule_id == new_rule.id
+                            ).values({
+                                models.DatasetLock.state: LockState.OK
+                            })
+                            session.execute(stmt)
                             session.flush()
                         if new_rule.notification == RuleNotification.YES:
                             generate_email_for_rule_ok_notification(rule=new_rule, session=session)
@@ -665,7 +750,14 @@ def add_rules(
                     else:
                         new_rule.state = RuleState.REPLICATING
                         if new_rule.grouping != RuleGrouping.NONE:
-                            session.query(models.DatasetLock).filter_by(rule_id=new_rule.id).update({'state': LockState.REPLICATING})
+                            stmt = update(
+                                models.DatasetLock
+                            ).where(
+                                models.DatasetLock.rule_id == new_rule.id
+                            ).values({
+                                models.DatasetLock.state: LockState.REPLICATING
+                            })
+                            session.execute(stmt)
 
                     # Add rule to History
                     insert_rule_history(rule=new_rule, recent=True, longterm=True, session=session)
@@ -693,7 +785,14 @@ def inject_rule(
     """
 
     try:
-        rule = session.query(models.ReplicationRule).filter(models.ReplicationRule.id == rule_id).with_for_update(nowait=True).one()
+        stmt = select(
+            models.ReplicationRule
+        ).where(
+            models.ReplicationRule.id == rule_id
+        ).with_for_update(
+            nowait=True
+        )
+        rule = session.execute(stmt).scalar_one()
     except NoResultFound as exc:
         raise RuleNotFound('No rule with the id %s found' % (rule_id)) from exc
 
@@ -710,7 +809,13 @@ def inject_rule(
         # Remove duplicates from the list of dictionaries
         dids = [dict(t) for t in {tuple(d.items()) for d in dids}]
         # Remove dids which already have a similar rule
-        dids = [did for did in dids if session.query(models.ReplicationRule).filter_by(scope=did['scope'], name=did['name'], account=rule.account, rse_expression=rule.rse_expression).count() == 0]
+        stmt = select(
+            models.ReplicationRule.id
+        ).where(
+            and_(models.ReplicationRule.account == rule.account,
+                 models.ReplicationRule.rse_expression == rule.rse_expression)
+        )
+        dids = [did for did in dids if session.execute(stmt.where(and_(models.ReplicationRule.scope == did['scope'], models.ReplicationRule.name == did['name']))).scalar_one_or_none() is None]
         if rule.expires_at:
             lifetime = (rule.expires_at - datetime.utcnow()).days * 24 * 3600 + (rule.expires_at - datetime.utcnow()).seconds
         else:
@@ -759,8 +864,13 @@ def inject_rule(
     # 3. Get the did
     with METRICS.timer('inject_rule.get_did'):
         try:
-            did = session.query(models.DataIdentifier).filter(models.DataIdentifier.scope == rule.scope,
-                                                              models.DataIdentifier.name == rule.name).one()
+            stmt = select(
+                models.DataIdentifier
+            ).where(
+                and_(models.DataIdentifier.scope == rule.scope,
+                     models.DataIdentifier.name == rule.name)
+            )
+            did = session.execute(stmt).scalar_one()
         except NoResultFound as exc:
             raise DataIdentifierNotFound('Data identifier %s:%s is not valid.' % (rule.scope, rule.name)) from exc
         except TypeError as error:
@@ -777,11 +887,25 @@ def inject_rule(
         rule.state = RuleState.STUCK
         rule.error = 'MissingSourceReplica'
         if rule.grouping != RuleGrouping.NONE:
-            session.query(models.DatasetLock).filter_by(rule_id=rule.id).update({'state': LockState.STUCK})
+            stmt = update(
+                models.DatasetLock
+            ).where(
+                models.DatasetLock.rule_id == rule.id
+            ).values({
+                models.DatasetLock.state: LockState.STUCK
+            })
+            session.execute(stmt)
     elif rule.locks_replicating_cnt == 0:
         rule.state = RuleState.OK
         if rule.grouping != RuleGrouping.NONE:
-            session.query(models.DatasetLock).filter_by(rule_id=rule.id).update({'state': LockState.OK})
+            stmt = update(
+                models.DatasetLock
+            ).where(
+                models.DatasetLock.rule_id == rule.id
+            ).values({
+                models.DatasetLock.state: LockState.OK
+            })
+            session.execute(stmt)
             session.flush()
         if rule.notification == RuleNotification.YES:
             generate_email_for_rule_ok_notification(rule=rule, session=session)
@@ -791,7 +915,14 @@ def inject_rule(
     else:
         rule.state = RuleState.REPLICATING
         if rule.grouping != RuleGrouping.NONE:
-            session.query(models.DatasetLock).filter_by(rule_id=rule.id).update({'state': LockState.REPLICATING})
+            stmt = update(
+                models.DatasetLock
+            ).where(
+                models.DatasetLock.rule_id == rule.id
+            ).values({
+                models.DatasetLock.state: LockState.REPLICATING
+            })
+            session.execute(stmt)
 
     # Add rule to History
     insert_rule_history(rule=rule, recent=True, longterm=True, session=session)
@@ -880,15 +1011,21 @@ def list_rule_history(
     :raises:        RucioException
     """
 
-    query = session.query(models.ReplicationRuleHistoryRecent.updated_at,
-                          models.ReplicationRuleHistoryRecent.state,
-                          models.ReplicationRuleHistoryRecent.locks_ok_cnt,
-                          models.ReplicationRuleHistoryRecent.locks_stuck_cnt,
-                          models.ReplicationRuleHistoryRecent.locks_replicating_cnt).filter_by(id=rule_id).order_by(models.ReplicationRuleHistoryRecent.updated_at)
+    stmt = select(
+        models.ReplicationRuleHistory.updated_at,
+        models.ReplicationRuleHistory.state,
+        models.ReplicationRuleHistory.locks_ok_cnt,
+        models.ReplicationRuleHistory.locks_stuck_cnt,
+        models.ReplicationRuleHistory.locks_replicating_cnt
+    ).where(
+        models.ReplicationRuleHistory.id == rule_id
+    ).order_by(
+        models.ReplicationRuleHistory.updated_at
+    )
 
     try:
-        for rule in query.yield_per(5):
-            yield {'updated_at': rule[0], 'state': rule[1], 'locks_ok_cnt': rule[2], 'locks_stuck_cnt': rule[3], 'locks_replicating_cnt': rule[4]}
+        for rule in session.execute(stmt).yield_per(5):
+            yield rule._asdict()
     except StatementError as exc:
         raise RucioException('Badly formatted input (IDs?)') from exc
 
@@ -909,22 +1046,27 @@ def list_rule_full_history(
     :raises:        RucioException
     """
 
-    query = session.query(models.ReplicationRuleHistory.id,
-                          models.ReplicationRuleHistory.created_at,
-                          models.ReplicationRuleHistory.updated_at,
-                          models.ReplicationRuleHistory.rse_expression,
-                          models.ReplicationRuleHistory.state,
-                          models.ReplicationRuleHistory.account,
-                          models.ReplicationRuleHistory.locks_ok_cnt,
-                          models.ReplicationRuleHistory.locks_stuck_cnt,
-                          models.ReplicationRuleHistory.locks_replicating_cnt).\
-        with_hint(models.ReplicationRuleHistory, "INDEX(RULES_HISTORY_SCOPENAME_IDX)", 'oracle').\
-        filter(models.ReplicationRuleHistory.scope == scope, models.ReplicationRuleHistory.name == name).\
-        order_by(models.ReplicationRuleHistory.created_at, models.ReplicationRuleHistory.updated_at)
-
-    for rule in query.yield_per(5):
-        yield {'rule_id': rule[0], 'created_at': rule[1], 'updated_at': rule[2], 'rse_expression': rule[3], 'state': rule[4],
-               'account': rule[5], 'locks_ok_cnt': rule[6], 'locks_stuck_cnt': rule[7], 'locks_replicating_cnt': rule[8]}
+    stmt = select(
+        models.ReplicationRuleHistory.id.label('rule_id'),
+        models.ReplicationRuleHistory.created_at,
+        models.ReplicationRuleHistory.updated_at,
+        models.ReplicationRuleHistory.rse_expression,
+        models.ReplicationRuleHistory.state,
+        models.ReplicationRuleHistory.account,
+        models.ReplicationRuleHistory.locks_ok_cnt,
+        models.ReplicationRuleHistory.locks_stuck_cnt,
+        models.ReplicationRuleHistory.locks_replicating_cnt
+    ).with_hint(
+        models.ReplicationRuleHistory, 'INDEX(RULES_HISTORY_SCOPENAME_IDX)', 'oracle'
+    ).where(
+        and_(models.ReplicationRuleHistory.scope == scope,
+             models.ReplicationRuleHistory.name == name)
+    ).order_by(
+        models.ReplicationRuleHistory.created_at,
+        models.ReplicationRuleHistory.updated_at
+    )
+    for rule in session.execute(stmt).yield_per(5):
+        yield rule._asdict()
 
 
 @stream_session
@@ -944,13 +1086,22 @@ def list_associated_rules_for_file(
     """
 
     rucio.core.did.get_did(scope=scope, name=name, session=session)  # Check if the did acually exists
-    query = session.query(models.ReplicationRule).\
-        with_hint(models.ReplicaLock, "INDEX(LOCKS LOCKS_PK)", 'oracle').\
-        join(models.ReplicaLock, models.ReplicationRule.id == models.ReplicaLock.rule_id).\
-        filter(models.ReplicaLock.scope == scope, models.ReplicaLock.name == name).distinct()
+    stmt = select(
+        models.ReplicationRule
+    ).distinct(
+    ).join(
+        models.ReplicaLock,
+        models.ReplicationRule.id == models.ReplicaLock.rule_id
+    ).with_hint(
+        models.ReplicaLock, 'INDEX(LOCKS LOCKS_PK)', 'oracle'
+    ).where(
+        and_(models.ReplicaLock.scope == scope,
+             models.ReplicaLock.name == name)
+    )
+
     try:
-        for rule in query.yield_per(5):
-            yield rule.to_dict()
+        for result in session.execute(stmt).yield_per(5):
+            yield result[0].to_dict()
     except StatementError as exc:
         raise RucioException('Badly formatted input (IDs?)') from exc
 
@@ -982,9 +1133,14 @@ def delete_rule(
 
     with METRICS.timer('delete_rule.total'):
         try:
-            rule = session.query(models.ReplicationRule)\
-                          .filter(models.ReplicationRule.id == rule_id)\
-                          .with_for_update(nowait=nowait).one()
+            stmt = select(
+                models.ReplicationRule
+            ).where(
+                models.ReplicationRule.id == rule_id
+            ).with_for_update(
+                nowait=nowait
+            )
+            rule = session.execute(stmt).scalar_one()
         except NoResultFound as exc:
             raise RuleNotFound('No rule with the id %s found' % rule_id) from exc
         if rule.locked and not ignore_rule_lock:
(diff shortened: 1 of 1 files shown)


lib/rucio/core/rule.py
```diff
@@ -674,7 +769,13 @@ def inject_rule(rule_id, *, session: "Session", logger=logging.log):
...
         dids = [dict(t) for t in {tuple(d.items()) for d in dids}]
         # Remove dids which already have a similar rule
-        dids = [did for did in dids if session.query(models.ReplicationRule).filter_by(scope=did['scope'], name=did['name'], account=rule.account, rse_expression=rule.rse_expression).count() == 0]
+        stmt = select(
+            models.ReplicationRule
+        ).where(
+            and_(models.ReplicationRule.account == rule.account,
+                 models.ReplicationRule.rse_expression == rule.rse_expression)
+        )
+        dids = [did for did in dids if session.execute(stmt.where(and_(models.ReplicationRule.scope == did['scope'], models.ReplicationRule.name == did['name']))).scalar_one_or_none() is None]
```
bari12, line 818: While this would do the same, I wonder if internally something different happens though in the ORM. 
The original version does a count, thus it only needs to transmit one value from the DB to the client. I wonder if in the `scalar_one_or_none` version more is transmitted, and then this is evaluated client side. Thus more data flows between DB and client (server)

lib/rucio/core/rule.py
```diff
@@ -890,12 +1027,20 @@ def list_associated_rules_for_file(scope, name, *, session: "Session"):
...
+        models.ReplicaLock,
+        models.ReplicationRule.id == models.ReplicaLock.rule_id
+    ).where(
+        and_(models.ReplicaLock.scope == scope,
+             models.ReplicaLock.name == name)
+    ).distinct()
+
     try:
-        for rule in query.yield_per(5):
+        for rule in session.execute(stmt).yield_per(5).scalars():
```
bari12: I wonder if .scalars() should be moved to the next line. Thus if the generator works like this correctly

lib/rucio/core/rule.py
```diff
@@ -944,9 +1094,14 @@ def delete_rule(rule_id, purge_replicas=None, soft=False, delete_parent=False, n
...
-                       .filter(models.ReplicaLock.rule_id == rule_id)\
-                       .with_for_update(nowait=nowait).yield_per(100)
+        stmt = select(
+            models.ReplicaLock
+        ).with_for_update(
+            nowait=nowait
+        ).where(
+            models.ReplicaLock.rule_id == rule_id
+        )
+        locks = session.execute(stmt).yield_per(100).scalars()
```
bari12: Again, is this correct to do the .scalars() at this point?

lib/rucio/core/rule.py
```diff
@@ -362,19 +386,40 @@ def add_rule(dids, account, copies, rse_expression, grouping, weight, lifetime,
...
                 if new_rule.grouping != RuleGrouping.NONE:
-                    session.query(models.DatasetLock).filter_by(rule_id=new_rule.id).update({'state': LockState.STUCK})
+                    stmt = update(
+                        models.DatasetLock
+                    ).where(
+                        models.DatasetLock.scope == new_rule.scope
+                    ).values(
+                        {'state': LockState.STUCK}
+                    )
+                    session.execute(stmt)
```
rdimaio: The filter condition is different here:
- old: `rule_id=new_rule.id`
- new: `models.DatasetLock.scope == new_rule.scope`

The result is probably the same, but it would be great if you could clarify why this was changed

lib/rucio/core/rule.py
```diff
@@ -427,33 +472,55 @@ def add_rules(dids, rules, *, session: "Session", logger=logging.log):
...
                     if archive is not None:
                         elem['name'] = archive.name
                         elem['scope'] = archive.scope
                         try:
-                            did = session.query(models.DataIdentifier).filter(models.DataIdentifier.scope == elem['scope'],
-                                                                              models.DataIdentifier.name == elem['name']).one()
+                            stmt = select(
+                                models.DataIdentifier
+                            ).where(
+                                and_(models.DataIdentifier.scope == elem['scopea'],
```
rdimaio: `scopea` seems to be a typo, should be `scope`

lib/rucio/core/rule.py
rdimaio: From the docs: https://docs.sqlalchemy.org/en/14/core/connections.html#sqlalchemy.engine.Result.scalar_one_or_none 

> Return exactly one scalar result or None. This is equivalent to calling [Result.scalars()](https://docs.sqlalchemy.org/en/14/core/connections.html#sqlalchemy.engine.Result.scalars) and then [Result.one_or_none()](https://docs.sqlalchemy.org/en/14/core/connections.html#sqlalchemy.engine.Result.one_or_none).

I wouldn't expect a significant performance difference between using `count` and `scalar_one_or_none`

lib/rucio/core/rule.py
```diff
@@ -818,9 +945,9 @@ def list_rules(filters={}, *, session: "Session"):
             stmt = stmt.where(getattr(models.ReplicationRule, key) == value)
 
     try:
-        for rule, data_identifier_bytes in session.execute(stmt).yield_per(5):
-            d = rule.to_dict()
-            d['bytes'] = data_identifier_bytes
+        for item in session.execute(stmt).yield_per(5):
+            d = item[0].to_dict()
+            d['bytes'] = item[1]
```
rdimaio: could you clarify why this change was made?

lib/rucio/core/rule.py
```diff
@@ -860,20 +993,24 @@ def list_rule_full_history(scope, name, *, session: "Session"):
...
-    query = session.query(models.ReplicationRuleHistory.id,
-                          models.ReplicationRuleHistory.created_at,
-                          models.ReplicationRuleHistory.updated_at,
-                          models.ReplicationRuleHistory.rse_expression,
-                          models.ReplicationRuleHistory.state,
-                          models.ReplicationRuleHistory.account,
-                          models.ReplicationRuleHistory.locks_ok_cnt,
-                          models.ReplicationRuleHistory.locks_stuck_cnt,
-                          models.ReplicationRuleHistory.locks_replicating_cnt).\
-        with_hint(models.ReplicationRuleHistory, "INDEX(RULES_HISTORY_SCOPENAME_IDX)", 'oracle').\
```
rdimaio, line 921: this `with_hint()` is missing from the new query

lib/rucio/core/rule.py
```diff
@@ -1943,7 +2263,14 @@ def update_rules_for_lost_replica(scope, name, rse_id, nowait=False, *, session:
         elif rule.locks_replicating_cnt == 0 and rule.locks_stuck_cnt == 0:
             rule.state = RuleState.OK
             if rule.grouping != RuleGrouping.NONE:
-                session.query(models.DatasetLock).filter_by(rule_id=rule.id).update({'state': LockState.OK})
+                stmt = update(
+                    models.DatasetLock
+                ).where(
+                    models.DatasetLock.rule_id == rule.id
+                ).values(
+                    state=LockState.OK
```
rdimaio: I'm not sure if this would work - you might need to pass it like this instead: `{'state': LockState.OK}` (I'm assuming this based on the other update statements in this commit)

lib/rucio/core/rule.py
```diff
@@ -2040,15 +2412,41 @@ def update_rules_for_bad_replica(scope, name, rse_id, nowait=False, *, session:
...
     if nlock:
-        session.query(models.RSEFileAssociation).filter(models.RSEFileAssociation.scope == scope, models.RSEFileAssociation.name == name, models.RSEFileAssociation.rse_id == rse_id).update({'state': ReplicaState.COPYING})
+        stmt = update(
+            models.RSEFileAssociation
+        ).where(
+            and_(models.DataIdentifier.scope == scope,
+                 models.RSEFileAssociation.rse_id == rse_id,
+                 models.DataIdentifier.name == name)
+        ).values(
+            {'availability': ReplicaState.COPYING}
```
rdimaio: I think this should be `state`, not `availability`?

lib/rucio/core/rule.py
```diff
@@ -2674,7 +3156,14 @@ def __evaluate_did_detach(eval_did, *, session: "Session", logger=logging.log):
             elif rule.locks_replicating_cnt == 0 and rule.locks_stuck_cnt == 0:
                 rule.state = RuleState.OK
                 if rule.grouping != RuleGrouping.NONE:
-                    session.query(models.DatasetLock).filter_by(rule_id=rule.id).update({'state': LockState.OK})
+                    stmt = update(
+                        models.DatasetLock
+                    ).where(
+                        models.DatasetLock.rule_id == rule.id
+                    ).values(
+                        state=LockState.OK
```
rdimaio, line 3164: This might need to be changed to `{'state': LockState.OK}` to work

lib/rucio/core/rule.py
```diff
@@ -2813,7 +3316,14 @@ def __evaluate_did_attach(eval_did, *, session: "Session", logger=logging.log):
                             insert_rule_history(rule=rule, recent=True, longterm=False, session=session)
                             # Try to update the DatasetLocks
                             if rule.grouping != RuleGrouping.NONE:
-                                session.query(models.DatasetLock).filter_by(rule_id=rule.id).update({'state': LockState.STUCK})
+                                stmt = update(
+                                    models.DatasetLock
+                                ).where(
+                                    models.DatasetLock.rule_id == rule.id
+                                ).values(
+                                    state=LockState.STUCK
```
rdimaio, line 3324: This might need to be changed to `{'state': LockState.STUCK}`

lib/rucio/core/rule.py
rdimaio: It should work anyway, based on the docs https://docs.sqlalchemy.org/en/20/core/dml.html#sqlalchemy.sql.expression.update

feel free to ignore this comment

lib/rucio/core/rule.py
erlingstaff: You're right. Changed now to use state instead

lib/rucio/core/rule.py
erlingstaff: Reintroduced with_hint() in newest commit

lib/rucio/core/rule.py
erlingstaff: I believe I had to change this when I changed the query statement, but I then reverted the query statement without also reverting this. Reverted the generator back to:

`for rule, data_identifier_bytes in session.execute(stmt).yield_per(5):`

lib/rucio/core/rule.py
erlingstaff: pretty sure this is a vim typo.. changed to just `scope`

lib/rucio/core/rule.py
erlingstaff: You're right, changed this to use the same filter conditions as previously:

`models.DatasetLock.rule_id == new_rule.id`

lib/rucio/core/rule.py
erlingstaff: Yes, you are correct. This does incur a somewhat larger data transfer (albeit it is just one row), changed it to a single value (id) instead of fetching the entire ReplicationRule object

lib/rucio/core/rule.py
erlingstaff: I am uncertain about how this is handled internally, will write a test script and figure it out. However, scalars are apparently not needed here, as the result is unpacked regardless. Removed .scalars() and returning

`rule.to_dict()`

lib/rucio/core/rule.py
erlingstaff: Scalars are not needed here either, removed it. As mentioned, I will figure out how yield_per().scalars() is handled for future use cases.

rdimaio (dismissed):

lib/rucio/core/rule.py
```diff
@@ -24,7 +24,7 @@
 from typing import TYPE_CHECKING, Any, Callable, Optional, Type, TypeVar
 
 from dogpile.cache.api import NO_VALUE
-from sqlalchemy import select, update
+from sqlalchemy import select, update, delete, desc
```
dchristidis: Please sort alphabetically.

lib/rucio/core/rule.py
```diff
@@ -246,23 +251,42 @@ def add_rule(dids, account, copies, rse_expression, grouping, weight, lifetime,
...
-                    .filter(models.RSEFileAssociation.scope == did.scope,
-                            models.RSEFileAssociation.name == did.name,
-                            models.RSEFileAssociation.state == ReplicaState.AVAILABLE,
-                            models.RSE.rse_type != RSEType.TAPE).count()
+                stmt = select(
+                    func.count(models.RSEFileAssociation.rse_id)
+                ).join(
+                    models.RSE,
+                    models.RSE.id == models.RSEFileAssociation.rse_id
+                ).where(
```
dchristidis: I wonder why you opted to reverse the order of this expression? In my mind, `models.RSEFileAssociation.rse_id == models.RSE.id` is what is expected because `RSEFileAssociation` is on the left of `RSE` during the `JOIN` operation (which, if I’m not mistaken, in rare cases can actually affect performance).

This comment applies throughout the file.

lib/rucio/core/rule.py
```diff
@@ -362,19 +386,40 @@ def add_rule(dids, account, copies, rse_expression, grouping, weight, lifetime,
...
                 new_rule.error = 'MissingSourceReplica'
                 if new_rule.grouping != RuleGrouping.NONE:
-                    session.query(models.DatasetLock).filter_by(rule_id=new_rule.id).update({'state': LockState.STUCK})
+                    stmt = update(
+                        models.DatasetLock
+                    ).where(
+                        models.DatasetLock.rule_id == new_rule.id
+                    ).values(
+                        {'state': LockState.STUCK}
+                    )
```
dchristidis: 1. We should avoid hard-coded strings.
 2. I would like to suggest that the dictionary brackets are put together with the parentheses. I believe this improves readability.

```python
.values({
    models.DatasetLock.state: LockState.STUCK
})
```

This comment applies throughout the file.

lib/rucio/core/rule.py
```diff
@@ -362,19 +386,40 @@ def add_rule(dids, account, copies, rse_expression, grouping, weight, lifetime,
...
                 if new_rule.grouping != RuleGrouping.NONE:
-                    session.query(models.DatasetLock).filter_by(rule_id=new_rule.id).update({'state': LockState.OK})
+                    stmt = update(
+                        models.DatasetLock
+                    ).where(
+                        models.DatasetLock.scope == new_rule.scope
+                    ).values(
+                        {'state': LockState.OK}
+                    )
+                    session.execute(stmt)
```
dchristidis: Could I ask you to double-check this? `DatasetLock.rule_id == new_rule.id` being replaced with `DatasetLock.scope == new_rule.scope`? Same for the query bellow.

lib/rucio/core/rule.py
```diff
@@ -246,23 +251,42 @@ def add_rule(dids, account, copies, rse_expression, grouping, weight, lifetime,
...
             if did.did_type == DIDType.FILE and did.constituent:
                 # Check if a single replica of this DID exists; Do not put rule on file if there are only replicas on TAPE
-                replica_cnt = session.query(models.RSEFileAssociation).join(models.RSE, models.RSEFileAssociation.rse_id == models.RSE.id)\
-                    .filter(models.RSEFileAssociation.scope == did.scope,
-                            models.RSEFileAssociation.name == did.name,
-                            models.RSEFileAssociation.state == ReplicaState.AVAILABLE,
-                            models.RSE.rse_type != RSEType.TAPE).count()
+                stmt = select(
+                    func.count(models.RSEFileAssociation.rse_id)
+                ).join(
```
dchristidis: I see what you’re doing here, but I think it makes the query confusing.

```python
>>> print(select(func.count(models.RSEFileAssociation.rse_id)))
SELECT count(atlas_rucio.replicas.rse_id) AS count_1 
FROM atlas_rucio.replicas
```

It makes the reader assume that `rse_id` is somehow significant. What I would suggest instead is:

```python
stmt = select(
    func.count()
).select_from(
    models.RSEFileAssociation
)
```

which makes the query more obvious:

```python
>>> print(select(func.count()).select_from(models.RSEFileAssociation))
SELECT count(*) AS count_1 
FROM atlas_rucio.replicas
```

This comment applies throughout the file.

lib/rucio/core/rule.py
```diff
@@ -427,33 +472,55 @@ def add_rules(dids, rules, *, session: "Session", logger=logging.log):
...
                         try:
-                            did = session.query(models.DataIdentifier).filter(models.DataIdentifier.scope == elem['scope'],
-                                                                              models.DataIdentifier.name == elem['name']).one()
+                            stmt = select(
+                                models.DataIdentifier
+                            ).where(
+                                and_(models.DataIdentifier.scope == elem['scope'],
+                                     models.DataIdentifier.name == elem['name'])
+                            )
+                            did = session.execute(stmt).scalars().one()
```
dchristidis: Why `scalars().one()` instead of `scalar_one()`?

lib/rucio/core/rule.py
```diff
@@ -836,14 +963,20 @@ def list_rule_history(rule_id, *, session: "Session"):
...
+        models.ReplicationRuleHistory.id,
+        models.ReplicationRuleHistory.state,
+        models.ReplicationRuleHistory.locks_ok_cnt,
+        models.ReplicationRuleHistory.locks_stuck_cnt,
+        models.ReplicationRuleHistory.locks_replicating_cnt
+    ).where(
+        models.ReplicationRuleHistory.id == rule_id
+    ).order_by(
+        models.ReplicationRuleHistory.created_at
+    )
```
dchristidis: Why is `ReplicationRuleHistoryRecent.updated_at` replaced with `models.ReplicationRuleHistory.id`?

lib/rucio/core/rule.py
```diff
@@ -836,14 +963,20 @@ def list_rule_history(rule_id, *, session: "Session"):
...
 
     try:
-        for rule in query.yield_per(5):
+        for rule in session.execute(stmt).yield_per(5):
             yield {'updated_at': rule[0], 'state': rule[1], 'locks_ok_cnt': rule[2], 'locks_stuck_cnt': rule[3], 'locks_replicating_cnt': rule[4]}
```
dchristidis: I think we can improve this:

```python
yield rule._asdict()
```

lib/rucio/core/rule.py
```diff
@@ -860,20 +993,26 @@ def list_rule_full_history(scope, name, *, session: "Session"):
...
+    ).where(
+        and_(models.ReplicationRuleHistory.scope == scope,
+             models.ReplicationRuleHistory.name == name)
+    ).order_by(
+        models.ReplicationRuleHistory.created_at,
+        models.ReplicationRuleHistory.updated_at
+    )
+    for rule in session.execute(stmt).yield_per(5):
         yield {'rule_id': rule[0], 'created_at': rule[1], 'updated_at': rule[2], 'rse_expression': rule[3], 'state': rule[4],
                'account': rule[5], 'locks_ok_cnt': rule[6], 'locks_stuck_cnt': rule[7], 'locks_replicating_cnt': rule[8]}
```
dchristidis: My previous comment applies here to, but you would have to use:
```python
models.ReplicationRuleHistory.id.label('rule_id')
```

lib/rucio/core/rule.py
```diff
@@ -890,13 +1029,21 @@ def list_associated_rules_for_file(scope, name, *, session: "Session"):
...
     rucio.core.did.get_did(scope=scope, name=name, session=session)  # Check if the did acually exists
-    query = session.query(models.ReplicationRule).\
-        with_hint(models.ReplicaLock, "INDEX(LOCKS LOCKS_PK)", 'oracle').\
-        join(models.ReplicaLock, models.ReplicationRule.id == models.ReplicaLock.rule_id).\
-        filter(models.ReplicaLock.scope == scope, models.ReplicaLock.name == name).distinct()
+    stmt = select(
+        models.ReplicationRule
+    ).with_hint(
+        models.ReplicationRule, "INDEX(RULES RULES_PK)", 'oracle'
+    ).join(
```
dchristidis: You’ve changed the hint. Is that wise?

lib/rucio/core/rule.py
```diff
@@ -890,13 +1029,21 @@ def list_associated_rules_for_file(scope, name, *, session: "Session"):
...
+        models.ReplicaLock,
+        models.ReplicationRule.id == models.ReplicaLock.rule_id
+    ).where(
+        and_(models.ReplicaLock.scope == scope,
+             models.ReplicaLock.name == name)
+    ).distinct()
+
```
dchristidis: I wonder if it’s preferable to have `distinct()` right after `select()` because that’s where it goes in actual SQL syntax (of course, the resulting query should be the same, but I find it more ‘natural’ this way.)

lib/rucio/core/rule.py
```diff
@@ -890,13 +1029,21 @@ def list_associated_rules_for_file(scope, name, *, session: "Session"):
...
     try:
-        for rule in query.yield_per(5):
-            yield rule.to_dict()
+        for result in session.execute(stmt).yield_per(5):
+            yield result[0].to_dict()
```
dchristidis, line 1104: Is there a reason you can’t use `scalars()` here?

lib/rucio/core/rule.py
```diff
@@ -920,9 +1067,14 @@ def delete_rule(rule_id, purge_replicas=None, soft=False, delete_parent=False, n
...
-            rule = session.query(models.ReplicationRule)\
-                          .filter(models.ReplicationRule.id == rule_id)\
-                          .with_for_update(nowait=nowait).one()
+            stmt = select(
+                models.ReplicationRule
+            ).with_for_update(
+                nowait=nowait
+            ).where(
+                models.ReplicationRule.id == rule_id
+            )
```
dchristidis: Similarly, I think `with_for_update()` should come after `where()`.

This comment applies throughout the file.

lib/rucio/core/rule.py
```diff
@@ -944,15 +1096,21 @@ def delete_rule(rule_id, purge_replicas=None, soft=False, delete_parent=False, n
...
+        )
+        results = session.execute(stmt).yield_per(100)
 
         # Remove locks, set tombstone if applicable
         transfers_to_delete = []  # [{'scope': , 'name':, 'rse_id':}]
         account_counter_decreases = {}  # {'rse_id': [file_size, file_size, file_size]}
 
-        for lock in locks:
+        for result in results:
+            lock = result[0]
```
dchristidis, line 1179: Here, too, I don’t quite understand the need for the rename and use a subscript, where `scalars()` would avoid it.

lib/rucio/core/rule.py
```diff
@@ -1789,18 +2060,25 @@ def get_expired_rules(total_workers, worker_number, limit=100, blocked_rules=[],
...
-        order_by(models.ReplicationRule.expires_at)  # NOQA
-
-    query = filter_thread_work(session=session, query=query, total_threads=total_workers, thread_id=worker_number, hash_variable='name')
+    stmt = select(
+        models.ReplicationRule.id,
+        models.ReplicationRule.rse_expression
+    ).where(
+        and_(models.ReplicationRule.expires_at < datetime.utcnow(),
+             models.ReplicationRule.locked == false(),
+             models.ReplicationRule.child_rule_id == None)  # NOQA: E711
```
dchristidis: Replace with:

```python
models.ReplicationRule.child_rule_id == null()
```

lib/rucio/core/rule.py
```diff
@@ -1789,18 +2060,25 @@ def get_expired_rules(total_workers, worker_number, limit=100, blocked_rules=[],
...
+    ).with_hint(
+        models.ReplicationRule, "index(rules RULES_EXPIRES_AT_IDX)", 'oracle'
+    ).order_by(
```
dchristidis: Similar to my previous comments, I would find it more natural if `with_hint()` is ordered before `where()`.

This comment applies throughout the file.

lib/rucio/core/rule.py
```diff
@@ -1789,18 +2060,25 @@ def get_expired_rules(total_workers, worker_number, limit=100, blocked_rules=[],
...
+    )
+    stmt = filter_thread_work(session=session, query=stmt, total_threads=total_workers, thread_id=worker_number, hash_variable='name')
 
     if limit:
-        fetched_rules = query.limit(limit).all()
-        filtered_rules = [rule for rule in fetched_rules if rule[0] not in blocked_rules]
-        if len(fetched_rules) == limit and not filtered_rules:
+        stmt = stmt.limit(limit)
+        result = session.execute(stmt).all()
+        filtered_rules = [rule for rule in result if rule[0] not in blocked_rules]
```
dchristidis: I wonder if `rule['id']` is preferable to `rule[0]`.

This comment applies throughout the file.

lib/rucio/core/rule.py
```diff
@@ -1910,9 +2203,32 @@ def update_rules_for_lost_replica(scope, name, rse_id, nowait=False, *, session:
...
-    locks = session.query(models.ReplicaLock).filter(models.ReplicaLock.scope == scope, models.ReplicaLock.name == name, models.ReplicaLock.rse_id == rse_id).with_for_update(nowait=nowait).all()
-    replica = session.query(models.RSEFileAssociation).filter(models.RSEFileAssociation.scope == scope, models.RSEFileAssociation.name == name, models.RSEFileAssociation.rse_id == rse_id).with_for_update(nowait=nowait).one()
-    requests = session.query(models.Request).filter(models.Request.scope == scope, models.Request.name == name, models.Request.dest_rse_id == rse_id).with_for_update(nowait=nowait).all()
+    stmt = select(
+        models.ReplicaLock
+    ).where(
+        and_(models.ReplicaLock.scope == scope,
+             models.ReplicaLock.name == name,
+             models.ReplicaLock.rse_id == rse_id)
+    ).with_for_update(nowait=nowait)
```
dchristidis: Shouldn’t `with_for_update(nowait=nowait)` be split into three lines? Same below.

lib/rucio/core/rule.py
```diff
@@ -1910,9 +2203,32 @@ def update_rules_for_lost_replica(scope, name, rse_id, nowait=False, *, session:
...
+    locks = session.execute(stmt).all()
```
dchristidis: You’ve forgotten a `scalars()` here. Same for the third query.

lib/rucio/core/rule.py
```diff
@@ -1943,7 +2266,14 @@ def update_rules_for_lost_replica(scope, name, rse_id, nowait=False, *, session:
...
             rule.state = RuleState.OK
             if rule.grouping != RuleGrouping.NONE:
-                session.query(models.DatasetLock).filter_by(rule_id=rule.id).update({'state': LockState.OK})
+                stmt = update(
+                    models.DatasetLock
+                ).where(
+                    models.DatasetLock.rule_id == rule.id
+                ).values(
+                    state=LockState.OK
+                )
```
dchristidis: I think it would be preferable if we only used dictionaries instead of keyword parameter values, for the same reason as one of my initial comments. So this should be:

```python
.values({
    models.DatasetLock.state: LockState.OK
})
```

This comment applies throughout the file.

lib/rucio/core/rule.py
```diff
@@ -2040,15 +2415,41 @@ def update_rules_for_bad_replica(scope, name, rse_id, nowait=False, *, session:
...
         # Insert rule history
         insert_rule_history(rule=rule, recent=True, longterm=False, session=session)
     if nlock:
-        session.query(models.RSEFileAssociation).filter(models.RSEFileAssociation.scope == scope, models.RSEFileAssociation.name == name, models.RSEFileAssociation.rse_id == rse_id).update({'state': ReplicaState.COPYING})
+        stmt = update(
+            models.RSEFileAssociation
+        ).where(
+            and_(models.DataIdentifier.scope == scope,
+                 models.RSEFileAssociation.rse_id == rse_id,
+                 models.DataIdentifier.name == name)
```
dchristidis: 1. `DataIdentifier` has no place here.
 2. It would be more natural for the order to be `scope`, `name`, `rse_id`.

lib/rucio/core/rule.py
```diff
@@ -2382,10 +2808,25 @@ def examine_rule(rule_id, *, session: "Session"):
...
+                stmt = select(
+                    models.RequestHistory
+                ).where(
+                    and_(models.RequestHistory.scope == lock.scope,
+                         models.RequestHistory.name == lock.name,
+                         models.RequestHistory.dest_rse_id == lock.rse_id)
+                ).order_by(
+                    desc(models.RequestHistory.created_at)
+                )
+                transfers = session.execute(stmt).all()
```
dchristidis: Missing `scalars()`.

lib/rucio/core/rule.py
```diff
@@ -2740,12 +3237,16 @@ def __evaluate_did_attach(eval_did, *, session: "Session", logger=logging.log):
...
-                .filter(models.DataIdentifierAssociation.scope == eval_did.scope,
-                        models.DataIdentifierAssociation.name == eval_did.name,
-                        models.DataIdentifierAssociation.rule_evaluation == True).all()  # noqa
-
+            stmt = select(
+                models.DataIdentifierAssociation
+            ).where(
+                and_(models.DataIdentifierAssociation.scope == eval_did.scope,
+                     models.DataIdentifierAssociation.name == eval_did.name,
+                     models.DataIdentifierAssociation.rule_evaluation == True)  # pylint: disable=C0121:singleton-comparison  # noqa: E712
```
dchristidis: Use `true()` instead.

lib/rucio/core/rule.py
```diff
@@ -2755,12 +3256,17 @@ def __evaluate_did_attach(eval_did, *, session: "Session", logger=logging.log):
...
-                    models.ReplicationRule.state != RuleState.WAITING_APPROVAL,
-                    models.ReplicationRule.state != RuleState.INJECT).with_for_update(nowait=True).all()
-
+                stmt = select(
+                    models.ReplicationRule
+                ).where(
+                    and_(or_(*rule_clauses),
+                         models.ReplicationRule.state != RuleState.SUSPENDED,
+                         models.ReplicationRule.state != RuleState.WAITING_APPROVAL,
+                         models.ReplicationRule.state != RuleState.INJECT)
```
dchristidis: I wonder if this is better expressed as:


```python
models.ReplicationRule.state.not_in([RuleState.SUSPENDED,
                                     RuleState.WAITING_APPROVAL,
                                     RuleState.INJECT])
```

lib/rucio/core/rule.py
```diff
@@ -860,20 +993,26 @@ def list_rule_full_history(scope, name, *, session: "Session"):
...
+        models.ReplicationRuleHistory.created_at,
+        models.ReplicationRuleHistory.updated_at,
+        models.ReplicationRuleHistory.rse_expression,
+        models.ReplicationRuleHistory.state,
+        models.ReplicationRuleHistory.account,
+        models.ReplicationRuleHistory.locks_ok_cnt,
+        models.ReplicationRuleHistory.locks_stuck_cnt,
+        models.ReplicationRuleHistory.locks_replicating_cnt
+    ).with_hint(
+        models.ReplicationRuleHistory, "INDEX(RULES_HISTORY_SCOPENAME_IDX)", 'oracle'
```
dchristidis: It never made sense to me that we have two quoting styles in the same line. Maybe it’s high time we fixed this? I also think we should use all-caps for SQL functions (this applies to other line like `index(CONTENTS CONTENTS_PK)`).

This comment applies throughout the file.

dchristidis (changes requested):
Looks promising! Most of my comments are of an aesthetic nature, but some are genuine issues and I’m concerned that our tests didn’t pick up on them.

lib/rucio/core/rule.py
erlingstaff: Sorted, now: "delete, desc, select, update"

lib/rucio/core/rule.py
erlingstaff: I agree, changed this and other occurences to have base first, like:
`models.RSEFileAssociation.rse_id == models.RSE.id `

lib/rucio/core/rule.py
erlingstaff: Changed all Update statements to follow this convention

lib/rucio/core/rule.py
erlingstaff: This looks like an error. Reverted to the previous query:
`DatasetLock.rule_id == new_rule.id`

lib/rucio/core/rule.py
erlingstaff: Changed all occurrences of func.count() to also have an accompanying select_from

lib/rucio/core/rule.py
erlingstaff: They are synonymous, but no real reason other than that. Changed it so that `scalar_one()` is always used now instead of `scalar().one()`

lib/rucio/core/rule.py
erlingstaff: This also looks like an error, fixed it in the newest commit

lib/rucio/core/rule.py
erlingstaff: Good idea, this looks a lot better

lib/rucio/core/rule.py
erlingstaff: Likely not, reverted to previous hint:
`models.ReplicaLock, 'INDEX(LOCKS LOCKS_PK)', 'oracle'`

lib/rucio/core/rule.py
erlingstaff: Moved this to right after select:

`stmt = select(models.ReplicationRule).distinct().[...]`

lib/rucio/core/rule.py
erlingstaff: This was the potential issue brought up in https://github.com/rucio/rucio/pull/6493#discussion_r1496013829, I decided to keep it like this (without `.scalars()`) in case it is a performance drain, but i am very open to rewrite this to:

`for result in session.execute(stmt).yield_per(5).scalars():`

lib/rucio/core/rule.py
erlingstaff: Refactored all queries with a `with_for_update` to order it after `where`

lib/rucio/core/rule.py
erlingstaff: Same situation as in https://github.com/rucio/rucio/pull/6493#discussion_r1519879430

lib/rucio/core/rule.py
erlingstaff: Done, and removed QA comment

lib/rucio/core/rule.py
erlingstaff: Refactored all `with_hint()` queries to order it before `where()`

lib/rucio/core/rule.py
erlingstaff: To solve this, I replaced all occurrences of this issue with indexing akin to this:

`filtered_rules = [rule for rule in result if rule._mapping['id'] not in blocked_rules]`

`rule['id']` was unavailable as this query does not retrieve the entire object (just models.ReplicationRule.id and models.ReplicationRule.rse_expression), so I used _mapping to index it.

lib/rucio/core/rule.py
erlingstaff: Fixed, now:

`locks = session.execute(stmt).scalars().all()`
`requests = session.execute(stmt).scalars().all()`

lib/rucio/core/rule.py
erlingstaff: Changed, now:
`transfers = session.execute(stmt).scalars().all()`

lib/rucio/core/rule.py
erlingstaff: Refactored the quoting style for `with_hint()` to use exclusively single quotes, like:

` with_hint(
        models.ReplicationRule, 'INDEX(RULES RULES_CHILD_RULE_ID_IDX)', 'oracle'`

`with_hint` Now also always using caps for SQL functions

lib/rucio/core/rule.py
erlingstaff: Changed this and other occurrences of it to a 3-line split to follow stylistic convention

lib/rucio/core/rule.py
erlingstaff: All `values()` now use the style you mentioned, i.e.:
`values({
                        models.DatasetLock.state: LockState.STUCK
                    })`

lib/rucio/core/rule.py
erlingstaff: 1. Error, fixed
2. I agree, makes sense, changed

lib/rucio/core/rule.py
```diff
@@ -427,33 +474,57 @@ def add_rules(dids, rules, *, session: "Session", logger=logging.log):
...
 
             # 2.1 If the did is a constituent, relay the rule to the archive
             if did.did_type == DIDType.FILE and did.constituent:  # Check if a single replica of this DID exists
-                replica_cnt = session.query(models.RSEFileAssociation).join(models.RSE, models.RSEFileAssociation.rse_id == models.RSE.id)\
-                    .filter(models.RSEFileAssociation.scope == did.scope,
-                            models.RSEFileAssociation.name == did.name,
-                            models.RSEFileAssociation.state == ReplicaState.AVAILABLE,
-                            models.RSE.rse_type != RSEType.TAPE).count()
+                stmt = select(
+                    func.count(models.RSEFileAssociation.rse_id)
```
rdimaio: This could be rewritten in the same way as the other queries outlined here https://github.com/rucio/rucio/pull/6493#discussion_r1517310447

erlingstaff: Good catch, changed this

lib/rucio/core/rule.py
dchristidis: Please do the same for `get_expired_rules()`.

lib/rucio/core/rule.py
dchristidis: I’m afraid I do not understand your comment. Yes, the query comes with two individual columns, but I don’t see why that requires anything different.

```python
>>> stmt = select(
...         models.ReplicationRule.id,
...         models.ReplicationRule.rse_expression
...     ).where(
...         and_(models.ReplicationRule.expires_at < datetime.utcnow(),
...              models.ReplicationRule.locked == false(),
...              models.ReplicationRule.child_rule_id == null())
...     ).order_by(
...         models.ReplicationRule.expires_at
...     ).with_hint(
...         models.ReplicationRule, 'INDEX(RULES RULES_EXPIRES_AT_IDX)', 'oracle'
...     )
>>> result = session.execute(stmt).all()
>>> result[0]['id']
'53f65ded554b43c1accaf8c981f90c58'
```

dchristidis: Forgive me, this console session is inaccurate. It was done using SQLAlchemy version 1.4.46 and the behaviour appears to have changed in more recent versions. However, as discussed privately, [dot notation should work](https://docs.sqlalchemy.org/en/20/core/connections.html#sqlalchemy.engine.Row._mapping).

> This [`Row`](https://docs.sqlalchemy.org/en/20/core/connections.html#sqlalchemy.engine.Row "sqlalchemy.engine.Row") behaves like a named tuple, in that it acts as a sequence but also supports attribute name access, e.g. `row.some_column`.

I’m having difficulty deciding what is best; please do as you prefer.

EDIT: See [here](https://docs.sqlalchemy.org/en/20/changelog/migration_14.html#change-4710-core) for additional information.

erlingstaff: Rewrote the indexing to use dot notation instead of `_mapping` for all occurrences

lib/rucio/core/rule.py
erlingstaff: That one seemed to have slipped by. Rewrote it now

lib/rucio/core/rule.py
dchristidis: I do not like it, but let’s keep it as you have it for now.

lib/rucio/core/rule.py
```diff
@@ -1157,13 +1365,34 @@ def repair_rule(rule_id: str, *, session: "Session", logger: LoggerFunction = lo
...
-        dataset_locks = session.query(models.DatasetLock).filter_by(rule_id=rule.id).all()
+        stmt = select(
+            models.ReplicaLock.rse_id
+        ).where(
+            models.ReplicaLock.rule_id == rule.id
+        ).group_by(
+            models.ReplicaLock.rse_id
+        )
+        results = session.execute(stmt).all()
+        validated_datasetlock_rse_ids = [rse_id.rse_id for rse_id in results]
```
dchristidis: 1. What is the need for `results` and the list comprehension? Why not keep the statement as it is and then do:
    ```python
    validated_datasetlock_rse_ids = session.execute(stmt).scalars().all()
    ```
 2. Doesn’t `distinct()` make more sense than `group_by()`? This is a comment about the original query.

lib/rucio/core/rule.py
```diff
@@ -1695,13 +1970,17 @@ def re_evaluate_did(scope: InternalScope, name: str, rule_evaluation_action: DID
...
+            func.sum(models.DataIdentifierAssociation.bytes),
+            func.count(1)
+        ).with_hint(
+            models.DataIdentifierAssociation, 'INDEX(CONTENTS CONTENTS_PK)', 'oracle'
+        ).where(
+            and_(models.DataIdentifierAssociation.scope == scope,
+                 models.DataIdentifierAssociation.name == name)
+        )
+        result = session.execute(stmt)
+        for bytes_, length in result:
```
dchristidis: Similar question: what is the need for `result` to exist?

lib/rucio/core/rule.py
```diff
@@ -1760,18 +2040,22 @@ def get_rules_beyond_eol(date_check: datetime, worker_number: int, total_workers
...
+        models.ReplicationRule.account
+    ).where(
+        models.ReplicationRule.eol_at < date_check
+    )
 
-    query = filter_thread_work(session=session, query=query, total_threads=total_workers, thread_id=worker_number, hash_variable='name')
-    return [rule for rule in query.all()]
+    stmt = filter_thread_work(session=session, query=stmt, total_threads=total_workers, thread_id=worker_number, hash_variable='name')
+    result = session.execute(stmt).all()
+    return [rule for rule in result]  # type: ignore
```
dchristidis: Similar question: why `result` and a list comprehension?

lib/rucio/core/rule.py
```diff
@@ -1747,7 +2027,7 @@ def get_updated_dids(total_workers: int, worker_number: int, limit: int = 100, b
         else:
             return filtered_dids
     else:
-        return [did for did in query.order_by(models.UpdatedDID.created_at).all() if (did.scope, did.name) not in blocked_dids]
+        return [did for did in session.execute(stmt.order_by(models.UpdatedDID.created_at)).all() if (did.scope, did.name) not in blocked_dids]  # type: ignore
```
dchristidis: I don’t understand how the hint for the return type is `list[tuple[str, str]]`. I think putting `#type: ignore` is the wrong approach.

I’m afraid that #6516 may have introduced some mistakes.

This comment applies throughout the file.

lib/rucio/core/rule.py
```diff
@@ -2424,12 +2891,14 @@ def get_evaluation_backlog(expiration_time: int = 600, *, session: "Session") ->
...
+            func.count(models.UpdatedDID.created_at),
+            func.min(models.UpdatedDID.created_at)
+        )
         result = session.execute(stmt).scalars().one()
         REGION.set('rule_evaluation_backlog', result)
-        return result
-    else:
-        return cached_backlog  # type: ignore
+        return result  # type: ignore
+    return cached_backlog  # type: ignore
```
dchristidis: 1. In this case, the type hint is correct. Can you try to use `result._tuple()` to [coerce the type checker](https://docs.sqlalchemy.org/en/20/core/connections.html#sqlalchemy.engine.Row._tuple)?

 2. Maybe you need to use the following for the cached value?
    ```python
    cached_backlog: Union[NO_VALUE, tuple[int, datetime]] = REGION.get('rule_evaluation_backlog', expiration_time=expiration_time)
    ```

lib/rucio/core/rule.py
```diff
@@ -2693,9 +3168,26 @@ def __evaluate_did_detach(eval_did: models.DataIdentifier, *, session: "Session"
...
+            stmt = select(
+                models.ReplicationRule
+            ).where(
+                and_(models.ReplicationRule.scope == did['scope'],
+                     models.ReplicationRule.name == did['name'])
+            ).with_for_update(
+                nowait=True
+            )
+            parent_rules = list(session.execute(stmt).scalars().all())
+            rules.extend(parent_rules)
```
dchristidis: 1. Doesn’t `all()` already return a `list`?
2. Why the need for `parent_rules`?

lib/rucio/core/rule.py
```diff
@@ -2707,8 +3199,13 @@ def __evaluate_did_detach(eval_did: models.DataIdentifier, *, session: "Session"
...
             rule_locks_ok_cnt_before = rule.locks_ok_cnt
-            query = session.query(models.ReplicaLock).filter_by(rule_id=rule.id)
-            for lock in query:
+            stmt = select(
+                models.ReplicaLock
+            ).where(
+                models.ReplicaLock.rule_id == rule.id
+            )
+            locks = session.execute(stmt).scalars().all()
+            for lock in locks:
```
dchristidis: Similar to my other comments, the following would have been more obvious for me:
```python
for lock in session.execute(stmt).scalars():
```

lib/rucio/core/rule.py
```diff
@@ -2808,12 +3322,16 @@ def __evaluate_did_attach(eval_did: models.DataIdentifier, *, session: "Session"
...
-                        models.DataIdentifierAssociation.rule_evaluation == True).all()  # noqa
-
+            stmt = select(
+                models.DataIdentifierAssociation
+            ).with_hint(
+                models.DataIdentifierAssociation, 'INDEX_RS_ASC(CONTENTS CONTENTS_PK)', 'oracle'
+            ).where(
+                and_(models.DataIdentifierAssociation.scope == eval_did.scope,
+                     models.DataIdentifierAssociation.name == eval_did.name,
+                     models.DataIdentifierAssociation.rule_evaluation == true())  # pylint: disable=C0121:singleton-comparison  # noqa: E712
```
dchristidis: I’m surprised, why are these directives to checkers necessary? I though that was the point of using `true()`.

dchristidis (changes requested):
This is shaping up nicely. Following my second full review, I am now more confident that the new queries are correct.

lib/rucio/core/rule.py
dchristidis: You can fully ignore this function. It was addressed in #6552.

lib/rucio/core/rule.py
erlingstaff: 1. No real reason other than personal coding style, will follow this convention from now on.
2. Agree, changed to use distinct instead:

`stmt = select(
            models.ReplicaLock.rse_id
        ).distinct(
        ).where(
            models.ReplicaLock.rule_id == rule.id
        )`

lib/rucio/core/rule.py
erlingstaff: Same as 1. https://github.com/rucio/rucio/pull/6493#discussion_r1523571821, changed to:
`for bytes_, length in session.execute(stmt):`

lib/rucio/core/rule.py
erlingstaff: Refactored to:
`return session.execute(stmt).all()  # type: ignore`

lib/rucio/core/rule.py
erlingstaff: Refactored to:
`rules.extend(session.execute(stmt).scalars().all())`

lib/rucio/core/rule.py
erlingstaff: Refactored to:
`for lock in session.execute(stmt).scalars().all():`

lib/rucio/core/rule.py
erlingstaff: They are not, I likely forgot to remove them when I rewrote this to use `true()` previously. Removed them now

lib/rucio/core/rule.py
dchristidis: My reasoning is to avoid having one-time-use intermediate variables unless there is a particular reason (e.g. it helps break down a long chain of operations).

dchristidis (approved):
I would have liked a third full review, but I think it has been more than enough.

Issue #6454, opened 2024-01-12
Type annotate the code


bari12:
### Description

Type annotate methods which have not been annotated yet.
Establish common types. Possibly re-write methods to conform to these common types.

PR #6497, opened 2024-02-15
Add type annotations and refactor to use TokenDict type

rdimaio:
#6454 - Adding various type annotations, and replacing the method:

```
def token_dictionary(token: models.Token):
    return {'token': token.token, 'expires_at': token.expired_at}
```

with a specific datatype `TokenDict` in order to be more specific with the return types.

### DIFF
--- a/lib/rucio/common/types.py
+++ b/lib/rucio/common/types.py
@@ -13,7 +13,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from datetime import datetime
 from typing import Any, Callable, Literal, Optional, TypedDict, Union
+
+from rucio.db.sqla.constants import AccountType, IdentityType
 
 
 class InternalType(object):
@@ -174,3 +177,96 @@ class RuleDict(TypedDict):
     activity: str
     notify: Optional[Literal['Y', 'N', 'C']]
     purge_replicas: bool
+
+
+class RSEAccountCounterDict(TypedDict):
+    account: InternalAccount
+    rse_id: str
+
+
+class RSEAccountUsageDict(TypedDict):
+    rse_id: str
+    rse: str
+    account: InternalAccount
+    used_files: int
+    used_bytes: int
+    quota_bytes: int
+
+
+class RSEGlobalAccountUsageDict(TypedDict):
+    rse_expression: str
+    bytes: int
+    files: int
+    bytes_limit: int
+    bytes_remaining: int
+
+
+class RSELocalAccountUsageDict(TypedDict):
+    rse_id: str
+    rse: str
+    bytes: int
+    files: int
+    bytes_limit: int
+    bytes_remaining: int
+
+
+class RSEResolvedGlobalAccountLimitDict(TypedDict):
+    resolved_rses: str
+    resolved_rse_ids: list[str]
+    limit: float
+
+
+class TokenDict(TypedDict):
+    token: str
+    expires_at: datetime
+
+
+class TokenValidationDict(TypedDict):
+    account: Optional[InternalAccount]
+    identity: Optional[str]
+    lifetime: Optional[datetime]
+    audience: Optional[str]
+    authz_scope: Optional[str]
+
+
+class AccountDict(TypedDict):
+    account: InternalAccount
+    type: AccountType
+    email: str
+
+
+class AccountAttributesDict(TypedDict):
+    key: str
+    value: Union[bool, str]
+
+
+class IdentityDict(TypedDict):
+    type: IdentityType
+    identity: str
+    email: str
+
+
+class UsageDict(TypedDict):
+    bytes: int
+    files: int
+    updated_at: Optional[datetime]
+
+
+class TokenOIDCAutoDict(TypedDict, total=False):
+    webhome: Optional[str]
+    token: Optional[TokenDict]
+
+
+class TokenOIDCNoAutoDict(TypedDict):
+    fetchcode: str
+
+
+class TokenOIDCPollingDict(TypedDict):
+    polling: bool
+
+
+class AccountUsageModelDict(TypedDict):
+    account: InternalAccount
+    rse_id: str
+    files: int
+    bytes: int
--- a/lib/rucio/common/utils.py
+++ b/lib/rucio/common/utils.py
@@ -40,7 +40,7 @@ from enum import Enum
 from functools import partial, wraps
 from io import StringIO
 from itertools import zip_longest
-from typing import TYPE_CHECKING, Callable, Optional
+from typing import Any, Callable, Optional, TYPE_CHECKING
 from urllib.parse import urlparse, urlencode, quote, parse_qsl, urlunparse
 from uuid import uuid4 as uuid
 from xml.etree import ElementTree
@@ -568,7 +568,7 @@ def str_to_date(string):
     return datetime.datetime.strptime(string, DATE_FORMAT) if string else None
 
 
-def val_to_space_sep_str(vallist):
+def val_to_space_sep_str(vallist: Any) -> str:
     """ Converts a list of values into a string of space separated values
 
     :param vallist: the list of values to to convert into string
--- a/lib/rucio/common/utils.py
+++ b/lib/rucio/common/utils.py
@@ -1198,7 +1198,7 @@ def detect_client_location():
             'longitude': longitude}
 
 
-def ssh_sign(private_key, message):
+def ssh_sign(private_key: str, message: str) -> str:
     """
     Sign a string message using the private key.
 
@@ -1206,14 +1206,13 @@ def ssh_sign(private_key, message):
     :param message: The message to sign as a string.
     :return: Base64 encoded signature as a string.
     """
-    if isinstance(message, str):
-        message = message.encode()
+    encoded_message = message.encode()
     if not EXTRA_MODULES['paramiko']:
         raise MissingModuleException('The paramiko module is not installed or faulty.')
     sio_private_key = StringIO(private_key)
     priv_k = RSAKey.from_private_key(sio_private_key)
     sio_private_key.close()
-    signature_stream = priv_k.sign_ssh_data(message)
+    signature_stream = priv_k.sign_ssh_data(encoded_message)
     signature_stream.rewind()
     base64_encoded = base64.b64encode(signature_stream.get_remainder())
     base64_encoded = base64_encoded.decode()
--- a/lib/rucio/core/account.py
+++ b/lib/rucio/core/account.py
@@ -17,7 +17,7 @@ from datetime import datetime
 from enum import Enum
 from re import match
 from traceback import format_exc
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any, cast, Iterator, Optional
 
 from sqlalchemy import select, and_
 from sqlalchemy.exc import IntegrityError
@@ -27,6 +27,7 @@ import rucio.core.account_counter
 import rucio.core.rse
 from rucio.common import exception
 from rucio.common.config import config_get_bool
+from rucio.common.types import InternalAccount, AccountAttributesDict, AccountDict, AccountUsageModelDict, IdentityDict, UsageDict
 from rucio.core.vo import vo_exists
 from rucio.db.sqla import models
 from rucio.db.sqla.constants import AccountStatus, AccountType
@@ -37,7 +38,7 @@ if TYPE_CHECKING:
 
 
 @transactional_session
-def add_account(account, type_, email, *, session: "Session"):
+def add_account(account: InternalAccount, type_: AccountType, email: str, *, session: "Session") -> None:
     """ Add an account with the given account name and type.
 
     :param account: the name of the new account.
@@ -63,7 +64,7 @@ def add_account(account, type_, email, *, session: "Session"):
 
 
 @read_session
-def account_exists(account, *, session: "Session"):
+def account_exists(account: InternalAccount, *, session: "Session") -> bool:
     """ Checks to see if account exists and is active.
 
     :param account: Name of the account.
@@ -81,7 +82,7 @@ def account_exists(account, *, session: "Session"):
 
 
 @read_session
-def get_account(account, *, session: "Session"):
+def get_account(account: InternalAccount, *, session: "Session") -> models.Account:
     """ Returns an account for the given account name.
 
     :param account: the name of the account.
@@ -102,7 +103,7 @@ def get_account(account, *, session: "Session"):
 
 
 @transactional_session
-def del_account(account, *, session: "Session"):
+def del_account(account: InternalAccount, *, session: "Session") -> None:
     """ Disable an account with the given account name.
 
     :param account: the account name.
@@ -115,15 +116,15 @@ def del_account(account, *, session: "Session"):
         models.Account.status == AccountStatus.ACTIVE
     )
     try:
-        account = session.execute(query).scalar_one()
+        account_result = session.execute(query).scalar_one()
     except exc.NoResultFound:
         raise exception.AccountNotFound('Account with ID \'%s\' cannot be found' % account)
 
-    account.update({'status': AccountStatus.DELETED, 'deleted_at': datetime.utcnow()})
+    account_result.update({'status': AccountStatus.DELETED, 'deleted_at': datetime.utcnow()})
 
 
 @transactional_session
-def update_account(account, key, value, *, session: "Session"):
+def update_account(account: InternalAccount, key: str, value: Any, *, session: "Session") -> None:
     """ Update a property of an account.
 
     :param account: Name of the account.
@@ -137,22 +138,22 @@ def update_account(account, key, value, *, session: "Session"):
         models.Account.account == account
     )
     try:
-        account = session.execute(query).scalar_one()
+        query_result = session.execute(query).scalar_one()
     except exc.NoResultFound:
         raise exception.AccountNotFound('Account with ID \'%s\' cannot be found' % account)
     if key == 'status':
         if isinstance(value, str):
             value = AccountStatus[value]
         if value == AccountStatus.SUSPENDED:
-            account.update({'status': value, 'suspended_at': datetime.utcnow()})
+            query_result.update({'status': value, 'suspended_at': datetime.utcnow()})
         elif value == AccountStatus.ACTIVE:
-            account.update({'status': value, 'suspended_at': None})
+            query_result.update({'status': value, 'suspended_at': None})
     else:
-        account.update({key: value})
+        query_result.update({key: value})
 
 
 @stream_session
-def list_accounts(filter_=None, *, session: "Session"):
+def list_accounts(filter_: Optional[dict[str, Any]] = None, *, session: "Session") -> Iterator[AccountDict]:
     """ Returns a list of all account names.
 
     :param filter_: Dictionary of attributes by which the input data should be filtered
@@ -213,7 +214,7 @@ def list_accounts(filter_=None, *, session: "Session"):
 
 
 @read_session
-def list_identities(account, *, session: "Session"):
+def list_identities(account: InternalAccount, *, session: "Session") -> list[IdentityDict]:
     """
     List all identities on an account.
 
@@ -244,11 +245,11 @@ def list_identities(account, *, session: "Session"):
     ).where(
         models.IdentityAccountAssociation.account == account
     )
-    return [row._asdict() for row in session.execute(query)]
+    return [cast(IdentityDict, row._asdict()) for row in session.execute(query)]
 
 
 @read_session
-def list_account_attributes(account, *, session: "Session"):
+def list_account_attributes(account: InternalAccount, *, session: "Session") -> list[AccountAttributesDict]:
     """
     Get all attributes defined for an account.
 
@@ -274,11 +275,11 @@ def list_account_attributes(account, *, session: "Session"):
     ).where(
         models.AccountAttrAssociation.account == account
     )
-    return [row._asdict() for row in session.execute(query)]
+    return [cast(AccountAttributesDict, row._asdict()) for row in session.execute(query)]
 
 
 @read_session
-def has_account_attribute(account, key, *, session: "Session"):
+def has_account_attribute(account: InternalAccount, key: str, *, session: "Session") -> bool:
     """
     Indicates whether the named key is present for the account.
 
@@ -298,7 +299,7 @@ def has_account_attribute(account, key, *, session: "Session"):
 
 
 @transactional_session
-def add_account_attribute(account, key, value, *, session: "Session"):
+def add_account_attribute(account: InternalAccount, key: str, value: Any, *, session: "Session") -> None:
     """
     Add an attribute for the given account name.
 
@@ -334,7 +335,7 @@ def add_account_attribute(account, key, value, *, session: "Session"):
 
 
 @transactional_session
-def del_account_attribute(account, key, *, session: "Session"):
+def del_account_attribute(account: InternalAccount, key: str, *, session: "Session") -> None:
     """
     Add an attribute for the given account name.
 
@@ -355,7 +356,7 @@ def del_account_attribute(account, key, *, session: "Session"):
 
 
 @read_session
-def get_usage(rse_id, account, *, session: "Session"):
+def get_usage(rse_id: str, account: InternalAccount, *, session: "Session") -> UsageDict:
     """
     Returns current values of the specified counter, or raises CounterNotFound if the counter does not exist.
 
@@ -373,13 +374,13 @@ def get_usage(rse_id, account, *, session: "Session"):
         models.AccountUsage.account == account
     )
     try:
-        return session.execute(query).one()._asdict()
+        return cast(UsageDict, session.execute(query).one()._asdict())
     except exc.NoResultFound:
-        return {'bytes': 0, 'files': 0, 'updated_at': None}
+        return UsageDict({'bytes': 0, 'files': 0, 'updated_at': None})
 
 
 @read_session
-def get_all_rse_usages_per_account(account, *, session: "Session"):
+def get_all_rse_usages_per_account(account: InternalAccount, *, session: "Session") -> list[AccountUsageModelDict]:
     """
     Returns current values of the specified counter, or raises CounterNotFound if the counter does not exist.
 
@@ -394,20 +395,20 @@ def get_all_rse_usages_per_account(account, *, session: "Session"):
         models.AccountUsage.account == account
     )
     try:
-        return [result.to_dict() for result in session.execute(query).scalars()]
+        return [cast(AccountUsageModelDict, result.to_dict()) for result in session.execute(query).scalars()]
     except exc.NoResultFound:
         return []
 
 
 @read_session
-def get_usage_history(rse_id, account, *, session: "Session"):
+def get_usage_history(rse_id: str, account: InternalAccount, *, session: "Session") -> list[UsageDict]:
     """
     Returns historical values of the specified counter, or raises CounterNotFound if the counter does not exist.
 
     :param rse_id:           The id of the RSE.
     :param account:          The account name.
     :param session:          The database session in use.
-    :returns:                A dictionary {'bytes', 'files', 'updated_at'}
+    :returns:                A list of dictionaries {'bytes', 'files', 'updated_at'}
     """
     query = select(
         models.AccountUsageHistory.bytes,
@@ -420,7 +421,6 @@ def get_usage_history(rse_id, account, *, session: "Session"):
         models.AccountUsageHistory.updated_at
     )
     try:
-        return [row._asdict() for row in session.execute(query)]
+        return [cast(UsageDict, row._asdict()) for row in session.execute(query)]
     except exc.NoResultFound:
         raise exception.CounterNotFound('No usage can be found for account %s on RSE %s' % (account, rucio.core.rse.get_rse_name(rse_id=rse_id, session=session)))
-    return []
--- a/lib/rucio/core/account_counter.py
+++ b/lib/rucio/core/account_counter.py
@@ -12,12 +12,14 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
 import datetime
-from typing import TYPE_CHECKING
+from typing import cast, TYPE_CHECKING
 
 from sqlalchemy import literal, insert, select
 from sqlalchemy.orm.exc import NoResultFound
 
+from rucio.common.types import InternalAccount, RSEAccountCounterDict
 from rucio.db.sqla import models, filter_thread_work
 from rucio.db.sqla.session import read_session, transactional_session
 
@@ -28,7 +30,7 @@ MAX_COUNTERS = 10
 
 
 @transactional_session
-def add_counter(rse_id, account, *, session: "Session"):
+def add_counter(rse_id: str, account: InternalAccount, *, session: "Session") -> None:
     """
     Creates the specified counter for a rse_id and account.
 
@@ -41,7 +43,7 @@ def add_counter(rse_id, account, *, session: "Session"):
 
 
 @transactional_session
-def increase(rse_id, account, files, bytes_, *, session: "Session"):
+def increase(rse_id: str, account: InternalAccount, files: int, bytes_: int, *, session: "Session") -> None:
     """
     Increments the specified counter by the specified amount.
 
@@ -55,7 +57,7 @@ def increase(rse_id, account, files, bytes_, *, session: "Session"):
 
 
 @transactional_session
-def decrease(rse_id, account, files, bytes_, *, session: "Session"):
+def decrease(rse_id: str, account: InternalAccount, files: int, bytes_: int, *, session: "Session") -> None:
     """
     Decreases the specified counter by the specified amount.
 
@@ -69,7 +71,7 @@ def decrease(rse_id, account, files, bytes_, *, session: "Session"):
 
 
 @transactional_session
-def del_counter(rse_id, account, *, session: "Session"):
+def del_counter(rse_id: str, account: InternalAccount, *, session: "Session") -> None:
     """
     Resets the specified counter and initializes it by the specified amounts.
 
@@ -82,7 +84,7 @@ def del_counter(rse_id, account, *, session: "Session"):
 
 
 @read_session
-def get_updated_account_counters(total_workers, worker_number, *, session: "Session"):
+def get_updated_account_counters(total_workers: int, worker_number: int, *, session: "Session") -> list[RSEAccountCounterDict]:
     """
     Get updated rse_counters.
 
@@ -104,7 +106,7 @@ def get_updated_account_counters(total_workers, worker_number, *, session: "Sess
 
 
 @transactional_session
-def update_account_counter(account, rse_id, *, session: "Session"):
+def update_account_counter(account: InternalAccount, rse_id: str, *, session: "Session") -> None:
     """
     Read the updated_account_counters and update the account_counter.
 
@@ -130,7 +132,7 @@ def update_account_counter(account, rse_id, *, session: "Session"):
 
 
 @transactional_session
-def update_account_counter_history(account, rse_id, *, session: "Session"):
+def update_account_counter_history(account: InternalAccount, rse_id: str, *, session: "Session") -> None:
     """
     Read the AccountUsage and update the AccountUsageHistory.
 
@@ -146,7 +148,7 @@ def update_account_counter_history(account, rse_id, *, session: "Session"):
 
 
 @transactional_session
-def fill_account_counter_history_table(*, session: "Session"):
+def fill_account_counter_history_table(*, session: "Session") -> None:
     """
     Make a snapshot of current counters
 
--- a/lib/rucio/core/account_counter.py
+++ b/lib/rucio/core/account_counter.py
@@ -91,17 +91,16 @@ def get_updated_account_counters(total_workers, worker_number, *, session: "Sess
     :param session:            Database session in use.
     :returns:                  List of rse_ids whose rse_counters need to be updated.
     """
-    query = session.query(models.UpdatedAccountCounter.account, models.UpdatedAccountCounter.rse_id).\
-        distinct(models.UpdatedAccountCounter.account, models.UpdatedAccountCounter.rse_id)
 
-    if session.bind.dialect.name == 'oracle':
-        hash_variable = 'CONCAT(account, rse_id)'''
-    else:
-        hash_variable = 'concat(account, rse_id)'
+    query = select(
+        models.UpdatedAccountCounter.account,
+        models.UpdatedAccountCounter.rse_id,
+    ).distinct(
+    )
 
-    query = filter_thread_work(session=session, query=query, total_threads=total_workers, thread_id=worker_number, hash_variable=hash_variable)
+    query = filter_thread_work(session=session, query=query, total_threads=total_workers, thread_id=worker_number, hash_variable='CONCAT(account, rse_id)')
 
-    return query.all()
+    return [cast(RSEAccountCounterDict, row._asdict()) for row in session.execute(query).all()]
 
 
 @transactional_session
--- a/lib/rucio/core/account_limit.py
+++ b/lib/rucio/core/account_limit.py
@@ -13,12 +13,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Optional, Union
+from uuid import UUID
 
 from sqlalchemy.orm.exc import NoResultFound
-from sqlalchemy.sql import func, select, literal
+from sqlalchemy.sql import func, literal, select
 from sqlalchemy.sql.expression import and_, or_
 
+from rucio.common.types import (InternalAccount, RSEAccountUsageDict,
+                                RSEGlobalAccountUsageDict,
+                                RSELocalAccountUsageDict,
+                                RSEResolvedGlobalAccountLimitDict)
 from rucio.core.account import get_all_rse_usages_per_account
 from rucio.core.rse import get_rse_name
 from rucio.core.rse_expression_parser import parse_expression
@@ -30,7 +35,7 @@ if TYPE_CHECKING:
 
 
 @read_session
-def get_rse_account_usage(rse_id, *, session: "Session"):
+def get_rse_account_usage(rse_id: str, *, session: "Session") -> list[RSEAccountUsageDict]:
     """
     Returns the account limit and usage for all accounts on a RSE.
 
@@ -90,7 +95,7 @@ def get_rse_account_usage(rse_id, *, session: "Session"):
 
 
 @read_session
-def get_global_account_limits(account=None, *, session: "Session"):
+def get_global_account_limits(account: Optional[InternalAccount] = None, *, session: "Session") -> dict[str, RSEResolvedGlobalAccountLimitDict]:
     """
     Returns the global account limits for the account.
 
@@ -121,7 +126,7 @@ def get_global_account_limits(account=None, *, session: "Session"):
 
 
 @read_session
-def get_global_account_limit(account, rse_expression, *, session: "Session"):
+def get_global_account_limit(account: InternalAccount, rse_expression: str, *, session: "Session") -> Union[int, float, None]:
     """
     Returns the global account limit for the account on the rse expression.
 
@@ -141,7 +146,7 @@ def get_global_account_limit(account, rse_expression, *, session: "Session"):
 
 
 @read_session
-def get_local_account_limit(account, rse_id, *, session: "Session"):
+def get_local_account_limit(account: InternalAccount, rse_id: str, *, session: "Session") -> Union[int, float, None]:
     """
     Returns the account limit for the account on the rse.
 
@@ -162,7 +167,7 @@ def get_local_account_limit(account, rse_id, *, session: "Session"):
 
 
 @read_session
-def get_local_account_limits(account, rse_ids=None, *, session: "Session"):
+def get_local_account_limits(account: InternalAccount, rse_ids: Optional[list[str]] = None, *, session: "Session") -> dict[str, int]:
     """
     Returns the account limits for the account on the list of rses.
 
@@ -196,7 +201,7 @@ def get_local_account_limits(account, rse_ids=None, *, session: "Session"):
 
 
 @transactional_session
-def set_local_account_limit(account, rse_id, bytes_, *, session: "Session"):
+def set_local_account_limit(account: InternalAccount, rse_id: str, bytes_: int, *, session: "Session") -> None:
     """
     Returns the limits for the account on the rse.
 
@@ -214,7 +219,7 @@ def set_local_account_limit(account, rse_id, bytes_, *, session: "Session"):
 
 
 @transactional_session
-def set_global_account_limit(account, rse_expression, bytes_, *, session: "Session"):
+def set_global_account_limit(account: InternalAccount, rse_expression: str, bytes_: int, *, session: "Session") -> None:
     """
     Sets the global limit for the account on a RSE expression.
 
@@ -232,7 +237,7 @@ def set_global_account_limit(account, rse_expression, bytes_, *, session: "Sessi
 
 
 @transactional_session
-def delete_local_account_limit(account, rse_id, *, session: "Session"):
+def delete_local_account_limit(account: InternalAccount, rse_id: str, *, session: "Session") -> bool:
     """
     Deletes a local account limit.
 
@@ -250,7 +255,7 @@ def delete_local_account_limit(account, rse_id, *, session: "Session"):
 
 
 @transactional_session
-def delete_global_account_limit(account, rse_expression, *, session: "Session"):
+def delete_global_account_limit(account: InternalAccount, rse_expression: str, *, session: "Session") -> bool:
     """
     Deletes a global account limit.
 
@@ -268,7 +273,7 @@ def delete_global_account_limit(account, rse_expression, *, session: "Session"):
 
 
 @transactional_session
-def get_local_account_usage(account, rse_id=None, *, session: "Session"):
+def get_local_account_usage(account: InternalAccount, rse_id: Optional[UUID] = None, *, session: "Session") -> list[RSELocalAccountUsageDict]:
     """
     Read the account usage and connect it with (if available) the account limits of the account.
 
@@ -311,7 +316,7 @@ def get_local_account_usage(account, rse_id=None, *, session: "Session"):
 
 
 @transactional_session
-def get_global_account_usage(account, rse_expression=None, *, session: "Session"):
+def get_global_account_usage(account: InternalAccount, rse_expression: Optional[str] = None, *, session: "Session") -> list[RSEGlobalAccountUsageDict]:
     """
     Read the account usage and connect it with the global account limits of the account.
 
--- a/lib/rucio/core/authentication.py
+++ b/lib/rucio/core/authentication.py
@@ -33,7 +33,7 @@ from rucio.common.exception import CannotAuthenticate, RucioException
 from rucio.common.types import InternalAccount, TokenDict, TokenValidationDict
 from rucio.common.utils import chunks, generate_uuid, date_to_str
 from rucio.core.account import account_exists
-from rucio.core.oidc import validate_jwt, token_dictionary
+from rucio.core.oidc import validate_jwt
 from rucio.db.sqla import filter_thread_work
 from rucio.db.sqla import models
 from rucio.db.sqla.constants import IdentityType
@@ -90,7 +90,7 @@ else:
 
 
 @transactional_session
-def get_auth_token_user_pass(account, username, password, appid, ip=None, *, session: "Session"):
+def get_auth_token_user_pass(account: InternalAccount, username: str, password: str, appid: str, ip: Optional[str] = None, *, session: "Session") -> Optional[TokenDict]:
     """
     Authenticate a Rucio account temporarily via username and password.
 
@@ -150,7 +150,7 @@ def get_auth_token_user_pass(account, username, password, appid, ip=None, *, ses
 
 
 @transactional_session
-def get_auth_token_x509(account, dn, appid, ip=None, *, session: "Session"):
+def get_auth_token_x509(account: InternalAccount, dn: str, appid: str, ip: Optional[str] = None, *, session: "Session") -> Optional[TokenDict]:
     """
     Authenticate a Rucio account temporarily via an x509 certificate.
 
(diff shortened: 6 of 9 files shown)


lib/rucio/api/authentication.py
```diff
@@ -230,14 +230,13 @@ def get_auth_token_ssh(account, signature, appid, ip=None, vo='def', *, session:
 
 
 @transactional_session
-def get_ssh_challenge_token(account, appid, ip=None, vo='def', *, session: "Session"):
+def get_ssh_challenge_token(account, ip=None, vo='def', *, session: "Session"):
```
dchristidis: This is quite a significant change that is currently invisible (think of a community that has implemented probes using the API layer, as is the official recommendation). Anything related to the removal of `appid` must be split into its own commit.

lib/rucio/common/types.py
```diff
@@ -14,6 +14,7 @@
 # limitations under the License.
 
 from typing import Any, Callable, Optional, TypedDict, Union
+from datetime import datetime
```
dchristidis: Pleas sort alphabetically.

lib/rucio/common/utils.py
```diff
@@ -570,7 +570,7 @@ def str_to_date(string):
     return datetime.datetime.strptime(string, DATE_FORMAT) if string else None
 
 
-def val_to_space_sep_str(vallist):
+def val_to_space_sep_str(vallist: list) -> str:
```
dchristidis: Doesn’t the `vallist: list` hint contradict the implementation (which also does something if it isn’t a list)?

lib/rucio/core/account.py
```diff
@@ -17,7 +17,8 @@
 from enum import Enum
 from re import match
 from traceback import format_exc
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any, Iterator, Optional
+import uuid
```
dchristidis: Importing entire modules should be placed above selective imports with `from`.

lib/rucio/core/account.py
```diff
@@ -37,7 +39,7 @@
 
 
 @transactional_session
-def add_account(account, type_, email, *, session: "Session"):
+def add_account(account: InternalAccount, type_: AccountType, email: str, *, session: "Session"):
```
dchristidis: Missing hint for the return type.

lib/rucio/core/account.py
```diff
@@ -102,7 +104,7 @@ def get_account(account, *, session: "Session"):
 
 
 @transactional_session
-def del_account(account, *, session: "Session"):
+def del_account(account: InternalAccount, *, session: "Session"):
```
dchristidis: Missing hint for the return type.

lib/rucio/core/account.py
```diff
@@ -115,15 +117,14 @@ def del_account(account, *, session: "Session"):
...
     )
     try:
-        account = session.execute(query).scalar_one()
+        account_result = session.execute(query).scalar_one()
+        account_result.update({'status': AccountStatus.DELETED, 'deleted_at': datetime.utcnow()})
     except exc.NoResultFound:
         raise exception.AccountNotFound('Account with ID \'%s\' cannot be found' % account)
 
-    account.update({'status': AccountStatus.DELETED, 'deleted_at': datetime.utcnow()})
-
```
dchristidis: You’ve also moved around statements later as well so that they are part of the try block. Please elaborate.

lib/rucio/core/account.py
```diff
@@ -115,15 +117,14 @@ def del_account(account, *, session: "Session"):
...
 
 @transactional_session
-def update_account(account, key, value, *, session: "Session"):
+def update_account(account: InternalAccount, key: str, value: Any, *, session: "Session"):
```
dchristidis: Missing hint for the return type.

dchristidis: Like my previous comment, the implementation suggests that value may not be a `str`.

lib/rucio/core/account.py
```diff
@@ -137,22 +138,22 @@ def update_account(account, key, value, *, session: "Session"):
...
-            account.update({'status': value, 'suspended_at': datetime.utcnow()})
-        elif value == AccountStatus.ACTIVE:
-            account.update({'status': value, 'suspended_at': None})
-    else:
-        account.update({key: value})
 
 
 @stream_session
-def list_accounts(filter_=None, *, session: "Session"):
+def list_accounts(filter_: Optional[dict] = None, *, session: "Session") -> Iterator[dict]:
```
dchristidis: Maybe `filter_: Optional[dict[str, Any]] = None` is more precise?

lib/rucio/core/account.py
```diff
@@ -213,7 +214,7 @@ def list_accounts(filter_=None, *, session: "Session"):
 
 
 @read_session
-def list_identities(account, *, session: "Session"):
+def list_identities(account: InternalAccount, *, session: "Session"):
```
dchristidis: Missing hint for the return type.

lib/rucio/core/account.py
```diff
@@ -248,7 +249,7 @@ def list_identities(account, *, session: "Session"):
 
 
 @read_session
-def list_account_attributes(account, *, session: "Session"):
+def list_account_attributes(account: InternalAccount, *, session: "Session") -> list[dict]:
```
dchristidis: Same comment: consider `-> list[dict[str, Any]]`

lib/rucio/core/account.py
```diff
@@ -298,7 +299,7 @@ def has_account_attribute(account, key, *, session: "Session"):
 
 
 @transactional_session
-def add_account_attribute(account, key, value, *, session: "Session"):
+def add_account_attribute(account: InternalAccount, key: str, value: Any, *, session: "Session"):
```
dchristidis: Missing hint for the return type.

lib/rucio/core/account.py
```diff
@@ -334,7 +335,7 @@ def add_account_attribute(account, key, value, *, session: "Session"):
 
 
 @transactional_session
-def del_account_attribute(account, key, *, session: "Session"):
+def del_account_attribute(account: InternalAccount, key: str, *, session: "Session"):
```
dchristidis: Missing hint for the return type.

lib/rucio/core/account.py
```diff
@@ -355,7 +356,7 @@ def del_account_attribute(account, key, *, session: "Session"):
 
 
 @read_session
-def get_usage(rse_id, account, *, session: "Session"):
+def get_usage(rse_id: uuid.UUID, account: InternalAccount, *, session: "Session") -> dict:
```
dchristidis: I don’t think this is right. `rse_id` should be `str`. This applies to many other places as well.

lib/rucio/core/account_counter.py
```diff
@@ -12,23 +12,26 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+import uuid
 import datetime
```
dchristidis: Please sort alphabetically.

lib/rucio/core/account_counter.py
```diff
@@ -82,7 +85,7 @@ def del_counter(rse_id, account, *, session: "Session"):
 
 
 @read_session
-def get_updated_account_counters(total_workers, worker_number, *, session: "Session"):
+def get_updated_account_counters(total_workers: int, worker_number: int, *, session: "Session") -> list["Query"]:
```
dchristidis: `-> list["Query"]` doesn’t look right.

lib/rucio/core/authentication.py
```diff
@@ -530,12 +530,8 @@ def validate_auth_token(token: str, *, session: "Session") -> "dict[str, Any]":
     return value
 
 
-def token_dictionary(token: models.Token):
-    return {'token': token.token, 'expires_at': token.expired_at}
-
-
```
dchristidis: Curious, why is removing `token_dictionary()` preferable? I’m not necessarily against it, but seems like a useful function.

lib/rucio/core/authentication.py
```diff
@@ -530,12 +530,8 @@ def validate_auth_token(token: str, *, session: "Session") -> "dict[str, Any]":
...
 @transactional_session
-def __delete_expired_tokens_account(account, *, session: "Session"):
+def __delete_expired_tokens_account(account: InternalAccount, *, session: "Session"):
```
dchristidis: Missing hint for the return type.

lib/rucio/core/oidc.py
```diff
@@ -1000,7 +1000,7 @@ def __change_refresh_state(token: str, refresh: bool = False, *, session: "Sessi
 
 
 @transactional_session
-def refresh_cli_auth_token(token_string: str, account: str, *, session: "Session"):
+def refresh_cli_auth_token(token_string: str, account: str, *, session: "Session") -> Optional[tuple]:
```
dchristidis: Why not be more precise with `-> Optional[tuple[str, int]]`?

dchristidis (changes requested):
Looks promising! It would be beneficial to prefix the commit subjects with `Auth: `.

lib/rucio/common/utils.py
rdimaio: Seems like `list` is the intended type for `vallist`, but instead of failling when a list is not passed, the implementation simply prints a string representation of whatever the object is. 

I think it would be better to instead raise an exception, as it would make it more explicit that `val_to_space_sep_str` is not actually returning a space-separated string. 

I don't want to make a functional change here, so I will just modify vallist to be `Any` for now, but we might want to consider modifying this implementation at some point.

lib/rucio/core/account.py
rdimaio: The original statement `account = session.execute(query).scalar_one()` would lead to a type error, because the type of `account` (`InternalAccount`) was not compatible with the type returned by `session.execute(query).scalar_one()` (I don't remember what this type is at the moment). 

To avoid this, I changed the statemente so that the output of `session.execute(query).scalar_one()` would be stored in a new variable `account_result`. 

As this variable is defined within the scope of the `try` block, I had to move the `update` statement under the `try` block as well.

lib/rucio/core/account_counter.py
rdimaio: This was the reference I used for that: https://stackoverflow.com/questions/55361186/type-hinting-sqlalchemy-query-result

> [Query.all()](https://github.com/sqlalchemy/sqlalchemy/blob/201c4a60e4b8af56d9c02a3675d1443ba4171c89/lib/sqlalchemy/orm/query.py#L3155-L3161) calls list() on the Query object itself

I'm not 100% sure about it either, but I'm not sure about what other type this would be.

lib/rucio/core/account.py
rdimaio: I was looking at this statement within the function:
```
models.AccountUsage.rse_id == rse_id
```

Where the model is:

```
class AccountUsage(BASE, ModelBase):
    """Represents account usage"""
    __tablename__ = 'account_usage'
    account: Mapped[InternalAccount] = mapped_column(InternalAccountString(get_schema_value('ACCOUNT_LENGTH')))
    rse_id: Mapped[uuid.UUID] = mapped_column(GUID())
...
```

So I assumed `rse_id` would need to match the type indicated in the model

lib/rucio/core/authentication.py
rdimaio: I can keep both the function and `TokenDict` (and just make `token_dictionary` return a `TokenDict` dict instead of a generic dict) but it seemed a bit unnecessary to make a function just for that. I can add it again if you prefer.

lib/rucio/core/account.py
rdimaio: `value` is set to `Any` here, not `str`. Could you clarify what you mean?

lib/rucio/api/authentication.py
dchristidis: Please review this again. It looks to me like you reverted the change in `core/authentication.py`, but not elsewhere.

(On a side note, I know you’re trying to be helpful, but I find it more beneficial to resolve conversations myself. This way, when I resolve it, I know for sure that it’s done and have no need to check it again. No need to reply or anything to demonstrate that you’ve done your part. Thanks!)

lib/rucio/core/account.py
dchristidis: You placed `uuid` correctly, but also introduced an unnecessary empty newline.

lib/rucio/core/account.py
dchristidis: I’m not sure I follow. The `try` block does not create a new scope.

```python
>>> print(foo)
NameError: name 'foo' is not defined
>>> try:
...     foo = 'foo'
... except Exception:
...     pass
... 
>>> print(foo)
foo
>>> 
```

lib/rucio/core/account.py
dchristidis: You’re right. I’m not sure what I was thinking; probably a mistake on my part.

lib/rucio/core/account.py
dchristidis: Good assumption, but it’s complicated. You can see that SQLAlchemy will always return `Optional[str]` when querying the database:

https://github.com/rucio/rucio/blob/cc0067cd2e5ae1c2c6c69cd8adc401db1bdcf015/lib/rucio/db/sqla/types.py#L70-L78

So in the API layer where the RSE name is converted to an RSE ID:

https://github.com/rucio/rucio/blob/cc0067cd2e5ae1c2c6c69cd8adc401db1bdcf015/lib/rucio/api/rse.py#L333-L334

It is most definitely a `str`.

lib/rucio/core/account_counter.py
dchristidis: But looking at the query:

https://github.com/rucio/rucio/blob/cc0067cd2e5ae1c2c6c69cd8adc401db1bdcf015/lib/rucio/core/account_counter.py#L94

I would have guessed something like `list[tuple[InternalAccount, str]]`. Did you try it and saw errors?

lib/rucio/core/authentication.py
dchristidis: It’s true that with hints you can be confident that the structure of the returned value will be correct, *if you’re testing for type checking*. But what about if someone accidentally does something like:

```python
return {'token': token.token, 'expires_at': datetime.utcnow()}
```

Far-fetched, I know, but hopefully it demonstrates that this function still has some value and is effectively more strict than just type checking.

lib/rucio/core/account_counter.py
rdimaio: I have not tried `list[tuple[InternalAccount, str]]` yet - I'll try it now to see if it works

lib/rucio/core/account.py
rdimaio: You're right - fixed

lib/rucio/core/account.py
rdimaio: I changed all of them to `str`, it only caused this error:

```
Found 2 new problems in /home/runner/work/rucio/rucio/lib/rucio/core/account_limit.py
  - 1 errors with message """Expression of type "Unknown | UUID" cannot be assigned to declared type "str | None"""".
    Candidates: line 293
  - 1 errors with message """Argument of type "str | None" cannot be assigned to parameter "__key" of type "UUID" in function "get"
                               Type "str | None" cannot be assigned to type "UUID"
                                 "str" is incompatible with "UUID"""".
    Candidates: line 294
```

under `get_local_account_usage`. I've now annotated that `rse_id` to be `uuid.UUID`, and left all the others to be `str`.

lib/rucio/core/authentication.py
dchristidis: Are you purposefully changing the location of `token_dictionary()` in the file?

lib/rucio/api/authentication.py
dchristidis: Still seeing changes related to this in `lib/rucio/web/rest/flaskapi/v1/auth.py` and `
tests/test_authentication.py`.

lib/rucio/core/account.py
dchristidis: I have the impression that it would be more correct to change the model to have `rse_id: Mapped[str]`. The `GUID` class literally has `impl = CHAR`. I’m consulting the SQLAlchemy documentation, in hope of finding a definitive answer. Could you also investigate this path?

lib/rucio/core/account_counter.py
dchristidis: We need to have a closer look at this. `.all()` returns instances of `sqlalchemy.engine.row.Row`. I’m thinking it would be preferable to:

 1. Convert the query to SQLAlchemy 2.0 syntax (unrelated to hints, but a good opportunity)
 2. Return dictionaries with `return [row._asdict() for row in query.all()]`
 3. Adapt `lib/rucio/daemons/abacus/account.py`

My motivation for this proposal is that the use of indices feels poor:

https://github.com/rucio/rucio/blob/cc0067cd2e5ae1c2c6c69cd8adc401db1bdcf015/lib/rucio/daemons/abacus/account.py#L72-L73

What do you think?

And on a side note, do you understand the usefulness of this? I think it’s identical and can go in `filter_thread_work()` directly.

https://github.com/rucio/rucio/blob/cc0067cd2e5ae1c2c6c69cd8adc401db1bdcf015/lib/rucio/core/account_counter.py#L97-L100

It was introduced in f3a073aa9da181070d6b41572838d28727528dd6.

lib/rucio/core/account.py
rdimaio: > I have the impression that it would be more correct to change the model to have `rse_id: Mapped[str]`. The `GUID` class literally has `impl = CHAR`. I’m consulting the SQLAlchemy documentation, in hope of finding a definitive answer. Could you also investigate this path?

I'm not sure about making this change - looking at the other models, all `id` fields seem to use `uuid.UUID`. For consistency, I think it would be best to keep this one as `uuid.UUID` as well.

lib/rucio/core/authentication.py
rdimaio: Fixed in latest revision

dchristidis: And what about `lib/rucio/core/oidc.py`?

lib/rucio/core/account.py
dchristidis: If `Mapped[str]` is correct for `mapped_column(GUID())`, then, by definition, all instances of `Mapped[uuid.UUID]` are incorrect and will have to be replaced.

lib/rucio/core/account_counter.py
rdimaio: I've done the first three changes suggested in the latest revision, but I think they're outside the scope of this PR, and I think the focus should be kept on typing (and make separate PRs for these behavioural changes when needed, which I think might happen fairly frequently with these typing-focused PRs) 

On `filter_thread_work`: I agree that it could be refactored like that, but it seems like that method is used in quite a few other places and I don't think refactoring would fit the scope of this PR - best to open an issue for it and we can work on it separately.

lib/rucio/core/account_counter.py
```diff
@@ -91,21 +93,24 @@ def get_updated_account_counters(total_workers, worker_number, *, session: "Sess
...
+             .distinct()
+             )
 
     if session.bind.dialect.name == 'oracle':
         hash_variable = 'CONCAT(account, rse_id)'''
     else:
         hash_variable = 'concat(account, rse_id)'
 
     query = filter_thread_work(session=session, query=query, total_threads=total_workers, thread_id=worker_number, hash_variable=hash_variable)
+    query_result = session.execute(query)
```
rdimaio: not 100% sure here if I had to filter `query` or `query_result`

lib/rucio/core/account_counter.py
dchristidis: Yes and no.

The SQLAlchemy 2.0 syntax, I admit, was over the top. There are five trivial queries in that file; they can all be worked on at the same time in a PR which aims to do it from the start. In retrospect, I should not have asked you to do it. I’ll let you choose whether to keep it or not. Note, however, that you’ll have to follow the coding style we’ve come to adopt for queries:

```python
query = select(
    models.UpdatedAccountCounter.account,
    models.UpdatedAccountCounter.rse_id,
).distinct(
)
```

Now, the change in the return type of `get_updated_account_counters()` *is* very much related to this PR. I’ve asked you to do it because I believe `list[tuple[InternalAccount, str]]` is incorrect. So we had three options:

 1. Not add type hints.
 2. Add incorrect type hints.
 3. Adjust and add correct type hints.

I chose option (3). However, you cannot hide it under `Auth: Migrate to SQLAlchemy 2.0 syntax; #6057`. You have to split it into its own commit.

Now, I did not understand your comment about `filter_thread_work()`. That function is completely fine as it is. What I am asking you to do is to:

 1. Remove those four lines.
 2. Replace `hash_variable=hash_variable` with `hash_variable='CONCAT(account, rse_id)'`.

~~In a separate commit, of course. Yes, it’s not related to type hints, but unlike the SQLAlchemy 2.0 syntax migration, there will be no dedicated issue or pull request to fix it (because it doesn’t happen anywhere else).~~

PS: Actually, we just have to not forget to apply this correction together with the 2.0 syntax change. So it ties together with the choice I mentioned at the beginning of this comment. I would have asked you to remove it, except for the fact that you’ve already done most of it.

lib/rucio/core/account_counter.py
dchristidis: `filter_thread_work()` takes a query and returns another one. Think of it as building on the existing query, like:

```python
query = query.where(models.UpdatedAccountCounter.rse_id == …)
```

I would like to question the introduction of `query_result` (which can be somewhat misleading because it does not contain any result. Who you oppose the following?

```python
return [row._asdict() for row in session.execute(query).all()]
```

lib/rucio/core/account_counter.py
rdimaio: I had misunderstood regarding `filter_thread_work()`, I thought you meant that this logic:

https://github.com/rucio/rucio/blob/cc0067cd2e5ae1c2c6c69cd8adc401db1bdcf015/lib/rucio/core/account_counter.py#L97-L100

could have gone directly within the `filter_thread_work()` function, rather than being done outside of it; now I understood that you meant that the `CONCAT` command is case insensitive and thus the `if/else` block is not needed. I'll make the change you proposed now

rdimaio: Regarding the query syntax - I opened this documentation issue https://github.com/rucio/documentation/issues/287 to consider adding guidelines like these to the documentation

lib/rucio/core/account.py
rdimaio: Addressed in https://github.com/rucio/rucio/pull/6497/commits/c20301e65f7f1a8c0c886beebdb4f17d01958d46

rdimaio: Update - this causes `test_lifetime_creation_core` to fail:

```
E   psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "CONFIGS_PK"
E   DETAIL:  Key (section, opt)=(lifetime_model, cutoff_date) already exists.
```

rdimaio: seems like it was a one-off issue - it passed when rerunning the test https://github.com/rucio/rucio/actions/runs/8005839275/job/21866803841?pr=6497#step:4:917

lib/rucio/core/account.py
```diff
@@ -137,22 +138,22 @@ def update_account(account, key, value, *, session: "Session"):
...
         raise exception.AccountNotFound('Account with ID \'%s\' cannot be found' % account)
-    if key == 'status':
-        if isinstance(value, str):
-            value = AccountStatus[value]
-        if value == AccountStatus.SUSPENDED:
-            account.update({'status': value, 'suspended_at': datetime.utcnow()})
-        elif value == AccountStatus.ACTIVE:
-            account.update({'status': value, 'suspended_at': None})
-    else:
-        account.update({key: value})
```
dchristidis: Similar to another comment, I do not understand the need for this code rearrangement.

lib/rucio/core/account.py
```diff
@@ -137,22 +138,22 @@ def update_account(account, key, value, *, session: "Session"):
...
 
 
 @stream_session
-def list_accounts(filter_=None, *, session: "Session"):
+def list_accounts(filter_: Optional[dict[str, Any]] = None, *, session: "Session") -> Iterator[dict]:
```
dchristidis: Do we want to prefer `Iterator[dict[str, Any]]`? We should aim to eventually replace all `dicts` with `TypedDict`s or the model themselves, but until then, is being a tiny bit more specific beneficial? This comment applies to other lines (and other files) too.

lib/rucio/core/account_counter.py
```diff
@@ -28,7 +30,7 @@
 
 
 @transactional_session
-def add_counter(rse_id, account, *, session: "Session"):
+def add_counter(rse_id: str, account: InternalAccount, *, session: "Session"):
```
dchristidis: Missing `-> None`. Same below. Same for the very last function of this file.

lib/rucio/core/account_limit.py
```diff
@@ -196,7 +197,7 @@ def get_local_account_limits(account, rse_ids=None, *, session: "Session"):
 
 
 @transactional_session
-def set_local_account_limit(account, rse_id, bytes_, *, session: "Session"):
+def set_local_account_limit(account: InternalAccount, rse_id: str, bytes_: int, *, session: "Session"):
```
dchristidis: Missing `-> None`. Also below.

lib/rucio/core/authentication.py
```diff
@@ -40,7 +41,7 @@
 
 if TYPE_CHECKING:
     from sqlalchemy.orm import Session
-    from typing import Any, Union
+    from typing import Any
```
dchristidis: We no longer recommending `Any` being imported conditionally. I would recommend moving it up with the others, then changing all uses of it (the hints to not be quoted in strings).

lib/rucio/core/authentication.py
```diff
@@ -214,7 +215,7 @@ def get_auth_token_gss(account, gsstoken, appid, ip=None, *, session: "Session")
 
 
 @transactional_session
-def get_auth_token_ssh(account, signature, appid, ip=None, *, session: "Session"):
+def get_auth_token_ssh(account: InternalAccount, signature: Union[str, bytes], appid: str, ip: Optional[str] = None, *, session: "Session") -> Optional[TokenDict]:
```
dchristidis: I think `Union[str, bytes]` may warrant one additional look. As far as I can tell, the current calls of this functions use `bytes`. So why keep supporting `str`?

lib/rucio/core/oidc.py
```diff
@@ -1444,5 +1444,5 @@ def oidc_identity_string(sub: str, iss: str):
     return 'SUB=' + str(sub) + ', ISS=' + str(iss)
 
 
-def token_dictionary(token: models.Token):
+def token_dictionary(token: models.Token) -> TokenDict:
```
dchristidis: Is there something preventing us from reusing the function in `authentication.py`?

dchristidis (changes requested):
I will also like to see some rework on how the changes are split into commits, but it might be easier to talk about it in person.

lib/rucio/core/oidc.py
```diff
@@ -1405,7 +1404,7 @@ def validate_jwt(json_web_token: str, *, session: "Session") -> dict[str, Any]:
             clprocess = subprocess.Popen(['curl', '-s', '-L', '-u', '%s:%s'
                                           % (oidc_client.client_id, oidc_client.client_secret),
                                           '-d', 'token=%s' % (json_web_token),
-                                          oidc_client.introspection_endpoint],
+                                          oidc_client.introspection_endpoint],  # type: ignore
```
rdimaio, line 1408: note: this one is pending discussion on Mattermost to decide whether we can delete this whole if block or not

rdimaio: discussed - best to ignore / avoid changing anything OIDC-related

lib/rucio/common/types.py
```diff
@@ -13,8 +13,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from uuid import UUID
```
dchristidis: Why are you reintroducing `UUID`? I thought we agreed that the correct type is `str`?

lib/rucio/common/utils.py
```diff
@@ -1200,7 +1200,7 @@ def detect_client_location():
             'longitude': longitude}
 
 
-def ssh_sign(private_key, message):
+def ssh_sign(private_key: str, message: Union[str, bytes]) -> str:
```
dchristidis: Do you see any current uses of this function where the `message` argument is `bytes`?

lib/rucio/core/account.py
```diff
@@ -27,6 +27,7 @@
 import rucio.core.rse
 from rucio.common import exception
 from rucio.common.config import config_get_bool
+from rucio.common.types import InternalAccount, AccountAttributesDict, AccountDict, AccountUsageModelDict, IdentityDict, UsageDict
```
dchristidis, line 30: Please use `isort` if you don’t want to do the work manually. :smile:

tests/test_authentication.py
```diff
@@ -146,7 +146,7 @@ def test_get_auth_token_ssh_fail(self, vo):
         except Duplicate:
             pass  # might already exist, can skip
 
-        signature = ssh_sign(PRIVATE_KEY, 'sign_something_else')
+        signature = base64.b64decode(ssh_sign(PRIVATE_KEY, 'sign_something_else'))
 
         result = get_auth_token_ssh(account='root', signature=signature, appid='test', ip='127.0.0.1', vo=vo)
```
dchristidis, line 148: I’m not too fond of the situation we ended up in. `ssh_sign()` converts `bytes` to `str`, which you have to convert back to `bytes` to use `get_auth_token_ssh()`.

https://github.com/rucio/rucio/blob/e85183fe6b77841ca01e2d68a1426526bef81b1b/lib/rucio/web/rest/flaskapi/v1/auth.py#L1201-L1229

I wonder if it’s worth propagating the the payload as it is, as an `str`, down to the core layer. There, inside `get_auth_token_ssh()`, we put the decoding operation and raise let it raise a `TypeError` if it fails. The code in the REST layer can then be simplified.

But probably this is going too far, so feel free to ignore.

dchristidis (changes requested):

rdimaio:
@dchristidis I've addressed all your most recent comments and rebased the PR into separate commits in a way that makes sense. 

I've also had to rebase some stuff from master, so "Compare" shows some of those changes as well, which might make it hard to review the diff.

lib/rucio/common/types.py
rdimaio: Removed it from the typeddict. Will only keep it where the tests fails, i.e.:

```
Found 2 new problems in /home/runner/work/rucio/rucio/lib/rucio/core/account_limit.py
  - 1 errors with message """Expression of type "Unknown | UUID" cannot be assigned to declared type "str | None"""".
    Candidates: line 293
  - 1 errors with message """Argument of type "str | None" cannot be assigned to parameter "__key" of type "UUID" in function "get"
                               Type "str | None" cannot be assigned to type "UUID"
                                 "str" is incompatible with "UUID"""".
    Candidates: line 294
```

lib/rucio/common/utils.py
```diff
@@ -40,7 +40,7 @@
 from functools import partial, wraps
 from io import StringIO
 from itertools import zip_longest
-from typing import TYPE_CHECKING, Callable, Optional
+from typing import Any, TYPE_CHECKING, Callable, Optional
```
dchristidis: `isort` is your friend!

lib/rucio/core/account.py
```diff
@@ -244,11 +245,11 @@ def list_identities(account, *, session: "Session"):
     ).where(
         models.IdentityAccountAssociation.account == account
     )
-    return [row._asdict() for row in session.execute(query)]
+    return [cast(IdentityDict, row._asdict()) for row in session.execute(query)]
```
dchristidis, line 248: Not so much a comment for you, but more so for me: understand the necessity to use `cast()`.

lib/rucio/core/account.py
```diff
@@ -373,13 +374,13 @@ def get_usage(rse_id, account, *, session: "Session"):
         models.AccountUsage.account == account
     )
     try:
-        return session.execute(query).one()._asdict()
+        return cast(UsageDict, session.execute(query).one()._asdict())
     except exc.NoResultFound:
-        return {'bytes': 0, 'files': 0, 'updated_at': None}
+        return UsageDict({'bytes': 0, 'files': 0, 'updated_at': None})
```
dchristidis, line 379: Similarly here.

lib/rucio/core/account_counter.py
```diff
@@ -12,12 +12,14 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
 import datetime
-from typing import TYPE_CHECKING
+from typing import cast, TYPE_CHECKING
```
dchristidis, line 17: Why do you hate sorting? Why do you hate order? :laughing:

lib/rucio/core/account_limit.py
```diff
@@ -268,7 +273,7 @@ def delete_global_account_limit(account, rse_expression, *, session: "Session"):
 
 
 @transactional_session
-def get_local_account_usage(account, rse_id=None, *, session: "Session"):
+def get_local_account_usage(account: InternalAccount, rse_id: Optional[UUID] = None, *, session: "Session") -> list[RSELocalAccountUsageDict]:
```
dchristidis, line 276: I still think that `UUID` is wrong and I would like to get to the bottom of this.

lib/rucio/core/authentication.py
```diff
@@ -30,17 +30,17 @@
 from rucio.common.cache import make_region_memcached
 from rucio.common.config import config_get_bool
 from rucio.common.exception import CannotAuthenticate, RucioException
+from rucio.common.types import InternalAccount, TokenDict, TokenValidationDict
 from rucio.common.utils import chunks, generate_uuid, date_to_str
 from rucio.core.account import account_exists
-from rucio.core.oidc import validate_jwt
+from rucio.core.oidc import validate_jwt, token_dictionary
```
dchristidis: Forgive me, it should be the other way around: the `token_dictionary()` in `rucio.core.authentication` should be kept and the one in `rucio.core.oidc` replaced.

dchristidis (changes requested):

lib/rucio/core/account_counter.py
rdimaio: What's the issue here?

lib/rucio/core/account.py
rdimaio: This is what occurs when I don't use `cast`:

```
Found 5 new problems in /home/runner/work/rucio/rucio/lib/rucio/core/account.py
  - 1 errors with message """Expression of type "list[Dict[str, Any]]" cannot be assigned to return type "list[IdentityDict]"""".
    Candidates: line 248
  - 1 errors with message """Expression of type "list[Dict[str, Any]]" cannot be assigned to return type "list[AccountAttributesDict]"""".
    Candidates: line 278
  - 1 errors with message """Expression of type "Dict[str, Any]" cannot be assigned to return type "UsageDict"
                               "Dict[str, Any]" is incompatible with "UsageDict"""".
    Candidates: line 377
  - 1 errors with message """Expression of type "list[dict[str, Any]]" cannot be assigned to return type "list[AccountUsageModelDict]"""".
    Candidates: line 398
  - 1 errors with message """Expression of type "list[Dict[str, Any]]" cannot be assigned to return type "list[UsageDict]"""".
    Candidates: line 424
```

See here for reference: https://github.com/rdimaio/rucio/actions/runs/8206991793/job/22447307300#step:9:10

lib/rucio/core/account.py
rdimaio: Replied here: https://github.com/rucio/rucio/pull/6497#discussion_r1518058114

lib/rucio/core/account.py
dchristidis: I think the `Dict[str, Any]` hint must be coming from [SQLAlchemy itself](https://docs.sqlalchemy.org/en/20/core/connections.html#sqlalchemy.engine.Row._asdict). But by using `cast()` we effectively lose type checking:

```console
$ cat --number test.py
     1	#!/usr/bin/env python3
     2	
     3	from typing import TypedDict, cast
     4	
     5	class Struct(TypedDict):
     6	    foo: str
     7	    bar: int
     8	
     9	def test1() -> Struct:
    10	    return {'foo': 'foo', 'bar': 0}
    11	
    12	def test2() -> Struct:
    13	    return {'baz': None}
    14	
    15	def test3() -> Struct:
    16	    return cast(Struct, {'baz': None})
$ mypy test.py 
test.py:13: error: Missing keys ("foo", "bar") for TypedDict "Struct"  [typeddict-item]
test.py:13: error: Extra key "baz" for TypedDict "Struct"  [typeddict-unknown-key]
Found 2 errors in 1 file (checked 1 source file)
```

Does that mean that we need to avoid the use of `_asdict()` and `to_dict()` altogether? How annoying.

lib/rucio/core/account_counter.py
dchristidis:
```console
$ isort --diff sort.py 
--- /var/home/dchristidis/sort.py:before	2024-03-08 20:42:16.143365
+++ /var/home/dchristidis/sort.py:after	2024-03-08 20:42:28.548160
@@ -1 +1 @@
-from typing import cast, TYPE_CHECKING
+from typing import TYPE_CHECKING, cast
```

lib/rucio/core/authentication.py
rdimaio: This causes a circular import:

```
ImportError: cannot import name 'validate_jwt' from partially initialized module 'rucio.core.oidc' (most likely due to a circular import) (/opt/venv/lib64/python3.9/site-packages/rucio/core/oidc.py)
```

execution for reference: https://github.com/rucio/rucio/actions/runs/8207133900/job/22447782460?pr=6497

lib/rucio/core/account.py
rdimaio: Instead of using `cast`, this seems to work fine:
```
UsageDict(**session.execute(query).one()._asdict())
```

see test commit for reference: https://github.com/rdimaio/rucio/commit/cff7fa20131248acc2c4a06538c20b024d3c0b98

And pyright report: https://github.com/rdimaio/rucio/actions/runs/8233368875/job/22512678528

dchristidis: I’m afraid that I failed to verify that it will raise an error in case of a mismatch:

```console
$ cat --number test.py 
     1	#!/usr/bin/env python3
     2	
     3	from typing import TypedDict
     4	
     5	from sqlalchemy import select
     6	from sqlalchemy.orm import Session
     7	
     8	from rucio.db.sqla.models import Identity
     9	from rucio.db.sqla.session import read_session
    10	
    11	
    12	# This is wrong.
    13	class IdentityDict(TypedDict):
    14	    foo: str
    15	    bar: int
    16	
    17	
    18	@read_session
    19	def list_identities(*, session: Session) -> list[IdentityDict]:
    20	    query = select(
    21	        Identity.identity
    22	    )
    23	    return [IdentityDict(**result._asdict()) for result in session.execute(query)]
    24	
    25	l = list_identities()
$ pyright test.py 
0 errors, 0 warnings, 0 informations
```

rdimaio: This is odd, I found some sources online that seem to indicate that it should work:
- https://stackoverflow.com/questions/75254643/why-are-typeddict-types-without-field-and-with-notrequired-field-incompatible
- https://stackoverflow.com/questions/74495944/no-type-hint-with-return-values-for-typeddict-in-pycharm

And using extra keys in a TypedDict object construction (in the example, `**result._asdict()` are all extra keys to `IdentityDict`) should result in an error, according to PEP-589: https://stackoverflow.com/questions/71814658/python-typing-does-typeddict-allow-additional-extra-keys - the answer also mentions that pyright implements this.

I tried using a `dataclass` instead of a `TypedDict`, and I also don't get any errors raised from the incompatibility:

```
❯ cat --number test.py  
     1  #!/usr/bin/env python3
     2  from sqlalchemy import select
     3  from sqlalchemy.orm import Session
     4
     5  from rucio.db.sqla.models import Identity
     6  from rucio.db.sqla.session import read_session
     7  from dataclasses import dataclass
     8
     9  @dataclass
    10  class IdentityStruct:
    11      first: str
    12      second: int
    13
    14  @read_session
    15  def list_identities(*, session: Session) -> list[IdentityStruct]:
    16      query = select(
    17          Identity.identity
    18      )
    19      return [IdentityStruct(**result._asdict()) for result in session.execute(query)]
    20
    21  l = list_identities()                                                                                                                      


❯ python3 -m pyright test.py 
0 errors, 0 warnings, 0 informations 
```

rdimaio: Some tests without the `sqla` components, just to check.

# 1. Incompatible TypedDict, constructor

```
#!/usr/bin/env python3
from typing import Any, TypedDict

class IncompatibleDict(TypedDict):
    foo: str
    bar: int

def print_dict(in_dict: dict[str, Any]) -> IncompatibleDict:
    d = IncompatibleDict(**in_dict)
    return(d)

test_dict = {
    "name": "test",
    "version": 3
}

l = print_dict(test_dict)
```

## Pyright doesn't complain
```
❯ python3 -m pyright test.py                                                                                                                      ─╯
0 errors, 0 warnings, 0 informations 
```

## Mypy complains that the dict type is unsupported for ** expansion
```
❯ python3 -m mypy test.py                                                                                                                         ─╯
test.py:9: error: Unsupported type "dict[str, Any]" for ** expansion in TypedDict  [typeddict-item]
Found 1 error in 1 file (checked 1 source file)
```

However, if I change the type annotation of `in_dict` from `dict[str, Any]` to `dict`, it behaves more as expected:
```
❯ python3 -m mypy test.py                                                                                                                         ─╯
test.py:9: error: Missing keys ("foo", "bar") for TypedDict "IncompatibleDict"  [typeddict-item]
Found 1 error in 1 file (checked 1 source file)
```

# 2. Incompatible TypedDict, cast
```
#!/usr/bin/env python3
from typing import Any, cast, TypedDict

class IncompatibleDict(TypedDict):
    foo: int
    bar: int

def print_dict(in_dict: dict[str, Any]) -> IncompatibleDict:
    d = cast(IncompatibleDict, in_dict)
    return(d)

test_dict = {
    "name": "test",
    "version": 3
}

l = print_dict(test_dict)
```

## Pyright doesn't complain
```
❯ python3 -m pyright test.py                                                                                                                      ─╯
0 errors, 0 warnings, 0 informations 
```

## Mypy doesn't complain
```
❯ python3 -m mypy test.py                                                                                                                         ─╯
Success: no issues found in 1 source file
```

# 3. Incompatible dataclass
```
#!/usr/bin/env python3
from typing import Any, cast
from dataclasses import dataclass

@dataclass
class IncompatibleClass:
    foo: int
    bar: int

def print_class(in_dict: dict[str, Any]) -> IncompatibleClass:
    d = cast(IncompatibleClass, in_dict)
    return(d)

test_dict = {
    "name": "test",
    "version": 3
}

l = print_class(test_dict)
```

## Pyright doesn't complain
```
❯ python3 -m pyright test.py                                                                                                                      ─╯
0 errors, 0 warnings, 0 informations 
```

## Mypy doesn't complain
```
❯ python3 -m mypy test.py                                                                                                                         ─╯
Success: no issues found in 1 source file
```

# Versions
```
❯ python3 -m pyright --version                                                                                                                    ─╯
pyright 1.1.353
```

```
❯ python3 -m mypy --version                                                                                                                       ─╯
mypy 1.9.0 (compiled: yes)
```

rdimaio: # 4. Incompatible model, runtime Pydantic test, works as expected

```
#!/usr/bin/env python3
from typing import Any, cast
from pydantic import BaseModel, ConfigDict

class IncompatibleModel(BaseModel):
    foo: int
    bar: int
    model_config = ConfigDict(extra='forbid')

def print_class(in_dict: dict[str, Any]) -> IncompatibleModel:
    return(IncompatibleModel(**in_dict))

test_dict = {
    "name": "test",
    "version": 3
}

l = print_class(test_dict)
```

```
❯ python3 test.py                                                                                                                                 ─╯
  File "/home/rdm/tech/rucio/rucio/lib/rucio/core/test.py", line 18, in <module>
    l = print_class(test_dict)
  File "/home/rdm/tech/rucio/rucio/lib/rucio/core/test.py", line 11, in print_class
    return(IncompatibleModel(**in_dict))
  File "/home/rdm/.local/lib/python3.9/site-packages/pydantic/main.py", line 171, in __init__
    self.__pydantic_validator__.validate_python(data, self_instance=self)
pydantic_core._pydantic_core.ValidationError: 4 validation errors for IncompatibleModel
foo
  Field required [type=missing, input_value={'name': 'test', 'version': 3}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.6/v/missing
bar
  Field required [type=missing, input_value={'name': 'test', 'version': 3}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.6/v/missing
name
  Extra inputs are not permitted [type=extra_forbidden, input_value='test', input_type=str]
    For further information visit https://errors.pydantic.dev/2.6/v/extra_forbidden
version
  Extra inputs are not permitted [type=extra_forbidden, input_value=3, input_type=int]
    For further information visit https://errors.pydantic.dev/2.6/v/extra_forbidden
```

# 5. Compatible model, runtime Pydantic test, works as expected:
```
#!/usr/bin/env python3
from typing import Any, cast
from pydantic import BaseModel, ConfigDict

class CompatibleModel(BaseModel):
    name: str
    version: int
    model_config = ConfigDict(extra='forbid')

def print_class(in_dict: dict[str, Any]) -> CompatibleModel:
    return(CompatibleModel(**in_dict))

test_dict = {
    "name": "test",
    "version": 3
}

l = print_class(test_dict)
print(l)
```

```
❯ python3 test.py                                                                                                                                 ─╯
name='test' version=3
```

rdimaio: Going back to static type checking - for test 4 (Incompatible model, runtime Pydantic test, works as expected), mypy and pyright do not detect issues by default:

```
❯ python3 -m mypy test.py                                                                                                                         ─╯
Success: no issues found in 1 source file
```

```
❯ python3 -m pyright test.py                                                                                                                      ─╯
0 errors, 0 warnings, 0 informations 
```

rdimaio: # 6. Incompatible TypedDict, passing parameters directly, no extra parameters
```
#!/usr/bin/env python3
from typing import Any, TypedDict

class IncompatibleDict(TypedDict):
    foo: str
    bar: int

def print_dict() -> IncompatibleDict:
    test_dict = {
        "name": "test",
        "version": 3
    }
    d = IncompatibleDict(foo=test_dict['name'], bar=test_dict["version"])
    return(d)


l = print_dict()
```

## Pyright doesn't complain
```
❯ python3 -m pyright lib/rucio/core/test.py                                                                                                       ─╯
0 errors, 0 warnings, 0 informations 
```

## Mypy complains
```
❯ python3 -m mypy lib/rucio/core/test.py                                                                                                          ─╯
lib/rucio/core/test.py:13: error: Incompatible types (expression has type "object", TypedDict item "foo" has type "str")  [typeddict-item]
lib/rucio/core/test.py:13: error: Incompatible types (expression has type "object", TypedDict item "bar" has type "int")  [typeddict-item]
Found 2 errors in 1 file (checked 1 source file)
```

# 7. Incompatible TypedDict, passing parameters directly, extra parameters
```
#!/usr/bin/env python3
from typing import Any, TypedDict

class IncompatibleDict(TypedDict):
    foo: str
    bar: int

def print_dict() -> IncompatibleDict:
    test_dict = {
        "name": "test",
        "version": 3
    }
    d = IncompatibleDict(foo=test_dict['name'], bar=test_dict["version"], extra="unexpected")
    return(d)


l = print_dict()
```

## Pyright complains
```
❯ python3 -m pyright lib/rucio/core/test.py                                                                                                       ─╯
/home/rdm/tech/rucio/rucio/lib/rucio/core/test.py
  /home/rdm/tech/rucio/rucio/lib/rucio/core/test.py:13:9 - error: No overloads for "__init__" match the provided arguments
    Argument types: (Unknown, Unknown, Literal['unexpected']) (reportCallIssue)
1 error, 0 warnings, 0 informations 
```

## Mypy complains
```
❯ python3 -m mypy lib/rucio/core/test.py                                                                                                          ─╯
lib/rucio/core/test.py:13: error: Extra key "extra" for TypedDict "IncompatibleDict"  [typeddict-unknown-key]
lib/rucio/core/test.py:13: error: Incompatible types (expression has type "object", TypedDict item "foo" has type "str")  [typeddict-item]
lib/rucio/core/test.py:13: error: Incompatible types (expression has type "object", TypedDict item "bar" has type "int")  [typeddict-item]
Found 3 errors in 1 file (checked 1 source file)
```

rdimaio: # 8. Incompatible Pydantic model, passing parameters directly, static type checking
```
#!/usr/bin/env python3
from pydantic import BaseModel, ConfigDict

class IncompatibleModel(BaseModel):
    foo: int
    bar: int

def print_class() -> IncompatibleModel:
    test_dict = {
        "name": "test",
        "version": 3
    }
    return(IncompatibleModel(foo=test_dict['name'], bar=test_dict['version']))

l = print_class()
```

Pydantic complains here, as we saw earlier. From the point of view of the static type checkers:

## Pyright doesn't complain
```
❯ python3 -m pyright lib/rucio/core/test.py                                                                                                       ─╯
0 errors, 0 warnings, 0 informations 
```

## Mypy complains
```
❯ python3 -m mypy lib/rucio/core/test.py                                                                                                          ─╯
lib/rucio/core/test.py:13: error: Argument "foo" to "IncompatibleModel" has incompatible type "object"; expected "int"  [arg-type]
lib/rucio/core/test.py:13: error: Argument "bar" to "IncompatibleModel" has incompatible type "object"; expected "int"  [arg-type]
Found 2 errors in 1 file (checked 1 source file)
```

rdimaio: # 9. Incompatible dataclass, passing parameters directly
```
#!/usr/bin/env python3
from typing import Any, cast
from dataclasses import dataclass, field

@dataclass
class IncompatibleClass:
    foo: int
    bar: int

def print_class() -> IncompatibleClass:
    test_dict = {
        "name": "test",
        "version": 3
    }
    d = IncompatibleClass(foo=test_dict['name'], bar=test_dict['version'])
    return(d)


l = print_class()
```

## Pyright doesn't complain
```
❯ python3 -m pyright lib/rucio/core/test.py                                                                                                       ─╯
0 errors, 0 warnings, 0 informations 
```

## Mypy complains (ambiguously)
```
❯ python3 -m mypy lib/rucio/core/test.py                                                                                                          ─╯
lib/rucio/core/test.py:15: error: Argument "foo" to "IncompatibleClass" has incompatible type "object"; expected "int"  [arg-type]
lib/rucio/core/test.py:15: error: Argument "bar" to "IncompatibleClass" has incompatible type "object"; expected "int"  [arg-type]
Found 2 errors in 1 file (checked 1 source file)
```

Mypy complains even if I change `foo` to `str` - its problem seems to be that the type for `foo` and `bar` is not explicitly mentioned.

# 10. Incompatible dataclass, passing parameters directly
```
#!/usr/bin/env python3
from typing import Any, cast
from dataclasses import dataclass, field

@dataclass
class IncompatibleClass:
    foo: int
    bar: int

def print_class() -> IncompatibleClass:
    name = "test"
    version = 3
    d = IncompatibleClass(foo=name, bar=version)
    return(d)

l = print_class()
```

## Pyright complains
```
❯ python3 -m pyright lib/rucio/core/test.py                                                                                                       ─╯
/home/rdm/tech/rucio/rucio/lib/rucio/core/test.py
  /home/rdm/tech/rucio/rucio/lib/rucio/core/test.py:13:31 - error: Argument of type "Literal['test']" cannot be assigned to parameter "foo" of type "int" in function "__init__"
    "Literal['test']" is incompatible with "int" (reportArgumentType)
1 error, 0 warnings, 0 informations 
```

## Mypy complains
```
❯ python3 -m mypy lib/rucio/core/test.py                                                                                                          ─╯
lib/rucio/core/test.py:13: error: Argument "foo" to "IncompatibleClass" has incompatible type "str"; expected "int"  [arg-type]
Found 1 error in 1 file (checked 1 source file)
```

# 11. Incompatible TypedDict, passing parameters directly, adding explicitly types for the input arguments
```
#!/usr/bin/env python3
from typing import Any, TypedDict

class IncompatibleDict(TypedDict):
    foo: int
    bar: int

def print_dict() -> IncompatibleDict:
    name: str = "test"
    version: int = 3
    d = IncompatibleDict(foo=name, bar=version)
    return(d)


l = print_dict()
```

## Pyright complains
```
❯ python3 -m pyright lib/rucio/core/test.py                                                                                                       ─╯
/home/rdm/tech/rucio/rucio/lib/rucio/core/test.py
  /home/rdm/tech/rucio/rucio/lib/rucio/core/test.py:11:30 - error: Argument of type "Literal['test']" cannot be assigned to parameter "foo" of type "int" in function "__init__"
    "Literal['test']" is incompatible with "int" (reportArgumentType)
1 error, 0 warnings, 0 informations 
```

## Mypy complains
```
❯ python3 -m mypy lib/rucio/core/test.py                                                                                                          ─╯
lib/rucio/core/test.py:11: error: Incompatible types (expression has type "str", TypedDict item "foo" has type "int")  [typeddict-item]
Found 1 error in 1 file (checked 1 source file)
```

rdimaio: # Conclusion
- Static type checking: mypy seems to be slightly stricter compared to pyright - I think it would be worth considering we move to mypy.
- Runtime type checking: I think this should be necessary to catch cases where static type checking would not be enough. I opened an issue to discuss this: https://github.com/rucio/rucio/issues/6544

In the long term, I think the best approach would be to use mypy+pydantic and use pydantic models for struct typing. For now, it seems to be fine to use `TypedDict`, passing arguments directly, instead of passing them as kwargs, and adding explicit type annotations for the input arguments, so that the static type checkers are able to notice any issues.

rdimaio: Relevant thread https://github.com/rucio/rucio/discussions/6615

rdimaio:
Sorry for leaving this unattended until now.

I am breaking down this PR into smaller PRs to help move things forward. Each PR will be for a different aspect of this PR:
- One PR for the `ssh_sign` changes: https://github.com/rucio/rucio/pull/6647
- One for the `get_auth_token_ssh` changes (in this one, we can explore this proposal further: https://github.com/rucio/rucio/pull/6497#discussion_r1516152819): https://github.com/rucio/rucio/pull/6649
- For the type hints (split into one per file):
  - authentication.py: https://github.com/rucio/rucio/pull/6742

Some PRs linked above have already been merged. I am closing this PR.

tests/test_authentication.py
rdimaio: Addressed in this follow-up PR: https://github.com/rucio/rucio/pull/6649

Issue #6499, opened 2024-02-16
Cannot install Rucio if Python is in a folder that has spaces


SergeyYakubov:
### Description

Getting this [error](https://github.com/galaxyproject/galaxy/actions/runs/7906137202/job/21580314449?pr=17156#step:10:879) when trying to install rucio-clients during a GitHUb workflow.

```
 /bin/sh: 1: /home/runner/work/galaxy/galaxy/galaxy: not found
      Traceback (most recent call last):
        File "/home/runner/work/galaxy/galaxy/galaxy root/.tox/unit-coverage/lib/python3.12/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 353, in <module>
          main()
        File "/home/runner/work/galaxy/galaxy/galaxy root/.tox/unit-coverage/lib/python3.12/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 335, in main
          json_out['return_val'] = hook(**hook_input['kwargs'])
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "/home/runner/work/galaxy/galaxy/galaxy root/.tox/unit-coverage/lib/python3.12/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 118, in get_requires_for_build_wheel
          return hook(config_settings)
                 ^^^^^^^^^^^^^^^^^^^^^
        File "/tmp/pip-build-env-wc5q9ur8/overlay/lib/python3.12/site-packages/setuptools/build_meta.py", line 325, in get_requires_for_build_wheel
          return self._get_build_requires(config_settings, requirements=['wheel'])
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "/tmp/pip-build-env-wc5q9ur8/overlay/lib/python3.12/site-packages/setuptools/build_meta.py", line 295, in _get_build_requires
          self.run_setup()
        File "/tmp/pip-build-env-wc5q9ur8/overlay/lib/python3.12/site-packages/setuptools/build_meta.py", line 311, in run_setup
          exec(code, locals())
        File "<string>", line 57, in <module>
        File "/tmp/pip-install-c17xxtd8/rucio-clients_6410b254503d43e98525cbe009afe017/setuputil.py", line 120, in get_rucio_version
          raise RuntimeError("Could not fetch Rucio version")
[...]

PR #6500, opened 2024-02-16
Fix setuputil.py to deal with spaces in filepath

SergeyYakubov:
a small fix to allow whitespaces in Python path. See #6499


rdimaio (approved):
LGTM

PR #6509, opened 2024-02-22
Replicas: Add type hints

### LINKED ISSUE #6454: Type annotate the code

rdimaio:
Part of #6454
Tests pass locally

### DIFF
--- a/lib/rucio/core/volatile_replica.py
+++ b/lib/rucio/core/volatile_replica.py
@@ -13,9 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-
+from collections.abc import Iterable
 from datetime import datetime
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 
 from sqlalchemy import and_, or_, exists, update, insert
 from sqlalchemy.orm.exc import NoResultFound
@@ -32,14 +32,13 @@ if TYPE_CHECKING:
 
 
 @transactional_session
-def add_volatile_replicas(rse_id, replicas, *, session: "Session"):
+def add_volatile_replicas(rse_id: str, replicas: Iterable[dict[str, Any]], *, session: "Session") -> None:
     """
     Bulk add volatile replicas.
 
     :param rse_id: the rse id.
-    :param replicas: the list of volatile replicas.
+    :param replicas: the iterable of volatile replicas.
     :param session: The database session in use.
-    :returns: True is successful.
     """
     # first check that the rse is a volatile one
     try:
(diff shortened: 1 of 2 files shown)


lib/rucio/core/volatile_replica.py
```diff
@@ -32,14 +32,13 @@
 
 
 @transactional_session
-def add_volatile_replicas(rse_id, replicas, *, session: "Session"):
+def add_volatile_replicas(rse_id: str, replicas: list[dict[str, Any]], *, session: "Session") -> None:
```
dchristidis: The only use of `replicas` is in:

```python
for replica in replicas:
```

I wonder if using `typing.Sequence` is a better practise for *function parameters*, when there’s nothing `list`-specific. But I do not propose to change this now, just food for thought.

dchristidis (changes requested):
It’s useful to prefix the commit message with a component. I would guess that `Replicas: ` is the best match here? Otherwise the changes contained in the commit are perfect.

lib/rucio/core/volatile_replica.py
rdimaio: That's a good point, I agree. I'll make this change now so that I can also amend the commit message to address the other comment

dchristidis (dismissed):

lib/rucio/core/volatile_replica.py
dchristidis: Forgive me, I gave you bad advice. `typing.Sequence` is deprecated and must be replaced with `collections.abc.Sequence`.

dchristidis: I’m now even more conflicted because `collections.abc.Iterable` might be more accurate.

Example: if `replicas` is a generator expression, then it is not a `Sequence`, but it is `Iterable`, which is all that matters to do a `for replica in replicas`.

rdimaio: I agree (relevant thread: https://stackoverflow.com/questions/72157296/what-is-the-difference-between-type-hinting-a-variable-as-an-iterable-versus-a-s)

Amended in latest revision

rdimaio:
Test failure seems unrelated:

```
 tests/test_judge_evaluator.py:520: in test_judge_double_container_with_existing_rule
    assert not get_replica_locks(**F)
E   AssertionError: assert not [<rucio.db.sqla.models.ReplicaLock object at 0x7f7caf7dc460>]
E    +  where [<rucio.db.sqla.models.ReplicaLock object at 0x7f7caf7dc460>] = get_replica_locks(**{'name': '/belle/HCXHHT-judge_evaluator-judge_double_container_with_existing_rule-/cont_e77000060ac143ce88854e57b3cc4844/dataset_4d975954d5d4492c9821573cb9fe46bd/file_a92a8acd339444c3a7bc2a91730587ec', 'scope': mock})
```

re-running

dchristidis (approved):

Issue #6070, opened 2023-02-06
allow wildcard * on a scope level for list_dids


txp314:
### Description

Hey guys,

last year at the rucio workshop in Lancaster we talked about the possibility to allow wildcards on the scope level.

By doing so queries like:
`rucio list-dids --filter 'type=file' '*:some_prefix*'`  would be possible.

At the moment a * at the scope level is not allowed.

### Motivation

If I remember correctly you said at the workshop that this is not a good idea for CERN, due to your huge number of files and scopes, which could make those kind of queries very slow and/or computationally intensive.

Nevertheless, this would be really helpful for other communities. For example in our case at a research institute for solar physics, we create a scope for every observation campaign. Within an observation campaign there is are files from different instruments, which can be distinguished by filename.

So, if we want to get all the data from an instrument no matter in which campaign we could do something like:

`rucio list-dids --filter 'type=file' '*:telescopeA*'`

I am sure this would be useful for others as well.  

 

### Change

The idea in Lancaster during a brief discussion on this topic was to make this feature configurable.
So the default stays the same - no wildcard searches on scopes, but if needed the wildcard search on scope level can be enabled in the configs.

I hope you consider this proposal. Thank you.

mlassnig:
I think this would be a very nice extension, and should be a server side configuration option so users can't work around it :-)

bari12:
Hi Alex,
Yes this is a good idea, but needs to be made configurable as Mario said.

bari12:
Please nominate a reviewer/co-author for this **LARGE/XLARGE** sized issue before the start of development (See [sprint planning guideline](https://rucio.github.io/documentation/developer/sprint_planning#size-estimation)).

maxnoe:
@bari12 I think this mostly touches 

- Core
- Database
- REST & API

which looking at the component leads would make @mlassnig the natural choice?

maxnoe:
@bari12 @mlassnig 

The API endpoint for this is `'/<scope>/dids/search`. 

A literal interpretation of this issue would keep this endpoint and treat the scope `"*"` in a special way.

Two related questions here:

- If that is the way to implement it, do we want to only allow `*`? As a user, I would kind of expect then that this endpoint supports patterns, so also `foo*` would be valid and match only scopes that start with `foo`. 
- Or is the real usecase here "query across all scopes the user has access to", which might be better implemented by a new endpoint than by passing a `*` to the `scope` for the existing endpoint.

maxnoe:
[...]

PR #6514, opened 2024-02-26
Core & Internals: Allowing scope-level wildcards #6070

erlingstaff:
This is currently a draft, and this functionality does not currently have it's own tests. I am looking for thoughts on this implementation before I write them.

TL;DR
- If * in scope name, check if the new config option "scope_wildcard_allowed" is set to true, return FAILURE if not
- Serialize all scopes into a comma-concatenated string: scope1,scope2,scope3
- Deserialize into a list of scopes, use SQLAlchemy's in_ to construct query (If only one scope, i.e. no wildcards or only one matched, then use old equals operator for efficiency)

Important to note: commas will never occur naturally in a rucio scope, as it is prohibited to create a scope with it:
```
$ rucio-admin scope add --scope test,scope --account root

Details: Problem validating scope: 'test,scope' does not match '^[a-zA-Z_\\-.0-9]{1,25}$'

Failed validating 'pattern' in schema:
    {'description': 'Scope name',
     'pattern': '^[a-zA-Z_\\-.0-9]{1,25}$',
     'type': 'string'}

On instance:
    'test,scope'
```


erlingstaff:
Closing - will revisit and open another PR at a later date

PR #6516, opened 2024-02-27
Typing annotations: `rule.py`

rdimaio:
Adding type annotations to `rule.py` and some correlated type annotations to `rule_grouping.py`. Part of #6454.

Tests pass locally.

### DIFF
--- a/lib/rucio/core/rule.py
+++ b/lib/rucio/core/rule.py
@@ -2420,7 +2420,8 @@ def get_evaluation_backlog(expiration_time: int = 600, *, session: "Session") ->
 
     cached_backlog = REGION.get('rule_evaluation_backlog', expiration_time=expiration_time)
     if cached_backlog is NO_VALUE:
-        result = session.query(func.count(models.UpdatedDID.created_at), func.min(models.UpdatedDID.created_at)).one()
+        stmt = select(func.count(models.UpdatedDID.created_at), func.min(models.UpdatedDID.created_at))
+        result = session.execute(stmt).scalars().one()
         REGION.set('rule_evaluation_backlog', result)
         return result
     else:
--- a/lib/rucio/core/rule.py
+++ b/lib/rucio/core/rule.py
@@ -1134,11 +1134,7 @@ def repair_rule(rule_id, *, session: "Session", logger=logging.log):
         if hard_repair:
             __find_surplus_locks_and_remove_them(datasetfiles=datasetfiles,
                                                  locks=locks,
-                                                 replicas=replicas,
-                                                 source_replicas=source_replicas,
-                                                 rseselector=rseselector,
                                                  rule=rule,
-                                                 source_rses=[rse['id'] for rse in source_rses],
                                                  session=session)
 
             session.flush()
@@ -2568,17 +2564,13 @@ def __find_missing_locks_and_create_them(datasetfiles, locks, replicas, source_r
 
 
 @transactional_session
-def __find_surplus_locks_and_remove_them(datasetfiles, locks, replicas, source_replicas, rseselector, rule, source_rses, *, session: "Session", logger=logging.log):
+def __find_surplus_locks_and_remove_them(datasetfiles, locks, rule, *, session: "Session", logger=logging.log):
     """
     Find surplocks locks for a rule and delete them.
 
     :param datasetfiles:       Dict holding all datasets and files.
     :param locks:              Dict holding locks.
-    :param replicas:           Dict holding replicas.
-    :param source_replicas:    Dict holding all source replicas.
-    :param rseselector:        The RSESelector to be used.
     :param rule:               The rule.
-    :param source_rses:        RSE ids for eglible source RSEs.
     :param session:            Session of the db.
     :param logger:             Optional decorated logger that can be passed from the calling daemons or servers.
     :raises:                   InsufficientAccountLimit, IntegrityError, InsufficientTargetRSEs
--- a/lib/rucio/core/rule.py
+++ b/lib/rucio/core/rule.py
@@ -22,7 +22,7 @@ from datetime import datetime, timedelta
 from os import path
 from re import match
 from string import Template
-from typing import TYPE_CHECKING, Any, Callable, Optional, Type, TypeVar, Sequence
+from typing import TYPE_CHECKING, Any, Callable, Literal, Optional, Type, TypeVar, Sequence
 
 from dogpile.cache.api import NO_VALUE
 from sqlalchemy import select, update
@@ -45,7 +45,7 @@ from rucio.common.exception import (InvalidRSEExpression, InvalidReplicationRule
                                     InvalidSourceReplicaExpression)
 from rucio.common.policy import policy_filter, get_scratchdisk_lifetime
 from rucio.common.schema import validate_schema
-from rucio.common.types import InternalScope, InternalAccount, LoggerFunction
+from rucio.common.types import InternalScope, InternalAccount, LoggerFunction, RuleDict
 from rucio.common.utils import str_to_date, sizefmt, chunks
 from rucio.core import account_counter, rse_counter, request as request_core, transfer as transfer_core
 from rucio.core.account import get_account
@@ -143,8 +143,8 @@ class AutoApprove(PolicyPackageAlgorithms):
 
 
 @transactional_session
-def add_rule(dids: Sequence[dict[str, Any]], account: InternalAccount, copies: int, rse_expression: str, weight: str, lifetime: int, locked: bool, subscription_id: str,
-             grouping: RuleGrouping = RuleGrouping.DATASET, source_replica_expression: Optional[str] = None, activity: str = 'User Subscriptions', notify: RuleNotification = RuleNotification.NO, purge_replicas: bool = False,
+def add_rule(dids: Sequence[dict[str, Any]], account: InternalAccount, copies: int, rse_expression: str, grouping: Literal['ALL', 'DATASET', 'NONE'], weight: str, lifetime: int, locked: bool, subscription_id: str,
+             source_replica_expression: Optional[str] = None, activity: str = 'User Subscriptions', notify: Optional[Literal['Y', 'N', 'C']] = None, purge_replicas: bool = False,
              ignore_availability: bool = False, comment: Optional[str] = None, ask_approval: bool = False, asynchronous: bool = False, ignore_account_limit: bool = False,
              priority: int = 3, delay_injection: Optional[int] = None, split_container: bool = False, meta: Optional[dict[str, Any]] = None, *, session: "Session", logger: LoggerFunction = logging.log) -> list[str]:
     """
@@ -163,7 +163,7 @@ def add_rule(dids: Sequence[dict[str, Any]], account: InternalAccount, copies: i
     :param subscription_id:            The subscription_id, if the rule is created by a subscription.
     :param source_replica_expression:  Only use replicas as source from this RSEs.
     :param activity:                   Activity to be passed on to the conveyor.
-    :param notify:                     Notification setting of the rule.
+    :param notify:                     Notification setting of the rule ('Y', 'N', 'C'; None = 'N').
     :param purge_replicas:             Purge setting if a replica should be directly deleted after the rule is deleted.
     :param ignore_availability:        Option to ignore the availability of RSEs.
     :param comment:                    Comment about the rule.
@@ -184,6 +184,8 @@ def add_rule(dids: Sequence[dict[str, Any]], account: InternalAccount, copies: i
         raise InvalidValueForKey("The number of copies for a replication rule should be greater than 0.")
 
     rule_ids = []
+
+    grouping_value = {'ALL': RuleGrouping.ALL, 'NONE': RuleGrouping.NONE}.get(grouping, RuleGrouping.DATASET)
 
     with METRICS.timer('add_rule.total'):
         # 1. Resolve the rse_expression into a list of RSE-ids
@@ -228,6 +230,8 @@ def add_rule(dids: Sequence[dict[str, Any]], account: InternalAccount, copies: i
             rseselector = RSESelector(account=account, rses=rses, weight=weight, copies=copies, ignore_account_limit=ask_approval or ignore_account_limit, session=session)
 
         expires_at = datetime.utcnow() + timedelta(seconds=lifetime) if lifetime is not None else None
+
+        notify_value = {'Y': RuleNotification.YES, 'C': RuleNotification.CLOSE, 'P': RuleNotification.PROGRESS}.get(str(notify or ''), RuleNotification.NO)
 
         for elem in dids:
             # 3. Get the did
@@ -286,13 +290,13 @@ def add_rule(dids: Sequence[dict[str, Any]], account: InternalAccount, copies: i
                                                   copies=copies,
                                                   rse_expression=rse_expression,
                                                   locked=locked,
-                                                  grouping=grouping,
+                                                  grouping=grouping_value,
                                                   expires_at=expires_at,
                                                   weight=weight,
                                                   source_replica_expression=source_replica_expression,
                                                   activity=activity,
                                                   subscription_id=subscription_id,
-                                                  notification=notify,
+                                                  notification=notify_value,
                                                   purge_replicas=purge_replicas,
                                                   ignore_availability=ignore_availability,
                                                   comments=comment,
@@ -383,7 +387,7 @@ def add_rule(dids: Sequence[dict[str, Any]], account: InternalAccount, copies: i
 
 
 @transactional_session
-def add_rules(dids: Sequence[dict[str, Any]], rules: Sequence[models.ReplicationRule], *, session: "Session", logger: LoggerFunction = logging.log) -> dict[tuple[InternalScope, str], list[str]]:
+def add_rules(dids: Sequence[dict[str, Any]], rules: Sequence[RuleDict], *, session: "Session", logger: LoggerFunction = logging.log) -> dict[tuple[InternalScope, str], list[str]]:
     """
     Adds a list of replication rules to every did in dids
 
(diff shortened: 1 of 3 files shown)


lib/rucio/core/rule.py
```diff
@@ -143,10 +143,10 @@ def default(rule: models.ReplicationRule, did: models.DataIdentifier, session: '
 
 
 @transactional_session
-def add_rule(dids, account, copies, rse_expression, grouping, weight, lifetime, locked, subscription_id,
-             source_replica_expression=None, activity='User Subscriptions', notify=None, purge_replicas=False,
-             ignore_availability=False, comment=None, ask_approval=False, asynchronous=False, ignore_account_limit=False,
-             priority=3, delay_injection=None, split_container=False, meta=None, *, session: "Session", logger=logging.log):
+def add_rule(dids: Sequence[dict[str, Any]], account: InternalAccount, copies: int, rse_expression: str, weight: str, lifetime: int, locked: bool, subscription_id: str,
```
rdimaio: Note: position change of `grouping` due to it becoming non-default argument. I checked all `add_rule` calls to make sure that every time the arguments are named every time the function is called.

rdimaio: I've reverted this change in latest commit as it caused some issues - these were likely not due to the argument becoming non-default, but rather about `grouping` being passed as a `RuleGrouping` rather than a `str`.

lib/rucio/core/rule.py
```diff
@@ -231,7 +231,7 @@ def add_rule(dids, account, copies, rse_expression, grouping, weight, lifetime,
 
         expires_at = datetime.utcnow() + timedelta(seconds=lifetime) if lifetime is not None else None
 
-        notify = {'Y': RuleNotification.YES, 'C': RuleNotification.CLOSE, 'P': RuleNotification.PROGRESS}.get(notify, RuleNotification.NO)
+        notify_value = {'Y': RuleNotification.YES, 'C': RuleNotification.CLOSE, 'P': RuleNotification.PROGRESS}.get(str(notify or ''), RuleNotification.NO)
```
rdimaio, line 234: `str(notify or '')` added to fix this issue:

```
 Found 1 new problems in /home/runner/work/rucio/rucio/lib/rucio/core/rule.py
  - 1 errors with message """Argument of type "Literal['Y', 'N', 'C'] | None" cannot be assigned to parameter "__key" of type "str" in function "get"
                               Type "Literal['Y', 'N', 'C'] | None" cannot be assigned to type "str"
                                 "None" is incompatible with "str"""".
    Candidates: line 234
```

bari12 (approved):

bari12:
I am squashing the commits together.

rdimaio:
> I am squashing the commits together.

I would keep them separate - I split them on purpose in that way to avoid having non-behavioural and behavioural changes in the same commit, following the guidelines of what I was asked to do for other PRs e.g. https://github.com/rucio/rucio/pull/6497/

lib/rucio/core/rule.py
```diff
@@ -15,14 +15,14 @@
...
-from collections.abc import Iterator
+from collections.abc import Iterable, Iterator
 from configparser import NoOptionError, NoSectionError
 from copy import deepcopy
 from datetime import datetime, timedelta
 from os import path
 from re import match
 from string import Template
-from typing import TYPE_CHECKING, Any, Callable, Optional, Type, TypeVar
+from typing import TYPE_CHECKING, Any, Callable, Literal, Optional, Type, TypeVar, Sequence
```
dchristidis, line 25: Please be advised that `typing.Sequence` is deprecated and should be replaced with `collections.abc.Sequence`. See https://docs.python.org/3/library/typing.html#typing.Sequence. This applies to the other file too.

rdimaio: Addressed here https://github.com/rucio/rucio/pull/6539

lib/rucio/core/rule.py
```diff
@@ -143,10 +143,10 @@ def default(rule: models.ReplicationRule, did: models.DataIdentifier, session: '
...
-             source_replica_expression=None, activity='User Subscriptions', notify=None, purge_replicas=False,
-             ignore_availability=False, comment=None, ask_approval=False, asynchronous=False, ignore_account_limit=False,
-             priority=3, delay_injection=None, split_container=False, meta=None, *, session: "Session", logger=logging.log):
+def add_rule(dids: Sequence[dict[str, Any]], account: InternalAccount, copies: int, rse_expression: str, grouping: Literal['ALL', 'DATASET', 'NONE'], weight: str, lifetime: int, locked: bool, subscription_id: str,
+             source_replica_expression: Optional[str] = None, activity: str = 'User Subscriptions', notify: Optional[Literal['Y', 'N', 'C']] = None, purge_replicas: bool = False,
+             ignore_availability: bool = False, comment: Optional[str] = None, ask_approval: bool = False, asynchronous: bool = False, ignore_account_limit: bool = False,
+             priority: int = 3, delay_injection: Optional[int] = None, split_container: bool = False, meta: Optional[dict[str, Any]] = None, *, session: "Session", logger: LoggerFunction = logging.log) -> list[str]:
```
dchristidis, line 149: 1. The style that we’re going for is:
    ```python
    def add_rule(
        dids: Sequence[dict[str, Any]],
        account: InternalAccount,
        copies: int,
        rse_expression: str,
        grouping: Literal['ALL', 'DATASET', 'NONE'],
        weight: str,
        lifetime: int,
        locked: bool,
        subscription_id: str,
        source_replica_expression: Optional[str] = None,
        activity: str = 'User Subscriptions',
        notify: Optional[Literal['Y', 'N', 'C']] = None,
        purge_replicas: bool = False,
        ignore_availability: bool = False,
        comment: Optional[str] = None,
        ask_approval: bool = False,
        asynchronous: bool = False,
        ignore_account_limit: bool = False,
        priority: int = 3,
        delay_injection: Optional[int] = None,
        split_container: bool = False,
        meta: Optional[dict[str, Any]] = None,
        *,
        session: "Session", logger: LoggerFunction = logging.log
    ) -> list[str]:
    ```
    It’s not necessary when everything can fit in one line, but type hints add a lot of noise, so I think one line per parameter makes a lot of sense and scales well.
    This comment applies throughout the PR.
 2. The existing code appears to accept `lifetime=None`. I think you must use `Optional` here.
 3. `subscription_id` can definitely be `None`.
 4. `P` is an acceptable value for `notify`. This has to be included in `RuleDict` too.
 5. *NOT FOR 34.0.0rc2:* Why not define `dids` more precisely? I think it only has two possible keys.

lib/rucio/core/rule.py
dchristidis: I get the `notify or ''`, but why is `str()` necessary?

Then further below you use:

```python
notify = {'Y': RuleNotification.YES, 'C': RuleNotification.CLOSE, 'P': RuleNotification.PROGRESS, None: RuleNotification.NO}.get(rule.get('notify'))
```

Why the difference?

lib/rucio/core/rule.py
```diff
@@ -765,7 +765,7 @@ def inject_rule(rule_id, *, session: "Session", logger=logging.log):
 
 
 @stream_session
-def list_rules(filters={}, *, session: "Session"):
+def list_rules(filters: dict[str, Any] = {}, *, session: "Session") -> Iterator[dict]:
```
dchristidis, line 768: 1. Is there some reason you avoided `dict[str, Any]` for the return type? To be honest, I don’t know how much benefit it offers against plain `dict`, but I am just wondering.
 2. We should not be using mutable objects as default arguments due to some weird esoteric Python behaviour (I thought I had opened an issue about this long time ago, but I can’t find it). So `filters` should be `None` by default, not an empty dictionary. We have this elsewhere as well, but I think this particular one is obvious to fix (whereas the others might not).

lib/rucio/core/rule.py
```diff
@@ -1779,7 +1775,7 @@ def get_rules_beyond_eol(date_check, worker_number, total_workers, *, session: "
 
 
 @read_session
-def get_expired_rules(total_workers, worker_number, limit=100, blocked_rules=[], *, session: "Session"):
+def get_expired_rules(total_workers: int, worker_number: int, limit: int = 100, blocked_rules: Sequence[str] = [], *, session: "Session") -> list[tuple[str, str]]:
```
dchristidis, line 1778: *NOT FOR 34.0.0rc2:* In `get_updated_dids()` above you used `blocked_dids: Iterable[tuple[str, str]] = []`. Why was that one an `Iterable` but this one is a `Sequence`? I think this applies elsewhere as well.

I think `Sequence` is an `Iterable` but not the other way around?

lib/rucio/core/rule.py
```diff
@@ -1717,7 +1713,7 @@ def re_evaluate_did(scope, name, rule_evaluation_action, *, session: "Session"):
 
 
 @read_session
-def get_updated_dids(total_workers, worker_number, limit=100, blocked_dids=[], *, session: "Session"):
+def get_updated_dids(total_workers: int, worker_number: int, limit: int = 100, blocked_dids: Iterable[tuple[str, str]] = [], *, session: "Session") -> list[tuple[str, str]]:
```
dchristidis, line 1716: The query has four columns, so the return type should be `list[tuple[str, InternalScope, str, DIDReEvaluation]]`.

lib/rucio/core/rule.py
```diff
@@ -1814,7 +1810,7 @@ def get_expired_rules(total_workers, worker_number, limit=100, blocked_rules=[],
 
 
 @read_session
-def get_injected_rules(total_workers, worker_number, limit=100, blocked_rules=[], *, session: "Session"):
+def get_injected_rules(total_workers: int, worker_number: int, limit: int = 100, blocked_rules: Sequence[str] = [], *, session: "Session") -> list[tuple[str, str]]:
```
dchristidis, line 1813: The return type is currently `list[tuple[str]]`. but the implementation is poor. I would suggest you change it to `list[str]` and make the necessary adjustments in the implementation and in the caller.

lib/rucio/core/rule.py
```diff
@@ -1849,7 +1845,7 @@ def get_injected_rules(total_workers, worker_number, limit=100, blocked_rules=[]
 
 
 @read_session
-def get_stuck_rules(total_workers, worker_number, delta=600, limit=10, blocked_rules=[], *, session: "Session"):
+def get_stuck_rules(total_workers: int, worker_number: int, delta: int = 600, limit: int = 10, blocked_rules: Sequence[str] = [], *, session: "Session") -> list[tuple[str, str]]:
```
dchristidis, line 1848: Exactly the same comment as above for the return type. Why do we do such things? :smiling_face_with_tear:

lib/rucio/core/rule.py
```diff
@@ -2242,7 +2238,7 @@ def insert_rule_history(rule, recent=True, longterm=False, *, session: "Session"
 
 
 @transactional_session
-def approve_rule(rule_id, approver=None, notify_approvers=True, *, session: "Session"):
+def approve_rule(rule_id: str, approver: Optional[str] = None, notify_approvers: bool = True, *, session: "Session") -> None:
```
dchristidis, line 2241: *NOT FOR 34.0.0rc2:* Does `approver: Optional[str] = None` offer any benefit against `approver: str = ''`? Same for `deny_rule()` below.

lib/rucio/core/rule.py
```diff
@@ -2418,23 +2414,26 @@ def examine_rule(rule_id, *, session: "Session"):
...
-        result = session.query(func.count(models.UpdatedDID.created_at), func.min(models.UpdatedDID.created_at)).one()
+    cached_backlog = REGION.get('rule_evaluation_backlog', expiration_time=expiration_time)
+    if cached_backlog is NO_VALUE:
+        stmt = select(func.count(models.UpdatedDID.created_at), func.min(models.UpdatedDID.created_at))
+        result = session.execute(stmt).scalars().one()
         REGION.set('rule_evaluation_backlog', result)
-    return result
+        return result
+    else:
+        return cached_backlog  # type: ignore
```
dchristidis, line 2432: See https://github.com/rucio/rucio/pull/6493/files#r1521715742. Please note how @erlingstaff has formatted the query. I think you should copy his work and apply the corrections on top of it so that he can ignore this function completely.

dchristidis (commented):
I’m submitting part of my review so that you can start working while I finish the rest of it.

lib/rucio/core/rule.py
```diff
@@ -2521,11 +2520,14 @@ def list_rules_for_rse_decommissioning(
 
 
 @transactional_session
-def __find_missing_locks_and_create_them(datasetfiles, locks, replicas, source_replicas, rseselector, rule, source_rses, *, session: "Session", logger=logging.log):
+def __find_missing_locks_and_create_them(datasetfiles: Sequence[dict[str, Any]], locks: dict[tuple[str, str], Sequence[models.ReplicaLock]],
+                                         replicas: dict[tuple[str, str], Any], source_replicas: dict[tuple[str, str], Any],
+                                         rseselector: RSESelector, rule: models.ReplicationRule, source_rses: Sequence[str], *,
+                                         session: "Session", logger: LoggerFunction = logging.log) -> None:
```
dchristidis, line 2526: 1. Since the key for `locks` includes a scope, shouldn’t it be `tupe[InternalScope, str]`? Same for `replicas`, `source_replicas`, and similar parameters in other functions.
 2. *NOT FOR 34.0.0rc2:* How come the value of `replicas` and `source_replicas` was left at `Any`? Also for similar functions.

lib/rucio/core/rule.py
```diff
@@ -2978,7 +2979,12 @@ def __evaluate_did_attach(eval_did, *, session: "Session", logger=logging.log):
 
 
 @transactional_session
-def __resolve_did_to_locks_and_replicas(did, nowait=False, restrict_rses=None, source_rses=None, only_stuck=False, *, session: "Session"):
+def __resolve_did_to_locks_and_replicas(did: models.DataIdentifier, nowait: bool = False, restrict_rses: Optional[Sequence[str]] = None,
+                                        source_rses: Optional[Sequence[str]] = None, only_stuck: bool = False, *,
+                                        session: "Session") -> tuple[list[dict[str, Any]],
+                                                                     dict[tuple[str, str], models.ReplicaLock],
+                                                                     dict[tuple[str, str], Any],
+                                                                     dict[tuple[str, str], str]]:
```
dchristidis, line 2987: The code suggests that the hint for the *third* member of the return type should be `dict[tuple[str, str], models.RSEFileAssociation]`. Did you try it and found issues? Same comment for the similar functions that follow.

lib/rucio/core/rule.py
```diff
@@ -3365,11 +3380,11 @@ def __create_rule_approval_email(rule: "models.ReplicationRule", *, session: "Se
 
 
 @transactional_session
-def _create_recipients_list(rse_expression: str, filter_=None, *, session: "Session"):
+def _create_recipients_list(rse_expression: str, filter_: Optional[str] = None, *, session: "Session") -> list[tuple[str, InternalAccount]]:
```
dchristidis, line 3383: Annoyingly, in the current implementation, `InternalAccount` should be `Union[str, InternalAccount]`.

lib/rucio/core/rule.py
```diff
@@ -3526,7 +3541,7 @@ def archive_localgroupdisk_datasets(scope, name, *, session: "Session", logger=l
 
 @policy_filter
 @read_session
-def get_scratch_policy(account, rses, lifetime, *, session: "Session"):
+def get_scratch_policy(account: str, rses: Sequence[dict[str, Any]], lifetime: int, *, session: "Session") -> int:
```
dchristidis, line 3544: 1. `account` is `InternalAccount`
 2. `lifetime` is `Optional[int]`

lib/rucio/core/rule.py
dchristidis: In addition, `weight` can be `None`. Also for `RuleDict`.

dchristidis (commented):
I think this is all my comments for `rule.py`. Some may apply also to `rule_grouping.py`

lib/rucio/core/rule.py
rdimaio: If I change `lifetime` to `Optional[int]`, then the linter complains that the return type should also be `Optional[int]`. I can rewrite this to only return `lifetime` if it's not `None`, but wanted to check first if that's correct. I think in this case `Optional[int]` might be the correct return type? 

`if not has_account_attribute(account=account, key='admin', session=session) and lifetime is None` and if none of the RSEs are SCRATCHDISKs, then `lifetime` would be `None`

dchristidis: Well observed. Yes, I believe returning `None` is a valid. All the callers of this functions handle it properly.

lib/rucio/core/rule.py
```diff
@@ -1755,7 +1751,7 @@ def get_updated_dids(total_workers, worker_number, limit=100, blocked_dids=[], *
 
 
 @read_session
-def get_rules_beyond_eol(date_check, worker_number, total_workers, *, session: "Session"):
+def get_rules_beyond_eol(date_check: datetime, worker_number: int, total_workers: int, *, session: "Session") -> list[dict[str, Any]]:
```
dchristidis, line 1754: Also the return type here is wrong.

rdimaio: Would this return type be correct?

```
                         list[tuple[InternalScope, 
                                         str,
                                         str,
                                         bool,
                                         str,
                                         datetime,
                                         datetime,
                                         InternalAccount]]
```

dchristidis: `ReplicationRule.expires_at` can be `None`. Technically, it’s the same for `eol_at`, but the query filters them out; probably it’s fine to keep that one as is.

The checker may complain about `tuple`, but that can be fixed in the SQLAlchemy migration.

lib/rucio/core/rule.py
rdimaio: 1. I agree with you and I've now made this change for the function signatures that were too long, but I would argue that until we have explicit style guidelines (https://github.com/rucio/documentation/issues/287) and checks in the CI that enforce these guidelines, the only style that can be inferred is the `pre-existing` one in the file; in this case it wasn't one parameter per line, but I think it might have been based on line length.

rdimaio:
Follow-up for the "NOT FOR 34.0.0rc2" comments: https://github.com/rucio/rucio/pull/6575

PR #6520, opened 2024-02-28
Client: Run client without configuration file; Fix #6410

### LINKED ISSUE #6410: Using Python client without configuration file
### Description

Currently, the Rucio client requires the presence of a configuration file to make it work properly. It would be very useful if that was not the case and the library worked properly without it.

### Motivation

We at Oak Ridge National Laboratory are trying to create a Rucio storage plugin for Galaxy (https://galaxyproject.org/). It works nicely, but we have to put a Rucio configuration file at a known location. This makes it hardly possible to use with dynamical parameters, in tests, etc. Even if the client allows to add some parameters during the constructor call, it still wants to read the config file and fails when it is not there. In some situations even `from rucio.client import Client` wants to read config file to get logging info (sometimes it works, though, I'm not sure what the difference is - probably depends on log configuration).

### Change

Add an option to get rid of the config file and use only constructor parameters. Get rid of using config file during import calls, pass config parameters used during client construction to inner functions/classes, ...

voetberg:
Changes:
            * Change `RuntimeError(Config Not Found)` to `ConfigNotFound` Error, config_get correctly handles those and returns a default client config. The `RuntimeError` was being caught in different places and causing conflict. 
            * Added typing for baseclient, initialization broken into functions to avoid typing conflicts for optionals, including an error check for the missing config. 
            * Added test to run multiple client commands without configuration file in place, including func decorator for extended testing if that's desired.

### DIFF
--- a/lib/rucio/client/baseclient.py
+++ b/lib/rucio/client/baseclient.py
@@ -23,11 +23,14 @@ import random
 import sys
 import time
 from configparser import NoOptionError, NoSectionError
+from logging import Logger
 from os import environ, fdopen, geteuid, makedirs, path
 from shutil import move
 from tempfile import mkstemp
+from typing import Any, Generator, Optional
 from urllib.parse import urlparse
 
+import requests
 from dogpile.cache import make_region
 from requests import Response, Session
 from requests.exceptions import ConnectionError
@@ -36,7 +39,7 @@ from requests.status_codes import codes
 from rucio import version
 from rucio.common import exception
 from rucio.common.config import config_get, config_get_bool, config_get_int
-from rucio.common.exception import CannotAuthenticate, ClientProtocolNotSupported, MissingClientParameter, MissingModuleException, NoAuthInformation, ServerConnectionException
+from rucio.common.exception import CannotAuthenticate, ClientProtocolNotSupported, ConfigNotFound, MissingClientParameter, MissingModuleException, NoAuthInformation, ServerConnectionException
 from rucio.common.extra import import_extras
 from rucio.common.utils import build_url, get_tmp_dir, my_key_generator, parse_response, setup_logger, ssh_sign
 
@@ -76,7 +79,17 @@ class BaseClient:
     TOKEN_PREFIX = 'auth_token_'
     TOKEN_EXP_PREFIX = 'auth_token_exp_'
 
-    def __init__(self, rucio_host=None, auth_host=None, account=None, ca_cert=None, auth_type=None, creds=None, timeout=600, user_agent='rucio-clients', vo=None, logger=None):
+    def __init__(self,
+                 rucio_host: Optional[str] = None,
+                 auth_host: Optional[str] = None,
+                 account: Optional[str] = None,
+                 ca_cert: Optional[str] = None,
+                 auth_type: Optional[str] = None,
+                 creds: Optional[dict[str, Any]] = None,
+                 timeout: Optional[int] = 600,
+                 user_agent: Optional[str] = 'rucio-clients',
+                 vo: Optional[str] = None,
+                 logger: Logger = LOG) -> None:
         """
         Constructor of the BaseClient.
         :param rucio_host: The address of the rucio server, if None it is read from the config file.
@@ -94,9 +107,8 @@ class BaseClient:
         """
 
         self.host = rucio_host
-        self.list_hosts = []
         self.auth_host = auth_host
-        self.logger = logger or LOG
+        self.logger = logger
         self.session = Session()
         self.user_agent = "%s/%s" % (user_agent, version.version_string())  # e.g. "rucio-clients/0.2.13"
         sys.argv[0] = sys.argv[0].split('/')[-1]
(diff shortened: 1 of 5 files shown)


lib/rucio/client/baseclient.py
```diff
@@ -76,7 +77,17 @@ class BaseClient:
...
+                 rucio_host: Optional[str] = None,
+                 auth_host: Optional[str] = None,
+                 account: Optional[str] = None,
+                 ca_cert: Optional[str] = None,
+                 auth_type: Optional[str] = None,
+                 creds: Optional[dict[str, Any]] = None,
+                 timeout: Optional[int] = 600,
+                 user_agent: Optional[str] = 'rucio-clients',
+                 vo: Optional[str] = None,
+                 logger: Callable = LOG):  # type: ignore
```
rdimaio: why is this `type: ignore` needed?

voetberg: I'm not exactly sure where the problem happens, but assigning the log type to a callable (only here in the baseclient for some reason....) produces: 
`Expression of type "Logger" cannot be assigned to parameter of type "(...) -> Unknown"
  Type "Logger" cannot be assigned to type "(...) -> Unknown"`

rdimaio: Instead of `Callable`, you can use `Logger` (`from logging import Logger`) and remove the `type: ignore` comment.

This is due to the default value assigned here, `LOG`, which is defined here: https://github.com/voetberg/rucio/blob/cf7717f3bfc3c9c0edb22c0c44b275c06d679c20/lib/rucio/client/baseclient.py#L49. `setup_logger` returns a `Logger` object. (offtopic: I noticed that the `LoggerFunction` we define in `types.py` is just a generic `Callable[..., Any]`; we might want to make this more specific later on)

lib/rucio/client/baseclient.py
```diff
@@ -262,30 +188,126 @@ def __init__(self, rucio_host=None, auth_host=None, account=None, ca_cert=None,
...
+                                    '\'x509_proxy\' not set in the configuration file.' % fname)
+                elif self.auth_type == 'ssh':
+                    creds['ssh_private_key'] = path.abspath(path.expanduser(path.expandvars(config_get('client', 'ssh_private_key'))))
+            except (NoOptionError, NoSectionError) as error:
+                if error.args[0] != 'client_key':
+                    raise MissingClientParameter('Option \'%s\' cannot be found in config file' % error.args[0])
+        return creds
 
-    def _get_exception(self, headers, status_code=None, data=None):
+    def _get_exception(self, headers: dict[str, str], status_code: Optional[int] = None, data=None) -> tuple[Any, Any]:
```
rdimaio:
```suggestion
    def _get_exception(self, headers: dict[str, str], status_code: Optional[int] = None, data=None) -> tuple[Any, str]:
```

For the return type of the exception class, I'm not 100% sure, but `type[exception.RucioException]` might work

lib/rucio/client/baseclient.py
```diff
@@ -804,7 +826,7 @@ def __get_token_saml(self):
         self.auth_token = result.headers['X-Rucio-Auth-Token']
         return True
 
-    def __get_token(self):
+    def __get_token(self) -> Optional[bool]:
```
rdimaio:
```suggestion
    def __get_token(self) -> None:
```

This function doesn't seem to return anything

tests/test_clients.py
```diff
@@ -30,55 +27,70 @@ def client_token_path_override(file_config_mock, function_scope_prefix, tmp_path
...
+        from rucio.client.baseclient import BaseClient
+
         creds = {'client_cert': self.usercert,
                  'client_key': self.userkey}
         BaseClient(account='root', ca_cert=self.cacert, auth_type='x509', creds=creds, vo=vo)
 
     def testx509NonExistingCert(self, vo):
         """ CLIENTS (BASECLIENT): authenticate with x509 with missing certificate."""
         creds = {'client_cert': '/opt/rucio/etc/web/notthere.crt'}
+        from rucio.client.baseclient import BaseClient
```
rdimaio, line 78: why is `BaseClient` imported separately for each test?

tests/test_clients.py
```diff
@@ -140,13 +158,47 @@ def test_import_without_config_file(self, vo):
...
-        configfile = Config().configfile
-        rename(configfile, f"{configfile}.tmp")
-        try:
-            exitcode, _, err = execute("python -c 'from rucio.client import Client'")
-            print(exitcode, err)
-            assert exitcode == 0
-            assert "RuntimeError: Could not load Rucio configuration file." not in err
-        finally:
-            # This is utterly important to not mess up the environment.
-            rename(f"{configfile}.tmp", configfile)
```
rdimaio, line 152: This is a pretty ominous comment: `# This is utterly important to not mess up the environment.`

The code block was added here: https://github.com/rucio/rucio/commit/9f78b001a1468e32538fb2afbfb534b3ceb91451

I think this is not needed anymore in this context, though

voetberg: Yeah I thought so as well. But I was too amused by the severity of the comment to remove it haha.

tests/test_clients.py
voetberg: When BaseClient is imported it runs `setup_logger`, which uses `config_get`, and thus, initializes the config, and renders the added test pointless. I can move the new tests to a separate file to avoid the separate imports, but I figured the added test belongs here.

lib/rucio/client/baseclient.py
voetberg: linting seems to agree, thanks for the suggestion

rdimaio:
I checked the type errors (https://github.com/rucio/rucio/actions/runs/8441860751/job/23121972309?pr=6520) - in my opinion, they can be ignored. Both errors existed before this change (verified on master branch) and they are due to not properly checking if a variable exists before accessing it. They can be fixed later on when type hints are added to those parts of the code.

rdimaio (approved):

voetberg:
> I checked the type errors (https://github.com/rucio/rucio/actions/runs/8441860751/job/23121972309?pr=6520) - in my opinion, they can be ignored. Both errors existed before this change (verified on master branch) and they are due to not properly checking if a variable exists before accessing it. They can be fixed later on when type hints are added to those parts of the code.

I'm of this opinion too, I was considering fixing them but figured it outside the scope of this PR. Annoying errors, but trying to trace a blame and ending up at this unrelated pr would be more annoying.

rdimaio:
LGTM but I think it's best if someone with more experience on the repository has a look as well

PR #6533, opened 2024-03-06
Tests: Remove writing curl output to /dev/null

rdimaio:
This test sometimes errors out due to write errors when curl writes output to /dev/null.
Reference: https://github.com/rucio/rucio/pull/6445#issuecomment-1977134159


bari12:
Merged via other PR

PR #6539, opened 2024-03-07
Typing: Import Sequence from collections.abc instead of typing; #6454

rdimaio:
Follow-up to https://github.com/rucio/rucio/pull/6516#discussion_r1516248381

### DIFF
--- a/lib/rucio/core/rule.py
+++ b/lib/rucio/core/rule.py
@@ -15,14 +15,14 @@
 
 import json
 import logging
-from collections.abc import Iterable, Iterator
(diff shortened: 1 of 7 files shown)


dchristidis (changes requested):
Kindly do the same on `rucio.core.rule_grouping`.

rdimaio:
> Kindly do the same on `rucio.core.rule_grouping`.

done in latest revision - I've also updated all the other instances of this in the repo

bari12 (approved):

PR #6552, opened 2024-03-13
Typing: Fix SQLA query and type hints in get_evaluation_backlog

### LINKED ISSUE #6550: Regression from #6516
### Description

The SQLAlchemy 2.0 syntax migration done as part of that PR changes the return value of `get_evaluation_backlog()`.

https://github.com/rucio/rucio/blob/cdd847fecb3041bc0007ba42735669fefeac6ffd/lib/rucio/core/rule.py#L2417-L2432

There is a misuse of `scalars()`.

### Steps to reproduce

Running the Reaper or the Necromancer causes an unhandled exception.

### Rucio Version

34.0.0rc1

### Additional Information

_No response_

rdimaio:
fixes #6550

### DIFF
--- a/lib/rucio/core/rule.py
+++ b/lib/rucio/core/rule.py
@@ -22,9 +22,9 @@ from datetime import datetime, timedelta
 from os import path
 from re import match
 from string import Template
-from typing import TYPE_CHECKING, Any, Callable, Literal, Optional, Type, TypeVar
+from typing import TYPE_CHECKING, Any, Callable, Literal, Optional, Type, TypeVar, Union
 
-from dogpile.cache.api import NO_VALUE
+from dogpile.cache.api import NoValue
 from sqlalchemy import select, update
 from sqlalchemy.exc import IntegrityError, StatementError
 from sqlalchemy.exc import NoResultFound  # https://pydoc.dev/sqlalchemy/latest/sqlalchemy.exc.NoResultFound.html
@@ -2422,14 +2422,16 @@ def get_evaluation_backlog(expiration_time: int = 600, *, session: "Session") ->
     :returns:     Tuple (Count, Datetime of oldest entry)
     """
 
-    cached_backlog = REGION.get('rule_evaluation_backlog', expiration_time=expiration_time)
-    if cached_backlog is NO_VALUE:
-        stmt = select(func.count(models.UpdatedDID.created_at), func.min(models.UpdatedDID.created_at))
-        result = session.execute(stmt).scalars().one()
+    cached_backlog: Union[NoValue, tuple[int, datetime]] = REGION.get('rule_evaluation_backlog', expiration_time=expiration_time)
+    if isinstance(cached_backlog, NoValue):
+        stmt = select(
+            func.count(models.UpdatedDID.created_at),
+            func.min(models.UpdatedDID.created_at)
+        )
+        result = session.execute(stmt).one()._tuple()
         REGION.set('rule_evaluation_backlog', result)
         return result
-    else:
-        return cached_backlog  # type: ignore
+    return cached_backlog
 
 
 @transactional_session


lib/rucio/core/rule.py
```diff
@@ -2414,18 +2415,18 @@ def examine_rule(rule_id: str, *, session: "Session") -> dict[str, Any]:
 
 
 @transactional_session
-def get_evaluation_backlog(expiration_time: int = 600, *, session: "Session") -> tuple[int, datetime]:
+def get_evaluation_backlog(expiration_time: int = 600, *, session: "Session") -> Row[tuple[int, datetime]]:
```
dchristidis: No, `Row[tuple[int, datetime]]` is not correct. `tuple[int, datetime]` is, but you may need to use `result._tuple()` to convince the type checker, see https://docs.sqlalchemy.org/en/20/core/connections.html#sqlalchemy.engine.Row._tuple.

rdimaio: ignore the latest force-push, hadn't seen this comment

done in latest revision

lib/rucio/core/rule.py
```diff
@@ -2414,22 +2414,24 @@ def examine_rule(rule_id: str, *, session: "Session") -> dict[str, Any]:
...
+def get_evaluation_backlog(expiration_time: int = 600, *, session: "Session") -> Union[NoValue, tuple[int, datetime]]:
     """
     Counts the number of entries in the rule evaluation backlog.
     (Number of files to be evaluated)
 
     :returns:     Tuple (Count, Datetime of oldest entry)
     """
 
-    cached_backlog = REGION.get('rule_evaluation_backlog', expiration_time=expiration_time)
+    cached_backlog: Union[NoValue, tuple[int, datetime]] = REGION.get('rule_evaluation_backlog', expiration_time=expiration_time)
```
rdimaio, line 2425: This annotation is not necessary for the type checker, but I think it's okay to keep it

lib/rucio/core/rule.py
```diff
@@ -2414,22 +2414,24 @@ def examine_rule(rule_id: str, *, session: "Session") -> dict[str, Any]:
 
 
 @transactional_session
-def get_evaluation_backlog(expiration_time: int = 600, *, session: "Session") -> tuple[int, datetime]:
+def get_evaluation_backlog(expiration_time: int = 600, *, session: "Session") -> Union[NoValue, tuple[int, datetime]]:
```
dchristidis: Under what circumstances can this function return `NoValue`?

rdimaio: You're right, it doesn't happen under any circumstances.

The type checker complains when `if cached_backlog is NO_VALUE:` is used as the condition - it still thinks that `cached_backlog` might be of type `NoValue` outside of the if statement. If I use `if isinstance(cached_backlog, NoValue):` instead, it works correctly, but the `NoValue` descriptor mentions that `The .NO_VALUE constant should be used.` What do you think?

lib/rucio/core/rule.py
dchristidis: I think we had to do something about it in the past because the type of `REGION.get()` is `Any`, but I can’t recall the specifics.

lib/rucio/core/rule.py
dchristidis: Good catch, `isinstance(cached_backlog, NoValue)` is the correct way to implement a type guard.

lib/rucio/core/rule.py
```diff
@@ -2414,22 +2414,24 @@ def examine_rule(rule_id: str, *, session: "Session") -> dict[str, Any]:
...
+        stmt = select(
+            func.count(models.UpdatedDID.created_at),
+            func.min(models.UpdatedDID.created_at)
+        )
+        result = session.execute(stmt).one()._tuple()
         REGION.set('rule_evaluation_backlog', result)
         return result
-    else:
-        return cached_backlog  # type: ignore
+    return cached_backlog
```
dchristidis, line 2434: I’m not asking you to change this, but I am wondering why linters think omitting the `else` clause is preferable.

dchristidis (approved):
Thanks for the quick fix.

lib/rucio/core/rule.py
rdimaio: It wasn't the linter - I omitted it because it was omitted in the changes you linked to as guidelines for this change: https://github.com/rucio/rucio/pull/6493/files#r1521715742

I'm okay with either having it removed or keeping it

rdimaio:
> Thanks for the quick fix.

Thanks for the quick reviews!

lib/rucio/core/rule.py
dchristidis: No, please keep it. I guess that @erlingstaff did it because of [R1705](https://pylint.pycqa.org/en/latest/user_guide/messages/refactor/no-else-return.html). I just can’t agree that it is *undeniably* better than the ‘problematic’ code.

bari12 (approved):

PR #6556, opened 2024-03-13
Refactoring and fixing type annotations in rule.py

### LINKED ISSUE #6454: Type annotate the code

rdimaio:
Addressing changes raised here https://github.com/rucio/rucio/pull/6516

#6454

### DIFF
--- a/lib/rucio/core/rule.py
+++ b/lib/rucio/core/rule.py
@@ -143,10 +143,32 @@ class AutoApprove(PolicyPackageAlgorithms):
 
 
 @transactional_session
-def add_rule(dids: Sequence[dict[str, Any]], account: InternalAccount, copies: int, rse_expression: str, grouping: Literal['ALL', 'DATASET', 'NONE'], weight: str, lifetime: int, locked: bool, subscription_id: str,
-             source_replica_expression: Optional[str] = None, activity: str = 'User Subscriptions', notify: Optional[Literal['Y', 'N', 'C']] = None, purge_replicas: bool = False,
-             ignore_availability: bool = False, comment: Optional[str] = None, ask_approval: bool = False, asynchronous: bool = False, ignore_account_limit: bool = False,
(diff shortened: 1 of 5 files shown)


lib/rucio/core/rule.py
```diff
@@ -1856,7 +1865,7 @@ def get_injected_rules(total_workers: int,
                        blocked_rules: Sequence[str] = [],
                        *,
                        session: "Session"
-                       ) -> list[tuple[str, str]]:
+                       ) -> list[str]:
```
dchristidis: I’m afraid I might have not made myself clear in https://github.com/rucio/rucio/pull/6516#discussion_r1523231979.

The current return type *is* `list[tuple[str]]`.

You can see it from:

https://github.com/rucio/rucio/blob/ec84fe0f2a9aefc429448778ba5d4f63a6cbe94b/lib/rucio/core/rule.py#L1844

`rule[0]` is just the rule ID, then `rule` is a single-element tuple. The Injector handles it as:

https://github.com/rucio/rucio/blob/ec84fe0f2a9aefc429448778ba5d4f63a6cbe94b/lib/rucio/daemons/judge/injector.py#L78-L90

But there is no benefit to doing this! So my request was to take the opportunity to 1) *change* `get_injected_rules()` to return `list[str]` and then 2) *adjust* the Injector to handle it. Same for `get_stuck_rules()` and the Repairer.

So, unfortunately, the correction to the hints in this PR is wrong again.

dchristidis (dismissed):
Under different circumstances I would have requested further changes. However, given the urgency to prepare the next release candidate, I believe it’s fine to merge it in its current state.

lib/rucio/core/rule.py
rdimaio: Addressed in latest revision

rdimaio:
Re-formatted to follow style outlined here https://github.com/rucio/rucio/pull/6516#discussion_r1523140076

dchristidis (approved):

Issue #6529, opened 2024-03-05
Update tests to use test fixtures


rdimaio:
`pytest` fixtures are defined here: https://github.com/rucio/rucio/blob/master/tests/conftest.py

Some tests have not been updated to use these fixtures yet - for consistency, it would be a good idea to update all outdated tests to use fixtures

PR #6561, opened 2024-03-13
Tests: Update test_rucio_bin to use fixtures, #6529

voetberg:
Near re-write of `test_rucio_bin` to change the hard-coding to fixtures and temp_factories. Extension of #6531 

Tried to leave functionality unchanged.


bari12:
I think this is a good idea. @rdimaio can you please have a look as well.

rdimaio:
> I think this is a good idea. @rdimaio can you please have a look as well.

It's marked as draft - @voetberg is it ready for review? Feel free to mark it as ready in that case

voetberg:
@rdimaio That's right it is a draft - I was having some issues getting everything to pass again so put this on the back-burner.

voetberg:
Need to solve the mystery of why the multivo non-auto oidc auth tests are failing, but I do not expect that to change the PR contents dramatically

Issue #6383, opened 2023-11-20
VO-specific tests for DUNE


jamesp-epcc:
### Description

We would like to add experiment-specific CI tests for DUNE, similar to the ones that already exist for ATLAS and BelleII. DUNE is in some ways more challenging; while ATLAS and BelleII both use generic container images, DUNE tests will require a MetaCat server and the DUNE policy package, so we will need a custom container image.

To keep things similar to the other tests, it will probably be easiest to have everything in a single container image, which should be feasible, but will require some work to create this image. The alternative would be to use a `docker-compose` type solution involving multiple containers, however this would be a departure from how the other tests work and would introduce complexity to the test workflow.

bari12:
@jamesp-epcc This is beyond the 2 year horizon, I would propose we close this. Any objections?

jamesp-epcc:
@bari12 yes, this should definitely be closed. I've actually implemented this now in the DUNE policy package repo since it's more appropriate to have DUNE-specific code in there. I'd forgotten there was still an open issue for it in Rucio itself.

PR #6563, opened 2024-03-14
Testing: Add DUNE VO tests #6383

jamesp-epcc:
This is a first attempt at adding some VO tests for DUNE. It tests all of the DUNE-specific functionality in our policy package and performs a complete replication workflow from deterministic to non-deterministic storage. Because we need an environment with multiple containers so that we can include a fake MetaCat server, this is modeled more on the integration tests than on the other VO tests.

I am open to suggestions on how to integrate this better with the other VO tests. For now I have added a separate `dune-tests` job to the top level workflow, because creating a full docker-compose environment doesn't fit well with the existing ATLAS and BelleII tests that just use a single container. However there might be a better way, especially if other VOs want to do something similar to DUNE in the future.

These tests rely on the MetaCat container, which is on my Docker account, and the DUNE policy package, which is on my GitHub account. But it might be possible to move these to more official homes, such as the official DUNE GitHub account, if you would prefer not to have the tests rely on code from personal accounts.

### DIFF
--- a/.github/workflows/vo_tests.yml
+++ b/.github/workflows/vo_tests.yml
@@ -20,9 +20,13 @@ jobs:
       - name: Identify Matrix
         id: matrix
         run: echo "matrix=$(./tools/test/votest_helper.py)" >> $GITHUB_OUTPUT
+      - name: Identify DUNE Matrix
+        id: dune-matrix
+        run: echo "::set-output name=matrix::$(./tools/test/matrix_parser.py < ./etc/docker/test/matrix_dune_tests.yml)"
     outputs:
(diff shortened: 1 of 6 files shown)


maany:
I'm working on https://github.com/rucio/rucio/issues/6667 . Once merged, James can rebase this PR and adapt the DUNE VOTests to use the framework.

bari12:
@maany is there an update on this?

maany:
@bari12 this fell off the radar. I can get something working for @jamesp-epcc by Monday

bari12:
I am closing this for now, with the change in the CI infrastructure, it will anyway need a substantial rewrite.

PR #6567, opened 2024-03-14
Transfers: metrics API endpoint, allow to group RSEs by attribute in response

### LINKED ISSUE #6564: Metrics API endpoint: allow option to return RSEs identified by RSE attribute
### Description

Related: https://gitlab.cern.ch/atlas-adc-ddm/flux-rucio/-/issues/10

Example calls:
- `requests/metrics?group=site`
- `requests/metrics?group=tier`

### Motivation

_No response_

### Change

_No response_

rdimaio:
Usage: `requests/metrics?group_response_by=[attr]`

Note: I didn't test this as I am not yet sure how to test API requests. I would appreciate some guidance on this if possible.

Fix #6564

### DIFF
--- a/lib/rucio/core/request.py
+++ b/lib/rucio/core/request.py
@@ -114,7 +114,6 @@ class RequestWithSources:
             transfertool: str,
             requested_at: Optional[datetime.datetime] = None,
     ):
-
         self.request_id = id_
         self.request_type = request_type
         self.rule_id = rule_id
@@ -1856,6 +1855,7 @@ def get_request_metrics(
         dest_rse_id: "Optional[str]" = None,
         src_rse_id: "Optional[str]" = None,
         activity: "Optional[str]" = None,
+        group_by_rse_attribute: "Optional[str]" = None,
         *,
         session: "Session"
 ):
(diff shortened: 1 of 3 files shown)


dchristidis (changes requested):
I’m concerned that the name `group_response_by` and the documentation around it can be misleading or not intuitive. I would be more precise that this is about RSE attributes.

Also, links should not be silently omitted if the RSE attribute does not exist (maybe use something like `rse.attributes.get(key, 'UNKNOWN')` instead ?)

rdimaio:
> I’m concerned that the name `group_response_by` and the documentation around it can be misleading or not intuitive. I would be more precise that this is about RSE attributes.
> 
> Also, links should not be silently omitted if the RSE attribute does not exist (maybe use something like `rse.attributes.get(key, 'UNKNOWN')` instead ?)

@dchristidis thanks for the comments. Would `group_response_by_rse_attribute` be an acceptable name or do you have other suggestions?

dchristidis:
The presence of ‘response’ feels unnecessary. Maybe `group_by_attribute` or `group_by_rse_attribute`?

rdimaio:
Please ignore the changes in `test_request.py` that I accidentally added in the first force-push; I reverted them in the second force-push

rdimaio:
Rebased in latest force-push

bari12 (approved):

PR #6578, opened 2024-03-18
Testing: sort imports and add isort-only Ruff to pre-commit and autotests

rdimaio:
Part of https://github.com/rucio/rucio/issues/6538

- b1debb9b46f8dc78ace269da882e633c9259a9e3: sorting all imports in the repository
- 49820420a5d39b6c92782cdb53ef40ca2d9edbe0: ruff configuration (only performs isort checking for now), and added ruff to pre-commit script and GH autotest

PR #6579, opened 2024-03-18
Metadata: Check on typing for client mode #6490

### LINKED ISSUE #6577: Regression from  #6490
### Description

As reported on Mattermost:

```
  File "/usr/local/bin/rucio", line 39, in <module>
  File "/usr/local/lib/python3.9/site-packages/rucio/client/__init__.py", line 16, in <module>
    from .client import Client  # noqa: F401
  File "/usr/local/lib/python3.9/site-packages/rucio/client/client.py", line 30, in <module>
    from rucio.client.metaconventionsclient import MetaConventionClient
  File "/usr/local/lib/python3.9/site-packages/rucio/client/metaconventionsclient.py", line 24, in <module>
    from rucio.db.sqla.constants import KeyType
ModuleNotFoundError: No module named 'rucio.db'
```

### Steps to reproduce

Use the RC Rucio client, with any subcommand.

### Rucio Version

40.0.0rc2

### Additional Information

_No response_

voetberg:


lib/rucio/client/metaconventionsclient.py
```diff
@@ -20,18 +20,20 @@
 
 from rucio.client.baseclient import BaseClient
 from rucio.client.baseclient import choice
-from rucio.common.utils import build_url
-from rucio.db.sqla.constants import KeyType
+from rucio.common.utils import build_url, is_client
 from typing import Union, Optional
 
+if not is_client:
+    from rucio.db.sqla.constants import KeyType
```
dchristidis: Thank you for your quick reaction. Do you wish to consider using `typing.TYPE_CHECKING` instead? This is my own personal opinion, so I am open to keeping it as it is, but `typing.TYPE_CHECKING` feels more natural because it’s a conditional import that is used exclusively for type checking (and if you’re type checking, how likely is it that you don’t have the entire source?)

voetberg: I won't protest to that suggestion :).`TYPE_CHECKING` seems to be the more common solution anyways.

lib/rucio/client/metaconventionsclient.py
```diff
@@ -21,8 +21,11 @@
 from rucio.client.baseclient import BaseClient
 from rucio.client.baseclient import choice
 from rucio.common.utils import build_url
-from rucio.db.sqla.constants import KeyType
-from typing import Union, Optional
+from typing import Union, Optional, TYPE_CHECKING
```
rdimaio, line 24: The `isort`-compliant order would be:

```suggestion
from typing import TYPE_CHECKING, Optional, Union
```

but I think we can ignore it here. I have opened a PR that also adds an `isort` check in the CI: https://github.com/rucio/rucio/pull/6578, so this will be fixed later on

rdimaio (approved):

bari12:
Merging it like this; I am just wondering in general if we should do it like this or park these types in an actual (client accessible) type module and not hide it behind `if TYPE_CHECKING`.

dchristidis:
Upon closer inspection, the use of `KeyType` (and anything in `constants.py`) was wrong. Instances of `Enum` are not JSON-serialisable, so they can’t be used. The correct hint would have been `Literal['ALL', 'COLLECTION', …]`.

So we don’t have to consider how to import them because we shouldn’t in the first place.

bari12:
Let's change this again then in a following minor release.
Though adding the Literals there also seems error-prone, since we would have to manually keep them in sync with the Enums (In case we keep them and do not change these as well). But my feeling is that we need to have stricter rules there between types used on the client side, types used in the core and types used in the database layer.

Issue #6588, opened 2024-03-21
Type annotate the code


rdimaio:
Follow-up to https://github.com/rucio/rucio/issues/6454

rdimaio:
Closing in favour of https://github.com/rucio/rucio/issues/7021

PR #6589, opened 2024-03-21
Testing: Add type annotations to topology.py

rdimaio:
Part of #6588

Regarding `SUPPORTED_PROTOCOLS_LITERAL`: If the literal is built from the list, static type checkers are not able to understand this correctly, so instead the list must be built from the literal. See here for more information: https://stackoverflow.com/a/64522240

### DIFF
--- a/lib/rucio/common/constants.py
+++ b/lib/rucio/common/constants.py
@@ -14,6 +14,7 @@
 
 import enum
 from collections import namedtuple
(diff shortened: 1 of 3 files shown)


rdimaio:
force-push done to fix circular import that I hadn't initially noticed

bari12 (approved):

PR #6591, opened 2024-03-22
Migrate core/replica.py to SQLAlchemy 2.0 syntax, #6586

### LINKED ISSUE #6586: Migrating core/replica.py to SQLAlchemy 2.0
### Description

Migrate the file replica.py in the core to use new SQLAlchemy 2.0 syntax.

### Motivation

https://docs.sqlalchemy.org/en/20/changelog/migration_20.html

### Change

Refactoring queries and other to-be deprecated SQLAlchemy 1.4 functionality to adhere to new SQLAlchemy 2.0 standards.

erlingstaff:
Fixes #6586

Part of the larger issue #6640

### DIFF
--- a/lib/rucio/core/replica.py
+++ b/lib/rucio/core/replica.py
@@ -34,7 +34,6 @@ from sqlalchemy import and_, delete, exists, func, insert, not_, or_, union, upd
 from sqlalchemy.exc import DatabaseError, IntegrityError
 from sqlalchemy.orm import aliased
 from sqlalchemy.orm.exc import FlushError, NoResultFound
-from sqlalchemy.sql import label
 from sqlalchemy.sql.expression import case, false, literal, literal_column, null, select, text, true
 
 import rucio.core.did
@@ -103,19 +102,28 @@ def get_bad_replicas_summary(rse_expression=None, from_date=None, to_date=None, 
         to_days = func.date_trunc('day', models.BadReplicas.created_at)
     else:
         to_days = func.strftime(models.BadReplicas.created_at, '%Y-%m-%d')
-    query = session.query(func.count(), to_days, models.BadReplicas.rse_id, models.BadReplicas.state, models.BadReplicas.reason)
+
+    stmt = select(
+        func.count(),
+        to_days,
+        models.BadReplicas.rse_id,
+        models.BadReplicas.state,
+        models.BadReplicas.reason
+    ).select_from(
+        models.BadReplicas
+    )
     # To be added : HINTS
     if rse_clause != []:
-        query = query.filter(or_(*rse_clause))
+        stmt = stmt.where(or_(*rse_clause))
     if from_date:
-        query = query.filter(models.BadReplicas.created_at > from_date)
+        stmt = stmt.where(models.BadReplicas.created_at > from_date)
     if to_date:
-        query = query.filter(models.BadReplicas.created_at < to_date)
-    summary = query.group_by(to_days, models.BadReplicas.rse_id, models.BadReplicas.reason, models.BadReplicas.state).all()
-    for row in summary:
-        if (row[2], row[1], row[4]) not in incidents:
-            incidents[(row[2], row[1], row[4])] = {}
-        incidents[(row[2], row[1], row[4])][str(row[3].name)] = row[0]
+        stmt = stmt.where(models.BadReplicas.created_at < to_date)
+    stmt = stmt.group_by(to_days, models.BadReplicas.rse_id, models.BadReplicas.reason, models.BadReplicas.state)
+    for count, to_days, rse_id, state, reason in session.execute(stmt):
+        if (rse_id, to_days, reason) not in incidents:
+            incidents[(rse_id, to_days, reason)] = {}
+        incidents[(rse_id, to_days, reason)][str(state.name)] = count
 
     for incident in incidents:
         res = incidents[incident]
@@ -161,27 +169,38 @@ def __exist_replicas(rse_id, replicas, *, session: "Session"):
     for clause in [path_clause, did_clause]:
         if clause:
             for chunk in chunks(clause, 10):
-                query = session.query(models.RSEFileAssociation.path,
-                                      models.RSEFileAssociation.scope,
-                                      models.RSEFileAssociation.name,
-                                      models.RSEFileAssociation.rse_id,
-                                      models.RSEFileAssociation.bytes,
-                                      func.max(case((models.BadReplicas.state == BadFilesStatus.SUSPICIOUS, 0),
-                                                    (models.BadReplicas.state == BadFilesStatus.BAD, 1),
-                                                    else_=0))).\
-                    with_hint(models.RSEFileAssociation, "INDEX(REPLICAS REPLICAS_PATH_IDX", 'oracle').\
-                    outerjoin(models.BadReplicas,
-                              and_(models.RSEFileAssociation.scope == models.BadReplicas.scope,
-                                   models.RSEFileAssociation.name == models.BadReplicas.name,
-                                   models.RSEFileAssociation.rse_id == models.BadReplicas.rse_id)).\
-                    filter(models.RSEFileAssociation.rse_id == rse_id).filter(or_(*chunk)).\
-                    group_by(models.RSEFileAssociation.path,
-                             models.RSEFileAssociation.scope,
-                             models.RSEFileAssociation.name,
-                             models.RSEFileAssociation.rse_id,
-                             models.RSEFileAssociation.bytes)
+                stmt = select(
+                    models.RSEFileAssociation.path,
+                    models.RSEFileAssociation.scope,
+                    models.RSEFileAssociation.name,
+                    models.RSEFileAssociation.rse_id,
+                    models.RSEFileAssociation.bytes,
+                    func.max(
+                        case(
+                            (models.BadReplicas.state == BadFilesStatus.SUSPICIOUS, 0),
+                            (models.BadReplicas.state == BadFilesStatus.BAD, 1),
+                            else_=0))
+                ).with_hint(
+                    models.RSEFileAssociation,
+                    'INDEX(REPLICAS REPLICAS_PATH_IDX',
+                    'oracle'
+                ).outerjoin(
+                    models.BadReplicas,
+                    and_(models.RSEFileAssociation.scope == models.BadReplicas.scope,
+                         models.RSEFileAssociation.name == models.BadReplicas.name,
+                         models.RSEFileAssociation.rse_id == models.BadReplicas.rse_id)
+                ).where(
+                    and_(models.RSEFileAssociation.rse_id == rse_id,
+                         or_(*chunk))
+                ).group_by(
+                    models.RSEFileAssociation.path,
+                    models.RSEFileAssociation.scope,
+                    models.RSEFileAssociation.name,
+                    models.RSEFileAssociation.rse_id,
+                    models.RSEFileAssociation.bytes
+                )
 
-                for path, scope, name, rse_id, size, state in query.all():
+                for path, scope, name, rse_id, size, state in session.execute(stmt).all():
                     if (scope, name, path) in replicas:
                         replicas.remove((scope, name, path))
                     if (None, None, path) in replicas:
@@ -212,19 +231,26 @@ def list_bad_replicas_status(state=BadFilesStatus.BAD, rse_id=None, younger_than
     :param session: The database session in use.
     """
     result = []
-    query = session.query(models.BadReplicas.scope, models.BadReplicas.name, models.BadReplicas.rse_id, models.BadReplicas.state, models.BadReplicas.created_at, models.BadReplicas.updated_at)
+    stmt = select(
+        models.BadReplicas.scope,
+        models.BadReplicas.name,
+        models.BadReplicas.rse_id,
+        models.BadReplicas.state,
+        models.BadReplicas.created_at,
+        models.BadReplicas.updated_at
+    )
     if state:
-        query = query.filter(models.BadReplicas.state == state)
+        stmt = stmt.where(models.BadReplicas.state == state)
     if rse_id:
-        query = query.filter(models.BadReplicas.rse_id == rse_id)
+        stmt = stmt.where(models.BadReplicas.rse_id == rse_id)
     if younger_than:
-        query = query.filter(models.BadReplicas.created_at >= younger_than)
+        stmt = stmt.where(models.BadReplicas.created_at >= younger_than)
     if older_than:
-        query = query.filter(models.BadReplicas.created_at <= older_than)
+        stmt = stmt.where(models.BadReplicas.created_at <= older_than)
     if limit:
-        query = query.limit(limit)
+        stmt = stmt.limit(limit)
 
-    for badfile in query.yield_per(1000):
+    for badfile in session.execute(stmt).yield_per(1000):
         if badfile.scope.vo == vo:
             if list_pfns:
                 result.append({'scope': badfile.scope, 'name': badfile.name, 'type': DIDType.FILE})
@@ -382,8 +408,16 @@ def add_bad_dids(dids, rse_id, reason, issuer, state=BadFilesStatus.BAD, *, sess
             new_bad_replica = models.BadReplicas(scope=scope, name=name, rse_id=rse_id, reason=reason, state=state,
                                                  account=issuer, bytes=size)
             new_bad_replica.save(session=session, flush=False)
-            session.query(models.Source).filter_by(scope=scope, name=name,
-                                                   rse_id=rse_id).delete(synchronize_session=False)
+            stmt = delete(
+                models.Source
+            ).where(
+                and_(models.Source.scope == scope,
+                     models.Source.name == name,
+                     models.Source.rse_id == rse_id)
+            ).execution_options(
+                synchronize_session=False
+            )
+            session.execute(stmt)
         else:
             if already_declared:
                 unknown_replicas.append('%s:%s %s' % (did['scope'], name, 'Already declared'))
@@ -468,18 +502,25 @@ def get_pfn_to_rse(pfns, vo='def', *, session: "Session"):
         if storage_element not in storage_elements:
             storage_elements.append(storage_element)
             se_condition.append(models.RSEProtocols.hostname == storage_element)
-    query = session.query(models.RSEProtocols.rse_id,
-                          models.RSEProtocols.scheme,
-                          models.RSEProtocols.hostname,
-                          models.RSEProtocols.port,
-                          models.RSEProtocols.prefix).\
-        join(models.RSE, models.RSE.id == models.RSEProtocols.rse_id).\
-        filter(and_(or_(*se_condition), models.RSEProtocols.scheme == scheme)).\
-        filter(models.RSE.deleted == false()).\
-        filter(models.RSE.staging_area == false())
+    stmt = select(
+        models.RSEProtocols.rse_id,
+        models.RSEProtocols.scheme,
+        models.RSEProtocols.hostname,
+        models.RSEProtocols.port,
+        models.RSEProtocols.prefix
+    ).join(
+        models.RSE,
+        models.RSEProtocols.rse_id == models.RSE.id
+    ).where(
+        and_(or_(*se_condition),
+             models.RSEProtocols.scheme == scheme,
+             models.RSE.deleted == false(),
+             models.RSE.staging_area == false())
+    )
+
     protocols = {}
 
-    for rse_id, protocol, hostname, port, prefix in query.yield_per(10000):
+    for rse_id, protocol, hostname, port, prefix in session.execute(stmt).yield_per(10000):
         if rse_id not in protocols:
             protocols[rse_id] = []
         protocols[rse_id].append('%s://%s:%s%s' % (protocol, hostname, port, prefix))
@@ -520,18 +561,28 @@ def get_bad_replicas_backlog(*, session: "Session"):
 
     :returns: a list of dictionary {'rse_id': cnt_bad_replicas}.
     """
-    query = session.query(func.count(models.RSEFileAssociation.rse_id), models.RSEFileAssociation.rse_id). \
-        with_hint(models.RSEFileAssociation, 'INDEX(DIDS DIDS_PK) USE_NL(DIDS) INDEX_RS_ASC(REPLICAS ("REPLICAS"."STATE"))', 'oracle'). \
-        filter(models.RSEFileAssociation.state == ReplicaState.BAD)
-
-    query = query.join(models.DataIdentifier,
-                       and_(models.DataIdentifier.scope == models.RSEFileAssociation.scope,
-                            models.DataIdentifier.name == models.RSEFileAssociation.name)).\
-        filter(models.DataIdentifier.availability != DIDAvailability.LOST).\
-        group_by(models.RSEFileAssociation.rse_id)
+    stmt = select(
+        func.count(),
+        models.RSEFileAssociation.rse_id
+    ).select_from(
+        models.RSEFileAssociation
+    ).with_hint(
+        models.RSEFileAssociation,
+        'INDEX(DIDS DIDS_PK) USE_NL(DIDS) INDEX_RS_ASC(REPLICAS ("REPLICAS"."STATE"))',
+        'oracle'
+    ).join(
+        models.DataIdentifier,
+        and_(models.RSEFileAssociation.scope == models.DataIdentifier.scope,
+             models.RSEFileAssociation.name == models.DataIdentifier.name)
+    ).where(
+        and_(models.DataIdentifier.availability != DIDAvailability.LOST,
+             models.RSEFileAssociation.state == ReplicaState.BAD)
+    ).group_by(
+        models.RSEFileAssociation.rse_id
+    )
 
     result = dict()
-    for cnt, rse_id in query.all():
+    for cnt, rse_id in session.execute(stmt).all():
         result[rse_id] = cnt
     return result
 
@@ -549,27 +600,36 @@ def list_bad_replicas(limit=10000, thread=None, total_threads=None, rses=None, *
     :returns: a list of dictionary {'scope' scope, 'name': name, 'rse_id': rse_id, 'rse': rse}.
     """
     schema_dot = '%s.' % DEFAULT_SCHEMA_NAME if DEFAULT_SCHEMA_NAME else ''
-    query = session.query(models.RSEFileAssociation.scope,
-                          models.RSEFileAssociation.name,
-                          models.RSEFileAssociation.rse_id). \
-        with_hint(models.RSEFileAssociation, 'INDEX(DIDS DIDS_PK) USE_NL(DIDS) INDEX_RS_ASC(REPLICAS ("REPLICAS"."STATE"))', 'oracle'). \
-        filter(models.RSEFileAssociation.state == ReplicaState.BAD)
 
-    query = filter_thread_work(session=session, query=query, total_threads=total_threads, thread_id=thread, hash_variable='%sreplicas.name' % (schema_dot))
-    query = query.join(models.DataIdentifier,
-                       and_(models.DataIdentifier.scope == models.RSEFileAssociation.scope,
-                            models.DataIdentifier.name == models.RSEFileAssociation.name)).\
-        filter(models.DataIdentifier.availability != DIDAvailability.LOST)
+    stmt = select(
+        models.RSEFileAssociation.scope,
+        models.RSEFileAssociation.name,
+        models.RSEFileAssociation.rse_id
+    ).with_hint(
+        models.RSEFileAssociation,
+        'INDEX(DIDS DIDS_PK) USE_NL(DIDS) INDEX_RS_ASC(REPLICAS ("REPLICAS"."STATE"))',
+        'oracle'
+    ).where(
+        models.RSEFileAssociation.state == ReplicaState.BAD
+    )
+
+    stmt = filter_thread_work(session=session, query=stmt, total_threads=total_threads, thread_id=thread, hash_variable='%sreplicas.name' % (schema_dot))
+
+    stmt = stmt.join(
+        models.DataIdentifier,
+        and_(models.RSEFileAssociation.scope == models.DataIdentifier.scope,
+             models.RSEFileAssociation.name == models.DataIdentifier.name)
+    ).where(
+        models.DataIdentifier.availability != DIDAvailability.LOST
+    )
 
     if rses:
-        rse_clause = list()
-        for rse in rses:
-            rse_clause.append(models.RSEFileAssociation.rse_id == rse['id'])
-        query = query.filter(or_(*rse_clause))
+        rse_clause = [models.RSEFileAssociation.rse_id == rse['id'] for rse in rses]
+        stmt = stmt.where(or_(*rse_clause))
 
-    query = query.limit(limit)
+    stmt = stmt.limit(limit)
     rows = []
-    for scope, name, rse_id in query.yield_per(1000):
+    for scope, name, rse_id in session.execute(stmt).yield_per(1000):
         rows.append({'scope': scope, 'name': name, 'rse_id': rse_id, 'rse': get_rse_name(rse_id=rse_id, session=session)})
     return rows
 
@@ -613,8 +673,16 @@ def get_did_from_pfns(pfns, rse_id=None, vo='def', *, session: "Session"):
             for pfn in parsed_pfn:
                 path = '%s%s' % (parsed_pfn[pfn]['path'], parsed_pfn[pfn]['name'])
                 pfndict[path] = pfn
-                condition.append(and_(models.RSEFileAssociation.path == path, models.RSEFileAssociation.rse_id == rse_id))
-            for scope, name, pfn in session.query(models.RSEFileAssociation.scope, models.RSEFileAssociation.name, models.RSEFileAssociation.path).filter(or_(*condition)):
+                condition.append(and_(models.RSEFileAssociation.path == path,
+                                      models.RSEFileAssociation.rse_id == rse_id))
+            stmt = select(
+                models.RSEFileAssociation.scope,
+                models.RSEFileAssociation.name,
+                models.RSEFileAssociation.path
+            ).where(
+                or_(*condition)
+            )
+            for scope, name, pfn in session.execute(stmt).all():
                 yield {pfndict[pfn]: {'scope': scope, 'name': name}}
 
 
@@ -659,15 +727,21 @@ def _list_files_wo_replicas(files_wo_replica, *, session: "Session"):
         for file in sorted(files_wo_replica, key=lambda f: (f['scope'], f['name'])):
             file_wo_clause.append(and_(models.DataIdentifier.scope == file['scope'],
                                        models.DataIdentifier.name == file['name']))
-        files_wo_replicas_query = session.query(models.DataIdentifier.scope,
-                                                models.DataIdentifier.name,
-                                                models.DataIdentifier.bytes,
-                                                models.DataIdentifier.md5,
-                                                models.DataIdentifier.adler32).\
-            filter_by(did_type=DIDType.FILE).filter(or_(*file_wo_clause)).\
-            with_hint(models.DataIdentifier, text="INDEX(DIDS DIDS_PK)", dialect_name='oracle')
-
-        for scope, name, bytes_, md5, adler32 in files_wo_replicas_query:
+        stmt = select(
+            models.DataIdentifier.scope,
+            models.DataIdentifier.name,
+            models.DataIdentifier.bytes,
+            models.DataIdentifier.md5,
+            models.DataIdentifier.adler32
+        ).with_hint(
+            models.DataIdentifier,
+            'INDEX(DIDS DIDS_PK)',
+            'oracle'
+        ).where(
+            and_(models.DataIdentifier.did_type == DIDType.FILE,
+                 or_(*file_wo_clause))
+        )
+        for scope, name, bytes_, md5, adler32 in session.execute(stmt):
             yield scope, name, bytes_, md5, adler32
 
 
@@ -1073,7 +1147,7 @@ def list_replicas(
             models.RSE.volatile,
         ).join(
             models.RSE,
-            and_(models.RSE.id == models.RSEFileAssociation.rse_id,
+            and_(models.RSEFileAssociation.rse_id == models.RSE.id,
                  models.RSE.deleted == false())
         )
 
@@ -1092,7 +1166,11 @@ def list_replicas(
                 stmt = stmt.where(models.RSE.id.in_([rse['id'] for rse in rses]))
             else:
                 rses_temp_table = temp_table_mngr(session).create_id_table()
-                session.execute(insert(rses_temp_table), [{'id': rse['id']} for rse in rses])
+                values = [{'id': rse['id']} for rse in rses]
+                insert_stmt = insert(
+                    rses_temp_table
+                )
+                session.execute(insert_stmt, values)
                 stmt = stmt.join(rses_temp_table, models.RSE.id == rses_temp_table.id)
 
         if not all_states:
@@ -1114,16 +1192,16 @@ def list_replicas(
         created temporary table.
         """
         resolved_files_temp_table = temp_table_mngr(session).create_scope_name_table()
+        selectable = rucio.core.did.list_child_dids_stmt(temp_table, did_type=DIDType.FILE)
 
         stmt = insert(
             resolved_files_temp_table
         ).from_select(
             ['scope', 'name'],
-            rucio.core.did.list_child_dids_stmt(temp_table, did_type=DIDType.FILE)
+            selectable
         )
-        result = session.execute(stmt)
 
-        return result.rowcount, resolved_files_temp_table
+        return session.execute(stmt).rowcount, resolved_files_temp_table
 
     def _list_replicas_for_collection_files_stmt(temp_table, replicas_subquery):
         """
@@ -1256,7 +1334,11 @@ def list_replicas(
         return
 
     input_dids_temp_table = temp_table_mngr(session).create_scope_name_table()
-    session.execute(insert(input_dids_temp_table), [{'scope': s, 'name': n} for s, n in dids])
+    values = [{'scope': scope, 'name': name} for scope, name in dids]
+    stmt = insert(
+        input_dids_temp_table
+    )
+    session.execute(stmt, values)
 
     num_files, num_collections, num_constituents = _inspect_dids(input_dids_temp_table, session=session)
 
@@ -1320,7 +1402,13 @@ def list_replicas(
         # Reuse input temp table. We don't need its content anymore
         random_dids_temp_table = input_dids_temp_table
         session.execute(delete(random_dids_temp_table))
-        session.execute(insert(random_dids_temp_table).from_select(['scope', 'name'], stmt))
+        stmt = insert(
+            random_dids_temp_table
+        ).from_select(
+            ['scope', 'name'],
+            stmt
+        )
+        session.execute(stmt)
 
         # Fetch all replicas for randomly selected dids and apply filters on python side
         stmt = _list_replicas_for_collection_files_stmt(random_dids_temp_table, replicas_subquery)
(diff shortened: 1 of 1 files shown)


lib/rucio/core/replica.py
```diff
@@ -101,16 +100,25 @@ def get_bad_replicas_summary(rse_expression=None, from_date=None, to_date=None,
...
     if from_date:
-        query = query.filter(models.BadReplicas.created_at > from_date)
+        stmt = stmt.where(models.BadReplicas.created_at > from_date)
     if to_date:
-        query = query.filter(models.BadReplicas.created_at < to_date)
-    summary = query.group_by(to_days, models.BadReplicas.rse_id, models.BadReplicas.reason, models.BadReplicas.state).all()
-    for row in summary:
+        stmt = stmt.where(models.BadReplicas.created_at < to_date)
+    stmt = stmt.group_by(to_days, models.BadReplicas.rse_id, models.BadReplicas.reason, models.BadReplicas.state)
+    for row in session.execute(stmt):
```
voetberg: Would be more readable to use ` for count, to_days, rse_id, reason in query:` instead of indexing the row.

lib/rucio/core/replica.py
```diff
@@ -159,27 +167,34 @@ def __exist_replicas(rse_id, replicas, *, session: "Session"):
...
-                             models.RSEFileAssociation.bytes)
-
-                for path, scope, name, rse_id, size, state in query.all():
+                stmt = select(
+                    models.RSEFileAssociation.path,
+                    models.RSEFileAssociation.scope,
+                    models.RSEFileAssociation.name,
+                    models.RSEFileAssociation.rse_id,
+                    models.RSEFileAssociation.bytes,
+                    func.max(case((models.BadReplicas.state == BadFilesStatus.SUSPICIOUS, 0),
```
voetberg: Should this case be broken onto separate lines? This line is awfully long and hard to parse.

lib/rucio/core/replica.py
```diff
@@ -1128,9 +1191,8 @@ def _resolve_collection_files(temp_table, *, session: "Session"):
             ['scope', 'name'],
             rucio.core.did.list_child_dids_stmt(temp_table, did_type=DIDType.FILE)
```
voetberg: This operation should be done outside the statement I think, this statement is rough on the eyes.

lib/rucio/core/replica.py
```diff
@@ -1431,12 +1493,18 @@ def __bulk_add_file_dids(files, account, dataset_meta=None, *, session: "Session
...
+        models.DataIdentifier.name,
+        models.DataIdentifier.bytes,
+        models.DataIdentifier.md5,
+        models.DataIdentifier.adler32,
+    ).with_hint(
+        models.DataIdentifier, 'INDEX(DIDS DIDS_PK)', 'oracle'
+    ).where(
+        or_(*condition)
+    )
+    available_files = [dict([(column, getattr(row, column)) for column in row._fields]) for row in session.execute(stmt)]
```
voetberg: This is a pretty nasty definition. Can I recommend: 
```
query = session.execute(stmt).scalars()
available_files  = [
   {column: row_attr for column, row_attr in row.items()} 
    for row 
    in query
]
```

lib/rucio/core/replica.py
```diff
@@ -1485,10 +1553,17 @@ def __bulk_add_replicas(rse_id, files, account, *, session: "Session"):
...
+        models.RSEFileAssociation.scope,
+        models.RSEFileAssociation.name,
+        models.RSEFileAssociation.rse_id,
+    ).with_hint(
+        models.RSEFileAssociation, 'INDEX(REPLICAS REPLICAS_PK)', 'oracle'
+    ).where(
+        or_(*condition)
+    )
+
+    available_replicas = [dict([(column, getattr(row, column)) for column in row._fields]) for row in session.execute(stmt)]
```
voetberg: Same question as [here](https://github.com/rucio/rucio/pull/6591/files#r1549889125)

lib/rucio/core/replica.py
```diff
@@ -1789,7 +1868,8 @@ def __cleanup_after_replica_deletion(scope_name_temp_table, scope_name_temp_tabl
     while parents_to_analyze:
         did_associations_to_remove = set()
 
-        session.query(scope_name_temp_table).delete()
+        stmt = delete(scope_name_temp_table)
+        session.execute(stmt)
         session.execute(insert(scope_name_temp_table), [sn._asdict() for sn in parents_to_analyze])
```
voetberg: Break this query into statement/session.execute(statement) please

lib/rucio/core/replica.py
```diff
@@ -1870,7 +1950,8 @@ def __cleanup_after_replica_deletion(scope_name_temp_table, scope_name_temp_tabl
                                   models.DataIdentifierAssociation.name == parent_name))))
 
         if did_associations_to_remove:
-            session.query(association_temp_table).delete()
+            stmt = delete(association_temp_table)
+            session.execute(stmt)
             session.execute(insert(association_temp_table), [a._asdict() for a in did_associations_to_remove])
```
voetberg: Can we also break this into a statement/execution here?

lib/rucio/core/replica.py
```diff
@@ -1916,9 +1997,11 @@ def __cleanup_after_replica_deletion(scope_name_temp_table, scope_name_temp_tabl
 
     # Get collection replicas of collections which became empty
     if clt_replicas_to_delete:
-        session.query(scope_name_temp_table).delete()
+        stmt = delete(scope_name_temp_table)
+        session.execute(stmt)
         session.execute(insert(scope_name_temp_table), [sn._asdict() for sn in clt_replicas_to_delete])
```
voetberg: Re: https://github.com/rucio/rucio/pull/6591/files#r1549891959

lib/rucio/core/replica.py
```diff
@@ -1957,7 +2040,8 @@ def __cleanup_after_replica_deletion(scope_name_temp_table, scope_name_temp_tabl
     # Update incomplete state
     messages, dids_to_delete = [], set()
     if incomplete_dids:
-        session.query(scope_name_temp_table).delete()
+        stmt = delete(scope_name_temp_table)
+        session.execute(stmt)
         session.execute(insert(scope_name_temp_table), [sn._asdict() for sn in incomplete_dids])
```
voetberg: Re: https://github.com/rucio/rucio/pull/6591/files#r1549891959

lib/rucio/core/replica.py
```diff
@@ -1993,7 +2082,8 @@ def __cleanup_after_replica_deletion(scope_name_temp_table, scope_name_temp_tabl
     # Remove Archive Constituents
     constituent_associations_to_delete = set()
     if affected_archives:
-        session.query(scope_name_temp_table).delete()
+        stmt = delete(scope_name_temp_table)
+        session.execute(stmt)
         session.execute(insert(scope_name_temp_table), [sn._asdict() for sn in affected_archives])
```
voetberg: https://github.com/rucio/rucio/pull/6591/files#r1549891959

lib/rucio/core/replica.py
```diff
@@ -2037,7 +2127,8 @@ def __cleanup_after_replica_deletion(scope_name_temp_table, scope_name_temp_tabl
             ).save(session=session, flush=False)
 
     if constituent_associations_to_delete:
-        session.query(association_temp_table).delete()
+        stmt = delete(association_temp_table)
+        session.execute(stmt)
         session.execute(insert(association_temp_table), [a._asdict() for a in constituent_associations_to_delete])
```
voetberg: https://github.com/rucio/rucio/pull/6591/files#r1549891959

lib/rucio/core/replica.py
```diff
@@ -2060,7 +2151,8 @@ def __cleanup_after_replica_deletion(scope_name_temp_table, scope_name_temp_tabl
                                              rse_id=rse_id, files=[sn._asdict() for sn in chunk], session=session)
 
     if dids_to_delete:
-        session.query(scope_name_temp_table).delete()
+        stmt = delete(scope_name_temp_table)
+        session.execute(stmt)
         session.execute(insert(scope_name_temp_table), [sn._asdict() for sn in dids_to_delete])
```
voetberg: https://github.com/rucio/rucio/pull/6591/files#r1549891959

lib/rucio/core/replica.py
```diff
@@ -2119,9 +2211,11 @@ def __cleanup_after_replica_deletion(scope_name_temp_table, scope_name_temp_tabl
         to_update = clt_to_set_not_archive.pop(0)
         if not to_update:
             continue
-        session.query(scope_name_temp_table).delete()
+        stmt = delete(scope_name_temp_table)
+        session.execute(stmt)
         session.execute(insert(scope_name_temp_table), [sn._asdict() for sn in to_update])
```
voetberg: I'm going to stop tagging all of these because I'm sure you find it annoying. But yeah. Please break into statement/query

lib/rucio/core/replica.py
```diff
@@ -2253,15 +2354,15 @@ def list_and_mark_unlocked_replicas(limit, bytes_=None, rse_id=None, delay_secon
                  models.RSEFileAssociation.name == temp_table_cls.name,
                  models.RSEFileAssociation.rse_id == rse_id)
         ).with_hint(
-            replicas_alias, "index(%(name)s REPLICAS_PK)", 'oracle'
+            replicas_alias, 'INDEX(%(name)s REPLICAS_PK)', 'oracle'
```
voetberg: Unsure the recommended style for this, but should the args of with_hint be on different lines?

lib/rucio/core/replica.py
```diff
@@ -2348,31 +2457,64 @@ def update_replicas_states(replicas, nowait=False, *, session: "Session"):
...
-            query = query.filter_by(lock_cnt=0)
+            stmt = stmt.where(
+                models.RSEFileAssociation.lock_cnt == 0
+            )
             # Exclude replicas use as sources
-            stmt = exists(1).where(and_(models.RSEFileAssociation.scope == models.Source.scope,
-                                        models.RSEFileAssociation.name == models.Source.name,
-                                        models.RSEFileAssociation.rse_id == models.Source.rse_id))
-            query = query.filter(not_(stmt))
+            sub_stmt = select(1).where(and_(models.RSEFileAssociation.scope == models.Source.scope,
```
voetberg: Can we stack these to match the other select statement formats?

lib/rucio/core/replica.py
```diff
@@ -2392,32 +2534,64 @@ def touch_replica(replica, *, session: "Session"):
...
+            models.RSEFileAssociation
+        ).where(
+            and_(models.RSEFileAssociation.rse_id == replica['rse_id'],
+                 models.RSEFileAssociation.scope == replica['scope'],
+                 models.RSEFileAssociation.name == replica['name'])
+        ).prefix_with(
+            "/*+ INDEX(REPLICAS REPLICAS_PK) */", dialect='oracle'
+        ).values({
+            models.RSEFileAssociation.accessed_at: accessed_at,
+            models.RSEFileAssociation.tombstone: case((models.RSEFileAssociation.tombstone.not_in([OBSOLETE,
```
voetberg: Should this `case` be broke into different lines? It's unwieldy as is.

lib/rucio/core/replica.py
```diff
@@ -2478,15 +2658,21 @@ def get_source_replicas(scope, name, source_rses=None, *, session: "Session"):
...
     if source_rses:
         if len(source_rses) < 10:
             rse_clause = []
             for rse_id in source_rses:
                 rse_clause.append(models.RSEFileAssociation.rse_id == rse_id)
             if rse_clause:
-                query = query.filter(or_(*rse_clause))
-    return [a[0] for a in query.all()]
+                stmt = stmt.where(or_(*rse_clause))
+    return session.execute(stmt).scalars().all()
```
voetberg, line 2766: Docs say `.all()` returns a sequence object, not a list. Does this mess with typing anywhere?

lib/rucio/core/replica.py
```diff
@@ -2508,115 +2694,142 @@ def get_and_lock_file_replicas_for_dataset(scope, name, nowait=False, restrict_r
...
-                   models.DataIdentifierAssociation.name == name)
+        stmt = select(
+            models.DataIdentifierAssociation.child_scope,
+            models.DataIdentifierAssociation.child_name,
+            models.DataIdentifierAssociation.bytes,
+            models.DataIdentifierAssociation.md5,
+            models.DataIdentifierAssociation.adler32,
+        ).with_hint(
+            models.DataIdentifierAssociation,
+            'INDEX_RS_ASC(CONTENTS CONTENTS_PK) NO_INDEX_FFS(CONTENTS CONTENTS_PK)', 'oracle'
```
voetberg: Should `oracle` be on a different line?

lib/rucio/core/replica.py
```diff
@@ -2508,115 +2694,142 @@ def get_and_lock_file_replicas_for_dataset(scope, name, nowait=False, restrict_r
...
+                    ).where(
+                        and_(models.DataIdentifierAssociation.child_scope == models.RSEFileAssociation.scope,
+                             models.DataIdentifierAssociation.child_name == models.RSEFileAssociation.name,
+                             models.RSEFileAssociation.state != ReplicaState.BEING_DELETED,
+                             models.DataIdentifierAssociation.scope == scope,
+                             models.DataIdentifierAssociation.name == name,
+                             or_(*rse_clause))
+                    )
 
     else:
```
voetberg, line 2827: Would it make more sense to make a base query and just change the filters with the if/else? None of the base queries the DataIdentifierAssociation Table seem to be that different between any of the parts of the control loop in this function.

lib/rucio/core/replica.py
```diff
@@ -2745,138 +2991,153 @@ def list_dataset_replicas(scope, name, deep=False, *, session: "Session"):
...
+            models.DataIdentifierAssociation,
+            'INDEX_RS_ASC(CONTENTS CONTENTS_PK) NO_INDEX_FFS(CONTENTS CONTENTS_PK)', 'oracle'
+        ).where(
+            and_(models.DataIdentifierAssociation.scope == scope,
+                 models.DataIdentifierAssociation.name == name)
+        )
 
         bytes_, length = 0, 0
-        for row in content_query:
+        for row in session.execute(stmt).all():
```
voetberg: `for bytes_, length in session.execute(stmt).scalars():`

lib/rucio/core/replica.py
```diff
@@ -2745,138 +2991,153 @@ def list_dataset_replicas(scope, name, deep=False, *, session: "Session"):
...
+            sub_query_stmt.c.accessed_at
+        ).where(
+            and_(sub_query_stmt.c.rse_id == models.RSE.id,
+                 models.RSE.deleted == false())
+        )
+
+        # Join everything together
+        final_stmt = stmt.union_all(full_query_stmt)
+        for row in session.execute(final_stmt).all():
             replica = row._asdict()
```
voetberg, line 3172: I think the more efficent version of the query.all() -> row._asdict() is to use [mappings](https://docs.sqlalchemy.org/en/20/core/connections.html#sqlalchemy.engine.Result.mappings)

lib/rucio/core/replica.py
```diff
@@ -3086,18 +3371,33 @@ def get_cleaned_updated_collection_replicas(total_workers, worker_number, limit=
             schema = BASE.metadata.schema + '.'
         session.execute(text('DELETE FROM {schema}updated_col_rep A WHERE A.rowid > ANY (SELECT B.rowid FROM {schema}updated_col_rep B WHERE A.scope = B.scope AND A.name=B.name AND A.did_type=B.did_type AND (A.rse_id=B.rse_id OR (A.rse_id IS NULL and B.rse_id IS NULL)))'.format(schema=schema)))  # NOQA: E501
```
voetberg, line 3403: Should this query be updated to use the data model while we're here?

lib/rucio/core/replica.py
```diff
@@ -3484,32 +3891,40 @@ def get_suspicious_files(rse_expression, available_elsewhere, filter_=None, logg
...
         available_replica = exists(select(1).where(and_(replicas_alias.state == ReplicaState.AVAILABLE,
                                                         replicas_alias.scope == bad_replicas_alias.scope,
                                                         replicas_alias.name == bad_replicas_alias.name,
                                                         replicas_alias.rse_id != bad_replicas_alias.rse_id)))
-        query = query.filter(available_replica)
+        stmt = stmt.where(available_replica)
 
     # Only return replicas that are the last remaining copy
     if available_elsewhere == SuspiciousAvailability["LAST_COPY"].value:
         last_replica = ~exists(select(1).where(and_(replicas_alias.state == ReplicaState.AVAILABLE,
```
voetberg: Nitpick. (Well all of these are nitpicks.) Can the `.where(....` get stacked under the select?

lib/rucio/core/replica.py
```diff
@@ -3519,16 +3934,18 @@ def get_suspicious_files(rse_expression, available_elsewhere, filter_=None, logg
...
-    query = query.filter(not_(other_states_present))
+    stmt = stmt.where(not_(other_states_present))
 
     # finally, the results are grouped by RSE, scope, name and required to have
     # at least 'nattempts' occurrences in the result of the query per replica.
     # If nattempts_exact, then only replicas are required to have exactly
     # 'nattempts' occurences.
     if nattempts_exact:
-        query_result = query.group_by(models.RSEFileAssociation.rse_id, bad_replicas_alias.scope, bad_replicas_alias.name).having(func.count() == nattempts).all()
+        stmt = stmt.group_by(models.RSEFileAssociation.rse_id, bad_replicas_alias.scope, bad_replicas_alias.name).having(func.count() == nattempts)
```
voetberg: Groupby args on different lines please

lib/rucio/core/replica.py
```diff
@@ -3555,23 +3972,46 @@ def get_suspicious_reason(rse_id, scope, name, nattempts=0, logger=logging.log,
...
+        bad_replicas_alias.scope,
+        bad_replicas_alias.name,
+        bad_replicas_alias.reason,
+        bad_replicas_alias.rse_id
+    ).where(
+        and_(bad_replicas_alias.rse_id == rse_id,
+             bad_replicas_alias.scope == scope,
+             bad_replicas_alias.state == 'S',
+             bad_replicas_alias.name == name,
+             ~exists(select(1).where(and_(bad_replicas_alias.rse_id == rse_id,
```
voetberg: Linebreak between `.where` and `and_` please

lib/rucio/core/replica.py
```diff
@@ -3595,27 +4035,31 @@ def set_tombstone(rse_id, scope, name, tombstone=OBSOLETE, *, session: "Session"
...
         try:
-            session.query(models.RSEFileAssociation).filter_by(scope=scope, name=name, rse_id=rse_id).one()
+            stmt = select(
+                models.RSEFileAssociation.tombstone
+            ).where(
+                and_(models.RSEFileAssociation.rse_id == rse_id,
+                     models.RSEFileAssociation.name == name,
+                     models.RSEFileAssociation.scope == scope)
+            )
+            _ = session.execute(stmt).scalar_one()
```
voetberg: You don't need the _ = here, you can just do the session.execute

voetberg (dismissed):
Put in a bunch of minor feedback. Nothing in there is breaking but there's a few stylistic things that I think could be improved

lib/rucio/core/replica.py
erlingstaff: Hi, thanks for the thorough review and sorry for leaving this until now.

Changed to:
```
    for count, to_days, rse_id, state, reason in session.execute(stmt):
```

lib/rucio/core/replica.py
erlingstaff: Changed to:

```
                    models.RSEFileAssociation.bytes,
                    func.max(
                        case((models.BadReplicas.state == BadFilesStatus.SUSPICIOUS, 0),
                             (models.BadReplicas.state == BadFilesStatus.BAD, 1),
                             else_=0))
                ).with_hint(
```

I was unsure whether to indent the case or not, let me know if you think the following is more readable:

```
                    models.RSEFileAssociation.bytes,
                    func.max(case(
                        (models.BadReplicas.state == BadFilesStatus.SUSPICIOUS, 0),
                        (models.BadReplicas.state == BadFilesStatus.BAD, 1),
                        else_=0))
                ).with_hint(
```

lib/rucio/core/replica.py
erlingstaff: Changed to:

```
        selectable = rucio.core.did.list_child_dids_stmt(temp_table, did_type=DIDType.FILE)

        stmt = insert(
            resolved_files_temp_table
        ).from_select(
            ['scope', 'name'],
            selectable
        )
```

lib/rucio/core/replica.py
erlingstaff: What do you think of using the ._asdict() method here? I changed this to:

```
    available_files = [res._asdict() for res in session.execute(stmt).all()]
```

lib/rucio/core/replica.py
erlingstaff: Again using _.asdict(), changed to:
```
    available_replicas = [res._asdict() for res in session.execute(stmt).all()]
```

lib/rucio/core/replica.py
erlingstaff: Changed to:

```
            values = [a._asdict() for a in did_associations_to_remove]
            stmt = insert(association_temp_table)
            session.execute(stmt, values)
```

lib/rucio/core/replica.py
erlingstaff: Changed to:

```
        values = [sn._asdict() for sn in parents_to_analyze]
        stmt = insert(scope_name_temp_table)
        session.execute(stmt, values)
```

lib/rucio/core/replica.py
erlingstaff: Changed to:

```
        values = [sn._asdict() for sn in clt_replicas_to_delete]
        stmt = insert(scope_name_temp_table)
        session.execute(stmt, values)
```

lib/rucio/core/replica.py
erlingstaff: Changed to:

```
        values = [sn._asdict() for sn in incomplete_dids]
        stmt = insert(scope_name_temp_table)
        session.execute(stmt, values)
```

lib/rucio/core/replica.py
erlingstaff: Changed to:

```
        values = [sn._asdict() for sn in affected_archives]
        stmt = insert(scope_name_temp_table)
        session.execute(stmt, values)
```

lib/rucio/core/replica.py
erlingstaff: Changed to:

```
        values = [a._asdict() for a in constituent_associations_to_delete]
        stmt = insert(association_temp_table)
        session.execute(stmt, values)
```

lib/rucio/core/replica.py
erlingstaff: Changed to:

```
        values = [sn._asdict() for sn in dids_to_delete]
        stmt = insert(scope_name_temp_table)
        session.execute(stmt, values)
```

lib/rucio/core/replica.py
erlingstaff: Changed the rest of the queries that didn't adhere to the stmt/execute pattern to do so

lib/rucio/core/replica.py
erlingstaff: I placed the args of all `with_hint`s on different lines, like so: 

```
                 models.RSEFileAssociation.rse_id == rse_id)
        ).with_hint(
            replicas_alias,
            'INDEX(%(name)s REPLICAS_PK)',
            'oracle'
        ).outerjoin(
```

Have we decided on a different recommended style for `with_hint`s since you left this comment? If so I am unaware of it, let me know

lib/rucio/core/replica.py
erlingstaff: Changed to:

```
            stmt = stmt.where(
                and_(models.RSEFileAssociation.lock_cnt == 0,
                     not_(exists(select(1)
                                 .where(and_(models.RSEFileAssociation.scope == models.Source.scope,
                                             models.RSEFileAssociation.name == models.Source.name,
                                             models.RSEFileAssociation.rse_id == models.Source.rse_id)))))
            )
```

lib/rucio/core/replica.py
erlingstaff: Changed to:

```
            models.RSEFileAssociation.accessed_at: accessed_at,
            models.RSEFileAssociation.tombstone: case(
                (models.RSEFileAssociation.tombstone.not_in([OBSOLETE, none_value]),
                 accessed_at),
                else_=models.RSEFileAssociation.tombstone)
        }).execution_options(
```

lib/rucio/core/replica.py
erlingstaff: I don't think so? I believe we only use `__iter__` on the results of this fn (`get_and_lock_file_replicas`) `Sequence` implements this. cc: typeguy @rdimaio

lib/rucio/core/replica.py
erlingstaff: (Same as: https://github.com/rucio/rucio/pull/6591#discussion_r1668793992)

Changed to:
```
        ).with_hint(
            models.DataIdentifierAssociation,
            'INDEX_RS_ASC(CONTENTS CONTENTS_PK) NO_INDEX_FFS(CONTENTS CONTENTS_PK)',
            'oracle'
        ).outerjoin(
```

lib/rucio/core/replica.py
erlingstaff: Good call, refactored the queries in this function (`get_and_lock_file_replicas_for_dataset`) quite a bit. Now using the following two base queries and changing them slightly in the control loop

```
    base_stmt = select(
        models.DataIdentifierAssociation.child_scope,
        models.DataIdentifierAssociation.child_name,
        models.DataIdentifierAssociation.bytes,
        models.DataIdentifierAssociation.md5,
        models.DataIdentifierAssociation.adler32,
    ).where(
        and_(models.DataIdentifierAssociation.scope == scope,
             models.DataIdentifierAssociation.name == name)
    )

    stmt = base_stmt.add_columns(
        models.RSEFileAssociation
    ).where(
        and_(models.DataIdentifierAssociation.child_scope == models.RSEFileAssociation.scope,
             models.DataIdentifierAssociation.child_name == models.RSEFileAssociation.name,
             models.RSEFileAssociation.state != ReplicaState.BEING_DELETED)
    )
```

lib/rucio/core/replica.py
erlingstaff: I don't think we need a loop here actually, what do you think about the following change?

From:
```
        bytes_, length = 0, 0
        for row in content_query:
            bytes_, length = row.bytes, row.length
 ```
 
 To:
 ```
         bytes_, length = session.execute(stmt).one()
         bytes_ = bytes_ or 0
 ```
 
 The previous query always only returned a single row, and set `bytes_` and `length` based on this. The reason I included `bytes_ = bytes_ or 0` is because `func.sum().label('bytes')` might return null, whilst `func.count().label('length') `cannot. This also means that .one() will never raise an exception as .count() always returns an integer, so if no values match the query it returns [NULL, 0]

File: lib/rucio/core/replica.py (hunk 21 above)
Comment (erlingstaff) [reply lib/rucio/core/replica.py:3172]: I am not sure we can do this, the dict-esque object `mappings` returns (RowMapping) is immutable, i.e it does not implement `__setitem__`

```
"__setitem__" method not defined on type "RowMapping”
```

File: lib/rucio/core/replica.py (hunk 22 above)
Comment (erlingstaff) [reply lib/rucio/core/replica.py:3403]: Changed this to:

```
    if session.bind.dialect.name == 'oracle':
        A = aliased(models.UpdatedCollectionReplica)
        B = aliased(models.UpdatedCollectionReplica)

        subquery = select(
            literal_column('ROWID')
        ).select_from(
            B
        ).where(
            and_(A.scope == B.scope,
                 A.name == B.name,
                 A.did_type == B.did_type,
                 or_(A.rse_id == B.rse_id,
                     and_(A.rse_id.is_(None),
                          B.rse_id.is_(None))))
        )

        stmt_del = delete(
            A
        ).where(
            literal_column('ROWID') > any_(subquery.scalar_subquery())
        ).execution_options(
            synchronize_session=False
        )
        session.execute(stmt_del)
```

Which compiles to:
```
DELETE FROM dev.updated_col_rep AS updated_col_rep_1 WHERE ROWID > ANY (SELECT ROWID
FROM dev.updated_col_rep AS updated_col_rep_2
WHERE updated_col_rep_1.scope = updated_col_rep_2.scope AND updated_col_rep_1.name = updated_col_rep_2.name AND updated_col_rep_1.did_type = updated_col_rep_2.did_type AND (updated_col_rep_1.rse_id = updated_col_rep_2.rse_id OR updated_col_rep_1.rse_id IS NULL AND updated_col_rep_2.rse_id IS NULL))
```

File: lib/rucio/core/replica.py (hunk 24 above)
Comment (erlingstaff) [reply lib/rucio/core/replica.py]: Changed to:

```
        stmt = stmt.group_by(
            models.RSEFileAssociation.rse_id,
            bad_replicas_alias.scope,
            bad_replicas_alias.name
        ).having(
            func.count() == nattempts
        )
```

File: lib/rucio/core/replica.py (hunk 25 above)
Comment (erlingstaff) [reply lib/rucio/core/replica.py]: Changed to:
```
             ~exists(select(1).where(
                 and_(bad_replicas_alias.rse_id == rse_id,
                      bad_replicas_alias.name == name,
                      bad_replicas_alias.scope == scope,
                      bad_replicas_alias.state != 'S'))))
```

File: lib/rucio/core/replica.py (hunk 26 above)
Comment (erlingstaff) [reply lib/rucio/core/replica.py]: Removed, now just
```
session.execute(stmt).scalar_one()
```

File: lib/rucio/core/replica.py (hunk 23 above)
Comment (erlingstaff) [reply lib/rucio/core/replica.py]: Sure, changed all of these queries to stack the `where`s under `select(1)`s, like so:

```
    if available_elsewhere == SuspiciousAvailability["LAST_COPY"].value:
        last_replica = ~exists(select(1)
                               .where(and_(replicas_alias.state == ReplicaState.AVAILABLE,
                                           replicas_alias.scope == bad_replicas_alias.scope,
                                           replicas_alias.name == bad_replicas_alias.name,
                                           replicas_alias.rse_id != bad_replicas_alias.rse_id)))
```

File: lib/rucio/core/replica.py (hunk 2 above)
Comment (voetberg) [reply lib/rucio/core/replica.py]: My understanding is that the preferred format is 
```
                    models.RSEFileAssociation.bytes,
                    func.max(
                        case(
                               (models.BadReplicas.state == BadFilesStatus.SUSPICIOUS, 0),
                               (models.BadReplicas.state == BadFilesStatus.BAD, 1),
                               else_=0))
                ).with_hint(
```
But either of the above are a huge improvement as well

File: lib/rucio/core/replica.py (hunk 4 above)
Comment (voetberg) [reply lib/rucio/core/replica.py]: If it passes the tests, I see no reason why not! _asdict doesn't seem to be private here according to the sqla docs. I'll resolve all my other comments you solved this way.

File: lib/rucio/core/replica.py (hunk 14 above)
Comment (voetberg) [reply lib/rucio/core/replica.py]: https://rucio.cern.ch/documentation/developer/dev_style_guide#whitespace 

I think we can assume your solution is right?

File: lib/rucio/core/replica.py (hunk 17 above)
Comment (rdimaio) [reply lib/rucio/core/replica.py:2766]: `Sequence` implements `__iter__` as a mix-in: https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes (and generally I think we would have seen errors in the CI if there was any issue)

Looking into the SQLA code, it seems like `ScalarResult.all()`  actually returns a `list` anyway:

`ScalarResult.all()` is typed as returning a `Sequence`: https://github.com/sqlalchemy/sqlalchemy/blob/8f47774de91b18e8b9a21d39662bffd239785b4e/lib/sqlalchemy/engine/result.py#L1793-L1801

But all it does is return `self._allrows`, which is inherited from `ResultInternal`, which specifically returns a `list`: https://github.com/sqlalchemy/sqlalchemy/blob/8f47774de91b18e8b9a21d39662bffd239785b4e/lib/sqlalchemy/engine/result.py#L545, and is type annotated as such

I'm not sure why they used the more generic `Sequence` type annotation for `ScalarResult.all()`. I'll open a PR to them to address this and see if it's a mistake or if there's a reason behind it

File: lib/rucio/core/replica.py (hunk 20 above)
Comment (voetberg) [reply lib/rucio/core/replica.py]: If it works it works! No problem with that

File: lib/rucio/core/replica.py (hunk 17 above)
Comment (rdimaio) [reply lib/rucio/core/replica.py:2766]: PR for reference: https://github.com/sqlalchemy/sqlalchemy/pull/11584

File: lib/rucio/core/replica.py (hunk 22 above)
Comment (voetberg) [reply lib/rucio/core/replica.py:3403]: cc: @mlassnig (Sorry if this is the wrong ping I think the comp leads page is out of date...)

File: lib/rucio/core/replica.py (hunk 17 above)
Comment (rdimaio) [reply lib/rucio/core/replica.py:2766]: There are reasons behind the `Sequence` type annotation, see PR for context. No issue on our side though - when we'll type hint `get_source_replicas` we can either annotate it to return a `Sequence` or `list` (in the latter case we'll have to use `typing.cast(list,...)` or `list(...)`

File: lib/rucio/core/replica.py (hunk 2 above)
Comment (erlingstaff) [reply lib/rucio/core/replica.py]: flake8 complained in the CI with this implementation - 
```
Running flake8                 
lib/rucio/core/replica.py:180:33: E126 continuation line over-indented for hanging indent
```

So I changed it to the following in order to satisfy flake8 and more closely follow the preferred format than the previous alternatives:
```
                    models.RSEFileAssociation.bytes,
                    func.max(
                        case(
                            (models.BadReplicas.state == BadFilesStatus.SUSPICIOUS, 0),
                            (models.BadReplicas.state == BadFilesStatus.BAD, 1),
                            else_=0))
                ).with_hint(
```

File: lib/rucio/core/replica.py (hunk 17 above)
Comment (erlingstaff) [reply lib/rucio/core/replica.py:2766]: Thanks, @rdimaio. I'll leave it as is for now then, the docstring says it returns a `list`, and regardless of what the SQLA docs say, `.all()` does actually return a `list` here

File: lib/rucio/core/replica.py (hunk 22 above)
Comment (erlingstaff) [reply lib/rucio/core/replica.py:3403]: Reverted this to use the text()-query like before after talking to @bari12, will open another PR to convert this to ORM at a later date

File: lib/rucio/core/replica.py [hunk 27]
```diff
@@ -1114,16 +1192,16 @@ def _resolve_collection_files(temp_table, *, session: "Session"):
...
             resolved_files_temp_table
         ).from_select(
             ['scope', 'name'],
-            rucio.core.did.list_child_dids_stmt(temp_table, did_type=DIDType.FILE)
+            selectable
         )
-        result = session.execute(stmt)
 
-        return result.rowcount, resolved_files_temp_table
+        return session.execute(stmt).rowcount, resolved_files_temp_table
```
rdimaio, line 1204: Since `stmt` is an `insert`, I'm a bit worried that by executing it as part of the return statement, it might cause the other variable returned (`resolved_files_temp_table`) to not reflect the changes of `insert`. Realistically I don't think this can possibly happen though, just throwing the thought out there.

lib/rucio/core/replica.py
```diff
@@ -2514,117 +2783,75 @@ def get_and_lock_file_replicas_for_dataset(scope, name, nowait=False, restrict_r
...
-    if session.bind.dialect.name == 'postgresql':
-        # Get content
-        content_query = session.query(models.DataIdentifierAssociation.child_scope,
-                                      models.DataIdentifierAssociation.child_name,
-                                      models.DataIdentifierAssociation.bytes,
-                                      models.DataIdentifierAssociation.md5,
-                                      models.DataIdentifierAssociation.adler32).\
-            with_hint(models.DataIdentifierAssociation,
-                      "INDEX_RS_ASC(CONTENTS CONTENTS_PK) NO_INDEX_FFS(CONTENTS CONTENTS_PK)",
-                      'oracle').\
```
rdimaio, line 2526: this is missing from `base_stmt` at the time when `base_stmt` is executed on line 2818. The functionality seems to be the same, but it might be a problem

lib/rucio/core/replica.py
```diff
@@ -2753,138 +3017,153 @@ def list_dataset_replicas(scope, name, deep=False, *, session: "Session"):
...
+                 models.CollectionReplica.did_type == DIDType.DATASET,
+                 models.CollectionReplica.rse_id == models.RSE.id,
+                 models.RSE.deleted == false())
+        )
+
+        for row in session.execute(stmt).all():
             yield row._asdict()
 
     else:
+        stmt = select(
```
rdimaio, line 3046: Missing original comment: `# find maximum values`

lib/rucio/core/replica.py
```diff
@@ -2753,138 +3017,153 @@ def list_dataset_replicas(scope, name, deep=False, *, session: "Session"):
...
+            'oracle'
+        ).where(
+            and_(models.DataIdentifierAssociation.scope == scope,
+                 models.DataIdentifierAssociation.name == name)
+        )
+
+        bytes_, length = session.execute(stmt).one()
+        bytes_ = bytes_ or 0
+
+        sub_query_stmt = select(
```
rdimaio, line 3064: Missing original comment: `# find archives that contain files of the requested dataset`

lib/rucio/core/replica.py
```diff
@@ -2753,138 +3017,153 @@ def list_dataset_replicas(scope, name, deep=False, *, session: "Session"):
...
+                 models.ConstituentAssociation.child_scope == models.DataIdentifierAssociation.child_scope,
+                 models.ConstituentAssociation.child_name == models.DataIdentifierAssociation.child_name,
+                 models.ConstituentAssociation.scope == models.RSEFileAssociation.scope,
+                 models.ConstituentAssociation.name == models.RSEFileAssociation.name,
+                 models.RSEFileAssociation.rse_id == models.RSE.id,
+                 models.RSEFileAssociation.state == ReplicaState.AVAILABLE,
+                 models.RSE.deleted == false())
+        ).subquery()
+
+        group_query_stmt = select(
```
rdimaio, line 3090: Missing original comment: `# count the metrics`

lib/rucio/core/replica.py
```diff
@@ -2753,138 +3017,153 @@ def list_dataset_replicas(scope, name, deep=False, *, session: "Session"):
...
+        ).group_by(
+            sub_query_stmt.c.dataset_scope,
+            sub_query_stmt.c.dataset_name,
+            sub_query_stmt.c.file_scope,
+            sub_query_stmt.c.file_name,
+            sub_query_stmt.c.rse_id,
+            sub_query_stmt.c.rse
+        ).subquery()
+
+        full_query_stmt = select(
```
rdimaio, line 3111: Missing original comment: `# bring it in the same column state as the non-archive query`

lib/rucio/core/replica.py
```diff
@@ -2753,138 +3017,153 @@ def list_dataset_replicas(scope, name, deep=False, *, session: "Session"):
...
+            func.max(group_query_stmt.c.updated_at).label('updated_at'),
+            func.max(group_query_stmt.c.accessed_at).label('accessed_at')
+        ).group_by(
+            group_query_stmt.c.dataset_scope,
+            group_query_stmt.c.dataset_name,
+            group_query_stmt.c.rse_id,
+            group_query_stmt.c.rse
+        )
+
+        sub_query_stmt = select(
```
rdimaio, line 3129: Missing original comment: `# find the non-archive dataset replicas`

lib/rucio/core/replica.py
```diff
@@ -3094,18 +3397,32 @@ def get_cleaned_updated_collection_replicas(total_workers, worker_number, limit=
             schema = BASE.metadata.schema + '.'
         session.execute(text('DELETE FROM {schema}updated_col_rep A WHERE A.rowid > ANY (SELECT B.rowid FROM {schema}updated_col_rep B WHERE A.scope = B.scope AND A.name=B.name AND A.did_type=B.did_type AND (A.rse_id=B.rse_id OR (A.rse_id IS NULL and B.rse_id IS NULL)))'.format(schema=schema)))  # NOQA: E501
     elif session.bind.dialect.name == 'mysql':
-        subquery1 = session.query(func.max(models.UpdatedCollectionReplica.id).label('max_id')).\
-            group_by(models.UpdatedCollectionReplica.scope,
-                     models.UpdatedCollectionReplica.name,
-                     models.UpdatedCollectionReplica.rse_id).subquery()
-        subquery2 = session.query(subquery1.c.max_id).subquery()
```
rdimaio, line 3101: This `subquery()` is missing from `subquery2` in the new version - not sure if it's needed

rdimaio (commented):
LGTM - there are some comments in the original code that I think should be kept to make it easier to understand those queries

lib/rucio/core/replica.py
erlingstaff: It shouldn't, line 2818 is only reached if the server is using a postgres db, and `with_hint` only adds the hint on `oracle`:
```
The dialect_name option will limit the rendering of a particular hint to a particular backend.
```

If postgres is not used then the hint is added (line 2830). Having the `with_hint` in the `base_stmt` likely has little to no overhead, I can add it there if you think that is clearer. It'd have no functional difference afaik

lib/rucio/core/replica.py
erlingstaff: Ran the test suite sequentially for both this feature branch and master and logged the output of this function, after truncating the randomized names/scopes there is no difference between the logfiles

```
shasum new_output_truncated.txt
6f41707e1041278b15bb22eb282f7c32602d99e4  new_output_truncated.txt

shasum old_output_truncated.txt
6f41707e1041278b15bb22eb282f7c32602d99e4  old_output_truncated.txt
```

[new_output_truncated.txt](https://github.com/user-attachments/files/16283388/new_output_truncated.txt)
[old_output_truncated.txt](https://github.com/user-attachments/files/16283390/old_output_truncated.txt)

Logged the outputs with:
```
    if num_collections:
        num_files_in_collections, resolved_files_temp_table = _resolve_collection_files(input_dids_temp_table, session=session)
        res_files_stmt = select(
            resolved_files_temp_table.scope.label("scope"),
            resolved_files_temp_table.name.label("name"),
        )
        res_files_result = session.execute(res_files_stmt).all()
        res_files_result_str = '\n'.join(str(r) for r in res_files_result)

        import fcntl
        file_loc = '/opt/rucio/testfile_new_2.txt'
        with open(file_loc, 'a') as f:
            fcntl.flock(f, fcntl.LOCK_EX)
            try:
                f.write(f"Number of files: {num_files_in_collections} --- Resolved files:\n{res_files_result_str}\n\n")
            finally:
                fcntl.flock(f, fcntl.LOCK_UN)
```

lib/rucio/core/replica.py
erlingstaff: Must've accidentally removed these when removing the commented out old queries, re-added them

lib/rucio/core/replica.py
erlingstaff: Added this again

lib/rucio/core/replica.py
erlingstaff: Added this again

lib/rucio/core/replica.py
erlingstaff: Added this again

lib/rucio/core/replica.py
erlingstaff: Added this again

lib/rucio/core/replica.py
erlingstaff: This subquery was used in `notin_()`, which does not want a `subquery()`:
```
Argument of type "Subquery" cannot be assigned to parameter "other" of type "Iterable[Any] | BindParameter[Any] | InElementRole" in function "notin_"
  Type "Subquery" is incompatible with type "Iterable[Any] | BindParameter[Any] | InElementRole"
    "Subquery" is incompatible with protocol "Iterable[Any]"
      "__iter__" is not present
    "Subquery" is incompatible with "BindParameter[Any]"
    "Subquery" is incompatible with "InElementRole"Pylance[reportArgumentType](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportArgumentType)
```

This is the same reason we have been seeing these warnings in the test output for a while:
```
tests/test_throttler.py::TestRequestCoreRelease::test_release_waiting_requests_grouped_fifo[file_config_mock{'conveyor': {'use_preparer': 'true'}}]
  /opt/rucio/lib/rucio/core/request.py:2210: SAWarning: Coercing Subquery object into a select() for use in IN(); please pass a select() construct explicitly
    models.Request.id.in_(cumulated_volume_subquery)  # type: ignore
```

However, when looking into this I saw that `notin_` is a legacy alias for `not_in`, so I changed this to `not_in` now (https://docs.sqlalchemy.org/en/20/core/sqlelement.html#sqlalchemy.sql.expression.ColumnElement.not_in)

lib/rucio/core/replica.py
rdimaio: thanks for checking this!

rdimaio (approved):

lib/rucio/core/replica.py
```diff
@@ -3576,23 +4032,47 @@ def get_suspicious_reason(rse_id, scope, name, nattempts=0, logger=logging.log,
...
+        bad_replicas_alias.reason,
+        bad_replicas_alias.rse_id
+    ).group_by(
+        bad_replicas_alias.rse_id,
+        bad_replicas_alias.scope,
+        bad_replicas_alias.name,
+        bad_replicas_alias.reason
+    ).having(
+        func.count() > nattempts
+    )
```
dchristidis, line 4071: There is a regression in this query. It does not make use of the existing `query` and, as such, a large portion of it is missing in the new version.

PR #6609, opened 2024-03-27
Testing: Add type annotations to config.py

rdimaio:
#6588

### DIFF
--- a/lib/rucio/common/config.py
+++ b/lib/rucio/common/config.py
@@ -18,7 +18,7 @@ import configparser
 import json
 import os
 from collections.abc import Callable
-from typing import TYPE_CHECKING, Optional, TypeVar, Union, overload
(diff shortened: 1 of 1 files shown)


rdimaio:
Regarding the Pyright failure:
```
Found 5 new problems in /home/runner/work/rucio/rucio/lib/rucio/core/credential.py
  - 5 errors with message """Object of type "None" is not subscriptable""".
    Candidates: line 152, 153, 154, 155, 199
```

I think this can be ignored - this issue existed prior to the type annotations change, and it's due to the fact that `cred` might not exist when it's accessed in those lines in `credential.py` (as `get_rse_credentials` might raise `ErrorLoadingCredentials`. `credential.py` has to check that `cred` exists first.

bari12 (approved):

PR #6635, opened 2024-04-04
Recovery: Fix typo in replica recoverer that leads to a crash #6634

### LINKED ISSUE #6634: Fix typo in suspicious replica recoverer that leads to a crash
Accidentally, the recoverer tries to declare same files as bad twice, leading to crash.

Fix: Swap a list of replicas to the correct list.

ChristophAmes:
Accidentally, the recoverer tries to declare same files as bad twice, leading to crash.

Fix: Swap a list of replicas to the correct list.


cserf (approved):
LGTM

Issue #6598, opened 2024-03-25
Possible redundant util function


rdimaio:
It seems like `render_json_list` is exactly the same as `render_json`.

https://github.com/rucio/rucio/blob/bed3f050bc6ee12dc4e6a92637a628604b265725/lib/rucio/common/utils.py#L614-L623

 Is `render_json_list` actually needed? If not, it should be removed.

Ban42:
Hi,
Since the PR I made is under review, could you assign the issue to me?
Many thanks

PR #6648, opened 2024-04-08
Bug: Patch #6598 remove redundant util function #6598

Ban42:
I have found that it was not possible to remove directly `render_json_list` from 'utils.py' as it was called in 'lib/rucio/client/didclient.py' and in 'lib/rucio/web/rest/flaskapi/v1/replicas.py'. 

Since the two functions perform the same operation with different parameter types, I just tried to find a solution that would run with parameters: a generic python list or a generic python dict. 

**N.B.** I tested it with custom lists and dicts as I do not know the content of a realistic parameter for these functions. If you want me to test them with realistic parameters, please provide me them.

### DIFF
--- a/lib/rucio/common/utils.py
+++ b/lib/rucio/common/utils.py
@@ -625,16 +625,15 @@ class APIEncoder(json.JSONEncoder):
         return json.JSONEncoder.default(self, obj)
 
 
-def render_json(**data: Any) -> str:
-    """ JSON render function
-    """
+def render_json(*args, **kwargs) -> str:
+    """ Render a list or a dict as a JSON-formatted string. """
+    if args and isinstance(args[0], list):
+        data = args[0]
+    elif kwargs:
+        data = kwargs
+    else:
+        raise ValueError("Error while serializing object to JSON-formatted string: supported input types are list or dict.")
     return json.dumps(data, cls=APIEncoder)
-
-
-def render_json_list(list_) -> str:
-    """ JSON render function for list
-    """
-    return json.dumps(list_, cls=APIEncoder)
 
 
 def datetime_parser(dct: dict[Any, Any]) -> dict[Any, Any]:
--- a/lib/rucio/client/didclient.py
+++ b/lib/rucio/client/didclient.py
@@ -21,7 +21,7 @@ from requests.status_codes import codes
 
 from rucio.client.baseclient import BaseClient, choice
 from rucio.common.exception import DeprecationError
-from rucio.common.utils import build_url, date_to_str, render_json, render_json_list
+from rucio.common.utils import build_url, date_to_str, render_json
 
 if TYPE_CHECKING:
     from collections.abc import Iterable, Iterator, Mapping, Sequence
@@ -142,7 +142,7 @@ class DIDClient(BaseClient):
         """
         path = '/'.join([self.DIDS_BASEURL])
         url = build_url(choice(self.list_hosts), path=path)
-        r = self._send_request(url, type_='POST', data=render_json_list(dids))
+        r = self._send_request(url, type_='POST', data=render_json(dids))
         if r.status_code == codes.created:
             return True
         else:
--- a/lib/rucio/web/rest/flaskapi/v1/replicas.py
+++ b/lib/rucio/web/rest/flaskapi/v1/replicas.py
@@ -23,7 +23,7 @@ from flask import Flask, Response, request
 from rucio.common.config import config_get, config_get_int
 from rucio.common.constants import SUPPORTED_PROTOCOLS
 from rucio.common.exception import AccessDenied, DataIdentifierAlreadyExists, DataIdentifierNotFound, Duplicate, InvalidObject, InvalidPath, InvalidType, ReplicaIsLocked, ReplicaNotFound, ResourceTemporaryUnavailable, RSENotFound, ScopeNotFound
-from rucio.common.utils import APIEncoder, parse_response, render_json_list
+from rucio.common.utils import APIEncoder, parse_response, render_json
 from rucio.core.replica_sorter import sort_replicas
 from rucio.db.sqla.constants import BadFilesStatus
 from rucio.gateway.quarantined_replica import quarantine_file_replicas
@@ -1066,7 +1066,7 @@ class SuspiciousReplicas(ErrorHandlingMethodView):
                 nattempts = int(params['nattempts'][0])
 
         result = get_suspicious_files(rse_expression=rse_expression, younger_than=younger_than, nattempts=nattempts, vo=request.environ.get('vo'))
-        return Response(render_json_list(result), 200, content_type='application/json')
+        return Response(render_json(result), 200, content_type='application/json')
 
 
 class BadReplicasStates(ErrorHandlingMethodView):


lib/rucio/common/utils.py
```diff
@@ -611,16 +611,13 @@ def default(self, obj):  # pylint: disable=E0202
         return json.JSONEncoder.default(self, obj)
 
 
-def render_json(**data):
-    """ JSON render function
+def render_json(list_=None, **data: Optional[Any]):
```
rdimaio: I would rewrite the function signature like this:

```suggestion
def render_json(list_: Optional[list[Any]] = None, **data) -> str:
```

`data` is a `dict`, but we generally have not added type hints for `kwargs` up until now. so I would avoid having a type annotation there for now.

rdimaio: This approach works, and I think it's clearly readable. 

The only issue I see that could arise is that `data` could have a key called `list_` which could clash with the `list_` positional argument, e.g.:
```
render_json([3,4,5], **{'a': 3, 'b': 5, 'list_': [9,8,10]})
```

which results in:

```
TypeError: render_json() got multiple values for argument 'list_'
```

In practice, this **should** never happen, as we either pass a list or pass a dict; however, if we want to be extra safe and avoid erroring out in this edge case, you could rewrite the method in this way:

```python
def render_json(*args, **kwargs) -> str:
    """ Render a list or a dict as a JSON-formatted string. """
    if args and isinstance(args[0], list):
        data = args[0]
    elif kwargs:
        data = kwargs
    else:
        raise ValueError("Error while serializing object to JSON-formatted string: supported input types are list or dict.")
    return json.dumps(data, cls=APIEncoder)
```

which supports all our current use cases and doesn't error out in that edge case:

```
print(render_json([3,4,5]))
print(render_json(**{'a': 3, 'b': 5}))
print(render_json([3,4,5], **{'a': 3, 'b': 5, 'list_': [9,8,10]}))

[3, 4, 5]
{"a": 3, "b": 5}
[3, 4, 5]
```

rdimaio (changes requested):
Thanks for the PR!

rdimaio:
Could you squash the commit 70a4d62489580f700d83926baab43e6c1f599039 into ae27bf61553acdfab4ff846ade654cdce90081d4? Thanks!

In theory, 49b5bc39d9043ea4e5cb1f27cfc99098047745f4 should (?) also be squashed into ae27bf61553acdfab4ff846ade654cdce90081d4, since ae27bf61553acdfab4ff846ade654cdce90081d4 removes `render_json_list` and would cause the build to fail since there are still existing references to that function. However, we don't really have a policy (that I know of) that states "every commit should build correctly", so I'm not sure if it would be justified to do this. @bari12 thoughts?

lib/rucio/common/utils.py
Ban42: I can perform this change, but are you sure this is the desired behavior? 
With this implementation, if I pass multiple lists or a list followed by anything, only the first list will be serialized, throwing away the rest of parameters.

```
print(render_json([3, 4, 5]))
print(render_json([3, 4, 5], [6, 7, 8]))
print(render_json([3,4,5], **{'a': 3, 'b': 5, 'list_': [9,8,10]}))

[3, 4, 5]
[3, 4, 5]
[3, 4, 5]
```

Maybe you meant:
```
def render_json(*args, **kwargs) -> str:
   ...
    if args and isinstance(args[0], list):
        data = args # <-- Look this line
    elif kwargs:
    ...
```
This solution however does not solve the problem of a list followed by a dict parameter.

Maybe we can do something like: 
```
def render_json(*list_, **dict_) -> str:
    """ Render a list or a dict as a JSON-formatted string.
    """
    serialized_result = ""
    if list_ and isinstance(list_[0], list):
        serialized_result += json.dumps(list_, cls=APIEncoder)
    
    if dict_:
        serialized_result += json.dumps(dict_, cls=APIEncoder)
    
    if (not(list_ and isinstance(list_[0], list)) and not(dict_)):
        raise ValueError("Error while serializing object to JSON-formatted string: supported input types are list or dict.")
    
    return serialized_result
```

This last case may be a little less elegant, but handles all these cases:

```
print(render_json([3, 4, 5]))
print(render_json([3, 4, 5], [6, 7, 8]))
print(render_json(**{'a': 3, 'b': 5}))
print(render_json(**{'a': 3, 'b': 5, 'list_': [9,8,10]}))
print(render_json([3,4,5], **{'a': 3, 'b': 5, 'list_': [9,8,10]}))

[[3, 4, 5]]
[[3, 4, 5], [6, 7, 8]]
{"a": 3, "b": 5}
{"a": 3, "b": 5, "list_": [9, 8, 10]}
[[3, 4, 5]]{"a": 3, "b": 5, "list_": [9, 8, 10]}
```

Ban42:
I was thinking of squashing all commits into one when merging as this PR changes few lines of code and anyway all commits together achieve a single functionality (i.e. it would not make sense to revert just one commit).
Just an opinion though, let me know which commits you would like to have and I will squash as you like.

lib/rucio/common/utils.py
Ban42: Will commit together with the changes defined below :)

rdimaio: > With this implementation, if I pass multiple lists or a list followed by anything, only the first list will be serialized, throwing away the rest of parameters.

I would say that's the intended functionality, at least for the way that we're currently using these `render_json*` methods: we either render a single list or a dict of kwargs. The change I proposed was just to avoid the `TypeError` in the edge case where someone might pass in both a list and kwargs, and one of the kwargs happens to be `list_` - this edge case goes against the intended use case and should never happen in the first place, but it's better to be safe and avoid having a runtime error here.

I get your point though, so I think it might be a good idea to have someone else give their opinion on this - I will ask someone in the team to chime in

Ban42: Sure, I was just giving my 2 cents, I am not the expert here.
Anyway, since this may take a while, would you mind assigning the related #6598 issue to me?

rdimaio: > Sure, I was just giving my 2 cents, I am not the expert here. Anyway, since this may take a while, would you mind assigning the related #6598 issue to me?

It doesn't let me do that. According to https://stackoverflow.com/a/41172554, it seems like you should comment on that issue first, and it can then be assigned to you

Ban42:
Hey,
In the end I did squash all commits into one and I adapted the code according to this suggestion:

> This approach works, and I think it's clearly readable.
> 
> The only issue I see that could arise is that `data` could have a key called `list_` which could clash with the `list_` positional argument, e.g.: 
> ...

In my opinion, since the intended functionality is to receive either a list or a string, this is solid enough.
Feel free to ask anyone's opinion, but on my side this is good to merge.

lib/rucio/common/utils.py
```diff
@@ -611,18 +611,18 @@ def default(self, obj):  # pylint: disable=E0202
         return json.JSONEncoder.default(self, obj)
 
 
-def render_json(**data):
-    """ JSON render function
+def render_json(list_: Optional[list[Any]] = None, **dict_) -> str:
+    """ JSON render function for list and dicts
```
rdimaio:
```suggestion
    """ Render a list or a dict as a JSON-formatted string.
```

I think the docstring should make it clearer that this function renders either a list or a dict, "for list and dicts" might indicate that you can pass multiple inputs and it would work.

lib/rucio/common/utils.py
```diff
@@ -611,18 +611,18 @@ def default(self, obj):  # pylint: disable=E0202
         return json.JSONEncoder.default(self, obj)
 
 
-def render_json(**data):
-    """ JSON render function
+def render_json(list_: Optional[list[Any]] = None, **dict_) -> str:
```
rdimaio: By using `list_` as a named argument, there is still the risk of a `TypeError`, as mentioned in the beginning here: https://github.com/rucio/rucio/pull/6648#discussion_r1596781366

Please rewrite the method like this:
```python
def render_json(*args, **kwargs) -> str:
    """ Render a list or a dict as a JSON-formatted string. """
    if args and isinstance(args[0], list):
        data = args[0]
    elif kwargs:
        data = kwargs
    else:
        raise ValueError("Error while serializing object to JSON-formatted string: supported input types are list or dict.")
    return json.dumps(data, cls=APIEncoder)
```

rdimaio (changes requested):

Ban42:
Hey,
This should now be all as requested, thank you for your time reviewing this :)

Ban42:
Hey,
Now the PR should be ready. About the python_ruff test, it's throwing an error and a warning on two files not touched by this PR, should I do something about it?
Cordially

rdimaio (approved):
LGTM - thank you for the help. You can ignore the `ruff` failure, that is due to a mistake on our side in an unrelated PR

Ban42:
Thanks,
I am not authorized to merge, is there any other step I should do or you will merge it for me?
Best

rdimaio:
> Thanks, I am not authorized to merge, is there any other step I should do or you will merge it for me? Best

No worries, no other step is needed on your side 😄 merging is mostly handled by @bari12 as he's also responsible for managing new releases; this will likely be merged next week

PR #6658, opened 2024-04-09
Database: Pass root_account defaults to config_get() instead of using re-definitions

rdimaio:
config_get() provides a `default` argument
that allows the user to provide a default value
in case an exception is raised when obtaining the requested information from the config.
This way, we can also simplify the logic
by removing the try/except block.

### DIFF
--- a/lib/rucio/db/sqla/util.py
+++ b/lib/rucio/db/sqla/util.py
@@ -145,33 +145,21 @@ def create_root_account():
 
     multi_vo = bool(config_get('common', 'multi_vo', False, False))
 
-    up_id = 'ddmlab'
-    up_pwd = 'secret'
-    up_email = 'ph-adp-ddm-lab@cern.ch'
-    x509_id = 'emailAddress=ph-adp-ddm-lab@cern.ch,CN=DDMLAB Client Certificate,OU=PH-ADP-CO,O=CERN,ST=Geneva,C=CH'
-    x509_email = 'ph-adp-ddm-lab@cern.ch'
-    gss_id = 'ddmlab@CERN.CH'
-    gss_email = 'ph-adp-ddm-lab@cern.ch'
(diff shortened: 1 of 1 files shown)


lib/rucio/db/sqla/util.py
```diff
@@ -145,33 +145,21 @@ def create_root_account():
...
-        # print 'Config values are missing (check rucio.cfg{.template}). Using hardcoded defaults.'
+    up_id = config_get('bootstrap', 'userpass_identity', default='ddmlab')
+    up_pwd = config_get('bootstrap', 'userpass_pwd', default='secret')
+    up_email = config_get('bootstrap', 'userpass_email', default='ph-adp-ddm-lab@cern.ch')
+    x509_id = config_get('bootstrap', 'x509_identity', default='emailAddress=ph-adp-ddm-lab@cern.ch,CN=DDMLAB Client Certificate,OU=PH-ADP-CO,O=CERN,ST=Geneva,C=CH')
+    x509_email = config_get('bootstrap', 'x509_email', default='ph-adp-ddm-lab@cern.ch')
+    gss_id = config_get('bootstrap', 'gss_identity', default='ddmlab@CERN.CH')
+    gss_email = config_get('bootstrap', 'gss_email', default='ph-adp-ddm-lab@cern.ch')
+    ssh_id = config_get('bootstrap', 'ssh_identity',
+                        default='ssh-rsa AAAAB3NzaC1yc2EAAAABIwAAAQEAq5LySllrQFpPL614sulXQ7wnIr1aGhGtl8b+HCB/'
```
voetberg, line 156: It's a little scary to see this hanging out in raw code, don't we have a token file with this in /etc?

bari12: We should probably remove all this from the core code. It is probably not needed at all.

rdimaio: We can remove these defaults, but in that case we should probably catch eventual exceptions that could be raised by `config_get` failures

bari12 (approved):

Issue #6515, opened 2024-02-27
Wrong specification for ‘Create a new replication rule’


dchristidis:
### Description

The Client API has:

https://github.com/rucio/rucio/blob/666faa57c306ac6d0db83d9eddb76044850e1b06/lib/rucio/client/ruleclient.py#L33-L35

The specification for the REST API is:

https://github.com/rucio/rucio/blob/666faa57c306ac6d0db83d9eddb76044850e1b06/lib/rucio/web/rest/flaskapi/v1/rules.py#L252-L256

Yes the Rucio CLI uses a dictionary:

https://github.com/rucio/rucio/blob/666faa57c306ac6d0db83d9eddb76044850e1b06/bin/rucio#L1171-L1175

And the internal API and core layers expect it as such:

https://github.com/rucio/rucio/blob/666faa57c306ac6d0db83d9eddb76044850e1b06/lib/rucio/api/rule.py#L93-L94

https://github.com/rucio/rucio/blob/666faa57c306ac6d0db83d9eddb76044850e1b06/lib/rucio/core/rule.py#L236-L241

### Steps to reproduce

1. Try to use `add_replication_rule` from the Client API where the `dids` parameter is a `list` of `str`s.

PR #6671, opened 2024-04-12
bug: Correct wrong specification for 'Create a new replication rule' …

GooseNight:
…#6515

This commit replaces the wrong specification for dids for 'Create a new replication rule', it was previously specified as a list of strings. It is now specified as a list of dictionaries containing two properties scope and name.

### DIFF
--- a/lib/rucio/client/ruleclient.py
+++ b/lib/rucio/client/ruleclient.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from collections.abc import Sequence
 from json import dumps, loads
 from typing import Any, Optional, Union
 from urllib.parse import quote_plus
@@ -30,7 +31,7 @@ class RuleClient(BaseClient):
 
     def add_replication_rule(
         self,
-        dids: list[str],
+        dids: Sequence[dict[str, str]],
         copies: int,
         rse_expression: str,
         priority: int = 3,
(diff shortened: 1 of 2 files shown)


lib/rucio/client/ruleclient.py
```diff
@@ -30,7 +30,7 @@ class RuleClient(BaseClient):
 
     def add_replication_rule(
         self,
-        dids: list[str],
+        dids: list[dict],
```
rdimaio: LGTM, but this would be more accurate:

```suggestion
        dids: Sequence[dict[str, str]],
```

Pyright will then likely complain about this code:

https://github.com/rucio/rucio/blob/339744866e9550f0f068f8e3c9574a95ccaa7d59/lib/rucio/api/rule.py#L92-L93

Because we're passing an `InternalScope` to something that should only be a `str`. 

To fix this, we can replace this `for d in dids` loop with this list comprehension:

```python
dids_with_internal_scope = [{'name': d['name'], 'scope': InternalScope(d['scope'], vo=vo)} for d in dids]
```

and then pass `dids=dids_with_internal_scope` to `rule.add_rule`.

(feel free to use a better name than `dids_with_internal_scope` - nothing better came to mind for me)

rdimaio: `Sequence` is not imported in `ruleclient` at the moment, so you will need to import it from `collections.abc`:

```python
from collections.abc import Sequence
```

rdimaio (approved):

PR #6679, opened 2024-04-15
Testing: Add type annotations to did_meta_plugin_interface; #6588

rdimaio:


lib/rucio/core/did_meta_plugins/did_meta_plugin_interface.py
```diff
@@ -35,7 +38,13 @@ def __init__(self):
         pass
 
     @abstractmethod
-    def get_metadata(self, scope, name, *, session: "Optional[Session]" = None):
+    def get_metadata(
+        self,
+        scope: "InternalScope",
```
voetberg, line 43: Can scope also be a string? I don't think there's anything prohibiting that.

rdimaio: From what I've seen, all the current plugins expect an `InternalScope`, as they then access attributes like `scope.internal` within the function, e.g.:

https://github.com/rucio/rucio/blob/339744866e9550f0f068f8e3c9574a95ccaa7d59/lib/rucio/core/did_meta_plugins/mongo_meta.py#L70

If we ever have a plugin that expects `scope` as a `str` then we can have another `get_metadata` method with a separate signature where `scope: str`, but for this one in particular I'd leave it `InternalScope` only to be more specific

voetberg (approved):

Issue #6677, opened 2024-04-14
Inconsistencies in Model Registration and Unregistration Functions in `models.py`


Geogouz:
I noticed couple of inconsistencies in the `models.py` related to how models are registered and unregistered.

For example:
1. **Duplicate Entries**: The model `ReplicationRule` is listed twice in the `register_models` function (even though SQLAlchemy shouldn't have a problem handling this).

https://github.com/rucio/rucio/blob/339744866e9550f0f068f8e3c9574a95ccaa7d59/lib/rucio/db/sqla/models.py#L1740-L1743

2. **Missing Model in Unregistration Function**: The `VirtualPlacements` model is included in the `register_models` function but is absent from the `unregister_models` function.

3. **Missing Models from both Functions**: The models `BadPFNs`, `TransferLimit`, `RSETransferLimit`, `DatasetLock`, `Distance`, `TransferStats` are not considered at all.

Are (any of) these intensional?

erlingstaff:
The current implementation is very confusing and not ideal. Even though there are models missing in register and unregister, as you noted, they all get created and deleted correctly (the commands in psql screenshot are run before/after running `./tools/destroy_database.py`, which calls `unregister_models`). 

<img width="800" alt="Pasted Graphic 1" src="https://github.com/rucio/rucio/assets/38101463/b39fc428-9da7-4dde-95fc-ccba2255ce2b">
 
 When unregister or register_models is run now, it loops over all the listed models and calls `model.metadata.drop_all(engine)`, where metadata is a db schema, from which all tables (models) are dropped. However, they all inherit the same metadata and schema from the `BASE` class.
 
 TL;DR:
 Currently `register_models` tries to register the same schema many times (one time for each model listed), likewise with `unregister_models`. The schema contains all the models. Simplified down to a single call `BASE.metadata.drop_all(engine)` /  `BASE.metadata.create_all(engine)` to register or unregister the schema. #6712

Issue #6628, opened 2024-04-03
Formalize docstring formatting


rdimaio:
Partly related to: https://github.com/rucio/documentation/issues/287.

It seems like there is no single style for docstrings in Rucio, e.g.:

For the docstring blocks:
- Sometimes single quotes are used: https://github.com/rucio/rucio/blob/a7f97edbddf96c0abb1739bacf35581838c73de8/lib/rucio/api/vo.py#L32
- Mostly, double quotes are used: https://github.com/rucio/rucio/blob/a7f97edbddf96c0abb1739bacf35581838c73de8/lib/rucio/api/rule.py#L138

For returns:
- In some places we use `:return:`, e.g.: https://github.com/rucio/rucio/blob/master/lib/rucio/common/didtype.py#L162
- In other places we use `:returns:`, e.g.: https://github.com/rucio/rucio/blob/master/lib/rucio/tests/common.py#L101

For exceptions:
- `:raises`: https://github.com/rucio/rucio/blob/master/lib/rucio/common/config.py#L337
- `:raises:`: https://github.com/rucio/rucio/blob/master/lib/rucio/api/rule.py#L224

I think we should agree on:
- What format we should use for docstrings
- Where do we want to add docstrings
- Whether to automatically generate documentation from docstrings

voetberg:
Some automatic generation is already done for the API/Client - https://github.com/rucio/documentation/blob/main/tools/run_in_docker/generate_client_api_docs.sh , but I do agree

rdimaio:
Reopening as that was only an initial PR, there is still some extra work to be done

voetberg:
In the interest of having fewer overlapping issues, can we narrow the scope of this? Namely, should this issue just focus on core documentation or just documentation style?

rdimaio:
> In the interest of having fewer overlapping issues, can we narrow the scope of this? Namely, should this issue just focus on core documentation or just documentation style?

I would say this issue should focus on documentation style - standardizing formatting, etc. in a way that can be analyzed automatically when linting/in CI

bari12:
Thank you for raising this. We have reviewed this issue as part of our regular issue review.

We see value in this request, but we do not have the capacity to work on it within the next twelve months. 

[...]

PR #6712, opened 2024-04-22
Database: Simplify register and unregister_models; #6677

erlingstaff:
Currently `register_models` tries to register the same schema many times (one time for each model listed), likewise with `unregister_models`. The schema contains all the models. Simplified down to a single call: `BASE.metadata.drop_all(engine)` to register or unregister the schema.

Also added type hints to the two affected functions, #6588 

More details in the comment on issue #6677


lib/rucio/db/sqla/models.py
```diff
@@ -1690,141 +1690,25 @@ class FollowEvents(BASE, ModelBase):
...
 
-def register_models(engine):
+def register_models(engine: Engine) -> None:
     """
-    Creates database tables for all models with the given engine
+    Creates database tables for all models in the `BASE.metadata` schema with the given engine.
+    Functionally this creates all models, as they all inherit from BASE and exist in the same `BASE.metadata` schema.
+
+    :param engine: engine to register the models with
+    :type engine: sqlalchemy.engine.Engine
```
rdimaio, line 1699: We haven't had a formal discussion on docstrings yet (context: https://github.com/rucio/rucio/issues/6628) but I think it's best if we avoid adding `type` information in the docstrings, as long as there's already type annotations in the function signature (like in this case), in order to avoid having to write duplicate information. 

Don't worry about it for this PR though; once a docstring guide is finalized and published then we can worry about standardising docstrings across the codespace, but for now this is a perfectly good docstring

rdimaio (approved):

rdimaio:
Great change - for reviewing context, here's a SO answer that suggests using `create_all` like done in the PR https://stackoverflow.com/a/70402756

PR #6714, opened 2024-04-22
Testing: Add type annotations to common/logging

rdimaio:
Part of #6588

On the `ECS_FIELDS` and `LOG_RECORDS` change, here's an execution of that code to show that the final `BUILTIN_FIELDS` is the same as before:

```
>>> from typing import TYPE_CHECKING, Any, Literal, Optional, Union, get_args
>>> ECS_FIELDS = Literal[
...     '@timestamp',
...     'message',
...     'log.level',
...     'log.origin.function',
...     'log.origin.file.line',
...     'log.origin.file.name',
...     'log.logger',
...     'process.pid',
...     'process.name',
...     'process.thread.id',
...     'process.thread.name'
... ]
>>>
>>> LOG_RECORDS = Literal[
...     'asctime',
...     'message',
...     'levelname',
...     'funcName',
...     'lineno',
...     'filename',
...     'name',
...     'process',
...     'processName',
...     'thread',
...     'threadName'
... ]
>>>
>>> BUILTIN_FIELDS = tuple((x, y) for x,y in zip(get_args(ECS_FIELDS), get_args(LOG_RECORDS)))
>>> BUILTIN_FIELDS
(('@timestamp', 'asctime'), ('message', 'message'), ('log.level', 'levelname'), ('log.origin.function', 'funcName'), ('log.origin.file.line', 'lineno'), ('log.origin.file.name', 'filename'), ('log.logger', 'name'), ('process.pid', 'process'), ('process.name', 'processName'), ('process.thread.id', 'thread'), ('process.thread.name', 'threadName'))
```

### DIFF
--- a/lib/rucio/common/logging.py
+++ b/lib/rucio/common/logging.py
@@ -20,12 +20,15 @@ import re
 import sys
 from collections.abc import Callable, Iterator, Mapping, Sequence
 from traceback import format_tb
-from typing import TYPE_CHECKING, Any, Optional
+from typing import TYPE_CHECKING, Any, Literal, Optional, Union, get_args
 
 from rucio.common.config import config_get, config_get_bool
 
 if TYPE_CHECKING:
-    from logging import LogRecord
+    from logging import LogRecord, _SysExcInfoType
+
+    from _typeshed import OptExcInfo
+    from flask import Flask
 
 
 # Mapping from ECS field paths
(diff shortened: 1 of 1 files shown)


lib/rucio/common/logging.py
```diff
@@ -160,11 +179,11 @@ def _timestamp_formatter(record_formatter: "LogDataSource", record: "LogRecord")
     yield record_formatter.ecs_fields[0], datetime.datetime.utcfromtimestamp(record.created).isoformat(timespec='milliseconds') + 'Z'
 
 
-def _ecs_field_to_record_attribute(field_name):
+def _ecs_field_to_record_attribute(field_name: Union[ECS_FIELDS, str]) -> Union[LOG_RECORDS, str]:
     """
     Sanitize the path-like field name into a symbol which can be the name of an object attribute.
     """
-    record = ECS_TO_LOG_RECORD_MAP.get(field_name)
+    record = ECS_TO_LOG_RECORD_MAP.get(field_name)  # type: ignore
```
rdimaio, line 186: The type ignore here is to ignore the error that occurs when a `str` that's not in `ECS_FIELDS` is passed to `_ecs_field_to_record_attribute`, in which case the function returns `field_name.replace('-', '_').replace('.', '_')`.

I personally don't like this logic being in this function, as the function name doesn't say anything about replacing those symbols within a string, and from the function name `_ecs_field_to_record_attribute` you would assume that this only takes in `ECS_FIELDS` and returns `LOG_RECORDS`.

I think it would be a good idea to change the way this works in future.

lib/rucio/common/logging.py
```diff
@@ -344,15 +363,15 @@ def __init__(
         self.output_json = output_json
         super().__init__(fmt=fmt, style='%', **_kwargs)
 
-    def format(self, record):
-        json_record = dict(itertools.chain.from_iterable(f.format(record) for f in self._desired_data_sources))
+    def format(self, record: "LogRecord") -> str:
+        json_record = dict(itertools.chain.from_iterable(f.format(record) for f in self._desired_data_sources))  # type: ignore
```
rdimaio, line 367: The `type: ignore` here is due to the fact that `LogDataSource.format` returns `None` if there is no `self._formatter`, and the static type checker is not able to guarantee that `f` has a `_formatter` specified in this case. 

This logic should likely be changed so that it only runs `LogDataSource.format` if `_formatter` exists, and otherwise returns `None`, but I think it's best to just`type: ignore` for now and address this later.

bari12 (approved):

PR #6715, opened 2024-04-22
Testing: Check variables exist before accessing them, and add type annotations to common/pcache

rdimaio:
Part of #6588.

### DIFF
--- a/lib/rucio/common/pcache.py
+++ b/lib/rucio/common/pcache.py
@@ -23,8 +23,17 @@ import subprocess
 import sys
 import time
 from socket import gethostname
+from typing import TYPE_CHECKING, Any, Optional
 from urllib.parse import urlencode
 from urllib.request import urlopen
+
+if TYPE_CHECKING:
+    from collections.abc import Iterable, Iterator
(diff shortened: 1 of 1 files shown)


lib/rucio/common/pcache.py
```diff
@@ -1072,7 +1095,7 @@ def do_http_post(self, url, data):
             # This will retry for up to 1 hour, at 2 minute intervals
             while retry < 30:
                 try:
-                    u = urlopen(url, data=urlencode(data))
+                    u = urlopen(url, data=urlencode(data))  # type: ignore
```
rdimaio, line 1098: This `type: ignore` is due to the fact that the type returned by `urlencode` is not compatible with the type accepted by `data` in `urlopen`. 

I'm not sure how this actually behaves at runtime, but due to the type mismatch I'm assuming that `urlencode` shouldn't be used in this context. I think it's best to address this separately.

bari12 (dismissed):

rdimaio:
Rebased in latest force-push

bari12 (approved):

PR #6721, opened 2024-04-23
Documentation: Fix typos in docstrings, comments and variable names; …

rdimaio:
Used the tool "codespell" to identify the typos. Will look into the possibility of adding something like this in the precommit script in order to do spell checking at commit time (probably best not to have it as part of GH Actions, though, as there might be annoying situations where we want to ignore it)

### DIFF
--- a/bin/rucio
+++ b/bin/rucio
@@ -166,7 +166,7 @@ def exception_handler(function):
                     used_account = '%s (from rucio.cfg)' % config_get('client', 'account')
                 except:
                     pass
-                try:  # are we overriden by the environment?
+                try:  # are we overridden by the environment?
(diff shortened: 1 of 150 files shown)


voetberg:
I think having this as a precommit would be nice, and it already exists for codespell - https://github.com/streetsidesoftware/cspell-cli

voetberg (approved):

rdimaio:
> I think having this as a precommit would be nice, and it already exists for codespell - https://github.com/streetsidesoftware/cspell-cli

That's a good idea, but I'm not sure if it could end up being annoying due to possible false positives (though in my experience of using it for this repository, it didn't raise any false positives)

Issue #6657, opened 2024-04-09
`try/except/pass` blocks where no logging is performed before passing


rdimaio:
### Description

There are instances of `try/except/pass` blocks where nothing is logged before passing, for example:

https://github.com/rucio/rucio/blob/b34fe26c68698cf9de166f36372cebf099f2ff57/lib/rucio/core/rule.py#L3893-L3896

For debugging purposes, it might make it easier if we add logs to instances like these (where reasonable).

## To do
For each statement where this happens:
- Determine whether the current `try/except/pass` logic makes sense, or if it should be refactored in some way
- If sticking to the `try/except/pass` logic, determine if it makes sense to log the exception (or some message) before passing
- If no logging should be done, add the inline comment `# noqa: S110`

## Linter log
```
❯ ruff check --select S110                                                                                                                                                                                                                                                                                               ─╯
bin/rucio:167:17: S110 `try`-`except`-`pass` detected, consider logging the exception
bin/rucio:171:17: S110 `try`-`except`-`pass` detected, consider logging the exception
lib/rucio/common/config.py:42:9: S110 `try`-`except`-`pass` detected, consider logging the exception
lib/rucio/common/pcache.py:603:9: S110 `try`-`except`-`pass` detected, consider logging the exception
lib/rucio/common/pcache.py:614:13: S110 `try`-`except`-`pass` detected, consider logging the exception
lib/rucio/common/pcache.py:835:13: S110 `try`-`except`-`pass` detected, consider logging the exception
lib/rucio/common/pcache.py:926:17: S110 `try`-`except`-`pass` detected, consider logging the exception
lib/rucio/common/pcache.py:928:9: S110 `try`-`except`-`pass` detected, consider logging the exception
lib/rucio/common/pcache.py:952:9: S110 `try`-`except`-`pass` detected, consider logging the exception
lib/rucio/common/pcache.py:971:9: S110 `try`-`except`-`pass` detected, consider logging the exception
lib/rucio/common/pcache.py:1012:9: S110 `try`-`except`-`pass` detected, consider logging the exception
lib/rucio/common/pcache.py:1054:13: S110 `try`-`except`-`pass` detected, consider logging the exception
lib/rucio/common/pcache.py:1163:13: S110 `try`-`except`-`pass` detected, consider logging the exception
lib/rucio/common/pcache.py:1175:13: S110 `try`-`except`-`pass` detected, consider logging the exception
[...]

PR #6722, opened 2024-04-23
Adding logging to otherwise bare try/except/pass's #6657

voetberg:
Adding logging to many of the try/pass/except statements. Adds a logger argument to some of the util functions.


lib/rucio/common/utils.py
```diff
@@ -260,11 +260,6 @@ def adler32(file):
     adler = 1
 
     can_mmap = False
-    # try:
-    #    with open(file, 'r+b') as f:
-    #        can_mmap = True
-    # except:
-    #    pass
```
rdimaio: It would be best to have this removal in a separate commit, as it doesn't match the scope of its current commit ("Adding logging to otherwise bare try/except/pass's")

bin/rucio
```diff
@@ -165,11 +165,12 @@ def exception_handler(function):
                 try:  # get the configured account from the configuration file
                     used_account = '%s (from rucio.cfg)' % config_get('client', 'account')
                 except:
-                    pass
+                    logger.debug("Failed to get account from rucio.cfg")
                 try:  # are we overridden by the environment?
                     used_account = '%s (from RUCIO_ACCOUNT)' % os.environ['RUCIO_ACCOUNT']
                 except:
-                    pass
+                    logger.debug("Failed to get account from envoriment variable 'RUCIO_ACCOUNT'")
```
```suggestion
                    logger.debug("Failed to get account from environment variable 'RUCIO_ACCOUNT'")
```

lib/rucio/common/config.py
```diff
@@ -39,7 +40,7 @@ def convert_to_any_type(value: str) -> Union[bool, int, float, str]:
     for conv in (int, float):
         try:
             return conv(value)
-        except:
+        except ValueError:
```
rdimaio, line 45: Changes like this one (where you're not adding logging to a bare except, but you're specifying the exception for a generic `except`) should go in a separate commit for clarity

lib/rucio/common/config.py
```diff
@@ -159,6 +160,7 @@ def config_get(
         use_cache: bool = True,
         expiration_time: int = 900,
         convert_type_fnc: Callable[[str], _T] = lambda x: x,
+        logger: Callable = logging.log
```
rdimaio:
```suggestion
        logger: "LoggerFunction" = logging.log
```

lib/rucio/common/pcache.py
```diff
@@ -1044,16 +1043,17 @@ def daemonize(self) -> None:
...
-        except OSError:
-            pass
+        except OSError as e:
+            self.log(DEBUG, f"Failed to change dir to '/': {e}")
         try:
             os.umask(0)
-        except OSError:
-            pass
+        except OSError as e:
+            self.log(INFO, f"Failed to unmask dir: {e}")
```
rdimaio:
```suggestion
            self.log(INFO, f"Failed to umask dir: {e}")
```

lib/rucio/common/pcache.py
```diff
@@ -1183,8 +1183,8 @@ def unlock_file(self, name: str) -> Optional[Any]:
         if 0:
             try:
                 os.unlink(name)
-            except:
-                pass
+            except Exception as e:
+                self.log(INFO, f"Filed to unlink {name}, {e}")
```
rdimaio:
```suggestion
                self.log(INFO, f"Failed to unlink {name}, {e}")
```

lib/rucio/common/pcache.py
```diff
@@ -1074,8 +1074,8 @@ def daemonize(self) -> None:
         for fd in range(0, maxfd + 1):
             try:
                 os.close(fd)
-            except:
-                pass
+            except Exception as e:
+                self.log(INFO, F"Failed to close {fd}: {e}")
```
rdimaio:
```suggestion
                self.log(INFO, f"Failed to close {fd}: {e}")
```

lib/rucio/common/pcache.py
```diff
@@ -1195,8 +1195,8 @@ def unlock_all(self) -> None:
             try:
                 f.close()
                 os.unlink(filename)
-            except:
-                pass
+            except Exception as e:
+                self.log(INFO, f"Failed to close and unlink {filename}, {e}")
```
rdimaio: I'm not sure if using f-string in logging is ideal, for optimization reasons. @dchristidis do you have any thoughts on this?

lib/rucio/core/oidc.py
```diff
@@ -1407,8 +1407,8 @@ def validate_jwt(json_web_token: str, *, session: "Session") -> dict[str, Any]:
             try:
                 token_dict['audience'] = inspect_claims['aud']
                 token_dict['authz_scope'] = inspect_claims['scope']
-            except:
-                pass
+            except Exception as e:
+                logging.debug(f"Could not access audience and and scope from oidc endpoint {oidc_client.introspection_endpoint}: {e}")
```
rdimaio:
```suggestion
                logging.debug(f"Could not access audience and scope from oidc endpoint {oidc_client.introspection_endpoint}: {e}")
```

lib/rucio/db/sqla/session.py
```diff
@@ -51,13 +51,17 @@
...
+
+if CURRENT_COMPONENT:
+    sql_connection = config_get(
+        '%s-database' % CURRENT_COMPONENT,
+        'default',
+        check_config_table=False,
+        raise_exception=False,
+        default='0').strip()
+    if (sql_connection != '0') and (len(sql_connection)):
+        DATABASE_SECTION = '%s-database' % CURRENT_COMPONENT
```
rdimaio, line 69: This is a functional change, it's best to put it in a separate commit

lib/rucio/db/sqla/session.py
```diff
@@ -207,10 +211,10 @@ def get_engine():
             conv = mysql_convert_decimal_to_float(pymysql=sql_connection.startswith('mysql+pymysql'))
             params['connect_args'] = {'conv': conv}
         for param, param_type in config_params:
-            try:
-                params[param] = param_type(config_get(DATABASE_SECTION, param, check_config_table=False))
-            except:
-                pass
+            param_from_config = config_get(DATABASE_SECTION, param, check_config_table=False, raise_exception=False, default=None)
+            if param_from_config is not None:
+                params[param] = param_type(param_from_config)
```
rdimaio, line 228: This should be in a separate commit (it's not a change that adds logging to a try/except block)

lib/rucio/rse/protocols/ngarc.py
```diff
@@ -42,7 +44,7 @@ def __del__(self):
 class Default(protocol.RSEProtocol):
     """ Implementing access to RSEs using ARC client."""
 
-    def __init__(self, protocol_attr, rse_settings, logger=None):
+    def __init__(self, protocol_attr, rse_settings, logger: Callable = logging.log):
```
rdimaio:
```suggestion
    def __init__(self, protocol_attr, rse_settings, logger: "LoggerFunction" = logging.log):
```

rdimaio (changes requested):

lib/rucio/common/pcache.py
voetberg: filename and e are already strings, and unless I'm mistaken, aren't f-strings just faster than other options? Not sure what you mean by optimization reasons.

dchristidis: Some linters will warn about this: https://pylint.pycqa.org/en/latest/user_guide/messages/warning/logging-fstring-interpolation.html

If the interpolation happens outside the function call, then it happens every time. If it happens inside the function call, then it happens only when necessary. So if your default level is `INFO` and this is a `DEBUG` message, then the interpolation is avoided.

Honestly, I don’t think that it makes a measurable difference performance-wise in our case, but it sounds like a good practice to follow.

lib/rucio/common/pcache.py
voetberg: 'unmask' is the correct spelling

rdimaio: it's not a spelling error; `umask` is the name of the command: https://docs.python.org/3/library/os.html#os.umask. Sorry for not being explicit about it!

lib/rucio/common/pcache.py
rdimaio: I opened an issue to enable this linting feature: https://github.com/rucio/rucio/issues/6826

lib/rucio/common/pcache.py
```diff
@@ -1045,15 +1044,16 @@ def daemonize(self) -> None:
...
         try:
             os.chdir("/")
         except OSError:
-            pass
+            self.log(DEBUG, "Failed to change dir to '/'")
         try:
             os.umask(0)
         except OSError:
-            pass
+            self.log(DEBUG, "Failed to unmask dir")
```
rdimaio:
```suggestion
            self.log(DEBUG, "Failed to umask dir")
```

https://docs.python.org/3/library/os.html#os.umask

rdimaio (changes requested):

rdimaio:
Not sure what caused the Pyright failure https://github.com/rucio/rucio/actions/runs/9547957312/job/26314076870?pr=6722 - I don't see any changes that should cause issues. We can try re-running when the workflow ends

voetberg:
Closed bc the rebase is more work than the original PR is - I will be redoing it

PR #6723, opened 2024-04-23
Hermes: Set values for username and password when use_ssl is set

### LINKED ISSUE #6358: Hermes: username not assigned when use_ssl: True causes an error
### Description

When configuring Hermes Daemon to `use_ssl: True`, Hermes has an error saying, `local variable 'username' referenced before assignment`

### Steps to reproduce

1. Set `messaging-hermes.use_ssl` to True
2. Start the Daemon

### Rucio Version

Daemons: 32.4.0

### Additional Information

- Operating System: OKD Cluster/Kubernetes
- Bug is server side (in hermes daemon)
- ActiveMQ/RabbitMQ is the broker with STOMP plugin

After looking through the hermes daemon code, it looks like in the function `setup_activemq`, the `username` variable is only assigned when `use_ssl` is `False`:
https://github.com/rucio/rucio/blob/d1c0faef08de89825950d4c8194e5a4535ffd6fe/lib/rucio/daemons/hermes/hermes.py#L190

The only instance of `username` being assigned is here:
https://github.com/rucio/rucio/blob/d1c0faef08de89825950d4c8194e5a4535ffd6fe/lib/rucio/daemons/hermes/hermes.py#L152-L155

This doesn't happen when `use_ssl` is `True`

d-ylee:
Fixes #6358 by making sure `username` and `password` variables are referenced and assigned before returning.

### DIFF
--- a/lib/rucio/daemons/hermes/hermes.py
+++ b/lib/rucio/daemons/hermes/hermes.py
@@ -149,6 +149,8 @@ def setup_activemq(logger: "Callable"):
 
     port = config_get_int("messaging-hermes", "port")
     vhost = config_get("messaging-hermes", "broker_virtual_host", raise_exception=False)
+    username = None
+    password = None
     if not use_ssl:
         username = config_get("messaging-hermes", "username")
         password = config_get("messaging-hermes", "password")


rdimaio (approved):

rdimaio:
For reviewing context, `username` and `password` are only used in this code when `if not use_ssl` is `True`.

The `username` and `password` returned in `run_once` by `setup_activemq`:

https://github.com/rucio/rucio/blob/5c0e140e568b9e46dccb3a877a14918558d72631/lib/rucio/daemons/hermes/hermes.py#L551

are passed to `deliver_to_activemq`:

https://github.com/rucio/rucio/blob/5c0e140e568b9e46dccb3a877a14918558d72631/lib/rucio/daemons/hermes/hermes.py#L665-L672

which only uses them `if not use_ssl`:

https://github.com/rucio/rucio/blob/5c0e140e568b9e46dccb3a877a14918558d72631/lib/rucio/daemons/hermes/hermes.py#L218-L224


I think we could consider rewriting these methods to avoid passing around `username` and `password` as `None` when `use_ssl` is `False` (maybe as kwargs), but this fix is fine to resolve the specific issue of variable reference before assignment

PR #6726, opened 2024-04-24
Testing: Update deprecated typing imports and configure Ruff to raise errors in case of deprecated typing imports

### LINKED ISSUE #6720: Remove deprecated `typing` imports
### Description

Several imports from `typing` are deprecated, and should be either:

1. Removed in favor of builtins, ex: `typing.Dict` -> `dict`
2. Re-imported from another module, ex: `from typing import Callable` ->
`from collections.abc import Callable`

The deprecations are relevant in Python versions ranging from 3.8 -> 3.12. This issue focuses on the imports deprecated in 3.8 or 3.9, as 3.9 is the minimum Python needed to run Rucio (as outlined in `setup_rucio.py`). When Rucio is updated, there might be new deprecations.

The following grep command finds all (i think) instances of deprecated imports:
```
grep -R --include="*.py" -E "from typing import.*(Callable|Iterator|Iterable|Sequence|AbstractSet|Dict|Tuple|List|Set|Type|FrozenSet|Tuple|DefaultDict|OrderedDict|ChainMap|Counter|Deque|AbstractSet|Collection|Container|ItemsView|KeysView|Mapping|MappingView|MutableMapping|MutableSet|MutableSequence|ValuesView|Corutine|AsyncGenerator|AsyncIterable|AsyncIterator|Awaitable|Generator|Reversible|ContextManager|AsyncContextManager|Match|Pattern)(,| )" .
```

This currently returns

image

More info:
https://docs.python.org/3/library/typing.html#deprecated-aliases
https://peps.python.org/pep-0585/#implementation

### Motivation

1. Standardize imports
2. Avoid DeprecationWarnings and larger changes in the future

### Change

1. Find all deprecated imports from `typing`, either manually or automatically using e.g. the aforementioned grep command
2. Change as appropriate

rdimaio:
Fixes #6720

### DIFF
--- a/lib/rucio/web/rest/flaskapi/v1/requests.py
+++ b/lib/rucio/web/rest/flaskapi/v1/requests.py
@@ -27,7 +27,7 @@ from rucio.web.rest.flaskapi.authenticated_bp import AuthenticatedBlueprint
 from rucio.web.rest.flaskapi.v1.common import ErrorHandlingMethodView, check_accept_header_wrapper_flask, generate_http_error_flask, parse_scope_name, response_headers, try_stream
 
 if TYPE_CHECKING:
-    from typing import Iterator
+    from collections.abc import Iterator
 
 
 class RequestGet(ErrorHandlingMethodView):
--- a/lib/rucio/client/baseclient.py
+++ b/lib/rucio/client/baseclient.py
@@ -22,12 +22,13 @@ import os
 import secrets
 import sys
 import time
+from collections.abc import Generator
 from configparser import NoOptionError, NoSectionError
 from logging import Logger
 from os import environ, fdopen, geteuid, makedirs, path
 from shutil import move
 from tempfile import mkstemp
-from typing import Any, Generator, Optional
+from typing import Any, Optional
 from urllib.parse import urlparse
 
 import requests
--- a/lib/rucio/common/plugins.py
+++ b/lib/rucio/common/plugins.py
@@ -14,8 +14,9 @@
 
 import importlib
 import os
+from collections.abc import Callable
 from configparser import NoOptionError, NoSectionError
-from typing import Any, Callable, Dict, Type, TypeVar
+from typing import Any, TypeVar
 
 from rucio.common import config
 from rucio.common.exception import InvalidAlgorithmName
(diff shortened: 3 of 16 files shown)


lib/rucio/web/rest/flaskapi/v1/requests.py
```diff
@@ -27,7 +27,7 @@
 from rucio.web.rest.flaskapi.v1.common import ErrorHandlingMethodView, check_accept_header_wrapper_flask, generate_http_error_flask, parse_scope_name, response_headers, try_stream
 
 if TYPE_CHECKING:
-    from typing import Iterator
+    from collections.abc import Iterator
```
dchristidis, line 30: I think we should recommend that modules from the Python standard library are not conditionally imported, even if used exclusively for type hints.

dchristidis (approved):

lib/rucio/web/rest/flaskapi/v1/requests.py
rdimaio: `flake8-type-checking` has a setting to achieve this: https://docs.astral.sh/ruff/settings/#lintflake8-type-checking

I opened this issue to track this work: https://github.com/rucio/rucio/issues/6728

rdimaio: On the other hand, this page makes a good point (https://docs.astral.sh/ruff/rules/typing-only-standard-library-import/):

> Unused imports add a performance overhead at runtime, and risk creating import cycles. If an import is only used in typing-only contexts, it can instead be imported conditionally under an if TYPE_CHECKING: block to minimize runtime overhead.

And there's this setting (https://docs.astral.sh/ruff/settings/#lint_flake8-type-checking_quote-annotations) which enables this:

> Whether to add quotes around type annotations, if doing so would allow the corresponding import to be moved into a type-checking block.

I think it might be worth following these guidelines and enabling the linter to check for imports that are only used in a type-checking context, and place them in an `IF TYPE_CHECKING` block to lower runtime overhead. What do you think?

dchristidis: Very nice. I will defer this decision to you.

The arguments in favour of conditional imports are known to me and I think they are valid for third-party libraries (e.g. `sqlalchemy.orm.Session`). I have my doubts as to how much they apply to the Python standard library and I have a (personal) preference against having to use quotes for the type hints. However, I am open to a different decision.

rdimaio: I opted for using `flake8-type-checking` with `quote-annotations=true`. PR: https://github.com/rucio/rucio/pull/6731

PR #6732, opened 2024-04-25
Testing: Add type annotations to core/did

rdimaio:
Part of #6588

### DIFF
--- a/lib/rucio/core/did.py
+++ b/lib/rucio/core/did.py
@@ -18,7 +18,7 @@ from datetime import datetime, timedelta
 from enum import Enum
 from hashlib import md5
 from re import match
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any, Literal, Optional, Union
 
 from sqlalchemy import and_, delete, exists, insert, or_, update
 from sqlalchemy.exc import DatabaseError, IntegrityError, NoResultFound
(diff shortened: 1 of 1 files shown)


lib/rucio/core/did.py
```diff
@@ -349,7 +349,15 @@ def attach_dids_to_dids(
         session.execute(insert(models.UpdatedDID), parent_dids)
 
 
-def __add_files_to_archive(parent_did, files_temp_table, files, account, ignore_duplicate=False, *, session: "Session"):
+def __add_files_to_archive(
+        parent_did: models.DataIdentifier,
+        files_temp_table: Any,
```
rdimaio, line 354: Note on the type `Any` for the temp tables across these functions:

These temp tables are generated via `rucio.db.sqla.util._create_temp_table`, which returns the class `DeclarativeObj`, which is defined inside `_create_temp_table` itself. 

Refactoring this class to be defined outside of `_create_temp_table` does not seem to be straightforward, as the class definition itself uses variables from `_create_temp_table`. These could be set in a constructor, but I'm not sure if the functionality would remain the same.

A workaround would be to create a `TypeVar` named `DeclarativeObj` to annotate these tables more specifically, but then type checkers complain that `TypeVar[DeclarativeObj` has no members `scope` and `name`.

The solution would be to refactor `DeclarativeObj` out of `_create_temp_table`, but I would like to do this in a follow-up PR with assistance from someone that has already worked closely on these methods.

bari12 (approved):

Issue #6670, opened 2024-04-11
upload client sometimes misses either declaring the file replica or declaring the file to a dataset


StevenCTimm:
### Description

On several occasions DUNE has observed that the uploadclient.py. of rucio v33.3 (as called in our utility justin-rucio-upload)

https://github.com/DUNE/dune-justin/blob/main/commands/justin-rucio-upload

either fails to register the file replica, or fails to attach the DID to the requested dataset.

This tends to happen when the rucio server is under high load.

[6093.90.justin-prod-sched02.dune.hep.ac.uk.txt](https://github.com/rucio/rucio/files/14947629/6093.90.justin-prod-sched02.dune.hep.ac.uk.txt)

The snippet of the debug output generated from the rucio upload client is below.
In short what you see is 5 attempts at rucio upload--3 to QMUL and 2 to RAL-PP the last of which
finally showed a success.  the QMUL RSE was very busy but we were also getting 503 errors against the
server at the time.

But what actually happened under the covers is that the file in question was attached to the dataset
but the replica was not declared to rucio.

A layman's review of uploadclient.py seems to indicate that there are very few if any checks for exceptions in the several lower level tasks that this client does.. i.e. there is an attach to the DID and no attached method to handle the exception--so it's quite possible that the 2 tasks are not atomic--what needs to be done is either to make the client code retry if it fails, or return an error if one of those two fails.

DEBUG:root:Num. of files that upload client is processing: 1
DEBUG:dogpile.cache.region:No value present for key: "host_to_choose_choice['https://dune-rucio.fnal.gov']"
DEBUG:dogpile.lock:NeedRegenerationException
DEBUG:dogpile.lock:no value, waiting for create lock
DEBUG:dogpile.lock:value creation lock <dogpile.cache.region.CacheRegion._LockWrapper object at 0x2ba369504ca0> acquired
DEBUG:dogpile.cache.region:No value present for key: "host_to_choose_choice['https://dune-rucio.fnal.gov']"
DEBUG:dogpile.lock:Calling creation function for not-yet-present value
DEBUG:dogpile.cache.region:Cache value generated in 0.000 seconds for key(s): "host_to_choose_choice['https://dune-rucio.fnal.gov']"
DEBUG:dogpile.lock:Released creation lock
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): dune-rucio.fnal.gov:443
[...]

PR #6754, opened 2024-05-03
Clients: don't return success from uploader when registration fails #6670

jamesp-epcc:
This PR ensures that the upload client will not return success if updating the replica state or attaching to a dataset fails. Previously if the Rucio server timed out during one of these operations the upload was still flagged as a success and there was no way for the caller to detect the registration failure.

### DIFF
--- a/lib/rucio/client/uploadclient.py
+++ b/lib/rucio/client/uploadclient.py
@@ -286,7 +286,6 @@ class UploadClient:
                     state_reason = str(error)
 
             if success:
-                num_succeeded += 1
                 trace['transferEnd'] = time.time()
                 trace['clientState'] = 'DONE'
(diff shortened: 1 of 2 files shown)


rdimaio (dismissed):
We should add a unit test at some point to assert `num_succeeded` is as expected

bari12 (changes requested):
LGTM, but as @rdimaio already mentioned, please add a unit test for this right away, so we make sure this does not regress in the future

jamesp-epcc:
I've added a unit test that simulates the replica state update timing out and ensures that an exception is thrown as a result.

rdimaio (approved):

bari12 (approved):

PR #6758, opened 2024-05-06
Metadata: Allow authenticated mongodb connections; Fix #6605

### LINKED ISSUE #6605: MongoDB metadata implementation does not allow using authentication
### Description

There is now way to configure username / password for the mongo db connection

### Motivation

Any real deployment will need authenticated access to a database

### Change

Add configuration options `mongo_service_user` and `mongo_service_password` in the `[metdata]` section, use to create and authenticated client.

I can open a PR

maxnoe:
Follow up to #6606 following the contributors guide

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -66,6 +66,7 @@ Individual contributors to the source code
 - Jensen Zhang <hack@jensen-zhang.site>, 2022
 - Aksel Lunde Aase <aksel.lunde.aase@gmail.com>, 2022
 - Elena Gazzarrini <gazzarrini.elena@gmail.com>, 2022-2023
+- Maximilian Linhoff, <maximilian.linhoff@tu-dortmund.de>, 2024
 
 Organisations employing contributors
 ------------------------------------
@@ -93,3 +94,4 @@ Organisations employing contributors
 - Imperial College London (UK)
 - University of Chicago (USA)
 - Purdue University (USA)
+- TU Dortmund University (Germany)
--- a/lib/rucio/core/did_meta_plugins/mongo_meta.py
+++ b/lib/rucio/core/did_meta_plugins/mongo_meta.py
@@ -36,7 +36,15 @@ IMMUTABLE_KEYS = [
 
 
 class MongoDidMeta(DidMetaPlugin):
-    def __init__(self, host=None, port=None, db=None, collection=None):
+    def __init__(
+        self,
+        host: "Optional[str]" = None,
+        port: "Optional[int]" = None,
+        db: "Optional[str]" = None,
+        collection: "Optional[str]" = None,
+        user: "Optional[str]" = None,
+        password: "Optional[str]" = None,
+    ):
         super(MongoDidMeta, self).__init__()
         if host is None:
             host = config.config_get('metadata', 'mongo_service_host')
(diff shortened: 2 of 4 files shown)


lib/rucio/core/did_meta_plugins/mongo_meta.py
```diff
@@ -47,7 +55,16 @@ def __init__(self, host=None, port=None, db=None, collection=None):
         if collection is None:
             collection = config.config_get('metadata', 'mongo_collection')
 
-        self.client = pymongo.MongoClient("mongodb://{}:{}/".format(host, port))
+        if user is None:
+            user = config.config_get("metadata", "mongo_user", default=None)
+        if user is not None and password is None:
+            password = config.config_get("metadata", "mongo_password")
+        if user is not None:
+            auth = "{user}:{password}@".format(user=user, password=password)
```
rdimaio: this could be `if user is not None and password is not None` to be extra sure that `password` exists

tests/test_did_meta_plugins.py
```diff
@@ -198,7 +198,9 @@ def mongo_meta():
         host='mongo',
         port=27017,
         db='test_db',
-        collection='test_collection'
+        collection='test_collection',
+        user="rucio",
+        password="mongo-meta",
```
rdimaio: It would be great if we had another `mongo_meta` fixture without `user` and `password` (and corresponding tests with that fixture), to ensure that unauthenticated mongodb connections still work

lib/rucio/core/did_meta_plugins/mongo_meta.py
maxnoe: actually, it can just be merged with  the if above. If `password is None` and doesn't appears in the configuration, we error.

tests/test_did_meta_plugins.py
maxnoe: That would need a second mongodb in the docker compose yaml I guess?

maxnoe: I added the second mongo definition and parametrized the test to test both authenticated and unauthenticated database connections

lib/rucio/core/did_meta_plugins/mongo_meta.py
maxnoe: done

AUTHORS.rst
```diff
@@ -66,6 +66,7 @@ Individual contributors to the source code
 - Jensen Zhang <hack@jensen-zhang.site>, 2022
 - Aksel Lunde Aase <aksel.lunde.aase@gmail.com>, 2022
 - Elena Gazzarrini <gazzarrini.elena@gmail.com>, 2022-2023
+- Maximilian Linhoff, <maximilian.linhoff@tu-dortmund.de>, 2024
```
rdimaio, line 69: The changes to `AUTHORS.rst` should be in a separate commit

lib/rucio/core/did_meta_plugins/mongo_meta.py
```diff
@@ -36,7 +36,15 @@
...
 class MongoDidMeta(DidMetaPlugin):
-    def __init__(self, host=None, port=None, db=None, collection=None):
+    def __init__(
+        self,
+        host: "Optional[str]" = None,
+        port: "Optional[int]" = None,
+        db: "Optional[str]" = None,
+        collection: "Optional[str]" = None,
+        user: "Optional[str]" = None,
+        password: "Optional[str]" = None,
```
rdimaio, line 46: Since these type hints only use built-in types and `Optional` from `typing`, we could move `from typing import Optional` outside of the `if TYPE_CHECKING` block and remove the quotes from these type annotations.

I'm just leaving this as a small note, no need to make any changes, as this is something that we can lint automatically with Ruff. We temporarily disabled the code that handles this because of false positives: https://github.com/rucio/rucio/blob/fb42cda0c8c49ec11a3fe624be2ca17d208887a8/pyproject.toml#L59, but once we enable it again this will be formatted automatically.

rdimaio (commented):
Ruff failures are due to an unrelated issue that was fixed in a separate PR (they will go away if you rebase, but there's no need to do that since there's no merge conflicts)

AUTHORS.rst
maxnoe: Sure

rdimaio (approved):

PR #6783, opened 2024-05-28
Testing: Add type annotations to client/accountclient

rdimaio:
Part of #6588.

About this failure:

```
Found 1 new problems in /home/runner/work/rucio/rucio/lib/rucio/client/downloadclient.py
  - 1 errors with message """Argument of type "Unknown | str | None" cannot be assigned to parameter "account" of type "str" in function "list_account_attributes"
                               Type "Unknown | str | None" is incompatible with type "str"
                                 "None" is incompatible with "str"""".
    Candidates: line 148
```

This is an issue within `downloadclient`, where `account` might be `None`. I think we should ignore this for the context of this PR, and address it in a separate PR where type hints are added to `downloadclient` and this issue is addressed.

### DIFF
--- a/lib/rucio/client/accountclient.py
+++ b/lib/rucio/client/accountclient.py
@@ -13,12 +13,16 @@
 # limitations under the License.
 
 from json import dumps
+from typing import TYPE_CHECKING, Any, Optional
 from urllib.parse import quote_plus
 
 from requests.status_codes import codes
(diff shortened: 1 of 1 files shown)


rdimaio:
Rebased in latest force push

voetberg (approved):

Issue #6780, opened 2024-05-28
archive metadata plugin sends empty collocation hints


labkode:
### Description

The archive metadata plugin sends empty collocation hints to FTS.

The following have been observed in ATLAS and CMS FTS databases for multiple transfers:

```
MySQL [newfts3atlas]> select archive_metadata, count(*) from t_file group by archive_metadata;
+---------------------------+----------+
| archive_metadata          | count(*) |
+---------------------------+----------+
| NULL                      |    15780 |
| {"collocation_hints": {}} | 14437895 |
+---------------------------+----------+
2 rows in set (29.215 sec)

MySQL [newfts3atlas]
```

I think it could be a side effect that when no policy is configured, the default is to send an empty dict?

If there are no collocation hints, the archive metadata should not be sent.

### Steps to reproduce

Observed in production env

labkode:
Thanks @voetberg !

PR #6786, opened 2024-05-28
Transfers: Ensure no archive metadata is used if a plugin is not configured #6780

voetberg:
View issue #6780


voetberg:
Failure is due to dependency sorting in unchanged files

bari12 (approved):

Issue #6741, opened 2024-04-30
Declare Bad File Replcas fails silently


dynamic-entropy:
### Description

The Python client call to declare file replicas as bad fails silently, without giving any results to the user about success or failure. 

The docs https://rucio.github.io/documentation/client_api/replicaclient#declare_bad_file_replicas for the command say that a list of dids for which the declaration failed is returned. 
However, this does not seem to be the case.

Are there use cases where the silent failing is an intended behaviour?

### Steps to reproduce

Use a random scope and name for a file with the command, with a valid RSE
```
>>> client.declare_bad_file_replicas([{'scope': 'cms', 'name': 'InvalidFileName', 'rse': 'T2_IT_Rome'}], reason='Lost on Disk')
{}
```

PR #6792, opened 2024-05-30
Recovery: Add catch if did does not exist #6741

voetberg:
Correcting replica returning for the python client when declaring bad replicas. If the replica does not exist and is not already declared, it is now added to the unknown replicas list for the client to return. 

Also adds some additional logging at debug level to show what pfn, paths, replicas are being declared in the logs.


rdimaio (commented):
The `logger`-related changes should be in a separate commit

bari12 (approved):

lib/rucio/core/replica.py
```diff
@@ -314,6 +313,9 @@ def __declare_bad_file_replicas(pfns, rse_id, reason, issuer, status=BadFilesSta
                 new_bad_replica.save(session=session, flush=False)
                 declared = True
 
+        else:
+            unknown_replicas.append(f"{path_pfn_dict[path]} Unknown replica")
```
faluchet, line 321: I believe this introduced a regression. 
The variable `path` , as returned by `__exist_replicas` in L306, can have the None value when `__exist` is false and the else branch is entered.

`path_pfn_dict[path]` will therefore raise an exception, that won't be caught anywhere.

Issue #6803, opened 2024-06-06
`DIDClient.add_containers()`: possible bug in function logic


rdimaio:
Regarding this line:

https://github.com/rucio/rucio/blob/ad19fb40f47f859c3c7f9e2786b4488f43f77ead/lib/rucio/client/didclient.py#L175

In `[dict(list(cnts.items()) + [('type', 'CONTAINER')]) for cnt in cnts]`, the `cnts.items()` seems suspicious - I think it should be `cnt.items()` instead. As it stands, `cnt` is unused in this loop. Is this intended?

bari12:
Honestly, I think this isn't working at all. event with `cnt`?

PR #6804, opened 2024-06-06
Testing: Add type annotations to client/didclient; 

rdimaio:
Part of #6588.

Regarding the test failures:

```
 Found 1 new problems in /home/runner/work/rucio/rucio/lib/rucio/client/didclient.py
  - 1 errors with message """Cannot access attribute "items" for class "Sequence[dict[str, Any]]"
                               Attribute "items" is unknown""".
```

For this one, we need to discuss this issue first https://github.com/rucio/rucio/issues/6803 to understand if there's a bug or not. It will go away once we resolve that issue. I am marking this PR as draft for now.

Edit: This has been fixed in: https://github.com/rucio/rucio/pull/6807

```
 Found 1 new problems in /home/runner/work/rucio/rucio/lib/rucio/rse/protocols/bittorrent.py
  - 1 errors with message """Argument of type "Any | Unknown | None" cannot be assigned to parameter "length" of type "int" in function "construct_torrent"
                               Type "Any | Unknown | None" is incompatible with type "int"
                                 "None" is incompatible with "int"""".
    Candidates: line 163
```

This one is due to `length` being used in `bittorrent.py` without checking if it exists first. To avoid functional changes in `bittorrent` for now, I think we should ignore it and address it at a later point.

### DIFF
--- a/lib/rucio/client/didclient.py
+++ b/lib/rucio/client/didclient.py
@@ -14,7 +14,7 @@
 
 from datetime import datetime
 from json import dumps
-from typing import Any, Optional
+from typing import TYPE_CHECKING, Any, Literal, Optional, Union
 from urllib.parse import quote_plus
 
 from requests.status_codes import codes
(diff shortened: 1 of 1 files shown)


rdimaio:
Rebased in latest force push. The only test failure will now be the second one described in the first post (the `bittorrent` issue), which I think should not be addressed as part of this PR

bari12 (approved):

PR #6821, opened 2024-06-13
Check for RSE before downloading a PFN when no RSE is specified

voetberg:
Makes a call to the replica client to get an RSE for the pfn to download from when there is no given rse.

### DIFF
--- a/bin/rucio
+++ b/bin/rucio
@@ -30,6 +30,7 @@ from configparser import NoOptionError, NoSectionError
 from copy import deepcopy
 from datetime import datetime
 from functools import wraps
+from typing import Optional
 
 from tabulate import tabulate
 
@@ -1077,6 +1078,24 @@ def download(args):
             logger.debug(args.dids)
         item_defaults['pfn'] = args.pfn
         item_defaults['did'] = did_str
+        if args.rses is None:
+            logger.warning("No RSE was given, selecting one.")
+            client = get_client(args)
+            replicas = client.list_replicas(
+                [{"scope": did_str.split(':')[0], "name": did_str.split(':')[-1]}],
+                schemes=args.protocol,
+                ignore_availability=False,
+                client_location=detect_client_location(),
+                resolve_archives=not args.no_resolve_archives
+            )
+
+            download_rse = _get_rse_for_pfn(replicas, args.pfn)
+            if download_rse is None:
+                logger.error("Could not find RSE for pfn %s", args.pfn)
+                return FAILURE
+            else:
+                item_defaults['rse'] = download_rse
+
         result = download_client.download_pfns([item_defaults], 1, trace_pattern, deactivate_file_download_exceptions=deactivate_file_download_exceptions)
 
     if not result:
@@ -1116,6 +1135,21 @@ def download(args):
         print('{0:40} {1:6d}'.format('Files that cannot be downloaded: ', not_downloaded_files))
 
     return SUCCESS
+
+
+def _get_rse_for_pfn(replicas, pfn) -> Optional[str]:
+    # Check each rse in the replica list for the pfn. If no pfn is found, returns None.
+    # If it is found, stop the generator and return the item.
+    for replica in replicas:
+        try:
+            download_rse = next(
+                rse for rse in replica['rses']
+                if pfn in replica['rses'][rse]
+            )
+        except StopIteration:
+            continue
+        else:
+            return download_rse
 
 
 @exception_handler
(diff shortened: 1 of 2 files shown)


voetberg:
Ruff failure is in transfertools - unrelated.

bin/rucio
```diff
@@ -1077,6 +1077,29 @@ def download(args):
...
+                client_location=detect_client_location(),
+                resolve_archives=not args.no_resolve_archives
+            )
+            download_rse = None
+            for replica in replicas:
+                for rse in replica['rses']:
+                    if replica['rses'][rse]:
+                        for pfn in replica['rses'][rse]:
+                            if pfn == args.pfn:
+                                download_rse = rse
```
rdimaio: This iterates over all replicas, even if a matching PFN is found early on in the loop.

I suggest refactoring according to this: https://stackoverflow.com/a/2364277

rdimaio (changes requested):

bin/rucio
voetberg: Yeah I didn't particularly like this either, I was just pulling from the existing logic in [list_file_replicas](https://github.com/rucio/rucio/blob/ba102506d470c417fd2b136304e4fa4f7fc3a870/bin/rucio#L428). All the logic here needs some love.

tests/test_bin_rucio.py
```diff
@@ -681,19 +681,25 @@ def test_download_pfn(self):
...
             assert re.search('Total files.*1', out) is not None
 
+        # Try to use the --pfn without rse
+        cmd = f"rucio -v download  --dir {download_dir.rstrip('/')}/duplicate --pfn {replica_pfn} {self.user}:{name}"
+        exitcode, out, err = execute(cmd)
+
+        assert "No RSE was given, selecting one." in err
+        assert exitcode == 0
+        assert re.search('Total files.*1', out) is not None
+
```
rdimaio, line 702: It would be a good idea to also add a test where the RSE is not given and the PFN is not found, to test for the branch where the iterator is exhausted.

bin/rucio
```diff
@@ -1077,6 +1077,33 @@ def download(args):
...
+                download_rse = next(
+                    next(
+                        rse for rse in replica['rses']
+                        if args.pfn in replica['rses'][rse]
+                    )
+                    for replica in replicas
+                )
+                item_defaults['rse'] = download_rse
+
+            # Raised after exhauting the iterator
```
```suggestion
            # Raised after exhausting the iterator
```

bin/rucio
```diff
@@ -1077,6 +1077,33 @@ def download(args):
...
+            try:
+                # Check each rse in the replica list for the pfn. If no pfn is found, raise StopIteration.
+                # If it is found, stop the generator and add the rse to 'item_defaults'
+                download_rse = next(
+                    next(
+                        rse for rse in replica['rses']
+                        if args.pfn in replica['rses'][rse]
+                    )
+                    for replica in replicas
+                )
```
rdimaio: My understanding here is that, if the inner `next` exhausts the iterator, it raises a `StopIteration`, so the outer loop will never go beyond the first iteration.

For readability, I think it would be best to just use `for` loops and either:
1. refactor the loop into a function (e.g. `get_rse_by_pfn` or something along those lines) and `return` the rse if found, no return otherwise
2. keep the loop here and using `break` when a matching rse is found

rdimaio (changes requested):

bin/rucio
```diff
@@ -1118,6 +1137,19 @@ def download(args):
     return SUCCESS
 
 
+def _get_rse_for_pfn(replicas, pfn) -> "Optional[str]":
```
rdimaio: small note: `Optional[str]` doesn't need the quotes around it. It's fine to have it like this anyway, so you can ignore this

bin/rucio
```diff
@@ -1118,6 +1137,19 @@ def download(args):
...
+    # Check each rse in the replica list for the pfn. If no pfn is found, returns None.
+    # If it is found, stop the generator and return the item.
+    download_rse = next((
+        next((
+            rse for rse in replica['rses']
+            if pfn in replica['rses'][rse]
+        ), None)
+        for replica in replicas
+    ), None)
```
rdimaio: This works fine, but I think that the outer `next()` is not necessary over a regular `for` loop (since it does not have a condition tied to it, it's just iterating).

For readability, I think this would be easier to understand:
```python
def _get_rse_for_pfn(replicas, pfn):
    # Check each rse in the replica list for the pfn. If no pfn is found, returns None.
    # If it is found, stop the generator and return the item.
    for replica in replicas:
        try:
            download_rse = next(
                rse for rse in replica['rses']
                if pfn in replica['rses'][rse]
            )
        except StopIteration:
            continue
        else:
            return download_rse
```

This is just my opinion though, I think it's also perfectly fine to keep the current version

rdimaio (dismissed):

bin/rucio
bari12: Yes, please still change this @voetberg

rdimaio (approved):

Issue #5129, opened 2022-01-10
Document and cleanup discrepancies between surl and lfn2pfn algorithms


bari12:
Motivation
----------
Currently the difference between the `lfn2pfn` (Deterministic) and `surl` (Non-deterministic) algorithms is very hard to understand. The naming comes mostly for historic reasons, but it is hard to understand/follow the motivation of it.

This needs to be both adressed in the documentation as well as the code itself. The `surl` algorithms should probably be re-named into something else (Mentioning non_deterministic) but even then, it needs to be well defined what a deterministic vs. non-deterministic algorithm can entail.

PR #6828, opened 2024-06-17
Core & Internals: rename SURL to (non-deterministic) PFN #5129

jamesp-epcc:
This PR renames 'surl' to something clearer throughout the code. When a non-deterministic PFN is intended (such as in the SURL algorithms), it has been renamed to `non_deterministic_pfn`. In other places it has been renamed simply to `pfn` (for example, `clean_surls` has become `clean_pfns`) where this makes more sense.

I will also submit a documentation PR to rename 'surl' and clarify the differences between deterministic and non-deterministic PFN algorithms.

### DIFF
--- a/bin/rucio-admin
+++ b/bin/rucio-admin
@@ -54,7 +54,7 @@ from rucio.common.exception import (
     ScopeNotFound,
 )
 from rucio.common.extra import import_extras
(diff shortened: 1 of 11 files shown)


bari12 (approved):
I think this is all good. Thanks James. 
Luckily the rse_attribute for the convention is `naming_convention`, so this all stays the same mostly. I think we need to write a notice in the release not though, for communities who have custom conventions, that they need to slightly adapt them. Can you please write one and send it to me! Thanks!

PR #6834, opened 2024-06-18
commons: support conda env #6833

### LINKED ISSUE #6833: Support conda env
Configuration now support virtualenv but not conda / mamba env.

PR coming

wiso:
Now common/config.py supports virtualenv. With this PR it supports also conda / mamba environments.

### DIFF
--- a/lib/rucio/common/config.py
+++ b/lib/rucio/common/config.py
@@ -708,7 +708,8 @@ def get_config_dirs() -> list[str]:
     """
     Returns all available configuration directories in order:
     - $RUCIO_HOME/etc/
-    - $VIRTUAL_ENV/etc/
+    - $VIRTUAL_ENV/etc/  
+    - $CONDA_ENV/etc
     - /opt/rucio/
     """
     configdirs = []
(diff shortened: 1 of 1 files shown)


wiso:
do I really need to add myself to the list of contributors for only 2 lines?

lib/rucio/common/config.py
```diff
@@ -709,6 +709,7 @@ def get_config_dirs() -> list[str]:
     Returns all available configuration directories in order:
     - $RUCIO_HOME/etc/
     - $VIRTUAL_ENV/etc/
+    - $CONDA_ENV/etc
```
rdimaio: AFAIK `CONDA_ENV` is not a default environment variable - did you mean `CONDA_PREFIX` here?

rdimaio:
> do I really need to add myself to the list of contributors for only 2 lines?

It's not mandatory, but you can do it if you want

lib/rucio/common/config.py
wiso: ops! the code actually is correct

rdimaio (approved):
LGTM - the commits should be squashed into a single commit, but we can do that on our side when merging

Issue #6822, opened 2024-06-13
rucio cli x509 options not working


maxnoe:
### Description

The `rucio` cli tool has 

- an option `--certificate` for the x509 certificate but doesn't use it
- has no option for the key

### Steps to reproduce

Look into `bin/rucio` or and `rucio --help`, observe that:

`args.certificate` is never used

there is no option to specify the key

### Rucio Version

current master and older

voetberg:
Easy to fix - question about testing this though. The bin tests assume we have auth already set up when CLI is used, is this something we should change in the future? (They just use userpass if I remember correctly).

PR #6838, opened 2024-06-20
Client: Pass forward x509 cert and key #6822

voetberg:
Add arguments for the x509 key, pass forward to the client auth

### DIFF
--- a/bin/rucio
+++ b/bin/rucio
@@ -234,6 +234,8 @@ def get_client(args):
                  'oidc_issuer': args.oidc_issuer,
                  'oidc_username': args.oidc_username,
                  'oidc_password': args.oidc_password}
+    elif auth_type == "x509":
+        creds = {'client_cert': args.certificate, "client_key": args.client_key}
     else:
         creds = None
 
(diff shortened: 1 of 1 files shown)


rdimaio (commented):
LGTM - it would be great to also have a unit test to confirm this works correctly

voetberg:
> LGTM - it would be great to also have a unit test to confirm this works correctly

Yeah totally agree - I think that would be a larger undertaking though. All of `test_bin_rucio` assumes auth has already taken place. (just with user/pass).

rdimaio (approved):

rdimaio:
> > LGTM - it would be great to also have a unit test to confirm this works correctly
> 
> Yeah totally agree - I think that would be a larger undertaking though. All of `test_bin_rucio` assumes auth has already taken place. (just with user/pass).

I agree. I opened this issue to track creating tests for the auth methods: https://github.com/rucio/rucio/issues/6839, we can merge this first and then add tests as part of that issue

dchristidis:
This PR introduced an annoying regression to `x509` authentication. The client reads from the configuration file only when `creds` is unset.

https://github.com/rucio/rucio/blob/0bbc069faf17f06eff1891aa90e634adea67b216/lib/rucio/client/baseclient.py#L264-L266

PR #6840, opened 2024-06-21
Testing: Add type annotations to client/uploadclient

rdimaio:
Part of #6588

### DIFF
--- a/lib/rucio/client/uploadclient.py
+++ b/lib/rucio/client/uploadclient.py
@@ -866,7 +866,7 @@ class UploadClient:
                 elif len(dirs) > 0:
                     containers.append({'scope': scope, 'name': root.split('/')[-1]})
                     self.logger(logging.DEBUG, 'Appended container with DID %s:%s' % (scope, path))
-                    attach.extend([{'scope': scope, 'name': root.split('/')[-1], 'rse': rse, 'dids': {'scope': scope, 'name': dir_}} for dir_ in dirs])
+                    attach.extend([{'scope': scope, 'name': root.split('/')[-1], 'rse': rse, 'did': {'scope': scope, 'name': dir_}} for dir_ in dirs])
                 elif len(dirs) == 0 and len(fnames) == 0:
                     self.logger(logging.WARNING, 'The folder %s is empty, skipping' % root)
                     continue
@@ -888,7 +888,7 @@ class UploadClient:
                 self.logger(logging.ERROR, 'It was not possible to create dataset with DID %s:%s' % (container['scope'], container['name']))
         for att in attach:
             try:
-                self.client.attach_dids(scope=att['scope'], name=att['name'], dids=[att['dids']])
+                self.client.attach_dids(scope=att['scope'], name=att['name'], dids=[att['did']])
                 self.logger(logging.INFO, 'DIDs attached to collection %s:%s' % (att['scope'], att['name']))
             except RucioException as error:
                 self.logger(logging.ERROR, error)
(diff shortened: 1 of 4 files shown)


lib/rucio/client/uploadclient.py
```diff
@@ -500,7 +530,7 @@ def _collect_file_info(self, filepath, item):
 
         return new_item
 
-    def _collect_and_validate_file_info(self, items):
+    def _collect_and_validate_file_info(self, items: "Iterable[FileToUploadDict]") -> list["FileToUploadWithCollectedInfoDict"]:
```
voetberg, line 533: Why list and not Sequence?

lib/rucio/client/uploadclient.py
```diff
@@ -634,7 +677,7 @@ def _upload_item(self, rse_settings, rse_attributes, lfn, source_dir=None, domai
                 pfn = self.client.get_signed_url(rse_settings['rse'], sign_service, 'write', pfn)
 
         # Create a name of tmp file if renaming operation is supported
-        pfn_tmp = '%s.rucio.upload' % pfn if protocol_write.renaming else pfn
+        pfn_tmp = cast(str, '%s.rucio.upload' % pfn if protocol_write.renaming else pfn)
```
voetberg, line 680: Why is cast necessary here? Are there cases where pfn isn't already a string?

voetberg (commented):
I have a few questions and not sure how I feel about the use of typing.cast, it's pretty uncommon and may cause confusion, a standard `# type: ignore` works just as well

lib/rucio/client/uploadclient.py
rdimaio: The general type annotation convention is to use abstract collection types e.g. `Iterable`, `Sequence` for function arguments, but to use specific types like `list` for the return type: https://docs.python.org/3/library/typing.html#typing.List

lib/rucio/client/uploadclient.py
rdimaio: The static type checker thinks that `pfn` could be `None`, which occurs in the event that:
- the `Exception` occurs on line 665
- `force_pfn` is `None`
- `if sign_service` is not `True`

I don't know if this edge case can realistically occur. I am using `cast` there because `pfn_tmp` is used in a few places afterwards that expect a `str` - if I were to `# type: ignore` then I would have to ignore multiple lines instead of just casting on one

rdimaio:
> I have a few questions and not sure how I feel about the use of typing.cast, it's pretty uncommon and may cause confusion, a standard `# type: ignore` works just as well

I can replace `typing.cast` with `# type: ignore` in most places. They fundamentally do the same thing - I think the comment might be faster, but `typing.cast` does nothing at runtime (https://docs.python.org/3/library/typing.html#typing.cast) so it should be pretty negligible.

I would mainly only keep `typing.cast` in contexts where there are multiple type errors for a single variable that the static type checker doesn't understand - instead of adding a comment on every line where it's used, it might be better to just `cast` in that case.

voetberg:
> > I have a few questions and not sure how I feel about the use of typing.cast, it's pretty uncommon and may cause confusion, a standard `# type: ignore` works just as well
> 
> I can replace `typing.cast` with `# type: ignore` in most places. They fundamentally do the same thing - I think the comment might be faster, but `typing.cast` does nothing at runtime (https://docs.python.org/3/library/typing.html#typing.cast) so it should be pretty negligible.
> 
> I would mainly only keep `typing.cast` in contexts where there are multiple type errors for a single variable that the static type checker doesn't understand - instead of adding a comment on every line where it's used, it might be better to just `cast` in that case.

Cool cool, yeah then in that case can you just add a comment explaining it here? I had to do a lot of digging to find a good justification for using cast anywhere (honestly have never seen it before haha, maybe I'm in the wrong python circles though) and I don't doubt future maintainers will have similar questions.

bari12 (approved):

Issue #6846, opened 2024-06-26
x509 auth: client does not read from configuration file whe creds are passed


rdimaio:
### Description

From https://github.com/rucio/rucio/pull/6838#issuecomment-2190913409:

> This PR introduced an annoying regression to `x509` authentication. The client reads from the configuration file only when `creds` is unset.
> 
> https://github.com/rucio/rucio/blob/0bbc069faf17f06eff1891aa90e634adea67b216/lib/rucio/client/baseclient.py#L264-L266

### Steps to reproduce

Attempt x509 auth with creds passed

PR #6848, opened 2024-06-26
Clients: Rework credential parsing logic for certs. Fixes #6845

### LINKED ISSUE #6845: Empty configuration option for client_x509_proxy, key, certificate results in cwd being used as client cert path
### Description

The docker clients image merges an existing configuration that as empty sections for many authentication related values:

https://github.com/rucio/containers/blob/48e664c5f58ba9a16dc1adfa35e9ee3f7b2c24dc/clients/rucio.default.cfg#L5-L11

When only some of the corresponding values are overriden, these options are empty in the resulting rucio.cfg.

When one then tries to override these options via the command line or environment variables, the options in the cfg still take precedence, are empty and because of code like this here:

https://github.com/rucio/rucio/blob/0bbc069faf17f06eff1891aa90e634adea67b216/lib/rucio/client/baseclient.py#L292-L309

this results in the current working directory being used as the certificate / key / proxy path.

This leads to very unhelpful errors when trying to authenticate like:

```
2024-06-25 08:50:53,863	ERROR	ConnectionError: ('Connection aborted.', IsADirectoryError(21, 'Is a directory'))
2024-06-25 08:50:53,866	ERROR	ConnectionError: ('Connection aborted.', IsADirectoryError(21, 'Is a directory'))
2024-06-25 08:50:53,868	ERROR	ConnectionError: ('Connection aborted.', IsADirectoryError(21, 'Is a directory'))
2024-06-25 08:50:53,868	ERROR	Cannot connect to the Rucio server.
```

or

```
2024-06-25 09:18:46,464	ERROR	Cannot authenticate.
Details: X.509 authentication selected, but private key (/home/user) permissions are liberal (required: 400 or 600, found: 755)
```

### Steps to reproduce

Start `rucio-clients` container with only some environmentvariables set. E.g. only the ones related to userpass authentication.

Then try to authenticate using one of the other methods.

### Rucio Version

_No response_

### Additional Information

_No response_

maxnoe:
Fixes #6845

### DIFF
--- a/lib/rucio/client/baseclient.py
+++ b/lib/rucio/client/baseclient.py
@@ -23,7 +23,7 @@ import secrets
 import sys
 import time
 from configparser import NoOptionError, NoSectionError
-from os import environ, fdopen, geteuid, makedirs, path
+from os import environ, fdopen, geteuid, makedirs
 from shutil import move
 from tempfile import mkstemp
 from typing import TYPE_CHECKING, Any, Optional
@@ -37,7 +37,7 @@ from requests.status_codes import codes
 
 from rucio import version
 from rucio.common import exception
-from rucio.common.config import config_get, config_get_bool, config_get_int
+from rucio.common.config import config_get, config_get_bool, config_get_int, config_has_section
 from rucio.common.exception import CannotAuthenticate, ClientProtocolNotFound, ClientProtocolNotSupported, ConfigNotFound, MissingClientParameter, MissingModuleException, NoAuthInformation, ServerConnectionException
 from rucio.common.extra import import_extras
 from rucio.common.utils import build_url, get_tmp_dir, my_key_generator, parse_response, setup_logger, ssh_sign
@@ -71,6 +71,14 @@ def choice(hosts):
     :return: A randomly selected host.
     """
     return secrets.choice(hosts)
+
+
+def _expand_path(path: str) -> str:
+    """Fully expand path, including ~ and env variables"""
+    path = path.strip()
+    if path == '':
+        return ''
+    return os.path.abspath(os.path.expanduser(os.path.expandvars(path)))
 
 
 class BaseClient:
@@ -173,7 +181,7 @@ class BaseClient:
             if self.ca_cert is None:
                 self.logger.debug('HTTPS is required, but no ca_cert was passed and X509_CERT_DIR is not defined. Trying to get it from the config file.')
                 try:
-                    self.ca_cert = path.expandvars(config_get('client', 'ca_cert'))
+                    self.ca_cert = _expand_path(config_get('client', 'ca_cert'))
                 except (NoOptionError, NoSectionError):
                     self.logger.debug('No ca_cert found in configuration. Falling back to Mozilla default CA bundle (certifi).')
                     self.ca_cert = True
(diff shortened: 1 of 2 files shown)


maxnoe:
How do we want to handle configured values that are invalid?

I would be in favor of raising an error instead of continuing to the next possible fallback in case something is set but not valid.

E.g. if `RUCIO_CLIENT_CERT` is set, but doens't point to a valid file, we raise an error instead of continuing to the next possible fallback.

maxnoe:
@rdimaio I just saw https://github.com/rucio/rucio/issues/6846 and the commits referenced therein, so you also started to refactor this part...

rdimaio:
> @rdimaio I just saw #6846 and the commits referenced therein, so you also started to refactor this part...

Yeah - I also think there might be some issues with the existing tests. In particular:

- this test expects `CannotAuthenticate` to be raised: https://github.com/rucio/rucio/blob/0bbc069faf17f06eff1891aa90e634adea67b216/tests/test_clients.py#L75-L81

But according to the code, it should be `MissingClientParameter`: https://github.com/rucio/rucio/blob/0bbc069faf17f06eff1891aa90e634adea67b216/lib/rucio/client/baseclient.py#L277-L278

- this test uses a `client_key` with permissions `644`: https://github.com/rucio/rucio/blob/0bbc069faf17f06eff1891aa90e634adea67b216/tests/test_clients.py#L67-L73, but the only allowed permission codes are `400` and `600`: https://github.com/rucio/rucio/blob/186354ce59567082c5526477c41533d96cf96fcb/lib/rucio/client/baseclient.py#L297-L298. This check is not performed in the current code since `creds` are passed. If you change the logic to always check for these permissions, it shows the test failure: https://github.com/rdimaio/rucio/actions/runs/9679508894/job/26708514046#step:4:1626. I'm still not 100% sure where this needs to be fixed

maxnoe:
@rdimaio In our tests for the different authentication methods in our setup, I also needed to make sure the rucio token is cleaned up before running a specific test, otherwise it would just reuse the existing token and not even try to authenticate using the given credentials.

Maybe the token should be stored per authentication method instead of just per user / per account?

See: https://gitlab.cta-observatory.org/cta-computing/dpps/dpps/-/blob/main/tests/bdms/test_basic_rucio_functionality.py?ref_type=heads#L26

rdimaio:
> @rdimaio In our tests for the different authentication methods in our setup, I also needed to make sure the rucio token is cleaned up before running a specific test, otherwise it would just reuse the existing token and not even try to authenticate using the given credentials.
> 
> Maybe the token should be stored per authentication method instead of just per user / per account?
> 
> See: https://gitlab.cta-observatory.org/cta-computing/dpps/dpps/-/blob/main/tests/bdms/test_basic_rucio_functionality.py?ref_type=heads#L26

Either option seems reasonable to me; I am more in favour of flushing the token before authentication, as it would be simpler to implement compared to storing and handling a token per authentication method. 

I think it would be best to get the opinion of the component leads for authentication and authorization - @dchristidis and @mlassnig, any thoughts?

voetberg:
@maxnoe Do you want to re-examine this now that #6856 has been completed?

bari12:
@maxnoe quick ping on this

maxnoe:
Will have a look again, sorry!

maxnoe:
I will now look into adapting the tests to address @rdimaio observations.

maxnoe:
The tests mentioned above by @rdimaio have been changed since the comments were made.

lib/rucio/client/baseclient.py
```diff
@@ -73,6 +73,14 @@ def choice(hosts):
     return secrets.choice(hosts)
 
 
+def _expand_path(p):
```
rdimaio: Please type annotate this function and use a more descriptive variable name

lib/rucio/client/baseclient.py
```diff
@@ -289,46 +297,61 @@ def _get_creds(self, creds: Optional[dict[str, Any]]) -> dict[str, Any]:
...
-                    if perms not in ['400', '600']:
-                        raise CannotAuthenticate('X.509 authentication selected, but private key (%s) permissions are liberal (required: 400 or 600, found: %s)' % (creds['client_key'], perms))
+                    raise MissingClientParameter('X.509 client key not found: %r' % creds['client_key'])
+
+                perms = oct(os.stat(creds['client_key']).st_mode)[-3:]
+                if perms not in ['400', '600']:
+                    raise CannotAuthenticate('X.509 authentication selected, but private key (%s) permissions are liberal (required: 400 or 600, found: %s)' % (creds['client_key'], perms))
 
             elif self.auth_type == 'x509_proxy':
+                # rucio speicific configuration takes precedence over GSI logic
```
rdimaio:
```suggestion
                # rucio specific configuration takes precedence over GSI logic
```

lib/rucio/client/baseclient.py
maxnoe: Done, though I changed more in this module due to it. The only pre-scriptive name I could think of was `path` and `os.path` was imported as `path`, which I already ran into while refactoring here.

So I removed this import and replaced usage of `path` with `os.path`. Let me now if this is ok or if I shoudl revert.

rdimaio (approved):

PR #6849, opened 2024-06-26
Core & Internals: Use maxLength instead of pattern, fix #6827

### LINKED ISSUE #6827: off-by-one error in database schema with respect to name jsonschema
### Description

The database schema uses a maxlength of `NAME_LENGTH`, while the default generic schema defines valid names as:

https://github.com/rucio/rucio/blob/ba102506d470c417fd2b136304e4fa4f7fc3a870/lib/rucio/common/schema/generic.py#L55-L59

which results in a maximum allowed length of `NAME_LENGTH + 1`

### Steps to reproduce

See above

### Rucio Version

current master and older

### Additional Information

_No response_

maxnoe:

### DIFF
--- a/lib/rucio/common/schema/atlas.py
+++ b/lib/rucio/common/schema/atlas.py
@@ -20,7 +20,8 @@ ACCOUNT_LENGTH = 25
 
 ACCOUNT = {"description": "Account name",
            "type": "string",
-           "pattern": "^[a-z0-9-_]{1,%s}$" % ACCOUNT_LENGTH}
+           "maxLength": ACCOUNT_LENGTH,
+           "pattern": "^[a-z0-9-_]+$"}
 
 ACCOUNTS = {"description": "Array of accounts",
             "type": "array",
@@ -48,7 +49,8 @@ SCOPE_LENGTH = 25
 
 SCOPE = {"description": "Scope name",
          "type": "string",
-         "pattern": "^[a-zA-Z_\\-.0-9]{1,%s}$" % SCOPE_LENGTH}
+         "maxLength": SCOPE_LENGTH,
+         "pattern": "^[a-zA-Z_\\-.0-9]+$"}
 
 R_SCOPE = {"description": "Scope name",
            "type": "string",
@@ -58,7 +60,8 @@ NAME_LENGTH = 250
 
 NAME = {"description": "Data Identifier name",
         "type": "string",
-        "pattern": "^[A-Za-z0-9][A-Za-z0-9\\.\\-\\_]{1,%s}$" % NAME_LENGTH}
+        "maxLength": NAME_LENGTH,
+        "pattern": "^[A-Za-z0-9][A-Za-z0-9\\.\\-\\_]*$"}
 
 R_NAME = {"description": "Data Identifier name",
           "type": "string",
--- a/lib/rucio/common/schema/belleii.py
+++ b/lib/rucio/common/schema/belleii.py
@@ -20,7 +20,8 @@ ACCOUNT_LENGTH = 25
 
 ACCOUNT = {"description": "Account name",
            "type": "string",
-           "pattern": "^[a-z0-9-_]{1,%s}$" % ACCOUNT_LENGTH}
+           "maxLength": ACCOUNT_LENGTH,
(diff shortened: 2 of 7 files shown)


lib/rucio/common/schema/atlas.py
```diff
@@ -58,7 +60,8 @@
 
 NAME = {"description": "Data Identifier name",
         "type": "string",
-        "pattern": "^[A-Za-z0-9][A-Za-z0-9\\.\\-\\_]{1,%s}$" % NAME_LENGTH}
+        "maxLength": NAME_LENGTH,
+        "pattern": "^[A-Za-z0-9][A-Za-z0-9\\.\\-\\_]*$"}
```
rdimaio, line 64: just to clarify - is there a reason why there's `*` at the end of the pattern here, compared to having `+` like for `SCOPE` and `ACCOUNT`?

for reference:
- `* matches the previous token between zero and unlimited times, as many times as possible, giving back as needed (greedy)`
- `+ matches the previous token between one and unlimited times, as many times as possible, giving back as needed (greedy)`

maxnoe: I defined all regexes in a way that they require at least 1 character to be present.

So for regexes that define a single character class, I used `+`.

For regexes that use a smaller, initial character set for the first letter and then a wider range for any following characters, I used `*`

rdimaio (commented):
`test_belle2_schema` fails: https://github.com/rucio/rucio/actions/runs/9681101823/job/26754980743?pr=6849

The `extract_scope` output for the two test cases in the test:

```
('other', 'invalid_name')
('other', '/belle/invalid@/did')
```



Using the old schema:

```
'invalid_name' does not match '^/[A-Za-z0-9\\.\\-\\_\\+\\/]{1,250}$'

Failed validating 'pattern' in schema['properties']['name']:
    {'description': 'Data Identifier name',
     'pattern': '^/[A-Za-z0-9\\.\\-\\_\\+\\/]{1,250}$',
     'type': 'string'}

On instance['name']:
    'invalid_name'
'/belle/invalid@/did' does not match '^/[A-Za-z0-9\\.\\-\\_\\+\\/]{1,250}$'

Failed validating 'pattern' in schema['properties']['name']:
    {'description': 'Data Identifier name',
     'pattern': '^/[A-Za-z0-9\\.\\-\\_\\+\\/]{1,250}$',
     'type': 'string'}

On instance['name']:
    '/belle/invalid@/did'
```

Using the updated schema, `/belle/invalid@/did` is considered to be a valid name

lib/rucio/common/schema/belleii.py
```diff
@@ -56,7 +58,8 @@
 
 NAME = {"description": "Data Identifier name",
         "type": "string",
-        "pattern": r"^/[A-Za-z0-9\.\-\_\+\/]{1,%s}$" % NAME_LENGTH}
+        "maxLength": NAME_LENGTH,
+        "pattern": r"^/[A-Za-z0-9\.\-\_\+\/]+"}
```
rdimaio:
```suggestion
        "pattern": r"^/[A-Za-z0-9\.\-\_\+\/]+$"}
```

This fixes the test:

```
'invalid_name' does not match '^/[A-Za-z0-9\\.\\-\\_\\+\\/]+$'

Failed validating 'pattern' in schema['properties']['name']:
    {'description': 'Data Identifier name',
     'maxLength': 250,
     'pattern': '^/[A-Za-z0-9\\.\\-\\_\\+\\/]+$',
     'type': 'string'}

On instance['name']:
    'invalid_name'
'/belle/invalid@/did' does not match '^/[A-Za-z0-9\\.\\-\\_\\+\\/]+$'

Failed validating 'pattern' in schema['properties']['name']:
    {'description': 'Data Identifier name',
     'maxLength': 250,
     'pattern': '^/[A-Za-z0-9\\.\\-\\_\\+\\/]+$',
     'type': 'string'}

On instance['name']:
    '/belle/invalid@/did'
```

maxnoe: Thanks!

rdimaio (approved):

bari12 (approved):
Thanks @maxnoe

bari12:
I will merge this for the 35 release though

Issue #6832, opened 2024-06-18
Kronos does not update the access_cnt for files


cserf:
### Description

Currently `kronos` update `accessed_at` but not `access_cnt` : https://github.com/rucio/rucio/blob/master/lib/rucio/core/replica.py#L2428
Would be good to update this value too

harsha-mangena:
Hey @cserf, Can I work on this issue.

PR #6851, opened 2024-06-26
[Fixes]Kronos update at `access_cnt` for file

harsha-mangena:
- refer #https://github.com/rucio/rucio/issues/6832
- added `access_cnt` param to maintain count as retrived.

### DIFF
--- a/tests/test_replica.py
+++ b/tests/test_replica.py
@@ -31,7 +31,7 @@ from rucio.common.exception import AccessDenied, DatabaseException, DataIdentifi
 from rucio.common.schema import get_schema_value
 from rucio.common.utils import clean_surls, generate_uuid, parse_response
 from rucio.core.config import set as cconfig_set
-from rucio.core.did import add_did, attach_dids, get_did, get_did_atime, list_files, set_status
+from rucio.core.did import add_did, attach_dids, get_did, get_did_atime, get_did_access_cnt, list_files, set_status
 from rucio.core.replica import add_bad_dids, add_replica, add_replicas, delete_replicas, get_bad_pfns, get_replica, get_replica_atime, get_replicas_state, get_RSEcoverage_of_dataset, list_replicas, set_tombstone, touch_replica, update_replica_state
 from rucio.core.rse import add_protocol, add_rse_attribute, del_rse_attribute
 from rucio.daemons.badreplicas.minos import minos
@@ -307,14 +307,14 @@ class TestReplicaCore:
 
         assert get_replica_atime({'scope': files1[0]['scope'], 'name': files1[0]['name'], 'rse_id': rse_id}) is None
         assert get_did_atime(scope=mock_scope, name=files1[0]['name']) is None
-        assert get_replica(rse_id=rse_id, scope=files1[0]['scope'], name=files1[0]['name'])['access_cnt'] == 0
+        assert 0 == get_did_access_cnt(scop=mock_scope, name=files1[0]['name'])
 
         for r in [{'scope': files1[0]['scope'], 'name': files1[0]['name'], 'rse_id': rse_id, 'accessed_at': now}]:
             touch_replica(r)
 
         assert now == get_replica_atime({'scope': files1[0]['scope'], 'name': files1[0]['name'], 'rse_id': rse_id})
         assert now == get_did_atime(scope=mock_scope, name=files1[0]['name'])
-        assert 1 == get_replica(rse_id=rse_id, scope=files1[0]['scope'], name=files1[0]['name'])['access_cnt']
+        assert 1 == get_did_access_cnt(scop=mock_scope, name=files1[0]['name'])
 
         for i in range(1, nbfiles):
             assert get_replica_atime({'scope': files1[i]['scope'], 'name': files1[i]['name'], 'rse_id': rse_id}) is None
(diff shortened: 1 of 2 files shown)


rdimaio (changes requested):
Thanks for the PR :)

The CI tests are failing:

```
Found 1 new problems in /home/runner/work/rucio/rucio/lib/rucio/core/replica.py
  - 1 errors with message """Cannot access attribute "access_cnt" for class "type[RSEFileAssociation]"
                               Attribute "access_cnt" is unknown""".
    Candidates: line 2413
```

```
 tests/test_replica.py:310: in test_touch_replicas
    assert get_replica(rse_id=rse_id, scope=files1[0]['scope'], name=files1[0]['name'])['access_cnt'] == 0
E   KeyError: 'access_cnt'
``` 


The reason is that `access_cnt` is an attribute in models.DataIdentifier, and the update statement for that is here:

https://github.com/rucio/rucio/blob/0bbc069faf17f06eff1891aa90e634adea67b216/lib/rucio/core/replica.py#L2424-L2428

So this is where you should put the access_cnt increment.

tests/test_replica.py
```diff
@@ -307,12 +307,14 @@ def test_touch_replicas(self, rse_factory, mock_scope, root_account):
 
         assert get_replica_atime({'scope': files1[0]['scope'], 'name': files1[0]['name'], 'rse_id': rse_id}) is None
         assert get_did_atime(scope=mock_scope, name=files1[0]['name']) is None
+        assert 0 == get_did_access_cnt(scop=mock_scope, name=files1[0]['name'])
```
rdimaio:
```suggestion
        assert get_did_access_cnt(scope=mock_scope, name=files1[0]['name']) == 0
```

tests/test_replica.py
```diff
@@ -307,12 +307,14 @@ def test_touch_replicas(self, rse_factory, mock_scope, root_account):
...
 
         for r in [{'scope': files1[0]['scope'], 'name': files1[0]['name'], 'rse_id': rse_id, 'accessed_at': now}]:
             touch_replica(r)
 
         assert now == get_replica_atime({'scope': files1[0]['scope'], 'name': files1[0]['name'], 'rse_id': rse_id})
         assert now == get_did_atime(scope=mock_scope, name=files1[0]['name'])
+        assert 1 == get_did_access_cnt(scop=mock_scope, name=files1[0]['name'])
```
rdimaio:
```suggestion
        assert get_did_access_cnt(scope=mock_scope, name=files1[0]['name']) == 1
```

tests/test_replica.py
```diff
@@ -31,7 +31,7 @@
 from rucio.common.schema import get_schema_value
 from rucio.common.utils import clean_surls, generate_uuid, parse_response
 from rucio.core.config import set as cconfig_set
-from rucio.core.did import add_did, attach_dids, get_did, get_did_atime, list_files, set_status
+from rucio.core.did import add_did, attach_dids, get_did, get_did_atime, get_did_access_cnt, list_files, set_status
```
rdimaio: Please sort the imports to pass the linter check (you can do this automatically via Ruff, see https://rucio.github.io/documentation/contributing#1-prerequisite)

rdimaio (changes requested):

tests/test_replica.py
```diff
@@ -307,12 +307,14 @@ def test_touch_replicas(self, rse_factory, mock_scope, root_account):
 
         assert get_replica_atime({'scope': files1[0]['scope'], 'name': files1[0]['name'], 'rse_id': rse_id}) is None
         assert get_did_atime(scope=mock_scope, name=files1[0]['name']) is None
+        assert get_did_access_cnt(scope=mock_scope, name=files1[0]['name']) == 0
```
rdimaio, line 310:
```suggestion
        assert get_did_access_cnt(scope=mock_scope, name=files1[0]['name']) is None
```

Currently this test is failing:
```
 _____________________ TestReplicaCore.test_touch_replicas ______________________
tests/test_replica.py:310: in test_touch_replicas
    assert get_did_access_cnt(scope=mock_scope, name=files1[0]['name']) == 0
E   AssertionError: assert None == 0
E    +  where None = get_did_access_cnt(scope=mock, name='file_fb6e29bb3bb44585931fd624261e3e9f')
```

My understanding is that the DID is not initialized with `access_cnt = 0`, so `access_cnt` would be `None`. 
This seems to be the case for the `accessed_at` times for the DID and replica, as seen from the assertions on line 308-309.

The `values(accessed_at=accessed_at, access_cnt=models.DataIdentifier.access_cnt + 1)` could potentially fail with `TypeError: unsupported operand type(s) for +: 'NoneType' and 'int'`. You should change that logic in `touch_replica` to match the one in `touch_dids`:

https://github.com/rucio/rucio/blob/b6854f44e1e3bf003497256aa0a1c6dc3e00628c/lib/rucio/core/did.py#L2865-L2866

But I'm also having doubts on whether it's a good idea to increment the `access_cnt` on the DID as part of `touch_replica` - maybe the DID `access_cnt` should only be incremented by `touch_dids`, and a separate replica `access_cnt` field should be added to the `RSEFileAssociation` and `CollectionReplica` tables, considering that they already have an `accessed_at` field. @bari12 thoughts?

rdimaio (changes requested):
See comment

PR #6854, opened 2024-06-27
Auth: Use OpenID Connect Discovery #6414

### LINKED ISSUE #6414: Use OpenID Connect Discovery
### Description

The current new-style token support builds the token endpoint URL by concatenating the issuer from the configuration file with the hard-coded `/token` path. Instead, it should make use of OIDC Discovery (i.e. `/.well-known/openid-configuration`).

### Motivation

_No response_

### Change

_No response_

dchristidis:
Resolves #6414.

### DIFF
--- a/lib/rucio/core/oidc.py
+++ b/lib/rucio/core/oidc.py
@@ -243,13 +243,22 @@ def __load_oidc_configuration() -> bool:
             data = json.load(f)
             OIDC_CLIENT_ID = data[ADMIN_ISSUER_ID]['client_id']
             OIDC_CLIENT_SECRET = data[ADMIN_ISSUER_ID]['client_secret']
-            OIDC_PROVIDER_ENDPOINT = urljoin(data[ADMIN_ISSUER_ID]['issuer'], 'token')
+            issuer = data[ADMIN_ISSUER_ID]['issuer']
     except Exception:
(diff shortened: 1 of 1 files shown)


lib/rucio/core/oidc.py
```diff
@@ -243,13 +243,22 @@ def __load_oidc_configuration() -> bool:
...
-        return True
+    try:
+        oidc_discover_url = urljoin(issuer, '.well-known/openid-configuration')
+        response = requests.get(oidc_discover_url)
+        response.raise_for_status()
+        payload = response.json()
+        OIDC_PROVIDER_ENDPOINT = payload['token_endpoint']
+    except (requests.HTTPError, requests.JSONDecodeError, KeyError):
+        logging.error('Failed to discover token endpoint', exc_info=True)
+        return False
```
rdimaio, line 259: as an idea: should it fall back to setting the endpoint to `urljoin(data[ADMIN_ISSUER_ID]['issuer'], 'token')` here, in case it fails to discover it in via the OIDC discover URL?

rdimaio (approved):

lib/rucio/core/oidc.py
dchristidis: I’m not sure whether `/token` is more common than any other path. I find it preferable to make OIDC Discovery a prerequisite.

Issue #6640, opened 2024-04-05
Finalize SQLAlchemy 1.0 -> 2.0 syntax migration


bari12:
### Description

Individual changes followed up in individual tickets.

Issue #6718, opened 2024-04-23
Remove `print` calls in tests


rdimaio:
Doing a quick search for `print(` in `tests/` returns more than 700 instances of printing in tests, e.g. as seen here:

https://github.com/rucio/rucio/blob/767dbc56e5540abf7c9b9f5d84b5c2ae4b848643/tests/test_replica_recoverer.py#L206-L217

Printing is useful for debugging, but once a test has been written and has the proper assertions, print statements should likely be removed.

`print` affects performance, as tested here: https://stackoverflow.com/questions/13288185/performance-effect-of-using-print-statements-in-python-script - we might be able to gain some performance in test speed by removing these calls.

rdimaio:
https://github.com/rucio/rucio/pull/7818

bari12:
Thank you for raising this. We have reviewed this issue as part of our regular issue review.

We see value in this request, but we do not have the capacity to work on it within the next twelve months.

In line with our [issue management guideline](https://rucio.cern/documentation/developer/issue_management), we are labelling it `wishlist` and closing it. This does not mean it is forgotten: wishlist items are revisited regularly during component reviews and roadmap planning, and this issue can be re-opened and added back to the pipeline as soon as interest and capacity align.

PR #6855, opened 2024-06-27
Testing: refactor tests to sqlalchemy2. #6640

erlingstaff:
Part of issue #6640

This PR updates the sqlalchemy queries in `tests/` to the new sqlalchemy2 standard.

### DIFF
--- a/tests/test_conveyor.py
+++ b/tests/test_conveyor.py
@@ -20,7 +20,7 @@ from unittest.mock import patch
 from urllib.parse import parse_qsl, urlencode, urlparse, urlunparse
 
 import pytest
-from sqlalchemy import update
+from sqlalchemy import and_, delete, select, update
 
 import rucio.daemons.reaper.reaper
 from rucio.common.constants import RseAttr
@@ -60,7 +60,16 @@ TEST_FTS_HOST = 'https://fts:8446'
 
 @transactional_session
 def __update_request(request_id, *, session=None, **kwargs):
-    session.query(models.Request).filter_by(id=request_id).update(kwargs, synchronize_session=False)
+    stmt = update(
+        models.Request
+    ).where(
+        models.Request.id == request_id
+    ).execution_options(
+        synchronize_session=False
+    ).values(
+        kwargs
+    )
+    session.execute(stmt)
 
 
 def __wait_for_replica_transfer(dst_rse_id, scope, name, max_wait_seconds=MAX_POLL_WAIT_SECONDS, transfertool=None):
@@ -119,12 +128,15 @@ def set_query_parameters(url, params):
 
 @read_session
 def __get_source(request_id, src_rse_id, scope, name, *, session=None):
-    return session.query(models.Source) \
-        .filter(models.Source.request_id == request_id) \
-        .filter(models.Source.scope == scope) \
-        .filter(models.Source.name == name) \
-        .filter(models.Source.rse_id == src_rse_id) \
-        .first()
+    stmt = select(
+        models.Source
+    ).where(
+        and_(models.Source.request_id == request_id,
+             models.Source.scope == scope,
+             models.Source.name == name,
+             models.Source.rse_id == src_rse_id)
+    )
+    return session.execute(stmt).scalar()
 
 
 @pytest.fixture
@@ -254,8 +266,19 @@ def test_multihop_intermediate_replica_lifecycle(vo, did_factory, root_account, 
 
         @transactional_session
         def _cleanup_all_usage_and_limits(rse_id, *, session=None):
-            session.query(models.RSELimit).filter_by(rse_id=rse_id).delete()
-            session.query(models.RSEUsage).filter_by(rse_id=rse_id, source='storage').delete()
+            stmt = delete(
+                models.RSELimit
+            ).where(
+                models.RSELimit.rse_id == rse_id
+            )
+            session.execute(stmt)
+            stmt = delete(
+                models.RSEUsage
+            ).where(
+                and_(models.RSEUsage.rse_id == rse_id,
+                     models.RSEUsage.source == 'storage')
+            )
+            session.execute(stmt)
 
         _cleanup_all_usage_and_limits(rse_id=jump_rse_id)
 
@@ -387,11 +410,14 @@ def test_multisource(vo, did_factory, root_account, replica_client, caches_mock,
 
     @read_session
     def __source_exists(src_rse_id, scope, name, *, session=None):
-        return session.query(models.Source) \
-            .filter(models.Source.rse_id == src_rse_id) \
-            .filter(models.Source.scope == scope) \
-            .filter(models.Source.name == name) \
-            .count() != 0
+        stmt = select(
+            models.Source
+        ).where(
+            and_(models.Source.rse_id == src_rse_id,
+                 models.Source.scope == scope,
+                 models.Source.name == name)
+        )
+        return session.execute(stmt).scalar_one_or_none() is not None
 
     # Entries in the source table must be created for both sources of the multi-source transfer
     assert __source_exists(src_rse_id=src_rse1_id, **did)
@@ -780,11 +806,9 @@ def test_preparer_throttler_submitter(rse_factory, did_factory, root_account, fi
             models.Request
         ).where(
             models.Request.id.in_(request_ids)
-        ).values(
-            {
-                models.Request.last_processed_at: timestamp
-            }
-        )
+        ).values({
+            models.Request.last_processed_at: timestamp
+        })
         session.execute(stmt)
 
     __set_process_timestamp([request['id'] for request in [request3, request4]], datetime.utcnow() - timedelta(days=2))
--- a/tests/test_judge_repairer.py
+++ b/tests/test_judge_repairer.py
@@ -17,6 +17,7 @@ from hashlib import sha256
 
 import pytest
 from dogpile.cache import make_region
+from sqlalchemy import and_, select
 
 from rucio.common.config import config_get, config_get_bool
 from rucio.common.types import InternalAccount, InternalScope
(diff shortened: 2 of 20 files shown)


tests/test_conveyor.py
```diff
@@ -387,11 +410,16 @@ def test_multisource(vo, did_factory, root_account, replica_client, caches_mock,
...
+        stmt = select(
+            func.count()
+        ).select_from(
+            models.Source
+        ).where(
+            and_(models.Source.rse_id == src_rse_id,
+                 models.Source.scope == scope,
+                 models.Source.name == name)
+        )
+        return session.execute(stmt).scalar() != 0
```
rdimaio: The `count` is not really needed in this query - I think it would be a better fit to check for `scalar_one_or_none` and checking if it's `None`, like done here https://github.com/rucio/rucio/blob/0bbc069faf17f06eff1891aa90e634adea67b216/lib/rucio/core/rule.py#L837

tests/test_meta_conventions.py
```diff
@@ -98,8 +99,13 @@ def test_add_key(self, rucio_client):
...
             key_name = 'datatype%s' % str(uuid())
             rucio_client.add_key(key_name, key_type['type'])
-            stored_key_type = session.get_session().query(models.DIDMetaConventionsKey).filter_by(key=key_name).one()['key_type']
-            assert stored_key_type, key_type['expected']
+            stmt = select(
+                models.DIDMetaConventionsKey
+            ).where(
+                models.DIDMetaConventionsKey.key == key_name
+            )
+            stored_key_type = session.get_session().execute(stmt).scalar_one()
```
rdimaio: To be more accurate, I think it should either be:

```
stored_key = session.get_session().execute(stmt).scalar_one()
```

or

```
stored_key_type = session.get_session().execute(stmt).scalar_one()['key_type']
```

tests/test_meta_conventions.py
```diff
@@ -130,8 +136,13 @@ def test_add_key():
...
         key_name = 'datatype%s' % str(uuid())
         add_key(key_name, key_type['type'])
-        stored_key_type = session.get_session().query(models.DIDMetaConventionsKey).filter_by(key=key_name).one()['key_type']
-        assert stored_key_type, key_type['expected']
+        stmt = select(
+            models.DIDMetaConventionsKey
+        ).where(
+            models.DIDMetaConventionsKey.key == key_name
+        )
+        stored_key_type = session.get_session().execute(stmt).scalar_one()
```
rdimaio: To be more accurate, I think it should either be:

```
stored_key = session.get_session().execute(stmt).scalar_one()
```

or

```
stored_key_type = session.get_session().execute(stmt).scalar_one()['key_type']
```

tests/test_multi_vo.py
```diff
@@ -881,16 +882,17 @@ def test_account_counters_at_different_vos(self, vo, second_vo, rse_client):
...
+            models.AccountUsage.account,
+            models.AccountUsage.rse_id
+        ).distinct(
+        ).where(
+            models.AccountUsage.account == new_acc_vo2
+        )
+        acc_counters = session.execute(stmt).one()
+        rse_id = acc_counters[1]
+        vo = get_rse_vo(rse_id)
+        assert vo == second_vo
```
rdimaio, line 895: `assert 1 == len(acc_counters)` has been removed; looking at the test, I don't think the assertion is necessary (since we're filtering by `account == new_acc_vo2`, and there's only one counter that matches that condition).

However, I think we should keep this assertion for now in order to maintain the same test functionality and to keep the scope of the PR purely to the SQLA migration.

tests/test_oauthmanager.py
```diff
@@ -76,27 +76,50 @@ def save_oidc_token(account, lifetime_access=0, lifetime_refresh=0, refresh_toke
...
-    result = session.query(models.Token).filter_by(account=account).all()  # pylint: disable=no-member
+    stmt = select(
+        models.Token
+    ).where(
+        models.Token.account == account
+    )
+    result = session.execute(stmt).scalars().all()
     for token in result:
         print(token.token, token.expired_at, token.refresh_token, token.refresh_expired_at, token.oidc_scope)
     return len(result)
```
rdimaio: To be consistent with the changes to `get_oauth_session_param_count`, this could be changed to use `func_count()` instead of doing `len(result)`

tests/test_message.py
```diff
@@ -115,14 +116,12 @@ def test_large_payload(core_config_mock, caches_mock):
...
-        .first()
+    stmt = select(
+        Message
+    ).where(
+        Message.event_type == event_type
+    )
+    msg = session.execute(stmt).scalar()
     assert msg.payload == 'nolimit'
     assert msg.payload_nolimit == json.dumps(dict_long_payload)
     messages = retrieve_messages(40)
```
rdimaio: Based on the assertions I think this could be simplified to just querying `payload` and `payload_nolimit` instead

tests/test_judge_repairer.py
```diff
@@ -319,15 +328,15 @@ def change_availability(new_value):
 
             if ignore_availability:
                 change_availability(False)
-                rule_id = add_rule(dids=[{'scope': scope, 'name': dataset}], account=self.jdoe, copies=1, rse_expression=rse, grouping=grouping, weight=None, lifetime=None, locked=False, subscription_id=None, ignore_availability=ignore_availability, activity='DebugJudge')[0]
+                rule_id = add_rule(dids=[{'scope': scope, 'name': dataset}], account=self.jdoe, copies=1, rse_expression=rse, grouping=grouping, weight=None, lifetime=None, locked=False, subscription_id=None, ignore_availability=ignore_availability, activity='DebugJudge')[0]  # NOQA: E501
```
rdimaio: It would be better to break this line down into multiple lines instead of adding `NOQA: E501`

tests/test_judge_repairer.py
```diff
@@ -319,15 +328,15 @@ def change_availability(new_value):
...
                 assert (RuleState.STUCK == get_rule(rule_id)['state'])
 
                 rule_repairer(once=True)
                 assert (RuleState.REPLICATING == get_rule(rule_id)['state'])
 
                 change_availability(True)
             else:
-                rule_id = add_rule(dids=[{'scope': scope, 'name': dataset}], account=self.jdoe, copies=1, rse_expression=rse, grouping=grouping, weight=None, lifetime=None, locked=False, subscription_id=None, ignore_availability=ignore_availability, activity='DebugJudge')[0]
+                rule_id = add_rule(dids=[{'scope': scope, 'name': dataset}], account=self.jdoe, copies=1, rse_expression=rse, grouping=grouping, weight=None, lifetime=None, locked=False, subscription_id=None, ignore_availability=ignore_availability, activity='DebugJudge')[0]  # NOQA: E501
```
rdimaio: It would be better to break this line down into multiple lines instead of adding `NOQA: E501`

rdimaio (changes requested):

tests/test_multi_vo.py
erlingstaff: `.one()` from `acc_counters = session.execute(stmt).one()` raises if the result is not exactly 1 row:

```
Return exactly one row or raise an exception.

Raises .NoResultFound if the result returns no
rows, or .MultipleResultsFound if multiple rows would be returned.
```

I can include the assertion again, but it wouldn't add any functionality

rdimaio: Ah nice, I didn't know! I'll resolve this

tests/test_oauthmanager.py
erlingstaff: If we replace this with a `func.count()` then we wouldn't be able to index the columns for the print statement (which may or may not be important, I don't know). It is possible to do something like:

```
def get_token_count(account):
    session = get_session()
    stmt = select(
        models.Token,
        func.count().over().label('token_count')
    ).select_from(
        models.Token
    ).where(
        models.Token.account == account
    )
    result = session.execute(stmt).all()
    for token, count in result:
        print(token.token, token.expired_at, token.refresh_token, token.refresh_expired_at, token.oidc_scope)
    return count
```

This also carries some overhead as the column 'label_count' is included in every row. It is also possible to do two queries, one for the count and one for the rows, this will have some overhead too. Maybe the best would be to remove the prints? It does not show in the test output nor in `logshow`

rdimaio: I'd be in favour of removing the print statement here (related: https://github.com/rucio/rucio/issues/6718)

erlingstaff: Removed the prints and refactored this to use func.count() instead in a separate commit - 00cffa126f49b448df14231361e4b7b4a36fa979,

Now:
```
def get_token_count(account):
    session = get_session()
    stmt = select(
        func.count()
    ).select_from(
        models.Token
    ).where(
        models.Token.account == account
    )
    return session.execute(stmt).scalar()
```

tests/test_conveyor.py
erlingstaff: Changed, now: 

```
    @read_session
    def __source_exists(src_rse_id, scope, name, *, session=None):
        stmt = select(
            models.Source
        ).where(
            and_(models.Source.rse_id == src_rse_id,
                 models.Source.scope == scope,
                 models.Source.name == name)
        )
        return session.execute(stmt).scalar_one_or_none() is not None
```

tests/test_meta_conventions.py
erlingstaff: Changed to:

```
stored_key = session.get_session().execute(stmt).scalar_one()
assert stored_key['key_type'], key_type['expected']
```

tests/test_meta_conventions.py
erlingstaff: Changed to:

```
stored_key = session.get_session().execute(stmt).scalar_one()
assert stored_key['key_type'], key_type['expected']
```

tests/test_message.py
erlingstaff: Changed to:

```
    stmt = select(
        Message.payload,
        Message.payload_nolimit
    ).where(
        Message.event_type == event_type
    )
    result = session.execute(stmt).first()
    assert result is not None
    assert result.payload == 'nolimit'
    assert result.payload_nolimit == json.dumps(dict_long_payload)
    messages = retrieve_messages(40)
    assert messages[0]['payload'] == dict_long_payload
```

Added another assert to keep Pylance happy as `.first()` might return None

tests/test_judge_repairer.py
erlingstaff: Changed, now the arguments are on separate lines

tests/test_judge_repairer.py
erlingstaff: Changed, now the arguments are on separate lines

rdimaio (approved):

Issue #6839, opened 2024-06-20
add tests to ensure different authentication methods work correctly


rdimaio:
Related: https://github.com/rucio/rucio/issues/6822

`test_bin_rucio` should have tests related to the different authentication methods, currently missing: https://github.com/rucio/rucio/blob/ba102506d470c417fd2b136304e4fa4f7fc3a870/tests/test_bin_rucio.py

rdimaio:
Can you elaborate on why this is not planned?

rdimaio:
Re-opening for discussion - we can close it once it's clear why it should be closed

bari12:
We closed this because it's just not worth doing at this time given the amount of other work which needs to be done. Was just a pure prioritisation question.

rdimaio:
> We closed this because it's just not worth doing at this time given the amount of other work which needs to be done. Was just a pure prioritisation question.

I think we should still keep it open and just not work on it for now, then. We had a bug that directly happened as a consequence of this (https://github.com/rucio/rucio/issues/6822), so if we close this and forget about ever doing this testing we might end up facing more bugs of this kind

bari12:
Thank you for raising this. We have reviewed this issue as part of our regular issue review.

We see value in this request, but we do not have the capacity to work on it within the next twelve months.

In line with our [issue management guideline](https://rucio.cern/documentation/developer/issue_management), we are labelling it `wishlist` and closing it. This does not mean it is forgotten: wishlist items are revisited regularly during component reviews and roadmap planning, and this issue can be re-opened and added back to the pipeline as soon as interest and capacity align.

PR #6856, opened 2024-06-27
Auth: Refactor BaseClient._get_creds to read from config even if creds are set

### LINKED ISSUE #6846: x509 auth: client does not read from configuration file whe creds are passed

rdimaio:
Fix #6846

### DIFF
--- a/lib/rucio/client/baseclient.py
+++ b/lib/rucio/client/baseclient.py
@@ -239,57 +239,62 @@ class BaseClient:
         return auth_type
 
     def _get_creds(self, creds: Optional[dict[str, Any]]) -> dict[str, Any]:
-        if self.auth_type == 'oidc':
-            if not creds:
-                creds = {}
-            # if there are default values, check if rucio.cfg does not specify them, otherwise put default
-            if 'oidc_refresh_lifetime' not in creds or creds['oidc_refresh_lifetime'] is None:
-                creds['oidc_refresh_lifetime'] = config_get('client', 'oidc_refresh_lifetime', False, None)
-            if 'oidc_issuer' not in creds or creds['oidc_issuer'] is None:
-                creds['oidc_issuer'] = config_get('client', 'oidc_issuer', False, None)
-            if 'oidc_audience' not in creds or creds['oidc_audience'] is None:
-                creds['oidc_audience'] = config_get('client', 'oidc_audience', False, None)
-            if 'oidc_auto' not in creds or creds['oidc_auto'] is False:
-                creds['oidc_auto'] = config_get_bool('client', 'oidc_auto', False, False)
-            if creds['oidc_auto']:
-                if 'oidc_username' not in creds or creds['oidc_username'] is None:
-                    creds['oidc_username'] = config_get('client', 'oidc_username', False, None)
-                if 'oidc_password' not in creds or creds['oidc_password'] is None:
-                    creds['oidc_password'] = config_get('client', 'oidc_password', False, None)
-            if 'oidc_scope' not in creds or creds['oidc_scope'] == 'openid profile':
-                creds['oidc_scope'] = config_get('client', 'oidc_scope', False, 'openid profile')
-            if 'oidc_polling' not in creds or creds['oidc_polling'] is False:
-                creds['oidc_polling'] = config_get_bool('client', 'oidc_polling', False, False)
-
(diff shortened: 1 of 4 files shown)


maxnoe:
Can we refactor this a bit further so that

* All options have a clear precedence of (from lowest to highest priority):
    * default value
    * configuration file
    * environment variable
    * `creds` argument

* take care that when checking the configuration file, no `No such section` error is thrown?
   I think for this, it might be easiest if the `config_get(section, name, default)` function would just return `default` in case the section doesn't exist instead of raising an error.


When using the cli or inside a docker container, it would be nice if rucio wouldn't need a configuration file at all when everything is defined either as arguments or via env variables. 


As far as I can tell there are a couple of options for which there are neither env variables nor cli options, just the configuration file value.

rdimaio:
> Can we refactor this a bit further so that
> 
>     * All options have a clear precedence of (from lowest to highest priority):
>       
>       * default value
>       * configuration file
>       * environment variable
>       * `creds` argument
> 
>     * take care that when checking the configuration file, no `No such section` error is thrown?
>       I think for this, it might be easiest if the `config_get(section, name, default)` function would just return `default` in case the section doesn't exist instead of raising an error.
> 
> 
> When using the cli or inside a docker container, it would be nice if rucio wouldn't need a configuration file at all when everything is defined either as arguments or via env variables.
> 
> As far as I can tell there are a couple of options for which there are neither env variables nor cli options, just the configuration file value.

I agree, but I think it might be best to do this in a follow-up PR so that we can merge your changes from https://github.com/rucio/rucio/pull/6848 first

maxnoe:
@rdimaio I wasn't yet totally sure how to implement #6848 properly, i.e. the error handling in case something is empty or not valid. So it has a large overlap.

I thought maybe doing this on one go, as one has to basically touch all lines here in this _get_creds function might be easiest.

lib/rucio/client/baseclient.py
```diff
@@ -239,57 +239,62 @@ def _get_auth_type(self, auth_type: Optional[str]) -> str:
...
+        if not creds:
             self.logger.debug('No creds passed. Trying to get it from the config file.')
             creds = {}
-            try:
-                if self.auth_type in ['userpass', 'saml']:
+
+        try:
+            if self.auth_type == 'oidc':
+                # if there are default values, check if rucio.cfg does not specify them, otherwise put default
+                if 'oidc_refresh_lifetime' not in creds or creds['oidc_refresh_lifetime'] is None:
```
voetberg, line 249: All of these if "key" not in creds or creds['key'] is None could be replaced with 

```
if creds.get('key') is None: 
```

For legibility

voetberg: This whole oidc auth block could also probably be a for loop, just looping through a list of the keys - A debug log would be able to track where errors happened in this case. The whole block is just a lot to read.

voetberg (changes requested):
I largely agree with @maxnoe 's comments here - There should be a clear logic for order of preference on the auth variables (I was having a hard time with #6839 for this exact reason). 

Maybe a better solution would be to have a smaller helper function that went through and checked each place a single auth variable can be found in order to keep _get_creds form exploding into hydra of a function

bari12:
Should we split this up into two different issues? I would like to fix the regression in the upcoming release.

rdimaio:
> Should we split this up into two different issues? I would like to fix the regression in the upcoming release.

I think it would be best to merge @maxnoe's PR first: https://github.com/rucio/rucio/pull/6848 - I can then adapt my PR to that one and address @voetberg 's comments

bari12:
I am merging this now, because the fix should be included in the 35.0.0 release. But please let's continue to include the other points mentioned in @maxnoe PR #6848 for the following release.

bari12 (approved):

PR #6875, opened 2024-07-04
Be able to get PFN to scope-name translation function from the policy package

haozturk:


haozturk:
Meant to create it against my fork's master for testing, but by mistake created it here. This isn't ready yet, closing it.

Issue #6881, opened 2024-07-08
Stop tracking .idea directory


Geogouz:
Stop tracking .idea (JetBrains IDE project directory) in .gitignore and remove an existing duplicate entry.

PR #6882, opened 2024-07-08
enhancement: Update .gitignore #6881

Geogouz:
Stop tracking .idea (JetBrains IDE project directory) in .gitignore and remove an existing duplicate entry.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,7 +50,6 @@ docs/
 *.log
 *.pyc
 *.swp
-*.swp
 *~
 .venv
 lib/rucio/vcsversion.py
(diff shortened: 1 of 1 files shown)


rdimaio (approved):

rdimaio:
Sidenote: you could also just ignore it locally (https://stackoverflow.com/questions/1753070/how-do-i-configure-git-to-ignore-some-files-locally), but in this case I think it's worth adding this to the `.gitignore` as it's a fairly common IDE

Geogouz:
Yes, plus it covers both PyCharm and IntelliJ.

PR #6903, opened 2024-07-11
Transfers: add overwrite_when_only_on_disk FTS logic

### LINKED ISSUE #6581: Updates to overwrite configuration on tape
### Description

This is the current `overwrite` scenario as I understand it.

1.  Currently, rucio does not allow overwriting on `Tape` (By default)
https://github.com/rucio/rucio/blob/1117e15fdd373ba6a0c24b22b7bd564c13f77241/lib/rucio/transfertool/fts3.py#L333-L334

2. Overwrite  on DISKs  (By Default) https://github.com/rucio/rucio/blob/1117e15fdd373ba6a0c24b22b7bd564c13f77241/lib/rucio/transfertool/fts3.py#L328
 
 (Both of those can be overridden by setting it explicitly on the RSE, @dchristidis @bari12 , kindly correct me on this).
 https://github.com/rucio/rucio/blob/1117e15fdd373ba6a0c24b22b7bd564c13f77241/lib/rucio/transfertool/fts3.py#L354
 
3.  If the file is found corrupted on the first try and it can be affirmed from the returned contents that it is corrupted it can be overwritten (if this is configured in the transfers section)
https://github.com/rucio/rucio/blob/1117e15fdd373ba6a0c24b22b7bd564c13f77241/lib/rucio/transfertool/fts3.py#L705

We would like to have two things changed in this if it makes sense for the community too. 
1. Make the `overwrite_corrupted_files` a per RSE attribute as opposed to a global transfer configuration. 
     - We have had opinions and arguments from sites that strongly want us to enable it and sites that want otherwise
     
2. Add an `overwrite-when-only-on-disk` configuration to be sent with the job params, also configurable at an RSE level for tape RSEs. 
Here is the final proposal document agreed on by the FTS team for future developments. https://cmsdm.web.cern.ch/documents/FTS3_CMS_DM_Overwrite-when-only-on-disk_proposal.pdf

### Motivation

Have the ability to configure sites independently.

### Change

_No response_

labkode:
The following PR allows Rucio to tag transfers to overwrite the disk buffer of tape endpoints and not the tape file when the following conditions are met:
- the RSE has the following attributes/settings configured:
  - RSE needs to be of type TAPE
  - overwrite_when_only_on_disk attribute is set to True
 - the FTS instance configures the tape endpoint with the the field `overwrite_disk_enabled` set to `1`

FTS transfer jobs will be created with this new job parameter and will be classified with the **Overwrite flag**  equal to `D` or `Q`, example below:

Screenshot 2024-07-11 at 16 06 58

The valid values as of 11/07/2024 are:
```
        Y -- overwrite enabled
        R -- overwrite enabled only on FTS retries
        M -- overwrite enabled only for intermediary hops
        D -- overwrite-when-only-on-disk feature enabled (overwrite if destination file has only disk locality)
        Q -- overwrite-hop + overwrite-when-only-on-disk behavior
```
 
In the FST logs the transfer will contain the following values:

```
...
    INFO    Thu, 11 Jul 2024 14:18:44 +0200; Transfer accepted
    INFO    Thu, 11 Jul 2024 14:18:44 +0200; Proxy: /tmp/x509up_hxxx
    INFO    Thu, 11 Jul 2024 14:18:44 +0200; VO: cms
    INFO    Thu, 11 Jul 2024 14:18:44 +0200; Job id: b655d08c-3f7f-11ef-a237-fa163ef11e82
    INFO    Thu, 11 Jul 2024 14:18:44 +0200; File id: 1933293193
    INFO    Thu, 11 Jul 2024 14:18:44 +0200; Source url: davs://eoscms.cern.ch:443/eos/cms/store/test/rucio/int/store/mc/HC/GenericTTbar/AODSIM/CMSSW_9_2_6_91X_mcRun1_realistic_v2-v2/00000/F04D2A27-1B76-E711-8DEF-02163E01501A.root
    INFO    Thu, 11 Jul 2024 14:18:44 +0200; Dest url: davs://eosctacms.cern.ch:8444//eos/ctacms/archivetest/cms/store/test/rucio/int/store/mc/HC/GenericTTbar/AODSIM/CMSSW_9_2_6_91X_mcRun1_realistic_v2-v2/00000/F04D2A27-1B76-E711-8DEF-02163E01501A.root
    INFO    Thu, 11 Jul 2024 14:18:44 +0200; Overwrite enabled: 0
--> INFO    Thu, 11 Jul 2024 14:18:44 +0200; Overwrite when only on disk: 1
    INFO    Thu, 11 Jul 2024 14:18:44 +0200; Disable delegation: 0
    INFO    Thu, 11 Jul 2024 14:18:44 +0200; Disable local streaming: 0
    INFO    Thu, 11 Jul 2024 14:18:44 +0200; Skip eviction of source file: 0
....
```

The PR includes some DEBUG calls to make it easier to debug in runtime.

Overwrite configuration options for transfers are depicted below:
Screenshot 2024-07-12 at 13 52 14

This PR also fixes a wrong hardcoded logic that set overwrite_hop when the transfer was not multi_hop: https://github.com/rucio/rucio/pull/6903/files#diff-10fdb713ecac24728050341f83e17dbe7cc3f1f5a317823860677917579caea2R449

### DIFF
--- a/lib/rucio/transfertool/fts3.py
+++ b/lib/rucio/transfertool/fts3.py
@@ -332,6 +332,8 @@ def build_job_params(
 
     # Overwriting by default is set to True for non TAPE RSEs.
     # Tape RSEs can force overwrite by setting the "overwrite" attribute to True.
+    # There is yet another configuration option: transfers->overwrite_corrupted_files that when is set to True
+    # it will retry failed requests with overwrite flag set to True
     overwrite, overwrite_when_only_on_disk, bring_online_local = True, False, None
     if first_hop.src.rse.is_tape_or_staging_required():
         # Activate bring_online if it was requested by first hop
@@ -343,14 +345,30 @@ def build_job_params(
         # This functionality should reduce the number of stuck files in the disk buffer that are not migrated to tape media (for whatever reason).
         # Please be aware that FTS does not guarantee an atomic operation from the time it checks for existence of the file on disk and tape and
         # the moment the file is overwritten, so there is a race condition that could overwrite the file on the tape media
-        overwrite = last_hop.dst.rse.attributes.get('overwrite', False)  # honour RSE configuration to force overwrite
+        overwrite = last_hop.dst.rse.attributes.get('overwrite', False)  # honour RSE configuration to force overwrite for all requests
         overwrite_when_only_on_disk = last_hop.dst.rse.attributes.get('overwrite_when_only_on_disk', False)
+
         # setting both flags is incompatible, so we opt in for the safest approach: "overwrite_when_only_on_disk"
         # this is aligned with FTS implementation: see
         if overwrite and overwrite_when_only_on_disk:
             overwrite = False
 
-    logger(logging.DEBUG, 'Is it tape?: %s %s overwrite_when_only_on_disk:%s' % (last_hop.dst.rse.name, last_hop.dst.rse.is_tape(), overwrite_when_only_on_disk))
+    if not (overwrite and overwrite_when_only_on_disk):
+        # if there are no overwrite flags set, we still need to check for the
+        # "transfers -> overwrite_corrupted_files setting. The logic behind this flag is that
+        # it will update the rws (RequestWithSources) with the "overwrite" attribute set to True
+        # after finding an 'Destination file exists and overwrite is not enabled' error message
+        overwrite_corrupted_files = last_hop.dst.rws.attributes.ge('overwrite', False)
+        if overwrite_corrupted_files:
+            overwrite = True  # both for DISK and TAPE
+
+    logger(logging.DEBUG, 'RSE:%s Is it Tape? %s overwrite_when_only_on_disk:%s overwrite:%s overwrite_corrupted_files=%s' % (
+        last_hop.dst.rse.name,
+        last_hop.dst.rse.is_tape(),
+        overwrite_when_only_on_disk,
+        overwrite,
+        overwrite_corrupted_files))
+
     logger(logging.DEBUG, 'RSE attributes are: %s' % (last_hop.dst.rse.attributes))
 
     # Get dest space token
--- a/lib/rucio/transfertool/fts3.py
+++ b/lib/rucio/transfertool/fts3.py
@@ -335,32 +335,32 @@ def build_job_params(
     # There is yet another configuration option: transfers->overwrite_corrupted_files that when is set to True
     # it will retry failed requests with overwrite flag set to True
     overwrite, overwrite_when_only_on_disk, bring_online_local = True, False, None
+
     if first_hop.src.rse.is_tape_or_staging_required():
         # Activate bring_online if it was requested by first hop
         # We don't allow multihop via a tape, so bring_online should not be set on any other hop
         bring_online_local = bring_online
+
     if last_hop.dst.rse.is_tape():
         # FTS v3.12.12 introduced a new boolean parameter "overwrite_when_only_on_disk" that controls if the file can be overwritten
-        # in TAPE enabled RSEs ONLY IF the file is on the disk buffer and not yet commited to tape media.
+        # in TAPE enabled RSEs ONLY IF the file is on the disk buffer and not yet committed to tape media.
         # This functionality should reduce the number of stuck files in the disk buffer that are not migrated to tape media (for whatever reason).
         # Please be aware that FTS does not guarantee an atomic operation from the time it checks for existence of the file on disk and tape and
         # the moment the file is overwritten, so there is a race condition that could overwrite the file on the tape media
-        overwrite = last_hop.dst.rse.attributes.get('overwrite', False)  # honour RSE configuration to force overwrite for all requests
+
+        # Setting both flags is incompatible, so we opt in for the safest approach: "overwrite_when_only_on_disk"
+        # this is aligned with FTS implementation: see (internal access only): https://its.cern.ch/jira/browse/FTS-2007
+
         overwrite_when_only_on_disk = last_hop.dst.rse.attributes.get('overwrite_when_only_on_disk', False)
+        overwrite = False if overwrite_when_only_on_disk else last_hop.dst.rse.attributes.get('overwrite', False)
 
-        # setting both flags is incompatible, so we opt in for the safest approach: "overwrite_when_only_on_disk"
-        # this is aligned with FTS implementation: see
-        if overwrite and overwrite_when_only_on_disk:
-            overwrite = False
-
-    if not (overwrite and overwrite_when_only_on_disk):
-        # if there are no overwrite flags set, we still need to check for the
-        # "transfers -> overwrite_corrupted_files setting. The logic behind this flag is that
-        # it will update the rws (RequestWithSources) with the "overwrite" attribute set to True
-        # after finding an 'Destination file exists and overwrite is not enabled' error message
-        overwrite_corrupted_files = last_hop.rws.attributes.get('overwrite', False)
-        if overwrite_corrupted_files:
-            overwrite = True  # both for DISK and TAPE
+    # We still need to check for the
+    # "transfers -> overwrite_corrupted_files setting. The logic behind this flag is that
+    # it will update the rws (RequestWithSources) with the "overwrite" attribute set to True
+    # after finding an 'Destination file exists and overwrite is not enabled' error message
+    overwrite_corrupted_files = last_hop.rws.attributes.get('overwrite', False)
+    if overwrite_corrupted_files:
+        overwrite = True  # both for DISK and TAPE
 
     logger(logging.DEBUG, 'RSE:%s Is it Tape? %s overwrite_when_only_on_disk:%s overwrite:%s overwrite_corrupted_files=%s' % (
         last_hop.dst.rse.name,
@@ -438,7 +438,7 @@ def build_job_params(
 
     # Refer to https://its.cern.ch/jira/browse/FTS-1749 for full details (login needed), extract below:
     # Why the overwrite_hop parameter is needed?
-    # Rucio decides that a multihop transfer is needed DISK1 --> DISK2 --> TAPE1 in order to put the file on tape. For some reason, the file already exist at DISK2.
+    # Rucio decides that a multihop transfer is needed DISK1 --> DISK2 --> TAPE1 in order to put the file on tape. For some reason, the file already exists on DISK2.
     # Rucio doesn't know about this file on DISK2. It could be either a correct or corrupted file. This can be due to a previous issue on Rucio side, FTS side, network side, etc (many possible reasons).
     # Normally, Rucio allows overwrite towards any disk destination, but denies overwrite towards a tape destination. However, in this case, because the destination of the multihop is a tape, DISK2 cannot be overwritten.
     # Proposed solution
@@ -453,6 +453,11 @@ def build_job_params(
         for transfer_hop in transfer_path[:-1]:
             # Only allow overwrite if all hops in multihop allow it
             h_overwrite = transfer_hop.rws.attributes.get('overwrite', True)
+
+            # if one hop cannot be overwritten, quit early
+            if h_overwrite is False:
+                break
+
             job_params['overwrite'] = h_overwrite and job_params['overwrite']
             # Allow overwrite_hop if all intermediate hops allow it (ignoring the last hop)
             overwrite_hop = h_overwrite and overwrite_hop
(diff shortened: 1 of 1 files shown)


lib/rucio/transfertool/fts3.py
```diff
@@ -354,7 +371,8 @@ def build_job_params(
                       'issuer': 'rucio',
                       'multi_sources': False,
                   },
-                  'overwrite': last_hop.rws.attributes.get('overwrite', overwrite),
+                  'overwrite': overwrite,
```
dynamic-entropy, line 393: Hello Hugo,
This current implementation LGTM.
 Do you think it would be better to unify/clean how necessary and optional attributes are set in the FTS job config? 

For e.g. `overwrite` and `overwrite_when_only_on_disk` are optional (i.e its okay to not set them, which would suggest a false value in the job config).
In which case we set them how `multihop` is set. 

Also, do you think its useful to set `overwrite_when_only_on_disk` in the `job_metadata`? (One reason behind this is that you can easily see if the flag is set in the job UI). 
I am not sure what transfer groupings are based on and if they need to be considered.

dynamic-entropy (commented):
I see a test for `overwrite_corrupted_files` failing.

I think its because now the global overwrite flag is not considered anymore. It needs to be set on the RSE. https://github.com/rucio/rucio/blob/40a5ae14680bc14f6be4738d3e75f351e5143ea1/tests/test_conveyor.py#L1351

lib/rucio/transfertool/fts3.py
labkode: Hi Rahul @dynamic-entropy 

I've added the flag to job metadata so is visible as you suggested, I think this is very useful.
The CI is passing now (it was related to the overwrite_corrupted_files setting that I was not aware of).
I've left an image in this PR description explaining which are the the overwrite flags and how they can be tuned.
Regarding group transfers, I haven't seen any showstopper, I think we can go ahead and merge this one. If there are improvements to be made beyond this PR, we can tackle them later.

dynamic-entropy: Yes, we need both the config to be set and the `rws` dictionary to be updated with `overwrite=True`  for a resubmit.

Thanks a lot.

lib/rucio/transfertool/fts3.py
```diff
@@ -321,20 +321,55 @@ def build_job_params(
...
+            overwrite = False
+
+    if not (overwrite and overwrite_when_only_on_disk):
+        # if there are no overwrite flags set, we still need to check for the
+        # "transfers -> overwrite_corrupted_files setting. The logic behind this flag is that
+        # it will update the rws (RequestWithSources) with the "overwrite" attribute set to True
+        # after finding an 'Destination file exists and overwrite is not enabled' error message
+        overwrite_corrupted_files = last_hop.rws.attributes.get('overwrite', False)
+        if overwrite_corrupted_files:
+            overwrite = True  # both for DISK and TAPE
```
dynamic-entropy: It's only considered for tape.
We always overwrite for Disks.

dynamic-entropy (approved):
LGTM

lib/rucio/transfertool/fts3.py
labkode: Thanks!

lib/rucio/transfertool/fts3.py
```diff
@@ -321,20 +321,55 @@ def build_job_params(
...
+    # it will retry failed requests with overwrite flag set to True
+    overwrite, overwrite_when_only_on_disk, bring_online_local = True, False, None
     if first_hop.src.rse.is_tape_or_staging_required():
         # Activate bring_online if it was requested by first hop
         # We don't allow multihop via a tape, so bring_online should not be set on any other hop
         bring_online_local = bring_online
     if last_hop.dst.rse.is_tape():
-        overwrite = False
+        # FTS v3.12.12 introduced a new boolean parameter "overwrite_when_only_on_disk" that controls if the file can be overwritten
+        # in TAPE enabled RSEs ONLY IF the file is on the disk buffer and not yet commited to tape media.
```
rdimaio: nit: 
```suggestion
        # in TAPE enabled RSEs ONLY IF the file is on the disk buffer and not yet committed to tape media.
```

lib/rucio/transfertool/fts3.py
```diff
@@ -321,20 +321,55 @@ def build_job_params(
...
+        # This functionality should reduce the number of stuck files in the disk buffer that are not migrated to tape media (for whatever reason).
+        # Please be aware that FTS does not guarantee an atomic operation from the time it checks for existence of the file on disk and tape and
+        # the moment the file is overwritten, so there is a race condition that could overwrite the file on the tape media
+        overwrite = last_hop.dst.rse.attributes.get('overwrite', False)  # honour RSE configuration to force overwrite for all requests
+        overwrite_when_only_on_disk = last_hop.dst.rse.attributes.get('overwrite_when_only_on_disk', False)
+
+        # setting both flags is incompatible, so we opt in for the safest approach: "overwrite_when_only_on_disk"
+        # this is aligned with FTS implementation: see
```
rdimaio: Regarding the "see" at the end - was there meant to be a link to the FTS implementation?

lib/rucio/transfertool/fts3.py
```diff
@@ -321,20 +321,55 @@ def build_job_params(
...
+        if overwrite and overwrite_when_only_on_disk:
+            overwrite = False
```
rdimaio: Could be simplified like this:

```
overwrite_when_only_on_disk = last_hop.dst.rse.attributes.get('overwrite_when_only_on_disk', False)

if overwrite_when_only_on_disk:
    overwrite = False
else:
    overwrite = last_hop.dst.rse.attributes.get('overwrite', False)
```

or as a one-liner:
```
overwrite = False if overwrite_when_only_on_disk else last_hop.dst.rse.attributes.get('overwrite', False)
```

This is because `overwrite` is always set to `False` if `overwrite_when_only_on_disk` is `True`, so we don't really even need to get it from the RSE attributes in that case.

lib/rucio/transfertool/fts3.py
```diff
@@ -321,20 +321,55 @@ def build_job_params(
...
+
+    if not (overwrite and overwrite_when_only_on_disk):
```
rdimaio: This `if` condition is not needed, because it's `True` under all circumstances:

## 1: if `last_hop.dst.rse.is_tape() == False`
`overwrite` and `overwrite_when_only_on_disk` are initialized like this:
```
overwrite, overwrite_when_only_on_disk, bring_online_local = True, False, None`
```
So if `last_hop.dst.rse.is_tape()` is `False`, `if not (overwrite and overwrite_when_only_on_disk):` will be `True`.

## 2: if `last_hop.dst.rse.is_tape() == True`
 `overwrite` is always set to `False` if `overwrite_when_only_on_disk` is `True` (as seen in the other comment). 

### 2.a: `overwrite_when_only_on_disk == True`
`overwrite` is `False`, so `if not (overwrite and overwrite_when_only_on_disk):` will be `True`.

### 2.b: `overwrite_when_only_on_disk == False`
`overwrite` can either be `True` or `False`, and `if not (overwrite and overwrite_when_only_on_disk):` will be `True`.

lib/rucio/transfertool/fts3.py
```diff
@@ -399,16 +436,31 @@ def build_job_params(
...
-    overwrite_hop = True
-    for transfer_hop in transfer_path[:-1]:
-        # Only allow overwrite if all hops in multihop allow it
-        h_overwrite = transfer_hop.rws.attributes.get('overwrite', True)
-        job_params['overwrite'] = h_overwrite and job_params['overwrite']
-        # Allow overwrite_hop if all intermediate hops allow it (ignoring the last hop)
-        overwrite_hop = h_overwrite and overwrite_hop
+    # Refer to https://its.cern.ch/jira/browse/FTS-1749 for full details (login needed), extract below:
+    # Why the overwrite_hop parameter is needed?
+    # Rucio decides that a multihop transfer is needed DISK1 --> DISK2 --> TAPE1 in order to put the file on tape. For some reason, the file already exist at DISK2.
```
rdimaio:
```suggestion
    # Rucio decides that a multihop transfer is needed DISK1 --> DISK2 --> TAPE1 in order to put the file on tape. For some reason, the file already exists on DISK2.
```

lib/rucio/transfertool/fts3.py
```diff
@@ -399,16 +436,31 @@ def build_job_params(
...
+        overwrite_hop = False
+
+    else:
+        overwrite_hop = True
+        for transfer_hop in transfer_path[:-1]:
+            # Only allow overwrite if all hops in multihop allow it
+            h_overwrite = transfer_hop.rws.attributes.get('overwrite', True)
+            job_params['overwrite'] = h_overwrite and job_params['overwrite']
+            # Allow overwrite_hop if all intermediate hops allow it (ignoring the last hop)
+            overwrite_hop = h_overwrite and overwrite_hop
```
rdimaio: This can be simplified to stop at the first iteration where `h_overwrite` is `False` to avoid having to iterate over all hops (because once an iteration where `h_overwrite` is `False` is reached, `job_params['overwrite']` and `overwrite_hop` will always be `False` afterwards)

rdimaio (changes requested):

lib/rucio/transfertool/fts3.py
rdimaio: A way it could be simplified:
```suggestion
         job_params['overwrite'] = overwrite_hop = all(transfer_hop.rws.attributes.get('overwrite', True) for transfer_hop in transfer_path[:-1])
```

labkode:
@rdimaio applied all your recommendations, and CI is green

lib/rucio/transfertool/fts3.py
```diff
@@ -399,16 +436,36 @@ def build_job_params(
...
+            h_overwrite = transfer_hop.rws.attributes.get('overwrite', True)
+
+            # if one hop cannot be overwritten, quit early
+            if h_overwrite is False:
+                break
+
+            job_params['overwrite'] = h_overwrite and job_params['overwrite']
+            # Allow overwrite_hop if all intermediate hops allow it (ignoring the last hop)
+            overwrite_hop = h_overwrite and overwrite_hop
+
```
rdimaio:
```suggestion
            job_params['overwrite'] = h_overwrite and job_params['overwrite']
            # Allow overwrite_hop if all intermediate hops allow it (ignoring the last hop)
            overwrite_hop = h_overwrite and overwrite_hop
            if h_overwrite is False:
                break
```

The `break` needs to be after the code block at 461-463, otherwise `job_params['overwrite']` and ` overwrite_hop` might remain `True` even though `h_overwrite` is `False`.

I'm more in favour of the `all()` approach:

```
# Set `overwrite_hop` to `True` only if all hops allow it
overwrite_hop = all(transfer_hop.rws.attributes.get('overwrite', True) for transfer_hop in transfer_path[:-1])

job_params['overwrite'] = overwrite_hop and job_params['overwrite']
```

Since it clarifies that all hops must allow overwriting, and it breaks early if one of them returns `False`

rdimaio (changes requested):

lib/rucio/transfertool/fts3.py
```diff
@@ -399,16 +436,36 @@ def build_job_params(
...
     if not job_params['overwrite'] and overwrite_hop:
         job_params['overwrite_hop'] = overwrite_hop
```
rdimaio, line 457: Out of scope of this PR, but I'm a bit confused by this condition. It only evaluates to `True` when:
- `job_params['overwrite']` is `False`
- `overwrite_hop` is `True`

I don't understand what this means in practice - do you know more about this?

lib/rucio/transfertool/fts3.py
labkode: Makes sense and is cleaner code

lib/rucio/transfertool/fts3.py
labkode: I think it means that if there is a need in a multi-hop transfer to overwrite the intermediate location and overwrite is to false (default for tape), the overwrite_hop is set to true so FTS can overwrite only the intermediate data but not the final one. I think this use case is there to allow transfers to tape that are multi-hop, and when one of the hops in the middle gets stuck, to create another request to overwrite these bad data and make it to the end hop without risking to overwrite data in tape media if in the mean time the intermediate data got unstuck.

rdimaio: Makes sense, thanks!

rdimaio (approved):

bari12:
Looks good, please squash the commits next time, I will manually squash them now :-)

bari12:
Oh and also mention #6581 in the commit message 😄

PR #6909, opened 2024-07-12
Core & Internals: Migrate core/config.py to SQLA20 #6640

erlingstaff:

Migrating core/config.py to SQLAlchemy 2.0

### DIFF
--- a/lib/rucio/core/config.py
+++ b/lib/rucio/core/config.py
@@ -16,7 +16,7 @@ from collections.abc import Callable
 from typing import TYPE_CHECKING, Any, Optional, TypeVar
 
 from dogpile.cache.api import NoValue
-from sqlalchemy import func
+from sqlalchemy import and_, delete, func, select, update
 
 from rucio.common.cache import make_region_memcached
 from rucio.common.exception import ConfigNotFound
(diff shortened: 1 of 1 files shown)


lib/rucio/core/config.py
```diff
@@ -307,15 +348,25 @@ def remove_section(section: str, *, session: "Session") -> bool:
     if not has_section(section=section, session=session):
         return False
     else:
-        for old in session.query(models.Config.value).filter_by(section=section).all():
+        stmt = select(
+            models.Config.value
+        ).where(
+            models.Config.section == section
+        )
+        for old in session.execute(stmt).all():
```
erlingstaff, line 356: This is translated "verbatim" to Sqlalchemy 2 syntax, but this query does not make much sense to me (both as it used to be and as it is now) - Selecting only `models.Config.value` and later indexing this 0 through 2 (`old[0]`, `old[1]`, `old[2]`) and using them as parameters for `section`, `opt`, and `value`. Perhaps I am misunderstanding, or perhaps this is just not used, in which case it should be removed.

rdimaio (approved):

lib/rucio/core/config.py
bari12: I think this makes no sense and does not, and never did, work. Probably this method is just not really used. But we should address it in a follow up PR. 👍

PR #6913, opened 2024-07-15
Core & Internals: Migrate core/did.py to SQLA2 #6640

erlingstaff:

Migrating core/did.py to SQLAlchemy 2.0

### DIFF
--- a/lib/rucio/core/did.py
+++ b/lib/rucio/core/did.py
@@ -79,13 +79,15 @@ def list_expired_dids(
         models.DataIdentifier.did_type,
         models.DataIdentifier.created_at,
         models.DataIdentifier.purge_replicas
+    ).with_hint(
+        models.DataIdentifier,
+        'INDEX(DIDS DIDS_EXPIRED_AT_IDX)',
+        'oracle'
     ).where(
         models.DataIdentifier.expired_at < datetime.utcnow(),
         not_(sub_query),
     ).order_by(
         models.DataIdentifier.expired_at
-    ).with_hint(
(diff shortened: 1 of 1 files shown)


lib/rucio/core/did.py
```diff
@@ -459,25 +469,33 @@ def __add_files_to_archive(
 
     # insert into archive_contents
     try:
-        dids_to_add and session.execute(insert(models.DataIdentifier), list(dids_to_add.values()))
-        archive_contents_to_add and session.execute(insert(models.ConstituentAssociation), list(archive_contents_to_add.values()))
+        values = list(dids_to_add.values())
+        stmt = insert(
+            models.DataIdentifier
+        )
+        dids_to_add and session.execute(stmt, values)
```
rdimaio, line 476: out of scope of this PR, but I've been wondering what the reason is for these statements - e.g. what is `dids_to_add and` doing here? I've seen them quite a few times around this codebase

rdimaio (approved):

lib/rucio/core/did.py
erlingstaff: I think its a short-circut evaluation to conditionally execute the query - if dids_to_add evaluates to `False` (i.e. empty/None) then the part after `and` is not evaluated. Same as writing
```
if dids_to_add:
    session.execute(stmt, values)
```

rdimaio: They seem to cause pyright `reportUnusedExpression` warnings though - maybe it's best if we expand them out at a later point to also make the code a bit clearer

Issue #6639, opened 2024-04-05
Re-write and standardization of client command structure


bari12:
### Description

Standardize the cli client commands 
- Create a Client 2.0 
- Possible commands will follow the database structure more closely
- Each command will have a common set of "verbs" to use (set/unset (update something existing), add/remove (modify the existence of something), list (read contents)
- Bring rucio-admin and rucio in under the same command (rucio)

### Motivation

 It is generally difficult to know commands without looking at the documentation, and the structure is very hap-hazard. No time like the present to bring it under a uniform structure. 

### Change

- Add depreciation warnings to the original client's output 
- Write a new client as diagrammed below 
- Add a table to documentation to translate the original client commands into the client 2.0 commands 
- Introduce the below verbs: 
--add/remove: Create new instance of object, update db with new row, delete object
--set/unset: Update something existing/Replace existing setting/Return setting to default or Null
--list: Read the state of the object (according to command args/requested verbosity) 

- Introduce the below commands: 
-- ping/whoami [Outliers, no verb. They are the verb.]
-- replica 
--- replica state 
--- replica pfn
--- replica tombstone 
-- did 
-- -did parent 
--- did history 
--- did metadata 
--- did lifetime-execption
-- rse 
--- rse usage 
--- rse distance 
--- rse attribute 
--- rse protocol 
--- rse limit 
--- rse qos-policy 
-- account 
    --- account attributes 
    --- account limits 
    --- account ban 
    --- account identities 
-- config 
-- subscription 
-- scope 

![Client Overhaul design diagram](https://github.com/rucio/rucio/assets/27649232/19d10309-9f58-426f-ace7-05a13730cb7d)

panta-123:
Let me know if you need extra manpower for this. I am willing to collaborate.

voetberg:
> Let me know if you need extra manpower for this. I am willing to collaborate.

Thank you @panta-123 ! If it's all the same to you, once I get the base structure and tests done, could I ask for your help with re-implementing some of the existing client commands in the new format?

panta-123:
@voetberg . Sure let me know once the base structure is ready.

PR #6923, opened 2024-07-16
Clients: Remap all cli commands to a uniform structure #6639

voetberg:
Remaps all the commands in bin/rucio and bin/rucio-admin to the same structure under the `rucio {verb} command subcommand`

Moves bin/rucio and bin/rucio-admin to lib/rucio/client/bin so the existing functions can be used by the new commands (under lib/rucio/client/commands) 

Adds a printed warning to the output of bin/rucio and bin/rucio-admin telling the user which new commands to use. 

Pyright failure are due to pyright picking up bin/rucio(-admin) again.


voetberg:
Closing in favor of revisiting this issue after a community vote has been placed

PR #6924, opened 2024-07-16
Be able to get PFN to scope-name translation function from the policy package

### LINKED ISSUE #6817: Get the PFN to scope:name translation function from policy package
### Description

This translation is needed to declare replicas suspicious [1,2,3]. Currently, the function which performs this translation is decided based on the VO [4]. Since CMS isn't multi-VO, we're using the default VO (`def`) which corresponds to the default function [5] which simply doesn't work for CMS [6]. 

[1] https://github.com/rucio/rucio/blob/bc29b8f9683f399ec74ed241819c2a6bf8dad2ed/lib/rucio/core/replica.py#L271
[2] https://github.com/rucio/rucio/blob/bc29b8f9683f399ec74ed241819c2a6bf8dad2ed/lib/rucio/rse/rsemanager.py#L29
[3] https://github.com/rucio/rucio/blob/bc29b8f9683f399ec74ed241819c2a6bf8dad2ed/lib/rucio/rse/protocols/protocol.py#L40
[4] https://github.com/rucio/rucio/blob/bc29b8f9683f399ec74ed241819c2a6bf8dad2ed/lib/rucio/rse/protocols/protocol.py#L46-L47
[5] https://github.com/rucio/rucio/blob/bc29b8f9683f399ec74ed241819c2a6bf8dad2ed/lib/rucio/rse/protocols/protocol.py#L60
[6] https://github.com/dmwm/CMSRucio/issues/692#issuecomment-2133123551 

### Motivation

To be able to mark replicas suspicious for CMS

### Change

After some discussions with @voetberg and @dynamic-entropy, we decided that the best way to handle this is to implement the relevant function in the policy package. The `RSEDeterministicScopeTranslation` class should check whether there's any function implemented in the policy package. If so, it should use it, o/w use the static functions implemented in the class (`_default` and `_atlas`). 

I'm planning to work on this change, so you can assign it to me, if that's okay. Also if somebody has a better idea, please let me know.

haozturk:
Fixes #6817 

This PR makes `RSEDeterministicScopeTranslation` get the parser function from the policy package if it's configured so. O/w it works as before, i.e. use the `atlas` or `default` function from the same class according to the VO.

### DIFF
--- a/lib/rucio/rse/protocols/protocol.py
+++ b/lib/rucio/rse/protocols/protocol.py
@@ -42,19 +42,21 @@ class RSEDeterministicScopeTranslation(PolicyPackageAlgorithms):
         Translates a pfn dictionary into a scope and name
     """
 
-    _algorithm_type = "pfn2scope_name"
+    _algorithm_type = "pfn2lfn"
 
     def __init__(self, vo: str = 'def'):
         super().__init__()
         self.register("def", RSEDeterministicScopeTranslation._default)
         self.register("atlas", RSEDeterministicScopeTranslation._atlas)
         policy_module = vo
+        logger = logging.getLogger(__name__)
         try:
             # Use the function defined in the policy package if it's configured so
             algo_type = self._algorithm_type
             algorithm_name = config.config_get('policy', self._algorithm_type)
         except (NoOptionError, NoSectionError, RuntimeError):
             # Don't use a function from the policy package. Use one defined in this class according to vo
+            logger.debug("PFN2LFN function will not be fetched from the policy package")
             algo_type = self.__class__.__name__
             if super()._supports(self.__class__.__name__, policy_module):
                 algorithm_name = policy_module
--- a/lib/rucio/rse/protocols/protocol.py
+++ b/lib/rucio/rse/protocols/protocol.py
@@ -20,7 +20,7 @@ import hashlib
 import logging
 from collections.abc import Callable, Mapping
 from configparser import NoOptionError, NoSectionError
-from typing import TypeVar, Any
+from typing import Any, TypeVar
 from urllib.parse import urlparse
 
 from rucio.common import config, exception
--- a/lib/rucio/rse/protocols/protocol.py
+++ b/lib/rucio/rse/protocols/protocol.py
@@ -41,16 +41,31 @@ class RSEDeterministicScopeTranslation(PolicyPackageAlgorithms):
     """
         Translates a pfn dictionary into a scope and name
     """
+
+    _algorithm_type = "pfn2scope_name"
+
     def __init__(self, vo: str = 'def'):
         super().__init__()
         self.register("def", RSEDeterministicScopeTranslation._default)
         self.register("atlas", RSEDeterministicScopeTranslation._atlas)
         policy_module = vo
-        # Uses the same policy as the DeterministicTranslation
-        if super()._supports(self.__class__.__name__, policy_module):
-            self.parser = self._get_one_algorithm(self.__class__.__name__, policy_module)
-        else:
-            self.parser = self._get_one_algorithm(self.__class__.__name__, "def")
+        try:
+            # Use the function defined in the policy package if it's configured so
+            algo_type = self._algorithm_type
+            algorithm_name = config.config_get('policy', self._algorithm_type)
+        except (NoOptionError, NoSectionError, RuntimeError):
+            # Don't use a function from the policy package. Use one defined in this class according to vo
+            algo_type = self.__class__.__name__
+            if super()._supports(self.__class__.__name__, policy_module):
+                algorithm_name = policy_module
+            else:
+                algorithm_name = "def"
+
+        self.parser = self.get_parser(algo_type, algorithm_name)
+
+    @classmethod
+    def get_parser(cls, algorithm_type, algorithm_name):
+        return super()._get_one_algorithm(algorithm_type, algorithm_name)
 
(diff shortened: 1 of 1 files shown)


lib/rucio/rse/protocols/protocol.py
```diff
@@ -41,16 +41,31 @@ class RSEDeterministicScopeTranslation(PolicyPackageAlgorithms):
...
-        # Uses the same policy as the DeterministicTranslation
-        if super()._supports(self.__class__.__name__, policy_module):
-            self.parser = self._get_one_algorithm(self.__class__.__name__, policy_module)
-        else:
-            self.parser = self._get_one_algorithm(self.__class__.__name__, "def")
+        try:
+            # Use the function defined in the policy package if it's configured so
+            algo_type = self._algorithm_type
+            algorithm_name = config.config_get('policy', self._algorithm_type)
+        except (NoOptionError, NoSectionError, RuntimeError):
```
voetberg, line 57: What causes this runtime error, is it just a missing config? It's a little generic, it could be useful to put a logging statement here just in case.

lib/rucio/rse/protocols/protocol.py
```diff
@@ -41,16 +41,31 @@ class RSEDeterministicScopeTranslation(PolicyPackageAlgorithms):
     """
         Translates a pfn dictionary into a scope and name
     """
+
+    _algorithm_type = "pfn2scope_name"
```
dynamic-entropy: I think, `pfn2lfn` would be a more standard name. 
Since this is supposed to be an inverse operation of `lfn2pfn` algorithms.

It is understood that `scope` is implicitly an output and input in the `pfn2lfn` and `lfn2pfn` policy package functions respectively.

lib/rucio/rse/protocols/protocol.py
haozturk: If the configuration is missing,  we should get `NoOptionError` or `NoSectionError` exceptions. I actually don't know what would throw `RuntimeError`. I've included it, since it was also included in many places in rucio code where a config is tried to be read, e.g. [1]. Not sure if we should keep it or not.

I can put a logging line. Should I make it `DEBUG` level? For communities that don't rely on policy package for this functionality, it might clutter their logging o/w

[1] https://github.com/rucio/rucio/blob/a62dfe18216cd0bcea238b5ed7b1197345a7f2fd/lib/rucio/daemons/reaper/reaper.py#L274

lib/rucio/rse/protocols/protocol.py
haozturk: okay, I don't have an objection to this. I'll do all the changes at once once the other thread is resolved.

lib/rucio/rse/protocols/protocol.py
voetberg: I would do it at debug level for sure, if we don't know what the Runtime is catching I would mark that as better safe than sorry. It is up to your digression though.

haozturk: I added the logging and I kept the RuntimeError exception. Does it look good now?

lib/rucio/rse/protocols/protocol.py
haozturk: I renamed it, does it look good now?

voetberg (dismissed):

lib/rucio/rse/protocols/protocol.py
```diff
@@ -41,16 +41,33 @@ class RSEDeterministicScopeTranslation(PolicyPackageAlgorithms):
...
+            algo_type = self.__class__.__name__
+            if super()._supports(self.__class__.__name__, policy_module):
+                algorithm_name = policy_module
+            else:
+                algorithm_name = "def"
+
+        self.parser = self.get_parser(algo_type, algorithm_name)
+
+    @classmethod
+    def get_parser(cls, algorithm_type, algorithm_name):
```
rdimaio:
```suggestion
    def get_parser(cls, algorithm_type: str, algorithm_name: str) -> Callable[..., Any]:
```

I'm not sure why the `pyright` check didn't complain about this - it should usually fail when the number of missing type annotations increases in a PR

lib/rucio/rse/protocols/protocol.py
```diff
@@ -41,16 +41,33 @@ class RSEDeterministicScopeTranslation(PolicyPackageAlgorithms):
...
+        logger = logging.getLogger(__name__)
+        try:
+            # Use the function defined in the policy package if it's configured so
+            algo_type = self._algorithm_type
+            algorithm_name = config.config_get('policy', self._algorithm_type)
+        except (NoOptionError, NoSectionError, RuntimeError):
+            # Don't use a function from the policy package. Use one defined in this class according to vo
+            logger.debug("PFN2LFN function will not be fetched from the policy package")
+            algo_type = self.__class__.__name__
+            if super()._supports(self.__class__.__name__, policy_module):
```
rdimaio: nit, for simplicity:

```suggestion
            if super()._supports(algo_type, policy_module):
```

lib/rucio/rse/protocols/protocol.py
haozturk: added them, thanks

lib/rucio/rse/protocols/protocol.py
haozturk: done

rdimaio (approved):

Issue #6826, opened 2024-06-17
Enable linting to avoid f-string logging


rdimaio:
We have a few instances of using f-strings in logging, e.g.: https://github.com/search?q=repo%3Arucio%2Frucio+%2Flog.*%2C+f%22%2F&type=code

As mentioned here: https://github.com/rucio/rucio/pull/6722#discussion_r1638565469, though there might be negligible changes in performance, it might be a good practice to follow.

Our linter (Ruff) has a code for matching these occurrences: https://docs.astral.sh/ruff/rules/logging-f-string/

dchristidis:
Very nice! We should apply this to all forms of interpolation, not only f-strings:
* [logging-string-format](https://docs.astral.sh/ruff/rules/logging-string-format/): `logging.info('foo {}'.format(bar))`
* [logging-percent-format](https://docs.astral.sh/ruff/rules/logging-percent-format/): `logging.info('foo %s' % bar)`
* [logging-string-concat](https://docs.astral.sh/ruff/rules/logging-string-concat/): `logging.info('foo ' + bar)`

PR #6926, opened 2024-07-16
Functional Tests :  Automatix crash if one destination RSE is not available #6925

### LINKED ISSUE #6925: Automatix crash if one destination RSE is not available
### Description

If `automatix` is run on a list of RSEs and one of these RSEs has issue, an exception can be raised into `upload_client.upload`. This exception is not caught and `automatix` won't be able to process the rest of the RSEs

### Steps to reproduce

-

### Rucio Version

_No response_

### Additional Information

_No response_

cserf:


lib/rucio/daemons/automatix/automatix.py
```diff
@@ -239,23 +241,32 @@ def run_once(heartbeat_handler: HeartbeatHandler, inputfile: str, **_kwargs) ->
...
-        else:
-            logger(logging.INFO, "Error uploading files")
-        for physical_fname in physical_fnames:
-            remove(physical_fname)
-        rmdir(tmpdir)
+        try:
+            upload_client = UploadClient(client)
+            ret = upload_client.upload(files)
+            if ret == 0:
+                logger(logging.INFO, f"{dsn} successfully registered on {rse}")
```
voetberg: Re issue #6826 , these should probably be %s style formating and not f-strings.

voetberg (changes requested):

lib/rucio/daemons/automatix/automatix.py
```diff
@@ -239,23 +241,32 @@ def run_once(heartbeat_handler: HeartbeatHandler, inputfile: str, **_kwargs) ->
...
+        logging.INFO,
+        f"It took {cycle_stopwatch.elapsed} seconds to upload datasets on {len(successes)} RSEs: {str(successes)}",
+    )
     logger(
         logging.INFO,
-        "It took %f seconds to upload one dataset on %s",
-        cycle_stopwatch.elapsed,
-        str(rses),
+        f"Datasets could not be uploaded on {len(failures)} RSEs: {str(failures)}",
     )
```
rdimaio: I think this should only be logged conditionally in case there's any failures (otherwise it'd just say "datasets could not be uploaded on 0 RSEs:"

lib/rucio/daemons/automatix/automatix.py
```diff
@@ -239,23 +241,32 @@ def run_once(heartbeat_handler: HeartbeatHandler, inputfile: str, **_kwargs) ->
...
+        finally:
+            for physical_fname in physical_fnames:
+                remove(physical_fname)
+            rmdir(tmpdir)
+    logger(
+        logging.INFO,
+        f"It took {cycle_stopwatch.elapsed} seconds to upload datasets on {len(successes)} RSEs: {str(successes)}",
+    )
     logger(
         logging.INFO,
```
rdimaio: `logging.ERROR` might be better - we expect this log to only appear if an error occurs

lib/rucio/daemons/automatix/automatix.py
```diff
@@ -239,23 +241,32 @@ def run_once(heartbeat_handler: HeartbeatHandler, inputfile: str, **_kwargs) ->
...
+                logger(logging.INFO, f"{dsn} successfully registered on {rse}")
+                METRICS.counter(name="addnewdataset.done").inc()
+                METRICS.counter(name="addnewfile.done").inc(nbfiles)
+                METRICS.timer(name='datasetinjection').observe(stopwatch.elapsed)
+                successes.append(rse)
+            else:
+                logger(logging.INFO, "Error uploading files")
+                failures.append(rse)
+        except Exception as error:
+            logger(logging.ERROR, f"Error uploading files on {rse}: {str(error)}")
```
rdimaio: Same point about f-string logging (https://github.com/rucio/rucio/issues/6826)

lib/rucio/daemons/automatix/automatix.py
```diff
@@ -239,23 +241,32 @@ def run_once(heartbeat_handler: HeartbeatHandler, inputfile: str, **_kwargs) ->
...
+            failures.append(rse)
+        finally:
+            for physical_fname in physical_fnames:
+                remove(physical_fname)
+            rmdir(tmpdir)
+    logger(
+        logging.INFO,
+        f"It took {cycle_stopwatch.elapsed} seconds to upload datasets on {len(successes)} RSEs: {str(successes)}",
```
rdimaio: Same point about f-string logging (https://github.com/rucio/rucio/issues/6826)

lib/rucio/daemons/automatix/automatix.py
```diff
@@ -239,23 +241,32 @@ def run_once(heartbeat_handler: HeartbeatHandler, inputfile: str, **_kwargs) ->
...
+    )
     logger(
         logging.INFO,
-        "It took %f seconds to upload one dataset on %s",
-        cycle_stopwatch.elapsed,
-        str(rses),
+        f"Datasets could not be uploaded on {len(failures)} RSEs: {str(failures)}",
```
rdimaio: Same point about f-string logging (https://github.com/rucio/rucio/issues/6826)

rdimaio (changes requested):

lib/rucio/daemons/automatix/automatix.py
rdimaio: (this comment refers to line 268, not 264 - the way the diff looks on GH is a bit weird)

lib/rucio/daemons/automatix/automatix.py
Ban42: Wouldn't it be better if there is always only one logging message in the format:
`"It took {cycle_stopwatch.elapsed} seconds to successfully upload {len(success)} out of {len(success)+len(failures)}. Successful RSEs : {str(successes)}, failed RSEs : {str(failures)}"` ?

The reason being that in case of partial success (5 successful uploads over 6) it may be weird for the user to read two different logs saying operation was successful and then operation was not successful. Just my 2cents.

rdimaio (approved):
LGTM, we can merge it like this, but we should consider implementing this review comment https://github.com/rucio/rucio/pull/6926#discussion_r1682765749 as it would be helpful to see how many RSEs failed out of all the RSEs attempted (e.g. if only one RSE fails it might be an issue with that RSE, if all RSEs fail there might be something else giong on)

PR #6927, opened 2024-07-16
Core & Internals: Migrate heartbeat to SQLA20 #6640

erlingstaff:

Migrating core/heartbeat.py to SQLAlchemy 2.0

### DIFF
--- a/lib/rucio/core/heartbeat.py
+++ b/lib/rucio/core/heartbeat.py
@@ -16,8 +16,7 @@ import datetime
 import hashlib
 from typing import TYPE_CHECKING, Optional
 
-from sqlalchemy import func, select
-from sqlalchemy.sql import distinct
+from sqlalchemy import and_, delete, func, select, update
 
 from rucio.common.exception import DatabaseException
 from rucio.common.utils import pid_exists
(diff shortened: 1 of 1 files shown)


lib/rucio/core/heartbeat.py
```diff
@@ -166,19 +192,28 @@ def live(
...
-                             Heartbeats.thread_id)\
-                   .order_by(Heartbeats.hostname,
-                             Heartbeats.pid,
-                             Heartbeats.thread_id)
-    result = query.all()
+    stmt = select(
+        Heartbeats.hostname,
+        Heartbeats.pid,
+        Heartbeats.thread_id
+    ).distinct(
```
rdimaio: this `distinct()` was not present in the original query, but this looks correct anyway

rdimaio (dismissed):

lib/rucio/core/heartbeat.py
bari12: I think there is basically no difference due to the group_by, but I would still remove it, it just doesn't add anything.

erlingstaff: Removed

rdimaio (approved):

PR #6929, opened 2024-07-16
Core & Internals: Migrate core/scope.py to SQLA20 #6640

erlingstaff:

1. Migrating core/scope.py to SQLAlchemy 2.0
2. Corrected a faulty return type from `str` to `InternalScope` that raised an error when migrating the queries.

### DIFF
--- a/lib/rucio/core/scope.py
+++ b/lib/rucio/core/scope.py
@@ -16,6 +16,7 @@ from re import match
 from traceback import format_exc
 from typing import TYPE_CHECKING, Any, Optional
 
+from sqlalchemy import and_, select
 from sqlalchemy.exc import IntegrityError
 
 from rucio.common.exception import AccountNotFound, Duplicate, RucioException, VONotFound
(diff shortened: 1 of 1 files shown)


lib/rucio/core/scope.py
```diff
@@ -79,7 +85,7 @@ def bulk_add_scopes(scopes, account, skipExisting=False, *, session: "Session"):
 
 
 @read_session
-def list_scopes(filter_: Optional[dict[str, Any]] = None, *, session: "Session") -> list[str]:
+def list_scopes(filter_: Optional[dict[str, Any]] = None, *, session: "Session") -> list[models.InternalScope]:
```
rdimaio: `InternalScope` is defined in `types`, best to import it from there: https://github.com/rucio/rucio/blob/a1d68d02d7d634827236d1e052a32b9a0c6bf516/lib/rucio/common/types.py#L112

rdimaio (changes requested):

lib/rucio/core/scope.py
erlingstaff: Importing `InternalScope` from `from rucio.common.types` now

rdimaio (approved):

PR #6930, opened 2024-07-16
Refactor SQLA models `UUID` type hints and add type annotations to account_limit

rdimaio:
Part of #6588. The `UUID` type hints are changed to `str` per discussion with Dimitrios

### DIFF
--- a/lib/rucio/gateway/account_limit.py
+++ b/lib/rucio/gateway/account_limit.py
@@ -56,10 +56,10 @@ def get_local_account_limits(account, vo='def', *, session: "Session"):
     :returns: The account limits.
     """
 
-    account = InternalAccount(account, vo=vo)
+    internal_account = InternalAccount(account, vo=vo)
 
     rse_instead_id = {}
-    for elem in account_limit_core.get_local_account_limits(account=account, session=session).items():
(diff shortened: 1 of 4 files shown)


lib/rucio/gateway/account_limit.py
```diff
@@ -56,16 +66,22 @@ def get_local_account_limits(account, vo='def', *, session: "Session"):
     :returns: The account limits.
     """
 
-    account = InternalAccount(account, vo=vo)
+    internal_account = InternalAccount(account, vo=vo)
```
voetberg, line 69: This rename should probably be a different commit (same goes for lines 98, 123, 152 etc)

rdimaio: It's directly related to the type annotation change (because by specifying `account: str`, the linter complains about this line as we're now passing an `InternalAccount` to an object that should be `str`), but I'll break these out in a separate commit for clarity

rdimaio:
Rebased in latest forcepush

lib/rucio/gateway/account_limit.py
rdimaio: Addressed in latest forcepush

bari12 (approved):

PR #6935, opened 2024-07-16
Core & Internals: Migrate core/lock to SQLA20 #6640

erlingstaff:

Migrating core/lock.py to SQLAlchemy 2.0

### DIFF
--- a/lib/rucio/core/lock.py
+++ b/lib/rucio/core/lock.py
@@ -18,7 +18,7 @@ from datetime import datetime
 from typing import TYPE_CHECKING, Any, Optional, Union
 
 from sqlalchemy.exc import DatabaseError
-from sqlalchemy.sql.expression import and_, or_
+from sqlalchemy.sql.expression import and_, or_, select, true, update
 
 import rucio.core.did
 import rucio.core.rule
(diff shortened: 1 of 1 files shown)


rdimaio (dismissed):

rdimaio (approved):

lib/rucio/core/lock.py
```diff
@@ -228,83 +260,60 @@ def get_files_and_replica_locks_of_dataset(scope: InternalScope, name: str, nowa
...
+        stmt = stmt.with_hint(
+            models.DataIdentifierAssociation,
+            'INDEX_RS_ASC(CONTENTS CONTENTS_PK) NO_INDEX_FFS(CONTENTS CONTENTS_PK)',
+            'oracle'
+        ).outerjoin(
+            models.ReplicaLock,
+            and_(models.DataIdentifierAssociation.child_scope == models.ReplicaLock.scope,
+                 models.DataIdentifierAssociation.child_name == models.ReplicaLock.name,
+                 or_(*rse_clause))
+        )
```
rdimaio, line 305: The changes in this section can be a bit hard to understand/review without context, I think it'd be helpful to have a description in the PR (or in the commit message) to explain these changes. 

My understanding is that the changes here reduce the duplicated queries and have a base statement that is then modified depending on the conditions - this makes sense, but it's not properly reflected in the PR title/body and commit message (which just indicate a migration to SQLA2.0). To make it easier to review (and for git history purposes), it would be best to separate these types of changes in separate commits:
1. changes that purely migrate to sqla2.0 syntax
2. changes that are just about formatting queries
3. changes that functionally modify/refactor queries in some way (e.g. the changes in this section)

I think it'd be annoying to rebase this now, but I think it would be good practice to do this for future PRs

erlingstaff: Good point, some context regarding this change: the reason I refactored this query specifically is that it caught my eye as essentially the same query that I changed in the core/replica.py PR (https://github.com/rucio/rucio/pull/6591#discussion_r1549913263) - I suppose both of these refactors should have been separate commits

PR #6939, opened 2024-07-16
Core & Internals: Migrate quarantined_replica to SQLA20 #6640

erlingstaff:

Migrating core/quarantined_replica.py to SQLAlchemy 2.0

### DIFF
--- a/lib/rucio/core/quarantined_replica.py
+++ b/lib/rucio/core/quarantined_replica.py
@@ -16,7 +16,7 @@ import datetime
 from collections.abc import Iterable
 from typing import TYPE_CHECKING, Any, Optional
 
-from sqlalchemy import and_, or_
+from sqlalchemy import and_, delete, or_, select
 from sqlalchemy.sql.expression import false, insert
 
 from rucio.common.utils import chunks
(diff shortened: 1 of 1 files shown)


lib/rucio/core/quarantined_replica.py
```diff
@@ -180,11 +207,17 @@ def list_rses_with_quarantined_replicas(filters: Optional[dict[str, Any]] = None
...
+    )
 
     if filters and filters.get('vo'):
-        query = query.filter(getattr(models.RSE, 'vo') == filters.get('vo'))
+        stmt = stmt.where(
+            models.RSE.vo == filters.get('vo')
+        )
 
-    return [rse for (rse,) in query]
+    return [str(rseid) for rseid in session.execute(stmt).scalars().all()]
```
erlingstaff, line 223: Just saw #6930, when that PR is merged this query should be rewritten to:
`    return list(session.execute(stmt).scalars().all())`

If #6930 is merged before this PR then I'll modify it in this PR, otherwise I will create a new PR addressing this at a later date

rdimaio (approved):

PR #6947, opened 2024-07-17
Refactor trace.validate_schema and add type annotations to trace and nongrid_trace

rdimaio:
Part of #6588

### DIFF
--- a/lib/rucio/core/trace.py
+++ b/lib/rucio/core/trace.py
@@ -349,8 +349,9 @@ def validate_schema(obj):
         if obj and 'eventType' in obj:
             event_type = SCHEMAS.get(obj['eventType'].lower())
             if not event_type:
-                validation_error = ValidationError(message=f"Trace schema for eventType {obj['eventType']} is not currently supported.")
-                validation_error.cause = "SCHEMA_NOT_FOUND"
+                validation_error = ValidationError(
+                    message=f"Trace schema for eventType {obj['eventType']} is not currently supported.",
+                    cause="SCHEMA_NOT_FOUND")
                 raise validation_error
             validate(obj, SCHEMAS.get(obj['eventType'].lower()), format_checker=FORMAT_CHECKER)
     except ValidationError as error:
--- a/lib/rucio/core/trace.py
+++ b/lib/rucio/core/trace.py
@@ -21,6 +21,7 @@ import json
 import logging.handlers
 import random
 import socket
+from typing import TYPE_CHECKING, Any, Union, overload
 
 import stomp
 from jsonschema import Draft7Validator, ValidationError, validate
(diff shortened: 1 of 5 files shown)


tests/test_trace.py
```diff
@@ -86,12 +86,8 @@ def test_trace_ip():
             validate_schema(obj)
 
 
-def test_non_existant_event_type_validation_rejection(caplog):
-    """
-    Test if an incoming trace with a non-supported schema logs a warning
-    instead of printing entire stack-trace
-    """
+def test_non_existant_event_type_validation_rejection():
```
voetberg, line 89: Why was the docstring removed here?

lib/rucio/core/trace.py
```diff
@@ -335,26 +353,23 @@ def trace(payload):
...
+        event_type = loaded_obj['eventType'].lower()
+        schema = SCHEMAS.get(event_type)
+        if schema is not None:
+            try:
+                validate(loaded_obj, schema, format_checker=FORMAT_CHECKER)
+            except ValidationError as error:
+                raise InvalidObject(error)
         else:
-            raise InvalidObject(error)
+            raise TraceValidationSchemaNotFound("Trace schema for eventType %s not found." % event_type)
```
voetberg: The added verbosity of "Trace scheme not supported" is better explanation of what the actual issue is I think. Sure, the schema couldn't be found, but it's because it's not supported, not because a schema file is missing somewhere.

voetberg (changes requested):
Minor nit-picks, dismiss if you don't like them.

tests/test_trace.py
rdimaio: Because it doesn't reflect what the test does anymore, and the new version of the test is clear enough to not need a docstring (in my opinion)

lib/rucio/core/trace.py
rdimaio: I agree, good point - I based the exception on the error cause that was previously present (`SCHEMA_NOT_FOUND`).

I think it depends on whether `eventType` has been validated before getting to `core` here. IMO the best setup would be this:
1. at the API level, `eventType` should be validated to see if it's a supported event type. If not, it raises `EventTypeNotSupported`
2. At the core level (here), assuming (due to point 1) that this can only be called with a valid `eventType`, we raise `TraceValidationSchemaNotFound` if we don't find a schema for this event type.

I'm not sure we're doing API validation at the moment, but I think ideally this might be preferred over raising `EventTypeNotSupported` here (and effectively doing validation in `core`)

voetberg: Yeah ideally this validation should be done in the API [but it doesn't seem to be](https://github.com/rucio/rucio/blob/master/lib/rucio/web/rest/flaskapi/v1/traces.py#L81). Unless you want to do that refactor in this PR it may be be just a nice band-aid to have the raised error more descriptive

rdimaio:
Rebased in latest force-push

lib/rucio/core/trace.py
rdimaio: Expanded the error message in the latest revision so that it mentions that the error could either be that the trace schema was not found or that the event type is unsupported

voetberg (approved):
lgtm

PR #6950, opened 2024-07-17
enhancement: BaseClient set ca_cert and vo defaults Fix #6934

### LINKED ISSUE #6934: Configuration handling for `ca_cert` and `vo` not falling back properly to defaults
### Description

I guess this is an edge scenario mostly for dev cases that use `userpass`, yet in the rucio lib, when initializing the `Client` instance without providing a `rucio.cfg` file and specifying `ca_cert` or `vo`, the fallback to the defaults (Mozilla CA bundle and `def` does not occur as expected. This causes the initialization process to fail.

On the contrary, if a `rucio.cfg` file is present but yet, `ca_cert` and `vo` are not defined within it, the fallback mechanisms work as intended.

Shouldn't the two approaches above lead to the same result?

**Suggested Fix:**
Catch also the `ConfigNotFound` exception (and not only `NoOptionError`, `NoSectionError`) when trying to fetch the `ca_cert` and `vo`:

https://github.com/rucio/rucio/blob/158081e7dcb66894a788fd98b24b4c85646c01f0/lib/rucio/client/baseclient.py#L163-L172

and

https://github.com/rucio/rucio/blob/158081e7dcb66894a788fd98b24b4c85646c01f0/lib/rucio/client/baseclient.py#L186-L198

### Steps to reproduce

_Use this within an environment that has no `rucio.cfg`._

```python

client_config = {
    'rucio_host': "https://rucio.ip",
    'auth_host': "https://rucio.ip",
    'auth_type': "userpass",
    'creds': {"username": "ddmlab", "password": "secret"},
    'account': "root"
}

Client(**client_config)

Geogouz:
- Allow fallbacks to default `ca_cert` and `vo` as mentioned in #6934.
- Minor docu update for the `BaseClient` constructor.

### DIFF
--- a/lib/rucio/client/baseclient.py
+++ b/lib/rucio/client/baseclient.py
@@ -94,14 +94,12 @@ class BaseClient:
         """
         Constructor of the BaseClient.
         :param rucio_host: The address of the rucio server, if None it is read from the config file.
-        :param rucio_port: The port of the rucio server, if None it is read from the config file.
         :param auth_host: The address of the rucio authentication server, if None it is read from the config file.
-        :param auth_port: The port of the rucio authentication server, if None it is read from the config file.
         :param account: The account to authenticate to rucio.
-        :param use_ssl: Enable or disable ssl for commucation. Default is enabled.
         :param ca_cert: The path to the rucio server certificate.
         :param auth_type: The type of authentication (e.g.: 'userpass', 'kerberos' ...)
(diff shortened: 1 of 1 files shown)


voetberg:
Can you change the doc strings to be a different commit? It's unrelated to the auth stuff. 

Otherwise, you should write a test to verify that this works as expected in https://github.com/rucio/rucio/blob/master/tests/test_clients.py

PR #6951, opened 2024-07-17
Testing: Add type annotations to core/replica_sorter

rdimaio:
Part of #6588

### DIFF
--- a/lib/rucio/common/types.py
+++ b/lib/rucio/common/types.py
@@ -239,6 +239,12 @@ class TokenValidationDict(TypedDict):
 
 
 class IPDict(TypedDict):
+    ip: Optional[str]
+    fqdn: Optional[str]
+    site: Optional[str]
+
+
+class IPWithLocationDict(TypedDict):
     ip: str
     fqdn: str
     site: str
--- a/lib/rucio/common/types.py
+++ b/lib/rucio/common/types.py
@@ -28,7 +28,7 @@ if TYPE_CHECKING:
     from datetime import datetime
 
     from rucio.common.constants import SUPPORTED_PROTOCOLS_LITERAL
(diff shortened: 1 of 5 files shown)


lib/rucio/common/types.py
```diff
@@ -236,10 +236,10 @@ class TokenValidationDict(TypedDict):
 
 class IPDict(TypedDict):
     ip: str
```
voetberg: Is there a reason for the IP to be required? There's a lot of `ignore`s further down the chain bc of ip

lib/rucio/core/replica_sorter.py
```diff
@@ -228,7 +245,11 @@ def __get_distance_custom(rse: Union[tuple, str], client_location: dict) -> floa
     return cache_val
 
 
-def site_selector(replicas, site, vo):
+def site_selector(
+        replicas: dict[str, Any],
```
voetberg: Is there really a dedicated replica dictionary? I think that would make more sense

rdimaio:
Rebased in latest force push

lib/rucio/core/replica_sorter.py
rdimaio: Addressed in latest revision - created a `ReplicaDict` type

lib/rucio/common/types.py
rdimaio: Good point; I've now split this type into two types:
- `IPWithLocationDict` is the one returned by `detect_client_location`, which always returns `ip`, `fqdn` and `site` as `str`
- `IPDict` is used in all other contexts, where any of those values might be `None`

rdimaio:
Latest revision:
- fix merge conflicts

bari12 (approved):

PR #6955, opened 2024-07-18
Recovery: possibly report non existing dids declared as bad

### LINKED ISSUE #6741: Declare Bad File Replcas fails silently

faluchet:
also when path is None - Fix #6741

### DIFF
--- a/lib/rucio/core/replica.py
+++ b/lib/rucio/core/replica.py
@@ -317,9 +317,6 @@ def __declare_bad_file_replicas(pfns, rse_id, reason, issuer, status=BadFilesSta
                 new_bad_replica.save(session=session, flush=False)
                 declared = True
 
-        else:
-            unknown_replicas.append(f"{path_pfn_dict[path]} Unknown replica")
-
         if not declared:
             if already_declared:
                 unknown_replicas.append('%s %s' % (path_pfn_dict.get(path, '%s:%s' % (scope, name)), 'Already declared'))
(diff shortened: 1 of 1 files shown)


faluchet:
Unfortunately, https://github.com/rucio/rucio/pull/6792 introduced a regression.

The variable path , as returned by __exist_replicas in L306 can have the None value when __exist is false in L309 and the else branch is entered in L320.

Therefore path_pfn_dict[path] may raise a KeyError exception, that won't be caught anywhere.

In such circumstances, the minos daemon crashes indefinitely while trying to process entries from the 'bad_pfns' table unless a targeted manipulation from an admin operator is performed. Observed in ATLAS with 150k+ cumulated entries on a bad day.

The usage of path_pfn_dict.get(path,"some default") is safer, although possibly still un-informative.
However, the handling of not declared replicas is performed later in the code, from L323 and on (where a check for None path indeed already existed). I opted to add the non existing did report in the same block, with an attempt to include the non-existing scope and name, as returned by __exist_replicas in L306.

rdimaio (dismissed):
LGTM, tests failed for minor flake8 issue about too many blank lines

rdimaio (approved):

Issue #6680, opened 2024-04-15
Usage of `assert` outside of tests


rdimaio:
## Description

Right now, we are using `assert` in production code, outside tests (see the linter log below).

Some usage examples:

https://github.com/rucio/rucio/blob/bc8fae0464e060d85c92bf93ee5603a78310b02f/bin/rucio-auditor#L58

https://github.com/rucio/rucio/blob/339744866e9550f0f068f8e3c9574a95ccaa7d59/bin/rucio-dumper#L128

https://github.com/rucio/rucio/blob/339744866e9550f0f068f8e3c9574a95ccaa7d59/lib/rucio/common/dumper/consistency.py#L44-L47

https://github.com/rucio/rucio/blob/339744866e9550f0f068f8e3c9574a95ccaa7d59/lib/rucio/core/lock.py#L83

This is not necessarily bad, but assertions are removed when Python is run with optimization requested (compiling to optimised byte code, `python -O ...` producing `*.opt-1.pyc` files). This is likely never going to be an issue for Rucio, but I still wanted to open an issue to discuss this. 

Additionally, when `assert` fails, it raises `AssertionError`. In the context of our usage of `assert`, I think we could raise more specific exceptions with clearer messages, which might also help from the point of view of debugging.

## To do
I would like us to decide on one of these options:
- We decide that the current production usage of `assert` is fine. In this case, we only need to set the linter to ignore `S101` (usage of `assert`) and no other changes will be needed.
- We decide that we want to replace our usage of `assert` with regular condition checking and custom exceptions. In this case, we'd need to make these changes.

## See also
- Ruff's docs: https://docs.astral.sh/ruff/rules/assert/
- Bandit's docs: https://bandit.readthedocs.io/en/latest/plugins/b101_assert_used.html


## Linter log
```
❯ ruff check --select S101                                                                                                                                                                                                                                                                                               ─╯
bin/rucio-admin:1198:5: S101 Use of `assert` detected
bin/rucio-auditor:51:5: S101 Use of `assert` detected
bin/rucio-auditor:58:5: S101 Use of `assert` detected
bin/rucio-auditor:72:5: S101 Use of `assert` detected
bin/rucio-dumper:99:9: S101 Use of `assert` detected
bin/rucio-dumper:117:9: S101 Use of `assert` detected
bin/rucio-dumper:128:13: S101 Use of `assert` detected
lib/rucio/client/lockclient.py:61:9: S101 Use of `assert` detected
[...]

PR #6982, opened 2024-07-29
Remove assert statements outside of tests 

voetberg:
Addresses #6680 

Removes asserts through a combination of swapping asserts to more verbose errors, minor refactors to avoid them in the first place, and manually ignoring the linter.

### DIFF
--- a/lib/rucio/common/dumper/__init__.py
+++ b/lib/rucio/common/dumper/__init__.py
@@ -81,7 +81,8 @@ def mkdir(dir_: "StrOrBytesPath") -> None:
     try:
         os.mkdir(dir_)
     except OSError as error:
-        assert error.errno == 17  # noqa: S101
+        if error.errno != 17:
+            raise error
 
 
 def cacert_config(config: "ModuleType", rucio_home: str) -> Optional[Union["FileDescriptorOrPath", Literal[False]]]:
--- a/lib/rucio/common/dumper/__init__.py
+++ b/lib/rucio/common/dumper/__init__.py
@@ -81,7 +81,7 @@ def mkdir(dir_: "StrOrBytesPath") -> None:
     try:
         os.mkdir(dir_)
     except OSError as error:
-        assert error.errno == 17
+        assert error.errno == 17  # noqa: S101
 
 
 def cacert_config(config: "ModuleType", rucio_home: str) -> Optional[Union["FileDescriptorOrPath", Literal[False]]]:
--- a/lib/rucio/common/dumper/consistency.py
+++ b/lib/rucio/common/dumper/consistency.py
@@ -42,7 +42,7 @@ class Consistency(data_models.DataModel):
             next_date_fname = data_models.Replica.download(
                 ddm_endpoint, next_date, cache_dir=cache_dir)
             if (prev_date_fname is None) or (next_date_fname is None):
-                raise ValueError("Both prev_data_fname and next_date_fname are required for subcommand='consistency'")
+                raise ValueError("Both prev_date_fname and next_date_fname are required for subcommand='consistency'; found prev_data_fname=%s and next_date_fname=%s" % (prev_date_fname, next_date_fname))
         elif subcommand == 'consistency-manual':
             pass
         else:
@@ -168,8 +168,8 @@ def min3(*values):
     Minimum between the 3 values ignoring None
     '''
     values = [value for value in values if value is not None]
-    if len(values) <= 0:
-        raise ValueError("Not enough values to sort.")  # Need a better error message, but this function doesn't have a lot of info.
+    if len(values) == 0:
+        raise ValueError("Input contains 0 non-null values.")
     return min(values)
 
 
@@ -304,7 +304,7 @@ def gnu_sort(file_path, prefix=None, delimiter=None, fieldspec=None, cache_dir=D
     memory and it is relatively fast if used with the environment variable
     LC_ALL set to C as in this function.
     '''
-    if (delimiter is None and fieldspec is not None) or (delimiter is not None and fieldspec is None):
+    if (delimiter is not None) ^ (fieldspec is not None):
         raise ValueError("Either both delimiter and fieldspec is set, or neither are.")
     if delimiter is None:
         cmd_line = 'LC_ALL=C sort {0} > {1}'
@@ -392,8 +392,12 @@ def _parse_args_consistency(args):
         error('The storage dump filename must be of the form '
               '"dump_YYYYMMDD" where the date correspond to the date '
               'of the newest files included')
+
+    date_str = date_str.group(1)
+    if date_str is None:
+        error('Invalid date {0}'.format(date_str))
     try:
-        args_dict['date'] = date = datetime.datetime.strptime(date_str.group(1), '%Y%m%d')
+        args_dict['date'] = date = datetime.datetime.strptime(date_str, '%Y%m%d')
     except ValueError:
         error('Invalid date {0}'.format(date_str))
 
--- a/lib/rucio/common/dumper/consistency.py
+++ b/lib/rucio/common/dumper/consistency.py
@@ -41,10 +41,12 @@ class Consistency(data_models.DataModel):
                 ddm_endpoint, prev_date, cache_dir=cache_dir)
             next_date_fname = data_models.Replica.download(
                 ddm_endpoint, next_date, cache_dir=cache_dir)
-            assert prev_date_fname is not None
-            assert next_date_fname is not None
+            if (prev_date_fname is None) or (next_date_fname is None):
+                raise ValueError("Both prev_data_fname and next_date_fname are required for subcommand='consistency'")
+        elif subcommand == 'consistency-manual':
+            pass
         else:
-            assert subcommand == 'consistency-manual'
+            raise ValueError("subcommand %s not accepted, choice from ('consistency','consistency-manual')" % subcommand)
 
         prefix_components = path_parsing.components(dumper.ddmendpoint_url(ddm_endpoint))
 
@@ -166,7 +168,8 @@ def min3(*values):
     Minimum between the 3 values ignoring None
     '''
     values = [value for value in values if value is not None]
-    assert len(values) > 0
+    if len(values) <= 0:
+        raise ValueError("Not enough values to sort.")  # Need a better error message, but this function doesn't have a lot of info.
     return min(values)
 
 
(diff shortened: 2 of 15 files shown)


lib/rucio/core/lock.py
```diff
@@ -96,21 +97,23 @@ def get_dataset_locks_bulk(dids: Iterable[dict[str, Any]], *, session: "Session"
...
                     yield lock_dict
+        else:
+            raise ValueError("Can only get locks for datasets and containers.")
 
 
 @stream_session
 def get_dataset_locks_by_rse_id(rse_id: str, *, session: "Session") -> Iterator[dict[str, Any]]:
     """
-    Get the dataset locks of an RSE.
+    Get the dataset locks of an RSE.l   r
```
rdimaio:
```suggestion
    Get the dataset locks of an RSE.
```

lib/rucio/core/oidc.py
```diff
@@ -96,11 +96,13 @@ def _token_cache_get(
...
-    except Exception:
-        METRICS.counter('token_cache.invalid').inc()
+    if isinstance(value, str):
+        try:
+            payload = JWT().unpack(value).payload()
+        except Exception:
+            METRICS.counter('token_cache.invalid').inc()
+            return None
+    else:
         return None
```
rdimaio: Beforehand, `METRICS.counter('token_cache.invalid').inc()` would be called even in case of an `AssertionError` (under `except Exception`), but in the new version it will just `return None` when `isinstance(value, str)` is false.

```suggestion
    if isinstance(value, str):
        try:
            payload = JWT().unpack(value).payload()
        except Exception:
            METRICS.counter('token_cache.invalid').inc()
            return None
    else:
        METRICS.counter('token_cache.invalid').inc()
        return None
```

There might be a way to simplify this to avoid duplicating `METRICS.counter('token_cache.invalid').inc()` and `return None`, but nothing comes to mind at the moment

lib/rucio/db/sqla/session.py
```diff
@@ -272,7 +273,6 @@ def get_maker() -> sessionmaker:
         May assign __MAKER if not already assigned.
     """
     global _MAKER, _ENGINE
-    assert _ENGINE
```
rdimaio, line 275: should there a be `RuntimeError` raised here in place of the assert?

lib/rucio/db/sqla/session.py
```diff
@@ -291,7 +291,8 @@ def get_session() -> scoped_session:
             get_maker()
         finally:
             _LOCK.release()
-    assert _MAKER
+    if not _MAKER:
+        raise RuntimeError("Database schema not in place, cannot make session.")
```
rdimaio: I think this would be a better descriptor of the error:

```suggestion
        raise RuntimeError("Session factory is not defined.")
```

lib/rucio/daemons/auditor/srmdumps.py
```diff
@@ -86,7 +86,7 @@ def mkdir(dir_: str) -> None:
     try:
         os.mkdir(dir_)
     except OSError as e:
-        assert e.errno == 17
+        assert e.errno == 17  # noqa: S101
```
rdimaio: Is there a reason to add `noqa` here instead of transforming it into a check?

lib/rucio/common/dumper/__init__.py
```diff
@@ -81,7 +81,7 @@ def mkdir(dir_: "StrOrBytesPath") -> None:
     try:
         os.mkdir(dir_)
     except OSError as error:
-        assert error.errno == 17
+        assert error.errno == 17  # noqa: S101
```
rdimaio: Is there a reason to add `noqa` here instead of transforming it into a check?

lib/rucio/common/dumper/data_models.py
```diff
@@ -235,9 +232,10 @@ def __init__(self, *args):
         if len(args) == 8:
             logger.warning('Extra parameter\nrse: %s\ndataset: %s\n', self.rse, self.name)
             self.state = args[7]
-        else:
+        elif len(args) <= 8:
             self.state = None
-        assert len(args) <= 8
+        else:
+            raise ValueError("Too many arguments, must be 8 or less. Instead passed %s" % len(args))
```
rdimaio: The functionality here is slightly different from before - before `self.state = None` was run if `len(args) != 8` (so even if `len(args > 8`), now it only runs if `len(args) <= 8`. 

To retain the same logic it might be worth doing:

```
if len(args) == 8:
    ...
else:
    self.state = None
if len(args) > 8:
   raise ValueError("Too many arguments, must be 8 or less. Instead passed %s" % len(args))
```

lib/rucio/common/dumper/data_models.py
```diff
@@ -264,7 +262,10 @@ def __init__(self, *args):
 
         if len(args) == 8:
             logger.warning('Missing parameter\nrse: %s\ndataset: %s\n', self.rse, self.name)
-        assert len(args) <= 9
+        elif len(args) <= 9:
+            pass
+        else:
+            raise ValueError("Too many arguments. Must be 9 or less, instead passed %s" % len(args))
```
rdimaio:
```suggestion
        elif len(args) > 9:
            raise ValueError("Too many arguments. Must be 9 or less, instead passed %s" % len(args))
```

lib/rucio/common/dumper/consistency.py
```diff
@@ -388,10 +392,8 @@ def _parse_args_consistency(args):
         error('The storage dump filename must be of the form '
               '"dump_YYYYMMDD" where the date correspond to the date '
               'of the newest files included')
-    date_str = date_str.group(1)
-    assert date_str is not None
     try:
-        args_dict['date'] = date = datetime.datetime.strptime(date_str, '%Y%m%d')
+        args_dict['date'] = date = datetime.datetime.strptime(date_str.group(1), '%Y%m%d')
     except ValueError:
         error('Invalid date {0}'.format(date_str))
```
rdimaio: `date_str.group(1)` could return `None`: https://docs.python.org/3/library/re.html#re.Match.group

So the return type of `date_str.group(1)` here is `Optional[str]`, and [`strptime`](https://docs.python.org/3/library/datetime.html#datetime.datetime.strptime) raises a `TypeError` if `None` is passed, not `ValueError`:

```
import datetime
from typing import Optional

date_str: Optional[str] = None
x = datetime.datetime.strptime(date_str, '%Y%m%d')
```

results in:

```
ERROR!
  File "<main.py>", line 6, in <module>
TypeError: strptime() argument 1 must be str, not None
```

So I think it would be best to add this before the `try/except` block:

```
if date_str is None:
    error('error parsing date from filename')
```

lib/rucio/common/dumper/consistency.py
```diff
@@ -301,7 +304,8 @@ def gnu_sort(file_path, prefix=None, delimiter=None, fieldspec=None, cache_dir=D
     memory and it is relatively fast if used with the environment variable
     LC_ALL set to C as in this function.
     '''
-    assert (delimiter is None and fieldspec is None) or (delimiter is not None and fieldspec is not None)
+    if (delimiter is None and fieldspec is not None) or (delimiter is not None and fieldspec is None):
```
rdimaio: This is `XOR(delimiter, fieldspec)`, so it could be simplified either like this:

```suggestion
    if (delimiter is not None) != (fieldspec is not None):
```

or like this:

```suggestion
    if (delimiter is not None) ^ (fieldspec is not None):
```

I think the first one is clearer in this context, especially cause the `ValueError` message perfectly describes the issue, but the second one is in theory the correct way to write a XOR operation

See also: https://stackoverflow.com/questions/432842/how-do-you-get-the-logical-xor-of-two-variables-in-python

lib/rucio/common/dumper/consistency.py
```diff
@@ -166,7 +168,8 @@ def min3(*values):
     Minimum between the 3 values ignoring None
     '''
     values = [value for value in values if value is not None]
-    assert len(values) > 0
+    if len(values) <= 0:
+        raise ValueError("Not enough values to sort.")  # Need a better error message, but this function doesn't have a lot of info.
```
rdimaio: I think this kind of message is fine considering the scope of the function - it's up to the caller to possibly propagate the exception with a more generic message, but in this context it makes sense to raise this kind of message. However, the name of the function indicates 3 values, even though this limit is not enforced anywhere in the function - that's a bit weird, but it's outside of the scope of this PR. I opened this issue for it https://github.com/rucio/rucio/issues/6990

len(values) can never be negative, so it should only be checked for 0.

```suggestion
    if len(values) == 0:
        raise ValueError("Input contains 0 non-null values.")
```

lib/rucio/common/dumper/consistency.py
```diff
@@ -41,10 +41,12 @@ def dump(cls, subcommand, ddm_endpoint, storage_dump, prev_date_fname=None, next
                 ddm_endpoint, prev_date, cache_dir=cache_dir)
             next_date_fname = data_models.Replica.download(
                 ddm_endpoint, next_date, cache_dir=cache_dir)
-            assert prev_date_fname is not None
-            assert next_date_fname is not None
+            if (prev_date_fname is None) or (next_date_fname is None):
+                raise ValueError("Both prev_data_fname and next_date_fname are required for subcommand='consistency'")
```
rdimaio: Could be worth mentioning which one of the two is `None` (or if both are) in the exception message. Maybe it could either be:

```suggestion
            if prev_date_fname is None):
                raise ValueError("prev_data_fname is required for subcommand='consistency'")
            elif next_date_fname is None):
                raise ValueError("next_date_fname is required for subcommand='consistency'")
```

or keeping it in a single check and logging both values

rdimaio (commented):
LGTM, just some small comments.

I'm not sure why `assert`s were used so much in daemons in the first place, I wonder if there's a good reason for that and whether we're breaking something here (I don't think so, though; failing assert would raise exceptions anyway)

lib/rucio/common/dumper/consistency.py
voetberg: I have never seen that ^ notation for XOR, thanks for teaching me something new and fun

lib/rucio/db/sqla/session.py
voetberg: Yeah totally should be, it is in my local branch I just missed adding it to the commit :sob:

lib/rucio/daemons/auditor/srmdumps.py
voetberg: I'm sure I had a reason but it would be so much better to just pass the error forward when the errno isn't 17. Changing that now.

lib/rucio/common/dumper/__init__.py
```diff
@@ -81,7 +81,8 @@ def mkdir(dir_: "StrOrBytesPath") -> None:
     try:
         os.mkdir(dir_)
     except OSError as error:
-        assert error.errno == 17  # noqa: S101
+        if error.errno != 17:
+            raise OSError(error)
```
rdimaio: can probably just raise `error` here since it's already an `OSError`

lib/rucio/daemons/auditor/srmdumps.py
```diff
@@ -86,7 +86,8 @@ def mkdir(dir_: str) -> None:
     try:
         os.mkdir(dir_)
     except OSError as e:
-        assert e.errno == 17  # noqa: S101
+        if e.errno != 17:
+            raise OSError(e)
```
rdimaio: same here, can raise `e` directly as it's already an `OSError`

rdimaio (dismissed):

rdimaio (approved):

Issue #6669, opened 2024-04-11
`psycopg2-binary` should not be used in production - `psycopg2` should be used instead


rdimaio:
### Description

For PostgreSQL-related functions, Rucio uses the `psycopg2-binary` library:

https://github.com/rucio/rucio/blob/339744866e9550f0f068f8e3c9574a95ccaa7d59/requirements.txt#L34

However, the [`psycopg2` docs](https://www.psycopg.org/docs/install.html#psycopg-vs-psycopg-binary) say:
> The psycopg2-binary package is meant for beginners to start playing with Python and PostgreSQL without the need to meet the build requirements.
> If you are the maintainer of a published package depending on psycopg2 you shouldn’t use psycopg2-binary as a module dependency. For production use you are advised to use the source distribution.

See also: https://stackoverflow.com/questions/70330567/what-is-the-different-about-psycopg2-and-psycopg2-binary-python-package

Additionally there is also [`psycopg3`](https://www.psycopg.org/psycopg3/), which introduces many new features (the differences between `psycopg2` and `psycopg3` are listed [here](https://www.psycopg.org/psycopg3/docs/basic/from_pg2.html)), like support for static typing.

To do:
- If possible, we might want to consider upgrading from `psycopg2` to `psycopg3`
- If not, at the very least we should move from `psycopg2-binary` to `psycopg2` as recommended in their docs.

This is currently a blocker for this issue: https://github.com/rucio/rucio/issues/6666, which requires the [`psycopg2.sql`](https://www.psycopg.org/docs/sql.html) module.

### Steps to reproduce

N/A

Ban42:
Hi,
I started having a look on this, these are some notes about my analysis:

- Currently psycopg is used in:

   - core/did_meta_plugins/postgres_meta.py
   - db/sqla/session.py
   - In the requirements (requirements.txt and setuputil.py)

   I tried to re-implement the functionalities of "postgres_meta.py" with psycopg3 and they seem to work. The only difference is that the result of the query in psycopg2 is an `OrderedDict` while in psycopg3 is a normal `dict`.

   In the next days, I will check "session.py" and reproduce its functionalities with psycopg3

- Apparently in psycopg3 the binary installation is not deprecated for production systems (even though is advised), which could be an extra reason to move to psycopg3. Source is their official [Documentation](https://www.psycopg.org/psycopg3/docs/basic/install.html#binary-installation)

[...]

PR #6993, opened 2024-08-02
Database: Update psycopg2-binary intp psycopg3 #6669

Ban42:
This PR aims at upgrading all the calls to psycopg2 functions into psycopg3 ones. 
Three changes are done here. Each one have an independent commit:

1. Update the function calls in "core/did_meta_plugins/postgres_meta.py"
2. Remove the custom type caster from postgres' DECIMAL into Python's float as it is built-in psycopg3
3. Update the requirements setup

N.B. Some tests in this PR are failing because the containers used in the tests install `psycopg2-binary` and don't have `psycopg` (package name for psycopg3)

### DIFF
--- a/etc/docker/test/extra/rucio_postgres14.cfg
+++ b/etc/docker/test/extra/rucio_postgres14.cfg
@@ -1,5 +1,5 @@
 [database]
-default = postgres://rucio:rucio@postgres14/rucio
+default = postgresql+psycopg://rucio:rucio@postgres14/rucio
 schema = dev
 pool_size = 20
 max_overflow = 20
--- a/etc/docker/test/extra/rucio_postgres14.cfg
+++ b/etc/docker/test/extra/rucio_postgres14.cfg
@@ -1,5 +1,5 @@
 [database]
-default = postgresql://rucio:rucio@postgres14/rucio
-schema=dev
+default = postgres://rucio:rucio@postgres14/rucio
+schema = dev
 pool_size = 20
 max_overflow = 20
--- a/lib/rucio/core/did_meta_plugins/postgres_meta.py
+++ b/lib/rucio/core/did_meta_plugins/postgres_meta.py
@@ -80,6 +80,9 @@ class ExternalPostgresJSONDidMeta(DidMetaPlugin):
             user=user,
             password=password)
 
+        # makes Rucio receive Decimal from Postgres as float
+        self.client.adapters.register_loader("numeric", psycopg.types.numeric.FloatLoader)
+
         # set search_path to include database schema by default
         cur = self.client.cursor()
         cur.execute("SET search_path TO %s", db_schema)
--- a/lib/rucio/core/did_meta_plugins/postgres_meta.py
+++ b/lib/rucio/core/did_meta_plugins/postgres_meta.py
@@ -16,8 +16,8 @@ import json
 import operator
 from typing import TYPE_CHECKING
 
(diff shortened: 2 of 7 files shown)


Ban42:
Hi @robbarnsley, here is the PR you asked for in https://github.com/rucio/rucio/issues/6669 .
Let me know if you also want me to modify the containers to make the test pass.

Thanks for your time :)

robbarnsley:
Hi @Ban42 , at a guess i think [this](https://github.com/rucio/rucio/blob/master/etc/docker/test/extra/rucio_postgres14.cfg) connection string may need changing as it looks like it's still using the psycopg2 dialect when the connection string refers to `postgresql://`

[The recommended install](https://pypi.org/project/psycopg/) also specifies a "binary" package, which may need to be added to the requirements..

Ban42:
Hi @robbarnsley ,
First of all, thanks for waiting my answer for long time, really appreciated.
Coming to code, just to be sure I understood the changes you asked me to do:

1. About [this](https://github.com/rucio/rucio/blob/master/etc/docker/test/extra/rucio_postgres14.cfg) connection string, not being an expert, I had a look at [connection URI in postgresql documentation](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING-URIS) and they claim that:

> The URI scheme designator can be either `postgresql://` or `postgres://`

So I believe this 'dialect' on connection string is something related to postgres, not psycopg. However, I changed it.

2. By requirements, I assumed the file `setuputil.py` and I performed a change there. Do I need to update requirements also in some other file?
3. You prefer me to put in setuputil.py `'postgresql': ['psycopg[binary,pool]']` as defined in [the page you linked](https://pypi.org/project/psycopg/) or you prefer `'postgresql': "psycopg[binary]"` as suggested in [psycopg documentation](https://www.psycopg.org/psycopg3/docs/basic/install.html#installation)? For now I assumed psycopg documentation.

Cheers!

robbarnsley:
Hi @Ban42 ,

No problem, thanks for making those changes.

I was looking here: https://docs.sqlalchemy.org/en/20/dialects/postgresql.html#module-sqlalchemy.dialects.postgresql.psycopg which states:

>  psycopg is the package and module name for version 3 of the psycopg database driver, formerly known as psycopg2. This driver is different enough from its psycopg2 predecessor that SQLAlchemy supports it via a totally separate dialect; support for psycopg2 is expected to remain for as long as that package continues to function for modern Python versions, and also remains the default dialect for the postgresql:// dialect series.

i.e. connection string should be `postgresql+psycopg://user:password@host:port/dbname[?key=value&key=value...]`

robbarnsley:
Regarding requirements, I think you need to also put this in the `requirements/requirements.server.in`.

It would be best to include `pool` also, although it's strictly optional it's pretty necessary for production environments.

Ban42:
Hi @robbarnsley,
I had a look at the SQLAlchemy documentation, thanks for pointing out. 
I did the requested changes but I am not 100% sure of the correct syntax in the requirements files.
Please let me know if there is some other actions needed on my side.
Cheers

bari12:
What do we do with this? Without the tests working this is really tricky to merge 😄

Geogouz:
> What do we do with this? Without the tests working this is really tricky to merge 😄

The problem that this issue is tackling has been solved by https://github.com/rucio/rucio/pull/7202

We can close it.

PR #6999, opened 2024-08-05
Core: Always sort RSEs alphabetically #6998

### LINKED ISSUE #6998: RSEs are not sorted
### Description

https://github.com/rucio/rucio/blob/8c4672c64b8784fd9f9cdc9e3516d7d83ecd22c9/lib/rucio/core/rse.py#L746-L749

The ordering is only done when `filters` is empty. That, by itself, appears to be an oversight. When called through the gateway layer, this block is never executed because `filter` always contains at least the `vo` key.

### Steps to reproduce

```python
from rucio.gateway.rse import list_rses
for rse in list_rses():
    print(rse['rse'])
```

### Rucio Version

_No response_

### Additional Information

_No response_

dchristidis:


rdimaio (commented):
lgtm - might be worth having a test that ensures the returned RSEs are sorted

rdimaio (approved):

dchristidis:
I have the impression such a test may never be conclusive; we can’t guarantee that, without the `ORDER BY` clause in the query, the database will return the results in the wrong order.

rdimaio:
> I have the impression such a test may never be conclusive; we can’t guarantee that, without the `ORDER BY` clause in the query, the database will return the results in the wrong order.

you're right, good point

Issue #6580, opened 2024-03-18
Allow permissions to optionally return a message


dynamic-entropy:
### Description

The current permission checks only tell a user that they are not allowed to issue a specific command. However they do not explain any reason why. 
For e.g. a permission check if a issuer can or cannot create a rule returns
`Account joe cannot add a replication rule`.

https://github.com/rucio/rucio/blob/dc97725de99d438966d412c81918a52755b63baa/lib/rucio/api/rule.py#L89-L90

This is often not enough when there are multiple criteria in an experiments policy package to determine if a rule can be created or not. 
This has become increasingly important for CMS due to user policy around rule creation that are `AutoApproved`. 

We would like to optionally, attach a reason when the permission fails to let the user if there is an unintended mistake or in cases when they should get required permissions from ops.Joe

### Change

Modify the `has_permissions` call to accept an optional (For backward compatibility and opt-in support) "Message" (reason for failure) in the return value and raise it with the `Access Denied` exception.

bari12:
I think this is a good idea and should not be too complicated to implement.

jamesp-epcc:
I've discussed this with Martin and I'm happy to take a look at it if people want me to? I think it should be possible to make it backwards compatible so that policy packages that implement their own permission checks and don't return a message will still work, probably by adding some code to the dispatcher in `core/permission/__init__.py`.

dynamic-entropy:
Hello James
It is good with us if you take this. I am not sure if @alexanderrichards has already started working on it.

jamesp-epcc:
I am finally looking at this now. I plan to do the following:

1. Add a new class `PermissionResult` containing a boolean and a string for the message. The return value from `has_permission` will be of this type.
2. Implement `__bool__()` to maintain compatibility with old code that expects a raw boolean.
3. The top level `has_permission` function in `__init__.py` will check the type returned by the policy packages. If it's a raw boolean, it will wrap the value in a `PermissionResult` object with an empty string, to maintain compatibility with old policy packages that return raw booleans.
4. The code in the gateway layer (mostly) that calls `has_permission` can be modified to expect a `PermissionResult` object and report the message when it throws an exception. This could be done incrementally.

I've already done some tests to ensure this will work, but I wanted to check it's acceptable before I spend time updating all the callers to report the message.

PR #7001, opened 2024-08-06
Policies: allow permission checks to optionally return message #6580

jamesp-epcc:
This PR allows permission checks to optionally return a message giving further details on the reason for failure. It also modifies the gateway layer to include this message in the exceptions it raises. I have tried to make it as backward compatible as possible - permission checks that return a simple boolean and callers that expect a boolean should both still work as before.

### DIFF
--- a/lib/rucio/gateway/account.py
+++ b/lib/rucio/gateway/account.py
@@ -59,8 +59,9 @@ def add_account(
     validate_schema(name='account', obj=account, vo=vo)
 
     kwargs = {'account': account, 'type': type_}
-    if not rucio.gateway.permission.has_permission(issuer=issuer, vo=vo, action='add_account', kwargs=kwargs, session=session):
-        raise rucio.common.exception.AccessDenied('Account %s can not add account' % (issuer))
+    auth_result = rucio.gateway.permission.has_permission(issuer=issuer, vo=vo, action='add_account', kwargs=kwargs, session=session)
+    if not auth_result.allowed:
+        raise rucio.common.exception.AccessDenied('Account %s can not add account. %s' % (issuer, auth_result.message))
 
     internal_account = InternalAccount(account, vo=vo)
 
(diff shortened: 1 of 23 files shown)


lib/rucio/gateway/credential.py
```diff
@@ -57,13 +57,15 @@ def get_signed_url(
...
-                                                                                                                                      operation,
-                                                                                                                                      url,
-                                                                                                                                      lifetime))
+    auth_result = permission.has_permission(issuer=account, vo=vo, action='get_signed_url', kwargs=kwargs, session=session)
+    if not auth_result.allowed:
+        raise exception.AccessDenied('Account %s can not get signed URL for rse=%s, service=%s, operation=%s, url=%s, lifetime=%s. %s' % (account,
+                                                                                                                                          rse,
+                                                                                                                                          service,
+                                                                                                                                          operation,
```
voetberg: Is there a reason for this to be across so many lines?

lib/rucio/gateway/account.py
```diff
@@ -59,8 +59,9 @@ def add_account(
     validate_schema(name='account', obj=account, vo=vo)
 
     kwargs = {'account': account, 'type': type_}
-    if not rucio.gateway.permission.has_permission(issuer=issuer, vo=vo, action='add_account', kwargs=kwargs, session=session):
-        raise rucio.common.exception.AccessDenied('Account %s can not add account' % (issuer))
+    auth_result = rucio.gateway.permission.has_permission(issuer=issuer, vo=vo, action='add_account', kwargs=kwargs, session=session)
+    if not auth_result.allowed:
```
voetberg, line 63: Maybe I don't completely understand how you've set this up, but shouldn't you be able to just do `if not auth_result`? `PermissionResult.__bool__` already returns allowed.

lib/rucio/gateway/credential.py
jamesp-epcc: I don't think so. It was already formatted like that so I just added to it, but I could reformat it.

lib/rucio/gateway/account.py
jamesp-epcc: Yes, that would also work. I thought this might make it clearer that the type is a `PermissionResult` rather than just a bool but I don't have strong feelings either way.

lib/rucio/gateway/credential.py
voetberg: Good a time as any to do so!

jamesp-epcc: Done!

voetberg (approved):

bari12:
I wonder here a bit with adding this `%s` for the message at the end of each exception message, or if we should do this more targeted?

jamesp-epcc:
I'm happy to reconsider how the messages are formatted/used. Could you give an example of what you would prefer?

bari12:
My thinking was mostly only to add it in cases where we know already that there is a return message, and then make it more targeted for that return message, but not to add it, per default to each.

jamesp-epcc:
That would make sense. I could rework the PR so that it just adds the infrastructure and maybe a couple of examples of the messages to show how they can be used. Though I guess the downside of this approach is that if some experiment added a return message in their own policy package, it wouldn't get displayed without changes to the main Rucio code.

PR #7005, opened 2024-08-06
Core: Add option to order_by parameters in core.did.list_parent_dids

### LINKED ISSUE #7002: `list_parent_dids`: add functionality to pick sorting parameter for query
### Description

Follow-up from a conversation with @mlassnig regarding the tape metadata plugin for ATLAS. In our context, we'd want to sort the `list_parent_dids` by `created_at`, but we can add a generic `sort_parameter` input to allow the function to be called with a different sorting parameter

In future, we might want to abstract this out into a decorator, so that we can allow caller functions to call query-based functions with specific sorting parameters rather than hardcoding them.

### Motivation

_No response_

### Change

_No response_

rdimaio:
fixes #7002

- not adding this to gateway and client for now. Before we do that, let's consider:
    - if we want to abstract this "dynamic `order_by` parameters" logic to a decorator, and:
    - if we want to provide this at gateway or client levels, instead of just having it in core for use by other core function.

- no new unit test for this particular case as `list_parent_dids` is currently not unit tested. We should address all these methods without unit tests as part of the test overhaul work.

### DIFF
--- a/lib/rucio/core/did.py
+++ b/lib/rucio/core/did.py
@@ -1869,11 +1869,15 @@ def list_parent_dids(
     stmt = select(
         models.DataIdentifierAssociation.scope,
         models.DataIdentifierAssociation.name,
-        models.DataIdentifierAssociation.did_type
+        models.DataIdentifierAssociation.did_type,
+        models.DataIdentifier.created_at
     ).where(
         and_(models.DataIdentifierAssociation.child_scope == scope,
-             models.DataIdentifierAssociation.child_name == name)
+             models.DataIdentifierAssociation.child_name == name,
+             models.DataIdentifier.scope == models.DataIdentifierAssociation.scope,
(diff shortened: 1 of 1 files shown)


lib/rucio/core/did.py
```diff
@@ -1861,19 +1862,29 @@ def list_parent_dids(
 
     :param scope:     The scope.
     :param name:      The name.
+    :param order_by:  List of parameters to order the query by.
```
voetberg: The docstring should list the allowed parameters or say explicitly what they're constrained by

rdimaio: Added in latest force-push

voetberg (approved):
LGTM, we should probably make issues adding this option to the client and the testing for list_parent_dids in particular so we don't lose track of that.

Issue #6987, opened 2024-07-30
Use the rich library to enhance the command-line client commands


ericbanzuzi:
### Description

Many of the Rucio commands are displayed or can be displayed in a table. Currently, the command-line client uses the tabulate library, which is not aesthetically good and displays long tables poorly. This will introduce rich library as the library to be used to display table formatted output,  and this introduces displaying the outputs in a pager by default if it does not fit the terminal size. The content of the outputs will remain the same.

We are not yet sure about how displaying the content in a pager by default will work. This will remain to be seen in a testing phase, where it will be evaluated whether it is better to leave displaying or not displaying the content in a pager as an option for the users through the configuration file or as an arguments while running the commands.  For example **Git** uses a pager by default. 

The colour palette used to highlight the different Rucio constants, as well as the chosen spinner style, 
will remain open for change. 

### Motivation

The outputs from the command-line client are generally not aesthetically good and the long tables are displayed poorly. The idea is to make the client more visually appealing and improve the user experience.

### Change

- Add rich as a dependency
- Add configuration option to use rich in the clients through the configuration file
- All commands where the output is a table now use rich
- Introduce coloured highlighting
- Introduce spinners for commands that may take longer to run  (around 1 to 5s)
- Introduce rich logging and exceptions for prettier logs in the command-line client
- Add two paths for the read commands: one that uses rich and the other that uses tabulate
- Display outputs in a pager by default
- Add argument --no-pager as an option

- Rich widgets to be used:
   Table
   Logging  (RichHandler)
   Spinners (Status)
   Text

- Changed commands in `bin/rucio`:
   whoami 
   get-metadata
   rule-info
   list-file-replicas
   list-dataset-replicas
   list-account-usage
   list-suspicious-replicas
   list-files
   list-rules
   list-rse-usage
   list-content
   list-rules-history
   list-dids
   list-parent-datasets
   list-account-limits 
   stat
   list-scopes
   list-rses
   

- Changed commands in `bin/rucio-admin`:
  account list
  account info
  account list-identities
  account list-attribute
  rse list
  rse info
  rse get-attribute
  rse get-distance
  rse list-qos-policies
  scope list
  subscription list
  data import
  data export
   

The changes will be added during minor releases but they will be hidden and only available for advanced users to test them in order to get feedback. The changes will be eventually released in the next major release as a new feature, and as the default option for output.

An example of a simple two column rich table (get-metadata):
[...]

PR #7012, opened 2024-08-08
Clients: Rich command-line client outputs #6987

ericbanzuzi:
For more details, view issue #6987.

###  Key changes:
-  Adds rich as a dependency
- Adds the CLI outputs using Rich for all the commands in `bin/rucio` and `bin/rucio-admin`.
- Adds a configuration option to choose between Rich or tabulate for CLI output.  Users can specify this in the config file by adding `cli = rich` in the `[experimental]` section. If the option `cli` does not exist or has a value other than `rich`, the tabulate CLI will be used by default.
- Implemented paging for outputs (`less -FRSXMK`) when the output exceeds the terminal size.
- Option `--no-pager ` to disable paging.

### Pager:

The paging is now done with `pydoc`, similarly as rich `console.pager()` does it. Here the steps of `console.pager()` is done manually so that the the command methods return the `SUCCESS` code before displaying the output and waiting for pager response. Rich only allows to use `console.pager()` as through a context manager, meaning that the method would only return `SUCCESS` when the user exits the pager.

### Dimming colors in rows:

I've decided to only apply dimming to have alternating coloring of rows only to the tables in which the numbe rof entries in unknown beforehand. This would mean that it is applied to all tables except the ones showing key-value pairs. The commands without alternating colors are: `rucio whoami`, `rucio get-metadata`, `rucio stat`, ` rucio list-rse-attributes`, `rucio rule-info`, `rucio-admin subscription list`, `rucio-admin account info`,  `rucio-admin rse get-attribute` and `attributes` `settings` `protocols` in `rucio-admin rse info`. I felt like dimming in the key-value tables was not such a natural effect.

### Styles:

I tried to choose a colour palette that highlights states with semantic colours to indicate whether something is positive, negative or neutral. For other types, the chosen colours are contrast colours, aiming to give clear separability without imposing emotional connotations. The chosen colours in a default mac terminal can be seen below:

cli_palette
cli_plaette_on_white

They might differ slightly in other terminal applications.

### DIFF
--- a/bin/rucio
+++ b/bin/rucio
@@ -32,11 +32,19 @@ from datetime import datetime
 from functools import wraps
 from typing import Optional
 
+from rich.console import Console
+from rich.padding import Padding
+from rich.status import Status
+from rich.text import Text
+from rich.theme import Theme
+from rich.traceback import install
+from rich.tree import Tree
 from tabulate import tabulate
 
 # rucio module has the same name as this executable module, so this rule fails. pylint: disable=no-name-in-module
 from rucio import version
 from rucio.client import Client
+from rucio.client.richclient import MAX_TRACEBACK_WIDTH, MIN_CONSOLE_WIDTH, CLITheme, generate_table, get_cli_config, get_pager, print_output, setup_rich_logger
 from rucio.common.client import detect_client_location
 from rucio.common.config import config_get, config_get_float
 from rucio.common.constants import ReplicaState
@@ -74,9 +82,15 @@ FAILURE = 1
 DEFAULT_SECURE_PORT = 443
 DEFAULT_PORT = 80
 
+# Setting up rich console.
+console = Console(theme=Theme(CLITheme.LOG_THEMES), soft_wrap=True)
+console.width = max(MIN_CONSOLE_WIDTH, console.width)
+
 logger = logging.log
 gfal2_logger = logging.getLogger("gfal2")
 tablefmt = 'psql'
+cli_config = get_cli_config()
+spinner = Status('Initializing spinner', spinner=CLITheme.SPINNER, spinner_style=CLITheme.SPINNER_STYLE, console=console)
 
 
 def setup_gfal2_logger(logger):
@@ -166,18 +180,21 @@ def exception_handler(function):
                 used_account = None
                 try:  # get the configured account from the configuration file
                     used_account = '%s (from rucio.cfg)' % config_get('client', 'account')
-                except:
+                except Exception:
                     pass
                 try:  # are we overridden by the environment?
                     used_account = '%s (from RUCIO_ACCOUNT)' % os.environ['RUCIO_ACCOUNT']
-                except:
+                except Exception:
                     pass
                 logger.error('Specified account %s does not have an associated identity.' % used_account)
             else:
-                logger.debug(traceback.format_exc())
+                if logger.level == logging.DEBUG and cli_config == 'rich':
+                    logger.exception(error)
+                else:
+                    logger.debug(traceback.format_exc())
                 contact = config_get('policy', 'support', raise_exception=False)
                 support = ('Please follow up with all relevant information at: ' + contact) if contact else ''
-                logger.error('\nThe object is missing this property: %s\n'
+                logger.error('The object is missing this property: %s\n'
                              'This should never happen. Please rerun the last command with the "-v" option to gather more information.\n'
                              '%s' % (str(error), support))
             return FAILURE
@@ -194,13 +211,20 @@ def exception_handler(function):
                 devnull = os.open(os.devnull, os.O_WRONLY)
                 os.dup2(devnull, sys.stdout.fileno())
                 return SUCCESS
-            logger.debug(traceback.format_exc())
-            logger.error(error)
+
+            if cli_config == 'rich':
+                if logger.level == logging.DEBUG:
+                    logger.exception(error)
+                else:
+                    logger.error(error)
+            else:
+                logger.debug(traceback.format_exc())
+                logger.error(error)
             contact = config_get('policy', 'support', raise_exception=False)
             support = ("If it's a problem concerning your experiment or if you're unsure what to do, please follow up at: %s\n" % contact) if contact else ''
             contact = config_get('policy', 'support_rucio', default='https://github.com/rucio/rucio/issues')
             support += "If you're sure there is a problem with Rucio itself, please follow up at: " + contact
-            logger.error('\nRucio exited with an unexpected/unknown error.\n'
+            logger.error('Rucio exited with an unexpected/unknown error.\n'
                          'Please rerun the last command with the "-v" option to gather more information.\n'
                          '%s' % support)
             return FAILURE
@@ -294,8 +318,14 @@ def whoami_account(args):
     """
     client = get_client(args)
     info = client.whoami()
-    for k in info:
-        print(k.ljust(10) + ' : ' + str(info[k]))
+    if cli_config == 'rich':
+        keyword_styles = {**CLITheme.ACCOUNT_STATUS, **CLITheme.ACCOUNT_TYPE}
+        table_data = [(k, Text(str(v), style=keyword_styles.get(str(v), 'default'))) for (k, v) in sorted(info.items())]
+        table = generate_table(table_data, col_alignments=['left', 'left'], row_styles=['none'])
+        print_output(table, console=console, no_pager=args.no_pager)
+    else:
+        for k in info:
+            print(k.ljust(10) + ' : ' + str(info[k]))
     return SUCCESS
 
 
@@ -330,6 +360,10 @@ def list_dataset_replicas(args):
             result[dsn] = {}
         result[dsn][replica['rse']] = [replica['rse'], replica['available_length'], replica['length']]
 
+    if cli_config == 'rich':
+        spinner.update(status='Fetching dataset replicas')
+        spinner.start()
+
     if len(args.dids) == 1:
         scope, name = get_scope(args.dids[0], client)
         dmeta = client.get_metadata(scope, name)
@@ -342,22 +376,39 @@ def list_dataset_replicas(args):
 
     if args.deep or len(datasets) < 2:
         for did in datasets:
-            dsn = "%s:%s" % (did['scope'], did['name'])
+            dsn = f"{did['scope']}:{did['name']}"
             for rep in client.list_dataset_replicas(scope=did['scope'], name=did['name'], deep=args.deep):
                 _append_result(dsn=dsn, replica=rep)
     else:
         for rep in client.list_dataset_replicas_bulk(dids=datasets):
-            dsn = "%s:%s" % (rep['scope'], rep['name'])
+            dsn = f"{rep['scope']}:{rep['name']}"
             _append_result(dsn=dsn, replica=rep)
 
     if args.csv:
         for dsn in result:
             for rse in list(result[dsn].values()):
                 print(rse[0], rse[1], rse[2], sep=',')
+
+        if cli_config == 'rich':
+            spinner.stop()
     else:
-        for dsn in result:
-            print('\nDATASET: %s' % (dsn))
-            print(tabulate(list(result[dsn].values()), tablefmt=tablefmt, headers=['RSE', 'FOUND', 'TOTAL']))
+        output = []
+        for i, dsn in enumerate(result):
+            if cli_config == 'rich':
+                if i > 0:
+                    output.append(Text(f'\nDATASET: {dsn}', style=CLITheme.TEXT_HIGHLIGHT))
+                elif len(result) > 1:
+                    output.append(Text(f'DATASET: {dsn}', style=CLITheme.TEXT_HIGHLIGHT))
+
+                table = generate_table(list(result[dsn].values()), headers=['RSE', 'FOUND', 'TOTAL'], col_alignments=['left', 'right', 'right'])
+                output.append(table)
+            else:
+                print(f'\nDATASET: {dsn}')
+                print(tabulate(list(result[dsn].values()), tablefmt=tablefmt, headers=['RSE', 'FOUND', 'TOTAL']))
+
+        if cli_config == 'rich':
+            spinner.stop()
+            print_output(*output, console=console, no_pager=args.no_pager)
     return SUCCESS
 
 
@@ -376,7 +427,7 @@ def list_file_replicas(args):
     if args.protocols:
         protocols = args.protocols.split(',')
 
-    table = []
+    table_data = []
     dids = []
     if args.missing and not args.rses:
         print('Cannot use --missing without specifying a RSE')
@@ -385,9 +436,13 @@ def list_file_replicas(args):
         print('The substitution parameter must equal --link="/pfn/dir:/dst/dir"')
         return FAILURE
 
+    if cli_config == 'rich':
+        spinner.update(status='Fetching file replicas')
+        spinner.start()
+
     for did in args.dids:
         scope, name = get_scope(did, client)
-        client.get_metadata(scope=scope, name=name)  # break with Exception before streaming replicas if DID does not exist
+        client.get_metadata(scope=scope, name=name)  # Break with Exception before streaming replicas if DID does not exist.
         dids.append({'scope': scope, 'name': name})
 
     replicas = client.list_replicas(dids, schemes=protocols,
@@ -401,14 +456,24 @@ def list_file_replicas(args):
     rses = [rse["rse"] for rse in client.list_rses(rse_expression=args.rses)]
 
     if args.metalink:
-        print(replicas[:-1])  # last character is newline, no need to print that
+        print(replicas[:-1])  # Last character is newline, no need to print that.
         return SUCCESS
 
     if args.missing:
         for replica, rse in itertools.product(replicas, rses):
             if 'states' in replica and rse in replica['states'] and replica['states'].get(rse) != 'AVAILABLE':
-                table.append([replica['scope'], replica['name'], "({0}) {1}".format(ReplicaState[replica['states'].get(rse)].value, rse)])
-        print(tabulate(table, tablefmt=tablefmt, headers=['SCOPE', 'NAME', '(STATE) RSE']))
+                if cli_config == 'rich':
+                    replica_state = f"[{CLITheme.REPLICA_STATE.get(ReplicaState[replica['states'].get(rse)].value, 'default')}]{ReplicaState[replica['states'].get(rse)].value}[/]"
+                    table_data.append([replica['scope'], replica['name'], '({0}) {1}'.format(replica_state, rse)])
+                else:
+                    table_data.append([replica['scope'], replica['name'], "({0}) {1}".format(ReplicaState[replica['states'].get(rse)].value, rse)])
+        if cli_config == 'rich':
+            table = generate_table(table_data, headers=['SCOPE', 'NAME', '(STATE) RSE'], col_alignments=['left', 'left', 'left'])
+            spinner.stop()
+            print_output(table, console=console, no_pager=args.no_pager)
+        else:
+            print(tabulate(table_data, tablefmt=tablefmt, headers=['SCOPE', 'NAME', '(STATE) RSE']))
+
     elif args.link:
         pfn_dir, dst_dir = args.link.split(':')
         if args.rses:
@@ -428,13 +493,23 @@ def list_file_replicas(args):
                 for pfn in replica['pfns']:
                     rse = replica['pfns'][pfn]['rse']
                     if replica['rses'].get(rse):
-                        print(pfn)
+                        if cli_config == 'rich':
+                            table_data.append([pfn])
+                        else:
+                            print(pfn)
         else:
             for replica in replicas:
                 for pfn in replica['pfns']:
                     rse = replica['pfns'][pfn]['rse']
                     if replica['rses'][rse]:
-                        print(pfn)
+                        if cli_config == 'rich':
+                            table_data.append([pfn])
+                        else:
+                            print(pfn)
+        if cli_config == 'rich':
+            table = generate_table(table_data, headers=['PFN'], col_alignments=['left'])
+            spinner.stop()
+            print_output(table, console=console, no_pager=args.no_pager)
     else:
         if args.all_states:
             header = ['SCOPE', 'NAME', 'FILESIZE', 'ADLER32', '(STATE) RSE: REPLICA']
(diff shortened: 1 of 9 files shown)


dchristidis (changes requested):
Very promising! I have two major comments that should be addressed first:

 1. It’s not beneficial to have identical definitions in both `rucio` and `rucio-admin`. Please move to `lib/rucio/richclient.py` and import as necessary.
 2. All new functions should have type hints.

There will be more comments to follow.

lib/rucio/richclient.py
```diff
@@ -0,0 +1,283 @@
...
+    if not logger.handlers:
+        console = console or Console()
+        add_handler(logger, console, verbose)
+
+    return logger
+
+
+def generate_table(table_format: box.Box, headers: list[str], rows: Sequence[Union[tuple[Any, Any], list[Any]]],
+                   col_alignments: Optional[list[JustifyMethod]] = None, show_header: bool = True,
+                   row_styles: Optional[list[str]] = None) -> Table:
```
dchristidis: 1. Type hints add some noise. When the definition becomes to long and needs to be wrapped, typically it becomes preferable to have one parameter per line. Especially so when some of the hints are long and complex.

    ```python
    def generate_table(
        table_format: box.Box,
        headers: list[str],
        rows: Sequence[Union[tuple[Any, Any], list[Any]]],
        col_alignments: Optional[list[JustifyMethod]] = None,
        show_header: bool = True,
        row_styles: Optional[list[str]] = None
    ) -> Table:
    ```

 2. I think the `tabulate.tabulate()` signature makes more sense.

    1. `table_format` should become a keyword parameter with a default value of `CLITheme.TABLE_FMT`.
    2. `headers` should also become a keyword parameter with a default value of `None`. It should be combined together with `show_header`: if set, then the arguments contains the header names; otherwise, don’t show the header at all.

 3. Maybe rows should be `Sequence[Sequence[Any]]`? `headers` should definitely be `Sequence[str]`. Probably others as well. Let’s discuss in person.

lib/rucio/richclient.py
```diff
@@ -0,0 +1,283 @@
...
+    """
+    Generates a rich Table object from given input rows.
```
dchristidis: From [PEP 257](https://peps.python.org/pep-0257/#multi-line-docstrings):

> Multi-line docstrings consist of a summary line just like a one-line docstring, followed by a blank line, followed by a more elaborate description.

lib/rucio/richclient.py
```diff
@@ -0,0 +1,283 @@
...
+    :returns: rich Table object
+    """
+    table = Table(box=table_format, show_header=show_header)
+    rich_widgets = (Text, Tree, Table)
+    if row_styles is None:
+        row_styles = ['none', 'dim']
+    table.row_styles = row_styles
+
+    if not col_alignments and len(rows) > 0:
+        # auto-detect on first row, numerical values on the right
```
dchristidis: From [PEP 8](https://peps.python.org/pep-0008/#comments):

> Comments should be complete sentences.

Please capitalise and end with a period.

lib/rucio/richclient.py
```diff
@@ -0,0 +1,283 @@
...
+        col_alignments = ['right' if str(value).isnumeric() else 'left' for value in rows[0]]
```
dchristidis: You’re using `value` here and `elem` below. Probably better to stick to one, and likely `column` or `col` so that it ties well with the term ‘row’.

lib/rucio/richclient.py
```diff
@@ -0,0 +1,283 @@
...
+    elif col_alignments and len(rows) > len(col_alignments):
+        while len(rows) > len(col_alignments):
+            col_alignments.append('left')
```
dchristidis: I don’t understand this. How is the size of `rows` related to the size of `col_alignments`?

lib/rucio/richclient.py
```diff
@@ -0,0 +1,283 @@
...
+
+    for i, header in enumerate(headers):
+        if col_alignments:
+            justify = col_alignments[i] if col_alignments[i] in ["left", "center", "right"] else "left"
+            table.add_column(header, overflow='fold', justify=justify)
+        else:
+            table.add_column(header, overflow='fold')
```
dchristidis: 1. Both single and double quotes. Please pick one and apply it consistently.
 2. No need for the if-else structure. Calling `add_column()` has `justify='left'` as a default.

lib/rucio/richclient.py
```diff
@@ -0,0 +1,283 @@
...
+
+    for row in rows:
+        row = [elem if isinstance(elem, rich_widgets) and str(elem) != 'None' else str(elem) if elem is not None and str(elem) != 'None' else '' for elem in row]
```
dchristidis: This list comprehension with two conditional expressions appears complex and might require some simplification.

dchristidis (changes requested):
Some of my comments apply to your entire PR.

lib/rucio/richclient.py
```diff
@@ -0,0 +1,285 @@
...
+    logger.setLevel(logger_level)
+
+    def add_handler(logger: logging.Logger,
+                    console: Optional[Console] = None,
+                    verbose: bool = False) -> None:
+
+        def time_formatter(timestamp: datetime) -> Text:
+            formatted_time = timestamp.strftime('%Y-%m-%d %H:%M:%S')
+            milliseconds = timestamp.microsecond // 1000
+            return Text(f'[{formatted_time},{milliseconds:03d}]')
```
dchristidis: What advantage does this have against `timestamp.isoformat(sep=' ', timespec='milliseconds')`?

lib/rucio/richclient.py
```diff
@@ -0,0 +1,285 @@
...
+
+        console = console or Console()
+        highlighter = ReprHighlighter()
+        handler = RichHandler(rich_tracebacks=True, markup=True, show_path=verbose, show_time=verbose, console=console, tracebacks_width=min(console.width, MAX_TRACEBACK_WIDTH),
+                              tracebacks_word_wrap=True, log_time_format=time_formatter, highlighter=highlighter)
+        logger.addHandler(handler)
+
+    # Setting handler and formatter.
+    if not logger.handlers:
+        console = console or Console()
```
dchristidis: Aren’t ensuring that `console` is initialised inside `add_handler()`?

lib/rucio/richclient.py
```diff
@@ -0,0 +1,285 @@
...
+
+
+def get_cli_config() -> str:
+    """
+    Returns the CLI type from the config file.
+
+    :returns: CLI type
+    """
+    try:
+        cli_type = config_get('client', 'cli').lower()
```
dchristidis: Please use the `experimental` section.

bin/rucio
```diff
@@ -293,8 +316,23 @@ def whoami_account(args):
...
+        if console.is_terminal:
+            if args.no_pager:
+                console.print(table)
+            else:
+                console.width = sys.maxsize
+                console.begin_capture()
+                console.print(table)
+        else:
+            console.width = sys.maxsize
+            console.print(table)
```
dchristidis: Isn’t this sequence of statements repeated multiple times? It looks to me like it should become a helper function.

dchristidis (changes requested):
There have been some changes merged affecting `rucio-admin` since you created your branch. Please rebase on top of the latest master.

lib/rucio/richclient.py
dchristidis: I believe this still stands? `col_alignments` is, due to the previous block, filed with `['left', 'left', …]` unless set as an arguments. What prevents you from calling `table.add_column(header, overflow='fold', justify=justify)` unconditionally?

ericbanzuzi: My initial thought process was that if the user accidentally provides a list `col_alignments` that is shorter than the elements in a row, the method could still fill the "missing" alignments with the default value `'left'` instead of breaking. E.g. you could have `row=['filled', 'test', 'row']` and `col_alignments=['left', 'right']` leading to `col_alignments=['left', 'right', 'left']`.

But yes I realised that you are right that if `col_alignments` will always be true since I always set it (if not provided) or fill the missing values. I will fix this.

lib/rucio/richclient.py
```diff
@@ -0,0 +1,308 @@
...
+        row = ['' if str(col) == 'None' else col for col in row]
+        table.add_row(*row)
+    return table
+
+
+def print_output(
+    *output: Any,
+    console: Console,
+    no_pager: bool = False
+) -> None:
```
ericbanzuzi: The **Check Python Type Annotations** test currently fails because of the use of `Any` here. Not sure how to change it as `Console.print()` in `rich.console`  accepts `Any` as the type to the print statement too. 

Should I perhaps use `*output: object`  instead of `*output: Any`? That would lead to a passing test.

lib/rucio/richclient.py
```diff
@@ -0,0 +1,308 @@
...
+        if headers:
+            for header in headers:
+                table.add_column(header)
+        return table
+
+    # Auto-detect on first row, numerical values on the right.
+    col_alignments = col_alignments or ['right' if str(col).isnumeric() else 'left' for col in rows[0]]
+    headers = headers or [''] * len(rows[0])
+
+    if col_alignments and len(headers) > len(col_alignments):
```
dchristidis: The second statement is repeated below. The empty line should be removed. And doesn’t the first statement guarantee that `col_alignments` won’t be evaluating to false?

lib/rucio/richclient.py
```diff
@@ -0,0 +1,308 @@
...
+        while len(headers) > len(col_alignments):
+            col_alignments.append('left')
+
+    headers = headers or [''] * len(rows[0])
+    for i, header in enumerate(headers):
+        table.add_column(header, overflow='fold', justify=col_alignments[i])
```
dchristidis: I wonder if `zip()` would look nicer here.

bin/rucio
```diff
@@ -73,9 +81,14 @@ FAILURE = 1
...
 DEFAULT_PORT = 80
 
+# Setting up rich console.
+console = Console(theme=Theme(CLITheme.LOG_THEMES), soft_wrap=True)
+console.width = max(MIN_CONSOLE_WIDTH, console.width)
+
 logger = logging.log
 gfal2_logger = logging.getLogger("gfal2")
 tablefmt = 'psql'
+cli_config = 'tabulate'
```
dchristidis: What is the purpose of this assignment?

lib/rucio/richclient.py
```diff
@@ -0,0 +1,308 @@
...
+def setup_rich_logger(
+    module_name: Optional[str] = None,
+    logger_name: Optional[str] = None,
+    logger_level: Optional[int] = None,
+    verbose: bool = False,
+    console: Optional[Console] = None
+) -> logging.Logger:
+    """
+    Factory method to set logger with handlers.
+
```
dchristidis: Please mention that it’s a copy of the `setup_logger()` with minor changes.

lib/rucio/richclient.py
```diff
@@ -0,0 +1,308 @@
...
+    Returns the CLI type from the config file.
+
+    :returns: CLI type
+    """
+    try:
+        cli_type = config_get('experimental', 'cli').lower()
+        if cli_type not in ['rich', 'tabulate']:
+            cli_type = 'tabulate'
+    except (NoOptionError, NoSectionError, RucioException):
+        cli_type = 'tabulate'
```
dchristidis: I think it would be better to set the function’s `raise_exception=False`.

lib/rucio/richclient.py
```diff
@@ -0,0 +1,308 @@
...
+    """
+    Returns the pager function based on the terminal availability.
+
+    :returns: pager
+    """
+    default_pager = 'less'
+    try:
+        result = subprocess.run([default_pager], stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
+        less_available = result.returncode == 0
+    except (FileNotFoundError, Exception):
```
dchristidis: If you’re catching Exception, is there a benefit to mentioning `FileNotFoundError`?

dchristidis (changes requested):

bin/rucio
```diff
@@ -1868,6 +2537,7 @@ def get_parser():
     oparser.add_argument('--robot', '-R', dest="human", default=True, action='store_false', help="All output in bytes and without the units. This output format is preferred by parsers and scripts.")
     oparser.add_argument('--user-agent', '-U', dest="user_agent", default='rucio-clients', action='store', help="Rucio User Agent")
     oparser.add_argument('--vo', dest="vo", metavar="VO", default=None, help="VO to authenticate at. Only used in multi-VO mode.")
+    oparser.add_argument("--no-pager", dest="no_pager", default=False, action='store_true', help="Do not pipe Rucio output into a pager.")
```
dchristidis: Use `help=argparse.SUPPRESS` for now.

lib/rucio/richclient.py
dchristidis: No, I think you’re doing the right thing. The CI treats `Any` as missing a type, but there could be functions such as this where literally any type is permissible and handled properly.

dchristidis (dismissed):
This PR is too large to to check exhaustively, but I believe its approach in general is well-engineered. Any defects can more realistically be identified through actual use.

The failing CI test should be ignored (refer to [this](https://github.com/rucio/rucio/pull/7012#discussion_r1727686501) discussion for more information).

Given its magnitude, it would be ideal if the component leads also reviewed this.

lib/rucio/richclient.py
```diff
@@ -0,0 +1,302 @@
+# Copyright European Organization for Nuclear Research (CERN) since 2012
```
voetberg: This file should probably live in the same folders with the other client files

bin/rucio
```diff
@@ -1868,6 +2537,7 @@ def get_parser():
     oparser.add_argument('--robot', '-R', dest="human", default=True, action='store_false', help="All output in bytes and without the units. This output format is preferred by parsers and scripts.")
     oparser.add_argument('--user-agent', '-U', dest="user_agent", default='rucio-clients', action='store', help="Rucio User Agent")
     oparser.add_argument('--vo', dest="vo", metavar="VO", default=None, help="VO to authenticate at. Only used in multi-VO mode.")
+    oparser.add_argument("--no-pager", dest="no_pager", default=False, action='store_true', help=argparse.SUPPRESS)
```
voetberg, line 2326: Why does no_pager need to be included in the args? If it's being supressed, do we want people to be able to change this?

bin/rucio
```diff
@@ -1801,19 +2464,25 @@ def add_lifetime_exception(args):
...
         client.add_exception(dids=datasets, account=client.account, pattern='', comments=reason, expires_at=expiration)
     except UnsupportedOperation as err:
         logger.error(err)
         return FAILURE
     except Exception:
-        logger.error('Failure to submit exception. Please retry')
-        logger.debug(traceback.format_exc())
+        if cli_config == 'rich':
+            if logger.level == logging.DEBUG:
+                logger.exception('Failure to submit exception. Please retry.')
```
voetberg: Because this message is the same across all errors, it would make more sense to define the error message as a str var before the logic for picking the logger.

bin/rucio
```diff
@@ -1702,19 +2343,41 @@ def list_datasets_rse(args):
...
-            table.append(['%s:%s' % (dsn['scope'], dsn['name']), '%s/%s' % (str(dsn['available_length']), str(dsn['length'])), '%s/%s' % (str(dsn['available_bytes']), str(dsn['bytes']))])
-        table.sort()
-        print(tabulate(table, tablefmt=tablefmt, headers=['DID', 'LOCAL FILES/TOTAL FILES', 'LOCAL BYTES/TOTAL BYTES']))
+    if cli_config == 'rich':
+        status = Status('Fetching datasets at RSE', spinner=CLITheme.SPINNER, spinner_style=CLITheme.SPINNER_STYLE, console=console)
+        with status:
+            if args.long:
+                table_data = []
+                for dsn in client.list_datasets_per_rse(args.rse):
+                    table_data.append(['%s:%s' % (dsn['scope'], dsn['name']),
```
voetberg: (This comment applies for all tables using %s formatting) Is there a reason for this not to be f-strings? Those are preferred for readability purposes.

bin/rucio
```diff
@@ -1670,26 +2282,55 @@ def list_account_usage(args):
...
-        remaining = 0 if float(item['bytes_remaining']) < 0 else float(item['bytes_remaining'])
-        table.append([item['rse'], sizefmt(item['bytes'], args.human), sizefmt(item['bytes_limit'], args.human), sizefmt(remaining, args.human)])
-    table.sort()
-    print(tabulate(table, tablefmt=tablefmt, headers=['RSE', 'USAGE', 'LIMIT', 'QUOTA LEFT']))
-
-    table = []
-    usage = client.get_global_account_usage(account=args.usage_account)
-    for item in usage:
-        if (args.rse and args.rse in item['rse_expression']) or not args.rse:
+        if args.rse:
```
voetberg: args.rse should default to None, so this logic is redundant, you can just pass args.rse whenever (ref: https://github.com/rucio/rucio/blob/e6097bd2ee4efb1ade8b4e262a28efdda6fb96a2/lib/rucio/client/accountclient.py#L330 )

bin/rucio
dchristidis: The `--no-pager` is inspired by the respective command-line option in Git. The pager is used by default. This command-line option allows one to write to the terminal instead.

I explicitly asked for it to be suppressed from the help output because it’s ineffective when the output with Rich is not enabled (which is introduced as an experimental opt-in feature in this PR). It will be switched to visible with an appropriate help message when the Rich output becomes the default.

lib/rucio/richclient.py
```diff
@@ -0,0 +1,302 @@
...
+        add_handler(logger, console, verbose)
+
+    return logger
+
+
+def generate_table(
+    rows: Sequence[Sequence[RenderableType]],
+    headers: Optional[Sequence[RenderableType]] = None,
+    row_styles: Optional[Sequence[StyleType]] = None,
+    col_alignments: Optional[list[JustifyMethod]] = None,
```
voetberg: Is there a rhyme or reason to the left/right alignments? It may be cleaner to automatically assign that based on the first row's numeric/not numeric status for each column

lib/rucio/richclient.py
```diff
@@ -0,0 +1,302 @@
...
+        else:
+            console.width = sys.maxsize   # Overwrite auto-detected console width.
+            console.begin_capture()
+            console.print(*output)
+    else:
+        console.width = sys.maxsize
+        console.print(*output)
+
+
+def get_cli_config() -> str:
```
voetberg: Is this the only way to set if using rich or tabulate? Having an override in the args would be useful for people who want to try a command or two without changing their entire setup

dchristidis: This was consciously done, as part of the release engineering. The first phase of testing will come from power users. We expect them to enable Rich in its entirety and, in a way, be forced to use it, not pick the output in each invocation.

voetberg:
I largely agree with Dimintros' comments. Nothing looks objectively wrong, just a few style/nit-picks I'll work through. I'd love to see some examples and maybe a follow up PR in documentation explaining how to enable this and why you'd want to.

bin/rucio
```diff
@@ -73,9 +81,14 @@ FAILURE = 1
 DEFAULT_SECURE_PORT = 443
 DEFAULT_PORT = 80
 
+# Setting up rich console.
+console = Console(theme=Theme(CLITheme.LOG_THEMES), soft_wrap=True)
```
voetberg, line 86: This setup, and the other setup in bin/rucio-admin, should probably be contained to a function in richclient and called within each function that uses the rich client to keep things more self-contained.

dchristidis:
> I'd love to see some examples and maybe a follow up PR in documentation explaining how to enable this and why you'd want to.

Please consult [Eric’s presentation](https://indico.cern.ch/event/1431526/#6-improving-the-rucio-cli-expe). It has many wonderful examples and the second-to-last slide describes how, if all goes well, this it to replace the existing implementation based on tabulate. Consequently, no additional documentation is necessary. If we time it with the new command-line structure that you’ve been working on, then we can update the existing documentation only once.

lib/rucio/richclient.py
voetberg: Understandable, I just think being able to try it out without the commitment will increase adoption. That can come at a later date though if this is part of the release planning.

dchristidis: Carrot vs stick. :laughing:

bin/rucio
```diff
@@ -327,36 +356,81 @@ def list_dataset_replicas(args):
...
         if dsn not in result:
             result[dsn] = {}
-        result[dsn][replica['rse']] = [replica['rse'], replica['available_length'], replica['length']]
-
-    if len(args.dids) == 1:
-        scope, name = get_scope(args.dids[0], client)
-        dmeta = client.get_metadata(scope, name)
-        _fetch_datasets_for_meta(meta=dmeta)
+        if cli_config == 'rich':
+            result[dsn][replica['rse']] = [replica['rse'], str(replica['available_length']), str(replica['length'])]
```
voetberg: Why can't this conversion be done when you're making the table? Seems like an unnecessary check

bin/rucio
```diff
@@ -327,36 +356,81 @@ def list_dataset_replicas(args):
...
+        else:
+            result[dsn][replica['rse']] = [replica['rse'], replica['available_length'], replica['length']]
+
+    if cli_config == 'rich':
+        status = Status('Fetching dataset replicas', spinner=CLITheme.SPINNER, spinner_style=CLITheme.SPINNER_STYLE, console=console)
```
voetberg: Is there a way to define a placeholder for status when rich isn't enabled? There's a ton of repeated code between the if/else and that's rife for inconsistency problems.

bin/rucio
```diff
@@ -384,75 +458,169 @@ def list_file_replicas(args):
...
-                                    sort=args.sort, domain=args.domain,
-                                    resolve_archives=not args.no_resolve_archives)
-    rses = [rse["rse"] for rse in client.list_rses(rse_expression=args.rses)]
-
-    if args.metalink:
-        print(replicas[:-1])  # last character is newline, no need to print that
-        return SUCCESS
+    if cli_config == 'rich':
+        status = Status('Fetching file replicas', spinner=CLITheme.SPINNER, spinner_style=CLITheme.SPINNER_STYLE, console=console)
+        with status:
```
voetberg: [See this comment](https://github.com/rucio/rucio/pull/7012/files#r1763387604) - I won't replicate it for each part of the code where it applies but keep it in mind please.

voetberg:
> > I'd love to see some examples and maybe a follow up PR in documentation explaining how to enable this and why you'd want to.
> 
> Please consult [Eric’s presentation](https://indico.cern.ch/event/1431526/#6-improving-the-rucio-cli-expe). It has many wonderful examples and the second-to-last slide describes how, if all goes well, this it to replace the existing implementation based on tabulate. Consequently, no additional documentation is necessary. If we time it with the new command-line structure that you’ve been working on, then we can update the existing documentation only once.

Wow @ericbanzuzi This is a fantastic presentation!! I'm now thoroughly excited

Now that I see the release plan, I have no criticism. In theory I'd like if we had a better way to communicate experimental features like this to users outside these meetings but I think that's above my pay grade

bin/rucio
ericbanzuzi: No particular reason,  I reused the approach from the original code base.

lib/rucio/richclient.py
ericbanzuzi: I included a line, **231**,  that does automatic assignment for the simple cases 
```python
col_alignments = col_alignments or ['right' if str(col).isnumeric() else 'left' for col in rows[0]]
```

However, this approach does not currently work properly in all cases. For example:
- File sizes in human-readable formats (e.g., `100 MB`) are not handled correctly.
- If the first row in a column contains `None`, even though other rows are numerical, it will default to left alignment.
- For tables where columns have mixed value types (e.g., the two column key-value pair tables with different types), it's better to manually assign alignments.

Hence, the use of manual left/right alignments for now.

bin/rucio
ericbanzuzi: I suppose it could be done in the `generate_table` as well. I just wanted to limit the accepted elements in a row to `rich.console.RenderableType` (similarly to `rich.table.Table.add_rows()`), which does not include integer/float values.

bin/rucio
ericbanzuzi: Not sure about a placeholder but maybe switching to `status.start()` instead of the context manager might help with that. Then the if/else conditions would need to be in the core loops collecting the data since some of the logic between storing the rows for tabulate and rich are slightly different.

voetberg: I think that's a better solution!

lib/rucio/richclient.py
dchristidis: @ericbanzuzi do you think you can find some time to address this one? I’m afraid that it is a blocker that I should have identified myself. The rest of @voetberg’s comments remain valid, but I feel like we can afford to work on them in the future.

ericbanzuzi: @dchristidis yes, I think I can do it over the weekend or the weekend next week. I can also include some of the smallest and simplest fixes in these comments: [first](https://github.com/rucio/rucio/pull/7012#discussion_r1763294876), [second](https://github.com/rucio/rucio/pull/7012#discussion_r1763304438) and [third](https://github.com/rucio/rucio/pull/7012#discussion_r1763382976).

ericbanzuzi:
I had some extra time, so on top what [I previously mentioned](https://github.com/rucio/rucio/pull/7012#discussion_r1809595953), I also tried to address [this](https://github.com/rucio/rucio/pull/7012#discussion_r1763387604) by reducing the amount of duplicate code using [this approach](https://github.com/rucio/rucio/pull/7012#discussion_r1763757708), and changed some of the %s formatting to f-strings ([see](https://github.com/rucio/rucio/pull/7012#discussion_r1763297884)).

voetberg (approved):
All the changes look good to me! Thank you for addressing my comments. 

After convo with @bari12 and @dchristidis, we decided this will probably have to be merged after #6923, but I can do that rebase myself, so as it exists now, this seems good!

bari12:
@voetberg I think we should do the other way around (Maybe this got lost in our chat). Merge Erics first and rebase your PR to it.

Issue #6973, opened 2024-07-25
Permissions: Unify the permissions for `add_did` and `add_dids`


rdimaio:
### Description

Right now, the permissions for `add_did` and `add_dids` are separate:

https://github.com/rucio/rucio/blob/d7c7645b10dd6b904108bd94bd48f6ce14a20d90/lib/rucio/core/permission/generic.py#L382-L419

Because of this, mismatches can occur, like in the current situation where `rucio.core.scope.is_scope_owner(scope=kwargs['scope'], account=issuer, session=session)` is checked for `perm_add_did` but not for `perm_add_dids`.

In the first place - I'd be in favour of removing these duplicated methods altogether (`add_dids` and `add_did`), but I can understand that we'd break backwards compatibility with older clients. If we want to keep them both, it would be a good idea to combine `perm_add_did` and `perm_add_dids`

### Steps to reproduce

N/A

rdimaio:
Discussed with Martin, reopening it to implement the changes for the generic permission

PR #7014, opened 2024-08-09
Permissions: Unify the permissions for `add_did` and `add_dids` #6973

ericbanzuzi:
Combine `perm_add_did` and `perm_add_dids` into one method.

### DIFF
--- a/lib/rucio/core/permission/atlas.py
+++ b/lib/rucio/core/permission/atlas.py
@@ -435,13 +435,7 @@ def perm_add_dids(issuer, kwargs, *, session: "Optional[Session]" = None):
     :returns: True if account is allowed, otherwise False
     """
     # Check the accounts of the issued rules
-    if not _is_root(issuer) and not has_account_attribute(account=issuer, key='admin', session=session):
-        for did in kwargs['dids']:
-            for rule in did.get('rules', []):
-                if rule['account'] != issuer:
-                    return False
-
-    return _is_root(issuer) or has_account_attribute(account=issuer, key='admin', session=session)
+    return all(perm_add_did(issuer, kwargs=did, session=session) for did in kwargs['dids'])
 
 
 def perm_attach_dids(issuer, kwargs, *, session: "Optional[Session]" = None):
--- a/lib/rucio/core/permission/belleii.py
+++ b/lib/rucio/core/permission/belleii.py
@@ -437,10 +437,7 @@ def perm_add_dids(issuer: "InternalAccount", kwargs: dict, *, session: "Optional
     :returns: True if account is allowed, otherwise False
     """
     # Check the accounts of the issued rules
-    for did in kwargs['dids']:
-        if not perm_add_did(issuer, kwargs=did, session=session):
-            return False
-    return True
+    return all(perm_add_did(issuer, kwargs=did, session=session) for did in kwargs['dids'])
 
 
 def perm_attach_dids(issuer: "InternalAccount", kwargs: dict, *, session: "Optional[Session]" = None) -> bool:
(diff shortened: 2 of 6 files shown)


lib/rucio/core/permission/atlas.py
```diff
@@ -76,7 +76,6 @@ def has_permission(issuer, action, kwargs, *, session: "Optional[Session]" = Non
             'get_auth_token_x509': perm_get_auth_token_x509,
             'get_auth_token_saml': perm_get_auth_token_saml,
             'add_account_identity': perm_add_account_identity,
-            'add_did': perm_add_did,
             'add_dids': perm_add_dids,
```
rdimaio: I think we should keep this line, but simply assign `perm_add_dids` instead of `perm_add_did`.

This is simlar to how it's done for `remove_did_from_followed` and `remove_dids_from_followed` later on this dict

lib/rucio/core/permission/belleii.py
```diff
@@ -75,7 +75,6 @@ def has_permission(issuer: "InternalAccount", action: str, kwargs: dict, *, sess
             'get_auth_token_x509': perm_get_auth_token_x509,
             'get_auth_token_saml': perm_get_auth_token_saml,
             'add_account_identity': perm_add_account_identity,
-            'add_did': perm_add_did,
```
rdimaio: same here

lib/rucio/core/permission/belleii.py
```diff
@@ -438,7 +407,26 @@ def perm_add_dids(issuer: "InternalAccount", kwargs: dict, *, session: "Optional
...
+
+        scope = did['scope']
+        if isinstance(did['scope'], str):
+            scope = InternalScope(did['scope'])
+
+        if not perm_default(issuer, did, session=session)\
+                and not has_account_attribute(account=issuer, key='did_admin', session=session)\
+                and not has_account_attribute(account=issuer, key='production_account', session=session)\
+                and not rucio.core.scope.is_scope_owner(scope=scope, account=issuer, session=session)\
+                and not (did.get('name', False) and did['name'].startswith('/belle/scout')):
```
rdimaio: I actually think that the approach that was there seems pretty good - maybe we should change the focus of the issue/PR into using this approach in the other permission files as well, so basically:

1. `perm_add_did` should have the logic to check permission for a single did
2. `perm_add_dids` should simply do `return all([perm_add_did(issuer, kwargs=did, session=session) for did in kwargs['dids']])`

and then we shouldn't have to change anything else other than this

rdimaio (changes requested):

lib/rucio/core/permission/atlas.py
rdimaio: note - you can ignore this if we decide to go with this approach: https://github.com/rucio/rucio/pull/7014#discussion_r1711532967

lib/rucio/core/permission/belleii.py
ericbanzuzi: Fixed this in the latest force-push

rdimaio (approved):

bari12:
Closing this, with the policy packages this isn't do-able like this anymore.

rdimaio:
> Closing this, with the policy packages this isn't do-able like this anymore.

I think it's still valid for the generic permission

rdimaio:
Discussed with Martin, adding the generic changes to a new PR

rdimaio:
Follow-up PR: https://github.com/rucio/rucio/pull/7178

rdimaio:
Thanks for the changes Eric!

Issue #6989, opened 2024-07-31
`session` parameter in `*_session` decorated functions is required, but effectively treated as optional


rdimaio:
### Description

This is closely related to https://github.com/rucio/rucio/issues/6986, but they are two separate issues (#6986 is about annotating the decorators effectively in order to unhide the signature of the decorated functions, this issue is about refactoring `session` in decorated functions)

The issue at the moment is that `session` is defined as a required parameter in the decorated function signatures, not as an optional parameter. For example:

```
@read_session
def get_rse_account_usage(rse_id, *, session: "Session"):
```

Here `session` is a required parameter. It only works because it's decorated, and the decorator always runs the function with the `session` parameter, but from a syntax perspective it's not correct.

Beforehand, the decorators were not annoted using `ParamSpec`, so the signature of the decorated function was `hidden` by linters and type checkers, as seen in the issue spotted by Dimitrios here: [rucio/rucio#6111](https://github.com/rucio/rucio/pull/6111#discussion_r1696431460)

After annotating the decorators properly, the linters can now properly access the signature of the decorated functions, and they can detect issues like the one above, but also that `session` is indeed a required parameter: `Argument missing for parameter "session"`

From the message of this related commit  https://github.com/rucio/rucio/commit/cfe3aafdee667120a49d1b39212e2c3a44aae60f:
> There is currently no way to fully and correctly inform pyright about
this signature rewrite. That's why the type of decorators is not
entirely defined. I'm not sure how well the type deduction in pyright
will do its job, but I consider that it's more important to get rid
of the 500 + false positive errors in pyright related to session

Note the this is not something directly related to type annotation, this is simply a characteristic of the default behaviour of decorators in Python. The same linter errors occur when `@functools.wraps` is used to copy over the function name and argument list to the decorator (https://stackoverflow.com/a/309000)

### Solution 1
All of these decorators get the session via `session_scoped = get_session()` if it doesn't exist. Instead of doing this, we could modify the function signatures of the functions decorated with these decorators to have `session=get_session()`, in order to set the default session value in the function signature itself.

From what I can tell the functionality would be the same - the only things that would need to be changed is that the `if isgeneratorfunction(function):` check should be done within the `try` block in `read_session` and `stream_session`, to ensure that we always `remove()` the session.

## Solution 2
An alternative could be to remove `session` from the function signature of all the decorated functions and simply treat it as a keyword argument - we then check the list of args in the decorators, and if session is not passed we create it as usual.
[...]

Issue #6986, opened 2024-07-30
Fix static code analysis behaviour for functions decorated with `@transactional_session`, `@read_session` and `@stream_session`


rdimaio:
### Description

As spotted by Dimitrios here: https://github.com/rucio/rucio/pull/6111#discussion_r1696431460

The type for `ignore_availability` in `add_rule()` is `bool`, not `Optional[bool]`, so the type checker should have raised an error. https://github.com/rucio/rucio/blob/adfcf6913a0cae5357fb6b44671cbbc542eda5b5/lib/rucio/core/rule.py#L179

Removing the `@transactional_session` decorator from `add_rule` results in the type checker working properly. I had a look around and it seems like this issue affects the functions decorated with `@transactional_session` and `@read_session`

### Steps to reproduce

N/A

rdimaio:
This problem is described here: https://mypy.readthedocs.io/en/stable/generics.html#declaring-decorators

rdimaio:
regarding the `@transactional_session` and `@read_session` decorators (https://github.com/rucio/rucio/blob/dc0a90ce37b91ff044285603d29fa5bf8b604371/lib/rucio/db/sqla/session.py#L462)

It seems like all the functions decorated with these decorators have a `*` in the function signature before `session` so that `session` must be a keyword argument (e.g.: https://github.com/rucio/rucio/blob/dc0a90ce37b91ff044285603d29fa5bf8b604371/lib/rucio/core/distance.py#L30)

This didn't cause explicit issues because the decorators did not use `ParamSpec`, so they were 'hiding' the type annotations of the decorated functions (relevant issue: https://github.com/rucio/rucio/issues/6986, mypy explanation: https://mypy.readthedocs.io/en/stable/generics.html#declaring-decorators)

After properly annotating the decorators using `ParamSpec`, all calls without a `session` parameter to the decorated functions now result in: `Argument missing for parameter "session"`, which is likely due to marking it as a forced keyword argument via the `*`.

rdimaio:
Found this related issue: https://github.com/rucio/rucio/issues/5947

rdimaio:
This also affects the decorators `count_it` and `time_it` in `core/monitor`, see: https://github.com/rucio/rucio/pull/7017/commits/e28b5430c06090f22f6cc29095a14621e4129d9a

```
Found 5 new problems in /home/runner/work/rucio/rucio/lib/rucio/core/request.py
  - 2 errors with message """No parameter named "session"""".
[...]

PR #7017, opened 2024-08-12
Testing: Add type annotations to core/monitor

rdimaio:
Part of #6588


rdimaio:
Type checking here fails due to these two issues:
- https://github.com/rucio/rucio/issues/6989
- https://github.com/rucio/rucio/issues/6986

I will remove the type annnotations from the decorators to 'hide' this problem for now, and update those issues to track `core/monitor` too.

bari12 (approved):

PR #7036, opened 2024-08-20
Client: Allow account to be None in BaseClient constructor

### LINKED ISSUE #7028: Account attribute in BaseClient not initialized when authenticating with x509_proxy
### Description

With the newest release, it is not possible to instantiate a `Client` object without specifying the account argument to the constructor, when authenticating with x509 proxy.
With the following `rucio.cfg` file and the `os.environ["RUCIO_HOME"]` parameter properly set:
```
[common]
[client]
rucio_host = http://cms-rucio.cern.ch
auth_host = https://cms-rucio-auth.cern.ch
auth_type = x509_proxy
ca_cert = /cvmfs/cms.cern.ch/grid/etc/grid-security/certificates/ 
request_retries = 3
```
Instantiating a `Client` object in a python shell throws the following error:
```
>>> Client()
  File "/usr/local/lib/python3.9/site-packages/rucio/client/client.py", line 76, in __init__
    super(Client, self).__init__(**args)
  File "/usr/local/lib/python3.9/site-packages/rucio/client/baseclient.py", line 200, in __init__
    self.auth_token_file_path, self.token_exp_epoch_file, self.token_file, self.token_path = self._get_auth_tokens()
  File "/usr/local/lib/python3.9/site-packages/rucio/client/baseclient.py", line 213, in _get_auth_tokens
    token_filename_suffix = "for_default_account" if self.account is None else "for_account_" + self.account
AttributeError: 'Client' object has no attribute 'account'
```
whereas the authentication works if the `account` argument is passed:
```
>>> Client(account=username)
<rucio.client.client.Client object at 0x7fcd519b1910>
```

The error traceback is suggesting that the initialization of the `self.account` attribute at lines L174-L184 of `baseclient.py` does not assign a value to `self.account` in this case:
https://github.com/rucio/rucio/blob/c0b58178c238556e553df0d7d200de5437bafaed/lib/rucio/client/baseclient.py#L174-L184

since the attribute `account` does not exist when `_get_auth_tokens()` is called at line L200:
https://github.com/rucio/rucio/blob/c0b58178c238556e553df0d7d200de5437bafaed/lib/rucio/client/baseclient.py#L200

The expected behavior is that one should be able to authenticate with x509 proxy without passing the account parameter, e.g. by initializing `self.account` to `None` in the constructor of `BaseClient`.
The authentication with x509 proxy with the same `rucio.cfg` file was working in a previous version.

### Steps to reproduce

1. Login to lxplus.cern.ch el9 machine
2. Create VOMS proxy certificate: `voms-proxy-init -voms cms -rfc --valid 168:0`
3. Open a python shell
4. Type the following commands:
   - import os
   - os.environ["RUCIO_HOME"] = "/cvmfs/cms.cern.ch/rucio/current"
   - from rucio.client import Client
   - import getpass
5. Try to instantiate a Client object: `Client()`
6. Try to instantiate a Client object by passing the username: `Client(account=getpass.getuser())`

Step 4 should throw the error while Step 5 should not, since the `account` argument is passed.

### Rucio Version

Rucio: 35.1.1

### Additional Information

Possible offending commit: https://github.com/rucio/rucio/commit/b3e3b31f7e3265491cfbe330aab561e1cda52198

rdimaio:
Partially reverts https://github.com/rucio/rucio/commit/b3e3b31f7e3265491cfbe330aab561e1cda52198 to fix #7028

The typing errors (https://github.com/rdimaio/rucio/actions/runs/10468813387/job/28990505604) are due to some methods expecting `account` to exist. In practice, these methods are called after `account` has been set, but since that's done at runtime, the static type checker does not see them. We can address these at a later point.

### DIFF
--- a/lib/rucio/client/baseclient.py
+++ b/lib/rucio/client/baseclient.py
@@ -137,6 +137,7 @@ class BaseClient:
             self.logger.debug('No trace_host passed. Using rucio_host instead')
 
         self.list_hosts = [self.host]
+        self.account = account
         self.ca_cert = ca_cert
         self.auth_token = ""
         self.headers = {}
@@ -171,9 +172,7 @@ class BaseClient:
                     self.logger.debug('No ca_cert found in configuration. Falling back to Mozilla default CA bundle (certifi).')
                     self.ca_cert = True
 
-        if account is not None:
-            self.account = account
-        else:
+        if account is None:
             self.logger.debug('No account passed. Trying to get it from the RUCIO_ACCOUNT environment variable or the config file.')
             try:
                 self.account = environ['RUCIO_ACCOUNT']


voetberg (approved):
👍

PR #7040, opened 2024-08-20
enhancement: BaseClient set ca_cert and vo defaults Fix #6934

### LINKED ISSUE #6934: Configuration handling for `ca_cert` and `vo` not falling back properly to defaults

Geogouz:
- Allow fallbacks to default `ca_cert` and `vo` as mentioned in #6934.

### DIFF
--- a/lib/rucio/client/baseclient.py
+++ b/lib/rucio/client/baseclient.py
@@ -171,6 +171,9 @@ class BaseClient:
                 except (NoOptionError, NoSectionError):
                     self.logger.debug('No ca_cert found in configuration. Falling back to Mozilla default CA bundle (certifi).')
                     self.ca_cert = True
+                except ConfigNotFound:
+                    self.logger.debug('No configuration found. Falling back to Mozilla default CA bundle (certifi).')
+                    self.ca_cert = True
 
         if account is None:
             self.logger.debug('No account passed. Trying to get it from the RUCIO_ACCOUNT environment variable or the config file.')
@@ -194,6 +197,9 @@ class BaseClient:
                     self.vo = config_get('client', 'vo')
                 except (NoOptionError, NoSectionError):
                     self.logger.debug('No VO found. Using default VO.')
+                    self.vo = 'def'
+                except ConfigNotFound:
+                    self.logger.debug('No configuration found. Using default VO.')
                     self.vo = 'def'
 
         self.auth_token_file_path, self.token_exp_epoch_file, self.token_file, self.token_path = self._get_auth_tokens()


lib/rucio/client/baseclient.py
```diff
@@ -167,7 +167,7 @@ def __init__(self,
                 self.logger.debug('HTTPS is required, but no ca_cert was passed and X509_CERT_DIR is not defined. Trying to get it from the config file.')
                 try:
                     self.ca_cert = path.expandvars(config_get('client', 'ca_cert'))
-                except (NoOptionError, NoSectionError):
+                except (NoOptionError, NoSectionError, ConfigNotFound):
                     self.logger.debug('No ca_cert found in configuration. Falling back to Mozilla default CA bundle (certifi).')
```
rdimaio: I think the logger message should be different in the event of a `ConfigNotFound`. Right now it'd log:
> 'No ca_cert found in configuration. ...'

But in the case of `ConfigNotFound`, it'd be more accurate to say "No configuration found."

lib/rucio/client/baseclient.py
```diff
@@ -193,7 +193,7 @@ def __init__(self,
                 self.logger.debug('No VO found. Trying to get it from the config file.')
                 try:
                     self.vo = config_get('client', 'vo')
-                except (NoOptionError, NoSectionError):
+                except (NoOptionError, NoSectionError, ConfigNotFound):
                     self.logger.debug('No VO found. Using default VO.')
                     self.vo = 'def'
```
rdimaio: Same here, should mention that the config wasn't found when `ConfigNotFound`

voetberg (approved):
👍

rdimaio (approved):

bari12:
Merging. But the commit message should be `<component>: <short_change_message> #<issue number>` (Enhancement is not one of our components) :-)

PR #7042, opened 2024-08-21
[DRAFT] Testing: typing.overload read_from_cache and write_to_cache in core/config #7018

### LINKED ISSUE #7018: Typing: `typing.overload` `read_from_cache` and `write_to_cache`
e.g.:

https://github.com/rucio/rucio/blob/b76b124f387afd994c556a996aaaa4e030bc914b/lib/rucio/core/config.py#L438

if `key` is `Literal['sections']`, then `value` should be `list[str]`: 

https://github.com/rucio/rucio/blob/b76b124f387afd994c556a996aaaa4e030bc914b/lib/rucio/core/config.py#L82

We could use `typing.overload` to overload these two cache-related function so that, for a given key, only the corresponding possible types are used for `value`, similarly to how it's done for  `get_rse_attribute` here: https://github.com/rucio/rucio/pull/6952/files

ericbanzuzi:
Resolves #7018

### DIFF
--- a/lib/rucio/core/config.py
+++ b/lib/rucio/core/config.py
@@ -12,11 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from collections.abc import Callable
-from typing import TYPE_CHECKING, Any, Optional, TypeVar
+from collections.abc import Callable, Sequence
+from typing import TYPE_CHECKING, Any, Literal, NewType, Optional, TypeVar, Union, overload
 
 from dogpile.cache.api import NoValue
 from sqlalchemy import and_, delete, func, select, update
+from sqlalchemy.engine import Row
 
 from rucio.common.cache import make_region_memcached
 from rucio.common.exception import ConfigNotFound
@@ -24,6 +25,11 @@ from rucio.db.sqla import models
 from rucio.db.sqla.session import read_session, transactional_session
 
 T = TypeVar('T')
+
+HasKey = NewType('HasKey', str)
+ItemsKey = NewType('ItemsKey', str)
+ValueKey = NewType('ValueKey', str)
+OptionsKey = NewType('OptionsKey', str)
 
 if TYPE_CHECKING:
     from sqlalchemy.orm import Session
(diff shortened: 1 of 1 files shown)


lib/rucio/core/config.py
```diff
@@ -34,24 +40,24 @@
...
-def _items_cache_key(section: str) -> str:
-    return 'items_%s' % section
+def _items_cache_key(section: str) -> ItemsKey:
+    return ItemsKey('items_%s' % section)
 
 
-def _value_cache_key(section: str, option: str) -> str:
-    return 'get_%s_%s' % (section, option)
+def _value_cache_key(section: str, option: str) -> ValueKey:
+    return ValueKey('get_%s_%s' % (section, option))
```
rdimaio, line 60: This is being refactored in this previous open PR: https://github.com/rucio/rucio/pull/7015/files#diff-3c88400ee4ff0b9fca1b3a4f6dd3f58e23b10991a3a4032cbdbcd7028ae5c264

lib/rucio/core/config.py
```diff
@@ -423,7 +429,32 @@ def remove_option(section: str, option: str, *, session: "Session") -> bool:
...
+    ...
+
+
+@overload
+def read_from_cache(key: ItemsKey, expiration_time: int = 900) -> Union[Sequence[Row], NoValue]:
+    ...
+
+
+@overload
+def read_from_cache(key: ValueKey, expiration_time: int = 900) -> Union[Any, NoValue]:
```
rdimaio, line 453:
```
Union[Any, NoValue]
```

Placing `Any` inside an `Union` is redundant: `Any` is compatible with all types https://docs.python.org/3/library/typing.html#typing.Any

rdimaio, line 453: Because all of these return types are `Union[..., NoValue]`, you could create a `TypeVar` to parametrize this type, see https://stackoverflow.com/a/46384213

e.g. something like

```python
T = TypeVar('T')

CachedValue = Union[T, NoValue]

...

def read_from_cache(key: HasKey, expiration_time: int = 900) -> CachedValue[bool]:

...
```

rdimaio: Something also worth mentioning is that the docstring for `NoValue` says:
```
    """Describe a missing cache value.

    The :data:`.NO_VALUE` constant should be used.

    """
```

So it might be better to use something like `Literal[NO_VALUE]`? Not completely sure about this, though. We use `NoValue` quite a bit at the moment, so you can probably ignore this for now and at a later point we could modify all `NoValue` annotations to `Literal[NO_VALUE]` if needed

lib/rucio/core/config.py
```diff
@@ -423,7 +429,32 @@ def remove_option(section: str, option: str, *, session: "Session") -> bool:
...
 
-def read_from_cache(key: str, expiration_time: int = 900) -> Any:
+@overload
+def read_from_cache(key: Literal['sections'], expiration_time: int = 900) -> Union[list[str], NoValue]:
+    ...
+
+
+@overload
+def read_from_cache(key: OptionsKey, expiration_time: int = 900) -> Union[list[str], NoValue]:
+    ...
```
rdimaio, line 439: These two have the same argument list and output type, so they could be unified into a single overload:

```suggestion
def read_from_cache(key: Union[OptionsKey, Literal['sections']], expiration_time: int = 900) -> Union[list[str], NoValue]:
    ...
```

lib/rucio/core/config.py
```diff
@@ -435,6 +466,31 @@ def read_from_cache(key: str, expiration_time: int = 900) -> Any:
...
 
 
+@overload
+def write_to_cache(key: Literal['sections'], value: list[str]) -> None:
+    ...
+
+
+@overload
+def write_to_cache(key: OptionsKey, value: list[str]) -> None:
+    ...
```
rdimaio, line 476: Can be unified into a single overload:

```suggestion
@overload
def write_to_cache(key: Union[OptionsKey, Literal['sections']], value: list[str]) -> None:
    ...
```

rdimaio (commented):
I suggest waiting until [this PR](https://github.com/rucio/rucio/pull/7015/files#diff-3c88400ee4ff0b9fca1b3a4f6dd3f58e23b10991a3a4032cbdbcd7028ae5c264) is merged as it will change `core/config`, so some refactoring will need to be done after that

Issue #7008, opened 2024-08-07
Improve test coverage


rdimaio:
Tracker for test coverage - follow-up to https://github.com/rucio/rucio/issues/6545

PR #7068, opened 2024-09-02
Core: Fix RucioException bug, refactor, add TestRucioException

### LINKED ISSUE #7065: RucioException crashes when kwargs are passed
### Description

https://github.com/rucio/rucio/blob/5d42f2c54bc60a213701bb6b77bf810835a280b4/lib/rucio/common/exception.py#L26-L54

The issue is with this line:

https://github.com/rucio/rucio/blob/5d42f2c54bc60a213701bb6b77bf810835a280b4/lib/rucio/common/exception.py#L34

`Exception` does not take `kwargs`. Passing a keyword argument to `RucioException` thus results in:

```
    def __init__(self, *args, **kwargs):
>       super(RucioException, self).__init__(*args, **kwargs)
E       TypeError: MockException() takes no keyword arguments
```

### Steps to reproduce

Pass a kwarg to `RucioException` or to one of its subclasses

### Rucio Version

_No response_

### Additional Information

_No response_

### LINKED ISSUE #7066: RucioException: Remove unused `kwargs` parsing logic
This logic:

https://github.com/rucio/rucio/blob/5d42f2c54bc60a213701bb6b77bf810835a280b4/lib/rucio/common/exception.py#L42-L46

Only works when `_message` contains conversation specifications (e.g. `%s`): https://docs.python.org/3.3/library/stdtypes.html#printf-style-string-formatting

Currently, no subclass exceptions are using this logic. The few custom exceptions that contain `%s` use it with a constant:

https://github.com/rucio/rucio/blob/5d42f2c54bc60a213701bb6b77bf810835a280b4/lib/rucio/common/exception.py#L799

Or with named keyword arguments:

https://github.com/rucio/rucio/blob/5d42f2c54bc60a213701bb6b77bf810835a280b4/lib/rucio/common/exception.py#L907-L910

This logic is currently unused, and it's not used by the base `RucioException`. I think it should be removed, as any subclass exception that would need to use this would likely name the keyword arguments that it would need for the string.

rdimaio:
- fix #7065 
- fix #7066 
- part of #7008

### DIFF
--- a/lib/rucio/common/exception.py
+++ b/lib/rucio/common/exception.py
@@ -31,7 +31,7 @@ class RucioException(Exception):
     """
 
     def __init__(self, *args, **kwargs):
-        super(RucioException, self).__init__(*args, **kwargs)
+        super(RucioException, self).__init__(*args)
         self._message = "An unknown exception occurred."
         self.args = args
         self.kwargs = kwargs
--- a/lib/rucio/common/exception.py
+++ b/lib/rucio/common/exception.py
@@ -30,20 +30,15 @@ class RucioException(Exception):
     with the keyword arguments provided to the constructor.
     """
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args):
         super(RucioException, self).__init__(*args)
         self._message = "An unknown exception occurred."
         self.args = args
-        self.kwargs = kwargs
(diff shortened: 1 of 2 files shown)


voetberg (approved):
👍

PR #7072, opened 2024-09-03
Testing: Remove unnecessary test_download_states

### LINKED ISSUE #7070: Remove unnecessary `test_download_states`
This test can be removed as it does not assert anything useful:

https://github.com/rucio/rucio/blob/5d42f2c54bc60a213701bb6b77bf810835a280b4/tests/test_download.py#L743-L754

rdimaio:
fix #7070

### DIFF
--- a/tests/test_download.py
+++ b/tests/test_download.py
@@ -22,7 +22,7 @@ from zipfile import ZipFile
 
 import pytest
 
-from rucio.client.downloadclient import DownloadClient, FileDownloadState
+from rucio.client.downloadclient import DownloadClient
 from rucio.common.config import config_add_section, config_set
 from rucio.common.exception import InputValidationError, NoFilesDownloaded, RucioException
 from rucio.common.types import InternalScope
@@ -738,20 +738,6 @@ def test_download_exclude_tape(rse_factory, did_factory, download_client):
          TemporaryDirectory() as tmp_dir, \
          pytest.raises(NoFilesDownloaded):
         download_client.download_dids([{'did': did_str, 'base_dir': tmp_dir}])
-
-
-def test_download_states():
-    """ Tests the available download states. """
-    FileDownloadState.PROCESSING
-    FileDownloadState.DOWNLOAD_ATTEMPT
-    FileDownloadState.DONE
-    FileDownloadState.ALREADY_DONE
-    FileDownloadState.FOUND_IN_PCACHE
-    FileDownloadState.FILE_NOT_FOUND
-    FileDownloadState.FAIL_VALIDATE
-    FileDownloadState.FAILED
-
-    assert len(FileDownloadState) == 8
 
 
 def test_download_traceless(rucio_client):


voetberg:
I agree this test itself is pointless, but wouldn't it make more sense to replace it with something that actually tests what it claims to test? The states 'Processing', 'Download Attempt' and 'Found in Pcache', are used by the download client never tested.

rdimaio:
> I agree this test itself is pointless, but wouldn't it make more sense to replace it with something that actually tests what it claims to test? The states 'Processing', 'Download Attempt' and 'Found in Pcache', are used by the download client never tested.

I'm not sure what the test was attempting to test in the first place. It calls the members of the enum and checks the length of the enum, but that's just checking the definition of the enum itself. Maybe we wanted to test against regressions in the state messages (e.g. let's say `FOUND_IN_PCACHE` changed to `FOUND_PCACHE`, maybe we wanted to fail if that happened), but in that case we should have tested the enum values, not the members.

It was added as part of this commit, together with the enum definition itself, but the commit message doesn't elaborate on what the test was meant to do: https://github.com/rucio/rucio/commit/fa501babaec19ae722265a5fae95215d0ee7d799 

If we really want to test this, I think this would be the way to do it: https://stackoverflow.com/a/29795561

But I don't think there's much value in testing non-functional behaviour like this - instead, we should make sure that these download states are all tested within the `downloadclient` tests as part of the state they represent. I think these replies make a good point:
- https://softwareengineering.stackexchange.com/a/358162
- https://stackoverflow.com/a/1103800

voetberg:
> But I don't think there's much value in testing non-functional behaviour like this - instead, we should make sure that these download states are all tested within the downloadclient tests as part of the state they represent. I think these replies make a good point: 

Sorry I wasn't clear, this was exactly what I was suggesting. Should we make an issue out of this?

rdimaio:
> > But I don't think there's much value in testing non-functional behaviour like this - instead, we should make sure that these download states are all tested within the downloadclient tests as part of the state they represent. I think these replies make a good point:
> 
> Sorry I wasn't clear, this was exactly what I was suggesting. Should we make an issue out of this?

I think that's a good idea - so far I've just been using https://github.com/rucio/rucio/issues/7008 as a catch-all issue, but I think it makes sense to have a specific issue here

bari12 (approved):

rdimaio:
> > > But I don't think there's much value in testing non-functional behaviour like this - instead, we should make sure that these download states are all tested within the downloadclient tests as part of the state they represent. I think these replies make a good point:
> > 
> > 
> > Sorry I wasn't clear, this was exactly what I was suggesting. Should we make an issue out of this?
> 
> I think that's a good idea - so far I've just been using #7008 as a catch-all issue, but I think it makes sense to have a specific issue here

I opened this issue to track this: https://github.com/rucio/rucio/issues/7191

PR #7074, opened 2024-09-04
Refactor utils, add tests

### LINKED ISSUE #7071: Clean up unused functions
e.g. `dids_as_dicts` is defined but never used:

https://github.com/search?q=repo%3Arucio%2Frucio%20dids_as_dicts&type=code

(in this case in particular - this logic also already exists as part of `DIDType`)

rdimaio:
Main refactoring changes:
- Refactor bittorrent specific functions into common/bittorrent: 27f2455ecd4ff150a4aa57ed67f53814b42c1256
- Refactor checksum specific functions into common/checksum: 055c2923e43ef3060f02b0429680ea39ed2eb8fb

Related issues:
- part of #7071 
- part of #7008


rdimaio:
Rebased in latest force push

rdimaio:
Rebased in latest forcepush

rdimaio:
Latest revision:
- fix merge conflicts

tests/test_upload.py
```diff
@@ -406,13 +407,17 @@ def upload_client_registration_fail():
     logger = logging.getLogger('upload_client')
     logger.addHandler(logging.StreamHandler())
     logger.setLevel(logging.DEBUG)
+
     # modify the client object used by upload_client so that replica registration fails
     class RegistrationFailureClient(Client):
+
```
voetberg, line 413: Are these spaces necessary?

lib/rucio/common/utils.py
```diff
@@ -63,8 +57,9 @@
         EXTRA_MODULES['paramiko'] = False
 
 if TYPE_CHECKING:
-    from collections.abc import Callable, Iterable, Iterator, Sequence
+    from collections.abc import Callable, Iterable, Iterator, Mapping, Sequence
     T = TypeVar('T')
+    Q = TypeVar('Q')
```
voetberg: T and Q are not very descriptive type names. Is this a common convention I don't know about? Should they be renamed to something more verbose?

tests/test_upload.py
rdimaio: they were automatically added to meet the formatting rules - this file had probably not been touched since we enabled the formatting check, which is why it did not fail before

lib/rucio/common/utils.py
rdimaio: That's the common convention I see around:
- https://docs.python.org/3/library/typing.html#typing.TypeVar
- https://stackoverflow.com/questions/74589610/whats-pylints-typevar-name-specification

Which seems to be close to the PEP 8 specification: https://peps.python.org/pep-0008/#type-variable-names

In the context of these typevars being used in `invert_dict`, I agree that they could be more explicit, I can rename them to have `Hashable` in the name (since they both need to be hashable for `invert_dict`) to work.
`T` is also used in `retrying`, where I'd leave it as `T` considering that it's more generic.

rdimaio:
Latest revision: rebase, fix merge conflicts

lib/rucio/common/utils.py
rdimaio: Addressed in latest revision

rdimaio:
Latest revision: rename `TypeVar`s in `invert_dict` to be more specific

voetberg (approved):
lgtm! Thanks or the changes

PR #7081, opened 2024-09-09
Handle new FTS error message

### LINKED ISSUE #7080: Support new FTS error messages
### Description

With FTS version  v3.12.12 the following error message changed from:
'Destination file exists and overwrite is not enabled'  to
'Destination file exists and is on tape' 

The Rucio codebase will be updated to handle both cases for backwards compatibility for sites using older versions of FTS.

### Steps to reproduce

Self-explanatory

### Rucio Version

_No response_

### Additional Information

_No response_

labkode:
Fixes https://github.com/rucio/rucio/issues/7080

### DIFF
--- a/lib/rucio/transfertool/fts3.py
+++ b/lib/rucio/transfertool/fts3.py
@@ -109,6 +109,9 @@ _SCITAGS_NEXT_REFRESH = datetime.datetime.utcnow()
 _SCITAGS_EXP_ID = None
 _SCITAGS_ACTIVITY_IDS = {}
 
+FTS_FILE_EXISTS_ERROR_MSG = 'Destination file exists and is on tape'  # used in FTS  >= 3.12.12
+FTS_FILE_EXISTS_ERROR_MSG_LEGACY = 'Destination file exists and overwrite is not enabled'  # Error message used in FTS < 3.12.12, checked in Rucio for backwards compatibility
+
 
 def _scitags_ids(logger: "LoggerFunction" = logging.log) -> "tuple[int | None, dict[str, int]]":
     """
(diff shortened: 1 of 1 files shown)


lib/rucio/transfertool/fts3.py
```diff
@@ -685,7 +685,7 @@ def _is_recoverable_fts_overwrite_error(cls, request: dict[str, Any], reason: st
         dst_type = file_metadata.get('dst_type', None)
         METRICS.counter('overwrite.check.{rsetype}.{rse}').labels(rse=file_metadata["dst_rse"], rsetype=dst_type).inc()
 
-        if 'Destination file exists and overwrite is not enabled' in (reason or ''):
+        if 'Destination file exists and overwrite is not enabled' in (reason or '') or 'Destination file exists and is on tape' in (reason or ''):
```
rdimaio: I'm a bit confused by the pre-existing check:
```[string] in (reason or '')```

`reason` is a `str`, and `bool(reason)` would only return `False` if it's empty, in which case it'd  just be: 
```[string] in ('' or '')```

If we expect `request` to optionally be `None`, then we should change the type annotation from `reason: str` to `reason: Optional[str]`, and we can rewrite the check like this:

```suggestion
        if reason and any(s in reason for s in ('Destination file exists and overwrite is not enabled', 'Destination file exists and is on tape')):
```

If we expect `request` to always be a valid `str`, then we can simplify to this:

```suggestion
        if any(s in reason for s in ('Destination file exists and overwrite is not enabled', 'Destination file exists and is on tape')):
```

lib/rucio/transfertool/fts3.py
```diff
@@ -615,7 +615,7 @@ def _transfer_link(self) -> str:
 
     def _find_attribute_updates(self, request: dict, new_state: RequestState, reason: str, overwrite_corrupted_files: Optional[bool] = None) -> Optional[dict[str, Any]]:
         attributes = None
-        if new_state == RequestState.FAILED and 'Destination file exists and overwrite is not enabled' in (reason or ''):
+        if new_state == RequestState.FAILED and ('Destination file exists and overwrite is not enabled' in (reason or '') or 'Destination file exists and is on tape' in (reason or '')):
```
rdimaio: Same as other comment - unsure why we're checking `in (reason or '')` instead of checking for `if reason`

rdimaio (dismissed):

lib/rucio/transfertool/fts3.py
rdimaio: I think it would also make sense to clarify that these error messages cannot both happen in the same site: `'Destination file exists and overwrite is not enabled'` is for FTS < 3.12.12 and `'Destination file exists and is on tape'` is for FTS >= 3.12.12, but without this context, one might think that these are separate errors that can both occur within the same FTS instance on a given site.

I think it could be a good idea to have these strings be variables somewhere in the file, e.g.:

```
fts_file_exists_error_msg = ''Destination file exists and is on tape'
fts_file_exists_error_msg_legacy = 'Destination file exists and overwrite is not enabled'  # Error message used in FTS < 3.12.12, checked in Rucio for backwards compatibility
```

And then accessing them in `_find_attribute_updates` and `_is_recoverable_fts_overwrite_error`. This way, we also avoid the duplication of writing out these strings both times each

labkode:
@rdimaio applied your suggestions, thanks

labkode (commented):
Got them

rdimaio (approved):

labkode:
@bari12 feel free to merge this one.

Issue #7021, opened 2024-08-12
Release 36 type annotations


rdimaio:
Follow-up to https://github.com/rucio/rucio/issues/6588

rdimaio:
Note, the following decorators:
- `count_it`, `time_it` in `core/monitor`
- `*_session` in `db/sqla/session`

 currently cannot be properly type annotated due to these issues:
- https://github.com/rucio/rucio/issues/6986
- https://github.com/rucio/rucio/issues/6989

Until those are resolved, you can ignore missing type annotations for those decorators

PR #7091, opened 2024-09-17
Testing: adding type annotations to core/transfer.py

erlingstaff:
Part of #7021

### DIFF
--- a/lib/rucio/core/transfer.py
+++ b/lib/rucio/core/transfer.py
@@ -20,7 +20,7 @@ import sys
 import time
 import traceback
 from collections import defaultdict
-from typing import TYPE_CHECKING, cast
+from typing import TYPE_CHECKING, Any, Optional, TypeVar, Union, cast
 
 from dogpile.cache import make_region
 from dogpile.cache.api import NoValue
@@ -31,6 +31,7 @@ from rucio.common import constants
 from rucio.common.config import config_get, config_get_list
 from rucio.common.constants import SUPPORTED_PROTOCOLS, RseAttr
 from rucio.common.exception import InvalidRSEExpression, RequestNotFound, RSEProtocolNotSupported, RucioException, UnsupportedOperation
+from rucio.common.types import InternalAccount, InternalScope, LoggerFunction, RequestDict
 from rucio.common.utils import construct_non_deterministic_pfn
 from rucio.core import did
 from rucio.core import message as message_core
@@ -51,16 +52,13 @@ from rucio.transfertool.mock import MockTransfertool
 from rucio.transfertool.transfertool import TransferStatusReport, Transfertool
 
 if TYPE_CHECKING:
-    from collections.abc import Callable, Iterable, Iterator, Mapping, Sequence
-    from typing import Any, Optional
+    from collections.abc import Iterable, Iterator, Sequence
 
     from sqlalchemy.orm import Session
 
-    from rucio.common.types import InternalAccount
     from rucio.core.topology import Topology
     from rucio.rse.protocols.protocol import RSEProtocol
 
-    LoggerFunction = Callable[..., Any]
 
 """
 The core transfer.py is specifically for handling transfer-requests, thus requests
@@ -75,22 +73,24 @@ WEBDAV_TRANSFER_MODE = config_get('conveyor', 'webdav_transfer_mode', False, Non
 
 DEFAULT_MULTIHOP_TOMBSTONE_DELAY = int(datetime.timedelta(hours=2).total_seconds())
 
-TRANSFERTOOL_CLASSES_BY_NAME: "dict[str, type[Transfertool]]" = {
+TRANSFERTOOL_CLASSES_BY_NAME: dict[str, type[Transfertool]] = {
     FTS3Transfertool.external_name: FTS3Transfertool,
     GlobusTransferTool.external_name: GlobusTransferTool,
     MockTransfertool.external_name: MockTransfertool,
     BittorrentTransfertool.external_name: BittorrentTransfertool,
 }
 
+DirectTransferImplementationT = TypeVar("DirectTransferImplementationT", bound="DirectTransferImplementation")
+
 
 class ProtocolFactory:
     """
     Creates and caches protocol objects. Allowing to reuse them.
     """
-    def __init__(self):
+    def __init__(self) -> None:
         self.protocols = {}
 
-    def protocol(self, rse: RseData, scheme: "Optional[str]", operation: str):
+    def protocol(self, rse: RseData, scheme: Optional[str], operation: str) -> "RSEProtocol":
         protocol_key = '%s_%s_%s' % (operation, rse.id, scheme)
         protocol = self.protocols.get(protocol_key)
         if not protocol:
@@ -107,7 +107,7 @@ class DirectTransferImplementation(DirectTransfer):
     during the migration.
     """
     def __init__(self, source: RequestSource, destination: TransferDestination, rws: RequestWithSources,
-                 protocol_factory: ProtocolFactory, operation_src: str, operation_dest: str):
+                 protocol_factory: ProtocolFactory, operation_src: str, operation_dest: str) -> None:
         super().__init__(sources=[source], rws=rws)
         self.destination = destination
 
@@ -118,7 +118,7 @@ class DirectTransferImplementation(DirectTransfer):
         self._dest_url = None
         self._source_urls = {}
 
-    def __str__(self):
+    def __str__(self) -> str:
         return '{sources}--{request_id}->{destination}'.format(
             sources=','.join([str(s.rse) for s in self.sources]),
             request_id=self.rws.request_id or '',
@@ -158,7 +158,7 @@ class DirectTransferImplementation(DirectTransfer):
         return self.protocol_factory.protocol(source.rse, source.scheme, self.operation_src)
 
     @staticmethod
-    def __rewrite_source_url(source_url, source_sign_url, dest_sign_url, source_scheme):
+    def __rewrite_source_url(source_url: str, source_sign_url: str, dest_sign_url: str, source_scheme: str) -> str:
         """
         Parametrize source url for some special cases of source and destination schemes
         """
@@ -184,7 +184,7 @@ class DirectTransferImplementation(DirectTransfer):
         return source_url
 
     @staticmethod
-    def __rewrite_dest_url(dest_url, dest_sign_url):
+    def __rewrite_dest_url(dest_url: str, dest_sign_url: str) -> str:
         """
         Parametrize destination url for some special cases of destination schemes
         """
@@ -200,7 +200,7 @@ class DirectTransferImplementation(DirectTransfer):
         return dest_url
 
     @classmethod
-    def _generate_source_url(cls, src: RequestSource, dst: TransferDestination, rws: RequestWithSources, protocol_factory: ProtocolFactory, operation: str):
+    def _generate_source_url(cls: type[DirectTransferImplementationT], src: RequestSource, dst: TransferDestination, rws: RequestWithSources, protocol_factory: ProtocolFactory, operation: str) -> str:
         """
         Generate the source url which will be used as origin to copy the file from request rws towards the given dst endpoint
         """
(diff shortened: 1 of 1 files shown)


lib/rucio/core/transfer.py
```diff
@@ -211,11 +211,11 @@ def _generate_source_url(cls, src: RequestSource, dst: TransferDestination, rws:
         source_sign_url = src.rse.attributes.get(RseAttr.SIGN_URL, None)
         dest_sign_url = dst.rse.attributes.get(RseAttr.SIGN_URL, None)
         source_url = list(protocol.lfns2pfns(lfns={'scope': rws.scope.external, 'name': rws.name, 'path': src.file_path}).values())[0]
-        source_url = cls.__rewrite_source_url(source_url, source_sign_url=source_sign_url, dest_sign_url=dest_sign_url, source_scheme=src.scheme)
+        source_url = cls.__rewrite_source_url(source_url, source_sign_url=source_sign_url, dest_sign_url=dest_sign_url, source_scheme=str(src.scheme))
```
rdimaio, line 214: `src.scheme` is `Optional[str]`, so `str(src.scheme)` would return `'None'` if `src.scheme` is `None`:
```
>>> x: Optional[str] = None
>>> str(x)
'None'
```

We need to be a bit careful with this: for example, if in `__rewrite_source_url` we were checking either:

```
if source_scheme:
```
or

```
if source_scheme is not None:
```

then in case of `src.scheme = None`, doing `source_scheme=str(src.scheme)` would change the outcome of both of those checks from `False` to `True`:

```
>>> x = None
>>> bool(x)
False
>>> bool(x is not None)
False
>>> bool(str(x))
True
>>> bool(str(x) is not None)
True
```

In this case, we are not doing these kinds of checks in `__rewrite_source_url`, so it's 'safe' to do this here; however, I think it's best to avoid calling `str()` directly on an object that is `Optional[str]`, because `str(None)` is likely not what we want to do. In this case, we could do either one of these options:

1. changing this line to:
```suggestion
        source_url = cls.__rewrite_source_url(source_url, source_sign_url=source_sign_url, dest_sign_url=dest_sign_url, source_scheme=(src.scheme or ''))
```

this way, `__rewrite_source_url` can still expect a `str` for `source_scheme`, but it can still check whether `source_scheme` is an empty string

2. changing `__rewrite_source_url` to accept an `Optional[str]` for `source_scheme`, and checking `if source_scheme` before using it.

I think the first option is better in this case

lib/rucio/core/transfer.py
```diff
@@ -211,11 +211,11 @@ def _generate_source_url(cls, src: RequestSource, dst: TransferDestination, rws:
...
         return source_url
 
     @classmethod
-    def _generate_dest_url(cls, dst: TransferDestination, rws: RequestWithSources, protocol_factory: ProtocolFactory, operation: str):
+    def _generate_dest_url(cls: type[DirectTransferImplementationT], dst: TransferDestination, rws: RequestWithSources, protocol_factory: ProtocolFactory, operation: str) -> str:
```
rdimaio, line 218: I'm not sure if we should annotate `cls`, in the same way that we don't currently annotate `self`. I think it's outside the scope of this PR though, so you can ignore this. At some point we can formalize this in the type annotation documentation and make changes to the code if needed

lib/rucio/core/transfer.py
```diff
@@ -400,24 +400,24 @@ def ensure_db_sources(
...
+) -> None:
     """
     Update the transfer info of a request.
     :param transfers:  Dictionary containing request transfer info.
     :param session:    Database session to use.
     """
 
     logger(logging.INFO, 'Setting state(%s), transfertool(%s), external_host(%s) and eid(%s) for transfers: %s',
-           state.name, transfertool, external_host, external_id, ', '.join(t.rws.request_id for t in transfers))
+           state.name, transfertool, external_host, external_id, ', '.join(str(t.rws.request_id) for t in transfers))
```
rdimaio, line 420: Same point regarding `str()` here - `RequestWithSources.request_id` is `Optional[str]`, so this would log the string `'None'` in case `t.rws.request_id` is `None`. Definitely not an issue in this case since it's just logging, but as a suggestion I think the logs might be clearer if we do something like:

```suggestion
           state.name, transfertool, external_host, external_id, ', '.join((t.rws.request_id or 'MISSING REQUEST ID') for t in transfers))
```

e.g. let's say all request IDs are missing - if all of them are `None` it's not as easy to figure out that the `None` are request IDs, since the first logger message just calls them "transfers". If we call them "missing request IDs", it's a bit clearer.

lib/rucio/core/transfer.py
```diff
@@ -520,7 +520,7 @@ def update_transfer_state(
         fields_to_update = tt_status_report.get_db_fields_to_update(session=session, logger=logger)
         if not fields_to_update:
             request_core.update_request(request_id, raise_on_missing=True, session=session)
-            return False
+            return 0
```
rdimaio, line 523: This is a functional change, please rebase it in a separate commit

lib/rucio/core/transfer.py
```diff
@@ -556,13 +556,13 @@ def update_transfer_state(
             return nb_updated
     except UnsupportedOperation as error:
         logger(logging.WARNING, "Request %s doesn't exist - Error: %s" % (request_id, str(error).replace('\n', '')))
-        return 0
     except Exception:
         logger(logging.CRITICAL, "Exception", exc_info=True)
+    return 0
```
rdimaio, line 561: This is also a functional change, related to the previous one - please rebase this one in a separate commit as well (ideally the same commit as the other change in this function)

lib/rucio/core/transfer.py
```diff
@@ -788,7 +788,7 @@ def __compress_multihops(
     for source, path in paths_by_source:
         if len(path) > 1:
             # find the index of the first hop starting from the end which is also a source. Path[0] will always be a source.
-            last_source_idx = next((idx for idx, hop in reversed(list(enumerate(path))) if hop.src.rse.id in source_rses), (0, None))
+            last_source_idx = next((idx for idx, hop in reversed(list(enumerate(path))) if hop.src.rse.id in source_rses), (0))
```
rdimaio, line 791: This is a functional change, please rebase it in a separate commit

lib/rucio/core/transfer.py
```diff
@@ -932,7 +932,7 @@ class _ClassNameDescriptor:
         """
         Automatically set the external_name of the strategy to the class name.
         """
-        def __get__(self, obj, objtype=None):
+        def __get__(self, obj: object, objtype: Optional[Any] = None) -> str:
```
rdimaio, line 935:
```suggestion
        def __get__(self, obj: object, objtype: Any = None) -> str:
```

`Any` allows for `None` as well, so `Optional[Any]` is not necessary

lib/rucio/core/transfer.py
```diff
@@ -972,68 +972,68 @@ def for_request(
...
-    def apply(self, ctx: RequestRankingContext, source: RequestSource) -> "Optional[int | _SkipSource]":
+    def apply(self, ctx: RequestRankingContext, source: RequestSource) -> Optional[Union[int, _SkipSource]]:
         return int(source.rse.is_tape_or_staging_required())  # rely on the fact that False < True
 
 
 class PathDistance(SourceRankingStrategy):
 
     class _RankingContext(RequestRankingContext):
-        def __init__(self, strategy: "SourceRankingStrategy", rws: "RequestWithSources", paths_for_rws: "Mapping[RseData, Sequence[DirectTransfer]]"):
+        def __init__(self, strategy: SourceRankingStrategy, rws: RequestWithSources, paths_for_rws: dict[RseData, list[DirectTransfer]]) -> None:
```
rdimaio, line 1032: I think this should be reverted, unless it was causing type errors before.

I think you made this change because you changed `build_or_return_cached` to always return `dict` instead of `Mapping` (which is correct, because that is the type returned and we should be as specific as possible with return types of private functions).

In the context of `_RankingContext`:  if you were to pass a `Mapping` for `paths_for_rws`, instead of specifically a `dict`, all usages of `_RankingContext` would still work, since we're only using the `get` method on `paths_for_rws`, and `get` exists on `Mapping`. So `_RankingContext` doesn't necessarily need a `dict` here, it can accept any `Mapping`.

lib/rucio/core/transfer.py
```diff
@@ -1073,7 +1073,7 @@ def __init__(self) -> None:
             self.files_done = 0
             self.files_failed = 0
 
-        def incorporate_stat(self, stat: "Mapping[str, int]") -> None:
+        def incorporate_stat(self, stat: dict[str, int]) -> None:
```
rdimaio, line 1076: same as the previous comment - this method can work just fine with a `Mapping` and does not specifically need a `dict`

lib/rucio/core/transfer.py
```diff
@@ -1125,19 +1125,19 @@ def apply(self, ctx: RequestRankingContext, source: RequestSource) -> "Optional[
...
-        transfertools: "Optional[list[str]]" = None,
+        transfertools: Optional[list[str]] = None,
         requested_source_only: bool = False,
         preparer_mode: bool = False,
         *,
         session: "Session",
-        logger: "Callable" = logging.log,
-):
+        logger: LoggerFunction = logging.log,
+) -> "tuple[dict[str, Sequence[DirectTransfer]], set[str], set[str], set[str], set[str]]":
```
rdimaio, line 1140: All of the `str`s in this return type are `rws.request_id`, so to be compliant with its type this should be:

```suggestion
) -> "tuple[dict[Optional[str], Sequence[DirectTransfer]], set[Optional[str]], set[Optional[str]], set[Optional[str]], set[Optional[str]]]":
```

To confirm this, on line 1029, where all the returned elements are initialized:

```python
candidate_paths_by_request_id, reqs_no_source, reqs_only_tape_source, reqs_scheme_mismatch = {}, set(), set(), set()
```

If you were to annotate their type at initialization time, e.g.:

```python
reqs_no_source: set[str] = set()
```

You'd get this error when a `request_id` is added to them (e.g., when `reqs_no_source.add(rws.request_id)` happens):

```
Type "str | None" is not assignable to type "str"
```

However, I think your return type annotation is correct, and we should keep it like this. The 'compliant' type annotation wouldn't make much sense - `dict` expects unique key and `set` expects unique elements, so what happens if there's multiple `request_id`s that are `None`?

I think there's a pre-existing issue that we need to fix somewhere else. My impression is that `RequestWithSources` should always be expected to have a `request_id`, so the best fix would likely be to change that type from `Optional[str]` to `str`. 
If my impression is wrong and `RequestWithSources` can have a missing `request_id`, then we'd need to modify this function (`build_transfer_paths`) to check that `rws.request_id is not None` before adding it to the returned dict/sets (and we would probably want to log a message in case `rws.request_id is None`).

lib/rucio/core/transfer.py
```diff
@@ -1353,7 +1353,7 @@ def update_transfer_priority(transfers_to_update, logger=logging.log):
             logger(logging.DEBUG, "Updated transfer %s priority in transfertool to %s: %s" % (transfer_id, priority, res['http_message']))
 
 
-def cancel_transfers(transfers_to_cancel, logger=logging.log):
+def cancel_transfers(transfers_to_cancel: dict[str, dict[str, int]], logger: LoggerFunction = logging.log) -> None:
```
rdimaio, line 1356:
```suggestion
def cancel_transfers(transfers_to_cancel: dict[str,  Iterable[str]], logger: LoggerFunction = logging.log) -> None:
```

`FTS3Transfertool.cancel` expects a `Sequence[str]` as its input, and `transfer_ids` is just iterated over, so an `Iterable[str]` works. I think the docstring suggests that `transfer_ids` is a `set`, but I'm not sure. 

(If you want to annotate it as a `set`, please use `collections.abc.Set`): 

> Note that to annotate arguments, it is preferred to use an abstract collection type such as [collections.abc.Set](https://docs.python.org/3/library/collections.abc.html#collections.abc.Set) rather than to use [set](https://docs.python.org/3/library/stdtypes.html#set) or [typing.Set](https://docs.python.org/3/library/typing.html#typing.Set).

https://docs.python.org/3/library/typing.html#typing.Set

rdimaio (changes requested):

lib/rucio/core/transfer.py
erlingstaff: Does the CI ignore `ANN401`? My flake8 does not like naked `Any` annotations

image

rdimaio: `ANN` is not enabled: https://github.com/rucio/rucio/blob/e6097bd2ee4efb1ade8b4e262a28efdda6fb96a2/pyproject.toml#L56-L87 - it might be your local Ruff install that complains about it, but the CI should be alright (we use `Any` in quite a few other places)

Geogouz:
Problems to be addressed pending

bziemons:

PR #7093, opened 2024-09-18
Testing: Add missing type annotations to core/scope.py

erlingstaff:
Part of #7021

### DIFF
--- a/lib/rucio/core/scope.py
+++ b/lib/rucio/core/scope.py
@@ -20,21 +20,23 @@ from sqlalchemy import and_, select
 from sqlalchemy.exc import IntegrityError
 
 from rucio.common.exception import AccountNotFound, Duplicate, RucioException, VONotFound
-from rucio.common.types import InternalScope
+from rucio.common.types import InternalAccount, InternalScope
 from rucio.core.vo import vo_exists
 from rucio.db.sqla import models
 from rucio.db.sqla.constants import AccountStatus, ScopeStatus
 from rucio.db.sqla.session import read_session, transactional_session
 
 if TYPE_CHECKING:
+    from collections.abc import Sequence
+
     from sqlalchemy.orm import Session
 
 
 @transactional_session
-def add_scope(scope, account, *, session: "Session"):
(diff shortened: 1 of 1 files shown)


lib/rucio/core/scope.py
```diff
@@ -64,12 +66,18 @@ def add_scope(scope, account, *, session: "Session"):
...
+    except Exception:
         raise RucioException(str(format_exc()))
 
 
 @read_session
-def bulk_add_scopes(scopes, account, skipExisting=False, *, session: "Session"):
+def bulk_add_scopes(
+    scopes: list[InternalScope],
+    account: InternalAccount,
+    skipExisting: Optional[bool] = False,
```
rdimaio, line 77:
```suggestion
    skipExisting: bool = False,
```

The `Optional` in `Optional[bool]` refers to the type of `skipExisting`, i.e. it could either be `None` or `bool`. In this context, we probably always expect this to be a `bool`, even when this parameter is not passed in a call to `bulk_add_scopes` (in which case it defaults to `False`).


Off-topic: `skipExisting` should be renamed to `skip_existing` for consistency with our naming convention

lib/rucio/core/scope.py
```diff
@@ -64,12 +66,18 @@ def add_scope(scope, account, *, session: "Session"):
...
             raise RucioException(e)
-    except:
+    except Exception:
         raise RucioException(str(format_exc()))
 
 
 @read_session
-def bulk_add_scopes(scopes, account, skipExisting=False, *, session: "Session"):
+def bulk_add_scopes(
+    scopes: list[InternalScope],
```
rdimaio, line 75:
```suggestion
    scopes: Iterable[InternalScope],
```

`scopes` is only used as an iterable in this function

rdimaio (changes requested):

Geogouz:
Problems to be addressed pending

bziemons:

* A review requested changes and there has been no activity for over seven days.

rdimaio:
Some additional context: these annotations ended up being added in a separate PR https://github.com/rucio/rucio/commit/08992377319c9a4ffbf24573fde49e0a93d125d4

PR #7094, opened 2024-09-18
Testing: Add missing type annotations to core/subscription.py

erlingstaff:
Part of #7021

### DIFF
--- a/lib/rucio/core/subscription.py
+++ b/lib/rucio/core/subscription.py
@@ -17,7 +17,7 @@ import logging
 import re
 from configparser import NoOptionError, NoSectionError
 from json import dumps
-from typing import TYPE_CHECKING, Optional
+from typing import TYPE_CHECKING, Any, Optional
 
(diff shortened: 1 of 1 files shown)


lib/rucio/core/subscription.py
```diff
@@ -332,11 +337,11 @@ def list_subscription_rule_states(name=None, account=None, *, session: "Session"
     )
 
     for row in session.execute(stmt).all():
-        yield row
+        yield row._tuple()
```
rdimaio, line 340: Relevant discussion: https://github.com/rucio/rucio/discussions/6615

rdimaio (approved):
LGTM - Regarding using `_tuple` for SQLA query return types, I think it's fine for now, but there are the limitations listed here https://github.com/rucio/rucio/discussions/6615 that need to be considered. Once SQLA 2.1 is out, we will move this back to just returning the `Row` directly, as it will also include the necessary type information for type checking

PR #7096, opened 2024-09-18
Clients: Creates 'BaseClient.authenticated_account' after authentication

### LINKED ISSUE #7062: Consider renaming `BaseClient.account` to `BaseClient.authenticated_account` for clarity
https://github.com/rucio/rucio/blob/5d42f2c54bc60a213701bb6b77bf810835a280b4/lib/rucio/client/baseclient.py#L140

This idea comes from a conversation with Martin. In `BaseClient`, we initialize `self.account` to the account used to authenticate to Rucio. However, we also use `account` as an input argument to some methods of `BaseClient` subclasses, e.g. in `SubscriptionClient.list_subscriptions`:

https://github.com/rucio/rucio/blob/5d42f2c54bc60a213701bb6b77bf810835a280b4/lib/rucio/client/subscriptionclient.py#L91

Using `account` to refer to both the authenticated client account and to the input account to some methods might be confusing. We should consider renaming `BaseClient.account` to `BaseClient.authenticated_account` to distinguish it from the methods that take `account` as an input.

voetberg:
Addresses #7062 

Changes references of BaseClient.account to BaseClient.authenticated_account, to differentiate between the client.account used as a client arg and the user account used for auth.

### DIFF
--- a/bin/rucio
+++ b/bin/rucio
@@ -1286,7 +1286,7 @@ def delete_rule(args):
         scope, name = get_scope(args.rule_id, client)
         rules = client.list_did_rules(scope=scope, name=name)
         if args.rule_account is None:
-            account = client.account
+            account = client.authenticated_account
         else:
             account = args.rule_account
         deletion_success = False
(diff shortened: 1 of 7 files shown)


rdimaio (approved):

bari12:
Sorry for the long delay, rebase please :-)

Issue #1839, opened 2018-11-29
Inconsistency in command-line options to specify RSEs


dchristidis:
Motivation
----------

It’s difficult memorising command-line options when they are named differently across commands. `delete-rule` uses `--rse_expression`, `list-file-replicas` has both `--rse` and `--expression` (with different meaning), `get` has `--rse` but it actually takes an RSE expression.

Modification
------------

First it has to be decided whether it’s worth breaking the current CLI interface to fix this.

bari12:
This should be consolidated. I would not be too afraid of backwards-compatibility, but maybe we should still add the new parameter, leave the old one in, but if it is used display some kind of message. A few (feature) versions later we then take it out. We used this concept already for several other CLI changes.

mlassnig:
Much of it is historical due to the API itself, but we can start changing the clients and bring the API, REST, and core slowly up to speed. At least for the RSE parameter I would suggest for conciseness to always use `--rse` but always parse RSE expressions, and get rid of the `--rse-expression` or `--expression` extra arguments.

vingar:
As rse expression can give a back a list, i.e. `--rses` looks more appropriate imo. starting with the same name for both `rse-expression` and `expression` would be already a good step.

bari12:
Sounds good to me:
- `--rse` if it is really only 1 RSE
- `--rses` if it is an expression. Although --rses suggests a bit that you can enumerate them, which is not really true (`RSEA|RSEB|RSEC` would work but not `RSEA,RSEB,RSEC`)

mlassnig:
But shouldn't we handle single RSE or multiple RSEs automatically? `--rse` is a basically a subset of `--rses` and having two different parameters is confusing.

bari12:
We would have to go through it, in most cases `--rses` is ok. But in cases where you really only want one RSE, e.g. marking a replica as bad/lost, --rses doesn't really make sense, as it suggests a list, which doesn't work.

mlassnig:
I would do `--rse` everywhere, and in case a list comes back instead of a single RSE fail to the client and tell "hey you selected more than one", instead of having two parameters. (In a way, like SQL cursors)

vingar:
ah but are you proposing to always resolve RSE by expression ? Most of the time the RSE expression corresponds to the RSE name. It might be worth to check the implementation as this will be the most popular call then...

bari12:
From a performance that should be fine, the expression is cached server-side, so it's not even going to the DB too much.

[...]

PR #7098, opened 2024-09-18
Clients: New structure for CLI Commands

voetberg:
#6639 

Completely rewrites the existing bin/rucio and bin/rucio-admin to bring them under the same structure. 

Full description of changes can be found here: https://github.com/voetberg/rucio/wiki/Commands-Rewrite-Summary 

Main structural changes: 
* Adds a deprecation warning to the existing cli commands, but allows them to still be run 
* Moves bin/rucio(-admin) to lib/rucio/client/commands/bin_legacy 
* Creates a CommandBase abstract class for all commands to use
* Requires all functions from bin/rucio(-admin) to accept args, client, and logger

Functional changes and mapping between old command and new commands are included in the rewrite summary. 

A documentation page based off that same summary will be added once this PR is finalized. 

Testing failure is due to pyright picking up old problems in bin/rucio(-admin) and declaring them as new problems (because they're in a new file)

Edit: 

View this dummy pr on my fork to comment on the structure itself. It's just the help menus of all commands and subcommands. Operations are omitted because it would get pretty busy in there. 

https://github.com/voetberg/rucio/pull/1/files


tests/test_bin_rucio.py
```diff
@@ -2175,7 +2175,7 @@ def test_update_rule_no_child_selfassign(self):
         cmd = r"rucio list-rules {0}:{1} | grep {0}:{1} | cut -f1 -d\ ".\
```
rdimaio, line 2176: These client commands in tests are using the legacy format - should we keep them like this or move them to the new format?

I think it's fine to have them using the legacy format for now, but later on we might want to move to the new version

tests/test_bin_translations.py
rdimaio, line 1: The name of this file (`test_bin_translations`) confuses me a bit - I assume it means that it tests the mapping from the legacy commands to the new commands, but I see that the tests use the new format (e.g. `rucio did content`)

If we are trying to test these translations, I think that a good way to test them would be to parametrize the tests, running each test two times, one time using the legacy command and one time using the modern command. I'm not sure if this is feasible for the tests that have been marked as `dirty` / `noparallel`, so a simpler way of testing this mapping could just be to test that the base mapping logic works, and then just testing all the commands once.

tests/test_bin_rucio.py
```diff
@@ -253,7 +253,7 @@ def test_rse_add_distance(self):
         exitcode, out, err = execute(cmd)
         print(out, err)
         assert exitcode == 0
-        cmd = 'rucio-admin rse add-distance --distance 1 --ranking 1 %s %s' % (temprse2, temprse1)
+        cmd = 'rucio-admin -v rse add-distance --distance 1 --ranking 1 %s %s' % (temprse2, temprse1)
```
rdimaio: Is the verbose flag needed here?

lib/rucio/client/commands/command.py
```diff
@@ -0,0 +1,228 @@
...
+        auth_args.add_argument(
+            "--oidc-polling",
+            default=False,
+            action="store_true",
+            help="If not specified, user will be asked to enter a code returned by the browser to the command line. "
+            + "If --polling is set, Rucio Client should get the token without any further interaction of the user. This option is active only if --auto is *not* specified.",
+        )  # NOQA: W503
+        auth_args.add_argument(
+            "--oidc-refresh-lifetime",
+            help="Max lifetime in hours for this an access token will be refreshed by asynchronous Rucio daemon. "
```
rdimaio:
```suggestion
            help="Max lifetime in hours for this access token; the token will be refreshed by an asynchronous Rucio daemon. "
```

rdimaio (commented):
Ideally, I think this PR should be broken down into multiple PRs, at the very least these ones:
1. one PR to add the basecommand logic for the new infrastructure
2. one PR for each command (`rule`, `replica`...)
3. one PR for the legacy mapping

I realize that this might potentially be a lot of overhead at this point, but I don't think it's feasible to properly review all the changes here as part of this single PR.

I am also fine with the idea of just going ahead and merging this. It seems to work and the overall architecture makes sense, and we can just rollback / address any bugs or regressions later on.

tests/test_bin_rucio.py
voetberg: I was debating this as well. I left them as is for the moment because I wanted to ensure the old commands still worked as expected (with the release schedule requiring them to still be functional). I'd like to move them to the new format soon though.

tests/test_bin_rucio.py
voetberg: It isn't lol. I think I forgot to revert it before submitting this pr

voetberg:
> Ideally, I think this PR should be broken down into multiple PRs, at the very least these ones:
> 
>     1. one PR to add the basecommand logic for the new infrastructure
> 
>     2. one PR for each command (`rule`, `replica`...)
> 
>     3. one PR for the legacy mapping

I would prefer this as well, but because all the commands are dependent on having the location of bin/rucio(-admin) changed. I wanted to both ensure there were no test failures in the final PR, and that nothing could be merged out of order.

dchristidis (commented):
The associated summary document is very helpful! I will commence my review exclusively from it.

> default operation - which will be run when running `rucio group`

I have to try this in action, but could you describe where did the idea came from? Are there any other CLI utilities that do this?

> common kwargs are given both single letter kwargs and full names (e.x. -r and --rse)

The history behind the different conventions is fascinating. I have my own very opinionated stance and, thus,  I can’t help but wonder whether this is beneficial. Especially since some options have pre-existing expectations (e.g. `-r` typically means recursive).

> Groups Rucio Concepts into logical command groups

The group for replication rules appears to be missing.

More importantly, it is difficult to understand what combinations are valid. Would it be possible to offer the equivalent of `openstack --help`? This applies to both the document, but also a wish for the behaviour of the client.

I’m afraid I have too many comments about the new command structure. But trying to do so in this GitHub comment make collaboration difficult, especially if other want to offer their views. We will need another tool where we start with all possible commands (see above) and then add comments below.

> Root

I was hoping we could have done something better here. The separation between root and command-specific options has always been a sore spot. For example:

* `rucio delete-rule --account=jdoe foo:bar` means delete the replication rules for the DID `foo:bar` owned by the account `jdoe`.
* `rucio --account=jdoe delete-rule foo:bar` means authentication as account `jdoe` and delete *all* replication rules for the DID `foo:bar`.

voetberg:
Easy things first: 

> The group for replication rules appears to be missing.

Added! 
 
> I have to try this in action, but could you describe where did the idea came from? Are there any other CLI utilities that do this? 

None immediately come to mind. This was done just because it's fairly easy to implement and typing out "rucio did" instead of "rucio list-dids" is faster. It can also be un-implimented easily if it's confusing. Closest thing I can think of is something like `git branch` which lists all branches and then you add a specific argument like "-d" to delete them. 

> I was hoping we could have done something better here. The separation between root and command-specific options has always been a sore spot. For example: 

This also bothers me, but I don't know if there is an easy way to do this, besides for ensuring all names in auth are different from all other ones. 
Only other thing I can think of is straight up removing the auth args and asking for a separate config file instead, which is probably very annoying. I do know of several groups that just have aliases set up to just store their auth params, it wouldn't be too different from that. 

> I’m afraid I have too many comments about the new command structure. But trying to do so in this GitHub comment make collaboration difficult, especially if other want to offer their views. We will need another tool where we start with all possible commands (see above) and then add comments below. 

I can make a dummy pr in my fork of rucio with just a text doc of all the help menus, it would be easier to comment line by line there. 

> More importantly, it is difficult to understand what combinations are valid.  

I think this is a problem resulting from the static document. In practice, I expect user to behavior to go something like: 
"Oh, I need to do something with DIDs, what are my options again?" 
```
rucio did -h 
usage: rucio did [-h] [--recursive] [--filter FILTER] [--short] [-d DID] {content,metadata,container,dataset,list,stat,remove,touch,parent,attach,detach,upload,download} ...

Default Operation: list
Manage Data IDentifiers. Modify and access specific files and groups of files. DIDs are accessed by the pattern `scope`:`name`, where name can be a wildcard, but scope must be specified.
Operations: ['list', 'stat', 'remove', 'touch', 'parent', 'attach', 'detach', 'upload', 'download']
Subcommands: ['content', 'metadata', 'container', 'dataset']
```

"Cool uhhh actually it was with containers. Let's check that." 
```
usage: rucio did container [-h] {add} ...

Manage Containers (high-level grouping construct).
Operations: ['add']
Subcommands: ['content', 'metadata', 'container', 'dataset']

Usage Example:
$ rucio did container add --did user.jdoe:test.cont.1234.1 # Make a new container named test.cont.1234.1

positional arguments:
  {add}
    add       Create a new container
```

"Ah, yes. `rucio did container add --did myscope:mynew.container`"

I think this bit of menuing is a compromise that is naturally inherited from the subject-object-verb structure. The object can change and not all verbs are implemented for all objects.

voetberg:
>> I’m afraid I have too many comments about the new command structure. But trying to do so in this GitHub comment make collaboration difficult, especially if other want to offer their views. We will need another tool where we start with all possible commands (see above) and then add comments below.

> I can make a dummy pr in my fork of rucio with just a text doc of all the help menus, it would be easier to comment line by line there.

 @dchristidis 
I made good on this - https://github.com/voetberg/rucio/pull/1/files

dchristidis:
> In practice, I expect user to behavior to go something like:

Again, I might need to try this out to be sure. But I do feel like OpenStack is a better model. For one thing, you can rapidly get a sense of all possible commands:

```console
openstack coe --help
Command "coe" matches:
  coe ca rotate
  coe ca show
  coe ca sign
  coe cluster config
  coe cluster create
  coe cluster delete
  coe cluster list
  coe cluster resize
  coe cluster show
  coe cluster template create
  coe cluster template delete
  coe cluster template list
  coe cluster template show
  coe cluster template update
  coe cluster update
  coe cluster upgrade
  coe nodegroup create
  coe nodegroup delete
  coe nodegroup list
  coe nodegroup show
  coe nodegroup update
  coe quotas create
  coe quotas delete
  coe quotas list
  coe quotas show
  coe quotas update
  coe service list
  coe stats list
```

For another, the current description suggests combinations that are not possible, such as `rucio account identity usage`.

voetberg:
> Again, I might need to try this out to be sure. But I do feel like OpenStack is a better model. For one thing, you can rapidly get a sense of all possible commands:

I do agree with this, it would be the cleaner solution. However, we're limited a bit by argparse (which we have to use to maintain backwards compatibility), which just doesn't have this functionality. We can hack and subclass it to bits but 🤷 Time/Effort trade-offs and all that. 

> For another, the current description suggests combinations that are not possible, such as `rucio account identity usage`.

Yeah, this is a bit confusing. I'm not exactly sure how best to communicate that the `subcommands` and `operations` list are at the same level and don't feed into each other. Everything that does this sort of thing successfully  works off custom cli tooling/parsing. Poor craftsman blaming his tools and all that, but I'm very open to suggestions.

dchristidis:
How unfortunate. Would you mind if I asked you to produce this list by hand, so that we can continue the discussion on the interface?

voetberg:
> How unfortunate. Would you mind if I asked you to produce this list by hand, so that we can continue the discussion on the interface?

Yeah that's no problem - check it out https://github.com/voetberg/rucio/commit/3f0c9c44fd21971f58acaf4522dbab82ae1aecdd#diff-d0b439c5fffdc7182aef7173c2790dfcea9ed5468175788c603740691f940e86R4

voetberg:
> Yeah, this is a bit confusing. I'm not exactly sure how best to communicate that the `subcommands` and `operations` list are at the same level and don't feed into each other. Everything that does this sort of thing successfully works off custom cli tooling/parsing. Poor craftsman blaming his tools and all that, but I'm very open to suggestions.

I lied - I tried and this took maybe 4 minutes to write something to generate this automatically. 
```
usage: rucio account [-h] [--type {GROUP,USER,SERVICE}] [-a ACCOUNT] [--id IDENTITY] [--filters FILTERS] {ban,attribute,limit,identity,list,add,info,remove,usage,set} ...

Methods to add or change accounts for users, groups, and services. Used to assign privileges.
Default Operation: list

Matching Commands:
rucio account list
rucio account add
rucio account info
rucio account remove
rucio account usage
rucio account set
rucio account ban add
rucio account ban remove
rucio account attribute list
rucio account attribute add
rucio account attribute remove
rucio account limit list
rucio account limit add
rucio account limit remove
rucio account identity list
rucio account identity add
rucio account identity remove
```

tests/test_bin_translations.py
voetberg: I never replied this - the name is bad it doesn't communicate what it's actually supposed to do. I plan for this to be a stop-over file to ensure the new commands are functional and map to what they're supposed to, not be a one-to-one translation. The old `test_bin` can be updated to the new command structure when we decide to formally depreciate it.  

It's renamed now to `tests/test_cli_client_structure.py`

voetberg:
Notes on test failures: 
* Pyright and Coverage failures are due to CI picking up existing problems from bin/rucio and declaring them as new since they're in a new place 
* docs failure is a little confusing though. I will put an additional PR in over there to specifically change how docs for the bin are generated

lib/rucio/client/commands/account.py
```diff
@@ -0,0 +1,240 @@
...
+    from rucio.client.commands.utils import OperationDict
+
+
+class Account(CommandBase):
+    def module_help(self) -> str:
+        return "Methods to add or change accounts for users, groups, and services. Used to assign privileges."
+
+    def usage_example(self) -> list[str]:
+        return [
+            "$ rucio account  # List all accounts on the instance" "$ rucio account add --account user_jdoe --type USER  # Create a new user account",
```
dchristidis: Missing comma and newline.

lib/rucio/client/commands/account.py
```diff
@@ -0,0 +1,240 @@
...
+            "$ rucio account update --account user_jdoe --email jdoe@cern.ch  # Update jdoe's email",
```
dchristidis, line 51: I think we’ve gone overboard with removing positional parameters. It’s not always beneficial (e.g. `ls /usr/bin/` is better than `ls --path=/usr/bin/`). In this case, `rucio account update` takes a single positional parameter, which is the selector.

lib/rucio/client/commands/account.py
```diff
@@ -0,0 +1,240 @@
...
+            "$ rucio account update --account jdoe --ban True  # Ban jdoe",
+            "$ rucio account history --account root  # Show all the usage history for the account root",
```
dchristidis, line 53: See https://github.com/voetberg/rucio/pull/1/files#r1824493441.

lib/rucio/client/commands/account.py
```diff
@@ -0,0 +1,240 @@
...
+        ]
+
+    def list_namespace(self, parser: "ArgumentParser") -> None:
+        parser.add_argument("--type", dest="account_type", default="USER", help="Account Type", choices={"USER", "GROUP", "SERVICE"})
```
dchristidis: The default listing should be all accounts, not just user accounts.

lib/rucio/client/commands/account.py
```diff
@@ -0,0 +1,240 @@
...
+        parser.add_argument("-a", "--account", dest="account", help="Account name")
+        parser.add_argument("--id", dest="identity", action="store", help="Identity (e.g. DN)")
+        parser.add_argument("--filters", dest="filters", action="store", help="Filter arguments in form `key=value,another_key=next_value`")
+
+    def add_namespace(self, parser: "ArgumentParser") -> None:
+        parser.add_argument("--type", dest="accounttype", default="USER", help="Account Type (USER, GROUP, SERVICE)")
```
dchristidis: I would urge against having a default at this time. Since we don’t provide `rucio account update --type=…`, an operator can make a mistake that can’t be fixed through the CLI.

lib/rucio/client/commands/account.py
```diff
@@ -0,0 +1,240 @@
...
+        parser.add_argument("-a", "--account", dest="account", help="Account name")
+        parser.add_argument("--email", dest="accountemail", help="Add an email address associated with the account")
+
+    def show_namespace(self, parser: "ArgumentParser") -> None:
+        parser.add_argument("--type", dest="accounttype", default="USER", help="Account Type (USER, GROUP, SERVICE)")
```
dchristidis: What is `--type` supposed to be doing?

lib/rucio/client/commands/account.py
```diff
@@ -0,0 +1,240 @@
...
+            "remove": {"call": self.remove, "docs": "Delete an account.", "namespace": self.remove_namespace},
+            "history": {"call": self.history, "docs": "See historical usage for an account", "namespace": self.history_namespace},
+            "update": {"call": self.update, "docs": "Change the basic account settings.", "namespace": self.update_namespace},
+        }
+
+    def implemented_subcommands(self) -> dict[str, type[CommandBase]]:
+        return {"attribute": Attribute, "limit": Limit, "identity": Identity}
+
+    def default_operation(self):
+        return self.list_
```
dchristidis: For the record, I do not find this beneficial.

lib/rucio/client/commands/account.py
```diff
@@ -0,0 +1,240 @@
...
+        else:
+            # Temp fix to allow specific kwargs
+            self.args.key = "email"
+            self.args.value = self.args.email
+            update_account(self.args, self.client, self.logger, self.console, self.spinner)
+
+
+class Attribute(Account):
+    def module_help(self) -> str:
+        return "Add additional key/value pairs associated with an account for organizational purposes."
```
dchristidis: ‘for organizational purposes’ is incorrect. Account attributes can have functional effects, depending on the policies.

lib/rucio/client/commands/account.py
```diff
@@ -0,0 +1,240 @@
...
+    def add(self):
+        set_limits(self.args, self.client, self.logger, self.console, self.spinner)
+
+    def remove(self):
+        delete_limits(self.args, self.client, self.logger, self.console, self.spinner)
+
+
+class Identity(Account):
+    def module_help(self) -> str:
+        return "Manage identities (DNs) on an account."
```
dchristidis: ‘DNs’ only applies to X.509 identities.

lib/rucio/client/commands/account.py
```diff
@@ -0,0 +1,240 @@
...
+
+    def namespace(self, parser: "ArgumentParser") -> None:
+        parser.add_argument("--account", dest="account", action="store", help="Account name", required=True)
+        parser.add_argument("--type", dest="authtype", action="store", choices=["X509", "GSS", "USERPASS", "SSH", "SAML", "OIDC"], help="Authentication type.")
+        parser.add_argument("--id", dest="identity", action="store", help="Identity as a DNs, provide in quotes.")
```
dchristidis: Ditto.

Plus, ‘provide in quotes’ is contextual and could apply at other instances (e.g. RSE expression), but we don’t mention it because it’s part of knowing how a shell functions.

lib/rucio/client/commands/account.py
```diff
@@ -0,0 +1,240 @@
...
+        parser.add_argument("--email", dest="email", action="store", help="Email address associated with the identity")
+        parser.add_argument("--password", dest="password", action="store", help="Password if authtype is USERPASS")
+        parser.add_argument("--human", default=True, help=SUPPRESS)
+
+    def _operations(self) -> dict[str, "OperationDict"]:
+        return {
+            "list": {"call": self.list_, "docs": "Show existing DNs for an account.", "namespace": self.namespace},
```
dchristidis: Again, ‘DN’ come from the X.509 ‘distinguished name’ and is not a synonym for a Rucio identity.

lib/rucio/client/commands/account.py
```diff
@@ -0,0 +1,240 @@
...
+            "add": {"call": self.add, "docs": "Grant identity access to an account.  Format IDs as 'KEY1=Value One,KEY2=Value Two' if DNs.", "namespace": self.namespace},
```
dchristidis: Ditto. We can’t be mentioning every instance where the shell might do something different.

lib/rucio/client/commands/command.py
```diff
@@ -0,0 +1,246 @@
...
+        # Options for defining remaining OIDC parameters
+        auth_args.add_argument("--oidc-user", dest="oidc_username", help="OIDC username")
+        auth_args.add_argument("--oidc-password", help="OIDC password")
+        auth_args.add_argument(
+            "--oidc-scope",
+            default="openid profile",
+            help="Defines which (OIDC) information user will share with Rucio. "
+            + 'Rucio requires at least -sc="openid profile". To request refresh token for Rucio, scope must include "openid offline_access" and '  # NOQA: W503
+            + "there must be no active access token saved on the side of the currently used Rucio Client.",
+        )  # NOQA: W503
```
dchristidis: Side comment: there’s no need for the operators because strings are automatically concatenated.

lib/rucio/client/commands/command.py
```diff
@@ -0,0 +1,246 @@
...
+
+    Commands(logger, client, args, console, spinner)()
+
+    if cli_config == 'rich':
+        spinner.stop()
+
+    if console.is_terminal and not args.no_pager:
+        command_output = console.end_capture()
+        if command_output != '':
+            signal.signal(signal.SIGINT, signal.SIG_IGN)  # Do not allow the user to
```
dchristidis: To do what? The suspense is killing me! :laughing:

lib/rucio/client/commands/config.py
```diff
@@ -0,0 +1,61 @@
...
+            "add": {"call": self.add, "docs": "Update the existing configuration settings. WARNING: Changes the global config. Can overwrite existing settings!", "namespace": self.namespace},
+            "remove": {"call": self.remove, "docs": "Remove a setting from the configuration. WARNING: Changes the global config.", "namespace": self.namespace},
+        }
+
+    def usage_example(self) -> list[str]:
+        return [
+            " $ rucio config add --section limitsscratchdisk --option testlimit --value 30  # Change the existing limitstractdisk section",
+            " $ rucio config list --section foo # Show the settings in section foo",
+            " $ rucio config  # Show all the different sections of the config",
+            " $ rucio config remove --section testsection -- option test  # Remove the value in testsection/test",
```
dchristidis: Erroneous space.

lib/rucio/client/commands/lifetime_exception.py
```diff
@@ -0,0 +1,56 @@
...
+    from logging import Logger
+
+    from rich.console import Console
+    from rich.status import Status
+
+    from rucio.client.client import Client
+    from rucio.client.commands.utils import OperationDict
+
+
+class LifetimeException(CommandBase):
```
dchristidis, line 31: Honestly, the part of the client related to the Lifetime Model just doesn’t fit with the rest of it. I think it would be better to keep the name as it is until we can decide what to do with it (there are thoughts of removing or hiding it from the client).

lib/rucio/client/commands/replica.py
```diff
@@ -0,0 +1,174 @@
...
+    def module_help(self) -> str:
+        return "Interact with a Data IDentifier at a specific Rucio Service Element."
+
+    def _operations(self) -> dict[str, "OperationDict"]:
+        return {
+            "list": {"call": self.list_, "docs": "List the replicas of a DID and its PFNs.", "namespace": self.list_namespace},
+            "dataset": {"call": self.dataset, "docs": "List replica datasets", "namespace": self.dataset_namespace},
+            "pfn": {"call": self.pfn, "docs": "Show the pfn for replicas of a DID", "namespace": self.pfn_namespace},
+            "tombstone": {"call": self.tombstone, "docs": "Add a tombstone which will mark the replica as ready for deletion by a reaper daemon", "namespace": self.tombstone_namespace},
+        }
```
dchristidis: I’ve put a lot of effort, but I’m afraid I don’t know what to recommend here. Most, if not all, of the `rucio replica` subcommands don’t fit with the structure and are unintuitive.

lib/rucio/client/commands/replica.py
```diff
@@ -0,0 +1,174 @@
...
+    def implemented_subcommands(self) -> dict[str, type[CommandBase]]:
+        return {}
+
+    def _operations(self) -> dict[str, "OperationDict"]:
+        return {
+            "suspicious": {"call": self.suspicious, "docs": "Show existing replicas marked as suspicious.", "namespace": self.sus_namespace},
+            "quarantine": {"call": self.quarantine, "docs": "Quarantine replicas.", "namespace": self.quarantine_namespace},
+            "bad": {"call": self.bad, "docs": "Declare bad replicas.", "namespace": self.bad_namespace},
+            "temp-unavailable": {"call": self.temp_unavailable, "docs": "Declare temporary unavailable replicas,", "namespace": self.unavail_namespace},
+        }
```
dchristidis: Similar here. I’m concerned that this will be a poor experience.

lib/rucio/client/commands/rse.py
```diff
@@ -0,0 +1,283 @@
...
+
+
+class RSE(CommandBase):
+    def module_help(self) -> str:
+        return "Manage Rucio Storage Elements - the sites where Rucio can place and access data."
+
+    def _operations(self) -> dict[str, "OperationDict"]:
+        return {
+            "list": {"call": self.list_, "docs": "Show all RSEs.", "namespace": self.list_namespace},
+            "show": {"call": self.show, "docs": "Show history of RSE usage (date created, date updated, etc)", "namespace": self.namespace},
```
dchristidis: The content of the parentheses is not helpful.

lib/rucio/client/commands/rse.py
```diff
@@ -0,0 +1,283 @@
...
+
+    def usage_example(self) -> list[str]:
+        return [
+            "$ rucio rse  # Show all current RSEs, can also access with `rucio rse list`",
+            "$ rucio rse list --rse 'deterministic=True' # Show all RSEs that match the RSE Expression 'deterministic=True'",
+            "$ rucio rse remove --rse RemoveThisRSE  # Disable an RSE by name",
+            "$ rucio rse add --rse CreateANewRSE  # add a new RSE named CreateANewRSE",
+            "$ rucio rse update --rse rse123456 --setting deterministic --value False  # Make an RSE Non-Deterministic",
+            "$ rucio rse show --rse rse123456  # See all the settings for rse123456"
+        ]
```
dchristidis: This is very confusing. `--rse` is an RSE expression for `list`, but an RSE name in the others.

Having kept `--rses` for `list` and passing the RSE as a positional argument for the others would have been more intuitive.

lib/rucio/client/commands/rse.py
```diff
@@ -0,0 +1,283 @@
...
+            "set": {"call": self.set_, "docs": "Update the distance between a pair of RSE that already have a distance between them."},
+        }
+
+    def usage_example(self) -> list[str]:
+        return [
+            "$ rucio rse distance list --rse rse1 --destination rse2  # View the existing distance between rse1 and rse2",
+            "$ rucio rse distance remove --rse rse1 --destination rse2  # Remove an existing distance",
+            "$ rucio rse distance add --rse rse1 --destination rse2 --distance 10  # Add the distance between two rses that do not already have a distance",
+            "$ rucio rse distance set --rse rse1 --destination rse2 --distance 20  # Update an existing distance",
+        ]
```
dchristidis: `--source` feels like a more appropriate name to be paired with `--destination`.

lib/rucio/client/commands/rule.py
```diff
@@ -0,0 +1,128 @@
...
+
+    def _operations(self) -> dict[str, "OperationDict"]:
+        return {
+            "add": {"call": self.add, "docs": "Add replication rule to define number of replicas at sites.", "namespace": self.add_namespace},
+            "remove": {"call": self.remove, "docs": "Delete a replication rule. Replicas created by the rule will not be impacted unless specified to do so.", "namespace": self.remove_namespace},
+            "show": {"call": self.show, "docs": "Retrieve information about a specific rule.", "namespace": self.show_namespace},
+            "history": {"call": self.history, "docs": "List history of different replica rules impacting a DID", "namespace": self._common_namespace},
+            "update": {"call": self.update, "docs": "Update replication rule, can be used to move a rule from one RSE to another.", "namespace": self.update_namespace},
+            "list": {"call": self.list_, "docs": "List replication rules.", "namespace": self.list_namespace},
+        }
```
dchristidis: These descriptions did not come from the original client, right? ‘Replicas created by the rule will not be impacted unless specified to do so’ is very misleading. I’m concerned we might need to review all the help content.

lib/rucio/client/commands/rule.py
```diff
@@ -0,0 +1,128 @@
...
+
+    def default_operation(self):
+        return self.list_
+
+    def usage_example(self) -> list[str]:
+        return [
+            "$ rucio rule add -d user.jdoe:test_did --copies 2 --rse SPAINSITES  # Create a rule that requires two copies of a did limited to Spanish Cites.",
```
dchristidis: Sites not Cites.

But again, `--rse` causes confusion.

lib/rucio/client/commands/rule.py
```diff
@@ -0,0 +1,128 @@
...
+            "$ rucio rule list --did user.jdoe:test_did  # show rules impacting a DID",
+            "$ rucio rule list --rule-id rule123456  # View a summary for an existing rule",
+            "$ rucio rule show --rule-id rule123456  # View a detailed overview for an existing rule.",
```
dchristidis: I don’t understand see the need for `rucio rule list --rule-id`. Either one needs a list of rules for a DID, or to show the details of a specific rule with its ID.

lib/rucio/client/commands/rule.py
```diff
@@ -0,0 +1,128 @@
...
+            "$ rucio rule update --rule-id rule123456 --suspend  # Suspend the execution of a rule",
+            "$ rucio rule update --rule-id rule123456 --move --rse NewRSE # Copy an existing rule to a new RSE",
+        ]
+
+    def _common_namespace(self, parser: "ArgumentParser") -> None:
+        parser.add_argument("-a", "--account", dest="rule_account", action="store", help="The account of the rule")
+        parser.add_argument("--activity", action="store", help="Activity to be used (e.g. User, Data Consolidation)")  # TODO More info on this
+        parser.add_argument("--rule-id", help="The rule ID, for accessing an existing rule.")
+        parser.add_argument("--lifetime", dest="lifetime", action="store", type=int, help="Rule lifetime (in seconds)")
+        parser.add_argument("--locked", action="store_true", help="Set the rule to locked - [WHAT IS THE CONSEQUENCE?]")  # TODO More info on this
```
dchristidis: TODOs are acceptable as comments, but not in the help description.

lib/rucio/client/commands/did.py
```diff
@@ -0,0 +1,291 @@
...
+
+    def _operations(self) -> dict[str, "OperationDict"]:
+        return {
+            "list": {"call": self.list_, "docs": "List the Data IDentifiers matching certain pattern. Only collection type DIDs are returned by default, use --filter 'type=all' to return all.", "namespace": self.list_namespace},
+            "show": {"call": self.show, "docs": "List attributes and statuses about data identifiers.", "namespace": self.show_namespace},
+            "add": {"call": self.add, "docs": "Create a new collection type data identifier", "namespace": self.add_namespace},
+            "remove": {"call": self.remove, "docs": "Delete a DID. Can be recovered for up to 24 hours after deletion.", "namespace": self.remove_namespace},
+            "update": {"call": self.update, "docs": "Touch one or more DIDs and set the last accessed date to the current date.", "namespace": self.update_namespace},
+            "upload": {"call": self.upload, "docs": "Upload a DID", "namespace": self.upload_namespace},
+            "download": {"call": self.download, "docs": "Download a DID", "namespace": self.download_namespace},
```
dchristidis: As previously mentioned, I think I would rather see `upload` and `download` as their own subcommands instead of part of `did`. The motivation is how they don’t fit with the structure and their sheer complexity.

dchristidis (commented):
I like the promise of the new structure and I am genuinely looking forward to testing a client with it. Kudos to Maggie for the work, both in terms of quantity and quality.

But I have concerns, effectively too many to comment on all of them. Obviously, all this is a personal opinion. I think, roughly, 80 % of the new structure is fantastic, 10 % is fine, but the remaining 10 % is a potentially painful regression in terms of usability. The underlying implementation and organisation is, on the whole, logical, easy to follow, and good for future development.

Consequently, I am unable to approve as it is (but neither am I recommending to postpone it).

lib/rucio/client/commands/rule.py
voetberg: Ah I meant to remove that, thanks for catching that.

lib/rucio/client/commands/rule.py
voetberg: This is an artifact of the existing `rucio list-rules` which does take Rule ID as an arg. We can hide it to avoid confusion though

lib/rucio/client/commands/replica.py
voetberg: I am also less than satisfied with these, but leaving them as is I think is an even worse experience (having 4 obviously different commands would mess with the concept of this PR in general). I would really love suggestions on how to improve this.

lib/rucio/client/commands/account.py
voetberg: All of these account-type argument defaults are from the old version. We can remove them.

lib/rucio/client/commands/command.py
voetberg: Yet another unfortunate victim of copy/paste 😔

lib/rucio/client/commands/command.py
voetberg: I think someone started talking to me while I was in the middle of writing that comment and forgot to go back and fix it lmao

lib/rucio/client/commands/rse.py
voetberg: I also think this is confusing. There is a number of places where both RSE and RSE Expressions are allowed. How best do we handle this? This confusion here is symptomatic of that.

dchristidis: > both RSE and RSE Expressions

An RSE is a unitary RSE expression. So the question becomes not when both are allowed, but only the former (i.e. the argument is an RSE name and not an RSE expression).

This was originally reported in #1839. The solution was to use `--rse` when a single RSE name is expected and `--rses` when it is an RSE expression.

lib/rucio/client/commands/rule.py
dchristidis: And, of course, it doesn’t work!

```console
$ rucio list-rules --id ad4f96dcc05d4f42a9c4b59827d67eee
2024-12-10 15:14:48,110	ERROR	
The object is missing this property: 'bytes'
This should never happen. Please rerun the last command with the "-v" option to gather more information.
```

I suggest getting rid of it. My be preferable to do so in a separate commit that that the difference in interface is not lost together with all the other changes.

lib/rucio/client/commands/account.py
voetberg: Because `list_account_usage` shows the same information `get_limits` just with the current utilization, to keep `account history` un-confusing, it is removed and `list_account_usage` is remapped to `account limit list`

lib/rucio/client/commands/rse.py
voetberg: I'm afraid I'm very unsatisfied with that solution. There's no reason that `rses` would imply it's anything other than a list of rse names. Perhaps the more restrictive version (e.g. only allows a single explicit rse name) should just be `--rse-name` and the generic rse expression or explicit rse name should be `--rse-expression` in all cases?

lib/rucio/client/commands/replica.py
dchristidis: Let’s go back to our basic building blocks:

* `add`
* `list`
* `remove`
* `show`
* `update`

And let’s focus on `list`: this is the one that we *have* to make work because it is a frequently-used command. `add` and `remove` we likely won’t implement (such functions exist, but it’s highly unusual to do so directly from the CLI). `show` and `update` are useful, but not as critical as `list`.

We have to implement two commands: `list-file-replicas` and `list-dataset-replicas`. It doesn’t help that the latter has a lot of complex options because it also returns the PFNs.

*I* wouldn’t mind if the output depended on the type of the DID: if a collection you get `list-dataset-replicas` and if a file you get `list-file-replicas`. But we would have to add some sort of option (e.g. `--files` or `--recursive`) so that, given a collection, you can get its file replicas.

I think this will require some unfortunate plumbing. But how acceptable would it be as an interface?

lib/rucio/client/commands/account.py
dchristidis: In the latest changes, doesn’t it have to be marked as required?

lib/rucio/client/commands/rse.py
mlassnig: Have to agree with @voetberg here. One of my biggest gripes with the current CLI is the inconsistency of `--rse`, `--rses` and `--rse-expression`.

There's two cases: one where we definitely only want to interact with a single RSE, in that case positional makes sense IMHO. (I could be convinced for `--rse` if you really want.) For everything else this is always a set based operation, so here we need to merge `--rses` and `--rse-expression` into a single mindful variant. Since this also includes RSE attributes, I would rather tend towards `--rse-expression` but I find that a bit clunky. How about `--rse-filter`?

lib/rucio/client/commands/account.py
voetberg: 🤦 Yeah

lib/rucio/client/commands/did.py
mlassnig: For the docs, I think that should be plural? (Should always work with multiple DIDs, for both upload and download?)

lib/rucio/client/commands/rse.py
mlassnig: Since I noticed it here. In the docs, there's a `.` at the end. We should make it consistent. Either always `.` or never `.` (my preference).

lib/rucio/client/commands/account.py
dchristidis: Forgive me, I still don’t understand this. `info_account` in the original `rucio-admin` does not take the account type into consideration (and I don’t see how it would make sense to do so).

lib/rucio/client/commands/account.py
dchristidis: Some of the `usage_example()` methods will need to be adapted as well.

lib/rucio/client/commands/account.py
dchristidis: Now there’s a double space, sorry.

lib/rucio/client/commands/account.py
voetberg: I got most of them in that commit but I did just append in one from RSE and Subs

lib/rucio/client/commands/account.py
dchristidis: I’m afraid this remains confusing. It doesn’t help help that there’s no term for it (the ‘identity’ of an identity). But for `USERPASS`, for example, it is a user name. For `GSS` it’s the principal.

Let’s ignore this for now, but make a note for the future.

lib/rucio/client/commands/account.py
dchristidis: Replace with ‘Show existing *identities* for an account.’ for now. But we should avoid using the verb ‘show’ for the subcommand `list`.

lib/rucio/client/commands/replica.py
voetberg: Okay I have a proposal (That may be okay but may be deeply annoying, depending on the royal "your" preference.) 

Again returning to building blocks. 
We only have `list` and `add` as general options (e.g. listing suspicious replicas or declaring it as suspicious, quarantines, or bad)

So for `add` we get possible states: 

`rucio replica add state {bad|unavailable|quarantine}` where the type of state is a locked choice positional arg. I am also amicable to calling it `replica update state {...}`

And for list we only get
`rucio replica list state {suspicious}` where hopefully suspicious can be built out into different options in the future.

lib/rucio/client/commands/rse.py
dchristidis: I would have gone with `--rse-expression` as well, but that’s what was decided the last time. I think the idea was that, wherever you can put multiple RSEs, you do so exclusively via an RSE expression.

So the problem here is that you can either follow that decision or propose something that is universally-accepted as superior. And the argument against `--rse-expression` is its length.

I have to say, I find the documentation provided in the original `rucio list-dataset-replicas` on-point. Hence, I’ve grown accustomed to `--rses`.

lib/rucio/client/commands/replica.py
voetberg: If the output depends on the type of DID, it is possible for a user to get the right answer to the wrong question without knowing it. I have a strong distaste for these sort of contextual commands for that reason. 

Would `rucio replica list {file|dataset}` be an acceptable compromise? 

I think `rucio replica pfn` can probably be removed in favor of `rucio replica [solution to the list problem] --pfn`, and `rucio replica tombstone` can be swapped with `remove` (with the note that it's controlled by a reaper daemon) though.

lib/rucio/client/commands/rse.py
voetberg: I'll see if I can run a quick search to remove all those trailing `.`'s, it's also my preference to not include them

lib/rucio/client/commands/rse.py
dchristidis: > One of my biggest gripes with the current CLI is the inconsistency of --rse, --rses and --rse-expression.

The current CLI *is* consistent. There’s no `--rse-expression` anymore.

lib/rucio/client/commands/replica.py
dchristidis: > Would `rucio replica list {file|dataset}` be an acceptable compromise?

I think I like this.

lib/rucio/client/commands/rse.py
mlassnig: ha! just shows how often i've been using it!

so... `--rses` for the set-based selection, and for single RSEs positional and/or `--rse` ?

dchristidis: Yes, that is correct. If there is some command that behaves differently, then that would qualify as a bug.

voetberg: > > One of my biggest gripes with the current CLI is the inconsistency of --rse, --rses and --rse-expression.
> 
> The current CLI _is_ consistent. There’s no `--rse-expression` anymore.

There are instances of --expression or `rse_expression` as a positional arg (`list-suspicious-replicas`, `add-replication-rule`).

I still think `rse` and `rses` is not sufficient in explaining the differences between the two, so many pluralized nouns are just stylistic or inconsistent (in may code bases, not just ours), it's strange to make people assume that this is _the_ instance where the `s` matters. But if this is the solutions those brave programmers of 1839 decided, I will defer to that decision.

voetberg: I should note, you can have multiple kwargs matching to a variable. We have it in `declare-temporary-unavialable-replica`, albeit for a depreciating argument. 
```
unavailable_parser.add_argument("--expiration-date", "--duration", new_option_string="--duration", dest="duration", action=StoreAndDeprecateWarningAction, type=int, help="Timeout in seconds when the replicas will become available again")

```

voetberg: For now, and because I'm stubborn about this, I put in the two argument version. 

RSE Name Only
` parser.add_argument("--rse", "--rse-name", dest="rse", help="...")`

RSE Expression
` parser.add_argument("--rses", "--rse-exp", dest="rses", help="..")`

Which look like

```
  --rse RSE, --rse-name RSE
                        Rucio Storage Element (RSE) name

  --rses RSES, --rse-exp RSES
                        The RSE expression

```

dchristidis: > There are instances of `--expression` or `rse_expression` as a positional arg (`list-suspicious-replicas`, `add-replication-rule`).

I’m afraid I stand by my comment. `list-suspicious-replicas` was introduced in parallel with the effort to unify the interface, hence it was missed. But I’m willing to bet my life savings that nobody is using this command. In `add-replication-rule` it is the description of an argument that matches the technical term. The only area where the interface is not consistent is some `rucio-admin` commands that offer it either as a keyword argument (e.g. `rucio-admin rse set-attribute` or as a positional argument (e.g. `rucio-admin rse add-protocol`).

I’m not deaf to your arguments; I would have chosen `--rse-expression` myself. But that was the proposal which was approved and accepted (recently, I might add) and the commands for the daemons follow it. Hence, we’re dealing with 1) a proposal that is not unequivocally superior to the current implementation and 2) change aversion.

lib/rucio/client/commands/rse.py
```diff
@@ -0,0 +1,268 @@
...
+)
+from rucio.client.commands.command_base import CommandBase
+
+if TYPE_CHECKING:
+    from argparse import ArgumentParser
+
+    from rucio.client.commands.utils import OperationDict
+
+
+class RSE(CommandBase):
```
dchristidis, line 46: While aware that naming things is hard, I’m not too enthusiastic about calling this class `RSE`. We have an `RSE` SQLAlchemy model and an `RSE` Flask blueprint.

Then we have `Limit` below but there’s another `Limit` in `account.py`.

I’m requesting that they be changed, just trying to raise awareness about something that could prove suboptimal in retrospect.

bari12:
Merging!
@voetberg please note down some of the missing change requests. I will create a ticket where we can track these and also mention it in the release notes.
Thanks a lot for this fantastic work, it's outstanding! 👏

Issue #7076, opened 2024-09-04
Remove all mutable global variables


rdimaio:
We have mutable global variables in a few places, e.g.:

https://github.com/rucio/rucio/blob/5d42f2c54bc60a213701bb6b77bf810835a280b4/lib/rucio/common/utils.py#L251-L260

Allowing to mutate the global state is a bit of an anti-pattern, and it makes it harder to properly test these functions.

We should consider refactoring mutable global variables and make them either immutable or non-global (depending on what the context is), and explicitly passing/injecting the value into functions that use those variables.

rdimaio:
Some relevant links:
- https://softwareengineering.stackexchange.com/a/148109
- https://stackoverflow.com/questions/59330578/how-to-avoid-using-global-variables
- https://stackoverflow.com/questions/19158339/why-are-global-variables-evil?noredirect=1&lq=1

PR #7120, opened 2024-10-04
Core: refactor config to cache the parser instead of using a global variable

rdimaio:
#7076 (though it's not a mutable global variable in this case)

### DIFF
--- a/lib/rucio/common/config.py
+++ b/lib/rucio/common/config.py
@@ -17,6 +17,7 @@
 import configparser
 import json
 import os
+from functools import cache
 from typing import TYPE_CHECKING, Any, Optional, TypeVar, Union, overload
 
 from rucio.common import exception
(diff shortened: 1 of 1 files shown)


voetberg:
I think your branch name has gotten mixed up (though I'm not sure if that makes it into the release notes tbh). 

Do we have any tests that test things have actually been cached? Because if codecov is complaining, it may be good enough reason to add one.

bari12:
Ping on this :-)

rdimaio:
> I think your branch name has gotten mixed up (though I'm not sure if that makes it into the release notes tbh).

Thanks for noticing - I had begun this branch working on another issue, but then the changes I was making for that issue converged into this PR instead. For cleanliness, I will close this PR and re-open another one from a branch with a better name

> Do we have any tests that test things have actually been cached? Because if codecov is complaining, it may be good enough reason to add one.

I've added them in the new PR here https://github.com/rucio/rucio/pull/7315

PR #7123, opened 2024-10-11
Clients: list sorted replicas in correct order #6141

jamesp-epcc:
This PR fixes the bug where the Rucio client loses the order of the sorted replicas returned from the server. It also allows the geo IP replica sorter to be configured so that an exception is raised if the geo IP database cannot be loaded properly for any reason. (The current behaviour is to silently ignore errors, resulting in replicas being sorted randomly instead). If we wanted to make this the default setting, we would need to make further changes as currently this breaks the integration tests. These are the only two bugs I am currently aware of relating to the replica sorting code.

### DIFF
--- a/bin/rucio
+++ b/bin/rucio
@@ -412,7 +412,7 @@ def list_file_replicas(args):
         pfn_dir, dst_dir = args.link.split(':')
         if args.rses:
             for replica, rse in itertools.product(replicas, rses):
-                if rse in list(replica['rses'].keys()) and replica['rses'][rse]:
+                if replica['rses'].get(rse):
                     for pfn in replica['rses'][rse]:
                         os.symlink(dst_dir + pfn.rsplit(pfn_dir)[-1], replica['name'])
         else:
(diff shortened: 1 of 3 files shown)


voetberg:
:+1: lgtm! 

Before I approve, can you also add a line to the documentation [here](https://github.com/rucio/documentation/blob/main/docs/operator/configuration_parameters.md) for the geoip error option?

jamesp-epcc:
Good point. I've added a PR for the documentation here: https://github.com/rucio/documentation/pull/376

voetberg (commented):
👍

bin/rucio
```diff
@@ -423,35 +423,36 @@ def list_file_replicas(args):
                             os.symlink(dst_dir + pfn.rsplit(pfn_dir)[-1], replica['name'])
     elif args.pfns:
         if args.rses:
-            for replica, rse in itertools.product(replicas, rses):
-                if rse in list(replica['rses'].keys()) and replica['rses'][rse]:
-                    for pfn in replica['rses'][rse]:
+            for replica in replicas:
+                for pfn in replica['pfns']:
+                    rse = replica['pfns'][pfn]['rse']
+                    if rse in list(replica['rses'].keys()) and replica['rses'][rse]:
```
```suggestion
                    if replica['rses'].get(rse):
```

jamesp-epcc: Thanks, that's an improvement

rdimaio (approved):

jamesp-epcc:
Do we think this could make it into the next Rucio release? I know there are some in the UK who are very keen to get GeoIP working better.

voetberg:
> Do we think this could make it into the next Rucio release? I know there are some in the UK who are very keen to get GeoIP working better.

@bari12 Can we try to make this into 36?

bari12:
Done! :-)

PR #7141, opened 2024-10-25
Dependencies: Remove virtualenv in favour of built-in venv; fix #7140

### LINKED ISSUE #7140: Unneeded dev dependency `virtualenv`
We define `virtualenv` in the dev requirements but it is not used in the code itself https://github.com/search?q=repo%3Arucio%2Frucio%20virtualenv&type=code

The only use I could find is when setting up Rucio clients here, which still instructs you to install `virtualenv` separately: https://github.com/rucio/documentation/blob/ef17e7d79e4a538aba80554fc0b949776f07f371/docs/user/setting_up_the_rucio_client.md?plain=1#L27

Python 3+ ships with `venv` built-in (https://stackoverflow.com/questions/41573587/what-is-the-difference-between-venv-pyvenv-pyenv-virtualenv-virtualenvwrappe), so I think `virtualenv` might be a leftover from Python 2 support.

Discussed on Mattermost, ok to remove it in favour of built-in `venv`

rdimaio:

### DIFF
--- a/requirements/requirements.dev.in
+++ b/requirements/requirements.dev.in
@@ -8,7 +8,6 @@ flake8==6.1.0                                               # Wrapper around PyF
 pycodestyle==2.11.0                                         # New package replacing pep8; python_version < '3.9'
 pylint==3.0.2
 astroid==3.0.1
-virtualenv==20.26.3                                         # Virtual Python Environment builder
 xmltodict==0.13.0                                           # Makes working with XML feel like you are working with JSON
 pytz==2023.3.post1                                          # World timezone definitions, modern and historical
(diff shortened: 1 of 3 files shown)


voetberg (approved):
Martins comment in mattermost (for reference): 

> I think we can remove it. I think the idea of the dev dependencies was to have everything in one place for a dev environment (based on virtualenv) - but it's not really the case anymore 

All looks good

rdimaio:
> Martins comment in mattermost (for reference):
> 
> > I think we can remove it. I think the idea of the dev dependencies was to have everything in one place for a dev environment (based on virtualenv) - but it's not really the case anymore
> 
> All looks good

Thanks for providing the quote, good idea!

PR #7154, opened 2024-11-01
Policies: Refactor policy package version checks to use specifier sets

### LINKED ISSUE #7082: Policy package version support: specifying each supported major version (current way of doing things) vs specifying a minimum valid version (proposal)
## Description

This is the way that Rucio currently checks that a policy package supports the current Rucio version:

https://github.com/rucio/rucio/blob/5d42f2c54bc60a213701bb6b77bf810835a280b4/lib/rucio/common/utils.py#L2115-L2119

This only works if the policy package's `SUPPORTED_VERSION` list specifies a major version. This is in line with the [policy package docs](https://rucio.github.io/documentation/operator/policy_packages/):

> `__init__.py` should include a `SUPPORTED_VERSION` field indicating the major version(s) of Rucio that your package supports. This is checked against the Rucio server version to ensure compatibility when loading the policy package.

The code parses the current version like this:

```
>>> components = 1
>>> versions = "35.0.0"
>>> ".".join(version.split(".")[:components])
'35'
```

and then just checks that the string `'35'` exists in `SUPPORTED_VERSION`. 

However, the latest version of the ATLAS policy package depends on `Rucio >= 35.1.0`:

https://gitlab.cern.ch/atlas-adc-ddm/rucio-policy-package/-/blob/master/src/atlas_rucio_policy_package/__init__.py?ref_type=heads#L1

This is because it depends on a fix that was added in Rucio `35.1.0.`

## Consideration
Right now, we specify each supported major version. With regards to the ATLAS policy package, this would mean that we should set the supported Rucio version to `36`. However, this:
1. is not completely correct (36 is not the only supported Rucio version for the policy package, since it supports some minor versions of Rucio 35, just not 35 as a whole)
2. delays deployment of the policy package to the next major release, even though it is already compatible with the current minor release

I think this sparks the discussion of whether it would be better to specify a minimum valid version instead. 

Considering that policy packages might depend on fixes and patches introduced in minor Rucio releases, I would propose to instead:
1. allow the operator to specify a minimum supported Rucio version in the policy package
2. allow the operator to optionally specify unsupported versions (this is akin to how Python requirements are specified, e.g. `pkg_name>2.0.0, !=2.2.0` allows the installation of any `pkg_name` version higher than `2.0.0` except `2.2.0`)
3. In the Rucio code, check `rucio_version not in UNSUPPORTED_VERSIONS` and `rucio_version >= minimum_supported_version`). To do this second comparison, we should use `packaging.version.Version` (https://stackoverflow.com/a/11887885)

### LINKED ISSUE #7083: Policy package version check: instead of parsing string manually, use `packaging.version.Version`
change should be done here https://github.com/rucio/rucio/blob/5d42f2c54bc60a213701bb6b77bf810835a280b4/lib/rucio/common/utils.py#L2115-L2119

according to https://stackoverflow.com/a/11887885

Note: this should be done regardless of what we decide to do here: https://github.com/rucio/rucio/issues/7082

rdimaio:
- fixes #7082 
- fixes #7083

### DIFF
--- a/lib/rucio/common/exception.py
+++ b/lib/rucio/common/exception.py
@@ -1068,14 +1068,14 @@ class PolicyPackageVersionError(PolicyPackageBaseException):
     """
     Policy package is not compatible with this version of Rucio.
     """
-    def __init__(self, package: str, rucio_version: str, supported_versions: list[str], *args):
+    def __init__(self, package: str, rucio_version: str, supported_versionset: str, *args):
         super(PolicyPackageVersionError, self).__init__(package, *args)
         self.rucio_version = rucio_version
-        self.supported_versions = supported_versions
+        self.supported_versionset = supported_versionset
         self._message = 'Policy package %s is not compatible with this Rucio version.\nRucio version: %s\nVersions supported by the package: %s' % (
             self.package,
             self.rucio_version,
-            self.supported_versions
+            self.supported_versionset
         )
         self.error_code = 103
 
--- a/lib/rucio/common/plugins.py
+++ b/lib/rucio/common/plugins.py
@@ -18,6 +18,8 @@ import os
 from configparser import NoOptionError, NoSectionError
 from typing import TYPE_CHECKING, Any, TypeVar
 
+from packaging.specifiers import SpecifierSet
+
 from rucio.common import config
 from rucio.common.client import get_client_vo, is_client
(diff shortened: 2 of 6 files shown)


voetberg (approved):
:+1:

jamesp-epcc (approved):

bari12 (approved):

PR #7156, opened 2024-11-06
Metadata: add UnsupportedMetadataPlugin exception handling Fix #7155

### LINKED ISSUE #7155: Improper exception handling for unsupported metadata plugins leads to generic error messages
### Description

When requesting metadata using an unsupported/disabled plugin (e.g., 'STRUCTURED_JSON'), the server correctly identifies the error but the exception handling chain fails to properly propagate the specific error to the client. This results in a generic RucioException instead of a specific, actionable error message.

The core issues identified:
1. Missing specific exception class (UnsupportedMetadataPlugin) for handling unsupported metadata plugin scenarios in both client and server
2. Server raises a NotImplementedError (a Python built-in exception) instead of a Rucio-specific exception
3. Client doesn't recognise properly the exception and falls back to generic RucioException that does not indicate the actual problem.

Current behavior:
```python
RucioException: An unknown exception occurred.
Details: no error information passed (http status code: 500)
```

Enhanced behavior:
```python
UnsupportedMetadataPlugin: The requested metadata plugin is not enabled on the server.
Details: Metadata plugin "STRUCTURED_JSON" is not enabled on the server.
```

The fix requires:
- Define UnsupportedMetadataPlugin exception in rucio.common.exception
- Update server to properly raise this exception
- (Ensure also that the client has matching exception defined)

### Steps to reproduce

1) Setup a fresh Rucio client installation
2) Try to get metadata using an unsupported plugin:

```python
from rucio.client.didclient import DIDClient

client_config = {
    'rucio_host': "..",
    'auth_host': "..",
    'auth_type': "userpass",
    'creds': {"username": "..", "password": ".."},
    'account': ".."
}

try:
    cl = DIDClient(**client_config)
    result = cl.get_metadata(scope='scope', name='name', plugin='STRUCTURED_JSON')
except Exception as _e:
    print(f"An error of type '{type(_e).__name__}' occurred while connecting to Rucio server: {_e}\n")
```

3) Observe generic RucioException instead of specific plugin error

### Rucio Version

Server: 35.0.0

### Additional Information

_No response_

Geogouz:
This PR addresses the issue #7155. When a `get_metadata` request is made now using an unsupported or disabled plugin, the server will raise a specific UnsupportedMetadataPlugin exception.

### DIFF
--- a/lib/rucio/common/exception.py
+++ b/lib/rucio/common/exception.py
@@ -1139,3 +1139,13 @@ class PolicyPackageIsNotVersioned(PolicyPackageBaseException):
         super(PolicyPackageIsNotVersioned, self).__init__(package, *args)
         self._message = 'Policy package %s does not include information about which Rucio versions it supports.' % self.package
         self.error_code = 109
+
+
+class UnsupportedMetadataPlugin(RucioException):
+    """
+    Raised when attempting to use a metadata plugin that is not enabled on the server.
+    """
+    def __init__(self, *args):
+        super(UnsupportedMetadataPlugin, self).__init__(*args)
+        self._message = "The requested metadata plugin is not enabled on the server."
+        self.error_code = 110
--- a/lib/rucio/core/did_meta_plugins/__init__.py
+++ b/lib/rucio/core/did_meta_plugins/__init__.py
@@ -84,8 +84,9 @@ def get_metadata(scope, name, plugin="DID_COLUMN", *, session: "Session"):
     :param scope: The scope of the did.
     :param name: The data identifier name.
     :param plugin: (optional) Filter specific metadata plugins.
-    :returns: List of metadata for did.
-    :raises: NotImplementedError
+    :param session: (optional) The database session in use.
+    :returns: Dictionary containing metadata for did.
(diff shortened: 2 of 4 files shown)


lib/rucio/common/exception.py
```diff
@@ -1139,3 +1139,13 @@ def __init__(self, package: str, *args):
         super(PolicyPackageIsNotVersioned, self).__init__(package, *args)
         self._message = 'Policy package %s does not include information about which Rucio versions it supports.' % self.package
         self.error_code = 109
+
+
+class UnsupportedMetadataPlugin(RucioException):
+    """
+    RucioException
```
rdimaio: Could you expand this docstring with more specific information to this exception?

rdimaio (dismissed):
LGTM - worth adding tests to `test_did_meta_plugins` to verify that this exception is raised in the expected circumstances (and not raised otherwise), but since `NotImplementedError` was not being tested either, I think it's fine to do this in a separate PR at some other point

lib/rucio/common/exception.py
Geogouz: yep, added

rdimaio (approved):

PR #7158, opened 2024-11-07
REST & API: Ensure proper Access-Control-Request-Headers handling in CORS middleware

### LINKED ISSUE #7148: Server, Auth: Patch CORS middleware to include requested headers in `Access-Control-Allow-Header` in the server's response
### Description

In order for x509 auth to work in Firefox, we need to explicitly include certificates in the webui code where requests are made to the `/auth/x509/webui` endpoint.

In doing so, we can no longer return `Access-Control-Allow-Headers` as `*` as stated in the firefox docs for their implementation of the browser fetch functionality.

https://github.com/rucio/rucio/blob/81c0d4712fdd3b5277aece71cd22190b961c3139/lib/rucio/web/rest/flaskapi/v1/common.py#L62-L76

### Motivation

This change is required to allow x509 authentication using the new WebUI in Firefox.

### Change

We will explicitly pass the `Access-Control-Request-Headers` from the WebUI and the code referenced above needs to be adapter to forward the supplied value to the `Access-Control-Allow-Headers` in the response.

MytsV:
Fix #7148 

Ensuring proper x509 authentication in the new version of the UI across all browsers requires concrete `Access-Control-Allow-Headers` instead of a wildcard (*). If the client certificate is explicitly included in a cross-origin request, the asterisk gets [treated as a literal character](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Headers#directives). Therefore, in accordance with https://github.com/rucio/rucio/blob/3f8b419609f7b776cff61267fcf6323054bf260c/lib/rucio/web/rest/flaskapi/v1/auth.py#L64
the CORS middleware should pipe the request headers from `Access-Control-Request-Headers`.

### DIFF
--- a/lib/rucio/web/rest/flaskapi/v1/common.py
+++ b/lib/rucio/web/rest/flaskapi/v1/common.py
@@ -69,7 +69,7 @@ class CORSMiddleware:
                 response: Response = Response(status=200)
                 response.headers['Access-Control-Allow-Origin'] = request.origin
                 response.headers['Access-Control-Allow-Methods'] = 'GET, POST, PUT, DELETE, OPTIONS'
-                response.headers['Access-Control-Allow-Headers'] = '*'
+                response.headers['Access-Control-Allow-Headers'] = request.environ.get('HTTP_ACCESS_CONTROL_REQUEST_HEADERS')  # type: ignore (value could be None)
                 response.headers['Access-Control-Allow-Credentials'] = 'true'
                 return response(environ, start_response)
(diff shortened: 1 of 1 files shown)


maany:
Looks good @MytsV! Could you check the Pyright report on the CI and send a fix :)

maany (approved):

maany:
@bari12 this is required for fixing the x509 issue on firefox :)

Issue #6842, opened 2024-06-21
"diff"-based policies, instead of fully replicating the base policy package code


rdimaio:
Follow-up to this comment: https://github.com/rucio/rucio/issues/5138#issuecomment-2162353861, in particular:

> My main feedback is that it seems very hard to maintain a policy package that only does simple changes like allowing / in lfns with respect to updates in the default schema / authorization policies.
The proposed way is to initially copy ~2000 lines of very repetitive python code.
It seems very hard to maintain changes that were made intentionally vs. changes that are just outdated versions of the generic scheme. It's also easy to miss new functions defined in newer rucio versions.
A good way to be able to only modify parts that need to be changed would be great.

maxnoe:
One more thing:

Mention that the policy package is required basically everywhere (I.e. also custom images of the daemons, init and clients are needed, not just server). The current version contains this sentence:

> The policy package code must be available to the Rucio server (and possibly other components such as daemons).

but we found that it is really needed, not just "possibly".


The description of some items was also not clear. The importance of `SCOPE_NAME_REGEXP` was not at all clear and we could only figure that out with help from @cserf

jamesp-epcc:
I think it would be quite easy to allow policy packages that don't include the schema or the permissions module if they don't need it, and Rucio could fall back to the generic module in those cases. It would be even better if we could allow those two modules to only override the parts that they actually need to change. That would be a bit more work but I think it should be possible.

maxnoe:
Could we re-open this issue? If I am not mistaken, it is still basically impossible, e.g. to only override the `NAME` schema.

Looking at the PR, the only granularity implemented at the moment is individually overriding `SCOPE_NAME_REGEX` and `validate_schema`. I tried to implement only overriding verifying `NAME`, but this doesn't work due to the nature of json schema.

The low-level objects are included in the higher-level ones, so I ended up again with the full policy package as before.

Maybe a possibility would be to not define the schemas at import time, but provide a schema "builder" with the individual components needed and then building the more complex schemas from there

maxnoe:
See https://github.com/maxnoe/rucio-dirac-dev/pull/2

rdimaio:
Good point, thanks for pointing it out. Reopening it for discussion.

> Maybe a possibility would be to not define the schemas at import time, but provide a schema "builder" with the individual components needed and then building the more complex schemas from there

[...]

PR #7160, opened 2024-11-11
Policies: allow diff-based policy packages #6842

jamesp-epcc:
This PR allows policy packages to omit the permission and/or schema modules if these do not require customisation. If they are not present, the built in generic modules will be used instead. It also allows the schema module to define only a subset of the schema values, with any missing ones being taken from the generic module.

### DIFF
--- a/lib/rucio/common/schema/__init__.py
+++ b/lib/rucio/common/schema/__init__.py
@@ -12,7 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import functools
 import importlib
+import logging
 from configparser import NoOptionError, NoSectionError
 from os import environ
 from typing import TYPE_CHECKING, Any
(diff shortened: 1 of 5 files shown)


rdimaio (commented):
LGTM, just two comments:
1. would it make sense to allow people to define a subset of `permission`, similarly to how it's done in `schema`?

I think this would be possible by checking if `permission_modules[issuer.vo].has_permission(issuer, action, kwargs, session=session)` returns `None` here https://github.com/rucio/rucio/blob/3fd6a7aae889d9aaa9469a2c270eb8faa11a477f/lib/rucio/core/permission/__init__.py#L157C14-L157C99 - if it returns `None`, that particular permission is missing from that VO's policy and we can fall back to the generic package to check.

2. Would it be possible to write tests to check the fallback logic?

jamesp-epcc:
1. That's a good idea. I hadn't thought of doing it that way. I'll definitely implement that
2. I'll think about how we can test this. I don't think this part of the code is really tested at present

jamesp-epcc:
I've implemented the fallback to the generic permission checks. Once I've added some tests I'll rebase and squash again

lib/rucio/common/schema/__init__.py
```diff
@@ -61,15 +64,35 @@
...
+        POLICY = 'rucio.common.schema.' + GENERIC_FALLBACK.lower()
+        generic_module = importlib.import_module(POLICY)
+    except ModuleNotFoundError:
+        raise exception.PolicyPackageNotFound(POLICY)
+    except ImportError:
+        raise exception.ErrorLoadingPolicyPackage(POLICY)
 
 
 def load_schema_for_vo(vo: str) -> None:
+    global generic_module
```
rdimaio: Instead of defining a `global` variable to hold the `generic_module`, I would be more in favour of having a separate function that returns the `generic_module` and decorating it with `@functools.cache`, for testability.

tests/test_policy_package.py
```diff
@@ -0,0 +1,60 @@
...
+        rucio.common.schema.schema_modules['def'] = importlib.import_module('tests.mocks.schema_diff')
+
+        # check that overriden value is as expected
+        assert rucio.common.schema.get_schema_value('SCOPE_LENGTH') == 50
+
+        # check that omitted value falls back to generic module
+        assert rucio.common.schema.get_schema_value('NAME_LENGTH') == 250
+        
+        # restore original schema module
+        rucio.common.schema.schema_modules['def'] = old_module
```
rdimaio: there's some existing policy package unit tests in `tests/rucio/common/test_plugins.py` - these tests could be moved there

lib/rucio/common/schema/__init__.py
jamesp-epcc: Done

tests/test_policy_package.py
jamesp-epcc: Also done. Didn't notice those in there!

rdimaio: Looks like these tests rely on finding the config file (likely due to importing some module that imports the config), so they don't work when moved to be part of the unit tests, which do not use any config file at the moment. Sorry about this, I should have checked before asking you to move them here. Could you please move them back?

jamesp-epcc: Yes, the schema initialisation uses the config file. I hadn't realised there was a distinction between unit tests that wouldn't have the config file available and the other tests. I will move these back.

rdimaio (approved):

PR #7163, opened 2024-11-12
Do not set platform for images that support many archictures (#7162)

### LINKED ISSUE #7162: Allow arm64 images in docker-compose.yml where supported
The file `etc/docker/dev/docker-compose.yml` sets all images to linux/amd64. This is unnecessary. Many of the images support multiple architectures and Docker will pull the right one. Only rucio and a few older images are not compiled for arm64. 
https://github.com/rucio/rucio/blob/3f8b419609f7b776cff61267fcf6323054bf260c/etc/docker/dev/docker-compose.yml#L60

Removing platform from most images will let them run natively on Apple M processors.

suranap:
Only a few images are limited to amd64 images. Most images have multiple architectures. When running on Apple M, Docker can choose the appropriate arm64 images.


Geogouz:
Hi, I would suggest to keep the platform definition for the images that do not support (at least yet) multiple architectures as you also suggest in https://github.com/rucio/rucio/issues/7162#issuecomment-2554830091. This removes the need for tweaking the environment on Mac, etc.

bari12:
@suranap any comment?

suranap:
If you guys are cool with this change, please merge it in. I'm not great at following up, hence the 3 week delay in replying. :-) The failing check is a version change in setup.py. Maybe I can rebase on master...

rdimaio (approved):

Issue #7118, opened 2024-10-03
Remove the `deep` option for `list_dataset_replicas`


dchristidis:
### Description

This has been brought up a few times and was mentioned again during the Rucio Community Workshop. ATLAS has been running a patch that ignores the client and forces `deep=True` for all requests, for over a year. No performance issue has been identified.

The goal is to force this option to be always enabled, effectively removing the need for its existence.

### Motivation

Reduces the need for collection replicas. Provides more accurate results.

### Change

Remove* the `deep` parameter from all layers, from the CLI all the way to the core. Use the code path that goes down to the level of file replicas instead of relying on collection replicas.

\* Need to take backwards-compatibility into consideration.

dchristidis:
As previously mentioned, ATLAS has been running a patch to force the `--deep` option for all incoming requests. This morning we encountered an incident with it for the second time.

For individual users, this has worked very well. But for a WFMS like PanDA, when working on large datasets, it can cause a high load on the database and affect the Rucio server response time. Consequently, it would appear that removing this option is not safe.

We may want to suggest something else, like introducing an option that is the opposite of `--deep`.

PR #7170, opened 2024-11-15
Simple patch to get rid of non-deep mode

ericvaandering:


bari12:
Closing this since the test fails. @0xquark will follow it up in #7118

PR #7174, opened 2024-11-19
Draft: Avoid multiple reaper workers working on the same replicas

### LINKED ISSUE #6512: Prevent multiple Reaper threads from working on the same replicas
### Description

### Motivation

### Change

labkode:
WIP
Fixes https://github.com/rucio/rucio/issues/6512

### DIFF
--- a/lib/rucio/core/replica.py
+++ b/lib/rucio/core/replica.py
@@ -2359,7 +2359,6 @@ def get_replica(rse_id, scope, name, *, session: "Session"):
     except NoResultFound:
         raise exception.ReplicaNotFound("No row found for scope: %s name: %s rse: %s" % (scope, name, get_rse_name(rse_id=rse_id, session=session)))
 
-
 @transactional_session
 def list_and_mark_unlocked_replicas(limit, bytes_=None, rse_id=None, delay_seconds=600, only_delete_obsolete=False, *, session: "Session"):
     """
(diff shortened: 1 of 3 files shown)


lib/rucio/core/replica.py
```diff
@@ -4156,3 +4155,74 @@ def get_RSEcoverage_of_dataset(scope, name, *, session: "Session"):
             result[rse_id] = total
 
     return result
+
+@transactional_session
+def refresh_replicas(rse_id=None, replicas=None, *, session: "Session"):
```
rdimaio: For the type annotations failure, there are missing type annotations here:

```suggestion
def refresh_replicas(
        rse_id: Optional[str] = None, 
        replicas: Optional[Iterable[dict[str, Any]]] = None, 
        *, 
        session: "Session"
) -> bool:
```

labkode:
Thanks @rdimaio , that was very helpful, I was wondering what type was "replica".

lib/rucio/core/replica.py
```diff
@@ -4156,3 +4155,79 @@ def get_RSEcoverage_of_dataset(scope, name, *, session: "Session"):
             result[rse_id] = total
 
     return result
+
+@transactional_session
+def refresh_replicas(
+        rse_id: Optional[str] = None, 
+        replicas: Optional[Iterable[dict[str, Any]]] = None,
```
rdimaio:
```suggestion
        replicas: Optional['Iterable[dict[str, Any]]'] = None,
```

PR #7176, opened 2024-11-19
REST & API: update_recursive_handling_for_scope_endpoint Fix #7175

### LINKED ISSUE #7175: Align recursive Parameter Handling and Documentation in Scope API Endpoint
### Description

I noticed that the docstring for the Scope class suggests that the recursive parameter in the `GET` request body is expected as part of the JSON payload. However:
- *Current Behavior*: The get method does not use a request body at all. Instead, it looks for recursive as a query parameter in the URL (`request.args`).
- *Expected Behavior*: The docstring should reflect the actual implementation, where recursive is passed as a query parameter and not part of the request body.

https://github.com/rucio/rucio/blob/6122bb972209d0349bc78ea15fa3ee1501f3ab95/lib/rucio/web/rest/flaskapi/v1/dids.py#L94-L105

https://github.com/rucio/rucio/blob/6122bb972209d0349bc78ea15fa3ee1501f3ab95/lib/rucio/web/rest/flaskapi/v1/dids.py#L146

At the same time, the logic for handling the recursive query parameter in the get method is a bit ambiguous and might not align with user expectations:

- `?recursive` or `?recursive=true` correctly enables recursion.
- `?recursive=false` also enables recursion, which is unexpected.

**Proposed Fix**
Replace the incorrect requestBody section with a proper description of query parameters (recursive and name), e.g.:

```python
        - name: name
          in: query
          description: The name of the data identifier (did).
          required: false
          schema:
            type: string
        - name: recursive
          in: query
          description: If true, retrieves child identifiers recursively for non-file types.
          required: false
          schema:
            type: boolean
```
and update the recursive parameter logic into something like:

`recursive = request.args.get('recursive', 'false').lower() in ['true', '1']`

### Steps to reproduce

Try a request using `?recursive=false`.

### Rucio Version

35.6.0

### Additional Information

_No response_

Geogouz:
This commit updates the `Scope` API endpoint to properly handle the recursive query parameter. The logic now evaluates the parameter’s value (`true`/`1` for enabling recursion) instead of checking its mere presence. The docstring has been also corrected to reflect the actual implementation, specifying that recursive and name are query parameters, not part of a JSON request body.

### DIFF
--- a/lib/rucio/web/rest/flaskapi/v1/dids.py
+++ b/lib/rucio/web/rest/flaskapi/v1/dids.py
@@ -88,28 +88,29 @@ class Scope(ErrorHandlingMethodView):
         - name: scope
           in: path
           description: The scope.
+          required: true
           schema:
             type: string
           style: simple
-        requestBody:
-          content:
-            application/json:
-              schema:
-                type: object
-                properties:
-                  name:
-                    description: The name of the did.
-                    type: string
-                  recursive:
-                    description: If specified, also returns the child ids recursively.
-                    type: boolean
+        - name: name
+          in: query
+          description: The name of the data identifier (did).
+          required: false
+          schema:
+            type: string
+        - name: recursive
+          in: query
+          description: If true, retrieves child identifiers recursively for non-file types.
+          required: false
+          schema:
+            type: boolean
         responses:
           200:
             description: OK
             content:
               application/x-json-stream:
                 schema:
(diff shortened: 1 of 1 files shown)


rdimaio (approved):
LGTM - to make the query clearer from the user's point of view, I'd be in favour of renaming `recursive` to something like `return_child_ids`

PR #7195, opened 2024-11-29
Replicas: Remove deprecated replica sorting algorithms

### LINKED ISSUE #6863: Replica sorting algorithms described in documentation are not implemented
The documentation for [`list_replicas` ](https://rucio.github.io/documentation/client_api/replicaclient/#list_replicas) describes three sorting algorithms that have not been implemented yet:

https://github.com/rucio/rucio/blob/0bbc069faf17f06eff1891aa90e634adea67b216/lib/rucio/core/replica_sorter.py#L340-L367

Opening this issue as I could not find an open issue tracking this work.

I think it would be also be worth discussing whether we actually need different algorithms for sorting replicas, or if the current one suffices.

- [ ] `closeness`
- [ ] `ranking`
- [ ] `dynamic`

Some extra notes:

## `closeness`
Replicas sorted by AGIS closeness. I assume that AGIS here stands for the ATLAS Grid Information System (https://twiki.cern.ch/twiki/bin/view/AtlasComputing/AtlasGridInformationSystem), in which case it seems to be ATLAS-specific information. 

We should probably not have this in the Rucio core package (unless we can generalize it in some way, which I doubt) - it should be part of the ATLAS policy package instead.

## `dynamic`
"Rucio Dynamic Smart Sort (tm)". What does this mean? The docstring says "Return a list of replicas sorted by dynamic network metrics.", but I'm not sure which dynamic network metrics this would involve. 

I could not find issues/PRs about this, so if someone has more context on what this is meant to do it would be welcome.

rdimaio:
Fix #6863

### DIFF
--- a/lib/rucio/client/downloadclient.py
+++ b/lib/rucio/client/downloadclient.py
@@ -325,8 +325,6 @@ class DownloadClient:
         :param deactivate_file_download_exceptions: Boolean, if file download exceptions shouldn't be raised
         :param sort: Select best replica by replica sorting algorithm. Available algorithms:
             ``geoip``       - based on src/dst IP topographical distance
-            ``closeness``   - based on src/dst closeness
-            ``dynamic``     - Rucio Dynamic Smart Sort (tm)
 
         :returns: a list of dictionaries with an entry for each file, containing the input options, the did, and the clientState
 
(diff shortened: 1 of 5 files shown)


voetberg (approved):
Failed test seems unrelated, lgtm 👍

PR #7199, opened 2024-12-02
Daemons: reaper, avoid multiple reaper workers working on the same re…

### LINKED ISSUE #6512: Prevent multiple Reaper threads from working on the same replicas

labkode:
Fixes #6512


tests/test_replica.py
```diff
@@ -288,6 +304,31 @@ def test_delete_replicas_from_datasets_old(self, rse_factory, mock_scope, root_a
...
 
+    def test_refresh_replicas(self, rse_factory, mock_scope, root_account):
+        """REPLICA (CORE): Refresh replicas"""
+        _, rse_id = rse_factory.make_mock_rse()
+        nbfiles = 5
+
+        # add 5 replicas
+        files = [{'scope': mock_scope, 'name': did_name_generator('file'), 'bytes': 1, 'adler32': '0cc737eb', 'meta': {'events': 10}} for _ in range(nbfiles)]
+        add_replicas(rse_id=rse_id, files=files, account=root_account, ignore_availability=True)
+        # adding a repilca will set updated_at time to unix epoch, so we expect is set
```
rdimaio, line 315:
```suggestion
        # adding a replica will set updated_at time to unix epoch, so we expect is set
```

tests/test_replica.py
```diff
@@ -288,6 +304,31 @@ def test_delete_replicas_from_datasets_old(self, rse_factory, mock_scope, root_a
...
+        # we update the state to BEING_DELETED
+        update_replica_state(rse_id, r['scope'], r['name'], ReplicaState.BEING_DELETED)
+        original = get_replica_updated_at(r)
+
+        time.sleep(2)
+        ok = refresh_replicas(rse_id, [r])
+        assert ok is True
+
+        got = get_replica_updated_at(r)
+        assert got > original
```
rdimaio: nit: I'd rename some of these variables to have clearer names:
`r` -> `replica_dict`
`ok` -> maybe just `assert refresh_replicas(rse_id, [r])`
`original` -> `original_timestamp`
`got` -> `updated_timestamp`

lib/rucio/daemons/reaper/reaper.py
```diff
@@ -141,8 +141,13 @@ def delete_from_storage(heartbeat_handler, hb_payload, replicas, prot, rse_info,
     rse_id = rse_info['id']
     noaccess_attempts = 0
     pfns_to_bulk_delete = []
+
+    refresh_start_time = time.time()
+    trigger_time = 60 * 8  # 8 minutes, we could make it configurable, but will change functions signature.
+
     try:
         prot.connect()
+        num_replicas_processed = 0  # counts how many replicas has been already being processed
```
rdimaio, line 150:
```suggestion
        num_replicas_processed = 0  # counts how many replicas have already been processed
```

lib/rucio/daemons/reaper/reaper.py
```diff
@@ -141,8 +141,13 @@ def delete_from_storage(heartbeat_handler, hb_payload, replicas, prot, rse_info,
     rse_id = rse_info['id']
     noaccess_attempts = 0
     pfns_to_bulk_delete = []
+
+    refresh_start_time = time.time()
```
rdimaio: By having this before `prot.connect()`, the `elapsed_time` for the first replica will always be longer than all other replicas in the `for replica in replicas` loop - it should be moved under `for replica in replicas:`

lib/rucio/daemons/reaper/reaper.py
```diff
@@ -217,6 +222,26 @@ def delete_from_storage(heartbeat_handler, hb_payload, replicas, prot, rse_info,
                 deletion_dict['duration'] = duration
                 add_message('deletion-failed', deletion_dict)
 
+            finally:
+                # this control loop will run indefinetely until all deletions have gone through (or failed).
```
rdimaio:
```suggestion
                # this control loop will run indefinitely until all deletions have gone through (or failed).
```

lib/rucio/daemons/reaper/reaper.py
```diff
@@ -217,6 +222,26 @@ def delete_from_storage(heartbeat_handler, hb_payload, replicas, prot, rse_info,
...
+                # it assumes that for each deletion a timeout will occur (which will fail the deletion).
+                # if that assumption is not true, we need to introduce a maximum retry counter to avoid the worker hanging
+                # on individual deletions.
+                num_replicas_processed += 1
+
+                # after each replica is deleted we evaluate how much time we have left to delete.
+                # if we are not able to delete all the replicas we got in, other workers will take the replicas because their update time will
+                # be more than 10 minutes (refer to function list_and_mark_unlocked_replicas).
+                # After 8 minutes, and if we still have some replicas to process, we bump the replicas updated_at field to current time (with will delay the selectability
```
rdimaio:
```suggestion
                # After 8 minutes, and if we still have some replicas to process, we bump the replicas updated_at field to current time (which will delay the selectability
```

lib/rucio/daemons/reaper/reaper.py
```diff
@@ -217,6 +222,26 @@ def delete_from_storage(heartbeat_handler, hb_payload, replicas, prot, rse_info,
...
+                # by other workers by 8+ minutes).
```
rdimaio: I'd rephrase it like this for clarity:

```suggestion
                # Every 8 minutes, if there are still replicas to process,
                # we bump the updated_at field of the unprocessed replicas to current time,
                # which will prevent other workers from selecting
                # the unprocessed replicas for the next 8+ minutes
```

lib/rucio/daemons/reaper/reaper.py
```diff
@@ -217,6 +222,26 @@ def delete_from_storage(heartbeat_handler, hb_payload, replicas, prot, rse_info,
...
+                elapsed_time = time.time() - refresh_start_time
+                if elapsed_time > trigger_time:  # 8 minutes have passed
+                    to_refresh = replicas[num_replicas_processed + 1]  # only refresh the replicas we have not yet deleted
+                    ok = refresh_replicas(rse_id=rse_id, replicas=to_refresh)
+                    if not ok:
+                        logger(logging.WARNING, "Failed to bump updated_at for replicas BEING_DELETED")
+                    refresh_start_time = time.time()  # reset it so we can trigger new refresh cycles.
```
rdimaio: This check could also be done at the beginning of the loop, so that there's no need to even check if there still are replicas to process - as it's done at the beginning of the loop, we know that there's at least one more replica to process.

lib/rucio/core/replica.py
```diff
@@ -4472,3 +4496,84 @@ def get_RSEcoverage_of_dataset(
...
+        session: "Session"
+) -> bool:
+    """
+    Updates the updated_at timestamp of the given replicas but don't wait if row is locked.
+
+    :param rse_id: the RSE containing the replicas to refresh.
+    :param replicas: a list with replicas (dictionary with the information of the affected replica).
+    :param session: The database session in use.
+
+    :returns: True, if successful, False otherwise.
```
rdimaio: What returns `True` and `False` in this function feels a bit arbitrary - e.g. if the input is missing, `True` is returned, but it doesn't feel right since no refresh action has been taken.

labkode: I only want the function to fail in case there is an error while executing the query as I don't consider a failure if there aren't rows to update. 

I will update the docs to specify that in case there are no rows to update, the function succeed, to make it clear.

tests/test_replica.py
rdimaio: unresolving - this is still there

rdimaio (approved):
LGTM, just two small threads

lib/rucio/daemons/reaper/reaper.py
```diff
@@ -141,9 +141,26 @@ def delete_from_storage(heartbeat_handler, hb_payload, replicas, prot, rse_info,
...
+        refresh_start_time = time.time()
+        num_replicas_processed = 0  # counts how many replicas has been already being processed
         for replica in replicas:
+            # Every 8 minutes, if there are still replicas to process,
+            # we bump the updated_at field of the unprocessed replicas to current time,
+            # which will prevent other workers from selecting
+            # the unprocessed replicas for the next 8+ minutes
+            elapsed_time = time.time() - refresh_start_time
+            if elapsed_time > trigger_time:  # 8 minutes have passed
+                to_refresh = replicas[num_replicas_processed + 1]  # only refresh the replicas we have not yet deleted
```
dchristidis, line 158: This is incorrect. Imagine a setting where at the 8-minute mark only half the replicas have been physically deleted. With this implementation, only the remaining half will be touched. As such, at the 10-minute mark, the first half may be picked up again by another Reaper thread.

lib/rucio/core/replica.py
```diff
@@ -402,7 +402,7 @@ def __declare_bad_file_replicas(
                     if f"{pfn} Unknown replica" not in unknown_replicas:
                         unknown_replicas.append('%s %s' % (pfn, 'Unknown replica'))
             elif scope or name:
-                unknown_replicas.append(f"{(scope,name)} Unknown replica")
+                unknown_replicas.append(f"{(scope, name)} Unknown replica")
```
dchristidis, line 405: I applaud and actively encourage aesthetic changes and refactoring, but they should be put together on a separate commit (so that a `git blame` explains that said commit has no behavioural changes). This applies to other parts of the PR too.

lib/rucio/daemons/reaper/reaper.py
```diff
@@ -141,9 +141,26 @@ def delete_from_storage(heartbeat_handler, hb_payload, replicas, prot, rse_info,
     rse_id = rse_info['id']
     noaccess_attempts = 0
     pfns_to_bulk_delete = []
+
+    trigger_time = 60 * 8  # 8 minutes, we could make it configurable, but will change functions signature.
+
```
dchristidis, line 146: It’s a core function; I don’t see any problems with changing the signature. If a community uses the `--delay-seconds` command-line argument to set it to five minutes, for example, then your patch becomes ineffective.

lib/rucio/daemons/reaper/reaper.py
```diff
@@ -217,6 +234,13 @@ def delete_from_storage(heartbeat_handler, hb_payload, replicas, prot, rse_info,
                 deletion_dict['duration'] = duration
                 add_message('deletion-failed', deletion_dict)
 
+            finally:
+                # this control loop will run indefinitely until all deletions have gone through (or failed).
+                # it assumes that for each deletion a timeout will occur (which will fail the deletion).
+                # if that assumption is not true, we need to introduce a maximum retry counter to avoid the worker hanging
+                # on individual deletions.
```
dchristidis, line 241: Please follow [PEP 8](https://peps.python.org/pep-0008/#comments):

1. The first word of a sentence should be capitalised.
2. Each sentence should end with a period.

This applies to other parts of the PR too.

lib/rucio/daemons/reaper/reaper.py
```diff
@@ -444,7 +468,7 @@ def run_once(
         auto_exclude_threshold: int,
         auto_exclude_timeout: int,
         heartbeat_handler: "HeartbeatHandler",
-        **_kwargs
+        **_kwargs: Optional["dict[str, Any]"]
```
dchristidis, line 471: This isn’t in line with [PEP 484](https://peps.python.org/pep-0484/#arbitrary-argument-lists-and-default-argument-values).

lib/rucio/core/replica.py
```diff
@@ -3195,6 +3195,30 @@ def get_replica_atime(
     return session.execute(stmt).scalar_one()
 
 
+@read_session
+def get_replica_updated_at(replica: dict[str, Any], *, session: "Session") -> Optional[datetime]:
+    """
+    Get the updated_at timestamp for a replica. Just for testing.
+    :param replica: list of dictionary {scope, name, rse_id}
```
dchristidis, line 3202: Not a list.

lib/rucio/core/replica.py
```diff
@@ -4472,3 +4496,84 @@ def get_RSEcoverage_of_dataset(
...
     return result
+
+
+@transactional_session
+def refresh_replicas(
+        rse_id: Optional[str] = None,
+        replicas: Optional['Iterable[dict[str, Any]]'] = None,
+        *,
+        session: "Session"
+) -> bool:
```
dchristidis, line 4507: I would have preferred to see the last part of `list_and_mark_unlocked_replicas()` factored out into a separate function. Is that not feasible?

lib/rucio/core/replica.py
```diff
@@ -3195,6 +3195,30 @@ def get_replica_atime(
     return session.execute(stmt).scalar_one()
 
 
+@read_session
+def get_replica_updated_at(replica: dict[str, Any], *, session: "Session") -> Optional[datetime]:
```
dchristidis, line 3199: `scalar_one()` raises an exception if the query does not match exactly one row. As such, the use of `Optional` is wrong here.

lib/rucio/core/replica.py
```diff
@@ -3195,6 +3195,30 @@ def get_replica_atime(
...
+
+    :returns: A datetime timestamp with the updated time.
+    """
+
+    stmt = select(
+        models.RSEFileAssociation.updated_at
+    ).with_hint(
+        models.RSEFileAssociation,
+        'INDEX(REPLICAS REPLICAS_PK)',
+        'oracle'
```
dchristidis, line 3213: I don’t think we need a hint here.

dchristidis (changes requested):

lib/rucio/daemons/reaper/reaper.py
labkode: Right, after checking the code, I understood this issue. The DB removal happens in the calling function, and the reaper would eventually get those replicas (that will have no physical data anymore) and fail to delete. 

I will refactor this code.

Good catch.

labkode:
Revamped in #8048

PR #7202, opened 2024-12-03
Database: psycopg2 to psycopg migration #6669

### LINKED ISSUE #6669: `psycopg2-binary` should not be used in production - `psycopg2` should be used instead

Geogouz:
This PR upgrades the PostgreSQL driver from psycopg2 to psycopg (3). There has already been some work to tackle this (see #6993) but since I am following a quite different approach and the integration of those commits would yield a messy branch, I am starting a new one here and we can continue the discussion here.

A few important remarks:

I have added functionality to handle and parse requirements with “extras” (e.g., psycopg[binary,pool]).

While all tests are currently passing, we should remain cautious. Since `Psycopg` serves as the underlying driver for all `PostgreSQL` connections (used both indirectly through `SQLAlchemy` and directly in some parts of the code), there might still be edge cases that our tests have not detected.

For instance, I had to explicitly do casts of the port value at endpoints to an integer because `Psycopg` enforces stricter type handling compared to `Psycopg2`. The following method:

https://github.com/rucio/rucio/blob/594959f33b7d91892bee4d2147c7bbe1ec8b4dde/lib/rucio/core/rse.py#L1584-L1591

even though expects an optional integer for port, our tests (and possibly even production usage) are passing a string without raising an error in other contexts.

Additionally, the `postgres_meta.py` module directly interacts with Psycopg (bypassing SQLAlchemy and its models). I had to update the SQL statements to align with Psycopg’s stricter requirements. These changes call maybe for more thorough testing to ensure that all functionality remains intact and behaves as expected for this plugin.

### DIFF
--- a/lib/rucio/core/did_meta_plugins/postgres_meta.py
+++ b/lib/rucio/core/did_meta_plugins/postgres_meta.py
@@ -16,8 +16,9 @@ import json
 import operator
 from typing import TYPE_CHECKING
 
-import psycopg2
-import psycopg2.extras
+import psycopg
+from psycopg import sql
+from psycopg.rows import dict_row
 
 from rucio.common import config, exception
 from rucio.common.types import InternalScope
@@ -73,17 +74,20 @@ class ExternalPostgresJSONDidMeta(DidMetaPlugin):
         self.jsonb_column = table_column_data
 
         self.table = table
-        self.client = psycopg2.connect(
+        self.client = psycopg.connect(
             host=host,
             port=port,
-            database=db,
+            dbname=db,
             user=user,
             password=password)
 
         # set search_path to include database schema by default
         cur = self.client.cursor()
-        statement = "SET search_path TO {};".format(db_schema)
-        cur.execute(statement)
+        cur.execute(
+            sql.SQL("SET search_path TO {}").format(
+                sql.Identifier(db_schema)
+            )
+        )
         cur.close()
 
         if not table_is_managed:                    # not managed by Rucio, so just verify table schema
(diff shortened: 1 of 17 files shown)


rdimaio:
> even though expects an optional integer for port, our tests (and possibly even production usage) are passing a string without raising an error in other contexts.

that's due to the `@transactional_session` decorator, which accidentally hides the function signature from linters, see https://github.com/rucio/rucio/issues/6986 for more context

robbarnsley:
Thanks @Geogouz. Apologies for taking so long to get around to this.

As far as I am aware, we (SKAO) are the only ones using the external postgres plugin & we're not running a production instance currently so the... "blast radius" resulting from any resulting issues around this will be quite limited.

That being said, I haven't seen any problems with my (admittedly cursory) checks of getting/setting metadata following these changes, and the the (limited) unit tests all pass, so, from that perspective i'm happy for this to be merged.

robbarnsley (dismissed):
LGTM!

bari12:
Please rebase and adapt commit messages to the style recommended in the https://rucio.cern.ch/documentation/contributing/

lib/rucio/core/did_meta_plugins/postgres_meta.py
```diff
@@ -97,17 +101,18 @@ def _try_create_metadata_table(self):
...
-            ("name", "varchar", "NOT NULL"),
-            ("data", "jsonb", "DEFAULT", "'{}'::jsonb"),
-            ("UNIQUE", "(scope, name)")  # unique scope+name table constraint, required for ON CONFLICT
+        table_clauses = [
+            sql.SQL("id bigint NOT NULL GENERATED ALWAYS AS IDENTITY"),
+            sql.SQL("vo varchar NOT NULL"),
+            sql.SQL("scope varchar NOT NULL"),
+            sql.SQL("name varchar NOT NULL"),
+            sql.SQL("data jsonb DEFAULT '{}'::jsonb"),
+            sql.SQL("UNIQUE (scope, name)")
```
rdimaio: I think we should keep the `# unique scope+name table constraint, required for ON CONFLICT` for clarity

Geogouz: Okie. I'll adapt also the commit messages. Pushed it for the moment to see how tests are doing.

rdimaio (approved):

bari12:
Merged as feature for 37

PR #7215, opened 2024-12-05
Daemons: Allow help menu to show even without config section set

voetberg:
Solves a documentation issue: https://github.com/rucio/documentation/issues/405

The Cache-Client daemon requires the `message-cache` section in the config to be set to use, but you can't know that without looking at the help menu. Which requires the section to be set. 

This change just sets defaults to `None` if the section is not present and adds a warning to the executable description.

### DIFF
--- a/bin/rucio-cache-client
+++ b/bin/rucio-cache-client
@@ -104,7 +104,7 @@ def get_parser():
          'operation': 'delete_replicas'
         }
 """
-    description = "This daemons is used to populate information of replicas on volatile storage."
+    description = "This daemon is used to populate information of replicas on volatile storage."
 
     if not config_has_section("message-cache"):
         defaults = {}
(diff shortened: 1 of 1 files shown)


bin/rucio-cache-client
```diff
@@ -104,17 +104,24 @@ def get_parser():
          'operation': 'delete_replicas'
         }
 """
-    oparser = argparse.ArgumentParser(description="This daemons is used to populate information of replicas on volatile storage.",
-                                      prog=os.path.basename(sys.argv[0]), add_help=True)
+    description = "This daemons is used to populate information of replicas on volatile storage."
```
rdimaio:
```suggestion
    description = "This daemon is used to populate information of replicas on volatile storage."
```

rdimaio (dismissed):

bin/rucio-cache-client
rdimaio: (this mistake was there before the PR)

rdimaio (approved):

PR #7218, opened 2024-12-05
Metadata: improve MongoDB connection parameter validation Fix #7217

### LINKED ISSUE #7217: Improve MongoDB connection parameter handling and validation
### Description

I think that the current MongoDB connection setup in `MongoDidMeta.__init__` is a bit problematic:

https://github.com/rucio/rucio/blob/3d5cbf5d8c0e37f793bbd6f01d1064e44cfe0b4f/lib/rucio/core/did_meta_plugins/mongo_meta.py#L49-L59

- Assumes valid MongoDB metadata configuration.
- Unclear password requirement (if user exists but password isn't found in config, it will fail).
- No validation that all required parameters are defined eventually before connecting.

For example, if a metadata section exists in the configuration but is set up for a different plugin, the code will still try to fetch MongoDB-specific parameters, connect and fail.

### Steps to reproduce

Run the tests using some non MongoDB metadata plugin.

### Rucio Version

36.0.0rc5

### Additional Information

_No response_

Geogouz:
This PR enhances the MongoDB connection initialization with better parameter validation, error handling, and authentication logic. See #7217

### DIFF
--- a/lib/rucio/core/did_meta_plugins/mongo_meta.py
+++ b/lib/rucio/core/did_meta_plugins/mongo_meta.py
@@ -28,10 +28,10 @@ if TYPE_CHECKING:
     from sqlalchemy.orm import Session
 
 IMMUTABLE_KEYS = [
-    '_id',              # index key
-    'scope',            # generated on insert
-    'name',             # generated on insert
-    'vo'                # generated on insert
+    '_id',  # index key
+    'scope',  # generated on insert
+    'name',  # generated on insert
+    'vo'  # generated on insert
 ]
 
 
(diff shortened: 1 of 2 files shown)


lib/rucio/core/did_meta_plugins/mongo_meta.py
```diff
@@ -46,28 +46,40 @@ def __init__(
...
+        for param in con_params:
+            if con_params[param] is None:
+                if config.config_has_option('metadata', param):
+                    con_params[param] = (
+                        config.config_get_int('metadata', param)
+                        if param == 'mongo_service_port'
+                        else config.config_get('metadata', param)
+                    )
+                else:
+                    raise exception.MissingClientParameter(f"Parameter '{param}' not found!")
```
rdimaio: is this the right exception to use? Right now we only use `MissingClientParameter` in the client

Geogouz: To be honest, it took sometime to decide on that but still, I also wasn't much convinced. Maybe `ConfigurationError` / `ConfigNotFound` (which I avoided because the configurations may be actually correct) or even a new `MissingServerParameter` exception class or something similar might be better. Dunno, open to suggestions basically.

rdimaio: Maybe `ConnectionParameterNotFound`?

rdimaio (approved):

PR #7219, opened 2024-12-08
Documentation: LifetimeClient Docstring Update; Fix rucio#7633

### LINKED ISSUE #7633: Rucio Python Client docstrings enhancement (`LifetimeClient`)
Sub-issue of https://github.com/rucio/documentation/issues/363 targeting specifically the `LifetimeClient`.

Geogouz:
Handles the sub-issue about the LifetimeClient Docstrings

### DIFF
--- a/lib/rucio/client/lifetimeclient.py
+++ b/lib/rucio/client/lifetimeclient.py
@@ -39,10 +39,32 @@ class LifetimeClient(BaseClient):
             states: Optional['Sequence[LifetimeExceptionsState]'] = None
     ) -> 'Iterator[dict[str, Any]]':
         """
-        List exceptions to Lifetime Model.
+        Lists lifetime model exceptions that allow extending data lifetimes beyond their configured policies.
 
-        :param id:         The id of the exception
-        :param states:     The states to filter
+        The lifetime model exceptions are used to override the default lifecycle policies for data identifiers
+        (files, datasets, containers, or archives) that need to be kept longer than usual. These exceptions
+        can be filtered by their ID or approval state (this feature is not available yet).
+
+        :param exception_id: The unique identifier of a specific exception. If provided, returns only that exception.
+        :param states: Filter exceptions by their states. Possible values are:
+                      - `A` (APPROVED): Exception was approved
+                      - `R` (REJECTED): Exception was rejected
+                      - `W` (WAITING): Exception is waiting for approval by an admin (or other authorized account)
+
+        :returns:
+            An iterator of dictionaries containing the exception details:
+             - `id`: The unique identifier of the exception
+             - `scope`: The scope of the data identifier
+             - `name`: The name of the data identifier
+             - `did_type`: Type of the data identifier:
+                    `F` (file), `D` (dataset), `C` (container), `A` (archive),
+                    `X` (deleted file), `Y` (deleted dataset), `Z` (deleted container)
+             - `account`: The account that requested the exception
+             - `pattern`: Pattern used for matching data identifiers
+             - `comments`: User provided comments explaining the exception
+             - `state`: Current state of the exception
+             - `created_at`: When the exception was created (returned as timestamp string)
+             - `expires_at`: When the exception expires (returned as timestamp string)
         """
 
         path = self.LIFETIME_BASEURL + '/'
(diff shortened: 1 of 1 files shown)


lib/rucio/client/lifetimeclient.py
```diff
@@ -39,10 +39,38 @@ def list_exceptions(
...
+             - `scope`: The scope associated with the data identifier
+             - `name`: The name of the data identifier
+             - `did_type`: Type of the data identifier:
+                    `F` (file), `D` (dataset), `C` (container), `A` (archive),
+                    `X` (deleted file), `Y` (deleted dataset), `Z` (deleted container)
+             - `account`: The account that requested the exception
+             - `pattern`: Pattern used for matching data identifiers
+             - `comments`: User provided comments explaining the exception
+             - `state`: Current state of the exception
+             - `created_at`: When the exception was created
```
voetberg: Worth stating that if created_at and expires_at are str or datetime obj's.

lib/rucio/client/lifetimeclient.py
```diff
@@ -39,10 +39,38 @@ def list_exceptions(
...
+
+        The lifetime model exceptions are used to override the default lifecycle policies for data identifiers
+        (files, datasets, containers, or archives) that need to be kept longer than usual. These exceptions
+        can be filtered by their ID or approval state.
+
+        :param exception_id: The unique identifier of a specific exception. If provided, returns only that exception.
+        :param states: Filter exceptions by their states. Possible values are:
+                      - `A` (APPROVED): Exception was approved
+                      - `R` (REJECTED): Exception was rejected
+                      - `W` (WAITING): Exception is waiting for approval
```
voetberg: Waiting approval from whom (what sort of admin approves them)? Useful information for troubleshooting.

lib/rucio/client/lifetimeclient.py
```diff
@@ -70,16 +98,35 @@ def add_exception(
...
+        maximum extension periods. The request includes details about which DIDs should have extended
+        lifetimes, who is requesting it, and why it's needed.
+
+        :param dids: List of dictionaries containing the data identifiers to be excepted.
+                    Each dictionary must contain:
+                    - `scope`: The scope name
+                    - `name`: The DID name
+        :param account: The account requesting the exception
+        :param pattern: Associated pattern for the exception request
+        :param comments: Justification for why the exception is needed
```
voetberg: Can you add an example of the sort of comments used here? (I think in the CLI it's listed as stuff like "needed for my analysis")

voetberg:
> First things first, though: @voetberg, may I ask if the following example aligns with the intended direction for docstring enhancements?  

Yes, this is pretty much exactly what I was thinking of when I proposed client docstring enhancements! Thank you. I left a view minor clarification questions but if you don't know the answers to those we can consult component leads.

lib/rucio/client/lifetimeclient.py
```diff
@@ -70,16 +98,35 @@ def add_exception(
...
-        returns:            The id of the exception.
+        Creates a lifetime model exception request to extend the expiration date of data identifiers (DIDs).
+
+        These exceptions allow requesting extensions to DIDs' lifetimes, subject to approval and configured
+        maximum extension periods. The request includes details about which DIDs should have extended
+        lifetimes, who is requesting it, and why it's needed.
+
+        :param dids: List of dictionaries containing the data identifiers to be excepted.
+                    Each dictionary must contain:
+                    - `scope`: The scope name
```
rdimaio:
```suggestion
                    - `scope`: The DID scope
```

might be clearer

rdimaio (dismissed):
LGTM but it would be great if you could add Maggie's suggestions

bari12:
Ping @Geogouz

lib/rucio/client/lifetimeclient.py
```diff
@@ -39,10 +39,38 @@ def list_exceptions(
...
+             - `comments`: User provided comments explaining the exception
+             - `state`: Current state of the exception
+             - `created_at`: When the exception was created (returned as timestamp string)
+             - `expires_at`: When the exception expires (returned as timestamp string)
+
+        :raises LifetimeExceptionNotFound: If no exceptions exist or match the filters
+        :raises UnsupportedRequestedContentType: If the client requests an unsupported content type
+        :raises CannotAuthenticate: If authentication to the server fails
+        :raises ServerConnectionException: If the server cannot be reached
+        :raises RucioException: Any other failure happening in the server
```
rdimaio: I'm not sure if it's worth adding these exceptions that can be raised by any client function - I see the usefulness of adding information on the function-specific exceptions (e.g. `LifetimeExceptionNotFound`), but the other ones here will be true for any client function (with the exception of the ping client, which does not need authentication, so `CannotAuthenticate` wouldn't be raised there), so I'm not sure if it would end up being a lot of duplicated docstring info. I don't have a specific opinion on this, but worth discussing.

rdimaio (dismissed):

lib/rucio/client/lifetimeclient.py
voetberg: I do think it is probably worth having a unified definition of custom exceptions to draw from (Probably in our dev docs), but wasn't able to find anything about best practices

Geogouz: One approach (that I followed at least in the UploadClient which I will finish today) is to mention only the Exceptions that are being specifically raised by the method itself (assuming that the sub-called methods would include in their docstrings their raised exceptions). But I am open to ideas. If we go eventually this way, I can apply this to all docstrings.

voetberg: This is the better solution it seems, at least as far as minimizing repeat work 😭 I'll trust your instincts on it

bari12: I agree, let's please update this and merge the PR then.

rdimaio (approved):

PR #7263, opened 2024-12-13
Patch 7236 dispatch docs

### LINKED ISSUE #7236: Testing: Replace doc building test with build action from the documentation repo
### Description

The `tools/generate_docs.py` is out of date with how documentation is actually build (And is causing confusing failures). It should be replaces with a [dispatch](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#repository_dispatch) event that tests the build from the docs repo. 

### Motivation

_No response_

### Change

_No response_

voetberg:
* Removes the generate-docs file and corresponding 'docs' suite
* Make a dispatch action in the .github folder - it will contact the docs repo 'build-doc' action (something should be in the docs repo to look for the dispatch request)


tools/test/test.sh
```diff
@@ -68,21 +68,6 @@ elif [ "$SUITE" == "syntax" ]; then
...
-    srchome
-    export RUCIO_CLIENT_API_OUTPUT="rucio_client_api.md"
-    export RUCIO_REST_API_OUTPUT="rucio_rest_api.md"
-    tools/generate_doc.py
-    test -s $RUCIO_CLIENT_API_OUTPUT
-    test -s $RUCIO_REST_API_OUTPUT
-
-    export REST_API_DOC_FILENAME="api_doc.yaml"
-    tools/generate_rest_api_doc.py > $REST_API_DOC_FILENAME
-    tools/test/check_rest_api_documentation.sh $REST_API_DOC_FILENAME
```
rdimaio, line 84: This `tools/test/check_rest_api_documentation.sh` script could also be removed as it's not used anywhere else (or we might want to keep it in case we need to check the API docs manually, but in that case it might make sense to move it to the documentation repo)

.github/workflows/dispatch_documentation.yml
```diff
@@ -0,0 +1,19 @@
...
+
+jobs:
+  workflow_trigger:
+    runs-on: ubuntu-latest
+
+ steps:
+      - name: Checkout Code
+        uses: actions/checkout@v2
+      - name: Dispatch Event to Trigger Docs Build
+        uses: GitHub/dispatch@v2
```
rdimaio, line 15: I cannot find this action online, is it a placeholder? This seems like the equivalent event https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#repository_dispatch - there's also [workflow_dispatch](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_dispatch), but I think the first one would be the one we'd need to use here

voetberg: Yes, it's a placeholder, should have clarified!

tools/test/test.sh
voetberg: Yeah, it should probably just be a step after the REST doc generation in the docs repo. The paired PR over that will make that change

voetberg: Yeah that file just runs the npx/cli command that is run in the main build_documentation.sh in the docs repo. It'll get covered by the dispatch

Issue #7278, opened 2024-12-17
Remove faulty docs test


voetberg:
### Description

the `docs` suite in the testing suite does not match the actual generation of docs from the documentation repo (mismatch between `tools/generate_docs.py` in rucio/rucio and `tools/build_documentation.sh` in rucio/documentation. It should be removed in favor of #7236 

### Steps to reproduce

n/a

PR #7279, opened 2024-12-17
Testing: Remove doc checking tests from main repo #7278

voetberg:
Paired with https://github.com/rucio/documentation/pull/426


rdimaio:
https://github.com/rucio/rucio/pull/7263 should be merged before this PR to avoid going a period of time without checking if the docs build correctly

edit: actually, thinking about it - this test is incorrect anyway, so there's no harm in removing it now; the build test in the docs repo should be fine for now

rdimaio (approved):

PR #7293, opened 2025-01-07
Improve handling of suspended accounts #7292

### LINKED ISSUE #7292: Improve handling of suspended accounts
### Description

* The core functions that handle accounts mostly require that the account state is `ACTIVE` (e.g. it’s not possible to delete an account that is suspended).
* The error message during authentication is too generic.

### Motivation

_No response_

### Change

_No response_

dchristidis:


dchristidis:
I would like to start with a question on style: there is a query is repeated multiple times. I think it would be beneficial to reuse a core function than to implement it multiple times. Opinions?

PR #7296, opened 2025-01-07
Database: Update template to latest standards #7295

### LINKED ISSUE #7295: Database migration scripts do not pass the Ruff check
### Description

The template `script.py.mako` has not been updated to our latest standards.

### Steps to reproduce

See #7294.

### Rucio Version

_No response_

### Additional Information

_No response_

dchristidis:
I won’t be surprised if the changes to the file-level docstring are unwanted (no surrounding whitespace, ending with a period), despite adhering to PEP 257. I don’t mind reverting.

### DIFF
--- a/lib/rucio/db/sqla/migrate_repo/script.py.mako
+++ b/lib/rucio/db/sqla/migrate_repo/script.py.mako
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 # Copyright European Organization for Nuclear Research (CERN) since 2012
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -13,10 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-''' ${message} '''
+"""${message}"""    # noqa: D400, D415
 
 import sqlalchemy as sa
-
 from alembic import context
 from alembic.op import execute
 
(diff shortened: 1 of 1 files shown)


rdimaio (dismissed):

lib/rucio/db/sqla/migrate_repo/script.py.mako
```diff
@@ -13,10 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-''' ${message} '''
+"""${message}."""
```
dchristidis: Unless I am mistaken, it should be fine to remove the surrounding whitespace, see [here](https://github.com/sqlalchemy/alembic/blob/8d13dc92ddf27b95403cd73d690bc8cb94d0904b/alembic/script/base.py#L863). But maybe forcing the period outside of `message` is too much.

rdimaio: Good point. I think it might be worth removing the period and adding a `# noqa: D400, D415` comment.

For reference on D400 and D415:
- https://docs.astral.sh/ruff/rules/missing-trailing-period/
- https://docs.astral.sh/ruff/rules/missing-terminal-punctuation/

These rules are pretty similar, and there's been discussion to integrate both of them in a single rule here: https://github.com/astral-sh/ruff/issues/8469, but for now we'd probably have to ignore both of them

rdimaio (approved):

PR #7300, opened 2025-01-08
Make Rucio handle the new FTS error message coming with overwrite-when-only-on-disk feature

### LINKED ISSUE #7299: Rucio doesn't handle the new FTS error message coming with `overwrite-when-only-on-disk` feature
### Description

After FTS implemented `overwrite-when-only-on-disk` feature, they introduced new error messages. If this feature is enabled and overwrite is not enabled, FTS throws `Destination file exists and is on tape` error instead of `Destination file exists and overwrite is not enabled` when the file is tape. When the latter is thrown, rucio checks the destination file report and marks the transfer as OK if the checksum is correct. It should do the same for the new error as well.

I reckon this was missed at https://github.com/rucio/rucio/pull/6903

We've been already running the following patch to avoid this bug. 

https://github.com/dynamic-entropy/rucio/pull/1/files

### Steps to reproduce

Enable `overwrite-when-only-on-disk` feature and transfer a file to a tape where the file is correctly stored.

### Rucio Version

Server: 35.6.0

### Additional Information

@labkode @ericvaandering  FYI

haozturk:
Fixes #7299 

CMS is running this patch for a few months now

### DIFF
--- a/lib/rucio/transfertool/fts3.py
+++ b/lib/rucio/transfertool/fts3.py
@@ -686,7 +686,10 @@ class Fts3TransferStatusReport(TransferStatusReport):
         dst_type = file_metadata.get('dst_type', None)
         METRICS.counter('overwrite.check.{rsetype}.{rse}').labels(rse=file_metadata["dst_rse"], rsetype=dst_type).inc()
 
-        if 'Destination file exists and overwrite is not enabled' in (reason or ''):
+        if reason and any(msg in reason for msg in (
+        'Destination file exists and overwrite is not enabled',
+        'Destination file exists and is on tape'
+        )):
             if cls._dst_file_set_and_file_correct(request, dst_file):
                 if dst_type == 'DISK' or dst_file.get('file_on_tape'):
                     METRICS.counter('overwrite.ok.{rsetype}.{rse}').labels(rse=file_metadata["dst_rse"], rsetype=dst_type).inc()


lib/rucio/transfertool/fts3.py
```diff
@@ -686,7 +686,7 @@ def _is_recoverable_fts_overwrite_error(cls, request: dict[str, Any], reason: st
         dst_type = file_metadata.get('dst_type', None)
         METRICS.counter('overwrite.check.{rsetype}.{rse}').labels(rse=file_metadata["dst_rse"], rsetype=dst_type).inc()
 
-        if 'Destination file exists and overwrite is not enabled' in (reason or ''):
+        if 'Destination file exists and overwrite is not enabled' in (reason or '') or 'Destination file exists and is on tape' in (reason or ''):
```
rdimaio: LGTM; could also be rewritten like this, which might look a bit nicer in case we need to add more messages in the future

```suggestion
        if reason and any(msg in reason for msg in (
        'Destination file exists and overwrite is not enabled',
        'Destination file exists and is on tape'
        )):
```

But it's definitely already clear in your version, so feel free to do either option

rdimaio (dismissed):

lib/rucio/transfertool/fts3.py
haozturk: Thanks Riccardo, yours is cleaner. I've applied it. Will squash the commits now

rdimaio:
This can be closed in favour of https://github.com/rucio/rucio/pull/7081 - apologies for the duplicate work

PR #7315, opened 2025-01-14
Refactor config to cache the parser instead of using a global variable, minor test refactorings, add tests for get_config and clean_cached_config

### LINKED ISSUE #7312: Config: if config fails to load, an exception other than ConfigNotFound should be raised
Right now, `ConfigNotFound` is raised:
https://github.com/rucio/rucio/blob/d6885e191473129f7dee0954c6bdaebd279e5a3b/lib/rucio/common/config.py#L793-L796

This is not correct though - the config file was found, but it cannot load properly.

rdimaio:
#7076 (though it's not a mutable global variable in this case) and #7312

Follow-up to https://github.com/rucio/rucio/pull/7120

### DIFF
--- a/lib/rucio/common/config.py
+++ b/lib/rucio/common/config.py
@@ -17,6 +17,7 @@
 import configparser
 import json
 import os
+from functools import cache
 from typing import TYPE_CHECKING, Any, Optional, TypeVar, Union, overload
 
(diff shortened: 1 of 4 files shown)


voetberg (approved):
These are all really clean changes, looks great!

PR #7343, opened 2025-01-20
Fix webdav upload to s3

### LINKED ISSUE #7342: S3 upload fails for webdav protocol impl
### Description

An S3 endpoint is configured to use `rucio.rse.protocols.webdav.Default`

Uploads fail.

### Steps to reproduce

1. Configure an S3 RSE, with the S3 Client ID and Client Secret stored within Rucio's configuration.
2. Add `https` as protocol with implementation of `rucio.rse.protocols.webdav.Default`.
3. Attempt to create a new file, targetting the S3 RSE.

### Rucio Version

Client: 36.0.0, Server: 35.6.0

### Additional Information

There are three problems and one annoyance:
- The rucio.rse.protocols.webdav.Default protocol impl sends the (OIDC) credential to the S3 endpoint with any requests.  This causes the S3 endpoint to reject such requests.
- The `connect` method tries to "ping" the server by issuing a HEAD request against an artificial URL, one corresponding to no possible Rucio file.  This works-by-accident for most WebDAV servers, but it doesn't with S3 endpoints. The S3 endpoint responds with a 4xx status code, which Rucio client interprets as an indication the S3 server is not working.
- The Rucio client error message (in response to the HEAD "ping" request failing) is rather misleading.  This isn't a critical bug, but it was annoying.
- The client attempts to upload the file with a temporary name and then rename the file (server-side).  Although this works for WebDAV endpoints, S3 endpoints do not support renaming data object.

paulmillar:
A series of four patches that address #7342 

Closes: #7342

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -68,6 +68,7 @@ Individual contributors to the source code
 - Elena Gazzarrini <gazzarrini.elena@gmail.com>, 2022-2023
 - Maximilian Linhoff, <maximilian.linhoff@tu-dortmund.de>, 2024
 - Eric Banzuzi, <eric.banzuzi@gmail.com>, 2024
+- Paul Millar, <paul.millar@desy.de>, 2025
 
 Organisations employing contributors
 ------------------------------------
@@ -96,3 +97,4 @@ Organisations employing contributors
 - University of Chicago (USA)
 - Purdue University (USA)
 - TU Dortmund University (Germany)
+- Deutsches Elektronen-Synchrotron DESY (Germany)
--- a/lib/rucio/rse/protocols/webdav.py
+++ b/lib/rucio/rse/protocols/webdav.py
@@ -213,7 +213,11 @@ class Default(protocol.RSEProtocol):
         # "ping" to see if the server is available
         try:
             res = self.session.request('HEAD', self.path2pfn(''), verify=False, timeout=self.timeout, cert=self.cert)
-            if res.status_code != 200:
+            # REVISIT: this test checks some URL that doesn't correspond to
+            # any valid Rucio file.  Although this works for normal WebDAV
+            # endpoints, it fails for endpoints using presigned URLs.  As a
+            # work-around, accept 4xx status codes when using presigned URLs.
+            if res.status_code != 200 and not (using_presigned_urls and res.status_code < 500):
                 raise exception.ServiceUnavailable('Problem to connect %s : %s' % (self.path2pfn(''), res.text))
         except requests.exceptions.ConnectionError as error:
             raise exception.ServiceUnavailable('Problem to connect %s : %s' % (self.path2pfn(''), error))
--- a/lib/rucio/rse/protocols/webdav.py
+++ b/lib/rucio/rse/protocols/webdav.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import logging
 import os
 import sys
 from dataclasses import dataclass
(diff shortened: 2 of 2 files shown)


Geogouz (commented):
**Comments**
- If an RSE is somehow configured with `sign_url, yet still requires some other form of auth, this logic would skip all credentials. Not sure how likely this could be now or in the future, but worth noting.
- Nevertheless, this seems to enable some WebDAV-S3 functionality, and in general, it looks like prior functionality is not affected, which is good.
- So those interested in using also for WebDAV, a `NoRename` subclass, could give it a try and experiment. But for them, I think there is more to it. For example, what happens with the `exists()` method when a potentially requested object does not exist? Or with the `get_space_usage` which is DAV-specific?

 **Suggested fixes:**
- Rebase 
- Function '__init__' does not have a parameter `rse_settting` but `rse_setttings`

paulmillar:
@Geogouz,

Thanks for your comments.

You're quite correct: the code avoids sending _any_ credential if the RSE is marked as `sign_url`.

This change solves one of the problems identified in #7342; specifically, that Rucio sends the OIDC access token with the request.  It seems that (for the S3 endpoints being used in interTwin) any use of the `Authorization` header (even with a credential that the server doesn't understand) will cause the endpoint to reject the PUT request.  For this specific case (OIDC-based AuthN), we have to avoid sending any credential in the `Authorization` header.

You're also quite correct in pointing out that, should the endpoint requires additional authentication then with this patch, such requests would fail.  However, I think this is rather unlikely.  First, I've not heard of any S3 endpoint that has this requirement.  Second, it would be "a bit odd" for a server to require two authorisation mechanisms (signing the URL and `Authorization` header).  Third, the code is (deliberately) written to make the behaviour as clear as possible; if there is a problem, it should be easy to update the code.

> For example, what happens with the exists() method when a potentially requested object does not exist?

In the `exists` method, the HEAD request would return a 404 status code, which is converted to a `False` response.  This requires that pre-signed URL is authorised to read the target object.

The `_upload_item` method (of the `UploadClient` class) takes care to create two pre-signed URLs: one with read access and the other with write access.  The method also takes care that the `exists` method is called using the pre-signed URL with read-access, so any GET request (and correspondingly, the actual HEAD request) is authorised.

(There's quite a bit I _could_ write about `_upload_item`, but that's not directly relevant to this patch ;-)

> Or with the get_space_usage which is DAV-specific?

You're correct: in `webdav.Default` class, the `get_space_usage` method uses PROPFIND, which won't work for S3 endpoints.

I checked how the `gfal.Default` class implements the `get_space_usage` method.  This returns the information about a "space token".  Space tokens are an old SRM technology that would only work for an SRM endpoint.  This means that, when using the "gfal" implementation, the `get_space_usage` method doesn't work with S3 endpoints.

Taking these together, I believe its safe to say that `get_space_usage` currently doesn't work for S3 endpoints, irrespective of which implementation is used (gfal or webdav).

Further, I checked where the protocol implementation's `get_space_usage` method is used.  Rucio itself doesn't make use of this information, but I found some references to it in the "probes" repo.  The `atlas/check_storage_space` probe uses this information, as do a few in the `attic` directory.  There could be other code that uses this, but overall it seems a rather niche usage.

In conclusion, it looks like the behaviour of `webdav.NoRename` class' implementation of `get_space_usage` is broken for S3 endpoints, but (likely) no worse that it is already broken when using the `gfal.NoRename`  class' implementation.  In any case, this functionality isn't used by Rucio client, server or daemons.

Good catch with the `rse_setttings` typo.  That should be fixed, now.

I've also rebased against the current tip of master, as requested.

Geogouz:
Thanks. Just to note for those who might face the same issue in the future, I recently came across a case (an edge one I assume) where the S3 service was actually returning 403 for paths that did not exist (instead of the expected 404), and yet, 200 for existing ones. Maybe the IAM or resource-based policy was granting `s3:HeadObject` only for particular prefixes or exact keys. So in such cases the `exist()`would be problematic (that's why I mentioned it). Nevertheless, I believe this PR is a good start for now.

Geogouz (approved):

Geogouz (approved):

Geogouz (approved):

Geogouz (approved):

Geogouz (approved):

paulmillar:
> Thanks. Just to note for those who might face the same issue in the future, I recently came across a case (an edge one I assume) where the S3 service was actually returning 403 for paths that did not exist (instead of the expected 404), and yet, 200 for existing ones. 

Yes, this seems to happen.

There are [various report of this](https://stackoverflow.com/questions/19037664/how-do-i-have-an-s3-bucket-return-404-instead-of-403-for-a-key-that-does-not-e), but I'm not sure how much they describe what happens with a pre-signed URL.

One of the issues address (in this series of patches) is similar.  The `webdav.Default` class `connect`method needed to be updated to accept 4xx status codes (at least, for pre-signed URLs).  The natural choice would have been to accept a 404 status code, but I went with accepting any 4xx status code because (IIRC) I was seeing the S3 endpoint returning a 403 status code.

> Maybe the IAM or resource-based policy was granting `s3:HeadObject` only for particular prefixes or exact keys. So in such cases the `exist()`would be problematic (that's why I mentioned it). Nevertheless, I believe this PR is a good start for now.

The `connect` method is already problematic, as it targets a URL that doesn't correspond to any valid Rucio replica, and yet it is expecting a HEAD request to succeed (i.e., a GET request should also succeed).

In WebDAV, there is no requirement for a GET request on a collection to succeed: it's not part of the specification.  Most WebDAV endpoints provide _something_ for GET requests on collections (often an HTML rendering of the collection), but this is not part of the WebDAV spec.

A better approach to check whether the server is available/working would be to send an OPTIONS request.  This would be similar to how pre-flight requests operate with CORS.  However, modifying that would potentially have a much larger impact (and corresponding risk), so I went for the lower risk option of just allowing 4xx status codes.

Issue #7309, opened 2025-01-13
New CLI: Can’t unlock a rule


dchristidis:
### Description

The `--locked` option no longer accepts an argument.

https://github.com/rucio/rucio/blob/44a26d27af980f6b0cfa5eca5b8872992415db81/lib/rucio/client/commands/rule.py#L55

### Steps to reproduce

```console
$ rucio rule update --locked=False 1D9F11066C3740EF939A871F132DD5EF
usage: rucio rule update [-h] [-a RULE_ACCOUNT] [--activity ACTIVITY] [--rule-id RULE_ID] [--lifetime LIFETIME]
                         [--locked] [--source-replica-expression SOURCE_REPLICA_EXPRESSION] [--comment COMMENT]
                         [--rses RSE_EXPRESSION] [--stuck] [--suspend] [--cancel-requests] [--priority PRIORITY]
                         [--child-rule-id CHILD_RULE_ID] [--boost-rule] [--move]
rucio rule update: error: argument --locked: ignored explicit argument 'False'
```

### Rucio Version

`36.0.0.post2`

voetberg:
Should be solved by #7253 - if this issue continues, we can reopen this

Issue #7302, opened 2025-01-08
New CLI: Calling commands with missing required arguments should result in clear error or help message


maxnoe:
### Description

I noticed that when using the new cli, running commands with missing required arguments results in sub-optimal error messages, for example:

```
$ rucio upload
2025-01-08 15:39:10,350	ERROR	'NoneType' object is not iterable
```

or

```
$ rucio did
# runs a few seconds, outputs nothing
```

### Steps to reproduce

See above

### Rucio Version

36.0.0post2

maxnoe:
Is this fixed now that the CLI is build with click? @voetberg

voetberg:
@maxnoe This should be resolved - but there's a non-zero chance there's stragglers, so if you find any strange errors like this lmk

Issue #7301, opened 2025-01-08
New CLI: inconsistent use of `add` verb


maxnoe:
### Description

In the new CLI, the add verb is used inconsistently and confusingly (at least to me).

For some commands, behaves both as adding and updating. Something I would rather connect to the verb "set", this includes `rucio rse attribute add`.

For some commands, it only performs "add" and there is a separate "set" to update existing values. This includes `rucio rse distance add/set`

### Steps to reproduce

```
$ rucio rse attribute -h
$ rucio rse distance -h
```

### Rucio Version

36

voetberg:
`attribute add` is a failure carried forward from the old version. [The old `rse attribute set` redirected to client.add_rse_attribute](https://github.com/rucio/rucio/blob/c71deebae79995646bc80f864eb0c4570c00d8c3/lib/rucio/client/commands/bin_legacy/rucio_admin.py#L504) Ideally `add` would only create new objects and `set` (or `update` as is the verbiage) would only change existing objects. This logic also applies to all of the `config [operation]`s more or less. 

`distance set` is just a mistake, that should be `distance update`. 

Related issue: https://github.com/rucio/rucio/issues/7228

voetberg:
Closed by #7344

Issue #7265, opened 2024-12-16
New CLI: Review replacement for `move-rule`


dchristidis:
### Description

`rucio move-rule` effectively does two things:

1. Creates a new replication rule, copying most of the properties of an existing rule, except the RSE expression. Some may be explicitly overridden by the user.
2. Sets the child rule ID and expiration timestamp on the existing rule.

The current replacement (i.e. `rucio rule update --move`) makes sense in that it is like updating the RSE expression of an existing rule, but does so it two steps, changing the rule ID in the process. Still, out of the two actions, the creation of a new rule seems more important.

Some alternatives that may merit consideration:

1. If #536 were implemented, the moving a rule could be implemented as updating its RSE expression and re-evaluating it instead. (Personally, I think this offers the best experience but is the least likely to happen due to the complexity.)
2. Add an additional `rucio rule move` subcommand.

### Motivation

Here’s the help page for the `update` subcommand:

```console
$ rucio rule update --help
usage: rucio rule update [-h] [-a RULE_ACCOUNT] [--activity ACTIVITY] [--rule-id RULE_ID] [--lifetime LIFETIME]
                         [--locked] [--source-replica-expression SOURCE_REPLICA_EXPRESSION] [--comment COMMENT]
                         [--rses RSE_EXPRESSION] [--stuck] [--suspend] [--cancel-requests] [--priority PRIORITY]
                         [--child-rule-id CHILD_RULE_ID] [--boost-rule] [--move]

options:
  -h, --help            show this help message and exit
  -a, --account RULE_ACCOUNT
                        The account of the rule
  --activity ACTIVITY   Activity to be used (e.g. User, Data Consolidation)
  --rule-id RULE_ID     The rule ID, for accessing an existing rule.
  --lifetime LIFETIME   Rule lifetime (in seconds)
  --locked              Set the rule to locked
  --source-replica-expression SOURCE_REPLICA_EXPRESSION
                        RSE Expression for RSEs to be considered for source replicas
  --comment COMMENT     Comment about the replication rule
  --rses, --rse-exp RSE_EXPRESSION
                        RSE to change for the rule
  --stuck               Set state to STUCK
  --suspend             Set state to SUSPENDED
  --cancel-requests     Cancel requests when setting rules to stuck
  --priority PRIORITY   Priority of the requests of the rule
  --child-rule-id CHILD_RULE_ID
                        Child rule id of the rule. Use "None" to remove an existing parent/child relationship
[...]

PR #7344, opened 2025-01-21
Upgrade CLI to use Click over Argparse

### LINKED ISSUE #7253: New CLI: Replace argparse with Click
### Description

I feel that it’s impossible to achieve a good experience while still working within the limitations of the `argparse` library.

This morning I had a quick look into [Click](https://click.palletsprojects.com/en/stable/) and I’m very impressed with the results. The help output is much closer to what I had imagine and it does so with significantly less boilerplate.

```python
#!/usr/bin/env python3

import click

@click.group()
@click.option('--account')
@click.pass_context
def main(ctx, account):
    ctx.ensure_object(dict)
    ctx.obj['account'] = account
    
@main.group()
def rule():
    pass
@rule.command()
def add():
    pass
@rule.command(name='list')
@click.option('--account')
@click.pass_context
def list_(ctx, account):
    print(f'Root --account: {ctx.obj["account"]}',
          f'Subcommand --account: {account}',
          sep='\n')
@rule.command()
def remove():
    pass
@rule.command()
def show():
    pass
@rule.command()
def update():
    pass

@main.command()
def whoami():
    pass
    
if __name__ == '__main__':
    main()
```

Example usage:
```console
$ ./click --help
Usage: click [OPTIONS] COMMAND [ARGS]...

Options:
  --account TEXT
  --help          Show this message and exit.

Commands:
  rule
  whoami
$ ./click rule --help
Usage: click rule [OPTIONS] COMMAND [ARGS]...

Options:
  --help  Show this message and exit.

Commands:
  add
  list
  remove
  show
  update
$ ./click rule list --help
Usage: click rule list [OPTIONS]

Options:
  --account TEXT
  --help          Show this message and exit.
$ ./click --account=foo rule list --account=bar
Root --account: foo
Subcommand --account: bar
```

### Motivation

_No response_

### Change

_No response_

voetberg:
Resolves #7253 

Changes the argument parsing of CLI commands to use [click](https://click.palletsprojects.com/en/stable/) over argparse. Makes changes to accommodate this update, including adding `--no-x` options for boolean flags, removing documentation from positional args, implementing lazy loading for `rucio` commands, creating an `Arguments` type to facilitate translation between the original CLI using argparse and the new click CLI. 

Related changes: 
* Adds the suffix "commands" to all cli command files (avoid confusion with core files) 
* Updates the "command map" for directing users to the updated command 
* Replaces many positional args that were removed in the initial rewrite (#7250)
* Adds many `TODO`s where more work should be done

Additional issues potentially addressed: 
#7235 
#7309 
#7265 
#7301 
#7302 

Testing error is due to PyLint being upside the auth arguments are not being supplied to `commands.main()` when called in `bin/rucio`, `bin/rucio-admin`. An ignore statement is included there but it looks like it isn't considered. Additional guidance there would be appreciated.


dchristidis:
> - Add the suffix "_command" to all cli files

Is this beneficial? On the server side, I always found it helpful that you expect to see the same filename in the REST API, gateway, and core layers.

But can we take a moment to plan the long-term structure? It appears that right now we have:

```
bin/rucio{,-admin}
lib/rucio/client/…
lib/rucio/client/commands/…
lib/rucio/client/commands/bin_legacy/…
```

`lib/rucio/client/…` likely has to stay as is. I applaud your choice to reduce `bin/rucio` to its simplest form, and have a new ‘commands’ pseudo-layer. I assume that the content of `lib/rucio/client/commands/bin_legacy/…` will eventually be moved to the ‘commands’? Meaning that both the command structure and their implementation will exist in the same functions?

But then the location of the `lib/rucio/client/commands/` directory looks somewhat out of place, and I don’t particularly like seeing, for example, `account_commands.py` and `accountclient.py`. Beyond the naming inconsistency, I think it creates a bit of confusion as to what is the CLI and what is the Client API layer. I wonder if `lib/rucio/cli/` might be preferable?

voetberg:
>> Add the suffix "_command" to all cli files

> Is this beneficial? On the server side, I always found it helpful that you expect to see the same filename in the REST API, gateway, and core layers.

I remember there being a conversation that everything being named the same was causing confusion, but I can no longer find written record of it, so I am just fine keeping them named without the `__commands` suffix if there are so objections. 

> I assume that the content of `lib/rucio/client/commands/bin_legacy/…` will eventually be moved to the ‘commands’? Meaning that both the command structure and their implementation will exist in the same functions?

Yes, that is the goal. I hope `bin_legacy` will cease to exist around 40.0.0. I have been having questions about this so I will make a small migration/upgrade plan on my fork for reference that we can iterate on. 

> I wonder if `lib/rucio/cli/` might be preferable?

I wouldn't have objections to moving the commands, but I don't have a strong reason to either. The changes were kept the `/client/` dir because they are all applications of client functionality and basically just shortcuts to combinations of client calls, so grouping them makes sense. I have absolutely 0 strong feelings on this though.

bari12:
I had to reflect a bit about this, it's not a clear cut, but to me we did not do ourselves a favour with calling what is effectively the API, `clients`. This creates a cascading mess now, but within this mess, I don't think we should use the directory to also store the `commands`. I get the motivation, that everything which somebody might need in a client environment (API, CLI) is within the `lib/rucio/client` path. But what will happen, I think, is that people get confused and will use these commands as APIs and mixing it up with the "normal" clients (the actual API).

In the long run, I think we should separate this by moving `clients` to `api` and move this commands layer somewhere else. Could be `lib/rucio/cli`

voetberg:
I am not opposed to this at all, I can change this PR around to reflect that.

lib/rucio/client/commands/utils.py
```diff
@@ -223,3 +214,21 @@ def setup_gfal2_logger():
...
+
+
+def click_decorator(func):
+    """Util to combine decorators always used together in CLI. Equivalent to:
+    @click.help_option("-h", "--help")
+    @click.pass_context
+    @exception_handler
+    def func....
+    """
+    return exception_handler(click.pass_context(click.help_option("-h", "--help")(func)))
```
dchristidis: Help me out here, because I fail to see the need for this.

* `@click.help_option("-h", "--help")` can be replaced with `@click.group(context_settings={"help_option_names": ["-h", "--help"]})` above `main()`.
* `@click.pass_context` can be used where appropriate.
* `@exception_handler` can also be put above `main()` and it will apply to all subcommands.

The only thing that I didn’t try in my own experiments was lazy-loading.

lib/rucio/client/commands/account_commands.py
```diff
@@ -0,0 +1,219 @@
...
+    Remove an account
+    (WARNING: Permanently disables the account. If you want to temporary disable, use `account update [account-name] --ban`)
+    """
+    delete_account(Arguments({"acnt": account_name}), ctx.obj.client, ctx.obj.logger, ctx.obj.console, ctx.obj.spinner)
+
+
+@account.command("update")
+@click.argument("account-name")
+@click.option("--email", help="Email address associated with the account")
+@click.option("--ban/--unban", default=None, help="Temperately disable/enable an account")
```
dchristidis: Temperately?

lib/rucio/client/commands/account_commands.py
```diff
@@ -0,0 +1,219 @@
...
+@click.help_option("-h", "--help")
+def attribute():
+    """View or modify account attributes"""
+
+
+@attribute.command("list")
+@click.argument("account-name")
+@click_decorator
+def attr_list(ctx, account_name):
+    "List the attributes for a given account. Note: This table is generally empty."
```
dchristidis: ‘This table is generally empty’ is community-specific.

lib/rucio/client/commands/account_commands.py
```diff
@@ -0,0 +1,219 @@
...
+    list_account_attributes(Arguments({"account": account_name}), ctx.obj.client, ctx.obj.logger, ctx.obj.console, ctx.obj.spinner)
+
+
+@attribute.command("add")
+@click.argument("account-name")
+@click.option("-o", "--option", type=(str, str), help="Key and value of the attribute", required=True)
```
dchristidis: This hinders the review process. You’re making a notable change to the command-line options that is hidden in a diff between a file that is entirely removed and a new file. I must ask you to make a commit that solely makes the use of Click without any other changes, unless absolutely mandatory (and those should be well-advertised in the commit message).

In this case, I find the change debatable. The user will need to remember that it’s `--option` to add but `--key` to remove. Personally, I don’t think it beneficial.

lib/rucio/client/commands/account_commands.py
```diff
@@ -0,0 +1,219 @@
...
+    """List the limits and current usage for an account"""
+    args = Arguments({"usage_account": account_name, "rse": rse})
+    list_account_usage(args, ctx.obj.client, ctx.obj.logger, ctx.obj.console, ctx.obj.spinner)
+
+
+@limit.command("add")
+@click.argument(
+    "account-name",
+)
+@click.option("--rse", "--rse-name", help="Full RSE name", required=True)
```
dchristidis: This is a weird case. With `local` scope it’s and RSE name, but with `global` scope it’s an RSE expression.

lib/rucio/client/commands/account_commands.py
```diff
@@ -0,0 +1,219 @@
...
+@account.group("identity")
+@click.help_option("-h", "--help")
+def identity():
+    """Manage identities for an account - used to login"""
+
+
+@identity.command("list")
+@click.argument("account-name", required=True)
+@click_decorator
+def id_list(ctx, account_name):
```
dchristidis: Isn’t this naming scheme difficult to follow? Why not name it `account_identity_list()` so that it matches the actual command?

lib/rucio/client/commands/account_commands.py
```diff
@@ -0,0 +1,219 @@
...
+    args = Arguments({"account": account_name})
+    list_identities(args, ctx.obj.client, ctx.obj.logger, ctx.obj.console, ctx.obj.spinner)
+
+
+@identity.command("add")
+@click.argument("account-name", required=True)
+@click.option("--type", "type_", type=click.Choice(["X509", "GSS", "USERPASS", "SSH", "SAML", "OIDC"]), help="Authentication type", required=True)
+@click.option("--id", help="Identity", required=True)
+@click.option("--email", help="Email address associated with the identity", required=True)
+@click.option("-pwd", "--password", help="Password if authtype is USERPASS")
```
dchristidis: Big no-no! There are three main conventions for command-line options: one-dash-one-letter (UNIX), one-dash-multiple-letters (X11), two-dashes-multiple-letters (GNU). The UNIX convention hash the added trick that options can be combined. Implementing all three in a utility sounds like a recipe for disaster (think about [tar](https://xkcd.com/1168/) and how unintuitive it is due to trying to support multiple conventions).

lib/rucio/client/commands/command.py
```diff
@@ -11,253 +11,252 @@
...
-
-    def __call__(self):
-        whoami_account(self.args, self.client, self.logger, self.console, self.spinner)
-
-
-class TestServer(CommandBase):
-    def _operations(self):
-        return {}
+# Taken directly from https://click.palletsprojects.com/en/stable/complex/#defining-the-lazy-group
+class LazyGroup(click.Group):
```
dchristidis: How does lazy-loading benefit us?

lib/rucio/client/commands/command.py
```diff
@@ -11,253 +11,252 @@
...
+        "rse": "rucio.client.commands.rse_commands.rse",
+        "rule": "rucio.client.commands.rule_commands.rule",
+        "scope": "rucio.client.commands.scope_commands.scope",
+        "subscription": "rucio.client.commands.subscription_commands.subscription",
+        "upload": "rucio.client.commands.upload_commands.upload_command",
+    },
+)  # TODO: Implement https://click.palletsprojects.com/en/stable/options/#dynamic-defaults-for-prompts for args from config or os
+@click.version_option(version.version_string(), message="%(prog)s %(version)s")
+@click.option("--config", help="The Rucio configuration file to use")
+@click.option("-v", "--verbose/--no-verbose", default=False, help="Print more verbose output")
```
dchristidis: I don’t see the need for having two options for a boolean when one of the is the default.

lib/rucio/client/commands/config_commands.py
```diff
@@ -0,0 +1,74 @@
...
+import click
+
+from rucio.client.commands.bin_legacy.rucio_admin import delete_config_option, get_config, set_config_option
+from rucio.client.commands.utils import Arguments, click_decorator
+
+
+@click.group()
+@click.help_option("-h", "--help")
+def config():
+    "Modify the rucio.cfg"
```
dchristidis: Not really. The modification applies to the config table in the database.

lib/rucio/client/commands/config_commands.py
```diff
@@ -0,0 +1,74 @@
...
+        $ rucio config add --section my-section -o key value
+    """
+    args = Arguments({"section": section, "option": option[0], "value": option[1]})
+    set_config_option(args, ctx.obj.client, ctx.obj.logger, ctx.obj.console, ctx.obj.spinner)
+
+
+@config.command("remove")
+@click.option("-s", "--section", help="Section", required=True)
+@click.option("-k", "--key", help="Key in section", required=True)
+@click.confirmation_option(prompt="Are you sure you want to delete this value?")
```
dchristidis: Again, this is a new feature. Please don’t hide it under the argparse-to-Click migration.

lib/rucio/client/commands/did_commands.py
```diff
@@ -0,0 +1,192 @@
...
+
+
+@did.command("remove")
+@click.option("--undo", is_flag=True, default=False, help="Undo erase DIDs. Only works if has been less than 24 hours since erase operation.")
+@click.argument("dids", nargs=-1)
+@click_decorator
+def remove(ctx, dids, undo):
+    """
+    This command sets the lifetime of the DID in order to expire in the next 24 hours.
+    After this time, the dataset is eligible for deletion.
```
dchristidis: ‘Eligible for deletion’ suggests that it is at someone’s discretion. Expired DIDs are force-deleted (and their replicas purged).

lib/rucio/client/commands/lifetime_exception_commands.py
```diff
@@ -0,0 +1,34 @@
...
+    """Interact with the lifetime exception model"""
+
+
+@lifetime_exception.command("add")
+@click.option("-f", "--input-file", help="File where the list of datasets requested to be extended are located")
+@click.option("--reason", help="The reason for the extension")
+@click.option("-x", "--expiration", help="The expiration date format YYYY-MM-DD")
+@click_decorator
+def add_(ctx, input_file, reason, expiration):
+    """Add an exception to the lifetime model (to have protections against deletion from reaper daemons)"""
```
dchristidis: This description is very wrong. Where did it come from?

lib/rucio/client/commands/rse_commands.py
```diff
@@ -0,0 +1,305 @@
...
+    args = Arguments({"rse": rse_name, "non_deterministic": non_deterministic})
+    add_rse(args, ctx.obj.client, ctx.obj.logger, ctx.obj.console, ctx.obj.spinner)
+
+
+@rse.command("remove")
+@click.argument("rse-name")
+@click.confirmation_option(prompt="Are you sure you want to delete this RSE?")
+@click_decorator
+def remove(ctx, rse_name):
+    """Permanently disable an RSE. WARNING: You may not be able to recover this RSE!"""
```
dchristidis: Again, what does ‘not be able to recover’ mean?

lib/rucio/client/commands/rse_commands.py
```diff
@@ -0,0 +1,305 @@
...
+@distance.command("update")
+@click.argument("source-rse")
+@click.argument("destination-rse")
+@click.option("--distance", default=1, type=int, help="Relative distance between RSEs")
+@click.option("--ranking", default=1, type=int, help="Ranking of link to prioritize transfers")
+@click_decorator
+def distance_update(ctx, source_rse, destination_rse, distance, ranking):
+    """Update the existing distance or ranking from SOURCE-RSE to DESTINATION-RSE"""
+    if (distance is None) and (ranking is None):
+        raise ValueError("Either [distance] or [ranking] must be specified")
```
dchristidis: Their default value is `1`. Under what circumstances will either be `None`?

Also why is ranking still referenced? It has been removed from the database.

lib/rucio/client/commands/rule_commands.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+@rule.command("move")
+@click.argument("rule_id")
+@click.option("--rses", "--rse-exp", help="RSE expression of new rule", required=True)
+@click.option("--activity", help="Update activity for moved rule", hidden=True)  # Should only do this using `update`
+@click.option("--source-rses", help="Update how replicas are sourced for the rule")
+@click_decorator
+def move(ctx, rule_id, rses, activity, source_rses):
+    """Duplicate a rule onto another RSE"""  # TODO: verify - does it delete the old one?
```
dchristidis: No, it doesn’t duplicate. The two rules exist in parallel with a parent-child relationship. When the child rule reaches `OK` state, the parent is deleted.

dchristidis (changes requested):

lib/rucio/client/commands/utils.py
voetberg: Admittedly I didn't know about the `"help_option_names"` setting until after submitting this PR, and you're right about the other options as well

lib/rucio/client/commands/account_commands.py
voetberg: "temporarily" 😶‍🌫️ Spell check does not have my best interests in mind it seems

lib/rucio/client/commands/account_commands.py
voetberg: This was pulled from the main oauth commands (to keep argument names consistent). Should it be changed there as well?

lib/rucio/client/commands/command.py
voetberg: To keep the `command.py` file from growing to unwieldy levels - it is largely an organizational feature in this case as the client themselves are light-weight to load in. I will argue for keeping lazy-loading though, for developer quality of life (I personally do not want to work with a file that could end up being over 6000 lines long when bin_legacy is removed.)

lib/rucio/client/commands/account_commands.py
voetberg: I will revert these and add them as a discussion topic.

lib/rucio/client/commands/account_commands.py
voetberg: This was grabbed from the original list-attr in rucio-admin, but I will remove it.

lib/rucio/client/commands/command.py
voetberg: I think I was still figuring out how click handles bools here - I will replace it with a `is_flag` option.

lib/rucio/client/commands/account_commands.py
voetberg: The `identity` commands were named with `id` so `id_add` would not conflict with the existing `identity_add` (from rucio-admin) - this should be solved with a `import as`. 

The general naming scheme is: 
* base group commands are named for their operation (`[group]_command.py::[operation]`)
* Subgroups are named for their subgroup and operation (`[group]_command.py::[subgroup]_[operation]`)

lib/rucio/client/commands/did_commands.py
voetberg: This was pulled from the original help menu, I will replace it with your explanation

lib/rucio/client/commands/lifetime_exception_commands.py
voetberg: Honestly not sure. Removing it now.

lib/rucio/client/commands/rse_commands.py
voetberg: The original from rucio-admin states "CAUTION: all information about the RSE might be lost!" which I think got mutated into the current version. I'll replace the pr comment with that

lib/rucio/client/commands/rse_commands.py
voetberg: It was not removed from the cli so it was replicated here (just a confusing string change warning is there). I will make that change

voetberg:
Remarks re: Failing tests 
* Type checks fail because of 2 things - incorrectly picking up `bin_legacy/rucio[-admin]` and not regarding the typing on arguments set by `click`. The first should be ignored, the second is open for discussion on if these arguments should have typing applied (double work, as `click` already sets the default type of `Union[str, None]` unless otherwise specified)
* The  Type Check PR also fails because of bin_legacy and should be ignored.

lib/rucio/client/commands/account_commands.py
dchristidis: Very well, thank you for the explanation. Then I’ll try to ignore the actual help content as much as possible and focus solely on the structure and implementation.

lib/rucio/client/commands/account_commands.py
dchristidis: Wow, I can’t believe this existed since 2012. Yes, this should be removed (even the idea that a password is passed as a command-line argument is ludicrous to me).

lib/rucio/client/commands/command.py
dchristidis: Thank you, no argument against it.

lib/rucio/client/commands/command.py
dchristidis: This applies to other options as well.

lib/rucio/client/commands/account_commands.py
voetberg: Two possible solutions are presented in https://github.com/rucio/rucio/issues/7505 - I will add a todo linking to that issue

lib/rucio/client/commands/rse_commands.py
dchristidis: See #5985 for reference. I’m not sure whether the client was ignored accidentally or on purpose, but it’s been long enough that we can remove it.

lib/rucio/cli/command.py
```diff
@@ -0,0 +1,267 @@
...
+@click.option("--account", "--issuer", "issuer", help="Rucio account to use.")
+@click.option("-S", "--auth-strategy", help="Authentication strategy (userpass, x509...)")
+@click.option("-T", "--timeout", type=float, help="Set all timeout values to seconds")
+@click.option("-U", "--user-agent", default="rucio-clients", help="Rucio User Agent")
+@click.option("--vo", help="VO to authenticate at. Only used in multi-VO mode")
+@click.option("--pager/--no-pager", default=False, hidden=True)
+@click.option("-u", "--user", help="Username for userpass")
+@click.option("-pwd", "--password", help="Password for userpass")
+@click.option("--oidc-user", help="OIDC username")
+@click.option("--oidc-password", help="OIDC password")
```
dchristidis: Both argparse and Click preserve the option order. There’s no logic to their ordering, at least none that I can tell. Should we consider ordering them alphabetically (base on the long option)?

lib/rucio/client/commands/account_commands.py
voetberg: View/add comments to https://github.com/rucio/rucio/issues/7523

lib/rucio/cli/command.py
voetberg: I think grouping user/pass options together is a good idea, but otherwise I don't see a problem with that

lib/rucio/client/commands/command.py
voetberg: I will argue for using this in some cases (Having explicit true/false flags is useful when you know the option exists but can't remember the default - a not unheard-of situation), but in the spirit of not making changes explicitly outside the migration I will put it up as a discussion topic.

dchristidis: But looking at how you’re using this feature, you end up in the opposite situation. You say ‘can’t remember the default’, instead I see ‘cant’ remember the prefix’.

```
@click.option("--locked/--unlocked", default=False, help="Rule locking")  # Kept so rules can be unlocked as well as locked
```

Before I would have had to remember that the option is called `--locked` and I can pass it either `True` or `False`. If I don’t remember what is the default, I can set to what I want anyway. With you suggestion, I have to remember than the opposite of `--locked` is `--unlocked`. And in the very last commit, some options were prefixed with `--no-` and some with `--not-`. Why do you think it more likely to not remember the default than the opposite?

Please revert that one as well and open an issue if you wish to pursue this further. I am but one voice and I should not be making such decisions. But that means that they should not be bundled with the argparse-to-click migration.

lib/rucio/client/commands/account_commands.py
dchristidis: I have my own thoughts, but this has no effect on the end user so rightfully the decision rests with you.

dchristidis (commented):
There’s much in this PR that merit a closer look. A lot of it is cruft accumulated over multiple years of Rucio development (and not always with a clear long-term vision). Some is not, but I’m now asking myself whether this PR fixes what it set out to do. The response is a resounding **YES**.

Please address the only remaining open comment and rebase on top of the latest master. I am prepared to give my approval (and I am very eager to test the release candidate in action).

Well done.

lib/rucio/cli/upload.py
```diff
@@ -0,0 +1,60 @@
...
+
+@click.command("upload")
+@click.argument("file-paths", nargs=-1)
+@click.option("--rse", "--rse-name", help="Rucio Storage Element (RSE) name", required=True)
+@click.option("--lifetime", type=int, help="Lifetime of the rule in seconds")
+@click.option("--expiration-date", help="The date when the rule expires in UTC, format: <year>-<month>-<day>-<hour>:<minute>:<second>. E.g. 2022-10-20-20:00:00")
+@click.option("--scope", help="Scope name.")
+@click.option("--impl", type=click.Choice([]), help="Transfer protocol implementation to use (e.g: xrootd, gfal.NoRename, webdav, ssh.Rsync, rclone)")
+# The --no-register option is hidden. This is pilot ONLY. Users should not use this. Will lead to unregistered data on storage!
+@click.option("--no-register", is_flag=True, default=False, hidden=True)
```
dchristidis, line 29: ~~I think you need to add an additional `"register"` argument here for the tests to pass.~~

Rather rename the function parameter below to `no_register`.

voetberg: 🤦 Totally missed that one. Corrected in latest force push

dchristidis (approved):

Issue #6016, opened 2022-11-18
Log into another VO using the password for another VO


gmatthews20:
### Description

Have a username and password for 2 different VOs, both with the same username and different passwords.
I can access the VO1 using the password set for VO1.
For the VO2 I can't access using the password set for VO2 but can access using the password for the VO1.
Can't access VO1 using the password of VO2.

When the password is removed from the VO2 I can no longer access VO2 using the password of VO1

This is happening on both the current webui and webui2.0

maany:
Hey @gmatthews20, would you like to investigate this to figure out the root cause?

gmatthews20:
I have had a look into this and I think the issue might be that the identities don't have a VO associated with them. So both accounts on each VO are being mapped to the same identity.

![image](https://user-images.githubusercontent.com/112955466/206197179-3065fc3b-f264-4eb6-ae67-dfbb1566fda4.png)

![image](https://user-images.githubusercontent.com/112955466/206196673-1fdabbc1-d136-40ed-9be0-e0c31abf06aa.png)

dchristidis:
As discussed in today’s Rucio meeting, since the issue is not limited to the Web UI, I’ve moved it to the main Rucio repository.

maany:
Hey @gmatthews20, I believe this happens because the second identity was not actually created. 

There is a `primary_key`  constraint on the `(identity, identity_type)` columns of the `Identities` Table.  

https://github.com/rucio/rucio/blob/0c04740b4a15b6a37dcc8430c66473aff97226e8/lib/rucio/db/sqla/models.py#L343

However, if you'd try to create a new identity that conflicts with the composite key in the Identities Table, no errors or warnings are raised to inform the user that the new identity was not created. 

So, while you would think that you have created two identities with the same username, the reality is that only the first identity got mapped to both the accounts `gmatthews@tst` and `gmatthews@ts2`

That's why you were able to login with the password for the first identity to the second account. ( damn, this can get confusing very fast). Also, when you "removed the password for VO2", you actually removed the password for the `gmatthews` identity that is mapped to both the accounts (and VOs), thereby not allowing you to login at all. 

Anyways, I believe the solution to this issue should be to raise an error if the new identity would trigger a `primary_key` constraint violation.

I would not be in support of changing the database schema to make Identities dependent on VOs, I think we should just handle the creation of identities in a better way.
 
@bari12 thoughts?

bari12:
[...]

PR #7350, opened 2025-01-22
Clients: allow upload without configuring the account

### LINKED ISSUE #7349: Client must specify the account when uploading data
### Description

For many operations, the client does not need to specify the account.  If no account is configured then rucio will use the default account for the user's identity.

However, this doesn't work when uploading data.  In this case, the account must be configured.

If the user has not configured the account information then many operations will succeed but attempts to upload will fail with the obscure error message:

    can only concatenate str (not "NoneType") to str

### Steps to reproduce

1. Remove `account` property from `rucio.cfg` file.
2. Ensure account is not otherwise specified.
3. Attempt to upload data (`rucio upload ...`)

### Rucio Version

Client: 36.0.0 Server: 35.6.0

### Additional Information

Commit 00326935c8 intended to make account an optional configuration property for the client, but was incomplete.

paulmillar:
Motivation:

Commit 00326935c8 updated the client so that it was no longer necessary for the client to specify the account; the support for this already existed in the server.

Unfortunately, this patch was incomplete: it is still required that the user configures the desired account when uploading data, even when the user's identity maps to a single identity.

Moreover, failure to configure the account leads to an obscure error message:

    can only concatenate str (not "NoneType") to str

This provides the user with no useful information on how to correct the problem.  Enabling verbose output leads to the client providing arguably incorrect information:

    This means the parameter you passed has a wrong type.

Modification:

The UploadClient class now discovers an account for the user's identity if no account was specified.  It does this by making a whoami query against the Rucio server, and using the supplied account.

If the server is unable to provide the account information for this user then the operation will fail with a clear error message that provides information on how to correct the situation.

Result:

It is now possible to upload data without the client configuring the desired account.  Under these circumstances, the default account for that user identity is used.

Closes: #7349

### DIFF
--- a/lib/rucio/client/uploadclient.py
+++ b/lib/rucio/client/uploadclient.py
@@ -81,6 +81,16 @@ class UploadClient:
         self.tracing = tracing
         if not self.tracing:
             logger(logging.DEBUG, 'Tracing is turned off.')
+        if self.client.account is None:
+            self.logger(logging.DEBUG, 'No account specified, querying rucio.')
+            try:
+                acc = self.client.whoami()
+                if acc is None:
+                    raise InputValidationError('account not specified and rucio has no account with your identity')
+                self.client.account = acc['account']
+            except RucioException as e:
+                raise InputValidationError('account not specified and problem with rucio: %s' % e)
+            self.logger(logging.DEBUG, 'Discovered account as "%s"' % self.client.account)
         self.default_file_scope: Final[str] = 'user.' + self.client.account
         self.rses = {}
         self.rse_expressions = {}
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -67,8 +67,10 @@ def pytest_make_parametrize_id(
 ) -> Optional[str]:
     if argname == 'file_config_mock':
         cfg = {}
-        for section, option, value in val['overrides']:
+        for section, option, value in val.get('overrides', []):
             cfg.setdefault(section, {})[option] = value
+        for section, option in val.get('removes', []):
+            cfg.setdefault(section, {})[option] = "[REMOVED]"
         return argname + str(cfg)
     if argname == 'core_config_mock':
         cfg = {}
@@ -604,13 +606,15 @@ def file_config_mock(request: pytest.FixtureRequest) -> "Iterator[ConfigParser]"
     """
     from unittest import mock
 
-    from rucio.common.config import Config, config_add_section, config_has_section, config_set
+    from rucio.common.config import Config, config_add_section, config_has_option, config_has_section, config_remove_option, config_set
 
     # Get the fixture parameters
     overrides = []
+    removes = []
     params = __get_fixture_param(request)
     if params:
         overrides = params.get("overrides", overrides)
+        removes = params.get("removes", removes)
 
     parser = Config().parser
     with mock.patch('rucio.common.config.get_config', side_effect=lambda: parser):
@@ -618,6 +622,9 @@ def file_config_mock(request: pytest.FixtureRequest) -> "Iterator[ConfigParser]"
             if not config_has_section(section):
                 config_add_section(section)
             config_set(section, option, value)
+        for section, option in (removes or []):
+            if config_has_section(section) and config_has_option(section, option):
+                config_remove_option(section, option)
         yield parser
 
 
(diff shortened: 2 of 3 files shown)


voetberg:
Can you write a test in `tests/test_upload.py` that replicates the original bug and then corrects it? Based on my understanding of how 'whoami' works (checking the issuer env var), I'm confused how this will get information the original client didn't already have.

paulmillar:
Hi @voetberg ,

Thanks for the feedback and the hint with `tests/test_upload.py`.

I've added an additional patch that updates the `tests/test_upload.py` file (and some ancillary, support functions in `tests/conftest.py`).  This should describe the problem the patch is tackling.  The updated tests fail with current tip of `master` branch under all deployment scenarios.  With this patch, the tests pass for most deployment scenarios 

Currently, there's an unresolved issue: the test is failing for two deployment scenarios, with the server rejecting the initial authentication.

The problem seems to be triggered when authenticating with username+password against a multi-VO server. The problem isn't present when authenticating against single-VO servers and I don't see this problem when authenticating using OIDC on a multi-VO instance in real life.

I'm still investigating (as I have spare time) but any hints would be appreciated.

voetberg:
I suspect this is an inherent limitation of multi-vo Rucio and how we are testing it. ([view this fixture for why its hitting the issuer=super_user](https://github.com/rucio/rucio/blob/e19e196bfa438b5c5b6c583a15945fe2b7743c43/tests/test_multi_vo.py#L75)) I would be comfortable just marking this as non-multi-vo (and adding a specific error message when you hit a cannont-auth error after running the account-getting logic)

paulmillar:
@voetberg ,

To be honest, I didn't completely understand your earlier comment.
 
However, I've tracked down the cause of this `super_root` problem with M-VO matrix test deployments, which turns out to be because of (what I believe to be) a bug in Rucio.  This bug doesn't seem to have been reported already, so I've opened #7394.

Having identified the problem and reported it, I felt it was reasonable to update the functional tests to skip the multi-VO testing: the logic skipping this test may be removed once #7394 is closed.

I've manually tested (an earlier version of) this patch on a real M-VO Rucio deployment, using OIDC.  This worked fine, so I believe the patch is "good" (at least, for users authenticating via OIDC and where their identity maps to a single account).  The problem seen in testing is only an artefact of the how the testing is done (username+password testing with the same identity mapped to the `super_root` and `root` accounts), which is triggering the above issue.

If a user has an identity that maps to different accounts within different VOs, then (because of the above issue) they would be forced to specify the account name (e.g., in the client section of the `rucio.cfg` file) for their non-default account.  This is no different from how things are currently so, for such users, this patch brings no disadvantage.

This is all to say that, I believe the patch is safe.

You mentioned adding a specific error message if the whoami request doesn't work.

I think the patch supports this.  If the whoami query fails then the code raises an InputValidationError with a message like `account not specified and problem with rucio: [...]`, where `[...]` is replaced by the error message from the server.

voetberg:
@paulmillar 
> To be honest, I didn't completely understand your earlier comment.

I think you did exactly what I suggested, but in the future, do not be afraid to reach out for clarification! 

> However, I've tracked down the cause of this super_root problem with M-VO matrix test deployments, which turns out to be because of (what I believe to be) a bug in Rucio. This bug doesn't seem to have been reported already, so I've opened https://github.com/rucio/rucio/issues/7394. 

I'll be honest, I'm not sure this is a bug (though it is strange behavior.) I will defer to @Thysk on this. 

Otherwise, this PR looks fine and I'll go ahead and approve it.

voetberg (approved):

paulmillar:
Does this pull request require any further input from me?

voetberg:
@paulmillar 

> Does this pull request require any further input from me?

Nope! Just waiting on a merge

@bari12 Would it be okay to merge this?

thysk:
I believe this is similar behaviour described here #6016 and its something that I have thought about for a time, but would require a schema change in the DB (or that was the thought me and Martin had when we talked about it in the ticket / issue review)

paulmillar:
@Thysk, I've looked into the issue you raised.  I believe this is actually unrelated.

The issue this PR is targeting (#7349) is only about taking advantage of a Rucio feature, where an identity has a corresponding default account (within the selected VO).

From my perspective, the issue you identified (#6016) describes a limitation on the API/DB schema that doesn't protect against misconfiguration with multiple username+passwords with the same username.  I've added a comment to that issue, that I hope clarifies this.

PR #7361, opened 2025-01-23
Protocols: Remove `gsiftp.py`

### LINKED ISSUE #7347: Remove `gsiftp.py`
https://github.com/rucio/rucio/blob/dfcd980f4f10720af38ea1570a6015b2478a5408/lib/rucio/rse/protocols/gsiftp.py#L60-L61

could be instead set via `protocol_attr`

rdimaio:
fix #7347


dchristidis (changes requested):
1. ‘Deprecate’ should mean that something is pending removal but remains available. As such, the use of the term in this case is improper.
2. Removing the file `gsiftp.py` is one thing. Removing anything related to the scheme was not agreed upon. The GFAL protocol implementation (i.e. `gfal.py`) and FTS can handle `gsiftp://` URLs just fine.

dchristidis (approved):

PR #7363, opened 2025-01-27
Client: Update help menu for rucio distance list / remove and account

### LINKED ISSUE #7362: CLI: help text sometimes shows incorrect arguments
### Description

In the case of the `rucio rse distance list --help` and `rucio rse distance remove --help` the argument `distance` is show but it's irrelevant in this command. This is because the help arguments are common for all `rucio rse distance` commands.

I will check other commands to see if this also happens.

I will make a PR to fix this.

This issue does not appear to have any functional impact.

### Steps to reproduce

```
❯ rucio rse distance list --help
usage: rucio rse distance list [-h] [--source SOURCE] [--destination DESTINATION] [--distance DISTANCE]
```

```
❯ rucio rse distance remove --help
usage: rucio rse distance remove [-h] [--source SOURCE] [--destination DESTINATION] [--distance DISTANCE]
```

### Rucio Version

36.1.0

### Additional Information

_No response_

lobis:
This fixes https://github.com/rucio/rucio/issues/7362.

Before this PR:

```
❯ rucio rse distance remove --help
usage: rucio rse distance remove [-h] [--source SOURCE] [--destination DESTINATION] [--distance DISTANCE]

optional arguments:
  -h, --help            show this help message and exit
  --source SOURCE       Source RSE name
  --destination DESTINATION
                        Destination RSE name
  --distance DISTANCE   Distance between RSEs
```

After this PR:
```
❯ rucio rse distance list --help
usage: rucio rse distance list [-h] [--source SOURCE] [--destination DESTINATION]

options:
  -h, --help            show this help message and exit
  --source SOURCE       Source RSE name
  --destination DESTINATION
                        Destination RSE name
```

Updated commands:

- `rucio rse distance list`
- `rucio rse distance remove`

It also has some minor updates to the `rucio account` command. Some very minor changes to the help text but mostly it's internal changes to unify the usage of argparse.

### DIFF
--- a/lib/rucio/client/commands/account.py
+++ b/lib/rucio/client/commands/account.py
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
 from argparse import SUPPRESS
 from typing import TYPE_CHECKING
 
@@ -60,20 +61,20 @@ class Account(CommandBase):
         parser.add_argument("--filters", dest="filters", action="store", help="Filter arguments in form `key=value,another_key=next_value`")
 
     def add_namespace(self, parser: "ArgumentParser") -> None:
-        parser.add_argument("--type", dest="accounttype", help="Account Type (USER, GROUP, SERVICE)", required=True)
+        parser.add_argument("--type", dest="account_type", help="Account Type", choices={"USER", "GROUP", "SERVICE"}, required=True)
         parser.add_argument("-a", "--account", dest="account", help="Account name", required=True)
-        parser.add_argument("--email", dest="accountemail", help="Add an email address associated with the account")
+        parser.add_argument("--email", dest="email", help="Add an email address associated with the account")
 
     def show_namespace(self, parser: "ArgumentParser") -> None:
         parser.add_argument("-a", "--account", dest="account", help="Account name", required=True)
 
     def update_namespace(self, parser: "ArgumentParser") -> None:
-        parser.add_argument("-a", "--account", help="Account name", required=True)
-        parser.add_argument("--email", help="Account email")
-        parser.add_argument("--ban", type=bool, choices=(True, False), help='Ban the account, to disable it. Use --ban False to unban.', default=None)
+        parser.add_argument("-a", "--account", dest="account", help="Account name", required=True)
+        parser.add_argument("--email", dest="email", help="Account email")
+        parser.add_argument("--ban", dest="ban", type=bool, choices=(True, False), help='Ban the account, to disable it. Use --ban False to unban.', default=None)
 
     def remove_namespace(self, parser: "ArgumentParser") -> None:
-        parser.add_argument("-a", "--account", dest="acnt", action="store", help="Account name", required=True)
+        parser.add_argument("-a", "--account", dest="account", action="store", help="Account name", required=True)
 
     def _operations(self) -> dict[str, "OperationDict"]:
         return {
@@ -120,7 +121,7 @@ class Attribute(Account):
         return {}
 
     def namespace(self, subparser):
-        subparser.add_argument("-a", "--account", help="Account name")
+        subparser.add_argument("-a", "--account", dest="account", help="Account name")
         subparser.add_argument("--key", dest="key", action="store", help="Attribute key")
         subparser.add_argument("--value", dest="value", action="store", help="Attribute value")
 
@@ -153,7 +154,7 @@ class Limit(Account):
         parser.add_argument("-a", "--account", dest="account", help="Account name", required=True)
         parser.add_argument("--rses", "--rse-exp", dest='rse', action="store", help="RSE expression")
         parser.add_argument("--bytes", action="store", help='Value can be specified in bytes ("10000"), with a storage unit ("10GB"), or "infinity"')
-        parser.add_argument("--locality", nargs="?", default="local", choices=["local", "global"], help="Global or local limit scope")
+        parser.add_argument("--locality", nargs="?", default="local", choices={"local", "global"}, help="Global or local limit scope")
         parser.add_argument("--human", default=True, help=SUPPRESS)
 
     def _operations(self) -> dict[str, "OperationDict"]:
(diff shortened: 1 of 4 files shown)


voetberg:
These tests are failing because when the variable name is changed in the `bin_legacy/rucio_admin` functions, they are not changed in their corresponding parser. 

`dest` in argparser is quite annoying IMHO, they need to be the same for every argparser group that uses the same function. I will mark where they need to be changed. 

When changing client code, I would recommend running the following test files at a minimum before committing (to avoid having to push code, see it fail in the testing suite, and have to push again). If you're working on this over a lunch break it's probably better just to run everything in `tests/` but that can take >=35 minutes to finish. 

```
tests/test_bin_rucio.py 
tests/test_cli_client_structure.py
tests/test_clients.py
tests/test_download.py
tests/test_impl_upload_download.py
tests/test_upload.py
```


Additionally, we use commit messages to build release notes, so it would be best to follow this [commit message format](https://rucio.cern.ch/documentation/contributing#4-commit-your-changes)

lib/rucio/client/commands/account.py
```diff
@@ -153,7 +154,7 @@ def namespace(self, parser: "ArgumentParser") -> None:
         parser.add_argument("-a", "--account", dest="account", help="Account name", required=True)
         parser.add_argument("--rses", "--rse-exp", dest='rse', action="store", help="RSE expression")
         parser.add_argument("--bytes", action="store", help='Value can be specified in bytes ("10000"), with a storage unit ("10GB"), or "infinity"')
-        parser.add_argument("--locality", nargs="?", default="local", choices=["local", "global"], help="Global or local limit scope")
+        parser.add_argument("--locality", nargs="?", default="local", choices={"local", "global"}, help="Global or local limit scope")
```
voetberg, line 157: Why is this change required? `choices` takes any sequential object https://docs.python.org/3/library/argparse.html#choices

lib/rucio/client/commands/account.py
```diff
@@ -60,20 +61,20 @@ def list_namespace(self, parser: "ArgumentParser") -> None:
...
+        parser.add_argument("--email", dest="email", help="Add an email address associated with the account")
 
     def show_namespace(self, parser: "ArgumentParser") -> None:
         parser.add_argument("-a", "--account", dest="account", help="Account name", required=True)
 
     def update_namespace(self, parser: "ArgumentParser") -> None:
-        parser.add_argument("-a", "--account", help="Account name", required=True)
-        parser.add_argument("--email", help="Account email")
-        parser.add_argument("--ban", type=bool, choices=(True, False), help='Ban the account, to disable it. Use --ban False to unban.', default=None)
+        parser.add_argument("-a", "--account", dest="account", help="Account name", required=True)
```
voetberg, line 72: The `dest` field by default is full arg name (in this case `--account` without the `--` (illegal characters for a python var name are stripped out)), so this is redundant.

lib/rucio/client/commands/bin_legacy/rucio_admin.py
```diff
@@ -82,7 +81,7 @@ def add_account(args, client, logger, console, spinner):
     Adds a new account. Specify metadata fields as arguments.
 
     """
-    client.add_account(account=args.account, type_=args.accounttype, email=args.accountemail)
```
voetberg, line 85: I agree that these are better names for the variables, but to make these changes please also change the parser in this file to use them.

https://github.com/rucio/rucio/blob/2e8894e561e4a8a2090543e65edc140f1f154505/lib/rucio/client/commands/bin_legacy/rucio_admin.py#L1526 and https://github.com/rucio/rucio/blob/2e8894e561e4a8a2090543e65edc140f1f154505/lib/rucio/client/commands/bin_legacy/rucio_admin.py#L1527

lib/rucio/client/commands/bin_legacy/rucio_admin.py
```diff
@@ -95,8 +94,8 @@ def delete_account(args, client, logger, console, spinner):
     Delete account.
 
     """
-    client.delete_account(args.acnt)
```
voetberg, line 98: Please change this variable dest - https://github.com/rucio/rucio/blob/2e8894e561e4a8a2090543e65edc140f1f154505/lib/rucio/client/commands/bin_legacy/rucio_admin.py#L1542

lib/rucio/client/commands/rse.py
```diff
@@ -162,10 +166,11 @@ def usage_example(self) -> list[str]:
             "$ rucio rse distance set --source rse1 --destination rse2 --distance 20  # Update an existing distance",
         ]
 
-    def namespace(self, parser: "ArgumentParser") -> None:
+    def namespace(self, parser: "ArgumentParser", distance: bool = True) -> None:
```
voetberg: I think a better solution would be to just make a new namespace, similar to how `account` has different namespaces for its functions.

voetberg (changes requested):
Please view my comments for changes

lib/rucio/client/commands/bin_legacy/rucio_admin.py
lobis: Addressed, thanks!

lib/rucio/client/commands/bin_legacy/rucio_admin.py
lobis: Addressed, thanks!

lib/rucio/client/commands/rse.py
lobis: Addressed in https://github.com/rucio/rucio/pull/7363/commits/9585a5f0f7bb1f7ff99847899ee27e1a99703e20

I used a different function for each namespace but avoided code duplication whenever possible when adding arguments to the parser.

lib/rucio/client/commands/account.py
lobis: Thanks for clarifying. Even if it's redundant I added it because I saw it in other invokations of `parser.add_argument` for account, to match it's style. IMHO it's a good idea to explicitly declare it to make search easy or for a future (unlikely) rename.

lib/rucio/client/commands/account.py
lobis: It should be functionally identical, but I saw that there was not a clear choice in the codebase to use sets or lists and thought I might start unifying to use sets, just for style / coherences purposes.

lobis:
> Please view my comments for changes

Thanks for the very detailed review! I have addressed all suggestions.

lib/rucio/client/commands/account.py
voetberg: Nothing wrong with that, but that change is unrelated to the purpose of this PR as described by the title. I would refrain from rolling in stylistic changes into PRs focused on functional changes (it muddies the git blame)

lobis: I will keep that in mind in the future, thanks for the feedback!

voetberg (approved):

bari12 (approved):

PR #7381, opened 2025-02-04
Testing: remove policy restriction for `test_default_hash` #7334

### LINKED ISSUE #7334: `test_default_hash` should run in all conditions, not just with ATLAS policy
This test:

https://github.com/rucio/rucio/blob/dfcd980f4f10720af38ea1570a6015b2478a5408/tests/test_rse_lfn2path.py#L46-L56

Is being skipped if the policy is not ATLAS.

However, ATLAS does not have a specific lfn2pfn algorithm (https://gitlab.cern.ch/atlas-adc-ddm/rucio-policy-package/-/blob/master/src/atlas_rucio_policy_package/__init__.py?ref_type=heads, CERN login needed), so the default hash algorithm should just be tested regardless of whether the policy is ATLAS or not

0xquark:
The test_default_hash was being skipped unless the policy was ATLAS, but ATLAS does not have a specific lfn2pfn algorithm. The default hash algorithm should be tested universally.

Removed the @pytest.mark.skipif condition to ensure the test runs in all conditions.

Fixes : #7334

### DIFF
--- a/tests/test_rse_lfn2path.py
+++ b/tests/test_rse_lfn2path.py
@@ -43,7 +43,6 @@ class TestDeterministicTranslation:
         )
         assert translator.path("foo", "bar") == "foo/4e/99/bar"
 
-    @pytest.mark.skipif(os.environ.get('POLICY') != 'atlas', reason='Test ATLAS hash convention')
     def test_default_hash(self):
         """LFN2PFN: Translate to path using default algorithm (Success)"""
         translator = RSEDeterministicTranslation(
--- a/tests/test_rse_lfn2path.py
+++ b/tests/test_rse_lfn2path.py
@@ -43,6 +43,7 @@ class TestDeterministicTranslation:
         )
         assert translator.path("foo", "bar") == "foo/4e/99/bar"
 
+    @pytest.mark.skipif(os.environ.get('POLICY') == 'belleii', reason='BelleII does not use hashed lfn2pfn')
     def test_default_hash(self):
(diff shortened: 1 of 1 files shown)


0xquark:
The test is failing for Belle2 but passing for ATLAS. Can someone share the Belle2 policy package with me? I want to check the hashing algorithm used in that policy. 

I found a temporary Belle2 policy package ([link](https://github.com/rucio/temporary-belle2-policy-package/blob/main/src/belleii_rucio_policy_package/lfn2pfn.py)), but it has its own "deterministic" translation that isn’t truly deterministic. It inherits from `RSEDeterministicTranslation` but overrides its behavior, returning the `name` instead. That’s why, in the test, it returns `bar` ([link](https://github.com/rucio/rucio/actions/runs/13132682037/job/36653775093?pr=7381)), which is just what was passed in.  

Can someone confirm if this is the cause of the failure? Should we update the Belle2 package to use the default hashing algorithm, or should I modify the test to account for Belle2 returning `bar`?

bari12:
This test doesn't make much sense for Belle II, since they do not use the hash lfn2pfn. All the BelleII `name`s are already full pathes (`/a/b/c/d/file.txt`) so it doesn't make sense to add an extra transformation for it.

0xquark:
Thanks @bari12! I’ve skipped this test for BelleII since it doesn’t use the hashed lfn2pfn.

rdimaio (approved):

PR #7384, opened 2025-02-05
Transfers: Do not ignore requests from expired rules with a child #7379

### LINKED ISSUE #7379: Requests never processed under certain condition
### Description

Here is a condition were some requests will never be processed by the conveyor.
I did a move-rule on the rule ` 060dc94316b34d72a959ad2411c19a81` to `dest_rse` `IN2P3CC-TMP-SE` :
```
rucio list-rules /belle/MC/fab/step1/MC16rd_prompt/prod00047806/s00/e0030/4S/r00000/eeee/10601400/mdst/sub15
ID                                ACCOUNT    SCOPE:NAME                                                                                          STATE[OK/REPL/STUCK]     RSE_EXPRESSION       COPIES    SIZE    EXPIRES (UTC)        CREATED (UTC)
--------------------------------  ---------  --------------------------------------------------------------------------------------------------  -----------------------  -------------------  --------  ------  -------------------  -------------------
ec6f08e2c41a4628a75b9d444507125c  diracprd   mc_tmp:/belle/MC/fab/step1/MC16rd_prompt/prod00047806/s00/e0030/4S/r00000/eeee/10601400/mdst/sub15  REPLICATING[1722/181/0]  ANY=true             1         N/A                          2025-01-29 21:52:19
75fa262f8d854caaa1696381801c25ee  rucioddm   mc_tmp:/belle/MC/fab/step1/MC16rd_prompt/prod00047806/s00/e0030/4S/r00000/eeee/10601400/mdst/sub15  REPLICATING[1373/530/0]  IN2P3CC-TMP-SE       1         N/A                          2025-01-30 17:17:15
060dc94316b34d72a959ad2411c19a81  rucioddm   mc_tmp:/belle/MC/fab/step1/MC16rd_prompt/prod00047806/s00/e0030/4S/r00000/eeee/10601400/mdst/sub15  REPLICATING[1373/530/0]  productionSEshare>0  1         N/A     2025-01-30 17:17:22  2025-01-29 21:52:44
```

There are 530 files that were not processed at all by the conveyor for the last 4 days. Checking the files in requests
```sql
rucio=> select count(*), dest_rse_id, rule_id, state from requests where name like '/belle/MC/fab/step1/MC16rd_prompt/prod00047806/s00/e0030/4S/r00000/eeee/10601400/mdst/sub15%' group by dest_rse_id, rule_id, state;
 count |             dest_rse_id              |               rule_id                | state
-------+--------------------------------------+--------------------------------------+-------
   530 | 23c29176-fd14-48e3-98c8-17d50bd3f1a6 | 060dc943-16b3-4d72-a959-ad2411c19a81 | Q
(1 row)
```

There are all in `Q` state (we don't have a preparer) and the `dest_rse` is online. There requests are linked to rule `060dc94316b34d72a959ad2411c19a81` but are also used for the rule `75fa262f8d854caaa1696381801c25ee` which is the child (both rules have the same `dest_rse`), but the requests won't be processed because the parent as an expiration date and https://github.com/rucio/rucio/blob/master/lib/rucio/core/request.py#L530
The condition mentioned above should be relaxed in case the rule has a child with the same destination as the parent

### Steps to reproduce

See explanation above

### Rucio Version

_No response_

### Additional Information

_No response_

dchristidis:


lib/rucio/core/request.py
```diff
@@ -527,7 +524,11 @@ def list_and_mark_transfer_requests_and_source_replicas(
         models.ReplicationRule,
         models.Request.rule_id == models.ReplicationRule.id
     ).where(
-        coalesce(models.ReplicationRule.expires_at, now) >= now
+        or_(models.ReplicationRule.child_rule_id != null(),
+            and_(models.ReplicationRule.child_rule_id == null(),
+                 models.ReplicationRule.expires_at == null()),
+            and_(models.ReplicationRule.child_rule_id == null(),
+                 models.ReplicationRule.expires_at > datetime.datetime.utcnow()))
```
rdimaio, line 531: Could be simplified to

```suggestion
            and_(
                or_(models.ReplicationRule.expires_at == null(),
                    models.ReplicationRule.expires_at > datetime.datetime.utcnow()),
                models.ReplicationRule.child_rule_id == null()))
```

To avoid the duplicate `models.ReplicationRule.child_rule_id == null()`

rdimaio (approved):

lib/rucio/core/request.py
dchristidis: It all depends on whether we want to put more emphasis on code readability or query simplicity. The use of functions in SQLAlchemy for logical operators effectively results in prefix notation, which I would assume most are not familiar with. As such, I believe more than two levels of nesting should be avoided.

Otherwise, the simplest form is with `COALESCE()`, which I like, but I had received a comment in the past that it’s not readable.

```python
.where(
    or_(models.ReplicationRule.child_rule_id != null(),
        and_(models.ReplicationRule.child_rule_id == null(),
             coalesce(models.ReplicationRule.expires_at, now) >= now))
)
```

I was also looking into a `CASE` expression which guarantees short-circuit evaluation but becomes even more difficult to read.

PR #7385, opened 2025-02-05
Protocols: Use supplied timeout instead of default in webdav #7351

### LINKED ISSUE #7351: Setting transfer timeout when downloading has no effect
### Description

The transfer timeout that one can specify for the download subcommand (as well as the transfer speed timeout and the global --timeout option) have no effect on at least the webdav protocol. In the `get` method, the transfer_timeout variable is unused:
https://github.com/rucio/rucio/blob/dfcd980f4f10720af38ea1570a6015b2478a5408/lib/rucio/rse/protocols/webdav.py#L257

Instead self.timeout is used:
https://github.com/rucio/rucio/blob/dfcd980f4f10720af38ea1570a6015b2478a5408/lib/rucio/rse/protocols/webdav.py#L269

This one is set earlier to a default of 300:
https://github.com/rucio/rucio/blob/dfcd980f4f10720af38ea1570a6015b2478a5408/lib/rucio/rse/protocols/webdav.py#L201

This requires of course that connect is called without an argument, which it does however in the actual download client:
https://github.com/rucio/rucio/blob/dfcd980f4f10720af38ea1570a6015b2478a5408/lib/rucio/client/downloadclient.py#L693

This means that if I want to only download a small file and several RSEs don't answer it gives me a hardcoded timeout of 300 seconds for each RSE that fails (2 times that is). It would be nice if one could properly configure this timeout value. I didn't check how it is for other protocols, but they same problem might exist.

### Steps to reproduce

Will omit since I already pointed at the code

### Rucio Version

I am using 34.1.0 but clearly this problem is also present in the current master branch

### Additional Information

_No response_

voetberg:
I was unsure how to test this one/if it is already being tested somehow. Please let me know if it needs some addition.

### DIFF
--- a/lib/rucio/rse/protocols/webdav.py
+++ b/lib/rucio/rse/protocols/webdav.py
@@ -259,14 +259,16 @@ class Default(protocol.RSEProtocol):
 
             :param pfn: Physical file name of requested file
             :param dest: Name and path of the files when stored at the client
-            :param transfer_timeout: Transfer timeout (in seconds) - dummy
+            :param transfer_timeout: Transfer timeout (in seconds)
 
             :raises DestinationNotAccessible, ServiceUnavailable, SourceNotFound, RSEAccessDenied
         """
         path = self.path2pfn(pfn)
         chunksize = 1024
+        transfer_timeout = self.timeout if transfer_timeout is None else transfer_timeout
+
         try:
-            result = self.session.get(path, verify=False, stream=True, timeout=self.timeout, cert=self.cert)
+            result = self.session.get(path, verify=False, stream=True, timeout=transfer_timeout, cert=self.cert)
             if result and result.status_code in [200, ]:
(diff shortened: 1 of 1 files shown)


rdimaio (commented):
In the `put()` docstring, there's still a `dummy` which should be removed:

```
:param transfer_timeout Transfer timeout (in seconds) - dummy
```

(it doesn't let me comment directly on that line cause it's not close enough to the diff)

voetberg:
> In the `put()` docstring, there's still a `dummy` which should be removed:
> 
> ```
> :param transfer_timeout Transfer timeout (in seconds) - dummy
> ```
> 
> (it doesn't let me comment directly on that line cause it's not close enough to the diff)

Corrected! Thanks for catching that

rdimaio (approved):

PR #7388, opened 2025-02-07
Policies: standardise permission policy check logic #7206

### LINKED ISSUE #7206: Standardize permission policy check logic
However, checking the code, it seems this is not uses consistently.

`lib/rucio/core/permission/__init__.py` has in fact complex logic to determine the policy, but in several places in the code, a simple `config_get("policy", "permission")` is done instead.

_Originally posted by @maxnoe in https://github.com/rucio/documentation/issues/391#issuecomment-2488719810_

See also: https://github.com/rucio/documentation/pull/391#issuecomment-2488721184

### LINKED ISSUE #7332: ATLAS policy is set when no permission option is defined in cfg under policy
https://github.com/rucio/rucio/blob/master/lib/rucio/common/policy.py#L36-L37

In this case, I think we should either error out or set the policy to `def`

jamesp-epcc:
This is my first attempt to make the policy logic a bit more sensible and consistent:

1. Removes direct queries of the config file, everything calls `get_policy` instead so that the check is in just one place.
2. `get_policy` now checks a couple of possible config keys and an environment variable (which was already used by the tests) instead of just one config key.
3. The default policy is now 'def' instead of 'atlas'.
4. The tests were using various different method for skipping ATLAS-specific test cases. I've tried to tidy this up a bit, at least enough to make the tests pass for this PR, but I am not an ATLAS person so not sure if this is the most sensible way.

Also addresses https://github.com/rucio/rucio/issues/7332

### DIFF
--- a/lib/rucio/common/policy.py
+++ b/lib/rucio/common/policy.py
@@ -13,10 +13,11 @@
 # limitations under the License.
 
 import json
+import logging
 import os
 from configparser import NoOptionError, NoSectionError
 from functools import wraps
-from typing import Any
+from typing import TYPE_CHECKING, Any
 
 from dogpile.cache import make_region
 from dogpile.cache.api import NoValue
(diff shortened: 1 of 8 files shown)


lib/rucio/transfertool/fts3.py
```diff
@@ -1369,10 +1370,7 @@ def set_se_config(
 
         params_dict = {storage_element: {'operations': {}, 'se_info': {}}}
         if staging is not None:
-            try:
-                policy = config_get('policy', 'permission')
-            except Exception:
-                self.logger(logging.WARNING, 'Could not get policy from config')
+            policy = get_policy()
```
rdimaio, line 1373: I think this is fine, but worth pointing out that the behaviour here will change when policy is not found:
- beforehand, we'd log a warning and error out on the next line `params_dict[storage_element]['operations'] = {policy: {'staging': staging}}`, because `policy` would be undefined
- now, we fall back to `def` and no warning is logged, and execution continues

lib/rucio/common/policy.py
```diff
@@ -34,7 +34,11 @@ def get_policy() -> str:
         try:
             policy = config_get('policy', 'permission')
         except (NoOptionError, NoSectionError):
-            policy = 'atlas'
+            try:
+                policy = config_get('permission', 'policy')
+            except (NoOptionError, NoSectionError):
+                policy = 'def'
```
rdimaio, line 44: It could be worth adding a warning log here to mention that the policy wasn't found and we are falling back to `def`

rdimaio (dismissed):

lib/rucio/common/policy.py
jamesp-epcc: I think that's a good idea. I'll add the warning.

lib/rucio/transfertool/fts3.py
jamesp-epcc: I've added a warning to `get_policy` as discussed

rdimaio (approved):

PR #7400, opened 2025-02-12
Re-enable libtorrent

### LINKED ISSUE #6853: libtorrent dependency problematic?
### Description

I was installing the dev dependencies in a fresh virtual env, which happened to use python 3.11.

This failed, as rucio has `libtorrent` in its server dependencies and `libtorrent` does not have a wheel for 3.11 and doesn't provide a source distribution.

This seems very concerning: no source distribution is available and the last update to the wheels was last year.

### Steps to reproduce

`pip install -r requirements/requirements.dev.txt` using python > 3.10

### Rucio Version

_No response_

### Additional Information

_No response_

rdimaio:
Fix #6853

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -12,7 +12,7 @@ jobs:
 
       strategy:
           matrix:
-            python-version: ["pypy3.9", "pypy3.10", "3.9", "3.10", "3.11", "3.12"]
+            python-version: ["3.9", "3.10", "3.11", "3.12"]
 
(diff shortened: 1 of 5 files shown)


rdimaio:
Looks like `libtorrent` has no distributions for `pypy`: https://pypi.org/project/libtorrent/2.0.11/#files

I think we can remove the `pypy` test cases to support for libtorrent again. It's a bit annoying, but we can assume most people use CPython anyway.

bari12 (approved):

PR #7402, opened 2025-02-12
ci: GitHub Actions concurrency limit

### LINKED ISSUE #7401: Allow only one concurrent GitHub Action per PR
Currently actions are triggered after every push to a branch in a PR, if another commit is pushed, both the old and new actions run concurrently.

GitHub actions can be configured to only allow one concurrent run, the latest one. If there are running actions and new commits are pushed, the old runs are cancelled in favour of the new ones.

This would lower the load on the CI infrastructure.

lobis:
Closes #7401

### DIFF
--- a/.github/workflows/autotest.yml
+++ b/.github/workflows/autotest.yml
@@ -6,6 +6,10 @@ on:
   schedule:
     - cron: '0 3 * * *'
 
+concurrency:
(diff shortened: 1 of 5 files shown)


lobis:
I added a `style` commit after the initial commit to test that the initial actions were cancelled as expected.

rdimaio (approved):

PR #7403, opened 2025-02-13
Simplify calls to config_get* functions

### LINKED ISSUE #7395: Simplify calls to config_get* functions
### Description

Many calls to this family of functions supply arguments for `default` and `raise_exception`. When the former is set to a non-`None` value, the latter is unnecessary.

https://github.com/rucio/rucio/blob/37cc33b21f2ca2d99bffd2da7e230b1e50fce1f2/lib/rucio/common/config.py#L216-L220

`__config_get_table()` is implemented similarly.

### Motivation

Simplify the code.

### Change

1. Improve the documentation strings of the functions to explain their behaviour.
2. For all calls to those functions:
    i. If the `default` parameter is being set to `None`, then remove it.
    ii. If the `raise_exception` parameter is being set to True, then remove it.
    iii. If the `defallt` parameter is being set to any value other than `None` and the `raise_exception` parameter is being set to False, then remove the latter.

Examples:

```
config_get_bool('common', 'multi_vo', raise_exception=False, default=False) → config_get_bool('common', 'multi_vo', default=False)
config_get('core', 'geoip_download_url', raise_exception=False, default=None) → config_get('core', 'geoip_download_url, raise_exception=False)
```

aryanbhosale:
Fixes #7395 

Removed raise_exception for non-none default values and updated the docstrings


rdimaio (commented):
Thanks for the PR. Could you also address points 2.i and 2.ii in the issue (https://github.com/rucio/rucio/issues/7395):
- i. If the default parameter is being set to None, then remove it.
- ii. If the raise_exception parameter is being set to True, then remove it.

For the docstrings, could you please:
- remove the types mentioned (e.g. `(str)`)? We already have type annotations for those, so we want to avoid redundant information.
- remove the defaults (not needed in the docstring, they can easily be seen from the function signature)

I think the docstrings in this state might be a bit too busy, but I'm not sure what @dchristidis was looking for with regards to docstrings improvements here (I assume it was mainly with regards to the behaviour of `raise_exception` and `default`, rather than fully rewriting the docstrings like this)

aryanbhosale:
> Thanks for the PR. Could you also address points 2.i and 2.ii in the issue (#7395):
> 
> * i. If the default parameter is being set to None, then remove it.
> * ii. If the raise_exception parameter is being set to True, then remove it.
> 
> For the docstrings, could you please:
> 
> * remove the types mentioned (e.g. `(str)`)? We already have type annotations for those, so we want to avoid redundant information.
> * remove the defaults (not needed in the docstring, they can easily be seen from the function signature)
> 
> I think the docstrings in this state might be a bit too busy, but I'm not sure what @dchristidis was looking for with regards to docstrings improvements here (I assume it was mainly with regards to the behaviour of `raise_exception` and `default`, rather than fully rewriting the docstrings like this)

Hey, thanks for the reviews, I have made changes to almost every file that contained 2i and 2ii as mentioned in the original issue;
and i've made the changes suggested by you in the docstrings too.

Am I on the right path here?

lib/rucio/common/stomp_utils.py
```diff
@@ -202,13 +202,13 @@ def _parse_config(self, config_section: str) -> StompConfig:
...
-        nonssl_port = config_get_int(config_section, 'nonssl_port', default=0, raise_exception=False)
-        timeout = config_get_float(config_section, 'timeout', default=None, raise_exception=False)
-        heartbeats = tuple(config_get_list(config_section, 'heartbeats', default=[0., 0.], raise_exception=False))
+        key_file = config_get(config_section, 'ssl_key_file', raise_exception=False)
+        cert_file = config_get(config_section, 'ssl_cert_file', raise_exception=False)
+        username = config_get(config_section, 'username', raise_exception=False)
+        password = config_get(config_section, 'password', raise_exception=False)
+        nonssl_port = config_get_int(config_section, 'nonssl_port', default=0)
+        timeout = config_get_float(config_section, 'timeout', raise_exception=False)
+        heartbeats = tuple(config_get_list(config_section, 'heartbeats', default=[0., 0.])
```
rdimaio: this should fix the syntax errors mentioned in the CI

```suggestion
        heartbeats = tuple(config_get_list(config_section, 'heartbeats', default=[0., 0.]))
```

lib/rucio/common/config.py
```diff
@@ -166,30 +166,30 @@ def config_get(
...
+    the configuration table is queried.
+
+    Parameters:
+      section: the named section.
+      option: the named option.
+      raise_exception: whether to raise an exception (NoOptionError, NoSectionError, or RuntimeError)
+                       if the option is not found. When a non-None default is provided, this parameter is redundant.
+      default: the value to return if the option is not found. If no value is desired, omit it.
+      clean_cached: clears the cached configuration instance if no value is found.
+      check_config_table: indicates whether to check the configuration table if the option is missing.
```
rdimaio: The meaning here changed. The old one says "if not set, avoid looking at config table even if it is called from server/daemon", while this says something else.

lib/rucio/common/config.py
```diff
@@ -166,30 +166,30 @@ def config_get(
...
+      session: the database session to use (only applicable for server/daemon calls).
```
rdimaio: "Only used if not found in config file" is missing.

lib/rucio/common/config.py
```diff
@@ -166,30 +166,30 @@ def config_get(
...
+      use_cache: whether to use caching for server/daemon calls.
```
rdimaio: "Only used if not found in config file" is missing.

lib/rucio/common/config.py
```diff
@@ -166,30 +166,30 @@ def config_get(
...
+      expiration_time: the cache expiration time in seconds.
```
rdimaio: "Only used if not found in config file" is missing.

lib/rucio/common/config.py
```diff
@@ -166,30 +166,30 @@ def config_get(
...
+      convert_type_fnc: a function used to convert the retrieved configuration value.
```
rdimaio: Old version was clearer, as it clarifies that the retrieved configuration value is always a string, and that's why a conversion function might be needed for certain cases.

rdimaio (commented):
LGTM for the functional changes, except for the syntax error indicated, which should fix the CI.

For the docstrings, I would refrain from modifying them to this extent. Based on the issue description, I would keep them the same as the original (including the original `:param` formatting) and just add infomration on the `raise_exception` and `default` behaviour

aryanbhosale:
Ive made the changes in the docstring and the CI passing change youd suggested. Is this fine now?

lib/rucio/common/config.py
```diff
@@ -171,25 +171,27 @@ def config_get(
...
     :param clean_cached: Deletes the cached config singleton instance if no config value is found
-    :param check_config_table: if not set, avoid looking at config table even if it is called from server/daemon
-    :param session: The database session in use. Only used if not found in config file and if it is called from
-                    server/daemon
-    :param use_cache: Boolean if the cache should be used. Only used if not found in config file and if it is called
-                      from server/daemon
-    :param expiration_time: Time after that the cached value gets ignored. Only used if not found in config file and if
-                            it is called from server/daemon
+    :param check_config_table: if not set, avoid looking at config table even if it is called from server/daemon;
+                               only used if not found in the config file.
```
rdimaio, line 184: The original docstring doesn't mention 'only used if not found in the config file.' for `check_config_table`, so this should be removed

lib/rucio/common/config.py
```diff
@@ -171,25 +171,27 @@ def config_get(
...
+    :param session: The database session in use; only used if not found in the config file and if called from server/daemon
+    :param use_cache: Boolean if the cache should be used; only used if not found in the config file and if called from server/daemon
+    :param expiration_time: Time after which the cached value gets ignored; only used if not found in the config file and if called from server/daemon
```
rdimaio, line 187: in general, these should just be kept the same way they were before, no need for any changes here

lib/rucio/common/config.py
```diff
@@ -171,25 +171,27 @@ def config_get(
...
     :param convert_type_fnc: A function used to parse the string config value into the desired destination type
 
     :returns: the configuration value.
 
-    :raises NoOptionError
-    :raises NoSectionError
-    :raises RuntimeError
+    :raises NoOptionError: if the option is not found and no default is provided.
+    :raises NoSectionError: if the section is not found and no default is provided.
+    :raises RuntimeError: if there is an error accessing the configuration.
```
rdimaio, line 194: sidenote: It actually seems like `RuntimeError` is never raised in this function, but that's an issue with the original docstring in the first place so you can ignore this. Later on we will refactor docstrings and use some linter to keep exceptions in sync with the function body (or remove exceptions from the docstrings)

rdimaio (commented):
LGTM overall, just small notes regarding the docstrings.

You can ignore the Pyright error in the CI, it will be fixed by https://github.com/rucio/rucio/pull/7420

Could you also rebase your changes so that:
- all your docstring changes are in one commit
- all your functional changes are either in a single commit, or in three separate commits, each addressing the points 2.i, 2.ii and 2.iii mentioned here (https://github.com/rucio/rucio/issues/7395)? I can understand it'd be a bit annoying to rebase them into three separate commits, so feel free to just have them all in a single commit; the main priority is to keep the docstring changes and the functional changes in two separate commits.

Please also check the contributing guide for a reference on the commit message format: https://rucio.cern.ch/documentation/contributing/#4-commit-your-changes

Thank you for your patience and thanks again for contributing.

Issue #7310, opened 2025-01-13
Rework OIDC user authentication


bari12:
### Description

The existing OIDC user authentication to Rucio needs to be re-factored. The CiLogon use-case needs to be implemented.

panta-123:
It would be good time for following:
1. jwkest : is not maintained anymore (https://github.com/IdentityPython/pyjwkest/issues/101). Move to pyjwt  (already available on rucio-server)
2. simplify the code structure.

bari12:
Hi @panta-123 @jamesp-epcc 
A bit more information of what should be done for this OIDC rework. I think we should create individual tickets for this, but for now I am just collecting it here.

1. Remove the oauthmanager daemon and every code which relies on it https://github.com/rucio/rucio/tree/master/lib/rucio/daemons/oauthmanager 
such as the `refresh_jwt_tokens` method https://github.com/rucio/rucio/blob/0952f92726b587093ff47436b6e0af4832c677af/lib/rucio/core/oidc.py#L1097
`__refresh_token_oidc` https://github.com/rucio/rucio/blob/0952f92726b587093ff47436b6e0af4832c677af/lib/rucio/core/oidc.py#L1153C5-L1153C25
`__get_admin_token_oidc` etc.
Please check with @dchristidis if you are in doubt if something needs to be removed/kept, I would guess the large majority of this code (in `core/oidc.py`) can be removed.

2. Rework the oidc client authentication
https://github.com/rucio/rucio/blob/0952f92726b587093ff47436b6e0af4832c677af/lib/rucio/client/baseclient.py#L583
There only the `oidc_auto` mode should stay. I am not sure about the polling sub-mode. The other one (Where you put your IDP password in the rucio config) must be removed.

3. Update documentation concerning all these removals.

panta-123:
@bari12 ,
 I am doing following (nearly done, writing pytest now): 
   1. removing depecncy on : pyoidc and jwkest.  Adding/using only pyJWT.
   2. Removed all code that is there for user token exchange for transfer.  -> relates to daemon removal.
   3. Integrated deletion of expired oauth request and expired TOKEN into the token workflow.
   
With 2 and 3 , we can remove oauthmanager daemon . (Do you want to create new tiket for daemon removal? can be done here as well)

As of your point 2:
The oidc_auto should be removed and is removed in my work: this is the one that uses username and password from IDP (which is not an oidc flow)
[...]

Issue #6406, opened 2023-12-06
Make new token support Multi-VO compatible


dchristidis:
### Description

The new-style token support introduced with #6183, #6187, #6390 and #6401 is not compatible with mult-VO Rucio instances.

### Change

* Load the entire `idpsecrets.json` file. Treat the nicknames, which in the past were chosen freely, as the name of the VOs.
* Add a `vo` parameter to the `request_token()` function. Stop using the `oidc.admin_issuer` configuration. Adapt the cache module as needed.
* Adapt callers of `request_token()` to specify the VO, if necessary.

thysk:
would this change as described not limit VOs to a single IDP?
Which for most communities would be fine I think
but for instance, testing with Dteam may want to be configured to many IDPs at once

dchristidis:
The issuer of a token is currently used to identify the VO. The storages use it to map a token to a particular location (e.g. mapping tokens from https://atlas-auth.web.cern.ch/ to `/eos/atlas/`). How would the lack of a one-to-one relation work in production?

bari12:
Thank you for raising this. We have reviewed this issue as part of our regular issue review.

We see value in this request, but we do not have the capacity to work on it within the next twelve months.

In line with our [issue management guideline](https://rucio.cern/documentation/developer/issue_management), we are labelling it `wishlist` and closing it. This does not mean it is forgotten: wishlist items are revisited regularly during component reviews and roadmap planning, and this issue can be re-opened and added back to the pipeline as soon as interest and capacity align.

PR #7406, opened 2025-02-13
oidc: rework oidc flow #7310

panta-123:
Rework:
1. Remove oic libary dependency and pyoidc stuff. replace it with pyjwt.
2. Remove auto for OIDC auth as it doesn't adheres to standard.
3. Removed all code used  for oauthmanager daemon and daemon it self.
4. The expired oauth-request and token entry deleted during new creation.
5. Have to modify all the test as base libary and structure are changed.


panta-123:
Somthing wrong with requirement.txt , Let me double check.

panta-123:
There is comment : https://github.com/rucio/documentation/pull/480#discussion_r1961311179
I have to make new commit to simply this. Let me do that.

rcarpa:
I would appreciate if the compatibility with keycloak is maintained in tests (for example, by using it for the user authentication flow instead of the mock client that you introduced). I plan to rely on it for finalising bittorrent support once qbittorrent 5.0+ is released by default in ubuntu.  Alternatively, if the support for keycloak is dropped on purpose from rucio and there are absolutely no plans to keep compatibility with it, please cleanup all tests using it.

etc/docker/dev/rucio/idpsecrets.json
```diff
@@ -1,12 +1,18 @@
...
+      "user_auth_client": [
+          {
+              "issuer": "https://mock-oidc-provider",
+              "client_id": "mock-client-id",
+              "client_secret": "secret",
+              "redirect_uris": "https://rucio/auth/oidc_code",
+              "issuer_nickname": "example_issuer"
+          }
+      ],
+      "client_credential_client": {
```
rcarpa, line 12: Is the authentication flow type really the best way to split the oidc configuration? I can imagine, in the future, requirements when 2 different parts of rucio use two different issuers for the same authentication flow. For example: 1) the current workflow of accessing a file on the RSE may use one issuer. While 2) connecting to another  external resource will also use the client_credentials flow, but with a different issuer. Examples: authenticated communication between rucio and SENSE; or fetching a geoip database with authentication; or authentication to Globus transfertool; why not to qbittorrent? 🤓 
Considering that this PR changes things in a non-backward compatible way, I think it’s a good opportunity to design it in a more forward-looking way. 
@bari12 , @dchristidis , what do you think about this proposal? Seems like a very easy change to do now; but difficult to change later, when people will start using it. 
I can even easily imagine the requirement for two different RSEs to use 2 different issuers popping up in the very close future.

panta-123:
> I would appreciate if the compatibility with keycloak is maintained in tests (for example, by using it for the user authentication flow instead of the mock client that you introduced). I plan to rely on it for finalising bittorrent support once qbittorrent 5.0+ is released by default in ubuntu. Alternatively, if the support for keycloak is dropped on purpose from rucio and there are absolutely no plans to keep compatibility with it, please cleanup all tests using it.

Keycloak compatibility is not dropped . Might have missed in the test , I will look into it.

etc/docker/dev/rucio/idpsecrets.json
dchristidis: For the time being, we are working under the assumption that each VO will map to exactly one issuer. We have not received any requests to the contrary. So the first level being the name of the VO or `def` in case of a single-VO Rucio instance is in order (see also #6406).

The second level being the type of flow would not have been my first choice, but I have no objections based on current needs.

Sadly, nothing about token development is stable. :disappointed:  We should not be afraid to change the format of the configuration file in major releases, provided that those changes are well-advertised in the release notes.

panta-123: In future, We can evolve the config to accomodate these examples mentioned in the comment.
Note that this config is extendable with very few changes.
1. If needed issuer per rse, we can add  issuer_nickname to `client_credential_client` and do the same as user_auth_client, with very few changes.
2. We can add other subfield to vo/def field to add any other client, can be any name and do the same. Also we can add 1 to this as well.

So lets make the changes as the request comes in but not now as the new request can be accomodated easily.

panta-123:
> > I would appreciate if the compatibility with keycloak is maintained in tests (for example, by using it for the user authentication flow instead of the mock client that you introduced). I plan to rely on it for finalising bittorrent support once qbittorrent 5.0+ is released by default in ubuntu. Alternatively, if the support for keycloak is dropped on purpose from rucio and there are absolutely no plans to keep compatibility with it, please cleanup all tests using it.
> 
> Keycloak compatibility is not dropped . Might have missed in the test , I will look into it.

We are aiming for 37 release. If there is no objection then tests addition can be done in later patch release

dchristidis (commented):
In its current state, this PR is difficult to review. The first commit attempts to change too many things at once, and does so with little guidance (the commit message body is empty). The most noteworthy offence is that it changes the structure of the configuration file with no warning; such changes should be well-advertised, so that they can be described in the release notes.

panta-123:
> In its current state, this PR is difficult to review. The first commit attempts to change too many things at once, and does so with little guidance (the commit message body is empty). The most noteworthy offence is that it changes the structure of the configuration file with no warning; such changes should be well-advertised, so that they can be described in the release notes.

The config changes is well documented here. https://github.com/rucio/documentation/pull/480
And I can sumarize these for release note as well may be with little guidence on how it is done.

etc/docker/dev/rucio/idpsecrets.json
rcarpa: Making "issuer_nickname" available for client_credential_client seems like a good idea. First of all, it makes it more future proof. Second, it makes handling of the two sections of the configuration file (idpsecrets) more uniform (both client_credential_client and user_auth_client will be a list), so the code can be simplified to always expect a list.

panta-123: @rcarpa , I am willing to incorporate this to the PR. But I am limbo right now on what do with this PR regarding the latest statement on the PR.

rcarpa: I agree with Dimitrios that this PR is difficult to review. Imagine yourself having to read through 6000 lines of changes done by somebody else. With very little to explain why some of those changes were even needed. It would be difficult for anybody. But, as the comments are mostly about the structure of the PR and not about the high-level change, the way forward seems pretty clear: "git rebase".... and split the change in multiple small commits with clear explanation in each commit.

panta-123: okay changing this to draft.
Will divide into two PR.
1. remove oauthmanager and auto from oidc flow. Keep others same.
2. Change dependency from oic and pyoidc and move to new structure flow.

rcarpa: Having two dependent PR is more complicated to manage if one of them needs to be merged first. Why not just have one PR with two well-done commits ?

panta-123: Okay. 
Will do following commit in order:
1. remove oauth-manager daemon and any refrence.
2. remove auto and auto refresh liftime from client and core.
3. change idsecret structure and  add class to read and validate the the secrets
4.  remove oic and pyoidc and chnage any refrence to it.
5. Remove old test and add new test and any remnant changes.

panta-123:
chunk commit in different PR: https://github.com/rucio/rucio/pull/7705

PR #7411, opened 2025-02-17
Client: refactor rich client `get_pager` to avoid hanging

### LINKED ISSUE #7410: Unable to run `rucio` command from PyCharm IDE
I am unable to run the `rucio` command via the PyCharm IDE, it appears to hang indefinitely.

lobis:
Fixes https://github.com/rucio/rucio/issues/7410.

Without this timeout it hangs indefinitely in PyCharm.

I put a `0.1` second timeout which should be almost imperceptible (it will only be triggered when `less` is not found, instead of hanging indefinitely) and it should not change behaviour for the case where `less` is indeed in the system. `0.1` seconds should be plenty of time to find it in the system.

### DIFF
--- a/lib/rucio/client/richclient.py
+++ b/lib/rucio/client/richclient.py
@@ -15,7 +15,7 @@
 import logging
 import pydoc
 import re
-import subprocess
+import shutil
 import sys
 from datetime import datetime
 from typing import TYPE_CHECKING, Any, Optional, Union
@@ -309,16 +309,9 @@ def get_pager() -> 'Callable[[str], None]':
     :returns: pager
     """
     default_pager = 'less'
-    try:
-        result = subprocess.run([default_pager], stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
-        less_available = result.returncode == 0
-    except Exception:
-        less_available = False
+    # Attempt to use the default pager if available.
+    if shutil.which(default_pager) is not None:
+        return lambda text: pydoc.pipepager(text, f'{default_pager} -FRSXKM')
 
-    def less_pager_function(text: str) -> None:
-        """Use the 'less' pager with -FRSXKM options."""
-        pydoc.pipepager(text, f'{default_pager} -FRSXKM')
-
-    if less_available:
-        return less_pager_function
+    # Fall back to pydoc.pager if the default pager is not available.
     return pydoc.pager


rdimaio (commented):
This looks okay to me, but I think there might potentially be situations where `less` does not terminate within 0.1 seconds (maybe on very slow devices) https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate

I'd be more in favour of rewriting this function. We want to check that `less` is available, so we could just do:
https://stackoverflow.com/questions/7522712/how-can-i-check-if-a-command-exists-in-a-shell-script/7522866#7522866

lobis:
> This looks okay to me, but I think there might potentially be situations where `less` does not terminate within 0.1 seconds (maybe on very slow devices) https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate
> 
> I'd be more in favour of rewriting this function. We want to check that `less` is available, so we could just do: https://stackoverflow.com/questions/7522712/how-can-i-check-if-a-command-exists-in-a-shell-script/7522866#7522866

I agree that this is not the best implementation.

I rewrote the function to use `shutil` (built in) to find the utility. This also avoids having to run a try/except block. I think it is cleaner this way.

lib/rucio/client/richclient.py
```diff
@@ -15,6 +15,7 @@
 import logging
 import pydoc
 import re
+import shutil
 import subprocess
```
rdimaio: This is unused and should be removed

```
lib/rucio/client/richclient.py:19:1: F401 'subprocess' imported but unused
```

lib/rucio/client/richclient.py
```diff
@@ -309,16 +310,14 @@ def get_pager() -> 'Callable[[str], None]':
...
+        return pydoc.pager
 
     def less_pager_function(text: str) -> None:
         """Use the 'less' pager with -FRSXKM options."""
         pydoc.pipepager(text, f'{default_pager} -FRSXKM')
 
-    if less_available:
-        return less_pager_function
-    return pydoc.pager
+    return less_pager_function
```
rdimaio: I'd personally rewrite it like this for clarity:

```python
    # if `less` is available, set default options
    if shutil.which('less') is not None:
        return lambda text: pydoc.pipepager(text, 'less -FRSXKM')
    # fall back to available pager
    else:
        return pydoc.pager
```

lobis: Agree, refactored!

rdimaio (commented):
Could you please also rename the PR to reflect the new change?

lobis:
Should I address the ruff warnings? (This is new right?)

lib/rucio/client/richclient.py
```diff
@@ -15,10 +15,10 @@
 import logging
 import pydoc
 import re
-import subprocess
+import shutil
 import sys
 from datetime import datetime
-from typing import TYPE_CHECKING, Any, Optional, Union
+from typing import TYPE_CHECKING, Any, Optional, Union, Callable
```
rdimaio:
```suggestion
from typing import TYPE_CHECKING, Any, Optional, Union
```

lib/rucio/client/richclient.py
```diff
@@ -302,23 +302,16 @@ def get_cli_config() -> str:
     return cli_type
 
 
-def get_pager() -> 'Callable[[str], None]':
+def get_pager() -> Callable[[str], None]:
```
rdimaio:
```suggestion
def get_pager() -> 'Callable[[str], None]':
```

rdimaio (approved):

Issue #7232, opened 2024-12-12
Rich/New CLI: Ensure all ‘list’ and ‘show’ commands have an option to output in CSV format


dchristidis:
### Description

One such example is `rucio list-rses` / `rucio rse list`. It is not uncommon to want to loop over a list. For example:

```shell
for rse in $(rucio list-rses --rses=type=LOCALGROUPDISK; do
    rucio-admin rse set-attribute --rse=$rse --key=greedyDeletion --value=True
done
```

The above works with tabulate, but not with Rich. Some commands already offer a `--csv` command-line option (e.g. `rucio list-rules`).

dchristidis:
Repeating my comment in https://github.com/rucio/rucio/pull/7421#discussion_r1967115141, we can call this CSV output but it’s probably to be safer to use a tab as a separator. There are multiple fields where a comma is likely to be part of the value (e.g. identities, RSE attributes, rule comments).

We could use the term TSV, but it’s not common.

voetberg:
I much prefer "tsv" in this case, I think it is more apt if we do not want to use commas in particular. Would it also make sense to make a main argument (similar to the "--pager" argument) to control the separator? Almost what you mentioned here https://github.com/rucio/rucio/pull/7421#discussion_r1967115141

dchristidis:
On the whole, I find it prudent to not add new options unless we have evidence that they are required.

voetberg:
Summary of discussion on related PRs/offline dicussion: 

* They should use a tab deliminator ('\t') which is a global variable, the `--csv` name is more familiar though so will be used for uniformity
* They should not have headers
* A single row should represent a entry (so --short options result in a single name per row) - general format is something like 
```print(*[CSV_SEPARATOR.join(row) for row in results], sep='\n)```

Which would result in 

```
item1a   item1b ...
item2a   item2b ...
...
```

bziemons:
Shouldn't this be re-assigned as a child of issue #7960 ? @voetberg can I help somehow with the backlog other than reviewing?

voetberg:
> Shouldn't this be re-assigned as a child of issue [#7960](https://github.com/rucio/rucio/issues/7960) ? [@voetberg](https://github.com/voetberg) can I help somehow with the backlog other than reviewing?

It should be reassigned - thanks for the reminder. If you would like to take a few of these issues please go ahead! I have no great attachment to any of them.

PR #7421, opened 2025-02-18
CSV Option for rule show and rule history 

### LINKED ISSUE #7419: Client: Add `--csv` option to rucio/info_rule, list_rules, list_rules_history
### Description

Add `csv` as an option to facilitate iterating over the result of a cli command. See parent issue for further description. 

### Motivation

_No response_

### Change

_No response_

voetberg:
Closes #7419

### DIFF
--- a/lib/rucio/client/commands/bin_legacy/rucio.py
+++ b/lib/rucio/client/commands/bin_legacy/rucio.py
@@ -1436,54 +1436,76 @@ def info_rule(args, client, logger, console, spinner):
     Retrieve information about a rule.
     """
 
-    if cli_config == 'rich':
+    if (cli_config == 'rich') and (not args.csv):
         spinner.update(status='Fetching rule info')
         spinner.start()
 
     if args.examine:
         output = []
         analysis = client.examine_replication_rule(rule_id=args.rule_id)
-        if cli_config == 'rich':
-            keyword_styles = {**CLITheme.BOOLEAN, **CLITheme.DID_TYPE, **CLITheme.RULE_STATE}
-            rule_status = " ".join([f'[{keyword_styles.get(word, "default")}]{word}[/]' for word in analysis['rule_error'].split()])
-            output.append(f'Status of the replication rule: {rule_status}')
+
+        if analysis['transfers']:
+            for transfer in analysis['transfers']:
+                output.append(
+                    [
+                        transfer["scope"],
+                        transfer["name"],
+                        {
+                            "RSE": str(transfer["rse"]),
+                            "Attempts": str(transfer["attempts"]),
+                            "Last retry": str(transfer["last_time"]),
+                            "Last error": str(transfer["last_error"]),
+                            'Available sources:': ', '.join([source[0] for source in transfer['sources'] if source[1]]),
+                            'Blocklisted sources:': ', '.join([source[0] for source in transfer['sources'] if not source[1]])
+                        },
+                    ]
+                )
+
+        if args.csv:
             if analysis['transfers']:
-                output.append('[b]STUCK Requests:[/]')
-                for transfer in analysis['transfers']:
-                    output.append(Padding.indent(Text(f"{transfer['scope']}:{transfer['name']}", style=CLITheme.SUBHEADER_HIGHLIGHT), 2))
-                    table_data = [['RSE:', str(transfer['rse'])],
-                                  ['Attempts:', str(transfer['attempts'])],
-                                  ['Last retry:', str(transfer['last_time'])],
-                                  ['Last error:', str(transfer['last_source'])],
-                                  ['Available sources:', ', '.join([source[0] for source in transfer['sources'] if source[1]])],
-                                  ['Blocklisted sources:', ', '.join([source[0] for source in transfer['sources'] if not source[1]])]]
+                for row in output:
+                    did = f"{row[0]}:{row[1]}"
+                    print(did, *row[2].values(), sep=',')
+
+        elif cli_config == 'rich':
+            rich_output = []
+            keyword_styles = {**CLITheme.BOOLEAN, **CLITheme.DID_TYPE, **CLITheme.RULE_STATE}
+
+            rule_status = " ".join([f'[{keyword_styles.get(word, "default")}]{word}[/]' for word in analysis['rule_error'].split()])
+            rich_output.append(f"Status of the replication rule: {rule_status}")
+            if analysis['transfers']:
+                rich_output.append('[b]STUCK Requests:[/]')
+                for transfer in output:
+                    rich_output.append(Padding.indent(Text(f"{transfer[0]}:{transfer[1]}", style=CLITheme.SUBHEADER_HIGHLIGHT), 2))
+                    table_data = [[key, value] for key, value in transfer[2].items()]
                     table = generate_table(table_data, row_styles=['none'], col_alignments=['left', 'left'])
-                    output.append(Padding.indent(table, 2))
+                    rich_output.append(Padding.indent(table, 2))
 
             spinner.stop()
-            print_output(*output, console=console, no_pager=args.no_pager)
+            print_output(*rich_output, console=console, no_pager=args.no_pager)
+
         else:
-            analysis = client.examine_replication_rule(rule_id=args.rule_id)
             print('Status of the replication rule: %s' % analysis['rule_error'])
             if analysis['transfers']:
                 print('STUCK Requests:')
-                for transfer in analysis['transfers']:
-                    print('  %s:%s' % (transfer['scope'], transfer['name']))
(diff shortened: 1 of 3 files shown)


rdimaio (commented):
I'm not familiar with the pre-existing client CSV functionality, but would it make sense to change the `--csv` argument so that a file path can be passed to it, so that the user can specify the file path and name?

Additionally, for testing, I think it would be great if we could also read from the `csv` generated to ensure that the output written to the csv corresponds to what we expected

voetberg:
> I'm not familiar with the pre-existing client CSV functionality, but would it make sense to change the --csv argument so that a file path can be passed to it, so that the user can specify the file path and name? 

The existing functionality is to display results as a csv (I think with the exception of declaring suspicious replicas? There may be more oddballs.), and the intent of the original issue was to allow people to list over results (https://github.com/rucio/rucio/issues/7232). I think this is up for discussion though. 

> Additionally, for testing, I think it would be great if we could also read from the csv generated to ensure that the output written to the csv corresponds to what we expected 

Certainly I can do this for rule show, but there are no existing tests for rule history, so I am a little lost as to what that's even supposed to show.

rdimaio:
> > I'm not familiar with the pre-existing client CSV functionality, but would it make sense to change the --csv argument so that a file path can be passed to it, so that the user can specify the file path and name?
> 
> The existing functionality is to display results as a csv (I think with the exception of declaring suspicious replicas? There may be more oddballs.), and the intent of the original issue was to allow people to list over results (#7232). I think this is up for discussion though.
> 
> > Additionally, for testing, I think it would be great if we could also read from the csv generated to ensure that the output written to the csv corresponds to what we expected
> 
> Certainly I can do this for rule show, but there are no existing tests for rule history, so I am a little lost as to what that's even supposed to show.

Thanks for the info. I had a look at the existing csv functionality in e.g. `list_dataset_replicas`, looks like we don't explicitly test that either at the moment. I think we can add this csv output functionality to these other commands under these PRs, but it would be good to add tests at some point to ensure we get an expected output for these commands

lib/rucio/client/commands/rule.py
```diff
@@ -77,6 +81,7 @@ def remove_namespace(self, parser: "ArgumentParser") -> None:
     def show_namespace(self, parser: "ArgumentParser") -> None:
         parser.add_argument("--rule-id", help="The rule ID, for accessing an existing rule.")
         parser.add_argument("--examine", action="store_true", help="Detailed analysis of transfer errors")
+        parser.add_argument("--csv", action="store_true", help="Write output to a CSV")
```
rdimaio: This doesn't write the output to a CSV (file), it just prints the output in a CSV format. I think for clarity we should say something along the lines of what is said in `bin_legacy`, e.g. here:

https://github.com/rucio/rucio/blob/e383672a7d5932bd6128c739980d88da9201ffb5/lib/rucio/client/commands/bin_legacy/rucio.py#L2235

so something like "Comma Separated Value output"

lib/rucio/client/commands/rule.py
```diff
@@ -47,6 +47,10 @@ def usage_example(self) -> list[str]:
             "$ rucio rule update --rule-id rule123456 --move --rses NewRSE # Copy an existing rule to a new RSE",
         ]
 
+    def history_namespace(self, parser: "ArgumentParser") -> None:
+        parser.add_argument("--did", help="DID to look for rules")
+        parser.add_argument("--csv", action="store_true", help="Write output to a CSV")
```
rdimaio: This doesn't write the output to a CSV (file), it just prints the output in a CSV format. I think for clarity we should say something along the lines of what is said in `bin_legacy`, e.g. here:

https://github.com/rucio/rucio/blob/e383672a7d5932bd6128c739980d88da9201ffb5/lib/rucio/client/commands/bin_legacy/rucio.py#L2235

so something like "Comma Separated Value output"

lib/rucio/client/commands/rule.py
voetberg: Good call. These `csv` options also don't show a header so maybe a better help menu would say something like `Output results as CSVs, headers are [headers]`.

dchristidis:
>> I'm not familiar with the pre-existing client CSV functionality, but would it make sense to change the --csv argument so that a file path can be passed to it, so that the user can specify the file path and name?
>
> The existing functionality is to display results as a csv (I think with the exception of declaring suspicious replicas? There may be more oddballs.), and the intent of the original issue was to allow people to list over results (https://github.com/rucio/rucio/issues/7232). I think this is up for discussion though.

No, the main use case is a simplified output that can be further processed with filter utilities. If the user desires to write the output to a file, they can use redirection.

lib/rucio/client/commands/bin_legacy/rucio.py
```diff
@@ -1631,30 +1631,27 @@ def list_rules_history(args, client, logger, console, spinner):
...
-                print('RSE expression : %s' % (rule['rse_expression']))
-                print('Time : %s' % (rule['updated_at']))
+            table_data.append(['Deletion', rule['account'], rule['rse_expression'], rule['updated_at']])
 
-    if cli_config == 'rich':
+    if args.csv:
+        print('ACTION', 'ACCOUNT', 'RSE EXPRESSION', 'TIME', sep=',')
+        for rule in table_data:
+            print(rule[0], rule[1], rule[2], rule[3], sep=',')
+
```
dchristidis: There is an execution path with both `cli_config == 'rich'` and `args.csv` evaluating to `True` where there’s a `spinner.start()` but no `spinner.stop()`. Is this safe?

lib/rucio/client/commands/bin_legacy/rucio.py
```diff
@@ -1631,30 +1631,27 @@ def list_rules_history(args, client, logger, console, spinner):
...
-                print('-' * 40)
-                print('Rule deletion')
-                print('Account : %s' % rule['account'])
-                print('RSE expression : %s' % (rule['rse_expression']))
-                print('Time : %s' % (rule['updated_at']))
+            table_data.append(['Deletion', rule['account'], rule['rse_expression'], rule['updated_at']])
 
-    if cli_config == 'rich':
+    if args.csv:
+        print('ACTION', 'ACCOUNT', 'RSE EXPRESSION', 'TIME', sep=',')
```
dchristidis: ‘CSV’ stands for ‘comma-separated values’, but it’s an ill-defined format, with a tab or a semi-colon being frequent choices for a separator. Choosing the comma will end up being problematic with `list-identities`.

In addition, I think it wise that the separator be passed not as an individual string per `print()` call, but as a global variable (e.g. `CSV_SEPARATOR`).

dchristidis: In this PR there’s a header but in #7426 there isn’t. The use of `--csv` should be consistent. We should choose one behaviour and apply it to all commands.

I can think of persuading arguments for both cases.

lib/rucio/client/commands/bin_legacy/rucio.py
```diff
@@ -1631,30 +1631,27 @@ def list_rules_history(args, client, logger, console, spinner):
...
         table_data = sorted(table_data, key=lambda entry: entry[-1], reverse=True)
         table = generate_table(table_data, headers=['ACTION', 'ACCOUNT', 'RSE EXPRESSION', 'TIME'])
         spinner.stop()
         print_output(table, console=console, no_pager=args.no_pager)
+    else:
+        for rule in table_data:
+            print(f"Rule {rule[0].lower()}")
+            print(f'Account : {rule[1]}')
+            print(f'RSE expression : {rule[2]}')
+            print(f'Time : {rule[3]}')
```
dchristidis: The elements of `table_data` are no longer separated by a line of hyphens, hindering legibility.

lib/rucio/client/commands/bin_legacy/rucio.py
voetberg: You're right, I should have some of these PRs as a draft until we make these sorts of decisions. 

Even though this PR has the printed headers, I would argue against them as per the spirit of the original issue, it would be annoying to have to filter them out when doing downstream operations. [As I said here](https://github.com/rucio/rucio/pull/7421#discussion_r1965674443) it's probably cleaner to just include in the help what the columns are if there's more than one.

lib/rucio/client/commands/bin_legacy/rucio.py
voetberg: 👍 Totally right, I lost that in a copy-paste. Reverting it now.

lib/rucio/client/commands/bin_legacy/rucio.py
voetberg: I haven't had a problem locally (I have rich enabled) but you're right, it's better to be safe than sorry.

PR #7426, opened 2025-02-20
Client: Add csv option to list rses for rucio/rucio-admin 

### LINKED ISSUE #7414: Add `--csv` option to rucio_admin/list_rses, rucio/list_rses
### Description

Add `csv` as an option to facilitate iterating over the result of a cli command. See parent issue for further description. 

### Motivation

_No response_

### Change

_No response_

voetberg:
Closes #7414

### DIFF
--- a/lib/rucio/client/commands/bin_legacy/rucio.py
+++ b/lib/rucio/client/commands/bin_legacy/rucio.py
@@ -1671,7 +1671,9 @@ def list_rses(args, client, logger, console, spinner):
         spinner.start()
 
     rses = client.list_rses(args.rses)
-    if cli_config == 'rich':
+    if args.csv:
+        print(*(rse['rse'] for rse in rses), sep='\n')
+    elif cli_config == 'rich':
         table = generate_table([[rse['rse']] for rse in sorted(rses, key=lambda elem: elem['rse'])], headers=['RSE'], col_alignments=['left'])
         spinner.stop()
         print_output(table, console=console, no_pager=args.no_pager)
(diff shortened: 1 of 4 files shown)


tests/test_bin_rucio.py
```diff
@@ -224,7 +224,8 @@ def test_add_rse_nondet(self):
         assert 'Added new non-deterministic RSE: %s\n' % tmp_val in out
 
     def test_list_rses(self):
-        """CLIENT(ADMIN): List RSEs"""
+        """CLIENT(USER/ADMIN): List RSEs"""
+        # TODO Test filter
```
rdimaio, line 228: Is this planned to be done in this PR or an upcoming PR?

lib/rucio/client/commands/bin_legacy/rucio.py
```diff
@@ -1671,7 +1671,9 @@ def list_rses(args, client, logger, console, spinner):
         spinner.start()
 
     rses = client.list_rses(args.rses)
-    if cli_config == 'rich':
+    if args.csv:
+        print(*(rse['rse'] for rse in rses), sep=',')
+    elif cli_config == 'rich':
```
rdimaio: Should this be `if` instead of `elif`? Otherwise if you have `cli_config == rich` and `--csv` is passed, you end up with non-rich output

lib/rucio/client/commands/bin_legacy/rucio_admin.py
```diff
@@ -327,7 +327,9 @@ def list_rses(args, client, logger, console, spinner):
         spinner.start()
 
     rses = client.list_rses()
-    if cli_config == 'rich':
+    if args.csv:
+        print(*(rse['rse'] for rse in rses), sep=',')
+    elif cli_config == 'rich':
```
rdimaio, line 332: Same comment about `if` and `elif` here

lib/rucio/client/commands/bin_legacy/rucio.py
voetberg: No, this is the intended behavior. If you have both as `if` you get the output twice, and the rich output would break any sort of iterative bash script. The `--csv` arg should take priority over a rich output.

tests/test_bin_rucio.py
voetberg: Upcoming PR, I was going to make a meta-issue for the testing problems I found while doing #7232

rdimaio (dismissed):

dchristidis (changes requested):
Writing the RSEs in a single line is not the desired behaviour. This is a single-column table; the equivalent CSV output should be one RSE per line.

dchristidis (approved):
I think there’s room for improvement with the help messages, but let’s come back to them at a later date. Thanks!

PR #7438, opened 2025-02-26
Protocols: Handle `exists(None)` in GFAL Fix #7398

### LINKED ISSUE #7398: Avoid GFAL error when `exists(None)` is called.
### Description

In our code (e.g., `rsemanager.exists(...)`), we do:

https://github.com/rucio/rucio/blob/37cc33b21f2ca2d99bffd2da7e230b1e50fce1f2/lib/rucio/rse/rsemanager.py#L272-L280

- With GFAL, calling `exists(None)` would lead to an internal `ctx.stat("None")` which raises `"Protocol not supported or path/url invalid: None."`
- This is currently caught in `rsemanager.exists(...)` by `except Exception: pass`, so we never fall back to the “write” protocol. The rest of the flow may work, but we see messy log errors about “stat(None).”
- For a quick solution, and if we want to preserve the current behavior (i.e. remain in the “read” protocol and skip fallback) but avoid trying with GFAL to stat `None`, maybe, at the beginning of gfal's `def exists(self, path):` we should do something like:

```
if path is None:
    # Ignoring exists(None) in GFAL plugin, returning False."
    return False
```

### Steps to reproduce

Trigger the `exists(...)` of rsemanager.py with gfal.

### Rucio Version

36.3.0

### Additional Information

_No response_

Geogouz:
With the introduced changes, GFAL no longer tries to execute `exists(None)` as it was always leading to an exception. It now returns the exception `RSEOperationNotSupported` which will silently lead to the same prior behavior as it is caught by:

https://github.com/rucio/rucio/blob/5dcf2f958ee80ae335a88b95f83291b1e5b241eb/lib/rucio/rse/rsemanager.py#L274-L280

Moreover, I removed the `return -1` from

https://github.com/rucio/rucio/blob/5dcf2f958ee80ae335a88b95f83291b1e5b241eb/lib/rucio/rse/protocols/gfal.py#L514-L518

because judging from the underlying C function `gfal2_stat(..)`, it is effectively unreachable. Basically, the helper function translates the C GError into a Python exception of type `gfal2.GError` and so, instead of returning any negative integer, it raises the exception.

[Source:](https://github.com/cern-fts/gfal2-python/blob/69466f56110ec87886d1749d0a8d4529d153c30d/src/Gfal2Context.cpp#L156-L165)

```
Stat Gfal2Context::stat_c(const std::string & path)
{
    ScopedGILRelease unlock;
    GError* tmp_err = NULL;
    Stat st;
    const int ret = gfal2_stat(cont->get(), path.c_str(), &st._st, &tmp_err);
    if (ret < 0)
        GErrorWrapper::throwOnError(&tmp_err);
    return st;
}
```

### DIFF
--- a/lib/rucio/rse/protocols/gfal.py
+++ b/lib/rucio/rse/protocols/gfal.py
@@ -28,7 +28,7 @@ from rucio.rse.protocols import protocol
 
 try:
     import gfal2  # pylint: disable=import-error
-except:
+except Exception:
     if 'RUCIO_CLIENT_MODE' not in os.environ:
         if not config.config_has_section('database'):
             raise exception.MissingDependency('Missing dependency : gfal2')
@@ -344,7 +344,12 @@ class Default(protocol.RSEProtocol):
 
         :raises SourceNotFound: if the source file was not found on the referred storage.
         """
+
         self.logger(logging.DEBUG, 'checking if file exists {}'.format(path))
+
+        if path is None:
+            # Action not supported
+            raise exception.RSEOperationNotSupported()
 
         try:
             status = self.__gfal2_exist(path)
@@ -515,7 +520,6 @@ class Default(protocol.RSEProtocol):
         try:
             if ctx.stat(str(path)):
                 return 0
-            return -1
         except gfal2.GError as error:  # pylint: disable=no-member
             if error.code == errno.ENOENT or 'No such file' in str(error):  # pylint: disable=no-member
                 return -1
--- a/lib/rucio/rse/rsemanager.py
+++ b/lib/rucio/rse/rsemanager.py
@@ -266,18 +266,18 @@ def exists(rse_settings: types.RSESettingsDict, files, domain='wan', scheme=None
         :raises RSENotConnected: no connection to a specific storage has been established
     """
 
-    ret = {}
-    gs = True  # gs represents the global status which indicates if every operation worked in bulk mode
-
     protocol = create_protocol(rse_settings, 'read', scheme=scheme, impl=impl, domain=domain, auth_token=auth_token, logger=logger)
     protocol.connect()
-    try:
-        protocol.exists(None)
-    except NotImplementedError:
+
+    from rucio.rse.protocols.protocol import RSEProtocol  # Placed it here to avoid possible circular imports
+    # Check if 'exists' is truly overridden on the read protocol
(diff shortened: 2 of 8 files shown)


mlassnig (dismissed):

lib/rucio/rse/protocols/gfal.py
```diff
@@ -28,7 +28,7 @@
 
 try:
     import gfal2  # pylint: disable=import-error
-except:
+except:  # noqa: E722
```
rdimaio:
```suggestion
except Exception:
```

This way you can avoid having the noqa, while also avoiding catching possible system exceptions that we might not want to catch: https://docs.astral.sh/ruff/rules/bare-except/

lib/rucio/rse/protocols/gfal.py
```diff
@@ -344,7 +344,12 @@ def exists(self, path):
 
         :raises SourceNotFound: if the source file was not found on the referred storage.
         """
-        self.logger(logging.DEBUG, 'checking if file exists {}'.format(path))
+
+        self.logger(logging.DEBUG, f'checking if file exists {path}')
```
rdimaio: Avoid f-string logging: https://github.com/rucio/rucio/issues/6826

Geogouz: Ok, makes sense.

lib/rucio/rse/rsemanager.py
```diff
@@ -266,18 +266,18 @@ def exists(rse_settings: types.RSESettingsDict, files, domain='wan', scheme=None
...
-    ret = {}
-    gs = True  # gs represents the global status which indicates if every operation worked in bulk mode
-
     protocol = create_protocol(rse_settings, 'read', scheme=scheme, impl=impl, domain=domain, auth_token=auth_token, logger=logger)
     protocol.connect()
-    try:
-        protocol.exists(None)
-    except NotImplementedError:
+
+    from rucio.rse.protocols.protocol import RSEProtocol  # Placed it here to avoid possible circular imports
```
Geogouz, line 272: Placed that here because in protocol we are importing rsemanager itself (which I guess is not the best practice):

https://github.com/rucio/rucio/blob/981f5455551cfab85e8cc1f3fbb89d7e1dfba98e/lib/rucio/rse/protocols/protocol.py#L24-L33

rdimaio (approved):

PR #7440, opened 2025-02-26
Lifetime Model: Fix counting of containers #7439

### LINKED ISSUE #7439: Resolved containers not being accounted in the summary
### Description

As reported by a user, the containers that are resolved into datasets are properly included in a Lifetime Model exception request, but the summary reports zero.

```
Total DIDs                                                                                             1282
DID not submitted because it is a file                                                                    0
DID that are containers and were resolved                                                              1282
DID not submitted because it is not part of the lifetime campaign                                      2589
DID successfully submitted including the one from containers resolved                                     0
```

### Steps to reproduce

Submit an exception requests using a list of containers.

### Rucio Version

_No response_

### Additional Information

_No response_

dchristidis:


rdimaio (approved):
LGTM - I opened a separate issue to discuss changing the key names in `summary` https://github.com/rucio/rucio/issues/7443, I don't like the way we're just using ints now (could have as well used a list at that point)

PR #7450, opened 2025-02-27
Add context manager for database sessions, migrate gateway/identity, add validation to relevant APIs

### LINKED ISSUE #7407: Create context managers for database session handling to replace current `*_session` decorators
Related:
- https://github.com/rucio/rucio/issues/6989
- https://github.com/rucio/rucio/issues/6986

As part of the initial PR, I will also migrate one initial data flow to ensure everything works as expected

### LINKED ISSUE #7448: API identity put calls don't fail if email or identity key name are not passed, despite them being required by the SQLA model
### Description

GSS for example:

https://github.com/rucio/rucio/blob/5dcf2f958ee80ae335a88b95f83291b1e5b241eb/lib/rucio/web/rest/flaskapi/v1/identities.py#L181-L192

`email` could be `None`. However, the `Identity` model requires the `email` to not be null:

https://github.com/rucio/rucio/blob/5dcf2f958ee80ae335a88b95f83291b1e5b241eb/lib/rucio/db/sqla/models.py#L349-L362

This is not something that we can substitute with a default value, so we should validate that these exist when the call is made.

### Steps to reproduce

N/A

### Rucio Version

_No response_

### Additional Information

_No response_

rdimaio:
- 1fb4c82adc12a7ec728bb49108b46fe99d581808: fixes https://github.com/rucio/rucio/issues/7407. I created a single context manager to replace read and write operations, as they mostly followed the same logic - the main difference is that `read` only rolls back, and `write` can commit. Additionally, `read` always rolls back at the end, which was not the case with the decorator; this is because it's good practice to always commit/rollback a transaction, see https://stackoverflow.com/questions/309834/should-i-commit-or-rollback-a-read-transaction
- 0909b1e53660e2a767c91fdda3b68a432e65187e: this is the initial gateway migration. I plan to migrate the entire gateway layer first, and then move on to core. I find this a bit easier, because gateway functions are generally only called by the API, whereas core functions are a bit more widespread.
- The other commits add API validation to parameters that were not being validated, but whose `Optional` type was not previously causing issues with the static code analyzers due to the fact that the session decorators were hiding the annotations from the analyzers.

### Why is there no stream session equivalent in the context manager?
Stream session and read session are functionally equivalent, from the session's point of view. How the function's return is handled should be handled by the function, it doesn't depend on the session (i.e. the context manager's job is to manage the session, the function's job is to return the data as needed).

### DIFF
--- a/lib/rucio/db/sqla/constants.py
+++ b/lib/rucio/db/sqla/constants.py
@@ -201,6 +201,6 @@ class TransferLimitDirection(Enum):
     DESTINATION = 'D'
 
 
-class DBSessionOperation(Enum):
+class DatabaseOperationType(Enum):
     READ = 'read'
     WRITE = 'write'
--- a/lib/rucio/db/sqla/constants.py
+++ b/lib/rucio/db/sqla/constants.py
@@ -199,3 +199,8 @@ class SubscriptionState(Enum):
 class TransferLimitDirection(Enum):
     SOURCE = 'S'
     DESTINATION = 'D'
+
+
+class DBSessionOperation(Enum):
+    READ = 'read'
+    WRITE = 'write'
--- a/lib/rucio/web/rest/flaskapi/v1/accounts.py
+++ b/lib/rucio/web/rest/flaskapi/v1/accounts.py
@@ -687,7 +687,7 @@ class Identities(ErrorHandlingMethodView):
         vo = request.environ.get('vo')
 
         if not issuer or not vo:
-            return 'Issuer and VO must be set.', 400
+            return generate_http_error_flask(400, ValueError.__name__, 'Issuer and VO must be set.')
 
         try:
             add_account_identity(
@@ -800,7 +800,7 @@ class Identities(ErrorHandlingMethodView):
         vo = request.environ.get('vo')
 
         if not issuer or not vo:
-            return 'Issuer and VO must be set.', 400
+            return generate_http_error_flask(400, ValueError.__name__, 'Issuer and VO must be set.')
 
         try:
             del_account_identity(identity, authtype, account, issuer, vo=vo)
--- a/lib/rucio/web/rest/flaskapi/v1/accounts.py
+++ b/lib/rucio/web/rest/flaskapi/v1/accounts.py
@@ -682,6 +682,13 @@ class Identities(ErrorHandlingMethodView):
         identity = param_get(parameters, 'identity')
         authtype = param_get(parameters, 'authtype')
         email = param_get(parameters, 'email')
+
+        issuer = request.environ.get('issuer')
+        vo = request.environ.get('vo')
+
+        if not issuer or not vo:
+            return 'Issuer and VO must be set.', 400
+
         try:
             add_account_identity(
                 identity_key=identity,
@@ -689,9 +696,9 @@ class Identities(ErrorHandlingMethodView):
                 account=account,
                 email=email,
                 password=param_get(parameters, 'password', default=None),
-                issuer=request.environ.get('issuer'),
+                issuer=issuer,
                 default=param_get(parameters, 'default', default=False),
-                vo=request.environ.get('vo'),
+                vo=vo,
             )
         except AccessDenied as error:
             return generate_http_error_flask(401, error)
@@ -788,8 +795,15 @@ class Identities(ErrorHandlingMethodView):
         parameters = json_parameters()
         identity = param_get(parameters, 'identity')
         authtype = param_get(parameters, 'authtype')
+
+        issuer = request.environ.get('issuer')
+        vo = request.environ.get('vo')
+
+        if not issuer or not vo:
+            return 'Issuer and VO must be set.', 400
+
         try:
-            del_account_identity(identity, authtype, account, request.environ.get('issuer'), vo=request.environ.get('vo'))
+            del_account_identity(identity, authtype, account, issuer, vo=vo)
         except AccessDenied as error:
             return generate_http_error_flask(401, error)
         except (AccountNotFound, IdentityError) as error:
--- a/lib/rucio/web/rest/flaskapi/v1/identities.py
+++ b/lib/rucio/web/rest/flaskapi/v1/identities.py
@@ -77,6 +77,12 @@ class UserPass(ErrorHandlingMethodView):
         if not username or not password or not email:
             return 'Username, Email and Password must be set.', 400
 
+        issuer = request.environ.get('issuer')
+        vo = request.environ.get('vo')
+
+        if not issuer or not vo:
+            return 'Issuer and VO must be set.', 400
+
         add_identity(username, 'userpass', email, password)
 
         add_account_identity(
@@ -85,8 +91,8 @@ class UserPass(ErrorHandlingMethodView):
             account=account,
             email=email,
             password=password,
-            issuer=request.environ.get('issuer'),
-            vo=request.environ.get('vo'),
+            issuer=issuer,
+            vo=vo,
         )
 
         return 'Created', 201
@@ -133,14 +139,20 @@ class X509(ErrorHandlingMethodView):
         if not dn or not email:
             return 'SSL_CLIENT_S_DN and email must be set.', 400
 
+        issuer = request.environ.get('issuer')
+        vo = request.environ.get('vo')
+
+        if not issuer or not vo:
+            return 'Issuer and VO must be set.', 400
+
         add_identity(dn, 'x509', email=email)
         add_account_identity(
             identity_key=dn,
             id_type='x509',
             account=account,
             email=email,
-            issuer=request.environ.get('issuer'),
-            vo=request.environ.get('vo'),
+            issuer=issuer,
+            vo=vo,
         )
 
         return 'Created', 201
@@ -187,14 +199,20 @@ class GSS(ErrorHandlingMethodView):
         if not gsscred or not email:
             return 'REMOTE_USER and email must be set.', 400
 
+        issuer = request.environ.get('issuer')
+        vo = request.environ.get('vo')
+
+        if not issuer or not vo:
+            return 'Issuer and VO must be set.', 400
+
         add_identity(gsscred, 'gss', email=email)
         add_account_identity(
             identity_key=gsscred,
             id_type='gss',
             account=account,
             email=email,
-            issuer=request.environ.get('issuer'),
-            vo=request.environ.get('vo'),
+            issuer=issuer,
+            vo=vo,
         )
 
         return 'Created', 201
--- a/lib/rucio/web/rest/flaskapi/v1/identities.py
+++ b/lib/rucio/web/rest/flaskapi/v1/identities.py
@@ -130,6 +130,9 @@ class X509(ErrorHandlingMethodView):
         dn = request.environ.get('SSL_CLIENT_S_DN')
         email = request.headers.get('X-Rucio-Email', default=None)
 
+        if not dn or not email:
+            return 'SSL_CLIENT_S_DN and email must be set.', 400
+
         add_identity(dn, 'x509', email=email)
         add_account_identity(
             identity_key=dn,
--- a/lib/rucio/web/rest/flaskapi/v1/identities.py
+++ b/lib/rucio/web/rest/flaskapi/v1/identities.py
@@ -74,8 +74,8 @@ class UserPass(ErrorHandlingMethodView):
         password = request.headers.get('X-Rucio-Password', default=None)
         email = request.headers.get('X-Rucio-Email', default=None)
 
-        if not username or not password:
-            return 'Username and Password must be set.', 400
+        if not username or not password or not email:
+            return 'Username, Email and Password must be set.', 400
 
         add_identity(username, 'userpass', email, password)
 
--- a/lib/rucio/web/rest/flaskapi/v1/identities.py
+++ b/lib/rucio/web/rest/flaskapi/v1/identities.py
@@ -16,7 +16,7 @@ from flask import Flask, jsonify, request
 
 from rucio.gateway.identity import add_account_identity, add_identity, list_accounts_for_identity
 from rucio.web.rest.flaskapi.authenticated_bp import AuthenticatedBlueprint
-from rucio.web.rest.flaskapi.v1.common import ErrorHandlingMethodView, check_accept_header_wrapper_flask, response_headers
+from rucio.web.rest.flaskapi.v1.common import ErrorHandlingMethodView, check_accept_header_wrapper_flask, generate_http_error_flask, response_headers
 
 
 class UserPass(ErrorHandlingMethodView):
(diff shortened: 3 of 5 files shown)


rcarpa:
Be careful with adding empty rollbacks on read. I'm not sure they are "free" on oracle. I vaguely remember some discussions with the DBA team on this subject. 

In what concerns the transition to context manager... It's sad to make the code more complex just because static analyzers cannot understand perfectly valid and idiomatic python ))

rdimaio:
> Be careful with adding empty rollbacks on read. I'm not sure they are "free" on oracle. I vaguely remember some discussions with the DBA team on this subject.
> 

SQLA should still roll the transaction back when `Session.close()` is called. Per their docs (https://docs.sqlalchemy.org/en/20/orm/session_transaction.html):

> The [Session](https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.Session) itself features a [Session.close()](https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.Session.close) method. If the [Session](https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.Session) is begun within a transaction that has not yet been committed or rolled back, this method will cancel (i.e. rollback) that transaction, and also expunge all objects contained within the [Session](https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.Session) object’s state.

I'm okay with removing the empty `rollback` as it's effectively unnecessary given the `close()` operation, but I don't think leaving it there would cause issues

> In what concerns the transition to context manager... It's sad to make the code more complex just because static analyzers cannot understand perfectly valid and idiomatic python ))

Can you clarify what you mean by making the code more complex? I see a few advantages of using the context manager here:
1. with the context manager, it is clearer at which level we begin the session. With the current decorator approach, all of this is obfuscated. This is effectively an antipattern discouraged by SQLA, see my comment here: https://github.com/rucio/rucio/issues/6989#issuecomment-2260606756
2. with the context manager, it's clear where in the function the session is actually needed and being used. We can open the session right before we need it and close it right after we don't need it anymore. With the decorator, approach, the session is open for the entirety of each decorated function, regardless of whether the session is actually still needed or not.

rcarpa:
Contrary to what it may seem, at first look at the code, rucio doesn't actually use the arbitrary nesting antipattern described in the documentation. The read/transactional/write_session decorators are doing exactly what the solution to the antipattern suggests in the documentation: initializing a session every single time it wasn't passed as an argument and using it. The decorator just rely on syntactic sugar provided by python to hide this detail in an elegant way.  Unfortunately, in a way which cannot be handled by the current expressiveness of the typing module for static type checks.

Read more attentively the example in the sqlalchemy documentation and you will see that your solution actually diverges from what they propose: you suggest to completely stop passing the session as an argument and retrieve it from the sessionmaker at every single operation, while the example in the documentation continues to pass it as an argument and perform some initialization if not yet initialized. This will probably have an impact on the performance: according to the current code, the session will be instantiated and commited at every single "with" block. Which is a HUGE change compared to doing it once, at the outermost call of a decorated function. This behavioral change alone may be very dangerous... just in terms of transactional data coherence in the database. 

The old behavior of rucio, with decorators, is closer to the solution recommended by sqlalchemy than what is done in this PR. 

Reducing the scope of the session usage to the part of code which actually needs it (rather than to the whole function) is a good initiative, but only in the case when the decorated (by read/transactional/write_session) function only needs the session a single time. As soon you use the "with" block twice in a function the solution from this PR will change the behavior by having multiple transactions instead of one. If you start moving half of the body of the function into a "with" block: you didn't gain anything other than adding one identation layer.

rcarpa:
Ignore my hidden comment. If there is a bigger effort behind the scene to completely get rid of decorators in core, the change makes sense.

rdimaio:
> Ignore my hidden comment. If there is a bigger effort behind the scene to completely get rid of decorators in core, the change makes sense.

Yes, the idea is to remove decorators from core, and to have no session management done in core itself. Each core function that needs a session should explicitly always take `session` as one of the arguments, and the session should be managed outside of core - e.g. in gateway.

rcarpa:
That will be a lot of work for daemons.

rdimaio:
> That will be a lot of work for daemons.

We will do things progressively, so gateway first, then daemons, and only remove the decorators from core once all parent callers have been migrated. In the meantime, decorators will still be part of the code, so we can easily just perform this migration alongside the existing decorator functionality.

rdimaio:
> > Be careful with adding empty rollbacks on read. I'm not sure they are "free" on oracle. I vaguely remember some discussions with the DBA team on this subject.
> 
> SQLA should still roll the transaction back when `Session.close()` is called. Per their docs (https://docs.sqlalchemy.org/en/20/orm/session_transaction.html):
> 
> > The [Session](https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.Session) itself features a [Session.close()](https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.Session.close) method. If the [Session](https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.Session) is begun within a transaction that has not yet been committed or rolled back, this method will cancel (i.e. rollback) that transaction, and also expunge all objects contained within the [Session](https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.Session) object’s state.
> 
> I'm okay with removing the empty `rollback` as it's effectively unnecessary given the `close()` operation, but I don't think leaving it there would cause issues
> 
> > In what concerns the transition to context manager... It's sad to make the code more complex just because static analyzers cannot understand perfectly valid and idiomatic python ))
> 
> Can you clarify what you mean by making the code more complex? I see a few advantages of using the context manager here:
> 
>     1. with the context manager, it is clearer at which level we begin the session. With the current decorator approach, all of this is obfuscated. This is effectively an antipattern discouraged by SQLA, see my comment here: [`session` parameter in `*_session` decorated functions is required, but effectively treated as optional #6989 (comment)](https://github.com/rucio/rucio/issues/6989#issuecomment-2260606756)
> 
>     2. with the context manager, it's clear where in the function the session is actually needed and being used. We can open the session right before we need it and close it right after we don't need it anymore. With the decorator, approach, the session is open for the entirety of each decorated function, regardless of whether the session is actually still needed or not.

Latest revision - I've removed the `rollback` for peace of mind - `close()` will rollback regardless

bari12 (dismissed):
For me this is good.
Waiting for 2 or 4 more eyes if this is the general direction we should take

lib/rucio/db/sqla/constants.py
```diff
@@ -199,3 +199,8 @@ class SubscriptionState(Enum):
 class TransferLimitDirection(Enum):
     SOURCE = 'S'
     DESTINATION = 'D'
+
+
+class DBSessionOperation(Enum):
```
mlassnig: Wouldn't this rather be

`class TransactionType(Enum)`

?

mlassnig: so resulting in something like

`with db_session(TransactionType.WRITE) as session:`
`with db_session(TransactionType.READ) as session:`

lib/rucio/web/rest/flaskapi/v1/accounts.py
```diff
@@ -682,16 +682,23 @@ def post(self, account: str) -> 'ResponseReturnValue':
         identity = param_get(parameters, 'identity')
         authtype = param_get(parameters, 'authtype')
         email = param_get(parameters, 'email')
+
+        issuer = request.environ.get('issuer')
+        vo = request.environ.get('vo')
+
+        if not issuer or not vo:
+            return 'Issuer and VO must be set.', 400
```
mlassnig: shouldn't this be `return generate_http_error_flask(400, 'Issuer and VO must be set.')` ?

lib/rucio/web/rest/flaskapi/v1/accounts.py
```diff
@@ -788,8 +795,15 @@ def delete(self, account: str) -> 'ResponseReturnValue':
         parameters = json_parameters()
         identity = param_get(parameters, 'identity')
         authtype = param_get(parameters, 'authtype')
+
+        issuer = request.environ.get('issuer')
+        vo = request.environ.get('vo')
+
+        if not issuer or not vo:
+            return 'Issuer and VO must be set.', 400
```
mlassnig: same

lib/rucio/web/rest/flaskapi/v1/identities.py
```diff
@@ -74,8 +74,14 @@ def put(self, account):
         password = request.headers.get('X-Rucio-Password', default=None)
         email = request.headers.get('X-Rucio-Email', default=None)
 
-        if not username or not password:
-            return 'Username and Password must be set.', 400
+        if not username or not password or not email:
+            return 'Username, Email and Password must be set.', 400
```
mlassnig: same

lib/rucio/web/rest/flaskapi/v1/identities.py
```diff
@@ -74,8 +74,14 @@ def put(self, account):
...
+
+        issuer = request.environ.get('issuer')
+        vo = request.environ.get('vo')
+
+        if not issuer or not vo:
+            return 'Issuer and VO must be set.', 400
```
mlassnig: same

lib/rucio/web/rest/flaskapi/v1/identities.py
```diff
@@ -130,14 +136,23 @@ def put(self, account):
...
         email = request.headers.get('X-Rucio-Email', default=None)
 
+        if not dn or not email:
+            return 'SSL_CLIENT_S_DN and email must be set.', 400
+
+        issuer = request.environ.get('issuer')
+        vo = request.environ.get('vo')
+
+        if not issuer or not vo:
+            return 'Issuer and VO must be set.', 400
```
mlassnig: same

lib/rucio/web/rest/flaskapi/v1/identities.py
```diff
@@ -130,14 +136,23 @@ def put(self, account):
         dn = request.environ.get('SSL_CLIENT_S_DN')
         email = request.headers.get('X-Rucio-Email', default=None)
 
+        if not dn or not email:
+            return 'SSL_CLIENT_S_DN and email must be set.', 400
```
mlassnig: same

lib/rucio/web/rest/flaskapi/v1/identities.py
```diff
@@ -181,14 +196,23 @@ def put(self, account):
         gsscred = request.environ.get('REMOTE_USER')
         email = request.headers.get('X-Rucio-Email', default=None)
 
+        if not gsscred or not email:
+            return 'REMOTE_USER and email must be set.', 400
```
mlassnig: same

lib/rucio/web/rest/flaskapi/v1/identities.py
```diff
@@ -181,14 +196,23 @@ def put(self, account):
...
+
+        issuer = request.environ.get('issuer')
+        vo = request.environ.get('vo')
+
+        if not issuer or not vo:
+            return 'Issuer and VO must be set.', 400
```
mlassnig: same

mlassnig (commented):
this looks good. added a few suggestions

lib/rucio/db/sqla/constants.py
rdimaio: I don't mind either name. My main reasoning for using "operation" instead of "transaction" is to remain consistent with SQLA's terminology, particularly with regards to https://docs.sqlalchemy.org/en/20/orm/session_transaction.html

A `Session` manages a transaction; we cannot actually change a transaction type, set it to 'write' or to 'read', we can only modify the `Session`'s behaviour. Whether we pass `TransactionType.WRITE` `TransactionType.READ`, the transaction is effectively the same - it's the `Session`'s behaviour that changes with this value

lib/rucio/web/rest/flaskapi/v1/accounts.py
rdimaio: Addressed in latest revision

lib/rucio/web/rest/flaskapi/v1/accounts.py
rdimaio: Addressed in latest revision

lib/rucio/web/rest/flaskapi/v1/identities.py
rdimaio: Addressed in latest revision

lib/rucio/web/rest/flaskapi/v1/identities.py
rdimaio: Addressed in latest revision

lib/rucio/web/rest/flaskapi/v1/identities.py
rdimaio: Addressed in latest revision

lib/rucio/web/rest/flaskapi/v1/identities.py
rdimaio: Addressed in latest revision

lib/rucio/web/rest/flaskapi/v1/identities.py
rdimaio: Addressed in latest revision

lib/rucio/db/sqla/constants.py
mlassnig: Ok, then how about `OperationType` ?

rdimaio: I think `OperationType` is a bit too vague, as it's not clear that it's about database operations. I propose we meet in the middle and use `DatabaseOperationType`, which is also in line with the terminology we use here https://rucio.github.io/documentation/operator/database/ - what do you think?

lib/rucio/web/rest/flaskapi/v1/identities.py
```diff
@@ -181,14 +196,23 @@ def put(self, account):
...
+            return generate_http_error_flask(400, ValueError.__name__, 'REMOTE_USER and email must be set.')
+
+        issuer = request.environ.get('issuer')
+        vo = request.environ.get('vo')
+
+        if not issuer or not vo:
+            return generate_http_error_flask(400, ValueError.__name__, 'Issuer and VO must be set.')
+
         add_identity(gsscred, 'gss', email=email)
         add_account_identity(
```
Geogouz, line 209: I guess if the second call fails, the first has already been committed, resulting in atomicity loss. Not sure how problematic this could be, but an idea could be to use one session for both add_identity() and add_account_identity().

lib/rucio/db/sqla/constants.py
rdimaio: Renamed to `DatabaseOperationType` in latest revision

lib/rucio/web/rest/flaskapi/v1/identities.py
rdimaio: I agree, but I'd say this issue is out of the scope of this PR. This issue existed beforehand as well, because session management is never done at the API level, so any API call that calls more than one gateway operation will have this issue.

Issue opened: https://github.com/rucio/rucio/issues/7513

rdimaio: I also opened this issue to discuss where session management should be done in the API-gateway system https://github.com/rucio/rucio/issues/7514

Geogouz: Yep, nice. Keeping it separately as an issue sounds better as it is indeed out of scope. There are actually also many other such occasions in general (not only with those two methods). Just spotted it also here and commented it (having atomicity in mind).

bari12 (approved):

PR #7456, opened 2025-03-05
Patch: add checks for existing upstream remote and Git hooks #7454

### LINKED ISSUE #7454: Add checks to prevent errors when setting remote and installing Git hooks in `configure_git.sh`
The Rucio documentation's [contributing section](https://rucio.github.io/documentation/contributing) mentions setting up a remote in the `1. Prerequisite` section. However, it also suggests an optional step to install Git hooks, which again tries to set the remote to https://github.com/rucio/rucio.git. This can cause an error (`error: remote upstream already exists`.) if the remote was already set in the previous step when running `./tools/configure_git.sh`.

To fix this, we add a check in `./tools/configure_git.sh` to ensure the remote upstream is only added if it doesn't already exist. Similarly, we add a check for pre-commit setup to prevent unnecessary actions.

```bash
if git remote get-url upstream &>/dev/null; then
    echo "Remote 'upstream' already exists. Skipping addition."
else
    echo "Adding remote 'upstream'..."
    git remote add upstream https://github.com/rucio/rucio.git
fi
```

0xquark:
Fixes #7454

### DIFF
--- a/tools/configure_git.sh
+++ b/tools/configure_git.sh
@@ -14,7 +14,20 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-git remote add upstream https://github.com/rucio/rucio.git
+# Check if the upstream remote already exists
+if git remote get-url upstream &>/dev/null; then
+    echo "Remote 'upstream' already exists. Skipping addition."
+else
+    echo "Adding remote 'upstream'..."
+    git remote add upstream https://github.com/rucio/rucio.git
+fi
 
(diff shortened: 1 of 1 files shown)


rdimaio (approved):
LGTM - just two small notes regarding the commit itself:
- generally we only use component names to label the beginning of the commit messages, not `patch`; however in this case I think the component is a bit ambiguous, and we're discussing moving to conventional commits (https://www.conventionalcommits.org/en/v1.0.0/), so you can ignore this
- you might want to re-commit with your GH email, right now the commit author shows as 'root'

PR #7459, opened 2025-03-05
`add_lifetime_exception` summary : Refactor summary dict keys into proper names #7443

### LINKED ISSUE #7443: `add_lifetime_exception` summary: refactor dict keys into proper names
See:
 
https://github.com/rucio/rucio/blob/5dcf2f958ee80ae335a88b95f83291b1e5b241eb/lib/rucio/client/commands/bin_legacy/rucio.py#L2003-L2044

`summary` just has 1, 2, 3, 4 as key names, but these aren't very clear. These correspond to:

```
Total DIDs                                                                                             1282
DID not submitted because it is a file                                                                    0
DID that are containers and were resolved                                                              1282
DID not submitted because it is not part of the lifetime campaign                                      2589
DID successfully submitted including the one from containers resolved                                     0
```

so I think it would be clearer if this dict was refactored into having keys that reflect what they correspond to

0xquark:
Fixes: #7443 
> This PR also amends my last commit with the correct Author.

The old `error_types` list has been replaced with a dictionary that maps proper names (like "total_dids" and "files_ignored") to their related error messages. This makes it easier to understand what each error type means when mapped to summary (also making it coupled while printing it). This way we also  eliminate reliance on positional indexes, which can be error-prone and hard to understand.

### DIFF
--- a/lib/rucio/client/commands/bin_legacy/rucio.py
+++ b/lib/rucio/client/commands/bin_legacy/rucio.py
@@ -1993,15 +1993,16 @@ def add_lifetime_exception(args, client, logger, console, spinner):
     dids_list = []
     containers = []
     datasets = []
-    error_types = ['Total DIDs',
-                   'DID not submitted because it is a file',
-                   'DID that are containers and were resolved',
-                   'DID not submitted because it is not part of the lifetime campaign',
-                   'DID successfully submitted including the one from containers resolved']
     for did in dids:
         scope, name = get_scope(did, client)
         dids_list.append({'scope': scope, 'name': name})
-    summary = {0: len(dids_list), 1: 0, 2: 0, 3: 0, 4: 0}
+    error_summary = {
+        "total_dids": {"description": "Total DIDs", "count": len(dids_list)},
+        "files_ignored": {"description": "DID not submitted because it is a file", "count": 0},
+        "containers_resolved": {"description": "DID that are containers and were resolved", "count": 0},
+        "not_in_lifetime_model": {"description": "DID not submitted because it is not part of the lifetime campaign", "count": 0},
+        "successfully_submitted": {"description": "DID successfully submitted including the one from containers resolved", "count": 0},
+    }
     chunk_limit = 500  # Server should be able to accept 1000
     dids_list_copy = deepcopy(dids_list)
     for chunk in chunks(dids_list_copy, chunk_limit):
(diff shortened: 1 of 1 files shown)


tools/configure_git.sh
rdimaio, line 1: Please rebase this unrelated commit (0bca0b186a4fe1d18d2f62e9faf8bf0edacc5188) to not be in the PR

lib/rucio/client/commands/bin_legacy/rucio.py
```diff
@@ -1993,15 +1993,18 @@ def add_lifetime_exception(args, client, logger, console, spinner):
...
+        "containers_resolved": "DID that are containers and were resolved",
+        "not_in_lifetime_model": "DID not submitted because it is not part of the lifetime campaign",
+        "successfully_submitted": "DID successfully submitted including the one from containers resolved"
+    }
     for did in dids:
         scope, name = get_scope(did, client)
         dids_list.append({'scope': scope, 'name': name})
-    summary = {0: len(dids_list), 1: 0, 2: 0, 3: 0, 4: 0}
+    summary = {key: 0 for key in error_type_mapping}  # Initialize all summary counts
+    summary["total_dids"] = len(dids_list)
```
rdimaio: instead of initializing `total_dids` to 0 first and then setting it to `len(dids_list)`, it can be initialized directly as `len(dids_list)`

lib/rucio/client/commands/bin_legacy/rucio.py
```diff
@@ -1993,15 +1993,18 @@ def add_lifetime_exception(args, client, logger, console, spinner):
...
-                   'DID that are containers and were resolved',
-                   'DID not submitted because it is not part of the lifetime campaign',
-                   'DID successfully submitted including the one from containers resolved']
+    error_type_mapping = {
+        "total_dids": "Total DIDs",
+        "files_ignored": "DID not submitted because it is a file",
+        "containers_resolved": "DID that are containers and were resolved",
+        "not_in_lifetime_model": "DID not submitted because it is not part of the lifetime campaign",
+        "successfully_submitted": "DID successfully submitted including the one from containers resolved"
+    }
```
rdimaio: I don't see why we would assign those strings to the values of this dict - they are essentially comments on what each counter represents, so they should just be comments.

0xquark: Aren't these dictionary values being printed along with the count? We're using the summary dictionary keys to map to proper descriptions, so we no longer rely on positional indexes (as in the previous implementation) when displaying the information.
(https://github.com/rucio/rucio/blob/dbb9c6faf4b02e48de2983fb31dc01e90e93eabc/lib/rucio/client/commands/bin_legacy/rucio.py#L2073)
Like For example, the output now looks like this:
```
Total DIDs                                                                                      1282
DID not submitted because it is a file                                                            0
DID that are containers and were resolved                                                        1282
```

0xquark:
@rdimaio's approach was to create an error_type dictionary that includes both the count and description, making it a single nested dictionary which makes sense.

rdimaio (approved):

PR #7472, opened 2025-03-07
Reaper: do not exit when no RSE are found, wait and retry instead

### LINKED ISSUE #7471: Reaper - Do not exit when no RSE are found
Currently the reaper daemon exits early when no RSE are found.

In k8s this results in the reaper pod being destroyed and subsequently recreated. This is understood as an error state.

When a deployment is first created, no RSE are available. This is a valid state and I think a reaper daemon should be able to exist here without throwing errors due to no RSE.

lobis:
Closes https://github.com/rucio/rucio/issues/7471

All functional changes are just a couple of lines in [0c50cdd](https://github.com/rucio/rucio/pull/7472/commits/0c50cdd0170d00e973007f79b221d245e0e25c5d).

I took the opportunity to do some cleanup, hopefully it does not pollute the blame too much.

### DIFF
--- a/bin/rucio-admin
+++ b/bin/rucio-admin
@@ -12,6 +12,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
 import sys
 
 from rucio.client.commands.bin_legacy.rucio_admin import main as main_legacy
--- a/bin/rucio-reaper
+++ b/bin/rucio-reaper
@@ -13,8 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-'''
-'''
+"""
+Reaper is a daemon to manage file deletion.
+"""
 
 import argparse
 import signal
--- a/lib/rucio/daemons/reaper/reaper.py
+++ b/lib/rucio/daemons/reaper/reaper.py
@@ -12,9 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-'''
+"""
 Reaper is a daemon to manage file deletion.
-'''
+"""
 
 import concurrent.futures.thread  # noqa (https://github.com/rucio/rucio/issues/6548)
 
--- a/lib/rucio/daemons/reaper/reaper.py
+++ b/lib/rucio/daemons/reaper/reaper.py
@@ -690,7 +690,6 @@ def run(
 
     :param threads:                The total number of workers.
     :param chunk_size:             The size of chunk for deletion.
-    :param threads_per_worker:     Total number of threads created by each worker.
     :param once:                   If True, only runs one iteration of the main loop.
     :param greedy:                 If True, delete right away replicas with tombstone.
     :param rses:                   List of RSEs the reaper should work against.
--- a/lib/rucio/daemons/reaper/reaper.py
+++ b/lib/rucio/daemons/reaper/reaper.py
@@ -474,7 +474,7 @@ def run_once(
 
     rses_to_process = get_rses_to_process(rses, include_rses, exclude_rses, vos)
     if not rses_to_process:
-        logger(logging.ERROR, 'Reaper: No RSEs found. Will sleep for 30 seconds')
+        logger(logging.WARNING, 'Reaper: No RSEs found, sleeping')
         return must_sleep
     else:
         rses_to_process = [RseData(id_=rse['id'], name=rse['rse'], columns=rse) for rse in rses_to_process]
--- a/lib/rucio/daemons/reaper/reaper.py
+++ b/lib/rucio/daemons/reaper/reaper.py
@@ -710,14 +710,6 @@ def run(
     if rucio.db.sqla.util.is_old_db():
         raise DatabaseException('Database was not updated, daemon won\'t start')
 
-    logging.log(logging.INFO, 'main: starting processes')
-    rses_to_process = get_rses_to_process(rses, include_rses, exclude_rses, vos)
-    if not rses_to_process:
-        logging.log(logging.ERROR, 'Reaper: No RSEs found. Exiting.')
-        return
-
-    logging.log(logging.INFO, 'Reaper: This instance will work on RSEs: %s', ', '.join([rse['rse'] for rse in rses_to_process]))
-
     logging.log(logging.INFO, 'starting reaper threads')
     threads_list = [threading.Thread(target=reaper, kwargs={'once': once,
                                                             'rses': rses,
--- a/bin/rucio-atropos
+++ b/bin/rucio-atropos
@@ -13,9 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-'''
+"""
 Atropos Daemon : End the life of the rules according to the Lifetime Model
-'''
+"""
 
 import argparse
 import datetime
--- a/bin/rucio-automatix
+++ b/bin/rucio-automatix
@@ -13,9 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-'''
+"""
 Automatix (Dataset injector) Daemon
-'''
+"""
 
 import argparse
 import signal
--- a/bin/rucio-cache-client
+++ b/bin/rucio-cache-client
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 """
-    Rucio Cache client.
+Rucio Cache client.
 """
 
 import argparse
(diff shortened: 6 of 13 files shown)


lib/rucio/daemons/reaper/reaper.py
```diff
@@ -712,10 +711,10 @@ def run(
...
 
     logging.log(logging.INFO, 'main: starting processes')
-    rses_to_process = get_rses_to_process(rses, include_rses, exclude_rses, vos)
-    if not rses_to_process:
-        logging.log(logging.ERROR, 'Reaper: No RSEs found. Exiting.')
-        return
+
+    while not (rses_to_process := get_rses_to_process(rses, include_rses, exclude_rses, vos)):
+        logging.warning(f"Reaper: No RSEs found. Sleeping for {sleep_time} seconds and retrying.")
+        time.sleep(sleep_time)
```
dchristidis: I would have approached this differently. 

In this scope, `rses_to_process` is used exclusively for one logging entry. It has no other functional purpose. However, this entry is misleading: it outputs the list of RSEs at the time the daemon starts, but that may change during its operation. In addition, the actual list of RSEs is calculated in `run_once()`. Consequently, I would remove this entirely.

Note that using f-strings and other formatting constructs in calls to `logging.log()` is an [anti-pattern](https://docs.astral.sh/ruff/rules/logging-f-string/) which we’re actively moving away from.

lib/rucio/daemons/reaper/reaper.py
```diff
@@ -131,7 +131,7 @@ def get_rses_to_process(
         rses_to_process = [rse for rse in rses_to_process if rse not in excluded_rses]
 
     REGION.set(cache_key, rses_to_process)
-    logging.log(logging.INFO, 'Reaper: This instance will work on RSEs: %s', ', '.join([rse['rse'] for rse in rses_to_process]))
+
```
dchristidis: Given my previous comment, this entry is not redundant.

lib/rucio/daemons/reaper/reaper.py
```diff
@@ -690,7 +690,6 @@ def run(
 
     :param threads:                The total number of workers.
     :param chunk_size:             The size of chunk for deletion.
-    :param threads_per_worker:     Total number of threads created by each worker.
```
dchristidis, line 693: Please specify in the commit message that this affects only the docstring.

bin/rucio-reaper
```diff
@@ -50,7 +51,7 @@ def get_parser():
                         help='The delay (seconds) to query replicas in BEING_DELETED state.')
     parser.add_argument('--scheme', action="store", default=None, type=str,
                         help='Force the reaper to use a particular protocol/scheme, e.g., mock')
-    parser.add_argument("--auto_exclude_threshold", "--auto-exclude-threshhold", new_option_string="--auto-exclude-threshhold", action=StoreAndDeprecateWarningAction, default=100, type=int,
+    parser.add_argument("--auto_exclude_threshold", "--auto-exclude-threshold", new_option_string="--auto-exclude-threshold", action=StoreAndDeprecateWarningAction, default=100, type=int,
```
dchristidis: I think you might have misunderstood the intention here. `--auto_exclude_threshold` is already the proper option, but `--auto-exclude-threshhold` is still accepted. As such, the typo issue is already fixed, but in a backwards-compatible way.

I would say that it has been long enough that we can remove the old option, but this has to be done in a separate issue and commit that is applied in a major release.

bin/rucio-admin
```diff
@@ -12,6 +12,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
```
dchristidis, line 15: No argument against what you’re trying to achieve here (pluse the use of double quotes), but please explain in your commit message the _why_ (e.g. referencing PEP 257).

dchristidis (changes requested):
Please follow the [Contributing Guide](https://rucio.cern.ch/documentation/contributing#4-commit-your-changes) on how to structure commit messages.

bin/rucio-admin
lobis: Addressed in https://github.com/rucio/rucio/pull/7472/commits/9c326130168f95c935ffe4b4d8025e32880e1ff4

lib/rucio/daemons/reaper/reaper.py
lobis: Addressed in https://github.com/rucio/rucio/pull/7472/commits/2b9210c2523745aa8fc737be205ecdb8e768274a

bin/rucio-reaper
lobis: Addressed in https://github.com/rucio/rucio/pull/7475

lib/rucio/daemons/reaper/reaper.py
lobis: Reverted

lib/rucio/daemons/reaper/reaper.py
lobis: Rewrote in https://github.com/rucio/rucio/pull/7472/commits/0cf34c28ea38c19f3d22827bb1c612f63d4225d9

lobis:
Looks like there are some failing tests (https://github.com/rucio/rucio/actions/runs/13760417106/job/38475165301?pr=7472), but it does not look like it has anything to do with the contents of this PR.

rdimaio:
> Looks like there are some failing tests (https://github.com/rucio/rucio/actions/runs/13760417106/job/38475165301?pr=7472), but it does not look like it has anything to do with the contents of this PR.

`test_judge_evaluator` tests can be flaky - re-running the failed jobs generally works

bin/rucio-reaper
dchristidis: Upon revisiting this, I misspoke, but it’s best to tackle it in its own dedicated issue.

lib/rucio/daemons/reaper/reaper.py
```diff
@@ -711,12 +711,10 @@ def run(
...
     logging.log(logging.INFO, 'main: starting processes')
-    rses_to_process = get_rses_to_process(rses, include_rses, exclude_rses, vos)
-    if not rses_to_process:
-        logging.log(logging.ERROR, 'Reaper: No RSEs found. Exiting.')
-        return
 
-    logging.log(logging.INFO, 'Reaper: This instance will work on RSEs: %s', ', '.join([rse['rse'] for rse in rses_to_process]))
+    while not get_rses_to_process(rses, include_rses, exclude_rses, vos):
+        logging.log(logging.WARNING, 'Reaper: No RSEs found. Sleeping for %s seconds and retrying.', sleep_time)
+        time.sleep(sleep_time)
```
dchristidis: Repeating my first comment, what is the current purpose of this block?

Let’s say that you take it out entirely. Inside `run_once()`, `rses_to_process` will be calculated, and if it is empty, if will write a log entry and sleep (note: specifying 30 seconds is incorrect and the level should have been a warning and not an error). If not empty, it will even write the list of RSEs it will work on and get to work. In short, the Reaper will likely work just fine.

So, what are we losing?

lobis: Now I understand better what you mean, addressed it in https://github.com/rucio/rucio/pull/7472/commits/58940f80a3a14b8df136916afec4bc361cdee07f.

Now threads are spawned even if no RSE are available.

dchristidis (changes requested):
The first two commits are acceptable.

The third could use a body, but I won’t push you.

The fourth requires a bit of work. There are now two log entries, `main: starting processes` and `starting threads`. Why do we need both? And the change to the `run_once()` function is unrelated; it should be in a separate commit (and note my comment that `logging.ERROR` is too high of a level).

And I must remind you of the Contributing Guide. The component name is `Deletion` and the commit subject should end with the issue number. Actually, I’m a proponent of applying the latter to all commits, but this is not followed consistently, so I won’t object.

lobis:
> The first two commits are acceptable.
> 
> The third could use a body, but I won’t push you.
> 
> The fourth requires a bit of work. There are now two log entries, `main: starting processes` and `starting threads`. Why do we need both? And the change to the `run_once()` function is unrelated; it should be in a separate commit (and note my comment that `logging.ERROR` is too high of a level).
> 
> And I must remind you of the Contributing Guide. The component name is `Deletion` and the commit subject should end with the issue number. Actually, I’m a proponent of applying the latter to all commits, but this is not followed consistently, so I won’t object.

Hopefully changes are addressed in the latest rebase.

dchristidis (approved):

PR #4885, opened 2021-10-05
Clients: Use kebab-case for all command-line options #4865

joeldierkes:
Kebab-case is the de facto GNU and Linux standard of naming command line
arguments. Since we are not consistent on naming our command line arguments we
adapt this style.

PR #7475, opened 2025-03-10
Reaper: Remove deprecated alternative spelling for options

### LINKED ISSUE #7474: Reaper - Remove deprecated aternative spelling for options
Options `--auto-exclude-threshold` and `--auto-exclude-timeout` currently support an alternative spelling using underscores / with a typo for backwards compatibility. It may be time to remove them.

lobis:
Closes [Reaper - Remove deprecated aternative spelling for options #7474](https://github.com/rucio/rucio/issues/7474).

### DIFF
--- a/bin/rucio-reaper
+++ b/bin/rucio-reaper
@@ -19,7 +19,6 @@
 import argparse
 import signal
 
-from rucio.common.utils import StoreAndDeprecateWarningAction
 from rucio.daemons.reaper.reaper import run, stop
 
 
@@ -32,7 +31,7 @@ def get_parser():
                         help='Runs one loop iteration')
     parser.add_argument("--threads", action="store", default=1, type=int,
                         help='Concurrency control: number of threads')
-    parser.add_argument("--chunk_size", "--chunk-size", new_option_string="--chunk-size", action=StoreAndDeprecateWarningAction, default=100, type=int,
+    parser.add_argument("--chunk-size", action="store", default=100, type=int,
                         help='The size used for a bulk deletion on on RSE')
     parser.add_argument('--sleep-time', action="store", default=60, type=int,
                         help='Minimum time between 2 consecutive cycles')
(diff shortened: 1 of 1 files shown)


rdimaio (dismissed):

dchristidis (commented):
The second case is clear. The first, not so much.

The original goal was to replace `--auto_exclude_threshold` with `--auto-exclude-threshold` (i.e. the underscores with dashes, which is considered a better practice). Sadly, this was introduced with a typo, as you correctly observed.

Your commit suggests that it’s only removing a deprecated option, but in practice you’re introducing a new breaking change in the CLI without a deprecation notice.

And on a different note, I would have prepared to see this done for all other options affected in PR #4885, not solely the Reaper.

lobis:
> The second case is clear. The first, not so much.
> 
> The original goal was to replace `--auto_exclude_threshold` with `--auto-exclude-threshold` (i.e. the underscores with dashes, which is considered a better practice). Sadly, this was introduced with a typo, as you correctly observed.
> 
> Your commit suggests that it’s only removing a deprecated option, but in practice you’re introducing a new breaking change in the CLI without a deprecation notice.

While technically a breaking change, one could argue that the misspelled option should't work in the first place. My gut feeling is that no users are using the misspelled option anyways, but I guess it's impossible to tell for sure. Even if they would, the error they would get should be straightforward to correct.

Personally I would be inclined to update the commit message to make clear that this is technically a breaking change but I would not add a deprecation warning for this (and then in the future remove the misspelled option), I think the overhead is not worth it.

@dchristidis let me know how I should proceed: should I just rewrite the commit message or add the deprecation warning?

dchristidis:
I agree, a deprecation notice does not feel that necessary. But the changes to the offending option should be split into two commits: one to remove the deprecated option, and one to correct the remaining option. ~~And, as I wrote, the former should not be specific to the Reaper.~~

Edit: The others are mostly in the client and there’s a lot of changes related to the CLI structure happening right now.

dchristidis (dismissed):

lobis:
Forgot to remove unused dependency. Now it is ready to be merged.

rdimaio (approved):

Issue #7458, opened 2025-03-05
rucio-admin account get-limits accepts non-existing accounts


airnandez:
### Description

I noticed that the command `rucio-admin account get-limits` accepts a non-existent account and returns a response that could be valid, which is misleading when I make typos. 

### Steps to reproduce

Example with a non existent account `xxxxx`:

```
❯ rucio ping
35.6.0

❯ rucio-admin account list | grep -c 'xxxxx'
0

❯ rucio-admin account get-limits xxxxx MY_RSE local
Quota on MY_RSE for xxxxx : 0.0 B

❯ rucio-admin account get-limits xxxxx MY_RSE global
Quota on MY_RSE for xxxxx : 0.0 B
```

### Rucio Version

35.6.0

dchristidis:
I’m not sure I agree with this being categorised as an issue in the Clients. Wouldn’t it more proper for the core functions to raise an exception when called with a non-existent account?

rdimaio:
> I’m not sure I agree with this being categorised as an issue in the Clients. Wouldn’t it more proper for the core functions to raise an exception when called with a non-existent account?

Validation of this kind IMO should be performed at the Gateway/API level, rather than core, so I would rather label it as gateway. I agree that the clients label is imprecise here - the issue shows up when using clients, but the root issue is not in the clients themselves

dchristidis:
I’m not sure this is input validation. The issue at hand is that there is a core function with a suboptimal interface: it will return an empty result both when called with an non-existent account and when there is an account that happens to not have any limits.

But we’re not consistent in this regard. `get_rse_usage()` does not complain if the provided RSE does not exist. On the other hand, `list_identities()` will raise an exception if the provided account does not exist. I think I prefer the implementation of the latter, but I won’t object to the former.

The downside of implementing it in the core is the expense that is added with each invocation, even in a context where the parameters are guaranteed to be valid.

rdimaio:
> I’m not sure this is input validation. The issue at hand is that there is a core function with a suboptimal interface: it will return an empty result both when called with an non-existent account and when there is an account that happens to not have any limits.
> 
> But we’re not consistent in this regard. `get_rse_usage()` does not complain if the provided RSE does not exist. On the other hand, `list_identities()` will raise an exception if the provided account does not exist. I think I prefer the implementation of the latter, but I won’t object to the former.
> 
> The downside of implementing it in the core is the expense that is added with each invocation, even in a context where the parameters are guaranteed to be valid.

[...]

PR #7476, opened 2025-03-10
Clients : Handle Non-Existent Accounts in `rucio-admin account get-limits`

0xquark:
Previously, `rucio-admin account get-limits` accepted nonexistent accounts and returned misleading responses. This fix adds a validation check to ensure that the account exists before retrieving limits in `accountclient`.
This PR also adds error handling and a validation test case for the same.

### DIFF
--- a/lib/rucio/client/accountclient.py
+++ b/lib/rucio/client/accountclient.py
@@ -256,6 +256,7 @@ class AccountClient(BaseClient):
         :param locality:       The scope of the account limit. 'local' or 'global'.
         """
 
+        self.get_account(account)
         if locality == 'local':
             return self.get_local_account_limit(account, rse_expression)
         elif locality == 'global':
--- a/lib/rucio/client/commands/bin_legacy/rucio_admin.py
+++ b/lib/rucio/client/commands/bin_legacy/rucio_admin.py
@@ -36,10 +36,7 @@ from rucio import version
 from rucio.client.commands.utils import exception_handler, get_client, setup_gfal2_logger, signal_handler
 from rucio.client.richclient import MAX_TRACEBACK_WIDTH, MIN_CONSOLE_WIDTH, CLITheme, generate_table, get_cli_config, get_pager, print_output, setup_rich_logger
 from rucio.common.constants import RseAttr
-from rucio.common.exception import (
-    ReplicaNotFound,
-    RSEOperationNotSupported,
-)
+from rucio.common.exception import AccountNotFound, ReplicaNotFound, RSEOperationNotSupported
 from rucio.common.extra import import_extras
 from rucio.common.utils import StoreAndDeprecateWarningAction, chunks, clean_pfns, construct_non_deterministic_pfn, extract_scope, get_bytes_value_from_string, parse_response, render_json, setup_logger, sizefmt
 from rucio.rse import rsemanager as rsemgr
@@ -238,11 +235,16 @@ def get_limits(args, client, logger, console, spinner):
     Grant an identity access to an account.
 
     """
-    locality = args.locality.lower()
-    limits = client.get_account_limits(account=args.account, rse_expression=args.rse, locality=locality)
-    for rse in limits:
-        print('Quota on %s for %s : %s' % (rse, args.account, sizefmt(limits[rse], True)))
-    return SUCCESS
+    try:
+        locality = args.locality.lower()
+        limits = client.get_account_limits(account=args.account, rse_expression=args.rse, locality=locality)
+        for rse in limits:
+            print('Quota on %s for %s : %s' % (rse, args.account, sizefmt(limits[rse], True)))
+        return SUCCESS
+
+    except AccountNotFound as e:
+        logger.error(f"Validation Error: {e}")
+        return FAILURE
 
 
 @exception_handler
--- a/tests/test_account_limits.py
+++ b/tests/test_account_limits.py
@@ -14,6 +14,7 @@
 
 import pytest
 
+from rucio.common.exception import AccountNotFound
 from rucio.core import account_limit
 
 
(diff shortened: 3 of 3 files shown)


lib/rucio/client/commands/bin_legacy/rucio_admin.py
```diff
@@ -238,11 +235,20 @@ def get_limits(args, client, logger, console, spinner):
...
+        limits = client.get_account_limits(account=args.account, rse_expression=args.rse, locality=locality)
+        for rse in limits:
+            print('Quota on %s for %s : %s' % (rse, args.account, sizefmt(limits[rse], True)))
+        return SUCCESS
+
+    except AccountNotFound as e:
+        logger.error(f"Validation Error: {e}")
+        return FAILURE
+
+    except Exception as e:
```
voetberg: The `@exception_handler` decorator handles most errors, and doing this overwrites a more specific error. The account not found error is fine, but this second except catch should be removed.

voetberg (changes requested):

lib/rucio/client/accountclient.py
```diff
@@ -256,6 +257,9 @@ def get_account_limits(self, account: str, rse_expression: str, locality: str) -
         :param locality:       The scope of the account limit. 'local' or 'global'.
         """
 
+        existing_accounts = [account_data['account'] for account_data in self.list_accounts()]
+        if account not in existing_accounts:
+            raise AccountNotFound(f"Account '{account}' does not exist.")
```
rdimaio: `list_accounts()`, with no args passed to it, lists all Rucio accounts - this seems unnecessary if we just want to verify that a single account exists.

IMO we should instead call `get_account(account)`, which raises `AccountNotFound` if the account is not found

0xquark: Thanks, that makes sense! I didn't realize we had this method and mistakenly thought list_accounts() was handling some sort of caching.

dchristidis:
> Previously, rucio-admin account get-limits accepted nonexistent accounts and returned misleading responses. This fix adds a validation check to ensure that the account exists before retrieving limits in accountclient.

When adding a replication rule, there’s not validation that the DID exists. The client submits the requests and the server will respond appropriately. Why is this case different?

lib/rucio/client/accountclient.py
```diff
@@ -256,6 +256,7 @@ def get_account_limits(self, account: str, rse_expression: str, locality: str) -
         :param locality:       The scope of the account limit. 'local' or 'global'.
         """
 
+        self.get_account(account)
```
rdimaio, line 259: Based on the conversation here https://github.com/rucio/rucio/issues/7458#issuecomment-2716838516 - it would be best to have this check be done at the core level.

In core, we have four relevant methods:
- `get_local_account_limit`
- `get_local_account_limits`
- `get_global_account_limit`
- `get_global_account_limits`
I think these are not named clearly. I've opened an issue to address this: https://github.com/rucio/rucio/issues/7483

You can see that the methods with 'limit' (singular) wrap the query in a `try/except` with a `except NoResultFound` clause. I think we should:
- make sure that we raise an exception here, rather than returning `None`
- have this same behaviour for the 'limit' (plural) methods.

@dchristidis - thoughts on this?

As an overview of the flow - `get-limits` calls this:
https://github.com/rucio/rucio/blob/62fd7f40ee0ac50becf0866401b00d642b39e6c0/lib/rucio/client/commands/bin_legacy/rucio_admin.py#L233-L245 

Which calls this: 
https://github.com/rucio/rucio/blob/62fd7f40ee0ac50becf0866401b00d642b39e6c0/lib/rucio/client/accountclient.py#L250-L265

Which calls this:
https://github.com/rucio/rucio/blob/62fd7f40ee0ac50becf0866401b00d642b39e6c0/lib/rucio/client/accountclient.py#L313-L327

Which calls this:
https://github.com/rucio/rucio/blob/62fd7f40ee0ac50becf0866401b00d642b39e6c0/lib/rucio/web/rest/flaskapi/v1/accounts.py#L526-L571

In the usage example, an RSE is passed, so it calls this:

https://github.com/rucio/rucio/blob/62fd7f40ee0ac50becf0866401b00d642b39e6c0/lib/rucio/gateway/account_limit.py#L77-L101

which calls this:

https://github.com/rucio/rucio/blob/62fd7f40ee0ac50becf0866401b00d642b39e6c0/lib/rucio/core/account_limit.py#L158-L181

dchristidis: > make sure that we raise an exception here, rather than returning `None`

Beware, it is possible that an account has no quota on some RSE. In which case, `None` is correct.

rdimaio: > > make sure that we raise an exception here, rather than returning `None`
> 
> Beware, it is possible that an account has no quota on some RSE. In which case, `None` is correct.

Then, I guess we can't get around running `get_account` to check for this

dchristidis: You could do a left join of `Account` with `AccountLimit`. It’s one query instead of two, but makes the query more complex to read (I’m not sure a developer looking at this for the first time will get the intention).

rdimaio: > You could do a left join of `Account` with `AccountLimit`. It’s one query instead of two, but makes the query more complex to read (I’m not sure a developer looking at this for the first time will get the intention).

I agree - I think it would be preferable to have a separate query for account existence and one for the account limits, it's clearer and I don't imagine it would make a huge difference

0xquark:
After our discussion, I realized it made sense to add a separate check at the core level to confirm the account exists before querying limits. I’ve added this check in [this commit](https://github.com/rucio/rucio/pull/7522/commits/7c4b6af541314f2012661ca0cebaef57140feb8d).

Geogouz:
What are the plans for this PR? In case we keep working on this, we should also link an issue.

0xquark:
It’s been quite a while, and rebasing everything would be a bit challenging, so I’ll open a new PR with the updates. Closing this one now.
(Also this is related to https://github.com/rucio/rucio/pull/7679)

PR #7479, opened 2025-03-11
clients: add validation for missing 'bytes' field in `add_replicas` 

### LINKED ISSUE #7468: `client.add_replicas` fails with non-explicit error when adding file without bytes field
```
from rucio.common.exception import DataIdentifierNotFound

client = Client()

rse = 'EOS_XRD_2'
scope = 'tutorial'
filename = 'test-register-file-1'

files = [
    {
        "scope": scope,
        "name": filename,
        "bytes": 10485760,
        "pfn": f"root://eosworkspace.cern.ch:1094//eos/workspace/r/rucioit/opendata/EOS_XRD_2/tutorial/71/70/{filename}",
    }
]


client.add_replicas(rse=rse, files=files)
```

If the bytes field is removed, the command fails with a generic 500 error. It took me some time to realise the `bytes` field was actually required.

At least this should return an error directing the user to supply the `bytes` field.

0xquark:
fixes #7468

### DIFF
--- a/lib/rucio/gateway/replica.py
+++ b/lib/rucio/gateway/replica.py
@@ -272,6 +272,11 @@ def add_replicas(rse, files, issuer, ignore_availability=False, vo='def', *, ses
 
     :returns: True is successful, False otherwise
     """
+    for file in files:
+        if "bytes" not in file:
+            raise exception.InvalidObject(
+                f"Missing required field: 'bytes'. The file '{file.get('name', 'unknown')}' must have a 'bytes' field specifying its size."
+            )
     for v_file in files:
         v_file.update({"type": "FILE"})  # Make sure DIDs are identified as files for checking
     validate_schema(name='dids', obj=files, vo=vo)
--- a/tests/test_replica.py
+++ b/tests/test_replica.py
@@ -27,7 +27,7 @@ from werkzeug.datastructures import Headers, MultiDict
 
 from rucio.client.ruleclient import RuleClient
 from rucio.common.constants import RseAttr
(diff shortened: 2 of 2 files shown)


lib/rucio/gateway/replica.py
```diff
@@ -272,6 +272,10 @@ def add_replicas(rse, files, issuer, ignore_availability=False, vo='def', *, ses
 
     :returns: True is successful, False otherwise
     """
+    for file in files:
+        if "bytes" not in file:
+            raise exception.InvalidObject("Missing required field: 'bytes'. Each file must have a 'bytes' field specifying its size.")
```
rdimaio: For debugging purposes, it could be helpful to include the file name in the exception message, so that it's clear which file is causing the issue (or if there's more than one - at least it points to the first offending file)

rdimaio (dismissed):

rdimaio (approved):

bari12:
I wonder if this should be rather changed in the schema? in the `validate_schema` step?
RIght now, we check against the `dids` schema, which does not work, but we could make a new one. I am not sure if this would improve the error message though, I suppose so.

0xquark:
IMO if we would be enforcing this "bytes" check in other places as well then we should do it in `validate_schema` step by creating a new one (considering modifying `dids` schema might break something) or else it would increase the code complexity only if we are doing this one-time (and the error message looks good to me but might be more general in case of a schema).

bari12:
I think there might be other locations where "replicas" are added yes.

bari12:
Let's go the schema path then. I will close this in the meanwhile

PR #7487, opened 2025-03-13
Policies: Improve `parse_scope_name` error message

### LINKED ISSUE #7486: Add better error message for scope name splitting
I spend more time than I would like to admit into figuring out why uploads / downloads where not working, it was due to the DID I was submitting not being correctly split into scope and name by the regex.

This was due to the fact that the regex does not act on the actual did string, but on a modification of it. I can't speak as to why this is needed (I'm sure there is a good reason) but at least the comments / error messages should be more clear.

lobis:
Add a more useful error message.

No functional changes.

### DIFF
--- a/lib/rucio/web/rest/flaskapi/v1/common.py
+++ b/lib/rucio/web/rest/flaskapi/v1/common.py
@@ -210,14 +210,20 @@ def parse_scope_name(scope_name: str, vo: Optional[str]) -> tuple[str, ...]:
     :raises ValueError: when scope_name could not be parsed.
     :returns: a (scope, name) tuple.
     """
+
     if not vo:
         vo = 'def'
 
-    # why again does that regex start with a slash?
-    scope_regex = re.match(get_schema_value('SCOPE_NAME_REGEXP', vo), '/' + scope_name)
+    # The ':' in did is replaced by '/', also an '/' is added. Why?
+    pattern = get_schema_value('SCOPE_NAME_REGEXP', vo)
+    text = '/' + scope_name
+
+    scope_regex = re.match(pattern, text)
     if scope_regex is None:
-        raise ValueError('cannot parse scope and name')
-    return scope_regex.group(1, 2)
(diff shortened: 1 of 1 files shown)


lib/rucio/web/rest/flaskapi/v1/common.py
```diff
@@ -210,14 +210,20 @@ def parse_scope_name(scope_name: str, vo: Optional[str]) -> tuple[str, ...]:
     :raises ValueError: when scope_name could not be parsed.
     :returns: a (scope, name) tuple.
     """
+
     if not vo:
         vo = 'def'
 
-    # why again does that regex start with a slash?
-    scope_regex = re.match(get_schema_value('SCOPE_NAME_REGEXP', vo), '/' + scope_name)
+    # The ':' in did is replaced by '/', also an '/' is added. Why?
```
rdimaio, line 217: > The ':' in did is replaced by '/'

This is done at the client level, e.g.: 

https://github.com/rucio/rucio/blob/62fd7f40ee0ac50becf0866401b00d642b39e6c0/lib/rucio/client/didclient.py#L116

I would argue that this is not necessary in cases where we need to pass `scope` and `name` together anyway, like in the case of `add_did`.

https://github.com/rucio/rucio/blob/62fd7f40ee0ac50becf0866401b00d642b39e6c0/lib/rucio/client/didclient.py#L116

 We might as well pass them with the usual `:` separator. I think it would be great if someone that had more insight into designing this could clarify why this decision was made at the time. @bari12 thoughts?

voetberg (approved):

lib/rucio/web/rest/flaskapi/v1/common.py
bari12: I think the `/` is purely added because the `SCOPE_NAME_REGEXP` always starts with a `/` because the url always starts with a `/` there. So to make sure the `re.match` call between the `SCOPE_NAME_REGEXP` and the `scope_name` matches.

PR #7489, opened 2025-03-13
Policies: full diff-based schema support #6842

### LINKED ISSUE #6842: "diff"-based policies, instead of fully replicating the base policy package code

jamesp-epcc:
This PR reworks the schema code so that proper diff-based schemas are possible. When a schema value is required, the code now walks the schema structure and substitutes included values dynamically. This means that it's possible to have a schema value in a policy package include a value from the generic fallback module, or vice versa. The policy package will always take priority.

I am open to suggestions regarding the syntax. I picked a string prefix that seems unlikely to conflict with anything but I'm happy to change it if people have better suggestions.

As part of this work, the `validate_schema` function has been moved out of the schema modules and into `__init__.py`. This means that the behaviour of the validation function itself can no longer be customised, but I don't think that should be a problem. I have also refactored the policy module loading a little so that single and multi VO cases now mostly share the same code path.

I will also submit an update to the policy package documentation to reflect this change.

### DIFF
--- a/lib/rucio/common/schema/__init__.py
+++ b/lib/rucio/common/schema/__init__.py
@@ -19,6 +19,8 @@ from configparser import NoOptionError, NoSectionError
 from os import environ
 from typing import TYPE_CHECKING, Any
 
+from jsonschema import ValidationError, validate
+
 from rucio.common import config, exception
 from rucio.common.plugins import check_policy_package_version
 
@@ -140,10 +142,16 @@ def load_schema_for_vo(vo: str) -> None:
 def validate_schema(name: str, obj: Any, vo: str = 'def') -> None:
     if vo not in schema_modules:
         load_schema_for_vo(vo)
-    if not hasattr(schema_modules[vo], 'validate_schema'):
-        _get_generic_schema_module().validate_schema(name, obj)
-        return
-    schema_modules[vo].validate_schema(name, obj)
+    if hasattr(schema_modules[vo], 'SCHEMAS') and name in schema_modules[vo].SCHEMAS:
+        schema = schema_modules[vo].SCHEMAS.get(name, {})
+    else:
+        # if schema not available in VO module, fall back to generic module
+        schema = _get_generic_schema_module().SCHEMAS.get(name, {})
+    try:
+        if obj:
+            validate(obj, schema)
+    except ValidationError as error:  # NOQA, pylint: disable=W0612
+        raise exception.InvalidObject(f'Problem validating {name}: {error}')
 
 
 def get_schema_value(key: str, vo: str = 'def') -> Any:
--- a/lib/rucio/common/schema/generic.py
+++ b/lib/rucio/common/schema/generic.py
@@ -12,9 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from jsonschema import ValidationError, validate
-
-from rucio.common.exception import InvalidObject
 
 ACCOUNT_LENGTH = 25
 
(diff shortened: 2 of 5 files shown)


lib/rucio/common/schema/__init__.py
```diff
@@ -50,61 +52,53 @@ def _get_generic_schema_module():
...
+            # allow adding or subtracting a constant
+            constant = 0
+            if "-" in name:
+                pos = name.find("-")
+                constant = -int(name[pos + 1:].strip())
+                name = name[:pos].strip()
+            if "+" in name:
+                pos = name.find("+")
+                constant = int(name[pos + 1:].strip())
+                name = name[:pos].strip()
```
rdimaio: I'm not sure what the use case would be for this add/subtract feature - I think it would be easier and simpler to set integer schema values directly, rather than adding / subtracting to a base value. 

e.g. I saw the multi_vo schema doing `"%%ACCOUNT_LENGTH - 4"`, where `ACCOUNT_LENGTH` is `29` - but at that point I think it would just make more sense to directly set it to `25`.

rdimaio (commented):
What is the string prefix needed for?

rdimaio:
I saw this https://github.com/rucio/documentation/pull/495/files - I assume the prefixes are needed for a policy package schema to reference the generic schema. I think it would be best not to do this for a couple reasons:
1. There is no clear view of what the value of a referenced generic name is from within the custom schema - e.g. if I have a custom schema that does `%%ACCOUNT_LENGTH - 4`, I don't have a clear view of what the result is within my schema. It would be easier to write my custom `ACCOUNT_LENGTH` directly.
2. Values in the generic schema could change, which could affect the custom schema.

In my opinion, I would prefer if the logic was like this:
1. the code needs to validate the schema for a certain object
2. `validate_schema` looks for the schema for that object in its preferred schema module
3. if not found, it falls back to `generic`

lib/rucio/common/schema/__init__.py
jamesp-epcc: Yes, this was added because the multi vo schema subtracts from the lengths (presumably to leave space to put the vo name in as well), but it might not be worth the complexity to support it.

jamesp-epcc:
The prefix is to differentiate values that might be overridden by another module from simple string constants. As well as allowing the policy package to reference values in the generic schema (which I agree could potentially cause problems if the generic schema values change in a way that breaks the policy package) they're also needed to allow the policy package to replace only parts of the generic schema. For example, if the policy package redefines `R_SCOPE`, then validating against `R_SCOPE` will work correctly. But validating against `R_DID` won't work properly without this because unless we make some provision for parts to be referenced dynamically, the generic schema's `R_SCOPE` will be used within `R_DID` instead of the policy package's.

bari12:
I agree with @rdimaio here.
To me this seems a little bit too complicated. You can do more complex things, that's true, but at the cost of clarity, which might lead to more errors.

jamesp-epcc:
I agree there are downsides to the complex approach and it might be overkill. I will re-implement using the simpler logic Riccardo suggests.

jamesp-epcc:
I've re-implemented using the simpler approach as discussed.

lib/rucio/common/schema/__init__.py
```diff
@@ -140,10 +142,16 @@ def load_schema_for_vo(vo: str) -> None:
...
+    if hasattr(schema_modules[vo], 'SCHEMAS') and name in schema_modules[vo].SCHEMAS:
+        schema = schema_modules[vo].SCHEMAS.get(name, {})
+    else:
+        # if schema not available in VO module, fall back to generic module
+        schema = _get_generic_schema_module().SCHEMAS.get(name, {})
+    try:
+        if obj:
+            validate(obj, schema)
+    except ValidationError as error:  # NOQA, pylint: disable=W0612
+        raise exception.InvalidObject(f'Problem validating {name}: {error}')
```
rdimaio, line 154: Validation only happens `if obj` is `True`, so I think it could be more optimal to have this `if obj` check be done as the first thing in the function, so that we avoid loading the schema and preparing for validation if we don't have `obj`

rdimaio (approved):
Really clean implementation, very nice!

rdimaio:
Followed up in https://github.com/rucio/rucio/pull/7555 to fix merge conflicts for rc1

PR #7490, opened 2025-03-14
Testing: Codecov, turn project coverage off

rdimaio:
We will revise this in the future,
but for now we should keep this off,
as it is not needed for any useful checks.
Patch coverage is the main feature we want to use, in order to check coverage in PRs.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,5 +1,6 @@
 coverage:
   status:
+    project: off
     patch:
       default:
         paths:


codecov.yml
```diff
@@ -1,5 +1,6 @@
 coverage:
   status:
+    project: off
```
rdimaio, line 3: for reference: https://docs.codecov.com/docs/commit-status#disabling-a-status

rdimaio (commented):
Ruff errors unrelated, they were introduced in some other PR

voetberg (approved):
:+1:

PR #7500, opened 2025-03-14
Docker & Kubernetes: Introduce `bootstrap_dev.sh` for setting up a local Dev/Testing environment; Fix #7462

### LINKED ISSUE #7462: Fix psycopg dependency install in server and dev installs
Raised by @mlassnig during the 2025-02-20 Rucio meeting: when resetting the database, there's a `psycopg` issue, which requires a manual `pip install psycopg[binary,pool]` to be fixed.

https://github.com/rucio/containers/blob/105a57a4e7f07343d0ae490d36bf264bb2328400/dev/Dockerfile#L81 - likely this needs to be changed to install `psycopg[binary,pool]` instead

Though in theory `/tmp/rucio/requirements/requirements.dev.txt` should be present, so this shouldn't be entered in the first place

Geogouz:
The introduced bootstrap script is designed to properly deploy a Rucio Dev/Testing environment in an ephemeral local branch.

Previously, the `docker-compose.yml` file always fetched the `latest` images, and assumed that (the tip of) master requirements were installed there correctly. However, this assumption (along with other issues) was problematic, as discussed [here](https://github.com/rucio/containers/issues/379). This script should now ensure that the development environment correctly corresponds to the version of the local branch.

The script also allows selecting a specific version for deployment (accepting for example `<release-id>`, `latest` or `master` for options), leveraging the various releases available on DockerHub. However, the functionality of `<release-id>` will work properly only for releases after the merging of current PR and [this PR](https://github.com/rucio/containers/pull/383) too.

For now you can try things like:
```
./tools/bootstrap_dev.sh --master --profile storage --profile monitoring
./tools/bootstrap_dev.sh --latest --profile storage
./tools/bootstrap_dev.sh --master
./tools/bootstrap_dev.sh --profile
```

The `docker-compose.yml` file now allows also users to define a custom Rucio `DOCKER_REPO` (for example, in my case for testing, I use images I have prepared and pushed to my Docker account).

There are several robustness checks and handling of edge cases in place, yet feel free to test it and report any issues that may require fixing. Moreover, in the future, we could enhance this script to properly invoke the `run_tests.sh` script in an idempotent way, and/or also integrate the functionality of other scripts we provide for preparing a contribution branch (described [here](https://rucio.github.io/documentation/contributing)). This way, we can have everything in one place as a toolbox.

Lastly, I have included in the script, documentation for those who want to understand its exact functionality. However, these details should probably also be added to our developer documentation page: [Setting Up a Demo](https://rucio.github.io/documentation/operator/setting_up_demo).

### DIFF
--- a/etc/docker/dev/docker-compose.yml
+++ b/etc/docker/dev/docker-compose.yml
@@ -1,8 +1,11 @@
 name: rucio
 
 services:
+  # ------------------------------------------------------------------
+  # Main Rucio dev containers
+  # ------------------------------------------------------------------
   rucioclient:
-    image: docker.io/rucio/rucio-dev:latest-alma9
+    image: "docker.io/${DOCKER_REPO:-rucio}/rucio-dev${RUCIO_TAG:+:release-${RUCIO_TAG}}"
     platform: linux/amd64
     entrypoint: ["/rucio_source/etc/docker/dev/rucio/entrypoint.sh"]
     command: ["sleep", "infinity"]
@@ -29,9 +32,10 @@ services:
       - REQUESTS_CA_BUNDLE=/etc/pki/tls/certs/ca-bundle.crt
       - RUCIO_SOURCE_DIR=/rucio_source
       - RDBMS
+
   rucio:
-    container_name: dev_rucio_1
-    image: docker.io/rucio/rucio-dev:latest-alma9
+    container_name: dev-rucio-1
+    image: "docker.io/${DOCKER_REPO:-rucio}/rucio-dev${RUCIO_TAG:+:release-${RUCIO_TAG}}"
     platform: linux/amd64
     entrypoint: ["/rucio_source/etc/docker/dev/rucio/entrypoint.sh"]
     command: ["httpd","-D","FOREGROUND"]
@@ -56,6 +60,10 @@ services:
       - REQUESTS_CA_BUNDLE=/etc/pki/tls/certs/ca-bundle.crt
       - RUCIO_SOURCE_DIR=/rucio_source
       - RDBMS
+
+  # ------------------------------------------------------------------
+  # Database, MQ, metrics, and other external services
+  # ------------------------------------------------------------------
   ruciodb:
     image: docker.io/postgres:14
     environment:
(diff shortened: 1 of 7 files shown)


etc/docker/dev/docker-compose.yml
```diff
@@ -372,8 +422,6 @@ services:
       - IAM_FORWARD_HEADERS_STRATEGY=native
       - IAM_KEY_STORE_LOCATION=file:/keystore.jwks
       - IAM_JWK_CACHE_LIFETIME=21600
-      # - IAM_X509_TRUST_ANCHORS_DIR=/etc/grid-security/certificates
-      # - IAM_X509_TRUST_ANCHORS_REFRESH=14400
```
rdimaio: Why were these removed?

tools/bootstrap_dev.sh
```diff
@@ -0,0 +1,532 @@
...
+#     HELPER: Check that we have an '$UPSTREAM_REMOTE' pointing to Rucio
+# ---------------------------------------------------------------------------
+function ensure_upstream_exists() {
+  if ! git remote get-url "$UPSTREAM_REMOTE" >/dev/null 2>&1; then
+    echo "ERROR: No '$UPSTREAM_REMOTE' remote found. Please set up your fork properly, e.g.:"
+    echo "       git remote add upstream git@github.com:rucio/rucio.git"
+    echo "       git fetch --all"
+    exit 1
+  fi
+}
```
rdimaio: Could we run these commands automatically (similarly to what we do in `ensure_tag_fetched()`) as part of the script in case UPSTREAM_REMOTE is not found? (to avoid making the user have to run them manually and re-run the script)

rdimaio (dismissed):
This is really well documented and structured, nice job!

etc/docker/dev/docker-compose.yml
Geogouz: After deploying the test environment using the `iam` profile, the container's logs (via `docker logs dev-indigoiam-login-service-1`) suggest that Indigo IAM starts successfully (without complaining about any missing or misconfigured X.509 trust anchors). So since these were also commented by default, I thought of just discarding those and if someone would like to play with a different configuration that uses those, they could just add them manually. But maybe this happens indeed more often than I think. Dunno.

tools/bootstrap_dev.sh
Geogouz: We can indeed add automation here for convenience but not sure whether the user would prefer to use ssh or https. Maybe we can add some prompt interaction to ask them about how they would like to set this up. Or is this an overkill?

rdimaio (approved):

PR #7502, opened 2025-03-17
Transfers: Add staging metadata and tests

### LINKED ISSUE #7451: Activity passing in Tape REST API during tape staging missing for  davs/https protocol missing
### Description

Data management teams that have switched to HTTPS/DAVS protocol are now using the tape REST API for tape staging.

In xrootd protocol activity was coming for free on the xrdfs prepare url, but it is not the case anymore with the tape REST API where it has now to be passed in the specific staging_metadata header.

FTS performs this header packaging on behalf of Rucio, but Rucio needs to pass the header json content to FTS via `--staging-metadata` FTS job submission option.

The content of this header is specified in the Tape REST API specifications, but the initial proposal needs to evolve to a simpler model.
Indeed initial proposal required HTTP Tape REST site name to be passed in the staging metadata: this is likely to generate various issues for no good reason and would require this HTTP tape rest sitename as a new RSE attribute for tape endpoints...

The simpler proposal would be to pass "default" in place of site name (agreed with Mihai and already implemented in EOSCTA).

Basically to pass rucio activity to staging http jobs, rucio would have to pass the following additional parameters at job submission time:

```bash
$ fts-rest-transfer-submit --staging-metadata '{"default": {"activity": "rucio-activity"}}' --overwrite --bring-online 86400 -s https://fts3-pilot.cern.ch:8446/ https://eosctaatlaspps.cern.ch:8444/eos/ctaatlaspps/ftstest/IBM1L9/ftstest1/1/test01000010 root://eospublic.cern.ch:1094//eos/opstest/dteam/DMC-1402.test
```

in this example eosctaatlaspps will map received actvity named "rucio-activity" and use the matching activity-mount-rule to appy corresponding activity staging mount priority to this file.

This will notably allow high priority staging requests to jump ahead of the queue during read contention periords (like data taking) or to pass in front of multi PBs massive staging campaigns.

### Motivation

This will notably allow high priority staging requests to jump ahead of the queue during read contention periords (like data taking) or to pass in front of multi PBs massive staging campaigns.

### Change

_No response_

rdimaio:
fix #7451 

Note: end-to-end integ test failure seems to be unrelated, looks like a containers issue that should be resolved by merging another PR in progress

### DIFF
--- a/lib/rucio/transfertool/fts3.py
+++ b/lib/rucio/transfertool/fts3.py
@@ -1049,6 +1049,9 @@ class FTS3Transfertool(Transfertool):
             for plugin in self.tape_metadata_plugins:
                 t_file = deep_merge_dict(source=plugin.hints(t_file['metadata']), destination=t_file)
 
+        if t_file['metadata']['src_type'] == 'TAPE':
+            t_file['staging_metadata'] = {"default": {"activity": rws.activity}}
+
         return t_file
 
     def submit(self, transfers: "Sequence[DirectTransfer]", job_params: dict[str, str], timeout: Optional[int] = None) -> str:
--- a/tests/test_transfer_plugins.py
+++ b/tests/test_transfer_plugins.py
@@ -334,3 +334,62 @@ def test_no_metadata_if_dest_is_not_tape(did_factory, rse_factory, root_account)
 
     assert not mock_metadata_plugin.called
     assert "archive_metadata" not in job_params
+
+
+def test_staging_metadata_if_src_is_tape(did_factory, rse_factory, root_account):
+    """
+    If the source RSE is tape,
+    staging metadata should be included.
+    """
+
+    mock_did = did_factory.random_file_did()
+    transfer_path = _make_transfer_path(mock_did, rse_factory, root_account, src_rse_is_tape=True)
+
+    fts3_tool = FTS3Transfertool(TEST_FTS_HOST)
+
+    job_params = build_job_params(
+        transfer_path=transfer_path,
+        bring_online=None,
+        default_lifetime=None,
+        archive_timeout_override=None,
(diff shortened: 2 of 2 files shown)


lib/rucio/transfertool/fts3.py
```diff
@@ -1049,6 +1049,9 @@ def _file_from_transfer(self, transfer: "DirectTransfer", job_params: dict[str,
             for plugin in self.tape_metadata_plugins:
                 t_file = deep_merge_dict(source=plugin.hints(t_file['metadata']), destination=t_file)
 
+        if t_file['metadata']['src_type'] == 'TAPE':
+            t_file['staging_metadata'] = {"default": {"activity": "rucio-activity"}}
```
voetberg: Is this option automatically enabled on all FTS3 instances? We don't want tape transferred rejected on this bias.

rdimaio: If implemented this way, yeah it'd be automatically enabled on every transfer from tape. We could make it configurable by having a flag in the config to enable this, or maybe it should be a RSE attribute (but maybe we should avoid adding more attributes when needed). Not sure if this is something that should always be enabled as part of the Rucio->FTS interaction though. @bari12 thoughts?

bari12: I think we can leave this by default, but I think we need to check with Mihai if this might have any negative effects for tapes who do not need/use it.

mpatrascoiu: Hello,

The Tap[e REST API makes the staging metadata optional. Storages shouldn't be bothered by it, but in case some are, then you'd need an "per-RSE" flag in Rucio.

As for FTS, we are ok with the metadata being there.

voetberg (approved):

PR #7504, opened 2025-03-17
Clients: Change `did show --parent` to `did list --parent`

### LINKED ISSUE #7233: New CLI: retrieving the list of parent DIDs should be under the ‘list’ command
### Description

The ‘show’ subcommand is supposed to be used for retrieving all the details of one specific entity. A DID can have many parents and listing them provides a very limited set of details. As such, `rucio did show --parent` should become `rucio did list --parent`.

### Motivation

_No response_

### Change

_No response_

voetberg:
Closes #7233

### DIFF
--- a/lib/rucio/client/commands/did.py
+++ b/lib/rucio/client/commands/did.py
@@ -32,6 +32,12 @@ class DID(CommandBase):
         parser.add_argument("--filter", help="Filter arguments in form `key=value,another_key=next_value`. Valid keys are name, type")
         parser.add_argument("--short", action="store_true", help="Just dump the list of DIDs")
         parser.add_argument("-d", "--did", nargs=1, help="Data IDentifier pattern")
+        parser.add_argument("--parent", action="store_true", help="List the parents of the DID")
+        # Both non-functional, but list_parents complains if not present
+        # TODO: Correct these options in rucio::list-parent-dids
+        # Issue 7230 https://github.com/rucio/rucio/issues/7230
+        parser.add_argument("--pfn", dest="pfns", nargs="+", help=argparse.SUPPRESS)
+        parser.add_argument("--guid", dest="guids", nargs="+", help=argparse.SUPPRESS)
 
     def add_namespace(self, parser: "ArgumentParser") -> None:
         parser.add_argument("--type", dest='dtype', choices=("container", "dataset"), help="Add collection type DID")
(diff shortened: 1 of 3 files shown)


lib/rucio/client/commands/did.py
```diff
@@ -32,6 +32,11 @@ def list_namespace(self, parser: "ArgumentParser") -> None:
         parser.add_argument("--filter", help="Filter arguments in form `key=value,another_key=next_value`. Valid keys are name, type")
         parser.add_argument("--short", action="store_true", help="Just dump the list of DIDs")
         parser.add_argument("-d", "--did", nargs=1, help="Data IDentifier pattern")
+        parser.add_argument("--parent", action="store_true", help="List the parents of the DID")
+        # Both non-functional, but list_parents complains if not present
+        # Planned to re-implement in a future release
+        parser.add_argument("--pfn", dest="pfns", nargs="+", help=argparse.SUPPRESS)
+        parser.add_argument("--guid", dest="guids", nargs="+", help=argparse.SUPPRESS)
```
rdimaio: Could be a good idea to create an issue for this and link it in the comment, for future reference

rdimaio: might also be worth beginning the comment with a `TODO:`

tests/test_cli_client_structure.py
```diff
@@ -302,22 +302,20 @@ def test_did_content(root_account, rucio_client):
...
     assert exitcode == 0
     assert "ERROR" not in err
 
-    cmd = f"rucio did show --parent --did {scope}:{dataset}"
+    cmd = f"rucio did list --parent --did {scope}:{dataset}"
     exitcode, out, err = execute(cmd)
     assert exitcode == 0
-    assert "ERROR" not in err
-    assert container not in out
+    assert container not in out  # Only checks for output not err, upstream error with mistaking claim scopes dne
```
rdimaio, line 318: Could you clarify regarding 'upstream error with mistaking claim scopes dne'?

rdimaio:
end-to-end integ test failure is unrelated, should go away if you rebase

lib/rucio/client/commands/did.py
voetberg: iirc there's no open issue for it, but I'll check and/or open one and link it here :+1:

tests/test_cli_client_structure.py
voetberg: Sorry, yeah that comment is pretty brief. 

There is an error happening in ?? somewhere claiming that the scope `mock#` doesn't exist (most likely because it's filtering out mock scopes? I need to verify but decided it was outside the scope of this pr) so there is an error in the log, but the output is correct.

rdimaio (approved):

PR #7506, opened 2025-03-17
Core: Improve error message when PFN does not match expected PFN. No …

lobis:
No functional changes.

Improve error message by returning the expected replicas, this way the user might be able to fix their request.

I had this change added as a patch to debug my script that adds replicas and found it very useful, so I think this can be added upstream. It clutters a bit the body but I think it's worth it.

### DIFF
--- a/lib/rucio/core/replica.py
+++ b/lib/rucio/core/replica.py
@@ -1829,18 +1829,24 @@ def add_replicas(
             else:
                 # Check that the pfns match to the expected pfns
                 lfns = [{'scope': i['scope'].external, 'name': i['name']} for i in files if i['pfn'].startswith(scheme)]
-                pfns[scheme] = clean_pfns(pfns[scheme])
+                pfns[scheme] = set(clean_pfns(pfns[scheme]))
+                expected_pfns = set()
 
                 for protocol_attr in rsemgr.get_protocols_ordered(rse_settings=rse_settings, operation='write', scheme=scheme, domain='wan'):
-                    pfns[scheme] = list(set(pfns[scheme]) - set(_expected_pfns(lfns, rse_settings, scheme, operation='write', domain='wan', protocol_attr=protocol_attr)))
+                    expected_pfns.update(_expected_pfns(lfns, rse_settings, scheme, operation='write', domain='wan',
+                                                        protocol_attr=protocol_attr))
+                pfns[scheme] -= expected_pfns
 
                 if len(pfns[scheme]) > 0:
                     for protocol_attr in rsemgr.get_protocols_ordered(rse_settings=rse_settings, operation='write', scheme=scheme, domain='lan'):
-                        pfns[scheme] = list(set(pfns[scheme]) - set(_expected_pfns(lfns, rse_settings, scheme, operation='write', domain='lan', protocol_attr=protocol_attr)))
+                        expected_pfns.update(_expected_pfns(lfns, rse_settings, scheme, operation='write', domain='lan',
+                                                            protocol_attr=protocol_attr))
+                    pfns[scheme] -= expected_pfns
 
                 if len(pfns[scheme]) > 0:
                     # pfns not found in wan or lan
-                    raise exception.InvalidPath('One of the PFNs provided does not match the Rucio expected PFN : %s (%s)' % (str(pfns[scheme]), str(lfns)))
+                    pfns_scheme = pfns[scheme]
+                    raise exception.InvalidPath(f"One of the PFNs provided {pfns_scheme!r} for {lfns!r} does not match the Rucio expected PFNs: {expected_pfns!r}")
 
     nbfiles, bytes_ = __bulk_add_replicas(rse_id=rse_id, files=files, account=account, session=session)
     increase(rse_id=rse_id, files=nbfiles, bytes_=bytes_, session=session)


lib/rucio/core/replica.py
```diff
@@ -1830,17 +1830,22 @@ def _expected_pfns(lfns, rse_settings, scheme, operation='write', domain='wan',
...
                     for protocol_attr in rsemgr.get_protocols_ordered(rse_settings=rse_settings, operation='write', scheme=scheme, domain='lan'):
-                        pfns[scheme] = list(set(pfns[scheme]) - set(_expected_pfns(lfns, rse_settings, scheme, operation='write', domain='lan', protocol_attr=protocol_attr)))
+                        expected_pfn = set(_expected_pfns(lfns, rse_settings, scheme, operation='write', domain='lan', protocol_attr=protocol_attr))
+                        expected_pfns = expected_pfns.union(expected_pfn)
+                        pfns[scheme] = list(set(pfns[scheme]) - expected_pfn)
 
                 if len(pfns[scheme]) > 0:
                     # pfns not found in wan or lan
-                    raise exception.InvalidPath('One of the PFNs provided does not match the Rucio expected PFN : %s (%s)' % (str(pfns[scheme]), str(lfns)))
+                    raise exception.InvalidPath(f"One of the PFNs provided {str(pfns[scheme])} for {str(lfns)} does not match the Rucio expected PFNs: {str(expected_pfns)}")
```
rdimaio: when using f-strings in exception messages, it's best to assign the message to a variable first for a cleaner traceback, see: https://docs.astral.sh/ruff/rules/f-string-in-exception/

lib/rucio/core/replica.py
```diff
@@ -1830,17 +1830,22 @@ def _expected_pfns(lfns, rse_settings, scheme, operation='write', domain='wan',
                 # Check that the pfns match to the expected pfns
                 lfns = [{'scope': i['scope'].external, 'name': i['name']} for i in files if i['pfn'].startswith(scheme)]
                 pfns[scheme] = clean_pfns(pfns[scheme])
+                expected_pfns = set()
 
                 for protocol_attr in rsemgr.get_protocols_ordered(rse_settings=rse_settings, operation='write', scheme=scheme, domain='wan'):
-                    pfns[scheme] = list(set(pfns[scheme]) - set(_expected_pfns(lfns, rse_settings, scheme, operation='write', domain='wan', protocol_attr=protocol_attr)))
+                    expected_pfn = set(_expected_pfns(lfns, rse_settings, scheme, operation='write', domain='wan', protocol_attr=protocol_attr))
+                    expected_pfns = expected_pfns.union(expected_pfn)
```
rdimaio: Imo it would be cleaner to do

```suggestion
                    expected_pfns.update(_expected_pfns(lfns, rse_settings, scheme, operation='write', domain='wan', protocol_attr=protocol_attr)
```

Just to skip having to do the `list`->`set` conversion first, and then the `union`. `set.update()` likely does this in the background anyway, but given that we're iterating in a loop, `update` feels more appropriate

lib/rucio/core/replica.py
```diff
@@ -1830,17 +1830,22 @@ def _expected_pfns(lfns, rse_settings, scheme, operation='write', domain='wan',
...
+                    pfns[scheme] = list(set(pfns[scheme]) - expected_pfn)
```
rdimaio: Couldn't this be placed after the for loop? i.e. instead of doing `set(pfns[scheme]) - expected_pfn` for every `expected_pfn`, we just do it all at once after the loop by doing `set(pfns[scheme]) - expected_pfns`

lib/rucio/core/replica.py
```diff
@@ -1830,17 +1830,22 @@ def _expected_pfns(lfns, rse_settings, scheme, operation='write', domain='wan',
...
 
                 if len(pfns[scheme]) > 0:
                     for protocol_attr in rsemgr.get_protocols_ordered(rse_settings=rse_settings, operation='write', scheme=scheme, domain='lan'):
-                        pfns[scheme] = list(set(pfns[scheme]) - set(_expected_pfns(lfns, rse_settings, scheme, operation='write', domain='lan', protocol_attr=protocol_attr)))
+                        expected_pfn = set(_expected_pfns(lfns, rse_settings, scheme, operation='write', domain='lan', protocol_attr=protocol_attr))
+                        expected_pfns = expected_pfns.union(expected_pfn)
+                        pfns[scheme] = list(set(pfns[scheme]) - expected_pfn)
```
rdimaio: same comments here about possibly using `set.update` and doing the `pfns[scheme] = list(set(pfns[scheme]) - expected_pfn)` after the `for` loop with all the `expected_pfns`

rdimaio (commented):
end-to-end integ test failure not related - it should go away if you rebase

lib/rucio/core/replica.py
lobis: Adressed, thanks!

lobis:
I rewrote the change to use a set instead of a list (there was no reason to use a list in the first place I think) so now its clearner IMHO, which should address the comments too.

lib/rucio/core/replica.py
```diff
@@ -1829,18 +1829,24 @@ def _expected_pfns(lfns, rse_settings, scheme, operation='write', domain='wan',
...
-                        pfns[scheme] = list(set(pfns[scheme]) - set(_expected_pfns(lfns, rse_settings, scheme, operation='write', domain='lan', protocol_attr=protocol_attr)))
+                        expected_pfns.update(_expected_pfns(lfns, rse_settings, scheme, operation='write', domain='lan',
+                                                            protocol_attr=protocol_attr))
+                    pfns[scheme] -= expected_pfns
 
                 if len(pfns[scheme]) > 0:
                     # pfns not found in wan or lan
-                    raise exception.InvalidPath('One of the PFNs provided does not match the Rucio expected PFN : %s (%s)' % (str(pfns[scheme]), str(lfns)))
+                    pfns_scheme = pfns[scheme]
+                    raise exception.InvalidPath(f"One of the PFNs provided {pfns_scheme!r} for {lfns!r} does not match the Rucio expected PFNs: {expected_pfns!r}")
```
rdimaio, line 1849: https://docs.astral.sh/ruff/rules/f-string-in-exception/ is still present here actually - to remediate it, the actual exception message should be passed to a variable, like this:

```
msg = f"One of the PFNs provided {pfns_scheme!r} for {lfns!r} does not match the Rucio expected PFNs: {expected_pfns!r}
```

but I think we can ignore this. If we ever want to enforce this, we can do it properly via ruff and address it then as part of a separate PR.

rdimaio (approved):

PR #7517, opened 2025-03-19
Richer error message when listing dids

lobis:
Improve error message when listing dids.

### DIFF
--- a/lib/rucio/client/commands/bin_legacy/rucio.py
+++ b/lib/rucio/client/commands/bin_legacy/rucio.py
@@ -463,8 +463,10 @@ def list_dids(args, client, logger, console, spinner):
         scope = args.did[0]
         name = '*'
 
-    if scope not in client.list_scopes():
-        logger.error('Scope not found.')
+    scopes = set(client.list_scopes())
+    if scope not in scopes:
+        msg = f"Scope '{scope}' not found when listing dids. Available scopes: {scopes}"
+        logger.error(msg)
         return FAILURE
 
     if args.recursive and '*' in name:
--- a/lib/rucio/client/commands/bin_legacy/rucio.py
+++ b/lib/rucio/client/commands/bin_legacy/rucio.py
@@ -26,7 +26,7 @@ import uuid
 from copy import deepcopy
 from datetime import datetime
 from logging import DEBUG
-from typing import Optional
+from typing import TYPE_CHECKING, Optional
 
 from rich.console import Console
 from rich.padding import Padding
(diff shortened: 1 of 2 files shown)


lib/rucio/client/commands/bin_legacy/rucio.py
```diff
@@ -464,8 +462,10 @@ def list_dids(args, client, logger, console, spinner):
         scope = args.did[0]
         name = '*'
 
-    if scope not in client.list_scopes():
-        logger.error('Scope not found.')
+    scopes = set(client.list_scopes())
+    if scope not in scopes:
+        msg = f"Scope '{scope}' not found when listing dids. Available scopes: {scopes}"
```
voetberg: I'm not sure listing all the scopes is always a good idea. We don't enforce how many scopes a single account has access to, that can make this error message quite long and unwieldy. Instead, we should invite the user to just use `scope list` on their own.

lib/rucio/client/commands/bin_legacy/rucio.py
```diff
@@ -73,15 +73,14 @@
 cli_config = get_cli_config()
 
 
-def get_scope(did, client):
+def get_scope(did: str, client) -> tuple[str, str]:
```
voetberg: `get_scope` is also in rucio-admin, this function while you're here can you also annotate that with the same type? 

Additionally, the `client` argument is always `Client`, and should be typed as such.

voetberg (changes requested):

lib/rucio/client/commands/bin_legacy/rucio.py
lobis: Added annotations!

lib/rucio/client/commands/bin_legacy/rucio.py
lobis: I also had my doubts about this, I agree that redirecting the user to list the scopes is a better idea.

voetberg:
Sorry, I think I misled you. Adding the annotation of the `get_scope` option is a good idea, but doing it for everything in `bin_legacy` muddies the waters on your PR - please revert those and open a different PR if you want to add annotations there. 

Additionally, I recommend reverting your commits when a change is requested, it keeps the commit log cleaner. In this case I would be okay just squashing them.

rdimaio (commented):
The type annotation errors are due to the addition of the `client: "Client"` hints. e.g. 

```
 Found 6 new problems in /home/runner/work/rucio/rucio/lib/rucio/client/commands/bin_legacy/rucio.py
  - 1 errors with message """"items" is not a known attribute of "None"""".
    Candidates: line 125
```

is due to:
1. adding the `Client` type annotation
2. `info = client.whoami()` is called
3. `info.items()` is called

`whoami` returns `Optional[dict[str, Any]]`, so the analyzer complains about the case where this return would be `None`.
https://github.com/rucio/rucio/blob/f2465f35f95dd95f62d6e5a4175ddbe0613ed9af/lib/rucio/client/accountclient.py#L149-L155

lobis:
Closing this as I see it was already addressed (possibly in https://github.com/rucio/rucio/pull/7344/files).

PR #7518, opened 2025-03-19
Feature: Emitting Hermes message on DID Archive.

### LINKED ISSUE #7150: Emit a message to Hermes when file DIDs are archived
### Description

Please modify Rucio to emit a message when file DIDs are archived in the `contents_history` table, effectively "deleting" them from Rucio's knowledge.

### Motivation

In order to keep external systems such as metadata catalogs in sync, it is necessary for Rucio to emit a message to Hermes that can be detected and acted upon.

### Change

Currently when the last replica for a file DID is deleted `__cleanup_after_replica_deletion()` will delete any `DataIdentifierAssociation` and then move those file DIDs from the `dids` table to `contents_history`. It would be simple to emit a message to the Hermes daemon at this point.

We can discuss if this should be `ERASE` with an extended payload describing the `did_type`, or if it should be a separate `ERASE_FILE` message in order to maintain compatibility with existing systems.

alexanderrichards:
fixes #7150

### DIFF
--- a/tests/test_did.py
+++ b/tests/test_did.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
 
 import pytest
 
@@ -20,6 +20,7 @@ from rucio.common import exception
 from rucio.common.exception import DataIdentifierAlreadyExists, DataIdentifierNotFound, DuplicateContent, FileAlreadyExists, FileConsistencyMismatch, InvalidPath, ScopeNotFound, UnsupportedOperation, UnsupportedStatus
 from rucio.common.types import InternalScope
 from rucio.common.utils import generate_uuid
+from rucio.core import did as did_core
 from rucio.core.did import (
     add_did,
     add_did_to_followed,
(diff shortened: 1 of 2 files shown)


rdimaio (commented):
Can we add a test to check this?

tests/test_did.py
```diff
@@ -69,6 +71,16 @@ def test_delete_dids(self, mock_scope, root_account):
...
         delete_dids(dids=dsns, account=root_account)
 
+    def test_list_content_history(self, mock_scope, monkeypatch):
+        """ DATA IDENTIFIERS (CORE): List Content History """
+        def mock_add_messages(*args, **kwargs):
+            raise Exception("Caught!")
+        tmp_dsn1 = did_name_generator('dataset')
+        with monkeypatch.context() as m:
+            m.setattr(rdid, "add_messages", mock_add_messages)
+            for d in list_content_history(scope=mock_scope, name=tmp_dsn1):
```
voetberg: Does this actually test the change? From the way this is written, I would expect this test to just raise an exception if it was working correctly.

tests/test_did.py
```diff
@@ -16,6 +16,7 @@
 
 import pytest
 
+import rucio.core.did as rdid
```
voetberg: Common convention is to import as "core_{component}" or "{component}_core" for legibility

bari12 (changes requested):
We really need more tests on this. (as @rdimaio suggested)
e.g. the payload is wrong, the `ERASE_FILE` should report the file (child_scope, child_name) - right now this just reports the dataset which had a file which was deleted, but you have no idea which file it is.

PR #7521, opened 2025-03-19
Client: Rewrite default scope extraction method

### LINKED ISSUE #7516: Unable to list dids with trailing forward slash `/` via CLI
We have DIDs with the following pattern: trailing `/` for datasets, no trailing `/` for containers.

```
❯ rucio did list --did 'ams:*' --filter "type=CONTAINER"
┌───────────────────────────────────────────────────┬────────────┐
│ SCOPE:NAME                                        │ [DID TYPE] │
├───────────────────────────────────────────────────┼────────────┤
│ ams:/MC                                           │ CONTAINER  │
│ ams:/MC/AMS02                                     │ CONTAINER  │
│ ams:/MC/AMS02/2011B                               │ CONTAINER  │
│ ams:/MC/AMS02/2011B/he.B620dev                    │ CONTAINER  │
│ ams:/MC/AMS02/2011B/he.B620dev/he.pl1.512.qmd     │ CONTAINER  │
│ ams:/MC/AMS02/2014                                │ CONTAINER  │
│ ams:/MC/AMS02/2014/AntiD.B1128                    │ CONTAINER  │
│ ams:/MC/AMS02/2014/AntiD.B1128/antid.pl1ph.021000 │ CONTAINER  │
└───────────────────────────────────────────────────┴────────────┘
❯ rucio did list --did 'ams:*' --filter "type=DATASET"
┌────────────────────────────────────────────────────┬────────────┐
│ SCOPE:NAME                                         │ [DID TYPE] │
├────────────────────────────────────────────────────┼────────────┤
│ ams:/MC/                                           │ DATASET    │
│ ams:/MC/AMS02/                                     │ DATASET    │
│ ams:/MC/AMS02/2011B/                               │ DATASET    │
│ ams:/MC/AMS02/2011B/he.B620dev/                    │ DATASET    │
│ ams:/MC/AMS02/2011B/he.B620dev/he.pl1.512.qmd/     │ DATASET    │
│ ams:/MC/AMS02/2014/                                │ DATASET    │
│ ams:/MC/AMS02/2014/AntiD.B1128/                    │ DATASET    │
│ ams:/MC/AMS02/2014/AntiD.B1128/antid.pl1ph.021000/ │ DATASET    │
└────────────────────────────────────────────────────┴────────────┘
```

These dids were added using the sdk (using `name` and `scope` fields explicitly, so no magic happens to split `scope:name`).

The issue is that when using the CLI, the trailing `/` is removed, resulting in datasets being impossible to query.

I am not sure what the root cause of this issue is, but I am pretty sure it happens on the client side, as the request that reaches the server already does not have the trailing `/`. I'm not sure if this is related to the policy package as this is a client-side issue.

```
❯ rucio did list --did 'ams:/MC/' --filter "type=ALL"
┌────────────┬────────────┐
│ SCOPE:NAME │ [DID TYPE] │
├────────────┼────────────┤
│ ams:/MC    │ CONTAINER  │
└────────────┴────────────┘
❯ rucio did list --did 'ams:/MC' --filter "type=ALL"
┌────────────┬────────────┐
│ SCOPE:NAME │ [DID TYPE] │
├────────────┼────────────┤
│ ams:/MC    │ CONTAINER  │
└────────────┴────────────┘
```

### LINKED ISSUE #7520: Review default scope extraction method
Current method has some ATLAS specific logic and backwards compatibility stuff (which I believe is really old from conversations with @bari12 ).

I think it can be simplified.

lobis:
This should only affect the client side and should provide a more straightforward scope extraction, without any unexpected effects (e.g. trimming `/`).

I added tests for it.

Closes #7516

### DIFF
--- a/lib/rucio/common/utils.py
+++ b/lib/rucio/common/utils.py
@@ -595,26 +595,33 @@ class ScopeExtractionAlgorithms(PolicyPackageAlgorithms):
     @staticmethod
     def extract_scope_default(did: str, scopes: Optional['Sequence[str]']) -> 'Sequence[str]':
         """
-        Default fallback scope extraction algorithm, based on the ATLAS scope extraction algorithm.
+        Default scope extraction algorithm. Extracts the scope from the DID.
 
         :param did: The DID to extract the scope from.
+        :param scopes: Not used in the default algorithm.
 
-        :returns: A tuple containing the extracted scope and the DID.
+        :returns: A tuple containing the extracted scope and the name.
         """
-        if did.find(':') > -1:
-            if len(did.split(':')) > 2:
-                raise RucioException('Too many colons. Cannot extract scope and name')
-            scope, name = did.split(':')[0], did.split(':')[1]
-            if name.endswith('/'):
-                name = name[:-1]
-            return scope, name
-        else:
+
+        # This block is ATLAS specific, to be removed in the future.
+        # More info at https://github.com/rucio/rucio/pull/7521
+        if did.find(':') == -1:
             scope = did.split('.')[0]
             if did.startswith('user') or did.startswith('group'):
                 scope = ".".join(did.split('.')[0:2])
             if did.endswith('/'):
                 did = did[:-1]
             return scope, did
+
+        parts = did.split(':')
+        if len(parts) != 2:
+            msg = f"Cannot extract scope and name from DID {did}. The DID should have exactly one colon but found {len(parts)} colons."
+            raise RucioException(msg)
+        scope, name = parts
+        if not scope or not name:
+            msg = f"Cannot extract scope and name from DID {did}. Found empty scope or name."
+            raise RucioException(msg)
+        return scope, name
 
     @staticmethod
     def extract_scope_dirac(did: str, scopes: Optional['Sequence[str]']) -> 'Sequence[str]':
--- a/tests/rucio/common/test_utils.py
+++ b/tests/rucio/common/test_utils.py
@@ -11,9 +11,10 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
 import pytest
 
-from rucio.common.utils import _encode_params_as_url_query_string, build_url, invert_dict
+from rucio.common.utils import ScopeExtractionAlgorithms, _encode_params_as_url_query_string, build_url, invert_dict
 
 
 class TestUtils:
@@ -49,3 +50,19 @@ class TestUtils:
     )
     def test_encode_params_as_url_query_string(self, params, doseq, expected_query_string):
         assert _encode_params_as_url_query_string(params, doseq) == expected_query_string
+
+    @pytest.mark.parametrize(
+        'did, scope, name',
+        [
+            ("scope:name", "scope", "name"),
+            ("scope:/this/is/a/path", "scope", "/this/is/a/path"),
+            ("scope:/this/is/a/path/", "scope", "/this/is/a/path/"),
+            ("scope:no/slash/at/start", "scope", "no/slash/at/start"),
+            ("scope:no/slash/at/start/", "scope", "no/slash/at/start/"),
+            ("scope:/", "scope", "/"),
+            ("scope://", "scope", "//"),
+            ("scope:/path/with//duplicated///slash", "scope", "/path/with//duplicated///slash"),
+        ]
+    )
+    def test_default_scope_extraction_algorithm(self, did, scope, name):
+        assert ScopeExtractionAlgorithms.extract_scope_default(did=did, scopes=None) == (scope, name)


voetberg:
Hi, the tests should be added before the PR goes through, so we know it works before merging

lobis:
> Hi, the tests should be added before the PR goes through, so we know it works before merging

I created this PR mostly to have material for discussion in the developer round table. All current tests work, that is the most important.

I will add more tests for the new default algorithm, but since I don't if this change will be accepted (as it may have unintended consequences I am unaware of), I will wait for the go ahead before adding them, that is what I meant to say.

rdimaio:
> > Hi, the tests should be added before the PR goes through, so we know it works before merging
> 
> I created this PR mostly to have material for discussion in the developer round table. All current tests work, that is the most important.
> 
> I will add more tests for the new default algorithm, but since I don't if this change will be accepted (as it may have unintended consequences I am unaware of), I will wait for the go ahead before adding them, that is what I meant to say.

You can also mark the PR as a draft PR to indicate it is not ready for review yet and you want to discuss it in the meeting first

voetberg (dismissed):
👍

lib/rucio/common/utils.py
```diff
@@ -595,26 +595,23 @@ def register(cls: type[ScopeExtractionAlgorithmsT], name: str, fn_extract_scope:
...
-            if name.endswith('/'):
-                name = name[:-1]
-            return scope, name
-        else:
-            scope = did.split('.')[0]
-            if did.startswith('user') or did.startswith('group'):
-                scope = ".".join(did.split('.')[0:2])
-            if did.endswith('/'):
-                did = did[:-1]
-            return scope, did
```
rdimaio: Leaving a comment here to block merging of this until we check with @cserf that it is safe to remove

bari12: @mlassnig as well. Is this scope "extraction" still something relevant to anyone in ATLAS, now 10+ years after DQ2?

mlassnig: I'm pretty sure that a majority of people in ATLAS just use the name and not the full DID when interacting with Rucio. This removal would break this?

lib/rucio/common/utils.py
```diff
@@ -595,26 +595,23 @@ def register(cls: type[ScopeExtractionAlgorithmsT], name: str, fn_extract_scope:
...
+        :param scopes: Not used in the default algorithm.
 
-        :returns: A tuple containing the extracted scope and the DID.
+        :returns: A tuple containing the extracted scope and the name.
         """
-        if did.find(':') > -1:
-            if len(did.split(':')) > 2:
-                raise RucioException('Too many colons. Cannot extract scope and name')
-            scope, name = did.split(':')[0], did.split(':')[1]
-            if name.endswith('/'):
```
mlassnig, line 608: Removal of the final slash for old-style DQ2 container names is probably fine. Waiting on confirmation from Misha and Rod.

lobis: Just to clarify, the change makes so that we **don't** remove the slash, so if they are relying on this automatic removal, they would need to remove it themselves.

lib/rucio/common/utils.py
lobis: I was not aware of this, is it allowed to provide a did such as "name" with no scope? How is the scope inferred automatically?

mishaborodin: both panda and prodsys are using python API which forces to use scope. For many user scripts which are using  CLI it's not the case, as Mario wrote I'm pretty sure they are using a simple name without scope.

lobis: The new default scope extraction method will fail if no scope is provided explicitly via `:`.

Looking at the old implementation (https://github.com/rucio/rucio/blob/master/lib/rucio/common/utils.py#L596-L617) , I think this should also be the case. In the old implementation the scope needs to be present either via `:` or via `.` with the user/group logic.

If they are using just names, I think they must be using a different scope extraction than the default (which can be set in the policy package).

mishaborodin: The thing with the scope in ATLAS - it's very long and redundant. We agree that for ATLAS scope is always first two parts of the name for user and group and first part for the rest scopes. Having the scope in any place like dashboards or spreadsheets just impossible, string becomes too long. And then users can't just copy-paste the name to do something in CLI. My opinion scope "guessing" for ATLAS dataset should stay.

lobis: I think there is a misunderstanding here. This PR only changes the default scope extraction method, I think ATLAS uses a policy package so this change should have no effect on ATLAS. Perhaps @rdimaio can comment on this?

rdimaio: > I think there is a misunderstanding here. This PR only changes the default scope extraction method, I think ATLAS uses a policy package so this change should have no effect on ATLAS. Perhaps @rdimaio can comment on this?

Correct - in ATLAS we use the same algorithm and that will not be changed, so even if we merge this, ATLAS functionality would remain the same: https://gitlab.cern.ch/atlas-adc-ddm/rucio-policy-package/-/blob/master/src/atlas_rucio_policy_package/scope.py?ref_type=heads

the default algorithm was originally a duplicate of the ATLAS algorithm, I placed it there in order to be able to take out the ATLAS algorithm from the main Rucio code

bari12: This is not correct. We do not (and should not) force policy package on the client side. This PR would remove the extraction on the client side.

To be honest, I am not super happy with having this "vo-only" code upstream. But if people in ATLAS still rely on it andf the only alternative is to force policy package on the client side (which we should avoid) then it is better to leave it in.

rdimaio: > This is not correct. We do not (and should not) force policy package on the client side. This PR would remove the extraction on the client side.
> 
> To be honest, I am not super happy with having this "vo-only" code upstream. But if people in ATLAS still rely on it andf the only alternative is to force policy package on the client side (which we should avoid) then it is better to leave it in.

Why is scope extraction handled in the client and not the server?

bari12: It's just trickier, because most REST endpoints will expect a scope to be given to them.
Now all these could possibly be changed to allow a `None` scope and then do the extraction, it's just not as trivial than doing it on the client side.

rdimaio: > Now all these could possibly be changed to allow a None scope and then do the extraction, it's just not as trivial than doing it on the client side.

I believe it makes more sense to do this (or to restructure the API to not accept a scope as a parameter at all, just the full DID, and always do the extraction server side). Otherwise customising the scope extraction algorithm in the policy packages is useless and would potentially cause confusion, since custom scope extraction algorithms are not used if the client is not using a policy package (can it even use a policy package?)

lobis: I updated this PR to preserve the atlas functionality whenever a `:` is not found in the did (just like before).

rdimaio (approved):

lib/rucio/common/utils.py
bari12: I am merging this now, within the 37 release line, hoping that the `/` removal does not break anything. If it does, we have to re-introduce it.
This is a bit of a grey zone, since technically this might be a breaking change which should only go in a major release. But I think other deployments will benefit from this a lot, so it's worth taking that risk.

PR #7522, opened 2025-03-19
Core: Standardize account limits methods

### LINKED ISSUE #7483: Standardize account limits methods for clarity
We have these functions to query account limits:
- `get_local_account_limit`: Returns the account limit for the account on the rse.
- `get_local_account_limits`: Returns the account limits for the account on the list of rses.
- `get_global_account_limit`: Returns the global account limit for the account on the rse expression.
- `get_global_account_limits`: Returns the global account limits for the account.

I think that we should rename these functions and standardize their usages. 
I would avoid this usage of both singular and plural, as it's very confusing. We could even refactor `get_local_account_limits` into `get_local_account_limit`, and only have a single method that takes either a single RSE expression or a list of RSEs.

0xquark:
Fixes #7483 

#### **For Local Limits**  
The two methods, `get_local_account_limit` and `get_local_account_limits`, have been merged into one: `get_local_account_limit`, used in both core and gateway levels. Now, the `RSE` parameter works like this:  

- **Single RSE** (`str`) → Returns **one limit**.  
- **Multiple RSEs** (`list[str]`) → Returns **multiple limits**.  
- **No RSE (None)** → Returns **all limits**.  

Test cases have been added or updated to check if the function works correctly for single, multiple, and all RSEs.

#### **For Global Limits**
The two methods, `get_global_account_limit` and `get_global_account_limits`, have been merged into one:
`get_global_account_Limit`, used in both core and gateway levels.

- **Single RSE Expression** (`str`) with account -> returns the global limit for that specific expression in that account
- If **no RSE expression(`none`)** is provided, it returns all global account limits for the specified account (or for all accounts if no account is given).

### DIFF
--- a/lib/rucio/core/account_limit.py
+++ b/lib/rucio/core/account_limit.py
@@ -156,44 +156,31 @@ def get_global_account_limit(account: "InternalAccount", rse_expression: str, *,
 
 
 @read_session
-def get_local_account_limit(account: "InternalAccount", rse_id: str, *, session: "Session") -> Union[int, float, None]:
+def get_local_account_limit(account: "InternalAccount", rse_ids: Union[str, list[str], None] = None, *, session: "Session") -> Union[int, float, dict[str, int], None]:
     """
-    Returns the account limit for the account on the rse.
+    Returns the local account limit for a given RSE or list of RSEs.
 
     :param account:  Account to check the limit for.
-    :param rse_id:   RSE id to check the limit for.
+    :param rse_ids:  Single RSE id or a list of RSE ids to check the limit for.
     :param session:  Database session in use.
-    :return:         Limit in Bytes.
+    :return:         Limit in Bytes (int/float) for a single RSE or
+                     Dictionary {'rse_id': bytes, ...} for multiple RSEs.
     """
-    try:
-        stmt = select(
-            models.AccountLimit
-        ).where(
-            and_(models.AccountLimit.account == account,
-                 models.AccountLimit.rse_id == rse_id)
-        )
-        account_limit = session.execute(stmt).scalar_one()
-        if account_limit.bytes == -1:
-            return float("inf")
-        else:
-            return account_limit.bytes
-    except NoResultFound:
-        return None
+    if isinstance(rse_ids, str):  # Single RSE case
+        try:
+            stmt = select(models.AccountLimit).where(
+                and_(models.AccountLimit.account == account, models.AccountLimit.rse_id == rse_ids)
+            )
+            account_limit = session.execute(stmt).scalar_one()
+            return float("inf") if account_limit.bytes == -1 else account_limit.bytes
+        except NoResultFound:
+            return None
 
-
-@read_session
-def get_local_account_limits(account: "InternalAccount", rse_ids: Optional[list[str]] = None, *, session: "Session") -> dict[str, int]:
-    """
-    Returns the account limits for the account on the list of rses.
-
-    :param account:  Account to check the limit for.
-    :param rse_ids:  List of RSE ids to check the limit for.
-    :param session:  Database session in use.
-    :return:         Dictionary {'rse_id': bytes, ...}.
-    """
-
+    # Multiple RSE case or no RSE specified
     account_limits = {}
-    if rse_ids:
+
+    # If rse_ids is a list of RSEs
+    if isinstance(rse_ids, list) and rse_ids:
         rse_id_clauses = []
         for rse_id in rse_ids:
             rse_id_clauses.append(and_(models.AccountLimit.rse_id == rse_id,
@@ -339,14 +326,14 @@ def get_local_account_usage(account: "InternalAccount", rse_id: Optional[str] = 
     )
     if not rse_id:
         # All RSESs
-        limits = get_local_account_limits(account=account, session=session)
+        limits = get_local_account_limit(account=account, rse_ids=None, session=session)
         counters = {c.rse_id: c for c in session.execute(stmt).scalars().all()}
     else:
         # One RSE
         stmt.where(
             models.AccountUsage.rse_id == rse_id
         )
-        limits = get_local_account_limits(account=account, rse_ids=[rse_id], session=session)
+        limits = get_local_account_limit(account=account, rse_ids=[rse_id], session=session)
         counters = {c.rse_id: c for c in session.execute(stmt).scalars().all()}
     result_list = []
 
--- a/lib/rucio/core/account_limit.py
+++ b/lib/rucio/core/account_limit.py
@@ -18,7 +18,8 @@ from sqlalchemy.exc import NoResultFound
 from sqlalchemy.sql import func, literal, select
 from sqlalchemy.sql.expression import and_, or_
 
-from rucio.core.account import get_all_rse_usages_per_account
+from rucio.common.exception import AccountNotFound
+from rucio.core.account import account_exists, get_all_rse_usages_per_account
 from rucio.core.rse import get_rse_name
 from rucio.core.rse_expression_parser import parse_expression
 from rucio.db.sqla import models
@@ -101,6 +102,8 @@ def get_global_account_limit(account: Optional["InternalAccount"] = None, rse_ex
     :param session:         Database session in use.
     :return:                Limit in Bytes for a single RSE expression, or a dictionary of all limits {'MOCK': {'resolved_rses': ['MOCK'], 'limit': 10, 'resolved_rse_ids': [123]}}.
     """
+    if account and not account_exists(account, session=session):
+        raise AccountNotFound(f"Account {account} does not exist")
     if rse_expression:
         # Fetch limit for a single RSE expression
         try:
@@ -145,6 +148,8 @@ def get_local_account_limit(account: "InternalAccount", rse_ids: Union[str, list
     :return:         Limit in Bytes (int/float) for a single RSE or
                      Dictionary {'rse_id': bytes, ...} for multiple RSEs.
     """
+    if not account_exists(account, session=session):
+        raise AccountNotFound(f"Account {account} does not exist")
     if isinstance(rse_ids, str):  # Single RSE case
         try:
             stmt = select(models.AccountLimit).where(
--- a/lib/rucio/core/account_limit.py
+++ b/lib/rucio/core/account_limit.py
@@ -26,6 +26,8 @@ from rucio.db.sqla import models
 from rucio.db.sqla.session import read_session, transactional_session
 
 if TYPE_CHECKING:
+    from collections.abc import Iterable
+
     from sqlalchemy.orm import Session
 
     from rucio.common.types import InternalAccount, RSEAccountUsageDict, RSEGlobalAccountUsageDict, RSELocalAccountUsageDict, RSEResolvedGlobalAccountLimitDict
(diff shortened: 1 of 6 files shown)


lib/rucio/gateway/account_limit.py
```diff
@@ -47,110 +47,74 @@ def get_rse_account_usage(
...
-    account: str,
-    rse: str,
-    vo: str = 'def',
-    *,
-    session: "Session"
+        account: str,
+        rse: Optional[str] = None,
+        vo: str = 'def',
+        *,
+        session: "Session"
```
rdimaio, line 56: nit: previous indentation was fine, but you can ignore this as we should instead force this indentation with a formatter

lib/rucio/gateway/account_limit.py
```diff
@@ -47,110 +47,74 @@ def get_rse_account_usage(
...
-    rse_expression: str,
-    vo: str = 'def',
-    *,
-    session: "Session"
-) -> dict[str, dict[str, RSEResolvedGlobalAccountLimitDict]]:
+        account: str,
+        rse_expression: Optional[str] = None,
+        vo: str = 'def',
+        *,
+        session: "Session"
```
rdimaio, line 91: nit: previous indentation was fine, but you can ignore this as we should instead force this indentation with a formatter

lib/rucio/gateway/account_limit.py
```diff
@@ -47,110 +47,74 @@ def get_rse_account_usage(
...
-    :returns: The account limits.
-    """
-    if account:
-        internal_account = InternalAccount(account, vo=vo)
+    if rse:
+        # Single RSE lookup
+        rse_id = get_rse_id(rse=rse, vo=vo, session=session)
+        return {rse: account_limit_core.get_local_account_limit(account=internal_account, rse_ids=rse_id, session=session)}
     else:
-        internal_account = InternalAccount('*', vo=vo)
```
rdimaio, line 125: This functionality is lost in the new version. I'm not sure how important it is, so maybe someone with more insight here can chime in. I would keep it just to be safe.

lib/rucio/gateway/account_limit.py
```diff
@@ -47,110 +47,74 @@ def get_rse_account_usage(
...
-    Lists the limitation names/values for the specified account name and rse expression.
+    Lists the global account limitation names/values for the specified account.
+    If an RSE expression is provided, fetches the limit for that expression.
+    Otherwise, fetches all global limits for the account.
 
     REST API: http://<host>:<port>/rucio/account/<account>/limits
 
     :param account:         The account name.
-    :param rse_expression:  The rse expression.
+    :param rse_expression:  The RSE expression (optional for fetching all limits).
```
rdimaio: "(optional for fetching all limits)" to me sounds like the RSE expression is used to fetch all limits, but it seems to be the other way around - I think this could be clearer

lib/rucio/core/account_limit.py
```diff
@@ -130,70 +138,33 @@ def get_global_account_limits(account: Optional["InternalAccount"] = None, *, se
...
+            return float("inf") if account_limit.bytes == -1 else account_limit.bytes
+        except NoResultFound:
+            return None
 
+    # Multiple RSE case or no RSE specified
     account_limits = {}
-    if rse_ids:
+
+    # If rse_ids is a list of RSEs
+    if isinstance(rse_ids, list) and rse_ids:
```
rdimaio: We already checked `isinstance(rse_ids, str)` before, so this `isinstance(rse_ids, list)` is not necessary - we just need to check that `rse_ids` is not `None`.

lib/rucio/core/account_limit.py
```diff
@@ -130,70 +138,33 @@ def get_global_account_limits(account: Optional["InternalAccount"] = None, *, se
...
-            return float("inf")
-        else:
-            return account_limit.bytes
-    except NoResultFound:
-        return None
-
-
-@read_session
-def get_local_account_limits(account: "InternalAccount", rse_ids: Optional[list[str]] = None, *, session: "Session") -> dict[str, int]:
+def get_local_account_limit(account: "InternalAccount", rse_ids: Union[str, list[str], None] = None, *, session: "Session") -> Union[int, float, dict[str, int], None]:
```
rdimaio:
```suggestion
def get_local_account_limit(account: "InternalAccount", rse_ids: Optional[str, 'Iterable[str]'] = None, *, session: "Session") -> Optional[int, float, dict[str, int]]:
```

1. Prefer `Optional` instead of `Union[..., None]`
2. `rse_ids` is used as an `Iterable` here, so it does not need to be a list - please update the docstring accordingly as well. In general, when type hinting parameters, it's best to be as generic as possible (i.e. `Iterable` instead of `list`), while it's best to be more specific when type hinting return types

lib/rucio/core/account_limit.py
```diff
@@ -339,14 +310,14 @@ def get_local_account_usage(account: "InternalAccount", rse_id: Optional[str] =
     )
     if not rse_id:
         # All RSESs
-        limits = get_local_account_limits(account=account, session=session)
+        limits = get_local_account_limit(account=account, rse_ids=None, session=session)
```
rdimaio: `rse_ids` is `None` by default, so this is not necessary

lib/rucio/core/account_limit.py
```diff
@@ -91,36 +92,43 @@ def get_rse_account_usage(rse_id: str, *, session: "Session") -> list["RSEAccoun
 
 
 @read_session
-def get_global_account_limits(account: Optional["InternalAccount"] = None, *, session: "Session") -> dict[str, "RSEResolvedGlobalAccountLimitDict"]:
+def get_global_account_limit(account: Optional["InternalAccount"] = None, rse_expression: Optional[str] = None, *,
+                             session: "Session") -> Union[int, float, dict[str, "RSEResolvedGlobalAccountLimitDict"], None]:
```
rdimaio:
```suggestion
                             session: "Session") -> Optional[int, float, dict[str, "RSEResolvedGlobalAccountLimitDict"]]:
```

Prefer `Optional` over `Union[..., None]`

lib/rucio/core/account_limit.py
```diff
@@ -91,36 +92,43 @@ def get_rse_account_usage(rse_id: str, *, session: "Session") -> list["RSEAccoun
...
     """
-    Returns the global account limits for the account.
+    Returns the global account limit for the given account and RSE expression, or all limits if no specific expression is provided.
 
-    :param account:  Account to check the limit for.
-    :param session:  Database session in use.
-    :return:         Dict {'MOCK': {'resolved_rses': ['MOCK'], 'limit': 10, 'resolved_rse_ids': [123]}}.
+    :param account:         Account to check the limit for (optional for fetching all accounts).
```
rdimaio: > (optional for fetching all accounts)

Is this true? If `account` is `None`, when we do `models.AccountGlobalLimit.account == account`, I think it would return 0 results since `account` in `AccountGlobalLimit` is `Mapped[InternalAccount]`, which cannot be `None`.

rdimaio: I assume this is only true for the ex `get_global_account_limits` path, whereas for the ex `get_global_account_limit` path we need the `account` to always be specified. In that case, it would be better if the docstring was more explicit about this, and we should add an `if account` check together with `if rse_expression` when going through the ex `get_global_account_limit` path.

lib/rucio/core/account_limit.py
```diff
@@ -91,36 +92,43 @@ def get_rse_account_usage(rse_id: str, *, session: "Session") -> list["RSEAccoun
...
+        # Fetch limit for a single RSE expression
+        try:
+            stmt = select(models.AccountGlobalLimit).where(
+                and_(models.AccountGlobalLimit.account == account,
+                     models.AccountGlobalLimit.rse_expression == rse_expression)
+            )
+            global_account_limit = session.execute(stmt).scalar_one()
+            return float("inf") if global_account_limit.bytes == -1 else global_account_limit.bytes
+        except NoResultFound:
+            return None
```
rdimaio:
```suggestion
    if account:
        if not account_exists(account, session=session):
            raise AccountNotFound(f"Account {account} does not exist")
        if rse_expression:
            # Fetch limit for a single RSE expression
            try:
                stmt = select(models.AccountGlobalLimit).where(
                    and_(models.AccountGlobalLimit.account == account,
                        models.AccountGlobalLimit.rse_expression == rse_expression)
                )
                global_account_limit = session.execute(stmt).scalar_one()
                return float("inf") if global_account_limit.bytes == -1 else global_account_limit.bytes
            except NoResultFound:
                return None
```

tests/test_account_limits.py
```diff
@@ -75,23 +75,42 @@ def test_get_global_account_usage(self, account, rse_factory, db_session):
...
         _, rse1_id = rse_factory.make_mock_rse()
         _, rse2_id = rse_factory.make_mock_rse()
+        _, rse3_id = rse_factory.make_mock_rse()
         limit1 = 100
         limit2 = 200
         account_limit.set_local_account_limit(account=account, rse_id=rse1_id, bytes_=limit1, session=db_session)
         account_limit.set_local_account_limit(account=account, rse_id=rse2_id, bytes_=limit2, session=db_session)
-        results = account_limit.get_local_account_limits(account=account, session=db_session)
+        account_limit.set_local_account_limit(account=account, rse_id=rse3_id, bytes_=limit2, session=db_session)
+        results = account_limit.get_local_account_limit(account=account, rse_ids=[rse1_id, rse2_id],session=db_session)
```
rdimaio: I don't see a need for a third mock RSE here, we already have two RSEs, so we can test the 'multiple RSE' path - to do this kind of testing you could run
1. get_local_account_limits with just rse1_id, see if it returns 1 result
2. get_local_account_limits with both rse1_id and rse2_id, see if it returns both results

lib/rucio/core/account_limit.py
0xquark: Thanks! I’ll keep that in mind.  

Just out of curiosity, is using `Optional` instead of `Union[..., None]` just a convention in the Rucio codebase, or does it have any benefits in terms of performance or readability? For me, they're both same and `Union[..., None]` offers backward compatibility in most old codebases. We might also switch to https://docs.python.org/3/library/typing.html#typing.Optional `str | None`.

rdimaio: It's a convention in the Rucio codebase - it doesn't change anything performance-wise, and I guess it's subjective whether it's more or less readable. 

> Union[..., None] offers backward compatibility in most old codebases.

I mean, that's not really something we would need to worry about, we're on Python 3.9+.

> We might also switch to https://docs.python.org/3/library/typing.html#typing.Optional str | None.

That was added in Python 3.10, and we need to maintain compatibility with Python 3.9 for now.

lib/rucio/gateway/account_limit.py
0xquark: It didn't made sense to me since `account` is not an optional param.

rdimaio: It matters in the context of the [`if/else` it's found in](
https://github.com/rucio/rucio/blob/34a85aa511aaca69d59fa422b3f6a201375855db/lib/rucio/gateway/account_limit.py#L122C1-L125C55):

```
    if account:
        internal_account = InternalAccount(account, vo=vo)
    else:
        internal_account = InternalAccount('*', vo=vo)
```

`account` could be `''`, in which case it would evaluate to `False`.

```
>>> account = ''
>>> bool(account)
False
```

In that case, it would be set as `'*'`.

0xquark: Should an empty string `' '` be considered a valid value for account? If not, maybe we should raise an error instead of assigning '*'.
We can also make the condition clearer by explicitly checking for None:
```
if account is not None:  
    internal_account = InternalAccount(account, vo=vo)```

rdimaio: > Should an empty string ' ' be considered a valid value for account?

I assume it shouldn't, but the SQLA model actually doesn't seem to check against that. I agree with you that it would be clearer if we checked `is not None` instead, and then somewhere in the call (I guess in core) we have the logic that actually finds all account limits if a single `account` is not specified.

I think it's a good idea if someone that frequently uses this `account_limit` feature could chime in on what the use case is for having this kind of logic. Maybe you could ask in either the support or developers channel on Mattermost

0xquark:
@rdimaio I get it now—we can't use `Optional` in the code above because it only supports one type plus `None` (like [Optional[X]](https://docs.python.org/3/library/typing.html). But in this case, we’re using multiple types (`Union[int, float, dict[str, "RSEResolvedGlobalAccountLimitDict"], None]`), so using `Union` makes sense.  

Should I revert back from `Optional` to `Union`? in my previous changes. The tests are failing due to this.

rdimaio:
> @rdimaio I get it now—we can't use `Optional` in the code above because it only supports one type plus `None` (like [Optional[X]](https://docs.python.org/3/library/typing.html). But in this case, we’re using multiple types (`Union[int, float, dict[str, "RSEResolvedGlobalAccountLimitDict"], None]`), so using `Union` makes sense.
> 
> Should I revert back from `Optional` to `Union`? in my previous changes. The tests are failing due to this.

The standard in Rucio is to use `Optional` to mean `Union[..., None]`, the solution here is to use `Union` as part of `Optional`, e.g. `Optional[Union[str, list[str]]]`. I had made a mistake in the initial code suggestions as I forgot to include the `Union`, apologies about that!

rdimaio (approved):
LGTM. https://github.com/rucio/rucio/pull/7522#discussion_r2010078061 this question is still unsolved (and I see no one replied on Mattermost), but I think we can ignore it at this point

PR #7529, opened 2025-03-20
Core: Identity LFN2PFN clean duplicated `/` in output path

### LINKED ISSUE #7528: Clean duplicated `/` in identity LFN2PFN
Some DID start with `/` and this causes a duplicated slash in the pfn which may cause issues.

lobis:
Closes #7528 

Before this PR, if a did name starts with a `/` (e.g. `scope:/path/`), the resulting pfn will have a double `/` in the path part. 

When manually registering a replica, you would also need to provide this double `/` in the path, otherwise pfn would not match.

This change trims all double `/` into single `/`, so all pfn using the identity lfn2pfn are expected to have single `/` in the path part of the pfn.

I acknowledge that this is a breaking change and I am not fully aware of all unintended consequences.

### DIFF
--- a/lib/rucio/rse/translation.py
+++ b/lib/rucio/rse/translation.py
@@ -217,9 +217,13 @@ class RSEDeterministicTranslation(PolicyPackageAlgorithms):
         del rse
         del rse_attrs
         del protocol_attrs
+
         if scope.startswith('user') or scope.startswith('group'):
             scope = scope.replace('.', '/')
-        return '%s/%s' % (scope, name)
+        path = f'{scope}/{name}'
+        while '//' in path:
+            path = path.replace('//', '/')
+        return path
 
     @classmethod
     def _module_init_(cls) -> None:


rdimaio (commented):
The CI changes seem unrelated, looks like a bug with `setuptools` not being found, if you re-run them I think it will work
```
ERROR: Can not execute `setup.py` since setuptools is not available in the build environment.
```

lobis:
> The CI changes seem unrelated, looks like a bug with `setuptools` not being found, if you re-run them I think it will work
> 
> ```
> ERROR: Can not execute `setup.py` since setuptools is not available in the build environment.
> ```

Thanks for letting me know, I think there is an issue with the container image, restarting does not help atm.

rdimaio (commented):
I feel like this is something that should be caught as part of `validate_schema`

rdimaio:
> > The CI changes seem unrelated, looks like a bug with `setuptools` not being found, if you re-run them I think it will work
> > ```
> > ERROR: Can not execute `setup.py` since setuptools is not available in the build environment.
> > ```
> 
> Thanks for letting me know, I think there is an issue with the container image, restarting does not help atm.

You're right - @Geogouz identified the issue and published a PR to fix it here https://github.com/rucio/rucio/pull/7525

Geogouz:
Rebasing should fix the problem

rdimaio (approved):

lib/rucio/rse/translation.py
```diff
@@ -217,9 +217,13 @@ def __identity(
         del rse
         del rse_attrs
         del protocol_attrs
+
         if scope.startswith('user') or scope.startswith('group'):
             scope = scope.replace('.', '/')
-        return '%s/%s' % (scope, name)
+        path = f'{scope}/{name}'
+        while '//' in path:
```
mlassnig, line 224: Double slashes unfortunately have semantic meaning for root protocol URLs (i.e.,   root://hostname//path/file != root://hostname/path/file).

We can't generally dismiss them because of that.

lobis: After making this PR I realized that this could become problematic. I also agree that this might break stuff unfortunately.

I think the other changes make this PR a bit redundant. It was necessary at first but now it just makes sure the PFN does not have duplicated slashes, if we want to allow this, we can close the PR

PR #7531, opened 2025-03-20
Policy / Core: Update default DID name expression to support `/`

### LINKED ISSUE #7530: Better out of the box support for path-like DIDs
Rucio with default settings does not support path-like dids, which are a common use case. This can currently be supported via policy packages.

A few changes to the defaults are needed to support path-like dids, which in principle could be done without breaking anything.

- https://github.com/rucio/rucio/pull/7531 - Allows `/` in DID names.
- https://github.com/rucio/rucio/pull/7521 - Allows having `/` in DID, otherwise some trailing `/` is stripped which makes using CLI difficult or impossible for some names.
- https://github.com/rucio/rucio/pull/7529 - Change to the identity LFN2PFN to prevent double `//` in PFN path, which happens when DID name starts with `/`. DID names can still have multiple `/` if user chooses to do it, but the PFN path will always have a single one.
- https://github.com/rucio/rucio/pull/7550
- https://github.com/rucio/rucio/pull/7552
- https://github.com/rucio/rucio/issues/7691

Also related:
- https://github.com/rucio/rucio/issues/7519
- https://github.com/rucio/containers/issues/387
- https://github.com/rucio/containers/pull/394
- https://github.com/rucio/rucio/issues/7724

We ended up closing https://github.com/rucio/rucio/pull/7529 due to potential issues, its not needed after all to make slashes work out of the box.

### LINKED ISSUE #7691: Support slashes `/` by default in DID names

lobis:
Update the regex to support `/` in name, including just the `/` as a did name.

This should only extend the set of valid did names, so it should (hopefully) have no effect on existing installations.

Since this is the default, it won't affect those users using a policy package.

Related to #7530 

Closes https://github.com/rucio/rucio/issues/7691

Also closes https://github.com/rucio/rucio/issues/7530

### DIFF
--- a/tests/test_policy_package.py
+++ b/tests/test_policy_package.py
@@ -16,9 +16,51 @@ import importlib
 
 import pytest
 
+import rucio.common.exception
 import rucio.common.schema
 import rucio.core.permission
+from rucio.common.constants import DEFAULT_VO
 from rucio.common.types import InternalAccount
+
+
+class TestPolicyPackageGeneric:
+
+    @classmethod
+    def setup_class(cls):
+        # TODO: remove this after the generic schema becomes the default one used in tests (https://github.com/rucio/rucio/issues/7819)
+        cls._rucio_default_vo_module = rucio.common.schema.schema_modules[DEFAULT_VO]
+        rucio.common.schema.schema_modules[DEFAULT_VO] = importlib.import_module('rucio.common.schema.generic')
+
+    @classmethod
+    def teardown_class(cls):
+        # TODO: remove this after the generic schema becomes the default one used in tests (https://github.com/rucio/rucio/issues/7819)
+        rucio.common.schema.schema_modules[DEFAULT_VO] = cls._rucio_default_vo_module
+
+    @pytest.mark.parametrize("name", [
+        'name_with_underscore',
+        'name-with-dash',
+        'name.with.dot',
+        'name/with/slash',
+        '/name/starting/with/slash',
+        'name/ending/with/slash/',
+    ])
+    def test_default_schema_did_name_valid(self, name, file_config_mock):
+        rucio.common.schema.validate_schema('name', name, vo=DEFAULT_VO)
+
+    @pytest.mark.parametrize("name", [
+        '.startingWithDot',
+        '-startingWithDash',
+        '_startingWithUnderscore',
+        '#hash',
+        'name with spaces',
+        'name*with*asterisk',
+        'name\\with\\backslash',
+        'name:with:colon',
+        'name,with,comma',
+    ])
+    def test_default_schema_did_name_invalid(self, name, file_config_mock):
+        with pytest.raises(rucio.common.exception.InvalidObject):
+            rucio.common.schema.validate_schema('name', name, vo=DEFAULT_VO)
 
 
 class TestPolicyPackage:
--- a/lib/rucio/common/schema/generic.py
+++ b/lib/rucio/common/schema/generic.py
@@ -56,7 +56,7 @@ NAME_LENGTH = 250
 NAME = {"description": "Data Identifier name",
         "type": "string",
         "maxLength": NAME_LENGTH,
-        "pattern": "^[A-Za-z0-9][A-Za-z0-9\\.\\-\\_]*$"}
+        "pattern": r"^[/A-Za-z0-9][/A-Za-z0-9\.\-_]*$"}
 
 R_NAME = {"description": "Data Identifier name",
           "type": "string",
--- a/lib/rucio/common/schema/generic_multi_vo.py
+++ b/lib/rucio/common/schema/generic_multi_vo.py
@@ -57,7 +57,7 @@ NAME_LENGTH = 250
 NAME = {"description": "Data Identifier name",
         "type": "string",
         "maxLength": NAME_LENGTH,
-        "pattern": "^[A-Za-z0-9][A-Za-z0-9\\.\\-\\_]*$"}
+        "pattern": r"^[/A-Za-z0-9][/A-Za-z0-9\.\-_]*$"}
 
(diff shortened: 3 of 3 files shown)


voetberg:
I know this PR is a rather small change, but it should probably have a few examples in test suites showing it works. I would recommend parameterizing something in https://github.com/rucio/rucio/blob/90643ffb6bfb57f9a55fce546f56ad1de8eec5b4/tests/test_policy_package.py to show a few cases of allowed and not allowed did names. It's a good sanity check and something to point to if a similar issue arises in the future.

lobis:
> Policy / Core: Add tests for did name default schema validation

I agree these tests are useful. I have implemented them in https://github.com/rucio/rucio/pull/7531/commits/2b3bd2b2e1620a4482a4018997a01a2529635723

I think there might be some funny business with the policy package tests as this is failing and printing the old policy package in the logs, I'll take a look

tests/test_policy_package.py
```diff
@@ -12,55 +12,38 @@
...
-
-        # check that schemas defined in our module override generic
-        rucio.common.schema.validate_schema('account', 'this-account-name-is-too-long-for-the-generic-schema-but-should-validate-against-the-mock-one')
-
-        # check that schemas not defined in our module fall back to generic
-        rucio.common.schema.validate_schema('r_name', 'name_to_validate')
-
-        # restore original schema module
-        rucio.common.schema.schema_modules['def'] = old_module
+    @pytest.mark.parametrize("name", [
```
voetberg, line 39: These can be merged into one test with a flag as a second parameter if they're valid names or not but this is fine too

tests/test_policy_package.py
```diff
@@ -12,55 +12,38 @@
...
+        'name_with_underscore',
+        'name-with-dash',
+        'name.with.dot',
+        'name/with/slash',
+        '/name/starting/with/slash',
+        'name/ending/with/slash/',
+    ])
+    def test_default_schema_did_name_valid(self, name: str):
```
voetberg: Tests don't need to have type annotations (no need to change this but to save you future work) - https://rucio.cern.ch/documentation/developer/type_annotation_guide#what-not-to-type-annotate

tests/test_policy_package.py
lobis: That was my original layout but I find this structure more clear :) It should take around the same amount of lines too

lobis:
Tests are failing because the CI is not running with the latest code. The test pass on a local run. After this PR is merged to master, the docker images should update and the new test should pass.

rdimaio:
> Tests are failing because the CI is not running with the latest code. The test pass on a local run. After this PR is merged to master, the docker images should update and the new test should pass.

Could you clarify? The caching has been reverted, so the CI should be running with the latest code

lobis:
> > Tests are failing because the CI is not running with the latest code. The test pass on a local run. After this PR is merged to master, the docker images should update and the new test should pass.
> 
> Could you clarify? The caching has been reverted, so the CI should be running with the latest code

I am not entirely sure what is going on. I just know that the tests are being run with the generic policy package from before this change took place, so it must be using old code somehow. I did some debugging and I'm pretty confident it's the case. I don't know exactly why. We should probably address this in a separate PR. My guess is there is a docker container somewhere which is not being built with this code and instead relying on some already existing tag (master?). Perhaps this is an easy fix if you know where to look...

rdimaio:
> > > Tests are failing because the CI is not running with the latest code. The test pass on a local run. After this PR is merged to master, the docker images should update and the new test should pass.
> > 
> > 
> > Could you clarify? The caching has been reverted, so the CI should be running with the latest code
> 
> I am not entirely sure what is going on. I just know that the tests are being run with the generic policy package from before this change took place, so it must be using old code somehow. I did some debugging and I'm pretty confident it's the case. I don't know exactly why. We should probably address this in a separate PR. My guess is there is a docker container somewhere which is not being built with this code and instead relying on some already existing tag (master?). Perhaps this is an easy fix if you know where to look...

We use the ATLAS policy package by default in tests - I believe the issue is that we need to make this change to the ATLAS policy package too (I see that the Belle II one passes fine, so we can leave that one alone). Of course this depends on whether this `/` should be supported in ATLAS - @bari12 what do you think?

lobis:
Thanks for the help @rdimaio !

Now the new tests run explicitly against the `generic` and `generic_multi_vo` schemas, so everything is passing except for a few VO...

lobis:
> I know this PR is a rather small change, but it should probably have a few examples in test suites showing it works. I would recommend parameterizing something in https://github.com/rucio/rucio/blob/90643ffb6bfb57f9a55fce546f56ad1de8eec5b4/tests/test_policy_package.py to show a few cases of allowed and not allowed did names. It's a good sanity check and something to point to if a similar issue arises in the future.

Added tests. Tooks more effort than initially expected, an issue was created (https://github.com/rucio/rucio/issues/7819) to improve this in the future.

tests/test_policy_package.py
```diff
@@ -16,11 +16,53 @@
...
 import rucio.core.permission
+from rucio.common.constants import DEFAULT_VO
 from rucio.common.types import InternalAccount
 
 
+class TestPolicyPackageGeneric:
+
+    @classmethod
+    def setup_class(cls):
+        # TODO: remove this after the generic schema becomes the default one (https://github.com/rucio/rucio/issues/7819)
```
rdimaio: nit: I'd specify "becomes the default one used in tests"

rdimaio (dismissed):

tests/test_policy_package.py
lobis: Addressed, unfortunately the review became stale

rdimaio (approved):

voetberg (approved):

PR #7533, opened 2025-03-20
tests: skip x509 auth cert tests outside GitHub actions

### LINKED ISSUE #7507: `test-curl` tests fails in dev envoriments
### Description

About half of the test_curl tests fail when running in a dev environment - because the dev environment does not have a x509 cert the way these tests expect. 

```
FAILED tests/test_curl.py::TestCurlRucio::test_get_auth_x509 - AssertionError: assert 'X-Rucio-Auth-Token' in ''
FAILED tests/test_curl.py::TestCurlRucio::test_get_auth_validate - AssertionError: assert 'X-Rucio-Auth-Token' in ''
FAILED tests/test_curl.py::TestCurlRucio::test_post_account - AssertionError: assert 'X-Rucio-Auth-Token' in ''
FAILED tests/test_curl.py::TestCurlRucio::test_get_accounts_whoami - AssertionError: assert 'X-Rucio-Auth-Token' in ''
FAILED tests/test_curl.py::TestCurlRucio::test_post_rse - AssertionError: assert 'X-Rucio-Auth-Token' in ''
```

### Steps to reproduce

- install the dev environment using [the instructions here](https://github.com/rucio/rucio/tree/master/etc/docker/dev)
- Run either `bash tools/run_tests.sh` or `pytest tests/test_curl.py` - observe the failed tests

### Rucio Version

_No response_

### Additional Information

I propose we mark these tests with a `skip_if` to only run them in the github runner or a specific environment.

0xquark:
- Marked x509-failed-tests with `pytest.mark.skipif` to skip tests outside CI (github actions) env for local runs.
- Github Actions has a [env variable](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables) `GITHUB_ACTIONS` which is always set to true while running as part of the workflow.

Fixes #7507 

Screenshot 2025-03-20 at 14 55 01

### DIFF
--- a/lib/rucio/tests/common.py
+++ b/lib/rucio/tests/common.py
@@ -45,6 +45,8 @@ skip_multivo = pytest.mark.skipif('SUITE' in os.environ and os.environ['SUITE'] 
                                   reason="does not work for multiVO")
 skip_non_belleii = pytest.mark.skipif(not ('POLICY' in os.environ and os.environ['POLICY'] == 'belleii'),
                                       reason="specific belleii tests")
+skip_outside_gh_actions = pytest.mark.skipif(os.getenv("GITHUB_ACTIONS") != "true",
+                                             reason="Skipping tests outside GitHub Actions")
 
 
 def is_influxdb_available() -> bool:
--- a/tests/test_curl.py
+++ b/tests/test_curl.py
@@ -18,7 +18,7 @@ import os
 import pytest
 
 from rucio.common.config import config_get, config_get_bool
-from rucio.tests.common import account_name_generator, execute, get_long_vo, rse_name_generator
+from rucio.tests.common import account_name_generator, execute, get_long_vo, rse_name_generator, skip_outside_gh_actions
 
 
 class TestCurlRucio:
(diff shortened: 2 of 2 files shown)


lib/rucio/tests/common.py
```diff
@@ -45,6 +45,8 @@
                                   reason="does not work for multiVO")
 skip_non_belleii = pytest.mark.skipif(not ('POLICY' in os.environ and os.environ['POLICY'] == 'belleii'),
                                       reason="specific belleii tests")
+skip_x509_tests = pytest.mark.skipif(os.getenv("GITHUB_ACTIONS") != "true",
```
voetberg: I am fairly certain this can be reused in other situation - can this instead be named something like "skip_outside_gh_actions"? The reason should also reflect this.

voetberg (changes requested):

lib/rucio/tests/common.py
0xquark: Thanks! Nice Suggestion

voetberg (approved):

rdimaio (approved):

PR #7540, opened 2025-03-24
Core: Method to register files with structure; Fixes rucio#7534

### LINKED ISSUE #7534: Implement registration of file / did with hierarchical structure
Implement a generic file / did registration mechanism similar to what we have in https://gitlab.cern.ch/rucio-it/rucio-tools.

lobis:
Partially addresses #7534

### DIFF
--- a/lib/rucio/core/did.py
+++ b/lib/rucio/core/did.py
@@ -3064,7 +3064,7 @@ def add_files_with_attachments(
     """
 
     # Input validation
-    required_file_keys = {"scope", "name", "rse", "pfn", "bytes", "adler32"}
+    required_file_keys = {"scope", "name", "pfn", "bytes", "adler32"}
     for file in files:
         for key in required_file_keys:
             if key not in file:
--- a/lib/rucio/core/did.py
+++ b/lib/rucio/core/did.py
@@ -3033,10 +3033,12 @@ def insert_deleted_dids(filter_: "ColumnExpressionArgument[bool]", *, session: "
         ).save(session=session, flush=False)
 
 @transactional_session
-def add_files_with_structure(
+def add_files_with_attachments(
         files: "Iterable[dict[str, Any]]",
-        account: "InternalAccount",
+        dids_attachments: "Sequence[dict[str, Any]]",
+        rse_id: str,
         *,
+        account: "InternalAccount",
         session: "Session"
 ):
     """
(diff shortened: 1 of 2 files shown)


lib/rucio/core/did.py
```diff
@@ -3002,3 +3002,53 @@ def insert_deleted_dids(filter_: "ColumnExpressionArgument[bool]", *, session: "
...
             constituent=did.constituent
         ).save(session=session, flush=False)
+
+@transactional_session
+def add_files_with_structure(
+        files: "Iterable[dict[str, Any]]",
+        account: "InternalAccount",
+        *,
+        session: "Session"
+):
```
rdimaio:
```suggestion
) -> None:
```

lib/rucio/core/did.py
```diff
@@ -3002,3 +3002,53 @@ def insert_deleted_dids(filter_: "ColumnExpressionArgument[bool]", *, session: "
...
+        if "scope" not in file:
+            raise exception.InputValidationError("Missing 'scope' in file")
+        if "name" not in file:
+            raise exception.InputValidationError("Missing 'name' in file")
+        scope = file["scope"]
+        name = file["name"]
+        parts = name.split("/")
+        directories = [
+            "/".join(parts[:i + 1]) for i in range(len(parts) - 1)
+        ]
```
rdimaio: It would be good to have a comment to explain and help visualize what this does, e.g.

```suggestion
        # get parent paths (e.g. 'a/b/c/d' -> ['a', 'a/b', 'a/b/c'])
        parts = name.split("/")
        directories = [
            "/".join(parts[:i + 1]) for i in range(len(parts) - 1)
        ]
```

lib/rucio/core/did.py
```diff
@@ -3002,3 +3002,53 @@ def insert_deleted_dids(filter_: "ColumnExpressionArgument[bool]", *, session: "
             is_archive=did.is_archive,
             constituent=did.constituent
         ).save(session=session, flush=False)
+
+@transactional_session
+def add_files_with_structure(
```
rdimaio: I think the method name could be a bit clearer - initially I thought that `structure` meant "file structure", maybe referring to the pfn, bytes, adler32, but after reading the method I understood it means that it adds a container DID for each parent path. Maybe something like `add_files_and_container_dids`

rdimaio (commented):
I get the idea and overall I think it makes sense, but I wonder if this kind of feature will only be needed by one community or if multiple communities will actually be interested in organising DIDs like this. I think it makes sense to do this kind of hierarchical structuring, but I just wonder why this was not brought up in the past by existing communities (or was it?)

0xquark:
> I get the idea and overall I think it makes sense, but I wonder if this kind of feature will only be needed by one community or if multiple communities will actually be interested in organising DIDs like this. I think it makes sense to do this kind of hierarchical structuring, but I just wonder why this was not brought up in the past by existing communities (or was it?)

Why not release this as a custom package? Anyone interested could install this specific version/package of Rucio. The only challenge I see is keeping it updated with new features and bug fixes.

rdimaio:
I assume this is the linked issue https://github.com/rucio/rucio/issues/7534 - if so, it's worth mentioning it in the PR or commit description so there's more context for reviewing

lobis:
> I get the idea and overall I think it makes sense, but I wonder if this kind of feature will only be needed by one community or if multiple communities will actually be interested in organising DIDs like this. I think it makes sense to do this kind of hierarchical structuring, but I just wonder why this was not brought up in the past by existing communities (or was it?)

This hierarchichal registration is implemented in the dirac client, but I think this should be available as part of the rucio core in a more general way.

(btw thanks for the review but the PR is still in draft, I created the PR to have a way to track discussions regarding the implementation).

I think many communities are interested in this: for instance this comment just popped: https://github.com/rucio/containers/issues/387#issuecomment-2747547837

lobis:
> I assume this is the linked issue #7534 - if so, it's worth mentioning it in the PR or commit description so there's more context for reviewing

I expect this PR to take a long time, so I will probably end up rewriting history to comply with contributors guide before merging, so please ignore non-compliant commits for now. (until PR is ready to be reviewed).

lobis:
> I assume this is the linked issue #7534 - if so, it's worth mentioning it in the PR or commit description so there's more context for reviewing

I expect this PR to take a long time, so I will probably end up rewriting history to comply with contributors guide before merging, so please ignore non-compliant commits for now. (until PR is ready to be reviewed).

maxnoe:
> This hierarchichal registration is implemented in the dirac client, but I think this should be available as part of the rucio core in a more general way.

I agree, but that for me means that we should not add a separate client here, but refactor the dirac client to be less specific to dirac, maybe introducing a couple of configuration options.

lobis:
I'll close this stale PR. The issue remains open.

PR #7545, opened 2025-03-24
Update `setup.cfg` to address installation error

lobis:
I removed the key `force-manifest` as I think it's not actually used anywhere, but I'm not 100% sure.

Closes #7544 

Is there any reason why we can't just remove this file? I feel that `pyproject.toml` should be enough.

PR #7549, opened 2025-03-24
fix: InvalidConfigError when using setuptools >= 58

lgray:
This fixes an error when building the sdist using the setuptools released this morning.

```
setuptools.errors.InvalidConfigError: Invalid dash-separated key 'description-file' in 'metadata' (setup.cfg), please use the underscore name 'description_file' instead.
```


lgray:
Oh this is a duplicate of #7545, closing.

Issue #7519, opened 2025-03-19
Do not use regex to split did, use scope extraction method


lobis:
Currently there are multiple places where a did is split into scope and name, which can seem confusing.

Policy packages support a regex that is used to extract scope and name from a given DID (it does not really act on the did, but a modification of it, prepending some `/` and replacing `:` by `/`, which makes it even more confusing to use).

There are already existing methods that split a did into scope and name. I believe these should be used as they can also be overriden in the policy package. Compared to a regex a python method will always provide more functionality and be easier to understand, can add loggin at intermediate stages, etc.

IMHO the regex in the policy package should be used only for validation of the syntax.

panta-123:
https://github.com/rucio/rucio/blob/c1ed21347cd716180771660835f37490f980c960/lib/rucio/web/rest/flaskapi/v1/common.py#L203
This is the place where the use of scope_name_regex is done.
I think just replacng scope_name_regex  in that with extract_scope method should be okay....... (on surfacial look)

jamesp-epcc:
I have had a look at this. I don't think simply replacing with `extract_scope` will work. `parse_scope_name` is used in the specific case where the scope and name are in a URL path used to access the REST API. In this case they are always separated by `/` and the whole path is usually built with code like this:

```
path = '/'.join([BASEURL, quote_plus(scope), quote_plus(name)])
```

So there will always be a forward slash in between them, which `SCOPE_NAME_REGEXP` will deal with. In contrast, the default scope extraction method used by `extract_scope` expects a colon in between, and if this is overridden with experiment-specific code it might be a different format altogether. As I understand it, these two functions are doing two different things: one is extracting the scope and DID out of a string that might be in an experiment-specific format, the other is extracting them from a URL path string in a simple known format, built by the Rucio client code.

lobis:
> I have had a look at this. I don't think simply replacing with `extract_scope` will work. `parse_scope_name` is used in the specific case where the scope and name are in a URL path used to access the REST API. In this case they are always separated by `/` and the whole path is usually built with code like this:
> 
> ```
> path = '/'.join([BASEURL, quote_plus(scope), quote_plus(name)])
> ```
> 
[...]

PR #7552, opened 2025-03-25
API: Add step in `parse_scope_name` that checks for encoded slashes

### LINKED ISSUE #7551: Support for encoded slashes in URL
Related to https://github.com/rucio/containers/issues/392

After some discussion regarding how slashes should be handled in paths, a comment (https://github.com/rucio/containers/issues/387#issuecomment-2747547837) suggested using the 'AllowEncodedSlashes NoDecode` option in Apache.

I think this is the most general approach to the problem as we want to support slashes in urls. With just allowing slashes in url (`AllowEncodedSlashes On`), they are mixed with proper path separators and sometimes they are stripped, causing many issues which have been partially addressed in other PRs.

With this setting, slashes are encoded and thus preserved. They need to be decoded at some point in order to be converted into `scope` and `name`.

In order to ensure backwards compatibility I propose having an optional parsing step that checks for encoded slashes. This would make the regex (https://github.com/rucio/rucio/issues/7519) obsolete.

lobis:
Closes #7551 

Related to #7519 

This change should be backwards compatibly as it only takes place whenever encoded slashes are sent. This should only happen if the `AllowEncodedSlashes NoDecode` is enabled in Apache.

With this change the regex to split scope and name is no longer needed, simplifying the logic.

### DIFF
--- a/lib/rucio/web/rest/flaskapi/v1/common.py
+++ b/lib/rucio/web/rest/flaskapi/v1/common.py
@@ -15,11 +15,13 @@
 import itertools
 import json
 import logging
+import os
 import re
 from configparser import NoOptionError, NoSectionError
 from functools import wraps
 from time import time
 from typing import TYPE_CHECKING, Any, Literal, Optional, TypeVar, Union
+from urllib.parse import unquote_plus
 
 import flask
 from flask.views import MethodView
@@ -45,6 +47,9 @@ if TYPE_CHECKING:
     from rucio.web.rest.flaskapi.v1.types import HeadersType
 
 ResponseTypeVar = TypeVar('ResponseTypeVar', bound=flask.wrappers.Response)
+
+RUCIO_HTTPD_ENCODED_SLASHES_NO_DECODE = os.environ.get('RUCIO_HTTPD_ENCODED_SLASHES_NO_DECODE',
+                                                       'false').lower() == 'true'
 
 
 class CORSMiddleware:
@@ -210,6 +215,18 @@ def parse_scope_name(scope_name: str, vo: Optional[str]) -> tuple[str, ...]:
     :raises ValueError: when scope_name could not be parsed.
     :returns: a (scope, name) tuple.
     """
+
+    if RUCIO_HTTPD_ENCODED_SLASHES_NO_DECODE:
+        if scope_name.count('/') != 1:
+            # scope and name are always separated by a single slash ('/', unencoded) in the request.
+            # If the server is configured with the 'NoDecode' option, other slashes will be encoded.
+            # This is just a sanity check that should never happen.
+            raise ValueError(f"Could not parse '{scope_name}' ({scope_name=}) with encoded '/' into scope and name.")
+
+        scope, name = scope_name.split('/', 1)
+        name = unquote_plus(name)
+
+        return scope, name
 
     if not vo:
         vo = 'def'


lib/rucio/web/rest/flaskapi/v1/common.py
```diff
@@ -210,6 +210,15 @@ def parse_scope_name(scope_name: str, vo: Optional[str]) -> tuple[str, ...]:
     :raises ValueError: when scope_name could not be parsed.
     :returns: a (scope, name) tuple.
     """
+
+    encoded_slash = r"%2F"
+    if encoded_slash in scope_name:
+        if scope_name.count('/') != 1:
```
rdimaio, line 220: I'm a bit confused - how can there be a `/` in the `scope_name` if this is the path where slashes in `scope_name` are encoded?

lib/rucio/web/rest/flaskapi/v1/common.py
```diff
@@ -210,6 +210,15 @@ def parse_scope_name(scope_name: str, vo: Optional[str]) -> tuple[str, ...]:
     :raises ValueError: when scope_name could not be parsed.
     :returns: a (scope, name) tuple.
     """
+
+    encoded_slash = r"%2F"
+    if encoded_slash in scope_name:
```
rdimaio: Since we're server-side, I wonder if it would be faster to check if `AllowEncodedSlashes NoDecode` is true

lib/rucio/web/rest/flaskapi/v1/common.py
```diff
@@ -210,6 +210,15 @@ def parse_scope_name(scope_name: str, vo: Optional[str]) -> tuple[str, ...]:
...
+        if scope_name.count('/') != 1:
+            raise ValueError(f"Could not parse '{scope_name}' ({scope_name=}) with encoded '/' into scope and name.")
+        scope, name = scope_name.split('/', 1)
+        name = name.replace(encoded_slash, '/')
```
rdimaio: Small edge case, but we probably want to prevent people from having `%2F` in their scope or name (which the current `SCOPE_NAME_REGEXP` allows), otherwise this would break.

lib/rucio/web/rest/flaskapi/v1/common.py
lobis: It is confusing.

The request sent looks something like `/did/scope/name%2Fwith%2Fslash%2F`.

Since we are using the `AllowEncodedSlashes NoDecode` then this will be preserved. If we were not using this, then the `scope_name` variable will be converted into: `scope/name/with/slash`. It would be ambiguous the distinction between scope and name and also the trailing slash is lost, so we cannot differentiate between did with and without trailing slash for some requests.

lib/rucio/web/rest/flaskapi/v1/common.py
lobis: It is a possibility. I prefer this appraoch which is not dependent on the apache server, but it's a matter of preference. How would you check this in a reliable manner?

lib/rucio/web/rest/flaskapi/v1/common.py
lobis: Good point, but actually the DID name regex will not allow `%2F` in the request as `%` is forbidden (atleast in the default settings). It is true that on a given policy package this could be an issue, but it's a more marginal possibility that can always be addressed by updating the did regex.

lib/rucio/web/rest/flaskapi/v1/common.py
rdimaio: Not sure - maybe if `AllowEncodedSlashes` is set via a value in the chart, then we could have it as an env variable too, and then we can check that. I think it's something that seems it would only need to be checked once - i.e. if `AllowEncodedSlashes NoDecode` is true for the server, I assume it's not going to change at any point while the server runs, so we could have it as an env variable or as some kind of flag that is only toggled once, rather than having to do a character search in a string for every `parse_scope_name` operation

lib/rucio/web/rest/flaskapi/v1/common.py
rdimaio: It might be more appropriate to use `unquote_plus`, in order to do the complete inverse of the `quote_plus` operation done in the client:

https://docs.python.org/3/library/urllib.parse.html#urllib.parse.unquote_plus

lib/rucio/web/rest/flaskapi/v1/common.py
rdimaio: Thanks, I see what you mean now. It might be worth adding a comment to clarify this just to make it a bit simpler to read through the code - e.g. something that explains that the single `/` that we expect to see is the one that divides scope and name

lib/rucio/web/rest/flaskapi/v1/common.py
lobis: Updated, let me know if you agree with this way of checking for the setting.

lib/rucio/web/rest/flaskapi/v1/common.py
lobis: Updated.

lib/rucio/web/rest/flaskapi/v1/common.py
lobis: Added a comment to explain.

rdimaio (dismissed):

lobis:
I had to rebase so previos approval was reset

rdimaio (approved):

PR #7555, opened 2025-03-25
Policies: support diff-based schemas

rdimaio:
Follow-up from https://github.com/rucio/rucio/pull/7489 to implement diff-based schemas for rc1. It's the same as that PR, with the merge conflicts fixed + [this change](https://github.com/rucio/rucio/pull/7489#discussion_r2010157833) implemented under approval from @bari12

### DIFF
--- a/lib/rucio/common/schema/__init__.py
+++ b/lib/rucio/common/schema/__init__.py
@@ -140,18 +140,18 @@ def load_schema_for_vo(vo: str) -> None:
 
 
 def validate_schema(name: str, obj: Any, vo: str = 'def') -> None:
-    if vo not in schema_modules:
-        load_schema_for_vo(vo)
-    if hasattr(schema_modules[vo], 'SCHEMAS') and name in schema_modules[vo].SCHEMAS:
-        schema = schema_modules[vo].SCHEMAS.get(name, {})
-    else:
-        # if schema not available in VO module, fall back to generic module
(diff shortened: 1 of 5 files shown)


bari12 (approved):

lib/rucio/common/schema/__init__.py
```diff
@@ -138,12 +140,18 @@ def load_schema_for_vo(vo: str) -> None:
 
 
 def validate_schema(name: str, obj: Any, vo: str = 'def') -> None:
-    if vo not in schema_modules:
-        load_schema_for_vo(vo)
-    if not hasattr(schema_modules[vo], 'validate_schema'):
-        _get_generic_schema_module().validate_schema(name, obj)
-        return
-    schema_modules[vo].validate_schema(name, obj)
+    if obj:
```
maxnoe, line 143: Is this `if obj` intentional in its depth? I.e. it doesn't validate empty dicts for required properties or empty lists that have a jsonschema that might require at least 1 item in the list or the empty string against some validation logic?

rdimaio: It's a leftover from the `validate_schema` function taken from the generic schemas, to ensure the same functionality. I guess we could remove it and validate eventual empty input as all other input and raise a `ValidationError` in case it doesn't pass validation.

PR #7556, opened 2025-03-25
Release: Build wheels #7546

### LINKED ISSUE #7546: Publish wheels
### Description

Rucio is only published as source distribution (sdist) to PyPI.

Please consider also publishing wheels, as that would prevent issues like the sdist becoming incompatible with a new version of setuptools, as happened now.

### Motivation

_No response_

### Change

_No response_

bari12:


rdimaio (dismissed):
LGTM - worth noting that `build --sdist --wheel` is different than `build`:
- `build` (default behaviour): build an SDist and then build a wheel from the SDist.
- `--sdist --wheel`: Produce and SDist and a wheel, both from the source distribution.

I think the default behaviour is fine here, as recommended in the docs https://build.pypa.io/en/stable/, but I'm not sure if it would prevent an issue like #7546 from happening - I think it would result in a build failure as the wheels would try to be built from the sdist. This is probably still fine, because at least we discover this early on before we even upload anything to PyPI.

tools/build_sdist.sh
rdimaio, line 1: Should we rename this script, as it no longers just builds the sdist?

bari12: Ah yes, should probably rename.
I think the default behaviour is fine.

rdimaio (approved):

Issue #7571, opened 2025-03-27
dogpile.cache version requirement is not enforced


dchristidis:
### Description

```console
Traceback (most recent call last):
  File "/usr/lib64/python3.9/configparser.py", line 789, in get
    value = d[option]
  File "/usr/lib64/python3.9/collections/__init__.py", line 941, in __getitem__
    return self.__missing__(key)            # support subclasses that define __missing__
  File "/usr/lib64/python3.9/collections/__init__.py", line 933, in __missing__
    raise KeyError(key)
KeyError: 'metrics_port'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/rucio/common/config.py", line 195, in config_get
    return convert_type_fnc(get_config().get(section, option))
  File "/usr/lib64/python3.9/configparser.py", line 792, in get
    raise NoOptionError(option, section)
configparser.NoOptionError: No option 'metrics_port' in section: 'monitor'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/bin/rucio-judge-evaluator", line 23, in <module>
    from rucio.daemons.judge.evaluator import run, stop
  File "/usr/local/lib/python3.9/site-packages/rucio/daemons/judge/evaluator.py", line 36, in <module>
    from rucio.core.monitor import MetricManager
  File "/usr/local/lib/python3.9/site-packages/rucio/core/monitor.py", line 94, in <module>
    METRICS_PORT = config_get_int('monitor', 'metrics_port', raise_exception=False, default=8080)
  File "/usr/local/lib/python3.9/site-packages/rucio/common/config.py", line 345, in config_get_int
    return config_get(
  File "/usr/local/lib/python3.9/site-packages/rucio/common/config.py", line 209, in config_get
    return __config_get_table(section=section, option=option, raise_exception=raise_exception,
  File "/usr/local/lib/python3.9/site-packages/rucio/common/config.py", line 663, in __config_get_table
    return core_config_get(section, option, default=default, session=session, use_cache=use_cache,
  File "/usr/local/lib/python3.9/site-packages/rucio/common/utils.py", line 1650, in _wrapper
    return fn(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/rucio/db/sqla/session.py", line 402, in new_funct
    return function(*args, session=session, **kwargs)
[...]

Issue #7330, opened 2025-01-17
Consider restructuring Rucio into a namespace package with better distinction of packages to distribute


rdimaio:
As suggested here https://github.com/rucio/rucio/issues/7322:

> Actually split the distribution packages.
Maybe using namespace packages for common, server and client specific code respectively.

## What we distribute

We currently distribute the following packages from the main `rucio/rucio` repository:
- [`rucio`](https://pypi.org/project/rucio/): Includes the entire `lib/rucio` folder, so server+clients+webUI, plus some extra stuff. Setup is done [here](https://github.com/rucio/rucio/blob/master/setup_rucio.py)
- [`rucio-clients`](https://pypi.org/project/rucio-clients/): Includes the Rucio Python and CLI clients. Setup is done [here](https://github.com/rucio/rucio/blob/master/setup_rucio_client.py)
- [`rucio-webui`](https://pypi.org/project/rucio-webui/): Includes the current webUI, which is being deprecated. Setup is done [here](https://github.com/rucio/rucio/blob/master/setup_webui.py)

## Disadvantages of current structure/distribution flow
- Since the packages are not separate, and we use `setup.py` to build a package, we need to either symlink or copy the specific `setup_*` file to `setup.py`, as seen in the distribution tool: https://github.com/rucio/rucio/blob/dfcd980f4f10720af38ea1570a6015b2478a5408/tools/build_sdist.sh#L55-L74
 
This causes issues like the one mentioned in https://github.com/rucio/rucio/issues/7322, where you can't have an editable install of both `rucio` and `rucio-clients`.
- We cannot migrate to `pyproject.toml` (https://github.com/rucio/rucio/issues/6971), because the relevant packages are not clearly separated (e.g. we don't have a `server` folder).

## Proposal

At a minimum, it would be great to break the project down into:
- rucio-client
  - CLI client
  - Python client 
- rucio-server
  - API
  - Gateway
  - Daemons
  - Transfertool
- rucio-common
  - Common code utilized by the other packages
- rucio-webui

Structuring this into a native Python namespace package (see [here](https://packaging.python.org/en/latest/guides/packaging-namespace-packages/).)

[...]

Issue #6971, opened 2024-07-25
Dependencies: deprecate `setup.py` and move to `pyproject.toml`


rdimaio:
`pyproject.toml` is the current specification standard for build system requirements for Python projects, defined in PEP-518: https://peps.python.org/pep-0518/#file-format

`setup.py` is not deprecated, but it's recommended to move to `pyproject.toml` if possible:
- https://packaging.python.org/en/latest/discussions/setup-py-deprecated/
- https://stackoverflow.com/questions/75588250/what-are-the-downsides-to-relying-purely-on-pyproject-toml
- https://packaging.python.org/en/latest/discussions/setup-py-deprecated/#is-pyproject-toml-mandatory

To do:
- [ ] Implement this first: https://github.com/rucio/containers/issues/344 (we want to more accurately replicate the package installation process in tests)
- [ ] Follow https://packaging.python.org/en/latest/guides/modernize-setup-py-project/#modernize-setup-py-project to see what can be moved to `pyproject.toml`

mgajek-cern:
Has a tradeoff analysis been done between using `setuptools` in `pyproject.toml` versus adopting tools like `Poetry`, `Hatch` or others? It would be helpful to understand the rationale, especially considering their benefits in dependency locking, environment management and modern workflows.

Official Python Packaging Authority (PyPA) guidance: https://packaging.python.org/en/latest/guides/tool-recommendations/

rdimaio:
> Has a tradeoff analysis been done between using `setuptools` in `pyproject.toml` versus adopting tools like `Poetry`, `Hatch` or others? It would be helpful to understand the rationale, especially considering their benefits in dependency locking, environment management and modern workflows.
> 
> Official Python Packaging Authority (PyPA) guidance: https://packaging.python.org/en/latest/guides/tool-recommendations/

I can see the idea in migrating to a more modern tool like Poetry, but I think it would be hard to do that with Rucio at this stage. Rucio packaging is still done in a fairly manual way until we complete the namespace package restructure, which would allow us to actually build the different packages separately without needing to rename the setup script everytime. You're welcome to investigate the migration, but I don't think it's worth considering this for now.

Geogouz:
According to the linked PR #7675, this issue shall be handled by that. However, there are TODOs here in the description that are pointing to other PRs (which are delayed). I assume those TODOs are just related to this issue and not required for addressing it. Right?

rdimaio:
[...]

PR #7575, opened 2025-03-27
Dependencies: add upper version limit to all server dependencies; #7571

rdimaio:
A more drastic solution in case https://github.com/rucio/rucio/pull/7573 doesn't suffice. Based on versions here https://github.com/rucio/rucio/blob/master/requirements/requirements.dev.in

These versions are temporarily duplicated. In the next release cycle, I think we should 

1. restructure into a namespace package: [rucio/rucio#7330](https://github.com/rucio/rucio/issues/7330)
2. fully move to pyproject.toml: [rucio/rucio#6971](https://github.com/rucio/rucio/issues/6971)

and we will be able to remove this duplication as part of the `setup.py` deprecation.


setuputil.py
```diff
@@ -61,41 +61,53 @@
...
-    'install_requires': clients_requirements_table['install_requires'] + [
-        'argcomplete',
+    'install_requires': [
+        'requests<=2.32.3',
+        'urllib3<=1.26.19',
+        'dogpile-cache<=1.2.2',
+        'packaging<=24.1',
+        'tabulate<=0.9.0',
+        'jsonschema<=4.23.0',
+        'dataclasses',
```
0xquark: Is there a reason we haven't set an upper limit for this (`dataclasses`)? If it's intentional, maybe adding a comment might be good.

rdimaio: There's no upper limit for it set in neither the `.in` or `.txt` files, same thing for `boto`. I'll add a comment

rdimaio: Added in latest revision

0xquark (approved):

PR #7577, opened 2025-03-27
[DRAFT]: Rucio restructure into namespace package

rdimaio:
See https://github.com/rucio/rucio/issues/7330 for reference.

This PR is not meant to be merged at any point - it only serves the purpose of showcasing how an eventual restructure would look like as far as directories, `pyproject.toml` files and dependency management. I suggest navigating the file structure here to get an idea: https://github.com/rdimaio/rucio/tree/test-restructure

This is inspired by the structure of DiracX, outlined here: https://github.com/DIRACGrid/diracx/blob/main/docs/VERSIONING.md#python-modules


rdimaio:
Closing, to be picked up again in future

PR #7589, opened 2025-03-28
Core: Add option to ignore duplicates in did to container attachments

### LINKED ISSUE #7585: Add option to ignore duplicate content in did to did attachment
Currently the `attach_dids_to_dids` method supports a `ignore_duplicate` flag.

When this flag is turned on, adding a file to a dataset when it already exists will not produce an exception. This is very useful for bulk attachments.

IMHO this option should also take effect when mapping containers to containers or datasets to containers. Otherwise this results in a `rucio.common.exception.DuplicateContent`.

Not sure why this was not added in the first place, perhaps there are some details I am not aware of.

lobis:
Closes #7585

### DIFF
--- a/lib/rucio/core/did.py
+++ b/lib/rucio/core/did.py
@@ -669,6 +669,7 @@ def __add_collections_to_container(
     collections_temp_table: Any,
     collections: "Mapping[tuple[InternalScope, str], Mapping[str, Any]]",
     account: "InternalAccount",
+    ignore_duplicate: bool = False,
     *,
     session: "Session"
 ) -> None:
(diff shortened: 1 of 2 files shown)


rdimaio (commented):
LGTM, but I am not sure either about: 

> Not sure why this was not added in the first place, perhaps there are some details I am not aware of.

I'll defer the approval to someone with more long-term context on this

lobis:
> I'll defer the approval to someone with more long-term context on this

I had some offline discussions with @bari12 and we think this was just forgotten / not added. The implementation is a bit more involved than the other `ignore_duplicate` case so perhaps that is why.

I added some tests and I think it should be good to go now.

rdimaio:
> > I'll defer the approval to someone with more long-term context on this
> 
> I had some offline discussions with @bari12 and we think this was just forgotten / not added. The implementation is a bit more involved than the other `ignore_duplicate` case so perhaps that is why.
> 
> I added some tests and I think it should be good to go now.

Sounds good, thanks for the context

rdimaio (approved):

bari12 (approved):

PR #7590, opened 2025-03-31
common: Rename `get_lfn2pfn_algorithm_default` to `get_lfn2pfn_algorithm` for naming consistency

### LINKED ISSUE #7465: Rename `lfn2pfn_algorithm_default` option to same naming convention as other algorithms, like `pfn2lfn` (i.e. remove `_default`)
https://github.com/rucio/rucio/blob/d10519c8e1c10260e0c52cc1df41774fec2d5754/lib/rucio/common/config.py#L727

0xquark:
Fix #7465

### DIFF
--- a/lib/rucio/common/config.py
+++ b/lib/rucio/common/config.py
@@ -724,7 +724,7 @@ def get_config_dirs() -> list[str]:
     return configdirs
 
 
-def get_lfn2pfn_algorithm_default() -> str:
+def get_lfn2pfn_algorithm() -> str:
     """Returns the default algorithm name for LFN2PFN translation for this server."""
     default_lfn2pfn = "hash"
     try:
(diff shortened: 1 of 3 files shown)


lobis:
My current understanding of this:

This is named `_default` because it is the fallback option, when no algorithmn is present in the RSE (can this actually happen? isn't always a algorithmn in the RSE?).

So according to my experience if you change the default algoritmn on the client side, nothing will happen since the RSE will always have an algorithmn that will override it.

I'm not sure what this actually does tbh.

bari12:
I agree with @lobis 
I think the `pfn2lfn` case is just different. But here this is about getting the `default` one, should no other alogirthm be specified. Renaming it is actually detrimental I would say. I am closing this, please let me know if you strongly feel different :smiling

PR #7592, opened 2025-03-31
Opendata: Initial implementation

### LINKED ISSUE #7559: Opendata support in Rucio
XL Issue to track the development of the Opendata native support in Rucio.

lobis:
Closes #7559 

This PR introduces the rucio opendata interface and has changes across different components:

- DB model: Three new opendata tables added along with the migration files.
- API: Two new endpoints: `opendata` at `/opendata` and `opendata_public` at `/opendata/public` introduced. `opendata` is the generic internal endpoint and `opendata_public` is an unauthenticated endpoint which allows read only access to open data dids marked as public.
- CLI/Client: a new top level command `rucio opendata` is introduced with basic functionality.
- Core: Added the necessary opendata methods.


labkode:
@lobis the logic looks good to me, for a Python-in-depth review I suggest you ask to review someone else.

lobis:
At this time tests are failing because of an unrelated CI issue.

lib/rucio/web/rest/flaskapi/v1/opendata_public.py
```diff
@@ -0,0 +1,53 @@
...
+from flask import Blueprint, Flask, Response
+
+from rucio.web.rest.flaskapi.v1.common import ErrorHandlingMethodView, check_accept_header_wrapper_flask, response_headers
+from rucio.web.rest.flaskapi.v1.opendata import OpenDataDIDsView, OpenDataView
+
+
+class OpenDataPublicView(ErrorHandlingMethodView):
+
+    @check_accept_header_wrapper_flask(["application/json"])
+    def get(self) -> "Response":
```
voetberg, line 24: These methods need to have documentation for the payload and the return so they show up in our openapi docs. https://rucio.cern.ch/documentation/developer/rest_api_doc

lib/rucio/cli/opendata.py
```diff
@@ -0,0 +1,118 @@
...
+        msg = f"Cannot extract scope and name from DID {did}. The DID should have exactly one colon but found {len(parts)} colons."
+        raise RucioException(msg)
+    scope, name = parts
+    if not scope or not name:
+        msg = f"Cannot extract scope and name from DID {did}. Found empty scope or name."
+        raise RucioException(msg)
+    return scope, name
+
+
+def is_valid_json(s: str) -> bool:
```
voetberg, line 30: This can be done in the argument type instead of optionally validating https://click.palletsprojects.com/en/stable/parameter-types/

lib/rucio/cli/opendata.py
```diff
@@ -0,0 +1,118 @@
...
+        return False
+
+
+@click.group()
+def opendata() -> None:
+    """Manage Open Data DIDs"""
+
+
+@opendata.command("list")
+# TODO instead of state, maybe use a flag for each valid state?
```
voetberg: Flags cannot be mutually exclusive so the logic gets really messy really fast if you do it this way. I recommend using `type=click.Choice([valid states])`

voetberg:
One design critique - having both the public endpoint and the regular endpoint is a strange choice. I would instead make the public routing an argument so the url is more like `/opendata/public=False/` and `/opendata/public=True/`  and route to different gateway functions depending, sort of the way you do in the client. That routing logic should happen closer to core if possible.

lib/rucio/cli/opendata.py
```diff
@@ -0,0 +1,118 @@
...
+
+import click
+
+from rucio.common.exception import RucioException
+
+if TYPE_CHECKING:
+    from click import Context
+
+
+def extract_scope_name(did: str) -> tuple[str, str]:
```
voetberg: This already exists in bin_legacy, I'd recommend moving it to `cli.utils` and working off that

lib/rucio/tests/common.py
```diff
@@ -150,6 +150,12 @@ def rfc2253_dn_generator() -> str:
     random_dn = "CN={}, O={}, C={}".format(random_cn, random_o, random_c)
     return random_dn
 
+def doi_generator() -> str:
```
voetberg: Is this used anywhere besides tests? If not, this should probably be a fixture in tests/conftest.py

tests/test_opendata.py
```diff
@@ -0,0 +1,331 @@
...
+        assert "files" not in opendata_did, "Files should not be present in the response"
+        assert "meta" not in opendata_did, "Meta should not be present in the response"
+
+        opendata_did = rucio_client.get_opendata_did(scope=str(mock_scope), name=name, files=True, meta=True, doi=True)
+        assert opendata_did["doi"] is None, "DOI should still be None"
+        assert "files" in opendata_did, "Files should be present in the response"
+        assert "meta" in opendata_did, "Meta should be present in the response"
+        meta = opendata_did["meta"]
+        assert meta == {}, "'meta' should be empty"
+
```
voetberg, line 384: Do you have CLI tests anywhere? They would probably also fail at this stage but they are important to verify

lib/rucio/gateway/opendata.py
```diff
@@ -0,0 +1,131 @@
...
+from rucio.common.utils import gateway_update_return_dict
+from rucio.core import opendata
+from rucio.core.opendata import check_valid_opendata_did_state, opendata_state_str_to_enum
+from rucio.db.sqla.session import read_session, transactional_session
+
+if TYPE_CHECKING:
+    from sqlalchemy.orm import Session
+
+
+@read_session
```
voetberg: Should these be replaced with context managers instead? Re: https://github.com/rucio/rucio/issues/7407#issue-2857092248

lib/rucio/cli/opendata.py
lobis: I will use the rucio built in extract scope method

lib/rucio/cli/opendata.py
lobis: Added new click json type (click does not have a built in one) and will use that instead

lib/rucio/tests/common.py
lobis: Refactored to use fixture

tests/test_opendata.py
lobis: Added CLI tests and will test the full API in CLI too

labkode:
> One design critique - having both the public endpoint and the regular endpoint is a strange choice. I would instead make the public routing an argument so the url is more like `/opendata/public=False/` and `/opendata/public=True/` and route to different gateway functions depending, sort of the way you do in the client. That routing logic should happen closer to core if possible.

I think having the `public` argument  in the URL routing would make this API the weirdest one I've seen :)

I think having the endpoint exposed under `/public/opendata` is not a bad idea and a regular pattern used in other systems, for example, it's common in filesystems to have a `public` folder to transmit the notion of openness to the world.

lib/rucio/web/rest/flaskapi/v1/opendata_public.py
lobis: Added docs!

lib/rucio/cli/opendata.py
lobis: Added valid states option

lib/rucio/client/opendataclient.py
voetberg, line 1: Please add docstrings for each of these methods!!

lib/rucio/cli/opendata.py
```diff
@@ -0,0 +1,105 @@
...
+    scope, name = extract_scope(did)
+    result = client.get_opendata_did(scope=scope, name=name, public=public, files=files, meta=meta, doi=True)
+    # TODO: pretty print using tables, etc
+    print(json.dumps(result, indent=4, sort_keys=True, ensure_ascii=False))
+
+
+@opendata.command("update")
+@click.argument("did")
+@click.option("--meta", type=JSONType(), required=False, help="OpenData JSON")
+# TODO: do not hardcode the list of valid states but import them
```
voetberg: This TODO looks like it's finished, no? Best to take it out

lib/rucio/cli/opendata.py
voetberg, line 1: Each of these methods should have descriptions show they have documentation in the `--help` menu

lib/rucio/cli/opendata.py
```diff
@@ -0,0 +1,105 @@
...
+@click.option("--state", type=click.Choice(valid_opendata_states, case_sensitive=False), required=False, help="State")
+@click.option("--doi", required=False, help="DOI")
```
voetberg: The help on this option is pretty redudent, can you say something about either a) the restrictions on the doi, or what it does besides being used for bookkeeping?

lib/rucio/cli/opendata.py
```diff
@@ -0,0 +1,105 @@
...
+@click.pass_context
+def update_opendata_did(ctx: "Context", did: str, meta: Optional[str], state: Optional[str],
+                        doi: Optional[str]) -> None:
+    client = ctx.obj.client
+    if not any([meta, state, doi]):
+        raise click.UsageError("At least one of --meta, --state, or --doi must be provided.")
```
voetberg: While `usageerror` works here, in most other CLI functions, this situation raises a `ValueError` instead

lobis: Replaced with `ValueError` for consistency.

lib/rucio/cli/opendata.py
lobis: Updated help text

lib/rucio/cli/opendata.py
lobis: Added docs

lib/rucio/cli/opendata.py
lobis: Removed comment

lib/rucio/client/opendataclient.py
lobis: Added docstrings

lib/rucio/client/opendataclient.py
```diff
@@ -0,0 +1,198 @@
...
+        """
+        if public and self.opendata_host_from_config is not None:
+            return self.opendata_host_from_config
+
+        return choice(self.list_hosts)
+
+    def list_opendata_dids(
+            self,
+            *,
+            state: Optional[str] = None,
```
voetberg: Is there a more restrictive version (some enum subtype?) of `state`?

lib/rucio/client/opendataclient.py
```diff
@@ -0,0 +1,198 @@
...
+            public: bool = False,
+    ) -> dict[str, Any]:
+        """
+        Returns a list of Opendata DIDs optionally filtered by state and public/private access.
+        """
```
voetberg, line 74: Please include descriptions for `state` and `public`. (Sorry for being so pedantic on this!!)

lib/rucio/client/opendataclient.py
```diff
@@ -0,0 +1,198 @@
...
+            exc_cls, exc_msg = self._get_exception(headers=r.headers, status_code=r.status_code, data=r.content)
+            raise exc_cls(exc_msg)
+
+    def add_opendata_did(
+            self,
+            *,
+            scope: str,
+            name: str,
+    ) -> bool:
+        """
```
voetberg, line 101: Please include definitions for scope and name

lib/rucio/client/opendataclient.py
```diff
@@ -0,0 +1,198 @@
...
+            raise exc_cls(exc_msg)
+
+    def remove_opendata_did(
+            self,
+            *,
+            scope: str,
+            name: str,
+    ) -> bool:
+        """
+        Removes an existing Opendata DID from the Opendata catalog.
```
voetberg: same as above! Please document your inputs!

lib/rucio/client/opendataclient.py
```diff
@@ -0,0 +1,198 @@
...
+            *,
+            scope: str,
+            name: str,
+            state: Optional[str] = None,
+            meta: Optional[dict] = None,
+            doi: Optional[str] = None,
+    ) -> bool:
+        """
+        Updates an existing Opendata DID in the Opendata catalog.
+        At least one of the optional parameters must be provided for update.
```
voetberg: Same as above. Please include docstrings for inputs

lib/rucio/client/opendataclient.py
lobis: Refactored to use literal instead of strings for typing

lib/rucio/client/opendataclient.py
lobis: Updated docstrings!

lib/rucio/client/opendataclient.py
lobis: Updated docstrings!

lib/rucio/client/opendataclient.py
lobis: Updated docstrings!

lib/rucio/client/opendataclient.py
lobis: Updated docstrings!

lib/rucio/client/opendataclient.py
voetberg: It is better not to include the types for the docstrings [ref](https://rucio.cern.ch/documentation/developer/dev_style_guide/) so they don't overwrite the actual types in the type hints. Remove those and we can call this done!

lobis: Got it, I think it should be good to go now!

lobis: I removed types from parameters and return

voetberg (dismissed):
lgtm!!

lobis:
I will rebase and tidy up commits without any change in code

lib/rucio/common/exception.py
```diff
@@ -1206,3 +1206,58 @@ def __init__(self, param: str, *args):
...
+
+    def __init__(self, *args):
+        super(OpenDataDataIdentifierNotFound, self).__init__(*args)
+        self._message = f"Data identifier not found in the open data catalog."
+        self.error_code = 116
+
+
+class OpenDataDataIdentifierAlreadyExists(OpenDataError):
+    """
+    Throws when the data identifier is not in the open data catalog.
```
rdimaio:
```suggestion
    Throws when the data identifier already exists in the open data catalog.
```

tests/test_opendata.py
```diff
@@ -0,0 +1,571 @@
...
+                                         session=db_write_session)
+
+        opendata.update_opendata_did(scope=mock_scope, name=name, state=OpenDataDIDState.PUBLIC,
+                                     session=db_write_session)
+
+        db_write_session.commit()
+
+        state = opendata.get_opendata_did(scope=mock_scope, name=name, session=db_write_session)["state"]
+
+        assert state == OpenDataDIDState.PUBLIC
```
rdimaio, line 191: This one could potentially be parametrized with valid and invalid state updates to reduce some duplication

lib/rucio/client/opendataclient.py
```diff
@@ -0,0 +1,249 @@
...
+            raise exc_cls(exc_msg)
+
+    def get_opendata_did(
+            self,
+            *,
+            scope: str,
+            name: str,
+            files: bool = False,
+            meta: bool = False,
+            doi: bool = True,
```
rdimaio: I think these names should sound a bit more 'boolean-like' e.g. `files` is generally used to represent a `list[dict]` of files in Rucio, so this might potentially be confusing

lib/rucio/db/sqla/migrate_repo/versions/a62db546a1f1_opendata_initial_model.py
```diff
@@ -0,0 +1,84 @@
...
+# Alembic revision identifiers
+revision = 'a62db546a1f1'
+down_revision = '30d5206e9cad'
+
+
+def upgrade():
+    op.create_table(
+        'dids_opendata',
+        sa.Column('scope', sa.String(length=25), nullable=False),
+        sa.Column('name', sa.String(length=255), nullable=False),
```
rdimaio: Instead of hardcoding these, we could `get_schema_value('SCOPE_LENGTH')` and `get_schema_value('NAME_LENGTH')`, like it's done in `1d1215494e95_add_quarantined_replicas_table.py`

lib/rucio/gateway/opendata.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+def get_opendata_did(
+        *,
+        scope: str,
+        name: str,
+        state: Optional["OPENDATA_DID_STATE_LITERAL"] = None,
+        files: bool = True,
+        meta: bool = False,
+        doi: bool = True,
```
rdimaio: same comment about bool variable names that should ideally sound more bool-like

rdimaio (dismissed):
In principle I think this PR is way too large to be properly reviewed. We can take the same approach we took with the client refactor and merge it anyway and work through potential bugs as we go along, and ideally this PR's scope will only affect OpenData-related things, but in future I think it would be best if we implement this kind of functionality in a more gradual way.

lobis:
> In principle I think this PR is way too large to be properly reviewed. We can take the same approach we took with the client refactor and merge it anyway and work through potential bugs as we go along, and ideally this PR's scope will only affect OpenData-related things, but in future I think it would be best if we implement this kind of functionality in a more gradual way.

I agree in principle but how to implement this in practice when the PR has many parts that need to work together (core needs database, client needs core, etc.). I guess we could make a chain of PRs (which need to live in the fork I guess) but I suspect this would not make things easier. How would you do it?

The PR is very isolated on purpose so I think it is safe to merge and fix issues as they come.

rdimaio:
> > In principle I think this PR is way too large to be properly reviewed. We can take the same approach we took with the client refactor and merge it anyway and work through potential bugs as we go along, and ideally this PR's scope will only affect OpenData-related things, but in future I think it would be best if we implement this kind of functionality in a more gradual way.
> 
> I agree in principle but how to implement this in practice when the PR has many parts that need to work together (core needs database, client needs core, etc.). I guess we could make a chain of PRs (which need to live in the fork I guess) but I suspect this would not make things easier. How would you do it?
> 
> The PR is very isolated on purpose so I think it is safe to merge and fix issues as they come.

I think it makes sense to do something like 
1. database + core + db/core tests
2. gateway + api + gateway/api tests
3. client + client tests

A big reason is that we have different component leads for all of these, so splitting the PRs like this would make it easier for them to review them in isolation

lib/rucio/gateway/opendata.py
lobis: Addressed!

lib/rucio/db/sqla/migrate_repo/versions/a62db546a1f1_opendata_initial_model.py
lobis: Addressed!

lib/rucio/client/opendataclient.py
lobis: Addressed!

tests/test_opendata.py
lobis: I tried to come up with a way to parametrize it but IMHO it was way more confusing to read (I don't mind duplication in tests near as much as in the code). I may refactor in another PR

lib/rucio/common/exception.py
lobis: Thanks!

rdimaio (approved):

PR #7594, opened 2025-03-31
common: Remove unused legacy config handling in `get_legacy_config`

### LINKED ISSUE #7376: Legacy config support
This function:

https://github.com/rucio/rucio/blob/e19e196bfa438b5c5b6c583a15945fe2b7743c43/lib/rucio/common/config.py#L220-L237

Unless I'm missing something, `LEGACY_SECTION_NAME` and `LEGACY_OPTION_NAME` are re-initialized to empty dicts every time this is called, and these dicts don't seem to be populated anywhere. In this case, the `get()` actions would always return `None`. 

This was added here: https://github.com/rucio/rucio/commit/c3e376863b80d4234355c6aed2e0f87b3aa603da, with the message

> We will rename some configs in the near future (e.g. https://github.com/rucio/rucio/issues/4795). This commit
introduces the boilerplate to fetch deprecated configuration names and
use them if the new ones do not exist.

I think the idea was to eventually populate `LEGACY_SECTION_NAME` and `LEGACY_OPTION_NAME` with legacy section and option names, but it has not been done so far.

Is this something we still want to support?

0xquark:
Fixes #7376

### DIFF
--- a/lib/rucio/common/config.py
+++ b/lib/rucio/common/config.py
@@ -31,8 +31,6 @@ if TYPE_CHECKING:
 
     from sqlalchemy.orm import Session
 
-LEGACY_SECTION_NAME = {}
-LEGACY_OPTION_NAME = {}
 
 
 def convert_to_any_type(value: str) -> Union[bool, int, float, str]:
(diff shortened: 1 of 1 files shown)


rdimaio (changes requested):
Please remove this legacy config feature as a whole (i.e. this function + its usages)

rdimaio (approved):

PR #7610, opened 2025-04-02
Testing: extend `pylint` rules in `ruff`

### LINKED ISSUE #7499: Extend Pylint Rules in Ruff
### Description

We recently switched from using pylint and flake8 to Ruff as our main linter. Right now, we've enabled Ruff's default rules for pylint, as shown in this [PR](https://github.com/rucio/rucio/pull/7481). However, we want to expand the pylint rules in Ruff to better match our coding standards and ensure consistent quality checks across the project.

The [issue](https://github.com/astral-sh/ruff/issues/970) in Ruff tracks the pylint rules that Ruff currently supports.

This issue will be used to track any PRs/Discussions related to extending linting support in Rucio.

0xquark:
Fixes #7499

### DIFF
--- a/lib/rucio/core/did_meta_plugins/filter_engine.py
+++ b/lib/rucio/core/did_meta_plugins/filter_engine.py
@@ -102,9 +102,7 @@ class FilterEngine:
             filters, _ = parse_did_filter_from_string_fe(filters, omit_name=True)
         elif isinstance(filters, dict):
             filters = [filters]
-        elif isinstance(filters, list):
-            pass  # already a list of filters
-        else:
+        elif not isinstance(filters, list):
             raise exception.DIDFilterSyntaxError("Input filters are of an unrecognised type.")
 
         filters = self._make_input_backwards_compatible(filters=filters)
--- a/lib/rucio/core/did_meta_plugins/filter_engine.py
+++ b/lib/rucio/core/did_meta_plugins/filter_engine.py
@@ -103,7 +103,7 @@ class FilterEngine:
         elif isinstance(filters, dict):
             filters = [filters]
         elif isinstance(filters, list):
-            filters = filters
+            pass  # already a list of filters
         else:
             raise exception.DIDFilterSyntaxError("Input filters are of an unrecognised type.")
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -64,6 +64,10 @@ select = [
     "PLE", #pylint
 ]
 
+extend-select = [
+    "F401" # unused-import
+]
+
 ignore = [
     "ALL",
     "E731", # Do not assign a lambda expression, use a def
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -65,7 +65,11 @@ select = [
 ]
 
 extend-select = [
-    "F401" # unused-import
+    "F401", # unused-import
+    "F841", # Unused variable
+    "W291", # Trailing whitespace (low-hanging fruit that improves cleanliness)
+    "E722", # Do not use bare except, specify exception
+    "D419", # Empty docstring (suggests bad documentation hygiene.)
 ]
 
 ignore = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -71,6 +71,12 @@ extend-select = [
     "E722", # Do not use bare except, specify exception
     "D419", # Empty docstring (suggests bad documentation hygiene.)
     "PLW0127", # Self-assigning variable
+    "PLW0128", # Redeclared assigned name
+    "UP004", # Useless object inheritance
+    "PLE0101", # return in __init__ – breaks instantiation logic.
+    "PLE0704", # Misplaced bare raise – makes debugging harder.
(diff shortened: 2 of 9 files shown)


pyproject.toml
```diff
@@ -64,6 +64,28 @@ select = [
...
+    "PLE0101", # return in __init__ – breaks instantiation logic.
+    "PLE0704", # Misplaced bare raise – makes debugging harder.
+    "PLE1142", # Await outside async function – common async error.
+    "PLR1711", # Useless return - stylistic but can be misleading.
+    "PLR0124", # Comparison with itself – almost always a bug, but rare.
+    "PLW3301", # Nested min/max – niche readability case.
+    "PLR1702", # Too many nested blocks
+    "PLE0604", # Ensures __all__ is correctly defined.
+    "PLE1300", # Catches invalid string format chars.
+    "PLE0241", # duplicate-bases – can cause runtime errors.
```
rdimaio: Individual PLE rules don't need to be added to `extend-select`, we are already selecting `PLE` as a whole in `select`

pyproject.toml
```diff
@@ -64,6 +64,28 @@ select = [
...
+    "F401",    # unused-import
+    "F841",    # Unused variable
+    "W291",    # Trailing whitespace (low-hanging fruit that improves cleanliness)
+    "D419",    # Empty docstring (suggests bad documentation hygiene.)
+    "PLW0127", # Self-assigning variable
+    "PLW0128", # Redeclared assigned name
+    "UP004",   # Useless object inheritance
+    "PLE0101", # return in __init__ – breaks instantiation logic.
+    "PLE0704", # Misplaced bare raise – makes debugging harder.
+    "PLE1142", # Await outside async function – common async error.
```
rdimaio: Individual PLE rules don't need to be added to `extend-select`, we are already selecting `PLE` as a whole in `select`

pyproject.toml
```diff
@@ -64,6 +64,28 @@ select = [
...
 ]
 
+extend-select = [
+    "F401",    # unused-import
+    "F841",    # Unused variable
+    "W291",    # Trailing whitespace (low-hanging fruit that improves cleanliness)
+    "D419",    # Empty docstring (suggests bad documentation hygiene.)
+    "PLW0127", # Self-assigning variable
+    "PLW0128", # Redeclared assigned name
+    "UP004",   # Useless object inheritance
```
rdimaio: Individual UP rules don't need to be added to `extend-select`, we are already selecting `UP` as a whole in `select`

pyproject.toml
```diff
@@ -64,6 +64,28 @@ select = [
...
+    "PLE1142", # Await outside async function – common async error.
+    "PLR1711", # Useless return - stylistic but can be misleading.
+    "PLR0124", # Comparison with itself – almost always a bug, but rare.
+    "PLW3301", # Nested min/max – niche readability case.
+    "PLR1702", # Too many nested blocks
+    "PLE0604", # Ensures __all__ is correctly defined.
+    "PLE1300", # Catches invalid string format chars.
+    "PLE0241", # duplicate-bases – can cause runtime errors.
+    "F702",    # not-in-loop
+    "PLW0133", # pointless-exception-statement
```
rdimaio: Please sort this alphabetically so that codes within the same group are next to each other (so that all `PLW` ones are one after the other and so on)

tests/test_module_import.py
```diff
@@ -17,7 +17,9 @@
 
 class TestModuleImport:
     def test_import(self):
-        """ """
+        """
+        RUCIO: Test if Rucio module can be imported without errors.
```
rdimaio, line 21: I personally would add `D419` (https://docs.astral.sh/ruff/rules/empty-docstring/) to the list of `'tests/*.py'` ignores in `[tool.ruff.lint.per-file-ignores]` in the cfg. I think a lot of tests are self-documented through their function name, and enforcing a docstring for them might be a bit too much (but it can be welcome in case it's useful; I think it's useful in the case of `test_import` here, for example)

lib/rucio/core/did_meta_plugins/filter_engine.py
```diff
@@ -103,7 +103,7 @@ def __init__(
         elif isinstance(filters, dict):
             filters = [filters]
         elif isinstance(filters, list):
-            filters = filters
+            pass  # already a list of filters
         else:
             raise exception.DIDFilterSyntaxError("Input filters are of an unrecognised type.")
```
rdimaio: Can be simplified:

```python
elif not isinstance(filters, list):
    raise exception.DIDFilterSyntaxError("Input filters are of an unrecognised type.")
```

pyproject.toml
0xquark: I thought select only had some default `PLE` rules, not all of them, but I double-checked and you're correct—it includes all of them. So, we don't need to explicitly manage `PLE` rules in `extend-select`.

rdimaio (approved):

Issue #7605, opened 2025-04-01
API: static code analyzers misunderstand `WSGIEnvironment` queries


rdimaio:
This is tied to https://github.com/rucio/rucio/issues/7447

Right now, in API routes, we obtain the environment values like `vo` and `issuer` via a `get`, e.g.:

https://github.com/rucio/rucio/blob/eeddb248feac80224c33819acb00d7f8335bbd10/lib/rucio/web/rest/flaskapi/v1/scopes.py#L87

`request.environ` is a `WSGIEnvironment`, and from typeshed:

```
WSGIEnvironment: TypeAlias = dict[str, Any]
```

So static code analyzers raise errors here, saying that `vo` and `issuer` could be `None`.

However, every private API blueprint we have is based on `AuthenticatedBlueprint`:

https://github.com/rucio/rucio/blob/eeddb248feac80224c33819acb00d7f8335bbd10/lib/rucio/web/rest/flaskapi/authenticated_bp.py#L19

Which, before each request, runs this:

https://github.com/rucio/rucio/blob/eeddb248feac80224c33819acb00d7f8335bbd10/lib/rucio/web/rest/flaskapi/v1/common.py#L140-L160

This essentially means that those `WSGIEnvironment` values are guaranteed to exist once `request_auth_env` is complete (otherwise we'd run into one of the exceptions in `request_auth_env` - e.g. `validate_auth_token` queries an auth token with required issuer account). However, static code analyzers don't understand this.

A simple refactoring here would be to access the key directly rather than using `get`, i.e. `request.environ.get('vo')` -> `request.environ['vo']` - this way, we take 'responsibility' that the key exists (and we know it does), and code analyzers are okay with that

rdimaio:
We should also mention this in the developer docs for API development, it'd make things clearer in the future. This will probably change after an eventual Flask -> FastAPI migration though

Issue #7409, opened 2025-02-17
Migrate `*_session` decorated gateway functions to use context managers instead


rdimaio:
Issue to track initial migration of `session`-decorated DB gateway functions to use context managers instead.

PR #7618, opened 2025-04-02
Migrate gateway/scope and gateway/dirac to session context manager, minor refactors

rdimaio:
Part of:
- #7409
- #7605 

In addition, note this commit 270288946b4febda8ba0392ea39428b30b097b93 - this is the reason why I had to make a single PR for both scope and dirac. DIRAC used the gateway version of `list_scopes` instead of the core version, so I had to change it here, as we want to avoid having gateway functions calling other gateway functions (except for `has_permission`, which does not have an API route and is effectively a 'gateway utility' rather than a real gateway function)


0xquark (approved):
LGTM!

PR #7619, opened 2025-04-02
Revert stomp refactor

bari12:

### DIFF
--- a/lib/rucio/daemons/hermes/hermes.py
+++ b/lib/rucio/daemons/hermes/hermes.py
@@ -21,8 +21,10 @@ import datetime
 import functools
 import json
 import logging
+import random
 import re
 import smtplib
+import socket
 import sys
 import threading
 import time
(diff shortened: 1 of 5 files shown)


lib/rucio/daemons/hermes/hermes.py
```diff
@@ -467,7 +748,15 @@ def run_once(heartbeat_handler: "HeartbeatHandler", bulk: int, **_kwargs) -> boo
         if "activemq" in message_dict:
             t_time = time.time()
             try:
-                messages_sent = conn_mgr.deliver_messages(messages=message_dict["activemq"])
+                messages_sent = deliver_to_activemq(
+                    messages=message_dict["activemq"],
+                    conns=conns,
```
rdimaio: we should add `# type: ignore (argument could be None)` for this one too

lib/rucio/daemons/hermes/hermes.py
```diff
@@ -137,16 +397,15 @@ def submit_to_elastic(
...
+    res = requests.post(
+        endpoint, data=text, headers={"Content-Type": "application/json"}, auth=auth
+    )
     return res.status_code
 
 
 def aggregate_to_influx(
     messages: "Iterable[dict[str, Any]]",
-    bin_size: str,
+    bin_size: int,
```
rdimaio: I think `bin_size` is expected to be `str`. The only usage I can find is `bin_size="1m"`.

rdimaio:
```
   - 1 errors with message """Argument of type "Literal['1m']" cannot be assigned to parameter "bin_size" of type "int" in function "aggregate_to_influx"
                               "Literal['1m']" is not assignable to "int"""".
    Candidates: line 679
```

rdimaio (commented):
```
 Found 4 new problems in /home/runner/work/rucio/rucio/lib/rucio/daemons/hermes/hermes.py
  - 2 errors with message """"exception" is not a known attribute of module "stomp"""".
    Candidates: line 270, 277
```

I think this can be ignored, `exception` is part of `stomp`, not sure why it's complaining here. It might go away if we do `from stomp.exception import ConnectFailedException, NotConnectedException` like in the current version.

rdimaio (approved):
LGTM - the `""""exception" is not a known attribute of module "stomp""""` error could possibly be fixed by doing `from stomp.exception import ...` but let's ignore it for now

bari12:
I'll ignore it for now. This will need to be reworked anyway in a future minor release.
This is just to get the major release out.

PR #7621, opened 2025-04-02
CLI: Correct Rich Pager logic

### LINKED ISSUE #7560: Use of Rich does not produce any output
### Description

Running `rucio whoami` terminates successfully but does not print anything to standard output. Same for other commands.

### Steps to reproduce

```console
$ rucio whoami
```

with the following configuration option:

```ini
[experimental]
cli = rich
```

### Rucio Version

37.0.0rc1

### Additional Information

_No response_

voetberg:
Closes #7560 

Output was not being printed because the logic for pager/no-pager was scrambled. No-pager was also not being passed forward as an arg.

### DIFF
--- a/lib/rucio/cli/account.py
+++ b/lib/rucio/cli/account.py
@@ -51,7 +51,7 @@ def add_(ctx, account_type, account_name, email):
     Example:
         $ rucio account add
     """
-    args = Arguments({"account_type": account_type, "account": account_name, "email": email})
+    args = Arguments({"no_pager": ctx.obj.no_pager, "account_type": account_type, "account": account_name, "email": email})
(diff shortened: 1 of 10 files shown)


0xquark (approved):
LGTM! 

Suggestion(nit): 
- Does including a small helper function for `Arguments({"no_pager": ctx.obj.no_pager, ...})` makes sense considering the repetition?

voetberg:
> LGTM!
> 
> Suggestion(nit):
> 
>     * Does including a small helper function for `Arguments({"no_pager": ctx.obj.no_pager, ...})` makes sense considering the repetition?

This would probably be cleaner, and I considered just changing the definition of Arguments to automatically include the `context` object but I think this is more obvious for trouble shooting purposes as we plan to just remove the use of the `Arguments` proxy as we transition away from using the legacy functions

PR #7628, opened 2025-04-03
Database: gateway/rule, migrate to session context manager

rdimaio:
Part of:
- #7409 
- #7605

### DIFF
--- a/lib/rucio/gateway/rule.py
+++ b/lib/rucio/gateway/rule.py
@@ -20,7 +20,8 @@ from rucio.common.schema import validate_schema
 from rucio.common.types import InternalAccount, InternalScope
 from rucio.common.utils import gateway_update_return_dict
 from rucio.core import rule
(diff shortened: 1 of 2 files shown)


lib/rucio/gateway/rule.py
```diff
@@ -29,16 +30,14 @@
...
 
 
-@read_session
-def is_multi_vo(*, session: "Session") -> bool:
+def is_multi_vo(session: "Session") -> bool:
     """
     Check whether this instance is configured for multi-VO
     returns: Boolean True if running in multi-VO
     """
     return config_get_bool('common', 'multi_vo', raise_exception=False, default=False, session=session)
```
rdimaio, line 38: Note: this `is_multi_vo` is only used as a convenenience function by other gateway functions within this file, so it's intended usage is to pass `session` to it, without creating a new `db_session` in the function itself.

0xquark (approved):
LGTM!

Issue #7609, opened 2025-04-02
Gateway: `issuer` is marked `Optional` and defaults to `None`, but it is required for permission checking (like other authenticated BPs)


rdimaio:
### Description

e.g.

https://github.com/rucio/rucio/blob/eeddb248feac80224c33819acb00d7f8335bbd10/lib/rucio/gateway/heartbeat.py#L54

### Steps to reproduce

N/A

rdimaio:
Found the same issues in `gateway/config` - expanding this issue to be general for gateway as a whole

PR #7639, opened 2025-04-08
gateway/subscription, migrate to session context manager

rdimaio:
Part of:
- #7409 
- #7605
- #7609


rdimaio:
Last revision: fix yield bug (see https://github.com/rucio/rucio/issues/7657)

bari12 (approved):

PR #7644, opened 2025-04-09
API: Allow to add / remove endpoints from config

### LINKED ISSUE #7643: Allow adding api endpoints to defaults
Currently the available api endpoints can be explicitly defined via a configuration but this only allows to set the endpoints, so a user that wants to enable one additional endpoint needs to write the whole list.

There should be an option to just add / remove and endpoint to the list

lobis:
Closes #7643

### DIFF
--- a/lib/rucio/web/rest/flaskapi/v1/main.py
+++ b/lib/rucio/web/rest/flaskapi/v1/main.py
@@ -15,6 +15,7 @@
 
 import importlib
 import logging
+from typing import TYPE_CHECKING
 
 from flask import Flask
 
@@ -22,6 +23,9 @@ from rucio.common.config import config_get
 from rucio.common.exception import ConfigurationError
 from rucio.common.logging import setup_logging
 from rucio.web.rest.flaskapi.v1.common import CORSMiddleware
+
+if TYPE_CHECKING:
+    from collections.abc import Iterable
 
 DEFAULT_ENDPOINTS = [
     'accountlimits',
@@ -49,7 +53,7 @@ DEFAULT_ENDPOINTS = [
 ]
 
 
-def apply_endpoints(app, modules):
+def apply_endpoints(app: Flask, modules: "Iterable[str]") -> None:
     for blueprint_module in modules:
         # Legacy patch - TODO Remove in 38.0.0
         if blueprint_module == "meta":
--- a/lib/rucio/web/rest/flaskapi/v1/main.py
+++ b/lib/rucio/web/rest/flaskapi/v1/main.py
@@ -76,9 +76,17 @@ def apply_endpoints(app: Flask, modules: "Iterable[str]") -> None:
             raise ConfigurationError(f'"{blueprint_module}" from the endpoints configuration value did not have a blueprint')
 
 endpoints = set(config_get_list('api', 'endpoints', raise_exception=False, default=[]))
+endpoints_add = set(config_get_list('api', 'endpoints_add', raise_exception=False, default=[]))
+endpoints_remove = set(config_get_list('api', 'endpoints_remove', raise_exception=False, default=[]))
+
+if endpoints and (endpoints_add or endpoints_remove):
+    raise ConfigurationError("Endpoints cannot be set in both 'endpoints' and 'endpoints_add'/'endpoints_remove'")
+
+if endpoints_add.intersection(endpoints_remove):
+    raise ConfigurationError("Endpoints cannot be in both 'endpoints_add' and 'endpoints_remove'")
 
 if not endpoints:
-    endpoints = DEFAULT_ENDPOINTS
+    endpoints = DEFAULT_ENDPOINTS - endpoints_remove | endpoints_add
 
 application = Flask(__name__)
 application.wsgi_app = CORSMiddleware(application.wsgi_app)
--- a/lib/rucio/web/rest/flaskapi/v1/main.py
+++ b/lib/rucio/web/rest/flaskapi/v1/main.py
@@ -19,7 +19,7 @@ from typing import TYPE_CHECKING
 
 from flask import Flask
 
-from rucio.common.config import config_get
+from rucio.common.config import config_get_list
 from rucio.common.exception import ConfigurationError
 from rucio.common.logging import setup_logging
 from rucio.web.rest.flaskapi.v1.common import CORSMiddleware
@@ -27,7 +27,7 @@ from rucio.web.rest.flaskapi.v1.common import CORSMiddleware
 if TYPE_CHECKING:
     from collections.abc import Iterable
 
-DEFAULT_ENDPOINTS = [
+DEFAULT_ENDPOINTS = {
     'accountlimits',
     'accounts',
     'auth',
@@ -50,7 +50,7 @@ DEFAULT_ENDPOINTS = [
     'rules',
     'scopes',
     'subscriptions',
-]
+}
 
 
 def apply_endpoints(app: Flask, modules: "Iterable[str]") -> None:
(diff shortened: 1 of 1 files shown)


lib/rucio/web/rest/flaskapi/v1/main.py
```diff
@@ -81,6 +81,30 @@ def apply_endpoints(app, modules):
...
+        endpoints_add = config_get('api', 'endpoints_add', raise_exception=False, default='')
+        endpoints_add = list(filter(bool, map(str.strip, endpoints_add.split(sep=','))))
+    except RuntimeError:
+        endpoints_add = []
+
+    try:
+        endpoints_remove = config_get('api', 'endpoints_remove', raise_exception=False, default='')
+        endpoints_remove = list(filter(bool, map(str.strip, endpoints_remove.split(sep=','))))
+    except RuntimeError:
+        endpoints_remove = []
```
dchristidis: Please replace with:

```python
endpoints_add = config_get_list('api', 'endpoints_add', raise_exception=False, default=[])
endpoints_remove = config_get_list('api', 'endpoints_remove', raise_exception=False, default=[])
```

I have a commit sitting in a branch which replaces the existing code, which is not a good example to be replicated.

lib/rucio/web/rest/flaskapi/v1/main.py
```diff
@@ -72,15 +72,26 @@ def apply_endpoints(app, modules):
...
 
-try:
-    endpoints = config_get('api', 'endpoints', raise_exception=False, default='')
-    endpoints = list(filter(bool, map(str.strip, endpoints.split(sep=','))))
-except RuntimeError:
-    endpoints = None
+endpoints = config_get_list('api', 'endpoints', raise_exception=False, default=[])
 
 if not endpoints:
     endpoints = DEFAULT_ENDPOINTS
```
dchristidis: If you a keen on changing the original, then set `default=DEFAULT_ENDPOINTS`.

lib/rucio/web/rest/flaskapi/v1/main.py
```diff
@@ -72,15 +72,26 @@ def apply_endpoints(app, modules):
...
+    if set(endpoints_add).intersection(set(endpoints_remove)):
+        raise ConfigurationError("Endpoints cannot be in both 'endpoints_add' and 'endpoints_remove'")
+
+    for endpoint in endpoints_remove:
+        if endpoint in endpoints:
+            endpoints.remove(endpoint)
+
+    for endpoint in endpoints_add:
+        if endpoint not in endpoints:
+            endpoints.append(endpoint)
```
dchristidis: How about:

```python
endpoints = list(set(endpoints) - set(endpoints_remove) | set(endpoints_add))
```

lib/rucio/web/rest/flaskapi/v1/main.py
dchristidis: In addition, please include the following in the commit message:

```
    This has two subtle changes to the behaviour:
    
     1. If the `api.endpoints` configuration is set to an empty value, then
        no endpoint will be enabled. The previous behaviour was to enable
        the defaults.
     2. If the value is not a proper comma-separated sequence of non-empty
        values (e.g. there are two consecutive commas), then an exception
        will be raised while trying to import the offending module. The
        previous behaviour was to silently ignore them.
```

lib/rucio/web/rest/flaskapi/v1/main.py
```diff
@@ -72,14 +72,17 @@ def apply_endpoints(app, modules):
...
 
 
-try:
-    endpoints = config_get('api', 'endpoints', raise_exception=False, default='')
-    endpoints = list(filter(bool, map(str.strip, endpoints.split(sep=','))))
-except RuntimeError:
-    endpoints = None
+endpoints = config_get_list('api', 'endpoints', raise_exception=False, default=DEFAULT_ENDPOINTS)
 
 if not endpoints:
```
dchristidis: I think there may be a clash between what I had in my mind and what you’re trying to achieve.

I think this conditional no longer makes sense. Suppose the following configuration:
```ini
[api]
endpoints_add = foo
```
It will be rendered ineffective. So when I suggested replacing `default=[]` with `default=DEFAULT_ENDPOINTS`, I expected this bit to have been removed.

The only question is how should a configuration like:
```ini
[api]
endpoints = foo,bar
endpoints_add = baz
```
be interpreted. In my opinion, I don’t see a need to actively prevent it. Resulting with `endpoints = ['foo', 'bar', 'baz']` seems acceptable.

Did you intend to have `api.endpoints_add` and `api.endpoints_removed` be silently ignored when `api.endpoints` was set?

lobis: Yes, the original idea was to silently ignore them when `endpoints` is explicitly set after some discussion with @bari12 , maybe he can comment here. I don't mind that they are not ignored after all.

I added an exception when the same point is used in `add` and `remove` to prevent difficult to predict behaviour (will it be added or removed first? etc.). Do you agree with this?

dchristidis: I don’t have any say in this, nor a strong opinion. But if that was your intent, then please ignore my previous comment about setting `default=DEFAULT_ENDPOINTS`.

lobis: In any case I think silently ignoring them is a bad idea, I will throw saying that you either use `endpoints` explicitly or add to defaults.

lib/rucio/web/rest/flaskapi/v1/main.py
```diff
@@ -72,15 +72,22 @@ def apply_endpoints(app, modules):
...
 
 
-try:
-    endpoints = config_get('api', 'endpoints', raise_exception=False, default='')
-    endpoints = list(filter(bool, map(str.strip, endpoints.split(sep=','))))
-except RuntimeError:
-    endpoints = None
+endpoints = config_get_list('api', 'endpoints', raise_exception=False, default=[])
+endpoints_add = config_get_list('api', 'endpoints_add', raise_exception=False, default=[])
+endpoints_remove = config_get_list('api', 'endpoints_remove', raise_exception=False, default=[])
```
dchristidis: Minor suggestion that you can choose to ignore: use `set()` here to avoid converting the lists twice later.

What I don’t like about it is that you will end up with similarly-named variables of different types.

lib/rucio/web/rest/flaskapi/v1/main.py
```diff
@@ -72,15 +72,22 @@ def apply_endpoints(app, modules):
...
+
+if endpoints and (endpoints_add or endpoints_remove):
+    raise ConfigurationError("Endpoints cannot be set in both 'endpoints' and 'endpoints_add'/'endpoints_remove'")
 
 if not endpoints:
     endpoints = DEFAULT_ENDPOINTS
```
dchristidis: I would recommend merging this with the assignment at the end of the block.

lobis:
Looks like tests are failing at https://github.com/rucio/rucio/blob/e0d7ae217ea7e3400be73468b44f3e067637e607/tests/test_config.py#L137

Not sure why it fails yet, but this test is marked for removal from a commit from 2021, can we remove it?

lib/rucio/web/rest/flaskapi/v1/main.py
dchristidis: Forgive me, I’m afraid I may have not made myself clear. I meant to use `set(config_get_list(…))`.

I approve of turning everything into a set, but careful that once again you’re incorporating too many changes of different nature in a single commit.

lobis: I agree that the scope of the PR changed, so I split it into multiple commits, hopefully it's clear enough.

dchristidis: The end result is good, but the first commit, in isolation, doesn’t seem right.

lib/rucio/web/rest/flaskapi/v1/main.py
```diff
@@ -46,10 +50,10 @@
     'rules',
     'scopes',
     'subscriptions',
-]
+}
 
 
-def apply_endpoints(app, modules):
+def apply_endpoints(app: Flask, modules: "Iterable[str]"):
```
dchristidis: Specify `-> None` to complete the type hints.

lib/rucio/web/rest/flaskapi/v1/main.py
lobis: I think all commits should make sense now individually, if not I think I'd rather merge them into a single commit.

dchristidis (approved):
Good work.

PR #7646, opened 2025-04-09
Database: Handle json serialization in `models.py`

### LINKED ISSUE #7642: Improve database JSON usage
Currently on the oracle and sqlite version we support there is not a native json type, we work around this by serializing into strings and deserializing.

This code could be moved to `types.py` to make the `JSON` class automatically serialize/deserialize into strings if a dialect with no native json support is used.

This would simplify code and reduce repetition.

lobis:
Closes #7642


Geogouz:
Yeah, various tests fail. Gotta see why. If you want I can help.

Btw, I guess you mean `types.py` instead of `models.py`.

lobis:
I close this PR for being stale but the underlying issue remains.

PR #7654, opened 2025-04-10
Documentation: UploadClient docstring enhancement; Fix rucio#7634

### LINKED ISSUE #7634: Rucio Python Client docstrings enhancement (`UploadClient`)
Sub-issue of https://github.com/rucio/documentation/issues/363 targeting specifically the `UploadClient`.

Geogouz:
Handles the sub-issue about the UploadClient Docstrings

### DIFF
--- a/lib/rucio/client/uploadclient.py
+++ b/lib/rucio/client/uploadclient.py
@@ -57,18 +57,32 @@ if TYPE_CHECKING:
 
 
 class UploadClient:
-
     def __init__(
-            self,
-            _client: Optional[Client] = None,
-            logger: Optional["LoggerFunction"] = None,
-            tracing: bool = True
+        self,
+        _client: Optional[Client] = None,
+        logger: Optional["LoggerFunction"] = None,
+        tracing: bool = True
     ):
         """
-        Initialises the basic settings for an UploadClient object
+        Initialize the UploadClient with the necessary configuration to manage file uploads.
 
-        :param _client:     - Optional: rucio.client.client.Client object. If None, a new object will be created.
-        :param logger:      - Optional: logging.Logger object. If None, default logger will be used.
+        This method is used to create a new UploadClient instance that can upload files. It
+        allows the use of an existing Rucio Client, a custom logger, and tracing for debug
+        information during the upload process.
+
+        Parameters
+        ----------
+        _client
+            An existing Rucio `Client` instance to reuse. If not provided, a new one is created.
+        logger
+            A logger function. If not provided, the default Python logger is used.
+        tracing
+            Indicates whether to enable tracing to capture upload activity details.
+
+        Raises
+        ------
+        InputValidationError
+            If the client account is not found or is invalid, preventing upload setup.
         """
         if not logger:
             self.logger = logging.log
@@ -78,7 +92,9 @@ class UploadClient:
         self.client: Final[Client] = _client if _client else Client()
         self.client_location = detect_client_location()
         # if token should be used, use only JWT tokens
-        self.auth_token: Optional[str] = self.client.auth_token if len(self.client.auth_token.split(".")) == 3 else None
+        self.auth_token: Optional[str] = (
+            self.client.auth_token if len(self.client.auth_token.split(".")) == 3 else None
+        )
         self.tracing = tracing
         if not self.tracing:
             logger(logging.DEBUG, 'Tracing is turned off.')
@@ -87,10 +103,14 @@ class UploadClient:
             try:
                 acc = self.client.whoami()
                 if acc is None:
-                    raise InputValidationError('account not specified and rucio has no account with your identity')
+                    raise InputValidationError(
+                        'Account not specified and rucio has no account with your identity'
+                    )
                 self.client.account = acc['account']
             except RucioException as e:
-                raise InputValidationError('account not specified and problem with rucio: %s' % e)
+                raise InputValidationError(
+                    f'Account not specified and problem with rucio: {e}'
+                )
             self.logger(logging.DEBUG, 'Discovered account as "%s"' % self.client.account)
         self.default_file_scope: Final[str] = 'user.' + self.client.account
         self.rses = {}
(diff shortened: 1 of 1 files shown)


voetberg:
It is worth noting the private methods here do [not get their docstrings rendered](https://rucio.cern.ch/documentation/html/client_api/uploadclient.html) - I would not be opposed to taking a lighter hand on documenting those if it cuts down on the first pass work load

Geogouz:
Yes, the OCD kicked in there and thought of doing everything to maybe help devs :/ .
UploadClient and DownloadClient are the two biggest modules so my hopes are that after finishing those, the workload will drop significantly. :D

Geogouz:
I'll add an extra commit to fix some type check problems.

lib/rucio/client/uploadclient.py
```diff
@@ -913,13 +1250,19 @@ def preferred_impl(
...
+
+        This method checks the local client configuration (under the `[upload] preferred_impl`
+        setting) and compares it against the list of protocols declared in `rse_settings`.
+        It attempts to find a protocol that supports the required I/O operations (read,
+        write, delete) in the specified domain. If multiple preferred protocols are listed
+        in the config, it iterates in order and returns the first viable match.
 
-            :param rse_settings: dictionary containing the RSE settings
-            :param domain:     The network domain, either 'wan' (default) or 'lan'
+        :param rse_settings: A dictionary describing RSE details, including available protocols and their domains.
```
voetberg: Apparently `preferred_impl` doesn't work https://github.com/rucio/rucio/blob/0d44febcfd5d0a773a24d60668177324c534bd18/tests/test_upload.py#L289  (though I cannot find an issue for this)

I was going to recommend giving an example for `rse_settings` but there doesn't seem to be a test with examples, so I think it's more trouble than its worth at this point

voetberg (changes requested):
[Uploadclient - Rucio Python Client Documentation.pdf](https://github.com/user-attachments/files/19687517/Uploadclient.-.Rucio.Python.Client.Documentation.pdf)
Here is the PDF of the rendered doc page

I think the content of this is all pretty great! No criticism there. 

The `items` description however is really difficult to read. If we can figure out a way to get indenting to work properly there, I'd be very happy. Visual IDEs also don't seem to know what to do them either.  

Screenshot-vscode-rendered

If possible, pulling a few examples (probably just annotating things from `tests/test_upload.py`) would be greatly beneficial. I need to play around the mkdocs settings to make sure they render properly but I don't think it should be difficult.

Geogouz:
Yes, I think finding a way how to render things with more structure would be a game changer here. There are restrictions or limitations involved currently, but I will give it also a check to see what we can do about it.

lib/rucio/client/uploadclient.py
Geogouz: Yep, noticed that too :D. But went with the flow basically and tried to describe what that isolated code tries to do (honouring the effort of the writer lol).

voetberg:
I reformatted using `numpy` style docstrings and got this result (quite a bit more readable, even if the `items` section is longer, in my opinion 🧅 )

[Uploadclient - Rucio Python Client Documentation.pdf](https://github.com/user-attachments/files/19852070/Uploadclient.-.Rucio.Python.Client.Documentation.pdf)

I would be okay going through and doing a baseline conversion to this style, at least for the client docstrings.

[Here is a gist of the changes I made](https://gist.github.com/voetberg/d108045ba93fa3f6bae623156f6c75dd)

cc: @bari12

Geogouz:
> I reformatted using `numpy` style docstrings and got this result (quite a bit more readable, even if the `items` section is longer, in my opinion 🧅 )
> 
> [Uploadclient - Rucio Python Client Documentation.pdf](https://github.com/user-attachments/files/19852070/Uploadclient.-.Rucio.Python.Client.Documentation.pdf)
> 
> I would be okay going through and doing a baseline conversion to this style, at least for the client docstrings.
> 
> [Here is a gist of the changes I made](https://gist.github.com/voetberg/d108045ba93fa3f6bae623156f6c75dd)
> 
> cc: @bari12

Looks much better indeed. I will follow that styling for these docs.

voetberg:
> Looks much better indeed. I will follow that styling for these docs.

This does require a change on the doc generation side, so we'd have to apply this style to all of the client docstrings. It wouldn't be hard but I want to make sure this is an okay change to make.

Geogouz:
Ok, let's finalise and merge first #7688 (to convert current state to _numpy_ style), and I can go on with the updated versions.

voetberg:
Hi @Geogouz looks like the blocker has been merged, have you gotten a chance to update this PR?

Geogouz:
Need to add couple examples and fix the autotests problem. Do you suggest that I place here (and for each client's-module corresponding PR) a screenshot of how the rendered result looks like? To not have to built from scratch everytime in order to review it?

Geogouz:
To avoid test problems, I included here the required #7762 (the 3 first commits) to not have to wait for its merging. Once both PRs are merged, they should collapse. So you can pay attention only to the last commit since the other ones are being reviewed by another PR.

Geogouz:
Should be ready @voetberg. Looks like this rendered:
[Uploadclient - Rucio Python Client Documentation.pdf](https://github.com/user-attachments/files/20677374/Uploadclient.-.Rucio.Python.Client.Documentation.pdf)

voetberg (dismissed):
Looks great! As long as @rdimaio is fine with the typing I give this my full approval! Thanks so much for you hard work on this

lib/rucio/client/uploadclient.py
```diff
@@ -388,8 +622,24 @@ def _pick_random_rse(rse_expression: str) -> dict[str, Any]:
...
-    def _add_bittorrent_meta(self, file: "Mapping[str, Any]") -> None:
-        pieces_root, pieces_layers, piece_length = bittorrent_v2_merkle_sha256(os.path.join(file['dirname'], file['basename']))
+    def _add_bittorrent_meta(
+            self,
+            file: "Mapping[str, Any]"
+    ) -> None:
+        """
+        Add Bittorrent v2 metadata to the file DID.
+
+        This method calculates the Bittorrent v2 pieces root, layers, and piece length for
```
rdimaio: nit: the correct capitalization is `BitTorrent`

lib/rucio/client/uploadclient.py
rdimaio, line 1: Leaving comment to block to merge and rebase https://github.com/rucio/rucio/pull/7765 first

rdimaio (dismissed):

rdimaio (approved):

Issue #7653, opened 2025-04-10
`list_new_dids`: `isinstance(did_type, Enum)` should be `isinstance(did_type, DIDType)`


rdimaio:
https://github.com/rucio/rucio/blob/e0d7ae217ea7e3400be73468b44f3e067637e607/lib/rucio/core/did.py#L1323

Issue #7447, opened 2025-02-27
Fix type annotations revealed during session decorator -> context manager migration


rdimaio:
### Description

Related: https://github.com/rucio/rucio/issues/6986

### Steps to reproduce

Remove `_session` decorators

rdimaio:
Next step: continue migration in https://github.com/rucio/rucio/issues/8198

PR #7659, opened 2025-04-10
Database: gateway/did, migrate to session context manager

rdimaio:
Part of:
- #7409
- #7605
- #7447
- #7653

### DIFF
--- a/lib/rucio/gateway/did.py
+++ b/lib/rucio/gateway/did.py
@@ -33,7 +33,7 @@ if TYPE_CHECKING:
 
 def list_dids(
     scope: str,
-    filters: 'Mapping[Any, Any]',
+    filters: 'Iterable[dict[Any, Any]]',
     did_type: str = 'collection',
     ignore_case: bool = False,
     limit: Optional[int] = None,
--- a/lib/rucio/gateway/did.py
+++ b/lib/rucio/gateway/did.py
@@ -24,16 +24,13 @@ from rucio.common.utils import gateway_update_return_dict
 from rucio.core import did, naming_convention
 from rucio.core import meta_conventions as meta_convention_core
 from rucio.core.rse import get_rse_id
-from rucio.db.sqla.constants import DIDType
-from rucio.db.sqla.session import read_session, stream_session, transactional_session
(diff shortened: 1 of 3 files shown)


lib/rucio/gateway/did.py
```diff
@@ -24,28 +24,23 @@
...
     from collections.abc import Iterable, Iterator, Mapping, Sequence
 
-    from sqlalchemy.orm import Session
 
-
-@stream_session
 def list_dids(
     scope: str,
-    filters: 'Mapping[Any, Any]',
+    filters: 'Iterable[dict[Any, Any]]',
```
0xquark, line 36: Suggestion/Viewpoint?
I know using iterables is a standard in our Rucio codebase, but is there a reason we don't mention or explain this in the docstrings? For example, Filters used to take a single dictionary of key-value pairs, but now it accepts a list (or any iterable) of dictionaries, with each being a separate filter group. We include information about the parameter in the docstrings but not about the type of parameter it accepts. I just want to understand the reasoning behind this.

0xquark (approved):

lib/rucio/gateway/did.py
rdimaio: If a function is type annotated, adding the same type annotation in the docstring is redundant information that could eventually end up being out of sync (e.g. if you have a parameter that is annotated as `int` and is said to be an int in the docstring, someone could change the annotation to `str` but forget to update the docstring).

In this case, the reason why the type annotation changed is that the `@stream_session` decorator was hiding the function signature from static code analysis (https://github.com/rucio/rucio/issues/7447, https://github.com/rucio/rucio/issues/6986), so the type checker never complained about the occurrence where it is used and a list is passed to it. The only usage of `gateway.did.list_dids` is here, and you can see that it should always pass a list:

https://github.com/rucio/rucio/blob/0d44febcfd5d0a773a24d60668177324c534bd18/lib/rucio/web/rest/flaskapi/v1/dids.py#L260-L269

0xquark: Got it! Thanks for breaking it down so clearly.

Issue #7463, opened 2025-03-06
Remove `match_define_requirements`


rdimaio:
This function is not necessary. `install_requires` and the requirements file are separate by design.

This doc explains the difference between install_requires and requirements files: https://packaging.python.org/en/latest/discussions/install-requires-vs-requirements/

1. in `install_requires` we list the unpinned, minimal requirements needed
2. in the requirements files we list all the primary and secondary dependencies, pinned (except for client which stays unpinned)
Since we already list both `install_requires` and `extras_require` at the top of `setuputil`, `match_define_requirements` does not really serve any purpose.

On top of not serving any purpose, it caused the following issues:
1. `dogpile.cache`-related bug in `35.0.0rc1`: https://github.com/rucio/rucio/issues/6969
2. `psycopg`-related bug in `36.3.0`: https://github.com/rucio/rucio/issues/7462

PR #7672, opened 2025-04-15
Client: unpin dogpile.cache (it remains pinned in server)

rdimaio:
See: https://github.com/conda-forge/rucio-clients-feedstock/issues/114

### DIFF
--- a/setuputil.py
+++ b/setuputil.py
@@ -20,7 +20,7 @@ clients_requirements_table = {
         'click',
         'requests',
         'urllib3',
-        'dogpile-cache<=1.2.2',
+        'dogpile-cache',
         'packaging',
         'tabulate',
         'jsonschema',


nsoranzo (approved):

maxnoe:
There seems to be a deeper bug here in your setuputils.py
Somewhere the requirement is inverted.

You have in `requirements.clients.txt`:

```
❯ grep dogpile requirements/requirements.client.txt
dogpile-cache>=1.2.2  
```
but in setuputil.py you have:
```
clients_requirements_table = {
    'install_requires': [
        'click',
        'requests',
        'urllib3',
        'dogpile-cache<=1.2.2',
```

maxnoe:
Also, why do you have all the version pins in `requirements/` if you then publish with completely unpinned versions via that `setuputil.py`? That is very surprising.

rdimaio:
> Also, why do you have all the version pins in `requirements/` if you then publish with completely unpinned versions via that `setuputil.py`? That is very surprising.

This is in line with the Python packaging philosophy of providing loose requirements (that still work) in `install_requires`, and a reproducible environment in the requirements files.

See https://packaging.python.org/en/latest/discussions/install-requires-vs-requirements/ for reference:

> Whereas install_requires defines the dependencies for a single project, [Requirements Files](https://pip.pypa.io/en/latest/user_guide/#requirements-files) are often used to define the requirements for a complete Python environment.
> 
> Whereas install_requires requirements are minimal, requirements files often contain an exhaustive listing of pinned versions for the purpose of achieving [repeatable installations](https://pip.pypa.io/en/latest/topics/repeatable-installs/#repeatability) of a complete environment.
> 
> Whereas install_requires requirements are “Abstract”, i.e. not associated with any particular index, requirements files often contain pip options like --index-url or --find-links to make requirements “Concrete”, i.e. associated with a particular index or directory of packages. [[1]](https://packaging.python.org/en/latest/discussions/install-requires-vs-requirements/#id4)
> 
> Whereas install_requires metadata is automatically analyzed by pip during an install, requirements files are not, and only are used when a user specifically installs them using python -m pip install -r.

maxnoe:
> This is in line with the Python packaging philosophy of providing loose requirements (that still work) in install_requires, and a reproducible environment in the requirements files.

But that's not what is happening here.

You have loose requirements and looser requirements. You only have the reproducible requirements for the server.

rdimaio:
> There seems to be a deeper bug here in your setuputils.py Somewhere the requirement is inverted.
> 
> You have in `requirements.clients.txt`:
> 
> ```
> ❯ grep dogpile requirements/requirements.client.txt
> dogpile-cache>=1.2.2  
> ```
> 
> but in setuputil.py you have:
> 
> ```
> clients_requirements_table = {
>     'install_requires': [
>         'click',
>         'requests',
>         'urllib3',
>         'dogpile-cache<=1.2.2',
> ```

This was done on purpose - `requirements/requirements.client.txt` requirements should only have a lower bound defined, not upper bound. 

However, upgrading `dogpile.cache` caused this server-side issue (in a daemon) https://github.com/rucio/rucio/issues/7571, and an upper bound was added in setuputil to resolve this for the 37.0.0 release. This was because the server requirements were previously built from the client requirements, see: https://github.com/rucio/rucio/pull/7573/files#diff-036d634440b9cc9646a72ad3cc4c98998e6a2a744cdae4196b20accb0f6be07dR64

This has now been changed so that the server requirements are pinned as expected, so we don't need this upper bound for the client dependency anymore.

maxnoe:
Anyway, since this is all aout to go anyway due to https://github.com/rucio/rucio/pull/7675, not really worth discussing.

rdimaio:
> > This is in line with the Python packaging philosophy of providing loose requirements (that still work) in install_requires, and a reproducible environment in the requirements files.
> 
> But that's not what is happening here.
> 
> You have loose requirements and looser requirements. You only have the reproducible requirements for the server.

I see your point, in theory we should exactly pin exact versions for the client reqs file (for a reproducible install). This wasn't the case before because we used to have this `match_define_requirements`-based logic which would actually cause `pip install` to install from the requirement file: https://github.com/rucio/rucio/issues/7463. I'll open an issue for this, since this isn't actually going to go away after we move to `pyproject.toml` - we'd likely still want to provide reproducible environments

maxnoe:
> since this isn't actually going to go away after we move to pyproject.toml - we'd likely still want to provide reproducible environments

I think this could be done now using uv, which integrates well with `pyproject.toml`. So you can just run `uv lock` on the pyproject and commit the resulting lock file.

nsoranzo (approved):
It would be nice if this could be merged, the new rucio-clients forces us to downgrade dogpile-cache in https://github.com/galaxyproject/galaxy/pull/20139/

bari12 (approved):

Issue #7322, opened 2025-01-16
Make it possible to have editable installs of rucio and rucio-clients at the same time


maxnoe:
### Description

Currently, this repository contains two python distribution packages, that are however using the same `pyproject.toml` and one has to change a symlink to steer which distribution is actually created.

This makes it essentially impossible to simply create editable installs for both rucio and rucio-clients.

### Motivation

Development setup, e.g. for the rucio-dirac integration need changing both client and server code

### Change

Actually split the distribution packages.

Maybe using namespace packages for common, server and client specific code respectively.

rdimaio:
I agree. This is related to this: https://github.com/rucio/rucio/issues/6971

I think we might need to restructure the folders a bit to allow for this - e.g. right now we don't have a specific `server` folder, which we would need to create in order to package it separately.

I like the way DiracX splits the packages into different namespace packages with separate `pyproject.toml` files, I think we could use this as reference: https://github.com/DIRACGrid/diracx

bari12:
Let's follow this up in #6971 
I'll close this ticket in the meanwhile.

maxnoe:
The pull request #7675 does only address #6971 as in it replaces the several `setup_xy.py` with `pyproject.xy.toml`.

It still requires the "symlink" dance, as the different pyproject do not have the standard name.

bziemons:
What is the current plan here? The move the `pyproject.toml` is okay, the `pyproject.server.toml` contains all binaries (why?), but that makes it a possible candidate for a generic `pyproject.toml` that would install server/clients/daemons. I would really like to make use of editable installs in an upcoming version of the dev container and I would like to hear opinions and plans. https://github.com/rucio/rucio/issues/6971 does not exactly tell me much about why certain things are like they are now.

maxnoe:
I think the best solution to enable this issue is by implementing https://github.com/rucio/rucio/issues/7330, splitting rucio into at least three namespaced packages `rucio-common`, `rucio-server` and `rucio-client`.

These can then all be installed at the same time in development mode. This requires re-organizing the repository structure, so that you have the three packages in subdirectories I think:

```
rucio-common/
   pyproject.toml
   lib/rucio/
      common
      utils
      ...
   
rucio-server/
   pyproject.toml
   lib/rucio/
     core/
     ...

rucio-client/
   pyproject.toml
   lib/rucio/
      client/
      ...
```

bziemons:
[...]

PR #7675, opened 2025-04-16
Deprecate setup.py and move to pyproject.toml

### LINKED ISSUE #6971: Dependencies: deprecate `setup.py` and move to `pyproject.toml`

### LINKED ISSUE #7677: Remove `client_syntax` checks and `ignoretool` usage
I think this is a leftover from the past, before we had `ruff`. It does exactly the same thing as the `python_ruff` CI check, but it only runs it for a subset of the files.

In the spirit of simplifying the test architecture (#7667), I think we should remove this.

rdimaio:
fixes https://github.com/rucio/rucio/issues/6971


pyproject.client.toml
```diff
@@ -0,0 +1,135 @@
...
+
+dependencies = [
+        'click',
+        'requests',
+        'urllib3',
+        'dogpile-cache<=1.2.2',
+        'packaging',
+        'tabulate',
+        'jsonschema',
+        'dataclasses',
```
nsoranzo:
```suggestion
```

As in https://github.com/rucio/rucio/pull/7665 .

pyproject.client.toml
```diff
@@ -0,0 +1,135 @@
...
+name = "rucio-clients"
+dynamic = ["version"]
+
+# For using SSO login option, install these RPM packages: libxml2-devel xmlsec1-devel xmlsec1-openssl-devel libtool-ltdl-devel
+
+dependencies = [
+        'click',
+        'requests',
+        'urllib3',
+        'dogpile-cache<=1.2.2',
```
nsoranzo:
```suggestion
        'dogpile-cache',
```

As in https://github.com/rucio/rucio/pull/7672 .

pyproject.server.toml
```diff
@@ -0,0 +1,167 @@
...
+dynamic = ["version"]
+
+dependencies = [
+  'requests<=2.32.3',
+  'urllib3<=1.26.19',
+  'dogpile-cache<=1.2.2',
+  'packaging<=24.1',
+  'tabulate<=0.9.0',
+  'jsonschema<=4.23.0',
+  'dataclasses',  # no upper limit is set in .in or .txt req files
```
nsoranzo:
```suggestion
```

Part of the standard library since Python 3.7 .

pyproject.client.toml
rdimaio: Addressed in latest revision

pyproject.client.toml
rdimaio: Addressed in latest revision

pyproject.server.toml
rdimaio: Addressed in latest revision

maxnoe:
Let me note that while this is a big improvement, it does not fix #7322, which was originally closed as a duplicate of #6971 

So I re-opened that for now.

rdimaio:
> Let me note that while this is a big improvement, it does not fix #7322, which was originally closed as a duplicate of #6971
> 
> So I re-opened that for now.

I agree - I discussed with Martin that we would initially do this change, then possibly a namespace package restructure (https://github.com/rucio/rucio/pull/7577) and then the editable installs would be possible. This is somewhat discussed https://github.com/rucio/rucio/issues/7330 and in theory by resolving https://github.com/rucio/rucio/issues/7330 we'd resolve #7322 too, but it doesn't hurt to have that one open too.

bari12:
Please rebase

pyproject.client.toml
```diff
@@ -0,0 +1,134 @@
...
+        'dogpile-cache',
+        'packaging',
+        'tabulate',
+        'jsonschema',
+        'rich',
+        'typing_extensions'
+]
+requires-python = ">=3.9"
+authors = [
+  {name = "Rucio", email = "rucio-dev@cern.ch"},
```
bari12: Change to `rucio-contact` instead (across all files)

rdimaio:
> Please rebase

Rebased in latest revision to fix merge conflicts, will now address your comment

pyproject.client.toml
rdimaio: Addressed in latest commit

bari12 (dismissed):

pyproject.server.toml
```diff
@@ -0,0 +1,172 @@
...
+
+[tool.setuptools.package-dir]
+"" = "lib"
+rucio = "lib/rucio"
+bin = "bin"
+etc = "etc"
+tools = "tools"
+requirements = "requirements"
+
+[tool.setuptools.packages.find]
```
maxnoe, line 167: I think this section must also be added to the other two files.

rdimaio: Addressed in latest revision

tools/generate_version.py
```diff
@@ -45,13 +45,11 @@ def run_git_command(cmd):
...
-    'final': %s,
-    'version': '%s',
-    'branch_nick': '%s',
-    'revision_id': '%s',
-    'revno': %s
-}
+IS_FINAL = %s
+VERSION = %s
+BRANCH_NICK = %s
+REVISION_ID = %s
```
Geogouz: Aren't these written as bare strings (e.g., VERSION = 38.0.0)? If so, since this is not valid Python syntax, I guess any fresh release build will fail to import `rucio.vcsversion` (the new `pyproject` files declare a dynamic version `attr = "rucio.vcsversion.VERSION"`).

pyproject.server.toml
```diff
@@ -0,0 +1,172 @@
...
+
+[build-system]
+requires = ["setuptools>=78.1.0", "wheel>=0.45.1", "packaging"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "rucio"
+dynamic = ["version"]
+
+dependencies = [
```
Geogouz, line 23: Is having here only upper bounds safe?

maxnoe: This is quite a big change and weird to do.

The current requirements has == requirements:

https://github.com/rucio/rucio/blob/master/requirements/requirements.server.in

rdimaio: I would say so. From [the packaging docks](https://packaging.python.org/en/latest/discussions/install-requires-vs-requirements/#requirements-files), the requirements files are meant to provide a reproducible environment, while the dependencies specified here should be more minimal and not as 'hard-pinned'. We could be more precise and add lower boundaries here, but how can we exactly test which lower boundaries to set for each dependency? I think it would be added complexity for not much gain, particularly since this is mostly deployed as part of a Docker image where `rucio` is likely the main Python package installed, so it can likely get the latest dependencies without conflicts

tools/generate_version.py
rdimaio: Good catch, I ran it now with `python tools/generate_version.py 39.0.0` and indeed it'd generate `VERSION = 39.0.0`, which then is seen as a `SyntaxError` at build time. 

Addressed in commit https://github.com/rucio/rucio/pull/7675/commits/b03c8b478d22c58e5042be2245dea051b97e3657, now running `python tools/generate_version.py 39.0.0` results in the following `vcsversion`:

```
'''
This file is automatically generated; Do not edit it. :)
'''
IS_FINAL = True
VERSION = '39.0.0'
BRANCH_NICK = 'patch-6971-pyproject'
REVISION_ID = '7353b74ddb1d43c8e7fe545c5104c61a624d85be'
REVNO = '14212'
```

then running `./tools/build_sdist_wheel.sh all`, packages are built as expected, and looking at the generated `PKG-INFO` the version is parsed as expected:

```
Metadata-Version: 2.4
Name: rucio
Version: 39.0.0
Summary: Rucio server package
Author-email: Rucio <rucio-contact@cern.ch>
Maintainer-email: Rucio <rucio-contact@cern.ch>
License-Expression: Apache-2.0
...
```

pyproject.server.toml
rdimaio: > This is quite a big change and weird to do.
> 
> The current requirements has == requirements:
> 
> https://github.com/rucio/rucio/blob/master/requirements/requirements.server.in

That's the requirements file, but the current `setuputil.py` uses upper boundaries in the same way:

https://github.com/rucio/rucio/blob/master/setuputil.py#L63-L113

Geogouz:
Shouldn't this PR make `pip install .` work though? I think I am missing the scope of this PR a bit.

rdimaio:
> Shouldn't this PR make `pip install .` work though? I think I am missing the scope of this PR a bit.

The idea is that the different packages should be installed separately https://packaging.python.org/en/latest/guides/packaging-namespace-packages/, and then you can point to those packages as dependencies in the top-level pyproject, like this https://github.com/DIRACGrid/diracx/blob/main/pyproject.toml

However, until the namespace restructure mentioned in https://github.com/rucio/rucio/issues/7330 is done, this cannot be done, so I think we need to replicate the dependencies from the server pyproject into the main pyproject for now as a temporary workaround

rdimaio:
> Shouldn't this PR make `pip install .` work though? I think I am missing the scope of this PR a bit.

Addressed in commit https://github.com/rucio/rucio/pull/7675/commits/42df7ba0465e87e2789b1eab7a13f1f014a03e57

Geogouz (approved):
Can't find something obvious. I am giving a green light by my side.

bari12 (approved):

PR #7676, opened 2025-04-16
Create trace gateaway

### LINKED ISSUE #7256: No gateway for traces
### Description

Trace REST endpoint goes directly to core w/o going through the gateway layer : 
https://github.com/rucio/rucio/blob/master/lib/rucio/web/rest/flaskapi/v1/traces.py#L24

### Steps to reproduce

See above

### Rucio Version

_No response_

### Additional Information

_No response_

voetberg:
Closes #7256

### DIFF
--- a/lib/rucio/gateway/trace.py
+++ b/lib/rucio/gateway/trace.py
@@ -1,0 +1,47 @@
+# Copyright European Organization for Nuclear Research (CERN) since 2012
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import datetime
+import json
+import uuid
+
+from rucio.core.trace import trace as core_trace
+
+
+def trace(request: str, trace_ip: str) -> None:
+    """
+    Sends the trace data to trace broker after adding additional fields
+    Adds the following fields:
+        - 'traceTimeentry': The current UTC timestamp.
+        - 'traceTimeentryUnix': The Unix timestamp with microsecond precision.
+        - 'traceIp': The client's IP address, either from 'X-Forwarded-For' header or remote address.
+        - 'traceId': A unique identifier for the trace, generated as a UUID without hyphens.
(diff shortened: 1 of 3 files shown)


lib/rucio/gateway/trace.py
```diff
@@ -0,0 +1,48 @@
...
+    if isinstance(request_data, dict):
+        request_data = [request_data]
+
+    for item in request_data:
+        item["traceIp"] = trace_ip
+        # generate entry timestamp
+        item["traceTimeentry"] = datetime.datetime.now(datetime.timezone.utc)
+        item["traceTimeentryUnix"] = item["traceTimeentry"].timestamp()
+        # generate unique ID
+        item["traceId"] = str(uuid.uuid4()).replace("-", "").lower()
```
rdimaio, line 46: `lower()` is unnecessary - `uuid4` creates UUIDs (https://docs.python.org/3/library/uuid.html) as specified in RFC 4122 (https://datatracker.ietf.org/doc/html/rfc4122.html), which says:

> The hexadecimal values "a" through "f" are output as
      lower case characters and are case insensitive on input.


But it doesn't hurt to keep it to be safe.

lib/rucio/gateway/trace.py
```diff
@@ -0,0 +1,48 @@
...
+    Sends the trace data to trace broker after adding additional fields
+    Adds the following fields:
+        - 'traceTimeentry': The current UTC timestamp.
+        - 'traceTimeentryUnix': The Unix timestamp with microsecond precision.
+        - 'traceIp': The client's IP address, either from 'X-Forwarded-For' header or remote address.
+        - 'traceId': A unique identifier for the trace, generated as a UUID without hyphens.
+
+    Args:
+        request_data: Request json given by client
+        trace_ip: TraceIP, either the client's IP address, of IP from "X-Forwarded-For" header
```
rdimaio:
```suggestion
        trace_ip: TraceIP, either the client's IP address, or IP from "X-Forwarded-For" header
```

tests/test_trace.py
```diff
@@ -129,3 +129,32 @@ def test_non_existant_event_type_validation_rejection():
...
+    """
+    TRACE (REST): submit a trace via POST with no data
+    """
+    response = rest_client.post('/traces/', data=None, content_type=[('Content-Type', 'application/octet-stream')])
+    assert response.status_code == 400
+
+
+def test_trace_sent_json_dump(rest_client):
+    """
+    TRACE (REST): submit a trace  identically to how it is sent in utils
```
rdimaio: nit: double whitespace
```suggestion
    TRACE (REST): submit a trace identically to how it is sent in utils
```

rdimaio (dismissed):

rdimaio (approved):

Issue #7260, opened 2024-12-13
Testing: move client commands used in tests to new client syntax


rdimaio:
This would remove the deprecation warnings seen in the test logs, e.g.: https://github.com/rdimaio/rucio/actions/runs/12317691933/job/34380752177

```
 WARNING: This method is being deprecated.
Please replace your command with `rucio rse attribute add`.
Added new RSE attribute for XRD3: available_for_multihop-True 
1h
1l
WARNING: This method is being deprecated.
Please replace your command with `rucio rse distance add`.
Set distance from XRD1 to XRD2 to 1
1h
1l
WARNING: This method is being deprecated.
Please replace your command with `rucio rse distance add`.
Set distance from XRD1 to XRD3 to 1
1h
1l
WARNING: This method is being deprecated.
Please replace your command with `rucio rse distance add`.
Set distance from XRD2 to XRD1 to 1
1h
...
```

Might be a good idea to wait a few months before doing this, as the client changes is fairly new and the command names might not yet be final

PR #7679, opened 2025-04-22
Clients: simplify adding and removing RSE distances

### LINKED ISSUE #7570: Simplify distance setting
### Description

The currrent (expert) procedure to remove a distance between to storages, due to no network connection for example, is
src=INFN-MILANO-ATLASC dst=SWT2_CPB
read -r -a srcs < <(rucio list-rses --rses "site=$src" || echo "Invalid site name: $src" >&2)
read -r -a dsts < <(rucio list-rses --rses "site=$dst" || echo "Invalid site name: $dst" >&2)
for i in "${srcs[@]}"; do for j in "${dsts[@]}"; do rucio-admin --account ddmadmin rse delete-distance "$i" "$j"; done; done

This would be repeated for the other direction.
Could we have this search/loop put inside the rucio-admin command such that distances to/from all RSEs on site A to/from all RSEs on site B are removed or set, e.g.
 rucio-admin rse delete-distance --all --bidi --sites INFN-MILANO-ATLASC SWT2_CPB
 

### Motivation

Reduce expert ops workload/cheat sheet, and ready for not so expert ops.

### Change

_No response_

0xquark:
Fixes #7570

### DIFF
--- a/lib/rucio/cli/bin_legacy/rucio_admin.py
+++ b/lib/rucio/cli/bin_legacy/rucio_admin.py
@@ -555,10 +555,16 @@ def add_distance_rses(args, client, logger, console, spinner):
     %(prog)s add-distance [options] SOURCE_RSE DEST_RSE
 
     Set the distance between two RSEs.
+    If --bidirectional is specified, the distance will be set in both directions.
     """
     params = {'distance': args.distance}
     client.add_distance(args.source, args.destination, params)
     print('Set distance from %s to %s to %d' % (args.source, args.destination, args.distance))
+    
+    if args.bidirectional:
+        client.add_distance(args.destination, args.source, params)
+        print('Set distance from %s to %s to %d' % (args.destination, args.source, args.distance))
+    
     return SUCCESS
 
 
@@ -607,12 +613,140 @@ def update_distance_rses(args, client, logger, console, spinner):
 @exception_handler
 def delete_distance_rses(args, client, logger, console, spinner):
     """
-    %(prog)s delete-distance [options] SOURCE_RSE DEST_RSE
+    %(prog)s delete-distance [options] [SOURCE] [DESTINATION]
+    %(prog)s delete-distance --src [SOURCE]
+    %(prog)s delete-distance --dest [DESTINATION]
+    %(prog)s delete-distance --bidirectional [SOURCE] [DESTINATION]
 
-    Update the existing distance entry between two RSEs.
+    Delete distance entries between RSEs. Supports multiple modes:
+    - Delete distance from one RSE to another
+    - Delete distances between all RSEs at one site to all RSEs at another site
+    - Delete all outgoing distances from an RSE/site (using --src)
+    - Delete all incoming distances to an RSE/site (using --dest)
+    - Delete distances in both directions (using --bidirectional)
+
+    Arguments support both RSE names and site names. If a site name is provided,
+    the command will operate on all RSEs belonging to that site.
     """
-    client.delete_distance(args.source, args.destination)
-    print('Deleted distance information from %s to %s.' % (args.source, args.destination))
+    # First assign src_flag and dest_flag to source and destination if provided
+    if hasattr(args, 'src_flag') and args.src_flag:
+        args.source = args.src_flag
+        
+    if hasattr(args, 'dest_flag') and args.dest_flag:
+        args.destination = args.dest_flag
+        
+    if not (args.source or args.destination):
+        logger.error("Either source, destination, or both must be specified")
+        return FAILURE
+    
+    src_rses = []
+    dst_rses = []
+    distances_to_delete = []
+    
+    # Check which arguments were provided and how to interpret them
+    if args.source and args.destination:
+        # If both source and destination are provided, we need to check if they're sites or RSEs
+        # First try to get them as RSEs
+        try:
+            client.get_rse(args.source)
+            src_rses = [args.source]
+        except Exception:
+            # Not an RSE, try as site
+            try:
+                src_rses = [rse['rse'] for rse in client.list_rses("site=%s" % args.source)]
+                if not src_rses:
+                    logger.error("Source '%s' is neither a valid RSE nor a site with RSEs" % args.source)
+                    return FAILURE
+            except Exception as e:
+                logger.error("Error retrieving RSEs for site %s: %s" % (args.source, str(e)))
+                return FAILURE
+            
+        try:
+            client.get_rse(args.destination)
+            dst_rses = [args.destination]
+        except Exception:
+            # Not an RSE, try as site
+            try:
+                dst_rses = [rse['rse'] for rse in client.list_rses("site=%s" % args.destination)]
+                if not dst_rses:
+                    logger.error("Destination '%s' is neither a valid RSE nor a site with RSEs" % args.destination)
+                    return FAILURE
+            except Exception as e:
+                logger.error("Error retrieving RSEs for site %s: %s" % (args.destination, str(e)))
+                return FAILURE
+    elif args.source:
+        # Delete all outgoing links from source
+        try:
+            client.get_rse(args.source)
+            src_rses = [args.source]
+        except Exception:
+            # Not an RSE, try as site
+            try:
+                src_rses = [rse['rse'] for rse in client.list_rses("site=%s" % args.source)]
+                if not src_rses:
+                    logger.error("Source '%s' is neither a valid RSE nor a site with RSEs" % args.source)
+                    return FAILURE
+            except Exception as e:
+                logger.error("Error retrieving RSEs for site %s: %s" % (args.source, str(e)))
+                return FAILURE
+        # For outgoing links, we need all possible destinations
+        dst_rses = [rse['rse'] for rse in client.list_rses()]
+    elif args.destination:
+        # Delete all incoming links to destination
+        try:
+            client.get_rse(args.destination)
+            dst_rses = [args.destination]
+        except Exception:
+            # Not an RSE, try as site
+            try:
+                dst_rses = [rse['rse'] for rse in client.list_rses("site=%s" % args.destination)]
+                if not dst_rses:
+                    logger.error("Destination '%s' is neither a valid RSE nor a site with RSEs" % args.destination)
+                    return FAILURE
+            except Exception as e:
+                logger.error("Error retrieving RSEs for site %s: %s" % (args.destination, str(e)))
+                return FAILURE
+        # For incoming links, we need all possible sources
+        src_rses = [rse['rse'] for rse in client.list_rses()]
+    
+    # Build list of distances to delete
+    for src in src_rses:
+        for dst in dst_rses:
+            if src != dst:  # Skip self-links
+                distances_to_delete.append((src, dst))
+                if args.bidirectional:
+                    distances_to_delete.append((dst, src))
+    
+    # De-duplicate the list (in case of bidirectional with same source and destination)
+    distances_to_delete = list(set(distances_to_delete))
+    
+    # Display the distances that will be deleted and ask for confirmation
+    if len(distances_to_delete) == 0:
+        logger.error("No distances to delete")
+        return FAILURE
+    
+    print("The following distances will be deleted:")
+    for src, dst in distances_to_delete:
+        print(f"  {src} → {dst}")
+    
+    if not args.yes:
+        confirmation = input("\nDo you want to proceed? (y/n): ")
+        if confirmation.lower() not in ['y', 'yes']:
+            print("Operation cancelled")
+            return SUCCESS
+    
+    # Delete the distances
+    for src, dst in distances_to_delete:
+        try:
+            client.delete_distance(src, dst)
+            print(f"Deleted distance information from {src} to {dst}")
+        except Exception as e:
+            error_str: str = str(e)
+            if "NoDistance" in error_str or "RSENotFound" in error_str:
+                logger.error(f"Distance from {src} to {dst} not found")
+            else:
+                logger.error(f"Failed to delete distance from {src} to {dst}: {error_str}")
+    
     return SUCCESS
 
 
@@ -1883,8 +2017,14 @@ def get_parser():
                                                                '"""""""""""""\n'
                                                                '::\n'
                                                                '\n'
+                                                               '    # Set distance in one direction\n'
                                                                '    $ rucio-admin rse add-distance JDOE_SCRATCHDISK JDOE_DATADISK\n'
-                                                               '    Set distance from JDOE_SCRATCHDISK to JDOE_DATADISK to 1/n'
+                                                               '    Set distance from JDOE_SCRATCHDISK to JDOE_DATADISK to 1\n'
+                                                               '\n'
+                                                               '    # Set distance in both directions (bidirectional)\n'
+                                                               '    $ rucio-admin rse add-distance --bidirectional JDOE_SCRATCHDISK JDOE_DATADISK\n'
+                                                               '    Set distance from JDOE_SCRATCHDISK to JDOE_DATADISK to 1\n'
+                                                               '    Set distance from JDOE_DATADISK to JDOE_SCRATCHDISK to 1\n'
                                                                '\n'
                                                                'Note::\n'
                                                                '\n'
@@ -1896,6 +2036,7 @@ def get_parser():
     add_distance_rses_parser.add_argument(dest='destination', action='store', help='Destination RSE name')
     add_distance_rses_parser.add_argument('--distance', dest='distance', default=1, type=int, help='Distance between RSEs')
     add_distance_rses_parser.add_argument('--ranking', '--ranking', dest='ranking', new_option_string='--ranking', default=1, type=int, action=StoreAndDeprecateWarningAction, help='Ranking of link')
+    add_distance_rses_parser.add_argument('--bidirectional', '--bidi', action='store_true', help='Add distances in both directions')
 
     # The update_distance_rses command
     update_distance_rses_parser = rse_subparser.add_parser('update-distance',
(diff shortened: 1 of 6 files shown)


tests/test_bin_rucio.py
```diff
@@ -283,6 +283,11 @@ def test_rse_add_distance(self):
         exitcode, out, err = execute(cmd)
         print(out, err)
         assert exitcode == 0
+        temprse3 = rse_name_generator()
+        cmd = 'rucio-admin rse add %s' % temprse3
```
0xquark: I noticed `rucio-admin` is still used in the tests, even though it's deprecated. I followed the same pattern for my tests—should I switch to using the new CLI instead? in tests.

tests/test_bin_rucio.py
```diff
@@ -303,6 +308,36 @@ def test_rse_add_distance(self):
...
+        cmd = 'rucio-admin rse add-distance --bidirectional --distance 1 %s %s' % (temprse1, temprse3)
+        print(self.marker + cmd)
+        exitcode, out, err = execute(cmd)
+        print(out, err)
+        assert exitcode == 0
+        assert "Set distance from %s to %s to 1" % (temprse1, temprse3) in out
+        assert "Set distance from %s to %s to 1" % (temprse3, temprse1) in out
+        
+        # Verify distances were set in both directions
+        cmd = 'rucio-admin rse get-distance %s %s' % (temprse1, temprse3)
```
0xquark: Is there a reason we only use assertions to check the commands, and not do any cross-verification? I feel like cross-verifying could help catch issues if something's wrong with the gateway or server.

voetberg: We cross check in some tests, but this is generally done through making a second call with the py-client. I believe calls back to the core are rare because a) This ideally is covered by tests for the individual components and b) The client test suite doesn't assume there is a DB connection exposed. 

@rdimaio can correct me if I'm wrong

tests/test_bin_rucio.py
voetberg: `rucio-admin` isn't yet deprecated, it's just *going* to be. For now I don't see an issue keeping the old format until it's fully retired

tests/test_bin_rucio.py
```diff
@@ -303,6 +308,36 @@ def test_rse_add_distance(self):
...
+        exitcode, out, err = execute(cmd)
+        print(out, err)
+        assert exitcode == 0
+        assert "distance=1" in out
+        
+        # Test with non-existent RSE
+        cmd = 'rucio-admin rse add-distance %s %s' % (temprse1, generate_uuid())
+        exitcode, out, err = execute(cmd)
+        print(out, err)
+        assert exitcode != 0
```
voetberg, line 338: `execute` generally only produces an exitcode of 0, 1, 2 - you want to check for 1 here (1 is "failed to successfully complete" and 2 is "failed to parse the CLI input")

tests/test_bin_rucio.py
```diff
@@ -610,24 +869,21 @@ def test_upload_expiration_date(self):
         cmd = 'rucio -v upload --legacy --rse {0} --scope {1} --expiration-date 2021----10-10-20:00:00 {2}'.format(self.def_rse, self.user, tmp_file)
         print(self.marker + cmd)
         exitcode, out, err = execute(cmd)
-        print(out)
-        print(err)
+        print(out, err)
```
voetberg: This probably shouldn't be included in this PR (it has nothing to do with RSE distance setting)

lib/rucio/cli/bin_legacy/rucio_admin.py
```diff
@@ -607,13 +613,207 @@ def update_distance_rses(args, client, logger, console, spinner):
...
+        error_count = 0
+        for src, dst in distances_to_delete:
+            try:
+                client.delete_distance(src, dst)
+                success_count += 1
+                if cli_config == 'rich':
+                    logger.info(f"Deleted distance information from {src} to {dst}")
+                else:
+                    print(f"Deleted distance information from {src} to {dst}")
+            except Exception as e:
```
voetberg: Catching all exceptions like this overwrites some of the error handling done in `exception_handler`, you should only catch `NoDistance` or `RSENotFound`.

lib/rucio/cli/rse.py
```diff
@@ -126,10 +126,12 @@ def distance_add(ctx, source_rse, destination_rse, distance):
 @distance.command("remove")
 @click.argument("source-rse")
 @click.argument("destination-rse")
+@click.option('-y', '--yes', is_flag=True, help='Automatically confirm deletion')
```
voetberg, line 130: You can use a [`confirmation_option`](https://click.palletsprojects.com/en/stable/api/#click.confirmation_option) to streamline this, but I don't know if this `confirmation` functionality is something we want period. I'm a little dubious about this just because I don't know how this works into people's existing workflows

lib/rucio/cli/rse.py
```diff
@@ -126,10 +126,12 @@ def distance_add(ctx, source_rse, destination_rse, distance):
...
 @click.pass_context
-def distance_remove(ctx, source_rse, destination_rse):
+def distance_remove(ctx: click.Context, source_rse: str, destination_rse: str, yes: bool) -> None:
     """Un-link SOURCE-RSE from DESTINATION-RSE by removing the distance between them"""
-    args = Arguments({"no_pager": ctx.obj.no_pager, "source": source_rse, "destination": destination_rse})
+    # Setting yes=True for non-interactive test environments, but keeping bidirectional=False to match legacy behavior
```
voetberg: I don't understand this comment, why are we limiting to `bidirection=False` for the newer client?

lib/rucio/cli/bin_legacy/rucio_admin.py
```diff
@@ -607,13 +613,207 @@ def update_distance_rses(args, client, logger, console, spinner):
...
+            if cli_config == 'rich':
+                print_output(f"{CLITheme.FAILURE_ICON} {error_msg}", console=console, no_pager=True)
+            else:
+                logger.error(error_msg)
+            return FAILURE
+        
+        if cli_config == 'rich':
+            distance_tree = Tree(f"The following {len(distances_to_delete)} distances will be deleted:")
+            for src, dst in sorted(distances_to_delete):
+                distance_tree.add(f"{src} → {dst}")
```
voetberg: I think this is the only place `→` is ever used in rucio, for consistency I'd use `->` instead.

lib/rucio/cli/bin_legacy/rucio_admin.py
```diff
@@ -607,13 +613,207 @@ def update_distance_rses(args, client, logger, console, spinner):
...
+        
+        # Resolve source and destination RSEs
+        src_rses = []
+        dst_rses = []
+        
+        if args.source:
+            # Check if source is an RSE
+            if args.source in all_rses:
+                src_rses = [args.source]
+            else:
```
voetberg: This is the only place in the client were we allow the definition of "rse" to also mean "site at an RSE", I think this is confusing and should not be allowed.

tests/test_bin_rucio.py
rdimaio: I'd use the new CLI in tests, though. When it will be deprecated, we'll have to migrate the tests to the new syntax, so we might as well simplify that process by writing new tests with the new syntax and without `rucio-admin` - https://github.com/rucio/rucio/issues/7260

tests/test_bin_rucio.py
rdimaio: This is a client-level test, and in my opinion here we should only care that the client works correctly, i.e. getting the expected output for a given input. All the layers called by this client call are expected to be tested individually, to ensure they're all individually working correctly, + the integ tests

lib/rucio/cli/bin_legacy/rucio_admin.py
```diff
@@ -1922,18 +2129,38 @@ def get_parser():
...
+                                                                  '\n'
+                                                                  '    # Delete all incoming links to an RSE\n'
+                                                                  '    $ rucio-admin rse delete-distance --dest JDOE_SCRATCHDISK\n'
                                                                   '\n')
     delete_distance_rses_parser.set_defaults(which='delete_distance_rses')
-    delete_distance_rses_parser.add_argument(dest='source', action='store', help='Source RSE name')
-    delete_distance_rses_parser.add_argument(dest='destination', action='store', help='Destination RSE name')
+    
+    # Create a mutually exclusive group for source/destination specification
+    src_dest_group = delete_distance_rses_parser.add_mutually_exclusive_group()
```
voetberg: This causes more problems than its worth, I think using just `source, nargs=?` `dest, nargs=?`

lib/rucio/cli/bin_legacy/rucio_admin.py
```diff
@@ -607,13 +613,207 @@ def update_distance_rses(args, client, logger, console, spinner):
...
+                        distances_to_delete.append((src, dst))
+        
+        # Remove duplicates
+        distances_to_delete = list(set(distances_to_delete))
+        
+        if cli_config == 'rich':
+            spinner.stop()
+        
+        # Handle case when no distances are found
+        if not distances_to_delete:
```
voetberg, line 774: It makes more sense to just raise an error here, the result is the same and the error message is cleaner.

lib/rucio/cli/bin_legacy/rucio_admin.py
```diff
@@ -607,13 +613,207 @@ def update_distance_rses(args, client, logger, console, spinner):
...
+                            distances_to_delete.append((dst, src))
+        
+        elif args.source:
+            # Delete all outgoing distances from source(s)
+            for src in src_rses:
+                if src in all_distances:
+                    for dst in all_distances[src]:
+                        distances_to_delete.append((src, dst))
+        
+        elif args.destination:
```
voetberg, line 759: This just duplicates the source rse selection, no? It would be cleaner to just do `if args.source or args.destination` and handle the logic in the rse loop.

rcarpa:
Would be nice to use this new feature to set distances in the integration tests here : https://github.com/rucio/rucio/blob/22868d219b45929a7e3016d12f0d5b67e18166d2/tools/docker_activate_rses.sh#L95

tests/test_bin_rucio.py
0xquark: I believe using the new CLI in tests is a good idea, but my concern is whether we've finalized the commands mentioned in [https://github.com/rucio/rucio/issues/7260](https://github.com/rucio/rucio/issues/7260). I don't want the tests to fail if we decide to change the names of some commands later.

tests/test_bin_rucio.py
0xquark: Shouldn't the test needs to be failed for both the reasons? I mean if we check only for 1 here, we are limiting the reason of failure and making the test more brittle, if the implementation or anything changes in future which results in validation error, it could result error code 2. 
With `assert exitcode != 0` we don't worry about the reason but more about the failure and expanding the scope. let me know what do you think?

lib/rucio/cli/rse.py
0xquark: I suggest we stick with the current approach of using `click.option` instead of switching to `confirmation_option` because it gives us more control. Right now, we first show which distances will be deleted, then ask for confirmation, and finally delete them.

With confirmation_option, it seems like the confirmation prompt would appear right when the command starts before showing what will be deleted—which isn’t ideal. I haven’t tested it, but that’s how I expect it to behave.

Current flow:
Show distances to delete → Ask for confirmation → Delete

Expected confirmation_option flow:
Ask for confirmation right away → Show distances → Delete (assuming)

Also, I’m not sure if this new option would support different modes like Rich or plain text.

You can see the current deletion confirmation implementation here:
https://github.com/rucio/rucio/blob/fc6fb2d40fcad06d1f0159fc479304791063efeb/lib/rucio/cli/bin_legacy/rucio_admin.py#L766-L776

The above `@click.option('-y', '--yes', is_flag=True, help='Automatically confirm deletion')` is for skipping the confirmation box in the command.

lib/rucio/cli/rse.py
0xquark: Before, the CLI didn’t have a bidirectional option. If we now make a breaking change where bidirectional is set to true, it would delete distances in both directions which we probably don’t want. Earlier, `distance_remove` only deleted the distance in one direction, so the old behavior should stay the same even with the new feature.

voetberg: I understand that, but why not just add an option for bidirectional and default it to False?

lib/rucio/cli/rse.py
voetberg: You can connect 'confirmation option' to the prompt if you are concerned with showing proper information first (examples here)[https://click.palletsprojects.com/en/stable/prompts/#confirmation-prompts], but I think I disagree with using confirmation prompts o n principal (unless they are included everywhere). It creates an inconsistent user experience. I would like this option removed for now and we can open it up for discussion at a later date.

0xquark: I believe deleting distances in bulk is a major and potentially risky action, so we should always confirm with the user before doing it. Showing a confirmation prompt is a good way to make sure they know exactly what will be deleted. In this case, it’s more important to focus on user safety and clarity than on having a perfectly consistent user experience.

lib/rucio/cli/bin_legacy/rucio_admin.py
0xquark: I think this is the only place in the client where you can use a site name instead of an RSE name. Everywhere else in the code, when it asks for an RSE, it expects a real RSE name. But here, the input can be either a site or an RSE so you can delete the distance between two specific RSEs, or between all RSEs at one site and all RSEs at another site. I’ll update the comments to explain this more clearly.

lib/rucio/cli/rse.py
0xquark: Now I Understood! Thanks. 
I was confused before with New CLI vs Legacy CLI. So the options for new cli are added using click in `rse.py` while the options for old cli are added through argparser in `bin_legacy/rucio_admin.py`. 
I'll add the `bidirectional` option configurable in the new cli.

lib/rucio/cli/bin_legacy/rucio_admin.py
0xquark: Do you mean to remove the `--src` and `--dest` flags? and instead use source and destination only. 
If we do this, in order to delete all the incoming links to a destination, we would need to specify it using an empty space for the source `rucio-admin rse delete-distance '' JDOE_SCRATCHDISK`. According to me, in terms of user experience, `rucio-admin rse delete-distance --dest JDOE_SCRATCHDISK` is much more better.

voetberg: I completely agree with that, but this additional option of having the both options in mutually exclusively groups is way more confusing than having just one or the other and introduces all that nested logic in the function itself. I would recommend leaving "source"/"destination" in rucio-admin, and moving the '--src' '--dest' flags to the new client in `cli/rse.py`

lib/rucio/cli/bin_legacy/rucio_admin.py
voetberg: Hi, please don't resolve comments until you pushed the change, it makes it harder to follow up

tests/test_bin_rucio.py
voetberg: In principal I don't think you're wrong, but in practice, this can hide problems with the test stemming form the CLI parsing, which is a different sort of check. You should check that it is specifically failing for the reason you expect it to fail - like you can check that the error message calls out that the RSE doesn't exist in addition to the exitcode.

voetberg: Which you do and I missed that it's right below this comment, lol. Nvm.

lib/rucio/cli/rse.py
voetberg: I really do think we *should* do that (I tried to add it to the CLI rewrite lol), but we don't. This is a change beyond the stated goal of the issue/pr so for tracking/release management purposes, it should be a different issue.

0xquark: What I suggest is, Let's have the confirmation block in this feature since this is something which was initially asked in the issue (https://github.com/rucio/rucio/issues/7570#issuecomment-2801240422) and open a new issue for major overwrite elsewhere in the cli.

lib/rucio/cli/bin_legacy/rucio_admin.py
0xquark: Great! I've moved the flags (`--src` and `--dest`) to be part of the new cli and only kept positional source and destination arguments in the legacy cli.

0xquark:
```
______________________ TestBinRucio.test_rse_add_distance ______________________
[gw1] linux -- Python 3.9.21 /opt/venv/bin/python
tests/test_bin_rucio.py:309: in test_rse_add_distance
    assert exitcode != 0
E   assert 0 != 0
------------------------------ Captured log setup ------------------------------
DEBUG    urllib3.connectionpool:connectionpool.py:1021 Starting new HTTPS connection (1): localhost:443
DEBUG    urllib3.connectionpool:connectionpool.py:474 https://localhost:443 "GET /auth/userpass HTTP/1.1" 200 0
DEBUG    urllib3.connectionpool:connectionpool.py:1021 Starting new HTTPS connection (2): localhost:443
DEBUG    urllib3.connectionpool:connectionpool.py:474 https://localhost:443 "GET /rses/MOCK4 HTTP/1.1" 200 927
DEBUG    urllib3.connectionpool:connectionpool.py:1021 Starting new HTTPS connection (1): localhost:443
DEBUG    urllib3.connectionpool:connectionpool.py:474 https://localhost:443 "POST /accountlimits/local/root/MOCK4 HTTP/1.1" 201 7
DEBUG    urllib3.connectionpool:connectionpool.py:474 https://localhost:443 "POST /rses/MOCK4/attr/istape HTTP/1.1" 201 7
----------------------------- Captured stdout call -----------------------------
Added new deterministic RSE: MOCK-YSQEASDEKE
 
Added new deterministic RSE: MOCK-CGGUTJOJWZ
 
Added new deterministic RSE: MOCK-AEJCMYGSLD
 
$> rucio rse distance add MOCK-YSQEASDEKE MOCK-CGGUTJOJWZ --distance 1
Set distance from MOCK-YSQEASDEKE to MOCK-CGGUTJOJWZ to 1
 
$> rucio rse distance add MOCK-CGGUTJOJWZ MOCK-YSQEASDEKE --distance 1
Set distance from MOCK-CGGUTJOJWZ to MOCK-YSQEASDEKE to 1
 
$> rucio rse distance add MOCK-CGGUTJOJWZ MOCK-YSQEASDEKE --distance 1
 2025-05-12 12:18:29,933	ERROR	An object with the same identifier already exists.
Details: Distance from MOCK-CGGUTJOJWZ to MOCK-YSQEASDEKE already exists!
 0
 ```
 Why does the new cli provides with exit code 0 for an error? I think this is a bug.
 @voetberg

voetberg:
> ```
> ______________________ TestBinRucio.test_rse_add_distance ______________________
> [gw1] linux -- Python 3.9.21 /opt/venv/bin/python
> tests/test_bin_rucio.py:309: in test_rse_add_distance
>     assert exitcode != 0
> E   assert 0 != 0
> ------------------------------ Captured log setup ------------------------------
> DEBUG    urllib3.connectionpool:connectionpool.py:1021 Starting new HTTPS connection (1): localhost:443
> DEBUG    urllib3.connectionpool:connectionpool.py:474 https://localhost:443 "GET /auth/userpass HTTP/1.1" 200 0
> DEBUG    urllib3.connectionpool:connectionpool.py:1021 Starting new HTTPS connection (2): localhost:443
> DEBUG    urllib3.connectionpool:connectionpool.py:474 https://localhost:443 "GET /rses/MOCK4 HTTP/1.1" 200 927
> DEBUG    urllib3.connectionpool:connectionpool.py:1021 Starting new HTTPS connection (1): localhost:443
> DEBUG    urllib3.connectionpool:connectionpool.py:474 https://localhost:443 "POST /accountlimits/local/root/MOCK4 HTTP/1.1" 201 7
> DEBUG    urllib3.connectionpool:connectionpool.py:474 https://localhost:443 "POST /rses/MOCK4/attr/istape HTTP/1.1" 201 7
> ----------------------------- Captured stdout call -----------------------------
> Added new deterministic RSE: MOCK-YSQEASDEKE
>  
> Added new deterministic RSE: MOCK-CGGUTJOJWZ
>  
> Added new deterministic RSE: MOCK-AEJCMYGSLD
>  
> $> rucio rse distance add MOCK-YSQEASDEKE MOCK-CGGUTJOJWZ --distance 1
> Set distance from MOCK-YSQEASDEKE to MOCK-CGGUTJOJWZ to 1
>  
> $> rucio rse distance add MOCK-CGGUTJOJWZ MOCK-YSQEASDEKE --distance 1
> Set distance from MOCK-CGGUTJOJWZ to MOCK-YSQEASDEKE to 1
>  
> $> rucio rse distance add MOCK-CGGUTJOJWZ MOCK-YSQEASDEKE --distance 1
>  2025-05-12 12:18:29,933	ERROR	An object with the same identifier already exists.
> Details: Distance from MOCK-CGGUTJOJWZ to MOCK-YSQEASDEKE already exists!
>  0
> ```
> 
> Why does the new cli provides with exit code 0 for an error? I think this is a bug. @voetberg

Ah, I see what is happening here. The legacy CLI doesn't throw an error it just returns an integer 1 to indicate an error, which the error handling doesn't catch because the new CLI doesn't return anything. I suppose that counts as a bug, but I'm not sure how best to go about solving it. 

You can either add a return statement for the new CLI (`return add_distance_rses(...)`, or throw an error (raise an `Duplicate` error) - we don't have a solid design for this so I'll let you decide.

dchristidis (commented):
I’m afraid both the design and the implementation are heading in the wrong direction.

Design-wise, the way to go from a single RSE to many RSEs is through an RSE expression. This is our construct and it should be used wherever possible. An RSE name is a valid RSE expression, so it’s backwards-compatible. Remember that the `site` RSE attribute is not guaranteed to be set. And the way to express all RSEs is through the RSE expression `*`.

Implementation-wise, all this logic should **EDIT: not** exist in the client. The core/gateway/API should be adapted to accept RSE expressions instead of RSE names.

0xquark:
It’s been quite a while, and rebasing everything would be a bit challenging, so I’ll open a new PR with the updates. Closing this one now.

PR #7697, opened 2025-05-05
Feature 7687 wlcg token discovery

### LINKED ISSUE #7687: OIDC Token authentication based on WLCG Disovery guideline
### Description

Investigate the implementation of [WLCG Bearer Token Discovery guideline ](https://zenodo.org/records/3937438)for Rucio authentication.

@bjwhite-fnal 

### Motivation

_No response_

### Change

_No response_

panta-123:
Add wlcg_token_discovery specs.
https://zenodo.org/records/3937438

it is invoked for auth_type oidc,
-  invoke wlcg_token_discovery if auth_type oidc
- if found it will send that to do validate_jwt and if ok stored and ok auth . 
- if no token from above  then go on with regular flow (ie. look for rucio token file  and so on (which includes oidc flow))

closes #7687

### DIFF
--- a/lib/rucio/common/utils.py
+++ b/lib/rucio/common/utils.py
@@ -1708,7 +1708,7 @@ def get_transfer_schemas() -> dict[str, list[str]]:
     return scheme_map
 
 
-def wlcg_token_discovery(user_id=None):
+def wlcg_token_discovery() -> Optional[str]:
     """
     Discovers a WLCG bearer token from the environment, following the specified precedence.
     Specs: https://zenodo.org/records/3937438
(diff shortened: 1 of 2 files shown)


lib/rucio/common/utils.py
```diff
@@ -1690,3 +1690,63 @@ def is_method_overridden(obj, base_cls, method_name):
     if getattr(type(obj), method_name, None) is getattr(base_cls, method_name, None):  # Caring for bound/unbound cases
         return False
     return True
+
+
+def wlcg_token_discovery():
```
rdimaio:
```suggestion
def wlcg_token_discovery() -> Optional[str]:
```

rdimaio (approved):
LGTM, but best to get approval from a token component lead too before merging

PR #7703, opened 2025-05-06
Rules: check stuck locks for rule ID that are on non-target RSEs

### LINKED ISSUE #7538: Record rule error for STUCK rule if rse_expression excludes lock_rse in the meanwhile
### Description

Situation: We have a replication rule, DATASET grouped, with an rse_expression with dynamic attributes. E.g. `type=DATADISK&datapolicynucleus=True\(cloud=US|cloud=CA)&freespace>300` thus freespace might change over time.

1. At the creation of the rule, a destination was chosen for some locks, the transfer failed, rule was set to `STUCK`
2. At the time of the first (and continous) repairs of the rule, the rse_expression evaluation did NOT include the originally chosen destination for (at least some) STUCK locks anymore.

Thus the rule just stays stuck. This is correct behaviour, however, there is no error message recorded why the rule is stuck in this case.

What happens is that in the `repair_rule` workflow the rse_expression gets evaluated here

https://github.com/rucio/rucio/blob/c1ed21347cd716180771660835f37490f980c960/lib/rucio/core/rule.py#L1287-L1297

Afterwards when selecting the datasetfiles, locks, replicas, and source replicas here

https://github.com/rucio/rucio/blob/c1ed21347cd716180771660835f37490f980c960/lib/rucio/core/rule.py#L1391-L1396

the STUCK locks are not included in the lists anymore, since they are excluded due to the `restrict_rses` value. This causes the further workflow not to process these locks anymore.

### Motivation

_No response_

### Change

The proposed change is, for `ALL` and `DATASET` grouped rules, after the rse_expression has been evaluated, to compare it against rse_ids of STUCK locks of the rule. If the sets mismatch, then record an error in the `rule.error` attribute. Still continue processing the rest of the rule, but if no other errors were detected, this error will stay recorded at the end of the workflow.

rdimaio:
fix #7538

### DIFF
--- a/lib/rucio/core/rule.py
+++ b/lib/rucio/core/rule.py
@@ -1261,6 +1261,7 @@ def repair_rule(
     #     created.
     # (C) Transfers fail and mark locks (and the rule) as STUCK. All STUCK locks have to be repaired.
     # (D) Files are declared as BAD.
+    # (E) Stuck locks are found on RSEs that do not belong to the target RSEs.
 
     # start_time = time.time()
     try:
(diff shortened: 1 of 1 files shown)


voetberg:
Is there an existing test that covers this?

rdimaio:
> Is there an existing test that covers this?

I could not find anything

bari12 (approved):
LGTM, please add a test though

rdimaio:
> LGTM, please add a test though

created https://github.com/rucio/rucio/issues/7882 as a follow-up

Issue #7581, opened 2025-03-27
remove oauth manager from the repo


panta-123:
Remove oauth manager daemon and not needed funtionality.

PR #7705, opened 2025-05-07
Feature 7310 rework OIDC auth proper commit

panta-123:
re-PR, splitting commit into chunks.

- Remove oic libary dependency and pyoidc stuff. replace it with pyjwt.
- changed the format of the idpsecret.json -> also results in multi-vo capabilities. 
- Remove auto for OIDC auth as it doesn't adheres to standard.
- Removed all code used for oauthmanager daemon and daemon it self.
- The expired oauth-request and token entry deleted during new creation.
- Have to modify all the test as base libary and structure are changed.
-


dchristidis (changes requested):
This is a step in the right direction, but the fourth commit makes it very difficult to review. It’s a combination of multiple objectives, some of them with unclear motivation (I’m sure there is a perfectly valid reason, but I can’t deduce it by looking at an individual line change and then trying to associate it with an item in the commit message). A refactor cannot be an almost atomic rewrite.

In order to not risk delaying your efforts any longer, I would highly recommend separating your objectives, one PR at a time. You’ve already opened issue #7581; open a PR with the first three commits and it will get merged very quickly (I have only a couple of suggestions about the commit messages, otherwise the code changes are well in order).

Then you have a couple of changes that look like prerequisites. The move away from the deprecated Python modules is an admirable and desirable goal; please make an issue and a PR, restricting yourself to maintaining the existing functionality. You can include other small improvements as separate commits (like removing the use of `curl` which always bugged me), but not too many at once.

Similarly for the change to the configuration file structure. You might wish to describe this beforehand in the issue. I don’t have an objection on the whole, but I may have questions and/or wish to offer minor suggestions.

Finally, once those are done and merged, we can see what remains and plan accordingly.

PR #7707, opened 2025-05-08
Handle postgresql lock not available exceptions in daemons

### LINKED ISSUE #7706: Database postgresql lock not available exceptions not handled correctly
We are seeing some lock not available exceptions in our postgresql database which are not handled properly.

Looking at some of the code I see that these are only handled for Oracle in some cases.

Also, the psql lock regex: `PSQL_LOCK_NOT_AVAILABLE_REGEX = r".*55P03.*"` appears to be outdated for our psql database with the psycopg connector. I will add a new regex.

lobis:
Closes #7706 

Add new exception and use it where the oracle equivalent-ish exception is used (`ORACLE_RESOURCE_BUSY_REGEX`).

Also add `MYSQL_LOCK_NOWAIT_REGEX` to (hopefully) make this also work on MySQL.

This patch fixes the failing daemons in our postgresql deployments.


labkode (approved):

lib/rucio/daemons/conveyor/finisher.py
```diff
@@ -140,7 +140,7 @@ def _handle_requests(
         logger(logging.DEBUG, 'Finish to update %s finished requests in %s seconds', len(reqs), total_stopwatch.elapsed)
 
     except (DatabaseException, DatabaseError) as error:
-        if re.match(ORACLE_RESOURCE_BUSY_REGEX, error.args[0]) or re.match(ORACLE_DEADLOCK_DETECTED_REGEX, error.args[0]) or MYSQL_LOCK_WAIT_TIMEOUT_EXCEEDED in error.args[0]:
+        if re.match(ORACLE_RESOURCE_BUSY_REGEX, error.args[0]) or re.match(ORACLE_DEADLOCK_DETECTED_REGEX, error.args[0]) or re.match(PSQL_PSYCOPG_LOCK_NOT_AVAILABLE_REGEX, str(error.args[0])) or MYSQL_LOCK_WAIT_TIMEOUT_EXCEEDED in error.args[0]:
             logger(logging.WARNING, 'Lock detected when handling request - skipping: %s', str(error))
```
rdimaio, line 144: I'm a bit confused here - `PSQL_PSYCOPG_LOCK_NOT_AVAILABLE_REGEX` indicates that this pattern matches when a lock is not available (i.e. when it's not found?), but the log message says "Lock detected when handling request"

lobis: Yes, maybe the message can be reworded (in a separate PR).

I guess the idea is that an existing lock has been detected so it's not possible to acquire the lock.

labkode: The error condition is that we're trying to adcquire a lock and the DB is busy and won't give us that lock.
I think line 144 can be changed to (in a different PR maybe):
```
cannot get lock, DB is busy, skipping ...
```

rdimaio: Issue created: https://github.com/rucio/rucio/issues/7711

rdimaio (commented):
Many of these changes also add a match for `MYSQL_LOCK_NOWAIT_REGEX` - however, this is not mentioned in either the PR description or commit messages, which only mention `PSQL_PSYCOPG_LOCK_NOT_AVAILABLE_REGEX`. Can you modify the relevant commit message and the PR descriptions to add this?

lobis:
> Many of these changes also add a match for `MYSQL_LOCK_NOWAIT_REGEX` - however, this is not mentioned in either the PR description or commit messages, which only mention `PSQL_PSYCOPG_LOCK_NOT_AVAILABLE_REGEX`. Can you modify the relevant commit message and the PR descriptions to add this?

Done!

rdimaio (approved):

rdimaio:
Checked with @bari12 , alright to merge. Additional note from @lobis over Mattermost - "it's been a patch in our deployments for 1 month without issue", so it seems safe

PR #7718, opened 2025-05-12
Docker & Kubernetes: Make `docker_compose.yml` compatible with Podman; Fix rucio#7715 

### LINKED ISSUE #7715: Make `docker-compose.yml` compatible with Podman
### Description

In a recent Rucio meeting it was mentioned that Podman could no longer read the dev's `docker-compose.yml` properly. This issue is about fixing this. The problem eventually is not the kebab-case used in the `container_name` fields, but rather the syntax used in the image definitions (namely, `"docker.io/${DOCKER_REPO:-rucio}/rucio-dev${RUCIO_TAG:+:release-${RUCIO_TAG}}"`) which is not supported by Podman.

I will push an update to handle this, retaining the ability to specify a Rucio release and at the same time utilize the `docker-compose.yaml` from both the `bootstrap_dev.sh` and either `docker-compose` or `podman-compose`.

However, I would like to add couple comments about this.

I am noticing that even after fixing this, Podman is still acting a bit weird (for reasons unrelated to this issue). Has anyone used it with success lately to run our tests?

Also, the new suggested definition `"docker.io/${DOCKER_REPO:-rucio}/rucio-dev:${RUCIO_DEV_PREFIX:-}${RUCIO_TAG:-latest}"` introduces a new var (namely the `RUCIO_DEV_PREFIX`)  because on dockerhub, we are using the prefix `release-` only for the dev images (and not for the rest ones). Would it make sense to apply a common naming for all docker images?

Lastly, I think we should include somewhere in our instructions about Podman usage (and also Docker too), a strong suggestion that users should better login before trying to deploy our containers (e.g. using `podman login`). This, because non-authenticated users tend to get (much more often) errors from dockerhub, due to pulling limits.

### Steps to reproduce

Try to use `podman-compose` with our `docker-compose.yaml`.

### Rucio Version

37.2.0

### Additional Information

_No response_

Geogouz:
As described in #7715, this PR makes `docker_compose.yml` compatible with Podman. It also switches to fixed project/container names to avoid possible errors when deploying those under a custom `--project-name`, and updates the examples in `bootstrap_dev.sh`.

### DIFF
--- a/etc/docker/dev/docker-compose.yml
+++ b/etc/docker/dev/docker-compose.yml
@@ -1,10 +1,11 @@
-name: rucio
+name: dev
 
 services:
   # ------------------------------------------------------------------
   # Main Rucio dev containers
   # ------------------------------------------------------------------
   rucioclient:
+    container_name: dev-rucioclient-1
     image: "docker.io/${DOCKER_REPO:-rucio}/rucio-dev${RUCIO_TAG:+:release-${RUCIO_TAG}}"
     platform: linux/amd64
     entrypoint: ["/rucio_source/etc/docker/dev/rucio/entrypoint.sh"]
@@ -65,6 +66,7 @@ services:
   # Database, MQ, metrics, and other external services
   # ------------------------------------------------------------------
   ruciodb:
+    container_name: dev-ruciodb-1
     image: docker.io/postgres:14
     environment:
       - POSTGRES_USER=rucio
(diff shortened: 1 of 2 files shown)


etc/docker/dev/docker-compose.yml
```diff
@@ -35,7 +36,7 @@ services:
 
   rucio:
     container_name: dev-rucio-1
-    image: "docker.io/${DOCKER_REPO:-rucio}/rucio-dev${RUCIO_TAG:+:release-${RUCIO_TAG}}"
+    image: "docker.io/${DOCKER_REPO:-rucio}/rucio-dev:${RUCIO_DEV_PREFIX:-}${RUCIO_TAG:-latest}"
```
rdimaio, line 39: Following discussion - this is necessary for Podman compatibility, as there is no better way to allow for this kind of logic for now. We can simplify in future if we decide to remove the `release-` prefix from the dev rucio container.

rdimaio (dismissed):

rdimaio (commented):
Please add a component label to each commit message

rdimaio (dismissed):

rdimaio (commented):
Please fix merge conflicts

rdimaio (approved):

PR #7730, opened 2025-05-19
Conveyor: Remove ftshosts configuration from multiple Rucio configuration templates

### LINKED ISSUE #7696: Is `ftshosts` configuration removed from conveyors?
### Description

Check, if yes, this should be removed from docs, config, etc.

### Steps to reproduce

-

### Rucio Version

_No response_

### Additional Information

_No response_

0xquark:
I went through the code and noticed that the conveyor daemons (like the submitter, poller, receiver, etc.) no longer use the ftshosts setting. Instead, they get the FTS server information from the RSE attributes (through `RseAttr.FTS`). The FTS3Transfertool class uses the `_pick_fts_servers` method to do this. Since the ftshosts config was still present in some files but wasn’t being used anymore, I cleaned those up and removed them in this PR.

Fixes #7696

### DIFF
--- a/etc/docker/test/extra/rucio_default.cfg
+++ b/etc/docker/test/extra/rucio_default.cfg
@@ -55,7 +55,6 @@ scheme = https,davs,gsiftp,root,srm,mock,file,magnet
 #hostcert = /etc/grid-security/hostcert.pem
 #hostkey = /etc/grid-security/hostkey.pem
 transfertool = fts3
-ftshosts = https://fts:8446
 cacert = /etc/grid-security/certificates/5fca1cb1.0
 usercert = /opt/rucio/etc/usercertkey.pem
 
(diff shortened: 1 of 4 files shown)


etc/docker/test/extra/rucio_default.cfg
```diff
@@ -55,7 +55,7 @@ scheme = https,davs,gsiftp,root,srm,mock,file,magnet
 #hostcert = /etc/grid-security/hostcert.pem
 #hostkey = /etc/grid-security/hostkey.pem
 transfertool = fts3
-ftshosts = https://fts:8446
+#ftshosts = https://fts:8446
```
rdimaio: Any reason for commenting it out rather than deleting it here?

0xquark: There's no particular reason for this, I just commented it out of consistency with the other options in the configuration.

rdimaio: The other commented out options mostly seem to be currently valid options, whereas this `ftshosts` option is deprecated - I think it's best if we fully remove it for clarity

rdimaio (approved):

PR #7733, opened 2025-05-19
Policies: remove redundant import when checking package version #7149

### LINKED ISSUE #7149: Redundant package import when checking policy version compatibility
When `check_policy_package_version` is used, we run `importlib.import_module(package)` on the policy package twice.

1. once in `check_policy_package_version`: 
https://github.com/rucio/rucio/blob/81c0d4712fdd3b5277aece71cd22190b961c3139/lib/rucio/common/utils.py#L2107-L2114

2. once outside of `check_policy_package_version`:
- In `plugins`:

https://github.com/rucio/rucio/blob/81c0d4712fdd3b5277aece71cd22190b961c3139/lib/rucio/common/plugins.py#L133-L134

This is harmless, but still redundant. It's worth refactoring to avoid this double import.

jamesp-epcc:
Refactors the code to avoid importing the top level policy package module twice.

### DIFF
--- a/lib/rucio/common/plugins.py
+++ b/lib/rucio/common/plugins.py
@@ -27,41 +27,34 @@ from rucio.version import current_version
 
 if TYPE_CHECKING:
     from collections.abc import Callable
+    from types import ModuleType
 
     from rucio.common.types import LoggerFunction
 
 PolicyPackageAlgorithmsT = TypeVar('PolicyPackageAlgorithmsT', bound='PolicyPackageAlgorithms')
 
 
-def check_policy_package_version(package: str, logger: 'LoggerFunction' = logging.log) -> None:
+def check_policy_module_version(module: 'ModuleType', logger: 'LoggerFunction' = logging.log) -> None:
 
     '''
-    Checks that the Rucio version supported by the policy package is compatible
+    Checks that the Rucio version supported by the policy module is compatible
     with this version. Raises an exception if not.
-    :param package: the fully qualified name of the policy package
+    :param module: the top level module of the policy package
     '''
     try:
-        supported_versionset = _get_supported_versions_from_policy_package(package)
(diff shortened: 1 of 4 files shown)


voetberg (dismissed):

rdimaio (commented):
I think it can be a bit confusing to have both `check_policy_module_version` and `check_policy_package_version` as available methods - from what I see, it's either:

1. import module first, then run `check_policy_module_version` (e.g. in `PolicyPackageAlgorithms`)
2. check package version first, then import module (e.g. in `core.permission`)

I think it would be best to standardize the way we do this version checking so that we only have a single way of doing things - I'm personally in favour of the first approach, so importing module first and then running `check_policy_module_version` - in this case, `check_policy_package_version` is no longer needed and can be removed and all its usages replaced with approach 1

lib/rucio/common/plugins.py
```diff
@@ -160,8 +168,8 @@ def _try_importing_policy(cls: type[PolicyPackageAlgorithmsT], vo: str = "") ->
             if not package:
                 package = str(config.config_get('policy', 'package' + ('' if not vo else '-' + vo)))
 
-            check_policy_package_version(package)
```
rdimaio, line 163: Blocking for discussion

jamesp-epcc:
That makes sense. I'll make those changes in this sprint.

jamesp-epcc:
Refactored to remove one of the functions. I don't think the integration test failure is related to my changes.

rdimaio (approved):
LGTM, CI error unrelated

rdimaio (commented):
Please rebase to get the changes that fix the unrelated CI issues - I assume the rest of the integ tests will pass given the scope of this PR, but best to play it safe before merging

PR #7741, opened 2025-05-22
Script to autogenerate documentation for configuration parameters, check in autotest

### LINKED ISSUE #325: Generate "Configuration parameters" documentation page automatically, to avoid mismatches between documentation and code

voetberg:
Paired with https://github.com/rucio/documentation/pull/564

Closes https://github.com/rucio/documentation/issues/325

[Uses the existing documentation page as a basis](https://github.com/rucio/documentation/blob/a8e87264eea5abdbf46fe9fb2675d603b1794f64/docs/operator/configuration_parameters.md)

### DIFF
--- a/lib/rucio/common/config_settings.py
+++ b/lib/rucio/common/config_settings.py
@@ -1,0 +1,395 @@
+# Copyright European Organization for Nuclear Research (CERN) since 2012
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from rucio.common.types import ConfigOption
+
+
+class Alembic:
+    name = "alembic"
+    cfg = ConfigOption(
+        name, "cfg",
+        "Path to the configuration file (.ini) for Alembic",
+    )
+
+
+class Auditor:
+    name = "auditor"
+    cache = ConfigOption(name, "cache", "Path to the folder to store the `rucio-auditor` cache. Example: `/opt/rucio/auditor-cache`")
+    results = ConfigOption(name, "results", "Path to the folder to store the `rucio-auditor` results. Example: `/opt/rucio/auditor-results`")
+    logdir = ConfigOption(name, "logdir", "Path of the directory for logs, joined with `auditor.log``")
+    threshold = ConfigOption(name, "threshold", "floating number used in a sanity check, comparing the number of entries with the total number of files on the RSE", float, 0.1)
+
+
+class BB8:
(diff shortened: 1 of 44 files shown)


rdimaio (commented):
Please rebase

rdimaio:
Please rebase to get the changes that fix the unrelated integ test error

voetberg:
@rdimaio rebased 

cc: @bari12 re our meeting earlier

voetberg:
@rdimaio I think this is ready to review now

rdimaio:
> @rdimaio I think this is ready to review now

I see a failing test in the CI https://github.com/rucio/rucio/actions/runs/17928027139/job/50978902584?pr=7741

voetberg:
> > @rdimaio I think this is ready to review now
> 
> I see a failing test in the CI https://github.com/rucio/rucio/actions/runs/17928027139/job/50978902584?pr=7741

🤦 Solved now

rdimaio (commented):
This `ConfigOption` change seems pretty big and not the same as simply autogenerating documentation, maybe we should have it in a separate PR just for ease of review

voetberg:
> This `ConfigOption` change seems pretty big and not the same as simply autogenerating documentation, maybe we should have it in a separate PR just for ease of review

I agree that it's rather large, but I don't think there's a way to do it in two PRs without having something failing. (The autodocs needs to the ConfigOptions, there's no way to separate them.) The original version of this PR was agreed to be a patch fix that could be made more robust so this is the more complete version.

lib/rucio/common/config_settings.py
rdimaio, line 1: I'm not super sure about this. I think we should probably move towards some kind of config validation mechanism for sure, but it couuld be that by introducing this we have an extra place where we need to maintain config settings and such. 

Also, I think the PR title should mention this very big change rather than just the autogenerated documentation aspect.

I wonder if this should be introduced gradually (like one config setting class at a time)

bari12: I am ok with going forward in one big change. It is a bit of a pain to break this down into small parts and make the documentation consistent at the same time. So for a major release, this is ok.

I am not 100% sure about the `config_get` signature. I do not like that, if a `ConfigOption` is used, that the `section` is just an empty string. What if it is set? Maybe there should be an entire separate command for using the `ConfigOption` which does not ask for a section/option, but just for a ConfigOption?

voetberg: I was considering that, but was unsure if that was too big a change. I can totally try it though and see how we feel about it

lib/rucio/common/plugins.py
voetberg, line 1: Quick comment that the get_default function was moved here to avoid a circular import problem

Geogouz:
There are some type checking issues needed to be addressed first.

dchristidis:

PR #7751, opened 2025-05-28
protocols: use GET requests to check for existing files with presigned urls

### LINKED ISSUE #7714: WebDAV protocol makes HEAD requests, which the S3 pre-signed URL generation doesn't support
### Description

Rucio creates a pre-signed URL to allow interactions with the an S3 endpoint.

In `connect` and `exists` methods, the webdav protocol support uses the HTTP HEAD request.  For S3 endpoints, a get_object pre-signed URL is used.  This is a pre-signed URL that supports the HTTP GET request.

Unfortunately, a pre-signed URL generated to support an HTTP  GET request doesn't allow a client to make an HTTP HEAD request.  Such requests fail with a 403 error.

Possible solutions:
  1. Update lib/rucio/core/credential.py to generate `head_object` pre-signed URLs.
  2. Change the HEAD requests (in `connect` and `exists`) to GET requests with a 1 byte Range request header.

### Steps to reproduce

1. try to upload a file to an S3 RSE, configured to use the webdav protocol.

### Rucio Version

_No response_

### Additional Information

_No response_

tbeerman:
fixes #7714

### DIFF
--- a/lib/rucio/rse/protocols/webdav.py
+++ b/lib/rucio/rse/protocols/webdav.py
@@ -253,8 +253,15 @@ class Default(protocol.RSEProtocol):
             :raise  ServiceUnavailable, RSEAccessDenied
         """
         path = self.path2pfn(pfn)
+
+        using_presigned_urls = self.rse['sign_url'] is not None
+
         try:
-            result = self.session.request('HEAD', path, verify=False, timeout=self.timeout, cert=self.cert)
+            # use GET instead of HEAD for presigned urls
+            if not using_presigned_urls:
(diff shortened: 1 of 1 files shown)


rdimaio (commented):
Does this also need to be done for `connect`?

tbeerman:
The connect doesn't really work either way with S3. It tries to do a HEAD on the root of the server. That returns a 405 Method not allowed, at least for the endpoints I tried. Doing a GET request doesn't work either, because it will return 403 Forbidden. It would only work with the pre-signed url.
So at the moment the connect just accepts non-200 return codes for S3. At least it means the server is reachable, which is the purpose of the function, I guess.

rdimaio (approved):

PR #7752, opened 2025-05-28
Rewrite of TestBinRucio using pytest fixtures

### LINKED ISSUE #7641: Update test/test_bin_rucio to use fixtures where applicable
### Description

Extension of #6529 

Take class attributes in TestBinRucio (like self.user, self.def_rse) and replace them with test fixtures. 

### Motivation

`test/test_bin_rucio` relies on a lot of custom logic for setting up RSEs, accounts, etc, and does not use things like features to make new objects (so most tests are dirty). 

Ideally most of these objects in tests should be replaced with fixtures, or the tests otherwise marked as dirty where the test cannot be completely cleaned up

This would also be a good chance to clean up a lot of extra print statements and unused variables. 

### Change

_No response_

voetberg:
Closes #7641 

List of changes:
* Isolate tests instead of using a class setup stage
* Remove calls to cli used for setup (e.g. rucio upload, rucio add did)
* Use tempfile were possible
* Check functionality with a py client call instead of trusting output
* Remove format strings and extra print statements where possible
* Split up tests testing multiple things

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -312,7 +312,7 @@ def random_account(vo: str) -> "Iterator[InternalAccount]":
 
 
 @pytest.fixture
-def random_account_factory(vo: str):
+def random_account_factory(vo: str) -> "Iterator[InternalAccount]":
     import random
     import string
 
@@ -325,7 +325,7 @@ def random_account_factory(vo: str):
 
     made_accounts = []
 
-    def make_account():
+    def make_account() -> InternalAccount:
         account = InternalAccount(''.join(random.choice(string.ascii_lowercase) for _ in range(10)), vo=vo)
         made_accounts.append(account)
         if os.environ.get('SUITE') == 'client':
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import functools
+import os
 import re
 import traceback
 from os import environ
@@ -288,17 +289,61 @@ def random_account(vo: str) -> "Iterator[InternalAccount]":
     import random
     import string
 
+    from rucio.client import Client
     from rucio.common.types import InternalAccount
     from rucio.core.account import add_account, del_account
     from rucio.db.sqla import models
     from rucio.db.sqla.constants import AccountType
     from rucio.tests.common_server import cleanup_db_deps
 
-    account = InternalAccount(''.join(random.choice(string.ascii_uppercase) for _ in range(10)), vo=vo)
-    add_account(account=account, type_=AccountType.USER, email=f'{account.external}@email.com')
-    yield account
-    cleanup_db_deps(model=models.Account, select_rows_stmt=models.Account.account == account)
-    del_account(account)
+    account = InternalAccount(''.join(random.choice(string.ascii_lowercase) for _ in range(10)), vo=vo)
+
+    if os.environ.get('SUITE') == 'client':
+        c = Client(vo=vo)
+        c.add_account(account=account.external, type_="user", email=f'{account.external}@email.com')
+        yield account
+        c.delete_account(account=account.external)
+
+    else:
+        add_account(account=account, type_=AccountType.USER, email=f'{account.external}@email.com')
+        yield account
+        cleanup_db_deps(model=models.Account, select_rows_stmt=models.Account.account == account)
+        del_account(account)
+
+
+@pytest.fixture
+def random_account_factory(vo: str):
+    import random
+    import string
+
+    from rucio.client import Client
+    from rucio.common.types import InternalAccount
+    from rucio.core.account import add_account, del_account
+    from rucio.db.sqla import models
+    from rucio.db.sqla.constants import AccountType
+    from rucio.tests.common_server import cleanup_db_deps
+
+    made_accounts = []
+
+    def make_account():
+        account = InternalAccount(''.join(random.choice(string.ascii_lowercase) for _ in range(10)), vo=vo)
+        made_accounts.append(account)
+        if os.environ.get('SUITE') == 'client':
+            c = Client(vo=vo)
+            c.add_account(account=account.external, type_="user", email=f'{account.external}@email.com')
+        else:
+            add_account(account=account, type_=AccountType.USER, email=f'{account.external}@email.com')
+        return account
+
+    yield make_account
+
+    for account in made_accounts:
+        if os.environ.get('SUITE') == 'client':
+            c = Client(vo=vo)
+            c.delete_account(account=account.external)
+        else:
+            cleanup_db_deps(model=models.Account, select_rows_stmt=models.Account.account == account)
+            del_account(account)
 
 
 @pytest.fixture(scope="module")
--- a/tests/temp_factories.py
+++ b/tests/temp_factories.py
@@ -317,11 +317,18 @@ class TemporaryDidFactory:
                 scope = scope.external
             # Remove rules associated with the dids
             for rule in self.client.list_associated_rules_for_file(scope=scope, name=name):
-                try:
-                    self.client.delete_replication_rule(rule_id=rule['id'], purge_replicas=True)
-                except UnsupportedOperation:
-                    # The rule has a child rule so we cannot delete it
-                    pass
+
+                def remove_rules(rule_id: str) -> None:
+                    try:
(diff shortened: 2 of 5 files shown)


rdimaio:

tests/test_upload.py
```diff
@@ -60,26 +61,13 @@ def scope(vo, containerized_rses, test_scope, mock_scope):
...
-        ),
-    ],
-    indirect=True,
-)
+@pytest.mark.parametrize("file_config_mock", [
+    {},  # Use rucio.cfg as-is.
+    pytest.param(
+        {  # Remove "account" from the "[client]" section.
+            "removes": [('client', 'account')]
+        }, marks=pytest.mark.skipif('SUITE' in os.environ and os.environ['SUITE'] == 'multi_vo', reason="See https://github.com/rucio/rucio/issues/7394"))
```
rdimaio, line 69: I saw the linked issue, but I see that it's closed and it's not immediately clear why we should skip this test based on it - could you elaborate a bit more in the reason?

tests/conftest.py
```diff
@@ -288,17 +289,61 @@ def random_account(vo: str) -> "Iterator[InternalAccount]":
...
     import string
 
+    from rucio.client import Client
+    from rucio.common.types import InternalAccount
+    from rucio.core.account import add_account, del_account
+    from rucio.db.sqla import models
+    from rucio.db.sqla.constants import AccountType
+    from rucio.tests.common_server import cleanup_db_deps
+
+    account = InternalAccount(''.join(random.choice(string.ascii_lowercase) for _ in range(10)), vo=vo)
```
rdimaio, line 299: nit: I think this is fine, but I guess this could break if it were to be tested when the schema max account length is less than 10. It miiight make sense to make the max range at most `ACCOUNT_LENGTH` from the schema used, but honestly I think it's fine like this

tests/conftest.py
```diff
@@ -288,17 +289,61 @@ def random_account(vo: str) -> "Iterator[InternalAccount]":
...
     from rucio.tests.common_server import cleanup_db_deps
 
-    account = InternalAccount(''.join(random.choice(string.ascii_uppercase) for _ in range(10)), vo=vo)
-    add_account(account=account, type_=AccountType.USER, email=f'{account.external}@email.com')
-    yield account
-    cleanup_db_deps(model=models.Account, select_rows_stmt=models.Account.account == account)
-    del_account(account)
+    made_accounts = []
+
+    def make_account():
```
rdimaio: please add a return type hint

tests/conftest.py
```diff
@@ -288,17 +289,61 @@ def random_account(vo: str) -> "Iterator[InternalAccount]":
...
+
+    else:
+        add_account(account=account, type_=AccountType.USER, email=f'{account.external}@email.com')
+        yield account
+        cleanup_db_deps(model=models.Account, select_rows_stmt=models.Account.account == account)
+        del_account(account)
+
+
+@pytest.fixture
+def random_account_factory(vo: str):
```
rdimaio: please add a return type hint

tests/conftest.py
```diff
@@ -288,17 +289,61 @@ def random_account(vo: str) -> "Iterator[InternalAccount]":
...
+
+    yield make_account
+
+    for account in made_accounts:
+        if os.environ.get('SUITE') == 'client':
+            c = Client(vo=vo)
+            c.delete_account(account=account.external)
+        else:
+            cleanup_db_deps(model=models.Account, select_rows_stmt=models.Account.account == account)
+            del_account(account)
```
rdimaio, line 346: Could this reuse the `random_account` fixture to avoid duplication?

tests/temp_factories.py
```diff
@@ -222,14 +293,38 @@ def upload_client(self) -> UploadClient:
...
+        for scope, name in self.created_dids:
+            if isinstance(scope, InternalScope):
+                scope = scope.external
+            # Remove rules associated with the dids
+            for rule in self.client.list_associated_rules_for_file(scope=scope, name=name):
+                try:
+                    self.client.delete_replication_rule(rule_id=rule['id'], purge_replicas=True)
+                except UnsupportedOperation:
+                    # The rule has a child rule so we cannot delete it
+                    pass
```
rdimaio: can we query all child rules and delete them so that we can delete this one too?

tests/test_download.py
```diff
@@ -753,3 +754,74 @@ def test_download_traceless(rucio_client):
     client = DownloadClient(client=rucio_client, logger=None, tracing=True)
     assert client.logger is not None
     assert client.tracing is True
+
+
+# Adapted from tests/test_bin_rucio download tests
```
rdimaio: I'm a bit confused by these 'adapted' comments - do they mean that:
1. we have both `download` tests in `test_bin_rucio` and here, and the tests here are based on the ones in `test_bin_rucio` (in which case - could you expand on why this was done?)
2. or, this test was originally in `test_bin_rucio` and it was then moved here. In this case, I guess the comment might not be needed

rdimaio (commented):
nit: typos in commit message of commit 9e0511f76fc427fd3e9e870c38868afe5add9c8d (behavoir -> behaviour, approiate -> appropriate) and 28710f627c97350377998e821c3f1eb9356c6838 (accessable -> accessible)

tests/test_download.py
voetberg: They are replacing tests in `test_bin_rucio` that did not test the `bin` functionality, but instead the py client functionality. The only changes were changing the setup to use fixtures, and instead directly using the py client function. They functionality was otherwise not tested, so it seemed more appropriate to move the test into download. To verify the kwarg mapping it may be prudent to add a version of this tests that just checks that the arg is accepted.

tests/conftest.py
voetberg: Yeah this is from the original `random_account` functionality, and it hasn't caused problems thus far so I didn't see a reason to change it

tests/test_upload.py
voetberg: Ah, the reason this shows up as a change is bc the linting was yelling about spacing on it. I haven't a clue about this test in particular. I can look into it though

tests/conftest.py
voetberg: No, `random_account` returns a single account when called as a fixture, and trying to make multiple accounts with it will result in the same account name multiple times. As it is a fixture, it cannot be called as a normal function with would make a different name each time.

An alternate would be to make `random_account` callable (e.g. each time it gets called it creates a new account) and it would make `random_account_factory` redundant.

tests/test_download.py
rdimaio: Thanks for clarifying 😄 do you think it's best to keep these comments in for now or do you think they might become confusing in the long term? I think that, if we want to keep them in, it might be a good idea to expand them with this information you just mentioned, just to provide more context

tests/test_upload.py
rdimaio: Ah no worries, if it was already there then it's fine, I didn't notice that in the diff, sorry!

tests/temp_factories.py
```diff
@@ -222,14 +293,45 @@ def upload_client(self) -> UploadClient:
...
+
+                def remove_rules(rule_id: str) -> None:
+                    try:
+                        self.client.delete_replication_rule(rule_id=rule_id, purge_replicas=True)
+                    except UnsupportedOperation:
+                        # The rule has a child rule so we cannot delete it
+                        rule_info = self.client.get_replication_rule(rule['id'])
+                        child_rules = rule_info['child_rule_id']
+                        # Try to remove the child rules recursively
+                        remove_rules(child_rules)
```
rdimaio: Should this be singular? it seems to be a single `child_rule`

rdimaio (dismissed):

tests/test_download.py
voetberg: Hmmm. To be honest I left those comments mostly for the reviewer's benefit, so as long as the logic on moving it tracks I can just remove them

rdimaio: I personally think they could be removed as they might cause confusion in future

voetberg: Removed

tests/temp_factories.py
voetberg: It is now!

rdimaio (approved):

Issue #7750, opened 2025-05-28
Conveyor tests failing, checksum mismatch issues


rdimaio:
## Status
Pending ticket with FTS: https://its.cern.ch/jira/browse/FTS-2286, ETA: ~2-3 weeks (end of September)

### Description

Likely related to fts container that got updated to 3.14.1

The conveyor test is failing in multiple separate branches for the same reasons:

https://github.com/rucio/rucio/actions/runs/15272179167/job/42950137778
https://github.com/rucio/rucio/actions/runs/15295224544/job/43022643167
https://github.com/rucio/rucio/actions/runs/15270964347/job/42954480981
https://github.com/rdimaio/rucio/actions/runs/15275127865/job/42959579553

### Steps to reproduce

Run the CI

rdimaio:
https://fts.web.cern.ch/fts/releases.html#FTS-3.14.0 FTS 3.14 was released on 2025-04-30

https://github.com/rucio/containers/releases/tag/37.3.0 Rucio 37.3.0 was released and built on 2025-05-12, and we did not see any issues at that point. (https://hub.docker.com/layers/rucio/test-fts/37.3.0/images/sha256-afc6917237f39ac57aa890dad43ab7739e6608c02e1e9001a9544f7d3a00d627)

Issues began with the 37.4.0 build (https://hub.docker.com/layers/rucio/test-fts/37.4.0/images/sha256-6347a544ddb46fc14f582e7045ba89f4f09d0c4578708712d95d1dd1da93aeb8), 2025-05-26. The only FTS update in the meantime was 3.14.1: https://fts.web.cern.ch/fts/releases.html#FTS-3.14.1

rdimaio:
These 3.14.x changes to FTS seem relevant to what is going on here https://gitlab.cern.ch/fts/fts3/-/commit/23a8e16b40e6c33389bfac99abf4276006656efa#d580eb4e1fc79ccf8b1b068c403d7823374e7012

rdimaio:
At least one of the failures is a simple fix. `test_checksum_validation` is failing because the checksum validation error message changed in 3.14.1: https://gitlab.cern.ch/fts/fts3/-/commit/23a8e16b40e6c33389bfac99abf4276006656efa#d580eb4e1fc79ccf8b1b068c403d7823374e7012_795_807

And the assertion is now failing:
```
tests/test_conveyor.py:1603: in test_checksum_validation
    assert 'User and destination checksums do not match' in request['err_msg']
[...]

PR #7755, opened 2025-06-01
Documentation: quote all single-line descriptions Fix #7754

### LINKED ISSUE #7754: Add the value of each OpenAPI description field in double quotes
### Description

As described in https://github.com/rucio/documentation/issues/567, the OpenAPI one-line values of the description fields should better be double quoted to avoid invalid parsing. I suggest making this change everywhere to bypass for now test problems with Git hooks but we should also address this in https://github.com/rucio/documentation/issues/567 more holistically to have a permanent solution.

### Steps to reproduce

.

### Rucio Version

_No response_

### Additional Information

_No response_

Geogouz:
**Suggested change**
Wrap every one-line description: value in double quotes across our OpenAPI specs.

**Why**
We could have quoted only the fields that currently contain reserved YAML characters, but applying the rule everywhere makes the style self-evident for future authors, likely avoiding subtle parse errors.

**Next step**
Add validation to enforce either proper escaping or mandatory double-quotes so the rule stays consistent (see https://github.com/rucio/documentation/issues/567).

### DIFF
--- a/lib/rucio/client/downloadclient.py
+++ b/lib/rucio/client/downloadclient.py
@@ -1323,7 +1323,7 @@ class DownloadClient:
     def _resolve_one_item_dids(self, item: dict[str, Any]) -> "Iterator[dict[str, Any]]":
         """
         Resolve scopes or wildcard DIDs to lists of full did names:
-        
+
         Parameters
         ----------
         item :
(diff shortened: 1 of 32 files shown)


rdimaio (approved):
CI errors are due to unrelated issue: https://github.com/rucio/rucio/issues/7750

PR #7759, opened 2025-06-03
Core: refactor SCHEME_MAP to avoid circular dependency; 

### LINKED ISSUE #7758: Refactor `SCHEME_MAP` into `transfer`
`SCHEME_MAP` is the only thing that's causing the `constants` module to rely on another internal module, because it needs to access the config to partially populate the map. This is causing some circular dependency issues in https://github.com/rucio/rucio/pull/7747.

rdimaio:
fixes #7758

### DIFF
--- a/lib/rucio/common/constants.py
+++ b/lib/rucio/common/constants.py
@@ -16,8 +16,6 @@ import enum
 from collections import namedtuple
 from typing import Literal, get_args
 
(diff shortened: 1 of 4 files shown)


rdimaio:
Rebased in latest revision

bari12 (approved):

PR #7760, opened 2025-06-03
Database: cx oracle upgrade; Fix #6467

### LINKED ISSUE #6467: cx_oracle -> python-oracledb dependency upgrade
The `cx_oracle` version (8.3) used in Rucio is from 2021 (https://oracle.github.io/python-cx_Oracle/) and there has been a new major release since then, under the name `python-oracledb:` https://python-oracledb.readthedocs.io/en/latest/user_guide/appendix_c.html

Their docs provide an upgrade guide: https://python-oracledb.readthedocs.io/en/latest/user_guide/appendix_c.html#steps-to-upgrade-to-python-oracledb - however, there are a few considerations to keep in mind, like the deprecations listed here https://python-oracledb.readthedocs.io/en/latest/api_manual/deprecations.html#deprecations. 

This is not urgent, but it seems that `cx_oracle` will not be updated any further (https://cx-oracle.readthedocs.io/en/latest/index.html) in favour of `python-oracledb`, so it might be worth working on this upgrade at some point.

Geogouz:
- Replaced the old `cx_oracle` extra with the new `oracledb` driver in setup utilities
- Updated server requirements
- Revised default Oracle connection strings to use the `oracle+oracledb://` dialect
- Instant Client removal from the Dockerfiles (needs to be handled also at the containers side)
    _Regarding this, I assumed that the Dockerfile previously installed the Oracle Instant Client libraries since the `cx_oracle` package is a Python wrapper around Oracle’s OCI libraries and so, the native libraries (provided by the Instant Client RPMs) had to be available at runtime. With the migration to `oracledb`, these libraries are no longer needed, so I cleared that part._
- Tests seem to be fine (https://github.com/Geogouz/rucio/actions/runs/15416890108/job/43381695039#step:4:1967)
- I hope the universe doesn't explode

### DIFF
--- a/requirements/requirements.server.txt
+++ b/requirements/requirements.server.txt
@@ -57,7 +57,7 @@ cryptography==44.0.2
     #   pyjwt
     #   pyspnego
     #   requests-kerberos
-cx-oracle==8.3.0
+oracledb==3.1.1
     # via -r requirements.server.in
 decorator==5.2.1
     # via
--- a/etc/alembic.ini.template
+++ b/etc/alembic.ini.template
@@ -18,7 +18,7 @@
 # Choose one of the databases+schema below
 # Oracle, PostgreSQL, MySQL/MariaDB are fully supported
 
-sqlalchemy.url=oracle://user:pass@servicename
+sqlalchemy.url=oracle+oracledb://user:pass@servicename
 version_table_schema=rucio
 
 #sqlalchemy.url=postgresql+psycopg://rucio:rucio@psql-hostname:5432/rucio
--- a/etc/alembic_offline.ini.template
+++ b/etc/alembic_offline.ini.template
@@ -18,7 +18,7 @@
 # Choose one of the databases+schema below
 # Oracle, PostgreSQL, MySQL/MariaDB are fully supported
 
-#sqlalchemy.url=oracle://user:pass@servicename
+#sqlalchemy.url=oracle+oracledb://user:pass@servicename
 #version_table_schema=rucio
 
 #sqlalchemy.url=postgresql+psycopg://rucio:rucio@psql-hostname:5432/rucio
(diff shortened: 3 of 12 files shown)


requirements/requirements.server.txt
```diff
@@ -57,7 +57,7 @@ cryptography==44.0.2
     #   pyjwt
     #   pyspnego
     #   requests-kerberos
-cx-oracle==8.3.0
+oracledb==3.1.1
```
maany, line 60: This would replace the client used by sql alchemy in the production to communicate with oracle. Need to check if there are special hints provided to sqla for oracle and whether they will still be supported by the new client.

rdimaio:

bari12:
Since @0xquark is working on the same file, I think you need to rebase again.
To me this look all good and on the integration deployment we should see rather quickly if it works or not.

Geogouz:
Ready

rdimaio:
CI likely failing due to https://github.com/rucio/rucio/issues/7812, please rebase

Geogouz:
> CI likely failing due to #7812, please rebase

Oki, done

etc/docker/test/fedora35.Dockerfile
rdimaio, line 1: You accidentally re-added the `fedora35` Dockerfile, which was removed as part of https://github.com/rucio/rucio/pull/7786

Geogouz: Wooops. Force-pushed the last commit with the file removed.

rdimaio (approved):
LGTM, but Mayank raised a good point, we should check that before we go ahead with this

Geogouz:
Rebased

requirements/requirements.server.txt
rdimaio: Checked with Martin on this:

> In the end it should be simple to change
https://github.com/rucio/rucio/blob/02f32e1ef166681511c4e37e6cf5df051b415549/lib/rucio/core/lock.py#L136
from oracle to oracle+oracledb

Let's be aware of this over the next release

PR #7767, opened 2025-06-11
Testing: Implement Docker Buildx with Actions Cache for Autotest and Integration tests caching

### LINKED ISSUE #7667: Re-Factor test infrastructure
### Description

Main problems
- There a lot of scripts that make the CI workflow complicated
- Sometimes upstream dockerfiles/docker-compose are used, sometimes dockerhub container tags are used for testing
- Testing of LTS versions is buggy and bound to divert due to problem above

First we need to clarify a list of (low-hanging) goals to achieve during the 38 roadmap.

### Motivation

_No response_

### Change

_No response_

0xquark:
Ref: #7667 

Ref: https://codimd.web.cern.ch/s/3Uy5yFOQU

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -45,6 +45,19 @@ jobs:
         with:
           path: dev/rucio
           fetch-depth: 0
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@b5ca514318bd6ebac0fb2aedd5d36ec1b5c232a2 # v3.10.0
+        with:
+          driver-opts: network=host
+      - name: Cache Docker layers
+        uses: actions/cache@5a3ec84eff668545956fd18022155c47e93e2684 # v4.2.3
+        with:
+          path: /tmp/.buildx-cache
+          key: ${{ runner.os }}-buildx-autotest-${{ matrix.cfg.DIST }}-${{ matrix.cfg.PYTHON }}-${{ hashFiles('etc/docker/test/*.Dockerfile', '**/requirements*.txt') }}
+          restore-keys: |
+            ${{ runner.os }}-buildx-autotest-${{ matrix.cfg.DIST }}-${{ matrix.cfg.PYTHON }}-
+            ${{ runner.os }}-buildx-autotest-${{ matrix.cfg.DIST }}-
+            ${{ runner.os }}-buildx-autotest-
       - name: Select tag for rucio containers
         shell: bash
         run: |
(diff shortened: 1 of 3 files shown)


Geogouz:
Nice. Should the commit though include the fixes keyword for that issue? Maybe we should leave the issue open in case we decide to do further enhancements.

Also, instead of CI, a valid [label](https://github.com/rucio/rucio/labels) should be better used I guess among the purple ones according to the [contribution guide](https://rucio.github.io/documentation/contributing/#4-commit-your-changes).

.github/workflows/integration_tests.yml
```diff
@@ -45,6 +45,19 @@ jobs:
         with:
           path: dev/rucio
           fetch-depth: 0
+      - name: Set up Docker Buildx
```
rdimaio, line 48: The steps for caching seem to be the same in both `autotest` and `integration_tests`. Can we implement this as a reusable workflow to reduce duplication?

For reference:
- https://docs.github.com/en/actions/sharing-automations/reusing-workflows
- https://docs.github.com/en/actions/sharing-automations/avoiding-duplication

tools/test/build_images.py
```diff
@@ -82,12 +93,19 @@ def build_images(matrix, script_args):
 
             args = ()
             env = {"DOCKER_BUILDKIT": "1"}
+            
+            # Use docker buildx build (or podman build if USE_PODMAN is set)
+            if use_podman:
+                docker_cmd = ['docker', 'build']
```
rdimaio: Not sure, but should this be
```suggestion
                docker_cmd = ['podman', 'build']
```

0xquark: I think i missed this somehow (even in the comment i wrote `podman build`) Thanks.

.github/workflows/integration_tests.yml
0xquark: I managed to use a composite action to set up Docker buildx and caching at a common place. But preserving the cache or setting environment variables for buildx happens during the image build step, which is different for autotests and integration tests. So, I couldn't apply the same setup to both. Since this is done through shell command, it's also tricky to reuse from other workflows with the same setup. I could make another composite action for this, but that would just add more complexity.

rdimaio (dismissed):

0xquark:
@rdimaio PR is still in draft - The caching is not properly working after the composite action. Even after the cache is restored and some of the layers builds from cache, the requirements are still building from scratch. Looking into this rn.

rdimaio:
> @rdimaio PR is still in draft - The caching is not properly working after the composite action. Even after the cache is restored and some of the layers builds from cache, the requirements are still building from scratch. Looking into this rn.

We can also revert to the previous version without composite action and merge that, and possibly do the composite action at a later point

0xquark:
> > @rdimaio PR is still in draft - The caching is not properly working after the composite action. Even after the cache is restored and some of the layers builds from cache, the requirements are still building from scratch. Looking into this rn.
> 
> We can also revert to the previous version without composite action and merge that, and possibly do the composite action at a later point

I'll take a look for a while, if nothing works - we revert it back and merge it on monday.

0xquark:
Composite Actions changes the build context and file metadata somehow even if the content is same which invalidates the docker hash for files therefore files and directories with docker `COPY` command rebuilds from scratch as the hash doesn't match anymore with the cache hash. The only solution to this is recognizing which metadata is being changed and try to set it as constant before the build but then this would be just extra code and it would not make sense if in case to avoid duplication of code, we're now adding more code to maintain then before and making it more complex. I think it makes to merge this now and have the caching included in our PRs and then work on improving it. 

Ref: https://stackoverflow.com/questions/49932954/why-is-the-copy-instruction-causing-a-cache-miss-in-my-docker-build
https://docs.docker.com/build/cache/invalidation/

rdimaio (approved):

PR #7768, opened 2025-06-12
Enforce that dataset and files cannot be at the same level during file upload

### LINKED ISSUE #7638: “files + subdirectories” check problem in `_recursive` of `uploadclient.py`
### Description

Trying to understand/explain in docstrings what users can expect exactly from `uploadclient.py`, I noticed couple potential problems. One issue is inside the `_recursive` method (which is actually a method name that should be changed): 

https://github.com/rucio/rucio/blob/4618b6bc894b3f8554115a6ff58bea415295ff8e/lib/rucio/client/uploadclient.py#L840-L849

Even though we currently mention (and I guess expect) "A folder only can have either other folders inside or files, but not both of them", the code only does the “both files and subdirectories” check for the very first directory it walks (i.e., when `i == 0`). In other words, at the top level, if the directory has both files and subdirectories, it raises an error. But for deeper nested subdirectories (when `i > 0`), the code never checks this condition:

```
for i, (root, dirs, fnames) in enumerate(os.walk(path)):
    if len(dirs) > 0 and len(fnames) > 0 and i == 0:
        # Error out only on the very first directory
        raise InputValidationError('A container can only have either collections or files, not both')
    if len(fnames) > 0:
        # Treat the directory as a dataset
        ...
    elif len(dirs) > 0:
        # Treat the directory as a container
        ...
```

For deeper subdirectories (`i > 0`), the code simply goes into `if len(fnames) > 0` (which treats that subdirectory as a dataset) or `elif len(dirs) > 0` (which treats that subdirectory as a container). If a deeper subdirectory does contain both files and subdirectories, it still falls into whichever of those two branches matches first and no error is raised.

Is this the expected? Can't this lead to an inconsistent structure?

### Steps to reproduce

.

### Rucio Version

37.0.0rc4

### Additional Information

_No response_

voetberg:
Closes #7638

### DIFF
--- a/lib/rucio/client/uploadclient.py
+++ b/lib/rucio/client/uploadclient.py
@@ -867,7 +867,7 @@ class UploadClient:
         This method iterates over the provided items, each describing a local path and
         associated upload parameters, checks that each item has a valid path and RSE, and
         computes basic file details such as size and checksums. If the item is a directory
-        and `recursive` is set, the method calls `_recursive` to traverse subdirectories,
+        and `recursive` is set, the method calls `_collect_files_recursive` to traverse subdirectories,
         creating or attaching them as Rucio datasets or containers.
 
         Parameters
@@ -935,7 +935,7 @@ class UploadClient:
                     logger(logging.WARNING,
                            'Skipping %s because it has no files in it. Subdirectories are not supported.' % dname)
             elif os.path.isdir(path) and recursive:
-                files.extend(cast("list[FileToUploadWithCollectedInfoDict]", self._recursive(item)))
+                files.extend(cast("list[FileToUploadWithCollectedInfoDict]", self._collect_files_recursive(item)))
             elif os.path.isfile(path) and not recursive:
                 file = self._collect_file_info(path, item)
                 files.append(file)
@@ -1324,10 +1324,7 @@ class UploadClient:
         if self.tracing:
             send_trace(trace, self.client.trace_host, self.client.user_agent)
 
-    def _recursive(
-            self,
-            item: "FileToUploadDict"
-    ) -> list["FileToUploadWithCollectedAndDatasetInfoDict"]:
+    def _collect_files_recursive(self, item: "FileToUploadDict") -> list["FileToUploadWithCollectedAndDatasetInfoDict"]:
         """
         Recursively inspects a folder and creates corresponding Rucio datasets or containers.
 
--- a/lib/rucio/client/uploadclient.py
+++ b/lib/rucio/client/uploadclient.py
@@ -1379,10 +1379,9 @@ class UploadClient:
         if path and isinstance(path, str):
             if path[-1] == '/':
                 path = path[0:-1]
-            i = 0
             path = os.path.abspath(path)
             for root, dirs, fnames in os.walk(path):
-                if len(dirs) > 0 and len(fnames) > 0 and i == 0:
+                if len(dirs) > 0 and len(fnames) > 0:
                     self.logger(logging.ERROR, 'A container can only have either collections or files, not both')
                     raise InputValidationError('Invalid input folder structure')
                 if len(fnames) > 0:
@@ -1403,7 +1402,6 @@ class UploadClient:
                 elif len(dirs) == 0 and len(fnames) == 0:
                     self.logger(logging.WARNING, 'The folder %s is empty, skipping' % root)
                     continue
-                i += 1
         # if everything went ok, replicate the folder structure in Rucio storage
         for dataset in datasets:
             try:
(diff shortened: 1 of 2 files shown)


lib/rucio/client/uploadclient.py
```diff
@@ -837,7 +837,7 @@ def _send_trace(self, trace: "TraceDict") -> None:
         if self.tracing:
             send_trace(trace, self.client.trace_host, self.client.user_agent)
 
-    def _recursive(self, item: "FileToUploadDict") -> list["FileToUploadWithCollectedAndDatasetInfoDict"]:
+    def _collect_files_recursive(self, item: "FileToUploadDict") -> list["FileToUploadWithCollectedAndDatasetInfoDict"]:
```
rdimaio: nit: I would have done the rename in a separate commit for clarity

tests/test_upload.py
```diff
@@ -444,3 +447,73 @@ def test_upload_registration_fail(rse, scope, upload_client_registration_fail, f
...
+    }, True),  # One valid container, another invalid container
+    ({
+        "container": {"container2": {"dataset": {"dataset2": ["file1", {"dataset3": ["file2"]}]}}},
+    }, True),  # Same issue but further down the tree
+    ({"container": {"dataset": ["file1"], "dataset2": ["file2"]}}, False),  # One valid container
+    ({
+        "container": {"dataset": ["file1"], "dataset2": ["file2"]},
+        "container2": {"dataset": ["file1"], "dataset2": ["file2"]}
+    }, False)  # Two valid containers
+])
```
rdimaio: Can you use the `ids=` feature to move the comments here into ids?

e.g.

https://github.com/rucio/rucio/blob/33f1aa716f7032bf4b1e9888532332bb643a352b/tests/rucio/common/test_types.py#L23-L37

This way, if one of these test cases fails, the test logs show up the ID of the failing case, which would make it simpler to debug it.

tests/test_upload.py
```diff
@@ -444,3 +447,73 @@ def test_upload_registration_fail(rse, scope, upload_client_registration_fail, f
...
+@pytest.mark.parametrize("structure,failure", [
+    ({"container": ["file1", "file2", {"dataset": ["file3", "file4"]}]}, True),  # Cannot have both files and datasets in a container
+    ({
+        "container": {
+            "dataset": ["file1", "file2"]
+            },
+        "container2": {
+            "dataset": ["file1", {"dataset2": ["file1"]}]
+        }
+    }, True),  # One valid container, another invalid container
```
rdimaio: I would name the invalid container `bad_container` or something similar to clarify which one is the invalid one (and maybe also the naming the valid containers as `good_container` / `valid_container` or something along those lines)

tests/test_upload.py
```diff
@@ -444,3 +447,73 @@ def test_upload_registration_fail(rse, scope, upload_client_registration_fail, f
...
+    ({
+        "container": {"container2": {"dataset": {"dataset2": ["file1", {"dataset3": ["file2"]}]}}},
+    }, True),  # Same issue but further down the tree
+    ({"container": {"dataset": ["file1"], "dataset2": ["file2"]}}, False),  # One valid container
+    ({
+        "container": {"dataset": ["file1"], "dataset2": ["file2"]},
+        "container2": {"dataset": ["file1"], "dataset2": ["file2"]}
+    }, False)  # Two valid containers
+])
+def test_upload_recursive(structure, failure, rse_factory, scope, upload_client, caplog):
```
rdimaio: I would change `failure` to something more descriptive like `should_fail`. I also think it might be better to make this a positive bool instead, something like `is_valid_structure`

tests/test_upload.py
```diff
@@ -444,3 +447,73 @@ def test_upload_registration_fail(rse, scope, upload_client_registration_fail, f
...
+        "container": {
+            "dataset": ["file1", "file2"]
+            },
+        "container2": {
+            "dataset": ["file1", {"dataset2": ["file1"]}]
+        }
+    }, True),  # One valid container, another invalid container
+    ({
+        "container": {"container2": {"dataset": {"dataset2": ["file1", {"dataset3": ["file2"]}]}}},
+    }, True),  # Same issue but further down the tree
```
rdimaio: same comment about `bad_container`

lib/rucio/client/uploadclient.py
voetberg: Addressed in latest force push

rdimaio:

rdimaio:
CI likely failing due to https://github.com/rucio/rucio/issues/7812, please rebase

rdimaio (approved):

PR #7777, opened 2025-06-17
Database: gateway/account, migrate to session context manager

rdimaio:
Part of:
- #7409 
- #7605


lib/rucio/gateway/account.py
Geogouz, line 1: Looks ok. There are couple import/docstring incosistencies but not introduced by this PR (e.g. in `get_account_info`, `list_accounts` or `get_usage_history`). Maybe we can handle them in a different one.

Geogouz (approved):
Looks good to me.

Issue #7716, opened 2025-05-11
Refactor current Metadata-related classes


Geogouz:
### Description

Current Metadata-related classes have couple issues including wrong OpenAPI definitions and confusing naming conventions for distinguishing single and multiple DIDs and Metadata insertions. For an easier transition to the new Metadata engine, we should better refactor a bit first the existing Classes, retaining the exact current functionality.

### Steps to reproduce

Visible in [`lib/rucio/web/rest/flaskapi/v1/dids.py`](https://github.com/rucio/rucio/blob/master/lib/rucio/web/rest/flaskapi/v1/dids.py)

### Rucio Version

37.2.0

voetberg:
@Geogouz Should this be closed as the two sub issues are resolved?

PR #7781, opened 2025-06-17
REST & API: `Meta` class refactoring and OpenAPI docstring update; Fix #7779

### LINKED ISSUE #7779: Merge `Meta` and `SingleMeta` web classes
### Description

Sub-issue of #7716

### Steps to reproduce

-

### Rucio Version

_No response_

### Additional Information

_No response_

Geogouz:
The main goal of this PR is to refactor the class responsible for handling single-DID metadata and to improve its API documentation. This is an intermediate step to make it easier to transition to a new system for handling metadata.

### DIFF
--- a/lib/rucio/web/rest/flaskapi/v1/dids.py
+++ b/lib/rucio/web/rest/flaskapi/v1/dids.py
@@ -22,7 +22,6 @@ from rucio.common.exception import (
     DatabaseException,
     DataIdentifierAlreadyExists,
     DataIdentifierNotFound,
-    Duplicate,
     DuplicateContent,
     FileAlreadyExists,
     FileConsistencyMismatch,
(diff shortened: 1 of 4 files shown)


lib/rucio/web/rest/flaskapi/v1/dids.py
```diff
@@ -1295,172 +1294,288 @@ def generate(vo):
...
         - name: scope_name
           in: path
-          description: "The scope and the name of the DID."
+          description: "The scope and the name of the DID (e.g., `scope:name`)."
+          required: true
+          style: simple
           schema:
             type: string
-          style: simple
         - name: plugin
```
rdimaio, line 1312: There's an existing docstring for `plugin` but I don't see a `plugin` parameter in this request, and there's a new `key` parameter that is not added to the docstring. I assume these are two different parameters, but just checking whether there's a mismatch here between the docstring name and parameter name

lib/rucio/web/rest/flaskapi/v1/dids.py
```diff
@@ -1295,172 +1294,288 @@ def generate(vo):
...
-          description: "The scope and the name of the DID."
+          description: "The scope and the name of the DID (e.g., `scope:name`)."
+          required: true
+          style: simple
           schema:
             type: string
+        - name: key
+          in: path
+          description: "The metadata key to set/update. If omitted, it defaults to `None` and Multi-key mode is used."
+          required: true
```
rdimaio, line 1421: Is it required if it can be omitted for multi-key mode?

lib/rucio/web/rest/flaskapi/v1/dids.py
```diff
@@ -1295,172 +1294,288 @@ def generate(vo):
...
           401:
-            description: "Invalid Auth Token"
+            description: "Unauthorized – invalid Auth Token."
           404:
-            description: "DID or key not found"
-          406:
-            description: "Not acceptable"
+            description: >
+              Not found – the specified DID or `key` does not exist, or no `key` query
+              parameter provided.
```
rdimaio, line 1566: 404 is fine here but since `key` is a required parameter, I feel like the right error code for 'no `key` query parameter provided' should be 400 or 422 (https://stackoverflow.com/questions/3050518/what-http-status-response-code-should-i-use-if-the-request-is-missing-a-required). I think it's fine to leave it like this too though, reading about it online it seems that 404 could be considered the right response in this use case.

lib/rucio/web/rest/flaskapi/v1/dids.py
```diff
@@ -1295,172 +1294,288 @@ def generate(vo):
...
+          405:
+            description: "Method Not Allowed – the 'key' parameter is not supported with DELETE."
           409:
-            description: "Feature is not in current database."
+            description: "Conflict – action not supported by the utilized metadata plugin."
         """
+        if key is not None:
+            return generate_http_error_flask(405,
+                                             'MethodNotAllowed',
+                                             'DELETE not allowing keys')
```
rdimaio, line 1581: I'm a bit confused here, why do we add a `key parameter if it cannot be used and it instead needs to be passed as part of `request.args`? I think it might be helpful to clarify this in the error message (e.g. passing key as parameter not allowed - please pass it as request body)

lib/rucio/web/rest/flaskapi/v1/dids.py
Geogouz: In that `GET` case, plugin is actually not a `path` parameter – it is an optional `query` parameter, retrieved later like this:

    plugin = request.args.get('plugin', default='DID_COLUMN')

Hence there is no need for it to appear in the function signature.

Geogouz:
Currently, the same `Meta` view class handles three routes that share a lot of logic:

	1. /<scope_name>/meta        (GET, POST, DELETE)
	2. /<scope_name>/meta/<key>  (POST only)
	
But this is going to be just a temporal state till we migrate to the new `Metadata` engine where things are going to become quite simpler and straightforward.

lib/rucio/web/rest/flaskapi/v1/dids.py
rdimaio: Makes sense; should we still add a note in the docstring for the `key` parameter?

lib/rucio/web/rest/flaskapi/v1/dids.py
Geogouz: OpenAPI requires that every path parameter is marked as `required: true`. For `single-key mode` where the route is `/<scope_name>/meta/<key>`, `required: true` is needed for the path parameter `key`. For `multi-key mode` where the route is `/<scope_name>/meta`, that block is basically irrelevant as `key` is automatically passed by default as `None` from:

    bp.add_url_rule('/<path:scope_name>/meta', defaults={'key': None}, view_func=meta_view, methods=['get', 'post', 'delete'])

Because both routes share the same Python function (and therefore one docstring) it can read as contradictory at first glance. The OpenAPI definition itself, however, is technically correct: a path parameter that exists must be required.

I tried to make it a bit clear adding "If omitted, it defaults to `None` and Multi-key mode is used.", but maybe we can also add an extra one‑liner to the description, something like:

_This key parameter applies only to the `/meta/<key>` endpoint (single‑key mode). When you call `/meta` without the extra path segment, multi‑key mode is used and this parameter is not present._

lib/rucio/web/rest/flaskapi/v1/dids.py
Geogouz: The scope of the parent issue https://github.com/rucio/rucio/issues/7716 is to merge/simplify/reduce current classes/methods (to allow for clearer transition) but leaving current functionality intact as much as possible in order not to affect current use cases. The new Metadata endpoints shall respect though what you pointed out.

lib/rucio/web/rest/flaskapi/v1/dids.py
rdimaio: Yeah, I think adding this information in the docstring would be helpful for future reference

lib/rucio/web/rest/flaskapi/v1/dids.py
Geogouz: Including the `key` parameter in the definition `delete(self, scope_name, key=None):` cannot be avoided because of the `defaults={'key': None}` where Flask always injects it as an extra keyword argument:

    bp.add_url_rule('/<path:scope_name>/meta', defaults={'key': None}, view_func=meta_view, methods=['get', 'post', 'delete'])
    
As in the `GET` method, in practice, a non-None key is currently not possible to reach that part for `DELETE` requests, because right now only `POST` offers such routing:

    bp.add_url_rule('/<path:scope_name>/meta/<key>', view_func=meta_view, methods=['post', ])

however, I placed those checks there as a defensive kinda measure in case someone extended in the future (and before we eventually remove completely those classes after the backwards compatibility period passes), the methods=['post', ] with also `get` or `delete`. Since this is indeed quite unlikely, we could remove those checks. But even if we leave those, I don't think it will do bad.

rdimaio: Thanks for the context, it makes sense to have this check then. Could you add a comment adding this context in the code itself?

lib/rucio/web/rest/flaskapi/v1/dids.py
Geogouz: I think current state should be ok with the added explanation about the key on top of the "if key is not None:" part.

rdimaio (approved):

PR #7790, opened 2025-06-21
Database: Remove `destroy_database.py` and adjust `reset_database.py`; Fix #7789

### LINKED ISSUE #7789: Do we need the `destroy_database.py`?
### Description

[`destroy_database.py`](https://github.com/rucio/rucio/blob/master/tools/destroy_database.py) is not used somewhere, and if someone does use it manually, I would suggest for simplification to adjust [`reset_database.py`](https://github.com/rucio/rucio/blob/master/tools/reset_database.py) slightly, to use that instead since those two files are almost identical.

On top of that, the role/naming of the involved methods `drop_everything()`, `destroy_database()` is also not clear unless one searches what the called functions actually do. So we should at least add some clarification about them.

### Steps to reproduce

.

### Rucio Version

37.5.0

### Additional Information

_No response_

Geogouz:
Suggested solution for #7789.

For clarity, I would also suggest to rename:

    drop_everything()  -->  to something like purge_db()
and

    destroy_database()  --> to something like drop_orm_tables()

But when we do that, tests will fail to import the new methods so we will have to update the client too.

With the introduced changes, current CI is not affected. Only manual invokes would need to adapt to the new flags.


tools/reset_database.py
```diff
@@ -13,28 +13,69 @@
...
-    if args.drop_everything:
-        drop_everything()
+    # ------------------------------------------------------------------
+    # 1. Decide how to reset
+    # ------------------------------------------------------------------
+    if args.purge_build or args.purge_stop:
+        drop_everything()  # Method should better be renamed to something like purge_db()
     else:
-        destroy_database()
+        destroy_database()  # Method should better be renamed to something like drop_orm_tables()
```
rdimaio: opened an issue for this https://github.com/rucio/rucio/issues/7795 - we can remove the comments imo

tools/reset_database.py
```diff
@@ -13,28 +13,69 @@
...
+        "-b", "--purge-build",
+        action="store_true",
+        help="Purge EVERYTHING (tables, constraints, schema) "
+             "and then rebuild a fresh schema with base VO + root account.",
+    )
+    g.add_argument(
+        "-s", "--purge-stop",
+        action="store_true",
+        help="Purge EVERYTHING and stop – do NOT recreate schema or accounts.",
+    )
```
rdimaio: Intituively `purge-stop` makes me think more that this would stop a purge in progress, rather than purging and then doing nothing else. I think it would be worth naming this to something clearer, like just `purge`. i think having `purge` and `purge-build` as options implies that `purge` doesn't build anything and just purges

rdimaio (commented):
LGTM but worth checking if anyone's workflow relies on using `destroy_database` programmatically (I hope not) before releasing this

rdimaio (approved):

PR #7803, opened 2025-06-26
add obsolete replica check to core/rse_counter

### LINKED ISSUE #4261: Move deletable and obsolete replica checks from probes into Abacus-RSE

martynia:
This PR converts `check_obsolete_replica`  probe code to a daemon using SQLAlchemy.  It also add handling of deletable replicas:

for obsolete replicas:

`rucio-abacus-rse-expired --run-once --obsolete`

for deletable replicas:

`rucio-abacus-rse-expired --run-once` 

N.B obsolete replica counts are included in expired replica counts.

### DIFF
--- a/lib/rucio/core/rse_counter_expired.py
+++ b/lib/rucio/core/rse_counter_expired.py
@@ -1,0 +1,74 @@
+# Copyright European Organization for Nuclear Research (CERN) since 2012
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from datetime import datetime, timezone
+from typing import TYPE_CHECKING
+
+from sqlalchemy import Select, func, select
+
+from rucio.core.rse import set_rse_usage
+from rucio.db.sqla import models
+from rucio.db.sqla.constants import OBSOLETE, ReplicaState
+from rucio.db.sqla.session import transactional_session
+
+if TYPE_CHECKING:
+    from sqlalchemy.orm import Session
+
+
+@transactional_session
+def check_obsolete_replicas(rse_id: str, *, session: "Session") -> None:
+    """
+    Get number of files and bytes used by obsolete replicas for a given RSE and
+    update RSE usage accordingly.
+
+    :param rse_id: id of the RSE to check.
+    :param session: Database session in use.
+    """
+
+    stmt = _select_query(
+    ).where(
+        models.RSEFileAssociation.rse_id == rse_id,
+        models.RSEFileAssociation.tombstone == OBSOLETE
+    )
+    bytes, files = session.execute(stmt).one()
+    set_rse_usage(rse_id, 'obsolete', bytes, None, files, session=session)
+
+
+@transactional_session
+def check_deletable_replicas(rse_id: str, *, session: "Session") -> None:
+    """
+    Get number of files and bytes used by deletable replicas for a given RSE and
+    update RSE usage accordingly. OBSOLETE replicas are included in the calculation.
+
+    :param rse_id: id of the RSE to check.
+    :param session: Database session in use.
+    """
+
+    stmt = _select_query(
+    ).where(
+        models.RSEFileAssociation.rse_id == rse_id,
+        models.RSEFileAssociation.tombstone < datetime.now(timezone.utc),
+        models.RSEFileAssociation.state.in_((ReplicaState.AVAILABLE, ReplicaState.UNAVAILABLE, ReplicaState.BAD))
+    )
+    bytes, files = session.execute(stmt).one()
+    set_rse_usage(rse_id, 'expired', bytes, None, files, session=session)
+
+
+def _select_query() -> Select:
+    query = select(
+        func.coalesce(func.sum(models.RSEFileAssociation.bytes), 0).label('bytes'),
+        func.count().label('files')
+    ).with_hint(
+        models.RSEFileAssociation, 'INDEX(replicas REPLICAS_RSE_ID_TOMBSTONE_IDX)', 'oracle'
+    )
+    return query
--- a/lib/rucio/daemons/abacus/rse_expired.py
+++ b/lib/rucio/daemons/abacus/rse_expired.py
@@ -1,0 +1,118 @@
+# Copyright European Organization for Nuclear Research (CERN) since 2012
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+This is a daemon to update RSE counters for deletable and obsolete replicas. The type of replica counted is determined
+by obsolete variable.
+"""
+
+import functools
+import logging
+import threading
+import time
+from typing import TYPE_CHECKING
+
+import rucio.db.sqla.util
+from rucio.common import exception
+from rucio.common.logging import setup_logging
+from rucio.core.rse import list_rses
+from rucio.core.rse_counter_expired import check_deletable_replicas, check_obsolete_replicas
+from rucio.daemons.common import HeartbeatHandler, run_daemon
+
+if TYPE_CHECKING:
+    from types import FrameType
+    from typing import Optional
+
+graceful_stop = threading.Event()
+DAEMON_NAME = 'abacus-rse-deletable'
+
+
+def rse_update(
+        once: bool = False,
+        obsolete: bool = False,
+        sleep_time: int = 3600
+) -> None:
+    """
+    Main loop to check and update the RSE Counters.
+    """
+    run_daemon(
+        once=once,
+        graceful_stop=graceful_stop,
+        executable=DAEMON_NAME,
+        partition_wait_time=1,
+        sleep_time=sleep_time,
+        run_once_fnc=functools.partial(
+            run_once,
+            obsolete=obsolete,
+        ),
+    )
+
+
+def run_once(
+        heartbeat_handler: HeartbeatHandler,
+        obsolete: bool,
+        **_kwargs: object
+) -> None:
+
+    start_time = time.time()
+    _, _, logger = heartbeat_handler.live()
+    rses = list_rses()  # NOQA
+    for rse in rses:
+        if graceful_stop.is_set():
+            break
+        start = time.time()
+        if obsolete:
+            check_obsolete_replicas(rse['id'])
+        else:
+            check_deletable_replicas(rse['id'])
+        logger(logging.DEBUG, '%s replica backlog query for RSE %s took %f s.' % ('Expired' if not obsolete else 'Obsolete', rse['id'], time.time() - start))
+
+    logger(logging.DEBUG, 'update of all RSEs took %f s.' % (time.time() - start_time))
+
+
+def stop(signum: "Optional[int]" = None, frame: "Optional[FrameType]" = None) -> None:
+    """
+    Graceful exit.
+    """
+
+    graceful_stop.set()
+
+
+def run(
+        once: bool = False,
+        obsolete: bool = False,
+        sleep_time: int = 3600
+) -> None:
+    """
+    Starts up the Abacus-RSE deletable/obsolete threads.
+    """
+    global DAEMON_NAME
+    if obsolete:
+        DAEMON_NAME = 'abacus-rse-obsolete'
+    setup_logging(process_name=DAEMON_NAME)
+
+    if rucio.db.sqla.util.is_old_db():
+        raise exception.DatabaseException('Database was not updated, daemon won\'t start')
+
+    if once:
+        logging.info('main: executing one iteration only')
+        rse_update(once, obsolete=obsolete)
+    else:
+        logging.info('main: starting the RSE usage thread')
+        thread = threading.Thread(target=rse_update, kwargs={'once': once, 'obsolete': obsolete, 'sleep_time': sleep_time})
+        thread.start()
+        logging.info('main: waiting for interrupts')
+        # Interruptible joins require a timeout.
+        while thread.is_alive():
+            thread.join(timeout=3.14)
--- a/bin/rucio-abacus-rse-expired
+++ b/bin/rucio-abacus-rse-expired
@@ -1,0 +1,45 @@
+#!/usr/bin/env python3
+# Copyright European Organization for Nuclear Research (CERN) since 2012
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Abacus rse is a daemon to update rse counters. This script can run independently 2 daemons, one for expired and one for OBSOLETE replicas.
+"""
+
+import argparse
+import signal
+
+from rucio.daemons.abacus.rse_expired import run, stop
+
+
+def get_parser():
+    """
+    Returns the argparse parser.
+    """
+    parser = argparse.ArgumentParser(description="The abacus-rse-obsolete daemon is responsible for updating the 'expired' or 'obsolete' RSE usage counter", formatter_class=argparse.RawDescriptionHelpFormatter)
+    parser.add_argument("--run-once", action="store_true", default=False, help='One iteration only')
+    parser.add_argument('--sleep-time', action="store", default=3600, type=int, help='Thread sleep time after each chunk of work (there is only one OBSOLETE replica thread)')
+    parser.add_argument('--obsolete', action="store_true", default=False, help='Count obsolete replicas rather than expired ones')
+    return parser
+
+
+if __name__ == "__main__":
+
+    signal.signal(signal.SIGTERM, stop)
+    parser = get_parser()
+    args = parser.parse_args()
+    try:
+        run(once=args.run_once, obsolete=args.obsolete, sleep_time=args.sleep_time)
+    except KeyboardInterrupt:
+        stop()
--- a/tests/test_abacus_rse_expired.py
+++ b/tests/test_abacus_rse_expired.py
@@ -1,0 +1,78 @@
+# Copyright European Organization for Nuclear Research (CERN) since 2012
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from datetime import datetime, timedelta, timezone
+
+import pytest
+from sqlalchemy import delete
+
+from rucio.core.replica import add_replica, get_replica, set_tombstone
+from rucio.core.rse import get_rse_usage
+from rucio.daemons.abacus.rse_expired import run
+from rucio.db.sqla import models
+from rucio.db.sqla.constants import OBSOLETE
+from rucio.db.sqla.session import get_session
+from rucio.tests.common import did_name_generator
+
+
+@pytest.mark.noparallel(reason='uses daemon, failing in parallel to other tests')
+class TestAbacusRSEObsolete:
+
+    def test_abacus_rse_obsolete(self, mock_scope, rse_factory, root_account):
+        """ ABACUS (RSE): Test update of RSE usage for obsolete replicas. """
+        # Get RSE usage of all sources
+        session = get_session()
+        for model in [models.UpdatedRSECounter, models.RSEUsage]:
+            stmt = delete(model)
+            session.execute(stmt)
+        session.commit()
+
+        # create an RSE and some replicas with a tombstone:
+        _, rse_id = rse_factory.make_mock_rse()
+        nbfiles = 3
+        for _ in range(nbfiles):
+            name = did_name_generator('file')
+            add_replica(rse_id, mock_scope, name, 4, root_account)
+            assert get_replica(rse_id, mock_scope, name)['tombstone'] is None
+            set_tombstone(rse_id, mock_scope, name)
+            assert get_replica(rse_id, mock_scope, name)['tombstone'] == OBSOLETE
+
+        # add one more expired replica on the RSE
+        name3 = did_name_generator('file')
+        add_replica(rse_id, mock_scope, name3, 27, root_account)
+        assert get_replica(rse_id, mock_scope, name3)['tombstone'] is None
+        set_tombstone(rse_id, mock_scope, name3, datetime.now(timezone.utc) - timedelta(hours=1))
+        assert get_replica(rse_id, mock_scope, name3)['tombstone'] is not None
+        # now one more replica, on a different RSE, but not obsolete:
+        _, rse_id2 = rse_factory.make_mock_rse()
+        name2 = did_name_generator('file')
+        add_replica(rse_id2, mock_scope, name2, 4, root_account)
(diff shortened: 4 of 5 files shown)


martynia:
The original Oracle probe stored selected data in `rse_usage`  with a source set to _'obsolete'_.  The `rse_counter` module operates on `rse_usage` and `rse_usage_history` tables with source set to _rucio_.  To keep _rucio_ generated counters separate from _obsolete_ counters and the current counter logic is problematic since a buffer table `updated_rse_counters`  has no _source_ column. 

I opted for independent counting of OBSOLETE replicas (`updated_rse_counters` table is not used). See my comment below.

martynia:
Obsolete replicas are handled independently of regular replicas. The counting system for regular replicas are not used for obsolete replicas; `abacus-rse` daemon counts are not affected by a new `abacus-obsolete-rse` daemon. In particular when obsolete replicas are discovered, they are not subtracted from regular replica counts.

The `abacus-obsolete-rse` daemon is started by `rucio-abacus-obsolete-rse script`.  The system updates `rse_usage` with source = 'obsolete' and optionally `rse_usage_history` tables. The `rse_usage`  update interval is configurable (default 1 hour).

rdimaio (commented):
Could you add some tests to ensure this works?

bin/rucio-abacus-obsolete-rse
```diff
@@ -0,0 +1,77 @@
...
+import signal
+
+from rucio.daemons.abacus.obsolete_rse import run, stop
+
+
+def get_parser():
+    """
+    Returns the argparse parser.
+    """
+    parser = argparse.ArgumentParser(description="The Abacus-Obsolete-RSE daemon is responsible for updating RSE usages for OBSOLETE replicas. It runs a custom query and does not use Abacus counters.", epilog='''
```
dchristidis: 1. I believe this would be better phrased as

    ```
    … for updating the 'obsolete' RSE usage counter.
    ```

    Then the remaining phrase is an implementation detail that should not be expose in the parser.

2. The entire epilogue is irrelevant and should be removed.

lib/rucio/core/rse_counter.py
```diff
@@ -183,3 +189,54 @@ def fill_rse_counter_history_table(*, session: "Session"):
...
+        rse_counter = session.execute(stmt).scalar_one()
+        rse_counter.used = rse.bytes
+        rse_counter.files = rse.files
+        rse_counter.updated_at = rse.updated_at
+    except NoResultFound:
+        models.RSEUsage(rse_id=rse.rse_id,
+                        used=rse.bytes,
+                        files=rse.files,
+                        updated_at=rse.updated_at,
+                        source='obsolete').save(session=session)
```
dchristidis: Can’t this function be replaced with a call to `set_rse_usage()`?

lib/rucio/core/rse_counter.py
```diff
@@ -11,16 +11,22 @@
...
 from sqlalchemy.exc import NoResultFound
+from sqlalchemy.sql.functions import coalesce
 
 from rucio.common.exception import CounterNotFound
 from rucio.db.sqla import filter_thread_work, models
+from rucio.db.sqla.constants import OBSOLETE
 from rucio.db.sqla.session import read_session, transactional_session
 
 if TYPE_CHECKING:
+    import datetime
```
dchristidis: I believe that `from datetime import datetime` would be preferable.

lib/rucio/core/rse_counter.py
```diff
@@ -183,3 +189,54 @@ def fill_rse_counter_history_table(*, session: "Session"):
     )
     for usage in session.execute(stmt).scalars().all():
         models.RSEUsageHistory(rse_id=usage['rse_id'], used=usage['used'], files=usage['files'], source=usage['source']).save(session=session)
+
+
+@read_session
+def check_obsolete_replicas(*, session: "Session") -> "Sequence[Row[tuple[Any, Any, Any, datetime.datetime]]]":
```
dchristidis: `Sequence` should be used only for parameters to allow for different sequence-like structures. For the return value, you know its exact type already. Consequently, use `list`.

lib/rucio/core/rse_counter.py
```diff
@@ -183,3 +189,54 @@ def fill_rse_counter_history_table(*, session: "Session"):
...
+    # Replicas subquery
+    repl_subq = (select(models.RSEFileAssociation.rse_id, func.count(1).label("files"), func.sum(models.RSEFileAssociation.bytes).label("bytes")).where(
+                    models.RSEFileAssociation.tombstone == OBSOLETE
+                  ).group_by(models.RSEFileAssociation.rse_id).subquery())
+
+    # Full query with outer_join()
+    stmt = select(rse_subq.c.rse_id, coalesce(repl_subq.c.files, 0).label("files"), coalesce(repl_subq.c.bytes, 0).label("bytes"), func.now().label("updated_at")).outerjoin(
+        repl_subq,
+        rse_subq.c.rse_id == repl_subq.c.rse_id
+    )
```
dchristidis: Please follow the [Style Guide](https://rucio.cern.ch/documentation/developer/dev_style_guide/#sqlalchemy-syntax).

bin/rucio-abacus-obsolete-rse
```diff
@@ -0,0 +1,77 @@
...
+  ------
+      files: 1
+      used: 213.481 kB
+      rse: MOCK
+      updated_at: 2018-12-03 08:58:33
+      source: rucio
+  ------
+''', formatter_class=argparse.RawDescriptionHelpFormatter)
+    parser.add_argument("--run-once", action="store_true", default=False, help='One iteration only')
+    parser.add_argument("--enable-history", action="store_true", default=False, help='Record RSE usage into history table every hour.')
```
dchristidis: I believe this option should be removed?

lib/rucio/daemons/abacus/obsolete_rse.py
```diff
@@ -0,0 +1,115 @@
...
+def run_once(
+        heartbeat_handler: HeartbeatHandler,
+        **_kwargs: object
+) -> None:
+    _, _, logger = heartbeat_handler.live()
+
+    # Select a bunch of rses for to update for this worker
+    start = time.time()  # NOQA
+    rses = check_obsolete_replicas()
+    logger(logging.DEBUG, 'RSE collection query time %f ' % (time.time() - start))
```
dchristidis: The comment, the variable name, and the log message appear to be misleading.

dchristidis (changes requested):
> the current counter logic is problematic since a buffer table updated_rse_counters has no source column.

There is no need for a buffered approach because only this daemon updates the `obsolete` RSE usage counter, whereas the `rucio` RSE usage counter gets updated simultaneously and repeatedly by various components.

I also think it would be more intuitive to call this `abacus-rse-obsolete`.

lib/rucio/core/rse_counter.py
martynia: Yes, it possibly could. It would be more elegant. But what should we do with the `free` parameter? It is optional in the schema, so that's why it worked in Oracle probe and I reimplemented it SQLAlchemy. Obsolete replicas are a subset of `source='rucio'` replicas, so one could call `get_rse_usage() `before just to get a number of free bytes. And I would skip `updated_at` timestamp in this case.

dchristidis: `free` should be `None`. `updated_at` is handled automatically.

lib/rucio/core/rse_counter.py
martynia: not used anymore.

bin/rucio-abacus-obsolete-rse
martynia: Done.

lib/rucio/core/rse_counter.py
martynia: `Sequence` is the type returned by SQLAlchemy result set. If changed to `list`, I'm getting:

Found 1 new problems in /home/runner/work/rucio/rucio/lib/rucio/core/rse_counter.py

  - 1 errors with message """Type "Sequence[Row[Tuple[Any, Any, Any]]]" is not assignable to return type "list[Row[tuple[Any, Any, Any]""""

dchristidis: I stand corrected. Please switch it back.

I’m surprised that the inner function very clearly returns a list, but then the outer function widens its type. This must be a conscious decision.
https://github.com/sqlalchemy/sqlalchemy/blob/39149a25837c045c19368e7b16597aafaef58ce1/lib/sqlalchemy/engine/result.py#L543
https://github.com/sqlalchemy/sqlalchemy/blob/39149a25837c045c19368e7b16597aafaef58ce1/lib/sqlalchemy/engine/result.py#L1389

martynia: OK, Done.

lib/rucio/daemons/abacus/obsolete_rse.py
martynia: corrected that.

lib/rucio/core/rse_counter.py
martynia: I fixed this (hopefully)

martynia:
> Could you add some tests to ensure this works?

I added 1 test for `check_obsolete_replicas`

bin/rucio-abacus-obsolete-rse
martynia: done.

lib/rucio/core/rse_counter.py
```diff
@@ -183,3 +188,41 @@ def fill_rse_counter_history_table(*, session: "Session"):
...
+    rse_subq = select(
+        models.RSE.id.label("rse_id")
+    ).where(
+        models.RSE.deleted == false()
+    ).subquery()
+
+    # Replicas subquery
+    repl_subq = select(
+        models.RSEFileAssociation.rse_id,
+        func.count(1).label("files"),
```
dchristidis: I would recommend calling `func.count()` without an argument.

In raw SQL, `COUNT(*)` and `COUNT(1)` are identical. In SQLAlchemy, it introduces a bind variable.

```python
>>> print(select(func.count()).select_from(RSEFileAssociation))
SELECT count(*) AS count_1 
FROM atlas_rucio.replicas
>>> print(select(func.count(1)).select_from(RSEFileAssociation))
SELECT count(:count_2) AS count_1 
FROM atlas_rucio.replicas
```

It is slightly harder to read when you have an exception or you’re trying to debug queries from the database side.

lib/rucio/daemons/abacus/obsolete_rse.py
```diff
@@ -0,0 +1,112 @@
...
+from rucio.core.rse import set_rse_usage
+from rucio.core.rse_counter import check_obsolete_replicas
+from rucio.daemons.common import HeartbeatHandler, run_daemon
+
+if TYPE_CHECKING:
+    from types import FrameType
+    from typing import Optional
+
+graceful_stop = threading.Event()
+DAEMON_NAME = 'abacus-obsolete-rse'
```
dchristidis: The top-level executable is `rucio-abacus-rse-obsolete`, but many internal constructs have `obsolete[-_]rse`.

dchristidis (commented):
When I attempted to rewrite the probe in SQLAlchemy (see https://github.com/rucio/probes/pull/145), I opted to have the query be RSE-specific and have an external loop. In my mind, this was the ‘unit of work’.

I don’t know which approach is better. I don’t have a particularly strong opinion. The one advantage of the RSE-specific implementation that I can think of is that you can have some useful logging messages (e.g. elapsed time per RSE).

lib/rucio/core/rse_counter.py
martynia: Done that.

lib/rucio/daemons/abacus/obsolete_rse.py
martynia: I refactored the daemon file.

martynia:
> When I attempted to rewrite the probe in SQLAlchemy (see [rucio/probes#145](https://github.com/rucio/probes/pull/145)), I opted to have the query be RSE-specific and have an external loop. In my mind, this was the ‘unit of work’.
> 
> I don’t know which approach is better. I don’t have a particularly strong opinion. The one advantage of the RSE-specific implementation that I can think of is that you can have some useful logging messages (e.g. elapsed time per RSE).

I was unaware that your attempt existed. I followed a suggestion to move the probe to counters, as I did, following the way in was done in the original probe. I'm wondering if an SQL-engine driven query is not faster than n RSE queries.

dchristidis:
> I'm wondering if an SQL-engine driven query is not faster than n RSE queries.

Maybe there is an argument to be made that goes beyond performance. For the `obsolete` RSE usage counter it shouldn’t make a noticeable difference. However, issue #4261 states that the same will be done for `expired`. For ATLAS, the probe can take over 10 minutes. So that’s a long period where there will be no log messages and the heartbeat will become stale. Effectively, the daemon will appear as stuck.

martynia:
> Maybe there is an argument to be made that goes beyond performance. For the `obsolete` RSE usage counter it shouldn’t make a noticeable difference. However, issue #4261 states that the same will be done for `expired`. For ATLAS, the probe can take over 10 minutes. So that’s a long period where there will be no log messages and the heartbeat will become stale. Effectively, the daemon will appear as stuck.

OK, I have adapted your code and created a comparison function. The times are comparable, however I have just a few replicas in my test environment. The full query uses outer join just to exclude deleted RSEs. To make a fair comparison, do you know how to filter out those in `list_rses()` ? The comment for this method is quite vague.

dchristidis:
> The full query uses outer join just to exclude deleted RSEs. To make a fair comparison, do you know how to filter out those in list_rses() ?

Forgive me, I’m not sure I understand. This function already excludes deleted RSEs.

martynia:
OK, if it does, that’s OK, without specifying filters, I mean.
J.

martynia:
I created a small comparison task of 2 methods of getting OBSOLETE replicas, one based on the original probe code, which creates a full query for all RSEs (which is in this PR) and another one which does that on a RSE by RSE basis and a look over RSEs is made in Python (based on https://github.com/rucio/probes/pull/145).
 This is the comparison task code:
```python
from sqlalchemy import func, select, delete

from rucio.db.sqla.session import get_session, transactional_session
from rucio.core.rse import list_rses, set_rse_usage
from rucio.core.rse_counter import check_obsolete_replicas
from rucio.db.sqla.constants import OBSOLETE

if TYPE_CHECKING:
    from collections.abc import Sequence

@transactional_session
def check_obsolete_replicas_by_rse(rse, *, session: "Session") -> None:

    query = select(
        func.coalesce(func.sum(models.RSEFileAssociation.bytes), 0).label('bytes'),
        func.count().label('files') 
    ).with_hint(
        models.RSEFileAssociation, 'INDEX(replicas REPLICAS_RSE_ID_TOMBSTONE_IDX)', 'oracle'
    ).where(
        models.RSEFileAssociation.rse_id == rse['id'],
        models.RSEFileAssociation.tombstone == OBSOLETE
    )

    bytes, files = session.execute(query).one()
    set_rse_usage(rse['id'], 'obsolete', bytes, None, files)   

def clear_rse_usage():

    session = get_session()
    for model in [models.RSEUsage]:
        stmt = delete(model)
        session.execute(stmt)
        session.commit()

def compare_obsolete_replica_checks():
    import time
    
    clear_rse_usage()

    start = time.time()  # NOQA
    rows = check_obsolete_replicas() #  full query for all RSEs
    for row in rows:
        set_rse_usage(row.rse_id, 'obsolete', row.bytes, None, row.files)

    time_1 = time.time() - start
    # print("time elapse - full query", time_1)

    clear_rse_usage()

    start = time.time()
    rses = list_rses()
    for rse in rses:
        check_obsolete_replicas_by_rse(rse)
    time_2 = time.time() - start
    # print("time elapse - query by RSE", time_2)

    return time_1, time_2

def replica_checks_stats(runs = 100):
    import statistics
    values_1 =[]
    values_2 =[]

    for run in range(runs):
        t1, t2 = compare_obsolete_replica_checks()
        values_1.append(t1)
        values_2.append(t2)

    mean_1, stdev_1 = statistics.mean(values_1), statistics.stdev(values_1)
    mean_2, stdev_2 = statistics.mean(values_2), statistics.stdev(values_2)

    print(f"Elapsed time mean {mean_1:.5f}, sigma {stdev_1:.5f}  for n= {runs} runs")
    print(f"Elapsed time mean {mean_2:.5f}, sigma {stdev_2:.5f}  for n= {runs} runs")
    print(f"Speed ratio: {mean_1/mean_2:.2f}")
```

A run in my test environment for a bunch of RSEs and 2 obsolete replicas gives:
```
>>>from rucio.daemons.abacus.rse_obsolete_compare import replica_checks_stats
>>>replica_checks_stats()

Elapsed time mean 0.03411, sigma 0.00610  for n= 100 runs
Elapsed time mean 0.05700, sigma 0.00681  for n= 100 runs
Speed ratio: 0.60
```

There are fluctuations, but the test above shows that the full query is 40% faster than the RSE one. Now a decision has to be made if we sacrifice speed for being able to monitor a daemon with a higher granularity and get early commits for successfully analysed RSEs.

dchristidis:
One thing to note about your result is that the overhead increases with the number of RSEs. It also decreases with the number of replicas, but for `obsolete` this is typically low.

martynia:
It increases with number of commits per RSE. With a single commit (which denies a purpose) the results are comparable. What are we going to do now ? Have you got an opportunity to test this in a production environment ?

dchristidis:
I’m afraid I do not understand. What do you mean by ‘commits per RSE’?

martynia:
`check_obsolete_replicas_by_rse` is within a transactional section, so every invocation of it triggers a commit.

dchristidis:
But why would the number of commits per RSE not be exactly one?

dchristidis:
Picture four (overly simplified) scenarios:

 1. You use 1 query to fetch 10 rows
 2. You use 10 queries to fetch 10 rows (1 row per query)
 3. You use 1 query to fetch 1M rows
 4. You use 10 queries to fetch 1M rows (100K rows per query)

Now replace the number of queries with the number of RSEs, and the number of rows with the number of replicas that match the appropriate state (i.e. obsolete or expired). And note that the table in question is indexed, so it’s not a full-table scan per query.

The measurements you provide (which in no way do I doubt) fall under scenario 2. Where I do have my doubts is whether they will hold for scenario 4.

I did no extensive measurements when I moved the ATLAS probes from the ones in the master branch to ones similar to the PR previously mentioned, but I can remember that I did not observe an obvious increase in run time.

bari12:
For me, after thinking about it, I am also coming down on the making the RSE the unit of work. It's just easier to monitor. We have this issue in other daemons where operators complain that they don't know if the thing is stuck, or just working, I think it is fair to give up on some performance (though, not clear if we would) for this reason.

Geogouz:
Could you please rebase and resolve the conflicts?

martynia:
@dchristidis I addressed the comments, rebased and all tests are passing. Can we go forward and approve this PR?

martynia:
@Geogouz changes applied as requested.

lib/rucio/daemons/abacus/rse_obsolete.py
```diff
@@ -0,0 +1,104 @@
...
+
+def run_once(
+        heartbeat_handler: HeartbeatHandler,
+        **_kwargs: object
+) -> None:
+
+    start_time = time.time()
+    rses = list_rses()  # NOQA
+    for rse in rses:
+        _, _, logger = heartbeat_handler.live()
```
Geogouz: `logger` is assigned inside the loop, but used after the loop uncoditionally. So if `list_rsed()` returns `[]`, theres gonna be an unbound error.

lib/rucio/core/rse_counter_obsolete.py
```diff
@@ -0,0 +1,47 @@
...
+        func.coalesce(func.sum(models.RSEFileAssociation.bytes), 0).label('bytes'),
+        func.count().label('files')
+    ).with_hint(
+        models.RSEFileAssociation, 'INDEX(replicas REPLICAS_RSE_ID_TOMBSTONE_IDX)', 'oracle'
+    ).where(
+        models.RSEFileAssociation.rse_id == rse_id,
+        models.RSEFileAssociation.tombstone == OBSOLETE
+    )
+    bytes, files = session.execute(query).one()
+    set_rse_usage(rse_id, 'obsolete', bytes, None, files)
```
Geogouz: Without passing the session here each RSE will make a separate transaction. Not a correctness break but it might bring DB overhead.

Geogouz:
From what I understood, this PR's scope touches only the obsolete-checks side of the issue. So the deletable part is not addressed, right?

lib/rucio/core/rse_counter_obsolete.py
martynia: Fixed.

lib/rucio/daemons/abacus/rse_obsolete.py
martynia: I modified the code and fixed it.

bziemons:
please incorporate your review fixes into the original commits, possibly using `git commit` with `--fixup` and `git rebase` with `--autosquash`. Consult Dimitrios' presentation from Feb, 5th this year for guidance https://indico.cern.ch/event/1641324/

martynia:
@bziemons I'll squash it etc, as soon as @Geogouz is happy with the fixes.

martynia:
@Geogouz The main effort was to move an OBSOLETE probe to a daemon. Adding a deletable replica handling would not be a big job, a question is if it should be done here or as a new PR.

- a deletable replica is one with a (any) tombstone, it should include OBSOLETE counts, right ?
- one could re-use some of the core code to modify a SELECtion slightly, thus minimising code repetition
- deletable replica handling would require a new (very similar to the OBSOLETE one) daemon to be able to have independent frequencies. Alternatively an all-in-one daemon, handling both replica types in one go - this seems a bit heavy to me. 

Which route should I take ?

lib/rucio/core/rse_counter_expired.py
```diff
@@ -0,0 +1,72 @@
...
+    :param session: Database session in use.
+    """
+
+    stmt = _select_query(
+    ).where(
+        models.RSEFileAssociation.rse_id == rse_id,
+        models.RSEFileAssociation.tombstone.is_not(None)
+    )
+    bytes, files = session.execute(stmt).one()
+    set_rse_usage(rse_id, 'deletable', bytes, None, files, session=session)
```
Geogouz: The `deletable` here sounds a bit problematic cause it writes to a new `rse_usage` source-key `deletable` (in `RSEUsage`, source is part of the primary key `rse_id`/`source`). So I guess it might break backward compatibility with existing consumers that still read `source='expired'`? Maybe it should remain `expired` or maybe be dual-written during some kind of migration? But I am missing what the end goal is/should be regarding that.

lib/rucio/daemons/abacus/rse_expired.py
```diff
@@ -0,0 +1,116 @@
...
+        obsolete: bool = False,
+        sleep_time: int = 3600
+) -> None:
+    """
+    Main loop to check and update the RSE Counters.
+    """
+    run_daemon(
+        once=once,
+        graceful_stop=graceful_stop,
+        executable=DAEMON_NAME,
```
Geogouz, line 52: I think running obsolete and deletable daemon modes at the same time could become operationally confusing because they share one heartbeat namespace. For example, what if one mode dies silently and the other keeps up?

lib/rucio/core/rse_counter_expired.py
martynia: This is a very good point. There is already a mechanism to count 'expired' replicas (`fill_rse_expired` in `rse.py`) but it is not used in rucio codebase (i.e in the main rucio repository) as far as my search is correct. My implementation is called by a daemon.
I can't see a reason why we would need both. Changing a key from 'deletable' to 'expired' is a small editing problem...

By the way, the query is almost identical, I'm checking for a not `None` value tombstone, the `rse.py` code is checking for a tombstone `value < utcnow()`. Can we have a tombstone value in future,(in which case those queries would not have the same outcome) ?

lib/rucio/daemons/abacus/rse_expired.py
martynia: This was something I did not fully understand. It should be 2 different daemons, then ? The main question remains though, do we need that daemon mode at all, taking into account there is code already in` rse.py` ?

martynia: Since a heartbeat handler takes an executable as an argument, is it not enough to set a DAEMON_NAME in `run()` separately for 2 cases ? Would this fix the issue ? I defined it as `global` there,  planning to assign a value to it based on `obsolete` variable value, but somehow forgot to do it.

lib/rucio/core/rse_counter_expired.py
Geogouz: Yes, and from what I can infer from https://github.com/rucio/rucio/blob/6187280beda6f0890005ffce74283802beeb25b4/lib/rucio/core/replica.py#L1693 future tombstones seem to be valid. So a positive delay creates future tombstones.  
Therefore, I can also imagine that the `models.RSEFileAssociation.tombstone.is_not(None)` predicate that is being used now may be problematic. Given #4261 is about moving probe logic into a daemon, my assumption would be to preserve legacy semantics.

lib/rucio/daemons/abacus/rse_expired.py
Geogouz: Yes, setting different heartbeat executable names per mode should fix I guess the namespace collision. It does not require two separate codebases/entrypoints.

Geogouz:
> a question is if it should be done here or as a new PR

Personally, I am fine with it as long as commits are well separated. Otherwise, we should reformulate the issue itself I guess.

martynia:
I implemented following changes:

1. changed a 'source' value stored in `rse_usage` to read 'expired' for deletable/expired replicas. It stays obsolete for OBSOLETE replicas
2. fixed the daemon name (`in run()`)
3. fixed the tombstone time comparison (to now() rather than NOT NULL.
4. modified comments in places to reflect changed deletable->expired

martynia:
Squashed and rebased.

PR #7815, opened 2025-07-02
Move CLI error handling to exception_handler where applicable. 

### LINKED ISSUE #7720: Create definition for "error" and "failure" in the CLI
### Description

Let me see if I can describe this properly. 

In the CLI, there are situation where errors are raised on the client side because the user has taken an invalid action. These can either be raised as errors and passed forward, or caught, added to the error log, and instead a "FAILURE" is returned. 

The difference between the two seems to have very little consequence for the user (the thing they tried to do does not happen, they are told why) but it changes how things are tested dramatically (does the actual error code get passed forward, or does it just throw a nondescript exitcode = 1?). 

This also has problems in the new CLI, because the new CLI does not return anything, just runs the execution. [See this comment for a description](https://github.com/rucio/rucio/pull/7679#issuecomment-2872844868) 

A concrete design philosophy should be outlined and adhered to to avoid confusion. 

### Steps to reproduce

N/A

### Rucio Version

_No response_

### Additional Information

_No response_

voetberg:
Closes #7720 

* Removes instances of a function in bin_legacy returning a FAILURE status when raising an error is more descriptive
* Raises a "InputValidationError" when conflicting arguments are given
* Raises a specific error when schemas cannot be validated or objects not found`
* Raises a general "RucioException" with a verbose description when no existing error matches
* Removes error catches that result in returning FAILURE
* Expands the "@exception_handler" decorator to catch input validation errors

Note: All exitcode=1's are still in place, just shifted to being handled in the exception_handler, instead of being handled twice.

Most of this is just mapping "Cannot use -x and -y" or "Must include one of -x, -y, -z" to "InputValidationError", which was present in upload and seemed to be working just fine. Tests didn't have to be touched.

### DIFF
--- a/lib/rucio/cli/bin_legacy/rucio.py
+++ b/lib/rucio/cli/bin_legacy/rucio.py
@@ -20,12 +20,10 @@ import os
 import signal
 import sys
 import time
-import traceback
 import unittest
 import uuid
 from copy import deepcopy
 from datetime import datetime
-from logging import DEBUG
 from typing import TYPE_CHECKING, Optional
 
 from rich.console import Console
@@ -45,13 +43,13 @@ from rucio.common.client import detect_client_location
 from rucio.common.config import config_get, config_get_float
 from rucio.common.constants import ReplicaState
 from rucio.common.exception import (
-    DIDFilterSyntaxError,
-    DuplicateCriteriaInDIDFilter,
     DuplicateRule,
     InputValidationError,
     InvalidObject,
     InvalidType,
+    RSENotFound,
     RucioException,
+    ScopeNotFound,
     UnsupportedOperation,
 )
 from rucio.common.extra import import_extras
@@ -109,8 +107,7 @@ def ping(args, client, logger, console, spinner):
     if server_info:
         print(server_info['version'])
         return SUCCESS
-    logger.error('Ping failed')
-    return FAILURE
+    raise RucioException('Ping failed')
 
 
 @exception_handler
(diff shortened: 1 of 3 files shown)


lib/rucio/cli/bin_legacy/rucio.py
```diff
@@ -468,35 +462,16 @@ def list_dids(args, client, logger, console, spinner):
...
-    except DIDFilterSyntaxError as error:
-        logger.error(error)
-        return FAILURE
-    except ValueError as error:
-        logger.error(error)
-        return FAILURE
-    except Exception as e:
-        logger.error(e)
-        return FAILURE
+    filters, type_ = parse_did_filter_from_string_fe(args.filter, name)
```
rdimaio, line 474: is it fine to not have these exception checks anymore?

lib/rucio/cli/bin_legacy/rucio.py
```diff
@@ -2067,22 +2021,8 @@ def add_lifetime_exception(args, client, logger, console, spinner):
...
-            if logger.level == DEBUG:
-                logger.exception(error_message)
-            else:
-                logger.error(error_message)
-        else:
-            logger.error(error_message)
-            logger.debug(traceback.format_exc())
-        return FAILURE
+
+    client.add_exception(dids=datasets, account=client.account, pattern='', comments=reason, expires_at=expiration)
```
rdimaio, line 2025: same here, is it fine to remove this try except logic?

lib/rucio/cli/bin_legacy/rucio.py
voetberg: All of these should be handled by the `exception_handler`. There's no additional logic besides logging, which the handler does.

rdimaio (approved):

lib/rucio/cli/bin_legacy/rucio.py
voetberg: This one is less straight forward, but I think it should also be fine and handled by the decorator. It removed the "Failure to submit exception", but the exception message from the client does include something of the type.

Issue #7822, opened 2025-07-07
Incorrect rendering of DiracClient docstring


maxnoe:
### Description

Bullet point list is not rendered correctly

### Steps to reproduce

Visit: https://rucio.github.io/documentation/html/client_api/diracclient.html

PR #7823, opened 2025-07-07
Clients: fix rst markup in DiracClient #7822

maxnoe:


rdimaio (approved):
nit: the commit message is a bit misleading, as the change does not just fix the markup but also adds the '* Attach metadata defined in ``parents_metadata``' line. I think it would be best to have the formatting changes be in a separate PR from the 'meaningful' changes

PR #7831, opened 2025-07-09
Remove references to depeciated meta endpoint #7821

### LINKED ISSUE #7821: API: `meta` endpoint should be removed in release 38
Ran into this comment: https://github.com/rucio/rucio/blob/6442f5e88941e4a2bc6c8cd63218d6ef69fc11f4/lib/rucio/web/rest/flaskapi/v1/main.py#L57-L61

I don't see an issue open for this. Is the plan still to remove this for release 38?

voetberg:
Closes #7821 

Should include a message in the release making note of this change - it could break things if people don't have their configs updated.

### DIFF
--- a/lib/rucio/web/rest/flaskapi/v1/main.py
+++ b/lib/rucio/web/rest/flaskapi/v1/main.py
@@ -14,7 +14,6 @@
 # limitations under the License.
 
 import importlib
-import logging
 from typing import TYPE_CHECKING
 
 from flask import Flask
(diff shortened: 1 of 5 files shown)


lib/rucio/web/rest/flaskapi/v1/main.py
```diff
@@ -57,12 +56,10 @@ def apply_endpoints(app: Flask, modules: "Iterable[str]") -> None:
     for blueprint_module in modules:
         # Legacy patch - TODO Remove in 38.0.0
         if blueprint_module == "meta":
-            logging.log(logging.WARNING, "Endpoint `meta` is depreciated and will be removed in future releases")
             blueprint_module = "meta_conventions"
```
rdimaio: should we also delete this `if` statement in general? i think I'd either:
- delete the statement as a whole, or
- keep both the statement and the warning message


I think deleting just the warning message could be confusing

voetberg: You're totally right - removed in latest force push

rdimaio (approved):

PR #7604, opened 2025-04-01
Core: Refactor VO 'def' usage into DEFAULT_VO and fix common.config circular import

rdimaio:
- b5626b3a584ad9ab0be182964de96ef9cbc1cad0: fix #7600
- ff907f8b24b4ec39ab0b0473624dd79209c17fe7: fix #7602

PR #7836, opened 2025-07-11
Policies: simplify registration of policy package algorithms #7405

### LINKED ISSUE #7405: Consider changing how policy package algorithms are registered, in order to avoid having to set them in the config
Policy package docs for reference: https://rucio.github.io/documentation/operator/policy_packages/

For example, for `pfn2lfn`:
- registration is done as part of `_module_init`
- initialization is done by first checking `policy.pfn2lfn` in the config; if it isn't passed, it falls back to the `vo` algorithm if supported, otherwise to `def`.

https://github.com/rucio/rucio/blob/e383672a7d5932bd6128c739980d88da9201ffb5/lib/rucio/rse/translation.py

When we instead get the `schema`, we just get the policy package module name and append `.schema` to it, and then try to import it:

https://github.com/rucio/rucio/blob/e383672a7d5932bd6128c739980d88da9201ffb5/lib/rucio/common/schema/__init__.py#L59-L92

This way, `policy.schema` doesn't need to be specifically configured.

I think it might be worth changing the logic for the policy package algorithms so that it's done in a similar way as `schema`, i.e. we put each algorithm in a 'predictable' location (e.g. `.pfn2lfn`) and attempt to import it first, and if it fails, we can fall back to the configuration option (e.g. `policy.pfn2lfn`). I think this would remove a bit of cfg boilerplate for most cases.

jamesp-epcc:
This PR allows algorithms to be implemented in policy packages simply by putting them in a function and module with a certain name. For example, an LFN2PFN algorithm can be placed in a function named `lfn2pfn` in a module named `lfn2pfn` in the policy package and Rucio will find it and use it as the default LFN2PFN algorithm. No registration or configuration is necessary. However other algorithms can still be used as before by specifying them in the configuration file or as RSE attributes.

Much of the complexity here is due to the requirement to support multi VO installations. In this case, each VO can have its own policy package, so we have to know the active VO to be able to decide which package's default algorithms should be used. In some cases the VO was not immediately available and had to be passed down through a few layers of code.

I will submit a documentation PR as well. However, my main concern is that this will give us 3 different ways of registering policy package algorithms - the original method using `get_algorithms` as used by DUNE and possibly others, the method currently described in the documentation involving subclassing a core algorithm from Rucio as used by ATLAS, and now this new method. In the interests of simplicity and consistency we should probably deprecate at least one of these methods.

### DIFF
--- a/lib/rucio/common/plugins.py
+++ b/lib/rucio/common/plugins.py
@@ -16,12 +16,13 @@ import importlib
 import logging
 import os
 from configparser import NoOptionError, NoSectionError
-from typing import TYPE_CHECKING, Any, TypeVar
+from typing import TYPE_CHECKING, Any, Optional, TypeVar
 
 from packaging.specifiers import SpecifierSet
 
 from rucio.common import config
 from rucio.common.client import get_client_vo
+from rucio.common.constants import DEFAULT_VO
 from rucio.common.exception import InvalidAlgorithmName, PolicyPackageIsNotVersioned, PolicyPackageVersionError
 from rucio.version import current_version
 
@@ -30,6 +31,8 @@ if TYPE_CHECKING:
     from types import ModuleType
 
     from rucio.common.types import LoggerFunction
+
+LOGGER = logging.getLogger('policy')
 
 PolicyPackageAlgorithmsT = TypeVar('PolicyPackageAlgorithmsT', bound='PolicyPackageAlgorithms')
 
@@ -74,11 +77,43 @@ class PolicyPackageAlgorithms:
     """
     _ALGORITHMS: dict[str, dict[str, 'Callable[..., Any]']] = {}
     _loaded_policy_modules = False
+    _default_algorithms: dict[str, 'Callable[..., Any]'] = {}
 
     def __init__(self) -> None:
         if not self._loaded_policy_modules:
             self._register_all_policy_package_algorithms()
             self._loaded_policy_modules = True
+
+    @classmethod
+    def _get_default_algorithm(cls: type[PolicyPackageAlgorithmsT], algorithm_type: str, vo: str = "") -> Optional['Callable[..., Any]']:
+        """
+        Gets the default algorithm of this type, if present in the policy package.
+        The default algorithm is the function named algorithm_type within the module named algorithm_type.
+        Returns None if no default algorithm present.
+        """
+        # check if default algorithm for this VO is already cached
+        type_for_vo = vo + "_" + algorithm_type
+        if type_for_vo in cls._default_algorithms:
+            return cls._default_algorithms[type_for_vo]
+
+        default_algorithm = None
+        try:
+            if vo == DEFAULT_VO:
+                vo = ''
+            package = cls._get_policy_package_name(vo)
+        except (NoOptionError, NoSectionError):
+            return default_algorithm
+
+        module_name = package + "." + algorithm_type
+        try:
+            module = importlib.import_module(module_name)
+
+            if hasattr(module, algorithm_type):
+                default_algorithm = getattr(module, algorithm_type)
+                cls._default_algorithms[type_for_vo] = default_algorithm
+        except ImportError:
+            LOGGER.info('Policy algorithm module %s could not be loaded' % module_name)
+        return default_algorithm
 
     @classmethod
     def _get_one_algorithm(cls: type[PolicyPackageAlgorithmsT], algorithm_type: str, name: str) -> 'Callable[..., Any]':
@@ -144,15 +179,17 @@ class PolicyPackageAlgorithms:
                     cls._try_importing_policy(vo['vo'])
 
     @classmethod
+    def _get_policy_package_name(cls: type[PolicyPackageAlgorithmsT], vo: str = "") -> str:
+        env_name = 'RUCIO_POLICY_PACKAGE' + ('' if not vo else '_' + vo.upper())
+        package = os.getenv(env_name, "")
+        if not package:
+            package = str(config.config_get('policy', 'package' + ('' if not vo else '-' + vo)))
+        return package
+
+    @classmethod
     def _try_importing_policy(cls: type[PolicyPackageAlgorithmsT], vo: str = "") -> None:
         try:
-            # import from utils here to avoid circular import
-
-            env_name = 'RUCIO_POLICY_PACKAGE' + ('' if not vo else '_' + vo.upper())
-            package = os.getenv(env_name, "")
-            if not package:
-                package = str(config.config_get('policy', 'package' + ('' if not vo else '-' + vo)))
-
+            package = cls._get_policy_package_name(vo)
             module = importlib.import_module(package)
             check_policy_module_version(module)
 
--- a/lib/rucio/common/utils.py
+++ b/lib/rucio/common/utils.py
@@ -46,7 +46,7 @@ import requests
 from typing_extensions import ParamSpec
 
 from rucio.common.config import config_get, config_get_bool
-from rucio.common.constants import BASE_SCHEME_MAP
+from rucio.common.constants import BASE_SCHEME_MAP, DEFAULT_VO
 from rucio.common.exception import DIDFilterSyntaxError, DuplicateCriteriaInDIDFilter, InputValidationError, InvalidType, MetalinkJsonParsingError, MissingModuleException, RucioException
 from rucio.common.extra import import_extras
 from rucio.common.plugins import PolicyPackageAlgorithms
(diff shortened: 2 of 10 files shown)


lib/rucio/common/plugins.py
```diff
@@ -74,12 +74,40 @@ class PolicyPackageAlgorithms:
...
+            if vo == 'def':
+                vo = ''
+            package = cls._get_policy_package_name(vo)
+            module_name = package + "." + algorithm_type
+            module = importlib.import_module(module_name)
+
+            if hasattr(module, algorithm_type):
+                default_algorithm = getattr(module, algorithm_type)
+                cls._default_algorithms[type_for_vo] = default_algorithm
+        except (NoOptionError, NoSectionError, ImportError):
```
voetberg: Could you add a logging statement here? It is useful to know when a policy package *isn't* loaded

voetberg (commented):
I don't think there's anything wrong with this PR, but it's hard to know what this change actually means. Can you show an example of registering a policy package under this method?

lib/rucio/common/plugins.py
jamesp-epcc: I have added one.

jamesp-epcc:
> I don't think there's anything wrong with this PR, but it's hard to know what this change actually means. Can you show an example of registering a policy package under this method?

Yes, I will include an example in the documentation update.

lib/rucio/common/plugins.py
```diff
@@ -74,12 +76,44 @@ class PolicyPackageAlgorithms:
...
 
+    @classmethod
+    def _get_default_algorithm(cls: type[PolicyPackageAlgorithmsT], algorithm_type: str, vo: str = "") -> Optional['Callable[..., Any]']:
+        """
+        Gets the default algorithm of this type, if present in the policy package.
+        The default algorithm is the function named algorithm_type within the module named algorithm_type.
+        Returns None if no default algorithm present.
+        """
+        # check if default algorithm for this VO is already cached
+        type_for_vo = algorithm_type + "-" + vo
```
rdimaio: The documentation https://rucio.github.io/documentation/operator/policy_packages/policy-package-development uses `vo_algorithm` as naming convention, so I think it would be best to use this convention here as well

lib/rucio/common/plugins.py
```diff
@@ -74,12 +76,44 @@ class PolicyPackageAlgorithms:
...
+
+        try:
+            module_name = package + "." + algorithm_type
+            module = importlib.import_module(module_name)
+
+            if hasattr(module, algorithm_type):
+                default_algorithm = getattr(module, algorithm_type)
+                cls._default_algorithms[type_for_vo] = default_algorithm
+        except ImportError:
+            LOGGER.warning('Policy package %s could not be loaded' % package)
```
rdimaio: From https://docs.python.org/3/library/exceptions.html


exception ImportError

Raised when the [import](https://docs.python.org/3/reference/simple_stmts.html#import) statement has troubles trying to load a module. Also raised when the “from list” in from ... import has a name that cannot be found.


So I think we should log `module` instead of `package`, and say that the module could not be imported

lib/rucio/common/plugins.py
```diff
@@ -74,12 +76,44 @@ class PolicyPackageAlgorithms:
...
+        Returns None if no default algorithm present.
+        """
+        # check if default algorithm for this VO is already cached
+        type_for_vo = algorithm_type + "-" + vo
+        if type_for_vo in cls._default_algorithms:
+            return cls._default_algorithms[type_for_vo]
+
+        default_algorithm = None
+        try:
+            if vo == 'def':
```
rdimaio: Please use `DEFAULT_VO`, see: https://github.com/rucio/rucio/pull/7604

lib/rucio/common/utils.py
```diff
@@ -390,17 +390,24 @@ class NonDeterministicPFNAlgorithms(PolicyPackageAlgorithms):
 
     _algorithm_type = 'non_deterministic_pfn'
 
-    def __init__(self) -> None:
+    def __init__(self, vo: str = 'def') -> None:
```
rdimaio: Please use `DEFAULT_VO`, see: https://github.com/rucio/rucio/pull/7604

lib/rucio/common/utils.py
```diff
@@ -510,15 +517,15 @@ def construct_non_deterministic_pfn_default(dsn: str, scope: Optional[str], file
 NonDeterministicPFNAlgorithms._module_init_()
 
 
-def construct_non_deterministic_pfn(dsn: str, scope: Optional[str], filename: str, naming_convention: Optional[str] = None) -> str:
+def construct_non_deterministic_pfn(dsn: str, scope: Optional[str], filename: str, naming_convention: Optional[str] = None, vo: str = 'def') -> str:
```
rdimaio: Please use `DEFAULT_VO`, see: https://github.com/rucio/rucio/pull/7604

lib/rucio/common/utils.py
```diff
@@ -551,17 +558,24 @@ class ScopeExtractionAlgorithms(PolicyPackageAlgorithms):
 
     _algorithm_type = 'scope'
 
-    def __init__(self) -> None:
+    def __init__(self, vo: str = 'def') -> None:
```
rdimaio: Please use `DEFAULT_VO`, see: https://github.com/rucio/rucio/pull/7604

lib/rucio/common/utils.py
```diff
@@ -642,9 +656,10 @@ def extract_scope_dirac(did: str, scopes: Optional['Sequence[str]']) -> 'Sequenc
 def extract_scope(
         did: str,
         scopes: Optional['Sequence[str]'] = None,
-        default_extract: str = 'def'
+        default_extract: str = 'def',
+        vo: str = 'def'
```
rdimaio: Please use `DEFAULT_VO`, see: https://github.com/rucio/rucio/pull/7604

lib/rucio/transfertool/fts3_plugins.py
```diff
@@ -85,7 +85,11 @@ def _collocation(collocation_func: 'Callable', hints: dict[str, Any]) -> dict[st
         """
         return {"collocation_hints": collocation_func(**hints)}
 
-    def _default(self, *hints: dict) -> dict:
+    def _default(self, hint_dict: dict[str, Any]) -> dict:
+        vo = hint_dict['vo'] if 'vo' in hint_dict else 'def'
```
rdimaio: Please use `DEFAULT_VO`, see: https://github.com/rucio/rucio/pull/7604

lib/rucio/core/rule.py
```diff
@@ -100,21 +100,22 @@ class AutoApprove(PolicyPackageAlgorithms):
 
     _algorithm_type = 'auto_approve'
 
-    def __init__(self, rule: models.ReplicationRule, did: models.DataIdentifier, session: 'Session') -> None:
+    def __init__(self, rule: models.ReplicationRule, did: models.DataIdentifier, session: 'Session', vo: str = 'def') -> None:
```
rdimaio: Please use `DEFAULT_VO`, see: https://github.com/rucio/rucio/pull/7604

lib/rucio/rse/translation.py
```diff
@@ -111,7 +117,8 @@ def __init__(
             self,
             rse: Optional[str] = None,
             rse_attributes: Optional["RSESettingsDict"] = None,
-            protocol_attributes: Optional[dict[str, Any]] = None
+            protocol_attributes: Optional[dict[str, Any]] = None,
+            vo: str = 'def'
```
rdimaio: Please use `DEFAULT_VO`, see: https://github.com/rucio/rucio/pull/7604

lib/rucio/common/plugins.py
jamesp-epcc: Thinking about it some more, I'm not actually sure if we should log this at all, except maybe at a very low level such as info or debug. This exception will happen if the module containing one of the default algorithms can't be loaded, but that's normal and expected if the package doesn't implement that algorithm. If the policy package itself can't be loaded then that should be picked up and logged elsewhere.

lib/rucio/common/plugins.py
jamesp-epcc: Yes, that makes sense.

rdimaio (commented):
LGTM for the main code, but it would be great if we could have some tests on this logic (perhaps in https://github.com/rucio/rucio/blob/master/tests/rucio/common/test_plugins.py)

bari12 (approved):

bari12:
I will merge this now for 38, but the tests indeed should be added.

rdimaio:
> I will merge this now for 38, but the tests indeed should be added.

Created https://github.com/rucio/rucio/issues/7881 to follow-up

PR #7838, opened 2025-07-12
conveyor: sslport only when use_ssl for reciever #7833

### LINKED ISSUE #7833: Conveyor receiver requires ssl port even if use_ssl is False
### Description

In conveyor receiver , if `use_ssl` is false it still requires a port for ssl.
https://github.com/rucio/rucio/blob/master/lib/rucio/daemons/conveyor/receiver.py#L157

Since `use_ssl` is always required config parameter , we can have the above line be inside if condition.

### Steps to reproduce

1. Run conveyor receiver with 
```
[messaging-fts3]
use_ssl = False
brokers = activemq1.example.org
username = your-username
password = your-password
nonssl_port = 61613
destination = /topic/transfer.fts_monitoring_complete
```
It raises an exception that it couldn't find the required parameter `port` (which is ssl port)

### Rucio Version

server:37.x

### Additional Information

_No response_

panta-123:
ssl port requirement only when use_ssl is True.
This change was made because when use_ssl is False ,  it  currently complains of not having ssl port but non_ssl port is already specified.
closes: #7833

### DIFF
--- a/lib/rucio/daemons/conveyor/receiver.py
+++ b/lib/rucio/daemons/conveyor/receiver.py
@@ -154,7 +154,8 @@ def receiver(
     except Exception:
         logging.info('could not find use_ssl in configuration -- please update your rucio.cfg')
 
-    port = config_get_int('messaging-fts3', 'port')
+    if use_ssl:
+        port = config_get_int('messaging-fts3', 'port')
     vhost = config_get('messaging-fts3', 'broker_virtual_host', raise_exception=False)
     if not use_ssl:
         username = config_get('messaging-fts3', 'username')


lib/rucio/daemons/conveyor/receiver.py
```diff
@@ -154,7 +154,8 @@ def receiver(
     except Exception:
         logging.info('could not find use_ssl in configuration -- please update your rucio.cfg')
 
-    port = config_get_int('messaging-fts3', 'port')
+    if use_ssl:
```
rdimaio, line 157: LGTM but leaving this comment to block until https://github.com/rucio/rucio/issues/7750 is fixed

dchristidis (commented):
I stumbled upon this accidentally and wished to offer two comments, though they can be ignored.

 1. The resulting structure feels suboptimal.
    ```python
    if condition:
        …
    if not condition:
        …
    ```

 2. Do we understand the motivation behind having both `port` and `nonssl_port`? The commit that introduced the new configuration option doesn’t offer any justification.

dchristidis:

Issue #7846, opened 2025-07-14
It seems like `fileclient.py` is not usable


Geogouz:
### Description

I think we currently don't have any ``/files/<scope>/<name>/rses`` endpoint for handling the single available `list_file_replicas(..)` method in the `fileclient.py`: 

If this client isn't some forgotten leftover, we should remove it. Otherwise, we should adjust the endpoint that this targets so that it works:

https://github.com/rucio/rucio/blob/d526f32f1e715a2684b39713b405ac13a03bce00/lib/rucio/client/fileclient.py#L46

### Steps to reproduce

Try to use the method.

### Rucio Version

37.7.0

Geogouz:
Maybe this endpoint is instead in use now for the intended functionality:

https://github.com/rucio/rucio/blob/d526f32f1e715a2684b39713b405ac13a03bce00/lib/rucio/web/rest/flaskapi/v1/dids.py#L2616-L2620

which is used by this method in `didclient.py`:

https://github.com/rucio/rucio/blob/d526f32f1e715a2684b39713b405ac13a03bce00/lib/rucio/client/didclient.py#L533-L538

rdimaio:
This was added as part of this commit in 2012 https://github.com/rucio/rucio/commit/56b465f4f22656f9882f8270702ead6c3de030f1#diff-3073ed46d274eafb8a6e43a6703509469508ba7a634d324352321b97ad290836

Looks like it used to hit this endpoint https://github.com/rucio/rucio/commit/56b465f4f22656f9882f8270702ead6c3de030f1#diff-1a4cb43102ef012f1472017a32662671bedb618bcc6ec67a2191db72dade1e3bR31-R57, which is essentially replicas.

The replicaclient has `list_replicas`:

https://github.com/rucio/rucio/blob/02f32e1ef166681511c4e37e6cf5df051b415549/lib/rucio/client/replicaclient.py#L155-L195

which seems to perform the same functionality.

IMO the `fileclient` looks to be a leftover that was needed for the 2012 demo mentioned in the commit message in which it was added, and doesn't seem to be needed anymore. Can someone that was around back then confirm/deny this (@bari12 , @mlassnig )? If this is indeed unused, we can just delete it

rdimaio:
We're not even importing it here: https://github.com/rucio/rucio/blob/master/lib/rucio/client/client.py

I think this is essentially deprecated, seems safe to remove.

rdimaio:
[...]

PR #7847, opened 2025-07-14
Documentation: FileClient docstring enhancement; Fix rucio#7841

### LINKED ISSUE #7841: Rucio Python Client docstrings enhancement (`FileClient`)
Sub-issue of https://github.com/rucio/documentation/issues/363 targeting specifically the `FileClient`.

Geogouz:
Handles the sub-issue about the FileClient Docstrings

### DIFF
--- a/lib/rucio/client/fileclient.py
+++ b/lib/rucio/client/fileclient.py
@@ -23,25 +23,62 @@ from rucio.common.utils import build_url
 
 
 class FileClient(BaseClient):
-    """Dataset client class for working with dataset"""
+    """
+    Client for retrieving replica information for file DIDs.
+
+    This lightweight client exposes a single helper to query the Rucio
+    catalogue for all physical replicas of a given file.  It is typically
+    used by command line tools and scripts which need to inspect where a file
+    is currently stored.
+    """
 
     BASEURL = 'files'
 
     def list_file_replicas(self, scope: str, lfn: str) -> list[dict[str, Any]]:
         """
-        List file replicas.
+        Return all known replica locations of a file DID.
+
+        The method issues a GET request to the ``/files/<scope>/<name>/rses``
+        endpoint of the Rucio REST API. The server replies with a JSON list
+        describing each RSE where the file is currently present. Each entry in
+        the returned list contains the RSE name and may include additional
+        attributes such as the physical file name (PFN), file size, checksum
+        and replica state.
+
+        _**Note:**_ This method is currently not available.
 
         Parameters
         ----------
-        scope :
-            The scope of the file.
-        lfn :
-            The LFN
+        scope
+            The scope part of the file DID (e.g. ``"user.alice"``).
+        lfn
+            The logical file name of the DID.
 
         Returns
         -------
+        list[dict[str, Any]]
+            A list of dictionaries describing each replica. Each entry at least
+            contains the ``rse`` key and may include additional replica attributes
+            provided by the server.
 
-            List of replicas.
+        Raises
+        ------
+        RucioException
+            If the HTTP response status is not ``200 OK``.
+
+        Examples
+        --------
+        ??? Example
+
+            Print all replica locations for ``mock:file_b6222d9fe8e5434e84bfc002845348acanother.zip``:
+
+            ```python
+            >>> from rucio.client.fileclient import FileClient
+
+            >>> fc = FileClient()
+            >>> fc.list_file_replicas('mock', 'file_b6222d9fe8e5434e84bfc002845348acanother.zip')
+            [{'rse': 'CERN-PROD', ..}, {'rse': 'BNL-OSG2', ..}, ...]
+            ```
         """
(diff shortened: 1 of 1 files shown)


Geogouz:
Looks like this:

image

lib/rucio/client/fileclient.py
```diff
@@ -24,24 +24,64 @@
 
 class FileClient(BaseClient):
     """Dataset client class for working with dataset"""
```
voetberg: Should this still be here?

lib/rucio/client/fileclient.py
```diff
@@ -24,24 +24,64 @@
...
-        lfn :
-            The LFN
+        scope
+            The scope part of the file DID (e.g. ``"user.alice"``).
+        lfn
+            The file name within the scope.
 
         Returns
         -------
+        list[dict[str, Any]]
```
voetberg, line 59: What is the dict key? I assume the name of the replica

lib/rucio/client/fileclient.py
```diff
@@ -24,24 +24,64 @@
...
         Parameters
         ----------
-        scope :
-            The scope of the file.
-        lfn :
-            The LFN
+        scope
+            The scope part of the file DID (e.g. ``"user.alice"``).
+        lfn
+            The file name within the scope.
```
voetberg: Not sure what "within the scope" means here. The LFN isn't part of the scope.

lib/rucio/client/fileclient.py
```diff
@@ -24,24 +24,64 @@
...
+        Return all known replica locations of a file DID.
+
+        The method issues a GET request to the ``/files/<scope>/<name>/rses``
+        endpoint of the Rucio REST API. The server replies with a JSON list
+        describing each RSE where the file is currently present. Each entry in
+        the returned list contains the RSE name and may include additional
+        attributes such as the physical file name (PFN), file size, checksum
+        and replica state.
+
+        _**Note:**_ This method is currently not enabled.
```
voetberg: What do you mean by this? Do you mean it's not enabled by default? (we don't turn on the 'files' endpoint by default?) Or is there some other reason this doesn't work?

lib/rucio/client/fileclient.py
```diff
@@ -24,24 +24,64 @@
...
     BASEURL = 'files'
 
     def list_file_replicas(self, scope: str, lfn: str) -> list[dict[str, Any]]:
         """
-        List file replicas.
+        Return all known replica locations of a file DID.
+
+        The method issues a GET request to the ``/files/<scope>/<name>/rses``
+        endpoint of the Rucio REST API. The server replies with a JSON list
+        describing each RSE where the file is currently present. Each entry in
```
voetberg, line 43: > The method issues a GET request to the ``/files/<scope>/<name>/rses``  endpoint of the Rucio REST API. The server replies with a JSON list describing each RSE where the file is currently present.

I think this detail is unnecessary, this is just how any serve/client system works.

lib/rucio/client/fileclient.py
```diff
@@ -24,24 +24,64 @@
...
+
+            Print all replica locations for ``mock:file_b6222d9fe8e5434e84bfc002845348acanother.zip``:
+
+            ```python
+            from rucio.client.fileclient import FileClient
 
-            List of replicas.
+            file_client = FileClient()
+            replicas = file_client.list_file_replicas('mock', 'file_b6222d9fe8e5434e84bfc002845348acanother.zip')
+            print(replicas)
```
voetberg: Instead of this `for replica in replicas` thing, can I recommend the numpy style of examples where you assume this is being done in interactive python and just show the output directly. 

(For example: 
image
)

lib/rucio/client/fileclient.py
Geogouz: Basically (also related to couple other comments here), I was going a bit blind with this one because we don't have actually such an endpoint (or at least I didn't spot it). So I basically described how I imagined this was intended to work. See this issue: https://github.com/rucio/rucio/issues/7846

voetberg: ??????????????????

lib/rucio/client/fileclient.py
Geogouz: Something like in the updated example screenshot?

lib/rucio/client/fileclient.py
rdimaio: +1

rdimaio (approved):

lib/rucio/client/fileclient.py
voetberg: I think instead of documenting something that doesn't work, we should delay it and propose to remove it instead. It seems like a strange way to spend time

Geogouz: Yeh, let's wait for the linked issue to see what we will decide to do about it. I guess this client is about to be deleted.

rdimaio: https://github.com/rucio/rucio/pull/7855

Geogouz:
Closed due to #7855

PR #7848, opened 2025-07-14
Documentation: CredentialClient docstring enhancement; Fix rucio#7842

### LINKED ISSUE #7842: Rucio Python Client docstrings enhancement (`CredentialClient`)
Sub-issue of https://github.com/rucio/documentation/issues/363 targeting specifically the `CredentialClient`.

Geogouz:
Handles the sub-issue about the CredentialClient Docstrings

### DIFF
--- a/lib/rucio/client/credentialclient.py
+++ b/lib/rucio/client/credentialclient.py
@@ -19,7 +19,15 @@ from rucio.common.utils import build_url
 
 
 class CredentialClient(BaseClient):
-    """Credential client class for working with URL signing"""
+    """
+    Client helper to request signed URLs from a Rucio server.
+
+    A ``CredentialClient`` used to obtain temporary signed URLs from the server.
+    Those URLs allow direct access to objects on a storage service (currently
+    Google Cloud Storage, Amazon S3 or OpenStack Swift) without further authentication.
+    The signature embeds the permitted operation and its validity period, after which
+    the link becomes unusable.
+    """
 
     CREDENTIAL_BASEURL = 'credentials'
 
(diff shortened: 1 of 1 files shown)


Geogouz:
Looks like this:

image

lib/rucio/client/credentialclient.py
```diff
@@ -32,33 +41,69 @@ def get_signed_url(
...
+            on Google Cloud Storage valid for ten minutes:
+
+            ```python
+            from rucio.client.credentialclient import CredentialClient
+
+            cred_client = CredentialClient()
 
-            The signed URL string
+            try:
+                signed_url = cred_client.get_signed_url(
```
voetberg: The try-except loop is probably not required for an example. I would also show exactly what that signed URL looks like (sanitized of course)

lib/rucio/client/credentialclient.py
```diff
@@ -32,33 +41,69 @@ def get_signed_url(
...
-        operation :
-            The desired operation (read, write, delete)
-        url :
-            The URL to sign
-        lifetime :
-            The desired lifetime of the URL in seconds, by default 3600
+        rse
+            The name of the RSE to which the URL refers.
+        service
+            Storage service identifier. Must be one of ``"gcs"``, ``"s3"`` or ``"swift"``.
```
voetberg, line 55: Should the types be changed to reflect this instead of adding them in the docstring like this?

rdimaio: I agree (although it could be considered out of scope for the docstring change). It would be good to use `SUPPORTED_SIGN_URL_SERVICES_LITERAL` for the type annotation here (from `common.constants`)

lib/rucio/client/credentialclient.py
```diff
@@ -32,33 +40,65 @@ def get_signed_url(
...
+        operation
+            Allowed operation for the signed URL: ``"read"``, ``"write"`` or ``"delete"``.
```
rdimaio, line 57: Same comment as above, worth using `RSE_BASE_SUPPORTED_PROTOCOL_OPERATIONS_LITERAL`

bari12 (approved):

Issue #7840, opened 2025-07-14
Enable Registry based caching in Tests / Runtime Image Builds.


0xquark:
This issue builds upon #7839 and is part of #7667.  We need to enable registry based caching (from ghcr) in order to build the runtime images in case of dependencies bump or using the same runtime image for different matrix tests. This would help in reducing CI Time and simplifying the test-suite.

Issue #7839, opened 2025-07-14
Refactor CI to Use Runtime-Mounted Source Instead of Baking Rucio Code in Images


0xquark:
Currently, Rucio’s CI builds Docker images that [bake the source code directly into the image ](https://github.com/rucio/rucio/blob/6d473205cb846c96fbf08b1e9edea7027cc9baa1/.github/workflows/autotest.yml#L171-L188)for testing. This tightly couples the source with the runtime environment, which results in:

- Longer CI times (image rebuilds for every source change)
- Less flexibility for development workflows (Sometimes, old containers gets used during testing)
- Inefficient / Inconsistent image caching (every source change invalidates layers)

The idea is to refactor the CI as part of #7667, to build the base runtime images (3.9, 3.10) with dependencies installed in order to mount the rucio source code later during testing phase. 

This needs to be done for:
- [x] Autotests
- [x] Integration tests
- [x] VO Tests

PR #7853, opened 2025-07-15
Testing: Refactor CI / Test Infrastructure to use Runtime Mounted Source Code and Registry Caching in Autotests.

0xquark:
Ref: #7839 #7840 

This PR updates the CI testing approach by switching from building Docker images with the source baked in to using pre-built base runtime images and mounting the PR’s source code at runtime in `autotests.yml`.

- The base images (for Python 3.9 and 3.10) are built and pushed to GHCR using `runtime.Dockerfile` and the `runtime_images.yml` workflow (only if the image isn’t already available).

- These runtime images are then used by autotests.yml to execute tests. Inside `run_tests.py`, `test.sh`, and `docker-compose.yml`, the paths are configured to use the mounted PR code, which is installed in editable mode via `pip install -e`.

- Image builds uses registry based caching (rebuilding only when dependencies change), otherwise pulling the existing image. In case of dependencies bump or change, It uses old layers from the cache which hasn't changed.

The whole workflow with caching takes around [30-35 mins](https://github.com/0xquark/rucio/actions/runs/16190100219) instead of around [1h](https://github.com/rucio/rucio/actions/runs/16216539701). This happens because we only build the image once so we save a lot of time while running other tests in the matrix using the same image.

The whole testing workflow for autotests is described below:
image

### DIFF
--- a/.github/workflows/autotest.yml
+++ b/.github/workflows/autotest.yml
@@ -135,6 +135,11 @@ jobs:
           tools/run_pyright.sh compare --Werror \
             ${{ env.PYRIGHT_ANCESTOR_REPORT }} \
             ${{ env.PYRIGHT_CURRENT_REPORT }}
+  # Build runtime images for tests
+  runtime_images:
+    if: github.event_name != 'schedule'
+    name: Build Runtime Images
+    uses: ./.github/workflows/runtime_images.yml
   setup:
     if: github.repository_owner == 'rucio' || github.event_name != 'schedule'
     runs-on: ubuntu-latest
@@ -160,7 +165,7 @@ jobs:
       branch: ${{ steps.branch.outputs.branch }}
       matrix: ${{ steps.matrix.outputs.matrix }}
   test:
-    needs: setup
+    needs: [runtime_images, setup]
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
@@ -168,23 +173,15 @@ jobs:
         cfg: ${{ fromJson(needs.setup.outputs.matrix) }}
     steps:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
-      - name: Build images
-        id: images
-        shell: bash
-        run: |
-          docker login https://ghcr.io -u ${{ github.actor }} -p ${{ secrets.GITHUB_TOKEN }}
-          i=0; until [ "$i" -ge 3 ]; do
-            IMAGES=$(echo '${{ toJson(matrix.cfg) }}' | ./tools/test/build_images.py \
-                --cache-repo ghcr.io/${{ github.repository }} \
-                --branch "${{ needs.setup.outputs.branch }}" ./etc/docker/test || echo "")
-            if [[ -n $IMAGES ]]; then break;
-            else
-              i=$((i+1)); sleep 5;
-              echo "::warning::Building images failed, retrying…"
-            fi
-          done
-          docker logout https://ghcr.io
-          if [[ -z "$IMAGES" ]]; then echo "::error::Building images failed ultimately"; exit 1; fi
-          echo "images=$IMAGES" >> $GITHUB_OUTPUT
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@74a5d142397b4f367a81961eba4e8cd7edddf772 # v3.4.0
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
       - name: Run test with cfg
-        run: 'echo ''{"matrix": ${{ toJson(matrix.cfg) }}, "images": ${{ steps.images.outputs.images }} }'' | ./tools/test/run_tests.py'
+        env:
+          PY39_RUNTIME_IMAGE: ${{ needs.runtime_images.outputs.py39_image }}
+          PY310_RUNTIME_IMAGE: ${{ needs.runtime_images.outputs.py310_image }}
+        run: |
+          echo '{"matrix": ${{ toJson(matrix.cfg) }}, "runtime_images": {"3.9": "${{ needs.runtime_images.outputs.py39_image }}", "3.10": "${{ needs.runtime_images.outputs.py310_image }}"} }' | ./tools/test/run_tests.py
--- a/.github/workflows/autotest.yml
+++ b/.github/workflows/autotest.yml
@@ -174,7 +174,7 @@ jobs:
       matrix:
         cfg: ${{ fromJson(needs.setup.outputs.matrix) }}
     steps:
-      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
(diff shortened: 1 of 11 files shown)


0xquark:
Updated Workflow (with fix to the permission issue related to packages):
image

0xquark:
This workflow has been tested in [this PR](https://github.com/rucio/rucio/pull/7870) on @rdimaio’s fork.

- When the workflow runs on a `push`, since it doesn’t find the image on GHCR and the PR targets the master branch. In this case, the `build_locally` flag is set to `true` (as we would not have write permission for the actions run on main repository), so the Autotests run using a locally built image within the `rucio/rucio` GitHub Actions. 

- On Riccardo’s fork, Actions don’t run for pull requests targeting `master`. Since it’s the first time the image is being built there, the workflow builds and pushes the image to the fork’s package registry (as the fork has write permissions). Here, the`build_locally` flag remains `false` because it’s not running on the PR targeted to `rucio/rucio master` branch, and the GHCR image isn’t yet available. [These Actions on the fork build the image once and reuse it across jobs](https://github.com/rdimaio/rucio/actions/runs/16368355030/job/46250741137).

- To confirm that the PR code is properly mounted, we intentionally introduced incorrect assertions to break tests. This successfully triggered failures, as seen in [this commit](https://github.com/rucio/rucio/pull/7870/commits/36ccac4e2b5fa55d4e6d4a973f4c7f53ffff2489). The failures occurred in both the [main Rucio Actions run](https://github.com/rucio/rucio/actions/runs/16369793381/job/46255249844?pr=7870) and the [fork’s Actions run](https://github.com/rdimaio/rucio/actions/runs/16369792548/job/46255254182). On the fork, the workflow [pulls the pushed image from GHCR](https://github.com/rdimaio/rucio/actions/runs/16369792548/job/46255238903) instead of rebuilding it. 

These tests cover most of the key scenarios for the workflow and confirm that everything works as intended. There shouldn’t be any issues after merging the PR. Thanks @rdimaio for helping me test on his fork.

.github/workflows/autotest.yml
```diff
@@ -160,31 +167,56 @@ jobs:
...
+          context: .
+          file: etc/docker/test/runtime.Dockerfile
+          target: final
+          build-args: |
+            PYTHON=3.10
+          push: false
+          tags: ${{ needs.runtime_images.outputs.py310_image }}
+          outputs: type=docker
+          cache-from: type=registry,ref=ghcr.io/${{ github.repository }}/rucio-dev-runtime:py310-buildcache
+        continue-on-error: false
```
rdimaio, line 212: Could this be simplified in some way? e.g. since we have `matrix.cfg.PYTHON` that provides the version number, can we pass that directly to the `build-args` `PYTHON` argument and use it in the build cache names (e.g. `rucio-dev-runtime:py{{ matrix.cfg.PYTHON }}-buildcache` so that we can simplify this in a single step? 

I know that ultimately we only run one of these two steps depending on whether we're building python 3.9 or 3.10, but these two steps seem to mostly be the same, so I'm wondering if they can be simplified like that

etc/docker/test/runtime.Dockerfile
```diff
@@ -0,0 +1,181 @@
...
+    ENV C_INCLUDE_PATH="/usr/include/python${PYTHON}"
+    ENV PYTHON_VENV="/opt/venv"
+    ENV PATH="${PYTHON_VENV}/bin:${PATH}"
+    ENV PYTHON_310_PATCH_VERSION="4"
+    ENV RUCIO_HOME="/opt/rucio"
+
+FROM base AS oracle-client
+    RUN dnf install -y libnsl libaio nodejs npm
+    RUN rpm -i https://download.oracle.com/otn_software/linux/instantclient/1912000/oracle-instantclient19.12-basiclite-19.12.0.0.0-1.x86_64.rpm && \
+        echo "/usr/lib/oracle/19.12/client64/lib" > /etc/ld.so.conf.d/oracle-instantclient.conf;
```
rdimaio, line 17: No longer needed https://github.com/rucio/containers/pull/421 - @Geogouz pinging you just to confirm

.github/workflows/runtime_images.yml
```diff
@@ -0,0 +1,181 @@
...
+            org.opencontainers.image.source=${{ github.server_url }}/${{ github.repository }}
+            org.opencontainers.image.revision=${{ steps.build_metadata.outputs.commit_sha }}
+            org.opencontainers.image.ref.name=${{ steps.build_metadata.outputs.commit_ref }}
+            rucio.image.python.version=3.10
+            rucio.image.build.date=${{ steps.build_metadata.outputs.build_date }}
+            rucio.image.build.workflow=${{ github.workflow }}
+            rucio.image.build.run_id=${{ github.run_id }}
+          cache-from: type=registry,ref=ghcr.io/${{ github.repository }}/rucio-dev-runtime:py310-buildcache
+          cache-to: type=registry,ref=ghcr.io/${{ github.repository }}/rucio-dev-runtime:py310-buildcache,mode=max
+        continue-on-error: true
```
rdimaio: can this be simplified in a way to avoid duplication?

maany (dismissed):

.github/workflows/autotest.yml
maany: @rdimaio that will be part of autotest simplification :) https://github.com/maany/rucio/blob/861bf9a7379cfe59754be940674aeacc7805deed/.github/workflows/simple-autotest.yml

.github/workflows/runtime_images.yml
0xquark: Let me try this by using a [Matrix](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/run-job-variations) for the versions, this way i think this could be more simplified. Would probably reduce a lot of boilerplate.

etc/docker/test/runtime.Dockerfile
rdimaio: Checked with @Geogouz - let's leave it in for now

.github/workflows/runtime_images.yml
0xquark: I tried using the Matrix method to simplify things by setting it up for both Versions 3.9 and 3.10. The main advantage of using Matrix is that jobs run at parallel, so we could save about 3-4 minutes because both images would start building together at same time.

But after trying it, I found out that Matrix jobs run separately and independent of each other so each job can overwrite the other’s output depending on which finishes last. This causes problems because only one image output will be saved, and the other will be empty. That makes it hard to pass the built image (with its hash) to the Autotests and other tests.

I looked for a solution, but it didn’t work because this feature [isn’t supported yet in GitHub runners and is requested](https://github.com/orgs/community/discussions/17245). Another idea was to save the image tag as an artifact and then download it in another job to combine the outputs. But this makes the setup more complicated and not worth the extra complexity.

0xquark: The normal way to use loop for both the builds (3.9, 3.10) works well for simplification. It reduced a lot of code and is now much more cleaner and is part of a single step:).
https://github.com/0xquark/rucio/actions/runs/16569563583/job/46857899016

rdimaio (dismissed):

rdimaio (commented):

bari12 (approved):

Issue #7572, opened 2025-03-27
Plugins and policy packages: using plugin entry points for discovery?


rdimaio:
See:
- https://packaging.python.org/en/latest/guides/creating-and-discovering-plugins/
- https://docs.astral.sh/uv/concepts/projects/config/#plugin-entry-points

As this is a standard library spec, it might be worth moving to this system for plugin and policy package support, rather than using our in-house plugin discovery system.

bari12:
@jamesp-epcc have you ever looked into that?

jamesp-epcc:
It's not something I've really looked into before. I had a quick look just now but I think I would need to consider in more detail if it could be useful for us. It would be good to follow the standard, but we do some slightly unusual things that might not be a good match for it.

bari12:
@bziemons I suggested you as reviewer for this to James. Let us know if you prefer otherwise.

PR #7857, opened 2025-07-15
Policies: schema and permission, remove fallback logic that checks main Rucio codebase

### LINKED ISSUE #7856: schema and permission, remove fallback logic that checks main Rucio codebase
For example, in the schema logic here:

https://github.com/rucio/rucio/blob/master/lib/rucio/common/schema/__init__.py#L62-L79

We fall back to the `policy` in `rucio.common.schema` here:

https://github.com/rucio/rucio/blob/02f32e1ef166681511c4e37e6cf5df051b415549/lib/rucio/common/schema/__init__.py#L71-L77

This was useful when we had `atlas` and `belleii` as schemas in the `rucio.common.schema` module, but now we only have `generic` and `generic_multi_vo`, which are already falled back on in this branch: https://github.com/rucio/rucio/blob/02f32e1ef166681511c4e37e6cf5df051b415549/lib/rucio/common/schema/__init__.py#L78-L79

So this fallback logic is not needed anymore.

rdimaio:
fix #7856

### DIFF
--- a/lib/rucio/common/schema/__init__.py
+++ b/lib/rucio/common/schema/__init__.py
@@ -59,23 +59,15 @@ def _get_generic_schema_module():
 if not _is_multivo():
     GENERIC_FALLBACK = 'generic'
 
-    if config.config_has_section('policy'):
-        try:
-            if 'RUCIO_POLICY_PACKAGE' in environ:
-                policy = environ['RUCIO_POLICY_PACKAGE']
-            else:
-                policy = config.config_get('policy', 'package', check_config_table=False)
-            package_module = importlib.import_module(policy)
-            check_policy_module_version(package_module)
-            policy = policy + ".schema"
-        except (NoOptionError, NoSectionError, ModuleNotFoundError):
-            # fall back to old system for now
-            try:
-                policy = config.config_get('policy', 'schema', check_config_table=False)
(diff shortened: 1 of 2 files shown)


lib/rucio/common/schema/__init__.py
mgajek-cern, line 1: Changes correctly addresses the issue described. 

---

Just out of curiosity - this is outside the scope of this PR, but wouldn't it make sense to extract the logic from these `__init__.py` files into dedicated modules (e.g. `policy_loader.py` or `schema_manager.py`)? We could potentially refactor this in a future PR and add proper test coverage for all the functionality.

rdimaio: Yeah, I think that would be a good idea. In general I think we're in the process of refactoring the policy package import logic by quite a bit - we were potentially thinking about doing this https://github.com/rucio/rucio/issues/7572, which might let us support policy packages in a way that's more in line with Python's way of handling packages

mgajek-cern: Good idea. I was thinking more about moving the heavy initialization logic out of `__init__.py` for better code organization, but #7572 sounds like a good opportunity to tackle both improvements together.

mgajek-cern (approved):

PR #7862, opened 2025-07-16
Tests: Add fixtures for db session (read/write)

lobis:
This PR adds some fixtures useful for tests which require a db session.

The original PR where this change was introduced (https://github.com/rucio/rucio/pull/7592) was failing with a non-descript error, I was able to track it down to the changes present in this PR.

There is a deeper issue I cannot figure out that is making this change fail the CI.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -23,6 +23,9 @@ from typing import TYPE_CHECKING, Any, Optional
 
 import pytest
 
+from rucio.db.sqla.constants import DatabaseOperationType
+from rucio.db.sqla.session import db_session
+
 if TYPE_CHECKING:
     from collections.abc import Callable, Iterable, Iterator
     from configparser import ConfigParser
@@ -80,6 +83,18 @@ def pytest_make_parametrize_id(
         return argname + str(cfg)
     # return None to let pytest handle the formatting
     return None
+
+
+@pytest.fixture
+def db_read_session():
+    with db_session(DatabaseOperationType.READ) as session:
+        yield session
+
+
+@pytest.fixture
+def db_write_session():
+    with db_session(DatabaseOperationType.WRITE) as session:
+        yield session
 
 
 @pytest.fixture(scope='session')


rdimaio (commented):
CI is failing, but changes seem unrelated

lobis:
> CI is failing, but changes seem unrelated

Any idea why?

tests/conftest.py
```diff
@@ -23,6 +23,9 @@
 
 import pytest
 
+from rucio.db.sqla.constants import DatabaseOperationType
+from rucio.db.sqla.session import db_session
```
rdimaio, line 27: > > CI is failing, but changes seem unrelated
> 
> Any idea why?

I think this is the issue: if you look at the other fixtures in `conftest`, imports are done within the fixtures. I assume there is some circular import issue here that might be misreported.

e.g. for reference, `random_account` is like this:

```
@pytest.fixture
def random_account(vo: str) -> "Iterator[InternalAccount]":
    import random
    import string

    from rucio.core.account import add_account, del_account
    from rucio.db.sqla.constants import AccountType
    from rucio.tests.common_server import cleanup_db_deps

    account = InternalAccount(''.join(random.choice(string.ascii_lowercase) for _ in range(10)), vo=vo)

    if os.environ.get('SUITE') == 'client':
        c = Client(vo=vo)
        c.add_account(account=account.external, type_="user", email=f'{account.external}@email.com')
        yield account
        c.delete_account(account=account.external)

    else:
        add_account(account=account, type_=AccountType.USER, email=f'{account.external}@email.com')
        yield account
        cleanup_db_deps(model=models.Account, select_rows_stmt=models.Account.account == account)
        del_account(account)
```

Based on this, I assume that this would work if you move the imports to be within the fixtures

lobis: Thanks this is it!

lobis:
Closing this PR in favour of https://github.com/rucio/rucio/pull/7862 where I also included this change

PR #7863, opened 2025-07-17
Protocols: Make 0 the top priority #6312

### LINKED ISSUE #6312: Make 0 the top protocol priority
### Description

Following on the work of #6201, we proceed to make `0` the top protocol priority.

### Motivation

_No response_

### Change

 1. ~~Add a database migration to update existing protocols.~~
 2. Wherever `0` was being ignored, stop doing so. The sorting should take care of everything else.

dchristidis:
Below is the proposed text for the release notes. The documentation should also be updated after this PR is merged.

### Summary

This release contains the second part of a long-term effort to change how the RSE protocol priorities are interpreted. Rucio operators should review these instructions and plan accordingly.

* Prior to Rucio **32.0.0**, priority `0` meant disabled, `1` meant top-priority, `2` meant second-priority, etc.
* Between Rucio **32.0.0** and **39.0.0** (not inclusive), priority `0` *and/or* `None` meant disabled, `1` meant top-priority, `2` meant second-priority, etc.
* Since Rucio **39.0.0**, priority `None` means disabled, `0` *means top-priority*, `1` means second-priority, etc.

### Client Backward Compatibility

Operators should be advised that the configuration that is applied affects which version of the Rucio client are supported.

* When using `0` to denote disabled, `1` to denote top-priority, and `None` is not used at all, Rucio client earlier than **39.0.0** are supported. Client **39.0.0** and later may misbehave for some operations.
* When using `None` to denote disabled, `1` to denote top-priority, and `0` is not used at all, Rucio client earlier than **32.0.0** may misbehave for some operations. Client **32.0.0** and later are supported.
* When using `None` to denote disabled and `0` to denote top-priority, Rucio client earlier than **39.0.0** may misbehave for some operations. Client **39.0.0** and later are supported.

For established Rucio instances, it is generally recommended to choose the second option (i.e. not use `0` at all). The instructions that follow will assume as such. The third option is preferable if there are no concerns about backward compatibility.

### Upgrade Instructions

 1. (Optional) Review any systems that automatically configure RSEs or synchronise the Rucio instance with some external service.
 2. Apply the database schema upgrade.
 3. Update all existing RSE protocols in the database. It may look something like the following:
    ```sql
    UPDATE rse_protocols SET read_lan = NULL WHERE read_lan = 0;
    UPDATE rse_protocols SET write_lan = NULL WHERE write_lan = 0;
    UPDATE rse_protocols SET delete_lan = NULL WHERE delete_lan = 0;
    UPDATE rse_protocols SET read_wan = NULL WHERE read_wan = 0;
    UPDATE rse_protocols SET write_wan = NULL WHERE write_wan = 0;
    UPDATE rse_protocols SET delete_wan = NULL WHERE delete_wan = 0;
    UPDATE rse_protocols SET third_party_copy_read = NULL WHERE third_party_copy_read = 0;
    UPDATE rse_protocols SET third_party_copy_write = NULL WHERE third_party_copy_write = 0;
    ```
 4. Follow other instructions and procedures to upgrade Rucio.
 5. (Optional) Make a reminder to start using `0` to denote top-priority at some point in the future when Rucio clients earlier than **39.0.0** are no longer used.


rdimaio (commented):
Please rebase to fix the merge conflict

tests/test_bad_replica.py
```diff
@@ -442,8 +442,8 @@ def test_get_pfn_to_rse(rse_factory, rse_client, vo):
                   'prefix': '/the/one/with/all/the/files',
                   'impl': 'rucio.rse.protocols.SomeProtocol.SomeImplementation',
                   'domains': {
-                      'lan': {'read': 4,
-                              'write': 1,
+                      'lan': {'read': 1,
+                              'write': 0,
```
rdimaio, line 446: I see that `read` here changed from 4 to 1, in contrast with the pattern in the other changes for commit f2e36bd10848868568a5aeb1f16a375b943a1f3e where they all increased priority by 1. I assume it's fine to set this to 1 instead of 3, because we only have two protocols, but just checking if it was on purpose or not

tests/test_rse.py
```diff
@@ -532,8 +532,8 @@ def test_add_protocol(self, vo, rucio_client):
                       'prefix': '/the/one/with/all/the/files',
                       'impl': 'rucio.rse.protocols.SomeProtocol.SomeImplementation',
                       'domains': {
-                          'lan': {'read': 4,
-                                  'write': 1,
+                          'lan': {'read': 2,
```
rdimaio, line 549: Same comment here, read went from 4 to 2 instead of 3 (I assume it's fine as we only have 3 protocols)

tests/test_rse.py
```diff
@@ -563,18 +563,18 @@ def test_add_protocol(self, vo, rucio_client):
...
         resp = mgr.get_rse_info(rse=protocol_rse, vo=vo)
         for p in resp['protocols']:
-            if ((p['port'] == 19) and (p['domains']['lan']['read'] != 1)) or \
-                    ((p['port'] == 20) and (p['domains']['lan']['read'] != 2)) or \
-                    ((p['port'] == 18) and (p['domains']['lan']['read'] != 1)) or \
-                    ((p['port'] == 17) and (p['domains']['lan']['read'] != 4)):
+            if ((p['port'] == 19) and (p['domains']['lan']['read'] != 0)) or \
+                    ((p['port'] == 20) and (p['domains']['lan']['read'] != 1)) or \
+                    ((p['port'] == 18) and (p['domains']['lan']['read'] != 0)) or \
+                    ((p['port'] == 17) and (p['domains']['lan']['read'] != 2)):
```
rdimaio, line 591: Same comment here

rdimaio (dismissed):

tests/test_bad_replica.py
dchristidis: Yes, in all these cases there was no apparent reason for not having consecutive protocol priorities.

lib/rucio/alembicrevision.py
rdimaio, line 1: Discussed offline - Adding a comment to block merging until the release 39 code freeze, to reduce the additional work by other developers.

rdimaio (dismissed):
Is it safe to merge now, as the code freeze is in effect?

dchristidis:
I would say to keep this as the last PR to be merged since I do not wish to inflict merge conflicts on anyone else, but I am fine either way.

Geogouz:
Should we introduce any data migration (or upgrade step) that converts existing `0` (disabled I assume) priority values to `NULL`? To basically preserve the “disabled” semantics before enabling 0 as top-priority. Because now, previously disabled protocols become top-priority.

lib/rucio/rse/rsemanager.py
Geogouz: Should we update this description (i.e. zero is not supported)?

lib/rucio/core/rse.py
Geogouz: This exception becomes now a bit misleading. Should we adjust it?

Geogouz:
Looks quite good to me. Couple things that are too minor.

lib/rucio/rse/rsemanager.py
dchristidis: Addressed in commit 96e221c.

lib/rucio/core/rse.py
dchristidis: Addressed in commit 2efa927.

dchristidis:
Rebased to resolve conflict in `lib/rucio/alembicrevision.py`.

rdimaio (approved):

PR #7889, opened 2025-07-29
Core: add missing execute stmt line to update accessed_at of locks

### LINKED ISSUE #7888: Last access time of dataset locks is not updated
### Description

https://github.com/rucio/rucio/blob/3edc50731e780b0ad8d62f92a98bc37be843b9ca/lib/rucio/core/lock.py#L530-L577

The first query is constructed but not executed. This is a regression from https://github.com/rucio/rucio/pull/6935.

### Steps to reproduce

N/A

### Rucio Version

_No response_

### Additional Information

_No response_

rdimaio:
fix #7888

### DIFF
--- a/lib/rucio/core/lock.py
+++ b/lib/rucio/core/lock.py
@@ -553,6 +553,9 @@ def touch_dataset_locks(dataset_locks: "Iterable[dict[str, Any]]", *, session: "
             }).execution_options(
                 synchronize_session=False
             )
+
+            session.execute(stmt)
+
             subq = select(
                 models.DatasetLock.rule_id
             ).where(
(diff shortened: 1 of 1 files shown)


lib/rucio/core/lock.py
```diff
@@ -553,6 +553,9 @@ def touch_dataset_locks(dataset_locks: "Iterable[dict[str, Any]]", *, session: "
             }).execution_options(
                 synchronize_session=False
             )
+
```
dchristidis, line 557: I would remove this empty line to tie the execution of the statement with its declaration.

Then I would also also remove the empty line between `subq` and `stmt` below too.

dchristidis (commented):
I’m starting to think that reusing the same variable name for multiple queries should be regarded as an antipattern. If we had used `update_dslock_stmt` and `update_rule_stmt` (or whatever convention we decide to go with), then Ruff would have caught this regression.

rdimaio:
> I’m starting to think that reusing the same variable name for multiple queries should be regarded as an antipattern. If we had used `update_dslock_stmt` and `update_rule_stmt` (or whatever convention we decide to go with), then Ruff would have caught this regression.

Good point, addressed in latest revision

lib/rucio/core/lock.py
rdimaio: Addressed in latest revision

dchristidis (approved):

Issue #7898, opened 2025-07-31
Testing: Docker RSE setup, passing both distance and ranking is unnecessary


rdimaio:
Done here:

https://github.com/rucio/rucio/blob/6ac13a2f74fe56afdf843dd8d9173e84d448b79c/tools/docker_activate_rses.sh#L94-L106

But actually ranking is just ignored if distance is present:

https://github.com/rucio/rucio/blob/6ac13a2f74fe56afdf843dd8d9173e84d448b79c/lib/rucio/web/rest/flaskapi/v1/rses.py#L1940-L1950

rdimaio:
Good first issue, just need to remove the `--ranking`s

rishavd3v:
@rdimaio Hi, I'd like to work on this issue. Can you assign it to me?

rdimaio:
> [@rdimaio](https://github.com/rdimaio) Hi, I'd like to work on this issue. Can you assign it to me?

Thanks for picking it up :) assigned, please make sure to follow the contributing guide https://rucio.github.io/documentation/contributing/

PR #7902, opened 2025-08-01
Testing: remove redundant --ranking flag from RSE setup #7898

rishavd3v:
### Summary

This PR removes the unnecessary `--ranking` flag from the `rucio-admin rse add-distance` commands in the `docker_activate_rses.sh` script.

### Context

According to issue #7898, if the `--distance` flag is provided, the `--ranking` parameter is ignored. Including both is redundant.


rdimaio (approved):

rdimaio:
LGTM, thanks for the contribution!

Merging even though we're in code freeze as this is just a testing change (and even then, the `ranking` is completely ignored anyway, see original issue)

PR #7903, opened 2025-08-01
Return `is_opendata` info in meta query

### LINKED ISSUE #7901: Fetch `is_opendata` via existing endpoints
A DID that is part of opendata has an entry in a dedicated opendata table and this can be used to know wether a DID is opendata or not.

It is useful to know if a DID is part of opendata, for example, in order to display some additional info in the webui.

Right now this could be done by doing an extra query for the DID or interest but IMHO this would mean too many additional queries that would be a miss since most DIDs are not opendata. For example, in the webui an additional query would have to be made for every DID in the details page.

I propose modifying the `did_column_meta` plugin to also return this `is_opendata` boolean flag by doing an outter left join. This should not have a measureable impact in performance. This is the query used at the moment by the webui to fetch metadata to display.

In the future we might distinguish between user-supplied metadata and internal rucio metadata (attributes?) and this `is_opendata` could be part of the internal attributes or some other more scalable model.

WebUI related PR: https://github.com/rucio/webui/pull/568

lobis:
Closes #7901 

It is also required for https://github.com/rucio/webui/issues/569 and https://github.com/rucio/webui/pull/568

### DIFF
--- a/lib/rucio/core/did_meta_plugins/did_column_meta.py
+++ b/lib/rucio/core/did_meta_plugins/did_column_meta.py
@@ -16,10 +16,10 @@ import operator
 from datetime import datetime, timedelta
 from typing import TYPE_CHECKING, Any
 
-from sqlalchemy import inspect, update
+from sqlalchemy import and_, false, inspect, null, update
 from sqlalchemy.exc import CompileError, InvalidRequestError, NoResultFound
 from sqlalchemy.sql import func
-from sqlalchemy.sql.expression import true
+from sqlalchemy.sql.expression import case, select, true
 
 from rucio.common import exception
 from rucio.core import account_counter, rse_counter
@@ -66,9 +66,32 @@ class DidColumnMeta(DidMetaPlugin):
         :returns: DID metadata as a dictionary.
         """
         try:
-            row = session.query(models.DataIdentifier).filter_by(scope=scope, name=name). \
-                with_hint(models.DataIdentifier, "INDEX(DIDS DIDS_PK)", 'oracle').one()
-            return row.to_dict()
+            opendata_subquery = select(
+                models.OpenDataDid.scope,
+                models.OpenDataDid.name
+            ).subquery()
+
+            stmt = select(
+                models.DataIdentifier,
+                case((opendata_subquery.c.scope.isnot(null()), true()), else_=false()).label("is_opendata")
+            ).outerjoin(
+                opendata_subquery,
+                and_(
+                    models.DataIdentifier.scope == opendata_subquery.c.scope,
+                    models.DataIdentifier.name == opendata_subquery.c.name
+                )
+            ).where(
+                models.DataIdentifier.scope == scope,
+                models.DataIdentifier.name == name
+            )
+
+            row = session.execute(stmt).one()
+
+            data_identifier, is_opendata = row
+            result = data_identifier.to_dict()
+            result["is_opendata"] = is_opendata
+            return result
+
         except NoResultFound:
             raise exception.DataIdentifierNotFound(f"Data identifier '{scope}:{name}' not found")
 
--- a/tests/test_opendata.py
+++ b/tests/test_opendata.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import json
 import re
 from configparser import NoOptionError
 
@@ -34,10 +35,17 @@ skip_unsupported_json = pytest.mark.skipif(
     reason="JSON support is not implemented in this database"
 )
 
-skip_unsupported_dialect = pytest.mark.skipif(
-    get_session().bind.dialect.name in ['oracle', 'sqlite'],
-    reason=f"Unsupported dialect: {get_session().bind.dialect.name}"
-)
+
+def skip_unsupported_dialect(*unsupported_dialects):
+    session = get_session()
+    if not session.bind:
+        return pytest.mark.skipif(False, reason="Database not bound")
+
+    dialect: str = session.bind.dialect.name
+    should_skip = dialect in unsupported_dialects
+    reason = f"Unsupported dialect: {dialect}" if should_skip else "Dialect supported"
+    return pytest.mark.skipif(should_skip, reason=reason)
+
 
 OPENDATA_RSE_EXPRESSION = 'OpenData=True'
 
@@ -212,7 +220,7 @@ class TestOpenDataCore:
 
         assert state == OpenDataDIDState.PUBLIC
 
-    @skip_unsupported_dialect
+    @skip_unsupported_dialect("oracle", "sqlite")
     def test_opendata_dids_meta_update(self, mock_scope, root_account, db_write_session):
         name = did_name_generator(did_type="dataset")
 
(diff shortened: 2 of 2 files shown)


lobis:
The CI was failing only on postgresql runs with only the `is_opendata` key missing, likely because `opendata_table_exists` is returning false even though the table exists (all the other opendata tests pass).

```
opendata_table_exists = session.get_bind() is not None and inspect(session.get_bind()).has_table(models.OpenDataDid.__tablename__)
```

This might be because `session.get_bind()` is returning `None` but I don't understand why.

In the end I modifed the code to return `is_opendata` false when the opendata table is not found and skip on postgresql but I write this comment to keep this weird behaviour documented somewhere as it might come back to bite us!

It is worth mentioning that on my dev environment (k8s deployment) the change works fine on posgresql.

Thanks to @rdimaio for his help identifying the issue!

rdimaio (dismissed):

rdimaio (commented):
Please rebase to fix merge conflicts (if this is ready for merging now, otherwise feel free to wait if there is something that blocks this)

tests/test_opendata.py
```diff
@@ -605,7 +619,174 @@ def test_opendata_api_add_remove(self, rest_client, auth_token, root_account, mo
...
+        # Check `is_opendata` returns False again after removal
+        response = rest_client.get(
+            meta_endpoint,
+            headers=request_headers,
+        )
+        assert response.status_code == 200, f"Expected 200 OK, got {response.status_code}"
+        response_data = json.loads(response.get_data(as_text=True))
+        is_opendata: bool = response_data["is_opendata"]
+        assert not is_opendata, "Expected is_opendata to be False after removal from Opendata"
+
```
Geogouz, line 738: Can you add here a second line to avoid `E302`?

tests/test_opendata.py
```diff
@@ -605,7 +619,174 @@ def test_opendata_api_add_remove(self, rest_client, auth_token, root_account, mo
...
+    @skip_unsupported_dialect("postgresql")
+    def test_is_opendata(self, rest_client, auth_token, root_account, mock_scope):
```
Geogouz: This would give `F811` redefinition of unused 'test_is_opendata' from line 679

tests/test_opendata.py
```diff
@@ -605,7 +619,174 @@ def test_opendata_api_add_remove(self, rest_client, auth_token, root_account, mo
...
```
Geogouz: This would give `F811` redefinition of unused 'test_is_opendata' from line 623

lobis: Thanks, during the rebase this code block ended up appearing three times, now its deduplicated.

Geogouz (dismissed):
Looks good to me

lib/rucio/core/did_meta_plugins/did_column_meta.py
```diff
@@ -66,9 +66,43 @@ def get_metadata(
...
         try:
-            row = session.query(models.DataIdentifier).filter_by(scope=scope, name=name). \
-                with_hint(models.DataIdentifier, "INDEX(DIDS DIDS_PK)", 'oracle').one()
-            return row.to_dict()
+            # We check for the existence of the OpenDataDid table to improve backwards compatibility with older database schemas.
+            # At some point this check can be removed, and the code can be simplified (only running the else branch).
+            opendata_table_exists = session.get_bind() is not None and inspect(session.get_bind()).has_table(
+                models.OpenDataDid.__tablename__)
+            if not opendata_table_exists:
+                row = session.query(models.DataIdentifier).filter_by(scope=scope, name=name). \
```
dchristidis: New code should not use the 1.4 syntax, per the [contribution guide](https://rucio.cern.ch/documentation/developer/dev_style_guide#sqlalchemy-syntax).

dchristidis (changes requested):

lib/rucio/core/did_meta_plugins/did_column_meta.py
lobis: Hopefully it's addressed now, otherwise please clarify further.

dchristidis: `session.query()` which is used twice in the `else` clause is 1.4 syntax.

lib/rucio/core/did_meta_plugins/did_column_meta.py
```diff
@@ -66,9 +66,69 @@ def get_metadata(
         :returns: DID metadata as a dictionary.
         """
         try:
-            row = session.query(models.DataIdentifier).filter_by(scope=scope, name=name). \
-                with_hint(models.DataIdentifier, "INDEX(DIDS DIDS_PK)", 'oracle').one()
-            return row.to_dict()
+            # We check for the existence of the OpenDataDid table to improve backwards compatibility with older database schemas.
+            # At some point this check can be removed, and the code can be simplified (only running the else branch).
+            opendata_table_exists = session.get_bind() is not None and inspect(session.get_bind()).has_table(
+                models.OpenDataDid.__tablename__)
```
dchristidis: Could you elaborate on the need for this? Typically we do not have to think about the compatibility between the Rucio version running server-side and the database.

lib/rucio/core/did_meta_plugins/did_column_meta.py
```diff
@@ -66,9 +66,69 @@ def get_metadata(
...
+
+                stmt = select(
+                    models.DataIdentifier,
+                    case(
+                        (
+                            opendata_subquery.c.scope.isnot(
+                                null()
+                            ),
+                            true()
+                        ),
```
dchristidis: I think joining these into one line would be preferable?

lib/rucio/core/did_meta_plugins/did_column_meta.py
```diff
@@ -66,9 +66,69 @@ def get_metadata(
...
+                ).where(
+                    models.DataIdentifier.scope == scope,
+                    models.DataIdentifier.name == name
+                )
+
+                row = session.execute(stmt).first()
+                if row is None:
+                    # Otherwise a 'TypeError: cannot unpack non-iterable NoneType object' is raised
+                    # This will be converted to a 'DataIdentifierNotFound' by the except block
+                    raise NoResultFound
```
dchristidis: Why would the `one()` method not work here?

dchristidis (changes requested):
I observe that some of the changes to `test_opendata.py` are not described by the main commit and should be probably split into independent commits.

* `skip_unsupported_dialect`
* `request_headers`
* HTTP status

lobis:
The motivation of this PR is mainly to have a way for the WebUI to know if a DID is open data for display purposes and in the future to fetch additional open data specific information.

The way this is implemented in this PR is a bit convoluted due to the db constraints and its currently not working for all dbs, which makes necessary to add some skips.

I am afraid this might break more important functionality, so I will pause the merging of this PR for now since the open data webui integration is not time sensitive. It would be preferrable to merge this change with all passing tests either by changing the implementation or by waiting out some update that makes things magically work.

dchristidis:

* The PR is in draft mode and there has been no activity for over seven days.

Issue #7010, opened 2024-08-07
Testing: Change name of `NO_XDIST` flag


rdimaio:
https://github.com/rucio/rucio/blob/8c4672c64b8784fd9f9cdc9e3516d7d83ecd22c9/tools/pytest.sh#L33-L58

`NO_XDIST=False` means that XDIST is true, so for example:

```
if [[ "$NO_XDIST" == "False" ]]; then
```

means "let's run `xdist`"

I think it would be easier to comprehend if it was named `RUN_XDIST`, `IS_XDIST` or something like that - the `NO_` part causes it to be a bit confusing due to the double negation

mgajek-cern:
@rdimaio Is this GH issue still valid ? If so I will add it to the current sprint and will address it in an initial PR to get familiar with the process

rdimaio:
> [@rdimaio](https://github.com/rdimaio) Is this GH issue still valid ? If so I will add it to the current sprint and will address it in an initial PR to get familiar with the process

Yeah, I think it's still valid, a better naming convention would be useful here in my opinion

PR #7908, opened 2025-08-05
Testing: rename `NO_XDIST` flag to `RUN_XDIST`

mgajek-cern:
Addresses #7010

### DIFF
--- a/tools/pytest.sh
+++ b/tools/pytest.sh
@@ -30,28 +30,29 @@ fi
 
 export PYTEST_DISABLE_PLUGIN_AUTOLOAD="True"
 
-RUN_XDIST="${RUN_XDIST:-False}"
+RUN_XDIST="${RUN_XDIST:-True}"
 if [[ "${RDBMS:-}" == "sqlite" ]]; then
   # no parallel tests on sqlite, because of random "sqlite3.OperationalError: database is locked"
   echo "Disabling parallel testing for sqlite"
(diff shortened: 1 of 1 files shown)


tools/pytest.sh
```diff
@@ -30,28 +30,28 @@ fi
 
 export PYTEST_DISABLE_PLUGIN_AUTOLOAD="True"
 
-NO_XDIST="${NO_XDIST:-False}"
+RUN_XDIST="${RUN_XDIST:-False}"
 if [[ "${RDBMS:-}" == "sqlite" ]]; then
   # no parallel tests on sqlite, because of random "sqlite3.OperationalError: database is locked"
   echo "Disabling parallel testing for sqlite"
-  NO_XDIST="True"
+  RUN_XDIST="True"
```
rdimaio: You should also flip the bool logic: if `NO_XDIST` is true, we don't run XDIST. But if `RUN_XDIST` is true, we run XDIST. These semantically have opposite meaning, so every instance should be flipped accordingly: e.g. `NO_XDIST` was `False` by default, so `RUN_XDIST` should be `True` by default, and so on.

mgajek-cern: Addressed in subsequent commit

mgajek-cern:
Closed due to incorrect branch name. See updated PR: https://github.com/rucio/rucio/pull/7916

PR #7910, opened 2025-08-06
Docker & Kubernetes: Add health checks and service dependencies in docker-compose

### LINKED ISSUE #7909: Docker & Kubernetes: Add health‑checks to database containers in `docker-compose`
### Description

When we start the dev environment/CI/tests using the `docker-compose`, the rucio service (especially if we invoke fast the tests) may race ahead of some containers (e.g. databases) and crash with connection refused errors.
We should better make Rucio container wait for the databases to spin up.

### Steps to reproduce

Set `oracle` as the preferred database and execute the tests (via `docker exec -it dev-rucio-1 tools/run_tests.sh`), immediately after `./tools/bootstrap_dev.sh -p storage -p monitoring -p oracle -x`.

### Rucio Version

Server 38.0.0rc2

### Additional Information

_No response_

Geogouz:
Fixes rucio/rucio#7909

Now rucio container waits first for the various services to become healthy (up until our defined time-window of tries) before it spins up, giving ultimately a warning when some dependency service isn't started or available

image

image

### DIFF
--- a/etc/docker/dev/docker-compose.yml
+++ b/etc/docker/dev/docker-compose.yml
@@ -71,6 +71,9 @@ services:
       postgres14:
         condition: service_healthy
         required: false
+      elasticsearch:
+        condition: service_healthy
+        required: false
 
   # ------------------------------------------------------------------
   # Database, MQ, metrics, and other external services
@@ -124,6 +127,11 @@ services:
     image: docker.io/elasticsearch:7.8.0
     environment:
       - discovery.type=single-node
+    healthcheck:
+      test: [ "CMD", "curl", "-f", "http://localhost:9200/_cluster/health?wait_for_status=yellow" ]
+      interval: 10s
+      timeout: 5s
+      retries: 10
 
   activemq:
     container_name: dev-activemq-1
--- a/etc/docker/dev/docker-compose.yml
+++ b/etc/docker/dev/docker-compose.yml
@@ -61,6 +61,16 @@ services:
       - REQUESTS_CA_BUNDLE=/etc/pki/tls/certs/ca-bundle.crt
       - RUCIO_SOURCE_DIR=/rucio_source
       - RDBMS
+    depends_on:
+      oracle:
+        condition: service_healthy
+        required: false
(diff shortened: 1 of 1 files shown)


rdimaio (dismissed):

etc/docker/dev/docker-compose.yml
```diff
@@ -61,6 +61,22 @@ services:
       - REQUESTS_CA_BUNDLE=/etc/pki/tls/certs/ca-bundle.crt
       - RUCIO_SOURCE_DIR=/rucio_source
       - RDBMS
+    depends_on:
+      oracle:
+        condition: service_healthy
+        required: false
+      ruciodb:
+        condition: service_healthy
+        required: false
```
rdimaio, line 70: Shouldn't these be `required: true` for "Now rucio container waits for the databases to become healthy first before it spins up." to work?

https://docs.docker.com/reference/compose-file/services/#depends_on

> required: When set to false Compose only warns you when the dependency service isn't started or available. If it's not defined the default value of required is true. Introduced in Docker Compose version [2.20.0](https://docs.docker.com/compose/releases/release-notes/#2200).

Geogouz: If we set it to true, then we wont be able to start rucio without that service and in the CI or in general we want to be able to do so. Practically there is enough time for the container to be allowed to start before Rucio container decides to give up waiting for it (in practice I have actually never seen it giving up and starting regardless (with an error that it didn't start)).

Geogouz:
Rebased

etc/docker/dev/docker-compose.yml
rdimaio: Makes sense; I think we should specify in the PR description that this only raises a warning, rather than actually waiting for the dbs to come online

rdimaio: And it would be also good if the PR title reflected that the health checks added are not just for ruciodb and oracle but for the other services as well

rdimaio (approved):

PR #7924, opened 2025-08-11
Testing: Fix self‑referential symlink traversal

### LINKED ISSUE #7920: Testing: Symlink causes pytest to loop through the same directory repeatedly
### Description

After a pytest profiling, I got to realise that in our dev environment, when pytest is launched from `tools/`, it follows the `tools/lib` symlink and repeatedly re-enters the same directory (e.g., tools/lib/rucio/rucio/...), causing a huge number of redundant path inspections making the test collection extremely slow (approx. 8 minutes spent just for the collection till the process breaks - due to max recursions probably?).

### Steps to reproduce

-

### Rucio Version

Server: 38.0.0rc2

### Additional Information

**Solution**
Setting `--rootdir=${RUCIO_DIR}` or telling pytest to ignore the symlink eliminates the recursion and restores near‑instant discovery.

Geogouz:
Fixes https://github.com/rucio/rucio/issues/7920

And hopefully saves 8m of our lives each time we run tests (at least locally).

Depends on #7931 (we need to merge that first)

### DIFF
--- a/tools/pytest.ini
+++ b/tools/pytest.ini
@@ -1,0 +1,6 @@
+[pytest]
+pythonpath =
+    .
+    lib
+testpaths =
+    tests
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,8 +4,11 @@ build-backend = "setuptools.build_meta"
 
 [tool.pytest.ini_options]
 pythonpath = [
-  ".",
-  "lib",
+    ".",
+    "lib",
+]
+testpaths = [
+    "tests",
 ]
 
 [tool.coverage.report]
--- a/tools/pytest.sh
+++ b/tools/pytest.sh
@@ -63,4 +63,4 @@ if [[ "$RUN_XDIST" == "True" ]]; then
   echo "Running pytest with pytest-xdist: " "${XDIST_ARGS[@]}"
 fi
 
-exec python -bb -m pytest -r fExX --log-level=DEBUG ${XDIST_ARGS[@]+"${XDIST_ARGS[@]}"} ${ARGS[@]}
+exec python -bb -m pytest --rootdir=${RUCIO_DIR} -c tools/pytest.ini -r fExX --log-level=DEBUG ${XDIST_ARGS[@]+"${XDIST_ARGS[@]}"} ${ARGS[@]}


Geogouz:
**Update:**
I see that couple tests are failing, namely:

```
FAILED tests/test_multi_vo.py::TestVORestAPI::test_auth_oidc
FAILED tests/test_multi_vo.py::TestVORestAPI::test_auth_oidc_polling
```

Since these `multi_vo` tests are not executed locally, there is no FAIL there. Yet, the outcome of those tests should in theory not be dependent on this PR's introduced changes.

**Sidenotes:**
I also have a feeling that before this change, the multi‑VO tests executed against the pre‑installed Rucio package in the container. After current PR's change, pytest imports modules directly from the repository source tree and the environment adjustment may have exposed the underlying bug.

Specifically regarding the problem with those tests, `lib/rucio/web/rest/flaskapi/v1/auth.py` looks like it has fragile error handling in `CodeOIDC.get`. When `/auth/oidc_code` raises an exception with no message, the handler accesses `error.args[0]`, triggering an `IndexError` and a 500 response.

tools/pytest.ini
rdimaio, line 1: Do we need this change in both this file and `pyproject.toml`? Can we delete this file and just use the settings from the `toml` file?

Geogouz: I also don't like it. Tried to make them applicable from the toml or by playing with different resource configurations but my efforts were in vain.

Basically, once this goes away `-c tools/pytest.ini` problems start.

etc/docker/test/alma9.Dockerfile
rdimaio, line 1: Adding comment to block merge until https://github.com/rucio/rucio/pull/7931 is merged

rdimaio (approved):

etc/docker/test/alma9.Dockerfile
```diff
@@ -94,6 +94,9 @@ FROM python as rucio-runtime
     COPY requirements requirements
     COPY .pep8 .pycodestyle pyproject.toml setup.py setup_rucio.py setup_rucio_client.py setup_webui.py setuputil.py ./
 
+    # Include packaging manifest so templates ship with the wheel
```
rdimaio: https://github.com/rucio/rucio/pull/7931 has been merged, please rebase

rdimaio (approved):

PR #7925, opened 2025-08-12
Clients, API: Refactor HTTP method strings for enums

### LINKED ISSUE #7620: Use HTTP verbs in client request types
The client is full of statements such as:

```
r = self._send_request(url, type_='DEL', data=render_json(**data))
```

Where the `_send_request` switches to the corresponding http verb using the `type_` argument.

Naively one would think (atleast I did) that this `type_` should be an http verb, such as `GET` or `DELETE`, however it's not always the case. For instance, for deletion you need to specify `DEL` instead of `DELETE`.

I spent some time until I figured this out and I don't see any reason why we cannot use the HTTP verbs as values for the `type_` arg instead.

lobis:
Closes #7620 

This PR modifies `_send_request` which is an internal api. The parameter `type_` has been renamed to `method` and now takes an enum instead of a string.

When an invalid method is used an exception is raised. Previously it was just ignored. This resulted in some hard to track bugs in my experience, since this is an internal method I don't see any reason why this should not raise exception in this case as it should always be called properly (and now this error condition can be statically checked).

I updated all occurrences of this method to match the new usage.

**Changes**

The `_send_request` method now does not have a default value for the method (used to be `GET`). It also does not have a type hint. This is done due to `pyright` complaining in the CI, I was unable to fix this or even add an ignore directive. After discussing with @bari12 offline this was deemed acceptable as this method is interal only.

### DIFF
--- a/lib/rucio/client/baseclient.py
+++ b/lib/rucio/client/baseclient.py
@@ -38,7 +38,7 @@ from requests.status_codes import codes
 from rucio import version
 from rucio.common import exception
 from rucio.common.config import config_get, config_get_bool, config_get_int, config_has_section
-from rucio.common.constants import DEFAULT_VO
+from rucio.common.constants import DEFAULT_VO, HTTPMethod
 from rucio.common.exception import CannotAuthenticate, ClientProtocolNotFound, ClientProtocolNotSupported, ConfigNotFound, MissingClientParameter, MissingModuleException, NoAuthInformation, ServerConnectionException
 from rucio.common.extra import import_extras
 from rucio.common.utils import build_url, get_tmp_dir, my_key_generator, parse_response, setup_logger, ssh_sign, wlcg_token_discovery
@@ -440,14 +440,14 @@ class BaseClient:
         self.logger.warning("Waiting {}s due to reason: {} ".format(sleep_time, reason))
         time.sleep(sleep_time)
 
-    def _send_request(self, url, headers=None, type_='GET', data=None, params=None, stream=False, get_token=False,
+    def _send_request(self, url, method, headers=None, data=None, params=None, stream=False, get_token=False,
                       cert=None, auth=None, verify=None):
         """
         Helper method to send requests to the rucio server. Gets a new token and retries if an unauthorized error is returned.
 
         :param url: the http url to use.
         :param headers: additional http headers to send.
-        :param type_: the http request type to use.
+        :param method: the http request type to use.
         :param data: post data.
         :param params: (optional) Dictionary or bytes to be sent in the url query string.
         :param get_token: (optional) if it is called from a _get_token function.
@@ -469,28 +469,28 @@ class BaseClient:
         if verify is None:
             verify = self.ca_cert or False  # Maybe unnecessary but make sure to convert "" -> False
 
-        self.logger.debug("HTTP request: %s %s" % (type_, url))
+        self.logger.debug("HTTP request: %s %s" % (method.value, url))
         for h, v in hds.items():
             if h == 'X-Rucio-Auth-Token':
                 v = "[hidden]"
             self.logger.debug("HTTP header:  %s: %s" % (h, v))
-        if type_ != "GET" and data:
+        if method != HTTPMethod.GET and data:
             text = self._reduce_data(data)
             self.logger.debug("Request data (length=%d): [%s]" % (len(data), text))
 
         result = None
         for retry in range(self.AUTH_RETRIES + 1):
             try:
-                if type_ == 'GET':
+                if method == HTTPMethod.GET:
                     result = self.session.get(url, headers=hds, verify=verify, timeout=self.timeout, params=params, stream=True, cert=cert, auth=auth)
-                elif type_ == 'PUT':
+                elif method == HTTPMethod.PUT:
                     result = self.session.put(url, headers=hds, data=data, verify=verify, timeout=self.timeout)
-                elif type_ == 'POST':
+                elif method == HTTPMethod.POST:
                     result = self.session.post(url, headers=hds, data=data, verify=verify, timeout=self.timeout, stream=stream)
-                elif type_ == 'DEL':
+                elif method == HTTPMethod.DELETE:
                     result = self.session.delete(url, headers=hds, data=data, verify=verify, timeout=self.timeout)
                 else:
-                    self.logger.debug("Unknown request type %s. Request was not sent" % (type_,))
+                    self.logger.debug("Unknown request type %s. Request was not sent" % (method,))
                     return None
                 self.logger.debug("HTTP Response: %s %s" % (result.status_code, result.reason))
                 if result.status_code in STATUS_CODES_TO_RETRY:
@@ -537,7 +537,7 @@ class BaseClient:
 
         url = build_url(self.auth_host, path='auth/userpass')
 
-        result = self._send_request(url, headers=headers, get_token=True)
+        result = self._send_request(url, method=HTTPMethod.GET, headers=headers, get_token=True)
 
         if not result:
             # result is either None or not OK.
@@ -590,7 +590,7 @@ class BaseClient:
             return False
 
         request_refresh_url = build_url(self.auth_host, path='auth/oidc_refresh')
-        refresh_result = self._send_request(request_refresh_url, get_token=True)
+        refresh_result = self._send_request(request_refresh_url,method=HTTPMethod.GET, get_token=True)
         if refresh_result.status_code == codes.ok:
             if 'X-Rucio-Auth-Token-Expires' not in refresh_result.headers or \
                     'X-Rucio-Auth-Token' not in refresh_result.headers:
@@ -644,7 +644,7 @@ class BaseClient:
         request_auth_url = build_url(self.auth_host, path='auth/oidc')
         # requesting authorization URL specific to the user & Rucio OIDC Client
         self.logger.debug("Initial auth URL request headers %s to files" % str(headers))
-        oidc_auth_res = self._send_request(request_auth_url, headers=headers, get_token=True)
+        oidc_auth_res = self._send_request(request_auth_url, method=HTTPMethod.GET, headers=headers, get_token=True)
         self.logger.debug("Response headers %s and text %s" % (str(oidc_auth_res.headers), str(oidc_auth_res.text)))
         # with the obtained authorization URL we will contact the Identity Provider to get to the login page
         if 'X-Rucio-OIDC-Auth-URL' not in oidc_auth_res.headers:
@@ -665,7 +665,7 @@ class BaseClient:
                                            \nthe Rucio authentication server for a token.")
                 print("----------------------------------------------")
                 while time.time() - start < timeout:
-                    result = self._send_request(auth_url, headers=headers, get_token=True)
+                    result = self._send_request(auth_url, method=HTTPMethod.GET, headers=headers, get_token=True)
                     if 'X-Rucio-Auth-Token' in result.headers and result.status_code == codes.ok:
                         break
                     time.sleep(2)
@@ -675,7 +675,7 @@ class BaseClient:
                 while count < 3:
                     fetchcode = input()
                     fetch_url = build_url(self.auth_host, path='auth/oidc_redirect', params=fetchcode)
-                    result = self._send_request(fetch_url, headers=headers, get_token=True)
+                    result = self._send_request(fetch_url, method=HTTPMethod.GET, headers=headers, get_token=True)
                     if 'X-Rucio-Auth-Token' in result.headers and result.status_code == codes.ok:
                         break
                     else:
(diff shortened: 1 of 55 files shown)


lobis:
For some reason I got a `pyright` warning on `lib/rucio/client/uploadclient.py` without having modified the file, I couldn't find a quick fix for it so I added an ignore.

voetberg (changes requested):
Overall good changes! But I think you are missing a few instances of these constants. 

https://github.com/rucio/rucio/blob/c6d7f6f94194d831c85f35f4b3c4ceac703043f4/lib/rucio/rse/protocols/webdav.py#L264

https://github.com/rucio/rucio/blob/c6d7f6f94194d831c85f35f4b3c4ceac703043f4/lib/rucio/core/credential.py#L100

https://github.com/rucio/rucio/blob/c6d7f6f94194d831c85f35f4b3c4ceac703043f4/lib/rucio/web/rest/flaskapi/v1/common.py#L179

I'd do a more throughout search to make sure it's not just the clients using these constants and widen the scope of this PR a bit.

lobis:
> Overall good changes! But I think you are missing a few instances of these constants.
> 
> https://github.com/rucio/rucio/blob/c6d7f6f94194d831c85f35f4b3c4ceac703043f4/lib/rucio/rse/protocols/webdav.py#L264
> 
> https://github.com/rucio/rucio/blob/c6d7f6f94194d831c85f35f4b3c4ceac703043f4/lib/rucio/core/credential.py#L100
> 
> https://github.com/rucio/rucio/blob/c6d7f6f94194d831c85f35f4b3c4ceac703043f4/lib/rucio/web/rest/flaskapi/v1/common.py#L179
> 
> I'd do a more throughout search to make sure it's not just the clients using these constants and widen the scope of this PR a bit.

Thanks for the review. I expanded the scope of this PR to include most instances where `'GET'`, `'POST'`, etc. are mentioned explicitly, leaving only a few cases where I thought makes sense to leave them as they are.

voetberg (dismissed):

lib/rucio/common/constants.py
```diff
@@ -219,3 +219,16 @@ class RseAttr:
...
+    """HTTP verbs used in Rucio requests."""
+
+    # Standard HTTP/1.1
+    HEAD = 'HEAD'
+    OPTIONS = 'OPTIONS'
+    PATCH = 'PATCH'
+    GET = 'GET'
+    POST = 'POST'
+    PUT = 'PUT'
+    DELETE = 'DELETE'
```
rdimaio: I would be in favour of using the enum provided by the built-in `http` library instead: https://docs.python.org/3/library/http.html#http.HTTPMethod

lobis: I agree, unfortunately this can only be done when the minimum supported python version is `3.11`, so I have renamed the Rucio constant to match the `http` one and added a comment in the enum stating that it is to be removed once we set 3.11 as the minimum version.

lib/rucio/common/constants.py
```diff
@@ -219,3 +219,19 @@ class RseAttr:
 
 OPENDATA_DID_STATE_LITERAL = Literal['draft', 'public', 'suspended']
 OPENDATA_DID_STATE_LITERAL_LIST = list(get_args(OPENDATA_DID_STATE_LITERAL))
+
+@enum.unique
+class HTTPMethod(enum.Enum):
+    """HTTP verbs used in Rucio requests."""
+
+    # TODO: remove this enum in favor of standard library `http` module when the minimum supported Python version is >= 3.11
+    # Just replacing the `from rucio.common.constants import HttpMethod` to `from rucio.common.constants import HTTPMethod` should be enough
```
rdimaio:
```suggestion
    # Just replacing the `from rucio.common.constants import HTTPMethod` to `from rucio.common.constants import HTTPMethod` should be enough
```

lib/rucio/common/constants.py
rdimaio: Good catch. I also created this issue to track this, as TODO comments can be easy to forget about: https://github.com/rucio/rucio/issues/7959

lib/rucio/common/constants.py
```diff
@@ -219,3 +219,19 @@ class RseAttr:
 
 OPENDATA_DID_STATE_LITERAL = Literal['draft', 'public', 'suspended']
 OPENDATA_DID_STATE_LITERAL_LIST = list(get_args(OPENDATA_DID_STATE_LITERAL))
+
+@enum.unique
+class HTTPMethod(enum.Enum):
+    """HTTP verbs used in Rucio requests."""
+
+    # TODO: remove this enum in favor of standard library `http` module when the minimum supported Python version is >= 3.11
```
rdimaio: Would be good to add a link to the issue for doing this https://github.com/rucio/rucio/issues/7959

rdimaio (dismissed):

lib/rucio/common/constants.py
lobis: Thanks for spotting this!

lib/rucio/common/constants.py
lobis: Added!

lib/rucio/web/rest/flaskapi/v1/rules.py
```diff
@@ -827,21 +828,21 @@ def blueprint() -> AuthenticatedBlueprint:
     bp = AuthenticatedBlueprint('rules', __name__, url_prefix='/rules')
 
     rule_view = Rule.as_view('rule')
-    bp.add_url_rule('/<rule_id>', view_func=rule_view, methods=['get', 'put', 'delete'])
+    bp.add_url_rule('/<rule_id>', view_func=rule_view, methods=[HTTPMethod.GET.value, HTTPMethod.PUT.value, HTTPMethod.DELETE.value])
     all_rule_view = AllRule.as_view('all_rule')
-    bp.add_url_rule('/', view_func=all_rule_view, methods=['get', 'post'])
+    bp.add_url_rule('/', view_func=all_rule_view, methods=[HTTPMethod.GET.value, HTTPMethod.POST.value])
```
rdimaio, line 833: Small question - do we expect any possible issue that might arise by substituting these lowercase methods with their uppercase equivalents?

lobis: No, I checked the flask source code and it converts to upper internally when submitting a string

lib/rucio/client/baseclient.py
```diff
@@ -374,7 +387,9 @@ def _get_creds(self, creds: Optional[dict[str, Any]]) -> dict[str, Any]:
 
         return creds
 
-    def _get_exception(self, headers: dict[str, str], status_code: Optional[int] = None, data=None) -> tuple[type[exception.RucioException], str]:
+    def _get_exception(self, headers: "CaseInsensitiveDict[str]", status_code: Optional[int] = None, data=None) -> \
+    tuple[
+        type[exception.RucioException], str]:
```
rdimaio: Instead of using the backwards slash, it would be better if you could use the multi-line convention that we use for long function signatures, e.g.:

https://github.com/rucio/rucio/blob/749e15c7319caa8a2c791c83215129b919dd8990/lib/rucio/core/rse.py#L138-L150

(thought I don't think we have this actually documented anywhere, apologies about that)

lib/rucio/client/baseclient.py
```diff
@@ -583,11 +617,14 @@ def __refresh_token_oidc(self) -> bool:
...
-            pass
         else:
-            return False
+            exp_epoch = float(self.token_exp_epoch)
+            refresh_before = float(self.auth_oidc_refresh_before_exp) * 60
+            if exp_epoch - refresh_before < time.time() < exp_epoch:
+                # attempt to refresh token
+                pass
+            else:
+                return False
```
rdimaio: I'm confused about this change, could you clarify?

lobis: The linter was complaining about comparing floats to ints if I recall correctly, I explicitly convert to float to avoid the error

rdimaio: I see - it's best to have these changes in a separate commit, otherwise it can be mistaken for a functional change that's related to the rest of the commit that it's part of (and in general, it's a lot easier to review and have a clean commit history when linting changes are separate from functional changes)

lib/rucio/client/baseclient.py
lobis: Addressed!

lib/rucio/client/baseclient.py
lobis: I agree, because this PR was such a big diff I might have made a mistake when reordering commits, I have done again some reordering, hopefully its more clear now.

lib/rucio/client/baseclient.py
```diff
@@ -583,11 +605,14 @@ def __refresh_token_oidc(self) -> bool:
         if self.token_exp_epoch is None:
             # check expiration time for a new token
             pass
-        elif time.time() > self.token_exp_epoch - self.auth_oidc_refresh_before_exp * 60 and time.time() < self.token_exp_epoch:
-            # attempt to refresh token
-            pass
         else:
-            return False
+            exp_epoch = float(self.token_exp_epoch)
```
rdimaio: But it looks like specifically cast this to int when we initially define it (if defined):

```
self.token_exp_epoch = int(token_epoch_file.readline())
```

So it might make more sense to just cast that to float from that point onwards, instead of casting initially as int and then as float here.

Overall I think it might make more sense to compare as datetime objects (https://stackoverflow.com/questions/58242561/how-to-compare-datetime-object-with-an-integer)

But I think on this point in particular, it would be better if a token component lead reviewed this

lib/rucio/client/baseclient.py
```diff
@@ -374,7 +389,12 @@ def _get_creds(self, creds: Optional[dict[str, Any]]) -> dict[str, Any]:
 
         return creds
 
-    def _get_exception(self, headers: dict[str, str], status_code: Optional[int] = None, data=None) -> tuple[type[exception.RucioException], str]:
+    def _get_exception(
+            self,
+            headers: "CaseInsensitiveDict[str]",
```
rdimaio: I think a type annotation change would best be separated in a separate commit, I think that ' Address linter warnings (no functional changes).' hides it a bit (even if this change was caused by a linter warning in the first place), but I think it's fine in this case.

lib/rucio/client/baseclient.py
```diff
@@ -447,7 +467,7 @@ def _send_request(self, url, headers=None, type_='GET', data=None, params=None,
 
         :param url: the http url to use.
         :param headers: additional http headers to send.
-        :param type_: the http request type to use.
+        :param method: the http request type to use (GET, POST, PUT, DELETE).
```
rdimaio: Why was this changed? The function signature says `type_`

rdimaio: I see that `type_` changes to `method` in the next commit. I know it's annoying, but for consistency, this change should be part of that commit, not this one.

lib/rucio/client/baseclient.py
```diff
@@ -489,29 +489,44 @@ def _send_request(self, url, headers=None, type_='GET', data=None, params=None,
...
+                f"HTTP method {method} is not supported. Supported methods are: {list(request_map.keys())}")
+
+        self.logger.debug("HTTP request: %s %s" % (method.value, url))
         for h, v in hds.items():
             if h == 'X-Rucio-Auth-Token':
                 v = "[hidden]"
             self.logger.debug("HTTP header:  %s: %s" % (h, v))
-        if type_ != "GET" and data:
+
+        if method != HTTPMethod.GET and data is not None:
```
rdimaio: `and data` and `and data is not None` are two different checks. Are we sure that we need to change it like this here?

lib/rucio/client/baseclient.py
rdimaio: (for reference, this comment was left on commit eafd7dc222c8dc03d80d3a7eaa6de209b00ff823)

lobis: Agree, I spent quite some time reorganizing but still you get errors sometimes, will join with the relevant commit

lib/rucio/client/baseclient.py
```diff
@@ -440,14 +440,14 @@ def _back_off(self, retry_number: int, reason: str) -> None:
         self.logger.warning("Waiting {}s due to reason: {} ".format(sleep_time, reason))
         time.sleep(sleep_time)
 
-    def _send_request(self, url, headers=None, type_='GET', data=None, params=None, stream=False, get_token=False,
+    def _send_request(self, url, headers=None, method : HTTPMethod = HTTPMethod.GET, data=None, params=None, stream=False, get_token=False,
```
rdimaio: type issue in CI likely related to typing with an enum, see https://stackoverflow.com/questions/52624736/type-annotations-for-enum-attribute

voetberg (commented):
LGTM, just squish those last few commits and I'd happily approve

voetberg (dismissed):

rdimaio:
> LGTM, just squish those last few commits and I'd happily approve

I like the idea of having `git squish` instead of `git squash`

lobis:
Had to rebase due to conflicts, asked for review again, thanks!

lib/rucio/client/baseclient.py
rdimaio, line 1: LGTM, but not sure if we should merge this in a minor release as it seems like it could potentially be breaking (though it shouldn't) - @bari12 thoughts?

rdimaio (approved):

lib/rucio/client/baseclient.py
lobis: I think it shouldn't break anything unless a mistake was done, but I agree it has to be looked at carefuly as it's a big diff.

lobis: image

voetberg (approved):

lib/rucio/client/baseclient.py
rdimaio: Checked with Martin, fine to include this in a minor release

PR #7927, opened 2025-08-12
Docker & Kubernetes: Update dev Oracle container configs; #7918

### LINKED ISSUE #7918: Docker & Kubernetes: Update the oracle container in the docker-compose
### Description

The currently used `dev-oracle-1` container inside dev's docker-compose is actually misconfigured as the passed environment entries have little to none effect:

https://github.com/rucio/rucio/blob/769023e6a399306f7b2323a703a4399855c9a4c5/etc/docker/dev/docker-compose.yml#L153-L167

We should fix this. Also we should ideally update the image with the latest available XE version (>= 19c).

### Steps to reproduce

-

### Rucio Version

Server: 38.0.0rc2

### Additional Information

_No response_

Geogouz:
Fixes #7918 

To make current parameters in the dev's Oracle container take effect, I apply those now via an SQL script. Before they were not recognised.

The new values have been roughly derived from a formula based on:
https://docs.oracle.com/en/database/oracle/oracle-database/19/refrn/SESSIONS.html

and

https://docs.oracle.com/en/database/oracle/oracle-database/19/refrn/TRANSACTIONS.html

The processes were lowered to 200 because if we sticked to 1000, the SGA overhead alone would quickly exhaust the available (capped to 2GB from oracle) memory.


etc/docker/dev/docker-compose.yml
```diff
@@ -158,13 +158,8 @@ services:
       - oracle
     environment:
       - ORACLE_PASSWORD=rucio
-      - ORACLE_ALLOW_REMOTE=true
```
rdimaio, line 200: Seems like this one is no longer configured anywhere, is that an issue?

etc/docker/dev/oracle/01_tune_oracle_limits.sql
```diff
@@ -0,0 +1,6 @@
+-- Keep in mind that XE is limited to 2 GB RAM and 2 CPU threads.
+-- Very large PROCESSES values consume SGA memory and can hit that ceiling quickly..
+ALTER SYSTEM SET processes      = 200   SCOPE=SPFILE;
+ALTER SYSTEM SET sessions       = 335   SCOPE=SPFILE;
+ALTER SYSTEM SET transactions   = 369   SCOPE=SPFILE;
```
rdimaio, line 5: The comments here give some good context, but I think it could be worth pointing out in the PR or commit message as well why you changed these 3 values (e.g. processes from 1000 to 200), particularly sessions and transactions

etc/docker/dev/docker-compose.yml
Geogouz: It's the default behavior

etc/docker/dev/oracle/01_tune_oracle_limits.sql
Geogouz: Had been roughly derived from a formula that was based on:
https://docs.oracle.com/en/database/oracle/oracle-database/19/refrn/SESSIONS.html

and

https://docs.oracle.com/en/database/oracle/oracle-database/19/refrn/TRANSACTIONS.html

The processes were lowered to 200 because if we sticked to 1000, the SGA overhead alone would quickly exhaust the available (capped to 2GB from oracle) memory.

rdimaio (approved):

PR #7932, opened 2025-08-14
Kronos: Increment DID access_cnt in touch_replica

### LINKED ISSUE #6832: Kronos does not update the access_cnt for files

0xquark:
Fixes #6832

Kronos updates `accessed_at` but not `access_cnt` for files when calling `touch_replica`. 
Updated touch_replica to increment the DID-level access_cnt using the same NULL-safe logic as `touch_dids` in https://github.com/rucio/rucio/blob/b6854f44e1e3bf003497256aa0a1c6dc3e00628c/lib/rucio/core/did.py#L2865-L2866

### DIFF
--- a/lib/rucio/core/replica.py
+++ b/lib/rucio/core/replica.py
@@ -2874,7 +2874,11 @@ def touch_replica(
         ).prefix_with(
             '/*+ INDEX(DIDS DIDS_PK) */', dialect='oracle'
         ).values({
-            models.DataIdentifier.accessed_at: accessed_at
+            models.DataIdentifier.accessed_at: accessed_at,
+            models.DataIdentifier.access_cnt: case(
+                (models.DataIdentifier.access_cnt == none_value, 1),
+                else_=(models.DataIdentifier.access_cnt + 1)
+            )  # type: ignore
         }).execution_options(
             synchronize_session=False
         )
--- a/tests/test_replica.py
+++ b/tests/test_replica.py
@@ -30,7 +30,7 @@ from rucio.common.constants import RseAttr
 from rucio.common.exception import AccessDenied, DatabaseException, DataIdentifierNotFound, InputValidationError, ReplicaIsLocked, ReplicaNotFound, RucioException, ScopeNotFound
 from rucio.common.utils import clean_pfns, generate_uuid, parse_response
 from rucio.core.config import set as cconfig_set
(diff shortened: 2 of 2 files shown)


0xquark:
Regarding the comment https://github.com/rucio/rucio/pull/6851#discussion_r1682712098

> I'm also having doubts on whether it's a good idea to increment the access_cnt on the DID as part of touch_replica - maybe the DID access_cnt should only be incremented by touch_dids, and a separate replica access_cnt field should be added to the RSEFileAssociation and CollectionReplica tables, considering that they already have an accessed_at field.

The replica table (RSEFileAssociation) doesn’t have an access_cnt column today; it only has accessed_at. Same for CollectionReplica. Adding per-replica counters would require schema migrations, model changes, and maybe a broader agreement on semantics. If we want to continue with the per-replica counters, I wanted to understand this

What exactly counts as an “access” at the RSE level? Every touch (including internal/daemon touches)? What events would be counted and the use-case for per-RSE access counts?

rdimaio:
> Regarding the comment [#6851 (comment)](https://github.com/rucio/rucio/pull/6851#discussion_r1682712098)
> 
> > I'm also having doubts on whether it's a good idea to increment the access_cnt on the DID as part of touch_replica - maybe the DID access_cnt should only be incremented by touch_dids, and a separate replica access_cnt field should be added to the RSEFileAssociation and CollectionReplica tables, considering that they already have an accessed_at field.
> 
> The replica table (RSEFileAssociation) doesn’t have an access_cnt column today; it only has accessed_at. Same for CollectionReplica. Adding per-replica counters would require schema migrations, model changes, and maybe a broader agreement on semantics. If we want to continue with the per-replica counters, I wanted to understand this
> 
> What exactly counts as an “access” at the RSE level? Every touch (including internal/daemon touches)? What events would be counted and the use-case for per-RSE access counts?

I think we need to decise on this first before we can merge. @bari12 thoughts? see this comment too for reference https://github.com/rucio/rucio/pull/6851#discussion_r1682712098

bari12 (approved):
To me this is fine; I see the inconsistency, but nobody ever asked for individual access counting on the replicas, so I would just leave it like that.

PR #7939, opened 2025-08-18
Database: Add Oracle thick mode initialization and fallback to thin mode

### LINKED ISSUE #7907: Support Oracle thick client in Rucio
### Description

Rucio currently uses Oracle’s thin client exclusively, leaving no way to leverage the thick client even when Oracle Instant Client is installed. We should add configuration options to enable and initialize the thick client at startup—falling back to thin mode if unavailable—and update documentation and tests accordingly.

### Steps to reproduce

.

### Rucio Version

Server 38.0.0rc2

### Additional Information

_No response_

Geogouz:
Fixes #7907

As discussed in the issue, this is a minimal temporary solution that should unfreeze ATLAS oracle-based instance.

_Note:_
The message at line [1483](https://github.com/rucio/rucio/actions/runs/17041440904/job/48306017375?pr=7939#step:4:1483) showing:
`
Could not start Oracle thick mode; falling back to thin: DPI-1047: Cannot locate a 64-bit Oracle Client library: "libclntsh.so: cannot open shared object file: No such file or directory". See [https://python-oracledb.readthedocs.io/en/latest/user\_guide/initialization.html](https://python-oracledb.readthedocs.io/en/latest/user_guide/initialization.html) for help
`
is expected to be like that in our tests currently.

### DIFF
--- a/lib/rucio/db/sqla/session.py
+++ b/lib/rucio/db/sqla/session.py
@@ -38,6 +38,8 @@ from rucio.db.sqla.constants import DatabaseOperationType
 
 EXTRA_MODULES = import_extras(['MySQLdb', 'pymysql'])
 
+LOG = logging.getLogger(__name__)
+
 if TYPE_CHECKING:
     from collections.abc import Callable, Iterator
     from typing import Optional, ParamSpec, TypeVar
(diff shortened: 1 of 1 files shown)


Geogouz:
Should be ready @bari12, @dchristidis.
[Locally](https://github.com/Geogouz/rucio/blob/e5dc15307823966e186902672b47f124a958f28a/lib/rucio/db/sqla/session.py#L246), I am also using a path that can be overridden in case the client is not at the default place, but since the working example you shared also didn't use any path, we should be fine. We can merge this and give it a test.

rdimaio (approved):

bari12 (approved):

PR #7940, opened 2025-08-19
Testing: Add `needs_iam` marker to run tests with the IAM profile

### LINKED ISSUE #7933: Add pytest marker `needs_iam` for tests that require the dev iam profile

0xquark:
Fixes #7933 

Most of our test suites (both locally and in CI) use profiles, I introduced a `DEV_PROFILES` variable to get the profiles while running the tests and skip tests with the markers that doesn't have the profile required to run the test while also warning the user about it. 

Only these three tests need IAM profiles:
```
-  test_conveyer::test_transfer_with_tokens
-  test_reaper::test_deletion_with_tokens
-  test_Conveyer::test_bittorrent_submission
```

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -123,7 +123,9 @@ jobs:
           sed -i 's;image: docker.io/rucio/rucio-dev.*;image: ${{ fromJSON(steps.images.outputs.images)[0] }};' \
               $GITHUB_WORKSPACE/dev/rucio/etc/docker/dev/docker-compose.yml
       - name: Start containers
-        run: docker compose -f $GITHUB_WORKSPACE/dev/rucio/etc/docker/dev/docker-compose.yml --profile storage --profile externalmetadata --profile iam up -d
+        run: |
+          export DEV_PROFILES=storage,externalmetadata,iam
+          docker compose -f $GITHUB_WORKSPACE/dev/rucio/etc/docker/dev/docker-compose.yml --profile storage --profile externalmetadata --profile iam up -d
       - name: Initialize tests
         shell: bash
         run: |
--- a/etc/docker/dev/docker-compose.yml
+++ b/etc/docker/dev/docker-compose.yml
@@ -61,6 +61,7 @@ services:
       - REQUESTS_CA_BUNDLE=/etc/pki/tls/certs/ca-bundle.crt
       - RUCIO_SOURCE_DIR=/rucio_source
       - RDBMS
+      - DEV_PROFILES
     depends_on:
       oracle:
         condition: service_healthy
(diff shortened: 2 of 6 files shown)


rdimaio (commented):
To check the changes: we can only check `test_deletion_with_tokens`, because the other test is currently being skipped due to an existing ticket.

`test_deletion_with_tokens` is being run as expected here:
https://github.com/rucio/rucio/actions/runs/17066724292/job/48385560557?pr=7940#step:21:29

I guess we don't have any end to end setups where iam is not set up. Could you run this locally without the iam profile to see if the exclusion is working as expected? Or possibly you could run a gh workflow without spinning up the iam profile (so changing `workflows/integration_tests`, to check this

0xquark:
> To check the changes: we can only check `test_deletion_with_tokens`, because the other test is currently being skipped due to an existing ticket.

For `test_transfer_with_tokens`, I removed the other markers and only included the`needs_iam` marker  to test when running it inside a dev container with the default profile. It worked as expected. 
image

Similarly for `test_deletion_with_tokens` , I tested it locally using the dev-container which worked as expected.
image

Also noticed that the `test_bittorrent_submission` also requires iam as it tries to get the OIDC tokens, which fails when the IAM profile is not running. Therefore, Added the marker for this test as well.

image

Apologies for not including the screenshots before.

> 
> `test_deletion_with_tokens` is being run as expected here: https://github.com/rucio/rucio/actions/runs/17066724292/job/48385560557?pr=7940#step:21:29
> 
> I guess we don't have any end to end setups where iam is not set up. Could you run this locally without the iam profile to see if the exclusion is working as expected? Or possibly you could run a gh workflow without spinning up the iam profile (so changing `workflows/integration_tests`, to check this
 
I've ran integration tests without the `iam` profile, lets see if there are any more places we need to make changes.

rdimaio:
> > To check the changes: we can only check `test_deletion_with_tokens`, because the other test is currently being skipped due to an existing ticket.
> 
> For `test_transfer_with_tokens`, I removed the other markers and only included the`needs_iam` marker to test when running it inside a dev container with the default profile. It worked as expected. image
> 
> Similarly for `test_deletion_with_tokens` , I tested it locally using the dev-container which worked as expected. image
> 
> Also noticed that the `test_bittorrent_submission` also requires iam as it tries to get the OIDC tokens, which fails when the IAM profile is not running. Therefore, Added the marker for this test as well.
> image
> 
> Apologies for not including the screenshots before.
> 
> > `test_deletion_with_tokens` is being run as expected here: https://github.com/rucio/rucio/actions/runs/17066724292/job/48385560557?pr=7940#step:21:29
> > I guess we don't have any end to end setups where iam is not set up. Could you run this locally without the iam profile to see if the exclusion is working as expected? Or possibly you could run a gh workflow without spinning up the iam profile (so changing `workflows/integration_tests`, to check this
> 
> I've ran integration tests without the `iam` profile, lets see if there are any more places we need to make changes.

Nice, thanks!

rdimaio (approved):

PR #7929, opened 2025-08-13
Testing: Add support for running tests after bootstrapping the dev environment; rucio#7928

Geogouz:
**Draft Notice:** This works but kept frozen until we refactor our CI. Then I will re-adjust.
---

Fixes #7928

**Summary**
This PR extends the `./tools/bootstrap_dev.sh` script to optionally run (using -t/--test) a selected test suite immediately after setup. The script lists the available test options, validates the chosen test number, and starts/stops any required services for the selected scenario automatically when autotests are used. Ideally the number of available tests should match at the end the one we have in our CI (e.g. 31). Keep in mind that this PR depends on #7910 so we should merge that first (and then I could rebase).

**Usage**
- Run a specific test right after bootstrapping (e.g. `./tools/bootstrap_dev.sh -t 1` - for using the local test suite)
- Tests run against a clean database; the dev DB volume is reset to avoid state carryover.
- The available/offered autotests are detected automatically based on our matrix.
- The -f/--filter option can be used with -t/--test to pass any pytest-style filter such as 'tests/file.py::TestClass::test_method' to run a subset of tests.
    - With autotests (i.e. when `-t > 1` is selected), this should work fine. However, when `-t 1` is used (i.e. local test suite), since currently the default docker-compose profile is only allowed to be spinned up for that, you should select a pytest-filter that is supposed to run with those default containers. This should ideally be changed in the future so that the proper profile is automatically selected. But this needs a bit of discussion of how we should do it.
- The -c/--cache-use option skips rebuilding images if a cached version exists.
- In the future, I can also make an official instructions guide about this script's usage (to be placed on our documentation page), but for now, the complete instructions can be found in more detail at the top of `./tools/bootstrap_dev.sh` itself.

    _Example:_ 
    Run a single pytest within `multi_vo (dist alma9, py 3.9, db postgres14)` test suite using a filter:
    `./tools/bootstrap_dev.sh --test 9 --filter tests/test_scope.py::test_scope_duplicate`
    or just
    `./tools/bootstrap_dev.sh -t 9 -f tests/test_scope.py::test_scope_duplicate`

**Note**
This is a quite safe-to-merge PR since it affects only the `bootstrap_dev.sh` functionality which is used only by devs. However, since this convenience tool is still under active development and may occasionally break. If you encounter unexpected behavior, please open an issue so we can investigate and fix it. I already know couple edge-cases where upgrades are needed.

**Currently supported tests**
``` text
Available tests:
   1) run local test suite (tools/run_tests.sh)
[...]

PR #7953, opened 2025-08-26
Tooling: Add CI-aligned devcontainer; See #7906

### LINKED ISSUE #7906: Tooling: Add `devcontainer.json` file
### Description

Add Dev Container support to the Rucio project by including a `.devcontainer/devcontainer.json` configuration file. This would enable seamless development environment setup in VS Code, GitHub Codespaces, and other Dev Container-compatible IDEs.

### Motivation

- **Developer onboarding**: New contributors currently need to manually set up Docker, understand docker-compose commands, and configure their IDE separately. Dev Containers would provide one-click environment setup.
- **Consistency**: Ensures all developers work in identical environments, reducing "works on my machine" issues.
- **Modern workflow**: Dev Containers are becoming the standard for containerized development, supported natively by VS Code, GitHub Codespaces, and JetBrains IDEs.
- **Cloud development**: Enables easy development in GitHub Codespaces or other cloud-based development environments.
- **IDE integration**: Provides proper Python tooling, debugging, and test runner configuration out of the box.

The existing Docker development setup is excellent but requires manual IDE configuration and docker-compose knowledge that creates friction for new contributors.

### Change

Add a `.devcontainer/devcontainer.json` file that:

1. **Leverages existing infrastructure**: References the current `etc/docker/dev/docker-compose.yml` to avoid duplicating container definitions
2. **Configures development tools**: Pre-installs and configures Python extensions (pylint, flake8, ruff) and test runners
3. **Sets up workspace**: Points to the correct working directory (`/opt/rucio`) and Python interpreter
4. **Provides guidance**: Includes post-create commands to help developers get started with `tools/run_tests.sh -i`
5. **Maintains flexibility**: Developers can still use different compose profiles (storage, monitoring) as needed

The implementation would be minimal and non-disruptive - a single JSON file that enhances the existing Docker workflow without changing it. Developers could continue using the traditional `docker compose` approach while new contributors benefit from modern IDE integration.

**Example structure:**
```json
{
  "name": "Rucio Development",
  "dockerComposeFile": "etc/docker/dev/docker-compose.yml",
  "service": "rucio",
  "workspaceFolder": "/opt/rucio",
  // ... IDE extensions and settings
}
```

This would make Rucio development more accessible while preserving all existing workflows.

mgajek-cern:
Closes #7906 

Major changes:

 - Add .devcontainer/alma9/ with devcontainer.json using rucio-dev:release-38.0.0
    - Configure VS Code Python extension and Node.js/Git features
    - Add common_functions.sh with Docker/system checks, service management, and CI command display
    - Add initialize.sh to start external services with configurable profiles (storage, externalmetadata, etc.)
    - Add setup.sh to install dev tools (ruff, flake8, pytest, pyright) and Rucio in editable mode
    - Add help.sh showing CI-aligned commands matching GitHub Actions workflows
    - Support profile customization via rucio_dev_profiles.cfg
    - Mount Docker socket, enable privileged mode, and connect container to dev_default network
    - Set PYTHONPATH, RUCIO_HOME, and RUCIO_POLICY_PACKAGE environment variables
 
Improves developer onboarding and ensures local testing matches CI environment.

---

Screenshots of dev container setup in VS Code:

- Screenshots on `docker exec -t dev-rucio-1 tools/run_tests.sh -ir` command:
image
image

- Screenshot on `pytest tests/rucio --cov=lib/rucio`:
image

- Screenshot on `Run Tests`:
image

- TODO(mgajek-cern): Video presenting evidence of debugging the server component and other components as outlined in https://github.com/rucio/rucio/pull/7953#issuecomment-3275297837

### DIFF
--- a/.devcontainer/alma9/common_functions.sh
+++ b/.devcontainer/alma9/common_functions.sh
@@ -1,0 +1,323 @@
+#!/bin/bash
+# -*- coding: utf-8 -*-
+# Copyright European Organization for Nuclear Research (CERN) since 2012
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Common functions for Rucio development environment setup
+
+# Color definitions
+BLUE='\033[0;34m'
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+YELLOW='\033[1;33m'
+CYAN='\033[0;36m'
+MAGENTA='\033[0;35m'
+NC='\033[0m'
+
+# Print a standardized header
+print_header() {
+    local title="$1"
+    echo -e "${BLUE}╔══════════════════════════════════════════════════════════════╗${NC}"
+    echo -e "${BLUE}║$(printf "%62s" "$title")║${NC}"
+    echo -e "${BLUE}╚══════════════════════════════════════════════════════════════╝${NC}"
+    echo ""
+}
+
+# Check system requirements
+check_system_requirements() {
+    echo -e "${CYAN}📋 Checking System Requirements:${NC}"
+    
+    # Check available disk space (current directory)
+    AVAILABLE_DISK_GB=$(df . | tail -1 | awk '{print int($4/1024/1024)}')
+    echo -e "   Disk Space: ${AVAILABLE_DISK_GB}GB available in current directory"
+    
+    if [ "$AVAILABLE_DISK_GB" -lt 20 ]; then
+        echo -e "${RED} ❌ Warning: Less than 20GB disk space available. Build may fail.${NC}"
+        echo -e "${YELLOW}    Consider freeing up disk space before continuing.${NC}"
+    else
+        echo -e "${GREEN} ✅ Sufficient disk space available${NC}"
+    fi
+    
+    # Check Docker daemon
+    echo -e "   Docker: Checking daemon status..."
+    if ! docker info > /dev/null 2>&1; then
+        echo -e "${RED} ❌ Docker daemon not accessible. Please ensure Docker is running.${NC}"
+        exit 1
+    else
+        echo -e "${GREEN} ✅ Docker daemon is running${NC}"
+    fi
+    echo ""
+}
+
+# Install Docker CLI for AlmaLinux
+install_docker_cli() {
+    echo -e "${BLUE}🐳 Installing Docker CLI and tools...${NC}"
+    if ! command -v docker &> /dev/null 2>&1; then
+        dnf config-manager --add-repo https://download.docker.com/linux/centos/docker-ce.repo
+        dnf install -y docker-ce-cli docker-compose-plugin
+        echo -e "${GREEN} ✅ Docker CLI installed${NC}"
+    else
+        echo -e "${GREEN} ✅ Docker CLI already available${NC}"
+    fi
+}
+
+# Install Python development tools
+install_python_dev_tools() {
+    echo -e "${BLUE}🐍 Installing Python development tools...${NC}"
+    pip install ruff flake8 flake8-annotations PyYAML pytest pytest-cov
+    echo -e "${GREEN} ✅ Python tools installed${NC}"
+}
+
+# Install Node.js tools
+install_nodejs_tools() {
+    echo -e "${BLUE}📦 Installing Node.js tools...${NC}"
+    if command -v npm &> /dev/null 2>&1; then
+        npm install --global pyright
+        echo -e "${GREEN} ✅ Node.js tools installed${NC}"
+    else
+        echo -e "${YELLOW} ⚠️ npm not found, skipping Node.js tools${NC}"
+    fi
+}
+
+# Install Rucio in development mode
+install_rucio_dev() {
+    echo -e "${BLUE}📚 Installing Rucio in development mode...${NC}"
+    pip install -e .
+    echo -e "${GREEN} ✅ Rucio installed in development mode${NC}"
+}
+
+# Check Docker daemon with timeout
+check_docker_daemon() {
+    echo -e "${BLUE}🔍 Checking Docker daemon...${NC}"
+    if timeout 30 bash -c 'until docker info > /dev/null 2>&1; do sleep 1; done'; then
+        echo -e "${GREEN} ✅ Docker daemon is ready${NC}"
+    else
+        echo -e "${RED} ❌ Docker daemon failed to start within 30 seconds${NC}"
+        exit 1
+    fi
+}
+
+cleanup_existing_containers() {
+    echo -e "${YELLOW}🔄 Cleaning up existing containers...${NC}"
+
+    # Remove containers prefixed with 'dev-'
+    echo -e "${YELLOW}🧹 Removing containers with prefix 'dev-'...${NC}"
+    containers=$(docker ps -a --filter "name=^dev-" --format "{{.ID}}")
+    if [ -n "$containers" ]; then
+        docker rm -f $containers
+        echo -e "${GREEN} ✅ Containers removed${NC}"
+    else
+        echo -e "${GREEN} ✅ No 'dev-' containers to remove${NC}"
+    fi
+
+    # Remove volumes prefixed with 'dev_'
+    echo -e "${YELLOW}🧹 Removing volumes with prefix 'dev_'...${NC}"
+    volumes=$(docker volume ls --filter "name=^dev_" --format "{{.Name}}")
+    if [ -n "$volumes" ]; then
+        docker volume rm $volumes
+        echo -e "${GREEN} ✅ Volumes removed${NC}"
+    else
+        echo -e "${GREEN} ✅ No 'dev_' volumes to remove${NC}"
+    fi
+
+    # Remove network named 'dev_default' if exists
+    echo -e "${YELLOW}🧹 Removing 'dev_default' network if exists...${NC}"
+    if docker network ls --filter "name=^dev_default$" --format "{{.Name}}" | grep -q 'dev_default'; then
+        docker network rm dev_default
+        echo -e "${GREEN} ✅ Network 'dev_default' removed${NC}"
+    else
+        echo -e "${GREEN} ✅ Network 'dev_default' not found${NC}"
+    fi
+
+    echo ""
+}
+
+# Start external services
+start_external_services() {    
+    # Accept profiles as first argument, fallback to env var, then default
+    local input_profiles="$1"
+    PROFILES=${input_profiles:-${RUCIO_DEV_PROFILES:-"storage,externalmetadata"}}
+    
+    echo -e "${CYAN}Starting external services for Rucio development environment...${NC}"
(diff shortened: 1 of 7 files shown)


mgajek-cern:
I’ve added several potential reviewers here since I’m new and not sure who’s best for this topic. I think having 1 to 3 reviews should be enough to cover it. If anyone’s unable to review or prefers to be removed, just let me know and I’ll update the list. Thanks

.devcontainer/alma9/common_functions.sh
```diff
@@ -0,0 +1,274 @@
...
+install_docker_cli() {
+    echo -e "${BLUE}🐳 Installing Docker CLI and tools...${NC}"
+    if ! command -v docker &> /dev/null 2>&1; then
+        dnf config-manager --add-repo https://download.docker.com/linux/centos/docker-ce.repo
+        dnf install -y docker-ce-cli docker-compose-plugin
+        echo -e "${GREEN} ✅ Docker CLI installed${NC}"
+    else
+        echo -e "${GREEN} ✅ Docker CLI already available${NC}"
+    fi
+}
```
rdimaio, line 73: Could be worth using podman instead from the getgo here, since we sort of have support for it around Rucio, and in theory it'd be better if we relied on that instead of Docker

.devcontainer/alma9/common_functions.sh
```diff
@@ -0,0 +1,274 @@
...
+    echo -e "${YELLOW}🧹 Removing 'dev_default' network if exists...${NC}"
+    if docker network ls --filter "name=^dev_default$" --format "{{.Name}}" | grep -q 'dev_default'; then
+        docker network rm dev_default
+        echo -e "${GREEN} ✅ Network 'dev_default' removed${NC}"
+    else
+        echo -e "${GREEN} ✅ Network 'dev_default' not found${NC}"
+    fi
+
+    echo ""
+}
```
rdimaio, line 144: Assuming this is all within the devcontainer, is there no way to just clean up stuff within the devcontainer namespace?

.devcontainer/alma9/devcontainer.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "Rucio alma9 Development Environment",
+  "image": "rucio/rucio-dev:release-38.0.0",
```
rdimaio, line 3: Maybe we should always point to latest here?

rdimaio (commented):
Overall this seems to useful and to make sense, but I'm only concerned about the effort that would be needed to maintain this logic in sync with the rest of the things - for example, maintaining the profiles here in sync with the Docker compose profiles, maintaining the integ test list in sync, etc. it would be great if these things could be put in some sort of single place where they can then be imported in both places, to avoid this issue.

I am also curious about the system requirements, particularly the disk space - 30+ GB for full testing seems excessive, did you test this?

I think the best person to review this would be @maany as he developed the Rucio VSCode dev environment https://github.com/maany/rucio-vscode-dev-env, so he might have a lot of context on these aspects

mgajek-cern:
> Overall this seems to useful and to make sense, but I'm only concerned about the effort that would be needed to maintain this logic in sync with the rest of the things - for example, maintaining the profiles here in sync with the Docker compose profiles, maintaining the integ test list in sync, etc. it would be great if these things could be put in some sort of single place where they can then be imported in both places, to avoid this issue.
> 
> I am also curious about the system requirements, particularly the disk space - 30+ GB for full testing seems excessive, did you test this?
> 
> I think the best person to review this would be @maany as he developed the Rucio VSCode dev environment https://github.com/maany/rucio-vscode-dev-env, so he might have a lot of context on these aspects

Thanks for the feedback @rdimaio. I'll try to sync with @maany this or subsequent week on the setup if feasible and share ideas on whether we should continue with these efforts

.devcontainer/alma9/common_functions.sh
mgajek-cern: Is the systemd/daemon support the main rationale for preferring Podman here, or are there other considerations (like existing Rucio infrastructure alignment, rootless containers, etc.) we should factor in?

The [documentation](https://github.com/rucio/rucio/blob/master/etc/docker/dev/README.rst) states:

> We provide a containerised version of the Rucio development environment for a quick start. Our containers are ready-made for Docker, but we provide best-effort support for Podman and podman-compose. If you use podman-compose, you need to add -t cntnet to the podman-compose commands. For simplicity all commands explained in this documentation are using docker-compose.

.devcontainer/alma9/common_functions.sh
mgajek-cern: The devcontainer can't reliably handle docker-compose setup with complex volume mount paths for certificates and keys in `etc/docker/dev/docker-compose.yml` that must resolve correctly on the host filesystem (will reproduce and share screenshots if needed). We use `initializeCommand` to run the compose stack on the host before devcontainer startup, ensuring proper path resolution. 

Within the devcontainer itself, we access the host's Docker daemon through the mounted docker.sock, allowing container management from inside the devcontainer.

.devcontainer/alma9/devcontainer.json
mgajek-cern: We could configure Dependabot to automatically update the image version in devcontainer.json files. This would provide reproducible environments by default while ensuring automatic updates when new stable releases are available, rather than always pulling the potentially unstable latest tag.

This approach gives you both stability and currency - developers get consistent environments, but you don't fall behind on important updates.

Geogouz:
My worries here is that I see too much overlapping functionality with the currently existing `bootstrap_dev.sh` which is on hold till our ongoing CI adjustments are ready, in order to be also adjusted if needed (see #7929). Currently that bootstrapping dev environment does work and is quite feature-full (but unfortunately its documentation lives at the moment only within the script itself). Nevertheless, this PR does offer couple features that are unique or follow nice practices.

Another point is that for a new developer, seeing a dev container under Rucio/containers, and yet, another dev container here, might bring questions.

So for the shake of keeping things minimal and concentrated, and avoiding to maintain two parallel bootstrapping paths that almost do the same thing (not to mention that Rucio's web-documentation would end up being quite awkward for the developers and unclear when to use what), I would suggest to keep a single approach by merging the best of those two worlds.

mgajek-cern:
> My worries here is that I see too much overlapping functionality with the currently existing `bootstrap_dev.sh` which is on hold till our ongoing CI adjustments are ready, in order to be also adjusted if needed (see #7929). Currently that bootstrapping dev environment does work and is quite feature-full (but unfortunately its documentation lives at the moment only within the script itself). Nevertheless, this PR does offer couple features that are unique or follow nice practices.
> 
> Another point is that for a new developer, seeing a dev container under Rucio/containers, and yet, another dev container here, might bring questions.
> 
> So for the shake of keeping things minimal and concentrated, and avoiding to maintain two parallel bootstrapping paths that almost do the same thing (not to mention that Rucio's web-documentation would end up being quite awkward for the developers and unclear when to use what), I would suggest to keep a single approach by merging the best of those two worlds.

I understand the overlap concern, but these serve fundamentally different workflows:

- `bootstrap_dev.sh`: Host-based container orchestration for testing/CI scenarios
- Devcontainer: Integrated IDE development environment with debugging, IntelliSense, and consistent workspace

Thanks for the feedback @Geogouz. The key differentiator is that devcontainers provide guaranteed debugging capabilities within the container environment, which `bootstrap_dev.sh` doesn't address. They're complementary rather than competing approaches.

To address the confusion concern, I could clearly document the distinct use cases and potentially reuse components from `bootstrap_dev.sh` within the devcontainer setup where appropriate.

Geogouz:
I would argue that both have basically the same host dependencies (namely, Docker daemon and Compose plugin). The current script spawns all the required containers on host, while the later spawns a single container on the host and then all others inside it. The practical difference I see (which is indeed a **very** nice feature to have out-of-the-box) is that with the proposed approach, VS Code (as I'm not sure if PyCharm fully supports `devcontainer.json` yet) attaches directly to the containerized workspace enabling breakpoints. By contrast, to use breakpoints now with the `bootstrap_dev.sh`, you would need to use something like `debugpy`.

So my dilemma boils down to.. assuming that we do expand `bootstrap_dev.sh` with a minimal commit, to automatically attach a debugger, does this PR still introduce new functionalities? And what about their cost/gain?
Being a PyCharm guy myself, I think ultimately @maany's opinion is more valuable than mine here (he is the VS Coder guru :D).

mgajek-cern:
> I would argue that both have basically the same host dependencies (namely, Docker daemon and Compose plugin). The current script spawns all the required containers on host, while the later spawns a single container on the host and then all others inside it. The practical difference I see (which is indeed a **very** nice feature to have out-of-the-box) is that with the proposed approach, VS Code (as I'm not sure if PyCharm fully supports `devcontainer.json` yet) attaches directly to the containerized workspace enabling breakpoints. By contrast, to use breakpoints now with the `bootstrap_dev.sh`, you would need to use something like `debugpy`.
> 
> So my dilemma boils down to.. assuming that we do expand `bootstrap_dev.sh` with a minimal commit, to automatically attach a debugger, does this PR still introduce new functionalities? And what about their cost/gain? Being a PyCharm guy myself, I think ultimately @maany's opinion is more valuable than mine here (he is the VS Coder guru :D).

You make a fair point about the architectural similarity and debugging capabilities. You're right that if we enhanced `bootstrap_dev.sh` with debugger attachment, the functional gap would narrow significantly.

The main remaining value would be the standardized devcontainer specification, which provides:
- Cross-platform IDE integration (VS Code native, PyCharm improving)
- Reproducible development environments without local setup variations
- Extension/tooling auto-installation

However, given the overlap you've identified and the ongoing CI work, I think the pragmatic approach is to keep this as a draft until `bootstrap_dev.sh` is finalized. Then we can make an informed decision about whether the devcontainer benefits justify maintaining both approaches or if we should integrate the best parts into a single solution.

I agree @maany's input as the VS Code expert will be crucial for this decision.

.devcontainer/alma9/common_functions.sh
rdimaio: I think people with more time in the project can chime in with a better reply than me here, but I think that at least in this context where it's a dev environment, podman might be preferable (even if just considering podman's open source policies compared to Docker's)

.devcontainer/alma9/devcontainer.json
rdimaio: But in general I think this would need to be aligned with the Rucio version that's locally pulled, since I assume the dev would want to test their changes here

rdimaio:
>  (as I'm not sure if PyCharm fully supports `devcontainer.json` yet)


Seems like it does, even if just in PyCharm Pro https://www.jetbrains.com/help/pycharm/connect-to-devcontainer.html

Geogouz:
> The main remaining value would be the standardized devcontainer specification, which provides:
> 
> * Cross-platform IDE integration (VS Code native, PyCharm improving)
> * Reproducible development environments without local setup variations
> * Extension/tooling auto-installation

Fair points.

> Seems like it does, even if just in PyCharm Pro https://www.jetbrains.com/help/pycharm/connect-to-devcontainer.html

Gonna give it a try to see how things go with Pycharm too.

mgajek-cern:
After discussing with @maany, we've decided to:

- Wait for GitHub issue #7965 to be resolved before proceeding
- Integrate those changes once they're merged
- Expand the acceptance criteria and confirm that debugging works properly for tests, daemons, CLIs and server component (include video evidence)

mgajek-cern:
@Geogouz Depends on https://github.com/rucio/rucio/issues/7965. Work on this will continue once a PR to that issue is merged or the issue is closed with an explanation. I’ll rebase afterwards and proceed or someone else can take over if needed.

mgajek-cern:
Lets close this pull request based on yesterdays agreement related to DRAFT PRs which might depend on previous issues or PRs for now. We might re-open this in the future or address the issue in a separate PR

PR #7966, opened 2025-08-29
Testing: Update file generation methods to use `tempfile` lib for file creation across tests

### LINKED ISSUE #6655: Use `tempfile` instead of custom methods when generating random files (when possible)
There seem to be some methods that have the same functionality as methods from the `tempfile` module in the standard library, e.g.:

https://github.com/rucio/rucio/blob/b34fe26c68698cf9de166f36372cebf099f2ff57/lib/rucio/tests/common.py#L146-L153

Tempfile library for reference: https://docs.python.org/3/library/tempfile.html

I think it's best to substitute these custom functions with the standard `tempfile` methods, wherever possible - some reasons:
- fewer methods to maintain on our side
- `tempfile` methods might be faster/more optimized
- there's no point in reinventing the wheel (unless we really need something custom that `tempfile` cannot do)

0xquark:
Fixes #6655

### DIFF
--- a/lib/rucio/common/test_rucio_server.py
+++ b/lib/rucio/common/test_rucio_server.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import os
+import tempfile
 import unittest
 from os import remove
 from os.path import basename
@@ -20,12 +22,13 @@ from rucio.common.utils import execute
 from rucio.common.utils import generate_uuid as uuid
 
 
-def file_generator(size=2048, namelen=10):
+def file_generator(size=2048):
     """ Create a bogus file and returns it's name.
     :param size: size in bytes
     :returns: The name of the generated file.
     """
-    fn = '/tmp/rucio_testfile_' + uuid()
+    fd, fn = tempfile.mkstemp(prefix='rucio_testfile_', suffix='')
+    os.close(fd)  # Close the file descriptor since we'll use dd to write to it
     execute('dd if=/dev/urandom of={0} count={1} bs=1'.format(fn, size))
     return fn
 
--- a/lib/rucio/tests/common.py
+++ b/lib/rucio/tests/common.py
@@ -204,12 +204,13 @@ def rfc2253_dn_generator() -> str:
     return random_dn
 
 
-def file_generator(size: int = 2, namelen: int = 10) -> str:
+def file_generator(size: int = 2) -> str:
     """ Create a bogus file and returns it's name.
     :param size: size in bytes
     :returns: The name of the generated file.
     """
-    fn = '/tmp/file_' + ''.join(choice(ascii_uppercase) for x in range(namelen))  # noqa: S311
+    fd, fn = tempfile.mkstemp(prefix='file_', suffix='')
+    os.close(fd)  # Close the file descriptor since we'll use dd to write to it
     execute('dd if=/dev/urandom of={0} count={1} bs=1'.format(fn, size))
     return fn
 
--- a/tests/test_rse_protocol_posix.py
+++ b/tests/test_rse_protocol_posix.py
@@ -14,6 +14,7 @@
 
 import os
 import shutil
+import tempfile
 
 import pytest
 
@@ -28,19 +29,20 @@ def setup_posix_test_env(rse_name, rse_settings, user):
 
     data = load_test_conf_file('rse_repository.json')
     prefix = data[rse_name]['protocols']['supported']['file']['prefix']
-    try:
-        os.mkdir(prefix)
-    except Exception as e:
-        print(e)
-    os.system('dd if=/dev/urandom of=%s/data.raw bs=1024 count=1024' % prefix)
+    os.makedirs(prefix, exist_ok=True)
+    # Use tempfile for file creation in the same directory
+    fd, temp_data_file = tempfile.mkstemp(dir=prefix, suffix='.raw', prefix='data_')
+    os.close(fd)
+    os.system('dd if=/dev/urandom of=%s bs=1024 count=1024' % temp_data_file)
+    data_file = os.path.join(prefix, 'data.raw')
+    shutil.move(temp_data_file, data_file)
     for f in MgrTestCases.files_remote:
         protocol = mgr.create_protocol(rse_settings, 'write')
         pfn = next(iter(mgr.lfns2pfns(rse_settings, {'name': f, 'scope': 'user.%s' % user}).values()))
         path = protocol.pfn2path(pfn)
         dirs = os.path.dirname(path)
-        if not os.path.exists(dirs):
-            os.makedirs(dirs)
-        shutil.copy('%s/data.raw' % prefix, path)
+        os.makedirs(dirs, exist_ok=True)
+        shutil.copy(data_file, path)
 
 
 @skip_rse_tests_with_accounts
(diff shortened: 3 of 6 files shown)


lib/rucio/tests/common.py
```diff
@@ -204,12 +204,13 @@ def rfc2253_dn_generator() -> str:
     return random_dn
 
 
-def file_generator(size: int = 2, namelen: int = 10) -> str:
+def file_generator(size: int = 2) -> str:
```
rdimaio, line 207: This seems to do the same as the `file_generator` function in `test_rucio_server`, can we remove one of the two to avoid this duplication?

lib/rucio/common/test_rucio_server.py
```diff
@@ -20,12 +22,13 @@
...
 
 
-def file_generator(size=2048, namelen=10):
+def file_generator(size=2048):
     """ Create a bogus file and returns it's name.
     :param size: size in bytes
     :returns: The name of the generated file.
     """
-    fn = '/tmp/rucio_testfile_' + uuid()
+    fd, fn = tempfile.mkstemp(prefix='rucio_testfile_', suffix='')
```
rdimaio, line 30: Looks like we're not saving this in the `tmp` directory anymore, do we want to make this change? Otherwise it seems we can just set dir here https://docs.python.org/3/library/tempfile.html. I guess it's not needed though:

> If dir is not None, the file will be created in that directory; otherwise, a default directory is used. The default directory is chosen from a platform-dependent list, but the user of the application can control the directory location by setting the TMPDIR, TEMP or TMP environment variables.

tests/test_upload.py
```diff
@@ -714,8 +713,11 @@ def recursive_create(dir_path, contents):
             if isinstance(value, list):
                 for item in value:
                     if isinstance(item, str):
-                        file_path = os.path.join(dir_path, directory,  ''.join(choice(ascii_uppercase) for _ in range(5)))
-                        os.makedirs(os.path.dirname(file_path), exist_ok=True)
+                        target_dir = os.path.join(dir_path, directory)
+                        os.makedirs(target_dir, exist_ok=True)
+                        fd, temp_file = tempfile.mkstemp(dir=target_dir)
+                        os.close(fd)
```
rdimaio, line 719: Although this is in testing and likely not an issue, it's worth mentioning that closing the file descriptor here can result in security issues: https://stackoverflow.com/questions/27680807/mkstemp-is-it-safe-to-close-descriptor-and-reopen-it-again - best to keep it open.

lib/rucio/tests/common.py
```diff
@@ -204,12 +204,13 @@ def rfc2253_dn_generator() -> str:
...
     """ Create a bogus file and returns it's name.
     :param size: size in bytes
     :returns: The name of the generated file.
     """
-    fn = '/tmp/file_' + ''.join(choice(ascii_uppercase) for x in range(namelen))  # noqa: S311
+    fd, fn = tempfile.mkstemp(prefix='file_', suffix='')
+    os.close(fd)  # Close the file descriptor since we'll use dd to write to it
```
rdimaio, line 213: Although this is in testing and likely not an issue, it's worth mentioning that closing the file descriptor here can result in security issues: https://stackoverflow.com/questions/27680807/mkstemp-is-it-safe-to-close-descriptor-and-reopen-it-again - best to keep it open.

lib/rucio/common/test_rucio_server.py
```diff
@@ -20,12 +22,13 @@
...
+    os.close(fd)  # Close the file descriptor since we'll use dd to write to it
```
rdimaio, line 31: Although this is in testing and likely not an issue, it's worth mentioning that closing the file descriptor here can result in security issues: https://stackoverflow.com/questions/27680807/mkstemp-is-it-safe-to-close-descriptor-and-reopen-it-again - best to keep it open.

tests/test_upload.py
```diff
@@ -714,8 +713,11 @@ def recursive_create(dir_path, contents):
...
+                        file_path = temp_file
                         execute(f'dd if=/dev/urandom of={file_path} count={2} bs=1')
```
rdimaio, line 721: seems unnecessary to add another name for this variable - we can just use `temp_file` instead

tests/test_rse_protocol_posix.py
```diff
@@ -28,19 +29,20 @@ def setup_posix_test_env(rse_name, rse_settings, user):
...
     prefix = data[rse_name]['protocols']['supported']['file']['prefix']
-    try:
-        os.mkdir(prefix)
-    except Exception as e:
-        print(e)
-    os.system('dd if=/dev/urandom of=%s/data.raw bs=1024 count=1024' % prefix)
+    os.makedirs(prefix, exist_ok=True)
+    # Use tempfile for file creation in the same directory
+    fd, temp_data_file = tempfile.mkstemp(dir=prefix, suffix='.raw', prefix='data_')
+    os.close(fd)
```
rdimaio, line 35: Although this is in testing and likely not an issue, it's worth mentioning that closing the file descriptor here can result in security issues: https://stackoverflow.com/questions/27680807/mkstemp-is-it-safe-to-close-descriptor-and-reopen-it-again - best to keep it open.

tests/test_rse_protocol_posix.py
```diff
@@ -28,19 +29,20 @@ def setup_posix_test_env(rse_name, rse_settings, user):
...
+    os.system('dd if=/dev/urandom of=%s bs=1024 count=1024' % temp_data_file)
+    data_file = os.path.join(prefix, 'data.raw')
+    shutil.move(temp_data_file, data_file)
```
rdimaio, line 38: We could just use a named temporary file here so we avoid having to do this move

tests/test_rse_protocol_rsync.py
```diff
@@ -67,12 +65,17 @@ def setup_rse_and_files(cls, containerized_rses, vo, tmp_path_factory):
         protocol = rsemanager.create_protocol(rse_settings, 'write')
         protocol.connect()
 
-        os.system('dd if=/dev/urandom of=%s/data.raw bs=1024 count=1024' % prefix)
+        # Use tempfile for file creation in the same directory
+        fd, temp_data_file = tempfile.mkstemp(dir=prefix, suffix='.raw', prefix='data_')
+        os.close(fd)
```
rdimaio, line 70: Although this is in testing and likely not an issue, it's worth mentioning that closing the file descriptor here can result in security issues: https://stackoverflow.com/questions/27680807/mkstemp-is-it-safe-to-close-descriptor-and-reopen-it-again - best to keep it open.

tests/test_rse_protocol_rsync.py
```diff
@@ -67,12 +65,17 @@ def setup_rse_and_files(cls, containerized_rses, vo, tmp_path_factory):
...
+        os.system('dd if=/dev/urandom of=%s bs=1024 count=1024' % temp_data_file)
+        data_file = os.path.join(prefix, 'data.raw')
+        os.rename(temp_data_file, data_file)
```
rdimaio, line 73: Can use namedtemporaryfile instead to avoid renaming here

lib/rucio/tests/common.py
rdimaio, line 207: The idea of the original issue was: do we even need this `file_generator` function, rather than just using `mkstemp` as needed?

Geogouz:
Could you rebase please?

0xquark:
I likely misunderstood this issue and used the wrong approach. I’m closing it now and will open a new PR with the changes.

PR #7972, opened 2025-09-02
CI: Enforce Conventional Commits using `commitlint` 

### LINKED ISSUE #7945: Enforce Conventional Commits in CI
This issue is part of [CI Improvements](https://github.com/rucio/rucio/issues/7667) 

To improve commit consistency, we should enforce the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) standard across the repository.

0xquark:
Fixes #7945 

I am using [commitlint](https://commitlint.js.org/) as a linter for our commit messages to enforce conventional commits. It is widely used in the community and [is properly maintained](https://github.com/conventional-changelog/commitlint). It also offers a pre-commit hook to enable it locally. 

We are using this format: `<type>(<scope>): <description>` 

eg. `feat(Core): Add new rule evaluation engine`

I've added some types based on the types used in [standardized conventional commits](https://www.conventionalcommits.org/en/v1.0.0/). 
The scope is based on the available labels in rucio and is taken from what the [rucio's contributing guide](https://rucio.github.io/documentation/contributing/) mentions. 

These are the rules that i am enforcing through this:
1. **Type**: Must be one of the allowed types listed above
2. **Scope**: Must be one of the predefined Rucio components (PascalCase)
3. **Description**: 
   - Should not end with a period
   - Should be concise but descriptive
4. **Issue Number**: Must reference a GitHub issue using [git-trailer](https://alchemists.io/articles/git_trailers) with `#<number>`
     - In order to add a git trailer, you need to commit using `git commit -m "chore(client): fix xyz" --trailer "Closes: #Issue Number"`
     - Only `closes` and `issue` trailer are currently supported. 
     - Note: You can also add a git-trailer by just adding the line at the end of the commit body
     
5. **Line Length**: Header must not exceed 100 characters (This prevents commit titles from being truncated in the GitHub UI. Additional details can always be included in the commit body. )

Attaching screenshot of how it works: 
image

Note: The only drawback I can foresee is that commitlint is written in JavaScript, while Rucio is a Python-heavy project. To use this tool, we would need to maintain a small JavaScript configuration and install it via npm, which requires maintaining a package.json file.

### DIFF
--- a/.github/workflows/commitlint.yml
+++ b/.github/workflows/commitlint.yml
@@ -9,6 +9,7 @@ concurrency:
 jobs:
   commitlint:
     runs-on: ubuntu-latest
+    if: github.actor != 'dependabot[bot]'
     steps:
       - name: Checkout code
         uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
--- a/.github/workflows/commitlint.yml
+++ b/.github/workflows/commitlint.yml
@@ -1,0 +1,33 @@
+name: Commitlint
+
+on: [push, pull_request]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  commitlint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        with:
+          fetch-depth: 0
+
+      - name: Setup Node.js
+        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
+        with:
+          node-version: '24'
+
+      - name: Install dependencies
+        run: npm install
+
+      - name: Validate current commit (push)
+        if: github.event_name == 'push'
+        run: npx commitlint --from HEAD~1 --to HEAD --verbose
+
+      - name: Validate PR commits
+        if: github.event_name == 'pull_request'
+        run: npx commitlint --from ${{ github.event.pull_request.base.sha }} --to ${{ github.event.pull_request.head.sha }} --verbose
+
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -1,0 +1,122 @@
+// NOTE: Any changes to this list should also be applied to the [GitHub labels](https://github.com/rucio/rucio/issues/labels) and the [documentation](https://rucio.cern.ch/documentation/contributing/)
+const rucioComponents = [
+  'Auth', // Authentication & Authorisation
+  'API', // REST & API
+  'CI', // Continuous Integration
+  'Clients',
+  'Consistency', // Consistency checks
+  'Core', // Core & Internals
+  'Containerization', // Docker, Kubernetes, and container-related functionality
+  'Database',
+  'DatasetDeletion', // Dataset deletion
+  'Deletion',
+  'DIRAC',
+  'Documentation',
+  'LifetimeModel', // Life time model
+  'Messaging',
+  'Metadata',
+  'Monitoring', // Monitoring, observability, and traces
+  'MultiVO', // Multi VO
+  'OpenData', // Open data
+  'Policies',
+  'Probes', // Probes & Alarms
+  'Protocols',
+  'Rebalancing',
+  'Recovery',
+  'Replicas',
+  'Rules',
+  'Subscriptions',
+  'Testing',
+  'Transfers',
+  'WebUI'
+];
+
+const rucioTypes = [
+  // Functional changes
+  'feat',     // New feature
+  'fix',      // Bug fix
+
+  // Non-functional changes
+  'docs',     // Documentation only changes
+  'style',    // Changes that do not affect the meaning of the code
+  'refactor', // Code change that neither fixes a bug nor adds a feature
+  'test',     // Adding missing tests or correcting existing tests
+  'ci',       // Changes to CI configuration files and scripts
+
+  // Miscellaneous
+  'revert',   // Reverts a previous commit
+];
+
+const gitTrailerPlugin = {
+  rules: {
+    'issue-trailer-required': (parsed, _when, _value) => {
+      const { body, footer } = parsed;
+      const fullMessage = [body, footer].filter(Boolean).join('\n');
+
+      if (!fullMessage) {
+        return [false, 'Commit message must include an issue-related Git trailer'];
+      }
+
+      // Issue-related trailer tokens (case-insensitive)
+      const issueTrailerTokens = [
+        'issue', 'closes'
+      ];
+
+      // Create pattern for issue-related trailers: "Token: Value" or "Token #Value"
+      const tokenPattern = issueTrailerTokens.join('|');
+      const issueTrailerPattern = new RegExp(`^(${tokenPattern})\\s*[:#]\\s*.+$`, 'im');
+
+      if (!issueTrailerPattern.test(fullMessage)) {
+        return [false, 'Commit message must include an issue-related Git trailer (e.g., "Issue: #123", "Closes: #456"). You can add a trailer using: git commit -m "message" --trailer "Issue: #123"'];
+      }
+
+      return [true];
+    },
+    'breaking-change-consistency': (parsed, _when, _value) => {
+      const { header = '', body, footer } = parsed;
+      const fullMessage = [body, footer].filter(Boolean).join('\n');
+
+      const headerHasBang = /^[^:]+!\s*:/.test(header);
+      const breakingChangeMatch = fullMessage.match(/^BREAKING CHANGE\s*:\s*(.*)$/im);
+      const hasBreakingFooter = Boolean(breakingChangeMatch);
+
+      if (headerHasBang && !hasBreakingFooter) {
+        return [false, 'Commit messages that use "!" must include a BREAKING CHANGE footer.'];
+      }
+
+      if (!headerHasBang && hasBreakingFooter) {
+        return [false, 'BREAKING CHANGE footer requires "!" in the commit header (e.g., "feat(API)!: ...").'];
+      }
+
+      if (hasBreakingFooter) {
+        const description = breakingChangeMatch[1].trim();
+        if (!description) {
+          return [false, 'BREAKING CHANGE footer must include a description (e.g., "BREAKING CHANGE: describe the impact").'];
+        }
+      }
+
+      return [true];
+    }
+  }
+};
+
+module.exports = {
+  extends: ['@commitlint/config-conventional'],
+  plugins: [gitTrailerPlugin],
+  rules: {
+    'type-enum': [2, 'always', rucioTypes], // Require a valid type
+    'scope-enum': [2, 'always', rucioComponents], // Require a valid scope
+    'scope-case': [0], // Disable case checking since we use custom scopes
+    'scope-empty': [2, 'never'], // Scope is required
+    'subject-case': [0], // Disable subject case checking
+    'subject-empty': [2, 'never'], // Subject is required
+    'subject-full-stop': [2, 'never', '.'], // Subject should not end with a period
+    'header-max-length': [1, 'always', 100], // Warn about exceeding 100 characters
+    'body-max-line-length': [0], // Disable body line length limit
+    'body-leading-blank': [1, 'always'], // Body should start with a blank line
+    'footer-leading-blank': [1, 'always'], // Footer should start with a blank line
+    'issue-trailer-required': [2, 'always'], // Issue-related Git trailer is required
+    'breaking-change-consistency': [2, 'always'] // Ensure BREAKING CHANGE footer usage is consistent
+  },
+  helpUrl: 'https://rucio.cern.ch/documentation/contributing/'
+};
--- a/.gitignore
+++ b/.gitignore
@@ -71,3 +71,7 @@ tools/test.file.1G
 
 # VSCode
 .vscode/
+
+# Node.js dependencies for commitlint
+node_modules/
+package-lock.json
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,3 +20,9 @@ repos:
   hooks:
     - id: ruff
       args: [ --fix ]
+- repo: https://github.com/alessandrojcm/commitlint-pre-commit-hook
+  rev: v9.22.0
+  hooks:
+    - id: commitlint
+      stages: [commit-msg]
+      additional_dependencies: ['@commitlint/config-conventional@^19.8.1']
--- a/package.json
+++ b/package.json
@@ -1,0 +1,12 @@
+{
+  "name": "rucio-commitlint",
+  "version": "1.0.0",
+  "description": "Rucio project commitlint configuration",
+  "devDependencies": {
+    "@commitlint/cli": "^19.8.1",
+    "@commitlint/config-conventional": "^19.8.1"
+  },
+  "scripts": {
+    "commitlint": "commitlint --from HEAD~1 --to HEAD --verbose"
+  }
+}


commitlint.config.js
```diff
@@ -0,0 +1,77 @@
...
+      // Check if header ends with #<number>
+      const issuePattern = /#\d+$/;
+      if (!issuePattern.test(header)) {
+        return [false, 'Commit message must end with issue number format: #<number>'];
+      }
+      
+      return [true];
+    }
+  }
+};
```
dchristidis: This is what we had discussed at the Rucio retreat, but since then I’ve been warming up to the idea of a Git trailer instead. It feels like a more appropriate method to include such information in the commit message.

mgajek-cern:
Good work. Given we're adding a new toolchain, would it be worth doing a quick comparison? Maybe something like commitlint vs a simple Python validator vs GitHub Actions-only? Just a basic pros/cons on complexity, features, and maintenance - nothing fancy. It would help us feel confident we're making the right call and give us something to reference later.

0xquark:
> Good work. Given we're adding a new toolchain, would it be worth doing a quick comparison? Maybe something like commitlint vs a simple Python validator vs GitHub Actions-only? Just a basic pros/cons on complexity, features, and maintenance - nothing fancy. It would help us feel confident we're making the right call and give us something to reference later.

Thanks @mgajek-cern! I agree a quick comparison would help, but I don’t think it’s worth the extra effort and time. My first thought is to use something that’s already well maintained. Since `commitlint` is widely used and actively maintained, I’d prefer that option. Creating our own GitHub Action or Python validator would just add another script for us to maintain, and we’re already trying to cut down on extra scripts with the test-infra refactoring. Among the commit linters I looked at, commitlint seems more popular in the community and offers better features and long-term support.
@rdimaio @bari12 thoughts?

rdimaio:
> > Good work. Given we're adding a new toolchain, would it be worth doing a quick comparison? Maybe something like commitlint vs a simple Python validator vs GitHub Actions-only? Just a basic pros/cons on complexity, features, and maintenance - nothing fancy. It would help us feel confident we're making the right call and give us something to reference later.
> 
> Thanks @mgajek-cern! I agree a quick comparison would help, but I don’t think it’s worth the extra effort and time. My first thought is to use something that’s already well maintained. Since `commitlint` is widely used and actively maintained, I’d prefer that option. Creating our own GitHub Action or Python validator would just add another script for us to maintain, and we’re already trying to cut down on extra scripts with the test-infra refactoring. Among the commit linters I looked at, commitlint seems more popular in the community and offers better features and long-term support. @rdimaio @bari12 thoughts?

commitlint is listed in the suggested tooling for conventional commits https://www.conventionalcommits.org/en/about/#tooling-for-conventional-commits and it apparently seems to be the most popular one as far as GH stars, and it seems to fit our use case without us needing to maintain an extra tool, so I'm inclined to vote for it too.

.github/workflows/commitlint.yml
```diff
@@ -0,0 +1,33 @@
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        with:
+          fetch-depth: 0
+
+      - name: Setup Node.js
+        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
+        with:
+          node-version: '20'
```
maany: Node 20 is in maintenance mode, could we use a more recent version

commitlint.config.js
```diff
@@ -0,0 +1,92 @@
+const rucioComponents = [
```
mgajek-cern, line 2: The 24 components will need maintenance as the project evolves. Consider if this list accurately reflects current Rucio architecture.

commitlint.config.js
```diff
@@ -0,0 +1,92 @@
...
+  'perf',     // Code change that improves performance
+  'test',     // Adding missing tests or correcting existing tests
+  'build',    // Changes that affect the build system or external dependencies
+  'ci',       // Changes to CI configuration files and scripts
+  'chore',    // Other changes that don't modify src or test files
+  'revert',   // Reverts a previous commit
+  'patch'     // Small fixes or improvements
+];
+
+const gitTrailerPlugin = {
```
mgajek-cern, line 50: Original issue wanted `#<issue_number>` in subject line, but this uses Git trailers instead. Need to align on which approach.

The original issue requested commits like: 

```
fix(Core): Add something #123
```

This implementation expects:

```
fix(Core): Add something

Fixes: #123
```

.github/workflows/commitlint.yml
```diff
@@ -0,0 +1,33 @@
...
+      - name: Install dependencies
+        run: npm install
+
+      - name: Validate current commit (push)
+        if: github.event_name == 'push'
+        run: npx commitlint --from HEAD~1 --to HEAD --verbose
+
+      - name: Validate PR commits
+        if: github.event_name == 'pull_request'
+        run: npx commitlint --from ${{ github.event.pull_request.base.sha }} --to ${{ github.event.pull_request.head.sha }} --verbose
```
mgajek-cern, line 33: Should we validate every commit in a PR, or focus on ensuring only the final/merge commit follows the convention? This would allow for informal development commits while maintaining clean project history.
```suggestion
    - name: Validate PR commits
      if: github.event_name == 'pull_request'
      run: npx commitlint --from HEAD~1 --to HEAD --verbose
```

mgajek-cern (commented):
My few cents.
Solid technical implementation. A few considerations came up during review

commitlint.config.js
0xquark: I've taken the list from what's mentioned in the [Rucio Documentation](https://rucio.github.io/documentation/contributing/), which links to the components from [labels](https://github.com/rucio/rucio/labels), and the list is updated with the latest components.

commitlint.config.js
0xquark: As discussed [here](https://github.com/rucio/rucio/pull/7972#discussion_r2322102298) and during Rucio's Community Meeting, we’re going with the git-trailers approach. I'll update the issue as well.

.github/workflows/commitlint.yml
0xquark: I think for the PRs, it makes sense to validate all the commits to ensure consistent history and not only the final commit when merging.

commitlint.config.js
```diff
@@ -0,0 +1,92 @@
+const rucioComponents = [
+  'Core',
+  'Clients', 
+  'Database',
+  'Authentication',
+  'Authorisation',
```
dchristidis: This should be one component.

commitlint.config.js
```diff
@@ -0,0 +1,92 @@
...
+  'docs',     // Documentation only changes
+  'style',    // Changes that do not affect the meaning of the code
+  'refactor', // Code change that neither fixes a bug nor adds a feature
+  'perf',     // Code change that improves performance
+  'test',     // Adding missing tests or correcting existing tests
+  'build',    // Changes that affect the build system or external dependencies
+  'ci',       // Changes to CI configuration files and scripts
+  'chore',    // Other changes that don't modify src or test files
+  'revert',   // Reverts a previous commit
+  'patch'     // Small fixes or improvements
```
dchristidis: How did you imagine the purpose of `patch`?

commitlint.config.js
```diff
@@ -0,0 +1,92 @@
...
+      }
+      
+      // Issue-related trailer tokens (case-insensitive)
+      const issueTrailerTokens = [
+        'issue-id', 'issue', 'issues',
+        'closes', 'close', 'closed',
+        'fixes', 'fix', 'fixed',
+        'resolves', 'resolve', 'resolved',
+        'refs', 'ref', 'references', 'reference'
+      ];
```
dchristidis, line 63: Is this flexibility desirable? I would prefer to stick to one required trailer (most likely `Issue`), so that I can easily use it in `git log` with a custom format.

commitlint.config.js
0xquark: Aren’t these two things different in Rucio? Authentication is about validating who the user is, while authorization is about what the user is allowed to do. I do agree that having a single, similar component could help avoid confusion.

commitlint.config.js
0xquark: A `patch` is a small change that isn’t a bug fix (`fix`) or a new feature (`feat`). It’s usually for minor tweaks or adjustments that need tracking but don’t clearly fit other categories. I added it as a fallback for minor changes we’re unsure how to classify, though i feel like it overlaps with other categories like style or refactor.

commitlint.config.js
0xquark: To me, this seems more about identifying the most useful use case. I’m not entirely sure which direction to take (it’s quite subjective). I personally prioritize developer experience, so I’d like to include these extra options (like `fixes` for closing an issue with my commit, or `refs` for referencing a specific issue). At the same time, you’re also right that it would be simpler to grep for just `Issue`.

commitlint.config.js
dchristidis: Not according to the [list of components](https://rucio.cern.ch/documentation/component_leads).

commitlint.config.js
dchristidis: Could you maybe group with some comments? Maybe these three under ‘functional changes’, most other under ‘non-functional changes’ and the rest under ‘miscellaneous’?

commitlint.config.js
dchristidis: I think [PEP 20](https://peps.python.org/pep-0020/#the-zen-of-python) applies here.

> There should be one-- and preferably only one --obvious way to do it.

But I will back off if others have very strong opinions.

0xquark: We talked about this in yesterday’s meeting and agreed on a practical and simple solution: use two trailer tokens `Issue` and `Closes.`

`Issue` will be used to reference an issue.

`Closes` will be used to close an issue through a commit.

Using `Closes` gives us the nice benefit of automatically closing issues when a commit is merged.
And since there are only two trailer tokens, it’ll be easier to grep through them as well.

commitlint.config.js
dchristidis: Maybe it’s better to use the shorthand `Auth`?

commitlint.config.js
dchristidis: I’m now confused again. `patch` is described as `Small fixes or improvements`. Isn’t that a wildcard for everything? Isn’t there a risk of abuse?

commitlint.config.js
dchristidis: I remain unconvinced, but won’t push any further.

commitlint.config.js
0xquark: I see your point, but I think the idea behind patch was to cover those small tweaks that don’t really count as a bug fix or a feature or a styling improvement. Stuff like adjusting a timeout, changing a small detail in behavior, or improving something minor without adding new functionality/fix.

It’s not meant to be a catch-all, just a way to track small improvements that don’t fit neatly under fix or feat or part of the grey area. As long as we keep the definition clear in our guidelines, I think it should be fine and won’t be abused.
Maybe I'll add a comment or FAQ about not using it as a fallback for everything or explicitly mention it when it is meant to be used.  With Patch, it groups things whose impact is low-risk and incremental (no major functional or structural change).

commitlint.config.js
dchristidis, line 2: I would recommend:

 1. Adding a comment before this assignment stating that any changes to this list should also be applied to the GitHub labels and the documentation. A similar comment should probably be added to the latter.
 2. Sorting them alphabetically.
 3. Wherever a shorthand has been applied, adding a comment with the original, e.g. `'Auth', // Authentication & Authorisation`

commitlint.config.js
dchristidis: But even if you’re changing the default timeout, wouldn’t you open a GitHub issue with the `Feature` template?

0xquark: I’d say it depends. If changing the timeout introduces new behavior, it’s a feat; if it’s just tweaking an existing value, it fits patch. That said, I get your point, any change is still an improvement in behavior, no matter how small or big. Now that I see it, I also agree about the potential for patch to be abused, so I’d support removing it from the list.

0xquark:
Documentation for Review: https://github.com/rucio/documentation/pull/643

commitlint.config.js
dchristidis: What type is one supposed to use when removing functionality? I’d say `feat`, but the description doesn’t fit.

I’m also sceptical of `perf`. For one thing, this should probably not be a functional change: you improve performance by doing *the same thing* better. For another, improving performance is `refactor` with a specific goal.

dchristidis: In addition, `feat` and `fix` don’t align very well with the [Contribution Guide](https://rucio.cern.ch/documentation/contributing/#what-should-i-know-before-i-get-started).

bari12: I understand the worry with `feat` and `fix` - the issue is that these are the only two scopes absolutely required in the conventional commit RFC, all the other scopes are optional. Now, we can say we ignore this, but there is obviously an advantage following the standard. I would propose we keep it like this, for now, and see how it goes. We can still change things later. If I would have to make a guess, the `refactor` scope will be used a lot and is probably the one which we might want to change.

bari12:
To be discussed tomorrow in the dev meeting. But my proposal would be to merge this after the feature freeze next week.

commitlint.config.js
dchristidis: I was coming at it from the opposite direction: `feat` and `fix` should stay, but the distinction between a feature and a patch may need to be adapted. Probably we should pick a name around `BREAKING CHANGE`, as described in the Conventional Commits.

Naming things is hard.

commitlint.config.js
```diff
@@ -0,0 +1,125 @@
...
+    }
+  }
+};
+
+module.exports = {
+  extends: ['@commitlint/config-conventional'],
+  plugins: [gitTrailerPlugin],
+  rules: {
+    'type-enum': [2, 'always', rucioTypes], // Require a valid type
+    'scope-enum': [2, 'always', rucioComponents], // Require a valid scope
```
dchristidis, line 108: What do you do in case you’re making a sweeping change across the entire codebase? For example, we will stop using `typing.Union` once we move to Python 3.12.

0xquark: I get what you mean about a global change not really fitting any specific component. But personally, I think big or breaking changes to the whole codebase are usually easier to deal with if they’re done gradually, with separate commits for each component. It makes the review process a lot more manageable, and if the reviewer wants to squash everything at the end, they still can.

I’m not against adding a global or all scope if we really need it, but I don’t think it’s something we should encourage as the default way to make changes.

dchristidis: That’s probably fine.

commitlint.config.js
dchristidis: No further comment from me on the proposed types and scopes. We can revise as we gain more experience using them.

Geogouz:
Could you rebase please?

commitlint.config.js
```diff
@@ -0,0 +1,122 @@
...
+  'Clients',
+  'Consistency', // Consistency checks
+  'Core', // Core & Internals
+  'Containerization', // Docker, Kubernetes, and container-related functionality
+  'Database',
+  'DatasetDeletion', // Dataset deletion
+  'Deletion',
+  'DIRAC',
+  'Documentation',
+  'Lifetime', // Life time model
```
dchristidis: Maybe `LifetimeModel` is preferable here.

commitlint.config.js
```diff
@@ -0,0 +1,122 @@
...
+      const issueTrailerTokens = [
+        'issue', 'closes'
+      ];
+
+      // Create pattern for issue-related trailers: "Token: Value" or "Token #Value"
+      const tokenPattern = issueTrailerTokens.join('|');
+      const issueTrailerPattern = new RegExp(`^(${tokenPattern})\\s*[:#]\\s*.+$`, 'im');
+
+      if (!issueTrailerPattern.test(fullMessage)) {
+        return [false, 'Commit message must include an issue-related Git trailer (e.g., "Issue: #123", "Closes: #456"). You can add a trailer using: git commit -m "message" --trailer "Fixes: #123"'];
```
dchristidis: The example should not have ‘Fixes’ is no longer a valid trailer.

dchristidis (changes requested):

commitlint.config.js
0xquark: Nice Catch! 👏

commitlint.config.js
```diff
@@ -0,0 +1,122 @@
...
+      const headerHasBang = /^[^:]+!\s*:/.test(header);
+      const breakingChangeMatch = fullMessage.match(/^BREAKING CHANGE\s*:\s*(.*)$/im);
+      const hasBreakingFooter = Boolean(breakingChangeMatch);
+
+      if (headerHasBang && !hasBreakingFooter) {
+        return [false, 'Commit messages that use "!" must include a BREAKING CHANGE footer.'];
+      }
+
+      if (!headerHasBang && hasBreakingFooter) {
+        return [false, 'BREAKING CHANGE footer requires "!" in the commit header (e.g., "chore(api)!: ...").'];
```
dchristidis: Similarly, ’chore’ has been removed and ‘api’ should match case.

dchristidis (changes requested):

dchristidis (approved):

PR #7974, opened 2025-09-02
Policies: remove confusing log message when loading algorithm module #7941

### LINKED ISSUE #7941: Confusing log message ‘Policy algorithm module XXX could not be loaded’
### Description

Having deployed Rucio 38.0.0rc3 we’re seeing a lot of log messages with content:

```
Policy algorithm module atlas_rucio_policy_package.lfn2pfn could not be loaded
```

The source code of the ATLAS rucio policy package is available in https://gitlab.cern.ch/atlas-adc-ddm/rucio-policy-package (requires a CERN account). The important bit is that there’s no `lfn2pfn` returned in the `get_algorithms()` function.

Related to #7836.

### Steps to reproduce

Bootstrap:

```console
$ python3 -m venv rucio
$ source rucio/bin/activate
$ pip3 install rucio==38.0.0rc3 atlas_rucio_policy_package
…
$ mkdir rucio/etc
$ cat > rucio/etc/rucio.cfg << EOF
[policy]
package = atlas_rucio_policy_package
EOF
```

Then:

```python
>>> import logging
>>> logging.basicConfig(level=logging.DEBUG)
>>> from rucio.rse.translation import RSEDeterministicTranslation
>>> t = RSEDeterministicTranslation()
>>> t.path('scope', 'name')
INFO:policy:Policy algorithm module atlas_rucio_policy_package.lfn2pfn could not be loaded
'scope/c0/c2/name'
```

### Rucio Version

_No response_

### Additional Information

_No response_

jamesp-epcc:
I think it is best to remove this log message. The case where no module is loaded is likely to be the most common one and normally doesn't mean anything is wrong. The message gets printed a lot and has already caused confusion for operators (see https://github.com/rucio/rucio/issues/7941).

### DIFF
--- a/lib/rucio/common/plugins.py
+++ b/lib/rucio/common/plugins.py
@@ -77,7 +77,7 @@ class PolicyPackageAlgorithms:
     """
     _ALGORITHMS: dict[POLICY_ALGORITHM_TYPES_LITERAL, dict[str, 'Callable[..., Any]']] = {}
     _loaded_policy_modules = False
-    _default_algorithms: dict[str, 'Callable[..., Any]'] = {}
+    _default_algorithms: dict[str, Optional['Callable[..., Any]']] = {}
 
     def __init__(self) -> None:
         if not self._loaded_policy_modules:
@@ -105,17 +105,23 @@ class PolicyPackageAlgorithms:
                 vo = ''
             package = cls._get_policy_package_name(vo)
         except (NoOptionError, NoSectionError):
+            cls._default_algorithms[type_for_vo] = default_algorithm
             return default_algorithm
 
         module_name = package + "." + algorithm_type
+        LOGGER.info('Attempting to find algorithm %s in default location %s...' % (algorithm_type, module_name))
         try:
             module = importlib.import_module(module_name)
 
             if hasattr(module, algorithm_type):
                 default_algorithm = getattr(module, algorithm_type)
-                cls._default_algorithms[type_for_vo] = default_algorithm
+        except ModuleNotFoundError:
+            LOGGER.info('Algorithm %s not found in default location %s' % (algorithm_type, module_name))
         except ImportError:
-            LOGGER.info('Policy algorithm module %s could not be loaded' % module_name)
+            LOGGER.info('Algorithm %s found in default location %s, but could not be loaded' % (algorithm_type, module_name))
+        # if the default algorithm is not present, this will store None and we will
+        # not attempt to load the same algorithm again
+        cls._default_algorithms[type_for_vo] = default_algorithm
         return default_algorithm
 
     @classmethod


dchristidis (commented):
I don’t think this to be the right approach. For each LFN-to-PFN translation (millions per hour for ATLAS), Python tries to import a non-existent module. Isn’t that wasteful?

jamesp-epcc:
It might be better then if the code only tries to load the module the first time (potentially with log message on failure) and then keeps track of which modules it has already tried to load and doesn't try the same ones again.

dchristidis:
Isn’t it possible to use the `get_algorithms()` function in a policy package’s `__init__.py` file as some sort of manifest of what is actually included in it? And not try to import anything that is not advertised in said manifest?

jamesp-epcc:
That's how it was done originally, but the purpose of the "default algorithms" was to allow an algorithm to be implemented simply by placing it in a known location, with no boilerplate required.

dchristidis:
I confess I am perplexed by this direction. Already I find the use of term ‘default algorithm’ to refer to the policy package confusing. Default is what is used in the absence of a selection; default is what ATLAS Rucio is using given the absence of a  `lfn2pfn()` function in the ATLAS policy package. I am also somewhat opposed to the use of the term boilerplate in this context. By not implementing some rudimentary checking, one cannot discern the intention: is the algorithm from the policy package not used because of some issue or because it is not supposed to exist in the first place?

But then again I’m not familiar enough with this effort to make an informed opinion, hence I will withdraw from it.

rdimaio:
> I confess I am perplexed by this direction. Already I find the use of term ‘default algorithm’ to refer to the policy package confusing. Default is what is used in the absence of a selection; default is what ATLAS Rucio is using given the absence of a `lfn2pfn()` function in the ATLAS policy package. I am also somewhat opposed to the use of the term boilerplate in this context. By not implementing some rudimentary checking, one cannot discern the intention: is the algorithm from the policy package not used because of some issue or because it is not supposed to exist in the first place?
> 
> But then again I’m not familiar enough with this effort to make an informed opinion, hence I will withdraw from it.

Default here refers to the default algorithm name that Rucio expects to find in a policy package, to avoid having to register custom algorithm names if not needed, so it's the "default name" (i.e. `pfn2lfn`) rather than "default algorithm" (i.e. the actual implementation of `pfn2lfn`). 

I understand this is a bit confusing, but generally we say "generic / fallback" when talking about the actual default algorithms in the Rucio codebase.

There's some more context here https://github.com/rucio/rucio/issues/7405

rdimaio (commented):
Maybe we should also modify the log message a bit to be clearer to operators - we could say something like:

"Attempting to find {algorithm_name} in default location {module}.{algorithm_name}..." (when first attempting to load it)

"{algorithm_name} algorithm not found in default location {module}.{algorithm_name}." (if not found)

and if it's found but it's causing an import error, we could say:

"{algorithm_name} algorithm found in default location {module}.{algorithm_name}, but could not be loaded"

jamesp-epcc:
That's a good idea. I have updated the code.

rdimaio (approved):

PR #7981, opened 2025-09-03
Policies: use literal for policy algorithm types #7858

### LINKED ISSUE #7858: use literal to store allowed algorithm names for type hints and validation
e.g. here: 

https://github.com/rucio/rucio/blob/02f32e1ef166681511c4e37e6cf5df051b415549/lib/rucio/common/plugins.py#L83-L95

We should only allow specific algorithm names. We should add validation of this, as well as using a literal that contains the allowed algorithm names.

jamesp-epcc:
This PR adds a Literal containing the allowed policy package algorithm types, and uses it for type checking and validation.

### DIFF
--- a/lib/rucio/common/utils.py
+++ b/lib/rucio/common/utils.py
@@ -46,7 +46,7 @@ import requests
 from typing_extensions import ParamSpec
 
 from rucio.common.config import config_get, config_get_bool
-from rucio.common.constants import BASE_SCHEME_MAP, DEFAULT_VO
+from rucio.common.constants import BASE_SCHEME_MAP, DEFAULT_VO, POLICY_ALGORITHM_TYPES_LITERAL
 from rucio.common.exception import DIDFilterSyntaxError, DuplicateCriteriaInDIDFilter, InputValidationError, InvalidType, MetalinkJsonParsingError, MissingModuleException, RucioException
 from rucio.common.extra import import_extras
 from rucio.common.plugins import PolicyPackageAlgorithms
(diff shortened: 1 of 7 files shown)


lib/rucio/common/utils.py
```diff
@@ -560,7 +560,7 @@ class ScopeExtractionAlgorithms(PolicyPackageAlgorithms):
     Handle scope extraction algorithms
     """
 
-    _algorithm_type = 'scope'
+    _algorithm_type: 'POLICY_ALGORITHM_TYPES_LITERAL' = 'scope'
```
mgajek-cern: Quick question about the type annotations - I see we're using quotes around `POLICY_ALGORITHM_TYPES_LITERAL` in some places. Since it's imported from constants, would it be cleaner to remove the quotes for better type checking? Just wondering if there was a specific reason for keeping them quoted.

Also, would it make sense to consider consistent variable naming here? For instance, using `algorithm_type` instead of `_algorithm_type` might follow more conventional naming patterns. 

```suggestion
    algorithm_type: POLICY_ALGORITHM_TYPES_LITERAL = 'scope'
```

jamesp-epcc: Removing the quotes is a good suggestion so I've done that. The leading underscore on the variable name has always been there, I think it's intended to signify a variable used internally by the class, but I don't have strong views on whether this should be kept.

mgajek-cern: Thanks for addressing the findings.

You're right. Though Python doesn't enforce encapsulation, the underscore marks this as a protected member by convention - something subclasses can use but external code shouldn't access directly. Removing it would break this widely-followed convention and create inconsistency with your existing codebase. Let's keep it as is - not actually an issue that needs fixing.

rdimaio (approved):

PR #7983, opened 2025-09-03
Update daemons debug messages

### LINKED ISSUE #7982: Improve daemon judge injector messages

lobis:
Closes https://github.com/rucio/rucio/issues/7982

### DIFF
--- a/lib/rucio/core/request.py
+++ b/lib/rucio/core/request.py
@@ -22,6 +22,7 @@ import threading
 import traceback
 from abc import ABCMeta, abstractmethod
 from collections import defaultdict, namedtuple
+from collections.abc import Sized
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any, Optional, Union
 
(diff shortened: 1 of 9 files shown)


lib/rucio/core/request.py
```diff
@@ -310,7 +311,11 @@ def queue_requests(
     :param logger:    Optional decorated logger that can be passed from the calling daemons or servers.
     :returns:         List of Request-IDs as 32 character hex strings.
     """
-    logger(logging.DEBUG, "queue requests")
+
+    if isinstance(requests, Sized):
+        logger(logging.DEBUG, f"Queuing {len(requests)} requests")
```
dchristidis: The use of an f-string here is not considered a good practice, see: https://docs.astral.sh/ruff/rules/logging-f-string/

lobis: Right, I think this was already mentioned, I'll do my best to keep it present

dchristidis (approved):

lib/rucio/core/request.py
dchristidis: I created #7985 so that this can be checked automatically. I will not ask you to fix the offending lines modified by your commits; they can be addressed as part of that issue.

rdimaio (approved):

PR #7997, opened 2025-09-08
Replicas: Skip if there are no replicas to quarantine - Fix #7943

### LINKED ISSUE #7943: Replica quarantining proceeds to contact database even if nothing remains to be done
https://github.com/rucio/rucio/blob/c6d7f6f94194d831c85f35f4b3c4ceac703043f4/lib/rucio/core/quarantined_replica.py#L77

We were having database exceptions that seemed very odd. We tracked it to this line where even if replicas is left empty after removing the quarantined replicas, the code continues. We "solved" this by just returning from the function if replicas is empty, but maybe there is a better way to handle this?

If there is not, we can make a PR from our simple fix.

ericvaandering:
A very simple patch to avoid database exceptions. The database exception is that RSE ID is null in the case where there are no new replicas to quarantine.


rdimaio (commented):
Please rebase to fix unrelated CI issue

dchristidis:

PR #8008, opened 2025-09-11
Add a `config update` method, restrict `config add` to new fields

### LINKED ISSUE #7977: CLI: Add a `config update` function, modify `config add`
* Create a `config update` function. This function will verify the section, option already exists, and will change the given field
* Modify existing `config add` function. This function will only add a new section & option, if it does not already exist, otherwise it will raise an error (SectionExistsError?)

voetberg:
Closes #7977 

Also changes the error handler routing, the exceptions raised by the new `update` and `add` methods were not being handled due to the lazy loading.

### DIFF
--- a/lib/rucio/cli/command.py
+++ b/lib/rucio/cli/command.py
@@ -69,7 +69,7 @@ class LazyGroup(click.Group):
     def invoke(self, ctx: click.Context):
         result = self._invoke_with_handler(ctx)
         if result not in (None, 0):
-            sys.exit(1 if result != 2 else 2)
+            sys.exit(1 if result != 2 else 1)
         return result
 
 
--- a/lib/rucio/cli/command.py
+++ b/lib/rucio/cli/command.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 import importlib
 import signal
+import sys
 import time
 
 import click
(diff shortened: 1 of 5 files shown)


lib/rucio/cli/command.py
```diff
@@ -69,7 +69,7 @@ def _invoke_with_handler(self, ctx: click.Context):
     def invoke(self, ctx: click.Context):
         result = self._invoke_with_handler(ctx)
         if result not in (None, 0):
-            sys.exit(1 if result != 2 else 1)
+            sys.exit(1 if result != 2 else 2)
```
rdimaio, line 72: I'm a bit confused by the 1 and 2 thing - do we have some spec somewhere that says in what way error codes 1 and 2 differ?

tests/test_cli_client_structure.py
```diff
@@ -184,7 +184,7 @@ def test_account_identities(rucio_client):
 
     cmd = f"rucio account identity add --account {tmp_account} --type NotAType --email {email} --id {email}"
     exitcode, _, _ = execute(cmd)
-    assert exitcode == 2  # Fails with the argparse validation
+    assert exitcode == 1
```
rdimaio, line 187: Is the failure reason different now? It would be good if there was still a comment to point out the failure reason

rdimaio (approved):

lib/rucio/cli/command.py
voetberg: I don't think we do. We do have a lot of tests that rely on getting an exitcode 2 (raised by argparse, a argument missing/incorrectly specified) but because we don't use argparse as the main cli parser it gets a little fuzzy.

tests/test_cli_client_structure.py
voetberg: No, the failure reason is exactly the same but I'm fairly certain I miss-specified it originally. It raised a 2 because it was forced to

PR #8016, opened 2025-09-15
WebUI: Add global account limits to the account RSE usage page #7935

### LINKED ISSUE #7935: Add rse type totals to web UI usage
### Description

Add total scratchdisk/type usage to account [usage](https://rucio-ui.cern.ch/account_rse_usage) as per "rucio list-account-usage lzheng"

### Motivation

Useful to link for user task problems with quota.

### Change

_No response_

dchristidis:

### DIFF
--- a/lib/rucio/web/ui/static/account_rse_usage.js
+++ b/lib/rucio/web/ui/static/account_rse_usage.js
@@ -24,14 +24,14 @@ load_data = function(account_chosen) {
     r.get_local_account_limits({
         account: account_chosen,
         success: function(data) {
-            new_data = {};
+            var new_data = {};
             $.each(data, function(index, limit) {
                 new_data[index] = {'bytes': 0, 'bytes_limit': limit, 'files': 0};
             });
             r.get_local_account_usage({
                 account: account_chosen,
                 success: function(data) {
-                    table_data = [];
+                    var table_data = [];
                     var rse_dict = Object();
                     var start_date = new Date(1547078400000);
                     var today = new Date().getTime();
@@ -144,14 +144,14 @@ load_data = function(account_chosen) {
     r.get_global_account_limits({
         account: account_chosen,
         success: function(data) {
-            new_data = {};
+            var new_data = {};
             $.each(data, function(index, limit) {
                 new_data[index] = {'bytes': 0, 'bytes_limit': limit, 'files': 0};
             });
             r.get_global_account_usage({
                 account: account_chosen,
                 success: function(data) {
-                    table_data = [];
+                    var table_data = [];
                     $.each(data, function(index, value) {
                         new_data[value['rse_expression']] = {'bytes': value['bytes'], 'bytes_limit': value['bytes_limit'], 'files': value['files']}
                     });
--- a/lib/rucio/web/ui/static/account_rse_usage.js
+++ b/lib/rucio/web/ui/static/account_rse_usage.js
@@ -10,10 +10,10 @@
  * - Cedric Serfon, <cedric.serfon@cern.ch>, 2019
  */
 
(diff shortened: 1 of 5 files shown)


lib/rucio/web/ui/static/account_rse_usage.js
```diff
@@ -10,22 +10,25 @@
  * - Cedric Serfon, <cedric.serfon@cern.ch>, 2019
  */
 
-var result_html = '<table id="resulttable" class="compact stripe order-column" style="word-wrap: break-word;"><thead><tr><th>RSE</th><th>Quota</th><th>Used</th><th>Available</th><th>Files</th></tr></thead><tfoot><tr><th>RSE</th><th>Quota</th><th>Used</th><th>Available</th><th>Files</th></tr></tfoot></table>'
+var result_local_html = '<table id="resulttable_local" class="compact stripe order-column" style="word-wrap: break-word;"><thead><tr><th>RSE</th><th>Quota</th><th>Used</th><th>Available</th><th>Files</th></tr></thead><tfoot><tr><th>RSE</th><th>Quota</th><th>Used</th><th>Available</th><th>Files</th></tr></tfoot></table>'
+var result_global_html = '<table id="resulttable_global" class="compact stripe order-column" style="word-wrap: break-word;"><thead><tr><th>RSE Expression</th><th>Quota</th><th>Used</th><th>Available</th><th>Files</th></tr></thead></table>'
```
rdimaio, line 14: Nit and out of scope, but I'd be in favour of renaming these `resulttable_*` to `result_table_*` for better readability and for consistency with the fact that we are now using the underscore to separate the last word anyway

lib/rucio/web/ui/static/rucio.js
```diff
@@ -306,11 +306,31 @@ RucioClient.prototype.list_accounts = function(options) {
...
+                 },
+                 error: function(jqXHR, textStatus, errorThrown)
+                 {
+                     options.error(jqXHR, textStatus, errorThrown);
+                 }
+        });
+};
+
+/* get global limits for an account */
+RucioClient.prototype.get_global_account_limits = function(options) {
```
rdimaio, line 330: Looks like this mostly has the same behaviour as the `get_local_account_limits` function, except for the API endpoint. Would it be worth making a generic `get_account_limits` a function to which we can also pass a flag/string for `local` or `global`, which then sets the API endpoint as needed? 

I think it only depends on whether we ever expect to have a different behaviour for global and local queries here. We could also make the argument that this is the old webUI and thus we don't care too much about avoiding duplication, since it's going to be replaced anyway.

lib/rucio/web/ui/static/rucio.js
```diff
@@ -1041,6 +1061,27 @@ RucioClient.prototype.get_account_usage = function(options) {
         });
 };
 
+/* get global account usage */
+RucioClient.prototype.get_global_account_usage = function(options) {
```
rdimaio, line 1065: Same point here about the duplication

rdimaio (dismissed):

lib/rucio/web/ui/static/rucio.js
dchristidis: I agree with everything you said.

All the comments would be valid and warrant changes, were we working on a piece of code that is expected to be maintained long term. The Web UI is in maintenance mode and, if it were up to me, would not be receiving new features at all. In that sense, refactoring should also be treated as out of scope.

Consequently, I opted to do the minimum amount of work that would allow to mark the issue resolved.

dchristidis:
I have observed a strange behaviour, were the results are mixed or appear in the wrong table. It’s as if jQuery’s `each()` function does not respect the scope of the `new_data` variable.

lib/rucio/web/ui/static/account_rse_usage.js
dchristidis: While I don’t disagree with you, there existing uses of this naming scheme:

```
$ git grep resulttable_ | sed --regexp-extended 's/.*(resulttable_[a-zA-Z_]+).*/\1/' | sort --unique
resulttable_details
resulttable_filter
resulttable_length
resulttable_overview
resulttable_tape_overview
```

rdimaio (approved):

PR #8021, opened 2025-09-15
Deletion: Do not serialise the message payload #8020

### LINKED ISSUE #8020: ERASE message payloads are serialised twice
### Description

This applies to situations where the last file of a dataset is deleted by the Reaper, which under normal circumstances triggers the deletion of the dataset itself.

Originally in commit e17fb3c32e, the payload is serialised then inserted into the database with SQLAlchemy’s `bulk_insert_mappings()`. Then in commit ce39a0bf87, that function is replaced with `add_message()`, which internally serialises the payload again.

### Steps to reproduce

Once one observes the effect, the steps to reproduce should be obvious.

```sql
SELECT payload FROM atlas_rucio.messages WHERE event_type = 'ERASE';
 payload
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 "{\"scope\": \"mc16_13TeV\", \"name\": \"mc16_13TeV.802425.Py8EG_A14NNPDF23LO_XHS_X400_S250_4b_VAR2UP.simul.HITS.e8448_e7400_a875_tid41072328_00\", \"account\": \"root\"}"
 {"account": "root", "scope": "panda", "name": "panda.um.user.lbrown.700900.Sh.DAOD_PHYSLITE.e8514_s4369_r16083_p6697.27Aug25v2.log.619841135.619841135"}
 …
```

### Rucio Version

_No response_

### Additional Information

_No response_

dchristidis:
I had trouble deciding whether to put this under `Messaging` or `Deletion`. Eventually, I opted for the latter.

### DIFF
--- a/lib/rucio/core/replica.py
+++ b/lib/rucio/core/replica.py
@@ -22,7 +22,6 @@ from curses.ascii import isprint
 from datetime import datetime, timedelta
 from hashlib import sha256
 from itertools import groupby
-from json import dumps
 from re import match
 from struct import unpack
 from traceback import format_exc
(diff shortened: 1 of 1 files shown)


rdimaio (approved):
Would it be worth adding a test here to test against regressions?

dchristidis:
I’m not sure. I didn’t find any example of message generation, as part of another operation, being tested. Let alone testing the message payload.

PR #8024, opened 2025-09-16
Clients: Support rich client and table output for opendata CLI

### LINKED ISSUE #8023: Support rich client in opendata CLI

lobis:
Closes #8023.

Add table and rich client support to `opendata did show` and `opendata did list` cli commands.

The did code where I got inspiration from defines the logic in `bin_legacy`. I decided to put the code in `cli`. Let me know if you think I should arrange things differently.

Example with rich client:

Screenshot 2025-09-16 at 16 25 29

Example without rich client:

Screenshot 2025-09-16 at 16 26 06

new `short` option without rich client:

```
❯ rucio opendata did list --short
demo:demo-1
```

new `short` option with rich client:

```
❯ rucio opendata did list --short
┌─────────────┐
│ SCOPE:NAME  │
├─────────────┤
│ demo:demo-1 │
└─────────────┘
```

(is this last one supposed to like like this?)

### DIFF
--- a/lib/rucio/cli/opendata.py
+++ b/lib/rucio/cli/opendata.py
@@ -81,17 +81,14 @@ def list_opendata_dids(ctx: "Context", state: Optional["OPENDATA_DID_STATE_LITER
         else:
             table_data.append([f"{did['scope']}:{did['name']}", did['state']])
 
-    if cli_config == 'rich':
-        if short:
-            table = generate_table([[did] for did, _ in table_data], headers=['SCOPE:NAME'], col_alignments=['left'])
-        else:
-            table = generate_table(table_data, headers=['SCOPE:NAME', '[STATE]'], col_alignments=['left', 'left'])
-        spinner.stop()
-        print_output(table, console=ctx.obj.console, no_pager=ctx.obj.no_pager)
+    if short:
+        for did, _ in table_data:
+            print(did)
     else:
-        if short:
-            for did, _ in table_data:
-                print(did)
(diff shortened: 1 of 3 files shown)


voetberg:
Looks good! Do you have a screenshot of what the output looks like here?

lobis:
> Looks good! Do you have a screenshot of what the output looks like here?

I added them in the description.

Is the `short` option supposed to look like this? Comparing it to `rucio did list` the opendata cli looks the same, but as I understand it the purpose of `short` is to produce a nice pipeable output right?

rdimaio (commented):
LGTM but I agree that short seems a bit off, I assume it would only return the list of DIDs with nothing else, based ont he docstring for the existing short flag:
> Just dump the list of DIDs.

lobis:
> LGTM but I agree that short seems a bit off, I assume it would only return the list of DIDs with nothing else, based ont he docstring for the existing short flag:
> 
> > Just dump the list of DIDs.

Created an issue to track this: https://github.com/rucio/rucio/issues/8026

voetberg:
Yeah I agree with Riccardo here, the `--short` option should overwrite the rich cli.

lobis:
I updated this PR to work on the `short` option as we expected and I will create another PR for `rucio did list` to also work this way. I think this one can be merged as is.

rdimaio (dismissed):

voetberg (dismissed):

rdimaio (approved):

PR #8035, opened 2025-09-18
Feature 7581 remove oauth manager daemon OIDC auto

panta-123:
- remove oauth manager daemon
- port only relevant fucntion needed from that daemon to oidc core
- remove username and password of oidc input functionality
- that means redirect uri `https://<auth_server_name>/auth/oidc_token` is not needed now. Having that in config works as well but not used.


lib/rucio/core/oidc.py
```diff
@@ -1276,6 +855,42 @@ def delete_expired_oauthrequests(total_workers: int, worker_number: int, limit:
     except Exception as error:
         raise RucioException(error.args) from error
 
+@transactional_session
+def __delete_expired_tokens_account(
```
dchristidis: This function is very similar to `rucio.core.authentication.__delete_expired_tokens_account()`. It would be preferable to generalise that and reuse it here. Also note that `rucio.core.authentication.delete_expired_tokens()`, which was used by the OAuthManager, has remained.

dchristidis (changes requested):
1. Please rebase your branch on top of master instead of merging the latter over the former.
2. Please follow best practices for the commit messages.

lib/rucio/core/oidc.py
panta-123: addressed in commit : https://github.com/rucio/rucio/pull/8035/commits/5c1754945c763c3f16ea8ac31c53918f27bc8bad

dchristidis (changes requested):
1. `bin/rucio-oauth-manager` has not been removed.
 2. A pull request is supposed to tell a story. Like any good story, the structure matters as much as the content. From what I can see, removing the OAuth Manager boils down to a sequence of five actions. Please split your commits accordingly.
    1. Rework some existing functions
    2. Add new functionality to `get_auth_oidc()`
    3. Remove the OAuth Manager
    4. Remove no longer used functions
    5. Remove tests
 3. I’m afraid I must insist on my previous request.
    > Please follow best practices for the commit messages.

    There are many articles and guides on this topic. Looking at the first commit:
    ```
    oidc: remove-oauthmanager-daemon-#7581

    - compltely remove oauth manager daemons
    - the transfer toke now are got using client credential grant already.
    - move delete_oauth_request and delete_expired_tokens from oauthmanager daemon to core
    ```
    1. There is no Rucio component `oidc`
    2. The commit summary is not a sentence
    3. The body contains multiple spelling and grammatical errors
    4. The body is not wrapped
    5. There is a lack of clarity and motivation (for the third item)

And that’s just for the removal of the OAuth Manager. I did not look closely at the removal of the `auto` variant of the client authentication. Is there a reason theses two are bundled together?

lib/rucio/core/oidc.py
dchristidis: Please keep in mind that doing so as a separate commit and bundling it with numerous other changes hinders the review process.

panta-123:
Thanks I guess. Like 38 release , the review comes close to the code freeze of the release , it makes comments not possible to address timely. 

Okay things can be improved on the commit message, but would have been possbible to address and I would have done that if review was done timely.

Any way I am not gonna address more changes in this PR and the PR will be closed.

panta-123:
Closed as PR needs rewrite.

PR #8043, opened 2025-09-22
Testing: Add `addinivalue_line` for flaky tests. #8042

### LINKED ISSUE #8042: Flaky tests not being rerun in CI
I noticed the following line in the CI:

```
  /rucio_source/tests/test_cli_client_structure.py:724: PytestUnknownMarkWarning: Unknown pytest.mark.flaky - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/stable/how-to/mark.html
    @pytest.mark.flaky(reruns=3, reruns_delay=5)
```

meaning that the flaky marker is potentially being ignored.

lobis:
Closes #8042

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -56,6 +56,10 @@ def pytest_configure(config: pytest.Config) -> None:
     config.addinivalue_line('markers', 'dirty: marks test as dirty, i.e. tests are leaving structures behind')
     config.addinivalue_line('markers', 'noparallel(reason, groups): marks test being unable to run in parallel to other tests')
     config.addinivalue_line('markers', 'needs_iam: requires the dev iam profile (OIDC/IdP)')
+    config.addinivalue_line(
+        "markers",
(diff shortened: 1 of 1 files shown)


rdimaio (commented):
I don't think this is the solution needed. This marker is already part of this dependency https://pytest-rerunfailures.readthedocs.io/latest/mark.html - I think the issue is that this is missing from the dev requirements here https://github.com/rucio/rucio/blob/master/setuputil.py, though it was added to the txt file

rdimaio:
spoke offline, example in the CI where this happens https://github.com/rucio/rucio/actions/runs/17919013097/job/50948941594?pr=8033#step:7:5760

https://stackoverflow.com/questions/60806473/pytestunknownmarkwarning-unknown-pytest-mark-xxx-is-this-a-typo

markers need to be explicitly registered - so original PR was correct

rdimaio (approved):

PR #8045, opened 2025-09-22
Core: Simplify timestamp diff calculation #8044

### LINKED ISSUE #8044: Simplify timestamp diff calculation
### Description

https://github.com/rucio/rucio/blob/40bfcc89a1e53462b5cbb309c73b73510f4937f5/lib/rucio/core/rule_grouping.py#L924-L925

This implementation is from when Rucio was being run with Python 2. Nowadays, we can use `timedelta.total_seconds()` instead.

### Motivation

_No response_

### Change

_No response_

dchristidis:
The original implementation was from when Rucio was being run with Python 2. Note that the `total_seconds()` method returns a float.

### DIFF
--- a/lib/rucio/core/rule_grouping.py
+++ b/lib/rucio/core/rule_grouping.py
@@ -921,8 +921,8 @@ def __is_retry_required(lock, activity):
     :param activity:             The activity of the rule.
     """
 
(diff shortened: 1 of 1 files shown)


rdimaio (approved):
some related info https://stackoverflow.com/questions/8777753/converting-datetime-date-to-utc-timestamp-in-python

dchristidis:
Ah, my bad. `total_seconds()` was introduced in Python 2.7, so the original implementation was for Python 2.6 and earlier.

Issue #6476, opened 2024-01-22
Replace uses of `datetime.datetime.utcnow()`


dchristidis:
### Description

This is a continuation from #5781. [`datetime.datetime.utcnow()`](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow) is deprecated since Python 3.12. The recommended replacement appears to be `datetime.datetime.now(datetime.UTC)`.

```python
>>> import datetime
>>> datetime.datetime.utcnow()
<stdin>:1: DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).
datetime.datetime(2024, 1, 22, 7, 11, 40, 566141)
```

We should review the past discussions and consider whether switching from naive to aware `datetime` objects has any unintentional side effects.

bari12:
@dchristidis can you please explain @erlingstaff the background of the discussion a bit so he can investigate.

dchristidis:
Passing an aware object to SQLAlchemy appears to be working correctly.

```python
>>> session.execute(select(Account.account).where(Account.created_at == datetime.datetime(2018, 5, 6, 12, 29, 37))).scalar_one()
2024-03-14 13:07:48,104 INFO sqlalchemy.engine.Engine SELECT atlas_rucio.accounts.account 
FROM atlas_rucio.accounts 
WHERE atlas_rucio.accounts.created_at = :created_at_1
2024-03-14 13:07:48,104 INFO sqlalchemy.engine.Engine [cached since 323.9s ago] {'created_at_1': datetime.datetime(2018, 5, 6, 12, 29, 37)}
dichrist
>>> session.execute(select(Account.account).where(Account.created_at == datetime.datetime(2018, 5, 6, 12, 29, 37, tzinfo=datetime.timezone.utc))).scalar_one()
2024-03-14 13:07:51,863 INFO sqlalchemy.engine.Engine SELECT atlas_rucio.accounts.account 
FROM atlas_rucio.accounts 
WHERE atlas_rucio.accounts.created_at = :created_at_1
2024-03-14 13:07:51,863 INFO sqlalchemy.engine.Engine [cached since 327.7s ago] {'created_at_1': datetime.datetime(2018, 5, 6, 12, 29, 37, tzinfo=datetime.timezone.utc)}
dichrist
```

bari12:
[...]

PR #8048, opened 2025-09-23
Daemons: reaper, avoid multiple reaper workers working on the same replicas; rucio#6512

### LINKED ISSUE #6512: Prevent multiple Reaper threads from working on the same replicas

labkode:
Fixes https://github.com/rucio/rucio/issues/6512

## Overview

This branch introduces two complementary mechanisms to reduce the likelihood of multiple reaper workers working on the same replicas that are about to be deleted. 

The two mechanisms are:
- A) Immediate cleaning of replicas from the Rucio DB (configurable)
- B) Refreshing replicas to be deleted (always enabled)

### Existing Mode (Default)
**Configuration:** `enable_immediate_cleanup = false` (default, can be omitted)

```ini
[reaper]
# Traditional mode - no additional configuration needed
# enable_immediate_cleanup = false  # Default value, can be omitted
delay_seconds = 600                  # Standard replica selection delay
chunk_size = 100                     # Number of replicas to process per batch
```

**Behavior:**
- Database cleanup happens once after all physical deletions (hundreds to thousands) complete
- Maintains the original behavior

### Immediate Cleanup Mode (Opt-in)
**Configuration:** `enable_immediate_cleanup = true`

```ini
[reaper]
enable_immediate_cleanup = true      # Enable immediate cleanup optimization
db_batch_size = 50                   # Batch size for immediate database cleanup (default: 50)
refresh_trigger_ratio = 80           # Percentage of delay_seconds before refreshing (default: 80)
delay_seconds = 600                  # Standard replica selection delay
chunk_size = 100                     # Number of replicas to process per batch
```

**Behavior:**
- Database cleanup happens in configurable batches during physical deletion
- Increases database load (more statements executed)
- Deletions are faster (removed from Rucio DB), therefore making them visible to external scripts.

## Configuration Parameters

| Parameter | Default | Description |
|-----------|---------|-------------|
| `enable_immediate_cleanup` | `false` | Enable/disable immediate database cleanup optimization |
| `db_batch_size` | `50` | Number of replicas to clean from database in each immediate batch |
| `refresh_trigger_ratio` | `80` | Percentage of `delay_seconds` after which to refresh remaining replicas (applies to both traditional and immediate cleanup modes) |
| `delay_seconds` | `600` | Standard delay for replica selection (existing parameter) |
| `chunk_size` | `100` | Number of replicas to process per iteration (existing parameter) |

## B)  Replica Refresh Control
Always enabled. 

The reaper uses a `delay_seconds` mechanism to prevent multiple workers from processing the same replicas. When replicas are marked as `BEING_DELETED`, other workers will not select them until `delay_seconds` have passed since their last update.

To prevent race conditions when processing takes longer than expected, the reaper can refresh the `updated_at` timestamp of remaining replicas:

```ini
[reaper]
delay_seconds = 600                  # Replicas become selectable by other workers after 10 minutes
refresh_trigger_ratio = 80           # Refresh remaining replicas after 80% of delay_seconds (8 minutes)
```

**How it works:**
1. Worker starts processing 100 replicas at time T=0
2. At T=8 minutes (80% of 10 minutes), if replicas are still being processed:
   - Worker calls `refresh_replicas()` on remaining unprocessed replicas
   - This updates their `updated_at` timestamp to current time
   - Other workers will wait another 10 minutes before selecting these replicas
3. Original worker continues processing without interference

### Refresh Configuration Examples

**Conservative (longer processing time allowed):**
```ini
[reaper]
delay_seconds = 900                  # 15 minutes before other workers can take over
refresh_trigger_ratio = 90           # Refresh after 13.5 minutes
```

**Aggressive (faster worker coordination):**
```ini
[reaper]
delay_seconds = 300                  # 5 minutes before other workers can take over  
refresh_trigger_ratio = 70           # Refresh after 3.5 minutes
```

**Multi-worker environment (balanced):**
```ini
[reaper]
delay_seconds = 600                  # Standard 10 minutes
refresh_trigger_ratio = 75           # Refresh after 7.5 minutes (leaves 2.5min buffer)
```

### High-Throughput Environment
Optimize for maximum performance with frequent immediate cleanups:

```ini
[reaper]
enable_immediate_cleanup = true
db_batch_size = 25                   # Smaller batches for more frequent cleanup
refresh_trigger_ratio = 70           # Refresh remaining replicas earlier
delay_seconds = 300                  # Shorter delay for faster processing
chunk_size = 200                     # Larger chunks for higher throughput
```

### Conservative Environment
Optimize for reliability with larger batches:

```ini
[reaper]
enable_immediate_cleanup = true
db_batch_size = 100                  # Larger batches, fewer database calls
refresh_trigger_ratio = 90           # Wait longer before refreshing
delay_seconds = 900                  # Longer delay for stability
chunk_size = 50                      # Smaller chunks for reliability
```

### Multi-Worker Environment
Optimize for coordination between multiple reaper workers:

```ini
[reaper]
enable_immediate_cleanup = true
db_batch_size = 30                   # Moderate batch size
refresh_trigger_ratio = 75           # Refresh before other workers can interfere
delay_seconds = 600                  # Standard delay
chunk_size = 100                     # Standard chunk size
```

### Log Messages

**Traditional Mode:**
```
DEBUG: Deletion complete for RSE CERN-PROD - processed 150 replicas, all 150 will be cleaned up by main loop (traditional mode)
DEBUG: Main loop cleanup SUCCESS - deleted 150 remaining replicas in 2.34 seconds
```

**Immediate Cleanup Mode:**
```
DEBUG: Starting deletion for RSE CERN-PROD with 150 replicas, enable_immediate_cleanup=True, db_batch_size=50, delay_seconds=600
DEBUG: Immediate cleanup SUCCESS: deleted 50 replicas from database (batch #1)
DEBUG: Immediate cleanup SUCCESS: deleted 50 replicas from database (batch #2)
DEBUG: Final cleanup SUCCESS: deleted 50 remaining replicas from database
DEBUG: Deletion complete for RSE CERN-PROD - processed 150 replicas, performed 3 immediate cleanups, total immediate cleaned: 150, remaining for main loop: 0
```

**Replica Refresh Messages:**
```
DEBUG: Refresh trigger time set to 480.0 seconds (80% of delay_seconds=600)
DEBUG: Refresh triggered after 485.2 seconds - refreshing 45 remaining replicas (out of 100 total)
DEBUG: Successfully refreshed 45 remaining replicas after 485.2 seconds
WARNING: Failed to bump updated_at for remaining replicas BEING_DELETED
```

### Configuration Verification

Check active configuration at startup:
```
DEBUG: Optimization configuration - enable_immediate_cleanup=True, db_batch_size=50, refresh_trigger_ratio=80%, delay_seconds=600, chunk_size=100, total_workers=4
```

### Replica Refresh Function

The `refresh_replicas()` function in `rucio.core.replica` provides the underlying mechanism:

```python
from rucio.core.replica import refresh_replicas

# Update the updated_at timestamp for replicas to prevent other workers from taking them
success = refresh_replicas(
    rse_id='CERN-PROD_DATADISK', 
    replicas=[
        {'scope': 'cms', 'name': 'file1.root'},
        {'scope': 'cms', 'name': 'file2.root'}
    ]
)
```

### Common Issues

**Workers taking over each other's work:**
```ini
# Solution: Reduce refresh trigger ratio or increase delay
[reaper]
delay_seconds = 900              # Increase to 15 minutes
refresh_trigger_ratio = 70       # Refresh after 70% (10.5 minutes)
```

**Database performance issues with immediate cleanup:**
```ini
# Solution: Increase batch size to reduce DB calls
[reaper]
enable_immediate_cleanup = true
db_batch_size = 100             # Larger batches, fewer DB operations
```

**Slow processing causing timeouts:**
```ini
# Solution: Increase delay and refresh earlier
[reaper]
delay_seconds = 1200            # 20 minutes total
refresh_trigger_ratio = 60      # Refresh after 12 minutes
```

### DIFF
--- a/lib/rucio/core/replica.py
+++ b/lib/rucio/core/replica.py
@@ -33,7 +33,17 @@ from sqlalchemy import and_, delete, exists, func, insert, not_, or_, union, upd
 from sqlalchemy.exc import DatabaseError, IntegrityError
 from sqlalchemy.orm import aliased
 from sqlalchemy.orm.exc import FlushError, NoResultFound
-from sqlalchemy.sql.expression import ColumnElement, case, false, literal, literal_column, null, select, text, true
+from sqlalchemy.sql.expression import (
+    ColumnElement,
+    case,
+    false,
+    literal,
+    literal_column,
+    null,
+    select,
+    text,
+    true,
+)
 
 import rucio.core.did
 import rucio.core.lock
@@ -41,17 +51,43 @@ from rucio.common import exception
 from rucio.common.cache import MemcacheRegion
 from rucio.common.config import config_get, config_get_bool
 from rucio.common.constants import DEFAULT_VO, RseAttr, SuspiciousAvailability
-from rucio.common.types import InternalAccount, InternalScope, IPDict, LFNDict, is_str_list
+from rucio.common.types import (
+    InternalAccount,
+    InternalScope,
+    IPDict,
+    LFNDict,
+    is_str_list,
+)
 from rucio.common.utils import add_url_query, chunks, clean_pfns, str_to_date
 from rucio.core.credential import get_signed_url
 from rucio.core.message import add_messages
 from rucio.core.monitor import MetricManager
-from rucio.core.rse import get_rse, get_rse_attribute, get_rse_name, get_rse_vo, list_rses
+from rucio.core.rse import (
+    get_rse,
+    get_rse_attribute,
+    get_rse_name,
+    get_rse_vo,
+    list_rses,
+)
 from rucio.core.rse_counter import decrease, increase
 from rucio.core.rse_expression_parser import parse_expression
 from rucio.db.sqla import filter_thread_work, models
-from rucio.db.sqla.constants import OBSOLETE, BadFilesStatus, BadPFNStatus, DIDAvailability, DIDType, ReplicaState, RuleState
-from rucio.db.sqla.session import BASE, DEFAULT_SCHEMA_NAME, read_session, stream_session, transactional_session
+from rucio.db.sqla.constants import (
+    OBSOLETE,
+    BadFilesStatus,
+    BadPFNStatus,
+    DIDAvailability,
+    DIDType,
+    ReplicaState,
+    RuleState,
+)
+from rucio.db.sqla.session import (
+    BASE,
+    DEFAULT_SCHEMA_NAME,
+    read_session,
+    stream_session,
+    transactional_session,
+)
 from rucio.db.sqla.util import temp_table_mngr
 from rucio.rse import rsemanager as rsemgr
 
@@ -2548,7 +2584,7 @@ def list_and_mark_unlocked_replicas(
     :param limit:                    Number of replicas returned.
     :param bytes_:                   The amount of needed bytes.
     :param rse_id:                   The rse_id.
-    :param delay_seconds:            The delay to query replicas in BEING_DELETED state
+    :param delay_seconds:            The delay to query replicas in BEING_DELETED state, default is 10 minutes
     :param only_delete_obsolete      If set to True, will only return the replicas with EPOCH tombstone
     :param session:                  The database session in use.
 
@@ -3199,6 +3235,29 @@ def get_replica_atime(
         models.RSEFileAssociation,
         'INDEX(REPLICAS REPLICAS_PK)',
         'oracle'
+    ).where(
+        and_(models.RSEFileAssociation.scope == replica['scope'],
+             models.RSEFileAssociation.name == replica['name'],
+             models.RSEFileAssociation.rse_id == replica['rse_id'])
+    )
+    return session.execute(stmt).scalar_one()
+
+
+@read_session
+def get_replica_updated_at(
+    replica: dict[str, Any],
+    *,
+    session: "Session"
+) -> datetime:
+    """
+    Get the updated_at timestamp for a replica.
+    :param replica: Dictionary {scope, name, rse_id}
+    :param session: Database session to use.
+
+    :returns: A datetime timestamp with the updated time.
+    """
+    stmt = select(
+        models.RSEFileAssociation.updated_at
     ).where(
         and_(models.RSEFileAssociation.scope == replica['scope'],
              models.RSEFileAssociation.name == replica['name'],
@@ -4484,3 +4543,82 @@ def get_rse_coverage_of_dataset(
             result[rse_id] = total
 
     return result
+
+
+@transactional_session
+def refresh_replicas(
+        rse_id: Optional[str] = None,
+        replicas: Optional['Iterable[dict[str, Any]]'] = None,
+        *,
+        session: "Session"
+) -> bool:
+    """
+    Updates the updated_at timestamp of the given replicas but don't wait if row is locked.
+
+    :param rse_id: the RSE containing the replicas to refresh.
+    :param replicas: a list with replicas (dictionary with the information of the affected replica).
+    :param session: The database session in use.
+
+    :returns: True, if successful, False otherwise.
+    """
+
+    if not replicas or not rse_id:
+        return True
+
+    tt_mngr = temp_table_mngr(session=session)
+    scope_name_temp_table = tt_mngr.create_scope_name_table()
+
+    values = [{'scope': replica['scope'], 'name': replica['name']} for replica in replicas]
+
+    try:
+        stmt = insert(
+            scope_name_temp_table
+        )
+        session.execute(stmt, values)
+
+        stmt = select(
+            func.count(),
+        ).join_from(
+            scope_name_temp_table,
+            models.RSEFileAssociation,
+            and_(models.RSEFileAssociation.scope == scope_name_temp_table.scope,
+                 models.RSEFileAssociation.name == scope_name_temp_table.name,
+                 models.RSEFileAssociation.rse_id == rse_id,
+                 models.RSEFileAssociation.state == ReplicaState.BEING_DELETED)
+        )
+
+        total_to_refresh = session.execute(stmt).scalar_one()
+        if total_to_refresh == 0:
+            # nothing to do
+            return True
+
+        stmt = update(
+            models.RSEFileAssociation
+        ).prefix_with(
+                '/*+ INDEX(REPLICAS REPLICAS_PK) */', dialect='oracle'
+        ).where(
+            exists(select(1)
+                   .select_from(scope_name_temp_table)
+                   .where(
+                       and_(models.RSEFileAssociation.scope == scope_name_temp_table.scope,
+                            models.RSEFileAssociation.name == scope_name_temp_table.name,
+                            models.RSEFileAssociation.rse_id == rse_id)))
+        ).where(
+            models.RSEFileAssociation.state == ReplicaState.BEING_DELETED,
+        ).values({
+            models.RSEFileAssociation.updated_at: datetime.utcnow()
+        }).execution_options(
+            synchronize_session=False
+        )
+
+        session.execute(stmt)
+
+        # clean up temporary table
+        stmt = delete(scope_name_temp_table)
+        session.execute(stmt)
+
+    except DatabaseError:
+        return False
+    except NoResultFound:
+        return True
+    return True
--- a/lib/rucio/daemons/reaper/reaper.py
+++ b/lib/rucio/daemons/reaper/reaper.py
@@ -138,7 +138,7 @@ def get_rses_to_process(
     return rses_to_process
 
 
-def delete_from_storage(heartbeat_handler, hb_payload, replicas, prot, rse_info, is_staging, auto_exclude_threshold, delay_seconds: int = 600, logger=logging.log):
+def delete_from_storage(heartbeat_handler, hb_payload, replicas, prot, rse_info, is_staging, auto_exclude_threshold, delay_seconds: int = 600, logger=logging.log) -> tuple[list[dict[str, Any]], int]:
     """
     Delete replicas from storage and manage database cleanup.
 
@@ -176,8 +176,8 @@ def delete_from_storage(heartbeat_handler, hb_payload, replicas, prot, rse_info,
                 this only contains files that failed immediate cleanup.
               - Number of replicas successfully processed (for metric accounting).
     """
-    deleted_files = []
-    successful_replicas = 0
+    deleted_files: list[dict[str, Any]] = []
+    successful_replicas: int = 0
     rse_name = rse_info['rse']
     rse_id = rse_info['id']
     noaccess_attempts = 0
@@ -870,7 +870,7 @@ def _run_once(
                     logger(logging.DEBUG, 'Main loop cleanup SUCCESS - deleted %d remaining replicas in %.2f seconds', len(deleted_files), time.time() - del_start)
                 else:
                     logger(logging.DEBUG, 'Main loop cleanup - no files remaining, all handled by immediate cleanup optimization')
-                METRICS.counter('deletion.done').inc(len(successful_replicas))
+                METRICS.counter('deletion.done').inc(successful_replicas)
 
                 # Debug: Track cycle metrics
                 cycle_total_replicas_processed += len(file_replicas)
--- a/lib/rucio/daemons/reaper/reaper.py
+++ b/lib/rucio/daemons/reaper/reaper.py
@@ -47,7 +47,7 @@ from rucio.core.heartbeat import list_payload_counts
 from rucio.core.message import add_message
 from rucio.core.monitor import MetricManager
 from rucio.core.oidc import request_token
-from rucio.core.replica import delete_replicas, list_and_mark_unlocked_replicas
+from rucio.core.replica import delete_replicas, list_and_mark_unlocked_replicas, refresh_replicas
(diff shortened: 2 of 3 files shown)


mgajek-cern (commented):
Just my few cents.

The core race condition fix looks solid, but this PR mixes two separate features that would be easier to review as separate changes.

### Key Issues

**Two features in one PR**: The race condition fix (original issue) gets mixed with a new "immediate cleanup" optimization. Consider splitting these.

**Replica comparison bug**
```python
remaining_replicas = [r for r in replicas if r not in processed_replicas]
```
Dictionary comparison is unreliable here. Use `(scope, name)` tuples instead.

**Memory inefficiency:** The `processed_replicas` list grows continuously and is searched linearly for each replica. With large chunk sizes (1000+ replicas), this becomes a performance bottleneck.

**Config inconsistency**: `refresh_trigger_ratio` is retrieved as int but used as percentage. Pick one approach.

### What Works Well

The `refresh_replicas()` function is well-designed and the timing logic makes sense. Good logging for production debugging.

### Suggestion

Split into two PRs: race condition fix first (simpler to review), then the optimization.

rdimaio (commented):
For reference, worth mentioning that this is a follow-up to https://github.com/rucio/rucio/pull/7199

labkode:
@mgajek-cern thank you for the review. I've considered splitting but I decided against it because both mechanisms touch mostly the same files and do the same thing: reduce the likelihood of overlapping work by multiple workers. The changes are done in a way that are backwards compatible with existing deployments.

Nonetheless, I'll follow-up on the other suggestions and the ones from the previous attempt in #7199.

rdimaio (commented):
Requesting Dimitrios' review as well as he has some unresolved comments from https://github.com/rucio/rucio/pull/7199

bari12:
@labkode please squash this now. It shouldn't be reviewed in a wip state, this just makes the work harder for the reviewers, who are already very thinly spread.

labkode:
@bari12 done.

lib/rucio/core/replica.py
```diff
@@ -4484,3 +4541,82 @@ def get_rse_coverage_of_dataset(
...
+            exists(select(1)
+                   .select_from(scope_name_temp_table)
+                   .where(
+                        and_(models.RSEFileAssociation.scope == scope_name_temp_table.scope,
+                            models.RSEFileAssociation.name == scope_name_temp_table.name,
+                            models.RSEFileAssociation.rse_id == rse_id)))
+        ).where(
+            models.RSEFileAssociation.state == ReplicaState.BEING_DELETED,
+        ).values({
+            models.RSEFileAssociation.updated_at: datetime.now(timezone.utc)
```
Geogouz: The `RSEFileAssociation.updated_at` column elsewhere is populated with naïve UTC (`datetime.utcnow()`), and so we will likely end up with comparisons between naïve and aware datetimes. I would suggest to keep the timestamp naïve and aligned with our existing usage, or even better.. migrate the column to become timezone-aware.

lib/rucio/daemons/reaper/reaper.py
```diff
@@ -215,15 +314,103 @@ def delete_from_storage(heartbeat_handler, hb_payload, replicas, prot, rse_info,
...
+                               len(pending_db_deletions))
+                        delete_replicas(rse_id=rse_id, files=pending_db_deletions)
+                        immediate_cleanups += 1
+                        total_immediate_cleaned += len(pending_db_deletions)
+                        logger(logging.DEBUG, 'Immediate cleanup SUCCESS: deleted %d replicas from database (batch #%d)',
+                               len(pending_db_deletions), immediate_cleanups)
+                        # Remove successfully cleaned up files from deleted_files to avoid duplicate processing
+                        for replica_dict in pending_db_deletions:
+                            if replica_dict in deleted_files:
+                                deleted_files.remove(replica_dict)
```
Geogouz, line 352: Doesn't the removal of those entries here mean that they are also missed as counted by the following `METRICS.counter('deletion.done').inc(len(deleted_files))`? Is this desired?

lib/rucio/core/replica.py
labkode: Ok I have reverted back to `utcnow()`. I see there is already an issue tracking this usage: https://github.com/rucio/rucio/issues/6476

lib/rucio/daemons/reaper/reaper.py
labkode: Good catch. I've updated the variable to count the processed replicas (deleted replicas). 
The counter will be foul when using the immediate_cleanup mode where deletion happens internally in the inner call to delete the physical replica.

tests/test_replica.py
```diff
@@ -287,6 +303,29 @@ def test_delete_replicas_from_datasets_old(self, rse_factory, mock_scope, root_a
...
+        # we update the state to BEING_DELETED
+        update_replica_state(rse_id, r['scope'], r['name'], ReplicaState.BEING_DELETED)
+        original = get_replica_updated_at(r)
+
+        time.sleep(2)
+        ok = refresh_replicas(rse_id, [r])
+        assert ok is True
+
+        got = get_replica_updated_at(r)
+        assert got > original
```
Geogouz, line 328: Can you add a breakline here please? There is a PR under review that will flag this once merged.

lib/rucio/daemons/reaper/reaper.py
```diff
@@ -215,15 +314,103 @@ def delete_from_storage(heartbeat_handler, hb_payload, replicas, prot, rse_info,
...
+                        ok = refresh_replicas(rse_id=rse_id, replicas=remaining_replicas)
+                        if not ok:
+                            logger(logging.WARNING, "Failed to bump updated_at for remaining replicas BEING_DELETED")
+                        else:
+                            logger(logging.DEBUG, 'Successfully refreshed %d remaining replicas after %.1f seconds',
+                                   len(remaining_replicas), elapsed_time)
+                        refresh_start_time = time.time() # reset it so we can trigger new refresh cycles.
+                    else:
+                        logger(logging.DEBUG, 'No remaining replicas to refresh after %.1f seconds', elapsed_time)
+
```
Geogouz, line 387: Same reason. Can you remove a breakline please?

lib/rucio/daemons/reaper/reaper.py
```diff
@@ -215,15 +314,103 @@ def delete_from_storage(heartbeat_handler, hb_payload, replicas, prot, rse_info,
...
+                    if remaining_replicas:
+                        logger(logging.DEBUG, 'Refresh triggered after %.1f seconds - refreshing %d remaining replicas (out of %d total)',
+                               elapsed_time, len(remaining_replicas), len(replicas))
+                        ok = refresh_replicas(rse_id=rse_id, replicas=remaining_replicas)
+                        if not ok:
+                            logger(logging.WARNING, "Failed to bump updated_at for remaining replicas BEING_DELETED")
+                        else:
+                            logger(logging.DEBUG, 'Successfully refreshed %d remaining replicas after %.1f seconds',
+                                   len(remaining_replicas), elapsed_time)
+                        refresh_start_time = time.time() # reset it so we can trigger new refresh cycles.
```
Geogouz: There should be two spaces before "# reset it so we can trigger new refresh cycles.". Can you add one?

lib/rucio/daemons/reaper/reaper.py
```diff
@@ -215,15 +314,103 @@ def delete_from_storage(heartbeat_handler, hb_payload, replicas, prot, rse_info,
...
+                # on individual deletions.
+                num_replicas_processed += 1
+
+                # After each replica is deleted we evaluate how much time we have left to delete.
+                # If we are not able to delete all the replicas we got in, other workers will take the replicas because their update time will
+                # be more than delay_seconds (refer to function list_and_mark_unlocked_replicas).
+                # After trigger_time has passed, and if we still have some replicas to process, we bump the replicas updated_at field to current time
+                # (which will delay the selectability by other workers by delay_seconds+ minutes).
+                elapsed_time = time.time() - refresh_start_time
+                if elapsed_time > trigger_time: # trigger_time has passed
```
Geogouz: There should be at least two spaces before inline comment. Can you add 1 to be ok with `E261`?

Geogouz:
Once we merge #8269, Rucio's ruff [pre-commit checks](https://rucio.github.io/documentation/contributing/#1-prerequisite) will be required to pass also on the CI. Right now this is not enforced because we have miss-configured things and so, only when someone has enabled the pre-commit checks locally, things are taken care of. I have flagged couple problems that will arise in the future but to avoid mentioning all of them one by one in the review (as there are some more), maybe you can enable the pre-commit locally to fix the problems related to the files you have modified (note: do not mind about the `E501` line too long). But if you don't have time, don't bother (they are just stylistic). Just address the flagged ones here and I can fix everything once we merge #8269. No problem.

lib/rucio/daemons/reaper/reaper.py
```diff
@@ -648,14 +852,25 @@ def _run_once(
...
-                METRICS.counter('deletion.done').inc(len(deleted_files))
+                # Then delete any remaining replicas that weren't handled by immediate cleanup
+                if deleted_files:
+                    logger(logging.DEBUG, 'Main loop cleanup - %d files remaining for database deletion after immediate cleanup optimization', len(deleted_files))
+                    del_start = time.time()
+                    delete_replicas(rse_id=rse.id, files=deleted_files)  # type: ignore (argument missing: session)
+                    logger(logging.DEBUG, 'Main loop cleanup SUCCESS - deleted %d remaining replicas in %.2f seconds', len(deleted_files), time.time() - del_start)
+                else:
+                    logger(logging.DEBUG, 'Main loop cleanup - no files remaining, all handled by immediate cleanup optimization')
+                METRICS.counter('deletion.done').inc(len(file_replicas))
```
Geogouz: Should that be `len(deleted_files)` instead?

labkode: No. the `deleted_files` can be empty if the immediate cleanup is enabled. file_replicas are all the files that were deleted either in the main loop or inside `delete_from_storage` call.

Geogouz: `file_replicas` is just the chunk pulled from the DB and not the subset that succeeded, right? So the increment occurs now for every chunk, regardless of whether individual replicas were actually deleted - maybe there was some exception. Hence, the `done` in `deletion.done` is a bit misleading now and should not imply successful deletion because failures still inflate the metric.

Geogouz: If this behaviour is expected/intended, I'm ok.

labkode: You're right, failures will be still counted as successful. Let me check how this was calculated before.

Geogouz: Since `deleted_files` can indeed be empty as you said, maybe the increment should reflect the replicas actually deleted based on pending deletions before they’re cleared or a separate success counter within `delete_from_storage`.

labkode:
@Geogouz thanks for the changes, all good from my side.

lib/rucio/daemons/reaper/reaper.py
```diff
@@ -215,15 +320,102 @@ def delete_from_storage(heartbeat_handler, hb_payload, replicas, prot, rse_info,
...
+                # on individual deletions.
+                num_replicas_processed += 1
+
+                # After each replica is deleted we evaluate how much time we have left to delete.
+                # If we are not able to delete all the replicas we got in, other workers will take the replicas because their update time will
+                # be more than delay_seconds (refer to function list_and_mark_unlocked_replicas).
+                # After trigger_time has passed, and if we still have some replicas to process, we bump the replicas updated_at field to current time
+                # (which will delay the selectability by other workers by delay_seconds+ minutes).
+                elapsed_time = time.time() - refresh_start_time
+                if elapsed_time > trigger_time:  # trigger_time has passed
```
Geogouz, line 372: For safety, and to avoid having other workers starved indefinitely if a worker is slow or stuck, should we better treat non-positive `delay_seconds` (which defaults to 0) as “no refresh” (skipping the refresh block when trigger_time <= 0)?

For example, right before:

`if refresh_trigger_ratio <= 0 or refresh_trigger_ratio > 1:`

we can add:

`refresh_enabled = delay_seconds > 0`

and then instead of:

```
    trigger_time = delay_seconds * refresh_trigger_ratio
    logger(logging.DEBUG, 'Refresh trigger time set to %.1f seconds (%.0f%% of delay_seconds=%d)',
           trigger_time, refresh_trigger_ratio * 100, delay_seconds)
```

have something like:

```
    trigger_time = delay_seconds * refresh_trigger_ratio if refresh_enabled else None
    if refresh_enabled:
        logger(logging.DEBUG, 'Refresh trigger time set to %.1f seconds (%.0f%% of delay_seconds=%d)',
               trigger_time, refresh_trigger_ratio * 100, delay_seconds)
    else:
        logger(logging.DEBUG, 'Replica refresh disabled because delay_seconds=%d', delay_seconds)
```

and then replace:

`if elapsed_time > trigger_time:  # trigger_time has passed`

with:

`if refresh_enabled and elapsed_time > trigger_time:  # trigger_time has passed`

bari12:
@labkode I am closing this now. Please resubmit once we have another draft ready.

Issue #8058, opened 2025-09-25
Enable OIDC authentication in development environment


ftorradeflot:
OIDC authentication in the development environment with Indigo IAM does not work out of the box, some additional steps are needed:
* create SCIM client in Indigo IAM
* add redirect_uris
* update idpsecrets.json
* add identity to the root user

It would be convenient to automate these steps

PR #8059, opened 2025-09-25
Tooling: Update dev environment to be able to use OIDC login with indigoiam #8058

ftorradeflot:
Implement the automation described in #8058

### DIFF
--- a/etc/docker/dev/iam/indigoiam_db.sql
+++ b/etc/docker/dev/iam/indigoiam_db.sql
@@ -1,8 +1,9 @@
--- MariaDB dump 10.19  Distrib 10.11.6-MariaDB, for debian-linux-gnu (x86_64)
+/*M!999999\- enable the sandbox mode */
+-- MariaDB dump 10.19  Distrib 10.11.14-MariaDB, for debian-linux-gnu (x86_64)
 --
 -- Host: localhost    Database: indigoiam
 -- ------------------------------------------------------
--- Server version	10.11.6-MariaDB-1:10.11.6+maria~ubu2204
+-- Server version	10.11.14-MariaDB-ubu2204
 
 /*!40101 SET @OLD_CHARACTER_SET_CLIENT=@@CHARACTER_SET_CLIENT */;
 /*!40101 SET @OLD_CHARACTER_SET_RESULTS=@@CHARACTER_SET_RESULTS */;
@@ -29,7 +30,7 @@ USE `indigoiam`;
 
 DROP TABLE IF EXISTS `access_token`;
 /*!40101 SET @saved_cs_client     = @@character_set_client */;
-/*!40101 SET character_set_client = utf8 */;
+/*!40101 SET character_set_client = utf8mb4 */;
 CREATE TABLE `access_token` (
   `id` bigint(20) NOT NULL AUTO_INCREMENT,
   `token_value` varchar(4096) DEFAULT NULL,
@@ -45,7 +46,7 @@ CREATE TABLE `access_token` (
   KEY `at_tv_idx` (`token_value`(767)),
   KEY `at_exp_idx` (`expiration`),
   KEY `at_ahi_idx` (`auth_holder_id`)
-) ENGINE=InnoDB AUTO_INCREMENT=33 DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci;
+) ENGINE=InnoDB AUTO_INCREMENT=181 DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci;
 /*!40101 SET character_set_client = @saved_cs_client */;
 
 --
@@ -54,6 +55,9 @@ CREATE TABLE `access_token` (
 
 LOCK TABLES `access_token` WRITE;
 /*!40000 ALTER TABLE `access_token` DISABLE KEYS */;
+INSERT INTO `access_token` VALUES
+(37,'eyJraWQiOiJyc2ExIiwiYWxnIjoiUlMyNTYifQ.eyJpc3MiOiJodHRwczpcL1wvaW5kaWdvaWFtXC8iLCJhdWQiOiI4NWU2ZjdhNS01ODBiLTRhMWMtYTZkMi0zOTA1NTE0MzA2M2QiLCJpYXQiOjE3NTgxOTkwNTAsImp0aSI6IjMxMGMzNDkxLWJjY2QtNDg4Ni1iMGU0LTA0NDc0NDcwNTYyNSJ9.OoWvvn7ipKq-ySaQFtZq2dSaEcgGUFe99x2aMhZQ75ohg9yLLNgbYIz8b9DKLKi3Wh9nf5CC-aP-wfinnOcsQPZSVCF0hCjs3ZySNdJvpNJvT9By6q9qjRzaPuH6Zrt47lMtkbJpr26SySMmOPA4PCQ0KNgt3rICkTJ2H7WuSh4',NULL,'Bearer',NULL,2,37,NULL,NULL),
+(38,'eyJraWQiOiJyc2ExIiwiYWxnIjoiUlMyNTYifQ.eyJpc3MiOiJodHRwczpcL1wvaW5kaWdvaWFtXC8iLCJhdWQiOiIxMTAzNDY2Ny01YzUwLTRjOWEtOGY0My02YTQ4MDkwZGNlZDIiLCJpYXQiOjE3NTgyMDI3MjAsImp0aSI6ImY5MjM4ZTZhLWVmMmItNDNmZS04MDI0LTBmYjJiZGZhMzQ1YSJ9.iDQgj6w4K4VwDRMRyTbRim5-hGiQN8bJbX4aqWW3vIpxD6VEx0RH_FcV_VbHjXbcXXwo1BNT9Rxn-GEnr2YGsxvwG7rOuGpZPUn1ZuLLpSQEOIKW2Pbw77qnh-SG3GDfcO1dqbsK4R6qUW6amyHX8q2BJfKZG2_HQXeNdvQm7XY',NULL,'Bearer',NULL,9,38,NULL,NULL);
 /*!40000 ALTER TABLE `access_token` ENABLE KEYS */;
 UNLOCK TABLES;
 
@@ -63,7 +67,7 @@ UNLOCK TABLES;
 
 DROP TABLE IF EXISTS `access_token_permissions`;
 /*!40101 SET @saved_cs_client     = @@character_set_client */;
-/*!40101 SET character_set_client = utf8 */;
+/*!40101 SET character_set_client = utf8mb4 */;
 CREATE TABLE `access_token_permissions` (
   `access_token_id` bigint(20) NOT NULL,
   `permission_id` bigint(20) NOT NULL
@@ -85,7 +89,7 @@ UNLOCK TABLES;
 
 DROP TABLE IF EXISTS `address`;
 /*!40101 SET @saved_cs_client     = @@character_set_client */;
-/*!40101 SET character_set_client = utf8 */;
+/*!40101 SET character_set_client = utf8mb4 */;
 CREATE TABLE `address` (
   `id` bigint(20) NOT NULL AUTO_INCREMENT,
   `formatted` varchar(256) DEFAULT NULL,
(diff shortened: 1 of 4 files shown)


mgajek-cern (commented):
Tested the PR locally and wanted to share my findings to help get it working.

## Test Commands Used
```bash
# Start dev environment with IAM profile
docker compose -f etc/docker/dev/docker-compose.yml --profile iam up -d

# Enter the Rucio container
docker exec -it dev-rucio-1 /bin/bash

# Initialize and activate default RSEs
tools/run_tests.sh -ir

# Initialize test accounts
python3 /rucio_source/tools/bootstrap_tests.py

# Verify the OIDC identity was added
rucio-admin --host https://rucio --auth-host https://rucio \
            --auth-strategy userpass \
            --user root --password ddmlab \
            account list-identities admin

# Test OIDC authentication
rucio --verbose --host https://rucio --auth-host https://rucio \
      --auth-strategy oidc \
      --oidc-issuer indigoiam \
      --oidc-user admin \
      --oidc-password admin \
      --oidc-auto \
      --oidc-scope "openid profile" \
      whoami
```

## Issues Found

1. **Identity mapping behavior** - OIDC authentication consistently maps to `root` instead of `admin`, even though the admin identity (SUB=73f16d93...) is correctly registered in the database.
2. **Credentials unclear** - The admin password doesn't seem to be "admin" in the IndigoIAM database. Could you clarify what the correct credentials should be, or would it help to document them in a README?
3. **Database dump approach** - The PR uses a pre-populated database dump, which works but might be fragile if someone rebuilds their environment. Have you considered initialization scripts that configure everything programmatically? That might better achieve the "automate setup" goal from the issue.
4. **Keycloak configuration** - I noticed the Keycloak section has what appear to be IndigoIAM credentials. Was this tested, or is it meant as a template for future use?

## Questions

**Could you share your testing workflow?** Understanding your exact sequence of commands would help me reproduce your successful setup.

**About fresh environments** - Would you be able to test this in a completely fresh environment (new checkout, no cached containers)? I'm wondering if there are local assumptions that don't hold elsewhere.

The foundation here is solid. Just needs some polish on the authentication flow and credential documentation. Happy to help test any updates.

ftorradeflot:
Answering to your comments (quoted)

> 
> # Verify the OIDC identity was added
> rucio-admin --host https://rucio --auth-host https://rucio \
>             --auth-strategy userpass \
>             --user root --password ddmlab \
>             account list-identities admin

The Rucio account is "root" that holds a userpass identity: ddmlab/secret
```
rucio-admin --user ddmlab --password secret account list-identities root
```

> # Test OIDC authentication
> rucio --verbose --host https://rucio --auth-host https://rucio \
>       --auth-strategy oidc \
>       --oidc-issuer indigoiam \
>       --oidc-user admin \
>       --oidc-password admin \
>       --oidc-auto \
>       --oidc-scope "openid profile" \
>       whoami

The "root" Rucio account also maps to an identity in Indigo IAM: admin/password

```
rm /tmp/root/.rucio_root/auth_token_for_account_root
rucio --verbose --account root \
    --auth-strategy oidc \
    --oidc-issuer indigoiam \
    --oidc-user admin \
    --oidc-password password \
    --oidc-auto \
    --oidc-scope "openid profile" \
    --oidc-audience rucio \
    whoami
```

> ## Issues Found
> 1. **Identity mapping behavior** - OIDC authentication consistently maps to `root` instead of `admin`, even though the admin identity (SUB=73f16d93...) is correctly registered in the database.

I don't see why this is an issue. Rucio's "root" account maps to "admin" in Indigo IAM.

> 2. **Credentials unclear** - The admin password doesn't seem to be "admin" in the IndigoIAM database. Could you clarify what the correct credentials should be, or would it help to document them in a README?

The password for the admin user in Indigo IAM is "password". I didn't make it up, it was already there and I don't remember how I figured it out. Maybe from the [Indigo IAM documentation](https://indigo-iam.github.io/v/v1.7.2/docs/getting-started/basic_setup/)

> 3. **Database dump approach** - The PR uses a pre-populated database dump, which works but might be fragile if someone rebuilds their environment. Have you considered initialization scripts that configure everything programmatically? That might better achieve the "automate setup" goal from the issue.

This approach was already there, I just updated the DB dump after making some changes in IAM. I understand this is the (only) way to have a frozen `idpsecrets.json` file. Otherwise you would have to dynamically create it during container start up, before Rucio initialization. It can probably be done but looks tricky, I don't think it's worth the effort.

> 4. **Keycloak configuration** - I noticed the Keycloak section has what appear to be IndigoIAM credentials. Was this tested, or is it meant as a template for future use?

Keycloak wasn't tested, I just modified the idpsecrets.json file to make sure Rucio doesn't raise an error when loading it. To my understanding Keycloak integration is not yet supported.

> 
> ## Questions
> **Could you share your testing workflow?** Understanding your exact sequence of commands would help me reproduce your successful setup.

My tests are pretty much what you did. I started these changes to test new features related to OIDC authentication in the jupyterlab extension, but these are out of the scope of this PR.

> **About fresh environments** - Would you be able to test this in a completely fresh environment (new checkout, no cached containers)? I'm wondering if there are local assumptions that don't hold elsewhere.

I tested with a fresh environment and it worked. Most of the images with the "latest" tag are produced within Rucio so I guess we should be fine.
 
> The foundation here is solid. Just needs some polish on the authentication flow and credential documentation. Happy to help test any updates.

mgajek-cern:
You're right about the credentials, it's `admin:password`, not `admin:admin`. Also, the command should list identities for the `root` account, not `admin` (which doesn't exist). The IAM user `admin` maps to Rucio's `root` account.

**About the access tokens:**
The `registration_access_token` in `idpsecrets.json` is used by Rucio to manage the SCIM client registration via IAM's API. It's part of the OAuth2 Dynamic Client Registration flow. These tokens will expire if the database is rebuilt or after extended runtime.

**Alternative approach:**
Instead of the SQL dump with frozen tokens, we could use an initialization script that programmatically registers the SCIM client on startup and writes fresh tokens. Basic example:

```bash
# Register SCIM client (no authentication required)
curl -X POST "https://indigoiam/register" \
  -H "Content-Type: application/json" \
  -d '{
    "client_name": "rucio_scim",
    "grant_types": ["client_credentials"],
    "scope": "scim:read",
    "contacts": ["admin@iam.test"]
  }' | jq -r '.client_id, .client_secret, .registration_access_token'
# Parse response and update idpsecrets.json
```

~~However, IAM's client registration API isn't well documented, and the current SQL dump approach works reliably for the dev environment.~~ 

The SQL dump approach trades flexibility for simplicity. We could implement dynamic client registration using IAM's [RFC 7591 endpoints](https://datatracker.ietf.org/doc/html/rfc7591), but it adds startup orchestration complexity for marginal benefit in a dev environment. The tokens only expire if you rebuild the database or run for extended periods, which is rare in local dev. If token expiration becomes a real issue, we can add an init script. 

The `registration_access_token` is optional. It's only used to modify the client registration itself (per RFC 7591), not for SCIM operations. Rucio uses `client_id` and `client_secret` to obtain access tokens for actual SCIM API calls

mgajek-cern (approved):

ftorradeflot:
Thanks @mgajek-cern !

lib/rucio/client/baseclient.py
```diff
@@ -425,6 +426,8 @@ def _load_json_data(self, response: requests.Response) -> 'Generator[Any, Any, A
                 yield response.text
 
     def _reduce_data(self, data, maxlen: int = 132) -> str:
+        if isinstance(data, dict):
+            data = json.dumps(data)
```
rdimaio, line 430: It's not clear to me how this change relates to the rest of the changes here - could you please clarify it, and possibly separate it in a different commit?

etc/docker/dev/rucio/idpsecrets.json
```diff
@@ -2,11 +2,25 @@
...
-    "issuer": "https://indigoiam/"
+    "issuer": "https://indigoiam/",
+    "redirect_uris": [
+            "https://rucio/auth/oidc_token",
+            "https://rucio/auth/oidc_code"
+    ],
+    "SCIM": {
+        "client_id": "11034667-5c50-4c9a-8f43-6a48090dced2",
+        "client_secret": "eIcPpu5HbjuECyTVUB0iBAwPgIDEaJUm8T4xpsn5tyWH53CTxoG9cH2Vo6etYMWegPeWffdbrEPwwkOJZCHUgQ",
+        "registration_access_token": "eyJraWQiOiJyc2ExIiwiYWxnIjoiUlMyNTYifQ.eyJpc3MiOiJodHRwczpcL1wvaW5kaWdvaWFtXC8iLCJhdWQiOiIxMTAzNDY2Ny01YzUwLTRjOWEtOGY0My02YTQ4MDkwZGNlZDIiLCJpYXQiOjE3NTgyMDI3MjAsImp0aSI6ImY5MjM4ZTZhLWVmMmItNDNmZS04MDI0LTBmYjJiZGZhMzQ1YSJ9.iDQgj6w4K4VwDRMRyTbRim5-hGiQN8bJbX4aqWW3vIpxD6VEx0RH_FcV_VbHjXbcXXwo1BNT9Rxn-GEnr2YGsxvwG7rOuGpZPUn1ZuLLpSQEOIKW2Pbw77qnh-SG3GDfcO1dqbsK4R6qUW6amyHX8q2BJfKZG2_HQXeNdvQm7XY"
```
rdimaio, line 13: Is it safe to have these secrets exposed and not injected?

etc/docker/dev/iam/indigoiam_db.sql
rdimaio, line 1: Could you clarify the differences in this dump, and possibly separate this change in a separate commit?

lib/rucio/client/baseclient.py
ftorradeflot: the client crashed when using --oidc-auto authentication because the credentials were received as a dictionary. It is true it is not related to the rest. I will move it to an independent commit

etc/docker/dev/rucio/idpsecrets.json
ftorradeflot: I don't think this poses any risk, since this is supposed to run locally with docker-compose. I just added the SCIM part, does it make any difference with the rest of the information? By "injected" you mean querying Indigo IAM and generating the idpsecrets.json on container startup? If this is the case I don't know if it's possible, but I don't think it's worth it

etc/docker/dev/iam/indigoiam_db.sql
ftorradeflot: The differences come from applying the guide for [Server Configuration for Open ID Connect AuthN/Z](https://rucio.github.io/documentation/operator/installing_server/#server-configuration-for-open-id-connect-authnz) and basically include:
* add a second client for the SCIM part
* add necessary grants to each client
* add necessary scopes
* add redirect uris

ftorradeflot: it goes together with the changes in idpsecrets.json and bootstrap_tests.py, I wouldn't move it into a separate commit

lib/rucio/client/baseclient.py
ftorradeflot: In fact, this should probably go into a separate PR, do you think it is worth it @rdimaio ?

rdimaio: I think it's fine to just have it as a separate commit, but it might be a good idea to have a separate PR for it. In general I think all that would be good to have here is just a sign-off from a client component lead to check that this change is fine and doesn't potentially break anything - @voetberg what do you think?

voetberg: I'd say it's fine as part of this PR, as long as it's marked as a different type of change in the commit.  From the tests it doesn't seem to interfere with anything downstream.

ftorradeflot: Done! is it ok now?

etc/docker/dev/iam/indigoiam_db.sql
rdimaio: The thing is that some changes (e.g. the `utf8mb4` and the `AUTO_INCREMENT` changes) feel a bit 'unexplained' - that said, I am not too familiar with the indigoiam side of things of Rucio, so I'll resolve this and let someone with more context review

ftorradeflot: About the utf8 to utf8mb4 change, I suppose it is because the version of MariaDB that I used for the dump is newer and has the utf8mb4 charset by default. The character set for the client is just changed during table creation and then it gets restored to the original one.

The autoincrement changes are probably because during my testing additional records were created and destroyed. It is just the starting point for the incremental ids.

Both types of changes could probably be removed without any impact on the funcionality. I don't think it is necessary, but it can be done.

Issue #7250, opened 2024-12-13
New CLI: Review keyword vs positional arguments


dchristidis:
### Description

I wish to return to my observation in https://github.com/rucio/rucio/pull/7098#discussion_r1876216479.

Many commands have a single required argument. Some require one but accept more of the same type. For both those cases, having to use a command-line option feels unnecessary.

Examples: `rucio rule show --rule-id …` and `rucio replica list file --did … --did …`.

voetberg:
For the second example, you don't need to use a option flag for each supplied option (it accepts them with a comma delimiter I believe), but this is still a good recommendation. I'll take a first pass stab at this (checking for either required kwargs, or commands with only one argument/argument with a flag.) 

There is also a way to allow both option and positional notation, and I think as long as they don't both show in the help menu, it wouldn't hurt.

voetberg:
To clarify on that second point - it looks like this. 

```
usage: rucio scope add [-h] [account] [scope]

positional arguments:
  account     Account name for filtering, attribution (Note: You can also use --account)
  scope       Name of the new scope to add (Note: You can also use --scope)

optional arguments:
  -h, --help  show this help message and exit
```

and any combination of `rucio scope add --account account --scope scope` or `rucio scope add account scope` works.

dchristidis:
My initial reaction to this was indifferent, but maybe because of using `scope add` as an example.

The most intuitive interface would have been for the `account` to default to the issuer account and `scope` to be the single required positional argument. However, that is a dangerous thing to do while we don’t have `scope update --account=…`.

voetberg:
Solved by #7253 - If this needs additional review, let us revisit this

dchristidis:
`rucio rule list` still requires `--did` instead of accepting it as a positional argument.

voetberg:
> `rucio rule list` still requires `--did` instead of accepting it as a positional argument.

`--did` was left as it was because `--rule-id` is still _technically_ a way to list rules (though it is a duplication of `rule show` and hidden.) Both should be removed at once.

dchristidis:
The fix was effectively reverted as part of #7709.

bari12:
Ongoing in Dimitrios & Martin review; More specific change requests needed

PR #8061, opened 2025-09-26
Additional options for rule list

### LINKED ISSUE #7970: Listing rules in CLI results in empty response
Rucio version: `v38.0.0`

Running this command `rucio rule list` results in an empty response, while running `rucio rule list --account root` properly returns the rules.

I was under the impression that the plain command should return all rules, if that is the case this might be a bug.

voetberg:
Closes #7970 

Second commit makes a few more logical changes that could or could not be included in this PR, I can revert it and add it to a different pull if so desired

### DIFF
--- a/lib/rucio/cli/rule.py
+++ b/lib/rucio/cli/rule.py
@@ -11,11 +11,16 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-import click
+from typing import Optional
 
-from rucio.cli.bin_legacy.rucio import add_rule, delete_rule, info_rule, list_rules, list_rules_history, move_rule, update_rule
+import click
+from tabulate import tabulate
+
+from rucio.cli.bin_legacy.rucio import add_rule, delete_rule, info_rule, list_rules_history, move_rule, update_rule
 from rucio.cli.utils import Arguments
+from rucio.client.richclient import CLITheme, generate_table, print_output
 from rucio.common.exception import InputValidationError
+from rucio.common.utils import extract_scope, sizefmt
 
 
 @click.group()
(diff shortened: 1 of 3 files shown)


lib/rucio/cli/rule.py
```diff
@@ -151,17 +156,109 @@ def update(
 
 @rule.command("list")
 @click.option("--did", help="Filter by DID")
-@click.option("--traverse", is_flag=True, default=False, help="Traverse the did tree and search for rules affecting this did")
+@click.option("--traverse", is_flag=True, default=False, help="Traverse the did tree and search for rules affecting this did. Must supply a DID.")
 @click.option("--csv", is_flag=True, default=False, help="Comma Separated Value output")
-@click.option("--file", help="Filter by file")
+@click.option("--lock", is_flag=True, default=False, help="See all rules locking a DID")
```
dchristidis, line 161: That is not quite right. ‘DID’ without further designation can mean any DID type: containers, datasets, files. Any rule is ‘locking’ a DID. The purpose of `--file` is to do something like a reverse search, providing a file DID and listing the rules that effectively contain it. Note how this differs from providing a file DID without the `--file` option, which list the rules that match it exactly.

Seeing how the description¹ of this subcommand is misleading, as well as the regression described in #7250, I’m given the impression that we’re moving in the wrong direction. I’ll try to prepare a design description of this subcommand and its envisioned use cases. And given my [comment](https://github.com/rucio/rucio/issues/7970#issuecomment-3345227794) about the associated issue, I would recommend not moving forward on this PR.

¹ ‘List all rules impacting a given DID’

voetberg:
I have heard a lot of conflicting opinions about how people want `rule list` to work, and I do think it warrants further discussion. However, I think that `--file` is an incredibly confusing kwarg as well, as files are not really a concept rules interact with before this point (e.g. you assign a rule with `did` as the object you act upon.) If '--file' is really just a reverse search for a did, shouldn't we just provide a flag called "--reverse-search" or something of the like?

I have also seen people very frustrated with the lack of discoverability for rules (talking with CMS colleagues and more casual users of Rucio), but there doesn't seem to be a consensus on how that should actually work. This almost certainly comes from the confusion about the arguments. I would love to iterate on a proposal for how to improve it with you.

voetberg:
Closing pending discussion

PR #8063, opened 2025-09-28
Docker & Kubernetes: Use `PROFILE_COUNT` for reliable profile selection in bootstrap_dev.sh #8062

### LINKED ISSUE #8062: Skip expanding an empty array when -p is provided without a profile name in `bootstrap_dev.sh`
### Description

As the title suggests, `bootstrap_dev.sh` needs to skip expanding an empty array when -p is provided without a profile name cause it makes the process crash.

### Steps to reproduce

Try to execute just `./tools/bootstrap_dev.sh -m -p` instead of something like `./tools/bootstrap_dev.sh -m -p storage`

### Rucio Version

Server: 38.2.0

### Additional Information

_No response_

Geogouz:
Fixes #8062

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -75,7 +75,7 @@ def pytest_runtest_setup(item: pytest.Item) -> None:
         # Check if we're running in a dev environment with IAM profile available
         dev_profiles = environ.get('DEV_PROFILES', '').split(',')
         dev_profiles = [profile.strip() for profile in dev_profiles if profile.strip()]
-        
+
         if 'iam' not in dev_profiles:
             pytest.skip("Test requires IAM profile - start dev environment with: --profile iam")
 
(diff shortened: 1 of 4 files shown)


mgajek-cern (commented):
The fix works well and resolves the crash.

One suggestion to simplify: instead of tracking both `PROFILES` array and `PROFILE_COUNT`, you could just guard the loop with the array length check:

```bash
if [[ ${#PROFILES[@]} -gt 0 ]]; then
  for prof in "${PROFILES[@]}"; do
    profile_args+=( --profile "$prof" )
  done
fi
```

This removes the need for `PROFILE_COUNT` entirely while accomplishing the same goal. Just a thought to reduce the moving parts. Either way, the functionality is solid.

**Note on the crash:** The script uses `set -u` which treats unset/empty variables as errors. When `-p` has no value, the empty `PROFILES` array can trigger "unbound variable" errors when referenced in certain contexts (like string interpolation or unguarded expansions), causing immediate script termination. The guard `if [[ ${#PROFILES[@]} -gt 0 ]]` prevents this by never attempting to expand the empty array.

Geogouz:
> The fix works well and resolves the crash. One suggestion to simplify:..

Yes. Such array length checks were already happening in couple places but I switched those too with the `PROFILE_COUNT` because I think readability-wise,
 
`if (( PROFILE_COUNT > 0 )); then`

is easier to the eye compared to:

`if [[ "${#PROFILES[@]}" -gt 0 ]]; then` and
`if [[ "${#profile_args[@]}" -eq 0 ]]; then`

Dunno. I'm 51% so if you are 52%, you win.

MGTheTrain:
> > The fix works well and resolves the crash. One suggestion to simplify:..
> 
> Yes. Such array length checks were already happening in couple places but I switched those too with the `PROFILE_COUNT` because I think readability-wise,
> 
> `if (( PROFILE_COUNT > 0 )); then`
> 
> is easier to the eye compared to:
> 
> `if [[ "${#PROFILES[@]}" -gt 0 ]]; then` and `if [[ "${#profile_args[@]}" -eq 0 ]]; then`
> 
> Dunno. I'm 51% so if you are 52%, you win.

Fair enough on readability. I usually lean toward simpler solutions, but it works either way.

MGTheTrain (approved):

rdimaio (approved):

PR #8065, opened 2025-09-29
Transfers: Make automatix.rses an RSE expression #8064

### LINKED ISSUE #8064: Make automatix.rses an RSE expression
### Description

Currently, it is a comma-separated list of RSEs, which is unconventional.

### Motivation

An RSE expression is the conventional mechanism for specifying one or more RSEs. It offers an additional benefit: an RSE expression can be crafted in such a way so that newly-added RSEs can be automatically included.

### Change

1. Remove the deprecated `automatix.sites` option
2. Interpret `automatix.rses` as an RSE expression

dchristidis:
Due to this being a backwards-incompatible change, it should not be included in minor releases. Below are the recommended items for the release notes:

* The deprecated configuration option `automatix.sites` has been removed. Use `automatix.rses` instead.
* The configuration option `automatix.rses` is now parsed as an RSE expression instead of a comma-separated list of RSEs.

### DIFF
--- a/lib/rucio/daemons/automatix/automatix.py
+++ b/lib/rucio/daemons/automatix/automatix.py
@@ -146,14 +146,13 @@ def automatix(inputfile: str, sleep_time: int, once: bool = False) -> None:
 
 
 def run_once(heartbeat_handler: HeartbeatHandler, inputfile: str, **_kwargs) -> bool:
-
     _, _, logger = heartbeat_handler.live()
+
     try:
         rses = config_get_list("automatix", "rses")
     except (NoOptionError, NoSectionError, RuntimeError):
         logger(logging.ERROR, "Option rses not found in automatix section")
         return True
-
     set_metadata = config_get_bool(
         "automatix", "set_metadata", raise_exception=False, default=True
     )
@@ -162,6 +161,7 @@ def run_once(heartbeat_handler: HeartbeatHandler, inputfile: str, **_kwargs) -> 
     )
     account = config_get("automatix", "account", raise_exception=False, default="root")
     scope = config_get("automatix", "scope", raise_exception=False, default="test")
+
     client = Client(account=account)
     vo = map_vo(client.vo)  # type: ignore
     filters = {"scope": InternalScope("*", vo=vo)}
--- a/lib/rucio/daemons/automatix/automatix.py
+++ b/lib/rucio/daemons/automatix/automatix.py
@@ -153,17 +153,9 @@ def run_once(heartbeat_handler: HeartbeatHandler, inputfile: str, **_kwargs) -> 
     except (NoOptionError, NoSectionError, RuntimeError):
         logging.log(
             logging.ERROR,
-            "Option rses not found in automatix section. Trying the legacy sites option",
+            "Option rses not found in automatix section",
         )
-        try:
-            rses = config_get_list("automatix", "sites")
-            logging.log(
-                logging.WARNING,
-                "Option sites found in automatix section. This option will be deprecated soon. Please update your config to use rses.",
-            )
-        except (NoOptionError, NoSectionError, RuntimeError):
-            logger(logging.ERROR, "Could not load sites from configuration")
-            return True
+        return True
 
     set_metadata = config_get_bool(
         "automatix", "set_metadata", raise_exception=False, default=True
--- a/lib/rucio/daemons/automatix/automatix.py
+++ b/lib/rucio/daemons/automatix/automatix.py
@@ -151,10 +151,7 @@ def run_once(heartbeat_handler: HeartbeatHandler, inputfile: str, **_kwargs) -> 
     try:
         rses = config_get_list("automatix", "rses")
     except (NoOptionError, NoSectionError, RuntimeError):
-        logging.log(
-            logging.ERROR,
-            "Option rses not found in automatix section",
-        )
+        logger(logging.ERROR, "Option rses not found in automatix section")
         return True
 
     set_metadata = config_get_bool(
--- a/lib/rucio/daemons/automatix/automatix.py
+++ b/lib/rucio/daemons/automatix/automatix.py
@@ -200,7 +200,7 @@ def run_once(heartbeat_handler: HeartbeatHandler, inputfile: str, **_kwargs) -> 
         try:
             filesize = dic["filesize"]
         except KeyError:
-            filesize = 1000000
+            filesize = 1_000_000
             logger(
                 logging.WARNING,
                 "No filesize defined in the configuration, will use 1M files",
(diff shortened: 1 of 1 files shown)


mgajek-cern (commented):
## Good
- **Clean removal** of deprecated `sites` option, no half-measures
- **Better error handling** - separate exceptions for missing config vs invalid expression
- **Code simplification** - removed nested try/catch blocks
- **Minor polish** - better number formatting and clearer log message

## Bad
- **Breaking change with no migration help** - Users with `rses = RSE1,RSE2,RSE3` will get a cryptic "invalid RSE expression" error with no hint to use `RSE1|RSE2|RSE3` instead
- **Silent behavior change** - `sorted()` changes RSE processing order, may break assumptions
- **Inconsistent logger calls** - `logger.log()` vs `logger()` mixed throughout

## Should Address
**Better error message for common mistake:**
```python
except InvalidRSEExpression:
    if ',' in rse_expr and '&' not in rse_expr:
        logger.log(logging.ERROR, 
            "rses config uses comma-separated format. "
            "Update to RSE expression: 'rse1|rse2|rse3'")
```

**Documentation/examples** - Update any config examples showing the old format
**Release notes** - Add that RSE processing order may change due to `sorted()`

dchristidis:
> Breaking change with no migration help - Users with `rses = RSE1,RSE2,RSE3` will get a cryptic "invalid RSE expression" error with no hint to use `RSE1|RSE2|RSE3` instead

I think that context is important. Automatix is something configured by the instance administrators; it does not apply to users. While yes, we could add something to aid the transition, I consider it unnecessary because 1) instance administrators are supposed to check the release notes and 2) saying that it is an invalid RSE expression is enough of a clue because a comma is not an acceptable character in an RSE expression. Calling the error message cryptic is an exaggeration. Furthermore, I believe very few communities use Automatix.

> Silent behavior change - `sorted()` changes RSE processing order, may break assumptions

Some change is unavoidable, because there is no deterministic order to how the database returns the list of RSEs that match an RSE expression. But even then, given how Automatix functions, there can be absolutely no side-effect to change the order of the RSEs. Running through them in alphabetical order makes it easier when viewing the logs.

> Inconsistent logger calls - `logger.log()` vs `logger()` mixed throughout

There is no `logger.log()`. I suppose you meant `logging.log()`? The latter is necessary in some parts of the code where a `logger` object does not exist (typically outside of a thread of the main loop). This is in line with other daemons.

dchristidis:
> There is no `logger.log()`. I suppose you meant `logging.log()`?

Actually, there’s no `logging.log()` either. Is this an LLM hallucination?

mgajek-cern:
> > Breaking change with no migration help - Users with `rses = RSE1,RSE2,RSE3` will get a cryptic "invalid RSE expression" error with no hint to use `RSE1|RSE2|RSE3` instead
> 
> I think that context is important. Automatix is something configured by the instance administrators; it does not apply to users. While yes, we could add something to aid the transition, I consider it unnecessary because 1) instance administrators are supposed to check the release notes and 2) saying that it is an invalid RSE expression is enough of a clue because a comma is not an acceptable character in an RSE expression. Calling the error message cryptic is an exaggeration. Furthermore, I believe very few communities use Automatix.
> 
> > Silent behavior change - `sorted()` changes RSE processing order, may break assumptions
> 
> Some change is unavoidable, because there is no deterministic order to how the database returns the list of RSEs that match an RSE expression. But even then, given how Automatix functions, there can be absolutely no side-effect to change the order of the RSEs. Running through them in alphabetical order makes it easier when viewing the logs.
> 
> > Inconsistent logger calls - `logger.log()` vs `logger()` mixed throughout
> 
> There is no `logger.log()`. I suppose you meant `logging.log()`? The latter is necessary in some parts of the code where a `logger` object does not exist (typically outside of a thread of the main loop). This is in line with other daemons.

> > There is no `logger.log()`. I suppose you meant `logging.log()`?
> 
> Actually, there’s no `logging.log()` either. Is this an LLM hallucination?


You're right about the three points. Should have reviewed my feedback more carefully before posting. Leave further review to someone more familiar with the daemon patterns in this codebase

lib/rucio/daemons/automatix/automatix.py
```diff
@@ -146,25 +148,17 @@ def automatix(inputfile: str, sleep_time: int, once: bool = False) -> None:
 
 
 def run_once(heartbeat_handler: HeartbeatHandler, inputfile: str, **_kwargs) -> bool:
-
     _, _, logger = heartbeat_handler.live()
+
     try:
-        rses = config_get_list("automatix", "rses")
+        rse_expr = config_get("automatix", "rses")
+        rses = sorted([rse['rse'] for rse in parse_expression(rse_expr)])
```
rdimaio, line 155: Adding a note that we should also update the docs https://rucio.github.io/documentation/operator/configuration_parameters/#automatix ("rses: Separated by commas.") when this change is released

lib/rucio/daemons/automatix/automatix.py
rdimaio, line 1: Adding a comment to block merging until major release (we could merge before but avoid cherrypicking until the major release happens, but I want to check with Martin on this)

rdimaio (dismissed):
Can this logic be tested in some way? I assume it was not tested before, since the tests should break due to this change

dchristidis:
> Can this logic be tested in some way?

We probably could, I’m not qualified to say whether we should. The inner code runs in a loop for each RSE. Consequently, having the RSE expression resolve to multiple RSEs does not increase the coverage. Providing an invalid RSE expression will lead to an error, but is that worth a test?

lib/rucio/daemons/automatix/automatix.py
dchristidis: Good catch, thank you.

rdimaio:
> > Can this logic be tested in some way?
> 
> We probably could, I’m not qualified to say whether we should. The inner code runs in a loop for each RSE. Consequently, having the RSE expression resolve to multiple RSEs does not increase the coverage. Providing an invalid RSE expression will lead to an error, but is that worth a test?

I think it would still be worth testing that everything works as expected (e.g. given an expression, automatix works correctly) but I can agree that this is out of scope if automatix was not tested in the first place

dchristidis:
> I think it would still be worth testing that everything works as expected (e.g. given an expression, automatix works correctly)

That is already being tested, but with an RSE expression that resolves to a single RSE.

lib/rucio/daemons/automatix/automatix.py
```diff
@@ -146,25 +148,17 @@ def automatix(inputfile: str, sleep_time: int, once: bool = False) -> None:
...
-            rses = config_get_list("automatix", "sites")
-            logging.log(
-                logging.WARNING,
-                "Option sites found in automatix section. This option will be deprecated soon. Please update your config to use rses.",
-            )
-        except (NoOptionError, NoSectionError, RuntimeError):
-            logger(logging.ERROR, "Could not load sites from configuration")
-            return True
-
+        logger.log(logging.ERROR, "Option rses not found in automatix section")
```
bziemons: this is probably wrong! Use `logger(…)`

dchristidis: :man_facepalming:

bziemons (approved):

PR #8066, opened 2025-09-29
Documentation: Update `files` get endpoint documentation #8053

### LINKED ISSUE #8053: Wrong documentation for `dids/scope/name/files` call
https://github.com/rucio/rucio/blob/0a51bcfbf8ac5493ef139610454f60c52e11434b/lib/rucio/web/rest/flaskapi/v1/dids.py#L1074

This only checks the contents of files, not replicas. The description seems wrong.

Geogouz:
Fixes #8053

### DIFF
--- a/lib/rucio/web/rest/flaskapi/v1/dids.py
+++ b/lib/rucio/web/rest/flaskapi/v1/dids.py
@@ -1078,79 +1078,66 @@ class Files(ErrorHandlingMethodView):
     def get(self, scope_name):
         """
         ---
-        summary: Get replicas
-        description: "List all replicas for a DID."
+        summary: List DID file descriptors
+        description: >
+          Returns as newline-delimited JSON, the file-level descriptors of the specified DID.
+          For datasets and containers this expands to the child files; for file DIDs it
+          returns the metadata of the file itself.
         tags:
           - Data Identifiers
         parameters:
         - name: scope_name
           in: path
-          description: "The scope and the name of the DID."
+          required: true
+          description: "The scope and name of the DID (e.g., `scope:name`)."
           schema:
             type: string
-          style: simple
         - name: long
           in: query
-          description: "Flag to trigger long output."
-          schema:
-            type: object
           required: false
+          description: "Flag to include extended metadata such as `lumiblocknr` where available."
+          schema:
+            type: boolean
+          allowEmptyValue: true
         responses:
           200:
             description: "OK"
             content:
               application/x-json-stream:
                 schema:
-                  oneOf:
-                    - description: "All replica information if `long` is defined."
-                      type: array
-                      items:
-                        type: object
-                        properties:
-                          scope:
-                            description: "The scope of the DID."
-                            type: string
-                          name:
-                            description: "The name of the DID."
-                            type: string
-                          bytes:
-                            description: "The size of the DID in bytes."
-                            type: integer
-                          guid:
-                            description: "The guid of the DID."
-                            type: string
-                          events:
-                            description: "The number of events of the DID."
-                            type: integer
-                          adler32:
-                            description: "The adler32 checksum."
-                            type: string
-                          lumiblocknr:
-                            description: "The lumi block nr. Only available if `long` is defined in the query."
-                            type: integer
-                    - description: "All replica information."
-                      type: array
-                      items:
-                        type: object
-                        properties:
-                          scope:
-                            description: "The scope of the DID."
-                            type: string
-                          name:
-                            description: "The name of the DID."
-                            type: string
-                          bytes:
-                            description: "The size of the DID in bytes."
-                            type: integer
-                          guid:
-                            description: "The guid of the DID."
-                            type: string
-                          events:
-                            description: "The number of events of the DID."
-                            type: integer
-                          adler32:
-                            description: "The adler32 checksum."
-                            type: string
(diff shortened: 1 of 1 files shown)


mgajek-cern (dismissed):
Diff correctly fixes the misleading documentation (replicas → file descriptors) and improves clarity throughout. Changes are accurate and well-structured

lib/rucio/web/rest/flaskapi/v1/dids.py
```diff
@@ -1077,79 +1077,68 @@ class Files(ErrorHandlingMethodView):
...
           - Data Identifiers
         parameters:
         - name: scope_name
           in: path
-          description: "The scope and the name of the DID."
+          required: true
+          description: "The scope and name of the DID (e.g., `scope:name`)."
           schema:
             type: string
-          style: simple
```
voetberg, line 1091: Why was this removed?

lib/rucio/web/rest/flaskapi/v1/dids.py
```diff
@@ -1077,79 +1077,68 @@ class Files(ErrorHandlingMethodView):
...
         - name: long
           in: query
-          description: "Flag to trigger long output."
-          schema:
-            type: object
           required: false
+          description: >
+            If present, include extended metadata such as `lumiblocknr` where available.
+            (Presence is treated as true; the value is ignored.)
```
voetberg: This is sort of confusing. I would just do `Flag to include extended metadata when available` and leave the type as an object. If the type isn't actually treated as a boolean, documenting it as such just leads to confusion.

lib/rucio/web/rest/flaskapi/v1/dids.py
```diff
@@ -1077,79 +1077,68 @@ class Files(ErrorHandlingMethodView):
...
+          schema:
+            type: boolean
+          allowEmptyValue: true
         responses:
           200:
             description: "OK"
             content:
               application/x-json-stream:
                 schema:
-                  oneOf:
```
voetberg, line 1104: Having the `oneOf`is useful, I wouldn't just combine these two schema, unless there is a way to explicitly define `lumiblocknr` as an optional.

lib/rucio/web/rest/flaskapi/v1/dids.py
Geogouz: I was between removing that to avoid redundancy or adding also `explode: false` since both are the default ones for `in: path`. In practice, there is no change whether we keep it or not so I'm fine with both. We should maybe define somewhere guidelines/rules regarding the writing of OpenAPI docs to be consistent (in many places we use this, while in others we don't).

lib/rucio/web/rest/flaskapi/v1/dids.py
Geogouz: I used "Only returned when `long` is requested and the information exists." for this, but indeed the previous approach might be better. I can switch it back.

rdimaio: I agree that this is a bit confusing - does it mean that whether `long` is `False` or `True`, the value itself is ignored and it's treated as `True` as long as it's passed? Out of scope for this docstring change, but why is it like this?

lib/rucio/web/rest/flaskapi/v1/dids.py
rdimaio: Personally I like the new version more, I think it's easier to read when most properties are not duplicated and we just mention that the `lumiblocknr` is provided when `long` is requested

Geogouz: I used "Only returned when long is requested and the information exists." for this, but indeed the previous approach might be better. I can switch it back.

lib/rucio/web/rest/flaskapi/v1/dids.py
Geogouz: Yes, any request such as `...?long=false` or `...?long=0` still makes the condition true because the parameter is present. We have to omit long if we want to keep it false.

rdimaio: Should we change this? I think it's a bit confusing

bari12: I would also describe it as Maggie wrote.

Geogouz: Switched for now to the use of "Flag to include ..", but I would expect that this will (or already has) lead to people passing `?long=true` (which is meaningless). Going through the rest of our endpoints, this `long` parameter is actually the only case (aside from `dynamic_depth` and `dynamic` in `/dids/{scope}/{name}`) where we treat it as a way of "presence". So it might be worth changing it in the future as @rdimaio mentioned.

Note: Given our current implementation, `long` affects only the presence of `lumiblocknr`. So I think mentioning it in the description removes early any ambiguity or "falsy" expectations.

lib/rucio/web/rest/flaskapi/v1/dids.py
Geogouz: We need to consider here that if we use "oneOf", we violate the JSON Schema and Swagger specs which require a "oneOf" instance to never match more than one subschema. And in our case, a “short” file object (long not set) is valid against both subschemas. I think we should stick to the current approach.

bari12 (approved):

PR #8069, opened 2025-09-29
Clients: optimise rucio upload #7968

### LINKED ISSUE #7968: Optimise rucio upload
### Description

`rucio upload` is currently quite a heavyweight operation involving multiple different calls to the Rucio server. On DUNE's production server this frequently causes exceptions when many files are being uploaded simultaneously and the server is under heavy load.

We have identified that the most expensive part of the upload operation is usually the `list_replicas` call. This call is also responsible for the longest running queries observed on DUNE's database server. However, in `rucio upload` this is used only to determine whether or not a replica of the file already exists on the target RSE before creating one - all of the other information returned by this call is thrown away. It should therefore be possible to speed up the upload operation by providing a lighter weight endpoint that simply checks for the existence of a certain replica without retrieving all the other information normally returned by `list_replicas`.

### Motivation

_No response_

### Change

_No response_

jamesp-epcc:
The purpose of this work is to optimise the upload client by reducing the number of calls made to the server. Specifically:

- `list_rses` is only called if the RSE to be used is not already known
- RSE attribute are cached so that they're only retrieved once per RSE rather than once per file
- a new `replica_exists` endpoint is used instead of the much more expensive `list_replicas` call that was previously used to check replica existence

It would be possible to save another call by removing the warning message that is printed if the scope is not found for the account, but I have removed this change as I cannot be certain there isn't a good reason for this warning.

### DIFF
--- a/lib/rucio/client/replicaclient.py
+++ b/lib/rucio/client/replicaclient.py
@@ -153,6 +153,29 @@ class ReplicaClient(BaseClient):
         exc_cls, exc_msg = self._get_exception(headers=r.headers, status_code=r.status_code, data=r.content)
         raise exc_cls(exc_msg)
 
+    def replica_exists(self, scope: str, name: str, rse: str) -> bool:
+        """
+        Checks whether a replica of the specified file exists on the specified RSE.
+
+        Parameters
+        ----------
+        scope :
+            The scope of the file.
+        name :
+            The name of the file.
+        rse :
+            The RSE name.
+        """
+        payload = {'rse': rse}
+        url = build_url(self.host, path='/'.join([self.REPLICAS_BASEURL, quote_plus(scope), quote_plus(name), 'exists']), params=payload)
+        r = self._send_request(url, method=HTTPMethod.GET)
+        if r.status_code == codes.ok:
+            return True
+        elif r.status_code == codes.not_found:
+            return False
+        exc_cls, exc_msg = self._get_exception(headers=r.headers, status_code=r.status_code, data=r.content)
+        raise exc_cls(exc_msg)
+
     def list_replicas(self, dids, schemes=None, ignore_availability=True,
                       all_states=False, metalink=False, rse_expression=None,
                       client_location=None, sort=None, domain=None,
--- a/lib/rucio/client/replicaclient.py
+++ b/lib/rucio/client/replicaclient.py
@@ -153,6 +153,27 @@ class ReplicaClient(BaseClient):
         exc_cls, exc_msg = self._get_exception(headers=r.headers, status_code=r.status_code, data=r.content)
         raise exc_cls(exc_msg)
 
+    def replica_exists(self, scope: str, name: str, rse: str) -> bool:
+        """
+        Checks whether a replica of the specified file exists on the specified RSE.
+
+        Parameters
+        ----------
+        scope :
+            The scope of the file.
+        name :
+            The name of the file.
+        rse :
+            The RSE name.
+        """
+        payload = {'rse': rse}
+        url = build_url(self.host, path='/'.join([self.REPLICAS_BASEURL, quote_plus(scope), quote_plus(name), 'exists']), params=payload)
+        r = self._send_request(url, method=HTTPMethod.GET)
+        if r.status_code == codes.ok:
+            return r.text == 'True'
+        exc_cls, exc_msg = self._get_exception(headers=r.headers, status_code=r.status_code, data=r.content)
+        raise exc_cls(exc_msg)
+
     def list_replicas(self, dids, schemes=None, ignore_availability=True,
                       all_states=False, metalink=False, rse_expression=None,
                       client_location=None, sort=None, domain=None,
--- a/lib/rucio/client/uploadclient.py
+++ b/lib/rucio/client/uploadclient.py
@@ -43,7 +43,6 @@ from rucio.common.exception import (
     RSEOperationNotSupported,
     RSEWriteBlocked,
     RucioException,
-    ScopeNotFound,
     ServiceUnavailable,
 )
 from rucio.common.utils import execute, generate_uuid, make_valid_did, retry, send_trace
@@ -353,7 +352,10 @@ class UploadClient:
         rse_expression = None
         for file in files:
             rse_expression = file['rse']
-            rse = self.rse_expressions.setdefault(rse_expression, _pick_random_rse(rse_expression))
+
+            if rse_expression not in self.rse_expressions:
+                self.rse_expressions[rse_expression] = _pick_random_rse(rse_expression)
+            rse = self.rse_expressions[rse_expression]
 
             if not self.rses.get(rse):
                 rse_settings = self.rses.setdefault(rse, rsemgr.get_rse_info(rse, vo=self.client.vo))
@@ -378,6 +380,7 @@ class UploadClient:
         registered_dataset_dids = set()
         num_succeeded = 0
         summary = []
+        rse_attribute_cache = {}
         for file in files:
             basename = file['basename']
             logger(logging.INFO, 'Preparing upload for file %s' % basename)
@@ -416,11 +419,15 @@ class UploadClient:
 
             # resolving local area networks
             domain = 'wan'
-            rse_attributes = {}
-            try:
-                rse_attributes = self.client.list_rse_attributes(rse)
-            except Exception:
-                logger(logging.WARNING, 'Attributes of the RSE: %s not available.' % rse)
+            if rse in rse_attribute_cache:
+                rse_attributes = rse_attribute_cache[rse]
+            else:
+                rse_attributes = {}
+                try:
+                    rse_attributes = self.client.list_rse_attributes(rse)
+                except Exception:
+                    logger(logging.WARNING, 'Attributes of the RSE: %s not available.' % rse)
+                rse_attribute_cache[rse] = rse_attributes
             if self.client_location and 'lan' in rse_settings['domain'] and RseAttr.SITE in rse_attributes:
                 if self.client_location['site'] == rse_attributes[RseAttr.SITE]:
                     domain = 'lan'
@@ -688,16 +695,6 @@ class UploadClient:
         logger = self.logger
         logger(logging.DEBUG, 'Registering file')
 
-        # verification whether the scope exists
-        account_scopes = []
-        try:
-            account_scopes = self.client.list_scopes_for_account(self.client.account)
-        except ScopeNotFound:
-            pass
-        if account_scopes and file['did_scope'] not in account_scopes:
-            logger(logging.WARNING,
-                   'Scope {} not found for the account {}.'.format(file['did_scope'], self.client.account))
-
         rse = file['rse']
         dataset_did_str = file.get('dataset_did_str')
         # register a dataset if we need to
@@ -738,8 +735,7 @@ class UploadClient:
                 raise DataIdentifierAlreadyExists
 
             # add the file to rse if it is not registered yet
-            replicastate = list(self.client.list_replicas([file_did], all_states=True))
-            if rse not in replicastate[0]['rses']:
+            if not self.client.replica_exists(file_scope, file_name, rse):
                 self.client.add_replicas(rse=rse, files=[replica_for_api])
                 logger(logging.INFO, 'Successfully added replica in Rucio catalogue at %s' % rse)
         except DataIdentifierNotFound:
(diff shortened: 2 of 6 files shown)


lib/rucio/client/uploadclient.py
```diff
@@ -726,23 +733,28 @@ def _register_file(
...
+                # add the file to rse if it is not registered yet
+                replicastate = list(self.client.list_replicas([file_did], all_states=True))
+                if rse not in replicastate[0]['rses']:
+                    self.client.add_replicas(rse=rse, files=[replica_for_api])
+                    logger(logging.INFO, 'Successfully added replica in Rucio catalogue at %s' % rse)
+            except DataIdentifierNotFound:
+                new_file = True
+
+        if new_file:
             logger(logging.DEBUG, 'File DID does not exist')
```
mgajek-cern, line 753: 1. As a minimal safety mechanism, should there be a warning log for the user instead:

```suggestion
      logger(logging.WARNING, 
           f'--new-file flag used: Skipping validation for {file_scope}:{file_name}. '
           f'If this DID already exists, data corruption will occur.')
```

lib/rucio/cli/upload.py
```diff
@@ -35,8 +35,9 @@
 @click.option("--lfn", help="Specify the exact LFN for the upload")
 @click.option("--transfer-timeout", type=float, default=config_get_float("upload", "transfer_timeout", False, 360), help="Transfer timeout (in seconds)")
 @click.option("-r", "--recursive", is_flag=True, default=False, help="Convert recursively the folder structure into collections")
+@click.option("--new-file", is_flag=True, default=False, help="Assume DID does not already exist")
 @click.pass_context
-def upload_command(ctx, file_paths, rse, lifetime, expiration_date, scope, impl, no_register, register_after_upload, summary, guid, protocol, pfn, lfn, transfer_timeout, recursive):
+def upload_command(ctx, file_paths, rse, lifetime, expiration_date, scope, impl, no_register, register_after_upload, summary, guid, protocol, pfn, lfn, transfer_timeout, recursive, new_file: bool):
     """Upload file(s) to a Rucio RSE"""
     args = Arguments(
```
mgajek-cern, line 41: 2. Another safety mechansim: one could consider an Interactive confirmation, e.g.

```suggestion
    # Safety check for --new-file flag
    if new_file and not ctx.obj.no_pager:  # Interactive session check
        click.confirm(
            'WARNING: --new-file skips existence and checksum validation. '
            'Using this on already-registered files will cause silent data corruption. '
            'Are you sure you want to continue?',
            abort=True
        )
    
    args = Arguments(
```

Better then 1.

mgajek-cern (commented):
The performance optimizations (RSE attribute caching) are solid. However, the `--new-file` flag introduces a safety concern by skipping both existence and checksum validation. If misused on existing files, this will cause silent data corruption.

Suggested safety mechanisms:

1. Warning log - Alert users when --new-file skips validation checks
2. Interactive confirmation - Require explicit user confirmation in terminal sessions before proceeding with unsafe operations
3. Lightweight API endpoint - Create a dedicated replica_exists(scope, name, rse) endpoint that returns True/False without the overhead of full list_replicas. This preserves most performance gains while maintaining data integrity.

Option 3 is the proper long-term solution. Options 1-2 are minimal safeguards that should be added regardless.

voetberg:
> The performance optimizations (RSE attribute caching) are solid. However, the `--new-file` flag introduces a safety concern by skipping both existence and checksum validation. If misused on existing files, this will cause silent data corruption.
> 
> Suggested safety mechanisms:
> 
>     1. Warning log - Alert users when --new-file skips validation checks
> 
>     2. Interactive confirmation - Require explicit user confirmation in terminal sessions before proceeding with unsafe operations
> 
>     3. Lightweight API endpoint - Create a dedicated replica_exists(scope, name, rse) endpoint that returns True/False without the overhead of full list_replicas. This preserves most performance gains while maintaining data integrity.
> 
> 
> Option 3 is the proper long-term solution. Options 1-2 are minimal safeguards that should be added regardless.

Hi Marvin, 

You are missing some context on this PR. Right now it's a draft because it's being tested for DUNE, this should be used exclusively uploading in bulk and by machine. Sure there are some optimizations to be made for safety but it's not ready for that. I'd hold off on reviewing it.

StevenCTimm:
It seems that what was once an unhandled exception when the flaskapi couldn't read from the http stream is now handled.
In testing rucio v38 we are now seeing 
[wsgi:error] [pid 45:tid 145] [client 100.64.0.7:44662] Truncated or oversized response headers received from daemon process 'rucio': /usr/local/lib/python3.9/site-packages/rucio/web/rest/flaskapi/v1/main.py

where once there was an unhandled exception.
Was that fixed elsewhere in the course of rucio v38 or is it a result of this PR somehow?  The above error happens only when the rucio server pod is under extreme load, 60 simultaneous rucio uploads or so.

jamesp-epcc:
> Was that fixed elsewhere in the course of rucio v38 or is it a result of this PR somehow? The above error happens only when the rucio server pod is under extreme load, 60 simultaneous rucio uploads or so.

It wasn't directly fixed by this PR, in the sense that I didn't add any code to handle those exceptions. It could be due to another change in v38, or it could be because the optimisations introduced here are reducing the server load and causing the issues to manifest in a different way from before.

StevenCTimm:
Indeed..request.py which is where the exception was happening, on line 422, remains unchanged and no new handlers have been added.  Perhaps we will have to push the single pod a bit harder to be able to reproduce the problem now.

jamesp-epcc:
I've made a change so that instead of the `--new-file` switch (which as Marvin pointed out is potentially dangerous), a much lighter weight `replica_exists` endpoint is called, removing the potentially expensive `list_replicas` call.

StevenCTimm:
James, could you please prepare a version of the patches against rucio v38_3, server and client and then please let Dennis Lee know where it is and where to get it?  Then we will install on integration and test appropriately.

jamesp-epcc:
Apologies Steve, I'm just back from holiday today. I have rebased my branch against 38.3.0 so the changes should apply cleanly to that version. Dennis could get it directly from the branch here: https://github.com/jamesp-epcc/rucio/tree/feature-7968-optimise-upload but if he would prefer a different format such as patch files, just let me know and I can prepare them.

jamesp-epcc:
Actually, please use this branch for the v38.3 version: https://github.com/jamesp-epcc/rucio/tree/upload_v38_3
Trying to keep the PR clean by keeping this branch based against master.
I should also mention that the `--new-file` option is now gone. I replaced it with a safer optimisation that doesn't need to be told whether the file is new.

jamesp-epcc:
I've removed the draft status from this as I think the optimisations currently contained here should be safe for production. I've also added a test for the newly added `replica_exists` call.

lib/rucio/web/rest/flaskapi/v1/replicas.py
```diff
@@ -1829,6 +1832,52 @@ def post(self):
...
+            rse = request.args.get('rse', default='')
+            if not replica_exists(
+                    did=did,
+                    rse=rse,
+                    issuer=request.environ['issuer'],
+                    vo=request.environ['vo']):
+                raise ReplicaNotFound()
+        except ValueError as error:
+            return generate_http_error_flask(400, error)
+        except (RSENotFound, ReplicaNotFound) as error:
```
voetberg: I'm not sure a 404 is the right error here. Even if it doesn't exist, the advertised purpose of the endpoint is to confirm if it doesn't exist, so it's not an error if it doesn't. Would it make more sense to return a 404 if the RSE doesn't exist, and a false if the replica doesn't?

tests/test_replica.py
```diff
@@ -716,6 +731,23 @@ def test_list_replicas_rse_filter(self, rse_factory, mock_scope, root_account, c
         [replica] = list(list_replicas([did2], rse_expression='group2=true'))
         assert len(replica['pfns']) == 1
 
+    def test_replica_exists(self, rse_factory, mock_scope, root_account):
```
voetberg, line 737: The rest API and client functionality also needs tests

lib/rucio/core/replica.py
```diff
@@ -1189,6 +1189,33 @@ def _list_replicas(
...
+        session: "Session"
+) -> bool:
+    """
+    Returns true if a replica of the given file exists on the given RSE.
+    :param scope: The scope of the file.
+    :param name: The name of the file.
+    :param rse_id: The ID of the RSE to check.
+    :param session: The database session in use.
+    """
+    stmt = select(
```
voetberg, line 1201: I don't know which would be more efficient but my first inclination that it's cleaner to just get the count and and make sure it's not 0. 

```suggestion
    stmt = func.count().select_from(
        models.RSEFileAssociation
    ).where(
        and_(models.RSEFileAssociation.rse_id == rse_id,
             models.RSEFileAssociation.scope == scope,
             models.RSEFileAssociation.name == name)
    )
    return session.execute(stmt) != 0
```

voetberg:
@jamesp-epcc Would you mind breaking this PR up into a few commits (tests, new features in clients, rest, and core, and the changes to the download client?), and including tests for the different components for the new replica_exists route?

jamesp-epcc:
> @jamesp-epcc Would you mind breaking this PR up into a few commits (tests, new features in clients, rest, and core, and the changes to the download client?), and including tests for the different components for the new replica_exists route?

Sure, I can do that. Apologies for the late reply, I have been away for a couple of weeks.

jamesp-epcc:
I've made the changes Maggie suggested and added a test for the client and API layers of the new endpoint.

lib/rucio/client/uploadclient.py
```diff
@@ -688,16 +695,6 @@ def _register_file(
         logger = self.logger
         logger(logging.DEBUG, 'Registering file')
 
-        # verification whether the scope exists
```
voetberg, line 691: To make sure I understand, this stuff is totally superseded by the `replica exists` call right? I do worry a bit about removing the warning here.

voetberg:
This is hyper-pedantic, but for the sake of keeping all changes documented can you add some extra description to https://github.com/rucio/rucio/pull/8069/changes/760ba441635f0e527f88cf34a97dc17ca90b9956, including the removal of the scope exists message, introduction of replica_exists as a check, and RSE attribute cache? Additionally, it may be a good idea to change the PR description to match how this PR has changed. 

I don't see any immediate downstream problems though, these seem like solid changes to me.

lib/rucio/client/uploadclient.py
jamesp-epcc: Having looked again at this, I'm not 100% certain it's a good idea to remove this. It seems relatively harmless to me but it must have been put in for a reason, so maybe there are cases where it's important. So for the moment I have removed this part of the change.

jamesp-epcc:
I've improved the description for the final commit and for the PR itself. As mentioned above I've also reverted the removal of that warning about the scope because after looking again, I can't be sure this is a good idea.

lib/rucio/client/replicaclient.py
```diff
@@ -153,6 +153,27 @@ def get_did_from_pfns(self, pfns, rse=None):
...
+        ----------
+        scope :
+            The scope of the file.
+        name :
+            The name of the file.
+        rse :
+            The RSE name.
+        """
+        payload = {'rse': rse}
+        url = build_url(self.host, path='/'.join([self.REPLICAS_BASEURL, quote_plus(scope), quote_plus(name), 'exists']), params=payload)
```
voetberg, line 170: One merger note - because this adds a new endpoint this breaks client and server backwards compatibility. I wonder if there is some sort of workaround here we can try to make sure this isn't a problem....

jamesp-epcc: That's true. Old clients should continue to work with the updated server but the updated client will no longer work with old servers. I don't know if there is a standard procedure for changes like this.

voetberg: I do not know if it is standardized, but what I tend to do is include a try-except loop with the new change, pointing to the old implementation as a fallback until the change is included in all server line releases. We can then make a separate issue to remove the compatibility and include a link to it as a comment in the code. In pseudo-code: 

```python

try: 
     replica_exists = self.client.replica_exists(...)
except WhateverEndpointNotFoundErrorWeUse:   # TODO: Remove via issue #12345678
     replica_exists = replica in list(self.client.list_replicas(...))
     
 [continue as normal]

```

jamesp-epcc: That seems sensible. I will implement this after I get back from my break.

PR #8076, opened 2025-09-30
Remove unused request gateway functions

### LINKED ISSUE #7625: Unused `request` gateway functions
Similarly to https://github.com/rucio/rucio/issues/7624

https://github.com/rucio/rucio/blob/master/lib/rucio/gateway/request.py

The following gateway functions are unused:
- `queue_requests`
- `cancel_request`
- `cancel_request_did`
- `get_next`

We should determine whether these should be:

- put in use, by adding API routes that call these functions, or
- removed

voetberg:
Closes #7625

### DIFF
--- a/tests/test_gateway_external_representation.py
+++ b/tests/test_gateway_external_representation.py
@@ -23,8 +23,9 @@ import rucio.core.account_counter as account_counter
 import rucio.gateway.account_limit as gateway_acc_lim
 import rucio.gateway.rse as gateway_rse
 from rucio.common.config import config_get_bool
-from rucio.common.types import InternalScope
+from rucio.common.types import InternalAccount, InternalScope
 from rucio.common.utils import gateway_update_return_dict, generate_uuid
+from rucio.core.request import queue_requests
 from rucio.core.vo import add_vo, vo_exists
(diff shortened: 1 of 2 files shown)


tests/test_gateway_external_representation.py
```diff
@@ -309,17 +310,13 @@ def test_gateway_request(self, vo, rse_factory, account_name, scope_name):
...
 
-        reqs = queue_requests(requests, issuer='root', vo=vo)  # this does not pass in the source rse
+        reqs = queue_requests(requests)  # this does not pass in the source rse
         reqs = list(reqs)
         assert 0 != len(reqs)
-        for r in reqs:
-            assert r['scope'] == scope_name
-            assert r['account'] == account_name
-            assert r['source_rse'] == rse1
-            assert r['dest_rse'] == rse2
```
rdimaio, line 322: Could you clarify why this is no longer tested?

voetberg: This isn't functionality of the core queue, which is now what's being used. There's no representational requirements for it either, so it makes more sense to remove this requirement than to change core. 

We still need to include the queue_messages call to test the other request functionality.

rdimaio (approved):

PR #8082, opened 2025-10-02
Support for Open Data record IDs

### LINKED ISSUE #8080: Support Open Data record ids
After some discussions, we decided to add support for Open Data record id, which is used in CERN Open Data Portal as a unique identifier of the Open Data DID (similar to DOI, but record ID is present from the start).

The record ID should be optional but globally unique at the Rucio DB. Right now the only way to achieve this is to store the record id in a distinct table similar to the DOI.

While this feature is motivated by the structure of the CERN Open Data Portal, it's general enough to be used by other Open Data indexers: we can expect indexers to have a unique numeric id for a record.

lobis:
This PR adds support for Records IDs in a similar way to how DOI is implemented.

Record IDs are an optional attribute that must be globally unique, in order to implement this in the DB, we must use a dedicated table.

### DIFF
--- a/lib/rucio/client/opendataclient.py
+++ b/lib/rucio/client/opendataclient.py
@@ -163,6 +163,7 @@ class OpenDataClient(BaseClient):
             state: Optional["OPENDATA_DID_STATE_LITERAL"] = None,
             meta: Optional[dict] = None,
             doi: Optional[str] = None,
+            record_id: Optional[int] = None,
     ) -> dict[str, Any]:
         """
         Update an existing Opendata DID in the Opendata catalog.
@@ -172,7 +173,8 @@ class OpenDataClient(BaseClient):
             name: The name of the DID.
             state: The new state to set for the DID.
             meta: Metadata to update for the DID. Must be a valid JSON object.
-            doi: DOI to associate with the DID. Must be a valid DOI string (e.g., "10.1234/foo.bar").
+            doi: DOI to associate with the DID. Must be a valid DOI string (e.g., "10.1234/foo.bar") and unique across all DIDs.
+            record_id: The record ID of the DID to update. This can be used to cross-reference with external systems. Must be unique across all DIDs.
 
         Returns:
             True if the update was successful.
@@ -185,8 +187,8 @@ class OpenDataClient(BaseClient):
         path = '/'.join([self.opendata_private_dids_base_url, quote_plus(scope), quote_plus(name)])
         url = build_url(self.get_opendata_host(public=False), path=path)
 
-        if not any([meta, state, doi]):
-            raise ValueError("Either 'meta', 'state', or 'doi' must be provided.")
+        if not any([meta, state, doi, record_id]):
+            raise ValueError("Either 'meta', 'state', 'doi' or 'record_id' must be provided.")
 
         data: dict[str, Any] = {}
 
@@ -198,6 +200,9 @@ class OpenDataClient(BaseClient):
 
         if doi is not None:
             data['doi'] = doi
+
+        if record_id is not None:
+            data['record_id'] = record_id
 
         r = self._send_request(url, method=HTTPMethod.PUT, data=render_json(**data))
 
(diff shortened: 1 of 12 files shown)


lobis:
(I cloesd the PR by mistake).

rdimaio (commented):
Please fix merge conflicts

lib/rucio/client/opendataclient.py
```diff
@@ -172,7 +173,8 @@ def update_opendata_did(
             name: The name of the DID.
             state: The new state to set for the DID.
             meta: Metadata to update for the DID. Must be a valid JSON object.
-            doi: DOI to associate with the DID. Must be a valid DOI string (e.g., "10.1234/foo.bar").
+            doi: DOI to associate with the DID. Must be a valid DOI string (e.g., "10.1234/foo.bar") and unique across all DIDs.
+            record_id: The record ID of the DID to update. This can be used to cross-reference with external systems. Must be unique across all DIDs.
```
rdimaio, line 177: Do you have a reference for this Open Data record ID concept? What's the difference between this and a DOI (I'm assuming all OpenData DID should have a unique DOI anyway?)

lobis: The Open Data record if is a generic unique numerical id associated to the entry.

It comes from the CERN Open Data portal where each record has a unique numeric id (e.g. https://opendata.cern.ch/record/160000). Each record also has a unique DOI but crucially the record id is generated at the start.

The feature comes to have a better integration with the cern open data portal but I think it should be generic enough to also integrate with other portals and of course its fully optional.

The main feature is that when a DID is tagged as open data, it is also inmediatelly given a record id, the same as in the open data portal, to then create a permanent relationship with the portal, otherwise this would be significantly hard since you cannot reverse search for did in the portal.

lib/rucio/db/sqla/migrate_repo/versions/a7e76cf4881d_opendata_record.py
```diff
@@ -0,0 +1,46 @@
...
+revision = 'a7e76cf4881d'
+down_revision = 'a62db546a1f1'
+
+
+def upgrade():
+    op.create_table(
+        'dids_opendata_record',
+        sa.Column('scope', sa.String(length=get_schema_value('SCOPE_LENGTH')), nullable=False),
+        sa.Column('name', sa.String(length=get_schema_value('NAME_LENGTH')), nullable=False),
+        sa.Column('record_id', sa.Integer(), nullable=False, unique=True),
```
rdimaio, line 32: is there no max length for record IDs?

tests/test_opendata.py
```diff
@@ -266,6 +266,67 @@ def test_opendata_doi_duplicate(self, mock_scope, root_account, doi_factory, db_
...
+                    session=db_write_session)
+            opendata.add_opendata_did(scope=mock_scope, name=name, session=db_write_session)
+
+        record_id = 1337
+
+        opendata.update_opendata_did(scope=mock_scope, name=name_first, record_id=record_id, session=db_write_session)
+
+        db_write_session.commit()
+
+        # opendata.update_opendata_did(scope=mock_scope, name=name_second, record_id=record_id, session=db_write_session)
```
rdimaio: nit: we should remove commented out code if it's not needed

lib/rucio/db/sqla/migrate_repo/versions/a7e76cf4881d_opendata_record.py
lobis: The idea is that this parameter matches an index on a third party system so I think limiting the range of values can be counterproductive. I think a 32-bit integer is enough to cover all systems (`sa.Integer()` is a 32-bit integer I believe).

Is there anything to gain from setting a max length that I may be missing?

rdimaio: Not to my knowledge, was just wondering

rdimaio (approved):

PR #8085, opened 2025-10-02
Fix: Convert tuple to string for file path parameters in replica CLI

ArturAkh:
When using --lfn, --as-file flags with 'rucio replica state update' commands, the replicas argument (captured by nargs=-1) is a tuple. However, the underlying functions expect a string file path to open.

This fix extracts the first element from the tuple when file-based flags are used (--lfn, --as-file) for the following commands:
- replica state update bad
- replica state update unavailable
- replica state update quarantine

Fixes the error: 'expected str, bytes or os.PathLike object, not tuple' that occurred when using these flags.

### DIFF
--- a/lib/rucio/cli/replica.py
+++ b/lib/rucio/cli/replica.py
@@ -129,11 +129,13 @@ def update_bad(ctx, replicas, reason, as_file, collection, lfn, scope, rse):
     """Mark a replica bad"""
     args = {"reason": reason, "allow_collection": collection, "scope": scope, "rse": rse}
     if as_file:
-        args["inputfile"] = replicas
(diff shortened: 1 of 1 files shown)


voetberg:
Hi! Thanks for the fix here. Before you submit this PR would you mind filing an issue for it and following the [contributor guidelines](https://rucio.cern.ch/documentation/contributing/)?

ArturAkh:
Issue created: https://github.com/rucio/rucio/issues/8099

ArturAkh:
Created new PR more in line with guidelines:

https://github.com/rucio/rucio/pull/8100

PR #8089, opened 2025-10-03
Core: Add new Open Data exception for duplicate DOI. #8088

### LINKED ISSUE #8088: Improve error handling on duplicate Open Data properties
Duplicate DOI should raise a better exception

lobis:
Closes #8088

### DIFF
--- a/lib/rucio/core/opendata.py
+++ b/lib/rucio/core/opendata.py
@@ -740,5 +740,18 @@ def update_opendata_doi(
         if result.rowcount == 0:
             raise ValueError(f"Error updating Opendata DOI for DID '{scope}:{name}'.")
 
+    except IntegrityError as error:
+        msg = str(error)
+
+        if (
+                search(r'ORA-00001: unique constraint \([^)]+\) violated', msg)
+                or search(r'UNIQUE constraint failed: dids_opendata_doi\.doi', msg)
+                or search(r'1062.*Duplicate entry.*for key', msg)
+                or search(r'duplicate key value violates unique constraint', msg)
+                or search(r'columns?.*not unique', msg)
+        ):
+            raise exception.OpenDataDuplicateDOI(doi=doi)
+
+        raise exception.OpenDataError()
     except DataError as error:
         raise exception.InputValidationError(f"Invalid data: {error}")
--- a/lib/rucio/common/exception.py
+++ b/lib/rucio/common/exception.py
@@ -1281,3 +1281,13 @@ class InvalidAccountType(RucioException):
         super(InvalidAccountType, self).__init__(*args)
         self._message = "Cannot create an account with an invalid type."
         self.error_code = 121
+
+class OpenDataDuplicateDOI(OpenDataError):
+    """
+    Throws when a data identifier with the same DOI already exists in the open data catalog.
+    """
+
+    def __init__(self, doi: str, *args):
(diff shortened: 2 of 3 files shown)


lobis:
~~Also used this PR to mark `test_judge_double_rule_on_container` test as flaky as I noticed the CI often fails because of this.~~

Edit: placed it here https://github.com/rucio/rucio/pull/8117

tests/test_judge_evaluator.py
voetberg, line 1: I know it's annoying, but for cleanliness this should probably be a different PR

lib/rucio/core/opendata.py
```diff
@@ -867,7 +740,18 @@ def update_opendata_doi(
         if result.rowcount == 0:
             raise ValueError(f"Error updating Opendata DOI for DID '{scope}:{name}'.")
 
+    except IntegrityError as error:
```
voetberg, line 743: What specifically raises these different DB exceptions? Each one of those situations should be tested if we're going to make it a specific error instead of just a OpenDataError.

lobis: Sorry but not sure I follow. 

These different exceptions are how a "duplicate entry" is raised in different dbs, and this is tested in the new test present in this PR.

If we get an integrite error but it does not match, we get a general `OpenDataError`.

tests/test_judge_evaluator.py
lobis: Placed it here: https://github.com/rucio/rucio/pull/8117

voetberg (approved):

lib/rucio/core/opendata.py
```diff
@@ -740,5 +740,18 @@ def update_opendata_doi(
...
+        msg = str(error)
+
+        if (
+                search(r'ORA-00001: unique constraint \([^)]+\) violated', msg)
+                or search(r'UNIQUE constraint failed: dids_opendata_doi\.doi', msg)
+                or search(r'1062.*Duplicate entry.*for key', msg)
+                or search(r'duplicate key value violates unique constraint', msg)
+                or search(r'columns?.*not unique', msg)
```
rdimaio, line 751: nit: I wonder if a single `search` with an `|` operator (https://stackoverflow.com/questions/28446090/using-or-operator-in-re-search-for-multiple-strings-in-python-3) would be faster than having separate `search` statements? Leaving this comment unresolved in case you wanna check, but otherwise feel free to resolve it and we can merge.

rdimaio (approved):

lib/rucio/core/opendata.py
lobis: I am not a fan of this pattern in general, I think it can be improved overall.

I am thinking of something along the lines of:

```
patterns = [
    r'ORA-00001: unique constraint \([^)]+\) violated',
    r'UNIQUE constraint failed:',
    r'1062.*Duplicate entry.*for key',
    r'duplicate key value violates unique constraint',
    r'columns?.*not unique',
]

if any(re.search(p, msg, re.IGNORECASE) for p in patterns):
    # handle duplicate key error
    ...
```

where `patters` are reused across similar exception checkings (`duplicate_patterns, etc.).

For this PR I would prefer to just merge it as is and I will create a new issue to track this refactoring.

https://github.com/rucio/rucio/issues/8118

PR #8091, opened 2025-10-03
Metadata: Feature add jsonschema validation postgres jsonb; Fix rucio#8036

### LINKED ISSUE #8036: Add jsonschema validation option in postgres meta plugin
### Description

I would be great to have jsonschema validation option for postgres jsonb meta plugins.
Needed in insert and update operation

### Motivation

_No response_

### Change

_No response_

panta-123:
Add optional json schema validation on set_metadata_bulk of postgres_meta jsonb plugin.

- schema is applied globally regardless of did type if turned on
- Can be turned on as using `postgres_json_schema_path`:
```
[metadata]
plugins = rucio.core.did_meta_plugins.postgres_meta.ExternalPostgresJSONDidMeta
.......
........
postgres_json_schema_path=/opt/rucio/etc/schema.json
```
- added a test for this logic

### DIFF
--- a/lib/rucio/core/did_meta_plugins/postgres_meta.py
+++ b/lib/rucio/core/did_meta_plugins/postgres_meta.py
@@ -12,12 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import os
 import json
 import operator
-import jsonschema
+import os
 from typing import TYPE_CHECKING, Any, Optional
 
+import jsonschema
 import psycopg
 from psycopg import sql
 from psycopg.rows import dict_row
@@ -28,8 +28,6 @@ from rucio.core.did_meta_plugins.did_meta_plugin_interface import DidMetaPlugin
 from rucio.core.did_meta_plugins.filter_engine import FilterEngine
 
 if TYPE_CHECKING:
-    from typing import Optional
-
     from sqlalchemy.orm import Session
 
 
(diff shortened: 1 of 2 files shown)


Geogouz:
I see several problems here. For example, validation is silently disabled when configuration is wrong, or passing a `json_schema` argument to `ExternalPostgresJSONDidMeta` never even sets `self.metadata_schema` (so `_validate_metadata` will raise `AttributeError` on the first metadata write). Also, not having detected such problems through our tests suggests that these are also not sufficient. But although these can be addressed, my main concern is that this is not going towards the best direction for a solution to the underlying issue.

I think that a proper metadata validation engine in Rucio (I will introduce one either on this or the following release) should be shared across all metadata backends rather than restricted to PostgreSQL. Among others, it should ideally:

- Support per-DID schema identifiers and versions, storing the schema reference alongside metadata so validation and migrations stay consistent regardless of storage technology.
- Enforce validation on every metadata write path (`set_metadata`, `set_metadata_bulk`, migrations) and fail fast or log critically when configuration prevents validation, instead of silently bypassing checks.
- Offer migration hooks and tooling to revalidate and upgrade stored metadata when schemas evolve, using backend-agnostic flows to keep behavior uniform.

My suggestion would be to wait for the next release.

panta-123:
@Geogouz ,  Will wait for new implmetation in next release.  If not in next release, we will be forced to use local patches which we want to avoid.

bziemons:
> @Geogouz , Will wait for new implmetation in next release. If not in next release, we will be forced to use local patches which we want to avoid.

then please close the PR yourself until you're actively working on this again and it is ready to review.

PR #8096, opened 2025-10-05
Metadata: fix delete metadata in postgres_json plugin Fix #8090

### LINKED ISSUE #8090: PostgreSQL DID metadata deletion removes key from every row
### Description

@robbarnsley in [`ExternalPostgresJSONDidMeta.delete_metadata`](https://github.com/rucio/rucio/blob/017f3c25605d14307272f646b3f3deedd26cfb91/lib/rucio/core/did_meta_plugins/postgres_meta.py#L261-L279), deleting metadata for one (scope, name) strips that key from every DID stored in the table. Is this indeed on purpose? 

Originally it was:

```
statement = "UPDATE {} ".format(self.table) + \
            "SET data = {}.data - '{}';".format(self.table, key)
```

We need to either adjust the query (or at least the documentation) because currently it sounds like a bug as pointed out by @panta-123.

### Steps to reproduce

.

### Rucio Version

_No response_

### Additional Information

_No response_

panta-123:
- add WHERE stm for delete metadata so that only key from that did is deleted. NOT the key from all rows as it is now
 - add a test for delete metadata.

### DIFF
--- a/lib/rucio/core/did_meta_plugins/postgres_meta.py
+++ b/lib/rucio/core/did_meta_plugins/postgres_meta.py
@@ -260,17 +260,19 @@ class ExternalPostgresJSONDidMeta(DidMetaPlugin):
 
     def delete_metadata(self, scope, name, key, *, session: "Optional[Session]" = None):
         """
-        Delete a key from metadata.
+        Delete a key from metadata of DID.
 
         :param scope: the scope of DID
         :param name: the name of the DID
         :param key: the key to be deleted
         :param session: the database session in use
         """
-        statement = sql.SQL("UPDATE {} SET data = {}.data - {}").format(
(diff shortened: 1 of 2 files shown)


Geogouz:
Looks fine. I am only having problems related to the test you introduced. Just realised that (along with the rest external postgres JSON tests) it is actually getting skipped in our CI (probably due to a decorator that requires rse configs). But that's not a problem introduced by your commit. I will fix it first and come back to this.

panta-123:
@Geogouz , Is the other issue of test being skipped fixed?

Geogouz:
> @Geogouz , Is the other issue of test being skipped fixed?

Can you rebase please to trigger the updated tests?

panta-123:
All checks passed.

Geogouz (approved):
Looks good

rdimaio (approved):

PR #8098, opened 2025-10-06
Documentation: Add docstrings and comments to `run_tests.py`; rucio#8093

### LINKED ISSUE #8093: Add docstrings to `tools/test/run_tests.py`
Right now, `tools/test/run_tests.py` offers little to none docstrings explanation of what is happening there. Hence, it is not easy for a newcomer to understand what and why is happening there. We need to change that.

Geogouz:
Fixes #8093

### DIFF
--- a/tools/test/run_tests.py
+++ b/tools/test/run_tests.py
@@ -13,6 +13,72 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+"""
+Test runner used by Rucio's CI/autotest tooling.
+
+``donkeyrider.py`` collects matrix entries from ``etc/docker/test/matrix.yml``
+(via :mod:`tools.test.matrix_parser`), resolves runtime images, and pipes a JSON
+payload into this module. From that payload we derive a sequence of *cases*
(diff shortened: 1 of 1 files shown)


tools/test/run_tests.py
```diff
@@ -13,6 +13,72 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+"""
+Test runner used by Rucio's CI/autotest tooling.
+
+``donkeyrider.py`` collects matrix entries from ``etc/docker/test/matrix.yml``
```
rdimaio, line 19: The docstring here mentions `donkeyrider`, but do we actually use `donkeyrider`? My understanding is that it was meant to be deprecated

Geogouz: `tools/run_autotests.sh` ->  `tools/test/donkeyrider.py` -> `tools/test/run_tests.py`
So unless the scripts are executed manually.. 
- `run_tests` is invoked (only) by `donkeyrider`.
- `donkeyrider` is invoked (only) by `run_autotests.sh`.

rdimaio (approved):

PR #8105, opened 2025-10-09
Rules: Apply exponential backoff on rules with a source replica expression #8104

### LINKED ISSUE #8104: Repairer exponential backoff does not apply on rules with a source replica expression
### Description

https://github.com/rucio/rucio/blob/db1461d98a4609e8e3a87b39bda8c23a749c3cc2/lib/rucio/core/rule_grouping.py#L893-L913

Insight from Martin:

> the thinking probably was that a rule with a source_replica_expression is only used in a chained form. Thus often when the rule is created, the source does not exist yet

### Steps to reproduce

Not easily. One needs to create a situation where a request can be continuously reattempted (e.g. a unique corrupted file), then observe that even after four days the Repairer creates a new request each time it processes the rule.

### Rucio Version

_No response_

### Additional Information

_No response_

dchristidis:
Previously, the Repairer would always try to create new requests on rules with a source replica expression, irrespective of when the rules were created.

The motivation behind this was chained rules. Suppose a DID is registered on RSE A, and some mechanism simultaneously creates replication rules on RSEs B and C, the latter with a source replica expression so that the requests use the B→C path and not the A→C path. Such a rule would be created in state STUCK, given that the replicas are not yet available on B. Under such conditions, the Repairer should not apply the exponential backoff in order for the rule on C to pick up on the replication on B as soon as possible.

This commit tweaks the condition so that the exponential backoff is not applied before a request is created and at the first time it is recreated (by the Repairer, not the Finisher). The latter is unintentional but unavoidable.

### DIFF
--- a/lib/rucio/core/rule_grouping.py
+++ b/lib/rucio/core/rule_grouping.py
@@ -690,7 +690,7 @@ def __repair_stuck_locks_with_none_grouping(datasetfiles, locks, replicas, sourc
                     associated_replica.lock_cnt = session.execute(stmt).scalar_one()
                     continue
                 # Check if this is a STUCK lock due to source_replica filtering
-                if source_rses:
+                if source_rses and not lock.repair_cnt:
                     associated_replica = [replica for replica in replicas[(file['scope'], file['name'])] if replica.rse_id == lock.rse_id][0]
                     # Check if there is an eligible source replica for this lock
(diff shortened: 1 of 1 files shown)


dchristidis:
Ruff’s errors are unrelated to the PR and should be ignored.

rdimaio (approved):

Issue #8102, opened 2025-10-07
InvalidPolicyPackageAlgorithmType: Invalid policy package algorithm type 'surl'.


StevenCTimm:
We attempted to upgrade the fermilab dune-rucio.fnal.gov and got this error on almost all daemons:

│ httpd-error-log [Tue Oct 07 19:13:59.412037 2025] [wsgi:info] [pid 36:tid 600] [remote 100.64.0.7:56818] mod_wsgi (pid=36, process='rucio', application='rucio'): Loading Python script file '/usr/local/lib/python3.9/site-packag │
│ httpd-error-log [Tue Oct 07 19:13:59.425914 2025] [wsgi:error] [pid 36:tid 600] [remote 100.64.0.7:56818] mod_wsgi (pid=36): Failed to exec Python script file '/usr/local/lib/python3.9/site-packages/rucio/web/rest/flaskapi/v1/ │
│ httpd-error-log [Tue Oct 07 19:13:59.425954 2025] [wsgi:error] [pid 36:tid 600] [remote 100.64.0.7:56818] mod_wsgi (pid=36): Exception occurred processing WSGI script '/usr/local/lib/python3.9/site-packages/rucio/web/rest/flas │
│ httpd-error-log [Tue Oct 07 19:13:59.426127 2025] [wsgi:error] [pid 36:tid 600] [remote 100.64.0.7:56818] Traceback (most recent call last):                                                                                       │
│ httpd-error-log [Tue Oct 07 19:13:59.426177 2025] [wsgi:error] [pid 36:tid 600] [remote 100.64.0.7:56818]   File "/usr/local/lib/python3.9/site-packages/rucio/web/rest/flaskapi/v1/main.py", line 90, in <module>                 │
│ httpd-error-log [Tue Oct 07 19:13:59.426180 2025] [wsgi:error] [pid 36:tid 600] [remote 100.64.0.7:56818]     apply_endpoints(application, endpoints)                                                                              │
│ httpd-error-log [Tue Oct 07 19:13:59.426183 2025] [wsgi:error] [pid 36:tid 600] [remote 100.64.0.7:56818]   File "/usr/local/lib/python3.9/site-packages/rucio/web/rest/flaskapi/v1/main.py", line 61, in apply_endpoints          │
[...]

PR #8113, opened 2025-10-10
Policies: remove surl algorithm name alias #8102

jamesp-epcc:
Last year we renamed `surl` algorithms to `non_deterministic_pfn` algorithms to make it clearer what they do. This code was added to allow policy packages that used `surl` to continue working for the moment, but recently this aliasing was broken by https://github.com/rucio/rucio/pull/7981 . We could fix the alias but unless it's going to majorly inconvenience anyone, my preference would be to just remove it as we don't want to maintain it forever. (DUNE was relying on it but I have now updated their policy package to use the new name).

### DIFF
--- a/lib/rucio/common/plugins.py
+++ b/lib/rucio/common/plugins.py
@@ -218,10 +218,6 @@ class PolicyPackageAlgorithms:
             if hasattr(module, 'get_algorithms'):
                 all_algorithms = module.get_algorithms()
 
-                # for backward compatibility, rename 'surl' to 'non_deterministic_pfn' here
-                if 'surl' in all_algorithms:
-                    all_algorithms['non_deterministic_pfn'] = all_algorithms['surl']
-
                 # check that the names are correctly prefixed for multi-VO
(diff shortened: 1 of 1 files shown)


rdimaio (approved):

lib/rucio/common/plugins.py
```diff
@@ -218,10 +218,6 @@ def _try_importing_policy(cls: type[PolicyPackageAlgorithmsT], vo: str = "") ->
             if hasattr(module, 'get_algorithms'):
                 all_algorithms = module.get_algorithms()
 
-                # for backward compatibility, rename 'surl' to 'non_deterministic_pfn' here
```
rdimaio, line 221: This was added in July 2024 https://github.com/rucio/rucio/commit/37b445dbbfd16262b4dafc474c978903a4031e83, so I think enough time passed to merge this as part of a minor release rather than waiting for the next major release.

rdimaio (approved):

PR #8115, opened 2025-10-10
Core: Invalid global account limits resolve zero RSEs; Fix #7726

### LINKED ISSUE #7726: Add protection around RSE expressions of global account limits
### Description

Firstly, at the time a new global account limit is added, the RSE expression should be validated and rejected if it does not evaluate to at least one RSE.

Secondly, an RSE expression may be valid at the time of creation, but later evaluate to an empty set. As such, any parts of the code that handle global account limits should deal with `InvalidRSEExpression` gracefully.

### Motivation

_No response_

### Change

_No response_

### LINKED ISSUE #7727: `perm_set_global_account_limit()` leads to an unhandled exception if the RSE expression is invalid
### Description

https://github.com/rucio/rucio/blob/28ca2074af175e345f2b0f72cccd4c693a48577c/lib/rucio/core/permission/generic.py#L886-L887

At that point in the code, there has been no input validation. Maybe it’s sufficient to catch `InvalidRSEExpression` in the API layer. This will also address #7726 if the exception is originating from the core layer.

### Steps to reproduce

Execute `rucio-admin account set-limits root 'some-invalid-expression' 1TB global` with a non-admin account.

### Rucio Version

_No response_

### Additional Information

_No response_

bziemons:
Previously, having an invalid RSE expression (including an expression resolving to no RSEs) in a global account limit results in an InvalidRSEException in any operation regarding this limit. Instead, an invalid RSE expression should result in no quota for the account and regular operations should work as expected.

Also fixes #7727

### DIFF
--- a/lib/rucio/core/account_limit.py
+++ b/lib/rucio/core/account_limit.py
@@ -19,7 +19,7 @@ from sqlalchemy.sql import func, literal, select
 from sqlalchemy.sql.expression import and_, or_
 
 from rucio.common.constants import DEFAULT_VO
-from rucio.common.exception import AccountNotFound
+from rucio.common.exception import AccountNotFound, InvalidRSEExpression
 from rucio.core.account import account_exists, get_all_rse_usages_per_account
 from rucio.core.rse import get_rse_name
 from rucio.core.rse_expression_parser import parse_expression
@@ -131,16 +131,23 @@ def get_global_account_limit(account: Optional["InternalAccount"] = None, rse_ex
 
     resolved_global_account_limits = {}
     for limit in global_account_limits:
-        if account:
-            resolved_rses = parse_expression(limit['rse_expression'], filter_={'vo': account.vo}, session=session)
-        else:
-            resolved_rses = parse_expression(limit['rse_expression'], session=session)
         limit_in_bytes = float('inf') if limit['bytes'] == -1 else limit['bytes']
-        resolved_global_account_limits[limit['rse_expression']] = {
-            'resolved_rses': [resolved_rse['rse'] for resolved_rse in resolved_rses],
-            'resolved_rse_ids': [resolved_rse['id'] for resolved_rse in resolved_rses],
-            'limit': limit_in_bytes
+        resolved_limit = {
+            'resolved_rses': [],
+            'resolved_rse_ids': [],
+            'limit': limit_in_bytes,
         }
+        try:
+            if account:
+                resolved_rses = parse_expression(limit['rse_expression'], filter_={'vo': account.vo}, session=session)
+            else:
+                resolved_rses = parse_expression(limit['rse_expression'], session=session)
+            resolved_limit['resolved_rses'] = [resolved_rse['rse'] for resolved_rse in resolved_rses]
+            resolved_limit['resolved_rse_ids'] = [resolved_rse['id'] for resolved_rse in resolved_rses]
+        except InvalidRSEExpression:
+            # invalid expressions resolve to no RSEs
+            pass
+        resolved_global_account_limits[limit['rse_expression']] = resolved_limit
     return resolved_global_account_limits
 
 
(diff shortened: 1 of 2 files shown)


bziemons:
@dchristidis let me guess.. you don't like it because there are no tests? :smirk:

lib/rucio/core/account_limit.py
```diff
@@ -131,16 +131,23 @@ def get_global_account_limit(account: Optional["InternalAccount"] = None, rse_ex
...
+        try:
+            if account:
+                resolved_rses = parse_expression(limit['rse_expression'], filter_={'vo': account.vo}, session=session)
+            else:
+                resolved_rses = parse_expression(limit['rse_expression'], session=session)
+            resolved_limit['resolved_rses'] = [resolved_rse['rse'] for resolved_rse in resolved_rses]
+            resolved_limit['resolved_rse_ids'] = [resolved_rse['id'] for resolved_rse in resolved_rses]
+        except InvalidRSEExpression:
+            # invalid expressions resolve to no RSEs
+            pass
```
Geogouz, line 149: With this, the configured byte limit is returned even when the expression cannot be parsed. So an invalid expression no longer blocks usage queries but still grants effective quota instead of “no quota” as stated in the commit message. Maybe a safer approach would be to set the limit to 0 when parsing fails (and perhaps when it resolves to zero RSEs), so the account gets no usable quota while downstream logic remains functional.

lib/rucio/core/account_limit.py
```diff
@@ -379,17 +386,22 @@ def get_global_account_usage(account: "InternalAccount", rse_expression: Optiona
...
-        ).where(
-            and_(models.AccountUsage.account == account,
-                 models.AccountUsage.rse_id.in_(resolved_rses))
-        ).group_by(
-            models.AccountUsage.account
-        )
-        usage = session.execute(stmt).first()
+        try:
+            rses = parse_expression(rse_expression, filter_={'vo': vo}, session=session)
+            resolved_rses = [resolved_rse['id'] for resolved_rse in rses]
```
Geogouz, line 391: When `resolved_rses` is empty (e.g. no RSEs matched), `usage` is `None` while `limit` stays whatever `get_global_account_limit` returned. Should we treat such empty resolutions as invalid and force `limit = 0, usage = (0, 0)`?

lib/rucio/core/permission/generic.py
```diff
@@ -881,13 +882,19 @@ def perm_set_global_account_limit(issuer: "InternalAccount", kwargs: dict[str, A
     """
     if _is_root(issuer) or has_account_attribute(account=issuer, key='admin', session=session):
         return True
+    try:
+        # parse expression, fail early if not working or no results
+        rses = parse_expression(kwargs['rse_expression'], filter_={'vo': issuer.vo}, session=session)
+    except InvalidRSEExpression:
```
Geogouz, line 888: Here, I guess expressions that resolve to zero RSEs are still allowed (permitting creation of global limits that apply to no RSEs). Should we add additional validation that at least one RSE is resolved? (same holds for the deletion bellow)

lib/rucio/core/account_limit.py
Geogouz, line 369: Should we describe the new invalid-expression handling maybe?

lib/rucio/core/permission/generic.py
```diff
@@ -881,13 +882,19 @@ def perm_set_global_account_limit(issuer: "InternalAccount", kwargs: dict[str, A
     """
```
Geogouz, line 882: Also here, not sure if some description may be helpful. Up to you.

Geogouz:
Other than the review comments, in general it looks good to me. Maybe adding some tests would make it top ˆˆ

bziemons:
I will close this until I can get back to work on this.

PR #8124, opened 2025-10-15
Fix --lfn of rucio replica update bad

### LINKED ISSUE #8123: --lfn option broken for `rucio replica update bad`
### Description

Only the first argument should be passed

### Steps to reproduce

Run `rucio replica update bad --scope <scope> --rse <rse> --lfn <lfn list>`

```
2025-10-15 15:11:54,173 ERROR   expected str, bytes or os.PathLike object, not tuple
```

### Rucio Version

38.3

### Additional Information

_No response_

maxnoe:
Fixes #8123

### DIFF
--- a/tests/test_bin_rucio.py
+++ b/tests/test_bin_rucio.py
@@ -1350,3 +1350,24 @@ def test_cli_declare_bad_replicas(cli, lfn, rse_factory, mock_scope, did_factory
 
     replicas = next(rucio_client.list_replicas([did], rse_expression=rse, all_states=True))
     assert replicas["states"][rse] == "BAD"
+
+
+def test_cli_declare_bad_replicas_invalid_usage():
+    """CLIENT(USER): Rucio declare bad replica invalid argument handling"""
+    base_cmd = ["rucio", "replica", "state", "update", "bad"]
+
+    def run(expected_error, args=None, expected_code=1):
+        args = args or []
+        cmd = shlex.join(base_cmd + args)
+        code, stdout, stderr = execute(cmd)
+
+        assert code == expected_code, f"Running {cmd} did not fail as expected. out:\n{stdout}\nerr\n{stderr}"
+        assert expected_error in stderr, f"Expected error message not found in stderr:\n{stderr}"
+
+    run("Missing option '--reason'", expected_code=2)
+
+    args = ["--reason", "test", "--lfn", "foo"]
+    run("Scope and RSE are required", args=args, expected_code=1)
+
+    args = ["--reason", "test", "--scope", "test", "--rse", "test", "--lfn", "foo", "bar"]
+    run("Exactly one", args=args, expected_code=1)
--- a/tests/test_bin_rucio.py
+++ b/tests/test_bin_rucio.py
@@ -12,9 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import logging
 import os
 import random
 import re
+import shlex
 import tempfile
 from datetime import datetime, timedelta, timezone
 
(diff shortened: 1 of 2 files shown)


maxnoe:
@voetberg What would be the right way to add tests for this?

voetberg:
> @voetberg What would be the right way to add tests for this?

I'd recommend adding the CLI command that originally raised the issue in `test_bin_rucio` and verify the runs correctly now. 

Idr if there is a equivalent test for the old cli, but if there is, you can just duplicate it with the new cli structure
```
def test_declare_lfn_bad(....): 
     ...
     cmd = f"rucio replica update bad --scope {scope} --rse {rse} --lfn {list of lfns}"
     exitcode, out, log = execute(cmd)
     assert exitcode != 0 
     assert "ERROR" in log
     assert "Exactly one positional argument expected in case of LFN list" in log
     
     cmd = f"rucio replica update bad --scope {scope} --rse {rse} --lfn {single_lfn}"
     exitcode, out, log = execute(cmd)
     assert exitcode == 0
     ... # checking that the lfn is actually bad via py client 
```

maxnoe:
There seem to be no existing test for this part of the CLI.

I'll have  a look tomorrow.

rdimaio (dismissed):

tests/test_bin_rucio.py
```diff
@@ -1292,3 +1295,60 @@ def test_download_file_check_by_size(rse_factory, mock_scope, did_factory):
...
+    else:
+        bad_replicas.append(f"{did['scope']}:{did['name']}")
+
+    cmd.extend(bad_replicas)
+    cmd = shlex.join(cmd)
+
+    code, stdout, stderr = execute(cmd)
+    assert code == 0, f"Running {cmd} failed. out:\n{stdout}\nerr\n{stderr}"
+    log.info(stdout)
+    log.warning(stderr)
```
voetberg: These logging statements won't be displayed if the test fails, and if it successes it won't display anything at all (the test suite suppresses them), so I sort of fail to see their usefulness

tests/test_bin_rucio.py
```diff
@@ -1292,3 +1295,60 @@ def test_download_file_check_by_size(rse_factory, mock_scope, did_factory):
...
+
+    client = ReplicaClient()
```
voetberg: I recommend using the `rucio_client` fixture instead of initializing a new client

tests/test_bin_rucio.py
```diff
@@ -1292,3 +1295,60 @@ def test_download_file_check_by_size(rse_factory, mock_scope, did_factory):
...
+@pytest.mark.parametrize(
+    ("cli", "lfn"),
+    [
+        ("new", True),
+        ("old", True),
+        ("new", False),
+        ("old", False),
+    ]
+)
+def test_cli_declare_bad_replicas(cli, lfn, rse_factory, mock_scope, did_factory, tmp_path):
```
voetberg: Could we also include a failure case verifying the valueerror is raised as expected when multiple lfns are supplied?

tests/test_bin_rucio.py
maxnoe: The output is displayed when the test fails by default.

This is very useful for local development, as they become visible with pytest options like `--log-cli-level=INFO` for example.

tests/test_bin_rucio.py
maxnoe: Done

tests/test_bin_rucio.py
voetberg: I agree now that the order of assertion vs log is flipped!

tests/test_bin_rucio.py
maxnoe: sure

maxnoe: Done

voetberg (approved):

rdimaio (approved):

PR #8128, opened 2025-10-16
Metadata: Refactor metadata plugins to centralize metadata plugin_name

### LINKED ISSUE #8127: Centralize plugin name handling in DID metadata plugin base class
[DidMetaPlugin](https://github.com/rucio/rucio/blob/5f276f4d5d971ff7cae6dcc2722b7b4e2bb15fbd/lib/rucio/core/did_meta_plugins/did_meta_plugin_interface.py#L29) doesn’t initialize a shared `plugin_name` attribute or provide a default getter, so every subclass repeats the same logic and assumptions about this field.

I suggest doing that to reduce redundancy and potential inconsistencies.

Geogouz:
Fixes rucio#8127

### DIFF
--- a/lib/rucio/core/did_meta_plugins/__init__.py
+++ b/lib/rucio/core/did_meta_plugins/__init__.py
@@ -96,8 +96,9 @@ def get_metadata(scope, name, plugin="DID_COLUMN", *, session: "Session"):
         return all_metadata
     else:
         for metadata_plugin in METADATA_PLUGIN_MODULES:
-            if metadata_plugin.get_plugin_name().lower() == plugin.lower():
+            if metadata_plugin.is_named(plugin):
                 return metadata_plugin.get_metadata(scope, name, session=session)
+
     raise exception.UnsupportedMetadataPlugin(f'Metadata plugin "{plugin}" is not enabled on the server.')
 
 
--- a/lib/rucio/core/did_meta_plugins/did_meta_plugin_interface.py
+++ b/lib/rucio/core/did_meta_plugins/did_meta_plugin_interface.py
@@ -13,13 +13,11 @@
 # limitations under the License.
 
 from abc import ABCMeta, abstractmethod
-from typing import TYPE_CHECKING, Literal
-
-from rucio.db.sqla.session import transactional_session
+from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
     from collections.abc import Iterator
-    from typing import Any, Optional, Union
+    from typing import Any, Literal
 
     from sqlalchemy.orm import Session
 
@@ -35,7 +33,23 @@ class DidMetaPlugin(metaclass=ABCMeta):
         """
         Initializes the plugin
         """
-        pass
+        self._plugin_name = None
+
+    @property
+    def name(self) -> str:
+        """
+        The getter method for the plugin's name.
+
+        :returns: The standardized (casefolded) name of this plugin.
+        :raises AttributeError: If '_plugin_name' is not defined in the subclass.
+        """
+        if self._plugin_name:
+            return self._plugin_name.casefold()
+        raise AttributeError("Subclasses of DidMetaPlugin must define the '_plugin_name' attribute.")
+
+    def is_named(self, plugin_name: str) -> bool:
+        """Return whether the plugin matches the provided name using case-insensitive comparison."""
+        return self.name == plugin_name.casefold()
 
     @abstractmethod
     def get_metadata(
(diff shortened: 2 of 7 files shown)


lib/rucio/core/did_meta_plugins/did_meta_plugin_interface.py
```diff
@@ -35,20 +33,32 @@ def __init__(self):
...
         """
-        pass
+        self._plugin_name = None
+
+    @property
+    def name(self) -> str:
+        """
+        The getter method for the plugin's name.
+
+        :returns: The standardized (uppercase) name of this plugin.
```
rdimaio: This is a bit confusing because the returned name is actually lowercase, though this says uppercase (I guess the uppercase here refers to the original case of the name)

lib/rucio/core/did_meta_plugins/__init__.py
```diff
@@ -96,8 +96,9 @@ def get_metadata(scope, name, plugin="DID_COLUMN", *, session: "Session"):
         return all_metadata
     else:
         for metadata_plugin in METADATA_PLUGIN_MODULES:
-            if metadata_plugin.get_plugin_name().lower() == plugin.lower():
+            if metadata_plugin.name == plugin.lower():
```
rdimaio: Casefold is the recommended way of doing case-insensitive string comparison since python 3.3 https://stackoverflow.com/questions/319426/how-do-i-do-a-case-insensitive-string-comparison

rdimaio (commented):
I'm a bit confused as to why we set all plugin names to uppercase but then we compare them as lowercase - should we set them to lowercase from the beginning?

lib/rucio/core/did_meta_plugins/__init__.py
Geogouz: In practice, currently, we indeed do not have any assurance that future plugin names will always be in ASCII (although these sound like really edge cases that would be easy to detect because while testing). So we have to choose between the `casefold` method and `lower` considering how "heavy", "safe", "intuitive" they are.

Practically, the main problem I see might be `casefold` (but even `lower` to be honest) to be forgotten and one doing something like `if metadata_plugin.name == plugin:`. Maybe an idea would be to introduce also something like `if metadata_plugin.is_named(plugin)` and enforce the casefolding or lowercasing there so that users shouldn't care. Moreover I don't like that `plugin` was used as var name for the name itself instead the plugin object (I can change that too and sneak all these changes in the same PR:/).

lib/rucio/core/did_meta_plugins/did_meta_plugin_interface.py
Geogouz: Forgotten 🥲

lib/rucio/core/did_meta_plugins/did_meta_plugin_interface.py
```diff
@@ -63,39 +77,38 @@ def set_metadata(
...
         :param value: the value.
-        :param did: The data identifier info.
-        :param recursive: Option to propagate the metadata change to content.
+        :param recursive: Instruction to propagate the metadata change recursively to content (False by default).
         :param session: The database session in use.
         """
         pass
 
-    @transactional_session
+    @abstractmethod
```
rdimaio, line 94: Does this change potentially cause issues with session management? In theory we should use the context manager to manage the session either in this function or in a parent function - if done in a parent function, this function should always be passed a session, so we would need to specify that the type is always `Session`

lib/rucio/core/did_meta_plugins/did_meta_plugin_interface.py
```diff
@@ -63,39 +77,38 @@ def set_metadata(
...
     def set_metadata_bulk(
         self,
         scope: "InternalScope",
         name: str,
         meta: dict[str, "Any"],
         recursive: bool = False,
         *,
-        session: "Optional[Session]" = None
+        session: "Session | None" = None
```
rdimaio, line 102: IMO I would wait until py3.10 is the min version before we do this syntax change, so that we can do it without using quotes.

lib/rucio/core/did_meta_plugins/did_meta_plugin_interface.py
Geogouz: Although now the base interface method itself no longer opens any session (same pattern followed in the rest of the available methods there), the entry point used [`rucio.core.did_meta_plugins.set_metadata_bulk`](https://github.com/Geogouz/rucio/blob/b1380dec7dfca987c1bc4f5a7ca1b376b4225552/lib/rucio/core/did_meta_plugins/__init__.py#L139-L140) (i.e. the dispatcher that routes requests to the correct plugin implementations) and all these plugin implementations (which are some times called directly via tests etc) remain wrapped in `@transactional_session`. So this relies always on the context established by the callers. But to remain compatible with the transactional decorators used on the implementations and to preserve the ability for callers to omit the session and let the decorator manage it, the abstract method would still need to accept `Session | None`. Otherwise, tightening the annotation to `Session` would misrepresent that contract (even though the runtime would keep working).

rdimaio (approved):

PR #8135, opened 2025-10-20
Change didtype to explicitly use scope extraction policy

### LINKED ISSUE #7979: DID constructor does not use policy `extract_scope` function to determine scope
### Description

class `DID` has complex rules to parse scope and name from a string, but does not actually use the function advertised for policy packages to do this: `extract_scope`.

This is confusing and e.g. means that `rucio download` does not support giving the did in this convention. 

### Steps to reproduce

Define a `extract_scope` in policy package.

Run `rucio download <name>` where name contains the scope and extract_scope correctly returns the scope.

Download fails. in our case for some reason, the scope is parsed as `/ctao`:

```
>>> c.download_dids([{"did": "/ctao.dpps.test/root/foo.dat"}])
ERROR:root:DID does not exist: /ctao:/ctao.dpps.test/root/foo.dat
```

Where `extract_scope` works fine:
```
>>> extract_scope("/ctao.dpps.test/root/foo.dat")
('root', '/ctao.dpps.test/root/foo.dat')
```

### Rucio Version

38.1, current master

### Additional Information

_No response_

voetberg:
Closes #7979

### DIFF
--- a/lib/rucio/common/didtype.py
+++ b/lib/rucio/common/didtype.py
@@ -15,10 +15,12 @@
 """
 DID type to represent a DID and to simplify operations on it
 """
-
+import logging
+from configparser import NoSectionError
 from typing import Any, Union
 
-from rucio.common.exception import DIDError
+from rucio.common.exception import ConfigNotFound, DIDError, InvalidAlgorithmName
+from rucio.common.utils import extract_scope
 
 
 class DID:
@@ -126,15 +128,20 @@ class DID:
         Construct the DID from a string.
         :param did: string containing the DID information
         """
-        did_parts = did.split(DID.SCOPE_SEPARATOR, 1)
-        if len(did_parts) == 1:
-            self.name = did
-            self._update_implicit_scope()
-            if not self.has_scope():
-                raise DIDError('Object construction from non-splitable string is ambigious')
-        else:
-            self.scope = did_parts[0]
-            self.name = did_parts[1]
+        try:
+            self.scope, self.name = extract_scope(did)
+        except (ImportError, InvalidAlgorithmName, ConfigNotFound, NoSectionError) as e:  # Only use when the policy can not be found
+            logging.debug("Failure using extract_scope policy for '%s': %s - Using fallback." % (did, type(e).__name__))
+            did_parts = did.split(DID.SCOPE_SEPARATOR, 1)
+            if len(did_parts) == 1:
+                self.name = did
+                self._update_implicit_scope()
+                if not self.has_scope():
+                    error = f"Could not parse scope from did string {did} - fallback policy expects only one '{DID.SCOPE_SEPARATOR}'"
+                    raise DIDError(error)
+            else:
+                self.scope = did_parts[0]
+                self.name = did_parts[1]
 
     def _did_from_dict(self, did: dict[str, str]) -> None:
         """
--- a/tests/rucio/common/test_didtype.py
+++ b/tests/rucio/common/test_didtype.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+
 import pytest
 
 from rucio.common.didtype import DID
@@ -62,10 +63,6 @@ class TestDIDType:
     def test_did_type_success(self, input_did, expected_scope, expected_name):
         assert input_did.scope == expected_scope
         assert input_did.name == expected_name
-
-    def test_non_implicit_single_string(self):
-        with pytest.raises(DIDError, match='Error using DID type\nDetails: Object construction from non-splitable string is ambigious'):
-            DID('non.implicit.single.string')
 
     def test_copy(self):
         x = DID('test.scope:test.name')
--- a/tests/test_policy_package.py
+++ b/tests/test_policy_package.py
@@ -13,7 +13,9 @@
 # limitations under the License.
 
 import importlib
+import logging
 import os
+from unittest import mock
 
 import pytest
 
@@ -21,6 +23,8 @@ import rucio.common.exception
 import rucio.common.schema
 import rucio.core.permission
 from rucio.common.constants import DEFAULT_VO
+from rucio.common.didtype import DID
+from rucio.common.exception import DIDError
 from rucio.common.plugins import PolicyPackageAlgorithms
 from rucio.common.types import InternalAccount
 
@@ -127,3 +131,40 @@ class TestPolicyPackage:
             os.environ['RUCIO_POLICY_PACKAGE'] = old_pp_env
         else:
             del os.environ['RUCIO_POLICY_PACKAGE']
+
+
+@pytest.mark.noparallel(reason='Changes scope extraction policies')
+@pytest.mark.parametrize("file_config_mock", [
+        {"overrides": [("common", "extract_scope", "my_extraction_algo")]},
+    ], indirect=True)
+def test_change_scope_extraction(file_config_mock, did_factory, scope_factory, vo, rse_factory):
+
+    rse, _ = rse_factory.make_posix_rse()
+    mock_scope, _ = scope_factory(vos=[vo])
+
+    def extract_scope(did, scopes, *args, **kwargs):
+        return mock_scope, did.split(":")[-1]
+
+    from rucio.common.utils import ScopeExtractionAlgorithms
+    ScopeExtractionAlgorithms.register("my_extraction_algo", extract_scope)
+
+    n_files = 3
+    did_names = [did['did_name'] for did in did_factory.upload_test_dataset(rse_name=rse, scope=mock_scope, nb_files=n_files)]
+
+    constucted_dids = [f"fake_scope:{did}" for did in did_names]
+
+    for did in constucted_dids:
+        derived_did = DID(did)
+        assert derived_did.scope == mock_scope
+        assert derived_did.name == did.split(':')[-1]
+
+
+def test_scope_extraction_invalid_name(caplog):
+    caplog.set_level(logging.DEBUG)  # Override log level
+    with mock.patch('rucio.common.didtype.extract_scope', side_effect=ImportError()):
+
+        DID("scope:name")
+        assert "Failure using extract_scope policy for 'scope:name'" in caplog.text
+
+        with pytest.raises(DIDError):
+            DID("invalid/scope/name/combo")


lib/rucio/common/didtype.py
```diff
@@ -126,15 +127,18 @@ def _did_from_str(self, did: str) -> None:
...
-            self._update_implicit_scope()
-            if not self.has_scope():
-                raise DIDError('Object construction from non-splitable string is ambigious')
-        else:
-            self.scope = did_parts[0]
-            self.name = did_parts[1]
+        try:
+            self.scope, self.name = extract_scope(did)
+        except (ImportError, InvalidAlgorithmName, ConfigNotFound, NoSectionError):  # Only use when the policy can not be found
+            did_parts = did.split(DID.SCOPE_SEPARATOR, 1)
```
mgajek-cern, line 135: We could add visibility into when fallback happens:

```suggestion
             logging.debug(f"extract_scope fallback for '{did}': {type(e).__name__}")
            did_parts = did.split(DID.SCOPE_SEPARATOR, 1)
```

lib/rucio/common/didtype.py
```diff
@@ -126,15 +127,18 @@ def _did_from_str(self, did: str) -> None:
...
-        if len(did_parts) == 1:
-            self.name = did
-            self._update_implicit_scope()
-            if not self.has_scope():
-                raise DIDError('Object construction from non-splitable string is ambigious')
-        else:
-            self.scope = did_parts[0]
-            self.name = did_parts[1]
+        try:
+            self.scope, self.name = extract_scope(did)
```
mgajek-cern, line 132: Noticed `extract_scope` returns `Sequence[str]` but the DID constructor expects a tuple of two strings.

The type hint suggests it could return any sequence, but we're unpacking it as exactly 2 elements.
Could we clarify the return type in `extract_scope` function within `common/utils.py`? Maybe:

```python
def extract_scope(...) -> tuple[str, str]:
```

lib/rucio/common/didtype.py
```diff
@@ -126,15 +127,18 @@ def _did_from_str(self, did: str) -> None:
...
+        except (ImportError, InvalidAlgorithmName, ConfigNotFound, NoSectionError):  # Only use when the policy can not be found
```
mgajek-cern: Should validation errors (like `ValueError`) from custom extraction algorithms trigger the fallback or propagate?

tests/rucio/common/test_didtype.py
```diff
@@ -64,8 +66,14 @@ def test_did_type_success(self, input_did, expected_scope, expected_name):
...
-        with pytest.raises(DIDError, match='Error using DID type\nDetails: Object construction from non-splitable string is ambigious'):
-            DID('non.implicit.single.string')
+        # This is technically valid under ATLAS construction
+        if "unit_tests" not in os.environ.get("GITHUB_WORKFLOW", ""):
+            did = DID('non.implicit.single.string')
+            assert did.name == 'non.implicit.single.string'
+            assert did.scope == 'non'
+        else:
+            with pytest.raises(DIDError):
+                did = DID('non.implicit.single.string')
```
mgajek-cern: We have a test that behaves differently in CI versus locally. Would it make sense to explicitly configure the policy in the test rather than relying on environment? 

---

E.g. update in `test_didtype.py` (**NOTE:** Just a draft example. Needs to be reviewed and verified):

```python
@pytest.mark.parametrize("extract_scope_config", [
    "atlas",  # Configured to extract scope from paths
    None,     # No custom extraction configured
], indirect=True)
def test_non_implicit_single_string(self, extract_scope_config):
    if extract_scope_config == "atlas":
        # With ATLAS-style extraction configured
        did = DID('non.implicit.single.string')
        assert did.scope == 'non'
        assert did.name == 'non.implicit.single.string'
    else:
        # Without custom extraction, ambiguous string should error
        with pytest.raises(DIDError):
            DID('non.implicit.single.string')
```

and a fixture in `tests/conftest.py`:

```python
@pytest.fixture
def extract_scope_config(request):
    """Fixture that configures extract_scope based on parameter"""
    from rucio.common.utils import ScopeExtractionAlgorithms
    
    config_type = request.param
    
    if config_type == "atlas":
        # Register ATLAS-style scope extraction algorithm
        def atlas_extract_scope(did, scopes, *args, **kwargs):
            # Simple ATLAS-style: first part before dot is scope
            parts = did.split('.', 1)
            if len(parts) > 1:
                return parts[0], did
            return '', did
        
        ScopeExtractionAlgorithms.register("test_atlas", atlas_extract_scope)
        
        # Mock config to use this algorithm
        original_config = None
        try:
            from rucio.common import config
            original_config = config.config_get('common', 'extract_scope', False, None)
        except:
            pass
        
        # Here you'd need to temporarily set config - this depends on your config mock setup
        # For now, just yield the config type
        yield config_type
        
        # Cleanup: unregister algorithm
        if hasattr(ScopeExtractionAlgorithms, '_algorithms'):
            ScopeExtractionAlgorithms._algorithms.pop("test_atlas", None)
    else:
        # No custom extraction configured
        yield None
```

---

Would it make sense to include integration tests that set up a custom `extract_scope` and run a command like `rucio download /ctao.dpps.test/root/foo.dat` or could you try running this manually and share a screenshot of the result?

mgajek-cern (commented):
Just a few remarks/questions/findings

lib/rucio/common/didtype.py
voetberg: I think they should trigger an error, it means they have an extraction policy and it didn't work for the given DID. It would just cause inconsistencies if the fallback could be used when the policy fails

tests/rucio/common/test_didtype.py
voetberg: I honestly think this is just a poorly constructed test to begin with, because there are too many cases where this Exact string could be valid or could not be valid depending on the policy in place. Maybe a better solution would be to split the test into two, one that does explicitly set a policy and one that explicitly removes it, and parameterize it based on allowed and not allowed scope/name constructions. 

The problem with setting it as you propose is that because this test is part of the unittest suite, and you can't set a value in the config (there is no config present in unittests)

mgajek-cern: There are config mocks available in `conftest.py` like `core_config_mock` and `file_config_mock` that can override configuration settings for tests, but let's keep it simple and go with your proposal to split into explicit test scenarios

E.g. 
https://github.com/rucio/rucio/blob/08962d88e487a0897d453d6fb375018c021f290b/tests/conftest.py#L669-L688

lib/rucio/common/didtype.py
voetberg: `extract_scope` also takes a "scopes" argument, there could be downstream plugins where these additional scopes are returned. I don't think we can say For Certain it only ever returns a tuple of two strings.

mgajek-cern: All callers expect exactly 2 strings (scope, name) when invoking `extract_scope(...)`. The type hint mismatch between Sequence[str] and actual usage patterns should/could be addressed in a separate ticket since it's a broader API consistency issue beyond the scope of this PR, but it's definitely worth tracking.

tests/rucio/common/test_didtype.py
voetberg: Based on the failures I've been seeing I just don't think this is how the unittest/file_config_mock interaction works in practice. I'm going to add some checks for logging statements so we can see if/when the fallback gets triggered in certain cases. Maybe the best way to handle this is to remove this particular example from the unit tests (as it triggers differently based on interactions) and fall back using the `test_policy_package` with different scope extraction algorithms.

voetberg: Quick examples of testing failures 

https://github.com/voetberg/rucio/actions/runs/18758317431/job/53515901716 -> Fails to pick up the algorithm set by `file_config_mock`because a "NoSectionError" is triggered before the mocked option can be picked up

https://github.com/rucio/rucio/actions/runs/18758318574/job/53515984424?pr=8135 -> There's a config, it will go to the default algorithm

mgajek-cern: Ran unit tests locally with `RUCIO_POLICY_PACKAGE=atlas_rucio_policy_package` set and again after unsetting the `RUCIO_POLICY_PACKAGE` environment variable. Following console logs:


```python
tests/rucio/common/test_didtype.py -v
============================================================= test session starts ==============================================================
platform linux -- Python 3.9.21, pytest-7.4.3, pluggy-1.5.0 -- /usr/bin/python3
cachedir: .pytest_cache
rootdir: /workspaces/rucio
configfile: pyproject.toml
plugins: cov-5.0.0, xdist-3.5.0
collected 45 items                                                                                                                             

tests/rucio/common/test_didtype.py::TestDIDType::test_did_type_success[empty_did] PASSED                                                 [  2%]
...

=================================================================== FAILURES ===================================================================
_____________________________________________________ TestDIDType.test_fallback_str_policy _____________________________________________________

self = <test_didtype.TestDIDType object at 0xffff7d8551f0>, caplog = <_pytest.logging.LogCaptureFixture object at 0xffff7d8112e0>

    @pytest.mark.noparallel(reason='We temporarily remove the config file.')
    @remove_config
    def test_fallback_str_policy(self, caplog):
        with caplog.at_level(logging.DEBUG):
            with pytest.raises(DIDError, match='Error using DID type\nDetails: Object construction from non-splitable string is ambigious'):
>               DID('non.implicit.single.string')
E               Failed: DID NOT RAISE <class 'rucio.common.exception.DIDError'>

tests/rucio/common/test_didtype.py:91: Failed
=========================================================== short test summary info ============================================================
FAILED tests/rucio/common/test_didtype.py::TestDIDType::test_fallback_str_policy - Failed: DID NOT RAISE <class 'rucio.common.exception.DIDError'>
========================================================= 1 failed, 44 passed in 0.56s ===
```

---

The test `test_fallback_str_policy` fails consistently across:
- Local development
- CI Integration tests/auto-tests

This suggests the fallback scenario being tested is unreachable in practice. The system appears to always have default scope extraction mechanisms available.

**Recommendation:** Remove `test_fallback_str_policy` entirely since it's testing dead code and keep `test_implemented_str_policy` which demonstrates actual functionality.

Are there any environment considerations for ensuring consistent unit test behavior across development/locally and CI?

lib/rucio/common/didtype.py
mgajek-cern: Draft GH issue has been created: https://github.com/rucio/rucio/issues/8152

mgajek-cern (dismissed):
Code coverage is at `93.75%` for `extract_scope` function, which looks reasonable. It’d be good to have someone else verify, only 1 branch appears to be uncovered:

image

lib/rucio/common/didtype.py
```diff
@@ -126,15 +128,19 @@ def _did_from_str(self, did: str) -> None:
...
+        try:
+            self.scope, self.name = extract_scope(did)
+        except (ImportError, InvalidAlgorithmName, ConfigNotFound, NoSectionError) as e:  # Only use when the policy can not be found
+            logging.debug(f"Failure using exact_scope policy for '{did}': {type(e).__name__} - Using fallback.")
+            did_parts = did.split(DID.SCOPE_SEPARATOR, 1)
+            if len(did_parts) == 1:
+                self.name = did
+                self._update_implicit_scope()
+                if not self.has_scope():
+                    raise DIDError('Object construction from non-splitable string is ambigious')
```
rdimaio: These typos were already there before the PR, but worth fixing here
```suggestion
                    raise DIDError('Object construction from non-splittable string is ambiguous')
```

Although I think it's still a bit cryptic. Maybe it could be changed to something like "Could not parse scope from string"?

tests/rucio/common/test_didtype.py
```diff
@@ -63,10 +64,6 @@ def test_did_type_success(self, input_did, expected_scope, expected_name):
         assert input_did.scope == expected_scope
         assert input_did.name == expected_name
 
-    def test_non_implicit_single_string(self):
```
rdimaio, line 66: I think we should still have a test that raises this error to test this branch

lib/rucio/common/didtype.py
voetberg: I don't love it either. May I instead suggest "Could not parse scope from did string {str} - fallback policy expects only one ':'"

lib/rucio/common/didtype.py
```diff
@@ -126,15 +128,19 @@ def _did_from_str(self, did: str) -> None:
...
-            self._update_implicit_scope()
-            if not self.has_scope():
-                raise DIDError('Object construction from non-splitable string is ambigious')
-        else:
-            self.scope = did_parts[0]
-            self.name = did_parts[1]
+        try:
+            self.scope, self.name = extract_scope(did)
+        except (ImportError, InvalidAlgorithmName, ConfigNotFound, NoSectionError) as e:  # Only use when the policy can not be found
+            logging.debug(f"Failure using exact_scope policy for '{did}': {type(e).__name__} - Using fallback.")
```
rdimaio:
```suggestion
            logging.debug(f"Failure using extract_scope policy for '{did}': {type(e).__name__} - Using fallback.")
```

lib/rucio/common/didtype.py
rdimaio: Sounds good, maybe instead of hardcoding `:` we can point to `DID.SCOPE_SEPARATOR` instead

rdimaio (approved):

Issue #8133, opened 2025-10-18
Transfer with tokens is not being received by the receiver / not being submitted properly


rdimaio:
### Description

Opening this as a separate issue from https://github.com/rucio/rucio/issues/7750 as most of the tests failing under #7750 were fixed by properly migrating to the FTS >3.14.0 ActiveMQ daemon, but `test_transfer_with_tokens` is still failing. 

Trying to debug this, it seems like the transfer is not being properly submitted. In the FTS container's logs, there is no folder created for the WEB1->XRD5 transfers (while e.g. `xrd3__xrd4` is created by `test_receiver_archiving)`, and the `fts3server.log` and `msg.log` (the ActiveMQ log flie) don't report any activity, even though they do for `test_receiver_archiving`.

The overall test setup is the same as `test_receiver_archiving`, while the RSEs used for SRC and DST are the same as the ones used in `test_bittorrent_submission`, and both of these test work fine. I think `test_receiver_archiving` is the best one to take into account as reference for the most similar one that passes. For some reason, it seems that using WEB1 as the RSE is causing this overall transfer to not work.

The full logs are:
```
=================================== FAILURES ===================================
_ test_transfer_with_tokens[file_config_mock{'oidc': {'admin_issuer': 'indigoiam'}}] _
[gw0] linux -- Python 3.9.21 /opt/venv/bin/python
tests/test_conveyor.py:1665: in test_transfer_with_tokens
    assert False  # Waited too long; fail the test
E   assert False
------------------------------ Captured log call -------------------------------
DEBUG    root:uploadclient.py:347 Num. of files that upload client is processing: 1
DEBUG    urllib3.connectionpool:connectionpool.py:1021 Starting new HTTPS connection (1): rucio:443
DEBUG    urllib3.connectionpool:connectionpool.py:474 https://rucio:443 "GET /rses/?expression=WEB1 HTTP/1.1" 200 None
DEBUG    root:uploadclient.py:375 Input validation done.
INFO     root:uploadclient.py:383 Preparing upload for file CQKZHNFLCB
DEBUG    urllib3.connectionpool:connectionpool.py:474 https://rucio:443 "GET /rses/WEB1/attr/ HTTP/1.1" 200 226
DEBUG    root:uploadclient.py:427 wan domain is used for the upload
DEBUG    root:uploadclient.py:689 Registering file
[...]

PR #7899, opened 2025-08-01
Fix conveyor tests and migrate to FTS >3.14.0 ActiveMQ daemon

rdimaio:
Fix #7750. The token test requires more work and should be investigated under https://github.com/rucio/rucio/issues/8133.

PR #8136, opened 2025-10-20
fts3: allow unmanaged token option for FTS transfers #7761

### LINKED ISSUE #7761: FTS3 3.14 introduces unmanaged-token  options
### Description

https://fts3-docs.web.cern.ch/fts3-docs/docs/install/upgrades/3.14.html#explicit-managed--unmanaged-tokens

With FTS3 3.14, we can send parameter to say FTS not to do token exchange/refresh.

### Motivation

This will be good for experiment using long lifetime token. 

### Change

We need to add optional parameter as: https://gitlab.cern.ch/fts/fts-rest-flask/-/blob/develop/src/fts3rest/fts3rest/lib/api/submit_schema.py#L85

This parameter choice should be taken from rse attributes or should be at the cfg level ? ( to be discussed)

Should add the logic here: https://github.com/rucio/rucio/blob/master/lib/rucio/transfertool/fts3.py#L401

panta-123:
- Enable setting of the FTS "unmanaged_tokens" job parameter when token auth is active and conveyor.fts_unmanaged_token is true.
- Widen job_params typing from dict[str, str] to dict[str, Any] without needing type-ignore for new paramter which is boolean
  ref: https://gitlab.cern.ch/fts/fts-rest-flask/-/blob/develop/src/fts3rest/fts3rest/lib/api/submit_schema.py#L85
- Update tests and local FTS dev config to reflect the behavior.
- Test is currently skipped using
```
@pytest.mark.skip(reason="Pending https://cern.service-now.com/service-portal?id=ticket&table=incident&n=INC4506150")
```


tests/test_conveyor.py
```diff
@@ -1664,6 +1664,58 @@ def _perform_request_update(self, msg, *, session=None, logger=logging.log):
             RECEIVER_GRACEFUL_STOP.clear()
 
 
+@pytest.mark.skip(reason="Pending https://cern.service-now.com/service-portal?id=ticket&table=incident&n=INC4506150")
```
rdimaio: We should merge https://github.com/rucio/rucio/pull/7899 first, so that we can remove this skip marker

rdimaio: #7899 merged, this skip can be removed

panta-123: okay will remove that in next commit.

panta-123: seems passed.
```
tests/test_conveyor.py::test_transfer_with_unmanaged_tokens[file_config_mock{'oidc': {'admin_issuer': 'indigoiam'}, 'conveyor': {'fts_unmanaged_token': 'true'}}] 
[gw0] [ 88%] PASSED tests/test_conveyor.py::test_transfer_with_unmanaged_tokens[file_config_mock{'oidc': {'admin_issuer': 'indigoiam'}, 'conveyor': {'fts_unmanaged_token': 'true'}}] 
```

etc/docker/dev/fts/fts3restconfig
```diff
@@ -19,6 +19,7 @@ Alias=rucio/fts
 MonitoringMessaging=false
 
 OAuth2 = True
+AllowNonManagedTokens = True
```
rdimaio, line 22: Note: we should also document this in the Rucio documentation

lib/rucio/transfertool/fts3.py
```diff
@@ -998,7 +998,7 @@ def group_into_submit_jobs(self, transfer_paths: "list[list[DirectTransfer]]") -
         )
         return jobs
 
-    def _file_from_transfer(self, transfer: "DirectTransfer", job_params: dict[str, str]) -> dict[str, Any]:
```
rdimaio, line 1001: nit: this could be `dict[str, Union[str, bool]]` to be more precise

lib/rucio/transfertool/fts3.py
```diff
@@ -1054,7 +1054,7 @@ def _file_from_transfer(self, transfer: "DirectTransfer", job_params: dict[str,
 
         return t_file
 
-    def submit(self, transfers: "Sequence[DirectTransfer]", job_params: dict[str, str], timeout: Optional[int] = None) -> str:
+    def submit(self, transfers: "Sequence[DirectTransfer]", job_params: dict[str, Any], timeout: Optional[int] = None) -> str:
```
rdimaio: nit: this could be `dict[str, Union[str, bool]]` to be more precise

lib/rucio/transfertool/transfertool.py
```diff
@@ -176,7 +176,7 @@ def group_into_submit_jobs(self, transfer_paths: "Iterable[list[DirectTransfer]]
         pass
 
     @abstractmethod
-    def submit(self, transfers: "Iterable[DirectTransfer]", job_params: dict[str, str], timeout: Optional[int] = None) -> str:
+    def submit(self, transfers: "Iterable[DirectTransfer]", job_params: dict[str, Any], timeout: Optional[int] = None) -> str:
```
rdimaio: nit: this could be `dict[str, Union[str, bool]]` to be more precise

tests/test_conveyor.py
```diff
@@ -1664,6 +1664,57 @@ def _perform_request_update(self, msg, *, session=None, logger=logging.log):
...
+@skip_rse_tests_with_accounts
+@pytest.mark.needs_iam
+@pytest.mark.noparallel(groups=[NoParallelGroups.XRD, NoParallelGroups.SUBMITTER, NoParallelGroups.RECEIVER])
+@pytest.mark.parametrize("file_config_mock", [
+    {"overrides": [
+        ('oidc', 'admin_issuer', 'indigoiam'),
+        ('conveyor', 'fts_unmanaged_token', 'true')
+    ]},
+], indirect=True)
+def test_transfer_with_unmanaged_tokens(vo, did_factory, root_account, caches_mock, file_config_mock):
```
rdimaio: Looks like this test is exactly the same as `test_transfer_with_tokens`, which right now is being skipped with `@pytest.mark.skip(reason="Pending https://github.com/rucio/rucio/issues/8133")` (https://github.com/rucio/rucio/blob/master/tests/test_conveyor.py#L1627) - once we stop skipping that test, we can merge these two tests and just parametrize the `file_config_mock`.

Sidenote: I wonder why this test works and `test_transfer_with_tokens` doesn't, since FTS upgraded to 3.14. Should we even still keep `test_transfer_with_tokens`?

rdimaio (commented):
LGTM but I think it's best to also get approval from a token component lead.

tests/test_conveyor.py
panta-123: @rdimaio 
This is confusing. Its the same test . with unmanaged token it just say fts not to  exchange token fro refresh.
Have you tried the test_transfer_with_tokens test locally in dev enviornment.

rdimaio: Yeah, for me it was failing both locally and remotely. I put the logs here https://github.com/rucio/rucio/issues/8133

rdimaio: note: fixed by https://github.com/rucio/rucio/pull/8148

rdimaio: We can probably merge the two tests now and just parametrize, as mentioned in the comment above

panta-123: merged into one

etc/docker/dev/fts/fts3restconfig
panta-123: Yes will be added to doc.

tests/test_conveyor.py
```diff
@@ -1627,7 +1627,15 @@ def on_submit(file):
...
-    {"overrides": [('oidc', 'admin_issuer', 'indigoiam')]},
+    # Managed token case (default)
+    {"overrides": [
+        ('oidc', 'admin_issuer', 'indigoiam'),
+    ]},
+    # Unmanaged token case
+    {"overrides": [
+        ('oidc', 'admin_issuer', 'indigoiam'),
+        ('conveyor', 'fts_unmanaged_token', 'true')
+    ]},
```
rdimaio, line 1638: Instead of the comments, you can also set custom IDs for both test cases to make debugging a bit easier in the pytest output, e.g. https://github.com/rucio/rucio/blob/0089ea405be6a4c3add7e72359e740248d8e36cb/tests/test_hermes.py#L94-L97

rdimaio (approved):

rdimaio (commented):
Adding blocking comment to wait until token component lead approval before merging

dchristidis (commented):
This requires a bit more consideration because a global configuration may not be fit for purpose. The use case that comes to mind is needing unmanaged tokens for disk transfers but managed tokens for tape.

panta-123:
> This requires a bit more consideration because a global configuration may not be fit for purpose. The use case that comes to mind is needing unmanaged tokens for disk transfers but managed tokens for tape.

Tape token is long way from now and plan is still being worked out on FTS and storage side . Currently its only disk and only on davs protocol. So that can't be taken into account in this PR. At least this PR lets user have some experience on the choices. And any changes on other future possibilities should be addressed at that point.

Geogouz:
The concrete `MockTransfertool` would still declare `dict[str, str]` in:

https://github.com/rucio/rucio/blob/9526397338e4e74900a2000a8d8390b90f458c03/lib/rucio/transfertool/mock.py#L75-L76

etc/docker/dev/fts/fts3restconfig
Geogouz: Docstrings should also reflect the new behavior.

panta-123:
> This requires a bit more consideration because a global configuration may not be fit for purpose. The use case that comes to mind is needing unmanaged tokens for disk transfers but managed tokens for tape.

@dchristidis , What do you suggest here?
- Make it rse attribute.
    -> But then we need to do that for all the disk rse.
- make it as global config. 
   -> put a filter on rse type when TAPE token is allowed .
   -> And also add documentation for this.

Geogouz:
Can you also rebase this please?

bziemons:
> This requires a bit more consideration because a global configuration may not be fit for purpose. The use case that comes to mind is needing unmanaged tokens for disk transfers but managed tokens for tape.

@dchristidis can you bring this discussion to the issue? Because there seems to be the exact opposite opinion https://github.com/rucio/rucio/issues/7761#issuecomment-3028161660

panta-123:
@dchristidis , this PR is hanging for some time. I would like to have this done . Please make your argument in the issue as mentioned. As we like to start using this and would not like to do random patch.

bari12:
I am wondering about the defaults here. Should the default be that we expect FTS to refresh or not. This PR implies that we expect FTS to refresh, I am not sure if this is the way to go.

Besides, we should make progress here. I get @dchristidis point that this might need more consideration for tape, but until we have more information there, perhaps we should unblock things with this PR.

PR #8140, opened 2025-10-21
Monitoring & Traces: Update `TIME_ENTRY` schema pattern to better support fractional seconds and timezones; #8139

### LINKED ISSUE #8139: `TIME_ENTRY` pattern doesn't fully cover `traceTimeentry`
### Description

Currently, "traceTimeentry" is prepared as a timezone‑aware UTC datetime:

https://github.com/rucio/rucio/blob/d59c52d31d8c3e14e02a7b622195cbc217c14773/lib/rucio/gateway/trace.py#L43

serialised via:

https://github.com/rucio/rucio/blob/d59c52d31d8c3e14e02a7b622195cbc217c14773/lib/rucio/core/trace.py#L307-L309

The schema that validates trace payloads, after the seconds it allows one arbitrary character and then digits only to the end.

https://github.com/rucio/rucio/blob/d59c52d31d8c3e14e02a7b622195cbc217c14773/lib/rucio/common/schema/generic.py#L168-L172

which means that something like this fails:

`2025-10-21T13:41:13.842589+00:00`

A pattern like the following one is better since it allows optional fractional seconds and either `Z` or a signed offset like `+00:00`:

`^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(?:\.\d+)?(?:Z|[+-]\d{2}:\d{2})?$`

Or if we want to always require timezones, something like:

`^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(?:\.\d{1,6})?(?:Z|[+-]\d{2}:\d{2})$`

### Steps to reproduce

.

### Rucio Version

Rucio 38.4.0

### Additional Information

_No response_

Geogouz:
Fixes #8139

### DIFF
--- a/lib/rucio/common/schema/generic.py
+++ b/lib/rucio/common/schema/generic.py
@@ -168,7 +168,7 @@ SPLIT_CONTAINER = {"description": "Rule split container mode",
 TIME_ENTRY = {
     "description": "Datetime, ISO 8601",
     "type": "string",
-    "pattern": r'^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.\d*$'
+    "pattern": r'^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(?:\.\d+)?(?:Z|[+-]\d{2}:\d{2})?$'
 }
 
 IP = {


rdimaio (approved):

rdimaio (commented):
AFAIK traces aren't particularly covered by tests at the moment, but would be good if we added a test for this to ensure that it now supports both the "previous" timestamps it supported as well as the fractional seconds and timezones

PR #8143, opened 2025-10-22
End-matching DID names allowed for listing rules

### LINKED ISSUE #7121: Adding a matching pattern for DIDs in list-rules command
### Description

The list-rules can list rules for a single DID, an entire account or an entire subscription. We would like to add DIDs matching pattern to it. In order to avoid heavy DB query, we suggest below:

`r list-rules --account register_service  raw:Dataset/LSSTCam/202408'*' `

### Motivation

Some of the service accounts have a lot of rules/files and we are only interested to some of them during debugging or a quick checking. Adding a did pattern to limit to the output will avoid the additional filter that the client has to do. In addition, it will help Rucio to reduce the output and speed up.   

### Change

The changes will be in the Client API to allow the did matching pattern and add a DID like clause to the DB query. The changes should be minimum as Rucio already does the most part of this.

voetberg:
Closes #7121

### DIFF
--- a/lib/rucio/cli/bin_legacy/rucio.py
+++ b/lib/rucio/cli/bin_legacy/rucio.py
@@ -1449,7 +1449,6 @@ def list_rules(args, client, logger, console, spinner):
 
     List rules.
     """
-
     if cli_config == 'rich':
         spinner.update(status='Fetching rules')
         spinner.start()
@@ -1467,22 +1466,26 @@ def list_rules(args, client, logger, console, spinner):
             rules.append(client.get_replication_rule(rule_id))
     elif args.did:
         scope, name = get_scope(args.did, client)
-        meta = client.get_metadata(scope=scope, name=name)
-        rules = client.list_did_rules(scope=scope, name=name)
+        if not name.endswith('*'):
+            meta = client.get_metadata(scope=scope, name=name)
+        else:
+            meta = {}
+        rules = client.list_replication_rules(filters={'scope': scope, 'name': name})
         try:
             next(rules)
-            rules = client.list_did_rules(scope=scope, name=name)
+            rules = client.list_replication_rules(filters={'scope': scope, 'name': name})
         except StopIteration:
             rules = []
+            did_type = meta.get('did_type')
             # looking for other rules
-            if meta['did_type'] == 'CONTAINER':
+            if did_type == 'CONTAINER':
                 for dsn in client.list_content(scope, name):
-                    rules.extend(client.list_did_rules(scope=dsn['scope'], name=dsn['name']))
+                    rules.extend(client.list_replication_rules(filters={"scope": dsn['scope'], "name": dsn['name']}))
                 if rules:
                     print('No rules found, listing rules for content')
-            if meta['did_type'] == 'DATASET':
+            if did_type == 'DATASET':
                 for container in client.list_parent_dids(scope, name):
-                    rules.extend(client.list_did_rules(scope=container['scope'], name=container['name']))
+                    rules.extend(client.list_replication_rules(filters={"scope": container['scope'], "name": container['name']}))
                 if rules:
                     print('No rules found, listing rules for parents')
     elif args.rule_account:
--- a/lib/rucio/core/rule.py
+++ b/lib/rucio/core/rule.py
@@ -1015,6 +1015,14 @@ def list_rules(
                 value = DIDType(value)
             elif key == 'grouping' and isinstance(value, str):
                 value = RuleGrouping(value)
+            elif key == 'name':
+                # First check if it is a real DID or a wildcard
+                if "*" in value:
+                    # Get the matching dids
+                    value = value.replace('*', '%')
+                    stmt = stmt.where(models.ReplicationRule.name.like(value))
+                    continue
+
             stmt = stmt.where(getattr(models.ReplicationRule, key) == value)
 
     try:
--- a/tests/test_bin_rucio.py
+++ b/tests/test_bin_rucio.py
@@ -641,6 +641,23 @@ def test_move_rule_with_arguments(rse_factory, did_factory, rucio_client):
     assert new_rule_source_replica_expression == rule_info["source_replica_expression"]
 
 
+@pytest.mark.parametrize('cmd', ["rucio list-rules mock:file_*", "rucio rule list --did mock:file_*"])
+def test_list_rules_wildcard(cmd, rse_factory, did_factory, mock_scope, root_account, rucio_client):
+
+    rse, _ = rse_factory.make_posix_rse()
+    rse_1, _ = rse_factory.make_posix_rse()
+    n_files = 1
+    dataset = did_factory.upload_test_dataset(rse_name=rse, scope=mock_scope.external, size=1, nb_files=n_files)
+
+    [rule_id] = rucio_client.add_replication_rule(
+        dids=[{'scope': d['did_scope'], 'name': d['did_name']} for d in dataset], account=root_account.external, copies=1, rse_expression=rse_1)
+
+    exitcode, out, err = execute(cmd)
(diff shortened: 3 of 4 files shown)


lib/rucio/core/rule.py
```diff
@@ -1015,6 +1015,14 @@ def list_rules(
                 value = DIDType(value)
             elif key == 'grouping' and isinstance(value, str):
                 value = RuleGrouping(value)
+            elif key == 'name':
+                # First check if it is a real DID or a wildcard
+                if "*" in value:
+                    # Get the matching dids
+                    value = value.replace('*', '%')
+                    stmt = stmt.where(models.ReplicationRule.name.like(value))
+                    continue
```
mgajek-cern, line 1024: **Should we support `?` as a single-character wildcard and handle literal `%`/`_` in DIDs?**

Suggested implementation:
```python
elif key == 'name':
    if "*" in value or "?" in value:
       # Prevent overly broad queries
        if value.strip() in ("*", "?", "%", "_"):
            raise ValueError("Pattern too broad - please be more specific")
        # Escape backslashes first to prevent double-escaping
        value = value.replace('\\', '\\\\')
        # Escape SQL wildcards so literal % and _ are treated as normal characters
        value = value.replace('%', r'\%').replace('_', r'\_')
        # Convert shell-style wildcards to SQL wildcards
        value = value.replace('*', '%').replace('?', '_')
        stmt = stmt.where(models.ReplicationRule.name.like(value, escape='\\'))
        continue
```

**Example:**
```python
value = "file?*_%"
```

After transformations: `"file_%\_\%"`

**SQL LIKE interpretation:**
- `file` = literal "file"
- `_` = any single character (from `?`)
- `%` = any sequence (from `*`)
- `\_` = literal underscore (escaped)
- `\%` = literal percent (escaped)

**Result:** Matches strings like `"fileA123_data%"` where `A` is any character and `123` is any sequence.

lib/rucio/cli/bin_legacy/rucio.py
```diff
@@ -1467,22 +1466,26 @@ def list_rules(args, client, logger, console, spinner):
             rules.append(client.get_replication_rule(rule_id))
     elif args.did:
         scope, name = get_scope(args.did, client)
-        meta = client.get_metadata(scope=scope, name=name)
-        rules = client.list_did_rules(scope=scope, name=name)
+        if not name.endswith('*'):
```
mgajek-cern, line 1469: This assumes wildcards only appear at the end. What if someone queries `file_*_test` or `*_important`? We might want:

```suggestion
     if "*" not in name:  # No wildcards at all
```

tests/test_bin_rucio.py
```diff
@@ -641,6 +641,23 @@ def test_move_rule_with_arguments(rse_factory, did_factory, rucio_client):
...
+
+    [rule_id] = rucio_client.add_replication_rule(
+        dids=[{'scope': d['did_scope'], 'name': d['did_name']} for d in dataset], account=root_account.external, copies=1, rse_expression=rse_1)
+
+    exitcode, out, err = execute(cmd)
+    assert rule_id in out
+    assert 'ERROR' not in err
+    assert exitcode == 0
+
+
```
mgajek-cern, line 660: Should following edge cases be tested as well:

```
    # Test multiple wildcards e.g. "file_*_test_*"
    # Test wildcard at beginning e.g. "*_important"
    # Test wildcard in middle e.g. "file_*_2024"
    # Test no matches e.g. "nonexistent_*"
```

lib/rucio/core/rule.py
```diff
@@ -1015,6 +1015,14 @@ def list_rules(
                 value = DIDType(value)
             elif key == 'grouping' and isinstance(value, str):
                 value = RuleGrouping(value)
+            elif key == 'name':
+                # First check if it is a real DID or a wildcard
+                if "*" in value:
+                    # Get the matching dids
+                    value = value.replace('*', '%')
+                    stmt = stmt.where(models.ReplicationRule.name.like(value))
```
mgajek-cern, line 1023: Should we add a `LIMIT` to wildcard queries to reduce excessive resource (like CPU memory, or I/O) consumption for an extended time leading to slow performance or system slowdowns, e.g. 

```suggestion
                    stmt = stmt.where(models.ReplicationRule.name.like(value))
                    stmt = stmt.limit(1000)  # Prevent runaway queries
```

lib/rucio/core/rule.py
```diff
@@ -1015,6 +1015,14 @@ def list_rules(
                 value = DIDType(value)
             elif key == 'grouping' and isinstance(value, str):
                 value = RuleGrouping(value)
+            elif key == 'name':
+                # First check if it is a real DID or a wildcard
```
mgajek-cern, line 1019: To be more specific we could instead have following comment:

```suggestion
               # Check if name contains wildcard pattern and convert to SQL LIKE syntax
```

lib/rucio/core/rule.py
```diff
@@ -1015,6 +1015,14 @@ def list_rules(
...
+                if "*" in value:
```
mgajek-cern, line 1020: `LIKE` queries with leading wildcards (%something) can't use indexes efficiently, e.g.

```sql
-- Assume an indexed column `name`

-- 1. Wildcard at the end (can use index efficiently)
SELECT * FROM files WHERE name LIKE 'report%';

-- 2. Wildcard at the beginning (index usually not used, slower)
SELECT * FROM files WHERE name LIKE '%report';
``` 


We could add following comment:

```
# Note: Wildcards at the beginning of the pattern (*something) 
# may result in slower queries as they cannot use database indexes
```

mgajek-cern (commented):
**NOTE:** Reviewing this PR based on the understanding that wildcards can appear anywhere (in the beginning, middle, or end) according to the issue.

Here are a few remarks/questions/findings:

voetberg:
> **NOTE:** Reviewing this PR based on the understanding that wildcards can appear anywhere (in the beginning, middle, or end) according to the issue.
> 
> Here are a few remarks/questions/findings:

Sorry for the misleading issue title. I'm using "wildcard" as a shorthand for "match the end of a DID pattern" - as the issue suggests.

mgajek-cern:
> Sorry for the misleading issue title. I'm using "wildcard" as a shorthand for "match the end of a DID pattern" - as the issue suggests.

Thanks for clarifying. Before someone else continues with the review of the simpler end-matching implementation, can you confirm this scope has been agreed upon with the issue creator (@yuyiguo)? 

Just want to make sure we're all aligned on the requirements (end-matching only vs. full wildcard support) before the next reviewer dives in.

voetberg:
> > Sorry for the misleading issue title. I'm using "wildcard" as a shorthand for "match the end of a DID pattern" - as the issue suggests.
> 
> Thanks for clarifying. Before someone else continues with the review of the simpler end-matching implementation, can you confirm this scope has been agreed upon with the issue creator (@yuyiguo)?
> 
> Just want to make sure we're all aligned on the requirements (end-matching only vs. full wildcard support) before the next reviewer dives in.

I talked with Yuyi IRL and we agreed that end-matching is the better option. I'll change the name of the PR appropriately

bari12:
I think this needs to support wildcards anywhere in the name, not just the end. 
I understand that Yuyis use-case does not require it, but if we add this functionality, the assumption will be that it works the same as a `rucio did list` - where we do allow it at any point in the pattern (except the beginning as @mgajek-cern pointed out) - Dealing with this differently in `rucio rule list` is a mistake I think, it should be equivalent.

lib/rucio/core/rule.py
bziemons: this would have been a discussion for the issue, but since this is already the PR I see your comment as out of scope. Open a new issue if you really need this feature.

PR #8162, opened 2025-10-30
feat(rest-api): add query parameter endpoint for listing accounts by identity

### LINKED ISSUE #8159: REST API: Add query parameter endpoint for listing accounts by identity to support   OIDC identities with slashes
### Description

The existing endpoint `GET /identities/<identity_key>/<type>/accounts` fails for
  OIDC identities because the identity key contains forward slashes in the issuer URL.

  **Example OIDC identity:**
  SUB=mayank, ISS=https://auth.cern.ch/auth/realms/cern

  **Why it fails:**
  - The OIDC identity format contains special characters that are problematic in URL
  paths: forward slashes `/`, commas `,`, equals signs `=`, and colons `:`
  - Flask's default route parameter converter `<identity_key>` does not match forward
  slashes - it treats `/` as a path delimiter
  - Even though the URL encodes slashes as `%2F`, Flask/Werkzeug automatically decodes
   them before routing, breaking the pattern match
  - The route fails regardless of Apache's `AllowEncodedSlashes` setting
  (on/off/nodecode)
  - **The OIDC identity format is fundamentally not suitable as a path parameter**
  when it appears in the middle of a URL path (before `/accounts`)

### Steps to reproduce

1. **Create an account with an OIDC identity mapping:**
     ```python
     from rucio.core.identity import add_account_identity, add_identity
     from rucio.db.sqla.constants import IdentityType

     identity_key = 'SUB=mayank, ISS=https://auth.cern.ch/auth/realms/cern'
     add_identity(identity_key, IdentityType.OIDC, email='test@example.com')
     add_account_identity(identity_key, IdentityType.OIDC,
  InternalAccount('testaccount'),
                         email='test@example.com')
2.  Attempt to query the existing endpoint:  # URL-encode the identity key
  
  ```
  IDENTITY="SUB%3Dmayank%2C%20ISS%3Dhttps%3A%2F%2Fauth.cern.ch%2Fauth%2Frealms%2Fcern"
  
  curl -H "X-Rucio-Auth-Token: $TOKEN" \
      "https://rucio-server/identities/${IDENTITY}/OIDC/accounts"
  ```
3. Observe the failure:
    - Returns 404 Not Found
    - Flask cannot match the route because the decoded slashes in the identity key
  break the path parsing
    - The decoded path becomes: /identities/SUB=mayank, 
  ISS=https://auth.cern.ch/auth/realms/cern/OIDC/accounts

### Rucio Version

38.5.0

### Additional Information

Add a new query parameter-based endpoint that works alongside the existing path
  parameter endpoint:

  **New endpoint:** `GET /identities/accounts?identity_key=<value>&type=<value>`

maany:
OIDC identities follow the format `SUB=<subject>, ISS=<issuer_url>` (e.g., `SUB=mayank, ISS=https://auth.cern.ch/auth/realms/cern`). The existing endpoint `GET /identities/<identity_key>/<type>/accounts` cannot handle these identities because Flask's route parameter converter treats forward slashes as path delimiters, even when URL-encoded. This is a fundamental limitation of using complex strings with special characters (/, =, :, ,) as path parameters in the middle of a URL.

### Changes

Added a new endpoint that accepts parameters via query string instead of path parameters:

**New route:** `GET /identities/accounts?identity_key=<value>&type=<value>`

### Implementation details:
- Created `AccountsByIdentity` view class in `lib/rucio/web/rest/flaskapi/v1/identities.py`
- Validates that both `identity_key` and `type` parameters are provided (returns 400 if missing)
- Returns the same JSON response format as the existing endpoint
- The existing `/<identity_key>/<type>/accounts` route remains unchanged for backward compatibility

### Why query parameters work:
Query parameters are not affected by Apache's `AllowEncodedSlashes` directive, which only applies to URL paths. This solution works reliably across all Apache configurations (on/off/nodecode).

Fixes #8159

### DIFF
--- a/lib/rucio/web/rest/flaskapi/v1/identities.py
+++ b/lib/rucio/web/rest/flaskapi/v1/identities.py
@@ -12,9 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from flask import Flask, jsonify, request
+from flask import Flask, Response, jsonify, request
 
 from rucio.common.constants import HTTPMethod
+from rucio.db.sqla.constants import IdentityType
 from rucio.gateway.identity import add_account_identity, add_identity, list_accounts_for_identity
 from rucio.web.rest.flaskapi.authenticated_bp import AuthenticatedBlueprint
 from rucio.web.rest.flaskapi.v1.common import ErrorHandlingMethodView, check_accept_header_wrapper_flask, generate_http_error_flask, response_headers
@@ -263,6 +264,67 @@ class Accounts(ErrorHandlingMethodView):
         return jsonify(accounts)
 
 
+class AccountsByIdentity(ErrorHandlingMethodView):
+    """ Retrieve list of accounts mapped to an identity using query parameters. """
+
+    @check_accept_header_wrapper_flask(['application/json'])
+    def get(self) -> Response:
+        """
+        ---
+        summary: List accounts by identity
+        description: "List all accounts mapped to an identity using query parameters. This endpoint supports identity keys containing special characters like slashes."
+        tags:
+          - Identity
+        parameters:
+        - name: identity_key
+          in: query
+          description: "Identity string. For OIDC identities, use format: SUB=<sub>, ISS=<issuer_url>"
+          schema:
+            type: string
+          required: true
+        - name: type
+          in: query
+          description: "Identity type."
+          schema:
+            type: string
+            enum: [X509, GSS, USERPASS, SSH, SAML, OIDC]
+          required: true
+        responses:
+          200:
+            description: "OK"
+            content:
+              application/json:
+                schema:
+                  type: array
+                  items:
+                    type: object
+                    description: "Account for the identity."
+          400:
+            description: "Missing or invalid required parameter"
+          401:
+            description: "Invalid Auth Token"
+          406:
+            description: "Not acceptable"
+        """
+        identity_key = request.args.get('identity_key')
+        type_ = request.args.get('type')
+
+        if not identity_key:
+            return generate_http_error_flask(400, ValueError.__name__, 'identity_key parameter is required.')
+
+        if not type_:
+            return generate_http_error_flask(400, ValueError.__name__, 'type parameter is required.')
+
+        try:
+            IdentityType[type_.upper()]
+        except KeyError:
+            valid_types = [t.name for t in IdentityType]
+            return generate_http_error_flask(400, ValueError.__name__, f'Invalid identity type. Valid types are: {valid_types}')
+
+        accounts = list_accounts_for_identity(identity_key, type_)
+        return jsonify(accounts)
+
+
 def blueprint() -> AuthenticatedBlueprint:
     bp = AuthenticatedBlueprint('identities', __name__, url_prefix='/identities')
 
@@ -274,6 +336,8 @@ def blueprint() -> AuthenticatedBlueprint:
     bp.add_url_rule('/<account>/gss', view_func=gss_view, methods=[HTTPMethod.PUT.value])
     accounts_view = Accounts.as_view('accounts')
     bp.add_url_rule('/<identity_key>/<type>/accounts', view_func=accounts_view, methods=[HTTPMethod.GET.value])
+    accounts_by_identity_view = AccountsByIdentity.as_view('accounts_by_identity')
+    bp.add_url_rule('/accounts', view_func=accounts_by_identity_view, methods=[HTTPMethod.GET.value])
 
     bp.after_request(response_headers)
     return bp
--- a/tests/test_identity.py
+++ b/tests/test_identity.py
@@ -14,6 +14,7 @@
 
 import random
 import string
+from urllib.parse import urlencode
 
 import pytest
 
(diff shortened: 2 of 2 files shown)


lib/rucio/web/rest/flaskapi/v1/identities.py
```diff
@@ -263,6 +263,60 @@ def get(self, identity_key, type_):
...
+                  type: array
+                  items:
+                    type: object
+                    description: "Account for the identity."
+          400:
+            description: "Missing required parameter"
+          401:
+            description: "Invalid Auth Token"
+          406:
+            description: "Not acceptable"
```
rdimaio, line 307: Where are these two raised?

rdimaio (dismissed):
LGTM but also pinging @voetberg as component lead and @lobis as he worked on issues related to slash encoding recently, so he might have some insight on how to do this with a single API route?

lobis:
> OIDC identities follow the format `SUB=<subject>, ISS=<issuer_url>` (e.g., `SUB=mayank, ISS=https://auth.cern.ch/auth/realms/cern`). The existing endpoint `GET /identities/<identity_key>/<type>/accounts` cannot handle these identities because Flask's route parameter converter treats forward slashes as path delimiters, even when URL-encoded. This is a fundamental limitation of using complex strings with special characters (/, =, :, ,) as path parameters in the middle of a URL.

> LGTM but also pinging @voetberg as component lead and @lobis as he worked on issues related to slash encoding recently, so he might have some insight on how to do this with a single API route?

If I remember correctly this fails due to the Apache server configuration where encoded slashes are replaced by non-encoded ones and thus Flask splits on them, but if the Apache configuration is updated, this should work.

At some point we discussed enforcing this setting in Apache, to never decode encoded slahses, I think this is another reason why we should do it.

In any case I agree with the decission to use query parameters instead, it's much cleaner and prevents this issues from happening in the first place. In principle we may be able to make it work with the existing route but I am sure there are some corner cases that would make this a worse idea than just using query parameters.

Is there any reason why we should still offer the possibiltiy to select a setting for the Apache `AllowEncodedSlashes` setting? I propose to fix it as `NoDecode`, it should save us some trouble down the road.

Here is the relevant issue: https://github.com/rucio/containers/issues/392

lobis (dismissed):

lib/rucio/web/rest/flaskapi/v1/identities.py
```diff
@@ -263,6 +263,60 @@ def get(self, identity_key, type_):
...
+        """
+        identity_key = request.args.get('identity_key')
+        type_ = request.args.get('type')
```
voetberg, line 310: Does list accounts check the type or should that be done here?

lib/rucio/web/rest/flaskapi/v1/identities.py
```diff
@@ -263,6 +263,60 @@ def get(self, identity_key, type_):
...
+        parameters:
+        - name: identity_key
+          in: query
+          description: "Identity string. For OIDC identities, use format: SUB=<sub>, ISS=<issuer_url>"
+          schema:
+            type: string
+          required: true
+        - name: type
+          in: query
+          description: "Identity type (e.g., OIDC, X509, USERPASS, GSS)."
```
voetberg: This should be an enum type with the allowed ID types listed

lib/rucio/web/rest/flaskapi/v1/identities.py
voetberg: (Sorry for the non-descriptive comment) Does it check that the type is an allowed parameter or will checking for a non-existing ID type just result in an empty response?

voetberg (changes requested):
This needs some tests in `test_identities`

Geogouz:
Looks good. The only thing basically.. should we validate (i.e. that it maps to a known `IdentityType`) the user-supplied `type` param before calling `list_accounts_for_identity`? Right now, an unrecognized value will propagate into `IdentityType[id_type.upper()]`, raising an uncaught `KeyError` and so I guess then, a 500 instead of something like 4xx.

Geogouz:
> Looks good. The only thing basically.. should we validate (i.e. that it maps to a known `IdentityType`) the user-supplied `type` param before calling `list_accounts_for_identity`? Right now, an unrecognized value will propagate into `IdentityType[id_type.upper()]`, raising an uncaught `KeyError` and so I guess then, a 500 instead of something like 4xx.

Or do we trust always the client? If not, maybe we can add a test about it.

maany:
Please do not close as part of the review sprint. I will add the missing test, fix the commit messages and push updates as part of the current dev sprint

lib/rucio/web/rest/flaskapi/v1/identities.py
maany: Updated to use the IdentityType enum :)

lib/rucio/web/rest/flaskapi/v1/identities.py
maany: Aye, I am explicitly handling ID type validation now

tests/test_identity.py
```diff
@@ -130,3 +131,81 @@ def test_verify_x509_identity():
...
+        if config_get_bool('common', 'multi_vo', raise_exception=False, default=False):
+            vo = {'vo': get_vo()}
+        else:
+            vo = {}
+        account_name = account_name_generator()
+        account = InternalAccount(account_name, **vo)
+        identity_key = uuid()
+        email = identity_key + '@email.com'
+
+        add_account(account, AccountType.USER, email)
```
voetberg: All the `add_account` calls in tests can be replaced with the `random_account` fixture to cut down on all the try/except teardowns and using the `vo` fixture avoids calling the config for multivo

lib/rucio/web/rest/flaskapi/v1/identities.py
```diff
@@ -283,9 +284,10 @@ def get(self) -> Response:
           required: true
         - name: type
           in: query
-          description: "Identity type (e.g., OIDC, X509, USERPASS, GSS)."
+          description: "Identity type."
```
voetberg, line 287: Unsure the level of pedantic-ness on this, but should the docstring changes here be a different commit type? Unsure how we are handling these in conventional commits. Change if you so desire but I am mostly just flagging it.

voetberg:
Please squash your last two commits, 6dc80a422414aa0a764c01bf7a17bbc1ff2a45a9 just fixes an error in 3a5215ed57555e4faedc3fe0eff3d1227fd5c5bb no?

tests/test_identity.py
maany: Good point @voetberg! It shall be done :)

voetberg (approved):

lib/rucio/web/rest/flaskapi/v1/identities.py
dchristidis: No, they should not. Same for the type hint in the second commit.

Both are amendments to the first commit, which is brand-new code. As such, the three commits should be squashed together.

maany: @dchristidis it's done ;)

PR #8170, opened 2025-11-04
Rucio 39 major dependency upgrade

### LINKED ISSUE #8165: Dependency upgrade for Rucio 39

rdimaio:
Fix #8165


Geogouz (approved):
Since tests are passing with the new dependencies applied, I think we can proceed.

Couple non-critical comments (most are probably even out of current's PR scope):

- `urllib3==1.26.19, <2` is syntactically valid but I think redundant. Doesn't `==1.26.19` already implies the version is `<2`? So I don't think `<2` adds something effective here. Maybe if the goal is to allow updates within v1 but block v2, we should use `urllib3>=1.26.19,<2`.

- Maybe switch from dogpile.cache to dogpile-cache for consistency (currently we are mixing them) / or even standardize on the canonical dash form for distribution names in all files.

- Behavior will probably differ between `pip install rucio‑server` and `pip install -r requirements.server.txt` because `setuputil.py` uses upper bounds while runtime uses exact pins in .in .txt.

- The`server_requirements_table.install_requires` of `setuputil.py` includes boto (no upper limit) but this is not present in `requirements.server.in` so a server installed via the Rucio package will pull boto (boto2), but a server built via `requirements.server.txt` won’t. Maybe we can add it to `requirements.server.in` and pin it; otherwise remove it from setuputil.py (or move under an optional extra like s3-boto2).

- We have both `urllib3>=2.5` (client) and `urllib3==1.26.19` (server/dev) so if someone installs both client and dev requirements together under the same environment, pip will complain.

rdimaio:
> Since tests are passing with the new dependencies applied, I think we can proceed.
> 
> Couple non-critical comments (most are probably even out of current's PR scope):
> 
>     * `urllib3==1.26.19, <2` is syntactically valid but I think redundant. Doesn't `==1.26.19` already implies the version is `<2`? So I don't think `<2` adds something effective here. Maybe if the goal is to allow updates within v1 but block v2, we should use `urllib3>=1.26.19,<2`.
> 
I think it's useful in the context of performing automatic upgrades, the `<2` puts an upper bound on which versions an automated upgrade tool should go to (e.g. by using `pip-compile --upgrade`)

>     * Maybe switch from dogpile.cache to dogpile-cache for consistency (currently we are mixing them) / or even standardize on the canonical dash form for distribution names in all files.

The canonical name seems to be `dopgile.cache` (https://pypi.org/project/dogpile.cache/), so I would rather keep that one.

>     * Behavior will probably differ between `pip install rucio‑server` and `pip install -r requirements.server.txt` because `setuputil.py` uses upper bounds while runtime uses exact pins in .in .txt.
> 

That's more or less by design in Python dependency management, only the `requirements` files should really pin exact versions: https://stackoverflow.com/a/71386370/6668625

>     * The`server_requirements_table.install_requires` of `setuputil.py` includes boto (no upper limit) but this is not present in `requirements.server.in` so a server installed via the Rucio package will pull boto (boto2), but a server built via `requirements.server.txt` won’t. Maybe we can add it to `requirements.server.in` and pin it; otherwise remove it from setuputil.py (or move under an optional extra like s3-boto2).
 
This one is a bit of a leftover, I see it added here https://github.com/rucio/rucio/commit/4c4deb1eaeb07b22becb932a3fe275806e69b332 4 years ago and already there it was coexisting with `boto3`. I would discuss this in a separate PR

>     * We have both `urllib3>=2.5` (client) and `urllib3==1.26.19` (server/dev) so if someone installs both client and dev requirements together under the same environment, pip will complain.

I agree potentially but I see that this has been the case since at least release 37 https://github.com/rucio/rucio/commit/f2465f35f95dd95f62d6e5a4175ddbe0613ed9af, so I don't think it actually poses an issue

rdimaio:
Confirmed offline good to merge

PR #8176, opened 2025-11-04
Migrate core/account_counter and core/rse_counter away from session decorators, make `session` a required argument

### LINKED ISSUE #8174: `update_rules_for_lost_replica`: bytes from lock might be None, but are always passed to the account counter's decrease
### Description

https://github.com/rucio/rucio/blob/7201d7057a93d220011f142dd7c2b8ebc2d5976d/lib/rucio/core/rule.py#L2508

In the `ReplicaLock`, bytes is an `Optional[int]`. However, bytes is a required argument to `decrease`, so this would error out when the lock has no bytes associated with it.

I would keep the same functionality, but log a warning when the decrease here is called with a non-existing bytes argument

### Steps to reproduce

NA

### Rucio Version

_No response_

### Additional Information

_No response_

rdimaio:
Part of #6989

### DIFF
--- a/lib/rucio/core/rule.py
+++ b/lib/rucio/core/rule.py
@@ -2505,7 +2505,10 @@ def update_rules_for_lost_replica(
             rule.locks_replicating_cnt -= 1
         elif lock.state == LockState.STUCK:
             rule.locks_stuck_cnt -= 1
-        account_counter.decrease(rse_id=rse_id, account=rule.account, files=1, bytes_=lock.bytes, session=session)
+        if lock.bytes:
+            account_counter.decrease(rse_id=rse_id, account=rule.account, files=1, bytes_=lock.bytes, session=session)
+        else:
+            raise ValueError('Lock %s has no bytes value associated to it: account_counter not decreased for RSE %s, account %s' % (lock, rse_id, rule.account))
         if rule.state == RuleState.SUSPENDED:
             pass
         elif rule.state == RuleState.STUCK:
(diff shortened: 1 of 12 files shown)


lib/rucio/core/rule.py
```diff
@@ -2505,7 +2505,10 @@ def update_rules_for_lost_replica(
             rule.locks_replicating_cnt -= 1
         elif lock.state == LockState.STUCK:
             rule.locks_stuck_cnt -= 1
-        account_counter.decrease(rse_id=rse_id, account=rule.account, files=1, bytes_=lock.bytes, session=session)
+        if lock.bytes:
+            account_counter.decrease(rse_id=rse_id, account=rule.account, files=1, bytes_=lock.bytes, session=session)
+        else:
+            raise ValueError('Lock %s has no bytes value associated to it: account_counter not decreased for RSE %s, account %s' % (lock, rse_id, rule.account))
```
rdimaio, line 2511: Related: https://github.com/rucio/rucio/issues/8182

Geogouz: Yes, Replica locks still map `bytes` as `Optional[int]`, so the ORM will happily surface `None` when the database row has a NULL value. The cleanup logic then queues those values into `account_counter.decrease(...)` and calls `sum(...)` over them, both of which assume real integers. The PR looks solid overall, but this problem (which is out of current PR's scope) should be handled ideally. Maybe by normalising `None` to `0`.

Geogouz (approved):

PR #8181, opened 2025-11-05
Core: naming_convention, migrate away from session decorators, pass session as required parameter

rdimaio:
Part of #6989


Geogouz (approved):
Looks good.

PR #8184, opened 2025-11-05
Core: vo, migrate away from session decorators, make session a required parameter

rdimaio:
Part of #6989


Geogouz (approved):
Looks safe to me.

Regarding the `0.00% of diff hit (target 33.96%)`, it's fair to bypass it I think.

PR #8185, opened 2025-11-05
DIRAC: migrate away from session decorator, make session a required parameter

rdimaio:
Part of #6989

### DIFF
--- a/lib/rucio/core/dirac.py
+++ b/lib/rucio/core/dirac.py
@@ -39,7 +39,7 @@ if TYPE_CHECKING:
 
 
 def _exists(
-    scope: str,
+    scope: InternalScope,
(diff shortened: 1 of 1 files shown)


lib/rucio/core/dirac.py
Geogouz: Should we drop the `issuer` param in this PR? It's probably some leftover.

Geogouz (commented):
Looks fine. Just the `issuer` thingy in case it is worth fixing in this PR.

lib/rucio/core/dirac.py
rdimaio: Addressed in latest revision

Geogouz (approved):

PR #8111, opened 2025-10-09
List and update scope ownership 

voetberg:
Closes #7830

PR #8187, opened 2025-11-06
Revert PR https://github.com/rucio/rucio/pull/8111 for backwards compatibility

### LINKED ISSUE #8186: Revert scope client change that does not meet backwards compatibility policy
### Description

Issue introduced here https://github.com/rucio/rucio/pull/8111

This change unfortunately does not meet the backwards compatibility policy outlined here https://rucio.github.io/documentation/started/releasepolicy/#client--server-compatibility

We will revert the entire PR for now just to remediate this issue, and we should then make gradual changes to allow for this change. From dchristidis:

> To implement this, we have to: first introduce the forward compatibility in the client, then wait about two years, then change the REST API.

### Steps to reproduce

Upgrade to Rucio 38.5.0

### Rucio Version

_No response_

### Additional Information

_No response_

rdimaio:
fix #8186

### DIFF
--- a/lib/rucio/cli/bin_legacy/rucio.py
+++ b/lib/rucio/cli/bin_legacy/rucio.py
@@ -36,7 +36,7 @@ from tabulate import tabulate
 
 # rucio module has the same name as this executable module, so this rule fails. pylint: disable=no-name-in-module
 from rucio import version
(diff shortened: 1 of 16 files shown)


rdimaio (commented):
Type annotation check can be ignored - it's due to removing some type annotations that are reverted as part of this overall reversal

dchristidis (approved):

PR #8193, opened 2025-11-07
Dependencies: upgrade using strip extras (upgrade for Rucio 39)

rdimaio:
To maintain consistency with previous version. Part of #8165, follow-up to #8170


Geogouz:
Right now, the removal of those extras sounds safe because either we require them directly or indirectly through other dependencies. However, since `pyjwt[crypto]==2.10.1` was initially added automatically to support `globus_sdk`, by removing this extra, there is a chance that Globus transfers (for example, when RefreshTokenAuthorizer refreshes a token and needs RSA verification) might face problems. In our current tests it seems that this is practically not excersized and that we are mocking this. But so far we didn't have any problems and I assume crypto was not introduced as a requirement now.

I would say that we can merge this and if something breaks in production, we should definitely add a test that represents their use scenario.

Geogouz (approved):

PR #8207, opened 2025-11-11
Core & Internals: Add centralized startup diagnostics; Fixes rucio#8197

### LINKED ISSUE #8197: Add startup self-check mechanism to block Rucio services when critical diagnostics fail
**Note: This issue is delayed for now and will be rescheduled for release 41 LTS. It needs to be split into multiple sub-issues/PRs to help with its reviewing.**

### Summary
Our Rucio services (REST server, daemons) start immediately after building their Flask app/blueprints and establishing heartbeats. However, there is no built-in hook that runs a mandatory diagnostic suite whose failure would abort the startup.

### Current behaviour and blockers
- The shared daemon bootstrap relies on `heartbeat.sanity_check`, which deliberately swallows `DatabaseException`, so daemons keep starting even when the database is unreachable and only fail later when they try to do real work.
- Also, some daemons invoke `sanity_check` directly instead of going through a single shared entry point, so there is no uniform place to register extra diagnostics that every service will honor.
- DID metadata and OpenData metadata are stored in JSON columns, and once Oracle 21c+ is introduced by some experiment, we must be able to check/ensure that the deployment has migrated those columns to Oracle’s native JSON types before we allow services to run (validating and handling the request accordingly on every metadata call would be too expensive).

### Proposal
Provide a light startup self-check framework (e.g., a registry where tests can be plugged in) that executes before the REST server or daemons begin handling workload. If any registered test fails, the process should exit with a non-zero status and log the failure. Ideally this should be configurable, so operators can enable/disable specific tests and extend the suite with environment-specific checks. For start, I would like to introduce there a common test for Oracle 21c+ that checks if the database is JSON ready. If it is not, then specific database upgrades would be required by the admin.

### Alternatives
If we do not introduce such an early check that is able to stop the services, I will either have to perform this check on every metadata call, or I will have to detect and store on Rucio's Rest server/daemons startups, the database's state and act on every metadata call accordingly, or we need to advertise somewhere (and trust that this will be respected) that once you upgrade to Oracle 21c+, you need to run once, a specific alembic migration.

Any opinion on that @bari12 @rdimaio ?

### Motivation

_No response_

### Change

_No response_

Geogouz:
Handles rucio#8197

I have placed at the top of `startup_checks.py` a documentation/guide about this framework and its use. Nevertheless, I am placing this also here to understand what this does. Once we agree/merge this, we can also place it in the dev documentation. In summary, with this testing framework, we can now add easily checks that unless they pass, the services are blocked from starting. And such a test is needed for the new Metadata system. I will now push another PR where this framework is used. But it can wait till we agree on this one.

Keep in mind that the startup check registry in `lib/rucio/common/startup_checks.py` is initialized as an empty dictionary `(_registry: dict[str, StartupCheck] = {})`, and there are currently no calls that populate it outside of documentation examples and unit tests. Hence, there are currently zero startup checks registered at process startup (i.e. no-op for the moment) and so it should be quite safe to merge. Nevertheless, the introduced tests are passing.

Startup diagnostics registry and runner
======================================

This module provides a small framework to **register** and **run** fast, synchronous “startup checks” at process start. Its use is to assert prerequisites (database reachable, configuration sanity, network access, etc.) *before* the process begins normal work. It supports **simple/central** configuration but allows also **advanced operator controls**.

TL;DR summary
----------------------------

- A *startup check* is a **fast**, **side‑effect‑free**, **synchronous** function with **no arguments**.
- Register checks with :func:`register_startup_check`.
- Run them at startup with :func:`run_startup_checks(tags={...})` (or :func:`rucio.daemons.common.run_daemon_startup_checks` for daemons).
- **Untagged checks run for every service** (“always”).
- If **any** check fails, a :class:`~rucio.common.exception.StartupCheckError` is raised and your startup should abort.
- **No configuration is required.** If there is no ``[startup_checks]`` in ``rucio.cfg``, all checks whose tags match the service **plus** all untagged checks run.

From simple to advanced setups
-------------------------------------------

**1) Minimal – no config, run by tags (default behaviour)**

- The various check definitions can be kept at one place/module or spread however prefered.
- Register checks with an optional tag; leave ``tags=None`` for “always”.
- In each service, call ``run_startup_checks(tags={'daemon'})`` or ``run_startup_checks(tags={'rest'})``.
- Any failure raises :class:`~rucio.common.exception.StartupCheckError` and should stop startup.

**2) Centralized catalog – keep *all* checks in one module (convention)**

Put all checks in one file (e.g. ``rucio.common.startup_checks_catalog``) and import it at process start.

```python
# rucio/common/startup_checks_catalog.py
from rucio.common.startup_checks import register_startup_check

def register_all() -> None:
    # ALWAYS (untagged) – runs for every service
    def _config_sanity() -> None:
        pass  # raise on failure
    register_startup_check(
        name="config-sanity",
        callback=_config_sanity,
        description="Validate core configuration."
    )

    # Single-tag checks by convention
    def _db_ready() -> None:
        pass
    register_startup_check(
        name="database",
        callback=_db_ready,
        tags={"daemon"},
        description="Database connectivity."
    )

    def _rest_key_material() -> None:
        pass
    register_startup_check(
        name="rest-key-material",
        callback=_rest_key_material,
        tags={"rest"},
        description="REST auth key/cert presence."
    )
```

Import the catalog before running checks:

```python
# Daemon entrypoint
from rucio.common import startup_checks_catalog
startup_checks_catalog.register_all()
from rucio.daemons.common import run_daemon_startup_checks
run_daemon_startup_checks(executable="my-daemon")  # runs {'daemon'} + untagged

# Flask/REST app factory
from rucio.common import startup_checks_catalog
from rucio.common.startup_checks import run_startup_checks
def create_app():
    startup_checks_catalog.register_all()
    run_startup_checks(tags={"rest"})  # raises on failure -> abort startup
    ...
```

**2b) Operator-provided checks – register without modifying Rucio code**

Operators can register deployment-specific checks by shipping a small Python module/package that calls :func:`register_startup_check` and making sure it is imported **in the same Python process** before :func:`run_startup_checks` executes. The module must be importable (installed package or directory on ``PYTHONPATH``).

Choose tags depending on which services should run the check. For example, use ``tags={"daemon"}`` for daemon-only checks, ``tags={"rest"}`` for REST-only checks, or include both to run in both contexts. Omitting ``tags`` leaves the check untagged (“always”), so it will run for any :func:`run_startup_checks` call.

Example deployment module::

```python
# rucio_site_checks/custom_checks.py
import os
from rucio.common.startup_checks import register_startup_check

def check_nfs_mount() -> None:
    if not os.path.ismount("/data/rucio"):
        raise RuntimeError("NFS not mounted: /data/rucio")

register_startup_check(
    "nfs-mount",
    check_nfs_mount,
    tags={"daemon", "rest"},
    description="Ensure /data/rucio is mounted before services start",
)
```

Then ensure it is imported early::

```python
# REST/WSGI app factory / WSGI entrypoint
from rucio.common.startup_checks import run_startup_checks
import rucio_site_checks.custom_checks  # registers checks
run_startup_checks(tags={"rest"})
```

Note: the import must happen in the *same* Python process which calls ``run_startup_checks(...)``. Importing in a shell wrapper which then spawns a new Python process will not register anything in the child process.

**3) Operator controls via config (optional)**

If you add a ``[startup_checks]`` section in ``rucio.cfg``, operators can enable/disable checks without code changes, globally or per service. If you don’t add it, behaviour remains as in (1)/(2).

- ``enabled`` / ``enabled_<TAG>`` – restrict execution to this union (after tag filtering).
- ``disabled`` / ``disabled_<TAG>`` – remove these from the set (**disabled wins** if both listed).
- ``strict = true`` – treat misconfiguration as fatal (unknown names; enabling only non‑applicable names; removing all applicable checks).
- ``timeout_ms`` – **soft** time budget for the whole run (warning if exceeded; execution continues).

Tags & selection
----------------

- Each check can have **zero or more** tags.
- When you call :func:`run_startup_checks(tags=...)`, the runner executes:
  * all checks whose tag set **intersects** the supplied tags, **and**
  * all **untagged** checks (i.e. “always”).
- If you prefer a literal ``"always"`` tag, you can use ``tags={"always"}`` and pass it at runtime,
  but leaving checks untagged is simpler and equivalent.

Public API
----------

``register_startup_check(name, callback, *, tags=None, description=None, replace=False)``
    Register a diagnostic.

  * **name** – unique (case‑insensitive). Trimmed; empty names rejected.
  * **callback** – synchronous callable with no parameters; raise on failure.
    Returning an awaitable is invalid.
  * **tags** – optional iterable of strings (case‑insensitive, trimmed). If omitted, the check applies to all services.
  * **description** – text logged when the check runs.
  * **replace** – set ``True`` to replace an existing registration with the same name.

``run_startup_checks(*, tags, logger=None)``
    Execute all checks applicable to ``tags`` after applying optional configuration.
    On failure, raises :class:`~rucio.common.exception.StartupCheckError`.

Behaviour & selection algorithm
-------------------------------

1. Build the default **applicable** set: all checks whose tag set intersects ``tags`` **plus** all untagged checks.
2. If *any* ``enabled``/``enabled_<TAG>`` is present, restrict to the **union** of those names (still must be applicable).
3. Remove names in ``disabled``/``disabled_<TAG>``. If a name is both enabled and disabled, **disabled wins**.
4. Unknown names are **warned** and ignored (or **fail** in strict mode).
5. If enabled lists select nothing applicable, the runner **falls back** to the default applicable set and logs it
   (this fallback is a **failure** in strict mode).

Thread‑safety & duplication protection
--------------------------------------

- Registration is process‑wide and **thread‑safe**.
- Names are unique **case‑insensitively**. Attempting to register a duplicate name raises unless ``replace=True``.

Configuration quick reference
-----------------------------

```ini
[startup_checks]
# Optional soft time budget for ALL checks (milliseconds)
timeout_ms = 500

# Execute only the following checks (case‑insensitive names)
enabled = database, cache, storage

# Per‑tag enable/disable (suffix is case‑insensitive)
enabled_REST = database
disabled_DAEMON = storage

# Be strict about configuration correctness
strict = false
```

### DIFF
--- a/lib/rucio/common/startup_checks.py
+++ b/lib/rucio/common/startup_checks.py
@@ -1,0 +1,489 @@
+# Copyright European Organization for Nuclear Research (CERN) since 2012
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import annotations
+
+import inspect
+import logging
+import threading
+import time
+from dataclasses import dataclass
+from typing import TYPE_CHECKING, Optional, Protocol
+
+from rucio.common.config import config_get_bool, config_get_int, config_get_list
+from rucio.common.exception import StartupCheckError
+
+if TYPE_CHECKING:
+    from collections.abc import Collection, Iterable, Mapping
+
+
+class StartupCheckCallback(Protocol):
+    def __call__(self) -> None: ...
+
+
+@dataclass(frozen=True)
+class StartupCheck:
+    """
+    Definition of a startup diagnostic.
+
+    Attributes
+    ----------
+    name:
+        Unique identifier of the diagnostic.
+    callback:
+        Callable invoked to run the diagnostic.
+    tags:
+        Normalised set of tags describing the environments where the
+        diagnostic should execute.
+    description:
+        Optional human readable text describing the diagnostic. Logged when
+        the check runs so that operators know what is happening.
+    """
+
+    name: str
+    callback: StartupCheckCallback
+    tags: frozenset[str]
+    description: Optional[str] = None
+
+
+_registry: dict[str, StartupCheck] = {}
+_registry_lock = threading.Lock()
+
+
+def _normalize_name(name: str) -> str:
+    """
+    Canonicalise a diagnostic name.
+
+    A lot of the public API accepts free-form strings. Before storing the
+    value we trim surrounding whitespace to avoid treating ``" db "`` as a
+    distinct check. The same helper is used for validation when registering
+    new checks and when processing configuration overrides, which keeps the
+    matching logic consistent.
+
+    Parameters
+    ----------
+    name:
+        Raw name supplied by the caller.
+
+    Returns
+    -------
+    str
+        The trimmed diagnostic name.
+
+    Raises
+    ------
+    ValueError
+        If the normalised name would be empty.
+    """
+
+    normalized = name.strip()
+    if not normalized:
+        raise ValueError("Startup check name cannot be empty")
+    return normalized
+
+
+def _normalize_tags(tags: Optional[Iterable[str]]) -> frozenset[str]:
+    """
+    Normalise a collection of tags.
+
+    Tags are used to scope checks to specific services (for example ``daemon``
+    or ``rest``). We trim whitespace and fold everything to lowercase. The return
+    type is a :class:`frozenset` so the result can be safely cached and shared
+    between helpers without the risk of accidental mutation.
+
+    Parameters
+    ----------
+    tags:
+        Iterable containing the raw tag identifiers. ``None`` or empty values
+        result in an empty set.
+
+    Returns
+    -------
+    frozenset of str
+        The cleaned, lower-cased tags with surrounding whitespace removed.
+    """
+
+    if not tags:
+        return frozenset()
+    return frozenset(tag.strip().lower() for tag in tags if tag and tag.strip())
+
+
+def _collect_configured_names(
+        option: str,
+        tags: frozenset[str],
+        tag_suffixes: Mapping[str, Collection[str]]
+) -> set[str]:
+    """
+    Collect the check names configured for a given option.
+
+    Configuration may list checks under generic ``enabled`` / ``disabled``
+    keys as well as under tag specific variants such as ``enabled_daemon``.
+    This helper gathers all the possible spellings for the requested tags and
+    collates the configured check names into a single cleaned set. The caller
+    is still responsible for reconciling these names with the registry.
+
+    Parameters
+    ----------
+    option:
+        Name of the configuration option (``"enabled"`` or ``"disabled"``).
+    tags:
+        Set of tags describing the current service.
+
+    Returns
+    -------
+    set of str
+        Cleaned set of check names extracted from the configuration.
+    """
+
+    # ConfigParser normalizes option names, but config table lookups are case-sensitive,
+    # so query common case variants to avoid missing mixed-case entries.
+    options_to_query = {
+        option,
+        option.lower(),
+        option.upper(),
+    }
+
+    for tag in tags:
+        suffixes = tag_suffixes.get(tag)
+        if not suffixes:
+            suffixes = {tag}
+        for suffix in suffixes:
+            tag_option = f'{option}_{suffix}'
+            options_to_query.add(tag_option)
+            options_to_query.add(tag_option.lower())
+            options_to_query.add(tag_option.upper())
+
+    values: set[str] = set()
+    for config_option in options_to_query:
+        values.update(config_get_list(
+            'startup_checks',
+            config_option,
+            raise_exception=False,
+            default=[],
+            check_config_table=False,
+        ))
+    return {value.strip() for value in values if value and value.strip()}
+
+
+def _select_checks(tags: frozenset[str]) -> tuple[list[StartupCheck], set[str]]:
+    """
+    Return the set of checks applicable to the supplied tags.
+
+    The registry is protected by a lock during iteration so that late
+    registrations do not race with an ongoing startup check run. The function
+    returns a snapshot rather than a live view because the caller will filter
+    the list based on configuration overrides.
+
+    Parameters
+    ----------
+    tags:
+        Set of normalised tags describing the current service.
+
+    Returns
+    -------
+    tuple
+        Two elements where the first item is the list of registered checks
+        whose tag set intersects *tags* (or which are unscoped) and the
+        second item is the set of all registered check names.
+    """
+
+    with _registry_lock:
+        checks = list(_registry.values())
+        all_names = set(_registry)
+
+    if not tags:
+        return checks, all_names
+
+    return [
+        check for check in checks
+        if not check.tags or check.tags.intersection(tags)
+    ], all_names
+
+
+def _prepare_tags(tags: Optional[Iterable[str]]) -> tuple[frozenset[str], dict[str, frozenset[str]]]:
+    """
+    Normalise tags and compute the variants used for configuration lookups.
+
+    Besides producing the cleaned :class:`frozenset` of tags, this helper also
+    pre-computes a map of case variants for each tag. Configuration files may
+    spell options with different casing (``ENABLED_DAEMON`` vs ``enabled_daemon``),
+    so keeping all the candidate variants upfront avoids repeating that work in
+    the hot path.
+    """
+
+    if not tags:
+        return frozenset(), {}
+
+    normalized: set[str] = set()
+    suffixes: dict[str, set[str]] = {}
+
+    for raw_tag in tags:
+        if not raw_tag:
+            continue
+        trimmed = raw_tag.strip()
+        if not trimmed:
+            continue
+        normalized_tag = trimmed.lower()
+        normalized.add(normalized_tag)
+        variants = suffixes.setdefault(normalized_tag, set())
+        variants.add(normalized_tag)
+        variants.add(trimmed)
+        variants.add(trimmed.lower())
+        variants.add(trimmed.upper())
+
+    return frozenset(normalized), {tag: frozenset(variants) for tag, variants in suffixes.items()}
+
+
+def run_startup_checks(
+        *,
+        tags: Optional[Iterable[str]] = None,
+        logger: Optional[logging.Logger] = None
+) -> None:
+    """
+    Run startup diagnostics applicable to the given service tags.
+
+    Collects all registered checks that apply to ``tags`` (plus untagged ones),
+    optionally filters them according to the ``[startup_checks]`` section in
+    ``rucio.cfg``, and executes them. On any failure, raises
+    :class:`~rucio.common.exception.StartupCheckError`.
+
+    Parameters
+    ----------
+    tags : Collection[str]
+        Tags describing the current service(s). Common values are ``{"daemon"}`` and ``{"rest"}``.
+    logger : logging.Logger | Callable[..., None] | None, optional
+        Component‑specific logger. If omitted, uses the module logger ``rucio.startup_checks``.
+
+    Behavior
+    --------
+    - **Default selection**: all checks whose tag set intersects the supplied ``tags``,
+      plus untagged checks.
+    - **Config filters (optional)** from ``[startup_checks]`` in ``rucio.cfg``:
+        * ``enabled`` / ``enabled_<TAG>`` restrict the set (union) after tag filtering.
+        * ``disabled`` / ``disabled_<TAG>`` remove names. **Disabled wins** when both apply.
+        * Name and tag comparisons are case‑insensitive; duplicates are deduplicated.
+        * Unknown names are ignored with a warning (or fail in strict mode).
+        * If enabled lists select nothing applicable, the runner falls back to the default set
+          and logs that fallback (unless strict mode is on).
+    - **Soft time budget**: ``startup_checks.timeout_ms`` sets a soft limit for the whole run.
+      Exceeding the cumulative runtime logs a warning once and execution continues.
+    - **Strict mode**: ``startup_checks.strict = true`` turns these into hard errors:
+        * unknown names in any ``enabled*``/``disabled*``,
+        * enabled lists containing only non‑applicable names,
+        * configuration that removes all applicable checks.
+
+    Returns
+    -------
+    None
+
+    Raises
+    ------
+    StartupCheckError
+        If any check fails, or if configuration is invalid under strict mode.
+
+    Logging
+    -------
+    - Logs each check's ``description`` (if provided) before execution.
+    - Logs a completion summary like:
+      ``Startup checks completed successfully: N ran, M disabled by config``.
+
+    Examples
+    --------
+    >>> run_startup_checks(tags={"daemon"})
+    """
+
+    normalized_tags, tag_suffixes = _prepare_tags(tags)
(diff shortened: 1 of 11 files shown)


lib/rucio/daemons/common.py
```diff
@@ -35,6 +36,25 @@
...
+
+    logger = logging.getLogger(executable)
+    resolved_hostname = hostname or socket.getfqdn()
+
+    with _daemon_startup_checks_lock:
+        if not _daemon_startup_checks_ran:
+            run_startup_checks(tags={'daemon'}, logger=logger)
+            _daemon_startup_checks_ran = True
+
+    heartbeat_core.sanity_check(executable=executable, hostname=resolved_hostname)
```
bari12, line 122: I don't think this should be part of the startup check.

bari12 (changes requested):

bari12:
What I don't get, is this only done for testing? I assume so, but in this case, I think this needs to have a clearer name. When I read the daemon file, this looks like checks ran in general

lib/rucio/daemons/common.py
Geogouz: I updated the method with a description and comments on what it does (and why it is safe). This PR at the moment does not introduce any registerable method with the tag "daemon". Therefore, the only checks that we will end up having (after merging the subsequent PR that comes next), is a single global ("always") check that is about the JSON support in the database. But if/when we later add also any other check method tagged as "daemons", these will automatically be respected by all daemons. With this framework, we can of course introduce also additional tests unique to specific daemons/services (i.e. tagged differently). Lastly, an important difference between this startup-checking framework and the `sanity_check` that is always triggered as the final step, is that the later allows for re-tries. On the contrary, if the startup_checks fail, the entire startup process exits with the problem logged. This, because the problem we just detected is critical and requires solution first so any retries are meaningless.

Geogouz: Also, by making this method the entry point of all daemons ensures that the (already existing) sanity check is automatically invoked and we do not have to remember adding it manually everytime a daemon is added. For example, couple daemons (e.g. `rucio-cache-consumer`) start their worker logic without using the `HeartbeatHandler` or calling `heartbeat_core.sanity_check(...)`, so they currently skip the sanity check entirely.

Geogouz:
> What I don't get, is this only done for testing? I assume so, but in this case, I think this needs to have a clearer name. When I read the daemon file, this looks like checks ran in general

The `startup_checks` module is now part of the runtime startup, not a simple testing helper. It is basically a framework for running diagnostics at process start (e.g., database reachability, config sanity) to abort startup early if prerequisites are missing (whether this is in production or testing).

**- Daemons** call `run_daemon_startup_checks`, which registers catalogued checks and executes those tagged for daemons (plus the untagged “always” checks) before the old heartbeat sanity kicks in automatically. This happens on the first entry into a daemon’s HeartbeatHandler in each process.

**- The REST Flask** app calls `run_startup_checks(tags={'rest'}, ...)` during application setup, so the corresponding diagnostics (plus the untagged “always” checks) run at server startup for the REST service.

The tests you see in `tests/rucio/common/test_startup_checks.py` only validate this behavior; they don’t limit the feature to tests. So the naming “startup checks” reflects their role as pre-flight diagnostics at service startup rather than a test-only mechanism.

Geogouz:
This PR introduces only the framework and does not add any check needed to be executed. So there should be no difference in the startup process. The first (and so far the only) check being introduced is in this future commit (which is part of another PR):

https://github.com/rucio/rucio/blob/e26771177cf8a330118cbb01740bb71cc365586e/lib/rucio/common/startup_checks_catalog.py

To explain what is going on there, the registration is happening here:

```python
def register_all() -> None:
    """Register all built-in startup checks."""

    register_startup_check(
        name='oracle-json-columns',
        callback=ensure_oracle_json_columns_are_native,
        description='Ensure Oracle JSON columns are migrated to native types on 21c+',
        replace=True,
    )
```

Since no `tags` are passed in `register_startup_check(..)` (`oracle-json-columns` is just the name of this check), this check is considered a global ("always") one that both Flask and Daemons will need to pass in order to be allowed to startup.

lib/rucio/daemons/common.py
```diff
@@ -35,6 +36,53 @@
...
```
mgajek-cern: `sanity_check()` runs outside the lock (every time), while diagnostics run once per process. Is this intentional? If yes, add comment explaining why. If no, should we move it inside the lock? Consider tradeoffs: performance vs consistency, concurrent daemon startup behavior, heartbeat table race conditions.

---

Move sanity_check() inside lock? | Pro | Con
-- | -- | --
Yes | Guaranteed once-per-process, consistent with diagnostics pattern | Longer lock hold, blocks concurrent daemon starts
No (current) | Fast, validates heartbeat each entry | Multiple concurrent calls possible

---

E.g. inside lock suggestion:

```suggestion
            run_startup_checks(tags={'daemon'}, logger=logger)
            heartbeat_core.sanity_check(executable=executable, hostname=resolved_hostname)
            _daemon_startup_checks_ran = True
```

lib/rucio/daemons/common.py
```diff
@@ -35,6 +36,53 @@
...
+    flag is flipped only after that runner returns successfully. Because
+    the flag is cached per process, later calls in the same process skip the
+    diagnostic runner and proceed straight to the heartbeat verification.
+
+    The helper can also be used to register checks by itself. If you follow the
+    "central catalog" convention from :mod:`rucio.common.startup_checks`, register
+    your checks (for example by importing ``startup_checks_catalog.register_all``)
+    before the first call to this function. Other registration strategies (local
+    definitions, ad-hoc imports, etc.) work as long as the checks are registered
+    before this helper runs.
```
mgajek-cern, line 73: Must all checks be registered before `run_startup_checks()`? Late registration could causes false "unknown check" warnings. Document if this isn't supported.

```suggestion
    before the first call to this function. IMPORTANT: All checks MUST be registered before the first call to this function. Late registration will cause false "unknown check" warnings in config validation and may result in checks not executing. Register at module import time or via a centralized catalog (e.g. ``startup_checks_catalog.register_all()``) loaded at process start.
```

lib/rucio/common/startup_checks.py
```diff
@@ -0,0 +1,749 @@
...
+
+    for tag in tags:
+        suffixes = tag_suffixes.get(tag)
+        if not suffixes:
+            suffixes = {tag}
+        for suffix in suffixes:
+            tag_option = f'{option}_{suffix}'
+            options_to_query.add(tag_option)
+            options_to_query.add(tag_option.lower())
+            options_to_query.add(tag_option.upper())
```
mgajek-cern, line 480: Why generate case variants per option when config keys are likely case-insensitive? Can we canonicalize to lowercase? If no, add brief inline comment.

Suggestion:

```suggestion
options_to_query = {option.lower()}
for tag in tags:
    options_to_query.add(f'{option.lower()}_{tag}')
```

lib/rucio/common/startup_checks.py
```diff
@@ -0,0 +1,749 @@
...
+            existing_case_insensitive = lower_to_name.get(normalized_name_lower)
+            if existing_case_insensitive is not None:
+                raise ValueError(
+                    'Startup check '
+                    f'"{existing_case_insensitive}" is already registered (matches new name "{normalized_name}")'
+                )
+            target_name = normalized_name
+        else:
+            existing_case_insensitive = lower_to_name.get(normalized_name_lower)
+            target_name = existing_case_insensitive or normalized_name
```
mgajek-cern, line 425: Why check exact match when case-insensitive check already covers it? Simplify to single check.

Suggestion:

```suggestion
with _registry_lock:
    lower_to_name = {name.lower(): name for name in _registry}
    
    if not replace:
        existing = lower_to_name.get(normalized_name_lower)
        if existing:
            raise ValueError(f'Check "{existing}" already registered (matches "{normalized_name}")')
        target_name = normalized_name
    else:
        target_name = lower_to_name.get(normalized_name_lower, normalized_name)
```

lib/rucio/common/startup_checks.py
```diff
@@ -0,0 +1,749 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# ruff: noqa: UP007,UP045
+
+"""
+Startup diagnostics registry and runner
```
mgajek-cern, line 18: Should there be an example on how operators register custom checks in their deployment without modifying Rucio code, e.g.

```
Custom Operator Checks
----------------------

Operators can register deployment-specific checks by creating a Python module
that imports and calls register_startup_check, then ensuring that module is
imported before services start:

# /etc/rucio/custom_checks.py
from rucio.common.startup_checks import register_startup_check

def _check_nfs_mount():
    assert os.path.ismount('/data/rucio'), "NFS not mounted"

register_startup_check('nfs-mount', _check_nfs_mount, tags={'daemon'})

# In daemon wrapper script or WSGI file:
import custom_checks  # This registers the check
```

tests/rucio/common/test_startup_checks.py
```diff
@@ -0,0 +1,573 @@
...
+    monkeypatch.setattr(startup_checks.time, 'perf_counter', lambda: next(times))
+
+    with caplog.at_level(logging.WARNING, logger='rucio.startup_checks'):
+        startup_checks.run_startup_checks(tags={'daemon'})
+
+    assert executed == ['alpha']
+    warning_messages = [record.getMessage() for record in caplog.records if record.levelno >= logging.WARNING]
+    assert any('exceeded soft timeout' in message for message in warning_messages)
+
+
```
mgajek-cern, line 464: Maybe you can add some more test cases.E.g. on:

-  Registry preservation on replace
- Late registration edge case
- Multiple tags intersection
-  Empty registry
- Config with only non-applicable checks (strict mode off)

mgajek-cern (commented):
Some findings and questions. Overall good progress

lib/rucio/daemons/common.py
Geogouz: Yes, the lock is explicitly meant to guard only the one-time diagnostics. This does not alter the previous functionality too (as the default functionality when no checks are registered). And the pros of having it there do not justify much the introduction of potentially long lock holds (e.g. with the DB).

lib/rucio/daemons/common.py
Geogouz: Aside from the points: 1) _"as long as the checks are registered"_, and 2) _"register your checks (for example by importing ``startup_checks_catalog.register_all``) before the first call to this function"_, the explicit “register then run” pattern is primarily also shown in the module docs for startup checks in `lib/rucio/common/startup_checks.py`, under the _“Import the catalog before running checks”_ section. I am in general in favour of including details within docstrings because they tend to save time from trying to go over the code, but wouldn't these be sufficient?

lib/rucio/common/startup_checks.py
Geogouz: Needed because of the config table lookups. Added some comment

lib/rucio/common/startup_checks.py
Geogouz: The “exact match” check is mostly about diagnostics. Here we distinguish two different developer mistakes:
- **Exact duplicate registration** (“this module got imported twice”, “register function called twice”, or “you accidentally registered the same check again”),
- **Case-only collision** (“two authors used different casing” ("DB" vs "db"), which is a different class of problem and the message helps you spot it).

Geogouz:
Btw @dchristidis, before merging and to ensure backwards compatibility, an important point to assess is whether adding the `sanity_check` through `run_daemon_startup_checks` in all daemons is safe. Because so far, several daemons were skipping such checks (or at least this was not implemented uniformly everywhere and I am not sure if this could be even on purpose). But for uniformity, I added it. I believe this is safe but just to be on the safe side..

lib/rucio/common/startup_checks.py
Geogouz: Added:
**2b) Operator-provided checks – register without modifying Rucio code**

Geogouz:
Converting it for a bit into a draft to protect from merging cause I am investigating a probable issue.

Geogouz:
Disabling the draft mode. I made a small addition to handle a potential edge case. I want to ensure that the main service will log and exit once a failure inside a multi-threaded run, occurs.

PR #8211, opened 2025-11-13
Clients: Ensure temp dir is created in flow of browser-based token retrieval. Fix #8210

### LINKED ISSUE #8210: Browser-based token retrieval fails due to missing temporary directory
### Description

I'd like to submit a pull request to fix creating the directory where a temporary file is created when using browser-based token retrieval. According to the guidelines, I need to create an issue first. I hope I may contribute this fix.

### Steps to reproduce

```
$ rucio whoami

Please use your internet browser, go to:

    https://test.rucio.km3net.org/auth/oidc_redirect?tkUGmSUGSdcA1bPQJDhqw8V    

and authenticate with your Identity Provider.
Copy paste the code from the browser to the terminal and press enter:
<pasted token here>

2025-11-13 07:43:38,070 ERROR   [Errno 2] No such file or directory: '/tmp/roel/.rucio_roel/tmpltk1vagj'
2025-11-13 07:43:38,070 ERROR
Rucio exited with an unexpected/unknown error.
Please rerun the last command with the "-v" option to gather more information.
If you're sure there is a problem with Rucio itself, please follow up at: https://github.com/rucio/rucio/issues
```

### Rucio Version

Rucio commit `35ad0767a7a840bcadaaa98d1f25812aa233f87a`.

### Additional Information

GNU/Linux, Debian 13.

roelj:
This commit extracts the creation of a temporary directory which is used in a call to ‘mkstemp’.  It then ensures that this new procedure ‘__create_temp_directory()’ is called before the call to ‘mkstemp’.

* lib/rucio/client/baseclient.py: Implement and use ‘__create_temp_directory’.

### DIFF
--- a/lib/rucio/client/baseclient.py
+++ b/lib/rucio/client/baseclient.py
@@ -242,6 +242,22 @@ class BaseClient:
         except ValueError:
             self.logger.debug('request_retries must be an integer. Taking default.')
 
+    def __create_temp_directory(self) -> None:
+        """
+        Ensure the temporary directory is there.
+        """
+        # check if rucio temp directory is there. If not create it with permissions only for the current user
+        if not os.path.isdir(self.token_path):
+            try:
+                self.logger.debug('rucio token folder \'%s\' not found. Create it.' % self.token_path)
+                try:
+                    makedirs(self.token_path, 0o700)
+                except FileExistsError:
+                    msg = f'Token directory already exists at {self.token_path} - skipping'
+                    self.logger.debug(msg)
+            except Exception:
+                raise
+
     def _get_auth_tokens(self) -> tuple[Optional[str], str, str, str]:
         # if token file path is defined in the rucio.cfg file, use that file. Currently this prevents authenticating as another user or VO.
         auth_token_file_path = config_get('client', 'auth_token_file_path', False, None)
@@ -734,6 +750,7 @@ class BaseClient:
             self.logger.debug("Resetting the token expiration epoch file content.")
             # reset the token expiration epoch file content
             # at new CLI OIDC authentication
+            self.__create_temp_directory()
             self.token_exp_epoch = None
             file_d, file_n = mkstemp(dir=self.token_path)
             with fdopen(file_d, "w") as f_exp_epoch:
@@ -975,18 +992,7 @@ class BaseClient:
         """
         Write the current auth_token to the local token file.
         """
-        # check if rucio temp directory is there. If not create it with permissions only for the current user
-        if not os.path.isdir(self.token_path):
-            try:
-                self.logger.debug('rucio token folder \'%s\' not found. Create it.' % self.token_path)
-                try:
-                    makedirs(self.token_path, 0o700)
-                except FileExistsError:
-                    msg = f'Token directory already exists at {self.token_path} - skipping'
-                    self.logger.debug(msg)
-            except Exception:
-                raise
-
+        self.__create_temp_directory()
         try:
             file_d, file_n = mkstemp(dir=self.token_path)
             with fdopen(file_d, "w") as f_token:
--- a/tests/test_clients.py
+++ b/tests/test_clients.py
@@ -253,3 +253,24 @@ class TestRucioClients:
 
         # rule
         assert list(client.list_replication_rules()) is not None
+
+    def test_run_client_oidc_directory_creation (self):
+        rucio_host = "https://rucio:443"
+        auth_host = "https://rucio:443"
+
+        from getpass import getuser
+        from shutil import rmtree
+        from rucio.client.client import Client
+
+        target_directory = f"{Client.TOKEN_PATH_PREFIX}{getuser()}"
+        rmtree (target_directory, ignore_errors=True)
+        client = Client(
+            rucio_host=rucio_host,
+            auth_host=auth_host,
+            auth_type="oidc")
+
+        # Ensure directory exists and permissions are set to 'rwx------'
+        try:
+            assert os.stat(target_directory).st_mode & 0o777 == 0o700
+        except FileNotFoundError:
+            pytest.fail(f"directory not created: {target_directory}")


roelj:
So.. the diff is a bit non-transparent. What I did is: I took the directory creation (first clode block) in `__write_token` and moved it to a new procedure named `__create_temp_directory`. Then call that first in `__write_token` and add a call to `__create_temp_directory` in the code path that assumes the directory is created in `__get_token_oidc`.

I hope the explanation helps speeding up the review process. :-)

lib/rucio/client/baseclient.py
```diff
@@ -984,6 +985,11 @@ def __write_token(self) -> None:
             except Exception:
                 raise
 
+    def __write_token(self) -> None:
```
Geogouz, line 991: Can you add an integration/unit test that deletes the token directory, triggers a token write, and asserts the directory is recreated with `0o700` permissions?

lib/rucio/client/baseclient.py
```diff
@@ -968,9 +969,9 @@ def __read_token(self) -> bool:
         self.logger.debug('got token from file')
         return True
 
-    def __write_token(self) -> None:
+    def __create_temp_directory (self) -> None:
```
Geogouz: Please remove the whitespace here to respect `E211`.

lib/rucio/client/baseclient.py
Geogouz: For the `E231` correctness, can you also add right before the `method=HTTPMethod.GET`, the missing whitespace please?

Geogouz:
Can you also rebase please?

lib/rucio/client/baseclient.py
roelj: I attempted to do so in [tests/test_clients.py](https://github.com/rucio/rucio/pull/8211/files#diff-ec43990925a219561939f41cff82831d387574bbb71cdd29df8f045c89c333b8).

lib/rucio/client/baseclient.py
roelj: Done in the new PR.

roelj:
I ran into this issue today again with Rucio 39. I hope we can at least accept the bugfix at some point. :-)

tests/test_clients.py
```diff
@@ -253,3 +253,24 @@ def test_run_client_no_config(self, vo):
...
+        target_directory = f"{Client.TOKEN_PATH_PREFIX}{getuser()}"
+        rmtree (target_directory, ignore_errors=True)
+        client = Client(
+            rucio_host=rucio_host,
+            auth_host=auth_host,
+            auth_type="oidc")
+
+        # Ensure directory exists and permissions are set to 'rwx------'
+        try:
+            os.stat(target_directory).st_mode & 0o777 == 0o700
```
bari12: You need to assert this. This way this is just a no-op statement.

bari12 (changes requested):

roelj:
I reformatted the patch so it's clearer what actually happens. I really hope it's OK now, because otherwise I'll just drop  the bugfix and document it as a bug that won't be fixed.

roelj:
It seems that the failing tests have nothing to do with my PR. And you know what, I give up. I will never contribute to Rucio ever again. How can an obvious simple fix take half a year and three major releases to end like this? I wasted hours on this, and so have you. Perhaps try to figure out a more humane way to contribute, but this is not how Free and Open Source software was intended to be: Empowering its users. This project setup is doing quite the opposite.

So either, you can add the bugfix with or without the unit tests, or you can close it and leave the bug in. Then I will document this issue as wont-fix. Up to you.

Geogouz:
I would argue that it is not that obvious as a fix (or review given the available time of the reviewers), because for example there are still many points that are quite alarming in this PR. I see some tests failing also because of them (which do explain what went wrong). For example you have unsorted local imports, unused `client`, and undefined name `os`. On top of that, I see that in the test you are deleting the default runtime token directory which doesn't sound safe to me. Then, to test this, you perform a real OIDC authentication (without mocking auth responses or isolating config) which may lead to flaky tests. So indeed this needs some more work.

That said, you’re right that PRs can take longer than ideal and we acknowledge the part of the responsibility that does fall on our side and are actively working on improvements. As one step forward, we’re planning to introduce an issue acceptance framework/policy soon, which should help provide greater confidence that accepted issues will be addressed within a reasonable timeframe.

roelj:
So.. This patch consists of:
1. The actual bugfix - ensuring that the temporary directory is created before the code assumes it exists.
2. The unit tests, testing for this behavior.

As far as I can tell, your complaints are about 2. It is unfortunate that because of that, the actual bugfix (1) will not be in Rucio. I honestly tried to contribute to the best of my efforts. It's sad to see that instead of helping each other making Rucio better, you choose to put more and more work on my shoulders when I attempted to solve a bug.

> As one step forward, we’re planning to introduce an issue acceptance framework/policy soon, which should help provide greater confidence that accepted issues will be addressed within a reasonable timeframe.

FWIW, I don't think that more policies and more overhead will improve the review situation in this case. It would've taken about 10 to 20 minutes of (all of your) review time to see what the problem was and whether the patch (the actual bug fix without the unit tests) solves the problem without side-effects. But instead you choose to put up a wall of procedures, policies and documentation to the frustration of a new contributor and to waste your own time. 

Close the PR if you'd like. Then I will document in our documentation that this bug will remain in Rucio.

bari12:
Hi,

Let me first apologise for the long time it takes to get a review. This is an ongoing issue for us, which we are trying our hardest to improve, but we are not always succeeding.

Concerning the tests, I am not really sure what you are advocating for. Our policy is to have everything tested, especially already identified issues, so that they do not re-appear at some point in time. Yes, this takes time, and sometimes it takes more time to write good tests, than the actual bugfix itself. But in the long run, this ensures stability and maintainability of the software. Your ultimatum to us to either accept the bugfix without tests or with broken tests is not really helpful. I really appreciate the effort you put into the PR and I apologise for the time it took to review, but we can't abandon our policies just because writing the tests takes some more time. 

I also agree with the sentiment that `Free and Open Source software was intended to be: Empowering its users`. But the users here are not only contributors, they are also endusers and organisations which put trust in the software that a certain software quality standard is upheld. I am happy to have a separate discussion on how to achieve this, but this PR is probably not the best time to do that.

Best,
Martin

roelj:
Regarding:
> Your ultimatum to us to either accept the bugfix without tests or with broken tests is not really helpful. I really appreciate the effort you put into the PR and I apologise for the time it took to review, but we can't abandon our policies just because writing the tests takes some more time.

I didn't mean to put you up for an ultimatum. So it seems unfortunate to me to phrase it as such. What I am trying to communicate is that I spent way more time trying my best to contribute, but am now no longer contributing because the overhead is disproportional to the bugfix I attempted to contribute.

Then I rambled a bit about why I think that is. It's your project, and your rules. On the other hand, it's my time and my effort. It seems we have reached the point where these are simply too far apart.

> I also agree with the sentiment that Free and Open Source software was intended to be: Empowering its users. But the users here are not only contributors, they are also endusers and organisations which put trust in the software that a certain software quality standard is upheld.

So the end-result of this conversation is that you have code that isn't unit-tested and contains a known bug.

The other outcomes were:
2. A fixed bug, but not unit-tested
3. A fixed bug, unit-tested

To me, it seems the policies are preventing better outcomes from our time and effort. It is not empowering its users. It is keeping users from fixing bugs they run into.

Thank you for your time and effort and good luck going forward with Rucio. :)

tamasgal:
I get the same error after pasting the token from the browser back to Rucio

```
2026-06-08 15:15:40,954	ERROR	[Errno 2] No such file or directory: '/var/folders/p6/5y_ncc2n73x31z05l5vbs2140000gn/T/tamasgal/.rucio_tamasgal/tmpm0ukccdz'
2026-06-08 15:15:40,955	ERROR
Rucio exited with an unexpected/unknown error.
Please rerun the last command with the "-v" option to gather more information.
If you're sure there is a problem with Rucio itself, please follow up at: https://github.com/rucio/rucio/issues
```

PR #8212, opened 2025-11-13
Core & Internals: Enhance dev/test logging; Fixes rucio#8208

### LINKED ISSUE #8208: Core & Internals: Surface Rucio service logs in dev/test containers for debugging
The dev container’s logging is a bit misconfigured and stdout logging doesn't reach properly the running container, leaving `mod_wsgi`’s default stdout restrictions in place. If you have tried to log/print something and couldn't find the output somewhere, you have already seen this problem.

We need to fix this because it is quite frustrating. Basically I want to direct all outputs to a single destination (accessible by something like `docker exec -it dev-rucio-1 tail -f /var/log/rucio/httpd_error_log`).

I have already pushed the changes required from the container side here:
https://github.com/rucio/containers/pull/450

Geogouz:
Handles #8208

This PR mainly enhances dev/test observability while keeping production defaults untouched. Some points:

- The logging helper now lets you choose the destination via config or environment variables, and it only redirects `stdout` to `stderr` when one of the new toggles is explicitly enabled. Without those toggles the runtime behaviour is identical to the previous release, so merging this change keeps production defaults intact.
- The Docker dev stack opts into the new behaviour (forces logging to stderr and unbuffers Python output), but this is limited to the development compose profile and doesn’t affect deployed instances.
- Also, test configuration files got an explicit logformat setting because without it, logs look ugly with redundant timestamps (one from apache and one from python) and `/t`'s instead of spaces. The merge utility now reads INI sources with interpolation disabled (we don't use/need it even) so that the introduced percent tokens survive intact the parsing. These files live also under the test-only tree and won’t be consumed by production services.

The only potentially risky part is `WSGIRestrictStdout Off`, which is explicitly annotated as a debugging aid and lives inside the test Apache config. It should stay disabled in real deployments, but its presence here doesn’t (shouldn't) affect production assets.

_Note:_

The use of the `logformat` is to get from the top state to the bottom state:

image

### DIFF
--- a/etc/docker/dev/docker-compose.yml
+++ b/etc/docker/dev/docker-compose.yml
@@ -62,6 +62,9 @@ services:
       - RUCIO_SOURCE_DIR=/rucio_source
       - RDBMS
       - DEV_PROFILES
+      - RUCIO_LOG_STREAM=stderr
+      - RUCIO_LOGGING_REDIRECT_STDOUT_TO_STDERR=true
+      - PYTHONUNBUFFERED=true
     depends_on:
       oracle:
         condition: service_healthy
--- a/etc/docker/test/extra/rucio.conf
+++ b/etc/docker/test/extra/rucio.conf
@@ -1,9 +1,18 @@
 SSLSessionCache  shmcb:/var/log/httpd/ssl_scache(512000)
 
 Listen 443
+# In mod_wsgi *daemon* mode, direct writes to sys.stderr (and, with WSGIRestrictStdout Off,
+# also sys.stdout) can end up in the *main* server error log instead of the vhost ErrorLog.
+# For the docker test setup we want all REST-related output in one place, so we point the
+# main ErrorLog to /var/log/rucio/httpd_error_log and let the vhost inherit it.
+ErrorLog /var/log/rucio/httpd_error_log
 
 WSGIRestrictEmbedded On
 WSGIDaemonProcess rucio processes=4 threads=4
+# Allow the app to access and write to sys.stdout (e.g., via print statements or direct writes).
+# Keep in mind that any output sent to sys.stdout gets redirected to sys.stderr instead.
+# This setting is mainly for debugging (not for production systems).
+WSGIRestrictStdout Off
 WSGIApplicationGroup %{GLOBAL}
 WSGIProcessGroup rucio
 
@@ -21,8 +30,6 @@ WSGIProcessGroup rucio
  SSLOptions +StdEnvVars
 
  LogLevel debug authz_core:info ssl:info socache_shmcb:info
-
- ErrorLog /var/log/rucio/httpd_error_log
  TransferLog /var/log/rucio/httpd_access_log
 
  WSGIScriptAlias /  /opt/rucio/lib/rucio/web/rest/main.py
--- a/lib/rucio/common/logging.py
+++ b/lib/rucio/common/logging.py
@@ -16,6 +16,7 @@ import functools
 import itertools
 import json
 import logging
+import os
 import re
 import sys
 from traceback import format_tb
@@ -392,18 +393,79 @@ def rucio_log_formatter(process_name: Optional[str] = None) -> RucioFormatter:
 
 def setup_logging(application: Optional["Flask"] = None, process_name: Optional[str] = None) -> None:
     """
-    Configures the logging by setting the output stream to stdout and
-    configures log level and log format.
+    Configure the logging sink/format/level.
+
+    By default, log records are emitted to stdout. Stream selection can be configured via:
+      - [common] logstream = stdout|stderr
+      - env RUCIO_LOG_STREAM = stdout|stderr (overrides config)
+
+    Additionally, *process-wide* sys.stdout writes (e.g. print(), direct sys.stdout.write())
+    can be redirected to sys.stderr via:
+      - [common] redirect_stdout_to_stderr = true|false
+      - env RUCIO_LOGGING_REDIRECT_STDOUT_TO_STDERR = 1/0/true/false/on/off (overrides config)
+
+    Important notes:
+      - redirect_stdout_to_stderr affects only sys.stdout writes, not the logging handler output.
+        To unify *all* output to a single stream, set both:
+          logstream = stderr
+          redirect_stdout_to_stderr = true
+      - Redirecting sys.stdout is a process-wide side effect and is intended for debugging /
+        containerized deployments. Avoid enabling it in production unless you explicitly want it.
     """
     config_loglevel = getattr(logging, config_get('common', 'loglevel', raise_exception=False, default='DEBUG').upper())
 
-    stdouthandler = logging.StreamHandler(stream=sys.stdout)
-    stdouthandler.setFormatter(rucio_log_formatter(process_name=process_name))
-    stdouthandler.setLevel(config_loglevel)
-    logging.basicConfig(level=config_loglevel, handlers=[stdouthandler])
+    # Decide the handler stream (default stdout)
+    # Keep references to the streams before we potentially swap them around so the
+    # handler can keep writing to the original destination.
+    original_stdout = sys.stdout
+    original_stderr = sys.stderr
+
+    stream_choice = (os.getenv('RUCIO_LOG_STREAM', '').strip().lower()
+                     or str(config_get('common', 'logstream', raise_exception=False, default='stdout')).strip().lower())
+    stream = original_stderr if stream_choice in ('stderr', 'err', '2') else original_stdout
+
+    # Optionally route all process-level stdout to stderr
+    redirect_cfg = config_get_bool('common', 'redirect_stdout_to_stderr', raise_exception=False, default=False)
+    redirect_env_raw = os.getenv('RUCIO_LOGGING_REDIRECT_STDOUT_TO_STDERR', '')
+    if redirect_env_raw.strip():
+        redirect = redirect_env_raw.strip().lower() in ('1', 'true', 'yes', 'on')
+    else:
+        redirect = redirect_cfg
+
+    if redirect and sys.stdout is not sys.stderr:
+        try:
+            sys.stdout.flush()
+        except Exception:
+            pass
+        sys.stdout = sys.stderr
+
+    handler = logging.StreamHandler(stream=stream)
+    handler.setFormatter(rucio_log_formatter(process_name=process_name))
+    handler.setLevel(config_loglevel)
+    logging.basicConfig(level=config_loglevel, handlers=[handler])
 
     if application:
-        application.logger.addHandler(stdouthandler)
+        app_logger = application.logger
+
+        # Clear only Flask's default handler(s) to avoid duplicate emissions.
+        # Do not wipe custom user-provided handlers.
+        removed_flask_default = False
+        try:
+            from flask.logging import default_handler  # type: ignore
+        except Exception:
+            default_handler = None
+
+        if default_handler and default_handler in app_logger.handlers:
+            app_logger.removeHandler(default_handler)
+            removed_flask_default = True
+
+        # In Flask defaults, propagate=False and the default handler is the only handler.
+        # After removing it, attach our handler so app logs continue to be emitted.
+        if removed_flask_default and not app_logger.handlers:
+            app_logger.addHandler(handler)
+            app_logger.propagate = False
+
+        app_logger.setLevel(config_loglevel)
 
 
 def formatted_logger(innerfunc: 'Callable', formatstr: str = "%s") -> 'Callable':
--- a/tests/rucio/common/test_setup_logging.py
+++ b/tests/rucio/common/test_setup_logging.py
@@ -1,0 +1,211 @@
+# Copyright European Organization for Nuclear Research (CERN) since 2012
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import io
+import logging
+import sys
+import types
+
+import pytest
+
+from rucio.common import logging as rucio_logging
+
+
+@pytest.fixture(autouse=True)
+def _reset_root_handlers_and_streams():
+    """Ensure each test has a clean logging configuration and doesn't leak std stream changes."""
+    original_stdout, original_stderr = sys.stdout, sys.stderr
+    logging.root.handlers = []
+    yield
+    logging.root.handlers = []
+    sys.stdout = original_stdout
+    sys.stderr = original_stderr
+
+
+def _mock_config_get(value_map):
+    def _mock(section, option, raise_exception=False, default=None):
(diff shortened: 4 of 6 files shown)


Geogouz:
Coverage handled

etc/docker/test/extra/rucio.conf
```diff
@@ -1,9 +1,17 @@
 SSLSessionCache  shmcb:/var/log/httpd/ssl_scache(512000)
 
 Listen 443
+# In mod_wsgi *daemon* mode, the daemon workers write stderr to the *main* Apache
+# error log (not the vhost error log). By pointing the main ErrorLog to the same file
+# as the vhost below, we unify all REST outputs
+ErrorLog /var/log/rucio/httpd_error_log
```
rdimaio: Note for reviewing: these changes are the same done here to the containers dev rucio.conf https://github.com/rucio/containers/blob/master/dev/rucio.conf

etc/docker/dev/docker-compose.yml
```diff
@@ -62,6 +62,9 @@ services:
       - RUCIO_SOURCE_DIR=/rucio_source
       - RDBMS
       - DEV_PROFILES
+      - RUCIO_LOG_STREAM=stderr
+      - RUCIO_REDIRECT_STDOUT_TO_STDERR=true
+      - PYTHONUNBUFFERED=true
```
rdimaio: With these env variables that affect the stdout (https://stackoverflow.com/questions/59812009/what-is-the-use-of-pythonunbuffered-in-docker-file): does this have any implication on the CLI usage of Rucio? e.g. if I go in the dev container and execute a rucio command, would it redirect the output of the command to stderr?

lib/rucio/common/logging.py
```diff
@@ -392,18 +393,47 @@ def rucio_log_formatter(process_name: Optional[str] = None) -> RucioFormatter:
 
 def setup_logging(application: Optional["Flask"] = None, process_name: Optional[str] = None) -> None:
     """
-    Configures the logging by setting the output stream to stdout and
-    configures log level and log format.
+    Configure logging sink/format/level.
+    Defaults to stdout (backwards compatible).
+    Dev/test/CI can override via:
```
rdimaio: Is this meant in the sense that this overriding can only be done in dev/test/CI? because if it's just a matter of changing the config or env variables, i assume you could easily do this in prod too

lib/rucio/common/logging.py
```diff
@@ -392,18 +393,47 @@ def rucio_log_formatter(process_name: Optional[str] = None) -> RucioFormatter:
...
+      - [common] logstream = stdout|stderr
+      - [common] redirect_stdout_to_stderr = true|false
+      - env RUCIO_LOG_STREAM=stdout|stderr
+      - env RUCIO_REDIRECT_STDOUT_TO_STDERR=1/true/on
```
rdimaio: could you clarify that the first 2 are config options and the last 2 are env variables?

etc/docker/dev/docker-compose.yml
Geogouz: `RUCIO_REDIRECT_STDOUT_TO_STDERR` is only consulted by Rucio’s `setup_logging` function in `lib/rucio/common/logging.py`. When this flag is truthy, `setup_logging` rebinds `sys.stdout` to `sys.stderr` before creating the logging handler, so all subsequent writes to stdout (not just log messages) will go to stderr for any process that calls `setup_logging`. The dev Docker Compose service enables this variable.

However, even with `RUCIO_REDIRECT_STDOUT_TO_STDERR=true` present there, running `rucio …` interactively in the dev container will not redirect normal command output to stderr because the `bin/rucio` entrypoint initializes logging via `rucio.common.utils.setup_logger` (which constructs its own `StreamHandler`).

Regarding the `PYTHONUNBUFFERED=true`, this setting is supposed to simply disable Python’s stdio buffering, causing both stdout and stderr writes to flush immediately.

lib/rucio/common/logging.py
Geogouz: `[common] logstream` and `[common] redirect_stdout_to_stderr` are configuration options read from the Rucio config file. They are the defaults that apply when the equivalent env vars are absent. On the other hand, `RUCIO_LOG_STREAM` and `RUCIO_REDIRECT_STDOUT_TO_STDERR` are environment variables and if set, they override the config-file values for the handler stream and for whether `stdout` is rebound to `stderr`.

lib/rucio/common/logging.py
Geogouz: “Dev/test/CI can override” is basically descriptive here, not restrictive. Productions could use these the same way. But current implementation leaves production behavior unchanged.

etc/docker/dev/docker-compose.yml
rdimaio: > RUCIO_REDIRECT_STDOUT_TO_STDERR is only consulted by Rucio’s setup_logging

I then would be in favour of renaming this flag to specify that this is only about logging - maybe even something like `RUCIO_LOGGING_REDIRECT_STDOUT_TO_STDERR`

lib/rucio/common/logging.py
rdimaio: Sorry for the ambiguity, I meant if you could clarify this in the docstring itself, not in the PR

Geogouz: ready

etc/docker/dev/docker-compose.yml
Geogouz: ready

rdimaio (dismissed):

tools/merge_rucio_configs.py
```diff
@@ -100,7 +100,10 @@ def merge_configs(source_file_paths, dest_file_path, use_env=True, logger=loggin
                         file_config = fix_multi_word_sections(yaml.safe_load(f))
                         parser.read_dict(file_config)
                 elif path.is_file() or file_path.suffix in ['.ini', '.cfg', '.config']:
-                    local_parser = configparser.ConfigParser()
+                    # The merge utility only needs literal values.  Disabling interpolation keeps
+                    # strings that contain percent-based tokens (for example logging formatters)
+                    # from being parsed as references to other options.
+                    local_parser = configparser.ConfigParser(interpolation=None)
```
bziemons: Thinking about config merging is done at container startup: is it clear that no community is using this feature? Does it have to be changed in this PR regarding logging?

etc/docker/test/extra/rucio.conf
bziemons, line 8: using the same file here as in the vhost is not explicitly said to be supported in the [apache documentation](https://httpd.apache.org/docs/2.4/mod/core.html#errorlog)

what interesting contents may we even find from the non-vhost logging?

bziemons (changes requested):

etc/docker/test/extra/rucio.conf
Geogouz: I will ditch the ErrorLog of the vhost to inherit the global one. Apache’s logging docs state that vhosts without logging directives still send logs to the main server logs. So we should be covered.

Geogouz: `mod_wsgi` initialization and interpreter lifecycle stuff at least. I have forgotten which issue I was working on and I needed to be considering that file, but for debugging, I think it can save time in general.

Geogouz: By the way.. keep in mind that the actual configuration being currently applied when the dev container used, is coming from the container side (see https://github.com/rucio/containers/blob/master/dev/rucio.conf).. and not this file I am touching here.

The reason why I am updating this file is to keep them a bit in sync. I do find the way things are configured (i.e. containers/rucio tangling) though still a bit ugly.. but this is another topic of discussion. Regardless, I could also ditch the ErrorLog of the vhost to inherit the global one also at the container side.

tools/merge_rucio_configs.py
Geogouz: There are couple sketchy points suggesting that we should better remove it. More specifically, if you see in 

https://github.com/rucio/rucio/blob/e9a5a52fbb5e5b9439eeaea57e691ba08756e68e/lib/rucio/common/logging.py#L384-L385

we expect to he able to read the `logformat` configuration. At the same time, we offer examples about this that look like..

https://github.com/rucio/rucio/blob/e9a5a52fbb5e5b9439eeaea57e691ba08756e68e/etc/rucio.cfg.template#L41

and

https://github.com/rucio/rucio/blob/e9a5a52fbb5e5b9439eeaea57e691ba08756e68e/etc/rucio_multihost.cfg#L5

However, from my understanding this can only work in `merge_configs(..)` if interpolation is disabled..

lib/rucio/common/logging.py
```diff
@@ -392,18 +393,79 @@ def rucio_log_formatter(process_name: Optional[str] = None) -> RucioFormatter:
...
+        except Exception:
+            default_handler = None
+
+        if default_handler and default_handler in app_logger.handlers:
+            app_logger.removeHandler(default_handler)
+            removed_flask_default = True
+
+        # In Flask defaults, propagate=False and the default handler is the only handler.
+        # After removing it, attach our handler so app logs continue to be emitted.
+        if removed_flask_default and not app_logger.handlers:
```
bziemons, line 464: why does this need the extra boolean `removed_flask_default` AND a comment explaining it (line 450+451)? Can you merge this with the previous if block (line 458+) with `if len(app_logger.handlers) == 0: …`

tools/merge_rucio_configs.py
bziemons: but those examples clearly escape the interpolation by using duplicate `%%`, I do not understand your point. Instead you also use escaped interpolation in your config, why?

tests/rucio/common/test_setup_logging.py
```diff
@@ -0,0 +1,211 @@
...
+
+
+def _install_flask_default_handler(monkeypatch, default_handler):
+    """
+    Ensure `from flask.logging import default_handler` resolves to `default_handler`
+    without requiring Flask to be installed and without replacing the real `flask`
+    module when it is available.
+    """
+    try:
+        # If Flask is installed, just patch the attribute on the real module.
```
bziemons, line 49: bogus comment, wrong scope

tests/rucio/common/test_setup_logging.py
```diff
@@ -0,0 +1,211 @@
...
+        import flask.logging as flask_logging  # type: ignore
+    except Exception:
+        # Flask is not installed (or cannot be imported). Provide minimal stubs
```
bziemons, line 52: why are we patching a non-existent Flask module that cannot be used in testing, because it isn't installed? Shouldn't this method be a noop, if Flask is not installed instead?

lib/rucio/common/logging.py
```diff
@@ -392,18 +393,79 @@ def rucio_log_formatter(process_name: Optional[str] = None) -> RucioFormatter:
...
     """
     config_loglevel = getattr(logging, config_get('common', 'loglevel', raise_exception=False, default='DEBUG').upper())
 
-    stdouthandler = logging.StreamHandler(stream=sys.stdout)
-    stdouthandler.setFormatter(rucio_log_formatter(process_name=process_name))
-    stdouthandler.setLevel(config_loglevel)
-    logging.basicConfig(level=config_loglevel, handlers=[stdouthandler])
+    # Decide the handler stream (default stdout)
+    # Keep references to the streams before we potentially swap them around so the
+    # handler can keep writing to the original destination.
```
bziemons, line 419: I do not understand these comments in the scope of this commit

bziemons (changes requested):

tools/merge_rucio_configs.py
Geogouz: > but those examples clearly escape the interpolation by using duplicate `%%`, I do not understand your point. Instead you also use escaped interpolation in your config, why?

Check here (and compare with the config in test-commit that I included just for the demonstration):
https://github.com/rucio/rucio/actions/runs/21726209221/job/62668847270?pr=8212#step:7:2126

My point is that, from what I understand at least, our `merge_configs()` cannot handle properly the interpolation happening when reading the config file. Despite the clear examples that you also noticed. Maybe we could deactivate the interpolation from there since it is about testing only.

tests/rucio/common/test_setup_logging.py
Geogouz: I would argue it is to the point.

If Flask is installed
`import flask.logging as flask_logging`

just patch the attribute on the real module
`monkeypatch.setattr(flask_logging, "default_handler", default_handler, raising=False)`

but I can reduce the commenting since preferred.

bziemons: > just patch the attribute on the real module
> `monkeypatch.setattr(flask_logging, "default_handler", default_handler, raising=False)`

this is exactly the part that is not where the comment is and happens in any case (try, catch).

Geogouz: To understand, your point is that (if it was not to be removed), you would suggest it to placed before the `try:` or before the `except`?

bziemons: I would probably add a comment before the try that explains what you are trying to do (ultimately ensure default_handler on a "flask" module is set). But this obviously depends on the response of [my other criticism as well](https://github.com/rucio/rucio/pull/8212#discussion_r2770005678).

Geogouz: got it

tests/rucio/common/test_setup_logging.py
Geogouz: In `setup_logging()`, if `default_handler` from Flask can’t be imported, the logic is effectively ndeed noop. But since Flask is actually optional, the patching here is basically to exercise in test environments without Flask, the part:

```
  if default_handler and default_handler in app_logger.handlers:
    app_logger.removeHandler(default_handler)
    if not app_logger.handlers:
        # Flask defaults to propagate=False, so keep emitting after removing its only handler.
        app_logger.addHandler(handler)
        app_logger.propagate = False
```
 
…which is the “Flask present” side of the behavior.   I think this is more holistic and we don’t have something to lose given that this is a test. But if we are fine doing the assumption that our testing environments will always have Flask around, we can simplify the helper to just import the real `flask.logging` and monkeypatch `default_handler` there. I'm fine either way.

lib/rucio/common/logging.py
Geogouz: Those comments describe the purpose of `original_stdout` and `original_stderr` in this commit. In the notes, I mention: “redirect_stdout_to_stderr affects only `sys.stdout` writes, not the logging handler output”.  

Given this, keeping references to the streams helps because later we may do “sys.stdout = sys.stderr”. So with this snapshot, the logging handler can be pinned to the chosen destination (happening inside `handler = logging.StreamHandler(stream=stream)`) even if `sys.stdout` later changes.

  In other words, what those comments say is “Capture stdout/stderr before optionally rebinding `sys.stdout`, so the logging handler stays pinned to the selected sink and doesn’t follow stdout redirection”.

tests/rucio/common/test_setup_logging.py
bziemons: What I do not understand is why set a `default_handler` on a Flask module which we cannot use for the tests we run. I am not assuming that we always have Flask around in tests, but if we do not have Flask why add a logging handler on it?

tools/merge_rucio_configs.py
bziemons: I think this fix for `tools/merge_rucio_configs.py` should be in a different issue and PR. It seems to me that the tool parses the configuration twice, so your fix is correct, but why didn't you say so in the first place?

PR #8213, opened 2025-11-13
Database: Support JSON natively in Rucio; Fixes rucio #7648

### LINKED ISSUE #7648: Support native json for Oracle DB when compatible version
Oracle introduces native json types in 21c.

Our current logic is to serialize the json as a string for oracle db, since json types were not previously supported.

Instead of just checking the dialect, I think we could have a switch based on the version of oracle (and if the feature is enabled, if it can be disabled, not sure if this is possible).

I also add the metadata tag to this issues as this is strongly related to metadata.

Geogouz:
**Draft Notice:** This works but kept frozen until we merge #8207 and #8212.
---

Handles #7648

In this PR, I have already included A) https://github.com/rucio/rucio/pull/8207 and B) https://github.com/rucio/rucio/pull/8212 which need to be merged first (and then I can rebase to reduce current PR-diffs). Therefore, below I am summarising what changes are introduced only by the new commits targeting the #7648 issue specifically. These are:

_- Testing: Comment out unsupported dialect check in `test_opendata.py` (316e932cdfd03fcaaa86e02a946db5d1dac07975)_
_- Core & Internals: Add Oracle JSON column validation and centralized startup checks (f0f3269c3641d692555f3abd1831f19e384ba589)_
_- Metadata: Make JSON type native and database agnostic in Rucio; rucio#7648 (d4c41b6356383fc5a113d30aa35733646dfc3276)_

**Summary of Changes**
- Introduced `rucio.common.startup_checks_catalog` to centralize there our startup checks. Although https://github.com/rucio/rucio/pull/8207 supports check definitions & registrations from everywhere, it might be a good idea to just have all our startup checks at one place.
- Added a global check (which is basically the first one that is using the new startup-check framework introduced by #8207) to enforce native Oracle JSON columns by querying for legacy CLOB-backed JSON fields in Oracle 21c+. Passing this check before a service starts, simplifies many metadata-related issues for us.
- Added the new startup check catalog into both daemon startup and the Flask REST application so these diagnostics run automatically when services launch.
- Updated the SQLAlchemy JSON type to select native JSON storage for Oracle 21c+ while preserving text serialization on earlier versions or other backends (to ensure consistent handling across dialects).
- Switched the `DidMeta` and `DeletedDidMeta` models to use the unified JSON column type and adjusted the JSON metadata plugin to rely on the new capability checks while forcing ORM state changes when updating metadata payloads.

**Also regarding Testing Updates**
- Added unit tests for the startup check catalog to cover Oracle detection.
- Removed the previously hard-coded skip markers from the OpenData integration tests so metadata and CLI workflows execute regardless of database dialect, reflecting broader JSON support.

### DIFF
--- a/etc/docker/dev/docker-compose.yml
+++ b/etc/docker/dev/docker-compose.yml
@@ -62,6 +62,9 @@ services:
       - RUCIO_SOURCE_DIR=/rucio_source
       - RDBMS
       - DEV_PROFILES
+      - RUCIO_LOG_STREAM=stderr
+      - RUCIO_LOGGING_REDIRECT_STDOUT_TO_STDERR=true
+      - PYTHONUNBUFFERED=true
     depends_on:
       oracle:
         condition: service_healthy
--- a/etc/docker/test/extra/rucio.conf
+++ b/etc/docker/test/extra/rucio.conf
@@ -1,9 +1,17 @@
 SSLSessionCache  shmcb:/var/log/httpd/ssl_scache(512000)
 
 Listen 443
+# In mod_wsgi *daemon* mode, the daemon workers write stderr to the *main* Apache
+# error log (not the vhost error log). By pointing the main ErrorLog to the same file
+# as the vhost below, we unify all REST outputs
+ErrorLog /var/log/rucio/httpd_error_log
 
 WSGIRestrictEmbedded On
 WSGIDaemonProcess rucio processes=4 threads=4
+# Allow the app to access and write to sys.stdout (e.g., via print statements or direct writes).
+# Keep in mind that any output sent to sys.stdout gets redirected to sys.stderr instead.
+# This setting is mainly for debugging (not for production systems).
+WSGIRestrictStdout Off
 WSGIApplicationGroup %{GLOBAL}
 WSGIProcessGroup rucio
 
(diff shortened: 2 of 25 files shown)


lobis:
LGTM but I will wait until the rebase is ready to properly review it. Thanks @Geogouz !

bari12:
Should this be reviewed? If so, please take it out of draft mode.

Geogouz:
> Should this be reviewed? If so, please take it out of draft mode.

Because this requires the review/merging of two other PRs (https://github.com/rucio/rucio/pull/8207 and https://github.com/rucio/rucio/pull/8212), I introduced it here as a draft to control the merging order. Right now, this PR includes all commits involved in these two other PRs (so it is working if one wants to test it). But once we merge the other ones, I will rebase this one to reduce the number of commits it contains and leave only the 3 commits.

martynia:
Could you make constraint name guessed from a column name (and possibly a table name) to make this scalable:
[def _oracle_json_constraint
](https://github.com/rucio/rucio/blob/2262ce73fffff1b163aa3bd56d756f4051d7496a/lib/rucio/db/sqla/models.py#L99)

I have 11 tables to add `JSON` fields to.

Geogouz:
> Could you make constraint name guessed from a column name (and possibly a table name) to make this scalable: [def _oracle_json_constraint ](https://github.com/rucio/rucio/blob/2262ce73fffff1b163aa3bd56d756f4051d7496a/lib/rucio/db/sqla/models.py#L99)
> 
> I have 11 tables to add `JSON` fields to.

Need to see how to best support such cases. But I can figure it out. Will let u know. I have started working on my quite big backlog of metadata related issues and soon I will reach this one too to adjust and reopen it.

PR #8214, opened 2025-11-13
Refactoring Stomp Controller

### LINKED ISSUE #6423: factorize duplicate messaging code into a common module or class

alexanderrichards:
fixes #6423

### DIFF
--- a/lib/rucio/common/stomp_controller.py
+++ b/lib/rucio/common/stomp_controller.py
@@ -1,0 +1,158 @@
+# Copyright European Organization for Nuclear Research (CERN) since 2012
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import json
+import logging
+import random
+import socket
+from typing import TYPE_CHECKING, Any, Optional
+
+import stomp
+
+if TYPE_CHECKING:
+    from collections.abc import Callable, Sequence
+
+    from rucio.core.monitor import MetricManager
+
+
+class StompController:
+    """
+    Common controller for managing Stomp connections to brokers.
+    """
+
+    def __init__(
+        self,
+        brokers: "Sequence[str]",
+        port: int,
+        use_ssl: bool = True,
+        vhost: Optional[str] = None,
+        username: Optional[str] = None,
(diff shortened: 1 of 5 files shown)


Geogouz:
We should link this to an issue that explains the motivation.

bziemons:

* There is no GitHub issue associated with this PR.

lib/rucio/daemons/tracer/kronos.py
```diff
@@ -361,50 +362,52 @@ def run_once_kronos_file(heartbeat_handler: HeartbeatHandler, stomp_conn_mngr: S
...
+                         subscription_id=subscription_id,
+                         excluded_usrdns=excluded_usrdns,
+                         dataset_queue=dataset_queue,
+                         bad_files_patterns=bad_files_patterns,
+                         logger=logger),
+        subscription_id=subscription_id,
+        ack='client-individual',
+        metric=METRICS,
+        headers={'activemq.prefetchSize': prefetch_size})
+    controller.disconnect()
```
bari12, line 410: This subscribes and then directly disconnects? I don't think that works

lib/rucio/common/stomp_controller.py
```diff
@@ -0,0 +1,158 @@
...
+        metric: "MetricManager|None" = None,
+        logger: "Callable|None" = None
+    ) -> None:
+        if logger is None:
+            logger = self.logger
+        conn = random.sample(self.connections, 1)[0]
+        if not conn.is_connected():
+            host_port = conn.transport._Transport__host_and_ports[0][0]
+
+            logger(logging.INFO, 'connecting to %s' % host_port[0])
```
bari12, line 129: all the host_port[0] don't work? Don't these just give the first character of the host variable? since
`host_port = conn.transport._Transport__host_and_ports[0][0]`

lib/rucio/daemons/tracer/kronos.py
```diff
@@ -361,50 +362,52 @@ def run_once_kronos_file(heartbeat_handler: HeartbeatHandler, stomp_conn_mngr: S
...
         vhost=vhost,
-        reconnect_attempts=reconnect_attempts,
+        username=username,
+        password=password,
         ssl_key_file=ssl_key_file,
         ssl_cert_file=ssl_cert_file,
         timeout=sleep_time,
-        heartbeats=(0, 5000),
-        logger=logger,
+        reconnect_attempts=reconnect_attempts,
```
bari12, line 390: Does this mean we lost heartbeats functionality in the new StomController?

bari12 (changes requested):
First, big apologies that my review took so long.
I have to say I don't really correctly follow some of these changes, but to me it looks like this just swallows quite a lot of things. It's very hard to judge since we clearly miss tests here. So my suggestion would be to really start out with much more tests for the entire messaging stack.

Issue #8168, opened 2025-11-03
Bug in `TestDidMetaElastic` when upgrading elasticsearch 8.15.1 -> 9.1.2,  TypeError: __init__() got an unexpected keyword argument 'timeout'


rdimaio:
### Description

```
==================================== ERRORS ====================================
__________ ERROR at setup of TestDidMetaElastic.test_set_get_metadata __________
[gw0] linux -- Python 3.9.21 /opt/venv/bin/python
tests/test_did_meta_plugins.py:312: in elastic_meta
    return ElasticDidMeta(
/rucio_source/lib/rucio/core/did_meta_plugins/elasticsearch_meta.py:93: in __init__
    self.client = Elasticsearch(**self.es_config)
E   TypeError: __init__() got an unexpected keyword argument 'timeout'
___________ ERROR at setup of TestDidMetaElastic.test_list_did_meta ____________
[gw1] linux -- Python 3.9.21 /opt/venv/bin/python
tests/test_did_meta_plugins.py:312: in elastic_meta
    return ElasticDidMeta(
/rucio_source/lib/rucio/core/did_meta_plugins/elasticsearch_meta.py:93: in __init__
    self.client = Elasticsearch(**self.es_config)
E   TypeError: __init__() got an unexpected keyword argument 'timeout'
__________ ERROR at setup of TestDidMetaElastic.test_delete_metadata ___________
[gw0] linux -- Python 3.9.21 /opt/venv/bin/python
tests/test_did_meta_plugins.py:312: in elastic_meta
    return ElasticDidMeta(
/rucio_source/lib/rucio/core/did_meta_plugins/elasticsearch_meta.py:93: in __init__
    self.client = Elasticsearch(**self.es_config)
E   TypeError: __init__() got an unexpected keyword argument 'timeout'
```

example run: https://github.com/rdimaio/rucio/actions/runs/19041509815/job/54380701996#step:28:36

### Steps to reproduce

NA

rdimaio:
Seems that `timeout` was deprecated in 9.0.0 https://www.elastic.co/docs/release-notes/elasticsearch/clients/python/breaking-changes

rdimaio:
timeout is now request_timeout

rdimaio:
Should we even upgrade to elastic 9 though, from a license / FOSS perspective?

rdimaio:
Migration to elastic 9 is actually blocked anyway https://github.com/rucio/rucio/issues/8169

PR #8221, opened 2025-11-18
Testing: upgrade_to_elastic_v9.0.0 #8169

### LINKED ISSUE #8169: Upgrading the dev container from Elastic 7 to 9 causes the container to hang indefinitely
See: https://github.com/rdimaio/rucio/actions/runs/19050623966/job/54409762618

panta-123:
- upgrade to elasticsearch v9
- add mem limit in elasticsearch.

container starts in forked action. Tests in forked were failing , so will see in upstream.

### DIFF
--- a/etc/docker/dev/docker-compose.yml
+++ b/etc/docker/dev/docker-compose.yml
@@ -133,9 +133,13 @@ services:
 
   elasticsearch:
     container_name: ${RUCIO_ELASTICSEARCH_CONTAINER_NAME:-dev-elasticsearch-1}
-    image: docker.io/elasticsearch:7.8.0
+    image: docker.elastic.co/elasticsearch/elasticsearch:9.0.0
     environment:
       - discovery.type=single-node
+      - xpack.security.enabled=false
+      - xpack.security.enrollment.enabled=false
+      - "ES_JAVA_OPTS=-Xms1g -Xmx1g"
+    mem_limit: 2g
     healthcheck:
       test: [ "CMD", "curl", "-f", "http://localhost:9200/_cluster/health?wait_for_status=yellow" ]
       interval: 10s
--- a/etc/docker/dev/docker-compose.yml
+++ b/etc/docker/dev/docker-compose.yml
@@ -138,6 +138,7 @@ services:
       - discovery.type=single-node
       - xpack.security.enabled=false
       - xpack.security.enrollment.enabled=false
+      - cluster.routing.allocation.disk.threshold_enabled=false
       - "ES_JAVA_OPTS=-Xms1g -Xmx1g"
     mem_limit: 2g
     healthcheck:
--- a/tests/test_hermes.py
+++ b/tests/test_hermes.py
@@ -59,7 +59,7 @@ class MyListener:
         {
             "table_content": [
                 ("hermes", "services_list", "influx,activemq,elastic,email"),
-                ("hermes", "elastic_endpoint", "http://elasticsearch:9200/ddm_events/doc/_bulk"),
+                ("hermes", "elastic_endpoint", "http://elasticsearch:9200/ddm_events/_bulk"),
                 ("hermes", "influxdb_endpoint", "http://influxdb:8086/api/v2/write?org=rucio&bucket=rucio"),
                 ("hermes", "influxdb_token", "mytoken"),
                 ("messaging-hermes", "destination", "/queue/events"),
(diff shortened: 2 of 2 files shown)


panta-123:
@rdimaio , All test passed. have to add some env contrain to make it less resource heavy and work under resource limit.

rdimaio (commented):
Looks good, but it might be worth getting the approval of component leads with components in which Elastic is involved (I guess messaging and metadata?)

panta-123:
Elastic metadata is what I am mantaining. The elastic metadata move to elastic 9 will be handled in https://github.com/rucio/rucio/issues/8168 .

So may be a review from monitoring component lead.

panta-123:
@mlassnig , tagging you for input as it involves monitoring  or may be @maany ?

Geogouz:
Can you rebase please? I can give it a check.

panta-123:
> Can you rebase please? I can give it a check.

done.

Geogouz:
Isn't the Hermès bulk path change breaking the backward compatibility?

panta-123:
> Isn't the Hermès bulk path change breaking the backward compatibility?

I belive not. The path is configurable. And works for elastic7 as it was before as there is no code changes.
Here just the testing infrastructure is updated to latest.

Geogouz:
> > Isn't the Hermès bulk path change breaking the backward compatibility?
> 
> I belive not. The path is configurable. And works for elastic7 as it was before as there is no code changes. Here just the testing infrastructure is updated to latest.

I assume that the Hermes daemon reads `hermes.elastic_endpoint` from configuration. See `elastic_endpoint = config_get("hermes", "elastic_endpoint", False, None)` inside the hermes daemon. I would expect that here, all current paths are configured for the previous endpoint. In that case, we should warn about this change, mentioning what might need to change in the configurations.

But maybe I'm missing something. @mlassnig should indeed have a better opinion about this.

panta-123:
> > > Isn't the Hermès bulk path change breaking the backward compatibility?
> > 
> > 
> > I belive not. The path is configurable. And works for elastic7 as it was before as there is no code changes. Here just the testing infrastructure is updated to latest.
> 
> I assume that the Hermes daemon reads `hermes.elastic_endpoint` from configuration. See `elastic_endpoint = config_get("hermes", "elastic_endpoint", False, None)` inside the hermes daemon. I would expect that here, all current paths are configured for the previous endpoint. In that case, we should warn about this change, mentioning what might need to change in the configurations.
> 
> But maybe I'm missing something. @mlassnig should indeed have a better opinion about this.

The operator needs to change the endpoint only if they upgrade to elastic 9. If no elastic upgrade everything is expected to work as it is.

tests/test_hermes.py
```diff
@@ -59,7 +59,7 @@ def on_message(self, frame):
         {
             "table_content": [
                 ("hermes", "services_list", "influx,activemq,elastic,email"),
-                ("hermes", "elastic_endpoint", "http://elasticsearch:9200/ddm_events/doc/_bulk"),
+                ("hermes", "elastic_endpoint", "http://elasticsearch:9200/ddm_events/_bulk"),
```
Geogouz, line 62: If something breaks with this kind of required path change, it suggests I guess that the container in our docker-compose was being used in a non-expected/supported way (and thus, we shouldn't mind).

Geogouz (dismissed):
Looks fine to me.

Geogouz:
Note for future PRs: 
whenever you think that no changes are required and tests are passing, feel free to remove that label on your own. It would be quicker for reviewers to discover this PR. (personally, I am filtering out PRs with this label and check only the rest)

etc/docker/dev/docker-compose.yml
```diff
@@ -133,9 +133,14 @@ services:
 
   elasticsearch:
     container_name: ${RUCIO_ELASTICSEARCH_CONTAINER_NAME:-dev-elasticsearch-1}
-    image: docker.io/elasticsearch:7.8.0
+    image:  docker.elastic.co/elasticsearch/elasticsearch:9.0.0
```
dchristidis: Apologies for nitpicking, but please amend your first commit to not introduce an additional space. Otherwise it can be merged.

dchristidis (changes requested):

etc/docker/dev/docker-compose.yml
bziemons: @panta-123 please make this change soon

panta-123: @bziemons , I did that changes and rebase.
Since commitlint was introduced recently, this PR was before that.
Do I need to change commit message for all commit in this PR?

panta-123: went ahead and change the commit message to new standard

dchristidis (approved):

PR #8222, opened 2025-11-18
Allow update to scope ownership, listing scopes with owner

### LINKED ISSUE #7830: No way to check or change ownership of a scope using Rucio client
### Description

There is currently no built-in functionality in Rucio to determine the owner of a given scope.

In practice, this creates issues in the following scenarios:

- A scope is mistakenly assigned to the wrong account during creation.

- The ownership of a scope needs to be transferred to another account (e.g., due to project transitions or deprecation of old accounts).

Currently, there is no straightforward way to change the ownership of a scope, nor to audit who owns a particular scope. This limits the flexibility and maintainability of Rucio namespace management.

### Motivation

- Clean up orphaned or misassigned scopes.

- Manage transitions of data ownership between projects or user accounts.

- Enable better governance and automation in data management.

### Change

- Add a command in the Rucio client to retrieve the owner/account associated with a scope.

- Enable a secure and auditable way to change the owner of a scope (possibly restricted to admin privileges only).

voetberg:
Second pass at #8111 Closes #7830 

*Should* be compatible between server and client versions - introduces a new endpoint for the client to call, so old versions of the client will still just get the scope names back for the list.

### DIFF
--- a/lib/rucio/cli/utils.py
+++ b/lib/rucio/cli/utils.py
@@ -41,6 +41,7 @@ from rucio.common.exception import (
     RSENotFound,
     RucioException,
     RuleNotFound,
+    ScopeNotFound,
     UnsupportedOperation,
 )
 from rucio.common.utils import extract_scope, setup_logger
@@ -262,11 +263,33 @@ class JSONType(click.ParamType):
             self.fail(f"Invalid JSON: {e}", param, ctx)
 
 
+def scope_exists(client: 'Client', scope: str) -> None:
+    possible_scopes = client.list_scopes()
+    if not len(list(possible_scopes)):
+        raise ScopeNotFound
+    if isinstance(list(possible_scopes)[0], str):  # TODO Backwards Compat - Remove in future releases - #8125
+        scopes = possible_scopes
+    else:
+        scopes = [s['scope'] for s in possible_scopes]  # type: ignore
+
+    if scope not in scopes:  # type: ignore - handled by the if isinstance
+        raise ScopeNotFound
+
+
 def get_scope(did: str, client: Client) -> tuple[str, str]:
     try:
         scope, name = extract_scope(did)
         return scope, name
     except TypeError:
-        scopes = client.list_scopes()
-        scope, name = extract_scope(did, scopes)
+        known_scopes = client.list_scopes()
+        if not len(list(known_scopes)):
+            raise ScopeNotFound
+        if isinstance(known_scopes, dict):
+            scopes = known_scopes.get('scope')  # type: ignore - does not accept 'scope' as a Literal['scope']
+            scope, name = extract_scope(did, scopes)
+        elif isinstance(known_scopes, list):
+            scope, name = extract_scope(did, known_scopes)  # type: ignore - Handled by the isinstance
+        else:
+            raise ScopeNotFound
+
         return scope, name
--- a/lib/rucio/client/downloadclient.py
+++ b/lib/rucio/client/downloadclient.py
@@ -34,7 +34,7 @@ from rucio.common.client import detect_client_location
 from rucio.common.config import config_get
 from rucio.common.constants import DEFAULT_VO
 from rucio.common.didtype import DID
-from rucio.common.exception import InputValidationError, NoFilesDownloaded, NotAllFilesDownloaded, RucioException
+from rucio.common.exception import InputValidationError, NoFilesDownloaded, NotAllFilesDownloaded, RucioException, ScopeNotFound
 from rucio.common.pcache import Pcache
 from rucio.common.utils import execute, extract_scope, generate_uuid, parse_replicas_from_file, parse_replicas_from_string, send_trace, sizefmt
 from rucio.rse import rsemanager as rsemgr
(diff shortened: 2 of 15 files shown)


lib/rucio/core/scope.py
```diff
@@ -193,3 +230,32 @@ def is_scope_owner(
...
+    if not check_scope(scope, session):
+        raise ScopeNotFound
+
+    stmt = update(
+        models.Scope
+    ).where(
+        models.Scope.scope == scope
+    ).values({
+        models.Scope.account: account
+    })
```
Geogouz, line 257: Shouldn't we update ownership with restricting the SQL `WHERE` clause by VO? If the same scope name exists in multiple VOs, calling this API will reassign all those scopes to the new account.

Geogouz: This doesn't coop well with multi-VO deployments (including `list_scopes_with_account` and the underlying `list_scopes`). Shouldn't it?

voetberg: It is my understanding that because scope is an Internal Scope object, which takes into account VO, this isn't a problem. You can see the same thing for several other Multi-VO implementations e.g., `add_scope`  doesn't include a VO  in it's new row addition.

Geogouz: aaah yes. Your understanding is correct indeed.

lib/rucio/client/scopeclient.py
Geogouz: Would `list[str]` be still accurate here though? Given that as soon as this reaches the `/scopes/owner` endpoint, it will receive basically dictionaries as items? `ScopeOwnershipList` does:

        for dictionary in scopes:
            res.append(dictionary)

lib/rucio/core/scope.py
```diff
@@ -123,6 +124,42 @@ def list_scopes(session: "Session", filter_: Optional[dict[str, Any]] = None) ->
     return list(session.execute(stmt).scalars().all())
 
 
+def list_scopes_with_account(filter_: Optional[dict[str, Any]] = None, *, session: "Session") -> "Iterable[dict[str, Any]]":
+    """
+    Lists all scopes.
+    :param filter_: Dictionary of attributes by which the input data should be filtered
+    :param session: The database session in use.
+
+    :returns: A of dictionaries with the scope and its owner
```
Geogouz: The description of `:returns: ` needs correction.

lib/rucio/web/rest/flaskapi/v1/scopes.py
```diff
@@ -140,14 +180,48 @@ def get(self, account):
...
+          200:
+            description: "OK"
+            content:
+              application/json:
+                schema:
+                  description: "All scopes for the account."
+                  type: array
+                  items:
+                    description: "A scope for the account."
+                    type: string
```
Geogouz: We emit objects though, right?

lib/rucio/client/scopeclient.py
voetberg: Yep, you're correct. Don't know why it didn't pick that up in the type checking

lib/rucio/web/rest/flaskapi/v1/scopes.py
voetberg: Yep, missed in the docstring copy-paste

lib/rucio/cli/utils.py
```diff
@@ -262,11 +263,24 @@ def convert(
...
 def get_scope(did: str, client: Client) -> tuple[str, str]:
     try:
         scope, name = extract_scope(did)
         return scope, name
     except TypeError:
-        scopes = client.list_scopes()
+        scopes_dict = client.list_scopes()
+        if isinstance(scopes_dict, dict):
+            scopes = scopes_dict.get('scope')  # type: ignore - does not accept 'scope' as a Literal['scope']
         scope, name = extract_scope(did, scopes)
```
Geogouz: In case `isinstance(scopes_dict, dict):` branch is skipped (e.g. `scopes_dict` is by accident something other than dict), `scopes` here will remain undefined.

lib/rucio/gateway/scope.py
```diff
@@ -92,3 +118,29 @@ def get_scopes(
...
+
+    :param account: New owner for the scope
+    :param scope: Scope to change
+    param issuer: User making the request
+    :param vo: VO to act on
+
+    """
+    kwargs = {'scope': scope, 'account': account}
+    with db_session(DatabaseOperationType.WRITE) as session:
+        auth_result = rucio.gateway.permission.has_permission(issuer=issuer, vo=vo, action='update_scope', kwargs=kwargs, session=session)
```
Geogouz, line 140: Right now, this is based on whatever policy-package fallbacks or `perm_default` we use in

https://github.com/rucio/rucio/blob/d7a8ffd3571c503f89318c2d078cb656ee44f914/lib/rucio/core/permission/generic.py#L44

I guess we should include `update_scope` there (same for `generic_multi_vo.py`).

lib/rucio/cli/utils.py
```diff
@@ -262,11 +263,24 @@ def convert(
             self.fail(f"Invalid JSON: {e}", param, ctx)
 
 
+def scope_exists(client: 'Client', scope: str) -> None:
+    possible_scopes = client.list_scopes()
+    if isinstance(possible_scopes[0], str):  # type: ignore  #TODO Backwards Compat - Remove in future releases - #8125
```
Geogouz: I think we should handle the empty-scope edge case because this would raise `IndexError`.

lib/rucio/client/downloadclient.py
```diff
@@ -1856,8 +1856,13 @@ def _split_did_str(self, did_str: str) -> tuple[str, str]:
             did_name = did[1]
         elif len(did) == 1:
             if self.extract_scope_convention == 'belleii':
-                scopes = [scope for scope in self.client.list_scopes()]
-                did_scope, did_name = extract_scope(did[0], scopes)
+                listed_scopes = [scope for scope in self.client.list_scopes()]
+                # TODO remove backwards compatiblity check. See #8125
+                if not isinstance(listed_scopes[0], str):
```
Geogouz, line 1863: Same empty-scope edge case should be handled here.

Geogouz:
@voetberg I see just couple minor (quite edge-case) fixes that we should better handle. Otherwise, this looks fine by me.

lib/rucio/gateway/scope.py
voetberg: Good point. Added in 6e958e43478b242d1287b3e26b94d94dc42c7e45

bari12 (approved):
- **Quality**: Agree 🟢
- **Security**: Agree 🟢
- **Testing**: Agree 🟢
- **Documentation**: Agree 🟢

# Notes for merger
-

Issue #8125, opened 2025-10-15
Remove backwards compataiblity for client/server listing scopes mismatch


voetberg:
As per the release policy, after #8111 there must compatibility between server and clients of the same line. 

Because of this, there is additional logic for handling listing either just scopes, or scopes and ownership. This should be removed after this release line.

PR #8225, opened 2025-11-19
Client: Generalize Belle-II specific check in download client

### LINKED ISSUE #7973: BelleII specific logic in download client
### Description

There are two positions in the download client that are explicitly testing for belleii in code to handle dids with slashes in them:

https://github.com/rucio/rucio/blob/4482eaa09bb40aa0dd9152fc0de2a569c037fd61/lib/rucio/client/downloadclient.py#L1654-L1655

https://github.com/rucio/rucio/blob/4482eaa09bb40aa0dd9152fc0de2a569c037fd61/lib/rucio/client/downloadclient.py#L1858-L1860

This should probably be replaced by a policy package based solution.

### Steps to reproduce

Inspect the code above

### Rucio Version

Current master

### Additional Information

_No response_

### LINKED ISSUE #8223: Downloading does not work with hierarchical did names and extract_scope change
### Description

After fixing https://github.com/rucio/rucio/issues/7979 via https://github.com/rucio/rucio/pull/8135, downloading is broken for the CTAO policy.

There seems to be no variant where downloading using the download client works.

Before #8135, did had to be `<scope>:<lfn>`.

Now the initial code in download_dids correctly calls extract_scope on the lfn, but deeper down, it again tries to simply split at a colon or dot:

https://github.com/rucio/rucio/blob/695df6e325b78991817ee91f439e3e7ac0159330/lib/rucio/client/downloadclient.py#L1348-L1350

```
(Pdb) download_client.download_dids([{'did': '/ctao.dpps.test/t_20251118_173322_851c/test_lfn'}])
------------------------------------------------------------- live log logreport -------------------------------------------------------------
2025-11-18 17:34:04.700 INFO root Processing 1 item(s) for input
2025-11-18 17:34:04.749 ERROR root DID does not exist: /ctao:/ctao.dpps.test/t_20251118_173322_851c/test_lfn
*** rucio.common.exception.DIDError: Error using DID type
Details: Object has invalid format after construction: ctao.dpps.test:/ctao:/ctao.dpps.test/t_20251118_173322_851c/test_lfn
(Pdb) download_client.download_dids([{'did': 't_20251118_173322_851c:/ctao.dpps.test/t_20251118_173322_851c/test_lfn'}])
2025-11-18 17:34:15.796 INFO root Processing 1 item(s) for input
2025-11-18 17:34:15.946 INFO root No preferred protocol impl in rucio.cfg: No section: 'download'
*** ValueError: DID 't_20251118_173322_851c:/ctao.dpps.test/t_20251118_173322_851c/test_lfn' does not match expected schema: /<VO Name>/<scope>/<path>.
```

### Steps to reproduce

See above, download a file with policy extract_scope defined.

### Rucio Version

38.5.1

### Additional Information

_No response_

maxnoe:
Now checks for presence of / in did name instead of configured extract scope function name.

Fixes #7973 

Fixes #8223, but needs changes in the policy package to still allow `<scope>:<lfn>`  dids in the extract scope function.

It would be great if we could make it work just with the `lfn`, but that seems to need more involved changes.

In general, the logic in download client is very hard to follow and dids are transformed multiple times and making API requests for each did (at least in the belle ii case: list all scopes for each did).

### DIFF
--- a/lib/rucio/client/downloadclient.py
+++ b/lib/rucio/client/downloadclient.py
@@ -34,7 +34,7 @@ from rucio.common.client import detect_client_location
 from rucio.common.config import config_get
 from rucio.common.constants import DEFAULT_VO
 from rucio.common.didtype import DID
-from rucio.common.exception import InputValidationError, NoFilesDownloaded, NotAllFilesDownloaded, RucioException, ScopeNotFound
+from rucio.common.exception import InputValidationError, NoFilesDownloaded, NotAllFilesDownloaded, RucioException
 from rucio.common.pcache import Pcache
 from rucio.common.utils import execute, extract_scope, generate_uuid, parse_replicas_from_file, parse_replicas_from_string, send_trace, sizefmt
 from rucio.rse import rsemanager as rsemgr
@@ -229,7 +229,6 @@ class DownloadClient:
         # tar -C <dest_dir_path> -xf <archive_file_path>  <did_name>
         extract_args = '-C %(dest_dir_path)s -xf %(archive_file_path)s %(file_to_extract)s'
         self.extraction_tools.append(BaseExtractionTool('tar', '--version', extract_args, logger=self.logger))
-        self.extract_scope_convention = config_get('common', 'extract_scope', False, None)
 
     def download_pfns(
         self,
@@ -304,6 +303,7 @@ class DownloadClient:
         logger = self.logger
         trace_custom_fields['uuid'] = generate_uuid()
 
+        scopes = self.client.list_scopes()
         logger(logging.INFO, 'Processing %d item(s) for input' % len(items))
         input_items = []
         for item in items:
@@ -322,7 +322,7 @@ class DownloadClient:
                 logger(logging.DEBUG, did_str)
                 raise InputValidationError('Cannot use PFN download with wildcard in DID')
 
-            did_scope, did_name = self._split_did_str(did_str)
+            did_scope, did_name = extract_scope(did_str, scopes)
             dest_dir_path = self._prepare_dest_dir(item.get('base_dir', '.'), did_scope, item.get('no_subdir'))
 
             item['scope'] = did_scope
@@ -1321,7 +1321,7 @@ class DownloadClient:
 
         return items
 
-    def _resolve_one_item_dids(self, item: dict[str, Any]) -> "Iterator[dict[str, Any]]":
+    def _resolve_one_item_dids(self, item: dict[str, Any], scopes: list[str]) -> "Iterator[dict[str, Any]]":
         """
         Resolve scopes or wildcard DIDs to lists of full DID names:
 
@@ -1329,6 +1329,8 @@ class DownloadClient:
         ----------
         item :
             One input item
+        scopes :
+            list of known scope names
         """
         dids = item.get('did')
         filters = item.get('filters', {})
@@ -1346,7 +1348,7 @@ class DownloadClient:
             dids = [dids]
 
         for did_str in dids:
-            scope, did_name = self._split_did_str(did_str)
+            scope, did_name = extract_scope(did_str, scopes)
             filters['name'] = did_name
             any_did_resolved = False
             for did in self.client.list_dids(scope, filters=filters, did_type='all', long=True):
@@ -1428,8 +1430,9 @@ class DownloadClient:
         did_to_input_items = {}
 
         # Resolve DIDs
+        scopes = self.client.list_scopes()
         for item in input_items:
-            resolved_dids = list(self._resolve_one_item_dids(item))
+            resolved_dids = list(self._resolve_one_item_dids(item, scopes=scopes))
             if not resolved_dids:
                 logger(logging.WARNING, 'An item did not have any DIDs after resolving the input: %s.' % item.get('did', item))
             item['dids'] = resolved_dids
@@ -1650,8 +1653,9 @@ class DownloadClient:
                     no_subdir = item.get('no_subdir', False)
                     file_did_path = file_did.name
                     if input_did != file_did:
-                        # if datasets were given: prepare the destination paths for each dataset
-                        if self.extract_scope_convention == 'belleii' and file_did_path.startswith('/'):
+                        # if datasets were given: prepare a subdirectory for each dataset
+                        if file_did_path.startswith('/'):
+                            # hierarchical case: use last element of dataset path as name (/foo/bar/dataset -> dataset)
                             file_did_path = file_did_path.split('/')[-1]
                         path = os.path.join(self._prepare_dest_dir(base_dir, input_did.name, no_subdir), file_did_path)
                     else:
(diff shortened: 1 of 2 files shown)


lib/rucio/client/downloadclient.py
```diff
@@ -1855,7 +1855,12 @@ def _split_did_str(self, did_str: str) -> tuple[str, str]:
...
-            if self.extract_scope_convention == 'belleii':
+            # no : in name, extract scope from did
+
+            # this branch used to be belle-ii specific (if self.extract_scope_convention == "belleii")
+            # but is also needed for other groups, e.g. CTAO, using hierarchical dids,
+            # so now we check for the presence of / to decide if we need to call extract scope
+            if "/" in did[0]:
                 scopes = [scope for scope in self.client.list_scopes()]
                 did_scope, did_name = extract_scope(did[0], scopes)
             else:
```
maxnoe: This branch looks like an in-place manual "extract_scope" implementation.

Could this be replaced by just always calling `extract_scope`?

lib/rucio/client/downloadclient.py
```diff
@@ -1855,7 +1855,12 @@ def _split_did_str(self, did_str: str) -> tuple[str, str]:
             did_scope = did[0]
             did_name = did[1]
         elif len(did) == 1:
-            if self.extract_scope_convention == 'belleii':
+            # no : in name, extract scope from did
+
+            # this branch used to be belle-ii specific (if self.extract_scope_convention == "belleii")
+            # but is also needed for other groups, e.g. CTAO, using hierarchical dids,
+            # so now we check for the presence of / to decide if we need to call extract scope
+            if "/" in did[0]:
```
Geogouz: Are there currently VOs that deliberately disable scope auto-extraction (or use a different convention) that also provide slash-containing names? In the init, we have `self.extract_scope_convention = config_get('common', 'extract_scope', False, None)`.

maxnoe: >  self.extract_scope_convention = config_get('common', 'extract_scope', False, None)

This is also wrong I would say, as extract_scope can also be set via only setting a policy package which defines a default implementation of extract_scope.

This case is not covered by the line you quoted.

maxnoe: But it's only used to check for `== "belleii"`

lib/rucio/client/downloadclient.py
```diff
@@ -1855,7 +1855,12 @@ def _split_did_str(self, did_str: str) -> tuple[str, str]:
             did_scope = did[0]
             did_name = did[1]
         elif len(did) == 1:
-            if self.extract_scope_convention == 'belleii':
+            # no : in name, extract scope from did
```
voetberg: Can you add the issue number in the comment so people can find context for this change?

maxnoe:
After discussion today with @cserf we found that the download client `_split_did_str` was basically a duplicate of the default extract scope algorithm, checking the history we found that it predates the introduction of a default extract scope algoirithm, so we assume this was a missed refactoring when that was introduced.

maxnoe:
Converting to draft until I have a test for the download client when using the belle-ii policy

maxnoe:
@bari12 @cserf 

I tried using the `did_factory` fixture, but it seems to create invalid dids in teh belle-ii case:

```
(Pdb) did_factory.random_file_did()
{'scope': mock, 'name': '/belle/OTDMJU-belleii-download_datasets-/cont_944a19b4f11e477082cbf1fd39cf886b/dataset_d028d33c0edf4a96a4c2009e6d801211/file_8f469ec2559c40a49e71b9b9938ede2d'}
```

The `'name'` should start with `/belle/mock`

maxnoe:
@voetberg This is connected to https://github.com/rucio/rucio/issues/7979

The code that is replaced here with calling `extract_scope` was the reason we had to revert your changes for #7979.

I think once this is merged, we can retry cherry-picking your changes and see if it does not break the download test I added here, which should solve the https://github.com/rucio/rucio/issues/8223 triggered by those changes.

maxnoe:
@voetberg there is a change in the method I removed here and replaced with `extract_scope`, I resolved this by still removing the method but was wondering whether that change about scopes needs to be preserved.

Or maybe this is a question for @cserf since the logic was only in the `belle-ii` branch of the code:

```diff
             if self.extract_scope_convention == 'belleii':
-                scopes = [scope for scope in self.client.list_scopes()]
-                did_scope, did_name = extract_scope(did[0], scopes)
+                listed_scopes = [scope for scope in self.client.list_scopes()]
+                # TODO remove backwards compatiblity check. See #8125
+                if len(listed_scopes) == 0:
+                    raise ScopeNotFound
+                if not isinstance(listed_scopes[0], str):
+                    scopes = [str(s['scope']) for s in listed_scopes]  # type: ignore
+                else:
+                    scopes = listed_scopes
+                did_scope, did_name = extract_scope(did[0], scopes)  # type: ignore
```

I would assume that the change is ok, since now we delegate to the `extract_scope`, so `belle-ii` could decide what to do in their `extract_scope` implementation.

maxnoe:
Updated to pass the known scopes to `extract_scopes` so that policy can do the same check that was now removed here.

voetberg:
> @voetberg there is a change in the method I removed here and replaced with `extract_scope`, I resolved this by still removing the method but was wondering whether that change about scopes needs to be preserved.
> 
> Or maybe this is a question for @cserf since the logic was only in the `belle-ii` branch of the code:
> 
> ```diff
>              if self.extract_scope_convention == 'belleii':
> -                scopes = [scope for scope in self.client.list_scopes()]
> -                did_scope, did_name = extract_scope(did[0], scopes)
> +                listed_scopes = [scope for scope in self.client.list_scopes()]
> +                # TODO remove backwards compatiblity check. See #8125
> +                if len(listed_scopes) == 0:
> +                    raise ScopeNotFound
> +                if not isinstance(listed_scopes[0], str):
> +                    scopes = [str(s['scope']) for s in listed_scopes]  # type: ignore
> +                else:
> +                    scopes = listed_scopes
> +                did_scope, did_name = extract_scope(did[0], scopes)  # type: ignore
> ```
> 
> I would assume that the change is ok, since now we delegate to the `extract_scope`, so `belle-ii` could decide what to do in their `extract_scope` implementation.

I would expect this is okay, as long as `belle-ii` keeps this in mind

cserf:
Thank you. The PR looks good to me

cserf (dismissed):

lib/rucio/client/downloadclient.py
```diff
@@ -229,7 +229,16 @@ def __init__(
...
+
+    def _list_scopes_str(self) -> list[str]:
+        '''Get list of known scope names.'''
+        # server versions >= 40 return dicts of scope, owner
+        scopes = [
+            scope['scope'] if isinstance(scope, dict) else scope
+            for scope in self.client.list_scopes()
+        ]
+
+        return scopes
```
dchristidis: Following #8487, is this still required?

maxnoe: No, need to rebase and just call list_scopes

maxnoe: done

dchristidis (approved):

PR #8226, opened 2025-11-19
Feature 8224 delete subscriptions

### LINKED ISSUE #8224: Delete subscription feature
### Description

Ability to delete (not just deactivate) subscriptions that an account has created.

### Motivation

Some subscriptions become obsolete and they simply generate noisy if "deactivated"

### Change

_No response_

juanpablosalas:
Include feature to delete subscriptions via the python client as explained in the issue #8224 .

### DIFF
--- a/lib/rucio/client/subscriptionclient.py
+++ b/lib/rucio/client/subscriptionclient.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from json import dumps
+from json import dumps, loads
 from typing import TYPE_CHECKING, Any, Literal, Optional, Union
 
 from requests.status_codes import codes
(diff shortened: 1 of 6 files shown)


dchristidis (commented):
Be advised that there is a foreign key constraint in the `rules` table.

Geogouz:
Tests should be passing

dchristidis (commented):
This won’t do. Other tables have foreign key constraints (e.g. `locks`) and you would be creating major inconsistencies (e.g. `replicas`). Refer to `delete_rule()`.

Even if you replaced the newly-introduced query with the core function, I don’t think deleting the rules from the server is the right approach.

To be honest, I’m not convinced deleting a subscription is the right thing to do. Probably using the `SoftModelBase` is more realistic, though the constraint on the subscription name and account combination will pose some difficulties.

bziemons:

PR #8229, opened 2025-11-20
Add filter and caching to Open Data replica list; Fixes rucio#8215

### LINKED ISSUE #8215: Add caching for Open Data file lists
The Open Data endpoint can optionally return the list of replica locations which is useful for integration with other Open Data applications.

This query is expensive (the reason why it's optional) but it's results shouldn't change. It has the potential to slow down the db if the number of replicas to fetch is massive.

A simple solution to this issue would be to provide caching of responses but this could still lead to db slow down.

Another more involved solution would be to store this list of replicas inside some Rucio db table acting as cache and periodically update the list via some async mechanism like a daemon, then limit the public open data endpoint to just fetch from the cache, never to trigger heavy computations to the db.

Additionally, only replicas in open data rses should be listed (based on some filters such as the one used for open data rules).

This needs some discussion.

### LINKED ISSUE #8228: Add RSE filter to Open Data replica list
The returned list should be filtered to only Open Data RSEs.

lobis:
A new config parameter `rse_expression` has been added to the `opendata` section.

The `rule_rse_expression` used to define the rse expression for the automatic replication rules, will now fallback to the new `rse_expression` parameter when  `rule_rse_expression` is not defined.

The method to fetch file list for open data entries is now filtered on the open data rses according to this config variable and throws if its not defined. This method is now cached and provides additional info such as cache hit status and request elapsed time, which will be useful when assessing the impact of this method.

The tests have been adapted, because of the dependency on `rse_expression` to return files, I added this config parameter to the base config file for tests. Other specific configs such as open data rules are still added via calls.

Related docs PR: 
- https://github.com/rucio/documentation/pull/666

### DIFF
--- a/etc/docker/test/extra/rucio_autotests_common.cfg
+++ b/etc/docker/test/extra/rucio_autotests_common.cfg
@@ -154,5 +154,8 @@ gcs = /opt/rucio/etc/google-cloud-storage-test.json
 idpsecrets = /opt/rucio/etc/idpsecrets.json
 admin_issuer = wlcg
 
+[opendata]
+rse_expression = OpenData=True
+
 [api]
 endpoints = accountlimits, accounts, archives, auth, config, credentials, dids, dirac, export, heartbeats, identities, import, lifetime_exceptions, locks, meta_conventions, ping, redirect, replicas, requests, rses, rules, scopes, subscriptions, traces, vos, opendata, opendata_public
--- a/lib/rucio/core/opendata.py
+++ b/lib/rucio/core/opendata.py
@@ -13,14 +13,17 @@
 # limitations under the License.
 
 import json
+import time
 from re import match, search
 from typing import TYPE_CHECKING, Any, Optional, Union, cast
 
+from dogpile.cache.api import NoValue
 from sqlalchemy import and_, delete, insert, update
 from sqlalchemy.exc import DataError, IntegrityError
 from sqlalchemy.sql.expression import bindparam, select
 
 from rucio.common import exception
+from rucio.common.cache import MemcacheRegion
 from rucio.common.config import config_get, config_get_bool, config_get_int
 from rucio.common.constants import DEFAULT_VO
 from rucio.common.exception import OpenDataError, OpenDataInvalidStateUpdate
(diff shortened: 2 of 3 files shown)


lobis:
After doing an offline review with @bari12 , I will do some significant refactoring of the changes. I will mark it as ready again when done.

Geogouz:
> After doing an offline review with @bari12 , I will do some significant refactoring of the changes. I will mark it as ready again when done.

Is this ready for review now?

lobis:
> > After doing an offline review with @bari12 , I will do some significant refactoring of the changes. I will mark it as ready again when done.
> 
> Is this ready for review now?

Yes!

Geogouz:
The new hard dependency on `opendata.rse_expression` breaks the backward compatibility but I am not sure how critical that is. You don't think of any problem there, right?

lobis:
> The new hard dependency on `opendata.rse_expression` breaks the backward compatibility but I am not sure how critical that is. You don't think of any problem there, right?

It technically breaks yes but it's okay, only me is using this api right now.

Geogouz:
The patch is named after https://github.com/rucio/rucio/issues/8228 but you (and me too now in the title) are linking manually https://github.com/rucio/rucio/issues/8215 . Is this correct? Will/should it close both?

Geogouz:
The caching stores results for two hours but what about high-frequency dataset mutations or replica changes? Should any cache-invalidation be used here to avoid incorrect reported data during this window?

lobis:
> The caching stores results for two hours but what about high-frequency dataset mutations or replica changes? Should any cache-invalidation be used here to avoid incorrect reported data during this window?

This would almost exclusively be used on closed DIDs (requirement for being opendata-public) so I don't think there would be an issue.

lobis:
> The patch is named after #8228 but you (and me too now in the title) are linking manually #8215 . Is this correct? Will/should it close both?

Yes, both would be closed with this PR.

Geogouz (dismissed):
This looks ready from my side.

dchristidis:
Please rebase.

dchristidis (approved):

dchristidis:
> It technically breaks yes but it's okay, only me is using this api right now.

Based on this statement, I will make the changes available in the next minor release.

PR #8230, opened 2025-11-20
Rules: Store when a rule is deleted #8227

### LINKED ISSUE #8227: Store when a replication rule is deleted
### Description

1. ~~Introduce a new `deleted_at` column in the `rules_history` table.~~ See #8242.
2. Change `core.rules.delete_rule()` so that the new column is populated by `insert_rule_history()`, at the time the rule is deleted.

### Motivation

This is the prerequisite for the client to be able to report more accurate information.

### Change

_No response_

dchristidis:
I’m deliberately deviating from the current contribution guide to experiment with conventional commits.


rdimaio (commented):
Could you add a test to check that this is added as expected and, when retrieved, corresponds to the expected timestamp?

dchristidis:
As it stands right now, there is no way to retrieve the new column. Hence, it is not possible to write a test. I purposefully kept it as small as possible in case I could persuade @bari12 for a freeze exception. If not (more likely), I will augment the PR with a fix for #8231. This will make it possible to write a test.

One more thing, because I was thinking about it earlier today: I have the impression we can only reliably check that the timestamp is set, not that it has an appropriate value. If you believe otherwise, I would love to know how.

rdimaio:
> One more thing, because I was thinking about it earlier today: I have the impression we can only reliably check that the timestamp is set, not that it has an appropriate value. If you believe otherwise, I would love to know how.

Ideally we would need to retrieve the value returned by `datetime.utcnow()` when set for the rule, if we want to use the 'real' timestamp, but otherwise we could also mock `datetime.utcnow()`'s return with a value that we know, and then at the end check that that's the value that was set.

For consistency (though it's a bit useless), maybe we could also add a check to ensure that the `deleted_at` set comes after `created_at`?

Geogouz:
When a rule is hard-deleted, `delete_rule()` sets [`rule.deleted_at = datetime.utcnow()`](https://github.com/dchristidis/rucio/blob/b7a4307ac5c2253d705d46b5e2589f962424bae7/lib/rucio/core/rule.py#L1199) before writing history records. Meanwhile, the history models ([`ReplicationRuleHistoryRecent`](https://github.com/dchristidis/rucio/blob/b7a4307ac5c2253d705d46b5e2589f962424bae7/lib/rucio/db/sqla/models.py#L1282) and [`ReplicationRuleHistory`](https://github.com/dchristidis/rucio/blob/b7a4307ac5c2253d705d46b5e2589f962424bae7/lib/rucio/db/sqla/models.py#L1331)) now include a `deleted_at` column, implying deletion times should be recorded in those tables. But from what I see, `insert_rule_history` does not include the `deleted_at` when copying fields into both history tables. So I guess those values would remain `null`. Is this intended?

dchristidis:
You’re very right, this was an omission on my part.

The fix for this revealed a few other issues (e.g. also the `priority` is not stored), but I don’t want the PR to be trying to do too many things at once. I will split it in separate issues.

Replaced with #8243.

PR #8234, opened 2025-11-21
Documentation - Autodoc-able RseAttrs

### LINKED ISSUE #551: Document common.constant.RSEAttr

voetberg:
Closes [#documentation/551](https://github.com/rucio/documentation/issues/551)

### DIFF
--- a/lib/rucio/core/rse.py
+++ b/lib/rucio/core/rse.py
@@ -1087,33 +1087,34 @@ def get_rses_with_attribute_value(
     return result
 
 
+# The ignores on the following overloads are for variables not being accepted as types
 @overload
-def get_rse_attribute(rse_id: str, key: Literal['sign_url'], use_cache: bool = True, *, session: "Session") -> Optional[SUPPORTED_SIGN_URL_SERVICES_LITERAL]:
+def get_rse_attribute(rse_id: str, key: Union[Literal['sign_url'], type[RseAttr.SIGN_URL]], use_cache: bool = True, *, session: "Session") -> Optional[SUPPORTED_SIGN_URL_SERVICES_LITERAL]:  # type: ignore
     ...
 
 
 @overload
-def get_rse_attribute(rse_id: str, key: Literal['sign_url'], use_cache: bool = True) -> Optional[SUPPORTED_SIGN_URL_SERVICES_LITERAL]:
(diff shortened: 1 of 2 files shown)


lib/rucio/core/rse.py
```diff
@@ -1088,32 +1088,32 @@ def get_rses_with_attribute_value(
...
 @overload
-def get_rse_attribute(rse_id: str, key: Literal['sign_url'], use_cache: bool = True, *, session: "Session") -> Optional[SUPPORTED_SIGN_URL_SERVICES_LITERAL]:
+def get_rse_attribute(rse_id: str, key: 'RSEAttrObj', use_cache: bool = True, *, session: "Session") -> Optional[SUPPORTED_SIGN_URL_SERVICES_LITERAL]:
     ...
 
 
 @overload
-def get_rse_attribute(rse_id: str, key: Literal['sign_url'], use_cache: bool = True) -> Optional[SUPPORTED_SIGN_URL_SERVICES_LITERAL]:
+def get_rse_attribute(rse_id: str, key: 'RSEAttrObj', use_cache: bool = True) -> Optional[SUPPORTED_SIGN_URL_SERVICES_LITERAL]:
     ...
```
Geogouz, line 1098: The overloads for `get_rse_attribute` now accept any `RSEAttrObj` but still promise to return only the limited `SUPPORTED_SIGN_URL_SERVICES_LITERAL` values (`'gcs', 's3', 'swift'`). However, that return type may only be correct when the key is specifically `sign_url`.

Geogouz: For other attributes, the real return type is `str` or `bool`, so the overloads mislead static typing.

Geogouz:
Hmm, integration tests are failing:

    FAILED tests/test_conveyor.py::test_receiver_archiving[caches_mock0] - AssertionError: assert 'FAILED' == 'ARCHIVING'
    FAILED tests/test_conveyor.py::test_transfer_with_tokens[file_config_mock{'oidc': {'admin_issuer': 'indigoiam'}}] - assert False

lib/rucio/core/rse.py
voetberg: Yeah this was a compromise. I could either change this and ignore the type failures with `sign_url` or the other way around. I wasn't sure which one was the correct one, because the overload leads to a different type failure

voetberg:
@dchristidis I recall you had opinions on the implementation here, so I added you as a reviewer. I won't be hurt if you remove yourself though

Issue #8078, opened 2025-10-01
Database: Introduce a test-suite to detect whether Alembic migration history is inconsistent


Geogouz:
**Note: This issue is planned to be tackled during this summer, so my estimated target for it is Rucio 42 release.**

So this issue is about replacing the current Alembic migration checks with a new migration validation framework.

## How migrations are used/tested today

Today, our Alembic checks mostly verify that revision markers line up and that a few migration paths run successfully.

In practice, we create a fresh DB from the current SQLAlchemy models, stamp it to the current Alembic head, and exercise a downgrade/upgrade path such as `head -> base -> head-1 -> head`.

That is useful, but it is still mostly a revision-path check. It does not fully guarantee that a supported migration lands on the expected schema state, or that a downgrade really returns to the same schema state it started from.

What I would like instead is a schema-focused framework: for every supported release-to-release migration path, we should be able to say that:

- upgrade lands on the exact target schema
- downgrade returns to the exact source schema

## Proposed contract

I suggest the following points:

- **Production migration compatibility is guaranteed only for PostgreSQL and Oracle.**  
  Additional DB’s mean additional complexity/reviewing and according to our discussions, we agreed that it doesn’t worth it. SQLite may still be used in tests, but it is not a production migration target. MySQL is out of scope.

- **On each release branch, we only guarantee (and may only retain) Alembic history back to the latest still-supported LTS major release.**  
  Trying to keep the full historical chain healthy forever will likely lead to many migration rewrites and effort (a problem in https://github.com/rucio/rucio/pull/8238).

- **Within that retained window, the supported source states are the Alembic heads of supported released tags.**  
  A release may contain multiple Alembic revisions in one linear chain, but only the revision that is the head at a supported released tag is treated as a supported migration entry point.

- **On a target branch `B`, direct upgrades are supported from every supported released-tag head in that retained window, using branch `B`’s migration chain.**  
  So if the retained window on `B` contains `A1`, `A2`, and `A3`, then `B` must support:
  - `A1 -> B`
  - `A2 -> B`
  - `A3 -> B`  
  This makes branch `B` responsible for upgrading all currently supported source releases directly to its own head.

- **Downgrade is part of the same contract for those supported source states.**  
  If `B` can upgrade from `A`, it must also be able to downgrade back to `A` and restore the same schema state.

- **The guarantee is schema-only.**  
  This framework validates schema state, not data contents. Any guarantees around data migrations should be handled separately with dedicated data/fixture tests.

[...]

PR #8238, opened 2025-11-25
Database: Refactor Alembic migrations to become uniform and honour the configured schema; Fix rucio#8145

### LINKED ISSUE #8145: Database: Refactor Alembic migrations to become uniform and honour the configured schema
**Note: This issue is delayed for now and the corresponding PR (#8238) will be closed and then reopened (maybe once #8078 is ready so that we can use that to safely review it).**

During Alembic migrations we have PostgreSQL ENUM types created via `op.execute("CREATE TYPE ...")` end up in the `public` schema, even though tables and other objects correctly respect the configured schema (for example, `dev`). This happens because the raw SQL statements omit an explicit schema qualifier, so PostgreSQL falls back to the first entry in `search_path` (`public`). As a result, enum creation and teardown ignore Alembic’s schema translation settings.

I would suggest the following steps:

**1. Add schema-aware ENUM helper:**
Introduce a utility in the Alembic environment (or a shared migrations helper module) that derives the configured schema and builds fully qualified `CREATE TYPE`, `DROP TYPE`, and `ALTER TYPE` statements.

**2. Refactor existing migrations to use the helper:**
Touch every migration that currently issues raw CREATE TYPE/DROP TYPE SQL and replace it with calls to the new helper.

**3. Add a forward migration to relocate legacy ENUMs:**
Create a new Alembic revision that creates schema-qualified copies of each ENUM that previously lived in public, altering every dependent column to reference the new types. And eventually drop any old public ENUMs.

_Regarding the downgrade.._
Historically, each downgrade in the Alembic history recreates the enum in exactly the same unqualified way as the upgrade, so rolling back would push the database back into the public schema. Because we are going to patch those older revisions to use the new helper, their downgraded state will also become schema-aware. Therefore, I would suggest moving the same enums back into whatever schema the immediately preceding revision expects (i.e. the configured schema, not public). Practically, the downgrade can reuse the helper to recreate the previous type name in the configured schema and repoint columns, just like the upgrade but with the old definitions.

Geogouz:
**Note: Since the linked issue #8145 is delayed for now, this PR will be closed and then reopened once #8078 is ready (so that we can use that to safely review it).**

This PR is a fix for rucio#8145.

I will provide separately bellow some guidance on what exactly this PR contains, nevertheless regarding how to assess current PR:

- To test that the local tests via `tools/run_tests.sh` are now idempotent, you can execute locally: 

    ``` ./tools/bootstrap_dev.sh -p -x ```

    and then execute repeatedly:

    ```docker exec -it dev-rucio-1 tools/run_tests.sh```

- Moreover, as demonstrated here: https://github.com/Geogouz/rucio/tree/patch-XXX-alembic-squashed, to see also that repeated migration cycles do not affect in general any CI test (involving different database dialects), you can enable multiple migration cycles before the execution of those tests. For that, inside the files `tools/run_multi_vo_tests_docker.sh` and `tools/run_tests.sh`, locate this part:

    ```sh
    echo "Running full alembic migration (cycle: ${migration_cycle}/10)"
    ALEMBIC_CONFIG="$RUCIO_HOME/etc/alembic.ini" tools/alembic_migration.sh
    if [ $? != 0 ]; then
        echo 'Failed to run alembic migration!'
        exit 1
    fi
    ```

    .. and surround that with a for loop:

    ```sh
    for migration_cycle in {1..5}; do
	    …
    done
    ```

    Hence, it should be:

    ```sh
    for migration_cycle in {1..5}; do
        echo "Running full alembic migration (cycle: ${migration_cycle}/5)”
        ALEMBIC_CONFIG="$RUCIO_HOME/etc/alembic.ini" tools/alembic_migration.sh
        if [ $? != 0 ]; then
            echo 'Failed to run alembic migration!'
            exit 1
        fi
    done
    ```

**Tip for the reviewers:**
I would not suggest to review this commit-by-commit (i.e. as if each commit was a different PR) because this way you will spend much more time (as later commits keep inevitably reshaping the migrations to ultimately bring some unification and uniformity). Instead, I would suggest to review first/mainly the `ddl_helpers.py` and `enum_ddl_helpers.py` modules. The methods provided there are not tangled with other methods (i.e. they are mostly self-standing) so there should be not much complexity in understanding their effect. The migrations are then basically refactored to utilise those methods.


Geogouz:
To support the review of this PR, I summarise below the changes introduced by the new DDL helper modules added alongside the refactored migrations. This document could be also used as a reference for writing new migrations, however, I would suggest to update our documentation page too.

# Alembic DDL helper guide

This PR introduces two new helper modules for Alembic migrations and refactors existing migrations to consume them:

* `lib/rucio/db/sqla/migrate_repo/ddl_helpers.py` centralises cross-dialect DDL helpers so migrations can share proper quoting, dialect detection, and safe “drop if exists” semantics.
* `lib/rucio/db/sqla/migrate_repo/enum_ddl_helpers.py` provides PostgreSQL-only enum DDL helpers with input validation, backend guardrails, and offline SQL rendering support.
* Existing migrations have been updated to replace bespoke inline DDL snippets with the shared helpers above.

Therefore, all above provide consistency, idempotence, and offline SQL parity across the repository.

---

## Motivation

* **Reduce duplicated DDL snippets** across migrations by offering reusable helpers that encapsulate backend-specific quoting and “exists / missing” checks.
* **Improve reliability for offline SQL generation** by deriving dialect and identifier quoting even when no live connection is present.
* **Provide a well-defined interface for PostgreSQL enum operations** (create, add value, drop) with validation that catches invalid identifiers or labels early.
* **Make migrations more readable and safer to re-run** by codifying idempotent patterns and centralising backend-specific quirks.

---

## Key changes

### 1. Shared DDL helpers (`ddl_helpers.py`)

**Consistent quoting and dialect detection**

The module centralises how we discover Alembic/SQLAlchemy state:

* `get_migration_context()` – best-effort retrieval of the active `MigrationContext` (tries `op.get_context()` first, then `context.get_context()`), returning `None` when no migration is running.
* `get_current_dialect()` – resolves the dialect name from:
  1. `ctx.dialect.name` when a context is active,
  2. `context.config`’s `dialect` option, or
  3. the scheme of `sqlalchemy.url` (normalised for things like `postgresql+psycopg`).
* `is_current_dialect(*names)` – convenience helper to check whether the active dialect matches any of the given names (case-insensitive).
* `get_server_version_info()` – returns the current dialect’s `server_version_info` tuple when available.
* `get_effective_schema()` – derives automaticlly the “default schema” Alembic uses (the one hosting the version table) from:
  1. `MigrationContext.version_table_schema`,
  2. `MigrationContext.opts['version_table_schema']`,
  3. `context.config.get_main_option('version_table_schema')`.

Identifier and literal quoting are centralised as well:

* `get_identifier_preparer()` – fetches the dialect’s `IdentifierPreparer`. It prefers the live dialect from the migration context, falls back to the configured dialect, and finally to `DefaultDialect`. Unknown dialects emit a one-time warning rather than failing.
* `quote_identifier(name)` – quotes identifiers using the active dialect’s rules. Falsy values return `""`, which keeps templated SQL simple.
* `quote_literal(value)` – wraps values in single quotes for SQL/PL blocks (doubling embedded quotes). `None` is rendered as `NULL`.

On top of those:

* `qualify_table(table_name, schema=None)` – returns a dialect-appropriate table identifier:
  * PostgreSQL/MySQL/SQLite – quoted via the dialect’s `IdentifierPreparer`.
  * Oracle – intentionally unquoted (`SCHEMA.TABLE`) to preserve Oracle’s implicit uppercasing semantics.
  * Unknown dialect – uses `schema.table` without quoting and emits a one-time warning.
* `qualify_index(index_name, schema=None)` – similar for indexes, but with index-specific behaviour:
  * PostgreSQL/etc – schema-qualified and quoted.
  * MySQL/MariaDB – **never** schema-qualified (only the quoted index name), because `DROP INDEX` is `DROP INDEX <name> ON <table>`.
  * Oracle – unquoted, again to preserve uppercasing.

These helpers keep online and offline behaviour aligned and remove dialect-specific quoting from individual migrations.

**Schema-defaulting wrappers for Alembic operations**

To avoid threading `schema=` everywhere, most Alembic operations now have a thin wrapper that injects the “effective schema” when callers do not specify one:

* `add_column`
* `alter_column`
* `create_check_constraint`
* `create_index`
* `create_primary_key`
* `create_table`
* `create_unique_constraint`
* `drop_column`
* `drop_constraint`
* `drop_index`
* `drop_table`
* `rename_table`

Each wrapper:

1. Calls `_with_default_schema(kwargs)` to set `schema=get_effective_schema()` when `schema` is missing or falsy.
2. Delegates to the underlying `alembic.op.*` function.

This keeps migrations consistent with the Alembic version-table configuration, while still allowing explicit schema overrides. Keep in mind that so far, many migrations were actually targeting the default `public` schema because no schema was being provided.

**Safe re-runs / idempotent drop & create**

Several helpers implement “drop if exists” and “create if missing” semantics, normalising backend-specific behaviour and error messages:

* `try_drop_constraint(constraint_name, table_name, *, type_=None)` – idempotent, backend-aware constraint drops, including special handling for MySQL foreign keys/checks and PostgreSQL primary keys (default `<table>_pkey` name).
* `try_drop_index(index_name, table_name=None)` – idempotent index drops using dialect-appropriate SQL (e.g. `DROP INDEX IF EXISTS` on Postgres, `DROP INDEX ... ON ...` on MySQL/MariaDB) and tolerant error handling.
* `drop_current_primary_key(table_name)` / `try_drop_primary_key(table_name, legacy_names=())` – remove the current PK in a safe, dialect-specific way (catalog lookups on PostgreSQL/Oracle; tolerant `ALTER TABLE ... DROP PRIMARY KEY` on MySQL/MariaDB), plus optional clean-up of legacy constraint names.
* `try_create_table(*args, **kwargs)` – idempotent table creation using `Inspector.get_table_names` (and, where supported, `get_temp_table_names`) to detect pre-existing regular or temporary tables. Falls back to unconditional creation in true offline scenarios.

Together, these helpers make migrations safer to re-run and significantly reduce dialect-specific boilerplate inside individual migration scripts.

---

### 2. PostgreSQL enum helpers (`enum_ddl_helpers.py`)

This module focuses on PostgreSQL enum DDL and exposes a small, intentional API via `__all__`. It is designed to be **PostgreSQL-only** while still supporting offline SQL rendering.

**Input validation & backend guardrails**

Internal helpers:

* `_validate_identifier(name, *, allow_qualified=False)`  
  Enforces PostgreSQL identifier rules: non-empty strings, no NUL bytes, length ≤ 63 bytes when encoded as UTF-8, and (optionally) well-formed `schema.name` qualified identifiers.
* `_validate_enum_labels(values)`  
  Normalises an iterable of labels into a tuple and ensures the iterable is non-empty, every label is a non-empty string (no `None`), length ≤ 63 bytes, no NUL bytes, and no duplicates.
* `_assert_postgresql()`  
  Uses `get_current_dialect()` and raises `NotImplementedError` when the dialect is **known** and not `"postgresql"`, while still permitting offline SQL rendering when the dialect cannot be determined.

**Operation-focused enum SQL helpers**

The public enum helpers are:

* `render_enum_name(name, schema=None)` – validates the unqualified enum name, chooses an effective schema (`schema` arg or `get_effective_schema()`), and returns a quoted identifier like `"schema"."name"` or `"name"`.
* `enum_values_clause(values)` – validates labels and returns a fragment like `'QUEUED', 'SUBMITTED', 'DONE'` using `quote_literal`, suitable for `CREATE TYPE ... AS ENUM`.
* `create_enum_sql(name, values, *, schema=None, if_not_exists=False)` – builds a single `CREATE TYPE ... AS ENUM` statement for PostgreSQL; `if_not_exists=True` is deliberately rejected, in favour of:
* `create_enum_if_absent_sql(name, values, *, schema=None)` – returns a `DO $$` block that runs `CREATE TYPE ... AS ENUM` and ignores `duplicate_object`, effectively emulating `CREATE TYPE IF NOT EXISTS`.
* `drop_enum_sql(name, *, schema=None, if_exists=True, cascade=False)` – builds `DROP TYPE` SQL, optionally with `IF EXISTS` and `CASCADE`.
* `alter_enum_add_value_sql(name, value, *, schema=None, before=None, after=None, if_not_exists=False)` – builds `ALTER TYPE ... ADD VALUE` with optional positioning via `BEFORE`/`AFTER` and supports idempotent behaviour via `IF NOT EXISTS` on PostgreSQL ≥ 9.3 or a `DO $$`-wrapped statement that ignores `duplicate_object` on older/unknown versions.

There are also thin execution wrappers:

* `try_create_enum_if_absent(...)` – executes `create_enum_if_absent_sql` via `op.execute`.
* `try_alter_enum_add_value(...)` – executes `alter_enum_add_value_sql` via `op.execute`.
* `try_drop_enum(...)` – executes `drop_enum_sql` via `op.execute`.

**Backend-agnostic enum type helper**

The main integration point for migrations is:

* `get_backend_enum(py_enum, *, name) -> sa.types.TypeEngine`

This:

1. Iterates over `py_enum` and collects `member.value`, validating the resulting label set via `_validate_enum_labels`.
2. **On PostgreSQL**:
   * Determines the target schema via `get_effective_schema()`.
   * Constructs a `postgresql.ENUM` with `create_type=False` so SQLAlchemy does **not** manage the `CREATE TYPE`.
   * Ensures the underlying enum type exists by calling `try_create_enum_if_absent(pg_enum)` before returning the type.
3. **On other backends**:
   * Returns a `sa.Enum(py_enum, name=name, native_enum=False, create_constraint=True, values_callable=lambda obj: [e.value for e in obj])`, i.e. a CHECK-constrained enum backed by a regular string column.

---

### 3. Migration refactors using the helpers

In addition to adding the helper modules themselves, existing migrations have been refactored to **consume** them. At a high level:

* **Direct Alembic calls → helper wrappers**

  Many migrations that previously called `op.add_column`, `op.create_table`, `op.drop_index`, etc. now call the schema-defaulting wrappers (`add_column`, `create_table`, `drop_index`, …) re-exported from `rucio.db.sqla.migrate_repo`.

* **Ad-hoc “try/except” drops → `try_*` helpers**

  Custom logic such as:

  ```python
  try:
      op.drop_constraint("SOME_CONSTRAINT", "some_table")
  except Exception:
      pass
  ```

  has been replaced by explicit calls to:

  ```python
  try_drop_constraint("SOME_CONSTRAINT", "some_table")
  ```

  and likewise for `try_drop_index` and `try_drop_primary_key`.

* **Inline enum DDL → enum helpers**

  Migrations that previously emitted raw `CREATE TYPE ... AS ENUM`, `ALTER TYPE ... ADD VALUE`, or `DROP TYPE` statements have been rewritten to use:

  * `create_enum_if_absent_sql` / `try_create_enum_if_absent`
  * `alter_enum_add_value_sql` / `try_alter_enum_add_value`
  * `drop_enum_sql` / `try_drop_enum`
  * `get_backend_enum` for column type definitions

* **Identifier quoting & schema qualification**

  Raw string interpolation for object names (tables, indexes, types) in migrations has been reduced in favour of `qualify_table`, `qualify_index`, `quote_identifier`, and `quote_literal`. Where the helpers are used, migrations no longer need to know about backend-specific quoting rules.

---

## Public API surface & imports

`__init__.py` re-exports the helpers via `__all__` so migrations import them from a single, stable place:

```python
from rucio.db.sqla.migrate_repo import (
    add_column,
    alter_column,
    alter_enum_add_value_sql,
    create_check_constraint,
    create_enum_if_absent_sql,
    create_enum_sql,
    create_index,
    create_primary_key,
    create_table,
    create_unique_constraint,
    drop_column,
    drop_constraint,
    drop_current_primary_key,
    drop_enum_sql,
    drop_index,
    drop_table,
    enum_values_clause,
    get_backend_enum,
    get_current_dialect,
    get_effective_schema,
    get_identifier_preparer,
    get_migration_context,
    get_server_version_info,
    is_current_dialect,
    qualify_index,
    qualify_table,
    quote_identifier,
    quote_literal,
    rename_table,
    render_enum_name,
    try_create_table,
    try_alter_enum_add_value,
    try_create_enum_if_absent,
    try_drop_constraint,
    try_drop_enum,
    try_drop_index,
    try_drop_primary_key,
)
```

---

## How to use in migrations (practical guidance)

1. **Import from the package root**

   Use:

   ```python
   from rucio.db.sqla.migrate_repo import (
       add_column,
       try_drop_constraint,
       get_backend_enum,
       try_alter_enum_add_value,
       # ...
   )
   ```

   instead of importing directly from `ddl_helpers` / `enum_ddl_helpers`.

2. **Prefer safe helpers**

   When dropping or altering objects that *might* already exist or not exist (replays, partially applied migrations, local testing), use the `try_*` helpers:

   * `try_drop_constraint`
   * `try_drop_index`
   * `try_drop_primary_key`
   * `try_create_table`
   * `try_create_enum_if_absent`
   * `try_alter_enum_add_value`

3. **Use schema-defaulting wisely**

   For objects in the “normal” migration schema (where the Alembic version table lives), rely on the wrappers to inject `schema=get_effective_schema()`. Only pass `schema` explicitly when targeting a different schema on purpose.

4. **Keep offline runs in mind**

   When emitting custom SQL via `op.execute`, prefer to build identifiers using `qualify_table`, `qualify_index`, `quote_identifier`, and `quote_literal` so offline SQL generation behaves identically to online runs.

---

## Compatibility notes

* For dialects outside the explicitly supported set (currently PostgreSQL, MySQL/MariaDB, Oracle, SQLite), many helpers log and no-op rather than raising. If we want hard failures for unknown backends, this should be discussed during review.
* Enum helpers rely on PostgreSQL’s `ALTER TYPE ... ADD VALUE IF NOT EXISTS` support on version ≥ 9.3 and fall back to a `DO $$` block elsewhere. On very old PostgreSQL versions, adding the same label twice without `if_not_exists=True` will still raise; migration ordering should be arranged accordingly.

---

## Documentation / next steps

As a follow-up, we should add documentation describing:

* best practices for introducing new Alembic migrations,
* when and how to use the new `ddl_helpers` and `enum_ddl_helpers`,
* patterns for safe cross-dialect DDL and PostgreSQL enum evolution.

dchristidis:
Depending on the order of merging, the migration script being introduced in #7863 will also need to be adapted.

lib/rucio/db/sqla/migrate_repo/enum_ddl_helpers.py
rdimaio, line 1: Could be worth specifying in the filename itself that this is only for postgres, even though it's clear from the top comment in the file

lib/rucio/db/sqla/migrate_repo/ddl_helpers.py
```diff
@@ -0,0 +1,1552 @@
...
+
+    Examples
+    --------
+    >>> rename_table("collections", "collections_legacy")
+    """
+    kwargs = _with_default_schema(dict(kwargs))
+    return op.rename_table(*args, **kwargs)
+
+
+__all__ = [
```
rdimaio, line 1524: Worth trying to import from this file with absolute import to avoid having `__all__` both here and in the `__init__` file

bziemons (changes requested):
twimc the documentation needs a little more work to be less superfluous, talked to @Geogouz in PM.

Geogouz:
I wonder, does a thumbs up emoji suffice to keep the PR Open :D? Doing this bit by bit. Not much time this period.

rcarpa:
Changing old migrations code is a very bad practice. Once a migration is applied on a database ... it is definitely applied. If a fix is required on the database, you don't change the code of the migration + rollback and re-apply, you create a new migration, which fixes the issue in both directions. 

Frankly, I would rather just throw away very old migrations which nobody will ever use (because nobody will update from such an old version of rucio) and that's all. Why even bother touching them ?

Geogouz:
> Changing old migrations code is a very bad practice. Once a migration is applied on a database ... it is definitely applied. If a fix is required on the database, you don't change the code of the migration + rollback and re-apply, you create a new migration, which fixes the issue in both directions.
> 
> Frankly, I would rather just throw away very old migrations which nobody will ever use (because nobody will update from such an old version of rucio) and that's all. Why even bother touching them ?

Yes, I agree with the principle. However, the intent/result here is not to change outcomes/schema (which makes changing old migrations dangerous), but to make legacy migrations safer and more portable when they do run. E.g. by making them re-applicable so that repeated runs becomes possible.

rcarpa:
We don't need to run them. Throw them away completely... just remove them all together from the repository. Re-root to whatever revision was at the oldest still supported LTS version of rucio ... and throw away everything else. No need to "fix" them. No need to run them in the CI. They are plain not needed. Cleanup time.

Geogouz:
> We don't need to run them. Throw them away completely... just remove them all together from the repository. Re-root to whatever revision was at the oldest still supported LTS version of rucio ... and throw away everything else. No need to "fix" them. No need to run them in the CI. They are plain not needed. Cleanup time.

I wouldn't disagree with getting rid of those old ones. But introducing helpers like `ddl_helpers.py` and `enum_ddl_helpers.py` is still relevant to guard the remaining or future ones. And fixing the remaining ones is still valuable because we would still have broken migrations. E.g. ones that drop constraints or indices that do not exist. So wrapping those let's say in a `try` pattern shouldn't harm.

rcarpa:
It's 2000 lines of code to "help" with writing a couple lines of SQL every 4 months. I see it more like an annoyance than something useful: having to read through all the custom helpers we have to find that single one which I need to substitute 2 lines of SQL. Without even looking in detail at the migrations, I expect that 90% of those helpers will be called 0 times once the old migrations are thrown away. 
However, at this stage, I think this is a completely separate discussion which has to be left for later. If we get rid of the old migrations, we can come back to it to see if they are useful for the couple dozens migrations which will still exist in the repository. And then reviewing this PR will also become MUCH easier.

Geogouz:
I would not be against removing those old migrations. I would also not be against getting rid inside the helpers of stuff like MySQL logic (which is still under discussion). But to remove protections just because some cleanup lead into not facing those any more, is not effective. These problems will come again and without those protections, we will eventually have to clean up stuff again. I am personally fine having a method that is used only by one migration, if this ensures proper history traversal.

There is an ongoing work with a new CI test suite being introduced, dedicated to Alembic. This one assesses (via schema-fingerprinting) whether the upgrade <-> downgrade flow leads to exactly the same outcome and our defined models. Once that in place, we could indeed heavily reduce those helpers because any alembic problems would be caught by the CI.

rcarpa:
I am not contradicting you. What I said is: lets first remove legacy un-needed migration. After that, lets remove the helper methods which are not needed anymore because nobody calls them anymore. And, after that, we'll review what remains.

Geogouz:
**Note: Since the linked issue #8145 is delayed for now, this PR will be closed and then reopened once #8078 is ready (so that we can use that to safely review it).**

PR #8240, opened 2025-11-25
Release management: .gitignore, use *.egg-info so that the webUI egg-info is also matched

rdimaio:
The webUI egg-info is created when building the web UI package via the `build_sdist_wheel.sh` script, but it is not matched by the `.gitignore` right now, though it should.


Geogouz (approved):
Makes sense

PR #8246, opened 2025-11-27
Feature 2410 generic checksum column; rucio#2410

### LINKED ISSUE #2410: Extend Rucio to generic checksum column
Motivation
----------
Right now Rucio only supports adler32 and md5 checksums via two specific columns for these checksums. This creates problems if a community has different checksums (e.g. CMS historically used crc checksums) and wants to store them. On the dev meeting on [2019-03-28](https://indico.cern.ch/event/796966/) it was discussed to extend Rucio with a generic checksum column of string type. Multiple checksums can be added (comma-separated) to this column, always specifying the checksum type. E.g. `md5:1234,adler32:5678,crc:123`

We would have to specify a strategy in migrating to this new checksum column.

martynia:
This pull request add a new `checksum` column to all relevant tables which store `Adler32` and `md5` checksums. The changes in the code are made, so the new `checksum` dict value is calculated on the client side from `Adler32` and `md5` values  and propagated to relevant tables along with values SELECT-ed from legacy columns. The current state of the code is that it does not use new `checksum` dict in any checks just yet, since it is not obvious to me how a migration should be done.


Geogouz:
Can you rebase also?

lib/rucio/db/sqla/migrate_repo/versions/8ab4d628cffb_add_generic_checksum_column.py
```diff
@@ -0,0 +1,57 @@
...
+    drop_column('dids', 'availability', schema=schema)
+    drop_column('deleted_dids', 'availability', schema=schema)
+    drop_column('quarantined_replicas', 'availability', schema=schema)
+    drop_column('quarantined_replicas_history', 'availability', schema=schema)
+    drop_column('contents', 'availability', schema=schema)
+    drop_column('archive_contents', 'availability', schema=schema)
+    drop_column('archive_contents_history', 'availability', schema=schema)
+    drop_column('contents_history', 'availability', schema=schema)
+    drop_column('requests', 'availability', schema=schema)
+    drop_column('requests_history', 'availability', schema=schema)
```
Geogouz: You are dropping the `availability` here which doesn't sound correct.

Geogouz:
Couple points:

**Regarding the API/schema alignment with the feature goal:**

The new JSON-schema validation for checksums still only permits `md5` and `adler32` keys, which contradicts the stated goal of supporting arbitrary checksum types (e.g., `crc32`). So any attempt to provide alternative algorithms through the generic column will fail validation, and the main problem the PR aims to solve remains unresolved.

I think we should broaden the schema to allow arbitrary checksum keys (or a well-defined extensible set), with appropriate patterns/validation, so clients can store non-`adler32`/`md5` hashes.

**Regarding the backward compatibility in client logic:**

Right now, the upload validation seems to still hard-require matching Adler-32 between local and remote metadata, ignoring the new checksum dictionary. So deployments that migrate to the generic checksum column without storing Adler-32 will see uploads rejected.

I think we should extend the client-side checksum comparison to use the new generic checksum dict (and fall back to legacy columns), so uploads remain compatible when Adler-32 is absent but another configured checksum is available.

martynia:
> Couple points:
> 
> **Regarding the API/schema alignment with the feature goal:**
> 
> The new JSON-schema validation for checksums still only permits `md5` and `adler32` keys, which contradicts the stated goal of supporting arbitrary checksum types (e.g., `crc32`). So any attempt to provide alternative algorithms through the generic column will fail validation, and the main problem the PR aims to solve remains unresolved.
> 
> I think we should broaden the schema to allow arbitrary checksum keys (or a well-defined extensible set), with appropriate patterns/validation, so clients can store non-`adler32`/`md5` hashes.

Good point, I modified the schema only to ensure that the new checksum column is tolerated at all.

> 
> **Regarding the backward compatibility in client logic:**
> 
> Right now, the upload validation seems to still hard-require matching Adler-32 between local and remote metadata, ignoring the new checksum dictionary. So deployments that migrate to the generic checksum column without storing Adler-32 will see uploads rejected.
> 
> I think we should extend the client-side checksum comparison to use the new generic checksum dict (and fall back to legacy columns), so uploads remain compatible when Adler-32 is absent but another configured checksum is available.

This is roughly what _checksum normaliser_ is meant to do on the server side.

As a first step I wanted to provide a new column, in JSON format, keep legacy columns and use the new checksum values (still with Adler32 and md5 only) en lieu of legacy column values in _consumer_ code (i.e we select all 3 columns, but prefer the JSON one.)

When talking about configuration, Rucio has a notion of ALLOWED_CHECKSUMS, but this constant is not implemented throughout code and as you mentioned Adler32 and md5 are hardwired all over the place. Are the checksum types not RSE-specific ? They are most likely VO-specific.

martynia:
> Can you rebase also?

to HEAD master?

Geogouz:
> > Can you rebase also?
> 
> to HEAD master?

Yes

lib/rucio/db/sqla/migrate_repo/versions/8ab4d628cffb_add_generic_checksum_column.py
martynia: Fixed

martynia:
Rebased as requested.

bziemons:
Please make sure the tests run and since the commit messages are not in any accepted format, please consult the [contributing guide for how to apply our flavour of conventional commits](https://rucio.cern.ch/documentation/contributing/#4-commit-your-changes).

martynia:
> Please make sure the tests run and since the commit messages are not in any accepted format, please consult the [contributing guide for how to apply our flavour of conventional commits](https://rucio.cern.ch/documentation/contributing/#4-commit-your-changes).

@bziemons Most of the errors come from the fact that CI is using Atlas schema. The tests run locally since I changed the config to use schema defined in `generic.py`. Obviously it is a temporary solution. Atlas schema is coming from PyPI and I cannot see how I could force CI to use the generic schema. Alembic history show that database schema was changed multiple time - how were people dealing with a hardwired Atlas schema?

bziemons:
@jamesp-epcc since this is a policies topic and I am not really proficient in these, could you comment on this? My understanding is that any change should be compatible with all existing policy packages, but I might be wrong.

bari12:
> My understanding is that any change should be compatible with all existing policy packages, but I might be wrong.

This is correct.

martynia:
If a change is in a generic policy and other policies are not inheriting those changes then it would be difficult to keep the compatibility downstream, especially for proprietary policies like atlas.

I commented atlas policy in autotests for the time being, the error in CI went away, but there is still a difference between CI and tests running locally. I explained the issue in the developers forum on Mattermost.

bari12:
I am disabling the bot on this @martynia 
But one way or the other we need to get the tests green :-D

PR #8249, opened 2025-11-28
Rules: Store the priority in the history tables #8244

### LINKED ISSUE #8244: The priority is not stored in the rule history tables
### Description

Discovered as part of #8230.

In `insert_rule_history()`, the columns are listed explicitly. `priority` is missing, which appears to be an omission from commit e896c7d34.

### Steps to reproduce

1. Delete a replication rule.
2. Inspect the `rules_history` table.

### Rucio Version

_No response_

### Additional Information

_No response_

dchristidis:
I’m deliberately deviating from the current contribution guide to experiment with conventional commits.

### DIFF
--- a/lib/rucio/core/rule.py
+++ b/lib/rucio/core/rule.py
@@ -2906,23 +2906,9 @@ def insert_rule_history(
     :param session:   The Database session.
     """
     if recent:
-        models.ReplicationRuleHistoryRecent(id=rule.id, subscription_id=rule.subscription_id, account=rule.account, scope=rule.scope, name=rule.name,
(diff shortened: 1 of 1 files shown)


rdimaio (commented):
Can we test this?

dchristidis:
Testing specifically for `priority` is probably not very useful in the long-term. To test for everything, we would need to put a value in all columns of the rule, in some cases somewhat artificially. And one or two columns we may need to exclude.

Let me think about it.

In the case of the last commit (i.e. raise a `RuntimeError`), I think it would be necessary to move `valid_options` outside of the function.

dchristidis:
Will reopen once the test is implemented.

PR #8260, opened 2025-12-04
Documentation: ScopeClient docstring enhancement; Fix rucio#7845

Geogouz:
Handles rucio#7845

### DIFF
--- a/lib/rucio/client/scopeclient.py
+++ b/lib/rucio/client/scopeclient.py
@@ -24,7 +24,15 @@ from rucio.common.utils import build_url
 
 class ScopeClient(BaseClient):
 
-    """Scope client class for working with rucio scopes"""
+    """Scope client class for working with Rucio scopes.
+
+    In Rucio, a *scope* is the namespace that groups datasets, files and
+    containers underneath an account. Scopes provide isolation and ownership
+    boundaries—clients prepend the scope to data identifiers (DIDs) to avoid
+    collisions and to enforce the account responsible for the data. Common
+    examples include ``user.<username>`` or project-level scopes such as
+    ``dataops``.
+    """
 
     SCOPE_BASEURL = 'accounts'
 
@@ -34,43 +42,77 @@ class ScopeClient(BaseClient):
             scope: str
     ) -> bool:
         """
-        Sends the request to add a new scope.
+        Create a new scope for an account.
+
+        Scopes represent the namespaces that own datasets and files. Once
+        created, the scope can be used in DIDs (e.g. ``<scope>:<dataset>``) to
+        make ownership and visibility explicit.
 
         Parameters
         ----------
-        account :
-            The name of the account to add the scope to.
-        scope :
-            The name of the new scope.
+        account
+            Name of the account that will own the scope.
+        scope
+            Name of the scope to create.
 
         Returns
         -------
-            True if scope was created successfully.
+        bool
+            Literal ``True`` when the scope is created successfully.
 
         Raises
         ------
         Duplicate
-            If scope already exists.
+            If the scope already exists for the account.
         AccountNotFound
-            If account doesn't exist.
+            If the provided account does not exist.
+
+        Examples
+        --------
+        ??? Example
+
+            Create a scope for the ``dataops`` account. If the scope already
+            exists for the account, a ``Duplicate`` exception is raised.
+
+            ```python
+            from rucio.client.scopeclient import ScopeClient
+
+            # The client will pick up authentication from the standard Rucio
+            # configuration (e.g. ~/.ruciorc or RUCIO_* environment variables).
+            client = ScopeClient()
+
+            client.add_scope(account="dataops", scope="user.test")
+            ```
         """
 
-        path = '/'.join([self.SCOPE_BASEURL, account, 'scopes', quote_plus(scope)])
+        path = '/'.join(
+            [self.SCOPE_BASEURL, account, 'scopes', quote_plus(scope)]
+        )
         url = build_url(choice(self.list_hosts), path=path)
         r = self._send_request(url, method=HTTPMethod.POST)
         if r.status_code == codes.created:
             return True
         else:
-            exc_cls, exc_msg = self._get_exception(headers=r.headers, status_code=r.status_code, data=r.content)
+            exc_cls, exc_msg = self._get_exception(
+                headers=r.headers,
+                status_code=r.status_code,
+                data=r.content
+            )
             raise exc_cls(exc_msg)
 
     def list_scopes(self) -> list[str]:
         """
-        Sends the request to list all scopes.
+        List all scopes belonging to the client's authenticated account.
 
         Returns
         -------
-        A list containing the names of all scopes.
+        list[str]
+            Names of all scopes configured in Rucio.
+
+        Raises
+        ------
+        RucioException
+            Raised when the request is not successful.
         """
 
         path = '/'.join(['scopes/'])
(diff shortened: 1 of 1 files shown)


lib/rucio/client/scopeclient.py
```diff
@@ -34,43 +42,77 @@ def add_scope(
...
+
+        Examples
+        --------
+        ??? Example
+
+            Create a scope for the ``dataops`` account. If the scope already
+            exists for the account, a ``Duplicate`` exception is raised.
+
+            ```python
+            from rucio.client.scopeclient import ScopeClient
```
voetberg, line 78: Is there a reason to use the specific scope client here instead of just the generic Client?

lib/rucio/client/scopeclient.py
```diff
@@ -34,43 +42,77 @@ def add_scope(
...
+                headers=r.headers,
+                status_code=r.status_code,
+                data=r.content
+            )
             raise exc_cls(exc_msg)
 
     def list_scopes(self) -> list[str]:
         """
-        Sends the request to list all scopes.
+        List all available scopes.
```
voetberg: "available" is sort of a keyword, I'm a little nervous this may imply there is a state attached to scopes where "unavailable" is also an option. Maybe - 
```suggestion
        List all scopes belonging to the client's authenticated account. 
```

lib/rucio/client/scopeclient.py
Geogouz: Basically for consistency/uniformity with the other finished clients so far. But it is a fair point. We could switch to the general one too since the other clients become accessible through that. But if we choose to do so, we should better update all examples across the various clients.

Geogouz:
The ruff error is unrelated and is being handled by rucio/rucio#8269, rucio/rucio#8267 and rucio/rucio#8265.

voetberg:
> The ruff error is unrelated and is being handled by rucio/rucio#8269, rucio/rucio#8267 and rucio/rucio#8265.

The error is because of a trailing space on the docstring `lib/rucio/client/scopeclient.py:105:73` which isn't related to the `noqa` checks. Please remove that trailing space.

Geogouz:
> > The ruff error is unrelated and is being handled by rucio/rucio#8269, rucio/rucio#8267 and rucio/rucio#8265.
> 
> The error is because of a trailing space on the docstring `lib/rucio/client/scopeclient.py:105:73` which isn't related to the `noqa` checks. Please remove that trailing space.

Yes. Wrongly stated. In the failed "python_ruff" check, the `#noqa` warnings are unrelated and handled elsewhere.

lib/rucio/client/scopeclient.py
Geogouz: Are we keeping this one eventually? Or should I switch to the generic one?

lib/rucio/client/scopeclient.py
Geogouz: The trailing space is coming from here. Should I just send a new squashed and rebased fixed version?

lib/rucio/client/scopeclient.py
voetberg: Let's keep it for now. Because we don't have rules stating else anywhere I don't see a problem with it. It's an easy thing to swap at a later date if we decide to.

voetberg (approved):

lib/rucio/client/scopeclient.py
```diff
@@ -34,43 +42,77 @@ def add_scope(
...
+            client = ScopeClient()
+
+            client.add_scope(account="dataops", scope="user.test")
+            ```
         """
 
-        path = '/'.join([self.SCOPE_BASEURL, account, 'scopes', quote_plus(scope)])
+        path = '/'.join(
+            [self.SCOPE_BASEURL, account, 'scopes', quote_plus(scope)]
+        )
```
rdimaio, line 90: nit: formatting changes should be in a separate commit to the docstring change

lib/rucio/client/scopeclient.py
```diff
@@ -34,43 +42,77 @@ def add_scope(
...
         r = self._send_request(url, method=HTTPMethod.POST)
         if r.status_code == codes.created:
             return True
         else:
-            exc_cls, exc_msg = self._get_exception(headers=r.headers, status_code=r.status_code, data=r.content)
+            exc_cls, exc_msg = self._get_exception(
+                headers=r.headers,
+                status_code=r.status_code,
+                data=r.content
+            )
```
rdimaio, line 100: nit: formatting changes should be in a separate commit to the docstring change

lib/rucio/client/scopeclient.py
```diff
@@ -80,28 +122,51 @@ def list_scopes(self) -> list[str]:
             scopes = loads(r.text)
             return scopes
         else:
-            exc_cls, exc_msg = self._get_exception(headers=r.headers, status_code=r.status_code, data=r.content)
+            exc_cls, exc_msg = self._get_exception(
+                headers=r.headers,
+                status_code=r.status_code,
+                data=r.content
+            )
```
rdimaio, line 129: nit: formatting changes should be in a separate commit to the docstring change

lib/rucio/client/scopeclient.py
```diff
@@ -112,5 +177,9 @@ def list_scopes_for_account(self, account: str) -> list[str]:
...
```
rdimaio, line 184: nit: formatting changes should be in a separate commit to the docstring change

rdimaio (approved):

lib/rucio/client/scopeclient.py
```diff
@@ -24,7 +24,15 @@
...
 
-    """Scope client class for working with rucio scopes"""
+    """Scope client class for working with Rucio scopes.
+
+    In Rucio, a *scope* is the namespace that groups datasets, files and
+    containers underneath an account. Scopes provide isolation and ownership
+    boundaries—clients prepend the scope to data identifiers (DIDs) to avoid
+    collisions and to enforce the account responsible for the data. Common
+    examples include ``user.<username>`` or project-level scopes such as
+    ``dataops``.
```
dchristidis, line 34: 1. Though correct, there is a bit of reuse of terms with a different meaning. In the Rucio publication, *namespace* somewhat refers to the entire Rucio catalogue. Scopes create *partitions* of the global namespace.
2. An em-dash is not very visible in monospaced font. Please use a semicolon instead.
3. I am sceptical about saying that the scope is *prepended* to the DID. A DID, but definition, is the combination of a *scope* and a *name*, separated by a colon. Some communities follow a convention that the *name* starts with the scope too, but that is not a requirement. There is also a mechanism to try to guess the *scope* from the *name*.
4. Are these examples truly common? I am genuinely asking.

lib/rucio/client/scopeclient.py
dchristidis: I have to agree here. Especially since this change may not be universally agreeable. We don’t have a set line-length limit. This applies to the other instances as well.

lib/rucio/client/scopeclient.py
```diff
@@ -34,43 +42,77 @@ def add_scope(
...
+        ??? Example
+
+            Create a scope for the ``dataops`` account. If the scope already
+            exists for the account, a ``Duplicate`` exception is raised.
+
+            ```python
+            from rucio.client.scopeclient import ScopeClient
+
+            # The client will pick up authentication from the standard Rucio
+            # configuration (e.g. ~/.ruciorc or RUCIO_* environment variables).
```
dchristidis, line 81: I’ve never heard of `~/.ruciorc` before. Is that an actual functionality?

dchristidis (changes requested):

dchristidis:

Geogouz:
Hmmm.. I force-pushed with corrections/adjustments and now I can't reopen it. Lesson learned for the future re-openings.. See here: https://github.com/rucio/rucio/pull/8468

lib/rucio/client/scopeclient.py
Geogouz: Indeed DaFab specific with overridden functionality for our pilot clients. Removing it because I don't expect this to end up upstream.

PR #8285, opened 2025-12-16
Testing: Fixes a silent failure in `bootstrap_dev.sh` due to Bash 5.1

### LINKED ISSUE #8284: Testing: `bootstrap_dev.sh` exits silently on Linux (Bash 5.1+) when using profiles
When running `bootstrap_dev.sh -p iam (or with any profile)` , the script exits immediately and silently on Linux systems running modern Bash versions (e.g., Bash 5.1).

## Steps to Reproduce
1. On a Linux machine (e.g., Any system with Bash 5.1+).
2. Run the bootstrap script with a profile:
   ```bash
   ./tools/bootstrap_dev.sh -p iam

On the first iteration, PROFILE_COUNT is 0. The post-increment operator returns 0. In Bash arithmetic contexts, a result of 0 is treated as "failure" (exit code 1). Since the script runs with set -e, this causes an immediate termination. Older Bash versions (like 3.2 on macOS) are more lenient and do not trigger set -e in this specific case.

0xquark:
Fixes #8284

### DIFF
--- a/tools/bootstrap_dev.sh
+++ b/tools/bootstrap_dev.sh
@@ -341,8 +341,7 @@ fi
 SPECIFIED_RELEASE=""
 USE_MASTER="false"
 USE_LATEST="false"
-PROFILES=()           # Named profiles
-PROFILE_COUNT=0       # Track how many named profiles were provided
+PROFILES=()           # Named profiles (array); count via ${#PROFILES[@]}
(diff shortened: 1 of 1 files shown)


Geogouz (approved):
Looks good. Also safe to merge.

rdimaio (approved):

PR #8236, opened 2025-11-24
Default permission policies: handle sign-url account attribute

jouvin:
This PR fixes the default permission policies to take into account the `sign-url` account attribute. As explained in the [issue](#8235), the original name `sign-gcs` doesn't make obvious what it is intended for. As it has never been used into default permission policies (or elsewhere in the standard code), this change is not considered a backward-compatibility issue and will have no impact on existing, non-default, permission policies.

## Checklist
- [x] Created issue: #8235
- [ ] Added tests
- [x] Updated documentation (https://github.com/rucio/documentation/pull/797)
- [ ] Added database migrations - not needed
- [ ] For backwards compatibility breaking changes, leave additional description, follow conventional commits
- [x] Picked a reviewer after submission (Leave empty, if unclear)
- [x] I understand that stale (failing tests, author not answering) PRs will be closed promptly

## Additional description for reviewer

*Note: This OPTIONAL is only relevant for the REVIEWER, please leave it in the PR*

<details>
<summary>Reviewer template </summary>
Reviewers should copy&paste the code-block below and fill it out for APPROVED pull requests.
If the PR does not meet the standards the project sets out, the reasons should be WELL EXPLAINED in a CHANGE REQUEST (The answers below do not need to be answered in that case)

- **Confidence in review**: I am confident in my review concerning the components this PR touches: [*High 🟢, Medium 🟡 Low 🔴*]
- **Confidence in scope**: I am confident that this fits into the scope of the project and should be included: [*High 🟢, Medium 🟡, Low 🔴*]
  - For Medium and Low, explain in notes why this should be included
- **Quality**: The approach is sound, maintainable and addresses the issue in the best way: [*Agree 🟢*]
- **Security**: This PR does NOT require increased attention in terms of security (E.g. new dependencies): [*Agree 🟢, Disagree 🔴*]
  - If *Disagree* explain in notes.
- **Backwards compatibility**: This PR does NOT introduce backwards compatibility breaking changes: [*Agree 🟢, Disagree 🔴*]
  - If *Disagree* explain in notes
- **Testing**: This PR is well tested: [*Agree 🟢*]
- **Documentation**: Relevant documentation or comments are updated or not required: [*Agree 🟢*]

```
- **Confidence in review**: High 🟢 Medium 🟡 Low 🔴
- **Confidence in scope**: High 🟢 Medium 🟡 Low 🔴
- **Quality**: Agree 🟢
- **Security**: Agree 🟢 Disagree 🔴
[...]

PR #8289, opened 2025-12-18
expose scopes to perm_xxx_replicas permissions

### LINKED ISSUE #8288: 38.5.1: default permission policy is unusable
### Description

According to [Rucio S3 documentation](https://rucio.github.io/documentation/operator/s3_rse_config), a user (account) without privileges should be able to upload files to scopes they owns Unfortunately the default permission policy (`core/permission/generic.py`) doesn't allow it because the scope is not passed to `has_permission()` function and thus cannot be tested by `perm_add_replicas()`, `perm_delete_replicas()` and `perm_update_replicas_states()` functions.

This issue is the upload-side of #8235...

### Steps to reproduce

1. Create an unprivileged account with a quota on some RSEs
2. Create a scope owned by this account
3. Execute `rucio upload --scope previous_created_scope ...` as the previously created user

### Rucio Version

Client 38.5.1, Server 38.5.1

### Additional Information

- Kubernetes
- Server-side bug

### LINKED ISSUE #8333: Fix default permissions for uploading (perm_add_replica and perm_update_replicas)
### Description

According to the Rucio doc, an unprivileged user should be able to upload files in their own scope and to an RSE on which it has quota. However this clashes with the default [permission schema](https://github.com/rucio/rucio/blob/master/lib/rucio/core/permission/generic.py).

In particular, it is required that:
- to add replicas one needs either to be admin, or the RSE end with MOCK, SCRATCHDISK, LOCALGROUPDISK (clearly ATLAS-specific) - [code](https://github.com/rucio/rucio/blob/2668081ff217efb9726314298f395b180ce02ff3/lib/rucio/core/permission/generic.py#L690-L704)
- to update replicas one needs to be admin - [code](https://github.com/rucio/rucio/blob/2668081ff217efb9726314298f395b180ce02ff3/lib/rucio/core/permission/generic.py#L731-L740)

To ensure consistent behaviour, requirements should be like:

`return _is_root(issuer) or has_account_attribute(account=issuer, key='admin', session=session) or rucio.core.scope.is_scope_owner(scope=kwargs['scope'], account=issuer, session=session)`

I suggest to drop the requirements on the RSE name

### Steps to reproduce

1. create an account `rucio-admin account add --type USER --email jdoe@blahblih.com jdoe`
2. create scope `rucio-admin scope add --account jdoe --scope user.jdoe`
3. add quota on RSE `rucio-admin account set-limits jdoe RSE_MOCK 10000000000000`
4. do rucio upload 
```
rucio -v upload pippo.txt --rse RSE_MOCK 
INFO    Preparing upload for file pippo.txt
INFO    Successfully added replica in Rucio catalogue at RSE_MOCK
INFO    Successfully added replication rule at RSE_MOCK
INFO    Trying upload with root to RSE_MOCK
INFO    Successful upload of temporary file. root://xxx/f6/e2/pippo.txt.rucio.upload
INFO    Successfully uploaded file pippo.txt
ERROR   Failed to update replica state for file pippo.txt
ERROR   None of the given files have been uploaded.
```

### Rucio Version

35.*

### Additional Information

_No response_

jouvin:
Fixes #8333.

Originally linked to #8288 closed in favor of #8333.

### DIFF
--- a/lib/rucio/core/permission/generic.py
+++ b/lib/rucio/core/permission/generic.py
@@ -689,20 +689,29 @@ def perm_declare_suspicious_file_replicas(issuer: "InternalAccount", kwargs: dic
 
 def perm_add_replicas(issuer: "InternalAccount", kwargs: dict[str, Any], session: "Session") -> bool:
     """
-    Checks if an account can add replicas.
+    Checks if an account can add replicas. If the RSE is not a special one (name terminated by some magic words)
+    or if the account is not an administrator, for each replica scope (passed in kwargs['scopes'] for convenience),
+    the ownership is checked: if the account is not owning all of them, permission is denied.
 
     :param issuer: Account identifier which issues the command.
     :param kwargs: List of arguments for the action.
     :param session: The DB session to use
     :returns: True if account is allowed, otherwise False
     """
-    return str(kwargs.get('rse', '')).endswith('SCRATCHDISK')\
-        or str(kwargs.get('rse', '')).endswith('USERDISK')\
-        or str(kwargs.get('rse', '')).endswith('MOCK')\
-        or str(kwargs.get('rse', '')).endswith('LOCALGROUPDISK')\
-        or _is_root(issuer)\
-        or has_account_attribute(account=issuer, key='admin', session=session)
 
+    if str(kwargs.get('rse', '')).endswith('SCRATCHDISK')\
+        or str(kwargs.get('rse', '')).endswith('MOCK')\
+        or _is_root(issuer)\
+        or has_account_attribute(account=issuer, key='admin', session=session):
+        permission_given = True
+    else:
+        permission_given = True
+        for scope in kwargs['scopes']:
+            if not rucio.core.scope.is_scope_owner(scope, account=issuer, session=session):
+                permission_given = False
+                break
+
+    return permission_given
 
 def perm_skip_availability_check(issuer: "InternalAccount", kwargs: dict[str, Any], session: "Session") -> bool:
     """
@@ -730,14 +739,26 @@ def perm_delete_replicas(issuer: "InternalAccount", kwargs: dict[str, Any], sess
 
 def perm_update_replicas_states(issuer: "InternalAccount", kwargs: dict[str, Any], session: "Session") -> bool:
     """
-    Checks if an account can delete replicas.
+    Checks if an account can add replicas. If the account is not an administrator,,
+    for each replica scope (passed in kwargs['scopes'] for convenience), the ownership is checked: if the account
+    is not owning all of them, permission is denied.
 
     :param issuer: Account identifier which issues the command.
     :param kwargs: List of arguments for the action.
     :param session: The DB session to use
     :returns: True if account is allowed, otherwise False
     """
-    return _is_root(issuer) or has_account_attribute(account=issuer, key='admin', session=session)
+    if  _is_root(issuer)\
+        or has_account_attribute(account=issuer, key='admin', session=session):
+        permission_given = True
+    else:
+        permission_given = True
+        for scope in kwargs['scopes']:
+            if not rucio.core.scope.is_scope_owner(scope, account=issuer, session=session):
+                permission_given = False
+                break
+
+    return permission_given
 
 
 def perm_queue_requests(issuer: "InternalAccount", kwargs: dict[str, Any], session: "Session") -> bool:
--- a/lib/rucio/gateway/replica.py
+++ b/lib/rucio/gateway/replica.py
@@ -13,8 +13,9 @@
 # limitations under the License.
 
 import datetime
-from typing import TYPE_CHECKING, Any, Literal, Optional, Union, cast
+from typing import TYPE_CHECKING, Any, Literal, Optional, Tuple, Union, cast
 
+from requests import Session
 from rucio.common import exception
 from rucio.common.constants import DEFAULT_VO, SuspiciousAvailability
 from rucio.common.schema import validate_schema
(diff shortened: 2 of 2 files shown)


jouvin:
It is my attempt to fix the problem that scopes are not passed to the `has_permission()` function when adding (or deleting) a replica when they are supposed to be the Rucio object on which permissions are applied when uploading or downloading DIDs. 

It is more intrusive than #8236 as it is necessary to build the scope list in `gateway/replica.py` and pass it in the `kwargs` argument to the permission module (`core/permission/generic.py` for example). The alternative would be to require the permission module to look in all replicas passed to get the scope and to what is done once in `gateway/replica.py` so IMO it is less optimal. And anyway scopes are not passed at all in `update_replicas_states()` in `gateway/replica.py` so a modification of this module would be necessary anyway...

I am not enough a Rucio expert to ensure that there is no side effect to this modification but it's my feeling: nothing passed previously is modified, just an entry `scopes` is added in the `kwargs` dict so it should be harmless. And the resulting default permission of this PR is a basic configuration that you can use to test your Rucio installation, including upload operations, without writing a permission policy as your first step when configuring a new instance. And production instance probably all have their own policy file so they will not be impacted by the updated `generic.py` (a similar update is required in `generic_muti_vo.py` once we agreed on `generic.py`). I think that exposing scopes to replica-related permission functions could also benefit to some production instance that had to work around the absence of scopes with account attributes....

Interested by feedback from experts!

lib/rucio/core/permission/generic.py
```diff
@@ -689,20 +689,29 @@ def perm_declare_suspicious_file_replicas(issuer: "InternalAccount", kwargs: dic
...
 
     :param issuer: Account identifier which issues the command.
     :param kwargs: List of arguments for the action.
     :param session: The DB session to use
     :returns: True if account is allowed, otherwise False
     """
-    return str(kwargs.get('rse', '')).endswith('SCRATCHDISK')\
-        or str(kwargs.get('rse', '')).endswith('USERDISK')\
+
+    if str(kwargs.get('rse', '')).endswith('SCRATCHDISK')\
```
maxnoe, line 702: Maybe refactor this a bit with early exits instead of ors to make it easier to reason about?

And maybe a constant for the `SPECIAL_RSE_SUFFIXES` 

```python
rse = kwargs.get("rse", "")

if _is_root(issuer):
    return True
if has_account_attribute(account=issuer, key='admin', session=session):
   return True
if any(rse.endswith(suffix) for suffix in SPECIAL_RSE_SUFFIXES):
    return True

# Also allow if user owns all affected scopes
if all(is_scope_owner(scope, issuer, session=session) for scope in kwargs["scopes"]):
    return True

return False
```

bziemons:

jouvin:
@bziemons sorry, I have been away from Rucio since the beginning of the year but I hope to restart in the coming weeks and will try to fix the problems.

bziemons:
> @bziemons sorry, I have been away from Rucio since the beginning of the year but I hope to restart in the coming weeks and will try to fix the problems.

no need to be sorry :) We just need to get on top of PRs again, so closing the ones that haven't moved for a while helps us focusing our reviewing efforts and will get PRs merged more quickly. Please re-open when the PR is ready to review as mentioned in the last sentence of my closing comment

jouvin:
I restart to work on this PR and reopen it. I forced pushed an updated version rebased on last master (release of 40.0.0rc1): no change in the modifications. I'll fix the issues reported by the CI asap.

lib/rucio/core/permission/generic.py
```diff
@@ -689,20 +689,29 @@ def perm_declare_suspicious_file_replicas(issuer: "InternalAccount", kwargs: dic
 
 def perm_add_replicas(issuer: "InternalAccount", kwargs: dict[str, Any], session: "Session") -> bool:
     """
-    Checks if an account can add replicas.
+    Checks if an account can add replicas. If the RSE is not a special one (name terminated by some magic words)
```
maxnoe, line 692: The first line should stay short followed by an empty line (https://rucio.cern.ch/documentation/developer/dev_style_guide, linking to https://numpydoc.readthedocs.io/en/latest/format.html#short-summary)

```suggestion
    Checks if an account can add replicas.
    
    If the RSE is not a special one (name terminated by some magic words)
```

lib/rucio/core/permission/generic.py
jouvin: I agree with your proposal but not that the special RSE suffixes are already there and it is not my addition...

jouvin:
@maxnoe I pushed changes addressing your remarks and suggestions. How I can reopen the PR?

Not sure why the new commit pushed doesn't show up... May be because the PR is closed...

I also don't know what I should do about Codecov report.

maxnoe:
You should have a button here, right?

jouvin:
No I don't... May be only admins can do it...

jouvin:
@bziemons I'd like to reopen this PR but I don't find how to do it. I don't have the button to reopen the PR...

bziemons:
> @bziemons I'd like to reopen this PR but I don't find how to do it. I don't have the button to reopen the PR...

Unfortunately, me neither:

image

I have never seen that message. @bari12 can you re-open the PR?

jouvin:
Yes, I force-pushed the source branch. I never had a problem with this but may be GitHub changed something when it happens and the PR is closed... I found this post that seems somewhat related: https://gist.github.com/robertpainsi/2c42c15f1ce6dab03a0675348edd4e2c

May be the easiest is to forget this PR... I'll create a new one and reference this one but I don't think there is so much information useful here. Mainly @maxnoe suggestions that I implemented.

bari12:
Same for me, it shows the same message and no re-open button. I can only imagine something happened with the origin branch which makes GH think that this will never be mergable?

jouvin:
I'll  open a new PR as suggested earlier

dchristidis:
To reopen a PR, the branch must be in the exact state it was at the time it was closed. If you’ve already modified, you can push the original branch: `git push --force origin d094f08:scope_in_upload_generic_permission` (unless the housekeeping of the local repository has already cleaned it).

PR #8305, opened 2026-01-05
Feature 7880 automatic mapping of users

### LINKED ISSUE #8413: Automatic mapping of unknown users identified via OIDC
### Description

To allow experiments to share data it would be useful to allow users identified using OIDC providers from other experiments to read data. In particular Einstein Telescope and Cosmic Explorer would like to share data without having to setup users for everyone in the other experiment.

To enable this everyone with a valid token from a configured OIDC provider will be mapped to a configured account.  

### Motivation

_No response_

### Change

_No response_

Twokubikmeter:
Implements of the last item in 7880. 
"Users from collaborating experiments should not need to be configured.
-- The instance should accept everyone authenticated against the oidc provider hosted by the collaborating experiment."

### DIFF
--- a/lib/rucio/core/oidc.py
+++ b/lib/rucio/core/oidc.py
@@ -1409,7 +1409,8 @@ def validate_jwt(json_web_token: str, *, session: "Session") -> dict[str, Any]:
         token_dict: Optional[dict[str, Any]] = __get_rucio_jwt_dict(json_web_token, session=session)
         if not token_dict:
             raise CannotAuthenticate(traceback.format_exc())
-        issuer = token_dict['identity'].split(", ")[1].split("=")[1]
+        token_payload = __get_keyvalues_from_claims(json_web_token, ['iss'])
+        issuer = token_payload['iss']
         oidc_client = OIDC_CLIENTS[issuer]
         issuer_keys = oidc_client.keyjar.get_issuer_keys(issuer)
         JWS().verify_compact(json_web_token, issuer_keys)
@@ -1454,8 +1455,15 @@ def oidc_identity_string(sub: str, iss: str):
     :param sub: users SUB claim from the Identity Provider
     :param iss: issuer (IdP) https url
 
-    :returns: OIDC identity string "SUB=<usersid>, ISS=https://iam-test.ch/"
+    :returns: OIDC identity string "SUB=<usersid>, ISS=https://iam-test.ch/" or a configured identity string for the iss
     """
+    shared_identity_configs = config_get("oidc", "shared_identity_configs", False, default=None)
+    if shared_identity_configs:
+        for section in [x.strip() for x in shared_identity_configs.split(',')]:
+            shared_identity_iss = config_get(section, "iss")
+            if shared_identity_iss == iss:
+                shared_identity_string = config_get(section, "shared_identity_string")
+                return shared_identity_string
     return 'SUB=' + str(sub) + ', ISS=' + str(iss)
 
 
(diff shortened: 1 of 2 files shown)


bziemons:
Depends on #8160

bziemons:
@Twokubikmeter please rebase and check tests afterwards.

dchristidis:
Neither the referenced issue nor the content of this PR provide sufficient context. Could you open a new issue and elaborate?

Twokubikmeter:
> Neither the referenced issue nor the content of this PR provide sufficient context. Could you open a new issue and elaborate?

I have created issue #8413

bari12:
I have spend some time thinking about this now, and also had a chat with @dchristidis about the wider impact and believe we should do this a bit different.

The main problem I see is that we hide yet another thing behind a config option. For an admin/operator it's extremely hard to see that this is happening. If you just list all identities of a Rucio instance, you would have no idea that somebody, of that `iss` would have access to your Rucio instance. Plus, it further leads a bit to the inflation of config options, which we try to avoid.

What I would propose is to do this different and actually add an identity for that behaviour, under a new `identity_type` to the identity table. Basically a row like:

```
identity_type   | identity
--------------- | -------------------------------------------
OIDC            | sub=martin@cern.ch, iss=https://iam.cern.ch
OIDC_GENERAL    | iss=https://einstein.telescope
```

Now this identity can be mapped against the guest account in the conventional way. In addition, you could give some subjects under iss=einstein.telescope even individual accounts.

I know this diverts from the path we originally discussed, apologies for that, but I think this is actually more consistent with the typcial identity->account mapping behaviour we have and would be the better solution.

`OIDC_GENERAL` we might need a better name though, this is just an idea.

Twokubikmeter:
I will do these changes and I unless you tell me not to I will reuse this pull request with the new version.

Twokubikmeter:
The database has a limit on identity_type of 8 characters. So I suggest OIDC_ALL. It is the most descriptive name I could make up under 8 letters.
identity_type   | identity
---------------- | -------------------------------------------
OIDC           | sub=martin@cern.ch, iss=https://iam.cern.ch
OIDC_ALL   | iss=https://einstein.telescope

maxnoe:
How is "users from other experiments shouldn't need to be manually created" different from "Users should be automatically created for configured Identity Providers"?


Isn't this basically re-implementing the standard OIDC sign-up flow in a different way?

Usually, it works like this:
An admin configures the server to allow logins from a specific identity provider by installing an OIDC client and on first login the user is asked with which provider they want to authenticate, they are forwarded to that provider, login and are sent back to the server, which receives the user scim payload to create the new account (or attach it to an existing one if the "add new login flow" is used).

Twokubikmeter:
> How is "users from other experiments shouldn't need to be manually created" different from "Users should be automatically created for configured Identity Providers"?
> 
> Isn't this basically re-implementing the standard OIDC sign-up flow in a different way?
> 
> Usually, it works like this: An admin configures the server to allow logins from a specific identity provider by installing an OIDC client and on first login the user is asked with which provider they want to authenticate, they are forwarded to that provider, login and are sent back to the server, which receives the user scim payload to create the new account (or attach it to an existing one if the "add new login flow" is used).

It would be preferable to not have users for the cooperating experiment and the experiment would not need a client on the OIDC provider from the other experiment.

bari12:
I will close the PR, please re-open once the new changes are there. For discussions, let's have them on the issue.

PR #8306, opened 2026-01-05
dynamic iam client registration #8252

### LINKED ISSUE #8252: Dynamically register IAM clients
### Description

Right now we mantain .sql script to add client_id and secret in indigoIAM with its related config [here](https://github.com/rucio/rucio/blob/master/etc/docker/dev/iam/indigoiam_db.sql) .

This makes it hard to mantain in any schema changes upstream and will be hard to keep up with upstream release.

We can use dynamic client registration via `h2-test` profile of indigoIAM, where we need two clients one
 with auth code flow and another with client_credential . 

### Motivation

To streamline the test infrastructure and reduce complexity.

### Change

We need to add a script which runs via a container or during initialization and register clients and write the credential of registered clients to json file in rucio idpsecrets.json format.
then use that in rucio server/daemon mounts.

Also need to update FTS DB entry [here](https://github.com/rucio/rucio/blob/master/etc/docker/dev/fts/entrypoint.sh#L26)

panta-123:
Move away from mantaining the manual sql for adding clients to indigoiam to dynamic client registration. This gives long term mantainence and keep up with latest release of indigoiam.

Dynamically register client via new init container.
- use client from `h2-test` profile of indigoiam and user admin token to register new clients.
- added new init container in docker compose which run a script.
- The script creates rucio IdP client and fts IdP client with approriate values.
- The script updates the idpsecrets.json and also update the FTS database to add the client.

The commit are seperated in small chunks to make it digestable for review. (can be sqashed to merge)


0xquark:
Please update your PR by rebasing it on the latest `master` branch. That should fix the CI issues.

0xquark (commented):
Hi @panta-123 !I think you used the wrong format for the conventional commits. According to the [documentation](https://rucio.github.io/documentation/contributing/), your commits should look like this:
```
feat(Containerization): Change......

Issue: #IssueNumber
```
Thanks for updating it.

panta-123:
@0xquark , The commit was based on conventional commits PR (not merged at that time), there was Docker . The PR now merged was seem to be updated to `Containerization`. I will make the ammend to commit accordingly.

bari12 (changes requested):
The fts/entrypoint doesn't this need to be updated as well? I feel like the whole things falls apart if this is not done in sync? 

- The init container is in the iam profile and depends only on indigoiam-login-service, yet main() unconditionally connects to ftsdb, which lives in the storage profile with no healthcheck. Bringing up iam without storage makes the init container always fail after already writing idpsecrets.json; even with both profiles there is no guarantee fts-database-upgrade.py has created t_token_provider first.

- you called the container dev-indigoiam_init-1 please use dashes instead of underscore :-D to keep this in sync

bziemons:
@panta-123 are you still working on this?

panta-123:
@bziemons 
I have still not come up with solution on how to address @bari12's comments.
If this PR is worth from your point of view, I can look into how to address the comment or I will be welcome to any suggestion.

panta-123:
I think the following can solve @bari12's  comment.
- add fts healthcheck 
- Make init-iam container depends on both indigoIAM (required=True) and fts (required=False)
```
indigoiam-init:
    ....
    ...
   depends_on:
      indigoiam-login-service:
          condition: service_healthy
      fts:
           condition: service_healthy
           required: false
```

Then
- storage profile first, IAM profile second: On second run fts is already healthy (table exists), so when init runs it writes the t_token_provider row straight away.

- IAM profile first, storage profile second: fts  isn't up yet, so init skips the DB write with a warning (still writes idpsecrets.json and exits cleanly); the row gets written on the next init run once storage profile is up.

- both together: required=false on fts contaner + the healthcheck make init wait until fts is healthy before writing — no race, no polling.

@bari12 , If you think this okay then I will make a overwrite commit.

PR #8320, opened 2026-01-09
Policies: don't fail silently if policy package doesn't load #7962

### LINKED ISSUE #7962: Rucio silently skips importing configured policy package in case of errors
### Description

I had by mistake included quotes around the policy package name:

```
[policy]
package = "bdms_rucio_policy"
```

Which where taken over and rucio was looking for a module including the quotes.

The result was rucio falling back to the generic schema without any message.

I think rucio should at least show a warning if a policy package is configured but fails to load.

### Steps to reproduce

Configure a policy package that fails to import

### Rucio Version

server 38 and current master

### Additional Information

_No response_

jamesp-epcc:
This modifies the code that loads schema and permission modules from policy packages to throw an exception if the configured policy package cannot be loaded. Previously this would silently fail.

### DIFF
--- a/lib/rucio/common/schema/__init__.py
+++ b/lib/rucio/common/schema/__init__.py
@@ -69,8 +69,18 @@ if not _is_multivo():
         package_module = importlib.import_module(policy)
         check_policy_module_version(package_module)
         policy = policy + ".schema"
-    except (NoOptionError, NoSectionError, ModuleNotFoundError):
+    except (NoOptionError, NoSectionError):
         policy = 'rucio.common.schema.' + GENERIC_FALLBACK.lower()
+    except ModuleNotFoundError:
+        if config.is_client():
+            # policy package may not be required/installed on client, but client may
+            # share config file with server, so only a warning for this case
+            LOGGER.warning('Unable to find policy package %s' % policy)
+            policy = 'rucio.common.schema.' + GENERIC_FALLBACK.lower()
+        else:
+            raise exception.PolicyPackageNotFound(policy)
+    except ImportError:
+        raise exception.ErrorLoadingPolicyPackage(policy)
 
     try:
         module = importlib.import_module(policy)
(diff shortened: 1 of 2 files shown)


lib/rucio/common/schema/__init__.py
```diff
@@ -69,8 +69,12 @@ def _get_generic_schema_module():
         package_module = importlib.import_module(policy)
         check_policy_module_version(package_module)
         policy = policy + ".schema"
-    except (NoOptionError, NoSectionError, ModuleNotFoundError):
+    except (NoOptionError, NoSectionError):
         policy = 'rucio.common.schema.' + GENERIC_FALLBACK.lower()
+    except ModuleNotFoundError:
```
voetberg, line 74: Should this send a warning to the log instead? I'm worried this would cause problems for client versions where the policy package doesn't have to be present. (as schema is part of common)

jamesp-epcc: Possibly. I was thinking maybe it should be an error for a policy package that's configured to fail to load. But do people commonly share config files between client and server? If so, maybe a warning is more appropriate since a policy package configured for the server might not be present on the client.

voetberg: I don't think it's unreasonable to assume people are sharing them between server and client. We could also add a "is_client" check to direct if it's a warning or an error.

jamesp-epcc: I think checking for client mode and deciding based on that is a good idea. I will implement this.

lib/rucio/common/schema/__init__.py
```diff
@@ -69,8 +69,17 @@ def _get_generic_schema_module():
...
+        if config.is_client():
```
voetberg, line 75: Because this block doesn't reset the policy to the generic fallback, this will just lead to an import error later in logic

jamesp-epcc: You are right... good catch. I'll fix that.

voetberg (approved):

Issue #6545, opened 2024-03-12
Introduce test coverage in PR reviews (and possibly in CI)


rdimaio:
Rucio used to have test coverage some time ago, but it was toggled off. I think it would be a good tool to have for PR review purposes; if there is a portion of code that is not covered, it might be a good indicator that some tests need to be written for that part of the code.

As we use pytest for testing, we can use https://github.com/pytest-dev/pytest-cov for coverage.

There are some criticisms for using coverage as a metric: the main reason behind this is that coverage only tells you whether code is covered or not, but it doesn't tell you much about the quality of the tests. I somewhat agree with this, and I would suggest only using coverage as a PR review tool, but we could also consider having a fairly achievable coverage threshold (60-70%) and aiming to always be over that threshold.

bziemons:
The current coverage metric does not suffice for reviewing pull requests and it just creates noise. We have decided to roll back the coverage display and re-add this once it is properly implemented.

PR #8321, opened 2026-01-09
Testing: Remove Code Coverage Display

### LINKED ISSUE #8330: Remove incomplete test coverage
> The current coverage metric does not suffice for reviewing pull requests and it just creates noise. We have decided to roll back the coverage display and re-add this once it is properly implemented. 

 _Originally posted by @bziemons in [#6545](https://github.com/rucio/rucio/issues/6545#issuecomment-3728910317)_

# Motivation

Every PR currently gets unhelpful noise by a coverage display that does only use coverage data from simple unit tests. The displayed test coverage is unhelpful and does not include tests that a PR submitter may have written. It is just noise at the moment.

bziemons:
This removes the code coverage upload within the unit test workflow. Addresses https://github.com/rucio/rucio/issues/6545#issuecomment-3728910317

@maany maybe you can comment on whether we need the unit test workflow. I didn't see any other workflow run tests on Python 3.11 and 3.12. I left the workflow in for now, because it takes about a minute per Python version to complete.

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -37,9 +37,4 @@ jobs:
             run: pip install -r requirements/requirements.dev.txt
 
           - name: Run unit tests
-            run: pytest tests/rucio --cov=lib/rucio
-
-          - name: Upload coverage reports to Codecov
-            uses: codecov/codecov-action@v5.5.2
-            with:
-              token: ${{ secrets.CODECOV_TOKEN }}
+            run: pytest tests/rucio


maany (dismissed):
Thanks @bziemons! As a follow-up, should we also remove the github secret and probably the account on CodeCov ( @bari12 )?

maany:
@bziemons Regarding unit tests, let’s discuss this in the next CI refactor call so we can evaluate them alongside the other test suites we may want to remove. I like that the unit tests are simple to set up, but we might be re-testing the same things already covered elsewhere.

bziemons:
> @bziemons Regarding unit tests, let’s discuss this in the next CI refactor call so we can evaluate them alongside the other test suites we may want to remove. I like that the unit tests are simple to set up, but we might be re-testing the same things already covered elsewhere.

I agree. Then we can go ahead with this MR and it can be merged.

dchristidis (changes requested):
Kindly create a new issue for this.

bziemons:
…

dchristidis:
Apologies, I messed this one up with #8283.

dchristidis (approved):

PR #8322, opened 2026-01-09
gateway: validate SubscriptionState in update_subscription #7820

### LINKED ISSUE #7820: `update_subscription` in `gateway/subscription`: validate `SubscriptionState`
See https://github.com/rucio/rucio/pull/7796#discussion_r2189777449

jainoshika:
### Description
Validate the `SubscriptionState` passed in `update_subscription` by catching `ValueError` and returning a proper gateway-level exception instead of letting the raw exception propagate.

### Motivation
The gateway currently assumes that the client always passes a valid `SubscriptionState`. This change makes the validation explicit and improves error handling for invalid input.

Fixes #7820

### DIFF
--- a/lib/rucio/gateway/subscription.py
+++ b/lib/rucio/gateway/subscription.py
@@ -133,7 +133,11 @@ def update_subscription(
                     for rule in metadata['replication_rules']:
                         validate_schema(name='activity', obj=rule.get('activity', 'default'), vo=vo)
             if 'state' in metadata and metadata['state'] is not None:
-                metadata['state'] = SubscriptionState(metadata['state'])
+                try:
+                    metadata['state'] = SubscriptionState(metadata['state'])
+                except ValueError as err:
+                    msg = f"Invalid subscription state: {metadata['state']}"
+                    raise InvalidObject(msg) from err
 
         except ValueError as error:
             raise TypeError(error)
--- a/tests/test_subscription.py
+++ b/tests/test_subscription.py
@@ -187,6 +187,34 @@ class TestSubscriptionCoreGateway:
         with pytest.raises(SubscriptionNotFound):
             update_subscription(name=subscription_name, account='root', metadata={'filter': {'project': ['toto', ]}}, issuer='root', vo=vo)
 
+    def test_update_subscription_invalid_state(self, vo, rse_factory):
+        """ SUBSCRIPTION (Gateway): Invalid subscription state should raise InvalidObject """
+        subscription_name = uuid()
+        rse1, _ = rse_factory.make_mock_rse()
+
+        add_subscription(
+            name=subscription_name,
+            account='root',
+            filter_={'project': self.projects},
+            replication_rules=[{'rse_expression': rse1, 'copies': 1, 'activity': self.activity}],
+            lifetime=100000,
+            retroactive=False,
+            dry_run=False,
+            comments='test',
+            issuer='root',
+            vo=vo
+        )
+
+        # try to update with an invalid state
+        with pytest.raises(InvalidObject):
+            update_subscription(
+                name=subscription_name,
+                account='root',
+                metadata={'state': 'NOT_A_REAL_STATE'},
+                issuer='root',
+                vo=vo
+            )
+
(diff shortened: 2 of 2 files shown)


voetberg:
Hi! Thanks for the contribution here. If you could also include a test documenting this change that would be great.

jainoshika:
Hi! Thanks for the review. I’ve added a gateway-level unit test to ensure that invalid `SubscriptionState` values in `update_subscription` raise a proper `InvalidObject` exception instead of propagating a raw `ValueError`.

jainoshika:
Hi @voetberg ! I added a gateway-level test that exercises `update_subscription` with an invalid state and asserts `InvalidObject` is raised.
However, Codecov still reports the new state-validation lines as uncovered.
Could you please let me know if there’s a preferred way to trigger this code path, or if I’m missing a prerequisite?

voetberg:
> Hi @voetberg ! I added a gateway-level test that exercises `update_subscription` with an invalid state and asserts `InvalidObject` is raised. However, Codecov still reports the new state-validation lines as uncovered. Could you please let me know if there’s a preferred way to trigger this code path, or if I’m missing a prerequisite?

Codecov only looks at tests/rucio/common. For this PR, I would ignore it. Your test is sufficient.

lib/rucio/gateway/subscription.py
```diff
@@ -133,7 +133,10 @@ def update_subscription(
                     for rule in metadata['replication_rules']:
                         validate_schema(name='activity', obj=rule.get('activity', 'default'), vo=vo)
             if 'state' in metadata and metadata['state'] is not None:
-                metadata['state'] = SubscriptionState(metadata['state'])
+                try:
+                    metadata['state'] = SubscriptionState(metadata['state'])
+                except ValueError as err:
+                    raise InvalidObject(f"Invalid subscription state: {metadata['state']}") from err
```
voetberg: To make the error message cleaner, please define the string before raising it as an error (avoids showing the error string without the f-string properly rendered). 

```suggestion
                    msg = f"Invalid subscription state: {metadata['state']}"
                    raise InvalidObject(msg) from err
```

jainoshika: Done, applied the suggested change. Thanks!

voetberg (approved):

dchristidis (changes requested):
Thank you for volunteering to work on this issue.

 1. Typically, we prefer not to merge a series of commits where one commit fixes an issue introduced in a former. To avoid this, kindly rebase your branch, squashing the third commit with the first in the process.
 2. Given the need for (1), it’s a good opportunity to also adhere more closely to the contribution guidelines regarding the commits messages. You can do this as part of the rebase.
    You have the option of using the existing guidelines, in which case the commit summaries should look like:
    `<component>: <change_message> #<issue number>`
    Alternatively, you can opt for conventional commits, in which case they will look like:
    `<type>(<scope>): <description>`
    And have the `Issue` trailer set appropriately.

jainoshika:
Thanks for the review @dchristidis.
I’ve rebased the branch and squashed the commits as suggested, and updated the commit message accordingly.
Please let me know if everything looks good now.

dchristidis:
Not quite. The summary of your first commit is:

```
gateway: validate SubscriptionState in update_subscription #7820
```

But there’s no ‘gateway’ in the [list of components](https://github.com/rucio/rucio/labels), which is linked in the [contribution guide](https://rucio.cern.ch/documentation/contributing/#4-commit-your-changes). As such, you have to replace it with ‘Subscriptions’ instead.

Your second commit follows a different convention than the first. Please choose one and apply it to both.

jainoshika:
Thank you for being patient with me. I’ve updated the component to 'Subscriptions' and applied the same commit message convention to both commits. Please let me know if everything looks good now.

dchristidis:
Very close. You only need to append the issue number to the second commit.

jainoshika:
Done, I’ve appended the issue number to the second commit as well. Thanks!

dchristidis (approved):
Well done.

PR #8323, opened 2026-01-09
Stop auto-forwarding of legacy-style CLI commands

### LINKED ISSUE #8294: Stop auto-forwarding of old-style CLI commands
### Description

Right now the CLI will run an old-style command with a warning. In the ongoing process of deprecating the old CLI, we should stop auto-forwarding and instead replace it with an exitcode 2 with a message to use a `--legacy` option and deprecation release.

### Motivation

_No response_

### Change

(open to workshopping on the actual wording): 

`> rucio list-rses`
`WARNING: rucio list-rses will be deprecated in version 4X.X.X. Use `--legacy` to auto-foward existing commands`
`rucio: error: 'list-rses' is not a valid command. Use `rucio rse list`.

voetberg:
Closes #8294 

Three notes: 
* PR is left as a draft with the discussion needed tag so we can hash out the exact wording on the warning message and the exact release number we'll pull the legacy version. 
* This PR is larger than the issues implies because tests had to be modified to avoid having all of the bin tests fail - I took liberties to change the majority of `test_bin_rucio` to use parametrized tests for both the legacy version and the current version of the CLI format, and combine in tests from `test_cli_structure` as they are largely over the same component. 
* The changes in the tests made some issues with the exception handler obvious, so changed where the `sys.exit` was called. This doesn't seem to have any unintended consequences but it is also worth discussion


voetberg:
Closing following offline discussion of changes to the CLI deprecation schedule. Will most likely be re-opened with changes to the approach in the future, but rebasing this would be more trouble than the effort rewriting it would take.

PR #8331, opened 2026-01-14
Testing: Remove obsolete autotest images

### LINKED ISSUE #8328: Remove old autotest image workflow and scripts
### Description

Since https://github.com/rucio/rucio/pull/7853 the Actions workflow `imagecache`, `tools/test/build_images.py` and `etc/docker/test/alma9.Dockerfile` are not used anymore and should be removed.

### Motivation

_No response_

### Change

_No response_

bziemons:
Since the introduction of the runtime dev images, the old autotest images do not need to be built anymore and all tooling using these can be removed.

The current code to build the runtime images can be found here: https://github.com/rucio/rucio/blob/a16f3ee2cb0908ec5b918376a5bfea07c5baf729/.github/workflows/runtime_images.yml#L59

Closes: #8328


tools/test/donkeyrider.py
Geogouz, line 1: What are we going to do with

https://github.com/bziemons/rucio/blob/b2c66aa3683c7d5fd59fd3442e52486fcc403124/tools/run_autotests.sh#L19

bziemons: remove it, probably missed it, thanks

bziemons:
this PR also needs adjustments to the documentation (contributing guide). See https://github.com/rucio/documentation/pull/718

tools/test/donkeyrider.py
Geogouz: Also something last, `tools/test/run_tests.py` still mentions `tools.test.build_images` and `donkeyrider` in `find_image` and `run_tests` docstrings. We should ideally update this and maybe reference the current mechanism.

bziemons: It is true that they are still mentioned, but I cannot make much sense out of the remaining mentions, because the donkeyrider never was the sole source for the parsed matrix, see autotest GH Actions worklflows.

Geogouz: Yes, the docstrings were accurate for the `donkeyrider.py` execution path, but not for the CI usage. 
Maybe something like?:

```
diff --git a/tools/test/run_tests.py b/tools/test/run_tests.py
index 583a434569c57e34945082ad0a42a22f6a257c67..bf76b5c33da2321529c9db67f7fd6fd8056280ff 100755
--- a/tools/test/run_tests.py
+++ b/tools/test/run_tests.py
@@ -145,91 +145,91 @@ def matches(small: dict, group: dict):
     Image metadata and matrix entries are represented as loose dictionaries. The helper
     isolates the "does this image satisfy the required characteristics?" check used by
     :func:`find_image`.
     """
     for key in small.keys():
         if key not in group or small[key] != group[key]:
             return False
     return True
 
 
 def stringify_dict(inp: dict):
     """
     Coerce mapping keys/values to strings for JSON/YAML consumption.
 
     The matrix originates from YAML where types may be integers or booleans; container
     environment variables expect strings, so we normalise to avoid differences between
     Python/YAML types later in the pipeline.
     """
     return {str(k): str(v) for k, v in inp.items()}
 
 
 def find_image(images: dict, case: dict):
     """
     Return the runtime image tag matching the matrix attributes.
 
-    ``images`` is the structure produced by :mod:`tools.test.build_images` (and forwarded
-    through ``donkeyrider``). Each entry describes the image in terms of distribution, Python
-    version, and optional identifiers. The metadata recorded for the image must be a subset
-    of the matrix case so that the case attributes satisfy the requirements encoded by the image.
+    ``images`` is a mapping of image tags to metadata derived from the JSON payload read in
+    :func:`main`. Each entry describes the image in terms of distribution, Python version, and
+    optional identifiers. The metadata recorded for the image must be a subset of the matrix
+    case so that the case attributes satisfy the requirements encoded by the image.
     """
     for image, idgroup in images.items():
         if matches(idgroup, case):
             return image
     raise RuntimeError("Could not find image for case " + str(case))
 
 
 def case_id(case: dict) -> str:
     """
     Generate a human-readable identifier from the matrix attributes.
 
     The resulting string is used for log file names and stderr prefixes, e.g.
     ``alma9-py3.9-client-postgres14``.
     """
     parts = [case["DIST"], 'py' + case["PYTHON"], case["SUITE"], case.get("RDBMS", "")]
     return '-'.join(filter(bool, parts))
 
 
 def case_log(caseid, msg, file=sys.stderr):
     """
     Print ``msg`` with the case identifier prefix for consistent logging.
     """
     print(caseid, msg, file=file, flush=True)
 
 
 def run_tests(cases: list, images: dict, tests: Optional[list[str]] = None):
     """
     Execute all matrix cases serially or in parallel.
 
     Parameters
     ----------
     cases:
         Case dictionaries produced by :mod:`tools.test.matrix_parser`. These contain env
         variables and flags such as ``RUN_HTTPD`` which choose the orchestration mode.
     images:
         Mapping used by :func:`find_image` to resolve cases to container images. When ``runtime_images``
-        are provided by :mod:`tools.test.donkeyrider`, they override the default build matrix.
+        are present in the JSON payload, they override the default ``images`` mapping.
     tests:
         Optional test selectors forwarded to ``tools/test/test.sh`` and, in turn, to ``tools/run_tests.sh``.
         When set, the runner drops to a filtered pytest invocation rather than executing the full suite.
 
     High-level behaviour
     --------------------
     * honours ``PARALLEL_AUTOTESTS`` (and related variables) to decide whether
       to spawn a :class:`multiprocessing.Pool`.
     * normalises each case environment to strings so it can be passed to the
       shell wrappers without surprises.
     * records stdout/stderr for each case in dedicated log files when running
       in parallel, making it possible to inspect failures after the workers exit.
     * surfaces worker failures via the ``sys.exit`` calls in :func:`run_case`.
       Unexpected exceptions caught by :func:`run_case_logger` cause the worker
       to return ``False`` so ``PARALLEL_AUTOTESTS_FAILFAST`` can stop the pool
       early; when fail-fast is disabled these errors are left to the logs for
       later inspection.
     """
     use_podman = 'USE_PODMAN' in os.environ and os.environ['USE_PODMAN'] == '1'
     parallel = 'PARALLEL_AUTOTESTS' in os.environ and os.environ['PARALLEL_AUTOTESTS'] == '1'
     failfast = 'PARALLEL_AUTOTESTS_FAILFAST' in os.environ and os.environ['PARALLEL_AUTOTESTS_FAILFAST'] == '1'
     copy_rucio_logs = 'COPY_AUTOTEST_LOGS' in os.environ and os.environ['COPY_AUTOTEST_LOGS'] == '1'
     logs_dir = pathlib.Path('.autotest')
     if parallel or copy_rucio_logs:
         logs_dir.mkdir(exist_ok=True)
```

Geogouz (approved):
Looks good to me - nice cleanup.

PR #8342, opened 2026-01-21
CI: Skip commitlint for `dependabot` in merge push

### LINKED ISSUE #8340: Commitlint does not skip on merge push for Dependabot PRs
Commitlint behaves inconsistently when merging Dependabot pull requests.

The root cause is that on merge, the `github.actor` is the merger not `dependabot[bot]`, even though the commits themselves are authored by Dependabot. This makes the CI fail as the commitlint does not skip for the commit. 

Instead, Using `github.event.head_commit.author.name` to fetch the author of the latest commit in the PR could help skip the commitlint CI check.

0xquark:
Fixes: #8340

### DIFF
--- a/.github/workflows/commitlint.yml
+++ b/.github/workflows/commitlint.yml
@@ -9,7 +9,7 @@ concurrency:
 jobs:
   commitlint:
     runs-on: ubuntu-latest
-    if: github.actor != 'dependabot[bot]'
+    if: github.actor != 'dependabot[bot]' && github.event.head_commit.author.name != 'dependabot[bot]'
     steps:
       - name: Checkout code
         uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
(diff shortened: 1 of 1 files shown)


.github/workflows/commitlint.yml
```diff
@@ -9,7 +9,7 @@ concurrency:
 jobs:
   commitlint:
     runs-on: ubuntu-latest
-    if: github.actor != 'dependabot[bot]'
+    if: github.actor != 'dependabot[bot]' && github.event.head_commit.author.name != 'dependabot[bot]'
```
dchristidis, line 12: Why are both needed?

dchristidis (commented):
Please also take the opportunity to remove the blank line at the end of the file.

.github/workflows/commitlint.yml
0xquark: `github.event.head_commit.author.name` is [only generated when the PR is Merged](https://github.com/orgs/community/discussions/26228), Because of that, we still need another way to run commitlint on the pull request itself, which uses `github.actor`.
So we check both the values. If one of them is missing, it’s treated as `null`, and the check will still run using the one that’s available.

0xquark:
> Please also take the opportunity to remove the blank line at the end of the file.

Isn't having an extra new line at the end considered a best practice?

dchristidis:
A newline yes, but not a blank line (i.e. two newlines).

0xquark:
> A newline yes, but not a blank line (i.e. two newlines).

I misread! Fixed

dchristidis (approved):

PR #8344, opened 2026-01-22
fix(API): Reject invalid GUID with proper error #8341

### LINKED ISSUE #8341: Unhandled exception when providing an invalid GUID
### Description

```python
>>> client.list_dids(scope='data18_13TeV', filters=[{'guid': 'foo'}], did_type='file')
  File "<python-input-17>", line 1, in <module>
    client.list_dids(scope='data18_13TeV', filters=[{'guid': 'foo'}], did_type='file')
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/var/home/dchristidis/.local/share/virtualenvs/rucio/lib64/python3.14/site-packages/rucio/client/didclient.py", line 90, in list_dids
    raise exc_cls(exc_msg)
rucio.common.exception.RucioException: An unknown exception occurred.
Details: no error information passed (http status code: 500)
```

Internally this causes a `ValueError: badly formed hexadecimal UUID string`, but it is not handled properly across the layers.

Similarly for:

```python
>>> client.get_dataset_by_guid(guid='foo')
  File "<python-input-24>", line 1, in <module>
    client.get_dataset_by_guid('foo')
    ~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^
  File "/var/home/dchristidis/.local/share/virtualenvs/rucio/lib64/python3.14/site-packages/rucio/client/didclient.py", line 916, in get_dataset_by_guid
    raise exc_cls(exc_msg)
rucio.common.exception.RucioException: An unknown exception occurred.
Details: no error information passed (http status code: 500)
```

### Steps to reproduce

See above.

### Rucio Version

_No response_

### Additional Information

_No response_

BitWiseBrain:
Closes: #8341
**Problem**:
When users provide malformed GUID strings to `get_dataset_by_guid()` or `list_dids()`, the API returns a generic 500 error with no useful information instead of a clear validation error.

**Changes**:
Added GUID validation using uuid.UUID() to check format before processing. Raises InvalidObject exception with a clear error message when the GUID is malformed

### DIFF
--- a/lib/rucio/gateway/did.py
+++ b/lib/rucio/gateway/did.py
@@ -15,6 +15,7 @@
 from copy import deepcopy
 from typing import TYPE_CHECKING, Any, Optional
 
+import uuid 
 import rucio.gateway.permission
 from rucio.common.constants import DEFAULT_VO, RESERVED_KEYS
 from rucio.common.exception import AccessDenied, InvalidObject, RucioException
@@ -65,11 +66,14 @@ def list_dids(
             or_group['scope'] = InternalScope(or_group['scope'], vo=vo)
 
     with db_session(DatabaseOperationType.READ) as session:
-        result = did.list_dids(scope=internal_scope, filters=filters, did_type=did_type, ignore_case=ignore_case,
-                               limit=limit, offset=offset, long=long, recursive=recursive, session=session)
+        try:
+            result = did.list_dids(scope=internal_scope, filters=filters, did_type=did_type, ignore_case=ignore_case,
+                                limit=limit, offset=offset, long=long, recursive=recursive, session=session)
 
-        for d in result:
-            yield gateway_update_return_dict(d, session=session)
+            for d in result:
+                yield gateway_update_return_dict(d, session=session)
+        except ValueError as e:
+            raise InvalidObject(f"Invalid GUID provided in filters. Details: {str(e)}")
 
 
 def add_did(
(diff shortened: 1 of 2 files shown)


bziemons (changes requested):
Instead of adding a parsing/checking step on the gateway-level, I think it would be nicer if we could properly handle the exception that was mentioned in the issue and convert it to a useful error message for a user.

Also please note that your first commit also contains a test that could be instead moved to the second commit and the second commit's type should be `test` as per "Choosing the Right Type" in the [contributing guide](https://rucio.cern.ch/documentation/contributing/). Also please refine the commit message of the second commit to reflect this change.

BitWiseBrain:
Hi! ive made changes as requested, i removed the parsing/ checking at the gateway level and instead handelled the error by raising an InvalidObject explaing the error with an error message, ive also changed the commits and commit messages to reflect the changes made.
Could you please re-review the changes made

tests/test_did.py
```diff
@@ -354,6 +354,10 @@ def test_update_new_dids(self, vo):
         with pytest.raises(DataIdentifierNotFound):
             did.set_new_dids([{'scope': 'dummyscope', 'name': 'dummyname', 'did_type': DIDType.DATASET}], None, vo=vo)
 
+    def test_get_dataset_by_guid_invalid_guid(self, vo):
```
bziemons, line 357: please move adding the test to the other commit

lib/rucio/gateway/did.py
```diff
@@ -15,6 +15,7 @@
 from copy import deepcopy
 from typing import TYPE_CHECKING, Any, Optional
 
+import uuid
```
bziemons, line 18: leftover import

bziemons (changes requested):
I am pretty sure that catching `ValueError` this high up and always assuming it must be the GUID in the wrong format is error-prone when it comes to different error sources.
I do not have a good alternative in my head right now, but have you tried the command from the issue and looked at the server stack trace to identify the code path which leads to the error?

dchristidis:

PR #8346, opened 2026-01-22
refactor(Clients): Copy implimentation of did commands to new format

### LINKED ISSUE #8298: Copy old CLI implimentation of *did* to new CLI

voetberg:
Closes #8298

### DIFF
--- a/lib/rucio/cli/did.py
+++ b/lib/rucio/cli/did.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import math
+from typing import Literal, Optional, Union
 
 import click
 from rich.text import Text
@@ -25,7 +26,7 @@ from rucio.common.utils import chunks, parse_did_filter_from_string_fe
 
 
 @click.group()
-def did():
+def did() -> None:
     """Manage Data Identifiers - the source data objects"""
 
 
@@ -45,7 +46,7 @@ def did():
 @click.argument("did-pattern", nargs=1, required=True)
 @click.option("--parent", default=False, is_flag=True, help="List the parents of the DID - must use a full DID scope and name")
 @click.pass_context
-def list_(ctx, did_pattern, recursive, filter_, short, parent):
+def list_(ctx: click.Context, did_pattern: str, recursive: bool, filter_: str, short: bool, parent: bool) -> None:
     """
     List the Data IDentifiers matching certain pattern.
     Only the collections (i.e. dataset or container) are returned by default.
@@ -119,7 +120,7 @@ def list_(ctx, did_pattern, recursive, filter_, short, parent):
 @did.command("show")
 @click.argument("dids", nargs=-1)
 @click.pass_context
-def show(ctx, dids):
+def show(ctx: click.Context, dids: tuple[str, ...]) -> None:
     """List attributes, statuses, or parents for data identifiers"""
     if ctx.obj.use_rich:
         ctx.obj.spinner.update(status='Fetching DID stats')
@@ -155,7 +156,7 @@ def show(ctx, dids):
 @click.option("--monotonic", is_flag=True, default=False, help="Monotonic status to True.")
 @click.option("--lifetime", type=int, help="Lifetime in seconds.")
 @click.pass_context
-def add_(ctx, did_name, dtype, monotonic, lifetime):
+def add_(ctx: click.Context, did_name: str, dtype: Literal['container', 'dataset'], monotonic: bool, lifetime: Optional[int]) -> None:
     """Create a new collection-type DID"""
     scope, name = get_scope(did_name, ctx.obj.client)
     if dtype == "container":
@@ -171,7 +172,7 @@ def add_(ctx, did_name, dtype, monotonic, lifetime):
 @click.option("--open", "operation", flag_value="open", help="Reopen a dataset or container (only for privileged users)")
 @click.option("--close", "operation", flag_value="close", help="Close a dataset or container.")
 @click.pass_context
-def update(ctx, dids, rse, operation):
+def update(ctx: click.Context, dids: tuple[str, ...], rse: Optional[str], operation: Literal['touch', 'open', 'close']) -> None:
     """Touch one or more DIDs and set the last accessed date to the current date, or mark them as open or closed."""
 
     if operation == "touch":
@@ -197,7 +198,7 @@ def update(ctx, dids, rse, operation):
 @click.option("--undo", is_flag=True, default=False, help="Undo erase DIDs. Only works if has been less than 24 hours since erase operation.")
 @click.argument("dids", nargs=-1)
 @click.pass_context
-def remove(ctx, dids, undo):
+def remove(ctx: click.Context, dids: tuple[str, ...], undo: bool) -> None:
     """
     This command sets the lifetime of the DID in order to expire in the next 24 hours.
     Expired DIDs are force-deleted (and their replicas purged).
@@ -232,14 +233,14 @@ def remove(ctx, dids, undo):
                 ctx.obj.logger.debug('Error: %s' % error)
 
 @did.group()
-def content():
+def content() -> None:
     """Manage contents of collection type DIDs"""
 
 
 @content.command("history")
 @click.argument("dids", nargs=-1, required=True)
 @click.pass_context
-def content_history(ctx, dids):
+def content_history(ctx: click.Context, dids: tuple[str, ...]) -> None:
     """List the content history of a collection-type DID"""
     table_data = []
     if ctx.obj.use_rich:
@@ -267,7 +268,7 @@ def content_history(ctx, dids):
 @click.option("-f", "--from-file", is_flag=True, default=False, help="[DIDs] is a file instead of a list of did names. The file should contain one did per line.")
 @click.argument("dids", nargs=-1)
 @click.pass_context
-def content_add_(ctx, to_did, from_file, dids):
+def content_add_(ctx: click.Context, to_did: str, from_file: bool, dids: tuple[str, ...]) -> None:
     """Attach a list [dids] of data identifiers (file or collection-type) to another data identifier (collection-type)"""
     scope, name = get_scope(to_did, ctx.obj.client)
     limit = 499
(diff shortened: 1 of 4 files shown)


voetberg:
Ruff errors are: 

```

Run ruff check --output-format=github .
Error: bin/rucio:125:1: W293 Blank line contains whitespace
Error: bin/rucio:145:16: E713 Test for membership should be `not in`
Error: bin/rucio:149:1: W293 Blank line contains whitespace
Error: bin/rucio:157:1: W293 Blank line contains whitespace
Error: bin/rucio:161:1: W293 Blank line contains whitespace
Error: bin/rucio:161:5: W292 No newline at end of file
Error: Process completed with exit code 1.
```
All of which were already present so I will not change them in this PR

lib/rucio/cli/did.py
```diff
@@ -38,26 +46,110 @@ def did():
 @click.argument("did-pattern", nargs=-1, required=True)
 @click.option("--parent", default=False, is_flag=True, help="List the parents of the DID - must use a full DID scope and name")
 @click.pass_context
-def list_(ctx, did_pattern, recursive, filter_, short, parent):
+def list_(ctx: click.Context, did_pattern: str, recursive: bool, filter_: str, short: bool, parent: bool):
```
bziemons: click.argument with nargs=-1 gives a `tuple[str, ...]` not `str`, see https://click.palletsprojects.com/en/stable/arguments/#multiple-arguments

lib/rucio/cli/did.py
```diff
@@ -38,26 +46,110 @@ def did():
...
     if parent:
-        for did in did_pattern:
-            list_parent_dids(Arguments({"no_pager": ctx.obj.no_pager, "did": did}), ctx.obj.client, ctx.obj.logger, ctx.obj.console, ctx.obj.spinner)
+        if ctx.obj.use_rich:
+            ctx.obj.spinner.update(status='Fetching parent DIDs')
+            ctx.obj.spinner.start()
+
+        if did_pattern:
+            table_data = []
+            scope, name = get_scope(did_pattern[0], ctx.obj.client)  # nargs=-1 results in an iterator, we only need the first (only) obj
```
bziemons: why do we only need the first object and why is it deemed the only object? Wouldn't nargs be 1 in that case?

Also this is different from the previous code, which iterated over all arguments: `for did in did_pattern:` – Why?

bziemons (changes requested):
Partial review. Please also check other type annotations where nargs=-1

lib/rucio/cli/did.py
voetberg: I had the majority given as list, but you are correct, tuple[str, ...] is more accurate

lib/rucio/cli/did.py
voetberg: I am honestly not sure, this may be a hold out from the original `list_dids` and `list_did_parents` conflict - where `list_dids` had nargs=1 and `list_did_parents` had nargs='?', though I believe this was also originally misguided because the default is just `None`. 

I think the original iterator was a mistake and `did-pattern` should just have nargs=1. I will make this a separate commit for clarity.

bziemons (changes requested):
I hate to be strict, but currently 7feeeacf1f0ec7405155581aed1ddb7a96618648 is the commit that introduces the breaking change:

> BREAKING CHANGE: User can now only supply one did-pattern to `rucio did list`

Can you rebase 4f474609c9050e5db5ab35373460052abbcba0e1 before that commit? That would fix it and 7feeeacf1f0ec7405155581aed1ddb7a96618648 would really be a refactor/copy then.

lib/rucio/cli/did.py
```diff
@@ -45,8 +45,7 @@ def list_(ctx, did_pattern, recursive, filter_, short, parent):
     With the filter option, you can specify a list of metadata that the Data IDentifier should match
     """
     if parent:
-        for did in did_pattern:
-            list_parent_dids(Arguments({"no_pager": ctx.obj.no_pager, "did": did}), ctx.obj.client, ctx.obj.logger, ctx.obj.console, ctx.obj.spinner)
+        list_parent_dids(Arguments({"no_pager": ctx.obj.no_pager, "did": did}), ctx.obj.client, ctx.obj.logger, ctx.obj.console, ctx.obj.spinner)
```
bziemons: commit 1ef4b2bd4b00c73dd898f37c152c00e5dfba4618 the parameter name is `did_pattern`, not `did`

lib/rucio/cli/did.py
```diff
@@ -268,29 +269,31 @@ def content_history(ctx, dids):
...
+    did_objs = [{'scope': get_scope(did, ctx.obj.client)[0], 'name': get_scope(did, ctx.obj.client)[1]} for did in dids_list]
+    if len(did_objs) <= limit:
+        ctx.obj.client.attach_dids(scope=scope, name=name, dids=did_objs)
     else:
         ctx.obj.logger.warning("You are trying to attach too much DIDs. Therefore they will be chunked and attached in multiple commands.")
         missing_dids = []
-        for i, chunk in enumerate(chunks(dids, limit)):
-            ctx.obj.logger.info("Try to attach chunk {0}/{1}".format(i, int(math.ceil(float(len(dids)) / float(limit)))))
+        for i, chunk in enumerate(chunks(did_objs, limit)):
+            ctx.obj.logger.info("Try to attach chunk {0}/{1}".format(i, int(math.ceil(float(len(did_objs)) / float(limit)))))
```
bziemons, line 295: can you uncouple these change from the commit which is adding type annotations?

lib/rucio/cli/did.py
voetberg: Corrected

lib/rucio/cli/did.py
```diff
@@ -104,8 +208,33 @@ def remove(ctx, dids, undo):
...
     """
-    erase(Arguments({"no_pager": ctx.obj.no_pager, "dids": dids, "undo": undo}), ctx.obj.client, ctx.obj.logger, ctx.obj.console, ctx.obj.spinner)
+    for did in dids:
+        if '*' in did:
+            ctx.obj.logger.warning("This command doesn't support wildcards! Skipping DID: %s" % did)
+            continue
+        try:
+            scope, name = get_scope(did, ctx.obj.client)
+        except RucioException as error:
+            ctx.obj.warning('DID is in wrong format: %s' % did)
```
bziemons: this should probably be `ctx.obj.logger.warning` ?

bziemons (changes requested):

lib/rucio/cli/did.py
voetberg: These are tied to the typing change - if they are not changed they obscure the typing and would produce a new typing error

lib/rucio/cli/did.py
voetberg: Correct.....

lib/rucio/cli/did.py
bziemons: okay, I am fine with them then.

bziemons (dismissed):

lib/rucio/cli/command.py
dchristidis: https://github.com/rucio/rucio/blob/e9a5a52fbb5e5b9439eeaea57e691ba08756e68e/lib/rucio/cli/command.py#L206

Isn’t this already done?

lib/rucio/cli/did.py
```diff
@@ -35,7 +35,7 @@ def did():
     """,
 )  # TODO Shorten this help and make supplying this easier
 @click.option("--short", is_flag=True, default=False, help="Just dump the list of DIDs.")
-@click.argument("did-pattern", nargs=-1, required=True)
+@click.argument("did-pattern", nargs=1, required=True)
```
dchristidis, line 46: Just to be sure, this means that the entire set of changes won’t be applied on the release branch. Is this fine with you or would you rather split this particular commit in a separate PR? The potential risk when the master and the release branch have diverged significantly is that a future PR may have to be opened twice, once for each branch (adapted accordingly).

lib/rucio/cli/did.py
```diff
@@ -91,10 +91,9 @@ def list_(ctx, did_pattern, recursive, filter_, short, parent):
 
         if recursive and '*' in name:
             raise InputValidationError('Option recursive cannot be used with wildcards.')
-        # TODO Modify statement, filters can never have "name" as it is not modified until this point
         else:
-            if filters:
-                if ('name' in filters) and (name != '*'):
+            if filter_:
+                if ('name' in filter_) and (name != '*'):
```
dchristidis, line 93: Given this, is it worth to keep the `filters = {}` assignment above?

lib/rucio/cli/did.py
```diff
@@ -45,7 +46,7 @@ def did():
 @click.argument("did-pattern", nargs=1, required=True)
 @click.option("--parent", default=False, is_flag=True, help="List the parents of the DID - must use a full DID scope and name")
 @click.pass_context
-def list_(ctx, did_pattern, recursive, filter_, short, parent):
+def list_(ctx: click.Context, did_pattern: str, recursive: bool, filter_: str, short: bool, parent: bool):
```
dchristidis: Reminder that not hinting the return type is a missed opportunity. Unless I am mistaken and you believe that it is inconsequential?

lib/rucio/cli/did.py
```diff
@@ -268,29 +269,31 @@ def content_history(ctx, dids):
...
 @click.pass_context
-def content_add_(ctx, to_did, from_file, dids):
+def content_add_(ctx: click.Context, to_did: str, from_file: bool, dids: tuple[str, ...]):
     """Attach a list [dids] of data identifiers (file or collection-type) to another data identifier (collection-type)"""
     scope, name = get_scope(to_did, ctx.obj.client)
     limit = 499
 
     if from_file:
-        if len(dids) > 1:
+        if len(dids) != 1:
```
dchristidis: I think this particular change is not stylistic.

lib/rucio/cli/did.py
```diff
@@ -268,29 +269,31 @@ def content_history(ctx, dids):
...
             raise ValueError('If --from-file option is active, only one file is supported. The file should contain a list of dids, one per line.')
         try:
             f = open(dids[0], 'r')
-            dids = [did.rstrip() for did in f.readlines()]
+            dids_list = [did.rstrip() for did in f.readlines()]
         except OSError as error:
             ctx.obj.logger.error("Can't open file '" + dids[0] + "'.")
             raise OSError from error
+    else:
+        dids_list = list(dids)
```
dchristidis, line 286: Maybe have a quick mention in the commit message that this and the variable renames around it were done for typing reasons?

dchristidis (commented):
Very well done. This PR was effortless to review. Thanks!

lib/rucio/cli/command.py
voetberg: Sure is, I'll pop that commit off the stack. Had it in multiple CLI PRs because I didn't know which one would get merged first.

lib/rucio/cli/did.py
voetberg: It probably does not get caught in anything but nothing wrong with adding it just in cases it becomes useful down the line. Good catch.

lib/rucio/cli/did.py
voetberg: Added as its own commit

lib/rucio/cli/did.py
voetberg: Probably not - addressed

dchristidis (approved):

Issue #8195, opened 2025-11-07
Account limits: `get_global_account_usage` and `get_local_account_usage` are read-only operations, but are passed a `transactional_session`


rdimaio:
### Description

https://github.com/rucio/rucio/blob/b36113b21f2847d5f715e4bc38a8cf2f25dbe792/lib/rucio/core/account_limit.py#L301-L403

### Steps to reproduce

NA

Issue #7267, opened 2024-12-16
Type annotation tracker


rdimaio:
Follow-up from #7021 for general tracking of remaining type annotations

rdimaio:
For baseclient, reference PR https://github.com/rucio/rucio/pull/7019

PR #8350, opened 2026-01-24
Core: account and account_limit, migrate away from session decorators, pass session as required parameter, fix type annotations

rdimaio:
Part of:

- #6989
- #7267

Fixes:

- #8195


rdimaio:
Rebased in latest force push

bari12 (approved):

PR #8351, opened 2026-01-25
ci(Testing): run lint hooks via pre-commit in CI

### LINKED ISSUE #7321: Consider pre-commit.ci
https://pre-commit.ci/

As an example of how it's used by DiracX: https://github.com/DIRACGrid/diracx/commits?author=pre-commit-ci%5Bbot%5D

### LINKED ISSUE #8245: Consolidate linting checks: Run pre-commit in CI to eliminate duplication
**Problem:** We have duplicate linting logic in `.pre-commit-config.yaml` and `.github/workflows/autotest.yml` (ruff,  add_header). This creates maintenance overhead and divergence risk.

**Proposal:** Replace individual CI jobs with a single `pre-commit run --all-files` command.

### Changes:

**`.pre-commit-config.yaml`:**
- Keep all existing hooks (ruff, add_header, etc.)

**`autotest.yml`:**
- Replace `add_header` and `python_ruff` jobs with single `lint` job that runs pre-commit
- Net reduction: ~15 lines

### Benefits:
- Single source of truth
- Identical local/CI validation  
- Easier maintenance
- Future linters require only few lines in pre-commit vs ~15 lines in CI

**Out of scope:** `python_annotations`, `python_pyright`, `test` jobs remain separate (require git history comparison or are too heavy for pre-commit).

**Derived from:** PR #8055

Geogouz:
- Consolidate header and Ruff checks into a single pre-commit run in the autotest workflow.
- Disable the add_header progress bar for cleaner CI logs.

Closes: #8245

I think something like this should be fine to have local pre-commit and CI in sync. We could maybe even go on and close #7321 unless we have good reasons to go that way (personally I am in favour of keeping things simpler for the moment).

Keep in mind that with these changes, our CI right now reflects what is happening locally (where there are auto-fixes). But the way we have things setupped now on our CI, I don't see for the moment some problem there. We just need to remember that in case someone did not use pre-commits locally, there is a change that changes are made in the ephemeral workspace for those corresponding jobs. If we need to stick to no --fix, we could add this logic into the `autotest.yml`.

### DIFF
--- a/.github/workflows/autotest.yml
+++ b/.github/workflows/autotest.yml
@@ -11,33 +11,25 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  add_header:
+  pre_commit:
     if: github.repository_owner == 'rucio' || github.event_name != 'schedule'
-    name: Add header lint
+    name: Lint (pre-commit)
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5.0.1
         with:
-          # We need the history to determine the file creation date.
           fetch-depth: 0
-      - name: Check headers
-        shell: bash
-        run: |
-          python3 tools/add_header --dry-run --disable-progress-bar
-  python_ruff:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5.0.1
       - name: Install Python
         uses: actions/setup-python@e797f83bcb11b83ae66e0230d6156d7c80228e7c # v6.0.0
         with:
           python-version: "3.11"
-      - name: Install dependencies
+      - name: Install pre-commit
         run: |
           python -m pip install --upgrade pip
-          pip install ruff
-      - name: Run Ruff
-        run: ruff check --output-format=github .
+          python -m pip install pre-commit
+      - name: Run pre-commit
+        run: |
+          pre-commit run --all-files --show-diff-on-failure
   python_annotations:
     if: |
       !startsWith(github.ref_name, 'release')
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,6 +11,8 @@ repos:
     description: Run the `add_header` script to verify the file headers
     entry: tools/add_header
     language: system
+    args: [--disable-progress-bar]
+    pass_filenames: false
 - repo: https://github.com/astral-sh/ruff-pre-commit
   rev: v0.14.9
   hooks:


.pre-commit-config.yaml
```diff
@@ -11,6 +11,7 @@ repos:
     description: Run the `add_header` script to verify the file headers
     entry: tools/add_header
     language: system
+    args: [--disable-progress-bar]
```
mgajek-cern, line 14: The add_header script uses git ls-files internally to discover files . By default, pre-commit passes filenames to hooks, which could interfere with the script's file discovery logic.

```suggestion
    args: [--dry-run, --disable-progress-bar]
    pass_filenames: false 
```

This makes the behavior explicit and prevents potential issues if the script doesn't handle passed filenames correctly.

mgajek-cern (commented):
Some suggestions

Re #7321: Would implement this in a separate PR

.github/workflows/autotest.yml
```diff
@@ -11,33 +11,25 @@ concurrency:
...
+      - name: Install pre-commit
         run: |
           python -m pip install --upgrade pip
-          pip install ruff
-      - name: Run Ruff
-        run: ruff check --output-format=github .
+          python -m pip install pre-commit
+      - name: Run pre-commit
+        run: |
+          pre-commit run --all-files --show-diff-on-failure
```
mgajek-cern, line 32: All current hooks correctly fail when making changes, so a git diff check isn't needed. Consider adding one if we integrate external hooks in the future.

```suggestion
          pre-commit run --all-files --show-diff-on-failure
          # NOTE: 'git diff --exit-code' check could be added here as defense-in-depth
          # to catch hooks that silently modify files. Currently not needed as all hooks
          # correctly fail (exit 1) when they make changes.
```

.pre-commit-config.yaml
Geogouz: Using `--dry-run` is quite a big behavioural change for the local checks (which are currently being corrected automatically and the user doesn't have to bother what header to place and where).
Although there is no practical changes with the use of `pass_filenames: false` (this is what is happening by default currently because we do not pass something), I don't see problems using it.

mgajek-cern: Let's remove the --dry-run flag then to preserve the local auto-fix behavior. 

Quick clarification on pass_filenames: it actually defaults to true in pre-commit, so files are being passed by default. See: https://pre-commit.com. Without pass_filenames: false, pre-commit passes only changed/staged files to the script, which could interfere with the script's internal git ls-files logic that's designed to check all repository files. Adding pass_filenames: false ensures the script always uses its own file discovery mechanism (as intended by its design), so it's a good addition for correctness even if it appears to work currently.

---

Optional: Future enhancement

We could consider inverting the script's behavior to match ruff: check-only by default, `--fix` for auto-corrections. This would eliminate the `--dry-run` flag and make local/CI configs identical. Simple change, potential UX improvement. Not blocking this PR though. Issue created: https://github.com/rucio/rucio/issues/8353

Geogouz: What I mean is that, since we were not passing any file arguments in `python3 tools/add_header --dry-run --disable-progress-bar`, the add_header (being not invoked as a pre-commit hook but rather manually) logic on the CI was falling back to repository discovery by default. Now with this PR, the CI is in sync with the local pre-commits, and is considering only the changed files (which I made the assumption that was enough). But practically, even if we scan everything it should be fine. Basically we agree.

Geogouz: Added it

mgajek-cern (approved):

PR #8367, opened 2026-01-29
feat(Subscriptions): Enhance subscription rule states endpoint to include subscription metadata

### LINKED ISSUE #8343: Enhance subscription rule states endpoint to include subscription metadata
### Description

Currently, the /subscriptions/{account}/rules/states endpoint only returns  aggregated rule counts per state for each subscription:                       
   ```                                                                             
  {                                                                             
    "account": "root",                                                          
    "name": "my_subscription",                                                  
    "state": "O",                                                               
    "count": 123                                                                
  }   
```

  This issue proposes enhancing the endpoint to optionally include subscription metadata (state, last_processed, lifetime, etc.) to provide better operational visibility without requiring additional API calls.                           
                                                                                
  Additionally, this issue addresses the **current limitation where subscriptions with zero rules are not returned** due to the INNER JOIN between subscriptions  and rules tables. 

### Motivation

This endpoint is used in the WebUI to display the list of subscriptions.
 
### Problem 1: Missing Subscription Context                                       
                                                                                
  In the WebUI, we need to know:                              
  - Is the subscription itself active? (state: ACTIVE/INACTIVE/BROKEN)          
  - When was it last processed? (last_processed: helps identify stuck           
  subscriptions)                                                                
  - Will it expire soon? (lifetime, expired_at: proactive monitoring)           
  - ...
   
Please see https://github.com/rucio/webui/issues/534 as an example            
                                                                                
  Currently, this requires a second API call to /subscriptions/{account}/{name} 
  for each subscription, **causing N+1 query problems** in the UI.    
                                                                                
  ### Problem 2: Subscriptions Without Rules Are Hidden                             
                                                                                
  The current query uses INNER JOIN, so subscriptions without any rules don't   
  appear in the response. This is problematic for:                              
  - Newly created subscriptions before the transmogrifier daemon runs           
  - Subscriptions that haven't matched any DIDs yet                             
  - Debugging subscription configuration issues                                 
                                                                                

### Option 1: Query Parameter                                 
                                                                                
  Add optional `?include_details=true` query parameter to existing endpoint:      
                                                                                
  Endpoint: `GET /subscriptions/{account}/rules/states?include_details=true`
                                                                                
  New Response:                                                            
  ```
{                                                                             
    "account": "root",                                                          
    "name": "my_subscription",                                                  
    "state": "O",                                                               
    "count": 123,                                                               
    "details": {                                                                
      "sub_state": "ACTIVE",                                                    
      "last_processed": "2025-01-21T10:30:00Z",                                 
      "lifetime": "2025-12-31T23:59:59Z",                                       
      "expired_at": null,                                                       
      "comments": "Production subscription for ATLAS data"                      
    }
                                                                 
  } 
```                                                                           

This should be fully backwards compatible and clients can opt-in as needed.                                                                              
  ### Option 2: New Endpoint (Alternative)                                          
                                                                                
  Create separate detailed endpoint:                                            
                                                                                
  Endpoint: `GET /subscriptions/{account}/rules/states/detailed  `                
              
This will also work, however, this leads to code duplication                                        

### Code Changes
This section is not fully thought though but should offer some insights on the changes needed based on my initial digging around:

In the  `lib/rucio/core/subscription.py`, for the method: `list_subscription_rule_states()` (lines 282-338)         

Change INNER JOIN to LEFT OUTER JOIN
Before: 
``` 
 ).join(                              # Before                      
      rule,                                                          
      subscription.id == rule.subscription_id                        
  )                                                                  
   ```

After
```                                                                  
  ).outerjoin(                         # After                       
      rule,                                                          
      subscription.id == rule.subscription_id                        
  )                                                                  
   ``` 
And fix the count function to handle NULL properly  i.e change    `func.count().label('count')` to                      
  `func.count(rule.id).label('count')` 
Then test if this returns 0 for  subscriptions with no rules                                        
                                                                     
Also add subscription details (When include_details=True) to the query:   
```  
columns.extend([                                           
              subscription.state.label('sub_state'),                 
              subscription.last_processed,                           
              subscription.lifetime,                                 
              subscription.expired_at,                               
              subscription.comments                                  
          ])  
```

cserf:
Closes: #8343

### DIFF
--- a/lib/rucio/gateway/subscription.py
+++ b/lib/rucio/gateway/subscription.py
@@ -201,14 +201,16 @@ def list_subscriptions(
 def list_subscription_rule_states(
     name: Optional[str] = None,
     account: Optional[str] = None,
+    include_details: bool = False,
     vo: str = DEFAULT_VO,
-) -> 'Iterator[SubscriptionRuleState]':
+) -> 'Iterator[Union[SubscriptionRuleState, dict[str, Any]]]':
     """Returns a list of with the number of rules per state for a subscription.
 
     :param name: Name of the subscription
     :param account: Account identifier
+    :param include_details: Include subscription metadata (state, last_processed, lifetime, expired_at, comments)
     :param vo: The VO to act on.
-    :returns: Sequence with SubscriptionRuleState named tuples (account, name, state, count)
+    :returns: Iterator of SubscriptionRuleState namedtuples (default) or dicts (when include_details=true)
     """
     if account is not None:
         internal_account = InternalAccount(account, vo=vo)
@@ -216,13 +218,25 @@ def list_subscription_rule_states(
         internal_account = InternalAccount('*', vo=vo)
 
     with db_session(DatabaseOperationType.READ) as session:
-        subs = subscription.list_subscription_rule_states(name, internal_account, session=session)
+        subs = subscription.list_subscription_rule_states(name, internal_account, include_details=include_details, session=session)
 
         for sub in subs:
-            # sub is an immutable Row so return new named tuple with edited entries
-            d = sub._asdict()
+            # sub is a dict from core layer
+            d = dict(sub)
+            # Externalize the internal account
             d['account'] = d['account'].external
-            yield SubscriptionRuleState(**d)
+
+            if include_details:
+                # Return dict with all details
+                yield d
+            else:
+                # Backward compatible: return namedtuple with only basic fields
+                yield SubscriptionRuleState(
+                    account=d['account'],
+                    name=d['name'],
+                    state=d['state'],
+                    count=d['count']
+                )
 
 
 def delete_subscription(
--- a/lib/rucio/web/rest/flaskapi/v1/subscriptions.py
+++ b/lib/rucio/web/rest/flaskapi/v1/subscriptions.py
@@ -498,6 +498,13 @@ class States(ErrorHandlingMethodView):
           schema:
             type: string
           style: simple
+        - name: include_details
+          in: query
+          description: "Include subscription metadata (state, last_processed, lifetime, expired_at, comments)."
+          schema:
+            type: boolean
+            default: false
+          style: form
         responses:
           200:
             description: "OK"
@@ -517,10 +524,31 @@ class States(ErrorHandlingMethodView):
                         type: string
                       state:
                         description: "The state of the rules."
-                        $ref: "#/components/schemas/RuleState"
+                        type: string
+                        nullable: true
+                        enum: ["REPLICATING", "OK", "STUCK", "SUSPENDED", "WAITING_APPROVAL", "INJECT"]
                       count:
                         description: "The number of rules with that state."
                         type: integer
+                      subscription_state:
+                        description: "The state of the subscription (only when include_details=true)."
+                        type: string
+                        enum: ["ACTIVE", "INACTIVE", "NEW", "UPDATED", "BROKEN"]
+                      last_processed:
+                        description: "Last processing timestamp (only when include_details=true)."
+                        type: string
+                        format: date-time
+                      lifetime:
+                        description: "Subscription lifetime (only when include_details=true)."
+                        type: string
+                        format: date-time
+                      expired_at:
+                        description: "Expiration timestamp (only when include_details=true)."
+                        type: string
+                        format: date-time
+                      comments:
+                        description: "Subscription comments (only when include_details=true)."
+                        type: string
           401:
             description: "Invalid Auth Token"
           404:
@@ -528,8 +556,10 @@ class States(ErrorHandlingMethodView):
           406:
             description: "Not acceptable"
         """
+        include_details = param_get_bool(request.args, 'include_details', default=False)
+
         def generate(vo):
-            for row in list_subscription_rule_states(name=name, account=account, vo=vo):
+            for row in list_subscription_rule_states(name=name, account=account, include_details=include_details, vo=vo):
                 yield dumps(row, cls=APIEncoder) + '\n'
 
         return try_stream(generate(vo=request.environ['vo']))
--- a/lib/rucio/client/subscriptionclient.py
+++ b/lib/rucio/client/subscriptionclient.py
@@ -235,3 +235,53 @@ class SubscriptionClient(BaseClient):
         else:
             exc_cls, exc_msg = self._get_exception(headers=result.headers, status_code=result.status_code, data=result.content)
             raise exc_cls(exc_msg)
+
+    def list_subscription_rule_states(
+            self,
+            account: str,
+            name: Optional[str] = None,
+            include_details: bool = False
+    ) -> "Iterator[dict[str, Any]]":
+        """
+        List the aggregated rule states for subscriptions.
+
+        Parameters
+        ----------
+        account :
+            Account identifier
+        name :
+            Name of the subscription (optional)
+        include_details :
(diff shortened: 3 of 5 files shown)


lib/rucio/gateway/subscription.py
```diff
@@ -201,28 +201,50 @@ def list_subscriptions(
...
+            # sub is a dict from core layer
+            d = dict(sub)
+            # Externalize the internal account
             d['account'] = d['account'].external
-            yield SubscriptionRuleState(**d)
+
+            # Convert enums to string values
+            # rule.state can be NULL when subscription has no rules (LEFT OUTER JOIN)
+            if d.get('state') is not None:
+                d['state'] = d['state'].value
```
maany: I believe this is a breaking change.  This runs for both code paths (with and without include_details). Previously, the RuleState enum (e.g., RuleState.REPLICATING) was passed as-is into the SubscriptionRuleState namedtuple, and APIEncoder serialized it via .name → "REPLICATING". Now it's pre-converted via .value → "R", changing the REST API output even on the default path.

maany (changes requested):
Thanks @cserf. I just noticed a potential breaking change, otherwise the code looks good to me.

maany:
Hey @cserf, just pinging to see if you can wrap this up in the coming days. I wanted to build some WebUI views around it. Thanks!

maany:
@cserf gentle ping on this one.

lib/rucio/gateway/subscription.py
cserf: I removed the conversion

lib/rucio/gateway/subscription.py
```diff
@@ -201,28 +201,47 @@ def list_subscriptions(
...
+            # sub is a dict from core layer
+            d = dict(sub)
+            # Externalize the internal account
             d['account'] = d['account'].external
-            yield SubscriptionRuleState(**d)
+
+            # Convert enums to string values
+            # subscription_state is NOT NULL in database, always present when include_details=True
+            if 'subscription_state' in d:
+                d['subscription_state'] = d['subscription_state'].value
```
maany: Hey @cserf, I believe this would produce single character values like "A", "I" etc. However, state (RuleState) will be converted by APIEncoder to "Active", "Inactive". This would create an inconsistent format on the  REST output where, state: "Replicating" and subscription_state is "A"

Could you remove the .value conversion here and let the APIEncoder handle the serialization consistently.

lib/rucio/web/rest/flaskapi/v1/subscriptions.py
```diff
@@ -524,15 +531,36 @@ def get(self, account, name=None):
                       count:
                         description: "The number of rules with that state."
                         type: integer
+                      subscription_state:
+                        description: "The state of the subscription (only when include_details=true)."
+                        type: string
+                        enum: ["A", "I", "N", "U", "B"]
```
maany: This would also need to be updated to indicate the response is "Active", "Inactive" etc. instead of single character state. 

For the docs of state( Rule State), could you also update them? I believe they are incorrect

maany (changes requested):
Hey @cserf thanks for the follow up. Here are some comments but I think after this I am ready to approve this PR.

 1. subscription_state serialization

    In lib/rucio/gateway/subscription.py:232 ( see my comment), subscription_state is manually converted via .value, producing single-character values like "A", "I". Based on the last round of changes, state (RuleState) is left as a enum and serialized by APIEncoder via .name, producing fullstring names like "REPLICATING", "OK".
  
    This means the REST API response will have inconsistent formats for enum fields:
    - state: "REPLICATING" (via APIEncoder .name)
    - subscription_state: "A" (via manual .value)
 
  
  2. REST documentation 
    The enum values in the OpenAPI spec don't match actual API output: 
    - state enum is listed as ["R", "O", "S", "U", "W", "I"] but the API actually returns ["REPLICATING", "OK", ...]
    - subscription_state enum should similarly be updated to full names once the above fix is applied: ["ACTIVE", "INACTIVE",  ...]

  3. Client Test
  Could you add tests for  list_subscription_rule_states method in lib/rucio/client/subscriptionclient.py in TestSubscriptionClient?  Please cover both paths with/without include_details=True.

  4.  REST API test 
    Could you add a test for the REST layer as well in test_subscription.py as well. ( both paths) 

I think with these tests we ca be very sure that what the server returns, what the docs say and what the clients expect are all aligned.

Thanks again :D

lib/rucio/web/rest/flaskapi/v1/subscriptions.py
```diff
@@ -520,19 +527,40 @@ def get(self, account, name=None):
                       state:
                         description: "The state of the rules."
                         type: string
-                        enum: ["R", "O", "S", "U", "W", "I"]
+                        enum: ["REPLICATING", "OK", "STUCK", "SUSPENDED", "WAITING", "INJECT"]
```
maany:
```suggestion
                        enum: ["REPLICATING", "OK", "STUCK", "SUSPENDED", "WAITING_APPROVAL", "INJECT"]
```

lib/rucio/web/rest/flaskapi/v1/subscriptions.py
```diff
@@ -520,19 +527,40 @@ def get(self, account, name=None):
                       state:
                         description: "The state of the rules."
                         type: string
```
maany, line 529:
```suggestion
                        type: string
                        nullable: true
```

Since we are now doing a left outer join, the state can be null for subscriptions that have no rules

maany (changes requested):
Looks great @cserf. Just 2 code suggestions regarding the OpenAPI spec. Otherwise, I am ready to approve the PR.

lib/rucio/web/rest/flaskapi/v1/subscriptions.py
cserf: Done

lib/rucio/web/rest/flaskapi/v1/subscriptions.py
cserf: Done

maany (dismissed):
- **Quality**: Agree 🟢
- **Security**: Agree 🟢
- **Testing**: Agree 🟢
- **Documentation**: Agree 🟢

# Notes for merger

The PR adds an `include_details` query parameter to the `/subscriptions/{account}/{name}/rules/states` endpoint, returning subscription metadata (state, last_processed, lifetime, expired_at, comments) alongside rule state counts. The default path (`include_details=false`) preserves backward compatibility by returning the same response format as before.

Key points reviewed:
- Enum serialization is consistent: both `state` (RuleState) and `subscription_state` (SubscriptionState) are serialized via `APIEncoder` using `.name`, producing full string values (`"REPLICATING"`, `"ACTIVE"`) rather than single-char `.value` (`"R"`, `"A"`).
- The LEFT OUTER JOIN change is scoped appropriately and `func.count(rule.id)` correctly returns 0 for subscriptions without rules.
- OpenAPI spec is updated to reflect the new query parameter and response fields.
- Gateway, REST, and client tests cover both the default and `include_details=true` paths, including edge cases (no rules, multiple rule states).

maany (approved):
Re-approving as it needed a rebase. Please see my previous review for the notes for the merger.

bziemons:
This test failure is probably related to this change: https://github.com/rucio/rucio/actions/runs/27830309748/job/82365318057

bari12:
@cserf can you please have a look 🙏

Issue #8369, opened 2026-01-30
speed up docker_activate_rses in end-to-end tests


rcarpa:
It is painfully slow for something which has to run in the CI all the time

PR #8370, opened 2026-01-30
Speed up docker activate rses. Closes: #8369

rcarpa:


lib/rucio/core/importer.py
```diff
@@ -126,10 +126,13 @@ def import_rses(
 
         # Attributes
         attributes = rse.get('attributes', {})
-        attributes[RseAttr.LFN2PFN_ALGORITHM] = rse.get('lfn2pfn_algorithm')
-        attributes[RseAttr.VERIFY_CHECKSUM] = rse.get('verify_checksum')
+        # For legacy reasons, the following two attributes can be set at the top level
```
voetberg, line 129: I would think that these should be a different commit at the very least, maybe even a different PR.

.github/workflows/integration_tests.yml
```diff
@@ -157,24 +157,23 @@ jobs:
...
-        run: docker exec -t dev-rucio-1 tools/pytest.sh -v --tb=short tests/test_rse_protocol_ssh.py
-      - name: Test Protocol Rsync
-        run: docker exec -t dev-rucio-1 tools/pytest.sh -v --tb=short tests/test_rse_protocol_rsync.py
-      - name: Test Protocol Rclone
-        run: docker exec -t dev-rucio-1 tools/pytest.sh -v --tb=short tests/test_rse_protocol_rclone.py
-      - name: Test Conveyor
-        run: docker exec -t dev-rucio-1 tools/pytest.sh -v --tb=short tests/test_conveyor.py
+      - name: Run Tests
+        run: |
+          docker exec -e GITHUB_ACTIONS="$GITHUB_ACTIONS"  -t dev-rucio-1 tools/pytest.sh -v --tb=short \
```
voetberg, line 162: I'm probably missing something, but where is the docker env using the github actions vars? Is this needed?

rcarpa: See the change to [tools/pytest.sh](https://github.com/rucio/rucio/pull/8370/commits/ed43f0750d8631bc141e4010c075b48eacf6f2e3#diff-040cfefed23c288ab347c7c6035cc45b072214ba2ac843354c4d0be22519c779)  in the same commit. 

I'm not sure if this logic is actually needed. However, because the variable is not passed to docker correctly, at this moment it doesn't work as somebody would expect it works by reading the code.

lib/rucio/core/importer.py
rcarpa: It is already in a separate commit.

PR #8372, opened 2026-01-30
Patch 7581 remove oauthmanager daemon

### LINKED ISSUE #7581: remove oauth manager from the repo

panta-123:
remove not used oauth manager daemon., its related code which are not needed and its tests.
No changes in current workflow.


lib/rucio/core/oidc.py
```diff
@@ -40,8 +40,9 @@
 from rucio.common.config import config_get, config_get_bool, config_get_int
 from rucio.common.exception import CannotAuthenticate, CannotAuthorize, RucioException
 from rucio.common.stopwatch import Stopwatch
-from rucio.common.utils import all_oidc_req_claims_present, build_url, val_to_space_sep_str
+from rucio.common.utils import all_oidc_req_claims_present, build_url, chunks, val_to_space_sep_str
 from rucio.core.account import account_exists
+from rucio.core.authentication import __delete_expired_tokens_account
```
dchristidis, line 44: If this function is imported in another file, shouldn’t it be renamed to not be mangled?

lib/rucio/core/oidc.py
```diff
@@ -1461,3 +1466,37 @@ def oidc_identity_string(sub: str, iss: str):
...
+    query = select(
+        models.OAuthRequest.state
+    ).where(
+        models.OAuthRequest.expired_at <= datetime.utcnow(),
+        models.OAuthRequest.account == account
+    ).with_for_update(
+        skip_locked=True
+    )
+
+    oauth_requests = session.execute(query).scalars().all()
```
dchristidis: Since `chunks()` can handle an iterator, I would think it more efficient to not use `all()` here?

lib/rucio/core/oidc.py
```diff
@@ -1461,3 +1466,37 @@ def oidc_identity_string(sub: str, iss: str):
...
+
+    for chunk in chunks(oauth_requests, 100):
+        delete_query = delete(
+            models.OAuthRequest
+        ).where(
+            models.OAuthRequest.state.in_(chunk)
+        )
+        session.execute(delete_query)
+
+    session.commit()
```
dchristidis: Typically we let the decorators do the work. Is there a reason you are manually committing the changes here?

lib/rucio/core/oidc.py
```diff
@@ -1461,3 +1466,37 @@ def oidc_identity_string(sub: str, iss: str):
...
+
+
+@transactional_session
+def _delete_oauth_request_by_account_and_expiration(
+    account: str,
+    *,
+    session: "Session"
+) -> None:
+    """
+    Delete an OAuth request by its account and expiration time.
```
dchristidis, line 1024: Maybe rename this helper function to `_delete_expired_oauth_requests()`? I’m not sure, I find the current combination of name and this documentation string unintuitive.

lib/rucio/core/authentication.py
```diff
@@ -575,6 +508,9 @@ def validate_auth_token(token: str, *, session: "Session") -> "TokenValidationDi
             # identify JWT access token and validate
             # & save it in Rucio if scope and audience are correct
             if len(token.split(".")) == 3:
+                # pylint: disable=C0415
+                # noqa: F401
+                from rucio.core.oidc import validate_jwt
```
dchristidis: Could you explain why these two are necessary? As far as I can tell, `PLC0415` is not enabled in our Ruff configuration and the imported function is always used.

And maybe I would recommend adding a comment about the circular import next to the statement, in addition to the existing justification in the commit message.

dchristidis (changes requested):
I’m making a first pass over your PR, but I will need to look it over again.

The split into four commits was well done, thanks!

Please use `feat` when removing functionality. This applies to the first commit as well, because Rucio’s behaviour is not the same before and after.

lib/rucio/core/oidc.py
panta-123: removed all()

lib/rucio/core/oidc.py
panta-123: That was an oversight. removed manual commit as decorator will hadle that

lib/rucio/core/authentication.py
panta-123: My code editor was flagging that but seems not need.
And added explicit comment

lib/rucio/core/oidc.py
panta-123: done.

lib/rucio/core/oidc.py
panta-123: I am not sure we have to rename as it doesn't collide to any other.

panta-123:
> I’m making a first pass over your PR, but I will need to look it over again.
> 
> The split into four commits was well done, thanks!
> 
> Please use `feat` when removing functionality. This applies to the first commit as well, because Rucio’s behaviour is not the same before and after.

changed to feat.

panta-123:
@dchristidis , code freeze is approaching. Please have a look so that we don't miss that for new release

dchristidis (approved):

Issue #8295, opened 2025-12-29
Copy implimentation of CLI to new CLI structure


voetberg:
### Description

The new CLI relies on the old CLI's functions (in cli/bin_legacy). These should be moved over to the actual new CLI to avoid a more complicated structure and prepare for the deprecation of the old CLI.  

### Change

change

```
@component.command("...")
@click.arg(...)
def do_a_thing(....)
    (bin/rucio.do_a_thing())
```

to 

```
@component.command("...")
@click.arg(...)
def do_a_thing(....)
    [does the thing]
```

bari12:
@voetberg I am closing this since all sub-issues are done. Please re-open if there is something left.

PR #8379, opened 2026-02-04
refactor(Clients): Copy implimentation of rule CLI to new format

### LINKED ISSUE #8301: Copy old CLI implimentation of *rule* to new CLI

voetberg:
Closes #8301

### DIFF
--- a/lib/rucio/cli/rule.py
+++ b/lib/rucio/cli/rule.py
@@ -333,6 +333,8 @@ def update(
     """Update an existing rule"""
     if suspend and stuck:
         raise InputValidationError("Cannot supply both '--stuck' and '--suspend'. Supply one or the other.")
+    if cancel_requests and not (suspend or stuck):
+        raise InputValidationError('--stuck or --suspend must be specified when running --cancel-requests')
 
     options = {}
     if lifetime:
@@ -352,8 +354,6 @@ def update(
     if source_rses:
         options['source_replica_expression'] = None if source_rses.lower() == 'none' else source_rses
     if cancel_requests:
-        if 'state' not in options:
-            raise InputValidationError('--stuck or --suspend must be specified when running --cancel-requests')
         options['cancel_requests'] = True
     if priority:
         options['priority'] = int(priority)
--- a/lib/rucio/cli/rule.py
+++ b/lib/rucio/cli/rule.py
@@ -95,7 +95,7 @@ def add_(ctx, dids, copies, rses, weight, asynchronous, lifetime, grouping, lock
                     )
                     rule_ids.extend(rule_id)
                 except DuplicateRule:
-                    print('Duplicate rule for %s:%s found; Skipping.' % (did['scope'], did['name']))
+                    print(f'Duplicate rule for {did["scope"]}:{did["name"]} found; Skipping.')
         else:
             raise error
 
@@ -174,18 +174,24 @@ def show(ctx, rule_id, examine):
             print_output(*output, console=ctx.obj.console, no_pager=ctx.obj.no_pager)
         else:
             analysis = ctx.obj.client.examine_replication_rule(rule_id=rule_id)
-            print('Status of the replication rule: %s' % analysis['rule_error'])
+            print(f'Status of the replication rule: {analysis["rule_error"]}')
             if analysis['transfers']:
                 print('STUCK Requests:')
-                for transfer in analysis['transfers']:  # TODO Reformat with f-strings and padding
-                    print('  %s:%s' % (transfer['scope'], transfer['name']))
-                    print('    RSE:                  %s' % str(transfer['rse']))
-                    print('    Attempts:             %s' % str(transfer['attempts']))
-                    print('    Last Retry:           %s' % str(transfer['last_time']))
-                    print('    Last error:           %s' % str(transfer['last_error']))
-                    print('    Last source:          %s' % str(transfer['last_source']))
-                    print('    Available sources:    %s' % ', '.join([source[0] for source in transfer['sources'] if source[1]]))
-                    print('    Blocklisted sources:  %s' % ', '.join([source[0] for source in transfer['sources'] if not source[1]]))
+                lpad = ' ' * 5
+
+                def _format_print(name, value):
+                    name += ":"
+                    print(f"{lpad}{name.ljust(20)}{value}")
+                for transfer in analysis['transfers']:
+                    print(f'  {transfer["scope"]}:{transfer["name"]}')
+                    _format_print('RSE', transfer['rse'])
+                    _format_print('Attempts', transfer['attempts'])
+                    _format_print('Last Retry', transfer['last_time'])
+                    _format_print('Last error', transfer['last_error'])
+                    _format_print('Last source', transfer['last_source'])
+                    _format_print('Available sources', ', '.join([source[0] for source in transfer['sources'] if source[1]]))
+                    _format_print('Blocklisted sources', ', '.join([source[0] for source in transfer['sources'] if not source[1]]))
+
     else:
         rule = ctx.obj.client.get_replication_rule(rule_id=rule_id)
         if ctx.obj.use_rich:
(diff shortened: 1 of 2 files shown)


lib/rucio/cli/rule.py
```diff
@@ -41,29 +48,78 @@ def rule():
 @click.option("--account", help="The account owning the rule")
 @click.option("--skip-duplicates", is_flag=True, default=False, help="Skip duplicate rules")
 @click.pass_context
-def add_(ctx, dids, copies, rses, weight, asynchronous, lifetime, grouping, locked, source_rses, notify, activity, comment, ask_approval, delay_injection, account, skip_duplicates):
+def add_(
+    ctx: click.Context,
+    dids:tuple[str, ...],
```
bziemons: missing space after `dids:`

tests/test_cli_client_structure.py
```diff
@@ -801,7 +801,6 @@ def test_rule(rucio_client, mock_scope):
     # Testing the two different lifetime type options
     cmd = f"rucio rule update {rule_id} --lifetime 10"
     exitcode, out, err = execute(cmd)
-    print(out, err)
```
bziemons, line 804: sneaky change, I think this change is unnecessary? Will only be relevant when the test fails, anyway.

tests/test_cli_client_structure.py
```diff
@@ -857,6 +856,12 @@ def test_rule(rucio_client, mock_scope):
     assert "ERROR" not in err
     assert len(out.split("\n")) == 3  # Creates two rules with independent IDs and one extra line at the end
 
+    test_rule = out.split("\n")[0]
+    cmd = f'rucio rule show {test_rule}'
+    exitcode, out, err = execute(cmd)
+    assert rule_rse in out   # Only thing we explicitly set, so it will be there
+    assert exitcode == 0
+
```
bziemons, line 865: this should be in a separate commit, preferably prepending its current commit and checked, whether the test passes without the other changes

lib/rucio/cli/rule.py
```diff
@@ -391,42 +391,48 @@ def list_(ctx, did, traverse, csv, file, account, subscription):
...
-                                   rule['account'],
-                                   f"{rule['scope']}:{rule['name']}",
-                                   f"[{CLITheme.RULE_STATE.get(rule['state'], 'default')}]{rule['state']}[/][{rule['locks_ok_cnt']}/{rule['locks_replicating_cnt']}/{rule['locks_stuck_cnt']}]",
-                                   rule['rse_expression'],
-                                   rule['copies'],
-                                   sizefmt(rule['bytes'], ctx.obj.human) if rule['bytes'] is not None else 'N/A',
-                                   rule['expires_at'],
-                                   rule['created_at']])
+                table_data.append(
+                    [rule['id'],
```
bziemons: for readability I would prefer `…append([`, but maybe I am the odd one

lib/rucio/cli/rule.py
```diff
@@ -277,7 +296,7 @@ def history(ctx, did):
 @click.option("--activity", help="Update activity for moved rule", hidden=True)  # Should only do this using `update`
 @click.option("--source-rses", help="Update how replicas are sourced for the rule")
 @click.pass_context
-def move(ctx, rule_id, rses, activity, source_rses):
+def move(ctx: click.Context, rule_id: str, rses: str, activity: Optional[str], source_rses: Optional[str]):
```
bziemons: missing return type, accidental?

lib/rucio/cli/rule.py
```diff
@@ -337,9 +337,11 @@ def update(
         options['comment'] = comment
     if account:
         options['account'] = account
-    if stuck:  # TODO Error when both are selected
+    if stuck:
         options['state'] = 'STUCK'
     if suspend:
+        if 'state' in options:
+            raise InputValidationError("Cannot supply both '--stuck' and '--suspended'. Supply one or the other'")
```
bziemons: I would prefer a `if stuck and suspend:` before `if stuck:` to this solution, because it doesn't need a key lookup and doesn't use nested ifs. Please reconsider.

bziemons (changes requested):

lib/rucio/cli/rule.py
voetberg: Accidental....

lib/rucio/cli/rule.py
voetberg: Wonder why the precommit linter didn't complain....

lib/rucio/cli/rule.py
voetberg: Agreed. I was following the convention set up line 350-52, but it's not very clean.

tests/test_cli_client_structure.py
voetberg: I have it noted in the commit. It should be a different style commit tho, right.

lib/rucio/cli/rule.py
voetberg: I'm not plussed about it either way so I'll just defer to my reviewer

tests/test_cli_client_structure.py
bziemons: since the commit message mentions this being a stray print, I think it is there for the reason to see the command output when the test fails. That means to me it is not really stray in that sense. Do you still want to remove it?

lib/rucio/cli/rule.py
```diff
@@ -73,27 +129,165 @@ def add_(ctx, dids, copies, rses, weight, asynchronous, lifetime, grouping, lock
...
+            analysis = ctx.obj.client.examine_replication_rule(rule_id=rule_id)
+            print(f'Status of the replication rule: {analysis["rule_error"]}')
+            if analysis['transfers']:
+                print('STUCK Requests:')
+                lpad = ' '*5
+                def _format_print(name, value):
+                    name += ":"
+                    print(f"{lpad}{name.ljust(20)}{value}")
+                for transfer in analysis['transfers']:
+                    print(f'  {transfer["scope"]}: {transfer["name"]}')
```
bziemons: there should be no space between the colon and the name

lib/rucio/cli/rule.py
```diff
@@ -73,27 +129,165 @@ def add_(ctx, dids, copies, rses, weight, asynchronous, lifetime, grouping, lock
...
+                    _format_print('RSE', transfer['rse'])
+                    _format_print('Attempts', transfer['attampts'])
```
bziemons: `attampts` typo?

lib/rucio/cli/rule.py
```diff
@@ -73,27 +129,165 @@ def add_(ctx, dids, copies, rses, weight, asynchronous, lifetime, grouping, lock
...
+                    _format_print('Last Retry', transfer['last_time'])
+                    _format_print('Last Error', transfer['last_error'])
```
bziemons: "Last source" output was removed? Any particular reason?

lib/rucio/cli/rule.py
```diff
@@ -73,27 +129,165 @@ def add_(ctx, dids, copies, rses, weight, asynchronous, lifetime, grouping, lock
...
+            output.append(f'Status of the replication rule: {rule_status}')
+            if analysis['transfers']:
+                output.append('[b]STUCK Requests:[/]')
+                for transfer in analysis['transfers']:
+                    output.append(Padding.indent(Text(f"{transfer['scope']}:{transfer['name']}", style=CLITheme.SUBHEADER_HIGHLIGHT), 2))
+                    table_data = [
+                        ['RSE:', str(transfer['rse'])],
+                        ['Attempts:', str(transfer['attempts'])],
+                        ['Last retry:', str(transfer['last_time'])],
+                        ['Last error:', str(transfer['last_source'])],
```
bziemons: this was already a problem in the previous code, but last error != last source

bziemons (changes requested):

tests/test_cli_client_structure.py
voetberg: I'm mostly just bothered that this is the only place where this kind of print statement is used in these tests, so it seems inconsistent. It's not that big a deal so I can drop this commit

lib/rucio/cli/rule.py
voetberg: Corrected in new commit

lib/rucio/cli/rule.py
voetberg: Nope, I assume it was a copy-paste error

lib/rucio/cli/rule.py
```diff
@@ -163,7 +163,8 @@ def show(ctx, rule_id, examine):
                         ['RSE:', str(transfer['rse'])],
                         ['Attempts:', str(transfer['attempts'])],
                         ['Last retry:', str(transfer['last_time'])],
-                        ['Last error:', str(transfer['last_source'])],
+                        ['Last error:', str(transfer['last_error'])],
+                        ['Last source:', str(transfer['last_source'])],
```
bziemons, line 186: I am wondering if this fix should be done on the legacy code before copying?

bziemons (dismissed):
- **Quality**: Agree 🟢
- **Security**: Agree 🟢
- **Testing**: Agree 🟢
- **Documentation**: Agree 🟢

# Notes for merger

Looks good now. Please mind the comment below and decide whether it is necessary.

lib/rucio/cli/rule.py
```diff
@@ -124,28 +327,45 @@ def move(ctx, rule_id, rses, activity, source_rses):
...
-            "state_suspended": suspend,
-            "cancel_requests": cancel_requests,
-            "priority": priority,
-            "child_rule_id": child_rule_id,
-            "boost_rule": boost_rule,
-        }
-    )
-    update_rule(args, ctx.obj.client, ctx.obj.logger, ctx.obj.console, ctx.obj.spinner)
+    if suspend and stuck:
+        raise InputValidationError("Cannot supply both '--stuck' and '--suspended'. Supply one or the other'")
```
dchristidis: 1. `--suspend` instead of `--suspended`
2. Period instead of single quote (at the end of the string)

lib/rucio/cli/rule.py
dchristidis: I’m also becoming curious about the approach to #8295. Do all future fixes go to the `cli` directory exclusively, letting `bin_legacy` stagnate? Do we maintain both implementations?

voetberg: This is something that definitely needs some thought. My approach right now is to apply new fixes forward to the new version because of the eventually planned full deprecation of the old version, and only do fixes for the old versions that have specific issues for them.

dchristidis: Very well. I fear this may cause some confusion, but it should be fine as long as this period proves to be short.

dchristidis (approved):

dchristidis:
The tests are failing on the `release-39` branch. Please check.

dchristidis:
@voetberg Did you get a chance to investigate the issue? Otherwise we will have to revert the commits on the release branch.

bari12:
Reverting, is failing tests.

Geogouz:
See here for a probable reason: https://github.com/rucio/rucio/issues/8301#issuecomment-3984278436

PR #8385, opened 2026-02-06
fix(Policies): make diff-based schemas work fully

### LINKED ISSUE #7731: Schema values overridden in policy package are not picked up unless at top level
### Description

Policy packages can provide a schema module that can override some or all of the values in the generic schema. However, if a policy package overrides a value (for example, `NAME`) that is used within higher level schema constructs (in this case `COLLECTION`, `DID`, `ATTACHMENT`, `ADD_REPLICA_FILE`, `DELETE_REPLICA_FILE`) the overridden value will not be used unless the policy package also overrides those higher level constructs. The previous work on this (see https://github.com/rucio/rucio/issues/6842 ) did not resolve this issue.

### Motivation

This would allow simpler and more concise policy packages without so much boilerplate code required. It also makes intuitive sense that an overridden value should be used everywhere in place of the generic value.

### Change

Fixing it would require some more intelligence in the schema code to find overridden values in the policy package and use them wherever they should be used. I had code that would walk the schema tree and automatically substitute overridden values where necessary but it was rejected as too complex. I'm happy to discuss simpler options that would solve this problem.

jamesp-epcc:
Closes: #7731

This PR makes "diff-based" schemas in policy packages work correctly so that items not at the top level of the schema hierarchy can be overridden. It does this in a simpler way than my previous attempt, by introducing a new `SchemaRef` class that is used when schema values reference other schema values. These references are then resolved by checking the policy package first and falling back to the generic schema when appropriate.

(The reason `SchemaRef` is in its own file instead of in `__init__.py` is to avoid circular imports).

### DIFF
--- a/lib/rucio/common/schema/__init__.py
+++ b/lib/rucio/common/schema/__init__.py
@@ -24,6 +24,7 @@ from jsonschema import ValidationError, validate
 from rucio.common import config, exception
 from rucio.common.constants import DEFAULT_VO
 from rucio.common.plugins import check_policy_module_version
+from rucio.common.schema.schema_ref import SchemaRef
 from rucio.db.sqla.constants import DatabaseOperationType
 from rucio.db.sqla.session import db_session
 
@@ -146,6 +147,25 @@ def load_schema_for_vo(vo: str) -> None:
     schema_modules[vo] = module
 
 
+def _resolve_references(schema: Any, vo: str = DEFAULT_VO) -> Any:
+    # if this is a schema reference, resolve it
+    if isinstance(schema, SchemaRef):
+        result = schema.resolve(vo)
+    # if this is a list or dictionary, resolve recursively
+    elif isinstance(schema, list):
+        result = []
+        for value in schema:
+            result.append(_resolve_references(value))
+    elif isinstance(schema, dict):
+        result = {}
+        for key in schema:
+            result[key] = _resolve_references(schema[key])
+    else:
+        # for other types (e.g. scalars), return unmodified
+        result = schema
+    return result
+
+
 def validate_schema(name: str, obj: Any, vo: str = DEFAULT_VO) -> None:
     if obj:
         if vo not in schema_modules:
@@ -155,6 +175,7 @@ def validate_schema(name: str, obj: Any, vo: str = DEFAULT_VO) -> None:
         else:
             # if schema not available in VO module, fall back to generic module
             schema = _get_generic_schema_module().SCHEMAS.get(name, {})
+        schema = _resolve_references(schema, vo)
         try:
             validate(obj, schema)
         except ValidationError as error:
@@ -165,8 +186,8 @@ def get_schema_value(key: str, vo: str = DEFAULT_VO) -> Any:
     if vo not in schema_modules:
         load_schema_for_vo(vo)
     if not hasattr(schema_modules[vo], key):
-        return getattr(_get_generic_schema_module(), key)
-    return getattr(schema_modules[vo], key)
+        return _resolve_references(getattr(_get_generic_schema_module(), key), vo)
+    return _resolve_references(getattr(schema_modules[vo], key), vo)
 
 
 def get_scope_name_regexps() -> list[str]:
--- a/lib/rucio/common/schema/__init__.py
+++ b/lib/rucio/common/schema/__init__.py
@@ -24,6 +24,7 @@ from jsonschema import ValidationError, validate
 from rucio.common import config, exception
 from rucio.common.constants import DEFAULT_VO
 from rucio.common.plugins import check_policy_module_version
+from rucio.common.schema.schema_ref import SchemaRef
 from rucio.db.sqla.constants import DatabaseOperationType
 from rucio.db.sqla.session import db_session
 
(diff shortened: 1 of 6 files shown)


lib/rucio/common/schema/__init__.py
```diff
@@ -164,9 +185,11 @@ def validate_schema(name: str, obj: Any, vo: str = DEFAULT_VO) -> None:
 def get_schema_value(key: str, vo: str = DEFAULT_VO) -> Any:
     if vo not in schema_modules:
         load_schema_for_vo(vo)
-    if not hasattr(schema_modules[vo], key):
-        return getattr(_get_generic_schema_module(), key)
-    return getattr(schema_modules[vo], key)
+    if hasattr(schema_modules[vo], key):
```
voetberg: I think this counts as a style change and should be a different commit

lib/rucio/common/schema/schema_ref.py
```diff
@@ -0,0 +1,40 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Any, Optional
+
+from rucio.common.constants import DEFAULT_VO
+
+
+class SchemaRef:
```
voetberg, line 20: Can you add a few more docstrings here? It becomes obvious as to what this is when it's part of the PR, but less as just part of the repo

lib/rucio/common/schema/__init__.py
```diff
@@ -146,6 +147,25 @@ def load_schema_for_vo(vo: str) -> None:
     schema_modules[vo] = module
 
 
+def _resolve_references(schema: Any, vo: str = DEFAULT_VO) -> Any:
```
voetberg, line 150: Is the type of both the schema and the return object really any? I thought they were both limited to either some sort of dictionary, sequence, string or numeric

tests/mocks/schema_diff.py
voetberg, line 1: Following some of the new PR convensions, it would probably be better if the tests were their own "test(Policies)" commits

lib/rucio/common/schema/__init__.py
jamesp-epcc: You're right, I should constrain these to the types that are actually used.

lib/rucio/common/schema/__init__.py
jamesp-epcc: I thought it would be best to avoid repeating the call to `_resolve_references`, and then in the process I thought we could also avoid repeating the `getattr` call. But I guess that does count as a style change to the original code. I can make the change in the original style and we can have another commit to change it if that seems worth doing.

lib/rucio/common/schema/schema_ref.py
jamesp-epcc: Sure

tests/mocks/schema_diff.py
jamesp-epcc: I can separate them into their own commit.

jamesp-epcc:
I had the return type of `_resolve_references` as a union of all the possible types for a schema value, but this caused the type checking to fail because not all of those types are valid for the schema argument of the `validate` function in the `jsonschema` library, so I have reverted them back to `Any`. I can't think of a good way to handle this that keeps the stricter type checking but without complicating the code for what seems like little gain. However I'm open to other suggestions.

voetberg:
This seems like a solid solution but I don't have a ton of expertise on package deployments to feel confident in a review. I would ask the comp lead to review but that's you :sob:

jamesp-epcc:
Reopening this as I'm hoping to get it merged during the next sprint.

jamesp-epcc:
Reopening this again as I would really like to try and get it merged in this release. I've had users contact me asking for this feature.

voetberg (dismissed):
Giving a tentative approval, I do not have a ton of expertise in this area but the solution seems workable to me

lib/rucio/common/schema/__init__.py
```diff
@@ -146,6 +147,25 @@ def load_schema_for_vo(vo: str) -> None:
...
+    # if this is a schema reference, resolve it
+    if isinstance(schema, SchemaRef):
+        result = schema.resolve(vo)
+    # if this is a list or dictionary, resolve recursively
+    elif isinstance(schema, list):
+        result = []
+        for value in schema:
+            result.append(_resolve_references(value))
```
bari12: In this recursive call, I think you need to bring in the VO again, otherwise this is lost?

lib/rucio/common/schema/__init__.py
```diff
@@ -146,6 +147,25 @@ def load_schema_for_vo(vo: str) -> None:
...
+    elif isinstance(schema, dict):
+        result = {}
+        for key in schema:
+            result[key] = _resolve_references(schema[key])
```
bari12: In this recursive call, I think you need to bring in the VO again, otherwise this is lost?

bari12 (changes requested):
Minor change request. see comment otherwise gtg

lib/rucio/common/schema/__init__.py
jamesp-epcc: Well spotted. You are right. I will fix that.

bari12 (approved):
- **Quality**: Agree 🟢
- **Security**: Agree 🟢
- **Testing**: Agree 🟢
- **Documentation**: Agree 🟢

# Notes for merger

PR #8386, opened 2026-02-09
 fix(Clients): Do not overwrite purge-replicas arg in deletion 

### LINKED ISSUE #8356: rucio delete-rule unexpectedly flips purge_replicas from True to False
### Description

While deleting a rule using the Rucio CLI, I noticed that the purge_replicas flag of the rule is being modified during deletion.

Specifically, a rule created with purge_replicas=True ends up, after deletion, recorded with purge_replicas=False.

### Steps to reproduce

1. Create a rule with purge_replicas set to True
2. Delete the rule using the cli (I tried with the old command `rucio delete-rule ....`)
3. The now-deleted rule has purge_replicas set to False

### Rucio Version

We upgraded to 39 since, but I believe this occurred while we were running 38.5.1

### Additional Information

_No response_

voetberg:
Closes #8356 

Adds an (already passing) test in `test_rules.py` that verifies this problem is happening purely at the argument specification level in the CLI. This test is not strictly required to show the issue has been fixed, but it's a good sanity check regardless.

### DIFF
--- a/lib/rucio/cli/bin_legacy/rucio.py
+++ b/lib/rucio/cli/bin_legacy/rucio.py
@@ -2532,7 +2532,7 @@ You can filter by key/value, e.g.::
     delete_rule_parser = subparsers.add_parser('delete-rule', help='Delete replication rule.')
     delete_rule_parser.set_defaults(function=delete_rule)
     delete_rule_parser.add_argument(dest='rule_id', action='store', help='Rule id or DID. If DID, the RSE expression is mandatory.')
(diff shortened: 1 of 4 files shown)


dchristidis (approved):
I have a question which falls outside the scope of this PR. This sends:

```
2026-02-12 14:51:53,048	DEBUG	baseclient.py	Request data (length=24): [{"purge_replicas": null}]
```

Currently, the parameter is not described at all in the endpoint’s OpenAPI specification. But if it were, we would have used `boolean`. In that case, I think `null` is not an acceptable value.

Do we need to think of some mechanism to avoid sending parameters when undefined by the user?

bari12:
Reverting, is failing tests.

PR #8442, opened 2026-03-12
feat(Rules): make time delta for stuck rules updated_at configurable in repairer (float version)

### LINKED ISSUE #8441: Make repairer daemon time delta for rule checks configurable
### Description

Related to #8436, a quick improvement would be to make the `delta` parameter configurable.

This might increase resource load, but would allow the transfers to start at least, i.e. configure it to be -1, the same value used when running the daemon with `--once` would allow it repair the rules always.

This might result in a lot of unnecessary repair attempts on permanently stuck rules, but at least for us, this would be a good tradeoff for now until #8436 is properly solved.

### Motivation

_No response_

### Change

_No response_

maxnoe:
…

Closes #8441

## Checklist
- [X] Created issue: #8441 
- [ ] Added tests
- [X] Updated documentation (https://github.com/rucio/documentation/pull/761)
- [X] Picked a reviewer after submission (Leave empty, if unclear)
- [X] I understand that stale (failing tests, author not answering) PRs will be closed promptly

How to best test this?

Documentation update is needed.

<details>

```
- **Quality**: Agree 🟢
- **Testing**: Agree 🟢
- **Documentation**: Agree 🟢

# Notes for merger



```
</details>

### DIFF
--- a/lib/rucio/core/rule.py
+++ b/lib/rucio/core/rule.py
@@ -2356,7 +2356,7 @@ def get_injected_rules(
 def get_stuck_rules(
     total_workers: int,
     worker_number: int,
-    delta: int = 600,
+    delta: float = 600.0,
     limit: int = 10,
     blocked_rules: Optional['Sequence[str]'] = None,
     *,
--- a/lib/rucio/daemons/judge/repairer.py
+++ b/lib/rucio/daemons/judge/repairer.py
@@ -30,6 +30,7 @@ from sqlalchemy.exc import DatabaseError
 
 import rucio.db.sqla.util
 from rucio.common import exception
+from rucio.common.config import config_get_float
 from rucio.common.exception import DatabaseException
 from rucio.common.logging import setup_logging
 from rucio.core.monitor import MetricManager
@@ -52,6 +53,11 @@ def rule_repairer(
     """
     Main loop to check for STUCK replication rules
     """
+    if once:
+        delta = -1.0
+    else:
+        delta = config_get_float("repairer", "updated_at_delta", default=1800.0)
+
     paused_rules = {}  # {rule_id: datetime}
     run_daemon(
         once=once,
@@ -62,14 +68,14 @@ def rule_repairer(
         run_once_fnc=functools.partial(
             run_once,
             paused_rules=paused_rules,
-            delta=-1 if once else 1800,
+            delta=delta,
         )
     )
 
 
 def run_once(
         paused_rules: dict[str, datetime],
-        delta: int,
+        delta: float,
         heartbeat_handler: HeartbeatHandler,
         **_kwargs
 ) -> None:


voetberg:
Commit component should probably be "Recovery"? 

@0xquark Is there a reason we don't have a daemons component included as an option for the commit lint?

lib/rucio/daemons/judge/repairer.py
```diff
@@ -52,6 +53,11 @@ def rule_repairer(
     """
     Main loop to check for STUCK replication rules
     """
+    if once:
+        delta = -1
```
voetberg: Can the typing issue be fixed by just changing `config_get_float` to `config_get_int`?

0xquark:
> Commit component should probably be "Recovery"?

Yes, Recovery would be the appropriate component here.

> 
> @0xquark Is there a reason we don't have a daemons component included as an option for the commit lint?

It's because we don't have a [specific component lead](https://rucio.github.io/documentation/component_leads) for all the daemons.

lib/rucio/daemons/judge/repairer.py
maxnoe: I think float is the correct type, I'll change the hint

maxnoe:
Transfers or Rules seems more fitting, at least our use case for making it configurable is not data recovery

lib/rucio/core/rule.py
```diff
@@ -2354,7 +2354,7 @@ def get_injected_rules(
 def get_stuck_rules(
     total_workers: int,
     worker_number: int,
-    delta: int = 600,
+    delta: float = 600.0,
```
dchristidis, line 2359: Why the switch from `int` to `float`?

maxnoe: It's the correct type. The value is passed to `timedelta(seconds=delta)`, which accepts a float, not an int.

The type hint was added later than the default of 1800, so I think it was added without checking the receiver of the data.

dchristidis: All evidence points to the contrary.

The [documentation](https://docs.python.org/3/library/datetime.html#datetime.timedelta) states:
> Arguments may be integers or floats, and may be positive or negative.

The type checker raises no errors:
```concole
$ cat foo.py 
import datetime

delta = 600
datetime.timedelta(seconds=delta)
$ ty check foo.py 
All checks passed!
```

And in general, the behaviour of numeric types is explained in [PEP 484](https://peps.python.org/pep-0484/#the-numeric-tower):
> when an argument is annotated as having type `float`, an argument of type `int` is acceptable

So if this change isn’t necessary, is it still useful? Personally, I do not think so. But even if I did, it should have been in a separate commit.

maxnoe: > when an argument is annotated as having type float, an argument of type int is acceptable

But this is exactly the motiviation for making the annotation float. "If annotated float, an int is acceptable". Not the other way around.

See the discussion above with @voetberg about using `config_get_float` vs. `config_get_int`.

The correct type is `float`. We could discuss if it should be `float | int`, but the text you quote defines `int` as acceptable input to `float` annotated parameters, so `float | int` simplifies to `float`.

As you wrote:

> Arguments may be integers or floats 

→ `float | int`

and

> when an argument is annotated as having type float, an argument of type int is acceptable

→ `float`

dchristidis: > But this is exactly the motiviation for making the annotation float. "If annotated float, an int is acceptable". Not the other way around.

I see no motivation for the other way around. Who would need to configure the delta to `600.5`? You can set a stricter type at the higher level, in a sense, similar to how a parameter can be of type `Mapping` and receive an argument of type `dict`.

> The correct type is float.

I simply do not agree with your statement.

maxnoe: So your point is that you want to use the type hint to tell people that fractional seconds here don't make much sense given the typical time scales? The usual values will even be higher than integral seconds.

A type hint doesn't seem like the right place for this to me.

maxnoe: The question for the type hint and for `config_get_int` vs. `config_get_float` to me is: would it be _wrong_ to pass 500.0 or 600.5: no it isn't. Maybe peculiar, but the code works as intended and the value is what the user expects it to be.

The other way around is clearly wrong: if I write 600.5 for a timedelta as a user, I would be highly annoyed if Rucio rejects this with the argument that an int is required and would be slightly surprised if it accepted the value but truncated it to 600.

dchristidis: > So your point is that you want to use the type hint to tell people that fractional seconds here don't make much sense given the typical time scales?

Yes, in my opinion, advertising it as a `float` with a default value of `600.0` is misleading because it suggests that the fractional part is significant.

> The other way around is clearly wrong: … [I] would be slightly surprised if it accepted the value but truncated it to 600.

How would you notice the difference?

voetberg: Is this discussion useful? I would say make this a new issue and just keep the corrected type-hint as it stands (Because it passes tests). This is a useful change that is getting hung up on details.

dchristidis: > I would say make this a new issue and just keep the corrected type-hint as it stands (Because it passes tests).

Passing the tests is prerequisite, not a sufficient condition.

Why make a new issue to potentially revert a change? The disagreement about the type hints and which function does not block addressing the original issue; it would be more appropriate to postpone the portion that is under dispute.

> This is a useful change that is getting hung up on details.

All the more important to omit those ‘details’.

maxnoe: The change is introducing a new configuration option. I want this done correctly. Adding a new option with `config_get_int` if I think the correct type is `config_get_float` just to make you happy puts the burden back on me to fix it in a separate issue / PR _after_ having the same discussion again.

Please, make up your mind on this. I hold that float is the correct type annotation for this. That you say fractional values do not make much sense for this configuration option (I agree!) does not change what type is correct here.

maxnoe: In a statically typed language, the type of the interface would either be `double` or there would be an overload for `int` and `double`. As you posted before, the correct soltution here according to the typing PEP is the float annotation.

dchristidis: > I want this done correctly.

So do I. You believe that `float` is the correct type; I believe that `int` the correct type.

> puts the burden back on me to fix it in a separate issue

Assuming it will need to be fixed, I volunteer to do it for you.

> just to make you happy

Please be mindful of what you say. My role as a reviewer and merger is to lead the project in the right direction, to the best of my ability. To suggest that I devote this effort for my own personal satisfaction would be unkind.

maxnoe: > To suggest that I devote this effort for my own personal satisfaction would be unkind.

I apologize, that is in no way how I want this to be understood, I meant "passing the review" of course.

How do we resolve this? I don't want to have something I think is incorrect commited under my authorship.

dchristidis: > I don't want to have something I think is incorrect commited under my authorship.

For the same reason, I cannot approve or merge what I perceive as incorrect.

Thankfully, we have @bari12 as the project lead who can break this impasse.

maxnoe: Just to speed this up, I now also opened the alternative version: https://github.com/rucio/rucio/pull/8553

bari12: Hi,
I am closing this and merging the int one. 
I can follow both arguments, but what I feel strongest about what is most natural to the user, and I just think it is unlikely that somebody would ever want to set a float value there. Secondly, we have similar parameters/settings in the code for delays/timedeltas, they are all ints.
Many thanks!

PR #8414, opened 2026-03-02
Fix(Core): use logging in dirac, replace Belle-II terminology

maxnoe:
*By submitting this PR, I confirm I have followed the [Contributing Guide](https://rucio.cern.ch/documentation/contributing/).*

Fixes #8171 

Fixes #7824

## Checklist
- [X] Created issue: #8171, #7824
- [x] Added tests
- [ ] Updated documentation (Please link PRs in documentation repository)
- [ ] Added database migrations (if needed)
- [ ] For backwards compatibility breaking changes, leave additional description, follow conventional commits
- [ ] Picked a reviewer after submission (Leave empty, if unclear)
- [ ] I understand that stale (failing tests, author not answering) PRs will be closed promptly

## Additional description for reviewer

*Note: This OPTIONAL is only relevant for the REVIEWER, please leave it in the PR*

<details>
<summary>Reviewer template </summary>
Reviewers should copy&paste the code-block below and fill it out for APPROVED pull requests.
If the PR does not meet the standards the project sets out, the reasons should be WELL EXPLAINED in a CHANGE REQUEST (The answers below do not need to be answered in that case)

- **Confidence in review**: I am confident in my review concerning the components this PR touches: [*High 🟢, Medium 🟡 Low 🔴*]
- **Confidence in scope**: I am confident that this fits into the scope of the project and should be included: [*High 🟢, Medium 🟡, Low 🔴*]
  - For Medium and Low, explain in notes why this should be included
- **Quality**: The approach is sound, maintainable and addresses the issue in the best way: [*Agree 🟢*]
- **Security**: This PR does NOT require increased attention in terms of security (E.g. new dependencies): [*Agree 🟢, Disagree 🔴*]
  - If *Disagree* explain in notes.
- **Backwards compatibility**: This PR does NOT introduce backwards compatibility breaking changes: [*Agree 🟢, Disagree 🔴*]
  - If *Disagree* explain in notes
- **Testing**: This PR is well tested: [*Agree 🟢*]
- **Documentation**: Relevant documentation or comments are updated or not required: [*Agree 🟢*]

```
- **Confidence in review**: High 🟢 Medium 🟡 Low 🔴
- **Confidence in scope**: High 🟢 Medium 🟡 Low 🔴
- **Quality**: Agree 🟢
- **Security**: Agree 🟢 Disagree 🔴
- **Backwards compatibility**: Agree 🟢 Disagree 🔴
- **Testing**: Agree 🟢
- **Documentation**: Agree 🟢

[...]

PR #8444, opened 2026-03-12
Change didtype to explicitly use scope extraction policy 

### LINKED ISSUE #7979: DID constructor does not use policy `extract_scope` function to determine scope

voetberg:
Closes #7979 
(To be changed from draft after #8414 is merged)

## Checklist
- [x] Created issue: 7979
- [x] Added tests

<details>

```
- **Quality**: Agree 🟢
- **Testing**: Agree 🟢
- **Documentation**: Agree 🟢

# Notes for merger



```
</details>

### DIFF
--- a/lib/rucio/common/didtype.py
+++ b/lib/rucio/common/didtype.py
@@ -15,10 +15,12 @@
 """
 DID type to represent a DID and to simplify operations on it
 """
-
+import logging
+from configparser import NoSectionError
 from typing import Any, Union
 
-from rucio.common.exception import DIDError
+from rucio.common.exception import ConfigNotFound, DIDError, InvalidAlgorithmName, PolicyPackageNotFound
+from rucio.common.utils import extract_scope
 
 
 class DID:
@@ -126,15 +128,11 @@ class DID:
         Construct the DID from a string.
         :param did: string containing the DID information
         """
-        did_parts = did.split(DID.SCOPE_SEPARATOR, 1)
-        if len(did_parts) == 1:
-            self.name = did
-            self._update_implicit_scope()
-            if not self.has_scope():
-                raise DIDError('Object construction from non-splitable string is ambigious')
-        else:
-            self.scope = did_parts[0]
-            self.name = did_parts[1]
+        try:
+            self.scope, self.name = extract_scope(did)
+        except (ImportError, InvalidAlgorithmName, ConfigNotFound, NoSectionError) as e:  # Only use when the policy can not be found
+            logging.error("Failure using extract_scope policy for '%s': %s" % (did, type(e).__name__))
+            raise PolicyPackageNotFound('ScopeExtractionAlgorithm')
 
     def _did_from_dict(self, did: dict[str, str]) -> None:
         """
--- a/lib/rucio/common/utils.py
+++ b/lib/rucio/common/utils.py
@@ -33,6 +33,7 @@ import threading
 import time
 import types
 from collections import OrderedDict
+from configparser import NoOptionError, NoSectionError
 from enum import Enum
 from functools import cache, update_wrapper, wraps
 from io import StringIO
@@ -47,7 +48,7 @@ from typing_extensions import ParamSpec
 
 from rucio.common.config import config_get, config_get_bool
 from rucio.common.constants import BASE_SCHEME_MAP, DEFAULT_VO, POLICY_ALGORITHM_TYPES_LITERAL
-from rucio.common.exception import DIDFilterSyntaxError, DuplicateCriteriaInDIDFilter, InputValidationError, InvalidType, MetalinkJsonParsingError, MissingModuleException, RucioException
+from rucio.common.exception import ConfigNotFound, DIDFilterSyntaxError, DuplicateCriteriaInDIDFilter, InputValidationError, InvalidType, MetalinkJsonParsingError, MissingModuleException, RucioException
 from rucio.common.extra import import_extras
 from rucio.common.plugins import PolicyPackageAlgorithms
 from rucio.common.types import InternalAccount, InternalScope, LFNDict, TraceDict
@@ -663,11 +664,16 @@ def extract_scope(
         default_extract: str = 'def',
         vo: str = DEFAULT_VO
 ) -> 'Sequence[str]':
-    scope_extraction_algorithms = ScopeExtractionAlgorithms(vo)
-    extract_scope_convention = config_get('common', 'extract_scope', False, None) or config_get('policy', 'extract_scope', False, None)
-    if extract_scope_convention is None or not ScopeExtractionAlgorithms.supports(extract_scope_convention):
-        extract_scope_convention = default_extract
-    return scope_extraction_algorithms.extract_scope(did, scopes, extract_scope_convention)
+    try:
+        scope_extraction_algorithms = ScopeExtractionAlgorithms(vo)
+        extract_scope_convention = config_get('common', 'extract_scope', False, None) or config_get('policy', 'extract_scope', False, None)
+        if extract_scope_convention is None or not ScopeExtractionAlgorithms.supports(extract_scope_convention):
+            extract_scope_convention = default_extract
+        extracted = scope_extraction_algorithms.extract_scope(did, scopes, extract_scope_convention)
+    except (ConfigNotFound, NoSectionError, NoOptionError):
+        logging.warning("Config option for scope extraction convention not found. Using default convention.")
+        extracted = ScopeExtractionAlgorithms.extract_scope_default(did, scopes)
+    return extracted
 
 
 def pid_exists(pid: int) -> bool:
--- a/tests/rucio/common/test_didtype.py
+++ b/tests/rucio/common/test_didtype.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+
 import pytest
 
 from rucio.common.didtype import DID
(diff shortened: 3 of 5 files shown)


lib/rucio/common/didtype.py
```diff
@@ -126,15 +128,20 @@ def _did_from_str(self, did: str) -> None:
...
-            self.name = did
-            self._update_implicit_scope()
-            if not self.has_scope():
-                raise DIDError('Object construction from non-splitable string is ambigious')
-        else:
-            self.scope = did_parts[0]
-            self.name = did_parts[1]
+        try:
+            self.scope, self.name = extract_scope(did)
+        except (ImportError, InvalidAlgorithmName, ConfigNotFound, NoSectionError) as e:  # Only use when the policy can not be found
```
maxnoe, line 133: I think this should never happen here. There is a default extract_scope that handles the `<scope>:<name>` case and some atlas specific conventions, essentially what is the fallback implementation here below.

The exceptions you list here are a clear sign of misconfiguration, and I think should be bubbled up, not silently fallback to something unintended.


https://github.com/rucio/rucio/blob/c8242dd60ebc563da9ed006c8041d93deb4b7cdf/lib/rucio/common/utils.py#L614-L643

voetberg:
Adding no-bot until this is out of draft

lib/rucio/common/didtype.py
voetberg: That is fair.If my memory serves this was done because some tests had that behavior (not bubbling up the misconfig error), so I was deferring to them. I think your logic tracks though.

tests/rucio/common/test_didtype.py
```diff
@@ -12,10 +12,19 @@
...
+
 import pytest
 
 from rucio.common.didtype import DID
 from rucio.common.exception import DIDError
+from rucio.common.utils import ScopeExtractionAlgorithms
+
+
+def extract_scope(did, scopes, *args, **kwargs):
+    return did.split(":")[-1]
```
maxnoe: extract_scope needs to return `scope, name`, not just the scope

voetberg: Turns out this even wasn't being used, this `test_didtype` unitest doesn't expect any config to be present so it didn't even try to load this

lib/rucio/common/didtype.py
voetberg: Okay, multi-vo does require a bit of fallback (the way the tests are configured it seems to require the fallback scope to be used. The policy package `ScopeExtractionAlgorithm` **needs** a config to be present to be initialized (view - https://github.com/rucio/rucio/blob/6a4e74aa0da1fe3ee42ef26ba40653ebb18c9804/lib/rucio/common/plugins.py#L82 ), which is not always the case. The default needs to be called on fallback.

maxnoe: > The policy package ScopeExtractionAlgorithm needs a config to be present to be initialized

Yes, but the configuration of a policy package should be enough here, since it can define a default scope extraction mechanism.

lib/rucio/common/utils.py
```diff
@@ -663,11 +664,16 @@ def extract_scope(
...
-    scope_extraction_algorithms = ScopeExtractionAlgorithms(vo)
-    extract_scope_convention = config_get('common', 'extract_scope', False, None) or config_get('policy', 'extract_scope', False, None)
-    if extract_scope_convention is None or not ScopeExtractionAlgorithms.supports(extract_scope_convention):
-        extract_scope_convention = default_extract
-    return scope_extraction_algorithms.extract_scope(did, scopes, extract_scope_convention)
+    try:
+        scope_extraction_algorithms = ScopeExtractionAlgorithms(vo)
+        extract_scope_convention = config_get('common', 'extract_scope', False, None) or config_get('policy', 'extract_scope', False, None)
+        if extract_scope_convention is None or not ScopeExtractionAlgorithms.supports(extract_scope_convention):
+            extract_scope_convention = default_extract
```
maxnoe, line 671: I think this is not correct, since now policy packages can define default `extract_scope` algorithms, that shouldn't need to be explicitly registered. 

I might misread the logic here, but I think this does not cover the case of a policy package being defined via `policy.package` which defines a default `extract_scope`.

maxnoe: Ok, checked in our test setup, this works. In the case I mentioned above, `extract_scope_convention` resolves to `"def"`, which does correctly use the default defined in the package, not the global default of vanilla rucio.

voetberg: Yes, that's what is intended. You convinced yourself before I even had to try!

maxnoe:
@voetberg the failing test is the new download test added in the belle-ii test suite to protect against the regression for the original issue why we reverted this change originally (https://github.com/rucio/rucio/issues/8223).


```
[625](https://github.com/rucio/rucio/actions/runs/24997508478/job/73208374810?pr=8444#step:7:4626)
=================================== FAILURES ===================================
____________________________ test_download_datasets ____________________________
[gw0] linux -- Python 3.9.23 /opt/venv/bin/python
tests/test_belleii.py:162: in test_download_datasets
    download_client.download_dids(dids)
lib/rucio/client/downloadclient.py:444: in download_dids
    did_to_input_items, file_items_with_sources = self._resolve_and_merge_input_items(copy.deepcopy(items), sort=sort)
lib/rucio/client/downloadclient.py:1544: in _resolve_and_merge_input_items
    file_did = DID(file_item['did'])
lib/rucio/common/didtype.py:68: in __init__
    raise DIDError('Object has invalid format after construction: {}'.format(str(self)))
E   rucio.common.exception.DIDError: Error using DID type
E   Details: Object has invalid format after construction: other:mock:/belle/mock/cont_ac25b7e9f4a348aaa1a0ebf782858865/dataset_93b76d0d613146ecaabbd67d03c72f73/file_7e19ab72cc48480c93e7ff2a0912c5ab
```

I think there are two solutions:

- Change the belle-2 scope extraction algorithm to handle the case of an explicitly provided scope with the colon notation 
- Figure out where this double scope extraction and combination happens and eliminate it.

the ctao policy handles the colon case now:

https://gitlab.cta-observatory.org/cta-computing/dpps/bdms/bdms-rucio-policy/-/blob/main/src/bdms_rucio_policy/scope.py?ref_type=heads#L35

cc @cserf

voetberg:
The error in the belleii tests is: 
> E   Details: Object has invalid format after construction: other:mock:/belle/mock/cont_ac25b7e9f4a348aaa1a0ebf782858865/dataset_93b76d0d613146ecaabbd67d03c72f73/file_7e19ab72cc48480c93e7ff2a0912c5ab

This doesn't look like a valid belleii DID to me, so there may be something wrong with the construction of DIDs by the test constructor? Pinging  @cserf

voetberg:
Oh we seem to be debugging at the same time lol

maxnoe:
> This doesn't look like a valid belleii DID to me, so there may be something wrong with the construction of DIDs by the test constructor? 

But that isn't the DID that is created, it's is created somewhere in the middle of the downloadclient, that apparently calls `extract_scope` and then somewhere else prepends `<scope>:` to the lfn again and then the DID constructor is called again. I couldn't completely figure out why while debugging the original issue and it is also fixed by handling a well-formed, but scope-prepended name in our extract_scope.

I.e. make `extract_scope` also work for:

```
mock:/belle/mock/cont_ac25b7e9f4a348aaa1a0ebf782858865/dataset_93b76d0d613146ecaabbd67d03c72f73/file_7e19ab72cc48480c93e7ff2a0912c5ab
```

voetberg:
> > This doesn't look like a valid belleii DID to me, so there may be something wrong with the construction of DIDs by the test constructor?
> 
> But that isn't the DID that is created, it's is created somewhere in the middle of the downloadclient, that apparently calls `extract_scope` and then somewhere else prepends `<scope>:` to the lfn again and then the DID constructor is called again. I couldn't completely figure out why while debugging the original issue and it is also fixed by handling a well-formed, but scope-prepended name in our extract_scope.
> 
> I.e. make `extract_scope` also work for:
> 
> ```
> mock:/belle/mock/cont_ac25b7e9f4a348aaa1a0ebf782858865/dataset_93b76d0d613146ecaabbd67d03c72f73/file_7e19ab72cc48480c93e7ff2a0912c5ab
> ```

I'm not sure that is actually the DID being made by the test though - The DIDs given by dataset generator is `'did_scope': 'mock', 'did_name': 'file_4dc11c7a14f1480db69e91694db3df43'`, and give the full did as `'dataset_name': 'mockdataset_d327d58f0f79474ebb6f6236eda4d85e'`. 

What is this test actually supposed to show? I'm just getting myself confused by the monkeypatch to the name prefix.

maxnoe:
The did is:

```
/belle/mock/cont_ac25b7e9f4a348aaa1a0ebf782858865/dataset_93b76d0d613146ecaabbd67d03c72f73/file_7e19ab72cc48480c93e7ff2a0912c5ab
```
which is in scope `mock`.

The pattern for the test files is `/belle/{scope}/cont_{hash}/dataset_{hash}/file_{hash}`

voetberg:
Converting back to draft until we can get this debugged

PR #8495, opened 2026-04-16
Fix(Client): Introduce ScopeClient.list_scope_owners to keep list_scopes backwards compatible

### LINKED ISSUE #8487: Keep `ScopeClient.list_scopes` backwards compatible with previous versions
### Description

In https://github.com/rucio/rucio/pull/8222, there was a change to add an API endpoint for getting the scopes with their owners to close #7830.

There is now a strange mismatch between the python and the REST API.

The change added a new endpoint in the REST API, but the client now tries to call either the new endpoint or the old, depending on the server version. Also resulting in a return type that depends on the server version.

This makes it necessary in general code (like the RucioFileCatalogClient in DIRAC) to check return types of `ScopeClient.list_scopes`, which is breaks backwards compatibility in case the server is >=40.

I think a better approach would be to add a new method on the client (i.e. `ScopeClient.list_scope_owners()`) mirroring the new method on the server, that raises an exception in cases the server is too old to support it.

### Motivation

Downstream usage of `list_scopes`, return type of `list_scopes` depends on server version.

### Change

_No response_

maxnoe:
Closes #8487

## Checklist
- [X] Created issue: #8487
- [x] Added tests
- [ ] I understand that stale (failing tests, author not answering) PRs will be closed promptly

<details>

```
- **Quality**: Agree 🟢
- **Testing**: Agree 🟢
- **Documentation**: Agree 🟢

# Notes for merger



```
</details>

### DIFF
--- a/lib/rucio/cli/bin_legacy/rucio.py
+++ b/lib/rucio/cli/bin_legacy/rucio.py
@@ -505,29 +505,36 @@ def list_scopes(args, client, logger, console, spinner):
 
     if args.account:
         scopes = client.list_scopes_for_account(args.account)
+        with_owner = False
     else:
-        scopes = client.list_scopes()
+        scopes = client.list_scope_owners()
+        with_owner = True
+
     if (cli_config == 'rich') and (not args.csv):
         if len(scopes) == 0:
             spinner.stop()
-        elif isinstance(scopes[0], dict):
-            table = generate_table(scopes, headers=['SCOPE', "ACCOUNT"], col_alignments=['left'])
+        elif not with_owner:
+            scopes = [[scope] for scope in sorted(scopes)]
+            table = generate_table(scopes, headers=['SCOPE'], col_alignments=['left'])
             spinner.stop()
             print_output(table, console=console, no_pager=args.no_pager)
         else:
-            scopes = [[scope] for scope in sorted(scopes)]
-            table = generate_table(scopes, headers=['SCOPE'], col_alignments=['left'])
+            scopes = [[s['scope'], s['account']] for s in scopes]
+            table = generate_table(scopes, headers=['SCOPE', "ACCOUNT"], col_alignments=['left'])
             spinner.stop()
             print_output(table, console=console, no_pager=args.no_pager)
     else:
         if len(scopes) == 0:
             pass
-        elif isinstance(scopes[0], str):  # TODO: Backwards compatibility - remove in v40 issue #8125
-            for scope in scopes:
-                print(scope)
         elif args.csv:
             for scope in scopes:
-                print(scope['scope'])
+                if not with_owner:
+                    print(scope)
+                else:
+                    print(f"{scope['scope']},{scope['account']}")
+        elif not with_owner:
+            for scope in scopes:
+                print(scope)
         else:
             scopes = [[s['scope'], s['account']] for s in scopes]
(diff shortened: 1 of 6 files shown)


lib/rucio/cli/bin_legacy/rucio.py
```diff
@@ -503,31 +503,45 @@ def list_scopes(args, client, logger, console, spinner):
...
+    rucio_39_compat = False
     if args.account:
         scopes = client.list_scopes_for_account(args.account)
     else:
-        scopes = client.list_scopes()
+        try:
+            # new api endpoint introduced in rucio 40
+            scopes = client.list_scope_owners()
+            scopes = sorted(scopes, key=lambda s: s["scope"])
+        except RucioException:
```
maxnoe: @voetberg @bari12 Should we add a particular exception here so that we can be more specific and do not catch the generic `RucioException` here?

lib/rucio/cli/bin_legacy/rucio.py
```diff
@@ -503,31 +503,45 @@ def list_scopes(args, client, logger, console, spinner):
         spinner.update(status='Fetching scopes')
         spinner.start()
 
+    rucio_39_compat = False
```
voetberg: To avoid that try-except you can instead define this using client.ping(), which should return the server version. I assume efficiency wise it's all about the same, but defining it ahead of time is easier to read.

voetberg: Although I may be misreading your logic, in which case you can use `client.user_agent` to get the client version

maxnoe: The server version here is relevant. Using try except avoids an extra request for the "happy" path.

lib/rucio/cli/bin_legacy/rucio.py
voetberg: I don't think there is anything in use (The only one I could see looking at our current ones is the unused "ClientParameterMismatch", but I don't think that's a perfect fit.)

If we want to make this a general policy on how to handle mismatches, we may want to add a "ClientServerIncompatiblity" error or something of the sort. Either way, I think something more specific than RucioException is a good idea.

lib/rucio/cli/bin_legacy/rucio.py
voetberg: I mean, you need to make the request, even if it returns an error, to use the try-except, which is why I think it all comes out in the wash

lib/rucio/cli/bin_legacy/rucio.py
maxnoe: I was confusing myself again with the compatibility policy, the branches / try except here are not needed since only the server needs to support older clients not the other way around.

bari12 (approved):

bari12 (approved):
- **Quality**: Agree 🟢
- **Security**: Agree 🟢
- **Testing**: Agree 🟢
- **Documentation**: Agree 🟢

# Notes for merger

bari12:
@voetberg We should add the owner to the `rucio scope show` call as well. Actually, I find it most important there.
One could make the argument that the owner should not be listed in `rucio scope list` at all, and keep that interface simple and just list the scopes, more detailed information is then just added to the `rucio scope show` call. 
After reflecting a bit about that I would say we leave the owner in, I think it provides enough value there, but it's a tricky call to make in terms of consistency with other `list` calls.

PR #8516, opened 2026-04-23
permissions/generic: check scopes in upload permissions 

### LINKED ISSUE #8333: Fix default permissions for uploading (perm_add_replica and perm_update_replicas)

jouvin:

Fixes #8333

This PR is a new version of #8289 that was closed beause it was inactive and that we didn't manage to reopen. Should be considered as a follow-up, implementing the suggestions by @maxnoe and fixing issues reported by Ruff.

It is my attempt to fix the problem that scopes are not passed to the `has_permission()` function when adding (or deleting) a replica when they are supposed to be the Rucio object on which permissions are applied when uploading or downloading DIDs. 

It is necessary to build the scope list in `gateway/replica.py` and pass it in the `kwargs` argument to the permission module (`core/permission/generic.py` for example). The alternative would be to require the permission module to look in all replicas passed to get the scope and to what is done once in `gateway/replica.py` so IMO it is less optimal. And anyway scopes are not passed at all in `update_replicas_states()` in `gateway/replica.py` so a modification of this module would be necessary anyway...

I am not enough a Rucio expert to ensure that there is no side effect to this modification but it's my feeling: nothing passed previously is modified, just an entry `scopes` is added in the `kwargs` dict so it should be harmless. And the resulting default permission of this PR is a basic configuration that you can use to test your Rucio installation, including upload operations, without writing a permission policy as your first step when configuring a new instance. And production instance probably all have their own policy file so they will not be impacted by the updated `generic.py` or `generic_multi_vo.py`. I think that exposing scopes to replica-related permission functions could also benefit to some production instance that had to work around the absence of scopes with account attributes....

It also requires #8236 to fully address problems with S3 RSEs.

## Checklist
- [x] Created issue: #8333
- [ ] Added tests

<details>

```
- **Quality**: Agree 🟢
- **Testing**: Agree 🟢
- **Documentation**: Agree 🟢

# Notes for merger



```
</details>

### DIFF
--- a/lib/rucio/core/permission/generic.py
+++ b/lib/rucio/core/permission/generic.py
@@ -14,13 +14,13 @@
 
 from typing import TYPE_CHECKING, Any
 
-import rucio.core.scope
 from rucio.common.constants import RseAttr
 from rucio.core.account import has_account_attribute, list_account_attributes
 from rucio.core.identity import exist_identity_account
 from rucio.core.lifetime_exception import list_exceptions
 from rucio.core.rse import list_rse_attributes
 from rucio.core.rse_expression_parser import parse_expression
+from rucio.core.scope import is_scope_owner
 from rucio.db.sqla.constants import IdentityType
 
 if TYPE_CHECKING:
@@ -298,7 +298,7 @@ def perm_update_scope(issuer: "InternalAccount", kwargs: dict[str, Any], session
     """
     return _is_root(issuer) \
         or has_account_attribute(account=issuer, key='admin', session=session) \
-        or rucio.core.scope.is_scope_owner(scope=kwargs['scope'], account=issuer, session=session)
+        or is_scope_owner(scope=kwargs['scope'], account=issuer, session=session)
 
 
 def perm_get_auth_token_user_pass(issuer: "InternalAccount", kwargs: dict[str, Any], session: "Session") -> bool:
@@ -413,7 +413,7 @@ def perm_add_did(issuer: "InternalAccount", kwargs: dict[str, Any], session: "Se
 
     return _is_root(issuer)\
         or has_account_attribute(account=issuer, key='admin', session=session)\
-        or rucio.core.scope.is_scope_owner(scope=kwargs['scope'], account=issuer, session=session)\
+        or is_scope_owner(scope=kwargs['scope'], account=issuer, session=session)\
         or kwargs['scope'].external == 'mock'
 
 
@@ -441,7 +441,7 @@ def perm_attach_dids(issuer: "InternalAccount", kwargs: dict[str, Any], session:
     """
     return _is_root(issuer)\
         or has_account_attribute(account=issuer, key='admin', session=session)\
-        or rucio.core.scope.is_scope_owner(scope=kwargs['scope'], account=issuer, session=session)\
+        or is_scope_owner(scope=kwargs['scope'], account=issuer, session=session)\
         or kwargs['scope'].external == 'mock'
 
 
@@ -461,7 +461,7 @@ def perm_attach_dids_to_dids(issuer: "InternalAccount", kwargs: dict[str, Any], 
         scopes = [did['scope'] for did in attachments]
         scopes = list(set(scopes))
         for scope in scopes:
-            if not rucio.core.scope.is_scope_owner(scope, issuer, session=session):
+            if not is_scope_owner(scope, issuer, session=session):
                 return False
         return True
 
@@ -477,7 +477,7 @@ def perm_create_did_sample(issuer: "InternalAccount", kwargs: dict[str, Any], se
     """
     return _is_root(issuer)\
         or has_account_attribute(account=issuer, key='admin', session=session)\
-        or rucio.core.scope.is_scope_owner(scope=kwargs['scope'], account=issuer, session=session)\
+        or is_scope_owner(scope=kwargs['scope'], account=issuer, session=session)\
         or kwargs['scope'].external == 'mock'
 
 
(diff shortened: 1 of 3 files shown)


lib/rucio/core/permission/generic.py
```diff
@@ -706,17 +706,30 @@ def perm_add_replicas(issuer: "InternalAccount", kwargs: dict[str, Any], session
...
     :returns: True if account is allowed, otherwise False
     """
-    return str(kwargs.get('rse', '')).endswith('SCRATCHDISK')\
-        or str(kwargs.get('rse', '')).endswith('USERDISK')\
-        or str(kwargs.get('rse', '')).endswith('MOCK')\
-        or str(kwargs.get('rse', '')).endswith('LOCALGROUPDISK')\
-        or _is_root(issuer)\
-        or has_account_attribute(account=issuer, key='admin', session=session)
+
+    SPECIAL_RSE_SUFFIXES = ['MOCK', 'SCRATCHDISK']
```
maxnoe: Why remove `USERDISK` and `LOCALGROUPDISK`?

lib/rucio/core/permission/generic.py
```diff
@@ -706,17 +706,30 @@ def perm_add_replicas(issuer: "InternalAccount", kwargs: dict[str, Any], session
...
+
+    if _is_root(issuer):
+        return True
+    elif has_account_attribute(account=issuer, key='admin', session=session):
```
maxnoe: This is a minor stylistic thing, but it is a bit weird to see early returns with `elif`.

lib/rucio/core/permission/generic.py
jouvin: A mistake of mine... I'll fix it. Thanks for noticing.

lib/rucio/core/permission/generic.py
jouvin: I am not sure to see what is the exact difference between early returns with `elif` rather than `if`, I thought it was clearer to express that only one can happen but I have not issue to revert to your initial proposal!

maxnoe: There is no functional difference, that's why I wrote "minor stylistic thing", but it creates a connection where there really is none anymore, because of the return, so If I wrote it, I'd write:

```
if _is_root(issuer):
    return True

if has_account_attribute(...):
    return True
...
```

but again, this is a style question, there is no difference in behavior.

jouvin: Fine with me, always happy to learn suggestions ffom others! Thanks. It's implemented.

jouvin:
I just fixed a few typos preventing the PR to work... We had an infrastructure problem that made our K8s infrastructure unavailable in the last 2 days. I deployed successfully this last version of the PR and confirm that it works as expected (but it requires #8236 to fully address problems with S3 RSEs) for upload.

jouvin:
@maxnoe @bziemons I tried to fix all check erros... but there is one left (https://github.com/rucio/rucio/actions/runs/24902196964/job/72981820468?pr=8516) where I am not sure what I should do. If I follow the suggestion (moving an import to a `TYPE_CHECKING` section), it breaks the code...

jouvin:
@bziemons please could you avoid that this PR closes automatically...

maxnoe:
>  If I follow the suggestion (moving an import to a TYPE_CHECKING section), it breaks the code...

I think the correct solution here is to move the import as suggested:

```
if TYPE_CHECKING:
```

and then use a string, not the "naked" class for the type annotation:

```
def foo(..., session: "Session"):
   ...
```

note the quotes around `"Session"`

bziemons:
@jouvin could you please update the PR description to use the PR template?

jouvin:
> @jouvin could you please update the PR description to use the PR template?

I modified the descriptioon. Is it what you expected... It was not totally obvious to me what description is expected and where it should be inserted. Let me know if I need to adjust something.

jouvin:
I hope I fixed the remaining error related to type checking, following @maxnoe suggestion. I'm waiting for the checks to run to confirm it (I don't see errors reported on the commit).

bari12:
I am re-opening this, it should not have been closed by the bot. I am checking why the bot closed it.

jouvin:
Sorry for being silent in the last months... What can be done to move this PR forward?

jouvin:
I missed the conflict, I'll fix it.

jouvin:
I updated the PR based on the master today (f14072c74ee84c3a77cfff6f4e523b5541a18553)... Tests are in progress but look good, except one that failed to run (container not found) so I guess it is not related to the changes.

I have not tested it yet on my Rucio instance, currently running 40.0.0, as I need to upgrade it first  to something more recent to have the conflicting code.

PR #8524, opened 2026-04-28
Feat add bidirectional distance #8504 

### LINKED ISSUE #8504: Add `--bidirectional` flags to RSE distance operations in CLI
Should be included in add, update, remove

Would change source and target in one operation

panta-123:
Added bidrectional support in rse distance add/update/remove . 

## Checklist
- [X] Created issue: #8504 
- [X] Added tests
- [X] Picked a reviewer after submission (Leave empty, if unclear)
- [X] I understand that stale (failing tests, author not answering) PRs will be closed promptly

<details>

```
- **Quality**: Agree 🟢
- **Testing**: Agree 🟢
- **Documentation**: Agree 🟢

# Notes for merger



```
</details>

### DIFF
--- a/lib/rucio/cli/rse.py
+++ b/lib/rucio/cli/rse.py
@@ -259,35 +259,97 @@ def distance_show(ctx: click.Context, source_rse: str, destination_rse: str) -> 
 @distance.command("add")
 @click.argument("source-rse")
 @click.argument("destination-rse")
-@click.option("--distance", default=1, type=int, help="Relative distance between RSEs")
+@click.option(
+    "--distance",
+    default=1,
+    type=int,
+    help="Relative distance between RSEs"
+)
+@click.option(
+    "--bidirectional",
+    is_flag=True,
+    default=False,
+    help="If set, also creates the reverse distance (destination → source)."
+)
 @click.pass_context
-def distance_add(ctx: click.Context, source_rse: str, destination_rse: str, distance: int) -> None:
+def distance_add(
+    ctx: click.Context,
+    source_rse: str,
+    destination_rse: str,
+    distance: int,
+    bidirectional: bool
+) -> None:
     """Create a new link from SOURCE-RSE to DESTINATION-RSE with a distance"""
     params = {'distance': distance}
-    ctx.obj.client.add_distance(source_rse, destination_rse, params)
-    print(f'Set distance from {source_rse} to {destination_rse} to {distance}')
+    ctx.obj.client.add_distance(
+        source_rse,
+        destination_rse,
+        parameters=params,
+        bidirectional=bidirectional
+    )
+
+    if bidirectional:
+        print(f"Set distances between {source_rse} <-> {destination_rse} to {distance}")
+    else:
+        print(f"Set distance from {source_rse} -> {destination_rse} to {distance}")
 
 
 @distance.command("remove")
 @click.argument("source-rse")
 @click.argument("destination-rse")
+@click.option(
+    "--bidirectional",
+    is_flag=True,
+    default=False,
+    help="If set, also removes the reverse distance (destination → source)."
+)
 @click.pass_context
-def distance_remove(ctx: click.Context, source_rse: str, destination_rse: str) -> None:
+def distance_remove(ctx: click.Context, source_rse: str, destination_rse: str, bidirectional: bool = False) -> None:
     """Un-link SOURCE-RSE from DESTINATION-RSE by removing the distance between them"""
-    ctx.obj.client.delete_distance(source_rse, destination_rse)
-    print(f'Deleted distance information from {source_rse} to {destination_rse}.')
+    ctx.obj.client.delete_distance(
+        source_rse,
+        destination_rse,
+        bidirectional=bidirectional)
+    if bidirectional:
+        print(f"Deleted distances between {source_rse} <-> {destination_rse}")
+    else:
+        print(f"Deleted distance from {source_rse} -> {destination_rse}")
 
 
 @distance.command("update")
 @click.argument("source-rse")
 @click.argument("destination-rse")
-@click.option("--distance", type=int, help="Relative distance between RSEs", required=True)
+@click.option(
+    "--distance",
+    type=int,
+    help="Relative distance between RSEs",
+    required=True
+)
+@click.option(
+    "--bidirectional",
+    is_flag=True,
+    default=False,
+    help="If set, also updates the reverse distance (destination → source)."
+)
 @click.pass_context
-def distance_update(ctx: click.Context, source_rse: str, destination_rse: str, distance: int) -> None:
+def distance_update(
+    ctx: click.Context,
+    source_rse: str,
+    destination_rse: str,
+    distance: int,
+    bidirectional: bool
+) -> None:
     """Update the existing distance from SOURCE-RSE to DESTINATION-RSE"""
     params = {"distance": distance}
-    ctx.obj.client.update_distance(source_rse, destination_rse, params)
-    print(f'Update distance information from {source_rse} to {destination_rse}:\n - Distance set to {params["distance"]}')
+    ctx.obj.client.update_distance(
+        source_rse,
+        destination_rse,
+        parameters=params,
+        bidirectional=bidirectional)
+    if bidirectional:
+        print(f"Updated distances between {source_rse} <-> {destination_rse} to {distance}")
+    else:
+        print(f"Updated distance from {source_rse} -> {destination_rse} to {distance}")
 
 
 @rse.group()
--- a/lib/rucio/client/rseclient.py
+++ b/lib/rucio/client/rseclient.py
@@ -1385,10 +1385,11 @@ class RSEClient(BaseClient):
         raise exc_cls(exc_msg)
 
     def add_distance(
-            self,
-            source: str,
-            destination: str,
-            parameters: dict[str, int]
+        self,
+        source: str,
+        destination: str,
+        parameters: dict[str, int],
+        bidirectional: bool = False
     ) -> Literal[True]:
         """
         Add a distance between two RSEs.
@@ -1404,6 +1405,8 @@ class RSEClient(BaseClient):
             The destination RSE name.
         parameters :
             Dicionary in the format {"distance": int}.
+        bidirectional:
+            If True, also adds the distance from dest to src.
 
         Returns
         -------
@@ -1429,19 +1432,21 @@ class RSEClient(BaseClient):
         path = [self.RSE_BASEURL, source, 'distances', destination]
         path = '/'.join(path)
         url = build_url(choice(self.list_hosts), path=path)
+        parameters["bidirectional"] = bidirectional
         r = self._send_request(url, method=HTTPMethod.POST, data=dumps(parameters))
         if r.status_code == codes.created:
             return True
         exc_cls, exc_msg = self._get_exception(headers=r.headers,
-                                               status_code=r.status_code,
-                                               data=r.content)
+                                            status_code=r.status_code,
+                                            data=r.content)
         raise exc_cls(exc_msg)
 
     def update_distance(
-            self,
-            source: str,
-            destination: str,
-            parameters: dict[str, int]
+        self,
+        source: str,
+        destination: str,
+        parameters: dict[str, int],
+        bidirectional: bool = False
     ) -> Literal[True]:
         """
         Update distances between RSEs.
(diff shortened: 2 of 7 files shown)


tests/test_rse.py
```diff
@@ -1490,25 +1490,42 @@ def test_rsemgr_possible_protocols(self):
...
         rucio_client.update_distance(source=source,
                                      destination=destination,
-                                     parameters={'distance': 0})
+                                     parameters={'distance': 5},
+                                     bidirectional=bidirectional)
 
-        for distance in rucio_client.get_distance(source=source, destination=destination):
-            print(distance)
-            assert distance['distance'] == 0
+        if bidirectional:
```
voetberg: I think because you have these extra bidirection checks, it's worth just separating them out into two different tests even if there is a bit of repeated code. It would make it easier to debug in the long turn, it's harder to follow failures inside the if statements

tests/test_bin_rucio.py
voetberg, line 1: Let's leave the tests for the legacy client untouched atm, I plan to port them over to the new format as a whole when we retire it, to ensure it all still works for the time being. Could you instead write the bidirectional CLI test in `test_cli_client_structure`? Sorry it's confusing.

lib/rucio/web/rest/flaskapi/v1/rses.py
```diff
@@ -2063,10 +2079,12 @@ def delete(self, source, destination):
           406:
             description: "Not acceptable"
         """
+        bidirectional = request.args.get('bidirectional', 'false').lower() == 'true'
```
voetberg: Please use "param_get_bool" instead

lib/rucio/web/rest/flaskapi/v1/rses.py
```diff
@@ -2013,11 +2021,13 @@ def put(self, source, destination):
         if distance is None:
             distance = param_get(parameters, 'ranking', default=None)
 
+        bidirectional = param_get(parameters, 'bidirectional', default=False)
```
voetberg: This should also be "param_get_bool"

lib/rucio/web/rest/flaskapi/v1/rses.py
```diff
@@ -1941,12 +1944,14 @@ def post(self, source, destination):
         distance = param_get(parameters, 'distance', default=None)
         if distance is None:
             distance = param_get(parameters, 'ranking', default=None)
+        bidirectional = param_get(parameters, 'bidirectional', default=False)
```
voetberg: Use "param_get_bool"

tests/test_rse.py
voetberg, line 1: This should also have some tests on core and gateway with the new parameter, just to be safe

lib/rucio/gateway/rse.py
```diff
@@ -478,13 +486,20 @@ def add_distance(source, destination, issuer, vo=DEFAULT_VO, distance=None):
...
         except exception.Duplicate:
             # use source and destination RSE names
             raise exception.Duplicate('Distance from %s to %s already exists!' % (source, destination))
 
 
-def update_distance(source, destination, distance, issuer, vo=DEFAULT_VO):
+def update_distance(
+    source: str,
+    destination: str,
+    issuer: str,
```
voetberg, line 498: Isn't the issuer an InternalAccount?

lib/rucio/client/rseclient.py
```diff
@@ -905,12 +920,20 @@ def update_distance(
...
-            A dictionary with property
+            A dictionary with 'distance' key and int value.
+        bidirectional :
+            If True, also updates the distance from dest to src.
         """
         path = [self.RSE_BASEURL, source, 'distances', destination]
         path = '/'.join(path)
         url = build_url(choice(self.list_hosts), path=path)
-        r = self._send_request(url, method=HTTPMethod.PUT, data=dumps(parameters))
+        distance = parameters["distance"]
```
voetberg: For the sake of not touching anything other than the bidirectional argument, `bidirectional` should probably just be added onto the parameters dict rather than making a new `data` dict

lib/rucio/cli/rse.py
```diff
@@ -259,35 +259,98 @@ def distance_show(ctx: click.Context, source_rse: str, destination_rse: str) ->
...
+def distance_remove(ctx: click.Context, source_rse: str, destination_rse: str, bidirectional: bool = False) -> None:
     """Un-link SOURCE-RSE from DESTINATION-RSE by removing the distance between them"""
-    ctx.obj.client.delete_distance(source_rse, destination_rse)
-    print(f'Deleted distance information from {source_rse} to {destination_rse}.')
+    ctx.obj.client.delete_distance(
+        source_rse,
+        destination_rse,
+        bidirectional=bidirectional)
+    if bidirectional:
+        print(f"Deleted distance between {source_rse} <-> {destination_rse}")
```
voetberg: Pedantic comment but it's probably "distances"? Because technically there were two distances deleted in the table

lib/rucio/cli/rse.py
```diff
@@ -259,35 +259,98 @@ def distance_show(ctx: click.Context, source_rse: str, destination_rse: str) ->
 @distance.command("add")
 @click.argument("source-rse")
 @click.argument("destination-rse")
-@click.option("--distance", default=1, type=int, help="Relative distance between RSEs")
+@click.option(
+    "--distance",
+    default=1,
+    type=int,
+    show_default=True,
```
voetberg: Commit cleanliness comment - `show_default` change should probably be a separate docs commit, because it just changes user-facing docs

voetberg (changes requested):
Few minor things. Unsure about the core implementation yet though, I'd like to have a core dev go over that

lib/rucio/web/rest/flaskapi/v1/rses.py
voetberg, line 1: Strictly speaking, both the restAPI is its own component, and following atomic commits, it's be better if changes to it were its own commit

tests/test_rse.py
panta-123: Done.

tests/test_rse.py
panta-123: Added tests for core.

lib/rucio/gateway/rse.py
panta-123: I think its str. issuer string is used to check permission.
For account we map to InternalAccount.

lib/rucio/cli/rse.py
panta-123: I decided not to add show_default in this PR.

lib/rucio/web/rest/flaskapi/v1/rses.py
panta-123: Done. Commit seperated

tests/test_bin_rucio.py
panta-123: did that too.

lib/rucio/client/rseclient.py
```diff
@@ -866,34 +866,44 @@ def delete_rse_limits(self, rse: str, name: str) -> Literal[True]:
...
-        :param destination: The destination.
-        :param parameters: A dictionary with property.
+        Parameters
+        ----------
+        source :
+            The source RSE.
+        destination :
+            The destination RSE.
+        parameters :
+            A dictionary with 'distance' key and int value.
```
voetberg: This should be an isolated documentation change commit

lib/rucio/client/rseclient.py
```diff
@@ -905,11 +915,14 @@ def update_distance(
         destination :
             The destination RSE.
         parameters :
-            A dictionary with property
+            A dictionary with 'distance' key and int value.
```
voetberg: Also should be a different documentation commit

lib/rucio/client/rseclient.py
panta-123: Done added a docs commit in between.

lib/rucio/client/rseclient.py
panta-123: Done added a docs commit in between.

voetberg (approved):
Everything looks good to me and the tests seem sufficient. I am unsure if there is something I could be missing in the sql calls, but it seems fine. 

- **Confidence in scope**:  Medium 🟡
- **Quality**: Agree 🟢
- **Security**: Agree 🟢 
- **Testing**: Agree 🟢
- **Documentation**: Agree 🟢

# Notes for merger

lib/rucio/core/distance.py
```diff
@@ -28,19 +28,24 @@
...
+    pairs = [(src_rse_id, dest_rse_id)]
+    if bidirectional:
+        pairs.append((dest_rse_id, src_rse_id))
 
     try:
-        new_distance = Distance(src_rse_id=src_rse_id, dest_rse_id=dest_rse_id, distance=distance)
-        new_distance.save(session=session)
+        for src, dest in pairs:
+            new_distance = Distance(src_rse_id=src, dest_rse_id=dest, distance=distance)
+            new_distance.save(session=session)
```
bari12, line 48: I think this won't work. If e.g. the reverse distance B->A already exists, the pair will insert:
A->B
B->A
and raise an Integrity (Duplicate) on B->A

tests/test_rse.py
```diff
@@ -253,6 +254,56 @@ def test_empty_rse(self, vo):
...
+        rse2 = rse_name_generator()
+        rse1_id = add_rse(rse1, vo=vo)
+        rse2_id = add_rse(rse2, vo=vo)
+        assert rse_exists(rse=rse1, vo=vo)
+        assert rse_exists(rse=rse2, vo=vo)
+        add_distance(src_rse_id=rse1_id, dest_rse_id=rse2_id, distance=1, bidirectional=True)
+
+        delete_distances(src_rse_id=rse1_id, dest_rse_id=rse2_id, bidirectional=True)
+        assert len(get_distances(src_rse_id=rse1_id, dest_rse_id=rse2_id)) == 0
+        assert len(get_distances(src_rse_id=rse1_id, dest_rse_id=rse2_id)) == 0
```
bari12: I think this should be reverse. Rse2 -> rse1

bari12 (changes requested):

lib/rucio/core/distance.py
panta-123: But doesn't the decorator `transactional_session` roll back the transaction if integrity error is raised. so if B->A raises Integrity then A->B is rolled back.

bari12: Kind of, the integrity error just invalidates the entire transaction, similar to a roll back. 
But I would say, that's not what the user expects? To me, the user would expect that both directions will be set?

panta-123: @bari12 , should I overwrite if distance already exists?

bari12: I think so. Add the moment this command is under the `add` command, which would make you think it should not, but we are planning to change this to `set` which should overwrite it in this case.

panta-123: okay, Done.

tests/test_rse.py
panta-123: It should and changed .

lib/rucio/core/distance.py
```diff
@@ -28,21 +28,24 @@
...
+    if bidirectional:
+        pairs.append((dest_rse_id, src_rse_id))
 
     try:
-        new_distance = Distance(src_rse_id=src_rse_id, dest_rse_id=dest_rse_id, distance=distance)
-        new_distance.save(session=session)
-    except IntegrityError:
-        raise exception.Duplicate()
+        for src, dest in pairs:
+            new_distance = Distance(src_rse_id=src, dest_rse_id=dest, distance=distance)
```
bari12, line 47: I don't think we should do it like that. I see why you did it based on our discussion, but implementing it like this really changes the backwards compatibility, as it changes the behaviour of a non bi-directional add.

I see why you did it like this after all my discussion about the `set` vs `add`, but I still don't think we should do it like that breaking the backwards compatibility.

Maybe the best thing to do is to actually not overwrite, and we keep this for the future for a new set command. Thus, also for the bidirectional, don't do the merge, but let it fail.

bari12: Sorry for the back and forth.

bari12 (changes requested):

lib/rucio/core/distance.py
panta-123: Done. I have put it back. So that to let it fail on duplicate.
Added test: https://github.com/rucio/rucio/pull/8524/changes#diff-fd8a0f818c1869331fd0c207e95ee5af06aecc573c24edeeb9edc99e852f888eR270-R275
to show that roll back if duplicate is correct.

bari12 (approved):
- **Quality**: Agree 🟢
- **Security**: Agree 🟢
- **Testing**: Agree 🟢
- **Documentation**: Agree 🟢

# Notes for merger

PR #8546, opened 2026-05-18
Disable core config cache in testing

### LINKED ISSUE #8490: core_config_mock can leave retrieved entries in config cache
As part of investigating flaky tests @maany and I have found that the use of `core_config_mock` can lead to some of the values that should be temporary to be retrieved from the cache after the test. This can lead to other tests failing due to unforeseen configuration, if they intend to use default values.

bziemons:
Fixes #8490

I had to touch the automatix test and add the vo mappings to the core_config_mock, because the multi-VO test was failing due to relying on the vo-map being in the configuration cache.

## Checklist
- [x] Created issue: #8490
- [ ] Added tests

<details>

```
- **Quality**: Agree 🟢
- **Testing**: Agree 🟢
- **Documentation**: Agree 🟢

# Notes for merger



```
</details>

### DIFF
--- a/lib/rucio/tests/common_server.py
+++ b/lib/rucio/tests/common_server.py
@@ -134,7 +134,8 @@ def cleanup_db_deps(
 def reset_config_table() -> None:
     """ Clear the config table and install any default entries needed for the tests.
     """
-    db_session = get_session()()
+    session_scoped = get_session()
+    db_session = session_scoped()
     db_session.query(models.Config).delete()
     set_config_defaults(session=db_session)
     db_session.commit()
--- a/lib/rucio/tests/common_server.py
+++ b/lib/rucio/tests/common_server.py
@@ -134,11 +134,15 @@ def cleanup_db_deps(
 def reset_config_table() -> None:
     """ Clear the config table and install any default entries needed for the tests.
     """
-    db_session = get_session()
+    db_session = get_session()()
     db_session.query(models.Config).delete()
+    set_config_defaults(session=db_session)
     db_session.commit()
-    core_config.set("vo-map", "testvo1", "tst")
-    core_config.set("vo-map", "testvo2", "ts2")
+
+
+def set_config_defaults(session: "Session") -> None:
+    core_config.set(section="vo-map", option="testvo1", value="tst", session=session)
+    core_config.set(section="vo-map", option="testvo2", value="ts2", session=session)
 
 
 def get_vo() -> str:
--- a/lib/rucio/core/config.py
+++ b/lib/rucio/core/config.py
@@ -12,9 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import os
 from typing import TYPE_CHECKING, Any, Optional, TypeVar
 
-from dogpile.cache.api import NoValue
+from dogpile.cache.api import NoValue, NO_VALUE
 from sqlalchemy import and_, delete, func, select, update
 
 from rucio.common.cache import CacheKey, MemcacheRegion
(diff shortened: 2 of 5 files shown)


Geogouz:
Just one suggested typo change which is minor.  Namely, the commit message says `RUCIO_DISABLE_CACHE_READ`, but the actual code uses `RUCIO_CONFIG_DISABLE_CACHE_READ`.

Also there are couple edge cases where we might have a problem but they sound quite extreme to me to care, so I am just mentioning those here in case in the far future these may give a hint.  

1) If `rucio.core.config` is somehow weirdly imported before `RUCIO_CONFIG_DISABLE_CACHE_READ` is set, the module has already computed `DISABLE_CACHE_READ=False` and will keep reading from cache in that process.  

2) Also, since `core_config_mock` always adds `vo-map.testvo1 = tst`
and `vo-map.testvo2 = ts2`, this fixture should not be used for tests that actually need `vo-map` to be absent or to override those default keys. Again too extreme.  

In general LGTM.

bziemons:
> Just one suggested typo change which is minor.  Namely, the commit message says RUCIO_DISABLE_CACHE_READ, but the actual code uses RUCIO_CONFIG_DISABLE_CACHE_READ.

Well spotted. I did change the name, because the former was too generic. I will update the commit.

> If rucio.core.config is somehow weirdly imported before RUCIO_CONFIG_DISABLE_CACHE_READ is set, the module has already computed DISABLE_CACHE_READ=False and will keep reading from cache in that process.  

I thought about this and decided that environment variables are usually set before running a program (python in this case) and this is not an actual issue. The place where I set it here makes sure that all pytest runs are without the config cache. Adding it to tools/pytest.sh (my initial thought) would have excluded any manual pytest runs or IDE integrations that run pytest directly.

> Also, since core_config_mock always adds vo-map.testvo1 = tst
and vo-map.testvo2 = ts2, this fixture should not be used for tests that actually need vo-map to be absent or to override those default keys. Again too extreme.  

These are the actual (unchangeable) defaults for all tests, though. Meaning they were globally applied, see the code where they are set. Since the config cache was still providing these values as well, no test so far has been run without these values. If a future test wants to specifically override or remove them, it can be done with the config mock now and without affecting other tests. This wasn't even possible before, so I think this is actually a positive change in that regard.

Geogouz:
- **Quality**: Agree 🟢
- **Security**: Agree 🟢
- **Testing**: Agree 🟢
- **Documentation**: Agree 🟢

# Notes for merger

Geogouz (dismissed):
LGTM

lib/rucio/tests/common_server.py
```diff
@@ -134,11 +134,15 @@ def cleanup_db_deps(
 def reset_config_table() -> None:
     """ Clear the config table and install any default entries needed for the tests.
     """
-    db_session = get_session()
+    db_session = get_session()()
```
dchristidis: This change gives me pause.

In a private exchange you explained that this is necessary due to type checking and pointed to a prior use of this in: https://github.com/rucio/rucio/blob/309cd90c8737e5f8d96cfa7e37e211e91eb0722c/tests/conftest.py#L654
As such, it should not block the PR. But I cannot help but wonder if there’s a better way here.

In our session decorators we have:
https://github.com/rucio/rucio/blob/cd0ba94466c907e8ced690a5d2b5d08f653afafe/lib/rucio/db/sqla/session.py#L405-L406
Maybe this is preferable? Maybe `get_session()` is a misleading name? Maybe the type hints of `set_config_defaults()` should be modified to accept `scoped_session`?

bziemons: note: the type checking issue is masked by decorator usage for most of our functions that take session as an argument as explained in https://github.com/rucio/rucio/issues/6986

bziemons: I do prefer adding an intermediary variable for getting the session object (which has the correct type for our type annotations). Will add a refactor commit

dchristidis (approved):

PR #7373, opened 2025-01-29
rest: handle bulk traces #7368

maany:
Fix #7368 

This PR enhances the POST /traces/ endpoint to accept a list of trace objects in the request body while maintaining backward compatibility with the existing single-trace format.

Enhancements to trace handling:

* [`lib/rucio/web/rest/flaskapi/v1/traces.py`](diffhunk://#diff-079a0ba14236814e622ee6688ffcbfc312582d32e3a4dde97a9a697f71df6626R36-R68): Introduced the `__handle_payload_item` method to process individual trace payload items, adding trace information such as timestamps, client IP, and a unique trace ID.

Support for bulk trace submissions:

* [`lib/rucio/web/rest/flaskapi/v1/traces.py`](diffhunk://#diff-079a0ba14236814e622ee6688ffcbfc312582d32e3a4dde97a9a697f71df6626L60-R124): Modified the `post` method to handle both single and multiple trace objects, validating the JSON data and processing each trace item accordingly.

Test suite updates:

* [`tests/test_trace.py`](diffhunk://#diff-6cc3dad51d58aecec1745e607a16e76b7c66a42c10ae778061bcc49c4ae817d1R46-R83): Added a new test `test_bulk_submit_traces` to verify the functionality of submitting multiple traces via a POST request.

Other changes:

* [`lib/rucio/web/rest/flaskapi/v1/traces.py`](diffhunk://#diff-079a0ba14236814e622ee6688ffcbfc312582d32e3a4dde97a9a697f71df6626R18-R26): Updated imports to include `Union` from `typing` and `json` for JSON handling.

**Open Question**
Should the Rucio client be updated to leverage this change?

PR #8578, opened 2026-06-03
Metadata: Fix bulk metadata plugin handling; rucio#7323

### LINKED ISSUE #7323: `get_metadata_bulk` ignores `plugin` when `inherit=False`
### Description

Follow-up from https://github.com/rucio/rucio/pull/7317

When `get_metadata_bulk` is called with `inherit=True`, a call is made to `get_metadata` for each unique DID, and we can pass the `plugin` down to this call:

https://github.com/rucio/rucio/blob/dfcd980f4f10720af38ea1570a6015b2478a5408/lib/rucio/core/did.py#L2099

However, when `inherit=False`, a direct SQLAlchemy query is made instead.

https://github.com/rucio/rucio/blob/dfcd980f4f10720af38ea1570a6015b2478a5408/lib/rucio/core/did.py#L2110-L2129

We should consider whether
1. this is the intended use case (I assume it's not and we want to implement this `plugin` functionality in the case of `get_metadata_bulk` as well)
2. if it is the intended use case, we should explicitly mention this in the docs

### Steps to reproduce

Call `get_metadata_bulk` with any plugin

### Rucio Version

_No response_

### Additional Information

_No response_

Geogouz:
Fixes `get_metadata_bulk` so `plugin` is respected when `inherit=False`.

Also aligns bulk metadata defaults to `DID_COLUMN` across core, client, and REST to preserve the existing default output for non-inherited calls.

## Checklist
- [x] Created issue: rucio#7323
- [x] Added tests
- [x] For backwards compatibility breaking changes, leave additional description, follow conventional commits

<details>

```
- **Quality**: Agree 🟢
- **Testing**: Agree 🟢
- **Documentation**: Agree 🟢

# Notes for merger



```
</details>

### DIFF
--- a/lib/rucio/core/did.py
+++ b/lib/rucio/core/did.py
@@ -2089,7 +2089,7 @@ def list_parent_dids_bulk(
 def get_metadata_bulk(
     dids: list["Mapping[Any, Any]"],
     inherit: bool = False,
-    plugin: str = 'JSON',
+    plugin: str = 'DID_COLUMN',
     *,
     session: "Session"
 ) -> "Iterator[dict[str, Any]]":
--- a/lib/rucio/core/did.py
+++ b/lib/rucio/core/did.py
@@ -2136,9 +2136,9 @@ def get_metadata_bulk(
         for dids in parent_list:
             result = {'scope': dids[0][0], 'name': dids[0][1]}
             for did in dids:
-                for key in meta_dict[did]:
+                for key, value in meta_dict[did].items():
                     if key not in result:
-                        result[key] = meta_dict[did][key]
+                        result[key] = value
             yield result
     else:
         condition = []
@@ -2157,7 +2157,15 @@ def get_metadata_bulk(
                     or_(*chunk)
                 )
                 for row in session.execute(stmt).scalars():
-                    yield row.to_dict()
+                    if plugin.casefold() == 'did_column':
+                        yield row.to_dict()
+                    else:
+                        meta = get_metadata(row.scope, row.name, plugin=plugin, session=session)
+                        result = {'scope': row.scope, 'name': row.name}
+                        for key, value in meta.items():
+                            if key not in result:
+                                result[key] = value
+                        yield result
         except NoResultFound:
             raise exception.DataIdentifierNotFound('No Data Identifiers found')
 
--- a/lib/rucio/web/rest/flaskapi/v1/dids.py
+++ b/lib/rucio/web/rest/flaskapi/v1/dids.py
@@ -1778,7 +1778,7 @@ class BulkDIDsMeta(ErrorHandlingMethodView):
                       - scope: "user"
                         name: "dataset_002"
                     inherit: true
-                    plugin: "JSON"
+                    plugin: "DID_COLUMN"
 
         responses:
           200:
--- a/lib/rucio/web/rest/flaskapi/v1/dids.py
+++ b/lib/rucio/web/rest/flaskapi/v1/dids.py
@@ -1765,9 +1765,9 @@ class BulkDIDsMeta(ErrorHandlingMethodView):
                   plugin:
                     description: >
                       Which metadata plug‑in to query
-                      (`"JSON"`, `"DID_COLUMN"`, `"ALL"`, etc.; default: `"JSON"`).
+                      (`"JSON"`, `"DID_COLUMN"`, `"ALL"`, etc.; default: `"DID_COLUMN"`).
                     type: string
-                    default: "JSON"
+                    default: "DID_COLUMN"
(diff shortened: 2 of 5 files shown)


lib/rucio/core/did.py
```diff
@@ -2157,7 +2157,15 @@ def get_metadata_bulk(
...
                 for row in session.execute(stmt).scalars():
-                    yield row.to_dict()
+                    if plugin.casefold() == 'did_column':
+                        yield row.to_dict()
+                    else:
+                        meta = get_metadata(row.scope, row.name, plugin=plugin, session=session)
+                        result = {'scope': row.scope, 'name': row.name}
+                        for key in meta:
+                            if key not in result:
+                                result[key] = meta[key]
```
maxnoe:
```suggestion
                        for key, value in meta.items():
                            if key not in result:
                                result[key] = value
```

maxnoe: Why do we check for presence of keys in result? If scope and name are there, we could just `yield meta`, right?

Geogouz: Sure. Also I know this is a bit awkward because it dictates who gets to have priority (currently taking precedence over later values), but I could live for now with this. In the new metadata engine we don't get to have such problems. No metadata is lost/overriden.

Geogouz: Changed also the other branch for uniformity

maxnoe: > Currently taking precedence over later values),

I don't understand where the precedence issue arises here. `result` only contains `scope` and `name`, then we update it with a single `meta` from `get_metadata`.

Geogouz: I refer to the `inherit=True` branch that was adjusted. So this behavior was not introduced in this PR. Just mentioning that it is something that I'm not a big fan, and just letting that be as we had it so far. But it has changed in the new engine. Should have clarified it better.

maxnoe: Then the question remains: why do we check for key presence in the `inherit=False` case?

Geogouz: Someone theoretically could have already defined 'scope' and 'name' there.
For example the plugin could return

```
{
    'name': 'something_random',
    'scope': 'something_else',
    'data_product_id': '...'
}
```

maxnoe (dismissed):
- **Quality**: Agree 🟢
- **Security**: Agree 🟢 
- **Backwards compatibility**: Disagree 🔴
- **Testing**: Agree 🟢
- **Documentation**: Agree 🟢

This is a backwards incompatible bugfix, but most of the calls to `get_metadata_bulk` will stay the same.

- `get_metadata_bulk(dids)`, will stay as the default plugin was changed to from `JSON` to `DID_COLUMN`
- `get_metadata_bulk(dids, inherit=False, plugin="JSON")` will now work as expected (the actual fix for #7373)
 
but: 
- `get_metadata_bulk(dids, inherit=True)` will change: it will now by default return the `DID_COLUMN` meta, not the `JSON` meta


I'm not sure if the last point is ok or if we should try to keep compatibilty here.

One way to do that would be to say `plugin=None` as the default which would result in `plugin="JSON"` for the inherit branch and `plugin="DID_COLUMN"` for the other.

Geogouz:
> One way to do that would be to say `plugin=None` as the default which would result in `plugin="JSON"` for the inherit branch and `plugin="DID_COLUMN"` for the other.

Hmm, tbh something like that might feel to me like preserving history at the cost of a bit stranger API.

And I am not sure how important this backwards compatibility is practically. Because one can simply pass the plugin as param. And we can state that in the release notice as a warning.

An alternative maybe could be to keep/align the bulk metadata default to `JSON` instead of `DID_COLUMN`. Then fixing #7323 would make plugin="JSON" actually return JSON for `inherit=False`, and the old `DID_COLUMN` output for `omitted-plugin` + `inherit=False` calls could be treated as accidental behavior caused by the bug (so no backwards incompatibility).

But then, I am also not big fan of that because in this case we would break the nice uniformity against the single-DID metadata which defaults to `DID_COLUMN`.

Geogouz:
Any suggestion/opinion regarding this @robbarnsley ?

lib/rucio/web/rest/flaskapi/v1/dids.py
bari12: DID_COLUMN

bari12 (changes requested):
One very small change in the docs, otherwise this is GTG

bari12 (approved):
- **Quality**: Agree 🟢
- **Security**: Agree 🟢
- **Testing**: Agree 🟢
- **Documentation**: Agree 🟢

# Notes for merger

PR #8596, opened 2026-06-16
Use native XRootD Python bindings

lobis:
## Summary

- refactor upload/download `preferred_impl` handling so configured protocol order is respected and protocol priority `0` is treated as supported
- replace the XRootD RSE protocol implementation shell calls with optional native `XRootD.client` Python bindings when `xrootd` is importable and version-supported
- keep GFAL and other protocol implementations available as fallback choices through the existing RSE protocol selection
- add `xrootd>=6.1.0` to dev requirements so test/development environments include the native bindings without making them a mandatory runtime dependency for Rucio installs

## Motivation

The existing `rucio.rse.protocols.xrootd.Default` implementation required local XRootD command-line tools for `root://` operations. The native Python bindings let this protocol avoid `xrdcp`/`xrdfs` subprocess calls while preserving the existing plugin/fallback model for deployments that continue to prefer GFAL or other implementations.

`xrootd` remains an optional runtime dependency. If the bindings are missing or unsupported, the native protocol reports the missing dependency through the normal Rucio protocol path rather than making XRootD a mandatory Rucio install dependency.

## Implementation notes

- native operations use `XRootD.client.FileSystem` and `CopyProcess`
- auth setup uses the current XRootD process-wide environment hooks, so native calls are serialized and previous auth environment values are restored after each operation
- X.509 proxy discovery handles explicit Rucio/client env values, configured proxy paths, and the default `/tmp/x509up_u<uid>` location
- tests cover version gating, proxy cleanup, auth environment restoration, byte checksum decoding, and rename behavior when parent directories already exist

## Validation

- `pre-commit run --files lib/rucio/rse/protocols/xrootd.py requirements/requirements.dev.in requirements/requirements.dev.txt tests/test_rse_protocol_xrootd.py`
- `python3 -m pytest tests/test_rse_protocol_xrootd.py::test_native_xrootd_requires_version_6_or_newer tests/test_rse_protocol_xrootd.py::test_native_xrootd_clears_unexpanded_proxy_from_env tests/test_rse_protocol_xrootd.py::test_native_xrootd_operation_restores_auth_environment tests/test_rse_protocol_xrootd.py::test_native_xrootd_stat_accepts_bytes_checksum tests/test_rse_protocol_xrootd.py::test_native_xrootd_rename_ignores_existing_directory -q`
- `PYTHONPATH=lib npx pyright --pythonpath .venv/bin/python lib/rucio/rse/protocols/xrootd.py`
- `npx commitlint --from origin/master --to HEAD --verbose`

### DIFF
--- a/lib/rucio/client/downloadclient.py
+++ b/lib/rucio/client/downloadclient.py
@@ -1950,16 +1950,18 @@ class DownloadClient:
             self.logger(logging.INFO, 'No preferred protocol impl in rucio.cfg: %s' % (error))
             return supported_impl
         else:
-            preferred_impls = list(preferred_impls.split(', '))
-            i = 0
-            while i < len(preferred_impls):
-                impl = preferred_impls[i]
-                impl_split = impl.split('.')
-                if len(impl_split) == 1:
-                    preferred_impls[i] = 'rucio.rse.protocols.' + impl + '.Default'
+            configured_impls = preferred_impls
+            preferred_impls = []
+            for impl in configured_impls.split(','):
+                impl = impl.strip()
+                if not impl:
+                    continue
+                if impl.startswith('rucio.'):
+                    preferred_impls.append(impl)
+                elif '.' in impl:
+                    preferred_impls.append('rucio.rse.protocols.' + impl)
                 else:
-                    preferred_impls[i] = 'rucio.rse.protocols.' + impl
-                i += 1
+                    preferred_impls.append('rucio.rse.protocols.' + impl + '.Default')
 
         for source in sources:
             if source['rse'] in checked_rses:
@@ -1971,13 +1973,18 @@ class DownloadClient:
                 self.logger(logging.DEBUG, 'Could not get info of RSE %s: %s' % (source['source'], error))
                 continue
 
-            preferred_protocols = [protocol for protocol in reversed(rse_settings['protocols']) if protocol['impl'] in preferred_impls]
+            preferred_protocols = []
+            for preferred_impl in preferred_impls:
+                preferred_protocols.extend([
+                    protocol for protocol in rse_settings['protocols']
+                    if protocol['impl'] == preferred_impl and protocol not in preferred_protocols
+                ])
 
             if len(preferred_protocols) == 0:
                 continue
 
             for protocol in preferred_protocols:
-                if not protocol['domains']['wan'].get("read"):
+                if protocol['domains']['wan'].get("read") is None:
                     self.logger(logging.WARNING, 'Unsuitable protocol "%s": "WAN Read" operation is not supported' % (protocol['impl']))
                     continue
                 try:
--- a/lib/rucio/rse/protocols/xrootd.py
+++ b/lib/rucio/rse/protocols/xrootd.py
@@ -12,26 +12,76 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from __future__ import annotations
+
 import logging
 import os
+from contextlib import contextmanager
+from importlib import import_module, metadata
+from threading import RLock
+from typing import TYPE_CHECKING, Any, cast
+
+from packaging.version import InvalidVersion, Version
 
 from rucio.common import exception
 from rucio.common.checksum import PREFERRED_CHECKSUM
-from rucio.common.utils import execute
+from rucio.common.config import config_get
 from rucio.rse.protocols import protocol
+
+if TYPE_CHECKING:
+    from collections.abc import Iterator
+    from types import ModuleType
+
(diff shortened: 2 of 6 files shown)


lib/rucio/client/downloadclient.py
```diff
@@ -1950,16 +1950,15 @@ def preferred_impl(self, sources: list[dict[str, Any]]) -> Optional[str]:
...
-                i += 1
+            preferred_impls = [
+                'rucio.rse.protocols.' + impl.strip() + '.Default'
+                if '.' not in impl.strip()
+                else impl.strip()
+                if impl.strip().startswith('rucio.')
+                else 'rucio.rse.protocols.' + impl.strip()
+                for impl in preferred_impls.split(',')
+                if impl.strip()
+            ]
```
dchristidis: This and other changes to `{down,up}loadclient.py` are not described by `feat(Protocols): use native XRootD bindings`. You need at least one commit of type `refactor`. Please refer to the contributing guide.

But I should state already that I have some reservations as to whether a list comprehension with two conditional statements is more comprehensible than the existing implementation.

lib/rucio/rse/protocols/xrootd.py
```diff
@@ -42,6 +80,132 @@ def __init__(self, protocol_attr, rse_settings, logger=logging.log):
...
+    def _clear_unexpanded_x509_proxy(self):
+        proxy = os.environ.get('X509_USER_PROXY')
+        if proxy and '$' in os.path.expandvars(proxy):
+            os.environ.pop('X509_USER_PROXY', None)
+
+    def _configure_auth(self):
+        xrootd_client = _client()
+        if self.auth_token:
+            os.environ['XrdSecPROTOCOL'] = 'ztn'
+            os.environ['BEARER_TOKEN'] = self.auth_token
```
dchristidis, line 175: Are environmental variables the only way to pass such things to XRootD? What happens in a multi-threaded process (e.g. the Reaper)?

lib/rucio/rse/protocols/xrootd.py
```diff
@@ -14,22 +14,60 @@
...
+
+    if not _is_supported_xrootd_version(_xrootd_client):
+        _xrootd_client = None
+        _xrootd_flags = None
+except Exception:
+    _xrootd_client = None
+    _xrootd_flags = None
+
+
+def _client() -> Any:
```
dchristidis: Maybe `types.ModuleType`?

lib/rucio/rse/protocols/xrootd.py
```diff
@@ -42,6 +80,132 @@ def __init__(self, protocol_attr, rse_settings, logger=logging.log):
         self.hostname = self.attributes['hostname']
         self.port = str(self.attributes['port'])
         self.logger = logger
+        self.__fs: Optional[Any] = None
+
+    @property
+    def _endpoint(self):
```
dchristidis: Kindly provide type hints to all new methods.

.github/workflows/integration_tests.yml
```diff
@@ -156,6 +156,7 @@ jobs:
             ln -sf /root/.ssh/ruciouser_sshkey.pub /root/.ssh/id_rsa.pub 2>/dev/null || true
 
             pip install --no-cache-dir -e /rucio_source
+            python3 -m pip install --no-cache-dir 'xrootd>=6.0.0'
```
dchristidis: I need someone from the Testing component to comment on this because it doesn’t look like the appropriate location to me.

lobis: Placed in [‎requirements/requirements.dev.in](https://github.com/rucio/rucio/pull/8596/changes/1742838898d7d8944462801e4d7a29738eedfedb#diff-28885c8845c76cd15ae72018dd14e0362b00d62548dbe46bbd8ac012a67ce319)

lib/rucio/rse/protocols/xrootd.py
lobis: Added

lib/rucio/rse/protocols/xrootd.py
lobis: All new methods are typed now altough some with `Any`

lib/rucio/rse/protocols/xrootd.py
lobis: Added some locking mechanism at https://github.com/rucio/rucio/pull/8596/changes#diff-b2412edd320079a7a038c10aaf089dc0ad455bec9fd2754ce6e463a6e6478e5aR177-R179 - its not the cleanest implementation but it should work - we should probably improve this from the xrootd side

bziemons:
this needs a rebase for the CI to work

PR #8627, opened 2026-06-24
Allow list_dids(scope=None) in postres did_meta plugins

maxnoe:

## Checklist
- [x] Created issue: #6070 
- [x] Added tests

<details>

```
- **Quality**: Agree 🟢
- **Testing**: Agree 🟢
- **Documentation**: Agree 🟢

# Notes for merger



```
</details>

### DIFF
--- a/lib/rucio/core/did_meta_plugins/postgres_meta.py
+++ b/lib/rucio/core/did_meta_plugins/postgres_meta.py
@@ -290,14 +290,15 @@ class ExternalPostgresJSONDidMeta(DidMetaPlugin):
         if isinstance(filters, dict):
             filters = [filters]
 
+        additional_filters = [('vo', operator.eq, scope.vo)]
+        if scope is not None:
+            additional_filters.append(('scope', operator.eq, scope.internal))
+
         try:
             # instantiate fe and create postgres query
             fe = FilterEngine(filters, model_class=None, strict_coerce=False)
             postgres_where_sql, filter_args = fe.create_postgres_query(
-                additional_filters=[
-                    ('scope', operator.eq, scope.internal),
-                    ('vo', operator.eq, scope.vo)
-                ],
+                additional_filters=additional_filters,
                 fixed_table_columns=self.fixed_table_columns,
                 jsonb_column=self.jsonb_column
             )
--- a/lib/rucio/core/did_meta_plugins/__init__.py
+++ b/lib/rucio/core/did_meta_plugins/__init__.py
@@ -223,6 +223,9 @@ def list_dids(scope=None, filters=None, did_type='collection', ignore_case=False
     if isinstance(filters, dict):
         filters = [filters]
 
+    if scope is None and not long:
+        raise ValueError("long must be True in case of scope=None")
+
     required_unique_plugins = set()                 # keep track of which plugins are required
     for or_group in filters:
         for key in or_group.keys():
--- a/lib/rucio/core/did_meta_plugins/did_column_meta.py
+++ b/lib/rucio/core/did_meta_plugins/did_column_meta.py
@@ -345,6 +345,12 @@ class DidColumnMeta(DidMetaPlugin):
                 filters_tmp.append(or_group.copy())
         filters = filters_tmp
 
+        additional_filters: list[tuple[Any, Any, Any]] = [
+            (models.DataIdentifier.suppressed, operator.ne, true()),
+        ]
+        if scope is not None:
+            additional_filters.append((models.DataIdentifier.scope, operator.eq, scope))
+
         # instantiate fe and create sqla query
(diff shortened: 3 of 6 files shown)


maxnoe:
The changes here are enough to make

```
>>> c = Client()
>>> dids = c.list_dids(scope="*", filters=[{"name": "*"}], long=True)
>>> for did in dids:
...     print(did)
...
{'scope': 'test', 'name': 'dataset1', 'did_type': 'DATASET', 'bytes': None, 'length': None}
{'scope': 'test', 'name': 'dataset2', 'did_type': 'DATASET', 'bytes': None, 'length': None}
{'scope': 'test', 'name': 'container', 'did_type': 'CONTAINER', 'bytes': None, 'length': None}
{'scope': 'test', 'name': 'dataset3', 'did_type': 'DATASET', 'bytes': None, 'length': None}
{'scope': 'foo', 'name': 'foo', 'did_type': 'DATASET', 'bytes': None, 'length': None}
{'scope': 'foo', 'name': 'dataset1234', 'did_type': 'DATASET', 'bytes': None, 'length': None}
```

work

maxnoe:
I leave this as draft for now as there were worries about performance implications and the request to gate this behind a server configuration.

Also, I didn't yet implement this in the other metadata plugins, only in did meta and postgres

lib/rucio/core/did_meta_plugins/postgres_meta.py
```diff
@@ -290,14 +290,15 @@ def list_dids(self, scope, filters, did_type='collection', ignore_case=False, li
         if isinstance(filters, dict):
             filters = [filters]
 
+        additional_filters = [('vo', operator.eq, scope.vo)]
```
maxnoe, line 293: note to self: this doesn't actually work in the scope=None case

Geogouz:
Max, let me know when you want me to take a look at this.

maxnoe:
One question would be why the postgres plugin filters by vo while the internal did meta plugin does not.

Geogouz:
The difference is mostly where VO lives. The Postgres metadata plugin stores VO as its own table column. See

https://github.com/rucio/rucio/blob/75976b432c4a878e9cd7d772ce2d87eeb1c516e2/lib/rucio/core/did_meta_plugins/postgres_meta.py#L104-L110

metadata writes store `scope.vo`

https://github.com/rucio/rucio/blob/75976b432c4a878e9cd7d772ce2d87eeb1c516e2/lib/rucio/core/did_meta_plugins/postgres_meta.py#L245-L252

and `list_dids` adds `vo == scope.vo`

https://github.com/rucio/rucio/blob/75976b432c4a878e9cd7d772ce2d87eeb1c516e2/lib/rucio/core/did_meta_plugins/postgres_meta.py#L293-L295

So for that backend, keeping a VO filter is what keeps the query inside the current VO when the scope filter is relaxed.

The internal DID metadata plugin queries `DataIdentifier`, where there is `scope` but no separate `vo` column. VO is part of the internal scope value where `InternalScopeString` stores `value.internal` and non-default VOs are encoded as `external@vo`

https://github.com/rucio/rucio/blob/75976b432c4a878e9cd7d772ce2d87eeb1c516e2/lib/rucio/common/types.py#L100-L112

That is why a normal scoped query using `DataIdentifier.scope == scope` already carries the VO restriction.

https://github.com/rucio/rucio/blob/91ac893976903562500b708238179cc4867b1f20/lib/rucio/core/did_meta_plugins/did_column_meta.py#L357-L359

The tricky part in this PR is the cross-scope case. Once `scope="*"` stops being a concrete `InternalScope`, the internal plugin only adds the scope predicate when `scope is not None`

https://github.com/rucio/rucio/blob/75976b432c4a878e9cd7d772ce2d87eeb1c516e2/lib/rucio/core/did_meta_plugins/did_column_meta.py#L348-L352

That also removes the implicit VO boundary.

So my suggestion would be to carry `vo` separately for cross-scope searches and instead of the plugin API only receiving this:

```python
list_dids(scope=None, filters=...)
```

it should receive enough context to still know the VO, for example:

```python
list_dids(scope=None, vo=vo, filters=...)
```

or use some equivalent internal representation where the scope is wildcarded but the VO is still known.

Then internal plugins can add a VO-safe scope predicate, while Postgres keeps filtering by `vo` and only relaxes the scope predicate.

maxnoe:
> So my suggestion would be to carry vo separately for cross-scope searches and instead of the plugin API only receiving this:

Thanks a lot for the detailed walk-through. I agree with the conclusion that `vo` should be an explicit parameter.

maxnoe:
We could add it backwards compatible with:

```
if vo is None and scope is None:
    raise ValueError("vo must be provided when not restricting scope")
else:
   vo = scope.vo
```

PR #8631, opened 2026-06-26
Core: Added overloads for account limit functions

### LINKED ISSUE #8194: Account limits: `get` functions can return many different types, but their usages expect only one type (mostly `dict`)
### Description

E.g. this function:

https://github.com/rucio/rucio/blob/b36113b21f2847d5f715e4bc38a8cf2f25dbe792/lib/rucio/core/account_limit.py#L148

returns, in single rse case:
- `float` if `if account_limit.bytes == -1`,
- `int` otherwise
- `None` if not found

in multi-rse case:
- `dict`

However, when this is called, there is almost always an assumption for only one of these files, e.g. here:

https://github.com/rucio/rucio/blob/b36113b21f2847d5f715e4bc38a8cf2f25dbe792/lib/rucio/gateway/account_limit.py#L73-L75

Where `limits.items()` works if the returned item is a `dict`, but would break with `float` and `int`.

I am finding many of these occurrences while working on #6989 - I am labelling them with `type: ignore` for now, but this should be addressed in some way. 

I think ideally we should overload the signature so that:
if `rse_ids` is an `Iterable[str]`, then it's multi-RSE, so the return type is `dict`. Otherwise, if `rse_ids` is `str,` it's single-RSE and the return type is `float | int | None` (though even in this case, I would prefer that we only had either float or int, not both possible)

### Steps to reproduce

na

### Rucio Version

_No response_

### Additional Information

_No response_

ASHISHTOMER0817:
Closes: #8194 

## Checklist
- [x] Created issue: #8194 
- [ ] Added tests

This PR adds overloads for `get_global_account_limit` and `get_local_account_limit` to provide more precise return types based on the input arguments. It also updates the corresponding gateway type annotations to reflect the expected return types.
There is no runtime behavior change, verified by pyright clean run.
While working on this, I noticed two related issues. (1) get_global_account_limit(account=None) in core/account_limit.py can silently collapse limits from different accounts into the same dict key if they share an rse_expression — not fixed in this PR, flagging for a separate discussion/issue since it's a behavioral question, not a typing one. (2) The gateway's get_global_account_limit had a stale return-type annotation that didn't match either of its actual branches — likely invisible to mypy/pyright before, since the old blanket Union on the core function made the mismatch unprovable. Fixed in this PR as a necessary consequence of correcting the core function's types.
cc @rdimaio — opened this per the approach you outlined in #8194.

<details>

```
- **Quality**: Agree 🟢
- **Testing**: Agree 🟢
- **Documentation**: Agree 🟢

# Notes for merger



```
</details>

### DIFF
--- a/lib/rucio/core/account_limit.py
+++ b/lib/rucio/core/account_limit.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import TYPE_CHECKING, Optional, Union
+from typing import TYPE_CHECKING, Optional, Union, overload
 
 from sqlalchemy.exc import NoResultFound
 from sqlalchemy.sql import func, literal, select
@@ -92,9 +92,14 @@ def get_rse_account_usage(rse_id: str, session: "Session") -> list["RSEAccountUs
     return result
 
 
-def get_global_account_limit(
-    session: "Session",
-    account: Optional["InternalAccount"] = None,
+@overload
+def get_global_account_limit(session: "Session", account: "InternalAccount", rse_expression: str) -> Optional[Union[int, float]]: ...
+
+@overload
+def get_global_account_limit(session: "Session", account: Optional["InternalAccount"] = None, rse_expression: Optional[str] = None) -> dict[str, "RSEResolvedGlobalAccountLimitDict"]: ...
+
+
+def get_global_account_limit(session: "Session", account: Optional["InternalAccount"] = None,
     rse_expression: Optional[str] = None
 ) -> Optional[Union[int, float, dict[str, "RSEResolvedGlobalAccountLimitDict"]]]:
     """
@@ -144,11 +149,15 @@ def get_global_account_limit(
     return resolved_global_account_limits
 
 
-def get_local_account_limit(
-    account: "InternalAccount",
-    session: "Session",
-    rse_ids: Optional[Union[str, "Iterable[str]"]] = None
-) -> Optional[Union[int, float, dict[str, int]]]:
+@overload
+def get_local_account_limit(account: "InternalAccount", session: "Session", rse_ids: str) -> Optional[Union[int, float]]: ...
+
+@overload
+def get_local_account_limit(account: "InternalAccount", session: "Session", rse_ids: Optional["Iterable[str]"] = None) -> dict[str, Union[int, float]]: ...
+
+
+def get_local_account_limit(account: "InternalAccount", session: "Session", rse_ids: Optional[Union[str, "Iterable[str]"]] = None
+) -> Optional[Union[int, float, dict[str, Union[int, float]]]]:
     """
     Returns the local account limit for a given RSE or list of RSEs.
 
@@ -326,7 +335,7 @@ def get_local_account_usage(account: "InternalAccount", session: "Session", rse_
         counters = {c.rse_id: c for c in session.execute(stmt).scalars().all()}
     result_list = []
 
-    for rse_id in set(limits).union(counters):  # type: ignore (https://github.com/rucio/rucio/issues/8194)
+    for rse_id in set(limits).union(counters):
         counter = counters.get(rse_id)
         if counter:
             counter_files = counter.files
@@ -335,14 +344,14 @@ def get_local_account_usage(account: "InternalAccount", session: "Session", rse_
             counter_files = 0
             counter_bytes = 0
 
-        if counter_bytes > 0 or counter_files > 0 or rse_id in limits.keys():  # type: ignore (https://github.com/rucio/rucio/issues/8194)
+        if counter_bytes > 0 or counter_files > 0 or rse_id in limits.keys():
             result_list.append({
                 'rse_id': rse_id,
                 'rse': get_rse_name(rse_id=rse_id, session=session),
                 'bytes': counter_bytes,
                 'files': counter_files,
-                'bytes_limit': limits.get(rse_id, 0),  # type: ignore (https://github.com/rucio/rucio/issues/8194)
-                'bytes_remaining': limits.get(rse_id, 0) - counter_bytes,  # type: ignore (https://github.com/rucio/rucio/issues/8194)
+                'bytes_limit': limits.get(rse_id, 0),
+                'bytes_remaining': limits.get(rse_id, 0) - counter_bytes,
             })
     return result_list
 
@@ -366,7 +375,7 @@ def get_global_account_usage(
         # All RSE Expressions
         limits = get_global_account_limit(account=account, session=session)
         all_rse_usages = {usage['rse_id']: (usage['bytes'], usage['files']) for usage in get_all_rse_usages_per_account(account=account, session=session)}
-        for rse_expression, limit in limits.items():  # type: ignore
+        for rse_expression, limit in limits.items():
             usage = 0
             files = 0
             for rse in limit['resolved_rse_ids']:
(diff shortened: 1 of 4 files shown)


ASHISHTOMER0817:
Hi @rdimaio, just checking in on this — all checks are passing and it's been about 2 weeks. Also flagged a possible related issue in the description (the account=None collapsing behavior) in case that's useful context whenever you get a chance to look. No rush, just bumping it up in case it slipped through.

rdimaio:
> Hi @rdimaio, just checking in on this — all checks are passing and it's been about 2 weeks. Also flagged a possible related issue in the description (the account=None collapsing behavior) in case that's useful context whenever you get a chance to look. No rush, just bumping it up in case it slipped through.

I am not actively contributing in the project at the moment, please check with others

ASHISHTOMER0817:
There is one Lint(pre-commit) error, this is not the result of my PR. This error is coming from file: `/lib/rucio/web/rest/flaskapi/v1/credential.py b/lib/rucio/web/rest/flaskapi/v1/credential.py`. I only modified `core/account_limit` and `gateway/account_limit`.

lib/rucio/core/account_limit.py
```diff
@@ -92,11 +93,14 @@ def get_rse_account_usage(rse_id: str, session: "Session") -> list["RSEAccountUs
...
-    account: Optional["InternalAccount"] = None,
-    rse_expression: Optional[str] = None
-) -> Optional[Union[int, float, dict[str, "RSEResolvedGlobalAccountLimitDict"]]]:
+@overload
+def get_global_account_limit(account: "InternalAccount", rse_expression: str, *, session: "Session") -> Optional[Union[int, float]]: ...
+
+@overload
+def get_global_account_limit(account: Optional["InternalAccount"] = None, rse_expression: Optional[str] = None, *, session: "Session") ->  dict[str, "RSEResolvedGlobalAccountLimitDict"]: ...
+
+@read_session
```
bari12: Why are you re-adding this decorator? We are trying to get rid of these. See https://github.com/rucio/rucio/pull/8350

bari12 (changes requested):

ASHISHTOMER0817:
@bari12 Good catch — I just pushed a fix removing those decorators. I should have checked the latest commits to the file before starting, as you noted in #8350. Thanks for the feedback.

lib/rucio/core/account_limit.py
```diff
@@ -144,11 +147,14 @@ def get_global_account_limit(
...
-def get_local_account_limit(
-    account: "InternalAccount",
-    session: "Session",
-    rse_ids: Optional[Union[str, "Iterable[str]"]] = None
-) -> Optional[Union[int, float, dict[str, int]]]:
+@overload
+def get_local_account_limit(account: "InternalAccount", session: "Session", rse_ids: str) -> Optional[Union[float, int]]: ...
+
+@overload
+def get_local_account_limit(account: "InternalAccount", session: "Session", rse_ids: Optional["Iterable[str]"] = None) -> dict[str, int]: ...
```
VladGaluska: Return type should be `dict[str, Union[int, float]]` as stated in the docs of the original function `:return:         Limit in Bytes (int/float) for a single RSE ...`

lib/rucio/core/account_limit.py
```diff
@@ -92,11 +92,14 @@ def get_rse_account_usage(rse_id: str, session: "Session") -> list["RSEAccountUs
...
-    rse_expression: Optional[str] = None
-) -> Optional[Union[int, float, dict[str, "RSEResolvedGlobalAccountLimitDict"]]]:
+@overload
+def get_global_account_limit(session: "Session", account: "InternalAccount", rse_expression: str) -> Optional[Union[int, float]]: ...
+
+@overload
+def get_global_account_limit(session: "Session", account: Optional["InternalAccount"] = None, rse_expression: Optional[str] = None) ->  dict[str, "RSEResolvedGlobalAccountLimitDict"]: ...
+
+
+def get_global_account_limit(session: "Session", account=None, rse_expression=None):
```
VladGaluska: This removed the type annotations from the `account` and `rse_expression` , as well as the return type. That means that these 2 parameters will be `Unknown` inside the function and no errors will be caught. Can you restore the original types here please? Also for `get_local_account_limit` below

lib/rucio/core/account_limit.py
```diff
@@ -92,11 +92,14 @@ def get_rse_account_usage(rse_id: str, session: "Session") -> list["RSEAccountUs
...
-def get_global_account_limit(
-    session: "Session",
-    account: Optional["InternalAccount"] = None,
-    rse_expression: Optional[str] = None
-) -> Optional[Union[int, float, dict[str, "RSEResolvedGlobalAccountLimitDict"]]]:
+@overload
+def get_global_account_limit(session: "Session", account: "InternalAccount", rse_expression: str) -> Optional[Union[int, float]]: ...
+
+@overload
+def get_global_account_limit(session: "Session", account: Optional["InternalAccount"] = None, rse_expression: Optional[str] = None) ->  dict[str, "RSEResolvedGlobalAccountLimitDict"]: ...
```
VladGaluska: Nitpick, but there's an extra space in `->  dict...`

VladGaluska: There is an issue here if the caller passes a variable that's defined as `Optional["InternalAccount"]`. At runtime it behaves correctly, but `pyright` will show the caller that the return type is the dict, so the caller might then do `limits.items()`. And for this case, the older implementation at least had the users figure out which of the result types they got. I checked and no caller does this right now, but it could happen in the future. 

This depends on how the maintainers would like this continued, as different projects approach this differently, but to me, this would warrant different names for the functions. We could add a third overload, but the same problem is at `get_local_account_limit` and there another overload wouldn't fix it

VladGaluska (commented):
Looking at the original issue, the following commits introduced `# type: ignore` suppressions, which should be removed as part of this pr:
- https://github.com/rdimaio/rucio/commit/2808d4011ab3cd8a9c617ed5992cc33600a36807
- https://github.com/rdimaio/rucio/commit/2b686b4558c3d877aa387571c73d898e7a16eda0
- https://github.com/rdimaio/rucio/commit/fa9247417b8d4c061c67aae365970cd5fc38bc7e
You can also find them by searching the project for `8194`

Additionally, the commit hygiene needs a bit of work. The commits should be squashed and the description should be an actual message, ending with a `Closes: #8631`  trailer.

VladGaluska:
@ASHISHTOMER0817 is this ready for re-review? I believe it's still missing some of the changed like the `# type: ignore` fixes. Thanks!

ASHISHTOMER0817:
@VladGaluska I missed one # type: ignore that wasn’t tagged with #8194. I’ve removed it; the #8194-related ignores are all gone now. Ready for re-review when you have time. Thanks!

VladGaluska:
I can't resolve the comments as I lack the rights, but I marked them with a 👍  where they were done. There are still unresolved/unaddressed comments though

lib/rucio/core/account_limit.py
ASHISHTOMER0817: Extra space has been removed.

lib/rucio/core/account_limit.py
ASHISHTOMER0817: changed return types to Union[int, float] in the overload function and the implementation function.

bari12 (approved):
- **Quality**: Agree 🟢
- **Security**: Agree 🟢
- **Testing**: Agree 🟢
- **Documentation**: Agree 🟢

# Notes for merger

Issue #8653, opened 2026-07-07
conveyor-receiver using default STOMP heartbeat by not sending heartbeats


d-ylee:
### Description

While connecting conveyor-receiver to a deployed ActiveMQ Artemis broker, the ActiveMQ Artemis broker is timing out the client connection after 60 seconds.

The stomp client has a default `heartbeats` tuple of (0,0), which tells the client to not send stomp heartbeats and also tell the server to not send stomp heartbeats to the client. `conveyor-receiver` does not explicitly set the `heartbeats` kwarg when setting up the connection with `stomp.Connection12`.

As a result, the ActiveMQ broker will timeout the client with the default timeout of 60 seconds. There is also no way for the broker to send heartbeat to `conveyor-receiver` because of the client configuration.

### Steps to reproduce

1. Deploy an ActiveMQ broker
2. Deploy `conveyor-receiver` that connects to the broker
3. Wait 60 seconds and see the timeout frame in ActiveMQ. `conveyor-receiver` also logs an error (no useful error messages).

### Rucio Version

Daemons: 38.5.5

### Additional Information

- This bug doesn't occur when connecting to the CERN brokers. I believe it could be the CERN broker configuration, but I am unsure what configuration CERN is using.
- Environment: Kubernetes, ActiveMQ Artemis via ArkMQ operator

- I have made a test fix by explicitly setting the `heartbeats`: https://github.com/d-ylee/rucio/blob/7dcef632c02f48a424496ed71323cd3c4f57e363/lib/rucio/daemons/conveyor/receiver.py#L170-L173
- Further testing will be needed to make sure this doesn't affect receiving messages or other things.
- This could also potentially be set in `rucio.cfg`, which would require new configuration fields.

PR #8654, opened 2026-07-07
fix(Transfers): Added heartbeats to STOMP connection; rucio#8653

d-ylee:

## Checklist
- [x] Created issue: #8653 
- [ ] Added tests
- [x] Updated documentation (Please link PRs in documentation repository): rucio/documentation#842

Should be backwards compatible, since the default heartbeats are set to 0 milliseconds.

<details>

```
- **Quality**: Agree 🟢
- **Testing**: Agree 🟢
- **Documentation**: Agree 🟢

# Notes for merger



```
</details>

### DIFF
--- a/lib/rucio/daemons/conveyor/receiver.py
+++ b/lib/rucio/daemons/conveyor/receiver.py
@@ -167,10 +167,14 @@ def receiver(
         auth_kwargs['passcode'] = config_get('messaging-fts3', 'password')
         logger(logging.INFO, 'using username/password authentication.')
 
+    client_send = config_get_int('messaging-fts3', 'client_send_heartbeat', default=0)
+    client_recv = config_get_int('messaging-fts3', 'client_recv_heartbeat', default=0)
+
     conns = []
     for broker in brokers_resolved:
         con = stomp.Connection12(host_and_ports=[(broker, port)],
                                  vhost=vhost,
+                                 heartbeats=(client_send, client_recv),
                                  reconnect_attempts_max=999)
         if use_ssl:
             con.set_ssl(key_file=key_file, cert_file=cert_file)
--- a/lib/rucio/daemons/conveyor/receiver.py
+++ b/lib/rucio/daemons/conveyor/receiver.py
@@ -42,6 +42,7 @@ if TYPE_CHECKING:
     from types import FrameType
 
     from sqlalchemy.orm import Session
+    from stomp import Connection
     from stomp.utils import Frame
 
     from rucio.common.types import LoggerFunction
@@ -58,6 +59,7 @@ class Receiver:
     def __init__(
             self,
             broker: str,
+            conn: "Connection",
             id_: str,
             total_threads: int,
             transfer_stats_manager: request_core.TransferStatsManager,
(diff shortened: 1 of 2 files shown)


lib/rucio/daemons/conveyor/receiver.py
```diff
@@ -167,10 +167,14 @@ def receiver(
         auth_kwargs['password'] = config_get('messaging-fts3', 'password')
         logger(logging.INFO, 'using username/password authentication.')
 
+    client_send = config_get('messaging-fts3', 'client_send_heartbeat', default=0)
```
voetberg: Can you add these options to the [Config Parameters](https://rucio.cern.ch/documentation/operator/configuration_parameters) table in the documention?

d-ylee: Added in this PR: rucio/documentation#842

voetberg:
I think the pre-commit here is wrong, can you do me a favor and rebase?

voetberg (dismissed):
Probably can't be tested, and this is a minor enough change I don't see it breaking anything. lgmt 👍

lib/rucio/daemons/conveyor/receiver.py
```diff
@@ -167,10 +167,14 @@ def receiver(
         auth_kwargs['passcode'] = config_get('messaging-fts3', 'password')
         logger(logging.INFO, 'using username/password authentication.')
 
+    client_send = config_get('messaging-fts3', 'client_send_heartbeat', default=0)
```
bari12: This needs to be a `config_get_int` otherwise this will build a string tuple.

lib/rucio/daemons/conveyor/receiver.py
```diff
@@ -167,10 +167,14 @@ def receiver(
...
+    client_recv = config_get('messaging-fts3', 'client_recv_heartbeat', default=0)
```
bari12: This too

bari12 (changes requested):

bari12 (changes requested):
There is one more thing I would change (AI review brought this up):

Since the receiver does not have a `on_heartbeat_timeout` method, which activemq would call once it runs into a heartbeat timeout, it would raise an `AttributeError`. The whole thing would still work, since the receiver handles it, but `AttributeError` would show up in the log which is not very clean.

The clean thing would be to deal with it like kronos does: https://github.com/rucio/rucio/blob/360dc0be4f6ecdd83dd811f96e0088d4a465c2fa/lib/rucio/daemons/tracer/kronos.py#L61-L93 thus add the method. 

Otherwise this is gtg

bari12 (changes requested):
I am getting a few AI Review hits here. I did not fully trace this yet, but it seems plausible. Let me know if you want to discuss this offline.

Now that receive heartbeats is enabled, a slow database callback can outlast the timeout. Reconnection starts another reader before the original callback finishes; the original reader then resumes against the replacement socket. Competing readers can corrupt or drop completion messages. The new heartbeat argument makes this asynchronous interruption possible.

The right fix might be to just create a new Connection12 object at each reconnect.

d-ylee:
@bari12 This is probably out of scope, but would it be better to use StompConnectionManager from `rucio.common.stomp_utils.py`?

https://github.com/rucio/rucio/blob/master/lib/rucio/common/stomp_utils.py

Issue #8649, opened 2026-07-07
Add GitHub Actions workflows that run the pytest test runner


maany:
## Description

This adds two new GitHub Actions workflows that run the test suites through the pytest runner from
PR1: one general autotest workflow and one dedicated to the VO tests. Both are new files. No existing
workflow is edited or turned off.

## Motivation

Once the suites can run with a single `pytest` command, it makes sense for CI to use that same
command rather than a separate set of shell and matrix scripts. Running the plugin in Actions checks
that the runner works on every change, and keeps the local and CI paths the same, so a green run
locally should mean a green run in CI. Because the workflows are additive, they can be tried
alongside the current CI before anything is retired.

## Change

Two new workflow files are added and nothing existing is touched. `simple-autotest.yml` runs the
standard suites through the pytest runner. `simplify_votests.yml` runs the VO tests (atlas and
belleii) through the same runner using `--suite=votest --policy=<X>`.

Issue #8648, opened 2026-07-07
Add a pytest-based test runner for the dev environment


maany:
## Description

The `tests/ruciopytest/` package already exists in Rucio and holds the pytest helpers for
no-parallel test grouping. This change extends that package with a test runner that runs any Rucio
test suite with a single `pytest` command plus a few arguments. The runner handles the setup and
cleanup a suite needs: it starts the dev `docker compose` stack, sets up the database and test data,
selects the tests to run, and shuts the stack down afterwards. No separate shell scripts or matrix
logic are needed to run a suite. The new runner sits alongside the existing helpers in the package,
and the current test setup keeps working as before.

Two small, safe changes come along with the runner because it depends on them. The first is a guard
around the statsd monitoring client. When Rucio creates its statsd client it resolves the configured
monitoring host name; if that name cannot be resolved (for example in a minimal test container that
has no monitoring host), the underlying socket call raises `socket.gaierror` and the process fails
before the tests even run. The guard catches that specific lookup error so that monitoring is skipped
quietly and the tests carry on, instead of crashing. It only affects the failure path, so normal
environments where the host resolves are unaffected. The second change lets the dev compose container
names be set from environment variables, so several test runs can run at the same time without their
containers clashing. The default names are kept.

## Motivation

At the moment, running a suite means calling shell scripts and CI matrix logic and starting and
stopping containers by hand. That is awkward to reproduce locally and hard to follow. A single
`pytest <args>` command that takes care of the containers means the suites run the same way locally
and in CI. That lowers the barrier for new contributors, and it gives one clear way to run the tests
instead of several scattered scripts. The compose change keeps the current dev workflow exactly as it
is today and only lets the runner keep parallel runs apart.

## Change

Most of the change is new modules under `tests/ruciopytest/` (the runner itself, its unit tests, a
parity baseline file, and a short README). Three files already in the package are updated as well:
`__init__.py` gains the shared pytest stash keys that wire the runner in, and the two existing
no-parallel helpers `xdist_noparallel_remote.py` and `xdist_noparallel_scheduler.py` are extended so
the `noparallel` marker also accepts positional group names, not only the `groups=` keyword. The
runner is registered through `tests/conftest.py`, and `tools/test/run_tests.py` supplies the
per-project container-name values it uses. A `.gitignore` entry keeps the runner's local output
directories (`.test-logs/`, `.test-forward/`, `test-results/`) out of version control. The two
[...]

Issue #8005, opened 2025-09-10
Port test scripts and test suite management to pytest


maany:
### Description

As we have a lot of scripts required to set up the testing environment, the goal is to use pytest plugin and fixtures instead to replicate the behavior. This leads a path towards local testing and also better integration with IDEs

maany:
Hey @bziemons, I am picking you as the reviewer since the issue is XL sized. Please let me know if that is okay. This picks up where https://github.com/rucio/rucio/pull/7782 left off.

Geogouz:
> Hey [@bziemons](https://github.com/bziemons), I am picking you as the reviewer since the issue is XL sized. Please let me know if that is okay. This picks up where [#7782](https://github.com/rucio/rucio/pull/7782) left off.


![](https://media1.tenor.com/m/CAeDJHOkumQAAAAC/pokemon-i-choose-you.gif)

bziemons:
> Hey [@bziemons](https://github.com/bziemons), I am picking you as the reviewer since the issue is XL sized. Please let me know if that is okay. This picks up where [#7782](https://github.com/rucio/rucio/pull/7782) left off.

I think for a new simpler workflow (like we previously talked about) that you already added in #7782 (as `.github/workflows/simple-autotest.yml`) we need a different issue than this (looking at the description).
Also you maybe meant to add this' parent #7965 as XL to the sprint and then plan to break down tasks from there?

PR #8687, opened 2026-07-20
Add a pytest-based test runner for the dev environment

maany:
## Description

The `tests/ruciopytest/` package already exists in Rucio and holds the pytest helpers for
no-parallel test grouping. This change extends that package with a test runner that runs any Rucio
test suite with a single `pytest` command plus a few arguments, and adds two GitHub Actions
workflows that drive the suites through it. The runner handles the setup and cleanup a suite needs:
it starts the dev `docker compose` stack, sets up the database and test data, selects the tests to
run, and shuts the stack down afterwards. No separate shell scripts or matrix logic are needed to
run a suite. The new runner sits alongside the existing helpers in the package, and the current test
setup keeps working as before.

Two small, safe changes come along with the runner because it depends on them. The first is a guard
around the statsd monitoring client. When Rucio creates its statsd client it resolves the configured
monitoring host name; if that name cannot be resolved (for example in a minimal test container that
has no monitoring host), the underlying socket call raises `socket.gaierror` and the process fails
before the tests even run. The guard catches that specific lookup error so that monitoring is skipped
quietly and the tests carry on, instead of crashing. It only affects the failure path, so normal
environments where the host resolves are unaffected. The second change lets the dev compose container
names be set from environment variables, so several test runs can run at the same time without their
containers clashing. The default names are kept.

## Motivation

At the moment, running a suite means calling shell scripts and CI matrix logic and starting and
stopping containers by hand. That is awkward to reproduce locally and hard to follow. A single
`pytest <args>` command that takes care of the containers means the suites run the same way locally
and in CI. That lowers the barrier for new contributors, and it gives one clear way to run the tests
instead of several scattered scripts. The compose change keeps the current dev workflow exactly as it
is today and only lets the runner keep parallel runs apart.

The two workflows are additive. They run the suites through the runner on every pull request without
editing or disabling any existing workflow, so the new path can be exercised alongside the current CI
before anything is retired.

## Change

Most of the change is new modules under `tests/ruciopytest/` (the runner itself, its unit tests, and
a short README). Three files already in the package are updated as well:
`__init__.py` gains the shared pytest stash keys that wire the runner in, and the two existing
no-parallel helpers `xdist_noparallel_remote.py` and `xdist_noparallel_scheduler.py` are extended so
the `noparallel` marker also accepts positional group names, not only the `groups=` keyword. The
runner is registered through `tests/conftest.py`, and `tools/test/run_tests.py` supplies the
per-project container-name values it uses. A `.gitignore` entry keeps the runner's local output
directories (`.test-logs/`, `.test-forward/`, `test-results/`) out of version control. The two
supporting fixes are the statsd guard in `lib/rucio/core/monitor.py` and the container-name
parameterization in `etc/docker/dev/docker-compose.yml`, with a new
`docker-compose.test.override.yml` that provides test-scoped project names. No existing compose
behaviour is changed and all current defaults are preserved.

Two new workflow files are added: `simple-autotest.yml` runs the standard suites through the runner,
and `simplify_votests.yml` runs the VO tests (atlas and belleii) through the same runner using
`--suite=votest --policy=<X>`. Neither edits nor disables an existing workflow.

The runner ships its own unit tests next to its code. Those tests cover the runner, not Rucio, so
they should not be collected by the Rucio suite (which runs inside the test container under xdist).
A `norecursedirs` entry in `pyproject.toml` and `tools/pytest.ini` keeps them out of a plain
`pytest tests/` run, and the existing unit-tests workflow runs them explicitly alongside the other
infra-free unit tests (`pytest tests/rucio tests/ruciopytest`). That one line in
`.github/workflows/unit_tests.yml` is the only change to an existing workflow. Naming a path on the
command line still collects them, so `pytest tests/ruciopytest` works as expected.

Running a suite locally is a single `pytest` command (for example `pytest --suite=remote_dbs tests/`);
the runner brings up the containers it needs and tears them down afterwards, so no wrapper script is
required. The `tests/ruciopytest/README.md` Quickstart covers this.

## Checklist
- [x] Created issue: #8648 (and #8649, both sub-issues of #8005)
- [x] Added tests

This closes two sub-issues of #8005: #8648 (the runner) and #8649 (the workflows that drive it). They
were drafted as separate issues, but the workflows are not useful without the runner and the runner
has no CI coverage without them, so they are submitted together rather than as two dependent pull
requests. The commits stay separated by concern, with `Issue:` trailers on the intermediate commits
and `Closes:` on the commit that completes each sub-issue.

No documentation-repository PR is linked. The runner is documented in-repo at
`tests/ruciopytest/README.md`; happy to raise a documentation PR as well if that is preferred.

No database migrations, and no backwards compatibility breaking changes: the compose container names
keep their current defaults, the statsd change only affects a path that previously crashed, and no
existing workflow is disabled.

### Note before merging: CI cost while both paths run in parallel

Because the new workflows are additive and nothing existing is turned off, the old and the new test
paths run **side by side on every pull request**. That is deliberate, but it is not free, and it
should be a conscious decision rather than a surprise after merging.

Rough numbers from a single run of this pull request. These are one observation, not a benchmark, and
should be treated as indicative only.

| Path | Legs | Runner minutes (total compute) |
| --- | --- | --- |
| Existing autotest and votest workflows | 14 | about 323 |
| New plugin workflows added here | 7 | about 149 |
| Combined | 21 | about 472 |

The meaningful figure is **compute, not elapsed time**. Total runner minutes rise by roughly 149
(about 46 percent), because every job consumes runner time whether or not it runs concurrently. That
number is reasonably stable between runs.

Elapsed time is much less reliable and is deliberately not quantified here. The legs run in parallel,
so elapsed time depends mostly on how quickly runners are available: in this run the whole set spanned
about 68 minutes, but the longest single job was about 39 minutes, so roughly 28 minutes of that was
jobs waiting to start rather than jobs running. That component varies with runner pool load and says
little about the workflows themselves.

The overlap is genuine duplication rather than extra coverage: `remote_dbs`, `multi_vo`, `client`,
and the two `votest` policies are each exercised twice, once through each path.

The intent of running both is to build confidence before anything is retired. The thing worth
watching is **mismatches**: a suite that fails on the old path but passes on the new one, or the
reverse. As long as the two paths agree, the new path can be trusted to replace the old one; a
divergence is a real signal that the runner is selecting or executing something differently.

If the preference is to keep both running for a period and compare, that is fine and needs no change
here. If the duplicated cost is not acceptable, the old workflows can be disabled instead, either as
part of this pull request or as a follow-up. That decision is left to maintainers rather than assumed
here, since retiring the existing CI is a bigger call than adding an opt-in path alongside it.

<details>

```
- **Quality**: Agree 🟢
- **Testing**: Agree 🟢
- **Documentation**: Agree 🟢

# Notes for merger



```
</details>

### DIFF
--- a/.github/workflows/simple-autotest.yml
+++ b/.github/workflows/simple-autotest.yml
@@ -1,0 +1,267 @@
+name: Simple Autotests
+
+on:
+  pull_request:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  runtime_images:
+    uses: ./.github/workflows/runtime_images.yml
+    permissions:
+      contents: read
+      packages: write
+
+  test:
+    needs: [runtime_images]
+    runs-on: ubuntu-latest
+    permissions:
+      checks: write
+      contents: read
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - suite: remote_dbs
+            leg: remote_dbs
+            python: "3.9"
+            rdbms: postgres14
+          - suite: remote_dbs
+            leg: remote_dbs
+            python: "3.10"
+            rdbms: postgres14
+          # multi_vo runs as ONE leg (legacy parity): RUCIO_MULTI_VO_LEG is left unset
+          # (no `vo:` field), so InfraManager.run_multi_vo() takes the sequential
+          # shared-DB path -- tst first, then ts2 only on tst success, no DB reset
+          # between VOs. One shared instance, matching run_multi_vo_tests_docker.sh.
+          - suite: multi_vo
+            leg: multi_vo
+            python: "3.9"
+            rdbms: postgres14
+          - suite: client
+            leg: client
+            python: "3.9"
+            rdbms: postgres14
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5.0.1
+
+      - name: Set up Python
+        uses: actions/setup-python@e797f83bcb11b83ae66e0230d6156d7c80228e7c # v6.0.0
+        with:
+          python-version: ${{ matrix.python }}
+          cache: pip
+          cache-dependency-path: requirements/requirements.dev.txt
+
+      - name: Install host plugin deps
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install -c requirements/requirements.dev.txt pytest pytest-xdist pyyaml
+
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@c94ce9fb468520275223c153574b00df6fe4bcc9 # v3.7.0
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+        continue-on-error: true
+
+      - name: Set up Docker Buildx
+        if: needs.runtime_images.outputs.build_locally == 'true'
+        uses: docker/setup-buildx-action@8d2750c68a42422c14e847fe6c8ac0403b4cbd6f # v3.12.0
+
+      - name: Compute image metadata
+        if: needs.runtime_images.outputs.build_locally == 'true'
+        id: imgmeta
+        run: |
+          echo "repo=$(echo '${{ github.repository }}' | tr '[:upper:]' '[:lower:]')" >> $GITHUB_OUTPUT
+          echo "pynodot=$(echo '${{ matrix.python }}' | tr -d '.')" >> $GITHUB_OUTPUT
+
+      - name: Build runtime image locally
+        if: needs.runtime_images.outputs.build_locally == 'true'
+        uses: docker/build-push-action@471d1dc4e07e5cdedd4c2171150001c434f0b7a4 # v6.15.0
+        with:
+          context: .
+          file: etc/docker/test/runtime.Dockerfile
+          target: final
+          build-args: |
+            PYTHON=${{ matrix.python }}
+          push: false
+          tags: ${{ matrix.python == '3.10' && needs.runtime_images.outputs.py310_image || needs.runtime_images.outputs.py39_image }}
+          outputs: type=docker
+          cache-from: type=registry,ref=ghcr.io/${{ steps.imgmeta.outputs.repo }}/rucio-dev-runtime:py${{ steps.imgmeta.outputs.pynodot }}-buildcache
+
+      # ----------------------------------------------------------------
+      # Host-side `client` leg provisioning (legacy autotest path).
+      # Gated to matrix.suite == 'client' so the other 4 legs are untouched.
+      # Mirrors tools/test/test.sh SUITE=client: full rucio install +
+      # `tools/run_tests.sh -i` bootstrap + rucio_client.cfg/cert wiring.
+      # ----------------------------------------------------------------
+      - name: Install full rucio on host (client)
+        if: matrix.suite == 'client'
+        continue-on-error: true
+        run: |
+          # Match the runtime container's env: editable rucio + server + dev reqs.
+          # requirements.client.txt is intentionally NOT installed here: it pins
+          # urllib3>=2.5.0 which is irreconcilable with the server/dev pin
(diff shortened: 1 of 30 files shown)


Geogouz:
I think this is going to be handy for this PR 🥲: https://rucio-review.cern.ch/pr/8687

.github/workflows/simple-autotest.yml
```diff
@@ -0,0 +1,267 @@
...
+          fi
+          echo "reachable=$reachable" >> "$GITHUB_OUTPUT"
+
+      - name: Run tests
+        if: matrix.suite != 'client' || steps.client_provision.outputs.reachable == 'true'
+        env:
+          RUCIO_TEST_IMAGE: ${{ matrix.python == '3.10' && needs.runtime_images.outputs.py310_image || needs.runtime_images.outputs.py39_image }}
+          RDBMS: ${{ matrix.rdbms }}
+          PYTHON: ${{ matrix.python }}
+          POLICY: ${{ matrix.policy }}
```
bziemons, line 172: policy does not seem to be defined in the matrix

.github/workflows/simple-autotest.yml
```diff
@@ -0,0 +1,267 @@
...
+  test:
+    needs: [runtime_images]
+    runs-on: ubuntu-latest
+    permissions:
+      checks: write
+      contents: read
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
```
bziemons, line 26: why this step towards GH dependence on defining the matrix? Isn't this in the way of running CI tests locally?

.github/workflows/simple-autotest.yml
```diff
@@ -0,0 +1,267 @@
...
+            tests/ 2>&1 | tee test-results/${{ matrix.leg }}-py${{ matrix.python }}.pytest.log
+
+      # ----------------------------------------------------------------
+      # Legacy-faithful in-container fallback for the client leg.
+      # Fires only when host-side provisioning left the server unreachable
+      # or host collection failed to import (reachable != 'true'). Reproduces
+      # exactly what tools/test/test.sh SUITE=client did: cp rucio_client.cfg
+      # then pytest the 3 client files INSIDE the rucio container. Writes junit
+      # to the same test-results/ path so the report step still renders.
+      # ----------------------------------------------------------------
```
bziemons, line 197: what is the motivation for the fallback? Why would the server be unreachable or fail the import test?

.github/workflows/simple-autotest.yml
bziemons, line 26: why is there no test with Oracle?

.github/workflows/simple-autotest.yml
```diff
@@ -0,0 +1,267 @@
+name: Simple Autotests
```
bziemons, line 1: not exactly sure about the file name. Is this to be replaced with a follow-up when replacing the previous workflows?

.github/workflows/simple-autotest.yml
```diff
@@ -0,0 +1,267 @@
...
+          include-hidden-files: true
+          path: |
+            .test-forward/
+            test-results/${{ matrix.leg }}-py${{ matrix.python }}.pytest.log
+          retention-days: 7
+          if-no-files-found: ignore
+
+      - name: Publish test report
+        if: always()
+        uses: mikepenz/action-junit-report@v6
```
bziemons, line 263: this version should be pinned (along with other missing pins)

.github/workflows/simple-autotest.yml
```diff
@@ -0,0 +1,267 @@
...
+          # stdout + JSONL report stream; include-hidden-files is required or the
+          # in-container traceback for a red leg is silently dropped.
+          include-hidden-files: true
+          path: |
+            .test-forward/
+            test-results/${{ matrix.leg }}-py${{ matrix.python }}.pytest.log
+          retention-days: 7
+          if-no-files-found: ignore
+
+      - name: Publish test report
```
bziemons, line 261: is this set up correctly? I do not see the report on this PR

bziemons (commented):
Review 1/x

Mostly questions. One additional here: Will the integration tests be runnable with the same framework? If so, how do you plan to integrate them?

PR #8708, opened 2026-07-23
fix(Core): allow OIDC login without a pre-declared account

### LINKED ISSUE #8689: OIDC login without X-Rucio-Account is no longer possible since 41.0.0 (webui mock-account flow removed)
### Description

Since 41.0.0 it is no longer possible to start an OIDC authentication flow via `GET /auth/oidc` without specifying the target account up front in the `X-Rucio-Account` header. This should be possible: the account can be derived from the OIDC identity once the IdP has authenticated the user, and requiring the account before authentication breaks clients that cannot know it in advance.

Up to and including 40.x, a request without `X-Rucio-Account` worked as designed. The REST layer defaults the account to `webui` (`lib/rucio/web/rest/flaskapi/v1/auth.py`), and `get_auth_oidc` in `lib/rucio/core/oidc.py` treated `webui` as a mock account, skipping the `account_exists` check; the real account was assigned to the identity later, during `get_token_oidc`.

Commit eb3aa6297 ("feat(Core)!: remove auto OIDC login flow from core", part of #8318, first released in 41.0.0rc1) removed this special case as a side effect of removing the `auto` login flow. Every request now goes through `account_exists`, and since `webui` is not a real account, any account-less request fails with:

```
401 CannotAuthenticate: Cannot get authentication URL from Rucio Authentication Server for account webui
```

Note that the REST layer still defaults to `webui`, so the current default of the endpoint is an account that the core is guaranteed to reject. Even if the account-less flow were not restored, this leftover would need cleaning up (with a clear "X-Rucio-Account header required" error instead).

The concrete use case affected: OAuth bridges such as rucio-mcp, which bridge MCP clients (Claude Code, Codex, the MCP TypeScript SDK) to Rucio's OIDC polling flow. These clients follow standard OAuth and have no mechanism to attach a per-user Rucio account name to the authorisation request before the user has authenticated. This worked end-to-end against ATLAS until the upgrade to 41.0.0rc1, and still works against VOs on pre-41 releases (verified against ESCAPE VRE on 35.8.0).

The suggestion is to restore the ability to authenticate without a pre-declared account, by resolving the account from the identity (e.g. its default account) at token time in `get_token_oidc`, as the pre-41 flow did for the polling case.

### Steps to reproduce

1. Against a 41.0.0rc1 server (e.g. `atlas-rucio-auth.cern.ch`):
   ```
   curl -i "https://atlas-rucio-auth.cern.ch/auth/oidc" \
        -H "X-Rucio-Client-Authorize-Polling: true"
   ```
   Result: `401`, `ExceptionClass: CannotAuthenticate`, "Cannot get authentication URL ... for account webui".
2. Same request with `-H "X-Rucio-Account: <existing account>"`: `200` with `X-Rucio-OIDC-Auth-URL`.
3. Same request as step 1 against a pre-41 server (e.g. ESCAPE VRE, 35.8.0): `200` with `X-Rucio-OIDC-Auth-URL`.

### Rucio Version

Server: 41.0.0rc1 (regressed), 40.x and earlier fine (verified 35.8.0)

### Additional Information

- Server side. Introduced by eb3aa6297, first contained in tag `41.0.0rc1`; related commits `c115db858` (clients) and `065580efc` (gateway/REST) from the #8318 series.
- Not related to IAM client registration; the `webui` account was never an IAM-side entity, only a core-side mock value.

_Partially written/tested with AI_

bari12:

## Description
Restore the pre-41.0.0 account-less /auth/oidc flow removed by eb3aa6297: requests without X-Rucio-Account skip the account existence check, and get_token_oidc assigns the default account of the authenticated identity.

Tests were assisted by AI

## Checklist

- [X] This PR closes #8689 
      
- [X] This PR contains no breaking changes, or the breaking change is described

## Notes for contributors

  should remain open.
  will be assigned.
  closed promptly.

<details>

```
- **Quality**: Agree 🟢
- **Testing**: Agree 🟢
- **Documentation**: Agree 🟢

# Notes for merger



```
</details>

### DIFF
--- a/lib/rucio/web/rest/flaskapi/v1/auth.py
+++ b/lib/rucio/web/rest/flaskapi/v1/auth.py
@@ -275,7 +275,8 @@ class OIDC(ErrorHandlingMethodView):
         parameters:
         - name: HTTP_X_RUCIO_ACCOUNT
           in: header
-          description: "Account identifier as a string."
+          description: "Account identifier as a string. If not given, the default
+                        account of the authenticated identity is used."
           schema:
             type: string
         - name: HTTP_X_RUCIO_CLIENT_AUTHORIZE_SCOPE
@@ -321,7 +322,9 @@ class OIDC(ErrorHandlingMethodView):
         headers.set('Pragma', 'no-cache')
 
         vo = extract_vo(request.headers)
-        account = request.environ.get('HTTP_X_RUCIO_ACCOUNT', 'webui')
+        # if no account is given, it will be derived from the identity
+        # once the user has authenticated at the Identity Provider
+        account = request.environ.get('HTTP_X_RUCIO_ACCOUNT', None)
         auth_scope = request.environ.get('HTTP_X_RUCIO_CLIENT_AUTHORIZE_SCOPE', "")
         audience = request.environ.get('HTTP_X_RUCIO_CLIENT_AUTHORIZE_AUDIENCE', "")
         issuer = request.environ.get('HTTP_X_RUCIO_CLIENT_AUTHORIZE_ISSUER', None)
--- a/tests/test_oidc.py
+++ b/tests/test_oidc.py
@@ -552,6 +552,70 @@ class TestAuthCoreAPIoidc:
 
     @patch('rucio.core.oidc.__get_init_oidc_client')
     @patch('rucio.core.oidc.__get_rucio_oidc_clients')
+    def test_get_access_token_oidc_account_less_polling_success(self, mock_clients, mock_oidc_client):
+        """ OIDC Request for access token without a pre-declared account - success
+
+            Runs the Test:
+
+            - initialising the auth flow with the 'webui' sentinel account,
+              i.e. without specifying a Rucio account up front
+            - filling the right identity into the token (mocking the IdP response)
+            - calling the get_token_oidc core function
+
+            End:
+
+            - checking that the token is in the DB and got assigned
+              to the default account of the identity
+        """
+        mock_oidc_client.side_effect = get_mock_oidc_client
+        sentinel_account = InternalAccount('webui', **self.vo)
+        auth_init_response = self.get_auth_init_and_mock_response(code_response=rndstr(), account=sentinel_account, polling=True, session=self.db_session)
+        oauth_session_row = get_oauth_session_row(sentinel_account, state=auth_init_response['state'], session=self.db_session)
+        assert oauth_session_row
+        # mocking the token response
+        access_token = rndstr()
+        NEW_TOKEN_DICT['access_token'] = access_token
+        NEW_TOKEN_DICT['id_token'] = {'sub': 'knownsub', 'iss': 'https://test_issuer/', 'nonce': auth_init_response['nonce']}
+        token_dict = get_token_oidc(auth_init_response['auth_query_string'], session=self.db_session)
+        assert token_dict
+        assert token_dict['polling'] is True
+        assert 'token' not in token_dict
+        # the token must have been assigned to the default account
+        # of the identity, not to the sentinel account
+        db_token = get_token_row(access_token, session=self.db_session)
+        assert db_token
+        # not asserting == self.accountstring: 'knownsub' is remapped to a fresh
+        # account each setup_method, so the oldest mapping may be an earlier test's
+        assert db_token.account.external != 'webui'
+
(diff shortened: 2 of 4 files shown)


lib/rucio/web/rest/flaskapi/v1/auth.py
```diff
@@ -321,7 +322,9 @@ def get(self) -> 'ResponseReturnValue':
         headers.set('Pragma', 'no-cache')
 
         vo = extract_vo(request.headers)
-        account = request.environ.get('HTTP_X_RUCIO_ACCOUNT', 'webui')
+        # if no account is given, it will be derived from the identity
+        # once the user has authenticated at the Identity Provider
+        account = request.environ.get('HTTP_X_RUCIO_ACCOUNT', None)
```
bziemons, line 327: not exactly sure why this change and https://github.com/rucio/rucio/commit/7a3206453a68b296aef46160acc77ed32246d7ec#diff-6580a9a2e98875f6892626668075efb5ad79c567807c9c5bff1f09ed16e49662R101-R103 is necessary🤨

tests/test_oidc.py
```diff
@@ -550,6 +550,68 @@ def test_get_access_token_oidc_cli_polling_success(self, mock_clients, mock_oidc
...
+        NEW_TOKEN_DICT['id_token'] = {'sub': 'knownsub', 'iss': 'https://test_issuer/', 'nonce': auth_init_response['nonce']}
+        token_dict = get_token_oidc(auth_init_response['auth_query_string'], session=self.db_session)
+        assert token_dict
+        assert token_dict['polling'] is True
+        assert 'token' not in token_dict
+        # the token must have been assigned to the default account
+        # of the identity, not to the sentinel account
+        db_token = get_token_row(access_token, session=self.db_session)
+        assert db_token
+        assert db_token.account.external != 'webui'
```
bziemons, line 589: wouldn't the more specific check `db_token.account.external == self.accountstring` make more sense?

bziemons (commented):
I think it looks fine, compared to the code that we had before the removal. Also, I like that there are tests for this. Please check the review comments and if you want to not action on them, resolve them.

lib/rucio/web/rest/flaskapi/v1/auth.py
bari12: My thought mainly was that we shouldn't really put this at the REST level, and keep that interface as "dumb" as possible, so it shouldn't implement any business logic. 

The inject (for empty accounts) now happens at the gateway layer https://github.com/rucio/rucio/pull/8708/changes#diff-6580a9a2e98875f6892626668075efb5ad79c567807c9c5bff1f09ed16e49662R103

But yes, you are right in a way that it isn't a pure revert (Though it wasn't advertised as one either).

bziemons: It is fine. The change is simple enough, but I wasn't expecting this.

tests/test_oidc.py
bari12: Hmm that would make it a bit stronger. I am wondering though if there is a risk that this gets flaky. E.g. the `knownsub` identity might be defaulted to another account, in another test, and `self.accountstring` might not be the default account at the point where the check runs. the other test on L549 even mentions that.

We could still do it by giving this test it's special own identity to test though. Not sure if this is worth it though.

bziemons: I see that comment now. Maybe copy that here then, instead. Then it is clear here, why it is not checked explicitly. I don't like that this side-effect of other tests exists, but this is the wrong place to discuss this.

bari12: I am not so sure, because i think the setup_method runs for every test? In any way, I will add a comment to it.

bziemons: That is correct, according to https://docs.pytest.org/en/stable/how-to/xunit_setup.html
I thought this works like `setup_class`, however, they are distinct.

Either way, it would be nice to update the surroundings, remove side-effects and make these tests work in parallel at some point in the future (possibly with fixtures).

bari12: Agreed. For now I just want to avoid that the thing goes flaky. I think the test (mostly) achieves what it should; should be fine.

bziemons (approved):
- **Quality**: Agree 🟢
- **Security**: Agree 🟢
- **Testing**: Agree 🟢
- **Documentation**: Agree 🟢

PR #8724, opened 2026-08-02
feat(Clients): add rucio rse distance list command

### LINKED ISSUE #8630: CLI: allow `rucio rse distance show` with only one RSE
### Description

To know which possible connections a particular RSE has, it would be useful to allow 

```
rucio rse distance show <rse>
```

which should return a list of distances for this RSE (both source and target)

### Motivation

_No response_

### Change

Currently, this raises an error about the missing required second RSE:

```
2026-06-26 13:38:08,003 ERROR   Missing parameter: destination_rse. Please check the command help (-h/--help).
```

ASHISHTOMER0817:
Closes: #8630

## Description

Add `rucio rse distance list source-rse` The command prints a DESTINATION /
DISTANCE table for matching RSE pairs.

Make `destination` optional in the client, REST, and gateway `get_distance`
paths so callers can list all distances from a source RSE via
`GET /rses/<source>/distances`. Pair lookup and existing `distance show`,
`set`, and `unset` behaviour are unchanged.

## Checklist

- [x] This PR closes #8630
      — Added `TestRSEClient::test_get_distance_without_destination` and
      `test_rse_distance_list`; existing `test_add_distance` and
      `test_rse_distance` cover regression.
- [ ] Documentation is updated (link the docs PR here), or no documentation change is needed
      — CLI `--help` on `distance list` is sufficient; no separate docs change.
- [ ] Database migrations are included, or the change touches no database schema
      — Uses existing distance tables; no schema change.
      — `distance show` and pair `get_distance(source, destination)` unchanged.

## Notes for contributors

  should remain open.
  will be assigned.
  closed promptly.

<details>

- **Quality**: This PR is well tested: [*Agree 🟢*]

### DIFF
--- a/lib/rucio/cli/rse.py
+++ b/lib/rucio/cli/rse.py
@@ -18,6 +18,7 @@ import click
 from rich.padding import Padding
 from rich.text import Text
 from rich.tree import Tree
+from tabulate import tabulate
 
 from rucio.cli.utils import RichCLITheme, RichUtils
 from rucio.common.exception import InputValidationError
@@ -254,6 +255,27 @@ def distance_show(ctx: click.Context, source_rse: str, destination_rse: str) -> 
             print(f'Distance information from {source_rse} to {destination_rse}: distance={distance}')
         else:
             print(f"No distance set from {source_rse} to {destination_rse}")
+
+
+@distance.command("list")
+@click.argument("source-rse")
+@click.pass_context
+def distance_list(ctx: click.Context, source_rse: str) -> None:
+    """Display all destinations and distances for SOURCE-RSE"""
+    distances = ctx.obj.client.get_distance(source_rse)
+    if not distances:
+        print(f"No distances found for source RSE '{source_rse}'.")
+        return
+    table_rows = [[row['dest_rse'], row['distance']] for row in distances]
+    if ctx.obj.use_rich:
+        table = RichUtils.generate_table(
+            table_rows,
+            headers=['DESTINATION', 'DISTANCE'],
+            col_alignments=['left', 'right']
+        )
+        RichUtils.print_output(table, console=ctx.obj.console, no_pager=ctx.obj.no_pager)
+    else:
+        print(tabulate(table_rows, tablefmt=ctx.obj.tablefmt, headers=['DESTINATION', 'DISTANCE']))
 
 
 @distance.command("set")
--- a/lib/rucio/client/rseclient.py
+++ b/lib/rucio/client/rseclient.py
@@ -1504,17 +1504,17 @@ class RSEClient(BaseClient):
     def get_distance(
             self,
             source: str,
-            destination: str
+            destination: Optional[str] = None
     ) -> list[dict[str, Union[str, int]]]:
         """
         Get distances between rses.
 
-        Param
+        Parameters
         ----------
         source :
             The source RSE.
         destination :
-            The destination RSE.
+            The destination RSE. If None, returns all distances from source RSE.
 
         Returns
         -------
@@ -1530,7 +1530,10 @@ class RSEClient(BaseClient):
             ** ranking ** [int]: Legacy name for distance, same value as distance.
 
         """
-        path = [self.RSE_BASEURL, source, 'distances', destination]
+        if destination is None:
+            path = [self.RSE_BASEURL, source, 'distances']
+        else:
+            path = [self.RSE_BASEURL, source, 'distances', destination]
         path = '/'.join(path)
         url = build_url(choice(self.list_hosts), path=path)
         r = self._send_request(url, method=HTTPMethod.GET)
--- a/lib/rucio/gateway/rse.py
+++ b/lib/rucio/gateway/rse.py
@@ -520,21 +520,24 @@ def update_distance(
                                                 distance=distance, bidirectional=bidirectional, session=session)
 
 
-def get_distance(source, destination, issuer, vo=DEFAULT_VO):
+def get_distance(source: str, issuer: str, destination: "Optional[str]" = None, vo: str = DEFAULT_VO):
     """
-    Get distances between rses.
+    Get distances between RSEs.
 
     :param source: The source RSE.
-    :param destination: The destination RSE.
+    :param destination: The destination RSE. If None, return all distances from source.
     :param issuer: The issuer account.
     :param vo: The VO to act on.
 
-    :returns distance: List of dictionaries.
+    :returns distance: List of dictionaries with distance information.
     """
     with db_session(DatabaseOperationType.READ) as session:
-        distances = distance_module.get_distances(src_rse_id=rse_module.get_rse_id(source, vo=vo, session=session),
-                                                  dest_rse_id=rse_module.get_rse_id(destination, vo=vo, session=session),
-                                                  session=session)
+        kwargs = {
+            'src_rse_id': rse_module.get_rse_id(source, vo=vo, session=session)
+        }
+        if destination is not None:
+            kwargs['dest_rse_id'] = rse_module.get_rse_id(destination, vo=vo, session=session)
+        distances = distance_module.get_distances(session=session, **kwargs)
 
         return [gateway_update_return_dict(d, session=session) for d in distances]
 
--- a/tests/test_cli_client_structure.py
+++ b/tests/test_cli_client_structure.py
@@ -760,9 +760,9 @@ def test_rse_protocol(file_config_mock):
 
 
 @with_each_cli_renderer
-def test_rse_distance(file_config_mock):
-    source_rse = "MOCK"
-    dest_rse = "MOCK2"
+def test_rse_distance(file_config_mock, rse_factory):
+    source_rse, _ = rse_factory.make_posix_rse()
+    dest_rse, _ = rse_factory.make_posix_rse()
 
     cmd = f"rucio rse distance unset {source_rse} {dest_rse}"
     exitcode, out, err = execute(cmd)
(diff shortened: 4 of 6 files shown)


ASHISHTOMER0817:
@dchristidis PR is ready for review. It is implementing a new CLI command - `rucio rse distance list` with optional destination rse filtering. 
All tests passing.

tests/test_rse.py
```diff
@@ -1600,6 +1600,26 @@ def test_add_distance_bidirectional(self, rucio_client):
         assert len(rucio_client.get_distance(source=source, destination=destination)) == 0
         assert len(rucio_client.get_distance(source=destination, destination=source)) == 0
 
+    @pytest.mark.dirty
+    def test_get_distance_without_destination(self, rucio_client):
+        """ RSE (CLIENTS): List RSE distance from source."""
+        source, dest1, dest2 = rse_name_generator(), rse_name_generator(), rse_name_generator()
```
voetberg: Instead of making RSE like this, I recommend using the `rse_factory` fixture instead, so they're cleaned up afterwards. You can make 3 posix RSEs like this: 

```python
source, _ = rse_factory.make_posix_rse()
dest1, _ = rse_factory.make_posix_rse()
dest2, _ = rse_factory.make_posix_rse()
```

tests/test_cli_client_structure.py
```diff
@@ -796,6 +797,33 @@ def test_rse_distance(file_config_mock):
     assert "ERROR" not in err
 
 
+@pytest.mark.noparallel(reason='uses global RSEs (MOCK/MOCK2)')
```
voetberg: The first test probably shouldn't use MOCK/MOCK2 either, but using the `rse_factory` fixture here is preferable to avoid the noparallel slowdown.

lib/rucio/client/rseclient.py
```diff
@@ -1524,7 +1524,11 @@ def get_distance(
             ** ranking ** [int]: Legacy name for distance, same value as distance.
 
         """
-        path = [self.RSE_BASEURL, source, 'distances', destination]
+
+        if destination is None:
+            path = [self.RSE_BASEURL, source, 'distances']
+        else:
+            path = [self.RSE_BASEURL, source, 'distances', destination]
```
voetberg, line 1531: For backwards compatibility, it would be a good idea to put in a specific "parameter miss" or otherwise missing endpoint check to pass forward.

lib/rucio/client/rseclient.py
```diff
@@ -1498,7 +1498,7 @@ def update_distance(
     def get_distance(
             self,
             source: str,
-            destination: str
+            destination: Optional[str] = None
     ) -> list[dict[str, Union[str, int]]]:
         """
         Get distances between rses.
```
voetberg, line 1505: Docstring also needs to be updated

lib/rucio/cli/rse.py
```diff
@@ -256,6 +256,40 @@ def distance_show(ctx: click.Context, source_rse: str, destination_rse: str) ->
             print(f"No distance set from {source_rse} to {destination_rse}")
 
 
+@distance.command("list")
+@click.option("-s", "--source-rse-expression", required=True, help="Source RSE expression to use as a filter")
+@click.option("-d", "--destination-rse-expression", required=False, help="Destination RSE expression to use as a filter")
+@click.pass_context
+def distance_list(ctx: click.Context, source_rse_expression: str, destination_rse_expression: Optional[str] = None) -> None:
```
voetberg: A few things: 
1. I'm not sold on this being implemented as expressions, I think it would be much cleaner to not include the destination as an option and keeping the source as a full RSE. 
2. The required source rse shouldn't be a option, it should be an argument (e.g. supplied as `rucio rse distance list RSE-NAME` and not `rucio rse distance list --source-rse-expression RSES`
3. If we keep them are rse-expressions, what's the intended behavior for supplying the destination? It'll filter out RSEs that don't match the destination? This seems like a fairly niche use case that makes general usage confusing.

lib/rucio/cli/rse.py
```diff
@@ -256,6 +256,40 @@ def distance_show(ctx: click.Context, source_rse: str, destination_rse: str) ->
...
+                dest = dest_item["rse"]
+                distance_info = ctx.obj.client.get_distance(source, dest)
+                if distance_info:
+                    row = distance_info[0]
+                    table_data.append([row["src_rse"], row["dest_rse"], str(row["distance"])])
+        else:
+
+            for row in ctx.obj.client.get_distance(source):
+                table_data.append([row["src_rse"], row["dest_rse"], str(row["distance"])])
+    if table_data:
```
voetberg: The Rich implementation should only be used under the `if ctx.obj.use_rich` condition, Rich is still the experimental phase and shouldn't be the default display.

lib/rucio/gateway/rse.py
```diff
@@ -532,10 +532,13 @@ def get_distance(source, destination, issuer, vo=DEFAULT_VO):
     :returns distance: List of dictionaries.
     """
     with db_session(DatabaseOperationType.READ) as session:
-        distances = distance_module.get_distances(src_rse_id=rse_module.get_rse_id(source, vo=vo, session=session),
+
+        if destination is None:
+            distances = distance_module.get_distances(src_rse_id=rse_module.get_rse_id(source, vo=vo, session=session), session=session)
+        else:
+            distances = distance_module.get_distances(src_rse_id=rse_module.get_rse_id(source, vo=vo, session=session),
```
voetberg: What's the difference between these two calls? Maybe I'm missing something obvious but these are identical.

lib/rucio/gateway/rse.py
```diff
@@ -532,10 +532,13 @@ def get_distance(source, destination, issuer, vo=DEFAULT_VO):
...
                                                   dest_rse_id=rse_module.get_rse_id(destination, vo=vo, session=session),
                                                   session=session)
-
```
voetberg, line 538: Please don't randomly remove break lines when you're not otherwise changing the code around it, it clutters the blame.

lib/rucio/cli/rse.py
ASHISHTOMER0817: What I understand from the feedback is we aren't passing destination at all, it will only include source(should be an argument) and list all the pairs of this source. am I going in the right direction?

lib/rucio/gateway/rse.py
ASHISHTOMER0817: The difference is that else branch also passes dest_rse_id whereas if branch only passes src_rse_id. So the former returns a single pair, the latter all distances from the source.

ASHISHTOMER0817:
@voetberg I have addressed the mentioned changes and also made a few small improvements that weren't requested:

Added type hints to gateway.get_distance (Optional[str] for destination)
Refactored the gateway call to use a kwargs dict so the src_rse_id / dest_rse_id distinction is clearer

lib/rucio/gateway/rse.py
voetberg: Oh sorry, my confusion was caused by gh cutting off the rest of the second `get_distances` call.

lib/rucio/cli/rse.py
voetberg: That's about what I mean. I would expect the signature to look something like

```python
@click.argument(source_rse_name)
@click.pass_context
def distance_list(ctx: click.Context, source_rse_name: str) -> None:
   '''Check all the distances between [SOURCE-RSE-NAME] and other RSEs.'''
    ...
```

And to return a table like: 

```
DESTINATION | DISTANCE
-----------------
RSE1        | 1
RSE2        | 1
```

lib/rucio/cli/rse.py
```diff
@@ -256,6 +257,27 @@ def distance_show(ctx: click.Context, source_rse: str, destination_rse: str) ->
             print(f"No distance set from {source_rse} to {destination_rse}")
 
 
+@distance.command("list")
+@click.argument("source-rse")
+@click.pass_context
+def distance_list(ctx: click.Context, source_rse: str) -> None:
+    """Display all distances from SOURCE-RSE"""
```
voetberg: Sorry we've reach the nit-pick phase of the review. Because of the non-bidirectional nature of distances, the wording here is a little confusing. Can I instead recommend (and feel free to workshop)

```suggestion
    """Display all destinations and distances for SOURCE-RSE"""
```

lib/rucio/client/rseclient.py
```diff
@@ -1524,12 +1525,17 @@ def get_distance(
...
+            path = [self.RSE_BASEURL, source, 'distances']
+        else:
+            path = [self.RSE_BASEURL, source, 'distances', destination]
         path = '/'.join(path)
         url = build_url(choice(self.list_hosts), path=path)
         r = self._send_request(url, method=HTTPMethod.GET)
         if r.status_code == codes.ok:
             return next(self._load_json_data(r))
+        if destination is None and r.status_code == codes.not_found:
+            raise RucioException('Listing distances without a destination requires a newer Rucio server.')
```
voetberg, line 1538: I think we can make the error more specific (RucioException is the vaguest error we can supply). I'd recommend instead changing the exception a 501 (Not Implemented server side) error and using the same  self.__get_exception method to grab it.

voetberg:
One last thing - looking at the commit history, changes from the review should be squashed onto the original commit instead of being added as separate commits later. There should also be separate commits for the things impacting each component (e.g. the core changes/gateway changes should not be in the same commit as the client facing changes, and the testing and documentation changes should also be their own commits)

lib/rucio/client/rseclient.py
ASHISHTOMER0817: Removed the client-side 404 special case; get_distance now always uses _get_exception. On upgraded servers this returns 200 or RSENotFound. Old servers without the route still return bare Flask 404 — should we add explicit 501 handling in rses.py for a specific case, or is the client simplification enough?

ASHISHTOMER0817:
@voetberg CI is failing on a flake unrelated to this PR: test_help_menus / ping --help with signal only works in main thread of the main interpreter under xdist.

Could you re-run the failed job when you have time and take a final look at my PR. https://github.com/rucio/rucio/actions/runs/33066523804/job/98497912429?pr=8724#logs

voetberg:
> @voetberg CI is failing on a flake unrelated to this PR: test_help_menus / ping --help with signal only works in main thread of the main interpreter under xdist.
> 
> Could you re-run the failed job when you have time and take a final look at my PR. https://github.com/rucio/rucio/actions/runs/33066523804/job/98497912429?pr=8724#logs

Done - can you make an issue for the flaky test?

ASHISHTOMER0817:
@voetberg -- Will open an issue for the flaky `test_help_menus` / signal-under-xdist failure.

ASHISHTOMER0817:
@voetberg Tracked the flaky `test_help_menus` failure in #8780 . Could you re-run the failed autotest job when you have time?

voetberg:
I did rerun it and it failed again.... maybe we should look more closely at it

PR #8731, opened 2026-08-06
Transfers: Return latest request history by DID #8722

### LINKED ISSUE #8722: get_request_history_by_did returns a random request instead of the latest one
### Description

`get_request_history_by_did` in `lib/rucio/core/request.py` selects from REQUESTS_HISTORY filtered by scope, name and rse_id. However, there is no order by clause and `scalar()` which is then called on the statement returns the first row. 
```python
stmt = select(
    models.RequestHistory
).where(
    and_(models.RequestHistory.scope == scope,
         models.RequestHistory.name == name,
         models.RequestHistory.dest_rse_id == rse_id)
)
# ...
tmp = session.execute(stmt).scalar()
```
For requests that fail, they will have multiple rows in the RequestHistory table. The inserted rows will always be in chronological order (oldest first). I tried this out with PostgreSQL and it gave me the oldest of the attempts. The client method documents it as "Return latest request details for a DID".
This means that, when debugging failed requests, on some environments, they might be getting the wrong requests.

`get_request_by_did` has the same pattern, but it's not as important as there should only be 1 live request, however I don't think it is enforced anywhere. 

### Steps to reproduce

I used a naive test to reproduce this:
```python
def test_get_request_history_by_did_returns_latest_real_workflow(rse_factory, mock_scope, root_account, db_session):
    _, source_rse_id = rse_factory.make_mock_rse(session=db_session)
    _, dest_rse_id = rse_factory.make_mock_rse(session=db_session)
    name = generate_uuid()
    add_replica(source_rse_id, mock_scope, name, 1, root_account, session=db_session)

    queue_requests([{
        'dest_rse_id': dest_rse_id,
        'src_rse_id': source_rse_id,
        'request_type': RequestType.TRANSFER,
        'request_id': generate_uuid(),
        'name': name,
        'scope': mock_scope,
        'rule_id': generate_uuid(),
        'retry_count': 0,
        'attributes': {
            'activity': 'User Subscription',
            'bytes': 1,
            'md5': '',
            'adler32': ''
        }
    }], session=db_session)

    # 4 failed attempts
    for _ in range(4):
        req = get_request_by_did(mock_scope, name, dest_rse_id, session=db_session)
        transition_request_state(req['id'], state=RequestState.FAILED, session=db_session)
        requeue_and_archive({'request_id': req['id']}, source_ranking_update=False, session=db_session)

    request = get_request_history_by_did(mock_scope, name, dest_rse_id, session=db_session)
    assert request['retry_count'] == 3, (
        f"expected the latest attempt (retry_count=3), got retry_count={request['retry_count']} "
        f"(created_at={request['created_at']})"
    )
```
Result on the dev env (PostgreSQL) is reproducible every time. 

### Rucio Version

_No response_

### Additional Information

Suggested fix:
- add `.order_by(models.RequestHistory.created_at.desc())` to the query
- update some of the docs to mention it will be the latest request (right now it's specified just in one of the pydocs)
- add (at least) the regression test

divyanshu-iitian:

## Description

Orders matching `RequestHistory` rows by descending creation time before selecting one, so `get_request_history_by_did` consistently returns the latest retry instead of a database-dependent row. The core and gateway docstrings now state this contract, and a deterministic regression test inserts two matching histories and verifies the newest retry is returned.

No user-facing documentation or database migration is needed: this corrects an existing retrieval contract without changing the schema or public signature.

## Checklist

- [x] This PR closes #8722

## Validation

- `ruff check lib/rucio/core/request.py lib/rucio/gateway/request.py tests/test_request.py`
- Linux / Python 3.13 / SQLite: focused regression test (`1 passed`)
- Linux / Python 3.13 / SQLite: affected core request tests (`5 passed`)

<details>
<summary>Reviewer template</summary>

```text
- **Confidence in review**: High / Medium / Low
- **Confidence in scope**: High / Medium / Low
- **Quality**: Agree
- **Security**: Agree / Disagree
- **Backwards compatibility**: Agree / Disagree
- **Testing**: Agree
- **Documentation**: Agree

# Notes for merger
```
</details>

### DIFF
--- a/lib/rucio/core/request.py
+++ b/lib/rucio/core/request.py
@@ -1192,7 +1192,7 @@ def get_request_history_by_did(
     session: "Session"
 ) -> dict[str, Any]:
     """
-    Retrieve a historical request by its DID for a destination RSE.
+    Retrieve the latest historical request by its DID for a destination RSE.
 
     :param scope:          The scope of the data identifier.
     :param name:           The name of the data identifier.
@@ -1214,6 +1214,11 @@ def get_request_history_by_did(
             stmt = stmt.where(
                 models.RequestHistory.request_type == request_type
             )
+
+        stmt = stmt.order_by(
+            models.RequestHistory.created_at.desc(),
+            models.RequestHistory.retry_count.desc(),
+        ).limit(1)
 
         tmp = session.execute(stmt).scalar()
         if not tmp:
--- a/tests/test_request.py
+++ b/tests/test_request.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 import json
-from datetime import datetime
+from datetime import datetime, timedelta
 from typing import Union
 
 import pytest
@@ -23,7 +23,7 @@ from rucio.common.constants import RseAttr
 from rucio.common.utils import generate_uuid, parse_response
 from rucio.core.distance import add_distance
 from rucio.core.replica import add_replica
-from rucio.core.request import TransferStatsManager, get_request_by_did, list_requests, list_requests_history, queue_requests, set_transfer_limit
+from rucio.core.request import TransferStatsManager, get_request_by_did, get_request_history_by_did, list_requests, list_requests_history, queue_requests, requeue_and_archive, set_transfer_limit, transition_request_state
 from rucio.core.rse import add_rse_attribute
 from rucio.db.sqla import constants, models
 from rucio.db.sqla.constants import RequestState, RequestType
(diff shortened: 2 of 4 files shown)


lib/rucio/core/request.py
```diff
@@ -1215,6 +1215,8 @@ def get_request_history_by_did(
                 models.RequestHistory.request_type == request_type
             )
 
+        stmt = stmt.order_by(models.RequestHistory.created_at.desc()).limit(1)
```
VladGaluska: In some cases, requests might fail within microseconds, and Oracle and MySQL and maybe others use seconds for `DATE` and `DATETIME` respectively. While very low odds for this to happen, order for requests microseconds away could differ. This should use `retry_count` as tiebreakers. I know this was my suggested fix, but I just randomly thought about this, sorry!

tests/test_request.py
```diff
@@ -125,6 +125,40 @@ def test_queue_requests_state(vo, file_config_mock, rse_factory, mock_scope, roo
     assert request['state'] == target_state
 
 
+def test_get_request_history_by_did_returns_latest(rse_factory, mock_scope, db_session):
```
VladGaluska: This test inserts the older row first and then the newer one, so this test would've passed before as well. That's fine, but please also add tests for:
- when newer RequestHistory is inserted first
- same created_at, uses retry_count as tiebreaker
Also, this flow uses handmade created_at. It would be nice to have some test like the one I added into the issue:
```python
def test_get_request_history_by_did_returns_latest_real_workflow(rse_factory, mock_scope, root_account, db_session):
    _, source_rse_id = rse_factory.make_mock_rse(session=db_session)
    _, dest_rse_id = rse_factory.make_mock_rse(session=db_session)
    name = generate_uuid()
    add_replica(source_rse_id, mock_scope, name, 1, root_account, session=db_session)

    queue_requests([{
        'dest_rse_id': dest_rse_id,
        'src_rse_id': source_rse_id,
        'request_type': RequestType.TRANSFER,
        'request_id': generate_uuid(),
        'name': name,
        'scope': mock_scope,
        'rule_id': generate_uuid(),
        'retry_count': 0,
        'attributes': {
            'activity': 'User Subscription',
            'bytes': 1,
            'md5': '',
            'adler32': ''
        }
    }], session=db_session)

    # 4 failed attempts
    for _ in range(4):
        req = get_request_by_did(mock_scope, name, dest_rse_id, session=db_session)
        transition_request_state(req['id'], state=RequestState.FAILED, session=db_session)
        requeue_and_archive({'request_id': req['id']}, source_ranking_update=False, session=db_session)

    request = get_request_history_by_did(mock_scope, name, dest_rse_id, session=db_session)
    assert request['retry_count'] == 3, (
        f"expected the latest attempt (retry_count=3), got retry_count={request['retry_count']} "
        f"(created_at={request['created_at']})"
    )
```
where it would sort of behave like the real scenario. This test would also catch the retry_count thing I described above on MySQL/Oracle probably.

VladGaluska (commented):
Nice work, a small thing about testing. The rucio supported matrix is 3.9 - 3.12 so in the future, your tests might work locally, but not in the pipeline.

tests/test_request.py
divyanshu-iitian: Addressed in 571820e. I added coverage for both insertion orders, an equal-created_at retry_count tiebreaker, and the four-failure queue/requeue/archive workflow. The complete tests/test_request.py suite passes on Linux with Python 3.12 (11 passed).

lib/rucio/core/request.py
divyanshu-iitian: Addressed in 571820e. The query now orders by created_at DESC and retry_count DESC, so databases with coarse timestamp precision still return the newest retry.

VladGaluska:
Can you please squash the second commit into the first one? Other than that, looks good!

divyanshu-iitian:
Done — I squashed the follow-up into the original commit and force-pushed the branch safely. The PR now contains a single commit (979ac41) with the same tested tree. Thanks for the review!

VladGaluska (approved):

PR #8735, opened 2026-08-06
feat(Clients): add --state filter to rucio rule list

### LINKED ISSUE #8120: Allow filtering rules by state in rucio rule list
### Description

I.e. to find STUCK rules, could look like this:

```
rucio rule list --account root --state STUCK
```

### Motivation

Debugging transfers

### Change

Add the option to the CLI

ASHISHTOMER0817:

## Description
Add a --state option to rucio rule list so rules can be filtered by state when listing by account, e.g.:

`rucio rule list --account root --state STUCK`
When --account and --state are both provided, the CLI uses the existing list_replication_rules client API with account and state filters. Without --state, behavior is unchanged (list_account_rules).

Closes #8120.

## Changes
**CLI (lib/rucio/cli/rule.py)**

Added --state to rucio rule list (documented values: STUCK, OK, REPLICATING).
When --account and --state are set, calls list_replication_rules({"account": account, "state": state.upper()}).
The CLI passes human-readable state names (e.g. STUCK), consistent with rucio rule update --stuck and the help text. No single-letter codes are exposed to users.

**Core (lib/rucio/core/rule.py)**

Fixed state filtering in `list_rules()` so string filters resolve by enum name first (`RuleState[value]`), then fall back to enum value (`RuleState(value)`). Invalid strings raise `RucioException`.

Why this core change was needed: `RuleState` stores single-character DB values (`STUCK = 'S'`, `OK = 'O'`, `REPLICATING = 'R'`). The old code used `RuleState(value)`, which only accepts those one-character values. Callers (CLI, REST `GET /rules/?state=STUCK`) pass full names, which caused filtering to fail. Lookup by enum name matches how `update_rule` already handles state and what users expect from the CLI.

The fallback keeps existing callers that already filter by the DB character (`S`, `O`, `R`) working.

The client layer is unchanged beyond using the existing `list_replication_rules` API; the fix belongs in core so REST and CLI callers both work with full state names.

**Tests (tests/test_cli_client_structure.py)**

Extended test_rule to cover --state STUCK and --state OK when listing by account.
Why a separate rule is used in the test: The test needs a rule in STUCK state to verify the filter, but test_rule continues to use the original rule_id for later steps (move, update, lock, lifetime, suspend, etc.). Marking that main rule as STUCK would interfere with those steps. A dedicated rule is created and marked STUCK only for the state-filter assertions, so the rest of the test stays unchanged.
## Checklist

- [x] This PR closes #8120 
--- Integration tests in `test_cli_client_structure.py`: a rule is set to STUCK via rucio rule update --stuck, then --state STUCK includes it and --state OK excludes it.
- [ ] Documentation is updated (link the docs PR here), or no documentation change is needed
- [ ] Database migrations are included, or the change touches no database schema

<details>

```
- **Quality**: Agree 🟢
- **Testing**: Agree 🟢
- **Documentation**: Agree 🟢

# Notes for merger



```
</details>

### DIFF
--- a/lib/rucio/cli/rule.py
+++ b/lib/rucio/cli/rule.py
@@ -373,10 +373,20 @@ def update(
 @click.option("--csv", is_flag=True, default=False, help="Comma Separated Value output")
 @click.option("--file", help="Filter by file")
 @click.option("--account", help="Filter by account")
+@click.option("--state", type=click.Choice(["STUCK", "OK", "REPLICATING", "SUSPENDED", "WAITING_APPROVAL", "INJECT"]), help="Filter by rule state (only with --account)")
 @click.option("--subscription", help="Filter by subscription name")
 @click.pass_context
-def list_(ctx: click.Context, did: Optional[str], traverse: bool, csv: bool, file: Optional[str], account: Optional[str], subscription: Optional[str]) -> None:
-    """List all rules impacting a given DID"""
+def list_(
+    ctx: click.Context,
+    did: Optional[str],
+    traverse: bool,
+    csv: bool,
+    file: Optional[str],
+    account: Optional[str],
+    state: Optional[Literal["STUCK", "OK", "REPLICATING", "SUSPENDED", "WAITING_APPROVAL", "INJECT"]],
+    subscription: Optional[str],
+) -> None:
+    """List rules by DID, file, account or subscription."""
     # Done here to raise error==2
     if not (did or file or account or subscription):
         raise InputValidationError("At least one option has to be given. Use -h to list the options.")
@@ -416,7 +426,10 @@ def list_(ctx: click.Context, did: Optional[str], traverse: bool, csv: bool, fil
                 if rules:
                     print('No rules found, listing rules for parents')
     elif account:
-        rules = ctx.obj.client.list_account_rules(account=account)
+        if state:
+            rules = ctx.obj.client.list_replication_rules({"account": account, "state": state})
+        else:
+            rules = ctx.obj.client.list_account_rules(account=account)
     elif subscription:
         if account is None:
             account = ctx.obj.client.account
--- a/lib/rucio/core/rule.py
+++ b/lib/rucio/core/rule.py
@@ -972,7 +972,7 @@ def list_rules(
 
     :param filters: dictionary of attributes by which the results should be filtered.
     :param session: The database session in use.
-    :raises:        RucioException
+    :raises:        InputValidationError, RucioException
     """
 
     stmt = select(
@@ -1007,12 +1007,13 @@ def list_rules(
                 continue
             elif key == 'state':
                 if isinstance(value, str):
-                    value = RuleState(value)
-                else:
                     try:
                         value = RuleState[value]
-                    except ValueError:
-                        pass
+                    except KeyError:
+                        try:
+                            value = RuleState(value)
+                        except ValueError as exc:
+                            raise InputValidationError('Invalid rule state: %s' % value) from exc
             elif key == 'did_type' and isinstance(value, str):
                 value = DIDType(value)
             elif key == 'grouping' and isinstance(value, str):
--- a/tests/test_cli_client_structure.py
+++ b/tests/test_cli_client_structure.py
@@ -880,6 +880,89 @@ def test_rule(rucio_client, mock_scope, file_config_mock):
     assert exitcode == 0
     assert rule_id in out
 
+    stuck_name = generate_uuid()
+    rucio_client.add_replica(rse=mock_rse, scope=scope, name=stuck_name, bytes_=4, adler32="deadbeef")
+    ok_name = generate_uuid()
+    rucio_client.add_replica(rse=mock_rse, scope=scope, name=ok_name, bytes_=4, adler32="deadbeef")
+    replicating_name = generate_uuid()
+    rucio_client.add_replica(rse=mock_rse, scope=scope, name=replicating_name, bytes_=4, adler32="deadbeef")
+    suspended_name = generate_uuid()
+    rucio_client.add_replica(rse=mock_rse, scope=scope, name=suspended_name, bytes_=4, adler32="deadbeef")
+    waiting_approval_name = generate_uuid()
+    rucio_client.add_replica(rse=mock_rse, scope=scope, name=waiting_approval_name, bytes_=4, adler32="deadbeef")
+    inject_name = generate_uuid()
+    rucio_client.add_replica(rse=mock_rse, scope=scope, name=inject_name, bytes_=4, adler32="deadbeef")
+
+    rule_ids = rucio_client.add_replication_rule(
+        dids=[
+            {"scope": scope, "name": stuck_name},
+            {"scope": scope, "name": replicating_name},
+            {"scope": scope, "name": suspended_name},
+        ],
+        copies=1,
+        rse_expression=rule_rse,
+    )
+    waiting_approval_rule_id = rucio_client.add_replication_rule(
+        dids=[
+            {"scope": scope, "name": waiting_approval_name},
+        ],
+        copies=1,
+        rse_expression=rule_rse,
+        ask_approval=True
+    )[0]
+    inject_rule_id = rucio_client.add_replication_rule(
+        dids=[
+            {"scope": scope, "name": inject_name},
+        ],
+        copies=1,
+        rse_expression=rule_rse,
+        asynchronous=True
+    )[0]
+
+    ok_rule_id = rucio_client.add_replication_rule(
+        dids=[{"scope": scope, "name": ok_name}],
+        copies=1,
+        rse_expression=mock_rse
+    )[0]
+
+    stuck_rule_id = rule_ids[0]
+    replicating_rule_id = rule_ids[1]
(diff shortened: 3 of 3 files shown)


ASHISHTOMER0817:
@voetberg Could you review the CLI and test parts when you have time?
@bari12 Could you take a look at the core list_rules() change? Thanks!

lib/rucio/cli/rule.py
```diff
@@ -373,9 +373,10 @@ def update(
 @click.option("--csv", is_flag=True, default=False, help="Comma Separated Value output")
 @click.option("--file", help="Filter by file")
 @click.option("--account", help="Filter by account")
+@click.option("--state", help="Filter by rule state (STUCK, OK, REPLICATING)")
```
voetberg: You can use the 'choice' parameter type to not allow any invalid states - https://click.palletsprojects.com/en/stable/parameter-types/#id3

lib/rucio/cli/rule.py
```diff
@@ -416,7 +417,10 @@ def list_(ctx: click.Context, did: Optional[str], traverse: bool, csv: bool, fil
                 if rules:
                     print('No rules found, listing rules for parents')
     elif account:
-        rules = ctx.obj.client.list_account_rules(account=account)
+        if state:
```
voetberg: If state can only be used when account is given, that should be included in the docstring

tests/test_cli_client_structure.py
```diff
@@ -880,6 +880,32 @@ def test_rule(rucio_client, mock_scope, file_config_mock):
     assert exitcode == 0
     assert rule_id in out
 
+    stuck_name = generate_uuid()
+    rucio_client.add_replica(rse=mock_rse, scope=scope, name=stuck_name, bytes_=4, adler32="deadbeef")
+
+    cmd = f"rucio rule add {scope}:{stuck_name} --copies 1 --rses {rule_rse}"
```
voetberg: `rule add` is already tested here, it should be a faster test to use `rucio_client` to add the rule

tests/test_cli_client_structure.py
```diff
@@ -880,6 +880,32 @@ def test_rule(rucio_client, mock_scope, file_config_mock):
...
+    assert exitcode == 0
+    assert "ERROR" not in err
+
+    cmd = f"rucio rule list --account {rucio_client.account} --state STUCK"
+    exitcode, out, err = execute(cmd)
+    assert exitcode == 0
+    assert "ERROR" not in err
+    assert stuck_rule_id in out
+
+    cmd = f"rucio rule list --account {rucio_client.account} --state OK"
```
voetberg: If you want to test multiple states, adding 3 rules and making them each of the different allowed states is preferable. It is also a good idea to test the response for not accepted state, that can cause some confusing errors if not caught correctly.

lib/rucio/cli/rule.py
ASHISHTOMER0817: Done — --state is now click.Choice(["STUCK", "OK", "REPLICATING"]).

lib/rucio/cli/rule.py
ASHISHTOMER0817: Done — documented in lib/rucio/cli/rule.py: --state help on line 376, command docstring on line 380.

tests/test_cli_client_structure.py
ASHISHTOMER0817: Done — setup now uses rucio_client.add_replication_rule / update_replication_rule instead of rucio rule add. See tests/test_cli_client_structure.py lines 890–906.

tests/test_cli_client_structure.py
ASHISHTOMER0817: Done — the test now creates three rules (STUCK / OK / REPLICATING), checks each --state includes the matching id and excludes the others, and --state FOO fails with Click’s “not one of” message. See tests/test_cli_client_structure.py lines 890–940.

ASHISHTOMER0817:
@voetberg Sorry this took a while. I got stuck on the tests and CI for sometime. Your comments should all be covered in the latest push. Thanks for the review.

bziemons:
@ASHISHTOMER0817 if you addressed the changes for a review, please request a re-review from the reviewer :)

lib/rucio/cli/rule.py
```diff
@@ -373,10 +373,11 @@ def update(
 @click.option("--csv", is_flag=True, default=False, help="Comma Separated Value output")
 @click.option("--file", help="Filter by file")
 @click.option("--account", help="Filter by account")
+@click.option("--state", type=click.Choice(["STUCK", "OK", "REPLICATING"]), help="Filter by rule state (only with --account)")
 @click.option("--subscription", help="Filter by subscription name")
 @click.pass_context
-def list_(ctx: click.Context, did: Optional[str], traverse: bool, csv: bool, file: Optional[str], account: Optional[str], subscription: Optional[str]) -> None:
-    """List all rules impacting a given DID"""
+def list_(ctx: click.Context, did: Optional[str], traverse: bool, csv: bool, file: Optional[str], account: Optional[str], state: Optional[str], subscription: Optional[str]) -> None:
```
voetberg:
```suggestion
def list_(ctx: click.Context, did: Optional[str], traverse: bool, csv: bool, file: Optional[str], account: Optional[str], state: Optional[Literal["STUCK", "OK", "REPLICATING]], subscription: Optional[str]) -> None:
```

lib/rucio/cli/rule.py
```diff
@@ -373,10 +373,11 @@ def update(
...
+    """List rules by DID, file, account, or subscription. The state filter can only be used with account."""
```
voetberg: "The state filter can only be used with account." ends up being replicated information with the argument docstrings, I wouldn't double up on that

ASHISHTOMER0817: Removed that sentence from the command docstring, since it is already in the --state help text.

lib/rucio/cli/rule.py
ASHISHTOMER0817: Done, state is now Optional[Literal["STUCK", "OK", "REPLICATING"]], so it only accepts those strings or None.

voetberg (dismissed):
- **Quality**: Agree 🟢
- **Security**: Agree 🟢
- **Testing**: Agree 🟢
- **Documentation**: Agree 🟢

lib/rucio/cli/rule.py
```diff
@@ -373,10 +373,11 @@ def update(
 @click.option("--csv", is_flag=True, default=False, help="Comma Separated Value output")
 @click.option("--file", help="Filter by file")
 @click.option("--account", help="Filter by account")
+@click.option("--state", type=click.Choice(["STUCK", "OK", "REPLICATING"]), help="Filter by rule state (only with --account)")
 @click.option("--subscription", help="Filter by subscription name")
 @click.pass_context
-def list_(ctx: click.Context, did: Optional[str], traverse: bool, csv: bool, file: Optional[str], account: Optional[str], subscription: Optional[str]) -> None:
-    """List all rules impacting a given DID"""
+def list_(ctx: click.Context, did: Optional[str], traverse: bool, csv: bool, file: Optional[str], account: Optional[str], state: Optional[Literal["STUCK", "OK", "REPLICATING"]], subscription: Optional[str]) -> None:
```
bari12: Why only these states and not also SUSPENDED, WAITING_APPROVAL, and INJECT?

lib/rucio/core/rule.py
```diff
@@ -1007,12 +1007,13 @@ def list_rules(
...
-                else:
                     try:
                         value = RuleState[value]
-                    except ValueError:
-                        pass
+                    except KeyError:
+                        try:
+                            value = RuleState(value)
+                        except ValueError as exc:
+                            raise RucioException(f'Invalid rule state: {value}') from exc
```
bari12: You shouldn't raise a general rucio exception here. This will issue a HTTP 500.
Better InputValidationError

bari12 (changes requested):

lib/rucio/cli/rule.py
ASHISHTOMER0817: I initially scoped this to the common transfer states (OK / REPLICATING / STUCK) from the issue example. Expanding --state to all RuleState values makes sense — I’ll implement that and add the required tests.

lib/rucio/core/rule.py
ASHISHTOMER0817: Agreed — RucioException would surface as an HTTP 500 here. I’ll switch this to InputValidationError in the next push.

voetberg (dismissed):
- **Quality**: Agree 🟢
- **Security**: Agree 🟢
- **Testing**: Agree 🟢
- **Documentation**: Agree 🟢

bari12 (changes requested):
I think this is almost GTG. But the `InputValidationError` you added on my request, it still needs to be caught in the REST layer...

ASHISHTOMER0817:
@bari12 Handling `InputValidationError` is now handled in REST layer. I found that we catch `RuleNotFound` on `AllRule.get`, but list_rules never raises it (empty filters just return no rows). What it does raise are `InputValidationError` (invalid state) and `RucioException` (malformed filter/IDs).

PR #8749, opened 2026-08-14
add is_opendata field to get_metadata

### LINKED ISSUE #7901: Fetch `is_opendata` via existing endpoints

lgiommi:
Issue: #7901

## Description

This PR resumes the closed PR #7903 

## Checklist

- [x] This PR closes #7901 
      
## Notes for contributors

  should remain open.
  will be assigned.
  closed promptly.

Treat `is_opendata` as a reserved, read-only metadata key. Its value is derived from the OpenData catalog and cannot be overridden through custom metadata.

<details>

```
- **Quality**: Agree 🟢
- **Testing**: Agree 🟢
- **Documentation**: Agree 🟢

# Notes for merger



```
</details>

### DIFF
--- a/lib/rucio/core/did.py
+++ b/lib/rucio/core/did.py
@@ -2195,15 +2195,39 @@ def get_metadata_bulk(
                 ).where(
                     or_(*chunk)
                 )
-                for row in session.execute(stmt).scalars():
-                    if plugin.casefold() == 'did_column':
-                        yield row.to_dict()
-                    else:
+                if plugin.casefold() == 'did_column':
+                    opendata_subquery = select(
+                        models.OpenDataDid.scope,
+                        models.OpenDataDid.name
+                    ).subquery()
+
+                    stmt = stmt.add_columns(
+                        case(
+                            (opendata_subquery.c.scope.isnot(null()), true()),
+                            else_=false()
+                        ).label("is_opendata")
+                    ).outerjoin(
+                        opendata_subquery,
+                        and_(
+                            models.DataIdentifier.scope == opendata_subquery.c.scope,
+                            models.DataIdentifier.name == opendata_subquery.c.name
+                        )
+                    )
+
+                    for row, is_opendata in session.execute(stmt):
+                        result = row.to_dict()
+                        result["is_opendata"] = is_opendata
+                        yield result
+
+                else:
+                    for row in session.execute(stmt).scalars():
                         meta = get_metadata(row.scope, row.name, plugin=plugin, session=session)
+
                         result = {'scope': row.scope, 'name': row.name}
                         for key, value in meta.items():
                             if key not in result:
                                 result[key] = value
+
                         yield result
         except NoResultFound:
             raise exception.DataIdentifierNotFound('No Data Identifiers found')
--- a/lib/rucio/core/did.py
+++ b/lib/rucio/core/did.py
@@ -31,6 +31,7 @@ from rucio.common.config import config_get_bool, config_get_int
 from rucio.common.constants import DEFAULT_VO
 from rucio.common.utils import chunks, is_archive
 from rucio.core import did_meta_plugins
+from rucio.core.did_meta_plugins.opendata_meta import add_is_opendata_column
 from rucio.core.message import add_message, add_messages
 from rucio.core.monitor import MetricManager
 from rucio.core.naming_convention import validate_name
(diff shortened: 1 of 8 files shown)


bziemons:
Please make sure all items on the checklist are okay and checked :) Thanks

Geogouz:
There are many edge cases that must be considered here. We can discuss this on Mattermost.

lgiommi:
I’d like to summarize the reasoning behind the current implementation, since the review rounds following the findings raised by the xReview tool exposed a compatibility/design tradeoff around `is_opendata`.

The goal of this PR is to expose `is_opendata` as a derived `DID_COLUMN` metadata field, based on membership in `dids_opendata`. The bulk `DID_COLUMN` path was updated as well, and the OpenData join was factored into a shared helper.

The main complication is that `is_opendata` was not previously a reserved metadata key, so existing deployments may already have custom JSON metadata with the same name.

An earlier implementation made `DidColumnMeta` directly manage `is_opendata`, but that changed metadata routing and introduced backward-compatibility issues in filtering and deletion of existing custom metadata. For this reason, the current implementation does not make `DidColumnMeta.manages_key()` claim `is_opendata`.

The latest review findings, however, showed that allowing new custom metadata values under the same name creates an inconsistent state: the derived `DID_COLUMN.is_opendata` value can disagree with a writable JSON `is_opendata` value, and the same conflict can be introduced through several paths such as `set_metadata`, bulk metadata updates, DID creation, replica creation, attachments, and DIRAC registration.

The current implementation treats `is_opendata` as a read-only derived metadata key at the central metadata dispatcher level. New writes to `is_opendata` are rejected with `InvalidMetadata`, while the routing behaviour of `DidColumnMeta` itself remains unchanged. This avoids reintroducing the previous filtering and deletion issues.

For `plugin="ALL"`, the derived `DID_COLUMN.is_opendata` value is now included and takes precedence over values returned by other metadata plugins. This ensures that `ALL` exposes the canonical OpenData membership status, as required by the latest review findings.

Existing legacy custom metadata named `is_opendata` is not removed automatically and can still exist in the custom metadata plugin, but it no longer overrides the derived value in `ALL`, and new conflicting values cannot be written.

This does introduce a compatibility change for applications which previously used `is_opendata` as an arbitrary writable metadata key. However, the latest review findings indicate that the intended contract is for `is_opendata` to represent an authoritative derived status rather than user-controlled metadata.

I’d appreciate your feedback on whether this is the intended long-term contract, especially regarding existing deployments which may already contain custom metadata named `is_opendata`.

lgiommi:
The latest xReview round is now flagging the backward-compatibility consequences of the current read-only/authoritative approach, which confirms the tradeoff described above.

It is also worth noting that `is_opendata` was not previously a defined or documented Rucio metadata field. While the generic custom metadata plugin technically allowed arbitrary keys with that name, I am not aware of any existing Rucio semantics or documented use of `is_opendata` before this feature.

Before changing the implementation again, I’d prefer to wait a suggestion from you on the intended contract for `is_opendata`, so I can avoid oscillating between fixes that solve one side while reintroducing findings on the other.

Geogouz:
Can we make (maybe) the assumption that nobody currently uses `is_opendata` as custom metadata to simplify things?

lgiommi:
> Can we make (maybe) the assumption that nobody currently uses `is_opendata` as custom metadata to simplify things?

For what I know the answer is yes

lgiommi:
Regarding the new P1 finding about partial commits in set_dids_metadata_bulk(): assuming that is_opendata has not been used previously as custom metadata, I think we can simplify the behaviour and always reject it when it appears in a write payload.
To preserve the atomicity of the bulk operation, I would then pre-validate all DIDs before performing any write, so that a later invalid is_opendata entry cannot leave earlier metadata changes committed by an external plugin.
Does this approach sound reasonable?

PR #8758, opened 2026-08-18
feat(Transfers): List request history by did

### LINKED ISSUE #8051: Allow to retrieve N last requests per DID from REQUESTS_HISTORY
### Description

In CMS we have operational use cases where we might benefit from editing the current get_request_history_by_did function to allow querying more than the last entry. Probably passing an extra parameter like n_latest ?. We’d like in fact to get the  N FTS links corresponding to failed transfer attempts. We know from this PR: https://github.com/rucio/rucio/commit/bf8d33a9b8cc8acb2151571fb1dff59ee7da3399, that there exist REST calls to retrieve the latest link by DID or a list by src_rse and dest_rse.
In our case the [REQUESTS_HISTORY](https://github.com/rucio/rucio/blob/master/etc/sql/oracle/schema.sql#L988) table is indexed by (scope, name, dest_rse) and partitioned by created_at, so the query can efficiently integrate these as arguments. Additionally, an offset parameter could be supported to allow consecutive calls when more than N records are needed.

### Motivation

- In CMS, we often need more than just the latest failed attempt to debug transfer issues: [#rucio/webui/issues/553](https://github.com/rucio/webui/issues/553).

- Having access to multiple historical FTS links is useful for investigating recurring failures and understanding transfer patterns.

- While we are aware of potential performance concerns, this feature would be valuable for operational debugging and monitoring.

- It could also be integrated into the WebUI 2.0 to provide better visibility into transfer histories.

### Change

- Update the current get_request_history_by_did with inputs:
scope
name
dest_rse 
limit (number of FTS links to return, default N)
offset (optional)

- The method should return multiple FTS links for failed attempts, not just the latest one.

VladGaluska:

## Description
Added `list_requests_history_by_did` via core, gateway, REST and the python client. The new endpoint returns a stream with the content type `application/x-json-stream`. It returns newest-first, with optional query parameters and a default `limit` of 10. As said in the issue, it's probably gonna be pretty slow on some instances, so this is intended for debugging and it's a known limitation.

I also added comprehensive tests for the database operations, client and rest gateway.

There are a few things I would like to note:
- There is a sort of breaking change I wasn't sure if I should mark as a breaking change or not. I kept clear of the existing endpoints as much as possible to preserve existing behavior, but the new path is gonna be called `/requests/history/list/<scope>/<name>/<rse>` which means that the old endpoint will not allow a `scope` called 'list'. I assumed this was fine because it also happens for `history`, but please let me know if you want to use a different name here.
- While doing this, I noticed that already are some list endpoints for this functionality (`/requests/list` and `/requests/history/list`. These are the live requests and history tables. I assume these 2 are complementary for some monitoring, but they could be replaced with this new endpoint with the query parameters. That would mean making the scope optional + the sort order configurable. I think this is worth it, but it's a big change, deprecating the old requests and making this new one the intended endpoint for this, so it's not really my choice to make. **If you do think this is a good idea, I can make the filters optional and create a follow-up issue.**
- For the `request_states` I kept it consistent with the other requests, but I find it a bit odd to use `F` or `S` for that stuff, especially since the response has the actual state name, e.g. `FAILED`. Let me know if you want me to change that.
- I also did a few refactorings, where code was pretty much duplicated, but I would like to pursue an issue where the project standardizes that. I'll probably open a discussion on doing this, as there's currently a lot of duplicated code everywhere that do similar things.

Also tested via curl:
image

Closes: #8051

## AI usage disclosure
Claude Code assisted with research, writing code and drafting. I reproduced, verified, reviewed, understand and stand by all changes (as described in the Rucio AI Policy).

## Checklist

- [x] This PR closes #8051
      
## Notes for contributors

  should remain open.
  will be assigned.
  closed promptly.

<details>

```
- **Quality**: Agree 🟢
- **Testing**: Agree 🟢
- **Documentation**: Agree 🟢

# Notes for merger



```
</details>

### DIFF
--- a/lib/rucio/client/requestclient.py
+++ b/lib/rucio/client/requestclient.py
@@ -24,11 +24,19 @@ from rucio.common.utils import build_url
 
 if TYPE_CHECKING:
     from collections.abc import Iterator, Sequence
+    from datetime import datetime
 
 
 class RequestClient(BaseClient):
 
     REQUEST_BASEURL = 'requests'
+
+    def _send_get_request(self, url: str, params: Optional[dict[str, Any]] = None) -> 'Iterator[dict[str, Any]]':
+        r = self._send_request(url, method=HTTPMethod.GET, params=params)
+        if r.status_code == codes.ok:
+            return self._load_json_data(r)
+        exc_cls, exc_msg = self._get_exception(headers=r.headers, status_code=r.status_code, data=r.content)
+        raise exc_cls(exc_msg)
 
     def list_requests(
             self,
@@ -45,13 +53,7 @@ class RequestClient(BaseClient):
         path = '/'.join([self.REQUEST_BASEURL, 'list']) + '?' + '&'.join(['src_rse={}'.format(src_rse), 'dst_rse={}'.format(
             dst_rse), 'request_states={}'.format(request_states)])
         url = build_url(choice(self.list_hosts), path=path)
-        r = self._send_request(url, method=HTTPMethod.GET)
-
-        if r.status_code == codes.ok:
-            return self._load_json_data(r)
-        else:
-            exc_cls, exc_msg = self._get_exception(headers=r.headers, status_code=r.status_code, data=r.content)
-            raise exc_cls(exc_msg)
+        return self._send_get_request(url)
 
     def list_requests_history(
             self,
@@ -70,20 +72,14 @@ class RequestClient(BaseClient):
         path = '/'.join([self.REQUEST_BASEURL, 'history', 'list']) + '?' + '&'.join(['src_rse={}'.format(src_rse), 'dst_rse={}'.format(
             dst_rse), 'request_states={}'.format(request_states), 'offset={}'.format(offset), 'limit={}'.format(limit)])
         url = build_url(choice(self.list_hosts), path=path)
-        r = self._send_request(url, method=HTTPMethod.GET)
-
-        if r.status_code == codes.ok:
-            return self._load_json_data(r)
-        else:
-            exc_cls, exc_msg = self._get_exception(headers=r.headers, status_code=r.status_code, data=r.content)
-            raise exc_cls(exc_msg)
+        return self._send_get_request(url)
 
     def list_request_by_did(
             self,
             name: str,
             rse: str,
             scope: Optional[str] = None
-    ) -> 'Iterator[dict[str, Any]]':
+    ) -> dict[str, Any]:
         """Return latest request details for a DID
         Parameters
         ----------
@@ -106,20 +102,14 @@ class RequestClient(BaseClient):
         if scope is not None:
             path = '/'.join([self.REQUEST_BASEURL, quote_plus(scope), quote_plus(name), rse])
         url = build_url(choice(self.list_hosts), path=path)
-        r = self._send_request(url, method=HTTPMethod.GET)
-
-        if r.status_code == codes.ok:
-            return next(self._load_json_data(r))
-        else:
-            exc_cls, exc_msg = self._get_exception(headers=r.headers, status_code=r.status_code, data=r.content)
-            raise exc_cls(exc_msg)
+        return next(self._send_get_request(url))
 
     def list_request_history_by_did(
             self,
             name: str,
             rse: str,
             scope: Optional[str] = None
-    ) -> 'Iterator[dict[str, Any]]':
+    ) -> dict[str, Any]:
         """
         Return latest request details for a DID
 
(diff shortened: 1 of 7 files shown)


lib/rucio/client/requestclient.py
```diff
@@ -144,13 +134,75 @@ def list_request_history_by_did(
...
+        created_before :
+            Only return requests created at or before this time.
+        offset :
+            Number of requests to skip.
+        limit :
+            Maximum number of requests to return. Defaults to 10.
+
+        Raises
+        ------
+        exc_cls
```
voetberg: Please list the specific exceptions, the user doesn't know what `_get_exception` is

lib/rucio/client/requestclient.py
```diff
@@ -144,13 +134,75 @@ def list_request_history_by_did(
...
+            From BaseClient._get_exception.
+
+        Returns
+        -------
+            An iterator over the matching historical requests.
```
voetberg: Please list the exact contents, knowing it's just "requests" isn't helpful for downstream development.

lib/rucio/client/requestclient.py
```diff
@@ -24,12 +24,20 @@
...
 if TYPE_CHECKING:
     from collections.abc import Iterator, Sequence
+    from datetime import datetime
 
 
 class RequestClient(BaseClient):
 
     REQUEST_BASEURL = 'requests'
 
+    def _send_get_request(self, url: str, params: Optional[dict[str, Any]] = None) -> 'Iterator[dict[str, Any]]':
```
voetberg: I would argue this refactor change is outside the scope of this PR, and that if we want to do this, it would make more sense to make it its own PR.

voetberg: (Mostly because I will be permanently annoyed this is only in `requestclient` and none of the other clients.

lib/rucio/client/requestclient.py
```diff
@@ -70,20 +72,14 @@ def list_requests_history(
...
-            raise exc_cls(exc_msg)
+        return self._send_get_request(url)
 
     def list_request_by_did(
             self,
             name: str,
             rse: str,
             scope: Optional[str] = None
-    ) -> 'Iterator[dict[str, Any]]':
+    ) -> dict[str, Any]:
```
voetberg: Pedantic - this should be a separate refactor commit

lib/rucio/client/requestclient.py
```diff
@@ -106,20 +102,14 @@ def list_request_by_did(
...
-            raise exc_cls(exc_msg)
+        return next(self._send_get_request(url))
 
     def list_request_history_by_did(
             self,
             name: str,
             rse: str,
             scope: Optional[str] = None
-    ) -> 'Iterator[dict[str, Any]]':
+    ) -> dict[str, Any]:
```
voetberg: Also should be in that separate refactor commit

lib/rucio/core/request.py
```diff
@@ -3064,6 +3064,83 @@ def list_requests_history(
...
+def list_requests_history_by_did(
+    scope: InternalScope,
+    name: str,
+    rse_id: str,
+    rule_id: Optional[str] = None,
+    states: Optional['Sequence[RequestState]'] = None,
+    created_after: Optional[datetime.datetime] = None,
+    created_before: Optional[datetime.datetime] = None,
+    offset: Optional[int] = None,
+    limit: int = 10,
```
voetberg: Slightly worried about having defaults set for limit on the client falling out of sync with the core. I would remove the default from here and just make it required in the web API.

lib/rucio/web/rest/flaskapi/v1/requests.py
```diff
@@ -30,6 +32,52 @@
 if TYPE_CHECKING:
     from collections.abc import Iterator
 
+# smallest maximum allowed value across the supported databases
+MAX_BIGINT = 2 ** 63 - 1
```
voetberg: Don't love having this constant here. Is there a way to get it from core or some other system var?

lib/rucio/web/rest/flaskapi/v1/requests.py
```diff
@@ -30,6 +32,52 @@
...
+def _parse_request_states_param(request_states: Optional[str]) -> Optional[list[RequestState]]:
+    if not request_states:
+        return None
+    try:
+        return [RequestState(state) for state in request_states.split(',')]
+    except ValueError as error:
+        raise ValueError('Request state value is invalid') from error
+
+
+def _parse_datetime_param(value: Optional[str], name: str) -> Optional[datetime.datetime]:
```
voetberg: Use `common.utils.str_to_date`

lib/rucio/web/rest/flaskapi/v1/requests.py
```diff
@@ -821,6 +869,234 @@ def generate(issuer, vo):
...
+            description: "Not found"
+          406:
+            description: "Not acceptable"
+        """
+        try:
+            scope, name = parse_scope_name(scope_name, flask.request.environ['vo'])
+        except ValueError as error:
+            return generate_http_error_flask(400, error)
+
+        try:
```
voetberg: I have some issues with the design of this. You're doing a ton of checks for value errors and then just raising a value error at end again to just raise a different value error. Can this not be streamlined?

lib/rucio/web/rest/flaskapi/v1/requests.py
```diff
@@ -821,6 +869,234 @@ def generate(issuer, vo):
...
+            created_before = _parse_datetime_param(
+                flask.request.args.get('created_before', default=None), 'created_before'
+            )
+            limit = _validate_paging_param(
+                'limit', _parse_int_param(flask.request.args.get('limit'), 'limit', default=10), minimum=1
+            )
+            offset = _validate_paging_param(
+                'offset', _parse_int_param(flask.request.args.get('offset'), 'offset', default=0), minimum=0
+            )
+            rule_id = _validate_and_normalize_rule_id_param(flask.request.args.get('rule_id', default=None))
```
voetberg: Why this check? We validate the rule id when we check the db anyway, let's not re-invent the wheel here

voetberg (changes requested):

PR #8763, opened 2026-08-21
feat(Policies): Simplify parsing of scope and name from URLs

### LINKED ISSUE #7519: Do not use regex to split did, use scope extraction method

jamesp-epcc:
Closes: #7519

## Description

This PR removes the old complex code for extracting scope and name from URLs in the REST API and instead extracts them by simply splitting at a '/' separator. This depends on the server having the `AllowEncodedSlashes` option set to `NoDecode` to ensure that any other slashes in the name will be encoded. This behaviour has always been an option but now that we have this setting always present in the server container (see https://github.com/rucio/containers/pull/470 ) we intend to make it the only option and remove the old code that uses `SCOPE_NAME_REGEXP`.

Tests and documentation have not been updated as this change is merely a simplification of the code and should not have externally visible effects.

With `NoDecode` set on the server and **without** this change, many BelleII tests fail because BelleII includes a leading slash on the DID names which is not handled correctly in that case. **With** this change, the vast majority of the BelleII tests pass again, though two tests in `test_meta_did.py` fail. These appear to be failing because they use a DID that does not comply with the BelleII schema, so I am not sure how they were ever able to pass when using the BelleII policy package. Possibly some unintended quirk of the old code caused them to pass. I am not a BelleII person so I am not sure how best to fix this, but I am open to suggestions.

I am not sure what is happening with the other (non-BelleII) test failures. I didn't see the ATLAS failure in my own repo so it might be nothing to do with this. And the linter's suggestion looks bizarre to me.

## Checklist

- [X] This PR closes #7519
      
- [X] Tests cover the change, or no tests are needed (explain why in the description)
- [X] Documentation is updated (link the docs PR here), or no documentation change is needed
- [X] Database migrations are included, or the change touches no database schema
- [X] This PR contains no breaking changes, or the breaking change is described

<details>

```
- **Quality**: Agree 🟢
- **Testing**: Agree 🟢
- **Documentation**: Agree 🟢

# Notes for merger



```
</details>

### DIFF
--- a/etc/docker/test/extra/rucio.conf
+++ b/etc/docker/test/extra/rucio.conf
@@ -27,6 +27,6 @@ WSGIProcessGroup rucio
 
  WSGIScriptAlias /  /opt/rucio/lib/rucio/web/rest/main.py
 
- AllowEncodedSlashes on
+ AllowEncodedSlashes NoDecode
 
 </VirtualHost>
--- a/lib/rucio/common/schema/__init__.py
+++ b/lib/rucio/common/schema/__init__.py
@@ -25,8 +25,6 @@ from rucio.common import config, exception
 from rucio.common.constants import DEFAULT_VO
 from rucio.common.plugins import check_policy_module_version
 from rucio.common.schema.schema_ref import SchemaRef
-from rucio.db.sqla.constants import DatabaseOperationType
-from rucio.db.sqla.session import db_session
 
 if TYPE_CHECKING:
     from types import ModuleType
@@ -35,9 +33,6 @@ LOGGER = logging.getLogger('policy')
 
 # dictionary of schema modules for each VO
 schema_modules: dict[str, "ModuleType"] = {}
-
-# list of unique SCOPE_NAME_REGEXP values from all schemas
-scope_name_regexps: list[str] = []
 
 
 # cached function to check for multivo
@@ -101,8 +96,6 @@ if not _is_multivo():
         raise exception.ErrorLoadingPolicyPackage(policy)
 
     schema_modules[DEFAULT_VO] = module
-    if hasattr(module, 'SCOPE_NAME_REGEXP'):
-        scope_name_regexps.append(module.SCOPE_NAME_REGEXP)
 
 
 def load_schema_for_vo(vo: str) -> None:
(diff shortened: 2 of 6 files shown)


jamesp-epcc:
Although this still has some failing tests, I am reopening it to enable discussion on how to address these. I am reluctant to make the change suggested by the linter as to me it just adds needless extra text and I don't see this idiom used elsewhere in the code. As for the BelleII tests I am not a BelleII person so I am unsure what the best solution is here.

bziemons:
> I am reluctant to make the change suggested by the linter as to me it just adds needless extra text and I don't see this idiom used elsewhere in the code.

I agree, the suggested changes neither make sense to me. It would be nice to know which ruff rule leads to this.

About the Belle II tests, please make sure you contact people directly. Otherwise it might take very long until they see this PR and react to it.

Geogouz:
It is the `F401` warning which has a valid cause because `DatabaseOperationType` and `db_session` become now unused. But Ruff’s automatic fix is indeed a bit off because while X as X declares an intentional public re-export, there is no code importing them through `rucio.common.schema`. I think we can simply drop those.

bziemons:
I agree. Since this is in `__init__.py` it is a special case (importing modules outside of the package the init is for; I would try to avoid this). The unused imports should be removed.

jamesp-epcc:
You are right. I was so confused by the lint suggestion that I failed to notice that those imports were actually unused. I've removed them now.

I will flag this up with BelleII and find out what's best to do about their tests.

jamesp-epcc:
Actually a lot more BelleII tests are failing now than before. I need to investigate why this is happening.

jamesp-epcc:
Most of the failures are because the tests are running with a cached image that doesn't include the `AllowEncodedSlashes` change. When I force the image to be rebuilt in my fork, the tests all pass except for the two I mentioned in the description. I will speak to BelleII about how best to handle those.

Issue #8743, opened 2026-08-12
STUCK rule counter and an optional way to switch the rule to SUSPENDED given enough trials


onuray-sancar:
### Description

An optional way to identify rules constantly switching back and forth between STUCK and REPLICATING (never switching to SUSPENDED), a STUCK counter, that later switches the rule to SUSPENDED after a given number of trials

### Motivation

Currently, the rules in rucio switch to SUSPENDED only if certain criteria are met. When a rule that is STUCK goes to the state of REPLICATING, the time recording the initial STUCK state is not saved[1]. So, if a rule goes back to STUCK again, the timer for the STUCK state is renewed. This means that the 14-day timer for a STUCK rule to switch to SUSPENDED may never be reached[2],  and a rule can be in this loop forever. We, at the CMS DM team, currently look at the SUSPENDED rules for handling them, and would like to have a way to handle rules that are forever in the state of STUCK.

[1]https://github.com/rucio/rucio/blob/9d8f9f1f1835e1781b5cf9e8404cd2f91597a8fc/lib/rucio/core/rule.py#L1564

[2]https://github.com/rucio/rucio/blob/9d8f9f1f1835e1781b5cf9e8404cd2f91597a8fc/lib/rucio/core/rule.py#L1289

### Change

We would like to have an optional STUCK counter that switches the rule to SUSPENDED after a certain amount of trials.

bari12:
I think, in the first place, this could be implemented relatively easy.
The only decision we would have to make is how we configure which "policy" is being used for the suspended transition.

juanpablosalas:
I think both should be available, on the one hand if the `stuck_at` field is older than two weeks it changes to suspended (this is the current behavior) and in addition to that if rule that does not fit this category has a particular lock that is being retried more than X (3?) times it should also change its status to suspended. As @onuray-sancar mentioned, this is how we envision it working for CMS but since other VOs may have different policies this second feature can be made optional.

bari12:
Yes, I just wonder if this is just something we should push off into the policy package via a hook.

PR #8785, opened 2026-08-31
feat(Core): Suspend rules stuck too many times #8743

FanxinSun:
Addresses #8743.

What was fixed: `repair_rule` cleared `stuck_at` as soon as the rule was no longer STUCK,
including when it only went back to REPLICATING. Nothing counted the repair attempts either,
and the line still has the old comment `# TODO: Increase some kind of Stuck Counter here`.

Outcome: `stuck_at` is now cleared only when the rule reaches OK. That is done by
`reset_stuck_state`, called from the five places where a rule becomes OK (four in
`core/rule.py`, one in `core/lock.py`), so the 14 day timer covers the whole time the rule was
stuck. A new `rules.stuck_cnt` column counts the repair attempts that left the rule STUCK, and
the rule is suspended once that count reaches `rules/stuck_cnt_to_suspend`. That setting is 0
by default, so nothing changes unless it is set, and the 14 day rule still works as before.
`_stuck_rule_suspend_reason` makes the choice, and the reason it gives is printed in the log.

Before and after, for a rule flipping between STUCK and REPLICATING:

    before:  stuck_at reset on every flip, so "STUCK for 14 days" is never true and the rule
             loops indefinitely
    after:   Replication rule <id> has been SUSPENDED (STUCK since <date>)
    cnt = 3: Replication rule <id> has been SUSPENDED (STUCK after 3 repair attempts)

`tests/test_rule_stuck.py` covers the four cases, as a separate module because `test_rule.py`
builds a `RuleClient()` at import time and needs a running server.

### DIFF
--- a/lib/rucio/core/lock.py
+++ b/lib/rucio/core/lock.py
@@ -394,6 +394,7 @@ def successful_transfer(scope: InternalScope, name: str, rse_id: str, nowait: bo
             pass
         elif rule.locks_replicating_cnt == 0 and rule.state == RuleState.REPLICATING:
             rule.state = RuleState.OK
+            rucio.core.rule.reset_stuck_state(rule)
             # Try to update the DatasetLocks
             if rule.grouping != RuleGrouping.NONE:
                 stmt = select(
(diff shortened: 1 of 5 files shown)


voetberg:
Hi! Please use the PR template and check your PR follows all the requirements listed there. Also be check out the [contributing guidelines](https://rucio.cern.ch/documentation/contributing/) and [AI policy](https://rucio.cern.ch/documentation/developer/dev_ai_policy)

